PR #174, opened 2018-11-08
Created new CommandScheduler for FT

spericas:
Created new CommandScheduler based on the configurable ScheduledThreadPool supplier from Helidon common.


microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandScheduler.java
```diff
@@ -0,0 +1,57 @@
...
+    private CommandScheduler() {
+        executor = ScheduledThreadPoolSupplier.create().get();
+    }
+
+    /**
+     * Returns the single instance of this class.
+     *
+     * @return The instance.
+     */
+    public static CommandScheduler instance() {
```
tomas-langer: Using a singleton that cannot be configured is not the best approach for Helidon.
Also - is this the only place where we use a ScheduledExecutorService? If not, maybe the instance should be shared (or the developer may want to share the instance)?
Please find a better place to hold an instance of CommandScheduler (or refactor to do differently) that would allow us to provide a configured ExecutorService or to load it from config.

spericas: Yes, the configuration part is TBD, I wanted to discuss this. There is no API here for someone to call a builder, so perhaps there should be a well-defined config section from which to read these properties. Logically, it does make sense for this to be a singleton.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandScheduler.java
```diff
@@ -0,0 +1,57 @@
...
+
+import net.jodah.failsafe.util.concurrent.Scheduler;
+
+/**
+ * Class CommandScheduler.
+ */
+public class CommandScheduler implements Scheduler {
+
+    private static CommandScheduler instance;
+
```
tomas-langer: Is the lazy loading really required? If not, use
`private static CommandScheduler instance = new CommandScheduler` and remove synchronized from getInstance()

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandScheduler.java
```diff
@@ -0,0 +1,57 @@
...
+    private ScheduledThreadPoolExecutor executor;
```
tomas-langer: Should be final

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceMpService.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Class FaultToleranceMpService.
+ */
+public class FaultToleranceMpService implements MpService {
+
+    @Override
+    public void configure(MpServiceContext mpServiceContext) {
+        MpConfig config = (MpConfig) ConfigProviderResolver.instance().getConfig();
+        CommandScheduler.create(config.getConfig().get("helidon.fault-tolerance.async-thread-pool"));
```
tomas-langer: Replace the lines with:
`        CommandScheduler.create(mpServiceContext.getHelidonConfig().get("helidon.fault-tolerance.async-thread-pool"));`

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandScheduler.java
tomas-langer: `private final ScheduledThreadPoolExecutor executor;`

tomas-langer:
Two small changes, otherwise LGTM.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandScheduler.java
spericas: That line does not exist in latest PR.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceMpService.java
spericas: Done

PR #177, opened 2018-11-08
Create a directory for specifications

barchetta:
This creates a directory to hold internal documents and specifications. I have intentionally not included any process or templates -- preferring to defer that until if and when it is needed.

### DIFF
--- a/docs-internal/README.md
+++ b/docs-internal/README.md
@@ -1,0 +1,23 @@
+
+# Internal Documents and Specifications
+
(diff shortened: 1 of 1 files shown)


romain-grecourt:
Should we include a template for a specs ? Or should we wait until we do need to write one ?

barchetta:
> Should we include a template for a specs ? Or should we wait until we do need to write one ?
> Or maybe you are doing it in separate PRs.

I'm thinking we wait until we need one.

PR #178, opened 2018-11-12
Development guidelines in common

tomas-langer:
Refactoring of common modules based on development guidelines


common/configurable/src/main/java/io/helidon/common/configurable/Resource.java
```diff
@@ -163,35 +150,12 @@ static Resource from(InputStream inputStream, String description) {
...
-                                })
-                                .orElseGet(() -> Resource.from(uri))))
-                .or(() -> config.get(keyPrefix + "-content-plain")
-                        .asOptionalString()
-                        .map(content -> Resource.fromContent("config:" + keyPrefix + "-content", content)))
-                .or(() -> config.get(keyPrefix + "-content")
-                        .asOptionalString()
-                        .map(Base64.getDecoder()::decode)
-                        .map(content -> Resource.fromContent("config:" + keyPrefix + "-content-b64", content)))
+    static Optional<Resource> create(Config config, String keyPrefix) {
```
romain-grecourt: Why do we have to pass a key prefix ? Why can't the caller pass the right config node ?
This is introducing some level of abstraction that is doing magic.

common/configurable/src/main/java/io/helidon/common/configurable/ResourceUtil.java
```diff
@@ -99,4 +105,47 @@ static InputStream toIs(URI uri, Proxy proxy) {
             throw new ResourceException("Failed to open strem to uri: " + uri, e);
         }
     }
+
+    static Optional<Resource> fromConfigPath(Config config, String keyPrefix) {
```
romain-grecourt: These methods basically define some conventions for config key suffixes and types.
I can see this being used internally to have consistency between all the different configurable entities, but I don't understand why we would expose the keyPrefix in our API.

tomas-langer: We should have a look at this separately. This PR is to align with development guidelines - this library is in GIT for a long time. I agree we may need to refactor this, though not as a part of this PR.

romain-grecourt: Roger that.

PR #180, opened 2018-11-14
Small memory improvement

spericas:
Small memory improvement that discards data associated with a command when it is being reset. If the circuit breaker is used again, the data will be re-created instead.


spericas:
@romain-grecourt Right, since the reset state is identical to the creation state, when a reset comes in, the instance is removed and if needed re-created at a later time.

PR #182, opened 2018-11-15
Add Creating Docker Images guide

barchetta:
This guide has no associated code -- it uses the Quickstart example, so I've put the adoc file directly in  `src/main/docs/guides/`

### DIFF
--- a/docs/src/main/docs/guides/05_Dockerfile.adoc
+++ b/docs/src/main/docs/guides/05_Dockerfile.adoc
@@ -1,0 +1,302 @@
+///////////////////////////////////////////////////////////////////////////////
+
+    Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+///////////////////////////////////////////////////////////////////////////////
+:adoc-dir: {guides-dir}
+
+= Creating Docker Images 
+:description: Helidon guide docker
+:keywords: helidon, guide, docker
+
+Building Docker images for Helidon applications.
+
+=== What You Will Learn
+
+You'll learn how to package your Helidon application and its runtime dependencies into a Docker
+image that also contains the Java runtime.
+
+=== What You Need
+
+|===
+|About 10 minutes
+| <<getting-started/01_prerequisites.adoc,Helidon Prerequisites>>
+|You'll also need Java 11 if you want to create custom JRE's using `jlink`
+|===
+
+=== Create Your Application Using The Helidon Quickstart
+
+Follow the instructions on the <<getting-started/02_base-example.adoc,Quickstart page>>
+to create a Helidon SE example. Once you've run the archetype to create
+the project come back here.
+
+=== Handling Runtime Dependencies
+
+Build the example:
+
+[source,bash,subs="verbatim,attributes"]
+----
+mvn clean package
+----
+
+When you run the maven build, you'll notice lines like the following:
+
+[source,bash,subs="verbatim,attributes"]
+----
+[INFO] Scanning for projects...
+. . .
+[INFO]
+[INFO] --- maven-dependency-plugin:2.9:copy-dependencies (copy-dependencies) @ quickstart-se ---
+[INFO] Copying netty-transport-4.1.22.Final.jar to /tmp/quickstart-se/target/libs/netty-transport-4.1.22.Final.jar
+. . .
+----
+
+The project uses the `maven-dependency-plugin` to copy the runtime dependencies to
+`target/libs/`. Additionally the `maven-jar-plugin` adds a `Class-Path` entry to the
+application's jar file so it can find those dependencies at runtime:
+
+[source,bash,subs="verbatim,attributes"]
+----
+unzip -p target/quickstart-se.jar META-INF/MANIFEST.MF 
+. . .
+Class-Path: libs/helidon-webserver-bundle-0.10.4.jar libs/helidon-webser
+ ver-0.10.4.jar libs/helidon-common-reactive-0.10.4.jar libs/helidon-com
+ mon-http-0.10.4.jar libs/helidon-common-key-util-0.10.4.jar libs/helido
+. . .
+----
+
+This means you can easily run the application jar:
+
+[source,bash,subs="verbatim,attributes"]
+----
+java -jar target/quickstart-se.jar
+----
+
+Now try the applicaion with `curl`:
+
+[source,bash,subs="verbatim,attributes"]
+----
+curl -X GET http://localhost:8080/greet
+{"message":"Hello World!"}
+----
+
+=== Creating a Docker Image
+
+Since the `target` directory has the runtime dependencies, the
+Dockerfile for our application is pretty simple. You can
+find the Dockerfile in the generated project at `src/main/docker/Dockerfile`.
+It should look like this:
+
+[source,yaml,subs="verbatim,attributes"]
+----
+FROM openjdk:8-jre-slim 
+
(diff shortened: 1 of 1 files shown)


docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+Build the example:
+
+[source,bash,subs="verbatim,attributes"]
+----
+mvn clean package
+----
+
+When you run the maven build, you'll notice lines like the following:
+
+[source,bash,subs="verbatim,attributes"]
```
tjquinno: In the recent review meeting of the RESTful web app guide we said we would use something other than `source` to display output to suppress the automatic "copy-to-clipboard" support. I've done a few very quick experiments and 
```
[listing]
----
example output here
----
```
should work nicely. the `[listing]` block also supports callouts.

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+[INFO] --- maven-dependency-plugin:2.9:copy-dependencies (copy-dependencies) @ quickstart-se ---
+[INFO] Copying netty-transport-4.1.22.Final.jar to /tmp/quickstart-se/target/libs/netty-transport-4.1.22.Final.jar
+. . .
+----
+
+The project uses the `maven-dependency-plugin` to copy the runtime dependencies to
+`target/libs/`. Additionally the `maven-jar-plugin` adds a `Class-Path` entry to the
+application's jar file so it can find those dependencies at runtime:
+
+[source,bash,subs="verbatim,attributes"]
```
tjquinno: How about splitting the command the user enters from the output? Then the copy-to-clipboard would grab just the command, not the command plus the output.

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+=== Creating a Docker Image
+
+Since the `target` directory has the runtime dependencies, the
+Dockerfile for our application is pretty simple. You can
+find the Dockerfile in the generated project at `src/main/docker/Dockerfile`.
+It should look like this:
+
+[source,yaml,subs="verbatim,attributes"]
+----
+FROM openjdk:8-jre-slim
```
tjquinno: Because this is exactly what's in the `Dockerfile` how about using `include::` to pull the text from the actual source into the .adoc file? That `include` directive would have to reference a generated `Dockerfile` that's in the repo somewhere. You might also want to add `# tag::build[]` and `# end::build[]` to the Docker file to bracket the part of the file you want in the .adoc output and then reference the tag in the  `.adoc` `include` directive.

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+FROM openjdk:8-jre-slim 
+
+RUN mkdir /app
+COPY libs /app/libs 
+COPY ${project.artifactId}.jar /app
+
+CMD ["java", "-jar", "/app/${project.artifactId}.jar"]
+----
+
+* We use a Java 8 JRE image provided by the OpenJDK project.
```
tjquinno: I have not tried it with a `Dockerfile`, but you might try using AsciiDoc callouts in the `Dockerfile`? E.g., at the end of the `FROM` line add `# <1>` and at the end of the `RUN mkdir /app` line add `# <2>`. Then after the end of the AsciiDoc code block terminator (`----`) you can use 
```
<1> We use a Java 8 JRE image...
<2> We then create a directory...
```
This does mean editing the `Dockerfile` once it has been generated into the repository.

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+the `maven-resources-plugin` to filter the Dockerfile while copying it to the
+target directory. So this property gets expanded to the project artifactId
+(`quickstart-se` for example).
+
+Since the quickstart project already contains the Dockerfile you can go ahead and try it:
+
+[source,bash,subs="verbatim,attributes"]
+----
+docker build -t quickstart-se target
+
```
tjquinno: Again, it might be good to separate the command we want the user able to copy and the sample output into different blocks.

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+Fat Jars are jar files that contain the application and all runtime
+dependencies. This is handy because it's one file with all you need
+to run your application.
+
+The problem with fat jars is that they are not optimal when used in
+a docker image. That's because the image layer that contains your
+application also contains all of its runtime dependencies, and that
+means more data to push to a docker registry every time you rebuild
+your application.
+
```
tjquinno: Another problem (I think) with fat jars is that the content of any dependent JAR that was signed loses its "signed-ness" when extracted and merged into the fat JAR. This is probably far less important than fat JARs being sub-optimal for Docker. If we don't want to mention that other issue explicitly maybe change `The problem with fat jars` to `A problem...`?

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+[source,yaml,subs="verbatim,attributes"]
+----
+RUN ["jlink", "--compress=2", "--strip-debug", "--no-header-files" \ 
+     "--add-modules", \
+     "java.base,java.logging,java.sql,java.desktop,java.management,java.naming,jdk.unsupported", \
+     "--output", "/var/tmp/myjre"]
+----
+
+=== How Big is This Stuff?
+
```
tjquinno: One last time, about separating the commands from their output into separate blocks.

docs/src/main/docs/guides/05_Dockerfile.adoc
```diff
@@ -0,0 +1,302 @@
...
+.Java 11 Docker Image for Helidon SE Quickstart Example
+[source,bash,subs="verbatim,attributes"]
+----
+docker images java11-quickstart-se:latest
+
+REPOSITORY             TAG                 IMAGE ID            CREATED             SIZE
+java11-quickstart-se   latest              1cd7b46bf454        5 days ago          143MB
+----
+
```
tjquinno: Do we think these sizes are small? Large? Can we add at least a sentence leading readers to the conclusion about the sizes we want them to draw?

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: Nice! I'll give that a try.

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: Yep, I was thinking that too. I'll see how that looks.

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: Since this is the only source file and it's small, I decided to keep it inline. I started with it external (over in examples) but it didn't seem worth the extra work. I'll revisit this, but I might not change it.

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: Unfortunately the # for comments in a Dockerfile must be in column one, so that limits what you can do. I played with this and decided for this simple file I would not use them.

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: Roger

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: OK

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: Yep

docs/src/main/docs/guides/05_Dockerfile.adoc
barchetta: OK, I'll come up with something.

And thanks Tim! Good comments.

cricket007:
I wonder if using [`jib`](https://github.com/GoogleContainerTools/jib) would create smaller images 🤔

PR #187, opened 2018-11-15
Config API changes document

tomas-langer:


tomas-langer:
See branch https://github.com/oracle/helidon/tree/config-api-changes
Now with changes for "Config API has Too Many Methods"
Some of the usages of config are refactored to the new API
I have re-introduces a method similar to "mapList" in current api - 
`<T> ConfigValue<List<T>> asList(Function<Config, T> mapper)`
as this removes at least 4 lines of boiler-plate code.

tomas-langer:
I am now working on unit tests - I need them to pass to see if the new API is usable.
Discovering a few questions on the way.
One of the most interesting ones:
There are three options for Config interface:

1. Config extends ConfigValue<String>
2. Config extends ConfigValue<Config>
3. Config

The first option gives us the possibility to treat String as a privileged citizen - we say that each config node is in fact a String node as well (with an optional value).

The second option gives use the possibility to move some of the methods to ConfigValue only and they will nicely work on Config itself (e.g.  <N> N as(Function<T, N> mapper to map from current type to any other type). This option requires us to treat String value as any other type - e.g. the user must do `config.asString()...`

The third option give us the freedom to name methods on ConfigValue as we want. Also Config itself would only have typed methods to access ConfigValues and would not act as a ConfigValue itself.
We would have a method (e.g. `Optional<String> value()`) to get the possible string value and everything else would have to be done through accessors.

Usage examples:
1: 
```
Supplier<Config> configSup = config.asNode().asSupplier();
Supplier<String> stringSup = config.asSupplier();
Supplier<Integer> intSup = config.asInt().asSupplier();
```

2:
```
Supplier<Config> configSup = config.asSupplier();
Supplier<String> stringSup = config.asString().asSupplier();
Supplier<Integer> intSup = config.asInt().asSupplier();
```
3:
```
Supplier<Config> configSup = config.asNode().asSupplier();
Supplier<String> stringSup = config.asString().asSupplier();
Supplier<Integer> intSup = config.asInt().asSupplier();
```

Opinions?

romain-grecourt:
I do like option 3 because of the explicitness.
If option 1 simplifies the most common use-cases significantly I'd vote for it.

tomas-langer:
I like option 3 also because I can use methods such as "get" in ConfigValue, which we cannot do now (as they clash with Config.get)

tomas-langer:
I have updated #207 - all unit tests in "config" and "config-beans" modules now pass.
The API is modified with a small number of shortcut methods:
ConfigValue asString() - similar for Boolean, Int, Long, Double

There are no shortcut methods to get the actual value - we may introduce them if needed, such as:
String getString() throws MissingValueException;
Long getLong() throws MissingValueException;

These methods can be added as default methods to Config interface at any time (with a simple enough unit test - the ConfigValue is tested on its own, so the unit tests would only need to validate we call the correct methods on ConfigValue).

PR #200, opened 2018-11-16
Make FilesystemWatchPollingStrategy.initWatchServiceModifiers public

### LINKED ISSUE #188: Make FilesystemWatchPollingStrategy.initWatchServiceModifiers public
Creating this bug as the result of investigating issue #181 

The class `io.helidon.config.internal.FilesystemWatchPollingStrategy` implements polling strategy using `java.nio.file.WatchService`.

WatchService implementation in the JDK is either platform specific if supported, is a based on a polling mechanism if not supported `sun.nio.fs.PollingWatchService`.

The polling interval of `sun.nio.fs.PollingWatchService` can be configured by passing instances of `java.nio.file.WatchEvent.Modifier` to `java.nio.file.Path.register`. `java.nio.file.WatchEvent.Modifier` is implemented by `com.sun.nio.file.SensitivityWatchEventModifier` that defines 3 constants:
- HIGH (2s)
- MEDIUM (10s)
- LOW (30s)
`sun.nio.fs.PollingWatchService` uses `MEDIUM` (10s) by default.

Java does not support native WatchService for MacOS. While MacOS is not a target production platform, it is used for development and this behavior leads to inconsistent test behaviors.
Instead of forcing users to be aware of the default behavior for WatchService on MacOS and have them add sleep >10s in their code, we should provide a way to pass `Modifier` instances to `FilesystemWatchPollingStrategy`.

There is actually a method `initWatchServiceModifiers` on `FilesystemWatchPollingStrategy` that does exactly that, but is package protected and used exclusively by tests. The intent was probably (you guessed it) for cross-platform consistency.

We should allow our users to do the same with their tests, and thus make this method public.

This means users would tap into our internal API (i.e `FilesystemWatchPollingStrategy`) and configure the polling interval for `sun.nio.fs.PollingWatchService` by passing a modifier instance of `com.sun.nio.file.SensitivityWatchEventModifier` which is considered internal Java API.

An example code would look like this:
```java
Path configFile = new File("/tmp/application.yaml").toPath();
FilesystemWatchPollingStrategy pollingStrategy = new FilesystemWatchPollingStrategy(configFile, null);
pollingStrategy.initWatchServiceModifiers(SensitivityWatchEventModifier.HIGH);
Config config = Config.from(ConfigSources
        .file("/tmp/application.yaml")
        .pollingStrategy(pollingStrategy)
        .build());
```

romain-grecourt:
Fixes #188

### DIFF
--- a/config/config/src/main/java/io/helidon/config/internal/FilesystemWatchPollingStrategy.java
+++ b/config/config/src/main/java/io/helidon/config/internal/FilesystemWatchPollingStrategy.java
@@ -117,7 +117,13 @@ public class FilesystemWatchPollingStrategy implements PollingStrategy {
                 });
     }
 
-    void initWatchServiceModifiers(WatchEvent.Modifier... modifiers) {
+    /**
+     * Add modifiers to be used when registering the {@link WatchService}.
+     * See {@link Path#register(WatchService, WatchEvent.Kind[],
+     * WatchEvent.Modifier...) Path.register}.
+     * @param modifiers the modifiers to add
+     */
+    public void initWatchServiceModifiers(WatchEvent.Modifier... modifiers) {
         watchServiceModifiers.addAll(Arrays.asList(modifiers));
     }
 


config/config/src/main/java/io/helidon/config/internal/FilesystemWatchPollingStrategy.java
```diff
@@ -117,7 +117,13 @@ private void fireEvent(WatchEvent<?> watchEvent) {
...
     }
 
-    void initWatchServiceModifiers(WatchEvent.Modifier... modifiers) {
+    /**
+     * Add modifiers to be used when registering the {@link WatchService}.
+     * See {@link Path#register(WatchService, WatchEvent.Kind[],
+     * WatchEvent.Modifier...) Path.register}.
+     * @param modifiers the modifiers to add
+     */
+    public void initWatchServiceModifiers(WatchEvent.Modifier... modifiers) {
```
tjquinno: Our `module-info.java` for config contains:

`    exports io.helidon.config.internal to io.helidon.config.git;`

so, among other things, this class (in Java 9 and later) is not visible outside the config module itself (except to our git module, obviously). I don't think we want to unconditionally export all (or any) of a package called `internal`.

`PollingStrategies` in `io.helidon.config` is pretty much the outward interface config exposes to the world for change notification. Do we need to expose the ability to set `Modifier`s by adding a new method to  `PollingStrategies.FileSystemWatchBuilder`, for example?

romain-grecourt: I've create issue #202 to track this.

PR #204, opened 2018-11-17
Proposal for generic type mapping.

### LINKED ISSUE #201: Create proposal for generic type mapping support
We need to provide a general mechanism to map generic types.

Currently we have methods like `as(Class<T> type)`, or `asList(Class<T> type)` but these have some limitations.

Come-up with an equivalent of JAXRS's GenericType, and an API guideline for the various Helidon APIs (WebServer, Config etc.).

romain-grecourt:
Fixes #201

### DIFF
--- a/docs-internal/generic-type.md
+++ b/docs-internal/generic-type.md
@@ -1,0 +1,78 @@
+# Helidon Generic Type Support Proposal
+
+Provide a utility class to express generic type literals and provide API guidance
+ for the Helidon APIs (config, webserver etc).
+
+## Problematic
+
+Currently (v0.10.5) entity mapping is done with `as` methods that accept a `Class<T>` argument:
+
+```java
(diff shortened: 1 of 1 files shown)


docs-internal/generic-type.md
```diff
@@ -0,0 +1,78 @@
...
+
+```java
+request.content().asList(JsonObject.class)
+    .thenAccept((List<JsonObject> jsons) -> { /* ... */ });
+```
+
+Get the request content as `Pet<Dog>`:
+
+```java
+request.content().as(new GenericType(Pet<Dog>)(){})
```
tomas-langer: Typo, should be `new GenericType<Pet<Dog>>() {}` I think

romain-grecourt: oops, good catch, I'll fix that

PR #208, opened 2018-11-21
Mp jwt auth

tomas-langer:


tomas-langer:
The SecurityContext would not be injected (in some TCK tests we would get an exception that we are not in request scope).
When we changed the producer to request scope the problem was fixed and everything is injected correctly.
I was not able to find the root cause of the problem - it is somewhere in integration of CDI and Jersey...

PR #209, opened 2018-11-27
WIP: Config api changes

tomas-langer:
Modified API.
Config-beans moved to a separate module.

All usages are fixed.


tomas-langer:
This is a major update - please have a look on usages of config (outside of config module itself), as they most clearly illustrate the modified API.

tomas-langer:
Simplified review:
/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
/security/integration/jersey/src/main/java/io/helidon/security/jersey/SecurityFeature.java
- shows that the ConfigValue implements all methods from Optional

/config/config/src/main/java/io/helidon/config/spi/ConfigMapperProvider.java
- new extension mechanism (provides more options)

/examples/todos/demo-frontend/src/main/java/io/helidon/demo/todos/frontend/BackendServiceClient.java
- when you want to get actual value with no default (and get an exception if missing)
 - this is the case where we need to code more than before

/examples/todos/demo-frontend/src/main/java/io/helidon/demo/todos/frontend/EnvHandler.java
- example with a default

/security/security/src/main/java/io/helidon/security/Security.java
- uses multiple methods to show a more complete picture

/webserver/examples/comment-aas/src/main/java/io/helidon/webserver/examples/comments/Main.java
- how to use config.as() without reflection (method reference)

/config/tests/integration-tests/src/test/java/io/helidon/config/tests/DemoTest.java
- usage in tests - both "simpleValue" approach and getting a value approach

romain-grecourt:
We need to coordinate with https://github.com/oracle/helidon/blob/master/docs-internal/generic-type.md ; some of the API here would be impacted.

tomas-langer:
Related to @romain-grecourt comment:

`ServerConfiguration serverConfig = config.get("webserver").as(ServerConfiguration::fromConfig).get();`

and 
`ServerConfiguration.fromConfig(config.get("webserver"));`

While you are right the second approach is more straight-forward, it does not give you the option to handle easily cases when the config is not present.
e.g.:
```
config.get("webserver").as(ServerConfiguration::fromConfig).ifPresent(this::configureServer)
```
We do not use this approach in webserver, it is used in other places.
Also the developer has a choice to use either, depending on whether the config is mandatory or optional.

tomas-langer:
Closing this pull request, as it was approved. Using pull request to the config api branch from now.

PR #215, opened 2018-11-29
Updates and additions for the MP RESTful web service guide

tjquinno:
This PR is for the guide for the MicroProfile RESTful web app.

Romain and Joe and I met a couple weeks ago about the SE (not MP) guide. These changes _do not_  incorporate the improvements we talked about in that meeting. I wanted to get this guide published, and then I can update the SE and MP guides together (see
https://github.com/oracle/helidon/issues/176) to reflect that meeting. 

As part of this change I have added tags to the SE guide's `.adoc` file and `include` those segments into the MP guide. I have created a separate issue to think about and do some refactoring if we feel it's worthwhile to pull the common portions out to yet another file and have both guide `.adoc` files `include` from the refactored common ones.


docs/src/main/docs/guides/01_SE_REST_web-service.adoc
```diff
@@ -20,7 +20,7 @@
 :example-src-url: https://github.com/oracle/helidon/tree/master/examples/guides/se-rest-web-service/src/main/java
 
 
-= Creating a ReSTful Web Service using Helidon SE
+= Helidon SE RESTful Web Service
```
kumar-dhanagopal: I suggest: "Create a RESTful Web Service Using Helidon SE"

The "Create" verb is a useful signpost to help users know in advance what content they can expect in the doc. They'll know that the doc contains a step-by-step procedure, as opposed to a conceptual outline, for example.

PS: In formal Oracle docs, the gerund form "Creating.." used to be the norm for titles. But that's changing now to the simpler direct verb form.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+:greet-application-class: {java-base}/GreetApplication.java
+:health-resource-class: {java-base}/HealthResource.java
+:pom: {adoc-dir}/../../../pom.xml
+:meta-inf-dir: {adoc-dir}/../resources/META-INF
+:beans-xml: {meta-inf-dir}/beans.xml
+:config-file: {meta-inf-dir}/microprofile-config.properties
+:logging-properties-file: {adoc-dir}/../resources/logging.properties
+:se-guide-adoc: {guides-dir}/se-restful-webservice/src/main/docs/index.adoc
+
+
```
kumar-dhanagopal: I would like to read through & review this content, but it'll take me some time. By when do you need this review to be completed?

docs/src/main/docs/guides/01_SE_REST_web-service.adoc
tjquinno: the issue we have is the left nav panel on the page is so narrow that the "SE" part and the MP part for the corresponding label for the MP guide are obscured. The user does not see a visible difference (although the tool tips do show the entire label).

We could try "Create Helidon SE RESTful Web Service" although stringing together so many nouns consecutively can become less clear - but here brevity is in order!

tjquinno: I like that a lot. I'll make that change for both guides and push those changes.

tjquinno: I made those changes locally and the nav panel shows: 

Helidon SE: Create a RESTful Web...
Helidon MP: Create a RESTful Web...

but that's probably as good as it gets. I'll push those changes.

tjquinno:
@kumar-dhanagopal I am going to be out of the office some on Friday and some on Monday. If I aim to do the merge on Tuesday of next week is that enough time?

kumar-dhanagopal:
> 
> 
> @kumar-dhanagopal I am going to be out of the office some on Friday and some on Monday. If I aim to do the merge on Tuesday of next week is that enough time?

Yes, that will give me time to do a quick review. Thank you! I'll send in my comments by CoB Monday, PT.

tjquinno:
One way to look at the pages live is to clone my forked repository and run a simple HTTP server locally (if I remember this correctly...):

```
cd to the top level of the just-cloned workspace
mvn install
mvn install -Psources
cd javadocs ; mvn generate-sources
cd ../docs ; mvn package
cd target/docs ; python -m SimpleHTTPServer 8080
```

Then in a browser go to `http://localhost:8080` and expand the `Guides` section in the left-hand nav panel.

If the sequence of commands does not work please get in touch with Romain while I am out.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+|`PUT localhost:8080/greet/greeting/Hola` |Changes the greeting used in subsequent responses
+|===
+
+We create the app in these main steps:
+
+. Write a basic Helidon MP app to respond to the HTTP requests.
+. Add code to perform simple health checks.
+. Add code to record simple metrics.
+
+(See the last section to find out how to get the entire finished code for this example.)
```
kumar-dhanagopal: CONTENT: Could you add a link to the download section?

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+|Maven 3.5 or later
+|===
+
+//Optional:
+//|===
+//|Docker 18.02 (use the Edge channel to run Kubernetes on your desktop)
+//|`kubectl` 1.7.4
+//|===
+
+== Create Your Application
```
kumar-dhanagopal: I suggest the title "Develop Your Application".

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+
+=== Create a new Maven Project
+You can create your maven project in these ways:
+
+* use your IDE to create a new Java maven project, or
+* run the standard maven archetype to create a new Java project using this command:
++
+[source,bash]
+----
```
kumar-dhanagopal: CONTENT: To enable users running Maven on Windows to copy and paste the archetype:generate commands at the CMD prompt, consider showing the entire command in one line in the doc, i.e., without the \ line separators.

kumar-dhanagopal: CONTENT: Add an instruction to enter the groupID, artifactId, version, and package when prompted.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
+
```
kumar-dhanagopal: STYLE: A general observation - Use either the title case or the sentence case, consistently for all the sections.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+= The MP RESTful Web Service Guide
+:description: Helidon MP guide restful web service
+:keywords: helidon, guide, example, mp
+:toc:
+
+Create and build a JAX-RS RESTful web service as your first Helidon MP application.
+
+== What You Will Learn
+You'll learn how to use Helidon MP quickly to create a JAX-RS RESTful web service that accepts these HTTP requests:
```
kumar-dhanagopal: STYLE: "quickly to create" --> to quickly create

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: STYLE: "entire finished code" --> "complete code"

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+. Add code to record simple metrics.
+
+(See the last section to find out how to get the entire finished code for this example.)
+
+== What You Need
+
+|===
+|About 15 minutes
+|An IDE or text editor
+|JDK 1.8 or later
```
kumar-dhanagopal: CONTENT: "JDK 1.8" --> "JDK 8"

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+[source,bash]
+----
+mvn archetype:generate \
+    -DarchetypeGroupId=org.apache.maven.archetypes \
+    -DarchetypeArtifactId=maven-archetype-quickstart \
+    -DarchetypeVersion=1.3
+----
+
+=== Update Your `pom.xml`
+==== The Essentials
```
kumar-dhanagopal: STYLE: Is this subheading really required here?

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+----
+<1> Incorporates the Helidon MicroProfile bundle.
+<2> Pulls in JSON-B support.
+<3> Adds jandex, a JAX-RS bean search optimizer.
+<4> Adds the activation API.
+
+If you run your project from the IDE, the IDE typically handles the main class and places
+dependent JARs on the runtime classpath for you and your pom is now ready to go. 
+
+=== Create a MP Config File
```
kumar-dhanagopal: STYLE: Create "an" ...

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+----
+mvn archetype:generate \
+    -DarchetypeGroupId=org.apache.maven.archetypes \
+    -DarchetypeArtifactId=maven-archetype-quickstart \
+    -DarchetypeVersion=1.3
+----
+
+=== Update Your `pom.xml`
+==== The Essentials
+Make sure your POM contains these sections:
```
kumar-dhanagopal: CONTENT: Ask the user to `cd` to the <artifactId> directory.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+
+. Declare the class with these annotations:
++
+[source,java]
+----
+include::{health-resource-class}[tag=classDecl]
+----
+
+. Declare an injected instance field to refer to the greeting message resource. This
+is the only input to the activeness check in our simple implementation.
```
kumar-dhanagopal: STYLE: "to the activeness check" --> Consider rephrasing as "to check whether the service is active"

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+[source,java]
+----
+include::{health-resource-class}[tag=greetingDecl]
+----
+<1> Indicates that JAX-RS should inject the field.
+<2> Because `GreetingMessage` is an application-scoped bean, JAX-RS will inject a
+reference to the single instance every time it creates this request-scoped resource.
+
+. Add the `checkHealth` method which decides whether the greeting resource is active
+(health) or not.
```
kumar-dhanagopal: STYLE: "(health)" --> "(healthy)"

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+_not_ decrement it when the method returns. 
+
+=== Rebuild and rerun your application
+Follow the same steps as before, remembering to stop any instance of your application
+that is still running.
+
+include::{se-guide-adoc}[tag=sendRequestsForMetrics]
+
+=== Send some requests
+Use the same `curl` commands from the beginning of the guide to send requests to
```
kumar-dhanagopal: CONTENT:  "same `curl` commands from the beginning of the guide " --> At the beginning, we provide only the HTTP calls, not the curl commands. 

So I suggest: "Use the following `curl` commands to send requests to the service:"

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+. As expected, the value for the counter is 4.
+
+A `curl` to `http://localhost:8080/metrics` lists not only our application-scoped 
+metric but all the _base_ and _vendor_ metrics as defined in the MP metrics spec. 
+For example, you will see a `vendor:requests_count` 
+counter. This will be larger than our counter because that counter also tracks 
+requests to the `/metrics` path itself; our `accessctr` counter tracks only requests to
+our application endpoints.
+
+== Or Download the Example Source
```
kumar-dhanagopal: STYLE: I suggest "(Optional) Download the Example Source Code"

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: About the backslashes... An explicit recommendation from an earlier review was to _add in_ the backslashes and break the example onto separate lines because the single-line version did not wrap -- it would scroll horizontally -- and therefore be awkward for the user to view all of it. 

That said, you raise a good point. One option (despite the mild maintenance pain it would induce) would be to have two code examples for long commands: one Unix-style and one Windows (using the ^ as the continuation character). 

I think this `mvn` command is the only one in this guide long enough to cause wrapping/scrolling issues.

If we think this is a good idea then for consistency we probably need to be prepared to adopt it in all content, not just the guides, on our site where we display long commands.

Thoughts?

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: Do you have a strong suggestion one way or the other?

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: A writer suggesting a split infinitive!

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: The [Oracle Style Guide](http://st-doc.us.oracle.com/dp/1/doctools.101/e23911/grammar.htm#i1005871) prescribes the title case for section headings in docs that we publish on docs.oracle.com. Helidon docs aren't published to docs.oracle.com, but I suggest we follow the same standard.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: > have two code examples for long commands: one Unix-style and one Windows (using the ^ as the continuation character).

I agree.

barchetta: To be honest, our guides are going to have a Unix bias. For example the Dockerfile guide I wrote uses `du`, `ls`, etc. And I'm hesitant to harm readability/simplicity by always including Windows syntax. I'm OK with 80% solutions.

But I do see how it would be good for these long mvn archetype commands. Tim, do you think we could arrange the two snippets side-by-side? That might look OK and not overly clutter the document.

tjquinno: I'll look into a side-by-side display.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: OK. Will do.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: How about "You'll learn how to use Helidon MP to create quickly a JAX-RS..."

Or we just don't worry about the split infinitive as you suggested. ("To boldly go" and all that.)

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
romain-grecourt: Side by side is probably not a good idea.
Having them in a list would make more sense.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
romain-grecourt: @kumar-dhanagopal if we should follow any standard it should not be proprietary. Either we find one that is widely used by open source projects or we make one for ourselves.

romain-grecourt: Also please avoid internal links on github.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: Yeah, i don't think we should worry about that too much.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: Good point about internal links, Romain... I forgot the context; sorry!

tjquinno: About the heading style... I bow to Kumar's greater wisdom on this. Whatever you say, @kumar-dhanagopal, we'll go with!

About internal links: @romain-grecourt Do internal links (internal cross-references, i.e. `<<xxx>>`) not work on GitHub?? There are several scattered throughout the guide. Do we need to change them to  `link:`-style? There are warnings in the AsciiDoctor user's guide against that practice, at least for inter-document cross-references (which we use). 
https://asciidoctor.org/docs/user-manual/#inter-document-cross-references

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: Done.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: Done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: Personally, I like sentence case, because it is less formal. Title style is old school, but its use won't disappear for a while. 

We can't go horribly wrong with either choice. Consistency is key.

I suggest we pick title case as our standard, unless anyone has strong concerns, and use it consistently for content we write going forward.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: Done as a list.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: removed

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
```diff
@@ -0,0 +1,487 @@
...
+We think the guide is most useful if you follow along step-by-step, building
```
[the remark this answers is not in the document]
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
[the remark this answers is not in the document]
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
kumar-dhanagopal: > ... Do internal links (internal cross-references, i.e. `<<xxx>>`) not work on GitHub?

I think @romain-grecourt 's comment was about links to Oracle-internal docs, like the style guide link in my earlier comment.

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: done

examples/guides/mp-restful-webservice/src/main/docs/index.adoc
tjquinno: I would vote for sentence case, I think, because of the "new-school-ness" as you pointed out. I don't have _really_ strong feelings about this, but as the "old-school" way grows even older as time goes on our content might seem increasingly old and remain that way as we strive for consistency if we adopt the title case convention now.

tjquinno: @kumar-dhanagopal Any reaction to my recent thoughts from yesterday... Adopting "old-school" title case would lock us in going forward and (possibly) make us seem older and older school as time goes on.

kumar-dhanagopal: Sorry, I missed responding to that...

Good point! 

I looked at a few other sites, to get an idea of the direction the "industry" is taking. 

Kubernetes: No real consistency: title case on some pages (https://kubernetes.io/docs/setup/) and sentence case in others (e.g., https://kubernetes.io/docs/setup/independent/install-kubeadm/)

FnProject: Again no consistency: https://github.com/fnproject/docs/blob/master/fn/general/introduction.md 

Docker: Sentence case (https://docs.docker.com)

Apache Kafka: Mostly title case, but stray cases of sentence case usage (http://kafka.apache.org/documentation/)

I don't have time right now to do more extensive research. But if we had to take a call right now, I agree with you with the sentence case would give our content a more conversational and modern look. Let's go with that.

tjquinno: Thanks for the research and your thoughts. I'll change the MP guide and check the SE one to adopt sentence case.

kumar-dhanagopal: Perhaps we can publish as is now, and pick up this task later.  I would be happy to check & update the case for all the section titles, Tim, if that's ok by you of course.

tjquinno: Ha! I'm already part-way through the MP changes (luckily there are not so many headings affected) and the changes to the SE guide will be similarly quick. There will be plenty of heading fix-up in other documents to tend to later!

tjquinno: Done

tjquinno:
@kumar-dhanagopal Just for my own planning, do you have an ETA for review phase 2 as described in #219? Thanks.

kumar-dhanagopal:
> 
> 
> @kumar-dhanagopal Just for my own planning, do you have an ETA for review phase 2 as described in #219? Thanks.

I'm going to try rebuilding the docs tomorrow. If that works, then I'll set aside time to complete the phase-2 review by Tue next week.

tjquinno:
Received email from Kumar with these notes and my emailed responses:

> GreetResource.java
When I added @Produces(MediaType.APPLICATION_JSON), my IDE (NetBeans) prompted me to import the following packages.
image
> I selected javax.ws.rs.core.MediaType as in the doc. But the IDE then displays a ‘Cannot find symbol’ error. What could be wrong?

I’m not sure. I just tried it again myself and it worked fine for me. That class is in the javax.ws.rs-api-2.1.jar (maven group javax.ws.rs, artifact javax.ws.rs-api) which is pulled in as a result of adding this dependency to the pom.xml:

        <dependency> 
            <groupId>org.glassfish.jersey.media</groupId>
            <artifactId>jersey-media-json-binding</artifactId>
        </dependency>

which is one of the dependencies the guide describes adding. Is that in your pom?

> Main class
For Main.class.getResourceAsStream("/logging.properties"));, the IDE prompted me to import from the following options:
image
> None of these packages is in the doc. Which should we import?

There should be no need to import anything. The Main.class is intended to refer to the current Main class, not some other one. 

> pom.xml
Should we replace the <properties> element in the generated pom.xml (attached) with what’s in the doc?
Should we replace the <dependencies> element in the generated pom.xml, or only add the dependencies from the doc and leave the auto-generated ones as is?
Should we keep the <build> element as is in the generated pom?

How did you create your pom, just out of curiosity?

The intent is that the pom contents in the guide are merged into whatever pom the generic Java maven archetype or your IDE created. So if what the guide says is already there, replace it; if absent, add it. I will clarify that in the guide text.

> This sentence:
“If you run your project from the IDE, the IDE typically handles the main class and places dependent JARs on the runtime classpath for you and your pom is now ready to go.”
à Otherwise, i.e., for users who use a text editor for example, what instructions should we add?

There is a section in the SE guide (last part of "Update your pom") for exactly this purpose that, for brevity after some other review discussions, I left out of the MP guide. I have added the same notes into the MP guide in light of your comment. 

> src/main/resources
This directory didn’t exist when I ran mvn archetype:generate. Is that the expected behavior? If yes, then in Create an MP config file, we should add a step to create that directory.

I think that’s normal. I suppose we _could_ add explicit directions for the reader to create missing directories (the developer also needs to create the directory for the MP config file) but the guide gives the full path (within the maven project) where each of these files is to reside, and to my mind it’s not asking too much for a developer to know to create a directory we specify for a file if the directory is not already there.

kumar-dhanagopal:
@tjquinno - Hi Tim, I finished reviewing (and trying out) the remainder of the MP RESTful web services guide.

1. In the code snippets for `GreetResource.java` and `HealthResource.java`, for the benefit of users who'll copy-paste code as is, consider doing one of the following:
    - Include the closing "}" as well. This is my preference.
    - Or, omit the class definition statement (assuming the IDE or user will add it) and keep only the annotations.

2. When introducing the packages to be imported, in most of the sections I see: "You or your IDE will add these imports along the way".  I suggest that we explicitly ask users to add the import statements listed in the doc. Otherwise, when the IDE prompts them to choose from the possible packages, users might pick an incorrect package and end up spending time troubleshooting, like I was trying to do.

3. In **Build and run**, we say: "...run these commands to access its **three functions** (order is important for the last two)".  By "three functions" I guess you meant (1) get generic greeting, (2) get personalized greeting, and (3) change the greeting.  But because the phrase "three functions" occurs so close to "last two" (which refers to #3 and a repeat of #2), at the first reading, it appears that the command table lists 4 functions. To solve this problem, I suggest dropping "three". Also, consider adding a "Function" column (perhaps as the first one in the table) and state the goal of each command in that column.

4. Regarding the line "**liveness** - often a more thorough assessment **of if** and how well the service can do its job":
     - "liveness" is an odd term. Would "health" work? If yes, pls note that the doc has multiple occurrences of "liveness", "alive" etc. I recommend changing all of them (even the /alive path in the code if possible) to "health", "healthy", etc. as appropriate.
     - Also, to avoid the awkward "of if" combination in that sentence :-), consider changing "assessment of if" to "assessment of whether".  As a general practice, I recommend using "whether" and not "if" in such situations.

5. In the sentence, "For this example we define our service as **active**, consider using "healthy" in place of "active". 

6. In **Update GreetApplication**, for the sake of clarify, I suggest adding a snippet showing the code after making this change:

```
@Override
        return CollectionsHelper.setOf(
                GreetResource.class,
                HealthResource.class
        );
```
7. "Rebuild and rerun your service". The first time I saw this section, I read “Follow the same steps you did before” and then jumped ahead to rebuilding/rerunning the app, **without reading the part about stopping the running app**. And of course, I ran into the ‘unable to bind port’ error. To prevent users from committing this error, I suggest turning that para into 2 steps:
    - 1. Stop any running instance of your app.
    - 2. Rebuild the app and then run it.

8. "Download the example source (optional)"
     - The convention in official docs is to write “(Optional)” first and then the heading text, i.e., "(Optional) Download the example source", so that users notice that the step is optional before they read the heading.
     - “writing the application yourself” --> consider changing to “building the application using the code snippets in this guide”

tjquinno:
@kumar-dhanagopal Thanks for the further comments. I've adopted most of your suggestions. Exceptions only noted below.

> 4. Regarding the line "**liveness** - often a more thorough assessment **of if** and how well the service can do its job":
>    
>    * "liveness" is an odd term. Would "health" work? If yes, pls note that the doc has multiple occurrences of "liveness", "alive" etc. I recommend changing all of them (even the /alive path in the code if possible) to "health", "healthy", etc. as appropriate.
> 5. In the sentence, "For this example we define our service as **active**, consider using "healthy" in place of "active".

"Liveness" is a term of art in the Kubernetes world (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/) 
and "health" is the broader category that includes readiness and liveness. I used those terms in our guides specifically to align with the K8s usage.

tjquinno:
I have pushed my latest updates to the fork, so if you want you can pull and rebuild to inspect the changes locally.

kumar-dhanagopal:
> "Liveness" is a term of art in the Kubernetes world (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/)
> and "health" is the broader category that includes readiness and liveness. I used those terms in our guides specifically to align with the K8s usage.

Thanks for clarifying that, @tjquinno ! Is it reasonable/safe to assume that microservices will be deployed typically (only?) to K8s containers?

In the sentence: ""For this example we define our service as **active**..." could you use "alive" instead of "active?

kumar-dhanagopal:
> 
> 
> I have pushed my latest updates to the fork, so if you want you can pull and rebuild to inspect the changes locally.

Sure. I have a few JCS/PSM tasks to take care of today and tomorrow. After that, I'll pull the updates, rebuild, and verify. Thank you!

tjquinno:
> Thanks for clarifying that, @tjquinno ! Is it reasonable/safe to assume that microservices will be deployed typically (only?) to K8s containers?

Not necessarily, but I think readers _will be_ familiar with these terms and their use in K8s.

kumar-dhanagopal:
> 
> 
> I have pushed my latest updates to the fork, so if you want you can pull and rebuild to inspect the changes locally.

Done. I verified the fixes in a local build. They look good!

Just one incremental comment ...

In **Update your pom.xml**, I think  the parenthetical phrase ...

> possibly in addition to others

... means that users _must_ add the properties & dependencies listed in the doc if they don't exist, but that they _can_ choose to leave any other existing properties & dependencies in the pom as is.  If that's the intent, then I suggest making it clearer. Consider rephrasing as: "Add the properties and dependencies listed here if they don't exist in your `pom.xml`).

romain-grecourt:
@kumar-dhanagopal @tjquinno Shouldn't Maven be capitalized ?

kumar-dhanagopal:
> 
> 
> @kumar-dhanagopal @tjquinno Shouldn't `Maven` be capitalized ?

Yes, it should. Good catch!

The following lines of text should be updated:

- Create a new maven project
- You can create your maven project in these ways:
- use your IDE to create a new Java maven project, or
- run the standard maven archetype to create a new Java project using this command:
- To run maven outside your IDE
- If you want to use maven yourself, 
- This is typical with maven projects 
- Instructs maven what main class to set in the
- Tells maven to package the dependency 
- Or, to use maven outside the IDE,

PR #217, opened 2018-11-30
Quickstart SE changes to use a single Configuration

tomas-langer:


tomas-langer:
Also
 - removed final from parameters, as we do not use it in Helidon.
 - created a single sendResponse method to remove duplicate code
 - print the correct URL for the server being started (so when you click on it in standard output, you get a response)

Issue #235, opened 2018-12-06
Helidon 0.11.0-SNAPSHOT build failed


othmaan:
Maven build fails with if using Helidon  version 0.11.0-SNAPSHOT.

## Environment Details
* Helidon Version: 0.11.0-SNAPSHOT
* Helidon SE or Helidon MP
* JDK version: 8 (191)
* OS: 10.14.2 Beta (18C48a)
* Docker version (if applicable): N/A

----------

## Steps to reproduce
Use below pom.xml
```
    <groupId>io.helidon.examples</groupId>
    <artifactId>quickstart-mp</artifactId>
    <version>0.11.0-SNAPSHOT</version>
    <packaging>jar</packaging>
    <name>${project.artifactId}</name>
```

and properties
```
 <properties>
        <helidon.version>0.11.0-SNAPSHOT</helidon.version>
        <package>io.helidon.examples.quickstart.mp</package>
        ...
    </properties>
```

Below build errors are thrown
```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Non-resolvable import POM: Could not find artifact io.helidon:helidon-bom:pom:0.11.0-SNAPSHOT @ line 153, column 25
[ERROR] 'dependencies.dependency.version' for io.helidon.microprofile.bundles:helidon-microprofile-1.2:jar is missing. @ line 194, column 21
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project io.helidon.examples:quickstart-mp:0.11.0-SNAPSHOT 
```

Only fix I found is to change `<helidon.version>` in pom properties, to be like that:
```
 <properties>
        <helidon.version>0.10.5</helidon.version>
        <package>io.helidon.examples.quickstart.mp</package>
        ...
    </properties>
```

romain-grecourt:
How did you generate your project ? Did you use the archetype ?
Note that in order to use a SNAPHOT version you'd have to build the helidon workspace first.

barchetta:
I'm closing this as pilot error. External projects should only use released versions (like 0.10.5). If for some reason you want to try a SNAPSHOT version of Helidon then you must build the Helidon master locally so that it populates your local repository with the SNAPSHOT.

Note if you want to build an example from the Helidon repository without building then entire repo, then it is best to pull the repository from a release tag (like "0.10.5") since that will use a released version of Helidon and not a SNAPSHOT version.

othmaan:
[...]

PR #236, opened 2018-12-06
Update pom.xml

othmaan:
- Change helidon version to fix build issue #235


romain-grecourt:
The workspace is streamlined on the same SNAPSHOT version. If you need a pom that depends on a released version of helidon you should use the maven archetype, see https://helidon.io/docs/latest/#/getting-started/02_base-example

PR #238, opened 2018-12-06
GenericType support for config mapping.

tomas-langer:


tomas-langer:
Also fixes priorities to work according to documentation

config/config/src/main/java/io/helidon/config/Config.java
```diff
@@ -1424,6 +1445,26 @@ default Builder sources(Supplier<ConfigSource> configSource,
...
+         *
+         * @param type   class of type the {@code mapper} is registered for
+         * @param mapper mapping function
+         * @param <T>    type the {@code mapper} is registered for
+         * @return an updated builder instance
+         * @see #addStringMapper(Class, Function)
+         * @see #addMapper(ConfigMapperProvider)
+         * @see #disableMapperServices
+         */
+        <T> Builder addMapper(Class<T> type, Function<Config, T> mapper);
```
romain-grecourt: Shouldn't we have have a method to register with a GenericType?

tomas-langer: I was thinking about it and there are a few reasons why I did not create one:

1. GenericType support is usually implemented for more than one type argument (e.g. I would implement Map<String, ?>, rather than just Map<String, Integer>) - so this type of registration does not work
2. For the GenericType mapping, there is actually a BiFunction<Config, ConfigMapper>, as one needs much more flexibility when mapping generic type (e.g. Map<String, Set<Foo>> must be mapped using another GenericType) - so this would require a method with a different signature

So in general - I feel that using GenericType is an advanced use case and is only available through ConfigMapperProvider (which can be registered with builder as well).

tomas-langer: Or we would need to add method like this:
`<T> Builder addMapper(Function<GenericType<T>, Optional<BiFunction<Config, ConfigMapper, T>>> mapperFunction);`
which is quite complicated...

romain-grecourt: Not sure I follow, why just simply add:

```java
<T> Builder addMapper(GenericType <T> type, Function<Config, T> mapper);
```

tomas-langer: Because that would only work for a very specific GenericType - e.g. for Map<String, Integer>. If you wanted to support Map<String, any other type>, you would not be able to register this

romain-grecourt: My point is since we now have GenericType we must now mirror all methods that map a generic type parameter from a `Class<T>`.

If there are limitations with GenericType and wildcard type parameters we should document this properly or provide a way to deal with this (maybe something like you mentioned above).

Having 2 type parameters (or more) is currently not supported because by the API. This is a limitation that forces the user to register specific mappers for all supported types for the second parameters. Having a wildcard parameters is a bit of weird shortcut IMO.

We could think of expanding the code in GenericType to support wildcard type parameter. This could perhaps be an escape hatch that would allow N type parameters.  

```java
addMapper(new GenericType<Map<String, GenericType<?>>();
GenericType<Map<String, GenericType<?>> type = new GenericType<Integer>(){}){};
config.as(type);
```

Clearly this code is more complex, but it is a trade off since it enables a complex use-case.
There might be ways to improve this for primitive classes (Long, String, Boolean etc..).

However I'm wondering if this is worth the effort.

PR #243, opened 2018-12-06
Corrected the 'java -version' command in the quickstart guides

kumar-dhanagopal:
The existing command 'java --version' returns an error in JDK 8.

Confirmed by @romain-grecourt and @tjquinno .


kumar-dhanagopal:
Updated the MP quickstart, as well as the getting-started doc

PR #247, opened 2018-12-08
WIP: Fixes #242 by eliminating raw type usage of Routing.Rules.

### LINKED ISSUE #242: Usage of raw type in public Service method
Before we bake the API completely, I'd like to see this method use `Routing.Rules<?>` instead of just `Routing.Rules` so we can avoid raw type warnings:

https://github.com/oracle/helidon/blob/a52d082939216e85cd6bebe8417a926e86d1f9db/webserver/webserver/src/main/java/io/helidon/webserver/Service.java#L33

I haven't checked to see if anything will break as a result.

ljnelson:
Signed-off-by: Laird Nelson <ljnelson@gmail.com>

There are two main components to this change, which fixes #242:
1. `Routing.Rules` has been defined to be `Routing.Rules<T extends Routing.Rules<T>>`.
2. All _usages_ of `Routing.Rules` as a raw type have been replaced with `Routing.Rules<?>`.

**This is an alternative to #248.  Only one of these two PRs (if any) should be merged.**

`mvn clean install` passes on my laptop from the root.

### DIFF
--- a/examples/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
+++ b/examples/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
@@ -56,7 +56,7 @@ public class GreetService implements Service {
      * @param rules the routing rules.
      */
     @Override
-    public void update(Routing.Rules rules) {
+    public void update(Routing.Rules<?> rules) {
         rules
             .get("/", this::getDefaultMessage)
             .get("/{name}", this::getMessage)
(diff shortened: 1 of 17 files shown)


barchetta:
Ack! This cure looks worse than the disease. Our API has now gotten uglier. It feels like some implementation detail (or obscure corner case) is impacting everyday use of the API.  I would not want to stand in front of a room of developers and try to explain why we have this wart (`Routing.Rules<?>`). 

I also acknowledge I don't fully understand the issue, but is there no way to change the implementation to avoid this (even if it makes the implementation uglier)?

I had a concern with some of the Config changes that we were making the API messier to simplify our implementation. This change adds to that concern.

ljnelson:
That's fair.  If removing the type parameter is on the table then I can try that in a separate PR and then we can pick which we like better.

If we leave things as they are, every user writing a `Service` by hand (and not just using lambdas) will encounter raw types warnings.  That is the equally nasty experience I was hoping to avoid.

ljnelson:
The collective hath spoken; closing in favor of #248.

PR #248, opened 2018-12-08
Fixes #242 by removing type parameter from Routing.Rules

### LINKED ISSUE #242: Usage of raw type in public Service method

ljnelson:
Signed-off-by: Laird Nelson <ljnelson@gmail.com>

This PR fixes #242 by removing the type parameter from `Routing.Rules` altogether.


tomas-langer:
I prefer this solution.

ljnelson:
Merging shortly.

PR #250, opened 2018-12-11
Update multipart proposal to fix a mistake in the buffered model.

romain-grecourt:
This highlights the inconsistency with sync/async in the buffered model.


romain-grecourt:
@spericas FYI

PR #262, opened 2018-12-13
Run TCK tests only when "tck" profile is active

spericas:
Note that all TCK modules are still available in the default profile, only their execution is controlled by the new profile. Make new profile active during pipeline builds.


spericas:
No, it just doesn't work well with Intellij. So let's leave as is for now until we find a solution that works better for everyone.

PR #265, opened 2018-12-15
[guides/docker] Docs on Containerizing completely within Maven via Jib

cricket007:
Figured this would go better into Helidon docs rather than https://github.com/GoogleContainerTools/jib/issues/347

### DIFF
--- a/docs/src/main/docs/getting-started/02_base-example.adoc
+++ b/docs/src/main/docs/getting-started/02_base-example.adoc
@@ -131,9 +131,9 @@ curl -X GET http://localhost:8080/greet/Jose
 
 == Build a Docker Image
 
-The project also contains a Docker file so that you can easily
-build and run a docker image. Because the example's runtime
-dependencies are already in `target/libs`, the Docker file is
+The project also contains a Dockerfile so that you can easily
+build and run a Docker image. Because the example's runtime
+dependencies are already in `target/libs`, the Dockerfile is
 pretty simple (see `target/Dockerfile`). To build the Docker
 image, you need to have Docker installed and running on your system.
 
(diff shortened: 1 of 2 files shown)


barchetta:
@cricket007 thanks for the PR! This is great info on Jib and we should definitely get this into our docs.  We also appreciate you suggesting adding a Helidon example to the Jib project. BTW, before we can accept a contribution you'll need sign the Oracle Contributor Agreement (https://www.oracle.com/technetwork/community/oca-486395.html).

Some general feedback on the PR: I would prefer if the Jib information were in its own section maybe at the bottom of the document. Something like "Creating Docker Images with Jib". Here is why:

One of our principles with Helidon is to be really boring and focus on use of vanilla tooling that developers already know. So no special Maven plugins required. That's the focus of the quickstart examples that this guide builds on. So I'd like that to remain the initial focus of the guide. Your change shifts the focus of the guide to using Jib.

That said, Jib is great, and I think having this info is valuable. So after the guide goes over the basics of building Helidon Docker images using vanilla tooling, I think it makes sense to add a section on building Docker images with Jib.

One thing I can do is create a PR from your branch with what I'm thinking (leveraging your commit), and then you can review that. Let me know what you think.

Thanks Again!

cricket007:
Hey @barchetta. Sure, I'll get the CLA signed.

Thanks for the feedback! 

Wishful thinking, but hopefully, Jib can one day be part of the vanilla tooling (for Java projects). In the meantime, you bring up a good point - the arrangement of the sections does shift the focus. It would be even worse if the Dockerfile wasn't already provided.

> One thing I can do is create a PR from your branch with what I'm thinking (leveraging your commit), and then you can review that

Fine by me!

barchetta:
I've created a new PR that includes the Jib content as a new section (See #290) so I'm closing this PR.

PR #270, opened 2018-12-19
WIP: examples reorg

### LINKED ISSUE #245: Examples: reorganize and cleanup
## Environment Details
* Helidon Version: 0.10.5
* Helidon SE and Helidon MP

----------

## Problem Description
Our examples are a bit messy. They're buried under each component which makes them difficult to navigate. Plus we're missing some basic stuff. Like do we have an example that shows how to handle posted json? 

Also, the examples are not buildable in master unless you do a priming build of the workspace (due to dependency on SNAPSHOT versions). We might be stuck with that, although at least one user has complained (see #235).

The ask is to:
- [x] Consider reorganizing the examples to the top level "examples" directory. Examples are currently spread out:
    * config/examples
    * microprofile/examples
    * security/examples
    * webserver/examples
    * examples/integrations
    * examples/guides
    * examples/\*quickstarts\*
    * examples/integrations
- [x] Consider pom organization. Some of our examples intentionally have a self-contained pom (quickstarts, todo) others have a parent pom.
- [x] Makes sure all have READMEs 
- [x] Think about reconciliation with guides

barchetta:
Fixes #245


barchetta:
I'm still working on the READMEs.

PR #282, opened 2018-12-22
Webserver dev guidelines

### LINKED ISSUE #192: WebServer: refactor based on development guidelines
Refactor the webserver module to follow the development guidelines in PR #135. 

The following tasks are part of this issue ("Validate" means check and if not compliant, fix):

- [x] Change builders that accept other builders to accept a supplier instead
- [x] Change scope to "provided" where reasonable to have a cleaner dependency graph
- [x] Validate bundle dependencies to not bring in too many third party libraries
- [x] Move SE bundles to a /bundles root module and change group ids accordingly??
- [x] Validate module names and package names to ensure they adhere to dev guidelines)
- [x] Validate naming convention of configuration keys used by components
- [x] Fix inconsistencies in getter and setter names
- [x] Validate we do not create instances using "new" and use factory methods
- [x] Refactor factory method names (use create)
- [x] Refactor config methods in builders (use config)
- [x] Validate java module files are in each module

And for tests:

- [x] Refactor unit tests to use Hamcrest assertions

tomas-langer:
Fixes #192


examples/guides/se-restful-webservice/src/main/java/io/helidon/guides/se/restfulwebservice/Main.java
```diff
@@ -19,30 +19,32 @@
...
+import io.helidon.webserver.ServerConfiguration;
+import io.helidon.webserver.ServerRequest;
+import io.helidon.webserver.ServerResponse;
+import io.helidon.webserver.WebServer;
+import io.helidon.webserver.json.JsonSupport;
+
+// tag::importsHealth1[]
 // end::importsHealth1[]
 // tag::importsStart[]
-import io.helidon.config.Config;
```
romain-grecourt: These tags are included in guides/01_SE_REST_web-service.adoc and your changes remove their content.
The pipeline failures seems to be that the sitegen template `vuetify/block_listing` does expect a non empty value and then freemarker throws an exception. The template should be improved to handle this better: either fail nicely or allow an empty value. I'll file a bug on the build tools repo,

romain-grecourt: filed https://github.com/oracle/helidon-build-tools/issues/14

tomas-langer: My mistake - automatic reordering of imports in IDE caused this. I have disabled it, reverted the change and now it should pass.

PR #296, opened 2019-01-03
Rename config encryption filter from config-secure.

tomas-langer:


config/encryption/src/main/java/io/helidon/config/encryption/ConfigEncryptionException.java
```diff
@@ -14,20 +14,20 @@
  * limitations under the License.
  */
 
-package io.helidon.config.secure;
+package io.helidon.config.encryption;
 
 /**
  * Secure config related exception.
```
barchetta: Nit: left "secure config" in some comments.

PR #309, opened 2019-01-09
detect and allow the default Accept header sent by HTTPURLConnection

### LINKED ISSUE #211: Failure to parse media type
Caused by: java.lang.IllegalArgumentException: Could not parse '*; q=.2'
	at io.helidon.common.http.MediaType.parse(MediaType.java:253)

Request from another service using Jersey client with default Accept header.

romain-grecourt:
Fixes #211

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/HashRequestHeaders.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/HashRequestHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2018, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


romain-grecourt:
@spericas The initial idea is to hardcode a workaround specific to `HTTPURLConnection` and avoid allowing such bogus headers (`*; q=.2`) in general.

I'd be fine either way, what do others think ? @tomas-langer @ljnelson @barchetta

spericas:
> @spericas The initial idea is to hardcode a workaround specific to `HTTPURLConnection` and avoid allowing such bogus headers (`*; q=.2`) in general.
> 

 Just to be clear, it wouldn't be `*; q=.2` but simply`*` regardless of whether there are params like `q` or not. I just find the addition of the conditional,

```
acceptValues.size() == 1 && HUC_ACCEPT_DEFAULT.equals(acceptValues.get(0))
```

a less elegant and more fragile solution. OTOH, it may be sufficient.

romain-grecourt:
@spericas Yup, sorry I was misleading with "*; q=;.2", I did that to highlight the part of the header sent by HUC that is bogus. I agree that the current solution is not elegant and is fragile, however the main goal is to workaround that specific JDK bug.

PR #315, opened 2019-01-10
Combine webserver and webserver-netty

### LINKED ISSUE #268: combine webserver and webserver-netty
Merge webserver and webserver-netty and remove the webserver SPI.

Before doing so, investigate netty5 and wether we would benefit from having the webserver SPI around in order to support both netty4 and netty5.

romain-grecourt:
Fixes #268


webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2018, 2019 Oracle and/or its affiliates. All rights reserved.
```
tomas-langer: Wrong copyright. Should be `2017, 2019`

webserver/webserver/src/main/java/io/helidon/webserver/netty/NettyWebServer.java
```diff
@@ -55,7 +55,7 @@
 /**
  * The Netty based WebServer implementation.
  */
-class NettyWebServer implements WebServer {
+public class NettyWebServer implements WebServer {
```
tomas-langer: This class should not be public. The NettyWebServer class should be in the same package as the rest (e.g. `io.helidon.webserver`).
According to the development guidelines we should use flat package structure.
Adding a public class makes this part of our API

webserver/webserver/src/main/java/io/helidon/webserver/netty/NettyWebServer.java
```diff
@@ -75,7 +75,16 @@
...
+     * Creates a new instance.
+     *
+     * @param config a server configuration instance
+     * @param routing       a default routing instance
+     * @param namedRoutings the named routings of the configured additional server sockets. If there is no
+     *                      named routing for a given named additional server socket configuration, a default
+     *                      routing is used.
+     */
+    public NettyWebServer(ServerConfiguration config,
                    Routing routing,
```
tomas-langer: This constructor also should not be public. We do not want users to invoke NettyWebServer, we want them to use `WebServer` factory methods or a builder.

webserver/webserver/src/test/java/io/helidon/webserver/utils/SocketHttpClient.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.helidon.webserver.testsupport;
+package io.helidon.webserver.utils;
```
tomas-langer: Why is this package renamed?

romain-grecourt: The package itself not renamed. I moved `io.helidon.webserver.testsupport.SocketHttpClient` to `io.helidon.webserver.utils.SocketHttpClient` to cut a cyclic dependency.

Currently webserver/webserver-netty depends on webserver/test-support, combining webserver-netty into webserver/webserver because some of the test use LoggingTestUtils and SocketHttpClient.

I removed the usage of LoggingTestUtils because that was not really used by the tests but only by some main methods. SocketHttpClient is used only by these tests so I moved it to webserver/webserver/src/test/java under `io.helidon.webserver.utils.SocketHttpClient` in the webserver test sources.

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
romain-grecourt: This is actually valid, but not our preferred form. I will fix that.

webserver/webserver/src/main/java/io/helidon/webserver/netty/NettyWebServer.java
romain-grecourt: Forgot to say "will fix" ;)

tomas-langer:
LGTM

PR #316, opened 2019-01-10
Add health check and metrics to quickstarts

barchetta:


examples/integrations/cdi/datasource-hikaricp-mysql/src/main/java/io/helidon/integrations/examples/datasource/hikaricp/jaxrs/TablesResource.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
```
tomas-langer: This is not correct, should be `2018, 2019` - this is true for all files that were modified

PR #318, opened 2019-01-10
Add health and metrics to quickstart examples.

barchetta:


barchetta:
@tomas-langer this is a cleaned up PR without the rogue files with bad copyrights.

PR #326, opened 2019-01-16
Media processing separated to a top level module.

### LINKED ISSUE #319: Media type support refactoring
As media type support (readers and writers) are to be used both from client and server, we should 
 move the media types support to a root module and move json-p as the first implementation.

Related to #100

tomas-langer:

Resolves #319


tomas-langer:
@romain-grecourt @spericas please have a look at this - quite a big change, not sure whether we can validate correct modularization without client and additional media types...

media/jsonp/server/pom.xml
```diff
@@ -45,13 +44,18 @@
...
-            <artifactId>javax.json</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.projectreactor</groupId>
-            <artifactId>reactor-core</artifactId>
+            <groupId>io.helidon.media.jsonp</groupId>
+            <artifactId>helidon-media-jsonp-common</artifactId>
+            <version>${project.version}</version>
         </dependency>
+        <!--<dependency>-->
```
romain-grecourt: Why leaving these commented out rather than deleted them ?

tomas-langer: It is a leftover when I wanted to try if these dependencies are needed. Forgot to delete them, will do that.

romain-grecourt:
@tomas-langer Not sure what we decided, but I'd say we should merge it now. We have to make some more changes when supporting clients, but at least the workspace is setup and we can add more media type support for the webserver right now.

PR #330, opened 2019-01-18
Fix #329 Use JsonBuilderFactory instead of Json.create

### LINKED ISSUE #329: Use JsonBuilderFactory instead of  Json.create*
## Environment Details
* Helidon Version: 0.11.0
* Helidon SE and MP
----------

## Problem Description
`Json.createObjectBuilder()` and `Json.createArrayBuilder()` are very inefficient because the `Json` factory methods do not cache the Json provider. It is much more efficient to create a `JsonBuilderFactory` once and use it to create Objects and Arrays:

```
private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null);
. . .
JsonObject jo = jsonFactory.createObjectBuilder()
                 .add("name", "Joe")
. . .
```

We need to update all of our examples and documentation to use `JsonBuilderFactory` so that our users don't make this mistake.

barchetta:
This replaces most of our use of `Json.create*` methods with appropriate builder methods. Fixes #329.

Security has some use of `Json.createValue()`. For these I fetch and cache the JsonProvider.

Here is the JSON-P issue describing the problem this PR is a workaround for:  https://github.com/eclipse-ee4j/jsonp/issues/154

### DIFF
--- a/docs/src/main/docs/webserver/08_json-support.adoc
+++ b/docs/src/main/docs/webserver/08_json-support.adoc
@@ -56,7 +56,7 @@ Routing.builder()
 [source,java]
 .Handler that receives and returns JSON objects
 ----
-private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null); <1>
+private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(Collections.emptyMap()); <1>
 
 private void sayHello(ServerRequest req, ServerResponse res, JsonObject json) { // <2>
         JsonObject msg = jsonFactory.createObjectBuilder()   // <3>
--- a/docs/src/main/docs/webserver/08_json-support.adoc
+++ b/docs/src/main/docs/webserver/08_json-support.adoc
@@ -56,16 +56,19 @@ Routing.builder()
 [source,java]
 .Handler that receives and returns JSON objects
 ----
-private void sayHello(ServerRequest req, ServerResponse res, JsonObject json) { // <1>
-        JsonObject msg = Json.createObjectBuilder()   // <2>
+private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null); <1>
+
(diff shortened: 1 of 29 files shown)


barchetta:
@tomas-langer can you take a quick look at the changes to security and see if you are OK with them? Thanks!

docs/src/main/docs/webserver/08_json-support.adoc
```diff
@@ -56,16 +56,19 @@ Routing.builder()
 [source,java]
 .Handler that receives and returns JSON objects
 ----
-private void sayHello(ServerRequest req, ServerResponse res, JsonObject json) { // <1>
-        JsonObject msg = Json.createObjectBuilder()   // <2>
+private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null); <1>
```
tomas-langer: I would rather use an empty map than null.

examples/guides/mp-restful-webservice/src/main/java/io/helidon/guides/mp/restfulwebservice/GreetResource.java
```diff
@@ -53,6 +54,9 @@
 @RequestScoped
 public class GreetResource {
 // end::classDecl[]
+
+    private static final JsonBuilderFactory JSON = Json.createBuilderFactory(null);
```
tomas-langer: As previous - use empty map rather than null.

docs/src/main/docs/webserver/08_json-support.adoc
spericas: Looking at the docs and the implementation, passing `null` or an empty map results in the exact same behavior.

barchetta: Yes, `null` is documented as valid, but I've gone ahead and changed these to `Collections.emptyMap()` since it's a bit more self-documenting.

The JSON-P bug is https://github.com/eclipse-ee4j/jsonp/issues/154

Issue #328, opened 2019-01-17
Move guide src/doc/index.adoc files to top-level README.adoc


tjquinno:
Having the guide content in top-level `README.adoc` files will allow users to find them in a natural location on GitHub. This, combined with #272 and oracle/helidon-build-tools/pull/17, will finish the GitHub-ization of the guides.

Issue #327, opened 2019-01-16
Incorporate include preprocessing into maven lifecycle for examples (incl. guides)


tjquinno:
The guides specifically, and probably the examples in general, should invoke the include preprocessor (#272) so `.adoc` files are GitHub-friendly.

This task is for the work to update the build to use the preprocessing where it makes sense.

See PR #331

PR #331, opened 2019-01-18
Add preinclude check to examples build

tjquinno:
We want `README.adoc` files at the top level of any project to be formatted for proper display on GitHub (which does not currently handle `include::`).

These changes add a check to the examples directory (and submodules within) to make sure that any `README.adoc` at the top level is preprocessed so included text is inserted into the file itself.

There is a new profile `adoc-check` which triggers this check. The configuration of the plug-in limits the checking to `README.adoc` files at the top level of each module.

See issues #327 and #328


examples/pom.xml
```diff
@@ -69,5 +69,42 @@
                 <module>integrations</module>
             </modules>
         </profile>
+        <profile>
+            <!--
+                Pipeline builds use this profile with the generate-sources phase to convert
```
romain-grecourt: How about we move this text under the README.md inside the /examples directory ?

tjquinno:
Yes. They are in a later commit.

tjquinno:
Recent build failures are due to the change to the plug-in version. Once 1.0.5 is released the pipeline for these changes should be OK again.

examples/pom.xml
tjquinno: Done.

PR #338, opened 2019-01-23
remove incorrect README.md for examples/guides/se-restful-webservice

romain-grecourt:


romain-grecourt:
Addresses one of the issue reported in #338

PR #358, opened 2019-01-31
Use variable to hold GreetService instance -- easier for the guide

tjquinno:
The guide adds an app-specific health check, and the code it adds to do so requires that the same instance of `GreetService` be plugged into the health apparatus as is registered with the `/greet` path. 

The original quick-start code instantiated `GreetService` in-line as an argument to `.register`, and having the quick-start code use a variable makes it easier in the guide.

### DIFF
--- a/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/Main.java
+++ b/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/Main.java
@@ -95,6 +95,7 @@ public final class Main {
      */
     private static Routing createRouting(Config config) {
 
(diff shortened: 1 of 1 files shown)


tomas-langer:
Please make sure the metrics work as expected. I remember I had some issues when the metrics registry was used before the MetricSupport was created.

tjquinno:
@tomas-langer I am working my way through the guide changes to keep it in step with the coding changes over the past few weeks. I have not gotten to trying the metrics part yet.

So is your conjecture that if we instantiate `GreetService` earlier we might also need to invoke `MetricsSupport.create` before instantiating `GreetService`?

tjquinno:
I created this issue #360 to see if there is a way to relax the ordering requirement.

Issue #214, opened 2018-11-29
Update guides to reflect updates to quick start examples


tjquinno:
Tomas is updating the quickstart examples. These form the basis for the SE and MP RESTful web service guides.

This task is for reflecting the changes he is making to the quickstart examples in the guide code and documents.

Issue #176, opened 2018-11-08
Update SE and MP guides to adopt new review feedback


tjquinno:
Romain and Joe supplied some good feedback about the first guide. This task incorporates that feedback into the first guide.

This is important to do now so future guides can use the SE RESTful web app one as a pattern.

Here is my list of the topics we covered:

1. Change guides to use the Helidon archetype.
2. Because #1 generates a lot of Java code, change the guide from largely adding code to largely explaining the code that is generated. Some code (e.g., metrics and health) will still be added.
3. Link to or include pre-req information from the separate, common file. 
   a. Maybe a simple approach is for the common AsciiDoc attribute assignment file to assign attributes containing pre-req versions so the guide .adoc files can refer to those.
   b. Another approach would be for the existing pre-reqs info in the Getting Started file to be tagged so individual guides can include only the pre-reqs they need.
4. Add default attribute settings in the guide so GitHub rendering works correctly. (Will evolve over time as the common AsciiDoc attribute assignment file appears.)
5. Change ReSTful to RESTful. (done as part of #215)
6. Migrate the guide files from src/main/docs/index.adoc to top-level README.adoc. Part of this item is to make sure the README.md renders properly, esp. regarding includes.
7. Break long command lines explicitly to avoid horizontal scrolling of code blocks on small screens. (done in #215)
8. See about inlining the startServer() code block.
9. Change output code blocks top reformatted (so suppress the built-in copy feature on our site). (done in #215)
10. Specify the exact git clone command in the last section for cloning the repo.
11. Fix list problem after the HTTP method table. (done in #168)
12. See Joe’s example from Getting Started to narrow tables where possible.
13. See if there is a way to strip indentation from included code snippets.

tjquinno:
@romain-grecourt @barchetta FYI some bad news...It turns out that GitHub does not support the AsciiDoc `include` directive.

https://github.com/github/markup/issues/1095

The issue makes a reference to perhaps being available "next year" (2019).

romain-grecourt:
@tjquinno  That's fine with me :D - I actually would prefer linking the pre-req ; it does not make the guide self contained but makes it lighter.

It goes against the rule of "links are bad", but we could be smart and have a little section at the top like "before you begin" where we list the things that must be done before following a guide.

tjquinno:
[...]

PR #364, opened 2019-02-01
Update SE restful web service guide and code

tjquinno:
A number of API and other changes have altered the quick-start code on which the SE restful web service guide is based.

This PR updates the Java code accordingly and also changes the guide so it:
- [x] uses the Helidon Maven archetype to generate the project
- [x] acts mostly as a guided tour of the archetype-generated code, rather than a sequence of copy-and-paste steps by which the reader would construct the app.

The last sections of the guide, as before, show how to modify the app to add an app-specific health check and an app-specific metric.

See issues #214 and #176

### DIFF
--- a/examples/guides/mp-restful-webservice/README.adoc
+++ b/examples/guides/mp-restful-webservice/README.adoc
@@ -171,8 +171,6 @@ For this example add these dependencies:
 If you run your project from the IDE, the IDE typically handles the main class and places
 dependent JARs on the runtime classpath for you and your pom is now ready to go.
 
-// _include::0--1:{se-guide-adoc}[tag=runMavenOutsideIDE]
-
 === Create an MP config file
 This file contains settings for the Helidon web server and the
 application. Note that the MP Config specification says that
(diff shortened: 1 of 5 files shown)


tjquinno:
Note: In the initial creation of the PR I _did not_ do the required preprocessing of the `README.adoc` file to pre-include all included content. The automatically-run wercker/build job will fail until I change this, but reviewing the changes should be far easier with the natural-style `include::` directives rather than the preprocessed form.

tjquinno:
I've included the preprocessed version of `README.adoc` because I think that will actually make comparing the edits easier for reviewers. (It also clears the pipeline build error that resulted from using the natural-form instead of the preprocessed-form file.)

tjquinno:
Note to reviewers: Please don't spend time on the MP `README.adoc` that is included in this change set. Soon I'll open a separate PR for the MP changes which will be much more extensive than what's shown here. I updated the MP guide as part of _this_ PR so the pipeline build test for the MP guide would pass (because it includes significant portions from the SE guide).

tjquinno:
Previous commit and push reflect comments about the SE tutorial recorded in PR #367.

Issue #380, opened 2019-02-07
Rename the "Guides" node on the helidon.io docs site to "Tutorials"


kumar-dhanagopal:
Per the discussion with the team in the sync-up meeting on Feb 7.

kumar-dhanagopal:
We discussed this in the US Dev team meeting on 2/8.  The consensus was to stick with "Guides"

PR #381, opened 2019-02-07
Renamed the "Guides" node to "Tutorials"

kumar-dhanagopal:
See #380


romain-grecourt:
@kumar-dhanagopal There is more changes needed to rename guides to tutorials (docs/src/main/docs/guides, examples/guides/). Let's discuss the terminology first in an issue.

Also note that openliberty uses the guides term in the same way we do currently.

PR #384, opened 2019-02-08
Gathering all resources from CDI when application is not defined.

### LINKED ISSUE #363: MP server CDI extension might discard non-ApplicationScoped root resource classes
See:

https://github.com/oracle/helidon/blob/5edc48edb60a3eb3f6402984533d4633029c82bc/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java#L47

I believe that will rule out otherwise valid `Application`-less root resource classes, right?

Consider a root resource class annotated with `@RequestScoped`, not `@ApplicationScoped`.  It should be picked up, right?  But because it doesn't have `@ApplicationScoped` on it, this extension will ignore it.

tomas-langer:
Resolves #363 

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -240,10 +240,13 @@ public interface Server {
 
             if (applications.isEmpty()) {
                 List<Class<?>> resourceClasses = extension.getResourceClasses();
+
                 if (resourceClasses.isEmpty()) {
                     LOGGER.warning("Failed to find JAX-RS resource to use");
                 }
                 resourceConfig = configForResourceClasses(resourceClasses);
+
+                extension.getProviderClasses().forEach(resourceConfig::register);
             } else {
                 applications.forEach(this::addApplication);
             }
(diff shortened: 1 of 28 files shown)


tomas-langer:
This pull requests should add the resources to synthetic application as expected. 
I have also added integration tests (not run by default - use "itg-tests" profile) that tests the possible combinations:

- There is an `@ApplicationScoped` Application class; resources are not annotated with bean defining annotation
- There is Application class that is not annotated (and is ignored), there are resources and a provider in `@RequestScoped` and `@ApplicationScoped` scopes and all are  picked up
- There is a combination of annotated Application and resource classes - only the application is used

romain-grecourt:
The PR needs to be rebased from master to pickup the fix for the copyright check.
@tomas-langer Is this supposed to go in for 1.0 ?

tomas-langer:
@romain-grecourt I leave that to Santiago & Laird. You can pick my work up if you like it and do the rebase (and possibly merge).
Unfortunately I leave for vacation tomorrow

spericas:
After discussing this with @ljnelson, we agreed that the corresponding issue needs a more detailed analysis and that this PR may no cover all cases correctly. Discussion will be postpone until after 1.0.

PR #388, opened 2019-02-08
Yasson support

ljnelson:
This PR creates Yasson support, following conventions established by Helidon's preëxisting JSON-P support.


ljnelson:
This PR is now ready for review.  Note that I've followed the conventions and approaches in our JSONP support and in my Jackson support PR (which was approved and merged).

media/jsonb/server/src/main/java/io/helidon/media/jsonb/server/JsonBindingSupport.java
```diff
@@ -0,0 +1,117 @@
...
+     * @param jsonbProvider a {@link BiFunction} that returns a {@link
+     * Jsonb} when given a {@link ServerRequest} and a {@link
+     * ServerResponse}; must not be {@code null}
+     *
+     * @return a new {@link JsonBindingSupport}
+     *
+     * @exception NullPointerException if {@code jsonbProvider} is
+     * {@code null}
+     */
+    public static JsonBindingSupport create(final BiFunction<? super ServerRequest,
```
tomas-langer: Is this really needed as the only method? If there is a possibity to use a single Jsonb instance created as a default instance, please add also method create() with no parameters.

tomas-langer: Also why is this request/response dependent? I do not know JSON-B (or Jackson), so please just explain why it makes sense to do this, rather than just providing an instance.

ljnelson: I'll add `create()` as well.  Normally I don't like `static` instances but it sounds like that's OK so I'll follow the JSON-P support conventions here.

Regarding request/response-based instance creation, often non-trivial applications using Jackson (I don't use Yasson, but I assume it's similar) need different mapping setups for differing JAX-RS resource classes—it is common to create an `ObjectMapper` for certain collections of resources.  Here, the analogous thing to do would be to allow the caller to decide whether a new `Jsonb` is called for (or a cached one) for a particular request/response pair.

ljnelson:
@tomas-langer Good to go?

PR #391, opened 2019-02-11
Upgrading to Hystrix 1.5.18

spericas:
Upgrading to Hystrix 1.5.18. New strategy to support circuit breaker state transitions as defined by the FT spec. Most of the Hystrix circuit breaker support is turned off and state transitions are implemented manually using Hystrix dynamic properties. The old strategy was difficult to extend with the new version of Hystrix and the new implementation is simpler to understand.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CircuitBreakerHelper.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CircuitBreakerHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,9 +18,10 @@ package io.helidon.microprofile.faulttolerance;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Logger;
 
-import com.netflix.hystrix.HystrixCommand;
+import com.netflix.config.ConfigurationManager;
 import org.eclipse.microprofile.faulttolerance.CircuitBreaker;
 
 /**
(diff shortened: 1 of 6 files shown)


microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CircuitBreakerHelper.java
```diff
@@ -252,27 +282,24 @@ long getInStateNanos(State queryState) {
...
-                delayTotal += delayPeriod;
-            }
-            if (delayTotal >= MAX_DELAY_CIRCUIT_BREAKER_OPEN) {
-                throw new InternalError("Inconsistent state for circuit breaker in " + command.getCommandKey());
-            }
+    private void openBreaker() {
+        if (!command.isCircuitBreakerOpen()) {
+            ConfigurationManager.getConfigInstance().setProperty(
+                    String.format(FORCE_OPEN, command.getCommandKey()), "true");
+        }
```
tjquinno: The revised `closeBreaker` method below sets `FORCE_OPEN` to false as well as `FORCE_CLOSED` to true.

Should this `openBreaker` method set `FORCE_CLOSED` to false as well as `FORCE_OPEN` to true?

I'm not very familiar with Hystrix so maybe I'm missing something.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandRetrier.java
```diff
@@ -273,4 +275,16 @@ private void updateMetricsAfter(Throwable cause) {
...
+    /**
+     * Hystrix throws a {@code RuntimeException}, so we need to check
+     * the message to determine if it is a breaker exception.
+     *
+     * @param t Throwable to check.
+     * @return Outcome of test.
+     */
+    private static boolean isHystrixBreakerException(Throwable t) {
+        return t instanceof RuntimeException && t.getMessage().contains("Hystrix "
+                + "circuit short-circuited and is OPEN");
```
tjquinno: Testing the contents of the exception message seems a bit fragile but there's probably no alternative. 

Is there an easy way to add a unit test that would fail if that message text were changed in some future release of Hystrix?

spericas: Yes, this is quite ugly. Hystrix is no longer being developed and we're likely going to move away from it. Also, if that message changes, we should see plenty of unit and TCK failures :)

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CircuitBreakerHelper.java
spericas: It's not necessary because if both are defined, FORCE_OPEN takes precedence according to their documentation. See:

https://github.com/Netflix/Hystrix/wiki/Configuration#circuitBreaker.forceOpen

PR #399, opened 2019-02-12
Change quickstart examples so that PUT uses a json body and not path param

barchetta:
The quickstart examples (also used by the maven archetypes) support changing of the greeting via a path param. This does not follow REST best practices.

This PR changes the examples to that you PUT json to change the greeting. Specifically:

```
 curl -X PUT -H "Content-Type: application/json" -d '{"greeting" : "Howdy"}' http://localhost:8080/greet/greeting
```

### DIFF
--- a/examples/quickstarts/helidon-quickstart-mp/src/main/java/io/helidon/examples/quickstart/mp/GreetResource.java
+++ b/examples/quickstarts/helidon-quickstart-mp/src/main/java/io/helidon/examples/quickstart/mp/GreetResource.java
@@ -108,16 +108,16 @@ public class GreetResource {
     public Response updateGreeting(JsonObject jsonObject) {
 
         if (!jsonObject.containsKey("greeting")) {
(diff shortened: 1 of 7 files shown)


examples/quickstarts/helidon-quickstart-mp/src/main/java/io/helidon/examples/quickstart/mp/GreetResource.java
```diff
@@ -96,20 +97,27 @@ public JsonObject getMessage(@PathParam("name") String name) {
     /**
      * Set the greeting to use in future messages.
      *
-     * @param newGreeting the new greeting message
+     * @param jsonObject JSON containing the new greeting
      * @return {@link JsonObject}
```
tjquinno: This java doc `@return` should change to reflect the new return type.

barchetta: Done

PR #425, opened 2019-02-20
Add functional tests

### LINKED ISSUE #409: Integration tests
Inspired by samples to test things the same way as users write applications.

barchetta:
Fixes #409


tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
```diff
@@ -0,0 +1,98 @@
...
+        }
+        bookStore.store(book);
+        return Response.ok().build();
+    }
+
+    @GET
+    @Path("{isbn}")
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getBook(@PathParam("isbn") String isbn) {
+        Optional<Book> optional = bookStore.find(isbn);
```
tomas-langer: Functional style fits better:
```java
return bookStore.find(isbn)
                .map(book -> Response.ok(book).build())
                .orElseGet(() -> Response.status(Response.Status.NOT_FOUND).build());
```

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @DELETE
+    @Path("{isbn}")
+    public Response deleteBook(@PathParam("isbn") String isbn) {
+        Optional<Book> optional = bookStore.find(isbn);
+        if (!optional.isPresent()) {
+            return Response.status(Response.Status.NOT_FOUND).build();
+        }
+        bookStore.remove(isbn);
+        return Response.ok().build();
```
tomas-langer: Functional style fits better with Optional:
```java
return bookStore.find(isbn)
                .map(book -> {
                    bookStore.remove(book);
                    return Response.ok().build();            
                })
                .orElseGet(() -> Response.status(Response.Status.NOT_FOUND).build());
```

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
```diff
@@ -0,0 +1,98 @@
...
+    @PUT
+    @Path("{isbn}")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public Response putBook(Book book) {
+        Optional<Book> optional = bookStore.find(book.getIsbn());
+        if (!optional.isPresent()) {
+            return Response.status(Response.Status.NOT_FOUND).build();
+        }
+        bookStore.store(book);
+        return Response.ok().build();
```
tomas-langer:
```
return bookStore.find(book.getIsbn())
                .map(found -> {
                    bookStore.store(book);
                    return Response.ok().build();
                })
                .orElseGet(() -> Response.status(Response.Status.NOT_FOUND).build());
```

tests/apps/bookstore/bookstore-se/src/main/java/io/helidon/tests/apps/bookstore/se/AuthorMapper.java
```diff
@@ -0,0 +1,91 @@
...
+class AuthorMapper {
+
+    private static final String FIRST = "first";
+    private static final String LAST = "last";
+
+    /*
+     * Use JsonBuilderFactory (and not the Json factory) to create builders.
+     * This lets us cache the builder factory and improves performance dramatically.
+     */
+    private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null);
```
tomas-langer: Incorrect name - constants should be all capital `JSON_FACTORY`

tests/apps/bookstore/bookstore-se/src/main/java/io/helidon/tests/apps/bookstore/se/BookMapper.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final String CATEGORY = "category";
+    private static final String PUBLISHER = "publisher";
+    private static final String COPYRIGHT = "copyright";
+    private static final String AUTHORS = "authors";
+
+    /*
+     * Use JsonBuilderFactory (and not the Json factory) to create builders.
+     * This lets us cache the builder factory and improves performance dramatically.
+     */
+    private static final JsonBuilderFactory jsonFactory = Json.createBuilderFactory(null);
```
tomas-langer: Should be `JSON_FACTORY`

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
tomas-langer: This could be made even nicer as follows:
declare NOT_FOUND as a constant, as it is reused multiple times:
`private static final Response NOT_FOUND = Response.status(Response.Status.NOT_FOUND).build();`

create a method for ok:
```java
private Response ok(Book book) {
        return Response.ok(book).build();
    }
```

change the code:
```
return bookStore.find(isbn)
                .map(this::ok)
                .orElse(NOT_FOUND);
```

tomas-langer:
Most of my comments are about functional style when using optional.
At least you should not start with the negative case, e.g. `if (!optional.isPresent) {...` is not very readable.

tests/apps/bookstore/bookstore-se/src/main/java/io/helidon/tests/apps/bookstore/se/BookMapper.java
barchetta: Will fix.

tests/apps/bookstore/bookstore-se/src/main/java/io/helidon/tests/apps/bookstore/se/AuthorMapper.java
barchetta: Will fix.

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
spericas: I don't agree that this is necessarily more readable; it is a matter of style. Also using lambdas to avoid flow control statements does result in _more objects being created_ (a problem that is apparent in our profiles!). In the example above, in particular, the second lambda has an implicit reference to "this" which would require a new instance of it on every call to the resource method.

barchetta: Having the lambda to `map` do more than just transform the Optional value feels wrong to me.

barchetta:
I've fixed the JSON_FACTORY naming.

As to the functional programming style comments. I understand the feedback, but the existing code is simple, readable and works.  I know in our project we're going to have folks that lean towards imperative and folks that lean toward functional. In some cases that difference will be important. In this case I don't think it is.  Unless there is strong object, I'd prefer to leave the code as-is and get the PR merged.

etc/copyright-exclude.txt
```diff
@@ -1,3 +1,4 @@
+.iml
```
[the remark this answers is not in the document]
tomas-langer: Can you please explain why you keep this in?

tomas-langer:
For Helidon SE, you cannot lean towards imperative, as you would fight with every single API of Helidon - it is designed to be functional.
So I think our examples should follow case.
If you feel that using `map` the way I do is not nice, you can still use `ifPresentOrElse`. 

Also if you want to use imperative, why don't you just return `null` from `find` method? If you care about performance on the level of object creation, you should not create an instance of `Optional` as well...

etc/copyright-exclude.txt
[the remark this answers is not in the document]
barchetta: I run copyright locally. I have .iml files in my local repository. I don't want the copyright check failing when I run it locally.  Having .iml listed in copyright-exclude.txt causes no problems.

barchetta:
> For Helidon SE, you cannot lean towards imperative, as you would fight with every single API of Helidon - it is designed to be functional.

This is JAX-RS code. It's in a Helidon MP app. It's not fighting with anything.  The code is simple and correct.

Are you requiring the changes to approve this PR?

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
barchetta: Optional is now gone.

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
barchetta: Optional is now gone.

barchetta:
Use of Optional has been removed from Bookstore

tests/apps/bookstore/bookstore-mp/src/main/java/io/helidon/tests/apps/bookstore/mp/BookResource.java
barchetta: Use of Optional has been removed.

spericas:
@barchetta Minor comment. I don't recall us using "! expr" instead of "!expr" in other parts of Helidon

PR #435, opened 2019-02-23
Update health documentation and add a page for k8s probes.

romain-grecourt:
Update the health documentation for Helidon SE.
Add a new page to document usages of the health API for Kubernetes.

PR #448, opened 2019-02-25
Getting started updates

romain-grecourt:
- Combine `/about` and `/getting-started` into `/about`
- Move the quick-start example page into the guides as 2 separate guides, one for SE and one for MP
- Split the docker guide in 2 guides, one for Docker and one for Jib.
- Update overview pages and links accordingly
- Update cross ref links
- Removed the "What you will learn" section in the guides and replaced it with content in the preambules
- Removed examples/guides
- Updated content for Docker guide and Jib guide


docs/src/main/docs/guides/04_dockerfile.adoc
```diff
@@ -0,0 +1,216 @@
...
+NOTE: ~140MB is less than the pre-built OpenJDK slim JRE images. Results might
+ differ a bit depending on your platform.
+
+=== What about Helidon MP?
+
+For Helidon MP you need to add a couple more modules to the `jlink` command:
+
+[source,yaml,subs="verbatim,attributes"]
+----
+RUN ["jlink", "--compress=2", "--strip-debug", "--no-header-files" \
```
barchetta: Type: There is a comma missing at the end of this line.

romain-grecourt: roger that, will fix.

tjquinno:
The group had talked about and I thought agreed on combining the existing quick-starts and guides _and_ at least considering moving the current detailed walk-through for SE and MP into separate, tutorial-style, documents.

This PR does not convert the current detailed guide information into tutorials. Is that planned for later or are we not going to do that?

romain-grecourt:
Yes, this is a little different than what was discussed. The main goal of this PR is to move the quickstarts as guides, see it rendered here: https://romain-grecourt.github.io/junkyard/getting-started/ (note the new about section and the guides section)

- I'm going to improve the docs for metrics (TODO) and health (pr #435) to fill-in some of the gap addressed by the current guides.
- The tour part is not addressed by this PR, we have a few options:
  1. add 2 separate guides
  2. update the quickstart guides with the tour informations since they are now guides
  3. add the tour information in the quickstart projects directly, in the README and in comments

I'm still convinced that option 3 is the best, however this is a team decision.

tjquinno:
Got it. I understand that we need to divide the work into manageable chunks. One concern I have is that by leaving the guided tour part until later we will have to go back and retrieve the guide information we want from the earlier, tutorial versions of the guides and either recreate the guides the way we want (you listed as option i), plug that info in some form back into the new quick-start `.adoc` files (option ii), or convert that info into comments and put it in the generated code as comments or the generated `README` (option iii). 

I guess I'm less concerned at this point which option we adopt and more concerned about what seems like additional work fetching the information from soon-to-be-deleted files, but maybe there's no way to avoid it without doing it all at once.

PR #457, opened 2019-02-26
MetricSupport and RegistryFactory can be accessed in any order now.

### LINKED ISSUE #360: MetricsSupport must be created before a service accesses the registry
## Environment Details
* Helidon Version: 0.11.0
* Helidon SE
* JDK version: 9.0.4
* OS: Mac OS High Sierra

----------

## Problem Description
The developer must invoke `MetricsSupport.create` before a metrics-enabled service tries to access the metrics registry or an NPE results.

## Steps to reproduce
In the SE restful web service guide code (helidon/examples/guides/se-restful-webservice) find the `Main.createRouting` method. As written, it first creates the `MetricsSupport` object and then instantiates `GreetService`. Reverse the order. Then try to `mvn package` the project. The test will fail with an NPE. Restore the original order and the test passes.

Ideally, this ordering would not be important. This issue is for investigating whether there is a way to remove the ordering requirement.

tomas-langer:

Resolves #360

### DIFF
--- a/metrics/metrics/src/main/java/io/helidon/metrics/RegistryFactory.java
+++ b/metrics/metrics/src/main/java/io/helidon/metrics/RegistryFactory.java
@@ -44,7 +44,7 @@ import org.eclipse.microprofile.metrics.MetricRegistry.Type;
 // this class is not immutable, as we may need to update registries with configuration post creation
 // see Github issue #360
 public final class RegistryFactory {
-    private static final RegistryFactory staticInstance = create();
+    private static final RegistryFactory INSTANCE = create();
 
     private final EnumMap<Type, Registry> registries = new EnumMap<>(Type.class);
     private final EnumMap<Type, Registry> publicRegistries = new EnumMap<>(Type.class);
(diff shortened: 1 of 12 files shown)


metrics/metrics/src/main/java/io/helidon/metrics/RegistryFactory.java
```diff
@@ -74,20 +80,61 @@ static RegistryFactory create() {
      * @return supplier of registry factory (to bind as late as possible)
      */
     public static Supplier<RegistryFactory> getRegistryFactory() {
-        return () -> staticInstance;
+        return () -> {
```
tjquinno: Could this simply invoke `getInstance()` for the caller, rather than throw an exception? 

Doing so might mask the case where the developer intended to use config but forgot to. But if the developer had invoked `getInstance()` explicitly instead of `getInstance(Config)` s/he would have to diagnose the same sort of problem.

tomas-langer: Refactored a bit more - now we mostly use a singleton instance, with support for customized RegistryFactory (and MetricSupport) - if for any reason somebody wants more than one RegistryFactory in their application accessible on different endpoints

tomas-langer:
Additional commit fixing spotbugs issue (changes in singleton made a few null checks obsolete.

PR #459, opened 2019-02-27
Avoid using Zipkin v1 in example config

shakuzen:
While Zipkin v1 is still available, we do not want to encourage new users to use it. Some users may copy/paste and edit the example config given, so this changes it to Zipkin v2 version/path.

### DIFF
--- a/docs/src/main/docs/tracing/02_zipkin.adoc
+++ b/docs/src/main/docs/tracing/02_zipkin.adoc
@@ -70,9 +70,9 @@ tracing:
     protocol: "https"
     host: "192.168.1.1"
     port: 9987
-    api-version: 1
-    # this is the default path for API version 1
-    path: "/api/v1/spans"
+    api-version: 2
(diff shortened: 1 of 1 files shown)


romain-grecourt:
@shakuzen You will to sign the OCA for us to accept your contribution. See https://www.oracle.com/technetwork/community/oca-486395.html

shakuzen:
@romain-grecourt I was hoping there would be a trivial change clause that allowed this to not need an OCA. I don't have access to a printer normally, so it may be a while before I could sign the OCA. Therefore I think it would be easiest if we close this pull request and and a committer makes the same change to avoid delay.

romain-grecourt:
This process is from another era ;)
We need a release to update the docs and the next one will be done in a couple of weeks.
I'll make this change on your behalf if you did not sign the OCA at the time of the next release ;)

romain-grecourt:
Closing this PR, this change will  be made as part of PR #531

PR #463, opened 2019-02-28
Some code cleanup and optimizations when writing data from Jersey

spericas:
More details:

- Improve code to copy write buffers coming from Jersey
- Let Netty re-schedule writes on the correct NIO thread
- Do not use latches to synchronize Jersey and Netty threads

Although all tests are passing we should keep an eye for any regressions resulting from these changes. Note that all these changes are exclusive to Jersey and Helidon MP.


spericas:
@barchetta Completed some manual testing with payloads over 5 MB without seeing any problems. Once the functional tests are in place, we should used those to further stress test these changes.

PR #465, opened 2019-03-03
JWT-Auth can be configured not to be secure by default

### LINKED ISSUE #455: Clarify MP security spec and update Helidon impl
## Environment Details
* Helidon Version: 1.0.0
* Helidon MP
* JDK version: 9.0.4
* OS: n/a

----------
There are ambiguities in the MP JWT-Auth spec. Tim discovered that the code generated by the MP starter works with other impls but Helidon MP rejects attempts to access the endpoints.

Specifically, with the application annotated with `@LoginConfig(authMethod = "MP-JWT")` Helidon MP reject access to endpoints that are _not_ annotated with `@RolesAllowed` and no suitable JWT in the request header, essentially acting as if each endpoint is annotated with `@DenyAll`.

In contrast, other MP impls permit such access, acting as if each endpoint is annotated with `@PermitAll`.

The spec does not state clearly what the behavior should be in this case.

@tomas-langer is going to pursue this on the MP mailing list.

tomas-langer:
only requires authentication for endpoints annotated with RolesAllowed.

resolves #455

### DIFF
--- a/microprofile/jwt-auth/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthAnnotationAnalyzer.java
+++ b/microprofile/jwt-auth/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthAnnotationAnalyzer.java
@@ -15,6 +15,11 @@
  */
 package io.helidon.microprofile.jwt.auth;
 
+import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.Method;
+
+import javax.annotation.security.RolesAllowed;
+
 import io.helidon.config.Config;
 import io.helidon.security.providers.common.spi.AnnotationAnalyzer;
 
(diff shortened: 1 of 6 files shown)


tomas-langer:
Controlled by property
`security.jersey.analyzers.jwt.secure-by-default`
Set to `false` not to require authentication for endpoints that are not annotated by `RolesAllowed`

microprofile/jwt-auth/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthAnnotationAnalyzer.java
```diff
@@ -50,25 +56,89 @@ public void init(Config config) {
...
         }
 
+        builder.register(new RegisterMpJwt(false));
         return builder.build();
     }
 
+    // resource class analysis
+    @Override
+    public AnalyzerResponse analyze(Class<?> maybeAnnotated, AnalyzerResponse previousResponse) {
+        if (isMpJwt(previousResponse) && isRolesAllowed(maybeAnnotated)) {
```
tjquinno: Minor suggestion - Refactor this code and the code in the other variant of `analyze` into a separate private method to avoid duplicating it. I realize it's not a lot but it's about the same length as `isMpJwt` which is already in its own method.

tomas-langer: Refactored.

Issue #487, opened 2019-03-08
ClientSecurityFeature.PROPERTY_CONTEXT settable from WebTarget/Client configuration


martin-sladecek:
Currently, the ClientSecurityFeature.PROPERTY_CONTEXT can be set only through request().property().
In certain cases, when the security context is created manually and can be used repeatedly for multiple calls, it might be better to set the property in Client/WebTarget.

tomas-langer:
PR is merged

Issue #471, opened 2019-03-06
Jersey client: propagate SecurityContext automatically (in synchronous jersey calls)


martin-sladecek:
When the jersey client is used inside a request scope (i.e. in synchronous REST call to jersey server), the SecurityContext should be automatically propagated to Jersey clients, instead of using ClientSecurityFeature.PROPERTY_CONTEXT property.

tomas-langer:
We should use similar approach as with propagation of Tracing - `AutoDiscoverable` client filter that uses ThreadLocal to propagate the information.

tomas-langer:
depends on #596

tomas-langer:
As we have now the common `Context`, we can manage this even outside of synchronous calls (as long as the calls are executed in executor services wrapped with `Contexts.wrap`).
Also as part of this implementation, we should support the case when `SecurityContext` can be configured application wide - both WebServer and MP Server now have methods to configure the "top-leve" `Context`. If `SecurityContext` is registered in such a `Context`, it would be automatically available for outgoing requests as long as they happen within a scope of the context.
It should also be possible to achieve this using `Contexts.runInContext()`.
The current features (providing `SecurityContext` through JAX-RS property will be kept intact and will override values in `Context`

PR #488, opened 2019-03-08
ClientSecurityFeature.PROPERTY_CONTEXT settable from WebTarget/Client (#487)

martin-sladecek:
Fix for #487. There's a similar method in ClientTracingFilter, but I don't know what should be the good place for shared jersey code, seems like there's no such module yet.


tomas-langer:
Security context can only be set on a request - it should not be set on target or on client.
Security context is "request scoped" - e.g. it is different for each request. 
Configuring it on client or webtarget may cause unexpected behavior - e.g. propagating security of a different user than the one currently using the service.
Also this problem should be mitigated by #471

martin-sladecek:
#471 will not solve my problem with custom security context. In my use-case, there is no request context and I want to call every outbound call with the same user. This can be a case for a k8s job or message queue consumer.
Since my security context is always the same for me, it would be very convenient to actually set it to my Client/WebTarget instance in some CDI producer, so I don't need to set security context everywhere.

Issue #446, opened 2019-02-25
SE config should map env vars the same way that MP does


batsatt:
## Environment Details
* Helidon Version: 1.0.1-SNAPSHOT
* Helidon SE

----------

## Problem Description

MP config does a name mapping for env vars that SE does not:
```
MpcSourceEnvironmentVariables() {
        System.getenv().forEach((env, value) -> {
            if (env.contains("_") && !env.contains(".") && ENV_VAR.matcher(env).matches()) {
                // allow either MY_ENV_VAR or MY.ENV.VAR (only for keys that DO NOT contain a dot)
                String alternativeKey = env.replace('_', '.');
                envVariables.put(alternativeKey, value);

                // and allow my.env.var:
                alternativeKey = alternativeKey.toLowerCase();
                envVariables.put(alternativeKey, value);
            }
            envVariables.put(env, value);
        });
    }
```
This mapping is extremely useful for config overrides (e.g. `SERVER_PORT --> server.port`) in bash, which does support dotted env var names.

SE config should perform the identical mapping.

batsatt:
Hmm. Even if we fix this, it turns out that some config options use hyphenated names (e.g. `server.executor-service.max-pool-size`) which are also problematic in bash env var names. 

Also, the MP config mapping doesn't even kick in here, since the pattern doesn't match hyphenated names:
```
ENV_VAR = Pattern.compile("([A-Za-z0-9]+_?)*")
```
One important reason to support an env var mapping is for the Docker use case, where it is typically easier to specify env vars than system properties, but... unless *all* config can be overridden with them, it makes this whole issue moot.

batsatt:
To support mapping to hyphenated names, I propose that env var names containing `_dash_` literals first be  replaced with `-`, followed by the existing mappings. Given an env var named:
```
SERVER_EXECUTOR_dash_SERVICE_MAX_dash_POOL_dash_SIZE
```
the following two variants would be produced:
```
SERVER.EXECUTOR-SERVICE.MAX-POOL-SIZE
server.executor-service.max-pool-size
```

Issue #507, opened 2019-03-14
SE config should prioritize system properties and env vars as does MP


batsatt:
## Environment Details
* Helidon Version: 1.0.0-SNAPSHOT
* Helidon SE

----------

## Problem Description

MP prioritizes system properties (ordinal 400) over environment variables (ordinal 300), while SE does the reverse, so produces _different_ merged results when a given key is present in both.

tomas-langer:
This should now be fixed, as Helidon SE Config implements MP Config as well.
Fixed for 2.0 only, as this is a backward incompatible chane.

PR #509, opened 2019-03-15
Make SE and MP share config env var mapping

batsatt:
Introduces mapping for hyphenated properties.


batsatt:
Fixes #446.
See related #507.

tjquinno:
@tomas-langer I find that Diff Settings -> Hide Whitespace Changes helps to filter many formatting changes out which simplifies reviewing the substantive changes. Obviously it will not suppress _all_ types of formatting changes but it helps a lot.

batsatt:
@tomas-langer I hear you on the whitespace, but... do we not want to be more consistent with it? In this case, the primary change is aligning chained method calls on the dot; everyone here in the US agreed that we should do so, but you may disagree.

We really need to have a more complete checkstyle and associated IDE style settings.

batsatt:
@tjquinno The `_dash_` mapping is just a way to support any hyphenated key like`a-b`, not specifically about supporting both `a-b` and `a.b`. You're right though, that together with the other mappings both are supported.

batsatt:
@tomas-langer Indent changes reverted.

batsatt:
@tjquinno Also, you are correct that there is a risk, hopefully very small, that a user both actually has a config key containing the `"_dash_"` substring and wants to set it in an environment variable.

I did think about adding a mechanism to disable this mapping for that case (e.g. a `"HELIDON_DASH_MAPPING"` env var that can be set to `false`, or maybe some meta config), but it struck me as a step too far. I could easily be convinced to add it though...

tjquinno:
My opinion is that adding a way to disable that mapping is indeed going too far. Going further down the path of "configuration which controls how configuration works" feels to me too close to an Escher art print!

tomas-langer:
Regarding the whitespace - we need to be consistent overall, or not reformat existing code at all.
What happens now is that you reformat and commit, then somebody else does a change in the same class, reformats and commits.
This is useless. 

My approach - use the same formatting (as long as you use IntelliJ, we have a codestyle xml to use). 
Never reformat existing code - only modified and only if the original formatting is bad (or you have added code that requires a line break etc.).
For those without IntelliJ this requires configuration of their IDE to follow the same rules (unfortunately not everything gets caught by checkstyle).

batsatt:
Yes, I believe we need both a more thorough checkstyle _and_ a matching IntelliJ codestyle xml,  preferably maintained in the source tree. This would all be easier if we could convince everyone to switch to IntelliJ; maybe we should start there ;)

PR #512, opened 2019-03-17
Don't skip the handling and forwarding of 404 responses with content …

Holubow:
…to downstream WebServer filters (oracle#390)

* Only skip the handling and forwarding of 404 responses in io.helidon.webserver.jersey.ResponseWriter if there is no content


romain-grecourt:
@Holubow You will to sign the OCA for us to accept your contribution. See https://www.oracle.com/technetwork/community/oca-486395.html

Holubow:
@romain-grecourt, I completed and submitted the OCA

PR #515, opened 2019-03-18
Fixed the referenced archetype

eldermoraes:
It was using the SE archetype, not the MP.


romain-grecourt:
Thanks for fixing this.

eldermoraes:
My pleasure!

PR #517, opened 2019-03-19
Remove exception stacktraces from unit tests (oracle#466)

Holubow:
* Specify surefire and failsafe redirectTestOutputToFile=true configuration properties in top-level parent pom.xml to suppress unit test exceptions


romain-grecourt:
@tomas-langer Are you okay with this change ? @barchetta FYI

tomas-langer:
If we keep the `-e` switch, this may be sufficient solution. Let's try it and if it hinders troubleshooting, we would have to come back and do the hard work (e.g. removing stacktraces that are expected).

PR #530, opened 2019-03-22
Provide configuration for enabled SSL protocols

### LINKED ISSUE #519: NettyWebServer uses a deprecated constructor for JdkSslContext

## Environment Details
* Helidon Version: 1.0.1
* Helidon SE 
* JDK version: irrelevant
* OS: irrelevant
* Docker version (if applicable): irrelevant

----------

## Problem Description
We would like to use Helidon, but with a restricted TLS version (1.2). Helidon does not allow this, and also uses a deprecated constructor of JdkSslContext in `NettyWebServer.java`.
Please change to use the non deprecated constructor, and also make it configurable which protocols to enable / use

## Steps to reproduce
# "Step by step instructions to reproduce the problem"
Start WebServer, 
do `openssl s_client -connect localhost:port -tls1`, this should fail if setup with "TLSv1.2"

romain-grecourt:
Provide configuration for enabled SSL protocols.
Upgrade netty to 4.1.34
Fixes #519

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,7 +138,7 @@
         <version.lib.microprofile-fault-tolerance-api>1.1.3</version.lib.microprofile-fault-tolerance-api>
         <version.lib.mockito>2.23.4</version.lib.mockito>
         <version.lib.mysql-connector-java>8.0.11</version.lib.mysql-connector-java>
-        <version.lib.netty>4.1.30.Final</version.lib.netty>
+        <version.lib.netty>4.1.34.Final</version.lib.netty>
         <version.lib.oci-java-sdk-objectstorage>1.2.44</version.lib.oci-java-sdk-objectstorage>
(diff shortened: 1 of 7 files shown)


romain-grecourt:
@barchetta This requires PLS approval work for the netty version update.

romain-grecourt:
@tomas-langer I agree, it makes sense to add the new "enabledProtocols" attribute to SSLContextBuilder rather than SocketConfiguration so that the config remains consistent.

However, `SSLContextBuilder` builds `javax.net.ssl.SSLContext` and does not hold the new attribute that we need to pass to the netty `JdkSslContext` constructor.

I'm thinking of creating a subclass of `javax.net.ssl.SSLContext` called `io.helidon.webserver.SSLContext` and have SSLContextBuilder return that. 

Do you have any thoughts or suggestions ?

PR #547, opened 2019-04-04
Quickstart update for native images

romain-grecourt:
Make the helidon-quickstart-se example "GraalVM ready" instead of having a separate example.
Use a custom Maven plugin to execute the `native-image` command.


examples/quickstarts/helidon-quickstart-se/src/main/docker/Dockerfile.native
```diff
@@ -20,21 +20,21 @@ WORKDIR /helidon
 ADD *.jar target/
 ADD ${libs.classpath.prefix}/*.jar target/
 
-RUN echo "$GRAALVM_HOME/bin/native-image -H:Name=helidon-native -H:IncludeResources='application.yaml|logging.properties' -H:+ReportExceptionStackTraces -classpath target/* ${main.class}"
-RUN $GRAALVM_HOME/bin/native-image \
+RUN echo "${GRAALVM_HOME}/bin/native-image -H:Name=${project.artifactId}-native -H:IncludeResources='application.yaml|logging.properties' -H:+ReportExceptionStackTraces -classpath target/* ${mainClass}"
+RUN ${GRAALVM_HOME}/bin/native-image \
         --static \
-        -H:Name=helidon-native \
+        -H:Name=${project.artifactId}-native \
```
tomas-langer: This does not have to be project specific. We can use the hardcoded value. That way the generated image is the same for every project (and easier to troubleshoot)

examples/quickstarts/helidon-quickstart-se/src/main/docker/Dockerfile.native
```diff
@@ -20,21 +20,21 @@ WORKDIR /helidon
 ADD *.jar target/
 ADD ${libs.classpath.prefix}/*.jar target/
 
-RUN echo "$GRAALVM_HOME/bin/native-image -H:Name=helidon-native -H:IncludeResources='application.yaml|logging.properties' -H:+ReportExceptionStackTraces -classpath target/* ${main.class}"
-RUN $GRAALVM_HOME/bin/native-image \
+RUN echo "${GRAALVM_HOME}/bin/native-image -H:Name=${project.artifactId}-native -H:IncludeResources='application.yaml|logging.properties' -H:+ReportExceptionStackTraces -classpath target/* ${mainClass}"
```
tomas-langer: This restricts the resources to be always application.yaml and logging.properties. 
Also for a general application, it may make sense to add configuration file such as application-graal.yaml that would be configured as the first configuration file, so we could set things such as port to be explicitly the same as the one in Dockerfile

examples/quickstarts/helidon-quickstart-se/README.md
```diff
@@ -88,3 +88,41 @@ kubectl get pods                    # Verify connectivity to cluster
...
+See https://github.com/oracle/helidon-build-tools/tree/master/helidon-maven-plugin
+ for more information.
+
+### Multi-stage Docker build
+
+We publish a Docker image that contains GraalVM to Docker hub:
+ `helidon/java-native`.
+
+```
+mvn package -Pnative-image -Dnative.image.skip
```
tomas-langer: This is the main target for a graal native image application. I think this should be part of the maven build as well.

examples/quickstarts/helidon-quickstart-se/src/main/docker/Dockerfile.native
romain-grecourt: As we discussed in today's meeting, this will be removed once I finalize the proper multi-stage docker build, the Dockerfile.native will also use mvn to build the native executable.

examples/quickstarts/helidon-quickstart-se/src/main/docker/Dockerfile.native
romain-grecourt: Same comment as above this will be removed once I show proper multi-stage docker build. There won't be such duplicated configuration.

examples/quickstarts/helidon-quickstart-se/README.md
romain-grecourt: No sure I understand your comment. Note that this part will change as well with proper multi-stage docker build. The instruction will be `docker build -f Dockerfile.native .`

romain-grecourt:
Quick note this PR is work-in-progress. I Will add proper multi-stage Docker build support.

examples/quickstarts/helidon-quickstart-se/README.md
```diff
@@ -85,6 +85,40 @@ Exercise the application as described above
...
+ `-DgraalVMHome=/path` to the Maven command.
+
+See https://github.com/oracle/helidon-build-tools/tree/master/helidon-maven-plugin
+ for more information.
+
+### Multi-stage Docker build
+
+```
+docker build -t helidon-quickstart-se-native -f Dockerfile.native .
+```
```
tomas-langer: Again, please add commands to run the application.

examples/quickstarts/helidon-quickstart-se/pom.xml
```diff
@@ -201,4 +174,31 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
```
tomas-langer: Health checks do not work in Graal native image as they are now.
In Graal, we would need to replace line 101 in `Main.java`:
Instead of `.add(HealthChecks.healthChecks())   // Adds a convenient set of checks`
we would need: `.add(HealthChecks.heapMemoryCheck())`
as file system health check fails on MacOS and deadlock healthcheck can work on neither environment, as this feature is not supported in Graal native image at all

tomas-langer: Maybe we should update implementation of these two health checks to check if in graal - and if so, just return "up" without any initialization...

PR #555, opened 2019-04-10
Changes required to FT implementation to support new features and TCKs

spericas:
Contains all the necessary changes to support FT 2.0 except for an issue related with bulkhead queues that will be addressed in a separate PR.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandRetrier.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandRetrier.java
@@ -17,12 +17,14 @@
 package io.helidon.microprofile.faulttolerance;
 
 import java.lang.reflect.Method;
+import java.time.temporal.ChronoUnit;
 import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Future;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
 import java.util.logging.Logger;
 
 import javax.interceptor.InvocationContext;
@@ -65,6 +67,8 @@ import static io.helidon.microprofile.faulttolerance.FaultToleranceMetrics.getHi
 public class CommandRetrier {
     private static final Logger LOGGER = Logger.getLogger(CommandRetrier.class.getName());
 
+    private static final int DELAY_CORRECTION = 250;
+
     private final InvocationContext context;
 
     private final RetryPolicy retryPolicy;
@@ -98,7 +102,8 @@ public class CommandRetrier {
             // Initial setting for retry policy
             this.retryPolicy = new RetryPolicy()
                                    .withMaxRetries(retry.maxRetries())
-                                   .withMaxDuration(retry.maxDuration(), TimeUtil.chronoUnitToTimeUnit(retry.durationUnit()))
+                                   .withMaxDuration(retry.maxDuration(),
+                                                    TimeUtil.chronoUnitToTimeUnit(retry.durationUnit()))
                                    .retryOn(retry.retryOn());
 
             // Set abortOn if defined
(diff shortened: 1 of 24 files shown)


microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FailsafeChainedFuture.java
```diff
@@ -54,21 +55,36 @@ public boolean isDone() {
...
-        return result;
     }
 
     @Override
     @SuppressWarnings("unchecked")
-    public T get(long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
-        final T result = delegate.get();
-        if (result instanceof Future) {
-            return ((Future<T>) result).get(timeout, unit);
+    public T get(long timeout, TimeUnit unit) throws CancellationException, InterruptedException,
```
tjquinno: This is begging for refactoring because the two methods are nearly identical. I wondered about creating a third method `get(Supplier<T>)` that would accept a different lambda from each of the existing `get` methods. Then I remembered (admittedly after playing around a bit in an IDE) that the exceptions thrown from the new `get` would mandate a try/catch block in the two lambda expressions so we'd end up not saving much. I remember bumping into this a few times and coming away frustrated each time.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceCommand.java
```diff
@@ -57,6 +58,7 @@
 public class FaultToleranceCommand extends HystrixCommand<Object> {
     private static final Logger LOGGER = Logger.getLogger(FaultToleranceCommand.class.getName());
 
+    static final int MAX_THREAD_WAITING_PERIOD = 2000;      // milliseconds
```
tjquinno: Should this max be configurable?

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
```diff
@@ -260,33 +332,31 @@ public Type getBaseType() {
...
-         */
+        @Override
         @SuppressWarnings("unchecked")
         public <R extends Annotation> R getAnnotation(Class<R> annotationType) {
-            if (CommandBinding.class.equals(annotationType)) {
-                return (R) LiteralCommandBinding.getInstance();
-            }
-            return delegate.getAnnotation(annotationType);
+            Optional<Annotation> optional = annotationSet.stream()
+                    .filter(a -> annotationType.isAssignableFrom(a.annotationType())).findFirst();
```
tjquinno: Will the formatting police want the `.findFirst` to be on the next line by itself??

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandRetrier.java
```diff
@@ -92,24 +102,33 @@ public CommandRetrier(InvocationContext context, MethodIntrospector introspector
...
             if (retry.jitter() > 0) {
-                long delay = TimeUtil.convertToNanos(retry.delay(), retry.delayUnit());
                 long jitter = TimeUtil.convertToNanos(retry.jitter(), retry.jitterDelayUnit());
 
-                /*
-                 * We need jitter <= delay so we compute factor for Failsafe so we split
-                 * the difference, essentially making jitter and delay equal, and then set
-                 * the factor to 1.0.
-                 */
+                // Need to compute a factory and adjust delay for Failsafe
```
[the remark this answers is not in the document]
spericas: Will fix.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandRetrier.java
```diff
@@ -165,17 +203,31 @@ public Object execute() throws Exception {
...
-     * retries are executed.
+     * retries are executed. If running in async mode, this method will execute
+     * on a different thread.
      *
      * @return Object returned by command.
      */
     private Object retryExecute() throws Exception {
+        // Config requires use of appropriate context class loader
+        if (contextClassLoader != null) {
+            Thread.currentThread().setContextClassLoader(contextClassLoader);
```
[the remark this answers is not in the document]
spericas: No need because this is a special thread pool that we use for async. So when a new thread is started, the loader is set and then returned to the pool awaiting its next use.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/CommandRetrier.java
```diff
@@ -276,6 +328,23 @@ private String createCommandKey() {
...
+     *
+     * @param commandKey Command key.
+     * @param key Property key.
+     * @param value Property value.
+     */
+    private void setProperty(String commandKey, String key, Object value) {
+        final String actualKey = String.format("hystrix.command.%s.%s", commandKey, key);
+        synchronized (ConfigurationManager.getConfigInstance()) {
+            final AbstractConfiguration configManager = ConfigurationManager.getConfigInstance();
+            if (configManager.getProperty(actualKey) == null) {
```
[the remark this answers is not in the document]
spericas: Because a command with the same key can have multiple instances running in parallel. Values only need to be set once.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FailsafeChainedFuture.java
spericas: I'll try to refactor it.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceCommand.java
spericas: Yes, there's actually a few others like this. I'll create another issue to address this later.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
spericas: I'm not sure, but I'll move it.

PR #558, opened 2019-04-10
Add openapi support to SE

tjquinno:
Basic approach:

1. Take advantage of smallrye/smallrye-open-api which implements MP OpenAPI support.
2. Create `OpenAPISupport` class in SE which developers can use to explicitly and easily add OpenAPI support to their apps. Our SE support _excludes_ annotation support. The most reasonable use of OpenAPI support in SE will be with a static OpenAPI document file built into the app. This fits well with the idea often described of writing the OpenAPI document _first_ and using that to drive development of the endpoints described in the document. 
3. MP support will use a Java service that sets up `OpenAPISupport` _with_ annotation handling (as well as the other mechanisms for providing OpenAPI model information).

Note that the smallrye implementation expects to find a jandex index to support its OpenAPI annotation handling. It does not even fail gracefully if the index is missing. The work savings to us in leveraging smallrye's impl is significant and so requiring the presence of the jandex index seems worth it. Early on, our MP support will check for the index and fail if it is absent. Later (presumably a short while later), our MP support can create the index just-in-time (in memory) if the developer did not build it into the application. I'd expect our first release that supports OpenAPI to build the index JIT if it is missing.


tjquinno:
The starting point of this PR contains only the SE changes. MP will come along shortly.

tomas-langer:
I have a problem with the dependency on small rye - mostly because it obviously requires Weld (as you note above - Jandex index is expected).
We do not have any explicit dependency on Weld in our project nor should we have. We actually plan to do a POC against other CDI implementation(s) in the future.
Is the dependency on Weld "hard", or can it use any CDI implementation?

romain-grecourt:
What are the required runtime dependencies for SE ? Does it drag weld and jandex even for SE ?

tjquinno:
I have been digging into the dependency situation since Tomas noted his concern. As a brute force experiment I deleted the entire `org/jboss/weld` directory from my local .m2 repository. Building, testing, and packaging the smallrye openapi parent project (includes implementation and TCK) caused maven to download weld-parent-36 (a pom) but nothing else. 

Still pursuing.

Issue #596, opened 2019-04-18
Context propagation across threads


tomas-langer:
Support for propagating a context (similar to the one in ServerRequest) across threads.
This is needed in the following scenarios:

- We need to propagate data from server request further down the stream - e.g. for tracing, security for outbound calls
- We need to propagate data across thread boundaries - e.g. when invoking an asynchronous client call, when running a method in a different thread for Fault tolerace

Tasks:

- [x] Prepare one pager
- [x] Implement as a common module
- [x] Refactor existing implementations (new tasks to be created) - tracing, security, webserver, fault tolerance, grpc, jersey client, jersey support

PR #599, opened 2019-04-18
Helidon common context One pager.

tomas-langer:
Ope pager with proposal of common context handling in Helidon.

The "why" is described in the one pager.

### DIFF
--- a/docs-internal/context.md
+++ b/docs-internal/context.md
@@ -1,0 +1,90 @@
+# Helidon Context Support Proposal
+
+Provide a way to propagate Context across thread boundaries.
+We already use the WebServer request context to store and retrieve
+information (such as in security).
+
+This proposal would allow us to use the context:
+
+- in any thread managed (or wrapped) by Helidon
+- to propagate information all the way from server inbound
+    request to a client outbound request
+- to handle correctly things such as parent SpanContext in tracing
+    where we currently have access only to the overall WebServer
+    parent span
+    
+## Problem 
+
+We currently have context available only when we can reach
+ServerRequest. There are cases, where we need to propagate
+context information and where dependency on WebServer is not desired,
+such as in Helidon DB.
+Also with introduction of gRPC server, we have another source
+of contexts, that should not depend on http modules.
+    
+## Proposal
+
+This proposal defines:
+
+- new common API
+- updates to existing modules
+
+### New API
+
+A new module `helidon-common-context` is to be created.
+Public API consists of:
+
+- `io.helidon.common.context.Context` - a copy of existing `ContextualRegistry` in `http`,
+    to be used as the main point of registering and retrieving contextual values
+- `io.helidon.common.context.Contexts` - a utility class with helpful static methods
+    to work with `Context`:
+    - `Optional<Context> context()`: to retrieve current context (if there is one)
+    - `ExecutorService wrap(ExecutorService)`: to wrap any executor service and create a context-aware executor service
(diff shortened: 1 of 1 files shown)


tomas-langer:
See #600 for first reference

ljnelson:
Do I understand right this is basically CDI scopes (outside of CDI obviously)?

tomas-langer:
This is similar to CDI request scope. The Context is actually hierarchical - the context can have a parent.
So you have a single point to access context related information (which may come from request specific context, or may be a part of the overall application (e.g. webserver) - such as `Tracer`)

docs-internal/context.md
```diff
@@ -0,0 +1,90 @@
...
+                            .accept(nextRequest, nextResponse));
+```
+
+Processing through an executor service (`JerseySupport`):
+```java
+ExecutorService executorService = (service != null) 
+                ? service 
+                : Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors() * 2);
+        
+this.service = Contexts.wrap(executorService);
```
batsatt: I assume this will do nothing if `executorService` is already wrapped?

tomas-langer: Yes, I have already pushed that fix

tomas-langer: Did not realize this is the one pager PR. You can have a look on the first impl in #600

tomas-langer:
@romain-grecourt @spericas can you please add your opinion? I would like to have at least two approvals before merging this and continuing with the implementation

tjquinno:
I know I am very late to this, and it largely looks very good and very useful.

A few quick things, and I apologize if these have been discussed elsewhere already.

Does `inContext` automatically clear the context in the thread used for the `Runnable` or `Callable` after that code has finished or failed due to an exception? 

Is there need for an explicit `clear` operation to disassociate a context from a thread?

@batsatt asked earlier about preventing multiple wrapping for executor services. 

Similarly, does `Contexts.inContext` _replace_ any existing context that might be present or does it `wrap` it, creating nested contexts? This could presumably happen if the `Runnable` or `Callable` passed to `inContext` itself invoked `inContext`. 

If nested contexts can exist, what does that imply about what the "current" context is...the lowest-level, some kind of merge of the nested contexts? 

On the other hand, if `inContext` _replaces_ any current context then is that something we want to allow...put another way, is that a bug? Should we enforce that a thread's context be `clear``ed before being overridden by another `inContext` invocation?

Does each `Context` have a unique identifier, probably most useful for logging and debugging.

Some of these might not make sense in the use cases we _intend_ now, but a general and powerful mechanism like this might be used in clever ways in the future that we don't anticipate now.

tomas-langer:
The `Context.inContext` is intended to clean up after itself.
If you look to the first implementation, I am using a stack.
The `inContext` method is implemented as follows:
```
public static void inContext(Context context, Runnable runnable) {
        push(context);
        try {
            runnable.run();
        } finally {
            pop();
        }
    }
```

tomas-langer:
The `clear` operation would require quite a big change, as the stack approach would fail in case you clear the context. 
As the implementation now uses a `finally` block, the context will get cleared correctly.
The hierarchy is intentional - you may want to create a child context and run further processing in the child context, rather than the original (parent) context.
The unique identifier sounds good, I will check the implementation and add it if not there.

tomas-langer:
Related to #596

tomas-langer:
I am merging this in. Let's discuss further comments on the implementation

PR #600, opened 2019-04-18
Common context

tomas-langer:
First implementation of the common context described in #599

### DIFF
--- a/common/context/src/main/java/io/helidon/common/context/Context.java
+++ b/common/context/src/main/java/io/helidon/common/context/Context.java
@@ -17,6 +17,8 @@
 package io.helidon.common.context;
 
 import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
 
 /**
@@ -55,7 +57,7 @@ public interface Context {
      * @return new instance
      */
     static Context create() {
-        return new ListContext();
+        return builder().build();
     }
 
     /**
@@ -67,7 +69,16 @@ public interface Context {
      * @return new instance
      */
     static Context create(Context parent) {
-        return new ListContext(parent);
+        return builder().parent(parent).build();
+    }
+
+    /**
+     * Fluent API builder for advanced configuration.
+     *
+     * @return a new builder
+     */
+    static Builder builder() {
+        return new Builder();
     }
 
     /**
@@ -144,4 +155,70 @@ public interface Context {
      * @throws NullPointerException If {@code classifier} is null.
      */
     <T> Optional<T> get(Object classifier, Class<T> type);
+
+    /**
+     * A unique id of this context within this runtime.
+     *
+     * @return id of this context
+     */
+    String id();
+
+    /**
+     * Fluent API builder for {@link Context}.
+     */
+    class Builder implements io.helidon.common.Builder<Context> {
+        private static final AtomicLong PARENT_CONTEXT_COUNTER = new AtomicLong(1);
+        private Context parent;
+        private String id;
+
+        @Override
+        public Context build() {
+            if (null == id) {
+                id = generateId();
+            }
+            return new ListContext(this);
+        }
+
+        private String generateId() {
+            if (null == parent) {
+                return String.valueOf(PARENT_CONTEXT_COUNTER.getAndIncrement());
+            }
+            if (parent instanceof ListContext) {
+                return parent.id() + ":" + ((ListContext) parent).contextCounter().getAndIncrement();
+            }
+            // we cannot depend on the parent, so let's use UUID
+            return parent.id() + ":" + UUID.randomUUID();
+        }
+
+
+        /**
+         * Parent of the new context.
+         * @param parent parent context
+         *
+         * @return updated builder instance
+         */
+        public Builder parent(Context parent) {
+            this.parent = parent;
+            return this;
+        }
+
+        /**
+         * Identification of the new context, should be unique within this runtime.
+         *
+         * @param id context identification
+         * @return updated builder instance
+         */
+        public Builder id(String id) {
+            this.id = id;
+            return this;
+        }
+
+        Context parent() {
+            return parent;
+        }
+
+        String id() {
+            return id;
+        }
+    }
 }
--- a/common/context/src/main/java/io/helidon/common/context/Context.java
+++ b/common/context/src/main/java/io/helidon/common/context/Context.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common.context;
+
+import java.util.Optional;
+import java.util.function.Supplier;
+
+/**
+ * A registry for context objects. Enables instance localization between several <i>services / components / ...</i> integrated in
+ * a particular known scope. ContextualRegistry instance is intended to be associated with a scope aware object such as
+ * WebServer, ServerRequest or ClientRequest.
+ *
+ * <p>Context contains also a notion of <i>classifiers</i>. Classifier is any object defining additional <i>key</i> for registered
+ * objects. To obtain such registered object, the same classifier (precisely, any equal object) has to be used.
+ *
+ * <p>Classifiers can be used as follows:<ol>
+ * <li>As an additional identifier for registered objects of common types, like a {@link String}, ...<br>
+ * <pre>{@code
+ * // User detail provider service
+ * registry.register("NAME_PARAM_ID", "Smith");
+ * registry.register("GENDER_PARAM_ID", "male");
+ * ...
+ * // User consumer service
+ * String name = registry.get("name", String.class);
+ * }</pre></li>
+ * <li>As an access control mechanism where only owners of the classifier can retrieve such contextual instance.<br>
+ * <pre>{@code
+ * // In some central security service.
+ * registry.register(securityFrameworkInternalInstance, new AuthenticatedInternalIdentity(...));
+ * ...
+ * // In some authorization filter known by a central security service
+ * AuthenticatedInternalIdentity auth = registry.get(securityFrameworkInternalInstance, AuthenticatedInternalIdentity.class);
+ * }</pre></li>
+ * </ol>
+ */
+public interface Context {
+
+    /**
+     * Creates a new empty instance.
+     *
+     * @return new instance
+     */
+    static Context create() {
+        return new ListContext();
+    }
+
+    /**
+     * Creates a new empty instance backed by its parent read-through {@link Context}.
+     *
+     * <p>Parent {@code registry} is used only for get methods and only if this registry doesn't have registered required type.
+     *
+     * @param parent a parent registry
+     * @return new instance
+     */
+    static Context create(Context parent) {
+        return new ListContext(parent);
+    }
+
+    /**
+     * Register a new instance.
+     *
+     * @param instance an instance to register
+     * @param <T>      a type of the registered instance
+     * @throws NullPointerException if the registered object is {@code null}
+     */
+    <T> void register(T instance);
+
+    /**
+     * Register a new instance using a provided supplier. The supplier is guarantied to be called at most once when it's
+     * requested by the {@link #get(Class)} method. The returned value is then registered and the supplier is never used again.
+     *
+     * @param type     a type of supplied instance
+     * @param supplier a supplier of the instance to register
+     * @param <T>      a type of supplied object
+     * @throws NullPointerException if the {@code type} or the {@code supplier} is {@code null}
+     */
+    <T> void supply(Class<T> type, Supplier<T> supplier);
+
+    /**
+     * Optionally gets registered instance by its type.
+     *
+     * <p>More specifically, it returns <b>the last</b> registered instance without specified <i>classifier</i> which can be cast
+     * to the requested type.
+     *
+     * @param type a type of requested instance
+     * @param <T>  a type of requested instance
+     * @return The last registered instance compatible with the specified type
+     */
+    <T> Optional<T> get(Class<T> type);
+
+    /**
+     * Register a new instance with specified classifier.
+     *
+     * <p>Registered instance can be obtained only using {@link #get(Object, Class)} method with a {@code classifier} equal with
+     * the one used during registration.
+     *
+     * @param classifier an additional registered instance classifier
+     * @param instance   an instance to register
+     * @param <T>        a type of the registered instance
+     * @throws NullPointerException if {@code classifier} or registered object is {@code null}
+     */
+    <T> void register(Object classifier, T instance);
+
+    /**
+     * Registers a new instance using a provided supplier. The supplier is guarantied to be called at most once when it's
+     * requested by the {@link #get(Object, Class)} method. The returned value gets registered and the supplier is never called
+     * again.
+     *
+     * <p>Registered instance can be obtained only using {@link #get(Object, Class)} method with a {@code classifier} equal with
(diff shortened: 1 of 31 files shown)


common/context/src/main/java/io/helidon/common/context/Context.java
```diff
@@ -0,0 +1,147 @@
...
+     * Register a new instance.
+     *
+     * @param instance an instance to register
+     * @param <T>      a type of the registered instance
+     * @throws NullPointerException if the registered object is {@code null}
+     */
+    <T> void register(T instance);
+
+    /**
+     * Register a new instance using a provided supplier. The supplier is guarantied to be called at most once when it's
```
batsatt: typo: guarantied -> guaranteed

common/context/src/main/java/io/helidon/common/context/Context.java
```diff
@@ -0,0 +1,147 @@
...
+     * requested by the {@link #get(Class)} method. The returned value is then registered and the supplier is never used again.
+     *
+     * @param type     a type of supplied instance
+     * @param supplier a supplier of the instance to register
+     * @param <T>      a type of supplied object
+     * @throws NullPointerException if the {@code type} or the {@code supplier} is {@code null}
+     */
+    <T> void supply(Class<T> type, Supplier<T> supplier);
```
batsatt: If you can do so without conflicts, I think you should use "register" instead of "supply" (everywhere). As it is, you're naming the how, not the what, and it is a little confusing, especially when the first word of the javadoc is "Register".

common/context/src/main/java/io/helidon/common/context/ListContext.java
```diff
@@ -0,0 +1,221 @@
...
+                return parent.get(type);
+            }
+        } else {
+            return Optional.of(result);
+        }
+    }
+
+    @Override
+    public <T> void register(Object classifier, T instance) {
+        Objects.requireNonNull(classifier, "Parameter 'classifier' is null!");
```
batsatt: My preference is to statically import methods like this to declutter the code and make it more readable; don't know if we have an established practice here or not.

common/context/src/main/java/io/helidon/common/context/ListContext.java
```diff
@@ -0,0 +1,221 @@
...
+        private volatile T instance;
+
+        RegisteredSupplier(Class<T> type, Supplier<T> supplier) {
+            this.type = type;
+            this.supplier = supplier;
+        }
+
+        @Override
+        public T get() {
+            if (missing) {
```
batsatt: Is it your intention here that it will continue to retry on subsequent calls if missing?

common/context/src/main/java/io/helidon/common/context/ContextAwareExecutorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+        return tasks.stream()
+                .map(this::wrap)
+                .collect(Collectors.toList());
+
+    }
+
+    protected  <T> Callable<T> wrap(Callable<T> task) {
+        Optional<Context> context = Contexts.context();
+
+        return context.<Callable<T>>map(value -> () -> Contexts.inContext(value, task))
```
batsatt: Seems overly complex/heavyweight (capturing lambda). A simple (and yes, old-fashioned) if/else or ternary using `context.isPresent()` would be clearer and more efficient.

common/context/src/main/java/io/helidon/common/context/ContextAwareExecutorImpl.java
```diff
@@ -0,0 +1,132 @@
...
+                // no need to wrap, no context
+                .orElse(task);
+    }
+
+    protected Runnable wrap(Runnable command) {
+        Optional<Context> context = Contexts.context();
+
+        // no need to wrap, no context
+        return context.<Runnable>map(value -> () -> Contexts.inContext(value, command))
```
batsatt: ditto above comment.

common/context/src/main/java/io/helidon/common/context/ListContext.java
```diff
@@ -0,0 +1,221 @@
...
+        cr.supply(type, supplier);
+    }
+
+    @Override
+    public <T> Optional<T> get(Object classifier, Class<T> type) {
+        Objects.requireNonNull(classifier, "Parameter 'classifier' is null!");
+        ClassifiedRegistry cr = classifiers.get(classifier);
+        if (cr != null) {
+            T result = cr.get(type);
+            if ((result == null) && (parent != null)) {
```
batsatt: These internal parens are not required, but... I'm assuming you find this clearer? If so, I'm pretty sure that this is not consistent across the project.

common/context/src/test/java/io/helidon/common/context/ContextAwareScheduledExecutorTest.java
```diff
@@ -0,0 +1,175 @@
...
+        AtomicReference<Exception> exceptionRef = new AtomicReference<>();
+
+        Contexts.inContext(context, () -> {
+            futureRef.set(wrapper.schedule(() -> {
+                try {
+                    Contexts.context()
+                            .get()
+                            .get(CountDownLatch.class)
+                            .get()
+                            .countDown();
```
batsatt: Wow, that's kind of an ugly/confusing sequence at first sight; it is what we get with layered Optionals though.

common/context/src/test/java/io/helidon/common/context/ContextAwareScheduledExecutorTest.java
```diff
@@ -0,0 +1,175 @@
...
+        Context context = Context.create();
+        context.register(TEST_STRING);
+
+        AtomicReference<ScheduledFuture<String>> futureRef = new AtomicReference<>();
+        AtomicReference<Exception> exceptionRef = new AtomicReference<>();
+
+        Contexts.inContext(context, () -> {
+            futureRef.set(wrapper.schedule(() -> {
+                try {
+                    return Contexts.context().get().get(String.class).get();
```
batsatt: ditto above comment

common/context/src/test/java/io/helidon/common/context/ContextAwareScheduledExecutorTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        Context context = Context.create();
+        context.register(cdl);
+
+        AtomicReference<ScheduledFuture<?>> futureRef = new AtomicReference<>();
+
+        Contexts.inContext(context, () -> {
+            futureRef.set(wrapper.scheduleAtFixedRate(() -> {
+                    Contexts.context()
+                            .get()
```
batsatt: :-)

common/context/src/test/java/io/helidon/common/context/ContextAwareScheduledExecutorTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        Context context = Context.create();
+        context.register(cdl);
+        AtomicReference<ScheduledFuture<?>> futureRef = new AtomicReference<>();
+
+
+        Contexts.inContext(context, () -> {
+            futureRef.set(wrapper.scheduleWithFixedDelay(() -> {
+                Contexts.context()
+                        .get()
```
batsatt: ;-)

common/http/src/main/java/io/helidon/common/http/ContextualRegistry.java
```diff
@@ -69,79 +68,4 @@ static ContextualRegistry create() {
...
-    /**
-     * Register a new instance using a provided supplier. The supplier is guarantied to be called at most once when it's
-     * requested by the {@link #get(Class)} method. The returned value is then registered and the supplier is never used again.
-     *
-     * @param type     a type of supplied instance
-     * @param supplier a supplier of the instance to register
-     * @param <T>      a type of supplied object
-     * @throws NullPointerException if the {@code type} or the {@code supplier} is {@code null}
-     */
-    <T> void supply(Class<T> type, Supplier<T> supplier);
```
batsatt: Ah. I see that you inherited this name and that this class is exposed via `Request`, so renaming would be a breaking change. Never mind.

common/context/src/main/java/io/helidon/common/context/Context.java
batsatt: Never mind; I see below that you hoisted this from a public API.

common/context/src/main/java/io/helidon/common/context/ListContext.java
tomas-langer: This implementation was moved from common http module - it comes from webserver. I did not modify the implementation at all (e.g. this is how our context works now in webserver)

common/context/src/main/java/io/helidon/common/context/ContextAwareExecutorImpl.java
tomas-langer: You are right. And also hard to understand. Fixed.

common/context/src/main/java/io/helidon/common/context/ContextAwareExecutorImpl.java
tomas-langer: Fixed as well

common/context/src/main/java/io/helidon/common/context/ListContext.java
tomas-langer: My setting of IDE - "add clarifying parentheses" - I do prefer to clearly separate the conditions.
We should add this to the list of known inconsistencies and eventually discuss and agree on a single approach.
BTW I am doing this since the beginning of the project...

tomas-langer: Creating a follow up issue for this: #648

common/context/src/test/java/io/helidon/common/context/ContextAwareScheduledExecutorTest.java
tomas-langer: In general we should not have public API methods that return null. We should always return `Optional` instead.
This is ugly because it is a test that intentionally wants to fail. In real world, you should never expect something to be in the context and fail with a NPE (or no such element exception in this case) - you must do something about it. 
So `Optional` is intentional and correct.

common/http/src/main/java/io/helidon/common/http/ContextualRegistry.java
tomas-langer: Yes, most of the context API is driven by this. We may revisit for version 2.0

common/context/src/main/java/io/helidon/common/context/ListContext.java
tomas-langer: I probably do not see the issue. If missing, missing is set to `false` - so next time the instance is returned directly.
This is a lazy supplier that caches the supplied value and as I see it, it is written correctly

common/context/src/main/java/io/helidon/common/context/ListContext.java
tomas-langer: Added to #648 , I disagree with you. Static imports may be fine, but sometimes they hide the source of the method and make it harder to read the code.

common/context/src/main/java/io/helidon/common/context/Context.java
tomas-langer: Fixed

common/context/src/test/java/io/helidon/common/context/ContextAwareScheduledExecutorTest.java
tomas-langer: Also a general answer to all the comments - as I have decided to have the API backward compatible and the webserver context implements the common context, we need to keep these APIs.

tomas-langer:
Resolves #596

tomas-langer:
Added context id. Also the existing `ListContextualRegistry` now fully delegates to the common `Context`, so we no longer duplicate the functionality.

common/context/src/main/java/io/helidon/common/context/Contexts.java
```diff
@@ -0,0 +1,129 @@
...
+     * Wrap an executor service to correctly propagate context to its threads.
+     *
+     * @param toWrap executor service
+     * @return a new executor service wrapping the provided one
+     */
+    public static ExecutorService wrap(ExecutorService toWrap) {
+        if (toWrap instanceof ContextAwareExecutorImpl) {
+            return toWrap;
+        }
+        if (toWrap instanceof ContextAwareScheduledExecutorImpl) {
```
aseovic: Is this necessary? `ContextAwareScheduledExecutorImpl` extends `ContextAwareExecutorImpl`, so it will be handled by the first `if` statement above -- it is effectively dead code, isn't it?

common/context/src/main/java/io/helidon/common/context/ContextAwareExecutorImpl.java
```diff
@@ -0,0 +1,134 @@
...
+        Optional<Context> context = Contexts.context();
+
+        if (context.isPresent()) {
+            return () -> Contexts.inContext(context.get(), task);
+        } else {
+            return task;
+        }
+    }
+
+    protected Runnable wrap(Runnable command) {
```
aseovic: Not so much a comment, but more of a general question on style preference.

This whole method *could* be replaced with a "functional" one-liner:
```
return Contexts.context()
                .map(ctx -> () -> Contexts.inContext(context.get(), command))
                .orElse(command);
```
Do we have a preference for one style over another?

The functional approach is slightly shorter, but the code below, as currently written, is easier to debug.

spericas: We have had a few discussions about this in the team, some feel the use of `map` in this context is somewhat artificial and not necessarily easier to read. Personally, I prefer the use of _actual_ language statements (which tends to reduce the number of capturing lambdas as well), but in the case above would have used a conditional expression `? :`.

common/context/src/main/java/io/helidon/common/context/Contexts.java
batsatt: Looks like it to me too, but I haven't had my coffee yet.

common/context/src/main/java/io/helidon/common/context/Contexts.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Run the runnable in the provided context.
+     * The runnable can use {@link #context()} to retrieve the context.
+     *
+     * @param context  context to run in
+     * @param runnable runnable to execute in context
+     */
+    public static void inContext(Context context, Runnable runnable) {
```
batsatt: When seen in use, it isn't immediately apparent that `Context.inContext(...)` means `run`. It would be clearer if this method was called `runInContext` or `invokeInContext`.

You could also consider moving this to a non-static method on `Context`, so the call becomes e.g.`context.run(runnable)` or `context.invoke(callable)`.

common/context/src/main/java/io/helidon/common/context/Contexts.java
tomas-langer: Yep, comes from the time the `extends` was missing.

common/context/src/main/java/io/helidon/common/context/ContextAwareExecutorImpl.java
tomas-langer: There is an issue we can add this to - #648 
In this specific case the lambda is not nice, as it does `ctx -> () -> ...` - this double lambding is not easy to read (e.g. a lambda that returns a lambda), and in this case the current code is nicer to read - there was actually a review comment asking me to switch from lambda to if/then/else

common/context/src/main/java/io/helidon/common/context/Contexts.java
tomas-langer: @spericas what do you think about this? I wanted to have the static methods grouped into the `Contexts` class.
You are right this could be an instance method on context itself. 
In that case, I would have to implement is as a default method. And if anybody ever decides to implement the `Context` interface and they also implement the `run` or `invoke` method, this may break our handling of contexts (as the method `push` and `pop` is package local and should not be a part of our public API.
Not sure whether we want to open the API in this way.

batsatt: Yes, there are definitely tradeoffs making them non-static. I'm OK keeping them static but still think they should be renamed.

tomas-langer: Renamed

spericas: @tomas-langer I also have a small preference for the non-static variant, but at this point I feel I need to experiment with all this in real use cases (like FT) before I can provide more detailed feedback, so I'm OK with how it looks now (method renaming helps).

PR #614, opened 2019-04-23
Common service loader

tomas-langer:
Added common library to load prioritized services using Java Service loader and custom implementations.
Fixed a TODO in microprofile Server, as it was missing prioritized service loader.
This is also a prerequisite for Helidon DB, and it could be used to replace existing Java Service loader usages in other modules (such as Config).

### DIFF
--- a/common/common/src/main/java/io/helidon/common/Prioritized.java
+++ b/common/common/src/main/java/io/helidon/common/Prioritized.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.common;
+
+/**
+ * Interface to define that this class is a class with priority.
+ * One of the uses is for services loaded by a ServiceLoader.
+ */
+@FunctionalInterface
+public interface Prioritized {
+    /**
+     * Priority of this class (maybe because it is defined
+     * dynamically, so it cannot be defined by an annotation).
+     * If not dynamic, you can use the {@code javax.annotation.Priority}
+     * annotation rather then implementing this interface as long as
+     * it is supported by the library using this {@code Prioritized}.
+     *
+     * @return the priority of this service
+     */
+    int priority();
+}
--- a/common/common/src/main/java/io/helidon/common/Prioritized.java
+++ b/common/common/src/main/java/io/helidon/common/Prioritized.java
@@ -18,9 +18,26 @@ package io.helidon.common;
 /**
  * Interface to define that this class is a class with priority.
  * One of the uses is for services loaded by a ServiceLoader.
+ * <p>
+ * A {@code Prioritized} with lower priority number is more significant than a {@code Prioritized} with a
+ * higher priority number.
+ * <p>
+ * For cases where priority is the same, implementation must define ordering of such {@code Prioritized}.
+ * <p>
+ * <b>Negative priorities are not allowed and services using priorities should throw an
+ * {@link java.lang.IllegalArgumentException} if such a priority is used (unless such a service
+ * documents the specific usage of a negative priority)</b>
+ * <p>
+ * A {@code Prioritized} with priority {@code 1} is more significant (will be returned before) priority {@code 2}.
  */
 @FunctionalInterface
 public interface Prioritized {
+    /**
+     * Default priority for any prioritized component (whether it implements this interface
+     * or uses {@code javax.annotation.Priority} annotation.
+     */
+    int DEFAULT_PRIORITY = 5000;
+
     /**
      * Priority of this class (maybe because it is defined
      * dynamically, so it cannot be defined by an annotation).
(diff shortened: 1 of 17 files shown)


tomas-langer:
The reason there is no method `HelidonServiceLoader.builder(ServiceInterface.class)`:
since jpms was introduces in Java 9, loading service implementation from another module would require a (more or less) complicated setup of module dependencies that are not part of the normal `module-info.java`. To work around this problem, the `ServiceLoader` is loaded from the correct module, and passed as a parameter to this utility.

common/common/src/main/java/io/helidon/common/Prioritized.java
```diff
@@ -0,0 +1,34 @@
...
+public interface Prioritized {
+    /**
+     * Priority of this class (maybe because it is defined
+     * dynamically, so it cannot be defined by an annotation).
+     * If not dynamic, you can use the {@code javax.annotation.Priority}
+     * annotation rather then implementing this interface as long as
+     * it is supported by the library using this {@code Prioritized}.
+     *
+     * @return the priority of this service
+     */
```
spericas: Javadoc should state whether priority 1 is greater than 100 or vice-versa. Also are priorities 0 or negative allowed?

tomas-langer: It is defined in the `ServiceLoader`, as that uses the `Prioritized` interface. I thought that the "meaning" of priority should be left to the user.

common/service-loader/src/main/java/io/helidon/common/serviceloader/HelidonServiceLoader.java
```diff
@@ -0,0 +1,360 @@
...
+    public static final String SYSTEM_PROPERTY_EXCLUDE = "io.helidon.common.serviceloader.exclude";
+
+    private static final Logger LOGGER = Logger.getLogger(HelidonServiceLoader.class.getName());
+
+    private final List<T> services;
+
+    /**
+     * Create a builder for customizable service loader.
+     *
+     * @param javaServiceLoader the Java Service loader used to get service implementations
```
batsatt: I think you should remove the `java` prefix (everywhere); it is already clear that the parameter is the system service loader. This prefix seems both redundant and we probably don't want to set a precedent here ;-).

common/service-loader/src/main/java/io/helidon/common/serviceloader/HelidonServiceLoader.java
```diff
@@ -0,0 +1,360 @@
...
+         * When configured to use Java Service loader, then the result is a combination of all service implementations
+         * loaded from the Java Service loader and those added by {@link #addService(Object)} or {@link #addService(Object, int)}.
+         * When set to {@code false} the Java Service loader is ignored.
+         * <p>
+         * This defaults to {@code true}.
+         *
+         * @param useJavaServiceLoader whether to use the Java Service loader
+         * @return updated builder instance
+         */
+        public Builder<T> useJavaServiceLoader(boolean useJavaServiceLoader) {
```
batsatt: This seems a little weird given that your static `builder()` and `create()` methods require a `ServiceLoader` parameter. Maybe just have variants of the static methods that do not take a `ServiceLoader`?

common/common/src/main/java/io/helidon/common/Prioritized.java
spericas: Perhaps some of that info can be copied here. Also what about negative priorities? It's useful to be explicit here and state what is and what is not permitted. The `@Priority` annotation states that values _are generally non-negative_ for example. I'd make that even stronger.

common/service-loader/src/main/java/io/helidon/common/serviceloader/HelidonServiceLoader.java
tomas-langer: Removing. I will replace it to `useSystemServiceLoader` when enabling/disabling it.

common/service-loader/src/main/java/io/helidon/common/serviceloader/HelidonServiceLoader.java
tomas-langer: That is an interesting point. The problem is that I was missing any reference to the actual type of the service. And I do not want to create a method `create(ServiceInterface.class)`, as that would mislead the user into expecting the `ServiceLoader` to be automatically used.
I only need the information when creating exceptions (as everything else is driven by generic declaration), so a string would be sufficient....

common/common/src/main/java/io/helidon/common/Prioritized.java
tomas-langer: Fixed.

common/service-loader/src/main/java/io/helidon/common/serviceloader/HelidonServiceLoader.java
tomas-langer: Fixed

common/service-loader/src/main/java/io/helidon/common/serviceloader/HelidonServiceLoader.java
batsatt: Gotcha. So... what is the use case for disabling use of the `ServiceLoader`? Doesn't that just devolve to ordering the added services (assuming they added them with a priority) and maybe applying excludes? 

Maybe just remove this flag unless there is a good use case?

tomas-langer: The use case is when you want to configure only explicit services and not the ones loaded by the service loader. And yes, it devolves to ordering by priority and applying excludes.
It is an edge case, yet I wanted to support it for advanced "by hand" configurations.

tomas-langer: I have just realized that we already have these cases - Config.Builder supports disabling of system service loader. 
As by the time I want to disable/enable service loader I already need a builder instance of `HelidonServiceLoader`, it could be easily used in config as well. If I remove this feature, config could not utilize this library.

tomas-langer:
Resolves #621

PR #620, opened 2019-04-24
Minor edits to MicroProfile docs. Fixes #413

### LINKED ISSUE #413: doc: helidon.io doc area discussing MP should mention the MP version we support
Currently nothing on the helidon.io doc pages mentions which version of MP we support.

barchetta:
Fixes #413 doc: helidon.io doc area discussing MP should mention the MP version we support

### DIFF
--- a/docs/src/main/docs/microprofile/01_introduction.adoc
+++ b/docs/src/main/docs/microprofile/01_introduction.adoc
@@ -16,23 +16,35 @@
 
 ///////////////////////////////////////////////////////////////////////////////
 
-= Microprofile Introduction
-:description: Helidon Microprofile introduction
+= MicroProfile Introduction
+:description: Helidon MicroProfile introduction
 :keywords: helidon, microprofile, micro-profile
 
-MicroProfile is a platform definition that is familiar to Java EE developers. If you have experience with JAX-RS, JSON-P, and CDI, you 
-may prefer to use this model.
-
-To extend the functionality of your MicroProfile application, you might also decide to use the Helidon core APIs, especially for 
-configuration and security.
+MicroProfile is a collection of enterprise Java APIs that should feel familiar to
+Java EE developers. MicroProfile includes existing APIs like JAX-RS, JSON-P and
+CDI, and adds additional APIs in areas such as configuration, metrics, fault
+tolerance and more.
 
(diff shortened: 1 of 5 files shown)


docs/src/main/docs/microprofile/01_introduction.adoc
```diff
@@ -16,23 +16,35 @@
...
-To extend the functionality of your MicroProfile application, you might also decide to use the Helidon core APIs, especially for 
-configuration and security.
+MicroProfile is a collection of enterprise Java APIs that should feel familiar to
+Java EE developers. MicroProfile includes existing APIs like JAX-RS, JSON-P and
+CDI, and adds additional APIs in areas such as configuration, metrics, fault
+tolerance and more.
 
 == Getting Started with Helidon MicroProfile
 
+MicroProfile is implemented by Helidon MP. The 1.0 release of Helidon supports
```
tjquinno: The `sitegen.yaml` file defines the AsciiDoc attribute `helidon-version` so we could use that here as in `{helidon-version}` instead of hard-coding 1.0 to minimize the need for future edits just to track our Helidon releases. 

We could add a definition for `mp-version` also so various places in the doc could refer to the MicroProfile version without requiring multiple updates when we boost the MP version we support.

docs/src/main/docs/microprofile/01_introduction.adoc
```diff
@@ -16,23 +16,35 @@
...
+:description: Helidon MicroProfile introduction
 :keywords: helidon, microprofile, micro-profile
 
-MicroProfile is a platform definition that is familiar to Java EE developers. If you have experience with JAX-RS, JSON-P, and CDI, you 
-may prefer to use this model.
-
-To extend the functionality of your MicroProfile application, you might also decide to use the Helidon core APIs, especially for 
-configuration and security.
+MicroProfile is a collection of enterprise Java APIs that should feel familiar to
+Java EE developers. MicroProfile includes existing APIs like JAX-RS, JSON-P and
```
tjquinno: Maybe change "like" to "such as" here.

barchetta: Like, totally, I'll do that.

Issue #581, opened 2019-04-16
Helidon DB Client


tomas-langer:
This is a tracking issue for Helidon DB API and implementation.

Overall the implementation should follow the Development guidelines and:

- The implementation is full reactive and nonblocking
- There should be no dependency of the API or core on Tracing (not even Opentracing API)
- There should be no dependency of the API or core on Metrics
- There should be no dependency of the API or core on Healthchecks
- Rest will be defined in One pager

Tasks:

- [x] One Pager
- [x] API design
- [x] Core implementation
- [x] Implementation over JDBC
- [x] Metrics support (library + example)
- [x] Tracing support (library + example)
- [x] Health check support (library + example)
- [x] Implementation over Mongo reactive driver

Documentation moved to documentation epic: #1515 
Native image moved to native image SE epic: #1656

tomas-langer:
Depends on #596

tomas-langer:
Depends on #473

tomas-langer:
# TODOs:
## All
 - [ ] Unit tests...
 - [ ] The code is now an advanced POC - must revisit most of the code to make sure it is production quality
 - [ ] Statement parsing (when using named params in JDBC and any params in Mongo) - the usage of a state machine is very complicated. Could we maybe live with a regular expression and limitations on what can be in the statement (such as "no comments allowed in statements")? If not, The implementation must be done again and according to our code guidelines (and also a lot lot more readable)
## Core and Common
 - [ ] Find a good way to stream results to web server (see `DbResultSupport` in both examples)
 - [ ] Make sure all error states are correctly propagated to interceptors and to user
## JDBC
- [ ] Use a connection pool (see `ConnectionPool` interface, line 90) - and add appropriate documentation of configurable options
- [ ] Implement support for mapping of `DbRowResult` (see mongoDB impl)
- [ ] Document configuration properties of the `JdbcDbProviderBuilder`
- [ ] Make sure the `Flow` API is correctly implemented

## Mongo
- [ ] In `MongoDbStatementDml` - see TODO
- [ ] Make sure the backpressure is correctly implemented in `MongoDbStatementQuery`

tomas-langer:
DB was not chosen wisely - it looks like a Database implementation.
We are creating a "Reactive DB Client" so we should rename the modules, packages and classes to reflect that

spericas:
@tomas-langer Yes. I'd say "Reactive DB API" or "Reactive DB Client API". To me the Client is what you write for each DB itself.

[...]

PR #624, opened 2019-04-26
Helidon DB One Pager.

tomas-langer:
Related to #581

### DIFF
--- a/docs-internal/db.md
+++ b/docs-internal/db.md
@@ -1,0 +1,364 @@
+# Helidon Database Proposal
+
+Provide an API that enables reactive database support to be used with Helidon SE.
+
+## Proposal
+
+The Helidon DB is an abstraction layer over
+- database configuration
+- statement configuration
+- statement processing
+- handling of results
+
+The Helidon DB in addition supports
+- queries using either indexed or named parameters
+- mapping of POJOs to database parameters using provided mapper(s)
+    - `.createNamedDmlStatement("insert-mon").namedParam(pokemon)` would work if `Pokemon` db mapper is registered
+- mapping of database query results to POJOs using provided mapper(s)
+    - `dbRow.as(Pokemon.class)` would work if `Pokemon` db mapper is registered
+- mapping of database columns to arbitrary types using provided mappers(s)
+    - `column.as(Long.class)` would work even for String column, as long as `String -> Long` mapper is registered
+   
+
+The Helidon DB is *NOT*
+- Statement abstraction 
+    - users define statements in the language understood by the database (such as `SQL`)
+    - there is *NO* query language other than the database native query language
+    
+As part of this proposal there is also added support for [generic mapping](generic-mapping.md)
+
+### Required features
+
+1. The API must be reactive 
+2. The API must support backpressure for cases where multiple results are returned (queries)
+3. There must be support for configuring Tracing without dependency on Tracing API
+4. There must be support for configuring Metrics without dependency on Metrics API
+5. There must be support for configuring Healthchecks without dependency on Healthcheck API
+6. The first implementation must work at least over JDBC
+  
+### API
+
+The API main interfaces/classes:
+- `HelidonDb` - the entry point to create an instance of a Helidon DB, uses the usual `Builder`/`Config` pattern
+    the provider to be used is either explicitly configured, or defined by name (see `DbProvider` below) or
+    the first available one is used (ordered by priority). The instance has two methods to execute statements -
+    `execute` and `inTransaction`
+- `HelidonDbExecute` - the entry point to execution of statements
+- `DbStatement` - the abstraction of any type of statement (with more specific `DbStatementQuery` etc.)
+- `DbRowResult` - the interface for query results (supports access via `Subscriber`, can collect rows in memory etc.)
+- `DbRow` - represents a single row in the database (or single object), provides mapping methods (using mappers from 
+        SPI `DbMapperProvider`)
+- `DbColumn` - represent a single column in the database (or an object property), provides mapping methods (using
+    generic `MapperProvider`)
+- `DbResult` - used for statements of unknown type, invokes `Consumer` of either a DML statement (`Consumer<Long>`)
+        or a query statement (`Consumer<DbRowResult>`)
+- `DbMapper` - defines possible mapping operations required to map arbitrary types to types needed by the database 
+- `DbMapperManager` - used by DB implementations to access `DbMapper`s configured by `DbMapperProvider`s
+- `DbInterceptor` and `DbInterceptorContext` provide support for integration with the DB for Metrics, Tracing and similar
+- `DbException` a runtime exception to use when something fails (for JDBC this would usually wrap a `java.sql.SqlException`)
+
+
+### SPI
+
+SPI is used to add support for additional drivers (such as JDBC, Mongo DB etc.).
+The SPI Classes:
+- `DbProvider` - a Java Service loader interface used to locate available implementations
+- `DbProviderBuilder` - builder used to configure the underlying database driver and behavior of the 
+    driver implementation (such as `DbMapperProvider`, statements etc.)
+- `DbMapperProvider` - a Java Service loader interface used to locate available implementations of mappers specific
+    to database handling
+
+## Possible Implementations
+
+Our plan is to provide Helidon Database API implementations as very thin layers over several existing database APIs which are not reactive or their reactive API is unnecessarily complex or does not match Helidon style of API:
+
+- JDBC
+- MongoDB reactive driver
+- Oracle NoSQL
+- R2DBC - currently milestone releases
+- ADBA - currently in Alpha version
+
+## Configuration
+
+There is a single configuration option expected when loading the `HelidonDb` from configuration - 
+`source`. If defined, it is used to locate the `DbProvider` with the same name. If undefined, the
+first provider is used (ordered by priority).
+
+Each database driver implementation may choose what is required in configuration, though the following
+ is recommended:
+- Connectivity details
+    - `url` - the URL to the database (this may be a `jdbc` URL, or any database specific way of locating an instance)
+    - `username` - the user to connect to the database. If a database supports username/password based authentication, use this 
+            property
+    - `password` - password of the user to connect. If a database supports username/password based authentication, use this 
+        property
+- Statements configuration
+    - `statements` - a configuration node that contains name-statement pairs
+    
+*Note on named statements*
+Helidon DB is using named statements as a preferred way of configuration, as we can simply reference a statement
+ in logging, exception handling, tracing etc. If you use arbitrary statements, the name is generated as an UUID and
+ cannot provide better detail.
+The statements can be configured either in configuration file, or using the `HelidonDb.Builder.statements(DbStatements)` method. 
+
+### Example
+The following Yaml file configures JDBC data source to MySQL with custom statements:
+```yaml
+helidon-db:
+    source: jdbc
+    url: jdbc:mysql://127.0.0.1:3306/pokemon?useSSL=false
+    username: user
+    password: password
+    statements:
+      # required ping statement (such as for Healthcheck support) 
+      ping: "DO 0"
+      # Insert new pokemon
+      insert-mon: "INSERT INTO pokemons VALUES(:name, :type)"
+      select-mon-by-type: "SELECT * FROM pokemons WHERE type = ?"
+      select-mon: "SELECT * FROM pokemons WHERE name = ?"
+      select-mon-all: "SELECT * FROM pokemons"
+      update-mon-type: "UPDATE pokemons SET type = :type WHERE name = :name"
+      delete-mon: "DELETE FROM pokemons WHERE name = ?"
+      delete-mon-all: "DELETE FROM pokemons"
+```
+
+This configuration file defines both connectivity and statements.
+
+It also shows an example of the two options for parameters:
+1. Indexed parameters, such as in statement `delete-mon`
+2. Named parameters, such as in statement `insert-mon`
+
+## Statement Types
+
+Two basic types of statements are recognized:
+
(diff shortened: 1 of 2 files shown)


tomas-langer:
One pager for Helidon DB and for Generic Mapping.

docs-internal/db.md
```diff
@@ -0,0 +1,364 @@
...
+- mapping of database columns to arbitrary types using provided mappers(s)
+    - `column.as(Long.class)` would work even for String column, as long as `String -> Long` mapper is registered
+   
+
+The Helidon DB is *NOT*
+- Statement abstraction 
+    - users define statements in the language understood by the database (such as `SQL`)
+    - there is *NO* query language other than the database native query language
+    
+As part of this proposal there is also added support for [generic mapping](generic-mapping.md)
```
spericas: I have mixed feelings about the _generic mapping_ idea. Seems a bit too generic to be useful; perhaps it needs a bit more motivation? Automatic mapping can have some interesting side effects that are hard to find. OTOH, `DbMapper` makes perfect sense.

docs-internal/db.md
```diff
@@ -0,0 +1,364 @@
...
+#### Query without parameters
+
+```java
+db.execute(exec -> exec.namedQuery("select-all"))
+                .consume(response::send)
+                .exceptionally(throwable -> sendError(throwable, response));
+```
+
+#### Get with indexed parameters
+The `get` method is a shortcut to a query method that expects zero to 1 results.
```
spericas: I guess this is `get` or `namedGet` as below?

docs-internal/db.md
```diff
@@ -0,0 +1,364 @@
...
+                )
+        )
+).exceptionally(throwable -> <fail somehow>)
+```
+
+### Interceptor implementation
+
+Example interceptor (for tracing):
+```java
+public class DbTracing implements DbInterceptor {
```
spericas: How are these interceptors registered? Are they global or scoped to a certain type of statement or even an individual statement?

docs-internal/db.md
```diff
@@ -0,0 +1,364 @@
...
+        list.add(value.getName());
+        list.add(value.getType());
+        return list;
+    }
+    
+}
+```
+
+## Open questions
+The support for writing rows to WebServer currently transforms the row to JsonObject and then writes it to
```
spericas: Are you suggesting changes to the WebServer or simply future enhancements to `DbRowResult` to support these use cases? I read the latter.

docs-internal/db.md
romain-grecourt: We do have similar APIs in 2 places already (config & webserver) ; I like the idea of having a generic API for converting types, even if that's remains only an internal API.

If a built-in converter for DbRow -> Publisher<DataChunk> is provided for common formats this would nice, (e.g. JSON, XML).

Maybe we accommodate the potential side effects by providing a way to supply different converters when needed.

docs-internal/db.md
```diff
@@ -0,0 +1,364 @@
...
+
+## Proposal
+
+The Helidon DB is an abstraction layer over
+- database configuration
+- statement configuration
+- statement processing
+- handling of results
+
+The Helidon DB in addition supports
```
romain-grecourt: Maybe instead "The Helidon DB also supports".

docs-internal/generic-mapping.md
```diff
@@ -0,0 +1,249 @@
...
+Mapping provides tools to map a type to another type.
+
+The types may be
+ - a java class (`String.class`)
+ - a generic type (`io.helidon.common.GenericType` for any java type, such as `Supplier<String>`)
+ 
+Mappings can be provided either through Java Service loader, or through
+ explicitly configured providers using a builder.
+ 
+The mapping function gets either a pair of `Class` objects, or a pair of `GenericType` objects
```
ljnelson: Consider using just `java.lang.reflect.Type` instead.

docs-internal/generic-mapping.md
```diff
@@ -0,0 +1,249 @@
+# Generic Mapping Proposal
+
+Provide an API to map an arbitrary Java type to another arbitrary Java type.
```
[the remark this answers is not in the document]
ljnelson: What is the problem generic mapping is trying to solve?

docs-internal/generic-mapping.md
```diff
@@ -0,0 +1,249 @@
...
+In case no mapper is found, the result should be cached as well.
+
+The main API class is `MapperManager`.
+```java
+/**
+ * Map from source to target.
+ *
+ * @param source     object to map
+ * @param sourceType type of the source object (to locate the mapper)
+ * @param targetType type of teh target object (to locate the mapper)
```
ljnelson: Typo

docs-internal/generic-mapping.md
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Map from source to target.
+ *
+ * @param source     object to map
+ * @param sourceType class of the source object (to locate the mapper)
+ * @param targetType class of teh target object (to locate the mapper)
+ * @return result of the mapping
+ * @throws io.helidon.common.mapping.MapperException in case the mapper was not found or failed
+ */
+<SOURCE, TARGET> TARGET map(SOURCE source, Class<SOURCE> sourceType, Class<TARGET> targetType);
```
ljnelson: Can I pass in an `Integer` reference with `Number.class` as the second parameter?  If not (hazy on the generics here), then consider eliminating the `sourceType` parameter since it would then be redundant.

docs-internal/db.md
tomas-langer: There are a few places this is useful in the DB itself. I think that similar approach could be used by Config.
And yes, the possibility to add additional mappings is available (using a builder).
I agree that this may be interesting when multiple modules use it (e.g. String -> DateTime) mapping - we just have to be consistent in our approach and if we need a String -> DateTime that is unusual, it should be specified in a module specific mapping (as I have done with the DbMapper in DB implementation)

docs-internal/db.md
tomas-langer: I meant a general `get` approach - there are four methods (`get`, `namedGet`, `createGet`, and `createNamedGet`.

docs-internal/db.md
tomas-langer: Current when creating the `HelidonDb` instance - there is a builder method to add such an interceptor. So it is for the whole db.

docs-internal/db.md
tomas-langer: I meant the first.
Currently web server only supports `Publisher<DataChunk>` - there is no way for us to tell the response that we are sending a publisher of types.
The problem is that the writer needs to understand the type (e.g. `Pokemon.class`) and also understand how it should be send over the network.
For JSON, this would require sending the initial Array definition, then for each `onNext` create a JSON object separated by a comma from the previous one, then in `onComplete` write the end of Array.
I could not wrap my head around this issue and find a nice solution - help please!

docs-internal/generic-mapping.md
```diff
@@ -0,0 +1,249 @@
...
+
+The types may be
+ - a java class (`String.class`)
+ - a generic type (`io.helidon.common.GenericType` for any java type, such as `Supplier<String>`)
+ 
+Mappings can be provided either through Java Service loader, or through
+ explicitly configured providers using a builder.
+ 
+The mapping function gets either a pair of `Class` objects, or a pair of `GenericType` objects
+ that define the `SOURCE` and the `TARGET` of the mapping.
```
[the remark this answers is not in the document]
tomas-langer: Not sure - objects that define, or pair that defines?

docs-internal/generic-mapping.md
```diff
@@ -0,0 +1,249 @@
...
+ 
+For `Class` parameteres, lookup is done as follows:
```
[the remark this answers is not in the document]
tomas-langer: There are two parameters

docs-internal/generic-mapping.md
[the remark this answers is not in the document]
tomas-langer: Currently each module that does some mapping of types (or conversions), it creates its own set of predefined mappings (e.g. String -> Long, String -> int etc.)
Also the user may define additional mappers.
We could use this in config, database, webserver -> any place that needs to convert one type into another.

docs-internal/db.md
```diff
@@ -0,0 +1,364 @@
...
+    - `.createNamedDmlStatement("insert-mon").namedParam(pokemon)` would work if `Pokemon` db mapper is registered
+- mapping of database query results to POJOs using provided mapper(s)
+    - `dbRow.as(Pokemon.class)` would work if `Pokemon` db mapper is registered
+- mapping of database columns to arbitrary types using provided mappers(s)
+    - `column.as(Long.class)` would work even for String column, as long as `String -> Long` mapper is registered
+   
+
+The Helidon DB is *NOT*
+- Statement abstraction 
+    - users define statements in the language understood by the database (such as `SQL`)
```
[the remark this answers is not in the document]
tomas-langer: ok

docs-internal/db.md
tomas-langer: ok

docs-internal/generic-mapping.md
tomas-langer: That is not very useful, as it does not carry generic information, so you could not define a function as
`T map(S object, Class<S> srcType, Class<T> targetType`
You would end up with
`Object map(Object source, Type srcType, Type targetType`

docs-internal/generic-mapping.md
tomas-langer: fixed

docs-internal/generic-mapping.md
tomas-langer: No you cannot. The type mapping is exact - so if you have a mapper Number -> String and you pass an Integer, mapping would fail

docs-internal/db.md
tomas-langer: full list provided

tomas-langer:
There is a very rough implementation of JDBC available on https://github.com/oracle/helidon/pull/625 if you want to have a look and play around with the sources.

docs-internal/generic-mapping.md
tomas-langer: We could probably change the lookup, though it would be much more complicated, as we would have to look for a mapper of each super class of the source type. 
Assume the following example:
`CharSequence chars = mapper.map(149, CharSequence.class)`
what if there is a mapper for Number -> String - is that a suitable one?

tomas-langer: I just realised that this can already be implemented due to the nature of `MapperProvider`.
The method has two parameteres - the source type and the desired type.
How you implement the lookup is really up to you (or the implementor of that provider).
So if there is a  source: `Intger.class` and target: `CharSequence.class` requested, and you are capable of mapping any number into a string, you could return a mapper for this pair. This is then cached by the `MapperManager`.

docs-internal/db.md
tomas-langer: I have now modified the interceptor implementation
 - you can configure interceptors for
    - database
    - statement type (e.g. query, get, update, insert, delete, DML, unknown)
    - statement name (specific statements)

I have also added support for service loader using `DbInterceptorProvider`

tomas-langer:
I have done a few changes:
 - `HelidonDb` is now just `Db` - we do not have `HelidonWebServer` nor `HelidonConfig`
    - I was thinking about `Database`, but that would make other classes very long (e.g. `DatabaseInterceptorContext`)
    - also all other interfaces in the package just use `Db` prefix
 - `HelidonDbExecute` is now just `DbExecute` 
 - `DbStatement` - I have removed methods with functions to map to parameters - the class is cluttered enough already
 - DbInterceptors now asynchronous

docs-internal/generic-mapping.md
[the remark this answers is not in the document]
romain-grecourt: Need help from a native speaker ; I think the subject is the "pair" thus it would need the s.

docs-internal/generic-mapping.md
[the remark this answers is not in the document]
romain-grecourt: `parameteres` -> `parameters`

romain-grecourt:
The name "Db" is not great IMO. How about "DBClient" or maybe something else ?

docs-internal/generic-mapping.md
ljnelson: For whatever it's worth, that is exactly what CDI, Guice, etc. and other parameterized-type-reflecting frameworks do when working with parameterized types in addition to just classes; you end up with almost what you indicate above, with the minor change that the return value is `T`: `T map(Object source, Type srcType, Type targetType)`.

docs-internal/generic-mapping.md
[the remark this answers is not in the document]
spericas: Je suis d'accord avec @romain-grecourt A pair is singular.

docs-internal/db.md
tomas-langer: Closing this comment, as we have merged generic mapper into master.
If we identify issues with this approach, we can add support for "qualifiers" as Romain mentioned...

docs-internal/generic-mapping.md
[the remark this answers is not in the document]
tomas-langer: Mapping is merged into master.

docs-internal/generic-mapping.md
tomas-langer: We  already use the `GenericType` in config, plan to use it in WebServer.
Also there are other specs, such as JAX-RS that uses its own `GenericType`
Also CDI is not using `Type`, but `TypeLiteral` that has the same function as our `GenericType`.
From `javax.enterprise.inject.Instance`: 
`<U extends T> Instance<U> select(TypeLiteral<U> subtype, Annotation... qualifiers);`

tomas-langer:
Closing, discussions should continue on the DB Client PR #657

Issue #627, opened 2019-04-27
JSON support implementations should share common code


batsatt:
The `JsonSupport` (jsonp) `JacksonSupport` and `JsonBindingSupport` classes in SE duplicate code to test if JSON is an acceptable response. This should be shared to avoid misalignment if changes are required in the future.

PR #628, opened 2019-04-27
Refactor JSON support classes to share a common base.

batsatt:


batsatt:
Fixes #627.

@m0mus Do you think that our Jackson and JSON-B support should also accept the two additional media types that JSON-P does: `application/javascript` and `text/javascript`?

ljnelson:
Not sure if it's relevant or not but at least for Jackson stuff have a glance nonetheless at https://github.com/FasterXML/jackson-jaxrs-providers/blob/jackson-jaxrs-providers-2.9.4/json/src/main/java/com/fasterxml/jackson/jaxrs/json/JacksonJsonProvider.java#L167-L192, which is based on real-world scenarios.

batsatt:
The FasterXml approach is interesting, testing _only_ the subtype and accepting any of 

- `json`
- `+json`
- `javascript`
- `x-javascript`
- `x-json`

This approach is even more inclusive than our JSON-P impl, but seems reasonable to me. 

@momus @ljnelson @spericas  Two questions:

1. Is there any reason that JSON-P should be a special case here?
1. If not, should we adopt the FasterXml approach for all of our impls?

spericas:
@batsatt I understand `json` and `+json` but isn't `application/javascript` for source code? Never seen the `x-...` ones before.

ljnelson:
@spericas This is exactly why I posted the comment: these are drawn from real-world uses of Jackson.  See https://github.com/FasterXML/jackson-jaxrs-providers/issues/14, for example.

spericas:
@ljnelson I was confused by the term JSONP, in the Jackson case it refers to JSON with Padding, right? Are we saying we want to support that?

ljnelson:
I don't know enough to have an opinion one way or the other.  I am just always leery of finding old, battle-tested code that makes reference to an edge case, and then deciding to omit the edge case.  It may make sense in this case, or it may not.

batsatt:
Yes, I too thought that `javascript` (and `ecmascript`) should be for JS source code. Clearly that is the main intention, but it is the odd backwards compatibility edge case that Laird is bringing up. Note, in particular, that @m0mus specifically mentions this one in the javadoc of `JsonProcessing.APPLICATION_JAVASCRIPT`.

m0mus:
My understanding is that `*/javascript`, `*/ecmascript` and there experiments `*/x-xxx` media types are designed for JavaScript program sources which are not JSON documents (see [here](http://www.rfc-editor.org/rfc/rfc4329.txt)). Using these media types for JSON documents is wrong, but many people are doing it. The question is shall we close our eyes on it and allow the incorrect usage? I think we should not by default. Maybe adding kinda switch in config allowing it is a good idea.

batsatt:
Santiago just pointed out to me that JSON with Padding requires handling of a `callback` parameter, such that the response is wrapped with that name and parens. We don't do that anywhere, so either we should do so or we should drop `javascript`.

batsatt:
Correctly supporting [JSON with Padding](https://en.wikipedia.org/wiki/JSONP) in SE should be relatively easy but still non-trivial and should not be included in this PR. Note that support is already present in MP via the `org.glassfish.jersey.server.JSONP` annotation. 

Given the [potential for abuse](https://en.wikipedia.org/wiki/JSONP#Security_concerns), an important requirement IMO is that the user should have to _explicitly_ enable support, just as required in MP by use of the `JSONP` annotation. 

Our `JsonSupport` (JSON Processing) impl supports the `javascript` media types, but is broken in that it does not perform the wrapping necessary to produce a valid JSON with Padding response.

Given the above, I conclude that we should 

1. Defer adding support for JSON with Padding in SE (see issue #635)
1. Remove support for the `javascript` media types in this PR.

While the latter could be seen as a breaking change, but... it breaks an already broken pathway.

PR #659, opened 2019-05-07
Upgrade Weld Core to 3.1.1.Final

barchetta:


barchetta:
@ljnelson can you do some testing with the CDI extensions with this change?

PR #663, opened 2019-05-09
Add the ability for the MP Health extension to discover HealthCheck instances not only from CDI beans but also from providers discovered using the ServiceLoader.

thegridman:
This is a straight forward change to the `HealthMpService` class so that the `HealthCheck` instances that it adds to the `HealthSupport.Builder` can be provided by an alternative means than just discovering CDI beans that implement `HealthCheck` and are annotated with `Health`. This functionality is required by the future gRPC MP work so that the gRPC MP extension can supply health checks for all of the deployed gRPC services as these health checks would not come from annotated CDI beans.

A new interface called `HealthCheckProvider` has been added that has a single method that returns an array of `HealthCheck` instances. The `HealthMpService` uses the Helidon service loader to discover implementations of `HealthCheckProvider` and adds the provided `HealthCheck`s to the `HealthSupport.Builder`.

A new functional test was added to verify that this all works, which necessitated adding the fail-safe plugin to the MP health module's `pom.xml`.

The existing `HealthCheckProducer` class has been renamed as it was quite close in meaning to `HealthCheckProvider` but it did not actually produce any health checks and is in fact a CDI producer for some JVM runtime classes.

### DIFF
--- a/microprofile/health/src/main/java/io/helidon/microprofile/health/HealthCheckProvider.java
+++ b/microprofile/health/src/main/java/io/helidon/microprofile/health/HealthCheckProvider.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.microprofile.health;
+
+import org.eclipse.microprofile.health.HealthCheck;
+
+/**
+ * A provider of {@link HealthCheck} instances.
+ * <p>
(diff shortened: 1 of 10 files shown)


microprofile/health/src/main/java9/module-info.java
```diff
@@ -22,6 +22,7 @@
     requires java.management;
 
     requires io.helidon.common;
+    requires io.helidon.common.serviceloader;
     requires io.helidon.health;
     requires io.helidon.microprofile.server;
```
tomas-langer: Missing the `uses HealthCheckProvider;` from module definition.

microprofile/health/src/main/java/io/helidon/microprofile/health/HealthCheckProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * using the {@link io.helidon.common.serviceloader.HelidonServiceLoader} and all of the
+ * {@link HealthCheck} instances are added to the health endpoint.
+ */
+public interface HealthCheckProvider {
+    /**
+     * Return the provided {@link org.eclipse.microprofile.health.HealthCheck}s.
+     *
+     * @return  the {@link org.eclipse.microprofile.health.HealthCheck}s
+     */
+    HealthCheck[] healthChecks();
```
tomas-langer: In general we should not use arrays in API - the only exception is varargs in `Builders`. Please refactor to use a `Collection` - either a `List`, or `Iterable`

tomas-langer:
Adding @romain-grecourt - could you please check the additions to the `pom.xml` files, if these are aligned with what we want to do?
Thanks

PR #666, opened 2019-05-09
ATN providers no longer throw an exception if cannot extract header

tomas-langer:

### DIFF
--- a/security/security/src/main/java/io/helidon/security/CompositeAuthenticationProvider.java
+++ b/security/security/src/main/java/io/helidon/security/CompositeAuthenticationProvider.java
@@ -38,8 +38,6 @@ import static io.helidon.security.SecurityResponse.SecurityStatus.SUCCESS;
  */
 final class CompositeAuthenticationProvider implements AuthenticationProvider {
     private static final AuthenticationResponse ABSTAIN_RESPONSE = AuthenticationResponse.abstain();
-    private static final CompletableFuture<AtnResponse> COMPLETED_ABSTAIN = CompletableFuture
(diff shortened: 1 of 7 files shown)


tomas-langer:
Resolves #666

security/security/src/main/java/io/helidon/security/CompositeAuthenticationProvider.java
```diff
@@ -70,14 +72,31 @@
 
     @Override
     public CompletionStage<AuthenticationResponse> authenticate(ProviderRequest providerRequest) {
-        CompletionStage<AuthenticationResponse> previous = CompletableFuture.completedFuture(AuthenticationResponse.abstain());
+        CompletionStage<AtnResponse> result = CompletableFuture
```
tomas-langer: This should use a constant - unnecessary creation of objects

security/security/src/main/java/io/helidon/security/CompositeAuthenticationProvider.java
```diff
@@ -70,14 +72,31 @@
...
         }
 
-        return previous.exceptionally(throwable -> {
+        return result.thenApply(atnResponse -> {
+            // when we get here, we should have all the sucesses and the response is the last one
+            AuthenticationResponse response = atnResponse.response;
+            List<AuthenticationResponse> successes = atnResponse.successResponses;
+
+            if (successes.isEmpty()) {
+                // we do not have any successful responses - abstain
```
tomas-langer: abstain only if optional, otherwise return the `atnResponse`

tomas-langer:
Rebased, added a small perf. optimization

Issue #646, opened 2019-05-03
Scope annotation not working alone


tomas-langer:
The `Scope` annotation works when defined more than once on a resource method.
If a single annotation exists, it fails in `ABAC` provider that there is no supported validator for it.
This is probably caused by the validator registering support only for `Scopes` and that is added by the javac only when there is more than one instance defined.

## Environment Details
* Helidon Version: 1.0.3
* Helidon MP
* JDK version: 1.8
----------

PR #672, opened 2019-05-10
Scope and Scopes are both supported now.

tomas-langer:


tomas-langer:
Resolves #646

Issue #661, opened 2019-05-08
OIDC Provider : Incorrect error if scopes do not match security requirement


asantaga:
## Environment Details
* Helidon Version: 1.0.3
* Helidon SE 
* JDK version: 1.8.0_211
* OS: Windows
* Docker version (if applicable):

----------

## Problem Description
Scenario
Using idcs-login example, the rest service is protected with TWO scopes.
If you generate a auth token from IDCS with only one scope and then execute the REST Service with this token you will get this error 

```
FINEST: FAILURE authn.authenticate 802e3482-81d6-45ce-8d46-b819b350e7e7:y  io.helidon.security.AuthenticationClientImpl lambda$authenticate$6 AuthenticationClientImpl.java 118 :: "Provider io.helidon.security.providers.oidc.OidcProvider. Message: Missing token, redirecting to identity server"
```

## Steps to reproduce
1. Deploy sample application ensure it works from a browser
2. Use a tool (like postman) to generate a Auth token from idcs using the https://<IDCS>/oauth2/v1/token rest call.  Ensure you only have one scope (first_scope)
3. Execute this against the REST endpoint (/jersey)
4. you wull receive the above error
5. Regenerate the token using two scopes 
6. Execute again, and it should work.

The error message is incorrect and should be valid token , invalid scopes

PR #673, opened 2019-05-10
OIDC provider now logs information about why the redirect happens.

tomas-langer:


tomas-langer:
Resolves #661

PR #691, opened 2019-05-14
Make default sizing of Jersey thread pool dynamic, based on # of cores

batsatt:
Makes the default sizing of the Jersey thread pool dynamic, based on the # of cores. Sets the core and max sizes the same for now since the ThreadPoolExecutor does not add threads except under extreme conditions.

Adds tracking of peak queue sizes via ThreadPool, a subclass of ThreadPoolExecutor.

Enables injection of ThreadPool instances by name.

### DIFF
--- a/common/configurable/src/main/java/io/helidon/common/configurable/ServerThreadPoolSupplier.java
+++ b/common/configurable/src/main/java/io/helidon/common/configurable/ServerThreadPoolSupplier.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


spericas:
I commented in the issue. What if we override the queue size where it makes sense to reduce the number of queued tasks? An app whose queues are very large is an app that likely needs tuning.

batsatt:
@tomas-langer The `ServerThreadPoolSupplier` was introduced specifically so that existing uses of `ThreadPoolSupplier` will continue to have the same behavior. Only `Server` and `JerseySupport` use the new supplier.

batsatt:
@spericas As I suggested in the issue, I think this is a little tricky and would rather get this fix in place now and start in on the longer term solution a.s.a.p.

spericas:
@batsatt Ok, in that case I'm fine with the change.

PR #693, opened 2019-05-15
Using logging.properties from classpath or from a file.

### LINKED ISSUE #443: MP should read logging.properties by default
## Environment Details
* Helidon Version: 1.0.0
* Helidon MP
* JDK version: 9.0.4
* OS: n/a

----------

## Problem Description
MP reads default configuration from predefined places, by spec.

It would be convenient for our MP implementation to read logging settings from `logging.properties` by default. 

The user can specify `-Djava.util.logging.config.file` on the `java` command line and that would remain an option, but having Helidon MP follow the convention of looking on the class path for `logging.properties` would be helpful.

tomas-langer:
Resolves #443

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
@@ -16,6 +16,7 @@
 
 package io.helidon.microprofile.server;
 
+import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
@@ -87,7 +88,7 @@ public final class Main {
         Path path = Paths.get("").resolve(LOGGING_FILE);
 
         if (Files.exists(path)) {
-            logConfigStream = Files.newInputStream(path);
+            logConfigStream = new BufferedInputStream(Files.newInputStream(path));
             source = "file: " + path.toAbsolutePath();
         } else {
             // second look for classpath (only the first one)
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
@@ -60,45 +60,51 @@ public final class Main {
     }
 
     private static void configureLogging() throws IOException {
+        String configClass = System.getProperty("java.util.logging.config.class");
         String configPath = System.getProperty("java.util.logging.config.file");
-        String source = "defaults";
+        String source;
 
-        if (configPath == null) {
-            // first set up logging - but only if not configured on command line
-            // Let's try to find a logging.properties
-            // first as a file in the current working directory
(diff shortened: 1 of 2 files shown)


microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
```diff
@@ -32,12 +45,62 @@ private Main() {
...
         server.start();
         port = server.port();
     }
 
+    private static void configureLogging() throws IOException {
+        String configPath = System.getProperty("java.util.logging.config.file");
+        String source = "defaults";
+
+        if (configPath == null) {
+            // first set up logging - but only if not configured on command line
```
ljnelson: Suppose I use the [`java.util.logging.class` System property](https://docs.oracle.com/en/java/javase/12/docs/api/java.logging/java/util/logging/LogManager.html).  I think you need to account for that as well.  That might be a better way to implement this whole thing, actually.

tomas-langer: I now take the `.class` property into account. I do not see the advantage in creating a new class when we can handle this in the same method...

microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
```diff
@@ -60,45 +60,51 @@ public static void main(String[] args) {
...
-                source = "file: " + path.toAbsolutePath();
-            } else {
-                // second look for classpath (only the first one)
-                logConfigStream = Main.class.getResourceAsStream("/" + LOGGING_FILE);
-                if (null != logConfigStream) {
-                    source = "classpath: /" + LOGGING_FILE;
-                }
-            }
+        if (Files.exists(path)) {
+            logConfigStream = Files.newInputStream(path);
```
ljnelson: Probably want to buffer this.

tomas-langer:
Rebased on latest master, as it fixes an intermittent issue that just happened.

tomas-langer:
@ljnelson could you please check if the latest changes are OK?

microprofile/server/src/main/java/io/helidon/microprofile/server/Main.java
```diff
@@ -32,12 +46,68 @@ private Main() {
...
+        InputStream logConfigStream;
+
+        Path path = Paths.get("").resolve(LOGGING_FILE);
+
+        if (Files.exists(path)) {
+            logConfigStream = new BufferedInputStream(Files.newInputStream(path));
+            source = "file: " + path.toAbsolutePath();
+        } else {
+            // second look for classpath (only the first one)
+            logConfigStream = Main.class.getResourceAsStream("/" + LOGGING_FILE);
```
ljnelson: Sorry; one last one where you might want to buffer.

PR #700, opened 2019-05-16
WIP: Improved flow control of data chunks

spericas:
WIP: Helidon subscribes to a library or application publisher by requesting `Long.MAX_VALUE`. This has the potential of buffering a large number of data chunks in a Netty IO loop if either the network or the client is slow. 

An example would be a DB publisher that is able to stream a large data set faster than the network can deliver it. Getting the right control flow between publishers and subscribers is non-trivial. We need to test this approach in a real world scenario.


webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -245,28 +254,39 @@ public void onNext(DataChunk data) {
...
                 channelFuture = ctx.write(httpContent);
             }
-
             channelFuture
                     .addListener(future -> {
                         data.release();
                         LOGGER.finest(() -> log("Data chunk sent with result: " + future.isSuccess()));
+                        if (future.isSuccess()) {
+                            chunksInFlight.decrementAndGet();
+                            chunksWritten.incrementAndGet();
```
ljnelson: An uninformed, naïve question: do these by any chance need to be mutated atomically together?

spericas:
Superseded by other work.

Issue #650, opened 2019-05-06
Config injection happens too soon


martin-sladecek:
## Environment Details
* Helidon Version: 1.0.3
* Helidon MP
* JDK version: 11.0.2
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description
When using `@Alternative` bean (probably also applies to CDI bean specialization), the Config injection attempt in ConfigCdiExtension happens before `@Alternative` bean is evaluated, so the original bean is also instantiated and the Configs are injected into it. This is undesired as the original bean might use configuration that is not present in the current environment, leading to an exception.  

## Steps to reproduce
1. Create `@Alternative CDI` bean with higher priority to any existing bean
2. Inject non-existent config property into original bean

ljnelson:
If I understand correctly, your issue, rephrased, is basically something like this: our portable extension evaluates which `InjectionPoint`s to process and then blindly processes them during `AfterBeanDiscovery`, without taking into account whether their "containing" bean is enabled or not.  The result is that we might be setting up producers for these "slots" without taking into consideration whether those slots belong to an enabled bean—they might belong to a bean that is superseded by an alternative.

That makes sense in general; while we create far too many producers of `ConfigProperty` values, doing so here doesn't really result in any harm.

But, it looks also like our portable extension validates injection points at `AfterBeanValidation` time not by calling `BeanManager#validate(InjectionPoint)` as should be the case but by somewhat oddly pre-instantiating beans and seeing if such instantiation fails.  This is clearly wrong and is a bug.

Thanks for using Helidon MP and for your report.

ljnelson:
Taking some notes here.

The MicroProfile Config specification often leaves some things up to interpretation.

In this case, the only validation guidance we get is kind of a double negative.  The javadocs for `ConfigProperty` say, in part, regarding the injection of `Optional` configuration values: "Contrary to natively injecting, if the property is not specified, this will not lead to a DeploymentException."  From this we can infer that the specification requires us to run some validation logic at `AfterDeploymentValidation` time.  This makes sense.  The TCK also looks for such `DeploymentException`s to be thrown if a requested configuration property value is not available.  Fine.

This means, though, that we must not simply gather injection points via the seemingly obvious `ProcessInjectionPoint` container lifecycle event (as we do incorrectly today), as these events happen before the container sorts through all the beans and considers alternatives, vetoes, and other whittling-down-of-the-overall-bean-population features.

[...]

PR #708, opened 2019-05-17
Ensures that our MP Config CDI extension processes only enabled injection points

ljnelson:
This PR ensures that `@ConfigProperty`-annotated injection points belonging only to enabled beans are processed.  This is a **partial** fix for #650.  Subsequent PRs will continue to fix the other multiple issues surfaced by that issue.  The TCK continues to pass.

Other minor changes:
* Serialization is repaired so that default information is read
* Generics warnings are eliminated
* Methods that should never have been `public` (i.e. are intended only for the CDI container to call and that should not be part of our public supported subclassable API) have been marked `@Deprecated`

### DIFF
--- a/microprofile/config/config-cdi/src/main/java/io/helidon/microprofile/config/cdi/ConfigCdiExtension.java
+++ b/microprofile/config/config-cdi/src/main/java/io/helidon/microprofile/config/cdi/ConfigCdiExtension.java
@@ -587,7 +587,8 @@ public class ConfigCdiExtension implements Extension {
     }
 
     /**
-     * A three tier description of a field type (main type, first generic type, second generic type).
+     * A three tier description of a field type (main type, first
+     * generic type, second generic type).
      */
     static final class FieldTypes {
         private TypedField field0;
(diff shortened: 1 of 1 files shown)


microprofile/config/config-cdi/src/main/java/io/helidon/microprofile/config/cdi/ConfigCdiExtension.java
```diff
@@ -131,8 +138,49 @@ any java class (except for parametrized types)
 
             pip.configureInjectionPoint()
                     .addQualifier(q);
+        }
+    }
+
+    private void harvestConfigPropertyInjectionPointsFromEnabledBean(@Observes ProcessBean<?> event) {
+        Bean<?> bean = event.getBean();
+        Set<InjectionPoint> beanInjectionPoints = bean.getInjectionPoints();
+        if (beanInjectionPoints != null && !beanInjectionPoints.isEmpty()) {
```
tjquinno: Mostly to satisfy my own curiosity... There are several null-checks in this method. The JavaDoc for the invoked methods does not say that null might be returned (although it doesn't say null _will not_ ever be returned either although one could infer that). Are you just being really careful or have you seen these methods return null in your past experience with CDI?

ljnelson: Mostly being careful.  I've seen `NullPointerException`s from non-market-leader CDI implementations in odd places and it has scarred me.  Also, a `Bean` in particular is an interface that an end user may implement, so she may return `null` from a variety of its methods and the specification does not forbid her from doing so.  So yes, being careful.

PR #711, opened 2019-05-18
Small minor performance improvements

TheRealHaui:
Small minor performance improvements.

### DIFF
--- a/metrics/prometheus/src/main/java/io/helidon/metrics/prometheus/PrometheusSupport.java
+++ b/metrics/prometheus/src/main/java/io/helidon/metrics/prometheus/PrometheusSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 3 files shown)


TheRealHaui:
I completely agree with you that measuring will be very complicated.
However: https://stackoverflow.com/questions/1508975/why-is-list-size0-slower-than-list-isempty-in-java

metrics/prometheus/src/main/java/io/helidon/metrics/prometheus/PrometheusSupport.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
  *
```
barchetta: Can you change this to `2017, 2019"? Thanks.

TheRealHaui: Done.

barchetta:
Thanks for contributing!

TheRealHaui:
You are welcome!

PR #712, opened 2019-05-20
Add MP OpenAPI support to Helidon MP

tjquinno:
This PR adds MP OpenAPI support to our MP server.

General areas of change:

1. Refactoring of SE OpenAPI support to reuse code common to SE and MP easily.
2. Some minor changes to our Arquillian deployable container.
3. Addition of the Helidon MP OpenAPI maven project under the MP project; it is also now part of the MP bundle.
4. Addition of the OpenAPI TCK runner project.

The MP OpenAPI support leverages much of the work already done for support in SE. A key difference is support for endpoints annotated with OpenAPI annotations. We layer on SmallRye's OpenAPI implementation, which requires a Jandex index of annotated types. Our support looks for and uses the `META-INF/jandex.idx` file if it exists on the class path. (This is the normal, preferred case in which the developer uses the Jandex maven plug-in to create the index and packages it into the app.) Otherwise, our OpenAPI support creates a Jandex index in-memory during app start-up by registering a CDI portable extension and recording each annotated type CDI processes.

Signed-off-by: Tim Quinn <tim.quinn@oracle.com>


microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param <X> annotated type
+     * @param event {@code ProcessAnnotatedType} event
+     */
+    <X> void processAnnotatedType(@Observes ProcessAnnotatedType<X> type) {
+        if (isIndexPresentOnClasspath) {
+            return;
+        }
+        Class<?> c = type.getAnnotatedType().getJavaClass();
+        annotatedTypes.add(c);
```
ljnelson: No action item here, but be aware as we discussed that the `ProcessAnnotatedType` event is fired during type discovery, and by the time you get to bean discovery the ecosystem may no longer do anything interesting with this type, i.e. it may be effectively orphaned.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /**
+     * Records each type that is annotated unless a Jandex index was found on
+     * the classpath (in which case we do not need to build our own in memory).
+     *
+     * @param <X> annotated type
+     * @param event {@code ProcessAnnotatedType} event
+     */
+    <X> void processAnnotatedType(@Observes ProcessAnnotatedType<X> type) {
```
ljnelson: This method can be `private`.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     * @return {@code IndexView} describing discovered classes
+     * @throws IOException in case of error reading an existing index file or
+     * reading class bytecode from the classpath
+     */
+    public IndexView indexView() throws IOException {
+        return isIndexPresentOnClasspath ? existingIndexFileReader() : indexFromHarvestedClasses();
+    }
+
+    private IndexView existingIndexFileReader() throws IOException {
+        try (InputStream jandexIS = getClass().getResourceAsStream(INDEX_PATH)) {
```
ljnelson: Consider `BufferedInputStream`.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+            }
+            LOGGER.log(Level.FINE, "Using Jandex index at {0}", INDEX_PATH);
+            return new IndexReader(jandexIS).read();
+        }
+    }
+
+    private IndexView indexFromHarvestedClasses() throws IOException {
+        Indexer indexer = new Indexer();
+        for (Class<?> c : annotatedTypes) {
+            try (InputStream is = contextClassLoader().getResourceAsStream(resourceNameForClass(c))) {
```
ljnelson: Consider `BufferedInputStream`.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+                index.printAnnotations();
+                index.printSubclasses();
+                LOGGER.log(level, baos.toString(Charset.defaultCharset().name()));
+            } finally {
+                System.setOut(oldStdout);
+            }
+        }
+    }
+
+    private ClassLoader contextClassLoader() {
```
ljnelson: This method can be `static`.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+        return Thread.currentThread().getContextClassLoader();
+    }
+
+    private String resourceNameForClass(Class<?> c) {
```
ljnelson: This method can be `static`.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/MPOpenAPIBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+ * Fluent builder for OpenAPISupport in Helidon MP.
+ */
+public final class MPOpenAPIBuilder extends OpenAPISupport.Builder {
+
+    private Optional<OpenApiConfig> openAPIConfig;
+    private Optional<IndexView> indexView;
+
+    @Override
+    public OpenApiConfig openAPIConfig() {
+        return openAPIConfig.get();
```
ljnelson: May not be reading things clearly but `openAPIConfig` can be `null` here, right?

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -445,165 +323,208 @@ public static Builder builder() {
...
+         * @param path non-null location of the static OpenAPI document file
          * @return updated builder instance
          */
-        public Builder addOperationServer(String operationID, String operationServer) {
-            Objects.requireNonNull(operationID, "operationID must be non-null");
-            Objects.requireNonNull(operationServer, "operationServer must be non-null");
-            apiConfigBuilder.addOperationServer(operationID, operationServer);
+        public Builder staticFile(String path) {
+            Objects.requireNonNull(path, "path to static file must be non-null");
+            staticFilePath = Optional.of(path);
```
ljnelson: Using an `Optional` here seems purposeless.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -445,165 +323,208 @@ public static Builder builder() {
...
+
+            if (specifiedMediaType == null) {
+                throw new IllegalArgumentException("OpenAPI file path "
+                        + path.toAbsolutePath().toString()
+                        + " is not one of recognized types: "
+                        + OpenAPIMediaTypes.recognizedFileTypes());
+            }
+            final InputStream is;
+            try {
+                is = Files.newInputStream(path);
```
ljnelson: Consider `BufferedInputStream`.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -445,165 +323,208 @@ public static Builder builder() {
...
+                        + " was specified but was not found");
+            }
+
+            try {
+                LOGGER.log(Level.FINE,
+                        () -> String.format(
+                                OPENAPI_EXPLICIT_STATIC_FILE_LOG_MESSAGE_FORMAT,
+                                path.toAbsolutePath().toString()));
+                return new OpenApiStaticFile(is, specifiedMediaType.format());
+            } catch (Throwable th) {
```
ljnelson: Catch `Exception` instead, maybe?

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -445,165 +323,208 @@ public static Builder builder() {
...
+                try {
+                    is.close();
+                } catch (IOException ex) {
+                    LOGGER.log(Level.FINE,
```
ljnelson: Consider adding a suppressed throwable instead of logging.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -445,165 +323,208 @@ public static Builder builder() {
...
+                            Path path = Paths.get(candidatePath);
+                            LOGGER.log(Level.FINE, () -> String.format(
+                                    OPENAPI_DEFAULTED_STATIC_FILE_LOG_MESSAGE_FORMAT,
+                                    path.toAbsolutePath().toString()));
+                            return new OpenApiStaticFile(is, candidate.format());
+                        }
+                        if (candidatePaths != null) {
+                            candidatePaths.add(candidatePath);
+                        }
+                    } catch (Throwable th) {
```
ljnelson: Maybe catch `Exception` instead.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -445,165 +323,208 @@ public static Builder builder() {
...
+                        if (is != null) {
+                            try {
+                                is.close();
+                            } catch (IOException ex) {
+                                LOGGER.log(Level.FINE,
```
ljnelson: Consider adding a suppressed throwable instead of logging.

openapi/src/main/java/io/helidon/openapi/SEOpenAPISupportBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Set various builder attributes from the specified {@code Config} object.
+     * <p>
+     * The {@code Config} object can specify {@value #CONFIG_PREFIX}.web-context
+     * and {@value #CONFIG_PREFIX}.static-file in addition to settings
+     * supported by {@link OpenAPIConfigImpl.Builder}.
+     *
+     * @param config the {@code Config} object possibly containing settings
+     * @return updated builder instance
+     */
```
ljnelson: `NullPointerException` can be thrown from this method.  If that's by design, please add an `@exception` tag.

openapi/src/main/java/io/helidon/openapi/internal/OpenAPIConfigImpl.java
```diff
@@ -294,6 +297,17 @@ public Builder addServer(String server) {
             return this;
         }
 
+        /**
+         * Sets whether annotation scanning should be disabled.
+         *
+         * @param value new setting for annotation scanning disabled flag
+         * @return updated builder
+         */
+        public Builder scanDisable(boolean value) {
```
ljnelson: Consider renaming to `scanningDisabled`.

ljnelson: …or `scanningEnabled`, to avoid the double negative situation.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -131,16 +133,16 @@ private boolean checkForIndexFile() throws IOException {
...
+    private void dumpIndex(Level level, Index index) throws UnsupportedEncodingException {
         if (LOGGER.isLoggable(level)) {
             LOGGER.log(level, "Dump of internal Jandex index:");
             PrintStream oldStdout = System.out;
-            try {
-                ByteArrayOutputStream baos = new ByteArrayOutputStream();
-                System.setOut(new PrintStream(baos));
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
+            try (PrintStream newPS = new PrintStream(baos, true, Charset.defaultCharset().name())) {
+                System.setOut(newPS);
```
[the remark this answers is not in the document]
tjquinno: It felt odd to write that code, but after I got over the shock that the `Index.printXXX` methods have `System.out` hardcoded in them, I rationalized that at least this happens only when the user has turned on `FINER` logging.

Also, `IndexWriter` writes a binary file; it's not human-readable. 

I looked briefly into crafting our own equivalent of the two `print` methods, but there's no supported way to get all the annos and all the subclasses.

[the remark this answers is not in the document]
ljnelson: I figured you had your reasons; that seems fine then.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Creates a new instance of the index builder.
+     *
+     * @throws IOException in case of error checking for the Jandex index file
+     */
+    public IndexBuilder() throws IOException {
+        isIndexPresentOnClasspath = checkForIndexFile();
+        if (isIndexPresentOnClasspath) {
+            LOGGER.log(Level.FINE, () -> String.format("Index file %s was located and will be used", INDEX_PATH));
```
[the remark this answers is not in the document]
tjquinno: The `INDEX_PATH` is `META-INF/jandex.idx` which will appear in the output in both log messages in this bit of code. Is that sufficient?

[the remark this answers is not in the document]
ljnelson: Probably; the use case I had in mind is some developer that doesn't know what part of the technology soup is failing; seeing `jandex` in the output in some form would be helpful in this case.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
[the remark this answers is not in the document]
tjquinno: Unfortunately, there's not a _single_ annotation. There might be a small set that would cover all types we need to harvest, but we'd need to be sure to construct that set correctly and we would couple our index creation tightly to the current MP OpenAPI spec.

I streamlined the code in that method as much as I could, but of course there's overhead in CDI to invoke it, even if it returns right away. 

Our doc will -- and the `INFO` message does -- emphasize the advantages of the developer using the Jandex maven plug-in which will potentially shrink the size of the index as well as move the work to build it into the build process and out of app start-up.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
tjquinno: Yes. Presumably this is no different from using a pre-constructed index which might contain types that, during CDI processing, get discarded.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
tjquinno: Fixed.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
ljnelson: Yes; good point.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Yes, better. Fixed.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Yes. I'm making the same change in the similar previous method. Fixed.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Done.

openapi/src/main/java/io/helidon/openapi/SEOpenAPISupportBuilder.java
tjquinno: True, if `config` is passed in as null, which "should never happen." Fixed.

openapi/src/main/java/io/helidon/openapi/internal/OpenAPIConfigImpl.java
tjquinno: I agree with your aesthetics in method naming here. That said, in `OpenAPIConfigImpl` for which this is a builder we are overriding the `scanDisable` method declared in SmallRye's impl. 

I think it's clearer to keep the same method name in the builder.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
tjquinno: The `IndexReader` we pass the stream to itself wraps a `BufferedInputStream` around ours.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
tjquinno: The invoked method does its own buffering.

ljnelson: Oh!  You mean `URL#openStream()` (called by `ClassLoader#getResourceAsStream()`) is guaranteed to buffer?  I didn't know that!  Learn something new every day.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
tjquinno: Fixed.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
tjquinno: Fixed.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/MPOpenAPIBuilder.java
tjquinno: Yes. In practice, our own code will have invoked the setters before the getters.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: The code uses the `Optional` aspect basically as a flag, whether an assignment was made to the static file path. If so, then `staticfile()` uses the assigned path; if not, it looks for a default path. We could have used a null value instead to indicate the lack of an explicit assignment.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Fixed.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: This time it's not documented whether the SmallRye code does its own buffering, so I've changed this.

microprofile/openapi/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+                <artifactId>jandex-maven-plugin</artifactId>
+                <executions>
+                  <execution>
+                    <id>make-test-index</id>
+                    <goals>
+                      <goal>jandex</goal>
+                    </goals>
+                    <phase>process-test-classes</phase>
+                    <configuration>
+                        <classesDir>${project.build.directory}/test-classes</classesDir>
```
romain-grecourt: This looks like a mix of 2 spaces indent and 4 spaces indent.

tjquinno: Fixed

PR #714, opened 2019-05-20
Use jandex index in MP quickstart, update jandex-plugin version

barchetta:


examples/quickstarts/helidon-quickstart-mp/pom.xml
```diff
@@ -50,6 +50,11 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>3.1</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.jboss.jandex</groupId>
+                    <artifactId>jandex-maven-plugin</artifactId>
+                    <version>1.0.5</version>
```
ljnelson: If you want to (or need to) use Jandex 2.1.1 (@tjquinno?) then this must be version 1.0.6.

barchetta: Good point. We moved to 2.1.1 in the workspace. I'll update these.

tjquinno: Our build (top-level project pom.xml) specifies Jandex 2.1.1.Final. Laird, do you have a reference for the 1.0.6 requirement for 2.1.1.Final?

tjquinno: @ljnelson ^

ljnelson: Yep; see https://github.com/wildfly/jandex-maven-plugin/commits/master and read up from April 3, 2019.

examples/quickstarts/helidon-quickstart-mp/pom.xml
```diff
@@ -113,6 +118,19 @@
...
+            <plugin>
+                <groupId>org.jboss.jandex</groupId>
+                <artifactId>jandex-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>make-index</id>
+                        <goals>
+                            <goal>jandex</goal>
+                        </goals>
+                        <phase>process-classes</phase>
```
tjquinno: This is the default phase for the plug-in. If you want to keep it here for clarity, fine. If you want to remove it for minimalism, fine.

examples/quickstarts/helidon-quickstart-mp/pom.xml
tjquinno: Because we specified 2.1.1.FInal already in our top-level pom's dep. mgt. section does that influence the plug-ins' dependencies too?

ljnelson: No.  If you wanted to effectively use 1.0.5 for some reason with Jandex 2.1.1, thus effectively creating a very clean Frankenstein jandex-maven-plugin that is exactly equivalent to 1.0.6 as far as I can tell, you would add a `<dependencies>` stanza to the relevant `<plugin>` element in `<pluginManagement>` that referenced Jandex 2.1.1.  I think it's simpler just to change the jandex-maven-plugin version.

tjquinno: Oh, yes. I wasn't suggesting we rely on that (even if it did work). I was just curious.

barchetta: I've updated the jandex plugin version in the top level pom.xml, and the jandex versions in the quickstart MP pom.xml. I've also updated the bookstore test app to use the jandex plugin.

PR #717, opened 2019-05-21
New functional test that verifies context propagation in FT

spericas:
Checks that a context is available in the different threads in which application code can execute.

### DIFF
--- a/tests/functional/context-propagation/pom.xml
+++ b/tests/functional/context-propagation/pom.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


tomas-langer:
It would be beneficial to test that we get the correct context.
For webserver (in JAX-RS resource) you can test that an instance of `SpanContext` is available in the `Context` - this way we make sure that the correct context was propagated.
This should work for any kind of `RequestScoped` bean

spericas:
> It would be beneficial to test that we get the correct context.
> For webserver (in JAX-RS resource) you can test that an instance of `SpanContext` is available in the `Context` - this way we make sure that the correct context was propagated.
> This should work for any kind of `RequestScoped` bean

Are you suggesting to check that `context.get(SpanContext.class)` exists? Is of a particular type? Or something different?

tomas-langer:
The first - that `context.get(SpanContext.class)` exists. The type depends on the type of tracing used, so we should not verify that.

PR #719, opened 2019-05-21
Ensures that JPA javadocs are present in overall Helidon javadocs

### LINKED ISSUE #718: JPA Javadocs are missing
The javadocs for JPA support are missing from the site documentation.

ljnelson:
Fixes #718.

### DIFF
--- a/integrations/cdi/jpa-cdi/pom.xml
+++ b/integrations/cdi/jpa-cdi/pom.xml
@@ -127,6 +127,24 @@
 
     <profiles>
         <profile>
+            <id>sources</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-javadoc-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>resource-bundle</id>
+                                <goals>
+                                    <goal>resource-bundle</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+        <profile>
             <id>xmlbind</id>
             <activation>
                 <jdk>[9,)</jdk>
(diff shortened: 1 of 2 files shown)


integrations/cdi/jpa-cdi/pom.xml
```diff
@@ -126,6 +126,24 @@
...
+            <id>sources</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-javadoc-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>resource-bundle</id>
+                                <goals>
+                                    <goal>resource-bundle</goal>
```
romain-grecourt: This solves some kind of problem with `includeDependencySources` right ?

ljnelson: Sort of: it is certainly supposed to.

The root problem: there's some generated code in here.  Among other things it generates a `package-info.java`.  The `package-info.java` has no package description.

What I've done that works fine in "normal" javadoc is: I also have a `src/main/javadoc/io/helidon/.../package.html` file with just the package description.  "Normal" javadoc whines about the two package information files (`package-info.java` and `package.html`) but processes them just fine and you get good javadocs along with the generated package-level annotations etc. etc.

By using the `resource-bundle` goal here I make sure that this `package.html` file is present and available to the master Helidon javadoc project.

Of course for some reason even though it _is_ available, and _should_ end up being exactly the same as if javadoc were run "normally", it doesn't actually make the package description anyway.  But I would like to leave this stanza here because it is "correct", in the sense that I do want that `package.html` available to any javadoc tool or plugin that would normally respect it.

romain-grecourt: Roger that.

PR #722, opened 2019-05-22
Updates to the recently-merged OpenAPI MP support

tjquinno:
I had omitted the new Helidon MP OpenAPI module from the BOM. 

This also captures a slightly longer but more accurate log message when we create the Jandex index just-in-time.

### DIFF
--- a/microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
+++ b/microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
@@ -66,7 +66,7 @@ public class IndexBuilder implements Extension {
                     + "so will build an in-memory index.%n"
                     + "This slows your app start-up and, depending on CDI configuration, "
                     + "might omit some type information needed for a complete OpenAPI document.%n"
-                    + "Consider using the Jandex plug-in to build the index "
+                    + "Consider using the Jandex maven plug-in to build the index "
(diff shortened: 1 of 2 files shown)


microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/IndexBuilder.java
```diff
@@ -62,9 +62,11 @@ public IndexBuilder() throws IOException {
...
             LOGGER.log(Level.INFO, () -> String.format(
-                    "OpenAPI support could not locate the index file %s "
-                    + "so will build an in-memory index. This slows your app start-up.%n"
-                    + "Consider using the jandex maven plug-in to build the index "
+                    "OpenAPI support could not locate the Jandex index file %s "
+                    + "so will build an in-memory index.%n"
+                    + "This slows your app start-up and, depending on CDI configuration, "
+                    + "might omit some type information needed for a complete OpenAPI document.%n"
+                    + "Consider using the Jandex maven plug-in to build the index "
                     + "and add it to your app at build-time",
```
spericas: "build time" instead of "build-time" given that it is not an adjective?

tjquinno: Very good catch. How about:

Consider using the Jandex maven plug-in during your build to create the index and add it to your app.

Issue #464, opened 2019-03-01
Request to endpoint /metrics fails occasionally


enbohm:
Requests to metrics path ```/metrics``` seems to fail from time to time and reports HTTP status 500. This occurs often when the Helidon application has been idle for a while.

## Environment Details
* Helidon Version: 1.0.0 & 2.2.0
* Helidon MP
* JDK version: 11 (build 11-28)
* OS: Linux
* Docker version: 18.09.02-CE

----------

## Problem Description
I'm using Prometheus (https://prometheus.io/) to gather metrics from various services (scrape configuration is to call ```/metrics``` endpoint twice per minute). Once or twice a day (often when the Helidon MP app has been idle for some time) the endpoint reports HTTP status 500 and an exception occurs which is caused by a NumberFormatException, see below screenshot. However, upcoming requests does work after a while.

![helidon-metrics-stacktrace](https://user-images.githubusercontent.com/4103786/53637893-bb471d80-3c24-11e9-84e0-f84363cb20ec.png)

## Steps to reproduce
Create a simple Helidon service and call ```/metrics``` on regular basic - after a few hours the endpoint reports HTTP status 500. Subsequent request will work but during the failure systems such as Prometheus will report that the service might be down which can cause negative alarms or other notifications.

## Addendum (reopening the issue)
The same general situation (long-running server -- unclear whether it has been idle or just long-running) gives the same symptom but during JSON formatting now.

spericas:
Looks like we may be computing a double whose string representation is "NaN" (not a number) that is later used to instantiate a BigDecimal. Not sure how an "N" could make it there otherwise.

tomas-langer:
division by zero?

spericas:
A division by zero gives you `Infinity`. An example of a `NaN` is `sqrt(-1)`. Looks like `WeightedSnapshot::getMean` is returning `NaN`. I think this can only happen when the values passed to the constructor of `WeightedSnapshot` include `NaN`'s (passed from `ExponentiallyDecayingReservoir`). I'm not really sure how that could happen, however we could defensively check for `Nan`'s and ignore them when part of a snapshot. Any other suggestions?

amannm:
probably the N from an <N> .... cough.... well good luck

spericas:
We've added some special handling for NaN's to prevent the 500 results. Please re-open with additional information if necessary.

tjquinno:
[...]

PR #723, opened 2019-05-22
Handle special case of NaN's to avoid NumberFormatException's 

spericas:
Handle special case of NaN's to avoid NumberFormatException's that can result in 500 error conditions related to metrics. See #464. This does not address the reason for those NaN's but at least prevents 500 errors to be returned. Some new tests added.


batsatt:
Might be a good idea to make `String.valueof(Double.Nan)` into a constant. 

What does Prometheus etc. do with a "NaN" value?

spericas:
@batsatt Fixed constant suggestion. From what I read, Prometheus "handles" NaN's, as in it does not throw exceptions. Regardless, the main objective here is to prevent a Helidon-based service to return 500.

PR #736, opened 2019-05-23
Initial commit for MP OpenAPI example

tjquinno:
MP OpenAPI example, inspired by the MP quick-start example but with no Docker info and added OpenAPI annotations to the endpoints.

### DIFF
--- a/examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
+++ b/examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
@@ -1,0 +1,201 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.microprofile.examples.openapi.basic;
+
+import java.util.Collections;
+
+import javax.enterprise.context.RequestScoped;
+import javax.inject.Inject;
+import javax.json.Json;
+import javax.json.JsonBuilderFactory;
+import javax.json.JsonObject;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.PUT;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.ExampleObject;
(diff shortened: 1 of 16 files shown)


examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
```diff
@@ -0,0 +1,201 @@
...
+    public GreetResource(GreetingProvider greetingConfig) {
+        this.greetingProvider = greetingConfig;
+    }
+
+    /**
+     * Return a worldly greeting message.
+     *
+     * @return {@link JsonObject}
+     */
+    @SuppressWarnings("checkstyle:designforextension")
```
tomas-langer: Can you please not put these annotations in the code? If you use Idea, you can simply suppress this warning for methods annotated with a specific annotation (probably in other IDEs there is a way for that as well).
We do not have a checkstyle warning that would fail on such a method, so it should not be in the sources.

examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
```diff
@@ -0,0 +1,201 @@
...
+    @GET
+    @Operation(summary = "Returns a generic greeting",
+            description = "Greets the user generically")
+    @APIResponse(description = "Simple JSON containing the greeting",
+            content = @Content(mediaType = "application/json",
+                               schema = @Schema(implementation = GreetingMessage.class
+                                            )
+                               )
+                )
```
spericas: Is this the formatting "suggested" in the MP samples? I find it odd to have closing parentheses formatted this way in annotations.

examples/microprofile/openapi-basic/src/test/java/io/helidon/microprofile/examples/openapi/basic/MainTest.java
```diff
@@ -0,0 +1,107 @@
...
+class MainTest {
+    private static Server server;
+
+    @BeforeAll
+    public static void startTheServer() throws Exception {
+        server = startServer();
+    }
+
+    @Test
+    void testHelloWorld() {
```
spericas: No test for "/openapi"?

examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
tjquinno: I'll remove those from this example. 

There are several other places in our examples -- notably the quickstarts. This example is based on the MP quickstart which is why these `@SuppressWarnings` annos were here.

examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
tjquinno: Good point. The MP OpenAPI spec has few examples but what they have looks like this. Better?
```
    @Path("/{name}")
    @GET
    @Operation(summary = "Returns a personalized greeting")
    @APIResponse(description = "Simple JSON containing the greeting",
            content = @Content(mediaType = "application/json",
                               schema = @Schema(implementation = GreetingMessage.class)))
    public JsonObject getMessage(@PathParam("name") String name) {...}
```

examples/microprofile/openapi-basic/src/test/java/io/helidon/microprofile/examples/openapi/basic/MainTest.java
tjquinno: I'll add at least one.

examples/microprofile/openapi-basic/src/main/java/io/helidon/microprofile/examples/openapi/basic/GreetResource.java
spericas: It is for me at least.

Issue #560, opened 2019-04-11
Fallback annotation breaks automated Tracing propagation


tomas-langer:
In case a method that uses Jersey client to call external service is annotated with `@Fallback`, the tracing is disconnected (client request is a root span).
This is caused by running in a different thread (an FT thread) rather than on the original JAX-RS thread.

This should be resolved in a way that we correctly transfer request context between threads.
Connect to work done for Helidon database for context support in executor services and for "in-context" execution.

## Environment Details
* Helidon Version: 1.0.2
* Helidon MP

spericas:
This makes sense, we need to fix it in the FT 2.0 code that is currently under review. You can assign it to me if you want.

tomas-langer:
depends on #596

PR #737, opened 2019-05-24
Changes to correctly trace using common context.

tomas-langer:


tomas-langer:
Resolves #560

tomas-langer:
Tracing switched from using its own thread local to common `Context`. I have added support for “application” context to be configured for MP Server and WebServer that is used as a parent for each request context

PR #738, opened 2019-05-24
Annotation analyzers now loaded via Helidon service loader,

### LINKED ISSUE #710: AnnotationAnalyzers load in indeterminate order
## Environment Details
* Helidon Version: 1.1.0
* Helidon MP
* JDK version: adoptopenjdk/openjdk11:jdk-11.0.3_7-alpine (on Docker) and JDK 11.0.2 (on Mac)
* OS: Alpine and Mac Mojave
* Docker version (if applicable): Docker for Mac 2.0.0.3

----------

## Problem Description
Two potential issues:

When the `AnnotationAnalyzer` load within `SecurityFilter` they do so via the `ServiceLoader` mechanism into a `List`. The `ServiceLoader` has no guarantee of order which means that list could be in any order and is indeterminate. When the analysers are processed, depending on the order in which they were loaded they can overwrite the answer of the previous one. Between running locally on the Mac and within docker, I've observed that on the mac the analysers were:

```
[
  io.helidon.microprofile.jwt.auth.JwtAuthAnnotationAnalyzer@68dbe225,
  io.helidon.security.abac.role.RoleAnnotationAnalyzer@145ef9e8
]
```

...and in Docker:

```
[
  io.helidon.security.abac.role.RoleAnnotationAnalyzer@7e1953f7,
  io.helidon.microprofile.jwt.auth.JwtAuthAnnotationAnalyzer@26f07261
]
```

In the latter case, the JWT analyser was overriding the Role analyser which gives different authentication responses. This makes it impossible to reliably run using authorisation using the same codebase across various operating systems and java versions.

Second issue is that `JwtAuthAnnotationAnalyzer` does not process `@PermitAll` annotation but does process `@RolesAllowed`. This means if the JWT analyser follows the Role analyser, it will override its decision. However, due to the previous issue it's not possible to guarantee the order in which this happens. In searching through the Microprofile JWT spec, it's somewhat light on how `@PermitAll` should work, however I did find [this](https://github.com/eclipse/microprofile-jwt-auth/blob/1.1/spec/src/main/asciidoc/interoperability.asciidoc).

## Steps to reproduce
Create an application that has two or more analysers, run it in various operating systems and java builds and observer, via debug, that the analysers are in a different order.

tomas-langer:
so priorities are used.
Add priority to JWT and Role annotation analyzers to make sure the
Role analyzer is last.

Resolves #710

### DIFF
--- a/security/integration/jersey/src/main/java/io/helidon/security/integration/jersey/SecurityFilter.java
+++ b/security/integration/jersey/src/main/java/io/helidon/security/integration/jersey/SecurityFilter.java
@@ -47,6 +47,7 @@ import javax.ws.rs.core.UriInfo;
 
 import io.helidon.common.CollectionsHelper;
 import io.helidon.common.OptionalHelper;
+import io.helidon.common.serviceloader.HelidonServiceLoader;
 import io.helidon.config.Config;
 import io.helidon.security.AuditEvent;
 import io.helidon.security.Security;
@@ -126,6 +127,9 @@ public class SecurityFilter extends SecurityFilterCommon implements ContainerReq
     }
 
     private void loadAnalyzers() {
+        HelidonServiceLoader.builder(ServiceLoader.load(AnnotationAnalyzer.class))
+                .build()
+                .forEach(analyzers::add);
         ServiceLoader.load(AnnotationAnalyzer.class)
                 .forEach(analyzers::add);
     }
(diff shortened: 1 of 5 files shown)


security/integration/jersey/src/main/java/io/helidon/security/integration/jersey/SecurityFilter.java
```diff
@@ -126,6 +127,9 @@ public SecurityFilter() {
     }
 
     private void loadAnalyzers() {
+        HelidonServiceLoader.builder(ServiceLoader.load(AnnotationAnalyzer.class))
+                .build()
+                .forEach(analyzers::add);
         ServiceLoader.load(AnnotationAnalyzer.class)
```
dansiviter: It this going to cause duplicates in the `analyzers` list?

PR #740, opened 2019-05-24
Infinite redirects can no longer happen with OIDC.

### LINKED ISSUE #644: OIDC infinite redirects
In some cases the OidcProvider causes infinite redirects.
Examples:
  
- Audience is not valid
- Issuer is not valid
- cookie is not created correctly

The provider should redirect always if:
 - the token is not valid anymore (time out)
 - we have insufficient scopes

The provider should redirect limited number of times in other cases.

tomas-langer:

Resolves #644

### DIFF
--- a/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
+++ b/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
@@ -228,6 +228,18 @@ import org.glassfish.jersey.client.authentication.HttpAuthenticationFeature;
  *     <td>helidon</td>
  *     <td>Realm returned in HTTP response if redirect is not enabled or possible.</td>
  * </tr>
+ * <tr>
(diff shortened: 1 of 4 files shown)


security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -1061,7 +1102,7 @@ public Builder scopeAudience(String audience) {
          * @param sameSite SameSite cookie attribute value
          * @return updated builder instance
          */
-        private Builder cookieSameSite(String sameSite) {
+        public Builder cookieSameSite(String sameSite) {
```
Verdent: Why was this method changed from private to public?

tomas-langer: This was a mistake in the first place - the methods should have been public, as they allow configuration that can be done using Config.

PR #749, opened 2019-05-29
Add openapi doc

tjquinno:
This adds a guide for using OpenAPI to the doc set.

The changes also add a card for the OpenAPI guide on the "Guides" overview page.

You can look at a rendered version of the guide by navigating to the `.adoc` file in my fork here: https://github.com/tjquinno/helidon/blob/add-openapi-doc/docs/src/main/docs/guides/07_openapi.adoc

It won't be identical to how it'd look on the web site but it's easier than reading the source.


docs/src/main/docs/guides/07_openapi.adoc
```diff
@@ -0,0 +1,240 @@
...
+application's start-up.
+
+Add the link:{jandex-plugin-doc}[Jandex maven plug-in] to the `<build><plugins>` 
+section of your `pom.xml`:
+
+[source,xml]
+----
+<plugin>
+    <groupId>org.jboss.jandex</groupId>
+    <artifactId>jandex-maven-plugin</artifactId>
```
ljnelson: Probably should have a `<version>` tag here, or include a `<pluginManagement>` stanza with the version specified.  I believe it must be 1.0.6 or later, yes? (Can't remember offhand.)

docs/src/main/docs/guides/07_openapi.adoc
```diff
@@ -0,0 +1,240 @@
...
+
+==== Adding Helidon MP OpenAPI support to your app
+Add these two sections to the `<dependencies>` portion of your `pom.xml`:
+
+[source,xml,subs="attributes+"]
+----
+<dependency> <!--1-->
+    <groupId>org.eclipse.microprofile.openapi</groupId>
+    <artifactId>microprofile-openapi-api</artifactId>
+    <version>{microprofile-openapi-version}</version>
```
ljnelson: Typo; need a `$` before the `{`

docs/src/main/docs/guides/07_openapi.adoc
```diff
@@ -0,0 +1,240 @@
...
+</dependency>
+
+<dependency> <!--2-->
+    <groupId>io.helidon.microprofile.openapi</groupId>
+    <artifactId>helidon-microprofile-openapi</artifactId>
+    <version>{helidon-version}</version>
```
ljnelson: Typo; need a `$` before the `{`

docs/src/main/docs/guides/07_openapi.adoc
```diff
@@ -0,0 +1,240 @@
...
+<2> Adds the Helidon MP OpenAPI runtime support.
+
+== OpenAPI for Helidon SE
+Helidon SE does not process annotations. You can still use OpenAPI with your
+Helidon SE app by using a static OpenAPI document, a model reader,
+and a filter.
+
+To use OpenAPI from your Helidon SE app:
+
+1. in your `pom` add a dependency for Helidon SE support,
```
ljnelson: Maybe `pom.xml` instead of just `pom`

docs/src/main/docs/guides/07_openapi.adoc
```diff
@@ -0,0 +1,240 @@
...
+in `application.yaml` (see <<se_config, the SE config table>> below).
+
+Here is the additional dependency for Helidon SE OpenAPI runtime support:
+
+[source,xml,subs="attributes+"]
+----
+<dependency>
+    <groupId>io.helidon.openapi</groupId>
+    <artifactId>helidon-openapi</artifactId>
+    <version>{helidon-version}</version>
```
ljnelson: Typo; need a `$` before the `{`

docs/src/main/docs/guides/07_openapi.adoc
tjquinno: Yes. I'll do that. Our example from which I drew all the code benefits from the inherited `<pluginManagement>` but a user won't be inheriting from our poms. Luckily it's pretty easy to pass the parameterized version we use in our build into the ADOC processing so when we build the site for a version of Helidon we capture the version we've used ourselves.

docs/src/main/docs/guides/07_openapi.adoc
tjquinno: Not really. Omitting the $ lets the AsciiDoctor processing substitute the attribute (property) assignment that's at the top of the `.adoc` file so the published guide actually has the real version in it.

docs/src/main/docs/guides/07_openapi.adoc
tjquinno: See earlier note about omitting the `$`.

docs/src/main/docs/guides/07_openapi.adoc
tjquinno: Same.

docs/src/main/docs/guides/07_openapi.adoc
tjquinno: Fixed.

tjquinno:
In light of Joe's comment and a similar discussion in slack, I'm revising the write-ups into separate SE and MP documents. More pushes are in the works.

tjquinno:
This recent push removes the draft single guide for SE and MP both and adds:

-  a new top-level nav item OpenAPI
-  a new SE-only page under that new nav item, and
-  a new MP-only page for OpenAPI under MicroProfile.

For ease of reading:
SE page: https://github.com/tjquinno/helidon/blob/add-openapi-doc/docs/src/main/docs/openapi/01_openapi.adoc

MP page: https://github.com/tjquinno/helidon/blob/add-openapi-doc/docs/src/main/docs/microprofile/08_openapi.adoc

paulparkinson:
Looks great. I'll modify my demos to include this and the division of mp and and se is definitely good. Super small items: MP: Can you remove "(1)" and "(2)" from the pom.xml snippets (make them separate blocks if you want to represent they are separate dependency types)and use the fully qualified names for OASFilter. Same for OASModelReader in SE.

tjquinno:
Hey, @paulparkinson. I've changed to use fully-qualified names for the OAS interfaces. Good point.

On your other suggestion -- My intent in using the callouts was to just explain why we add each of the two dependencies. That they have different scopes is related but incidental. 

Did you find the callouts distracting? Just plain not helpful? I'm trying to understand and address your underlying concern while still conveying in some way what I wanted to about the deps.

paulparkinson:
I'm just a big fan of copy paste as-is , that's all. :) The callouts are helpful - could you just put them outside the gray/<dependency> blocks?

tjquinno:
Oh oh oh, I get what you mean now.

That is a bug (I think) in GitHub's AsciiDoc rendering. The callout symbols are in XML comments in the `.adoc` source but when displayed on the GitHub pages the comment chars are gone, leaving the `(1)` and `(2)` out there unguarded.

Note that in the Java example the callout symbols are still behind the `//` comment introducer.

When I build the docs and run them locally, our rendered pages exclude the callout tags entirely from the copied text. 

I'll see if there's an open issue about this against GitHub's `.adoc` rendering.

tjquinno:
I did not find an existing issue against GitHub's `.adoc` rendering, so I've created this one:

https://github.com/github/markup/issues/1276

tjquinno:
Dan thought the GitHub/markup issue was better posted against Asciidoctor, so I closed the earlier issue and opened https://github.com/asciidoctor/asciidoctor/issues/3319

PR #757, opened 2019-06-05
WIP: Tracing refactoring

### LINKED ISSUE #461: Tracing support refactoring
- [x] when tracing is enabled, security traces usernames logged in - this should be optional (may leak information to wrong audience)
- [x] tracing propagation has issues when security is in place - traces are disconnected probably due to wrong parent SpanContext being used

tomas-langer:
Resolves #461 

Tracing is configurable
Security tracing uses the same (shared) library from each integration (except for GRPC that does not support common Context yet)
A new `TracingSupport` was created for Web Server - it must be registered as a first service to correctly create the parent span and brings configuration capabilities per server path
Microprofile tracing is now configurable (expects `tracing` config node also for global and path specific tracing configuration)


tomas-langer:
See https://github.com/oracle/helidon/pull/757/files#diff-03640123cd68d9def8a8b316f3cc8ef0 for documentation of the new (and old) features

tomas-langer:
Changing to WIP as `SystemService` PR is in progress and if succeeds, TracingSupport should be done as a `SystemService`

tomas-langer:
Closing this pull request as the approach has changed.

PR #758, opened 2019-06-05
Fixes support for application calls to flush() when using StreamingOutput

spericas:
Fixes support for application calls to flush() when using StreamingOutput.

### DIFF
--- a/webserver/jersey/src/test/java/io/helidon/webserver/jersey/JerseyExampleResource.java
+++ b/webserver/jersey/src/test/java/io/helidon/webserver/jersey/JerseyExampleResource.java
@@ -18,7 +18,9 @@ package io.helidon.webserver.jersey;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.stream.Collectors;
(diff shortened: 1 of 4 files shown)


webserver/jersey/src/test/java/io/helidon/webserver/jersey/JerseyExampleResource.java
```diff
@@ -212,4 +216,22 @@ public Response deleteNotFound(@Context UriInfo uriInfo) {
...
+    @GET
+    @Path("/streamingOutput")
+    @Produces("application/stream+json")
+    public StreamingOutput getHelloOutputStream() {
+        return out -> {
+                try {
+                    out.write(("{ value: \"first\" }\n").getBytes(StandardCharsets.UTF_8));
+                    out.flush();
+
+                    Thread.sleep(2000);     // wait before sending next chunk
```
ljnelson: Minor nit: I know what you're doing here, but it will slow down the CI/CD builds.

spericas: I'll reduce the wait time, I don't think it needs to be that long.

Issue #698, opened 2019-05-16
Create helidon context in gRPC server implementation


tomas-langer:
Add support for common `Context` to gRPC implementation.
It should at least register the `Tracer` and `SpanContext`. Also the context should be available when Security is invoked, so it can register additional instances

PR #769, opened 2019-06-10
Create helidon context in gRPC server implementation

thegridman:
This pull request is for issue #698.

This change adds a ability to set a Helidon `io.helidon.common.context.Context` when creating the gRPC server by calling the `context(Context context)` method on the `GrpcServerConfiguration.Builder` when creating the server configuration that will be used to configure the gRPC server.

The Thread factory used by the gRPC server thread pool creates threads that have the context set so that the context can be obtained using the static `io.helidon.common.context.Contexts.context()` method in call handling code. The context can also be obtained from the gRPC call context inside a gRPC service method using the `io.helidon.grpc.core.ContextKeys.HELIDON_CONTEXT.get()` method.

If tracing is enabled, the `Tracer` and the current `SpanContext` are set into the `Context` at the start of a gRPC server call so that they are available during handling of the call.

If security is enabled then the `SecurityContext` is also set into the `Context`.

### DIFF
--- a/grpc/server/src/main/java/io/helidon/grpc/server/GrpcServer.java
+++ b/grpc/server/src/main/java/io/helidon/grpc/server/GrpcServer.java
@@ -21,7 +21,7 @@ import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Supplier;
 
-import io.helidon.common.http.ContextualRegistry;
+import io.helidon.common.context.Context;
 import io.helidon.grpc.core.PriorityBag;
 
 import io.grpc.ServerInterceptor;
@@ -51,7 +51,7 @@ public interface GrpcServer {
      *
      * @return a server context
      */
-    ContextualRegistry context();
+    Context context();
 
     /**
      * Starts the server. Has no effect if server is running.
--- a/grpc/server/src/main/java/io/helidon/grpc/server/GrpcServer.java
+++ b/grpc/server/src/main/java/io/helidon/grpc/server/GrpcServer.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Supplier;
 
+import io.helidon.common.http.ContextualRegistry;
 import io.helidon.grpc.core.PriorityBag;
 
 import io.grpc.ServerInterceptor;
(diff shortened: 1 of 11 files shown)


grpc/server/src/main/java/io/helidon/grpc/server/GrpcServer.java
```diff
@@ -45,6 +46,13 @@
      */
     GrpcServerConfiguration configuration();
 
+    /**
+     * Gets a {@link GrpcServer} context.
+     *
+     * @return a server context
+     */
+    ContextualRegistry context();
```
tomas-langer: This should not use `ContextualRegistry` but `Context` - `ContextualRegistry` is used for backward compatibility in WebServer

grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerImpl.java
```diff
@@ -116,6 +128,8 @@
      */
     private Map<String, ServiceDescriptor> services = new ConcurrentHashMap<>();
 
+    private final ContextualRegistry contextualRegistry;
```
tomas-langer: Use `Context` not `ContextualRegistry`

grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerImpl.java
```diff
@@ -213,6 +229,11 @@ public GrpcServerConfiguration configuration() {
         return config;
     }
 
+    @Override
+    public ContextualRegistry context() {
```
tomas-langer: `Context` instead of `ContextualRegistry`

grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerImpl.java
```diff
@@ -480,8 +479,8 @@ protected SslContextBuilder sslContextBuilder(SslConfiguration sslConfig) {
                                                                      Metadata headers,
                                                                      ServerCallHandler<ReqT, RespT> next) {
 
-            Context context = Context.current().withValue(ContextKeys.HELIDON_CONTEXT, context());
-            return Contexts.interceptCall(context, call, headers, next);
+            io.grpc.Context context = io.grpc.Context.current().withValue(ContextKeys.HELIDON_CONTEXT, context());
```
tomas-langer: I think this is incorrect. Please make sure that each request runs in its own context instance.
I would expect that within a request your create a new context with the server context as its parent

thegridman:
I don't have permissions to merge; someone else will need to do it.

PR #773, opened 2019-06-11
WIP: Content-length optimization

spericas:
Added new flag to DataChunk to mark them as immutable. Updated and created some new tests.

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/DataChunk.java
+++ b/common/http/src/main/java/io/helidon/common/http/DataChunk.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -66,7 +66,19 @@ public interface DataChunk {
      * @return a reusable data chunk with no release callback
      */
     static DataChunk create(boolean flush, ByteBuffer data) {
-        return create(flush, data, Utils.EMPTY_RUNNABLE);
+        return create(flush, data, Utils.EMPTY_RUNNABLE, false);
+    }
+
+    /**
+     * Creates a reusable data chunk.
+     *
+     * @param flush a signal that chunk should be written and flushed from any cache if possible
+     * @param data  a data chunk. Should not be reused until {@code releaseCallback} is used
+     * @param immutable indicates underlying buffer is not reused
+     * @return a reusable data chunk with no release callback
+     */
+    static DataChunk create(boolean flush, ByteBuffer data, boolean immutable) {
+        return create(flush, data, Utils.EMPTY_RUNNABLE, immutable);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


common/http/src/main/java/io/helidon/common/http/DataChunk.java
```diff
@@ -194,4 +225,29 @@ default void release() {
...
+        return dup;
+    }
+
+    /**
+     * Returns {@code true} if the underlying byte buffer of this chunk is read
+     * only or {@code false} otherwise.
+     *
+     * @return Immutability outcome.
+     */
+    default boolean isImmutable() {
```
romain-grecourt: How about calling the flag readOnly instead ?

webserver/webserver/src/test/java/io/helidon/webserver/TransferEncodingTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Start the Web Server
+     *
+     * @param port the port on which to start the server; if less than 1,
+     * the port is dynamically selected
+     * @throws Exception in case of an error
+     */
+    private static void startServer(int port) throws Exception {
+        webServer = WebServer.create(
```
romain-grecourt: I know this looks inspired from PlainTest, but @tomas-langer is advocating for not starting actual servers in unit tests in favor of mocking.

IIRC the idea is to have tests that do start real servers to be under /tests/functional or /tests/integration.
I don't see much under /tests at the moment, so this would mean more work to get this going.

Perhaps you can get a pass for now until we setup the integration tests properly.
Thoughts ?

common/http/src/main/java/io/helidon/common/http/DataChunk.java
spericas: Sure, we can use read only. But for consistency with other methods in that class, it should probably be named `isReadOnly`.

webserver/webserver/src/test/java/io/helidon/webserver/TransferEncodingTest.java
spericas: I struggle with this all the time. Recently, for another issue, I created a new functional test called `context-propagation`, but this isn't a very efficient way of writing tests because, as of now, every functional test is there for essentially _one_ purpose. If we had something more generic (like webserver) that we could easily extend, it would be a much more efficient way of writing tests.

common/http/src/main/java/io/helidon/common/http/DataChunk.java
romain-grecourt: Right that's what I meant, sorry for the confusion.

webserver/webserver/src/test/java/io/helidon/webserver/TransferEncodingTest.java
romain-grecourt: I agree, we need to setup a module under tests for the webserver integration / functional tests.

olotenko:
https://github.com/oracle/helidon/blob/ca0e11f/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java#L227 - use of `limit()` is less correct than using `remaining()`.

PR #781, opened 2019-06-13
System services

### LINKED ISSUE #780: WebServer system services
We need to be able to extend WebServer to achieve:

1. Add objects to application wide `Context`
2. Intercept every request to the `WebServer`

Currently we have a problem achieving the second point, as every handle may finish the processing (e.g. calls `response.send()` and not `req.next()`).

Proposed solution:

- Create a `SystemService` interface that can be registered with server configuration
- Add support for loading these through Java service loader (default is not to load the services)
- Invoke the `SystemServices` before processing routing on server

This is marked as P1, as other features depend on such an approach (AccessLog support, Tracing refactoring)

tomas-langer:

Resolves #780

### DIFF
--- a/docs/src/main/docs/webserver/11_system-services.adoc
+++ b/docs/src/main/docs/webserver/11_system-services.adoc
@@ -1,0 +1,71 @@
+///////////////////////////////////////////////////////////////////////////////
+
+    Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+///////////////////////////////////////////////////////////////////////////////
+
+:description: Helidon Reactive Webserver Access Log
+:keywords: helidon, webserver, system service, extension
+
+= WebServer System Services
+
+System services are extensions of a WebServer that can:
+
+1. Register instances to the application `io.helidon.common.context.Context`
+2. Intercept the request and response without terminating it
+
+A system service can be configured explicitly:
+
+[source,java]
+.Example reading configuration and registering a system service
+----
+ServerConfiguration.builder(config.get("server"))
+                .register(AccessLogSystemService.create())
+                .build();
+----
+
+== Using Service Loader to load System Services
+System services can also be loaded from classpath using the Java service loader mechanism:
+
+[source,java]
+.Example reading configuration and loading system services from classpath
+----
+ServerConfiguration.builder(config.get("server"))
+                .useSystemServiceLoader(true)
+                .build();
+----
+
+Note that system loader has to be explicitly enabled, otherwise the services are not loaded.
+
+The order of services (even the ones registered explicitly) is based on the
+service priority, defined either by annotation `@Priority` or by implementing
+interface `Prioritized`.
+
+Service loader uses the class `io.helidon.webserver.spi.SystemServiceProvider` that
+supports additional configuration of system services.
+
+The configuration is expected under the config node provided to `ServerConfiguration.Builder`
+   under node `system-services`, key of the specific service is defined in that service provider.
+
+[source,yaml]
+.Example configuration of system services when using a Java service loader
+----
+server:
+  system-services:
+    access-log:
+        fromat: "common"
+----
+
+
--- a/docs/src/main/docs/webserver/11_system-services.adoc
+++ b/docs/src/main/docs/webserver/11_system-services.adoc
@@ -24,7 +24,10 @@
 System services are extensions of a WebServer that can:
 
 1. Register instances to the application `io.helidon.common.context.Context`
-2. Intercept the request and response without terminating it
+2. Intercept the request - register readers and writers
+3. Intercept the response - update response headers
+4. System services cannot terminate processing and send a response
+
 
 A system service can be configured explicitly:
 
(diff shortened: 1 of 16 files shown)


romain-grecourt:
If we make `SystemService` extend `Service` then we could update the webserver services for the media support in a backward compatible way. I thought you were going to do that, any reason you did not ?

spericas:
As I stated in the other issue, I'm a bit concerned that we are only thinking about these systems services as _request processors_ (actually "processor" may be a better name too). Cross-cutting concerns in system like Helidon should also, at least, include support for _response processors_. Are we only seeing _half_ of the picture?

docs/src/main/docs/webserver/11_system-services.adoc
```diff
@@ -0,0 +1,71 @@
...
+The configuration is expected under the config node provided to `ServerConfiguration.Builder`
+   under node `system-services`, key of the specific service is defined in that service provider.
+
+[source,yaml]
+.Example configuration of system services when using a Java service loader
+----
+server:
+  system-services:
+    access-log:
+        fromat: "common"
```
batsatt: typo: fromat --> format

webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
```diff
@@ -100,13 +110,223 @@ public void route(BareRequest bareRequest, BareResponse bareResponse) {
             RoutedRequest nextRequests = new RoutedRequest(bareRequest, response, webServer, crawler, errorHandlers, span);
             // only register the span context once on the top level request, as others are cloned from it
             nextRequests.context().register(span.context());
-            Contexts.runInContext(nextRequests.context(), (Runnable) nextRequests::next);
+
+            // we must execute system services
+            Contexts.runInContext(nextRequests.context(), () -> {
```
batsatt: If it will be common to have zero system services, perhaps this whole block should be inside a conditional to avoid adding request overhead.

tomas-langer: There is one line that is always executed `nextRequests.next()`

tomas-langer:
Extending a `Service` was originally planned, but then we (during our meeting) said that this should be executed with each and every requests, and as such it should not really support registering additional routes.
You should be able to stuff in the `handle` method that has access to request and response and can register appropriate readers and writers.
I am not sure about @spericas comment - you have access to both request and response, so what is the use case you mean?

spericas:
@tomas-langer Can you access a response _after_ it has been updated by the application's handler?

tomas-langer:
@spericas I understand. No you cannot do that with current design

webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
batsatt:
```
List<SystemService> systemServices = webServer.configuration().systemServices();
if (systemServices.isEmpty()) {
    Contexts.runInContext(nextRequests.context(), (Runnable) nextRequests::next);
} else {
    Contexts.runInContext(nextRequests.context(), () -> {
        ServerRequest systemRequest = systemRequest(nextRequests);
        ServerResponse systemResponse = systemResponse(response);
        for (SystemService service : systemServices) {
            service.handle(systemRequest, systemResponse);
        }
        nextRequests.next();
    });
}
```
I understand that the extra complexity is not ideal (and certainly unwarranted if the list will generally be non-empty), but I do worry about unnecessary overhead in request handling slowing us down.

tomas-langer: ok

tomas-langer: There is no difference in performance between your proposed solution and my PR. You are doing exactly the same decision, only outside of the `Context.runInContext`

tomas-langer:
I have added support for request and response processing.
I think the name `SystemService` is not good, as it implies operating system services. We should come up with a better one. Maybe `WebServerExtension`?
@batsatt I have modified `send` methods in WebServer as they were doing the same thing. This should not impact performance (maybe in a positive way, as the empty publisher no longer goes through reactive streams adapter)

batsatt:
Agreed that `SystemService` is not a good name. Jersey uses the term "filter", though it separates out requests and responses by class: `ContainerRequestFilter` and `ContainerResponseFilter`. 

This seems like a good term so long as it is possible for an implementation of `processRequest()` or `processResponse()` to throw an exception to terminate the request. 

Maybe `WebFilter` or `WebServerFilter`?

spericas:
I tend to agree with @batsatt. These are basically filters, and probably _should_ be allowed to terminate processing, exceptionally or not. When we had this discussion, I brought authentication as an example of early termination. 

@tomas-langer What's the rationale for not supporting those use cases? 

On the subject of interfaces, having single-method interfaces enables the use of lambdas, but that may not be a significant benefit here.

romain-grecourt:
Is such WebFilter any different than a normal Handler ?

tomas-langer:
I did not consider security as a use case for these, as the security configuration is path base - I actually need to register it per path.
Nevertheless it would be quite good to know it is going to be executed before the business logic...

The complexity would be quite high though - 
 1. a request filter (can terminate, registered for each and every path and method)
 2. a response filter (called right before data is written to the response)
 3. a web server context enhancer (update application context)
 4. a Service that can add processing per method and path

If we allow termination, we need to support request.next() - I wanted to avoid that as well.

If you want to allow all above, we should probably end up with 
 1. A Service - this allows registration per path and method for each request
 2. A "response" service - this allows registrations as above, only invoked before send (after business logic processing)
 3. An extension that can configure application wide context

We could probably start with 1. and 3., and see if there is a use case for 2.

tomas-langer:
Also there is already a "filter" in WebServer response, which is a DataChunk Processor

romain-grecourt:
Register the system services against in ServerConfiguration feels odd. How about doing it directly within the WebServer builder ?

tomas-langer:
Removing this pull request as this is a complex issue that we should handle for version 2.0.
I will update the issue.

Issue #759, opened 2019-06-06
gRPC missing module-info.java


tomas-langer:
gRPC modules are missing `module-info.java`

thegridman:
There is an issue with this because out of the box grpc-java has split package issues so you cannot use it with Java 9 modules. The solution suggested on the GitHub issues that mention this in grpc-java is to re-package grpc-java by shading into a single jar, with either an automatic module name or a `module-info.java` file added. 

To make this work in Helidon we would need to do something like introduce a new sub-module under the helidon-grpc-project that does this repackaging and produces a Helidon grpc-java module. This would be the only module that has a dependency on the real grpc-java jars. The rest of the Helidon modules that currently depend on grpc-java would depend on this new module instead and could then all have `module-info.java` files added.

It is not ideal but is the only solution available as grpc-java have no intention reorganizing their jars until maybe grpc-java 2.0.

thegridman:
After discussing the approach on a call with Tomas I have done the changes and created a PR https://github.com/oracle/helidon/pull/789

PR #789, opened 2019-06-17
Add module-info.java files to gRPC modules (issue #759)

thegridman:


grpc/io.grpc/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+    limitations under the License.
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
```
romain-grecourt: The indentation seems to be 2 instead of 4 here.

thegridman:
The reason for repackaging is because the grpc-java project is not compatible with Java 9 modules due to having split package issues. The grpc-java project are not fixing this and the suggested workaround is to do what we have done here and repackage some of the grpc-java jars into a single jar and add a module-info.

grpc/pom.xml
```diff
@@ -34,10 +34,11 @@
...
-    <modules>
-        <module>core</module>
-        <module>server</module>
-        <module>client</module>
-        <module>metrics</module>
-    </modules>
+  <modules>
+    <module>io.grpc</module>
+    <module>core</module>
+    <module>server</module>
```
romain-grecourt: This looks like 2 spaces indent, it should be 4.

thegridman:
Can someone look at whether this can be merged please.

romain-grecourt:
@barchetta ping.

PR #801, opened 2019-06-20
credentials support added to git config

### LINKED ISSUE #744: Git Configuration does not allow credentials
## Environment Details
* Helidon Version: 1.1

----------

## Problem Description
When using the GitConfigSource, there is no facility to specify credentials of the git repository. So only public git repositories appear supported. 

GitConfigSourceBuilder is final so does not allow us to customise or extend it. Any thoughts on how to proceed? We can implement the config service but I think this use case is common enough to be part of Helidon and the code change is very small. 

We just need to call the call the setCredentialProvider with some values on Line 165

>         PullCommand pull = git.pull()
                .setCredentialsProvider(cp)
                .setRebase(true);

Verdent:
Fixes #744

### DIFF
--- a/config/git/src/main/java/io/helidon/config/git/GitConfigSourceBuilder.java
+++ b/config/git/src/main/java/io/helidon/config/git/GitConfigSourceBuilder.java
@@ -71,7 +71,7 @@ public final class GitConfigSourceBuilder
     private static final String URI_KEY = "uri";
     private static final String BRANCH_KEY = "branch";
     private static final String DIRECTORY_KEY = "directory";
-    private static final String USER = "user";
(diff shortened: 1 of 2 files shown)


config/git/src/main/java/io/helidon/config/git/GitConfigSourceBuilder.java
```diff
@@ -68,17 +71,21 @@
     private static final String URI_KEY = "uri";
     private static final String BRANCH_KEY = "branch";
     private static final String DIRECTORY_KEY = "directory";
+    private static final String USER = "user";
```
tomas-langer: use `username` instead

PR #808, opened 2019-06-21
Update javadocs plugin. Make javadocs compatible with HTML5 

barchetta:
This PR updates our javadocs so that we can switch to Java 11. Specifically:

1. Upgrades the javadocs-plugin to 3.1.0
2. Updates our javadoc to be HTML5 compliant


pom.xml
```diff
@@ -270,6 +270,7 @@
                     <artifactId>maven-javadoc-plugin</artifactId>
                     <version>${version.plugin.javadoc}</version>
                     <configuration>
+                        <source>8</source>
```
ljnelson: This gives me pause.  How come?

barchetta: Without it you get module errors from javadoc about stuff referencing stuff in the automatic module.

PR #809, opened 2019-06-21
Tracing configuration module.

### LINKED ISSUE #461: Tracing support refactoring

tomas-langer:

Resolves #461


tomas-langer:
Good start for review should be the documentation:
https://github.com/oracle/helidon/blob/ee463a8e3e17cf940362f30b19f0174535bfd029/docs/src/main/docs/tracing/01_tracing.adoc

docs/src/main/docs/tracing/01_tracing.adoc
```diff
@@ -120,35 +120,171 @@ To create a custom span that is a child of the WebServer request:
...
+|`jersey-client-call`   |`http.status_code` |HTTP status code of client response
+|`jersey-client-call`   |`http.url`         |Full URL of the request (such as `http://localhost:8080/greet`)
+|===
+
+=== Traced spans configuration
+
+Each component and its spans can be configured using Config.
+The traced configuration has the following layers:
+
+- `EnvTracingConfig` - the overall configuration of traced components of Helidon
```
batsatt: Would it make sense to simply call this `TracingConfig`? The `Env` prefix is a little confusing.

batsatt: Oh, I see you are also introducing a `TracingConfiguration` class in `webserver`. And it looks like you've used the terms "global", "overall" and "application" elsewhere to refer to the `EnvTracingConfig`. Hmm. Lots of terms here to disentangle. Maybe:

1. Rename `TracingConfiguration` to `WebTracingConfig` (using the shorter suffix as used elsewhere).
2. Rename `EnvTracingConfig` to `TracingConfig` (or `AppTracingConfig`).

webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
```diff
@@ -353,7 +290,14 @@ public void next() {
                     RoutedResponse nextResponse = new RoutedResponse(response);
                     RoutedRequest nextRequest = new RoutedRequest(this, nextResponse, nextItem.path, errorHandlers);
                     LOGGER.finest(() -> "(reqID: " + requestId() + ") Routing next: " + nextItem.path);
-                    requestSpan.log(nextItem.handlerRoute.diagnosticEvent());
+                    Span span = span();
```
romain-grecourt: The span was a field and now has to be looked-up from the context every time.
Is the performance impact negligible ?

docs/src/main/docs/tracing/01_tracing.adoc
tomas-langer: `TracingConfiguration` is already used to configure tracing in gRPC server, so I wanted to be consistent with that. Nevertheless the names are not ideal for the other classes.

webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
tomas-langer: I hope so. Otherwise there is a problem, as we do not have the interceptor to correctly handle all the events in the new service...

docs/src/main/docs/tracing/01_tracing.adoc
batsatt: We could also rename the gRPC class, e.g. `GrpcTracingConfig` (or the full suffix if we want to do that everywhere).

tracing/jersey-client/src/main/java/io/helidon/tracing/jersey/client/ClientTracingFilter.java
```diff
@@ -153,15 +155,24 @@ public void filter(ClientRequestContext requestContext) {
             return;
         }
 
+        // also we may configure tracing through other means
+        SpanTracingConfig spanConfig = TracingConfigUtil.spanConfig("jax-rs", SPAN_OPERATION_NAME);
```
Verdent: I would say that "jax-rs" should be declared as constant.

webserver/webserver/src/main/java/io/helidon/webserver/RouteListRoutingRules.java
```diff
@@ -110,6 +110,23 @@ private boolean isEmpty() {
...
 
+    @Override
+    public RouteListRoutingRules register(TracingConfiguration tracingConfiguration) {
+        onNewWebServer(ws -> ws.context().register(tracingConfiguration.envConfig()));
+
+        Service[] services = {tracingConfiguration.service()};
+        Record record = new Record(null, services);
+
+        // need tracing service to be the very first one, as it must start the span before other handlers are invoked
+        if (records.isEmpty()) {
```
Verdent: This can be reduced only to records.add(0, record). If you are adding element to the first index of empty List, it will work the same as if you call add(record).

webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
tomas-langer: I checked the implementation and it is much better to use a qualifier for performance reasons (as the list registry is not very optimal - it iterates until it finds an object that is of the correct type).

tomas-langer: So I will refactor the code to have a qualified instance of SpanContext

docs/src/main/docs/tracing/01_tracing.adoc
tomas-langer: I have created a new class `GrpcTracingConfig` and kept the old one for backward compatibility marked as `Deprecated`

tracing/config/src/main/java/io/helidon/tracing/config/TracingConfigUtil.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param component component tracing this span
+     * @param spanName name of the span to trace
+     * @param defaultEnabled whether tracing should be enabled by default
+     * @return span configuration, including configuration of span logs
+     */
+    public static SpanTracingConfig spanConfig(String component, String spanName, boolean defaultEnabled) {
+        return Contexts.context()
+                .flatMap(ctx -> ctx.get(TracingConfig.class))
+                .map(traceableEnvironment -> traceableEnvironment.component(component, defaultEnabled))
```
batsatt: No longer an "environment", so maybe just `tracingConfig`?

tracing/config/src/main/java9/module-info.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Tracing configuration for traced components.
+ *
```
batsatt: Shouldn't this be `TracingConfig`?

webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
```diff
@@ -0,0 +1,307 @@
...
+import io.opentracing.tag.Tags;
+import io.opentracing.util.GlobalTracer;
+
+/**
+ * Tracing configuration for webserver.
+ * Tracing configuration has two components - an overall (application wide) {@link io.helidon.tracing.config.TracingConfig}
+ *  and a path specific {@link PathTracingConfig}.
+ */
+public abstract class WebTracingConfig {
+    /**
```
batsatt: Remove "environment".

webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
```diff
@@ -0,0 +1,307 @@
...
+     * Environment tracing configuration.
+     *
```
batsatt: Maybe just delete this.

tracing/config/src/main/java/io/helidon/tracing/config/TracingConfig.java
```diff
@@ -0,0 +1,227 @@
...
+     */
+    public static final TracingConfig ENABLED = TracingConfig.builder().build();
+    /**
+     * Traced conifg that is disabled for all components, spans and logs.
+     */
+    public static final TracingConfig DISABLED = TracingConfig.builder().enabled(false).build();
+
+    /**
+     * A new traced configuration.
+     *
```
batsatt: "environment" --> "configuration"

tracing/config/src/main/java/io/helidon/tracing/config/TracingConfig.java
```diff
@@ -0,0 +1,227 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import io.helidon.common.OptionalHelper;
+import io.helidon.config.Config;
+
+/**
+ * Configuration of traced environment.
```
batsatt: Since the "Env" prefix was removed, should eliminate uses of the term "environment" to avoid confusion. Perhaps just "Tracing configuration" or perhaps include a description of the scope, e.g. "global", "application" etc.

webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
tomas-langer: The method is needed, some documentation should be present. Reworded and added details

PR #810, opened 2019-06-21
Change build pipeline to Java 11

barchetta:
Updates our build pipeline from Java 9 to Java 11. This required upgrading spotbugs and making some spotbugs and javadoc fixes.

### DIFF
--- a/integrations/serviceconfiguration/serviceconfiguration-ucp-accs/src/main/javadoc/overview.html
+++ b/integrations/serviceconfiguration/serviceconfiguration-ucp-accs/src/main/javadoc/overview.html
@@ -24,5 +24,5 @@
      target="_parent">when running on the Oracle Application Container
      Cloud Service</a> system.</p>
 
-    @see com.oracle.service.configuration.ucp.accs.UCPServiceConfigurationACCSProvider
+    @see io.helidon.service.configuration.ucp.accs.UCPServiceConfigurationACCSProvider
 </body>
--- a/pom.xml
+++ b/pom.xml
@@ -302,6 +302,7 @@
                         <additionalJOptions combine.children="append">
                             <JOption>-J-Dhttp.agent=maven-javadoc-plugin</JOption>
                         </additionalJOptions>
+                        <additionalOptions>--frames</additionalOptions>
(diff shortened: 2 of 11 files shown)


integrations/serviceconfiguration/serviceconfiguration-ucp-accs/src/main/javadoc/overview.html
```diff
@@ -24,5 +24,5 @@
      target="_parent">when running on the Oracle Application Container
      Cloud Service</a> system.</p>
 
-    @see com.oracle.service.configuration.ucp.accs.UCPServiceConfigurationACCSProvider
+    @see io.helidon.service.configuration.ucp.accs.UCPServiceConfigurationACCSProvider
```
ljnelson: Whoops; sorry about that!

pom.xml
```diff
@@ -302,6 +302,7 @@
                         <additionalJOptions combine.children="append">
                             <JOption>-J-Dhttp.agent=maven-javadoc-plugin</JOption>
                         </additionalJOptions>
+                        <additionalOptions>--frames</additionalOptions>
```
tjquinno: So this is just for the Java 8 variant of the build. I know that frames have disappeared from the 11 version of JavaDoc.

integrations/serviceconfiguration/serviceconfiguration-ucp-accs/src/main/javadoc/overview.html
barchetta: no problemo!

pom.xml
barchetta: Java 11 still supports frames, but the default is no-frames (https://bugs.openjdk.java.net/browse/JDK-8196202)  So this gives us frames with Java 11. And this delightful warning:
```
[WARNING] Javadoc Warnings
[WARNING] javadoc: warning - You have specified to generate frames, by using the --frames option.
[WARNING] The default is currently to not generate frames and the support for
[WARNING] frames will be removed in a future release.
[WARNING] To suppress this warning, remove the --frames option and avoid the use of frames.
```

Issue #639, opened 2019-05-02
Support analog to extended persistence contexts


ljnelson:
The [soon-to-be-available JPA/JTA support](https://github.com/oracle/helidon/tree/master/integrations/cdi/jpa-weld) does not have any support currently for any kind of persistence context propagation (all `EntityManager`s are effectively `@Dependent`-scoped).  It needs to in order to be fully compliant with the spirit of container-managed JPA.

ljnelson:
Implemented by #812.

PR #812, opened 2019-06-24
Addresses #639

ljnelson:


ljnelson:
As you can see, this is a draft pull request.  Rest assured that when all of its commits have been squashed and merged and the final commit is ready to go it will have a complete and detailed commit message, like all of my other PRs.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/DelegatingEntityManager.java
```diff
@@ -0,0 +1,434 @@
...
+     * occurs
+     *
+     * @see #acquireDelegate()
+     *
+     * @see #DelegatingEntityManager(EntityManager)
+     */
+    protected EntityManager delegate() {
+        final EntityManager returnValue;
+        if (this.delegate == null) {
+            returnValue = this.acquireDelegate();
```
tjquinno: I see that the `acquireDelegate` JavaDoc explains that its implementation must be fast because it is invoked each time a delegated method is invoked (if no delegate was passed to the constructor initially). So it seems intentional that we don't invoke `acquireDelegate` once (if there is none) and then cache and reuse that one subsequently. I'm sure there's a good reason; I'm just double-checking to make sure this is what we want.

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCountedExtension.java
```diff
@@ -0,0 +1,169 @@
...
+    /*
+     * Utility methods.
+     */
+
+
+    private <T> Producer<T> createReferenceCountingProducer(final Producer<T> delegate,
+                                                            final BeanManager beanManager) {
+        final Producer<T> returnValue = new DelegatingProducer<T>(delegate) {
+
+                private volatile Set<InjectionPoint> referenceCountedInjectionPoints;
```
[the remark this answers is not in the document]
ljnelson: My intention here was:

1. At line 115, atomically grab a local reference to the instance field.
2. Stuff that local instance full of stuff.
3. At line 142, atomically set the local reference back to the instance field.

It is possible that two threads could do this work at the same time.  Only one of them will win.  Unless CDI is really broken, they'll both end up doing exactly the same work.  Locking didn't seem worth it.

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCountedContext.java
```diff
@@ -0,0 +1,461 @@
...
+    /**
+     * A coupling of an object (a contextual instance), the {@link
+     * CreationalContext} in effect when it was created, the {@link
+     * Contextual} that {@linkplain
+     * Contextual#create(CreationalContext) created} it, and a
+     * reference count.
+     *
+     * <h2>Thread Safety</h2>
+     *
+     * <p>Instances of this class are <strong>not</strong> safe for
```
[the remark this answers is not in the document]
ljnelson: That's right, or at least that's my intention.  All of these (`private`) `Instance` instances will be stored in the (thread-local) `ALL_INSTANCES` map.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/DelegatingEntityManager.java
ljnelson: Double-checking is always appreciated.  Yes, the JPA specification basically spells this out.  In many different cases, you are required to hand out an `EntityManager` "shell", proxy, whatever you want to call it—and then on every invocation you have to do things like check to see if you're in a transaction, etc.

You inadvertently raise a good point which is that in one version of this PR I had a couple of use cases where I would pass a delegate to the constructor.  Those cases died off gradually—offhand I can't remember if I still use that constructor/feature at all.  I'll check.

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCountedExtension.java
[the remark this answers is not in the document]
tjquinno: Got it. I didn't study the body of the critical region enough to see that concurrent threads should compute the same result.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/DelegatingEntityManager.java
ljnelson: (Ah, I do use still use that feature in the `NonTransactionalEntityManager` delegate, whose sole job is to throw `TransactionRequiredException`s from certain methods.)

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/ContainerManaged.java
```diff
@@ -0,0 +1,60 @@
...
+ * defined by the JPA specification.
+ */
+@Qualifier
+@Retention(RetentionPolicy.RUNTIME)
+@Target({}) // can only be programmatically added
+@interface ContainerManaged {
+
+    /**
+     * An {@link AnnotationLiteral} that implements {@link
+     * CdiTransactionScoped}.
```
batsatt: Copy-and-paste error: should be ContainerManaged.

ljnelson: Thank you!

ljnelson: Fixed.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/Extended.java
```diff
@@ -0,0 +1,62 @@
...
+ * are associated with a JTA transaction.
+ *
+ * <p>This qualifier must not be combined with {@link
+ * CdiTransactionScoped}, {@link JpaTransactionScoped} or {@link
+ * NonTransactional}.</p>
+ */
+@Qualifier
+@Retention(RetentionPolicy.RUNTIME)
+@Target({}) // can only be programmatically added
+@interface Extended {
```
batsatt: Name seems too generic; based on the class description, shouldn't it have "transaction" in the name?

batsatt: Oh, just noticed the `@Target` comment. If this is an implementation detail, not exposed to users, then the name is obviously less relevant.

ljnelson: For the record, I think "extended" is awful, but it is standard JPA-speak.  An extended persistence context is one that spans transactions.  My pattern here was to use the [short name of the analogous Java EE/JPA annotation](https://javaee.github.io/javaee-spec/javadocs/javax/persistence/PersistenceContextType.html#EXTENDED).  And you are correct this is an implementation detail.

batsatt: gotcha

integrations/cdi/jpa-cdi/src/test/java/io/helidon/integrations/cdi/jpa/TestMessages.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package io.helidon.integrations.cdi.jpa;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+final class TestMessges {
```
batsatt: typo

ljnelson: Fixed

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCountedContext.java
```diff
@@ -0,0 +1,474 @@
...
+     * Static fields.
+     */
+
+
+    /**
+     * A {@link ThreadLocal} containing a {@link Map} whose keys are
+     * instances of this class, and whose values are {@link Map}s
+     * indexing {@link Instance}s by {@link Contextual} instances.
+     */
+    private static final ThreadLocal<Map<ReferenceCountedContext, Map<Contextual<?>, Instance<?>>>> ALL_INSTANCES =
```
batsatt: I'm struggling to understand the relationship between `ReferenceCountedContext` instances and thread specific values. My initial reaction to this `ThreadLocal` was to wonder why the `Map<Contextual<?>, Instance<?>>` cannot just be a field of the `ReferenceCountedContext` itself.

I assume that there really is a 1:N relationship, but there are too many indirections going on for my tired brain to see it...

ljnelson: Yes.  Somewhere I read and internalized that `ThreadLocal`s should _always_ be `static`—like _always_, even at the expense of a design like this.

If we hold that constant for the moment, then the key to the `Map` held by the `ThreadLocal` is simply a `ReferenceCountedContext` instance.  In normal circumstances, there will only be one entry (in a CDI container there is normally only one instance of a `Context` implementation).

What you are seeing here is the idiom suggested by those championing the use of `static` `ThreadLocal`s even when logically their contents are supposed to be affiliated with a particular instance.

So what we're left with is: a thread-safe association between a `ReferenceCountedContext` (usually just one) and a `Map` that indexes `Instance`s by `Contextual`s (CDI `Bean`s).

You are correct that logically speaking a `Context` instance, should "have" a single `Map` that indexes `Instance`s by `Contextual`s—per thread.

See http://www.0xcafefeed.com/2004/06/of-non-static-threadlocals-and-memory/ for one arbitrarily suggested discussion about `static` and non-`static` `ThreadLocal`s that influenced my design here.

batsatt: Yes, I always make `ThreadLocal` instances `static`. That wasn't my confusion here, but I think you answered the question: a single `ReferenceCountedContext` instance can be shared across threads, but the state must be thread-specific.

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCounted.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.enterprise.context.NormalScope;
+
+/**
+ * An annotation indicating the bean that is annotated belongs to the
```
batsatt: Seems like this needs some explanation of why/when this should be used.

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCountedContext.java
ljnelson: Exactly right.

integrations/cdi/common-cdi/reference-counted-context/src/main/java/io/helidon/integrations/cdi/referencecountedcontext/ReferenceCounted.java
ljnelson: Fixed.

PR #782, opened 2019-06-13
Test with Jersey Release Candidate.

tomas-langer:
Signed-off-by: Tomas Langer <tomas.langer@oracle.com>

Issue #793, opened 2019-06-18
Cannot run HelidonSE modular with HealthCheckSupport


tomas-langer:
When using `module-info.java`, the mdoule `io.helidon.health.checks` cannot be used.

```
Error occurred during initialization of boot layer
java.lang.module.FindException: Module microprofile.config.api not found, required by io.helidon.health.checks
```

PR #813, opened 2019-06-25
Upgrade Jersey to 2.29

tomas-langer:
Resolves #782


tomas-langer:
Also resolves #793

tomas-langer:
I had to create modules for Jersey to correctly exclude jakarta and include javax.

PR #824, opened 2019-07-02
Support for a gitlab pipeline

barchetta:


barchetta:
Let's go with "fix-" for now. If consensus arises to change it later we can .

PR #826, opened 2019-07-03
MP Opentracing Implementation

tomas-langer:

### DIFF
--- a/jersey/common/src/main/java/io/helidon/jersey/common/InvokedResourceImpl.java
+++ b/jersey/common/src/main/java/io/helidon/jersey/common/InvokedResourceImpl.java
@@ -194,8 +194,8 @@ final class InvokedResourceImpl implements InvokedResource {
                 return false;
             }
             ClassAnnotationKey that = (ClassAnnotationKey) o;
-            return annotationClass.equals(that.annotationClass) &&
-                    handlingClass.equals(that.handlingClass);
+            return annotationClass.equals(that.annotationClass)
+                    && handlingClass.equals(that.handlingClass);
         }
 
         @Override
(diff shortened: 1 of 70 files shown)


jersey/common/src/main/java/io/helidon/jersey/common/InvokedResourceImpl.java
```diff
@@ -0,0 +1,316 @@
...
+
+import org.glassfish.jersey.server.ExtendedUriInfo;
+import org.glassfish.jersey.server.model.Invocable;
+import org.glassfish.jersey.server.model.ResourceMethod;
+
+final class InvokedResourceImpl implements InvokedResource {
+    private static final ConcurrentHashMap<MethodAnnotationKey, Optional<? extends Annotation>> METHOD_ANNOTATION_CACHE =
+            new ConcurrentHashMap<>();
+    private static final ConcurrentHashMap<ClassAnnotationKey, Optional<? extends Annotation>> CLASS_ANNOTATION_CACHE =
+            new ConcurrentHashMap<>();
```
spericas: Do this caches every shrink? Could this be an issue in long-running services?

microprofile/tests/tck/tck-opentracing/src/test/java/io/helidon/microprofile/opentracing/tck/UrlResourceProvider.java
```diff
@@ -0,0 +1,44 @@
...
+            return URI.create("http://localhost:8080/").toURL();
+        } catch (MalformedURLException e) {
+            return null;
+        }
+    }
+
+    @Override
+    public boolean canProvide(Class<?> type) {
+        return type.isAssignableFrom(URL.class);
+    }
```
spericas: This assignment is probably backwards. It states that _type_ is a supertype of _URL_. Also _URL_ is a final class.

tomas-langer: yep, will fix

jersey/common/src/main/java/io/helidon/jersey/common/InvokedResourceImpl.java
tomas-langer: These cache keys are class and method combination. Is there something generated in Jersey? Otherwise this is a finite set of keys depending on number of Resource classes and methods.

spericas: "Finite" numbers can still be very large :) But this cache's size should have a small upper bound.

PR #845, opened 2019-07-17
Corrects traversal logic for MpConfig when acquiring configuration property names

### LINKED ISSUE #844: MpConfig will never include Helidon Config property names in the return value of getPropertyNames()
`MpConfig` as a MicroProfile `Config` implementation acquires and caches all property names at construction time for returning later from its `getPropertyNames()` method.  This is an area of the MicroProfile Config specification that is undefined and untested by the TCK (must it cache? must it not cache? is caching permitted? prohibited?) so this is probably fine in itself.

To do so, it interrogates its MicroProfile `ConfigSource`s for _their_ property names, and concatenates that set together with what is hoped is the full key names of all leaf nodes in a Helidon `Config` tree.  The combined set should represent all property names known to the `MpConfig` instance at construction time.

(You can see where `application.yaml` is added as a Helidon configuration source after all the MicroProfile `ConfigSource`s here:

https://github.com/oracle/helidon/blob/173f2c07161c27f4548a77ec37ec154e0c6e0160/microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfigBuilder.java#L143-L147

)

The MicroProfile `ConfigSource` interrogation works fine.  The Helidon interrogation does not; its code can be found here:

https://github.com/oracle/helidon/blob/173f2c07161c27f4548a77ec37ec154e0c6e0160/microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfig.java#L78-L80

What I think is happening is that the `Predicate` that is supplied to `traverse` is `Config::isLeaf`, which I think returns `false` for the root node of the tree, and hence it appears that that node and all of its descendants are ignored.

Another way to put this is when that predicate is `Config::isLeaf`, and `traverse` is supplied with it, then `traverseSubNodes` is never entered in the code excerpt below, even when, for example, `application.yaml` containing valid YAML properties is found and loaded by the Helidon Config subsystem:

https://github.com/oracle/helidon/blob/173f2c07161c27f4548a77ec37ec154e0c6e0160/config/config/src/main/java/io/helidon/config/ConfigComplexImpl.java#L57-L79

ljnelson:
Fixes #844.

### DIFF
--- a/microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfig.java
+++ b/microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfig.java
@@ -73,12 +73,13 @@ public final class MpConfig implements org.eclipse.microprofile.config.Config {
         this.mpConfigSources = mpConfigSources;
 
         this.propertyNames =
-                Stream.concat(mpConfigSources.stream()
-                                      .flatMap(cs -> cs.getPropertyNames().stream()),
-                              config.traverse(Config::isLeaf)
-                                      .map(Config::key)
-                                      .map(Config.Key::toString))
-                        .collect(Collectors.toSet());
+            Stream.concat(mpConfigSources.stream()
+                              .flatMap(cs -> cs.getPropertyNames().stream()),
+                          config.traverse()
+                              .filter(Config::isLeaf)
+                              .map(Config::key)
+                              .map(Config.Key::toString))
+                .collect(Collectors.toSet());
 
         this.converters = converters;
         this.helidonConverter = new AtomicReference<>();


microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfig.java
```diff
@@ -73,12 +73,13 @@
...
         this.propertyNames =
-                Stream.concat(mpConfigSources.stream()
-                                      .flatMap(cs -> cs.getPropertyNames().stream()),
-                              config.traverse(Config::isLeaf)
-                                      .map(Config::key)
-                                      .map(Config.Key::toString))
-                        .collect(Collectors.toSet());
+            Stream.concat(mpConfigSources.stream()
+                              .flatMap(cs -> cs.getPropertyNames().stream()),
+                          config.traverse()
```
tjquinno: I think you correctly described the intent of the original code -- harvest the leaves -- but it was overly aggressive in discarding nodes. But does the proposed change harvest _all_ keys, not just leaf ones? Is that OK?

ljnelson: The original code never allowed the subsquent `map` operations to apply to anything.  The fix now makes it such that the `map` operations apply only to leaf nodes, which is what we want.

tjquinno: Oh, rats, I missed the new `.filter` invocation. Sheesh.

PR #847, opened 2019-07-18
Enables support of application.yaml use case in data source injection extensions

ljnelson:
This PR changes some internals of `AbstractDataSourceExtension` to acquire configuration so that Helidon-specific post-MicroProfile-Config-initialization property names are recognized, enabling a user to configure data sources out of the box using `application.yaml`.


ljnelson:
Fixes #847.

Issue #859, opened 2019-07-26
Update quickstart-se to use .exceptionally for good practice


camerondurham:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon SE
* JDK version: 1.8.0_211
* OS: Windows 10
* Maven version: 3.6.1

----------

## `PUT` requests lags with no message body

The quickstart should handle `PUT` requests with no message body with an `exceptionally(...)` block.

### Steps to reproduce errors

```
java -jar target/quickstart-se.jar
curl -X PUT -H "Content-Type: application/json"  http://localhost:8080/greet/greeting
```

## Output from `mvn package`
```
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Non-resolvable import POM: Could not find artifact io.helidon:helidon-bom:pom:1.2.0-SNAPSHOT @ line 126, column 25
[ERROR] 'dependencies.dependency.version' for io.helidon.bundles:helidon-bundles-webserver:jar is missing. @ line 147, column 21
[ERROR] 'dependencies.dependency.version' for io.helidon.config:helidon-config-yaml:jar is missing. @ line 151, column 21
[ERROR] 'dependencies.dependency.version' for io.helidon.health:helidon-health:jar is missing. @ line 155, column 21g. @ line 155, column 21                                                                     missing. @ line 159, column 21
[ERROR] 'dependencies.dependency.version' for io.helidon.health:helidon-health-checks:jar ising. @ line 163, column 21 missing. @ line 159, column 21
[ERROR] 'dependencies.dependency.version' for io.helidon.metrics:helidon-metrics:jar is missing. @ line 163, column 21
```

### Steps to reproduce `pom.xml` errors

```
git clone https://github.com/oracle/helidon.git
cd helidon/examples/quickstarts/helidon-quickstart-se
mvn package
```

camerondurham:
Resolved by #860

PR #860, opened 2019-07-26
#859 Add error-handling best practice in quickstart-se

camerondurham:
Closes my issue #859 in `examples/quickstarts/helidon-quickstart-se`

### DIFF
--- a/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
+++ b/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
@@ -108,11 +108,20 @@ public class GreetService implements Service {
 
     private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
 
-         if (ex.getCause() instanceof JsonParsingException) {
+         if (ex.getCause() instanceof JsonParsingException && 
+             ex.getMessage().contains("Invalid token")) {
 
-            LOGGER.log(Level.FINE, "Error parsing JSON body", ex);
+            LOGGER.log(Level.FINE, "Invalid type in JSON object", ex);
             JsonObject jsonErrorObject = JSON.createObjectBuilder()
-                .add("error", "Error parsing JSON body")
+                .add("error", "Invalid type in JSON object")
+                .build();
+            response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
+         } else if (ex.getCause() instanceof JsonParsingException && 
+             ex.getMessage().contains("Unexpected char")) {
+
+            LOGGER.log(Level.FINE, "Invalid JSON format", ex);
+            JsonObject jsonErrorObject = JSON.createObjectBuilder()
+                .add("error", "Invalid JSON format")
                 .build();
             response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
         } else if (ex.getCause() instanceof JsonException){
--- a/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
+++ b/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
@@ -18,6 +18,8 @@ package io.helidon.examples.quickstart.se;
 
 import java.util.Collections;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Logger;
+import java.util.logging.Level;
 
 import javax.json.Json;
 import javax.json.JsonBuilderFactory;
@@ -54,6 +56,8 @@ public class GreetService implements Service {
     private final AtomicReference<String> greeting = new AtomicReference<>();
 
     private static final JsonBuilderFactory JSON = Json.createBuilderFactory(Collections.emptyMap());
+
+    private static final Logger LOGGER = Logger.getLogger(GreetService.class.getName());
 
     GreetService(Config config) {
         greeting.set(config.get("app.greeting").asString().orElse("Ciao"));
@@ -101,6 +105,27 @@ public class GreetService implements Service {
         response.send(returnObject);
     }
 
+    private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
+                if (ex.getCause() instanceof JsonException){
+
+                LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
+                    String error = String.format("Error with JSON: %s", ex.getCause().toString());
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
+                        .add("error", error)
+                        .build();
+                    response.status(Http.Status.BAD_REQUEST_400).send(jsonErrorObject);
+                } else {
+
+                LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder() .add("error", ex.getLocalizedMessage()) .build();
+                    response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
+                    request.next(ex);
+                }
+
+                LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
+                return null;
+    }
+
     private void updateGreetingFromJson(JsonObject jo, ServerResponse response) {
 
         if (!jo.containsKey("greeting")) {
@@ -125,24 +150,7 @@ public class GreetService implements Service {
                                        ServerResponse response) {
         request.content().as(JsonObject.class)
             .thenAccept(jo -> updateGreetingFromJson(jo, response))
-            .exceptionally((Throwable ex) -> {
-                if (ex.getCause() instanceof JsonException){
-                    String error = String.format("Error with JSON: %s", ex.getCause().toString());
-                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
-                        .add("error", error)
-                        .build();
-                    response.status(Http.Status.BAD_REQUEST_400).send(jsonErrorObject);
-                } else {
-                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
-                        .add("error", ex.getLocalizedMessage())
-                        .build();
-                    request.next(ex);
-                }
-
-                ex.printStackTrace();
-
-                return null;
-            });
+            .exceptionally(ex -> processErrors(ex, request, response));
     }
 
 }
--- a/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
+++ b/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
@@ -108,16 +108,16 @@ public class GreetService implements Service {
 
     private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
 
-         if (ex.getCause() instanceof JsonParsingException && 
-             ex.getMessage().contains("Invalid token")) {
+         if (ex.getCause() instanceof JsonParsingException
+             && ex.getMessage().contains("Invalid token")) {
 
             LOGGER.log(Level.FINE, "Invalid type in JSON object", ex);
             JsonObject jsonErrorObject = JSON.createObjectBuilder()
                 .add("error", "Invalid type in JSON object")
                 .build();
             response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
-         } else if (ex.getCause() instanceof JsonParsingException && 
-             ex.getMessage().contains("Unexpected char")) {
+         } else if (ex.getCause() instanceof JsonParsingException
+             && ex.getMessage().contains("Unexpected char")) {
 
             LOGGER.log(Level.FINE, "Invalid JSON format", ex);
             JsonObject jsonErrorObject = JSON.createObjectBuilder()
--- a/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
+++ b/examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
@@ -107,6 +107,7 @@ public class GreetService implements Service {
     }
 
     private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
+
          if (ex.getCause() instanceof JsonParsingException) {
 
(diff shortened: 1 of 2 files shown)


romain-grecourt:
Hi @camerondurham, the version change part of your PR is not valid.
I'll let other people chime in regarding the exception handling change.

The examples in the master branch depend on the current SNAPSHOT version and thus require a full build.
If you wish to try-out examples for a given release, you can follow the instructions documented here: https://github.com/oracle/helidon/blob/master/examples/README.md

camerondurham:
HI @romain-grecourt , thank you for the update. Should I instead remove this request and make only the `.exceptionally(...)` fix?

romain-grecourt:
Yes, go ahead.

camerondurham:
@romain-grecourt My PR now only updates the quickstart.

romain-grecourt:
There are a couple of issues with your suggested change.

1. If your submit invalid JSON, it would also return an 400 with "empty body"
2. A good practice would be to show the use of `request.next(throwable)` to delegate the error to error handlers.

Here is what I came up with:

```java
        request.content().as(JsonObject.class)
                .thenAccept(jo -> updateGreetingFromJson(jo, response))
                .exceptionally((Throwable ex) -> {
                    if (ex.getCause() instanceof JsonException) {
                        request.next(new BadRequestException("Invalid JSON",
                                (JsonException) ex.getCause()));
                    } else {
                        request.next(ex);
                    }
                    return null;
                });
```

camerondurham:
Thanks for pointing out those issues. Of course, I agree that your change is much more effective. Sorry for being unfamiliar with this process, but would you be making this change yourself or should I modify my PR? @romain-grecourt

romain-grecourt:
Ideally you'd make this change in your PR so that you could get a Helidon contrib under your Belt.

romain-grecourt:
@barchetta @tomas-langer Any thoughts on the suggested changes ?

camerondurham:
Thanks, that sounds great. I'm adding @namanshenoy to my fork since he's another Oracle employee I worked with on this.

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -122,7 +122,14 @@ private void updateGreetingFromJson(JsonObject jo, ServerResponse response) {
      */
     private void updateGreetingHandler(ServerRequest request,
                                        ServerResponse response) {
-        request.content().as(JsonObject.class).thenAccept(jo -> updateGreetingFromJson(jo, response));
+        request.content()
+               .as(JsonObject.class)
+               .thenAccept(jo -> updateGreetingFromJson(jo, response))
+               .exceptionally(t -> {
+                   JsonObject jsonErrorObject = JSON.createObjectBuilder().add("error", "No body provided").build();
```
tomas-langer: The `exceptionally` branch is indeed needed, though it should not disregard the `throwable` (or it should have a more general content).

It just prints "No body provided` and `BAD_REQUEST` status - yet the throwable may be a `NullPointerException`, `CancelledException` or basically any other uncaught exception thrown while processing the future, or an explicitly set throwable when completing the future exceptionally. 


Options:

1. Send "Internal server error" (500) and error with text "request failed" or similar and log the message
2. Check exception type (you would need to actually check the cause, as CompletionStage exceptions are wrapped) and create a customized message and status code depending on the failure; and log the message
3. Use the exception provided (`t`) and write its stacktrace to the response - this is only for debugging/development purposes (and should be documented so)

tomas-langer: Hi, you have marked this as resolved, though the code is not changed. I think both @romain-grecourt and me requested a very similar change.

camerondurham: Hi @tomas-langer, thanks for mentioning this. I accidentally resolved your request and will make the changes as soon as possible. Thanks for being patient with my first PR.

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -47,82 +48,101 @@
...
-    /**
-     * Set the greeting to use in future messages.
-     * @param request the server request
-     * @param response the server response
-     */
-    private void updateGreetingHandler(ServerRequest request,
-                                       ServerResponse response) {
-        request.content().as(JsonObject.class).thenAccept(jo -> updateGreetingFromJson(jo, response));
-    }
+		/**
```
romain-grecourt: Please avoid refactoring the entire file when not necessary.
Also make sure to use 4 spaces indent.

camerondurham: Thanks, I'll minimize the changes and use 4 space indent!

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -122,7 +123,26 @@ private void updateGreetingFromJson(JsonObject jo, ServerResponse response) {
...
+                        .build();
+                    response.status(Http.Status.BAD_REQUEST_400).send(jsonErrorObject);
+                } else {
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
+                        .add("error", ex.getLocalizedMessage())
+                        .build();
+                    request.next(ex);
+                }
+
+                ex.printStackTrace();
```
romain-grecourt: Exception.printStackTrace() should be avoided, instead this should be done with a logger. There is no logger in this class, but it's a good opportunity to add one: `private static final Logger LOGGER = Logger.getLogger(GreetService.class.getName())`

Also, I think it makes sense the create a separate method for the exception handling to make it re-usable. Even though here it's only used once, it shows a good practice as part of the quickstart.

E.g.

```java
    private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
        if (ex.getCause() instanceof JsonException) {
            String error = String.format("Error with JSON: %s", ex.getCause().toString());
            JsonObject jsonErrorObject = JSON.createObjectBuilder()
                    .add("error", error)
                    .build();
            response.status(Http.Status.BAD_REQUEST_400).send(jsonErrorObject);
        } else {
            JsonObject jsonErrorObject = JSON.createObjectBuilder()
                    .add("error", ex.getLocalizedMessage())
                    .build();
            response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
            request.next(ex);
        }
        LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
        return null;
    }
```

```java
    private void updateGreetingHandler(ServerRequest request,
                                       ServerResponse response) {
        request.content().as(JsonObject.class)
                .thenAccept(jo -> updateGreetingFromJson(jo, response))
                .exceptionally(ex -> processErrors(ex, request, response));
    }
```

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -122,7 +123,26 @@ private void updateGreetingFromJson(JsonObject jo, ServerResponse response) {
...
+            .thenAccept(jo -> updateGreetingFromJson(jo, response))
+            .exceptionally((Throwable ex) -> {
+                if (ex.getCause() instanceof JsonException){
+                    String error = String.format("Error with JSON: %s", ex.getCause().toString());
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
+                        .add("error", error)
+                        .build();
+                    response.status(Http.Status.BAD_REQUEST_400).send(jsonErrorObject);
+                } else {
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
```
romain-grecourt: jsonErrorObject is not used.

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
camerondurham: My latest commit addresses this change. Is sending the error message sufficient or should I add a more customized message? Thanks @romain-grecourt !

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -100,6 +105,27 @@ private void sendResponse(ServerResponse response, String name) {
         response.send(returnObject);
     }
 
+    private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
+                if (ex.getCause() instanceof JsonException){
+
+                LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
```
romain-grecourt: This is not indented properly.

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -100,6 +105,27 @@ private void sendResponse(ServerResponse response, String name) {
...
+                    String error = String.format("Error with JSON: %s", ex.getCause().toString());
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder()
+                        .add("error", error)
+                        .build();
+                    response.status(Http.Status.BAD_REQUEST_400).send(jsonErrorObject);
+                } else {
+
+                LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
```
romain-grecourt: This is not indented properly.

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -100,6 +105,27 @@ private void sendResponse(ServerResponse response, String name) {
...
+                    JsonObject jsonErrorObject = JSON.createObjectBuilder() .add("error", ex.getLocalizedMessage()) .build();
+                    response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
+                    request.next(ex);
+                }
+
+                LOGGER.log(Level.FINE, "Exception occurred while processing payload", ex);
```
romain-grecourt: You are effectively logging the exception twice. If the log message is not customized based on the exception type, then you can remove the first 2 and leave this one.

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
romain-grecourt: I'm not sure about sending the exception message back in the response body, I'll let @tomas-langer share his opinion about this.

I have a few more comments about the current changes, see my next review.

romain-grecourt:
@tomas-langer What do you think about returning the exception message in the response body ?

examples/quickstarts/helidon-quickstart-se/src/main/java/io/helidon/examples/quickstart/se/GreetService.java
```diff
@@ -100,6 +106,26 @@ private void sendResponse(ServerResponse response, String name) {
...
+         if (ex.getCause() instanceof JsonParsingException) {
+
+            LOGGER.log(Level.FINE, "Error parsing JSON body", ex);
+            JsonObject jsonErrorObject = JSON.createObjectBuilder()
+                .add("error", "Error parsing JSON body")
+                .build();
+            response.status(Http.Status.INTERNAL_SERVER_ERROR_500).send(jsonErrorObject);
+        } else if (ex.getCause() instanceof JsonException){
+
+            LOGGER.log(Level.FINE, "Empty request body", ex);
```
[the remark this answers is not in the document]
camerondurham: @romain-grecourt Thanks for the good input. Sending a JSON Array throws a `ClassCastException`. If you send a JSON object with multiple keys, objects or arrays instead of strings, a `JsonParsingException` is thrown.

Right now, I'm still sending the same parsing message for these since the exception details are either `Unexpected char ...` or `Invalid token=...`. I can work on adding more custom messages depending on the parsing error if it's more appropriate.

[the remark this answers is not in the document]
camerondurham: I'm sending minimally more informative messages for `Invalid token` errors vs `Unexpected char` errors.

`JsonParsingException` cases:

* Invalid token: JSON object is fine but the JSON schema wasn't expected
* Unexpected char if JSON is completely invalid

Are these messages appropriate?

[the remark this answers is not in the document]
romain-grecourt: I don't think this we should specialize the error messages that much. Instead I'd say you should simply handle the two kinds of error that should return a 400 or a 500 and use a generic error message.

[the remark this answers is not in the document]
camerondurham: @romain-grecourt Just to clarify, you mean to only have responses for the 500 status `ClassCastException` and `JsonParsingException` and the 400 status `JsonException`?
Thanks!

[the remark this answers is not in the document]
romain-grecourt: I don't think we should check `JsonParsingException`, since it inherits `JsonException`. `ClassCastException` here is a side effect of the implementation for the JSONP reader, it does not seem right to have this code here ; instead the reader should be improved to catch the `ClassCastException` and re-throw something more specific (maybe a `JsonException`).

I.e, I suggest you only detect JsonException to return a 400 status, al other exceptions return a 500 status.
Use a generic message and do not include the exception message, e.g. `Invalid JSON` or `Internal error` and log the exception at a FINE level.

Feel free to create an issue to track the improvements to the JSONP reader for better handling JsonObject and JsonArray casting issues.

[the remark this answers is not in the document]
camerondurham: @romain-grecourt Thanks! I updated with your requested changes and will look into filing that issue for the JSONP reader.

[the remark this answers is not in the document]
romain-grecourt: That looks good to me.

PR #861, opened 2019-07-26
WIP: Add OpenAPI parameter-processing util method (and test)

tjquinno:
OpenAPI supports multiple values for named parameters, using various formats. (For some background, see 

https://swagger.io/docs/specification/describing-parameters/
https://swagger.io/docs/specification/serialization/
https://www.ibm.com/support/knowledgecenter/en/SS4SVW_beta/designing/api_defining_mapping_http_parameters.html)

This PR adds a utility method to the Helidon OpenAPI project that developers can use to parse such a multi-valued parameter into a `List`.

The utility method will also be used in the generated Helidon SE classes from the OpenAPI generator tooling.


openapi/src/main/java/io/helidon/openapi/ParameterUtil.java
```diff
@@ -0,0 +1,84 @@
...
+
+        /**
+         * Returns the {@code Separator} value that matches the specified lower-
+         * case name.
+         *
+         * @param formatName lower-case name for the separation style
+         * @return {@code Separator} value matching the lower-case name
+         * @throws IllegalArgumentException if no {@code Separator} matches the name
+         */
+        private static Separator match(String formatName) {
```
batsatt: I think you can simplify this whole class by leveraging `Enum.valueOf(Separator.class, formatName.toUpperCase())` as your `match` method. This way you can eliminate the `formatName` field and also support mixed-case names (unless those are prohibited).

tjquinno: Yeah, I considered that. The code as written is a bit more general but you're right, given the current choices for format that would be simpler. I'll make that change. If I need to generalize it later we'll cross that bridge then.

tjquinno:
Thanks to @batsatt for the review comments so far. I have marked this as WIP to forestall further reviews -- for the moment -- because I plan to improve the utility class to be even more helpful in dealing with the different _locations_ (e.g., path, query, header, cookie), _styles_ (simple, label, matrix), and _explode_ (t/f) settings. 

Doing so will further simplify the OpenAPI generator template for Helidon SE as well as be more useful to developers _not_ using the generator who want to deal with endpoint parameters.

Stay tuned.

openapi/pom.xml
```diff
@@ -58,6 +58,10 @@
             <groupId>org.jboss</groupId>
             <artifactId>jandex</artifactId>
         </dependency>
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+        </dependency>
```
romain-grecourt: This might cause some complication with the newer JDKs (>= 11) where jaxb is not available.
You may need to use profiles to set jaxb-api with scope provided (< 11) or add the both the api and impl (>= 11)

openapi/src/main/java/io/helidon/openapi/ParameterParserImpl.java
```diff
@@ -0,0 +1,414 @@
...
+            super(paramName, exploded, includeID);
+            this.delimiter = delimiter;
+        }
+
+        @Override
+        String valueToParse(String originalValue) {
+            if (exploded()) {
+                return originalValue;
+            }
+                return originalValue.substring(prefix().length());
```
romain-grecourt: nit: extra indent.

openapi/src/main/java/io/helidon/openapi/ValueConverterImpl.java
```diff
@@ -0,0 +1,200 @@
...
+    @Override
+    public Optional<Long> toIntegerInt64(String value) throws NumberFormatException {
+        return toLong(value);
+    }
+
+    @Override
+    public Optional<Long> toIntegerInt64(Optional<String> value) {
+        return toIntegerInt64(value.orElse(null));
+    }
+
```
romain-grecourt: nit: extra empty line

openapi/src/test/java/io/helidon/openapi/ParameterParserTest.java
```diff
@@ -0,0 +1,222 @@
...
+    @Test
+    public void testSimpleStyle() {
+        final ParameterParser parser = ParameterParser.builder("id", Location.PATH)
+                .build(); // uses default SIMPLE style
+
+        final String input = "a,b,c";
+        final List<String> expected = CollectionsHelper.listOf("a,b,c".split(","));
+
+        final List<String> result = parser.parse(input);
+        assertEquals(expected, result);
```
romain-grecourt: Our dev guidelines mandate the use of assertThat.
In this case it would be: `assertThat(parser.parse(input), hasItems("a", "b", "c"));`

openapi/src/test/java/io/helidon/openapi/ParameterParserTest.java
```diff
@@ -0,0 +1,222 @@
...
+        ParameterParser parser = ParameterParser.builder("id", Location.QUERY)
+                .style(Style.SPACE_DELIMITED)
+                .exploded(false)
+                .build();
+
+        String input = "a%20b%20c";
+        final List<String> expected = CollectionsHelper.listOf("a,b,c".split(","));
+
+        assertEquals(expected, parser.parse(input));
+
```
romain-grecourt: nit: extra empty lines.

Issue #690, opened 2019-05-14
Default Jersey thread pool should resize based on heuristics (e.g. queue depth)


batsatt:
The `ThreadPoolExecutor` used by default does not resize except under extreme conditions, and the algorithm cannot easily be changed. See #689 for context.

An algorithm that can start off with relatively few threads and adapt to load should be investigated.

tomas-langer:
Please do not do this.
The number of threads can be specified using configuration.
For production environments, the initial thread count should be the final thread count.
For dev environment you can live with a small number.

Variable number of threads is a nightmare to tune and troubleshoot. In most production environments you end up with a fixed thread pool anyway to avoid cold starts (experience from tuning WLS environments).

batsatt:
You misunderstand: configuration will continue to work as is. Users would still be completely free to specify max == min to achieve fixed size behavior. When the configuration _does_ allow for growth, however, the hard-wired algorithm in `ThreadPoolExecutor` can only be affected by controlling the queue capacity, requiring a very small queue for growth to ever occur under reasonable conditions.

I was unaware of the specifics of the hard-wired growth algorithm and I strongly suspect others are as well; this can easily lead to very surprising behavior where growth (above min) 'never' occurs (_exactly_ as happened with our previous default configuration). I want to eliminate this surprise by choosing a different growth strategy by default, while adding the ability to select a strategy via configuration.

The goals here are to: 

1. Provide more control over when the pool should grow above the configured min ("core") size, replacing the hard-wired algorithm with a simple growth strategy interface (effectively a named predicate) that can be selected via configuration.
2. Provide 3 simple growth strategy implementations, e.g. `whenQueueIsFull` (prefer queueing over growth, same as hard-wired strategy), `whenPoolHasRoom` (prefer growth over queuing) and one based on some simple heuristic like queue size or growth rate. 
3. Provide reasonable default behavior with sizing based on cores, a large queue size ( _maybe_ based on memory) and the heuristic growth strategy. 

Note that if a fixed pool is configured, the growth strategy will never be used.

We can debate whether the default pool size should be fixed or allow growth. My strong preference is to set a reasonable minimum but allow for growth to defer the need for tuning.

spericas:
I agree with Bryan that the default behavior is somewhat confusing and surprising, especially the fact that the queue, as I understand it, is of size `MAX_VALUE`. I also agree that we need something better than the default we have now. 

[...]

PR #871, opened 2019-08-01
Thread pool growth

batsatt:
As discussed in issue #690, this work addresses Jersey thread pool sizing in MP, both the default and currently configurable behavior. 

While it is clear that most production uses will require tuning and likely result in use of a fixed size pool, out-of-the-box behavior should reasonably handle most development and light production uses, deferring the need to tune. Successful tuning is difficult and highly dependent on both application stability and some understanding of expected load, so the longer in the development cycle it can be deferred, the better.

At one end of the solution space is configuring a fixed size pool based on the available cores, using a "small" multiplier to avoid extra resource consumption. Choosing a good multiplier is a crap-shoot, however, as we have no way of anticipating application thread consumption. 

At the far other end of the space is an "autonomous" pool that reacts to conditions while fully considering current resources. This would be complex, easily wrong, and likely entail a performance penalty.
 
A better middle ground solution is simply to choose _two_ multipliers rather than one, configuring the minimum ("core") size with one and the maximum with the other. Using a range enables the default configuration to cover a much wider set of use cases, and is an obvious (and frequently used) choice. 

Our own implementation originally used this approach, and we support explicit configuration of core and max sizes; however, as described in #690, _it does not grow as expected_. 

This leaves us with two solutions, IMO:

1. Support only a single `size` option specifying a fixed size pool. Transition using deprecation/warnings. Pick a single multiplier for the default. During the transition, if `max-pool-size` is configured warn and ignore it. If `core-pool-size` is configured, warn and use it as the size to maintain existing behavior. If the new `size` option and either of the old ones are present, fail with an exception. Remove the warnings, `core-pool-size` and `max-pool-size` settings at some point.
2. Fix the thread pool so that it behaves as expected, with some lightweight mechanism to grow under sufficient load up to max size. Pick two multipliers for the default and continue to support the core/max configuration without violating the principal of least surprise.

This PR is an implementation of the second solution.

While it includes new configuration options for controlling growth (see `growthThreshold` and `growthRate` in `ThreadPoolSuppler.Builder`), we could defer exposing these until more experience is gained, preserving our ability to change the growth algorithm. We could also simplify them to a single more abstract option, e.g. a low/medium/high rate and be less precise in describing what they (currently) mean.


common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
```diff
@@ -131,18 +239,603 @@ public int getQueueCapacity() {
...
+        if (handler instanceof RejectionPolicy) {
+            super.setRejectedExecutionHandler(handler);
+        } else {
+            throw new IllegalArgumentException(handler.getClass() + " must be an instance of " + RejectionPolicy.class);
+        }
+    }
+
+    @Override
+    public void setMaximumPoolSize(int maximumPoolSize) {
+        throw new UnsupportedOperationException();
```
tomas-langer: This is a backward incompatible change.

webserver/webserver/src/main/java/io/helidon/webserver/ServerThreadPoolSupplier.java
```diff
@@ -14,22 +14,37 @@
  * limitations under the License.
  */
 
-package io.helidon.common.configurable;
+package io.helidon.webserver;
```
tomas-langer: This is a backward incompatible change.
Also this may be used by any server, not just the webserver.

common/configurable/pom.xml
```diff
@@ -58,13 +63,18 @@
...
+            <artifactId>hamcrest-all</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
```
tomas-langer: Do we really need a new dependency for a single test? Do we have approval for this?

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
batsatt: True. Given that setting this setting is generally pointless anyway, even without my changes, I guess making it a No-Op is a better choice, perhaps with a log message.

webserver/webserver/src/main/java/io/helidon/webserver/ServerThreadPoolSupplier.java
batsatt: The issue is the desire to ensure that when a task _is_ rejected by the pool, the result will be a `503` rather than a `500`. I wanted to make a generic change in `RequestRouting` that could handle any `HttpException`, wrapped or not, and still return the contained status. This approach requires that an `HttpException` should be thrown by the rejection handler. Given that this class was only recently introduced, the probability of breakage seems quite low, but is obviously non-zero.

I can think of two ways to put this class back in the original module:

1. Add a new exception (e.g. `ServerBusyException`) in the original module and make the corresponding (less generic) change in `RequestRouting`, or
2. Changing this class so that it does not default the rejection handler, and change the usage sites in `Server` and `JerseySupport` to pass one that throws the `HttpException`.

common/configurable/pom.xml
batsatt: Yes we have approval. I had the same thought and tried replacing it with a simple mechanism, but I could not get the growth test to pass. I did spend some time at it, but ultimately decided it wasn't worth any more effort. 

This library has some [really nice features](https://github.com/awaitility/awaitility/wiki/Usage) which I fully expect will be useful elsewhere.

webserver/webserver/src/main/java/io/helidon/webserver/ServerThreadPoolSupplier.java
tomas-langer: Default rejection handler was in the original implementation as well - can we keep that the same? I would opt for option 2.

common/configurable/pom.xml
tomas-langer: OK

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
batsatt: Changed this to log a warning.

webserver/webserver/src/main/java/io/helidon/webserver/ServerThreadPoolSupplier.java
batsatt: I realized that there was no need to add a new exception class; I updated `RequestRouting` to handle `java.util.concurrent.RejectedExecutionException`. Moved the class back to the helidon-common-configurable module.

PR #883, opened 2019-08-08
Allow different servers within an app to report different OpenAPI doc…

### LINKED ISSUE #880: OpenAPI generates same response for both server ports (different functions on each)
## Environment Details
* Helidon Version: 1.2.0
* Helidon SE or Helidon MP MP
* JDK version: Oracle JDK 12
* OS: macOS 10.13.6

----------

## Problem Description
My application is configured to use two ports to separate admin functions from the rest of the API.  On either port, I get the same OpenAPI document.  (It may have switched between the API for the base port 8080 and the admin port 8081, but it seems to be just the admin API now)

Portion of Main.java where I start the two servers:

```
Config config = ConfigProvider.getConfig();
Server server = Server.builder().addApplication(Authorize.class).addApplication(DataApplication.class).port(8080).build().start();
Server server2 = Server.builder().addApplication(Admin.class).port(8081).build().start();

``` 
Microprofile config properties excerpt:
```

server.static.classpath.location=/WEB
server.static.classpath.context=/
app.greeting=Hello
...

server.port=8080
server.host=0.0.0.0

```
Problem occurs every time.

## Steps to reproduce
Reproducing the problem, I think:

1. Start with quickstart MP application.  Add OpenAPI support 
2. Move one of classes implementing javax.ws.rs.core.Application  to a second server.
3. Start both in Main.java with separate ports.
4. Check http://localhost:port1/openapi and http://localhost:port2/openapi to see whether the OpenAPI docs are different.

tjquinno:
Fixes #880 

Helidon SE allows multiples servers in an application, but the OpenAPI support implementation relied directly on the single instance of an OpenAPI model that is managed by the SmallRye implementation (which we leverage). Only when a client accessed the `/openapi` endpoint did `OpenAPISupport` fetch the previously-constructed model from SmallRye for use in preparing the response from that instance of the `OpenAPISupport` service.

This meant that when multiple servers would register OpenAPI support, the model from the _last_ one to register would "win." Clients accessing `/openapi` on _any_ server's port would all get the same OpenAPI document for that last server to register.

The changes in this PR capture and save the `OpenAPI` model that SmallRye creates at the time the `OpenAPISupport` instance routing is updated (rather than later, as before). When the different instances of `OpenAPISupport` respond to `GET` requests to `/openapi` (ultimately invoking `formatDocument`) each instance uses with its own saved model.

This works fine for SE. 

MP itself (not just OpenAPI) does _not_ support multiple servers in one app. These changes would work fine for MP as well _except_ that, by default, under MP the `OpenAPISupport` class scans _all_ the code for endpoint annotations. The annotation scanning is not scoped by which server registered the current instance of `OpenAPISupport` so -- _if_ someone tried to use multiple servers in MP -- then _all_ OpenAPI documents reported by _any_ port would include information for _all_ annotated endpoints.

There might be ways of dealing with this, but given that multi-server MP apps are invalid anyway, I made no attempt to solve that in this PR.


openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -86,6 +87,7 @@
     private final Builder builder;
     private final OpenApiConfig openAPIConfig;
 
+    private OpenAPI model = null;
```
batsatt: Can you move the call to `initializeOpenAPIDocument` into the ctor and make this final? 

If it does need to be deferred, the explicit setting of `null` here is of course redundant and slightly smelly; It may not be strictly necessary, but you might consider using a  `final AtomicReference` to hold the model.

tjquinno: I've reworked things a bit to do all initialization in the constructor, which allows `model` to be `final` and eliminates the need to store the builder at all.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -108,63 +106,58 @@ public void update(final Routing.Rules rules) {
...
         }
     }
 
     private boolean isAnnotationProcessingEnabled(OpenApiConfig config) {
         return !config.scanDisable();
     }
 
-    private void expandModelUsingAnnotations(final OpenApiConfig config) throws IOException {
-        if (builder.indexView() != null) {
+    private void expandModelUsingAnnotations(final OpenApiConfig config, IndexView indexView) throws IOException {
```
batsatt: You should be consistent with the use of `final` for parameters. And I would be remiss if I didn't say that using it at all is controversial, and we have been encouraged not to.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -84,16 +84,14 @@
...
-    private OpenAPI model = null;
+    private final OpenAPI model;
     private final ConcurrentMap<Format, String> cachedDocuments = new ConcurrentHashMap<>();
 
     private OpenAPISupport(final Builder builder) {
         webContext = builder.webContext();
-        this.builder = builder;
-        this.openAPIConfig = builder.openAPIConfig();
+        model = prepareModel(builder.openAPIConfig(), builder.indexView(),
+                    builder.staticFile());
```
batsatt: Just a nit: this looks like it would fit just fine on the previous line.

tjquinno: Fixed.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: I removed all `final` from parameter declarations in `OpenAPISupport`.

PR #890, opened 2019-08-11
Added guide for setting up datasources in Helidon MP.

ljnelson:


docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,345 @@
...
+<1> Your application will now require the Oracle Universal Connection
+Pool classes on its compile and runtime classpaths.
+
+<2> The CDI container will use the Oracle Universal Connection Pool
+configuration described elsewhere in this document to create a new or
+retrieve an existing
+https://docs.oracle.com/en/database/oracle/oracle-database/19/jjuar/oracle/ucp/jdbc/PoolDataSource.html[`PoolDataSource`]
+implementation instance whose name is specified by the `@Named`
+annotation, and will set this field's value to it.
+
```
barchetta: Do we have some examples to point to? Like:

For some complete examples see:

* [Hikari wtih MySQL](https://github.com/oracle/helidon/tree/{helidon-version}/examples/integrations/cdi/datasource-hikaricp-mysql)
* [Hikari with H2](https://github.com/oracle/helidon/tree/{helidon-version}/examples/integrations/cdi/datasource-hikaricp-h2)

ljnelson: Hmm; we do, but they're kind of complicated.  Given that lines 342-344 above are already commenting on an example do you think I should link to them?  I'm inclined not to.  More thoughts welcome.

barchetta: I like the idea of pointing the user to a working example (or examples) that they can easily build and run.  If the above are too complicated (I'm not sure they are), then maybe a new example using UCP?

ljnelson: Added some links to examples.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -343,3 +348,19 @@ https://docs.oracle.com/en/database/oracle/oracle-database/19/jjuar/oracle/ucp/j
 implementation instance whose name is specified by the `@Named`
 annotation, and will set this field's value to it.
 
+== Examples
+
+Helidon features a few examples of projects that use data sources.
+
+* https://github.com/oracle/helidon/tree/master/examples/integrations/cdi/datasource-hikaricp-h2[An
```
romain-grecourt: Can you point at examples in the release tag using the project-version attribute instead ?

ljnelson: Oh, that's a good idea.

docs/src/main/docs/guides/07_datasource.adoc
ljnelson: @barchetta Does this look OK?

ljnelson:
@barchetta Changes are good, ja?

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+=== Add the Hikari Connection Pool to the Runtime Classpath
+
+If you choose to integrate the Hikari connection pool into your
+Helidon MP application, then add an appropriate
+https://maven.apache.org/ref/3.6.1/maven-model/maven.html#class_dependency[`<dependency>`]
+element as a child element of the `<dependencies>` element in your
+`pom.xml`, referencing the HikariCP extension:
+
+[source,xml,subs="attributes+"]
```
romain-grecourt: This is not needed anymore.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+|<<about/03_prerequisites.adoc,Helidon Prerequisites>>
+|The Maven coordinates for your database vendor's JDBC-compliant database driver library
+|===
+
+== Add Your Database Driver to Your Helidon MP Application's Runtime Classpath
+
+Make sure your database vendor's driver library is present on your
+application's runtime classpath.  In a Maven project, add an
+appropriate
+https://maven.apache.org/ref/3.6.1/maven-model/maven.html#class_dependency[`<dependency>`]
```
romain-grecourt: Referencing the maven model documentation for `dependency` seems a bit overkill.
Just saying "add a dependency" is enough.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+   Universal Connection Pool] into Helidon MP applications
+
+Each connection pool offers tradeoffs in terms of size, efficiency and
+features.
+
+=== Add the Hikari Connection Pool to the Runtime Classpath
+
+If you choose to integrate the Hikari connection pool into your
+Helidon MP application, then add an appropriate
+https://maven.apache.org/ref/3.6.1/maven-model/maven.html#class_dependency[`<dependency>`]
```
romain-grecourt: Just saying "add a dependency" is enough IMO.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+    <artifactId>helidon-integrations-cdi-datasource-hikaricp</artifactId>
+    <scope>runtime</scope> <!--1-->
+</dependency>
+----
+<1> Note the `<scope>` is `runtime`.
+
+=== Add the Oracle Universal Connection Pool to the Runtime Classpath
+
+If, instead, you shoose to integrate the Oracle Universal Connection Pool into your Helidon MP application, then add an appropriate
+https://maven.apache.org/ref/3.6.1/maven-model/maven.html#class_dependency[`<dependency>`]
```
romain-grecourt: Just saying "add a dependency" is enough IMO.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+element as a child element of the `<dependencies>` element in your
+`pom.xml`, referencing the Oracle Universal Connection Pool extension:
+
+[source,xml,subs="attributes+"]
```
romain-grecourt: This is not needed anymore.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+== Add Your Database Driver to Your Helidon MP Application's Runtime Classpath
+
+Make sure your database vendor's driver library is present on your
+application's runtime classpath.  In a Maven project, add an
+appropriate
+https://maven.apache.org/ref/3.6.1/maven-model/maven.html#class_dependency[`<dependency>`]
+element as a child element of the `<dependencies>` element in your
+`pom.xml`, referencing the driver library.  Here's an example that
+includes the
+https://search.maven.org/classic/#artifactdetails%7Ccom.h2database%7Ch2%7C1.4.199%7Cjar[H2
```
romain-grecourt: Somehow I don't feel good about having links to search.maven.org. Maybe we could instead reference the one we use using attribute substitution ?

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+[source,xml,subs="attributes+"]
+----
+<dependency>
+    <groupId>io.helidon.integrations.cdi</groupId>
+    <artifactId>helidon-integrations-cdi-datasource-hikaricp</artifactId>
+    <scope>runtime</scope> <!--1-->
+</dependency>
+----
+<1> Note the `<scope>` is `runtime`.
```
romain-grecourt: How about couple more words about the to explain why the runtime scope ?
This is I guess common to all CDI extensions we provide so maybe we could document that in a common place or say something like "Note the `<scope>` is `runtime` since it is not needed during compilation"

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+== Configure the Connection Pool
+
+Regardless of which connection pool you integrate into your Helidon MP
+application via the mechanisms described above, you will need to tell
+that connection pool's machinery how many pools to set up, what to
+name each of them so you can refer to them inside your code, and how
+each of them should connect to a database.
+
+Helidon MP, because it is a conformant MicroProfile specification
```
romain-grecourt: This paragraph should be part of the Helidon MP config documentation IMO.

docs/src/main/docs/guides/07_datasource.adoc
```diff
@@ -0,0 +1,364 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+///////////////////////////////////////////////////////////////////////////////
+
+= Setting Up Data Sources
```
romain-grecourt: This document isn't really a guide per the current definition. (create a project, do something, exercise something).
Maybe this can moved to the normal document as-is.

ljnelson: Hmm.  OK.  This document is a prerequisite for the JPA guide too.  So that's now on hold.

How would you suggest I improve this document's structure?

docs/src/main/docs/guides/07_datasource.adoc
ljnelson: OK; would you like me to remove it?

docs/src/main/docs/guides/07_datasource.adoc
ljnelson: Fixed

docs/src/main/docs/guides/07_datasource.adoc
ljnelson: Fixed

docs/src/main/docs/guides/07_datasource.adoc
ljnelson: Fixed; I just removed the note

docs/src/main/docs/guides/07_datasource.adoc
romain-grecourt: How about splitting this into 2 documents, one for each data source and document the steps needed to exercise each data source in a small project.
Not sure how that fits with the JPA guides, maybe these are multi-part guides ? (these would be part1, the jpa guide would be part 2?)

docs/src/main/docs/guides/07_datasource.adoc
romain-grecourt: Let's see what others have to say about this, @barchetta thoughts ?

barchetta: Yes, this could be more concise. Don't need to explain MicroProfile config, just that the extension uses it for configuration, and so we'll be putting properties here. And for more information see blah, blah, blah.

ljnelson: Totally fair.  Do note that I'm also trying to explain why I'm about to show you some `application.yaml`, but you could supply the same information as system properties or environment variables, in which case the syntax and names of the properties in question change.  I'm not sure if that's something I need to spell out here or not.

ljnelson:
Sounds good, so one path through the guide, picking one connection pool, one database, etc. and leaving all the other choose-your-own-adventure paths as exercises for the reader?

Issue #881, opened 2019-08-08
MP does not support multiple servers


tjquinno:
A separate issue #880 describes using _two_ separate servers in an MP application.

This is _not_ a supported use case. Parts of the MP implementation are single-threaded.

Helidon MP needs to document and/or enforce this more explicitly.

tomas-langer:
Two tasks:

- [ ] Document this behavior

- [ ] Make sure only a single `Server` runs within a VM

PR #892, opened 2019-08-12
Documentation of admin port configuration in Helidon MP.

tomas-langer:

### DIFF
--- a/docs/src/main/docs/microprofile/02_server-configuration.adoc
+++ b/docs/src/main/docs/microprofile/02_server-configuration.adoc
@@ -29,6 +29,7 @@ There are 3 default MicroProfile Config sources:
 * `System.getProperties()`
 * `System.getenv()`
 * all `META-INF/microprofile-config.properties` files on the class path
+* `application.yaml` on the classpath (read by default by Helidon Config)
 
 In this example, the configuration is in a file, and it includes Helidon configuration options.
 
@@ -57,23 +58,29 @@ ssl:
     keystore-passphrase: "abcd"
 ----
 
+== Configure "admin" port
+For this example, we will use a `yaml` file. We configure a port 8011 that would server health and metrics endpoints,
+and then configure health and metrics to use it.
 
-For Helidon config, the default file is `application.yaml` on the classpath
- (e.g. src/main/resources/application.yaml).
+Alternative ports can be used by Server extensions that support it. This support is currently limited to MP Health and
+MP Metrics implementations in Helidon.
+
+You can create a custom Server extension and run routes on alternative ports - this support is limited to WebServer routing.
 
 [source,yaml]
 .application.yaml - Server configuration
 ----
 server:
+  port: 7011
+  host: "some.host"
   sockets:
-    - secure:
-        port: 443
-        # supports all socket related properties of server
-        backlog: 1024
-        receive-buffer: 0
-        timeout: 60000
-        ssl:
-          ....
-    - another:
-        port: 12041
+    admin:
+      port: 8011
+      bind-address: "some.host"
+
+metrics:
+  routing: "admin"
+
+health:
+  routing: "admin"
 ----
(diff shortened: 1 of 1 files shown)


tomas-langer:
Related to #881

docs/src/main/docs/microprofile/02_server-configuration.adoc
```diff
@@ -57,23 +58,29 @@ ssl:
     keystore-passphrase: "abcd"
 ----
 
+== Configure "admin" port
+For this example, we will use a `yaml` file. We configure a port 8011 that would server health and metrics endpoints,
```
romain-grecourt: `would server health`, this looks like a typo.

docs/src/main/docs/microprofile/02_server-configuration.adoc
```diff
@@ -57,23 +58,29 @@ ssl:
...
+and then configure health and metrics to use it.
 
-For Helidon config, the default file is `application.yaml` on the classpath
- (e.g. src/main/resources/application.yaml).
+Alternative ports can be used by Server extensions that support it. This support is currently limited to MP Health and
```
romain-grecourt: I think it would preferred to define the notion of an  "admin" port at the beginning of the section.
Also use admin or alternative, but not both terms.

docs/src/main/docs/microprofile/02_server-configuration.adoc
```diff
@@ -57,23 +58,29 @@ ssl:
...
+MP Metrics implementations in Helidon.
+
+You can create a custom Server extension and run routes on alternative ports - this support is limited to WebServer routing.
```
romain-grecourt: Server extensions are not documented right ?
I think it's better omit this sentence for now.

tjquinno: At whatever point this note is added, maybe add a sentence explaining the implications of the support being limited to WebServer routing...maybe as simple as "you cannot use JAX-RS" or something like that.

docs/src/main/docs/microprofile/02_server-configuration.adoc
tjquinno: I think here, where the doc _first_ introduces additional ports, it should mention that such additional ports are restricted in what they can do. The restrictions can be explained in more detail later (see my note below).

docs/src/main/docs/microprofile/02_server-configuration.adoc
tomas-langer: Reworded the whole section

docs/src/main/docs/microprofile/02_server-configuration.adoc
tomas-langer: I have removed the section about server extensions and clearly explained what it can be used for (I hope)

docs/src/main/docs/microprofile/02_server-configuration.adoc
tomas-langer: I have split this into two parts

- overall explanation of ports in MP and restrictions
- example description

docs/src/main/docs/microprofile/02_server-configuration.adoc
```diff
@@ -57,23 +58,34 @@ ssl:
     keystore-passphrase: "abcd"
 ----
 
+== Configure "admin" port
```
romain-grecourt: I would rename this title to `Configuring additional ports`.

docs/src/main/docs/microprofile/02_server-configuration.adoc
```diff
@@ -57,23 +58,34 @@ ssl:
...
+Helidon can expose multiple ports. For the purpose of Helidon MP, this support has the following limitations:
```
romain-grecourt: Using just `Helidon` in the begining is a bit ambiguous, since this is document is for Helidon MP only this can be more explicit.
Also the use of `purpose` is a little awkward. 

How about `Helidon MP can expose multiple ports, with the following limitations:`

tomas-langer:
Merging as build failure was related to slow environment (Metrics) and this is a documentation fix only.

PR #902, opened 2019-08-14
Logging a warning when more than one MP server is started.

### LINKED ISSUE #881: MP does not support multiple servers

tomas-langer:

Together with #892 resolve #881

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -165,6 +165,7 @@ public interface Server {
         private Supplier<? extends ExecutorService> defaultExecutorService;
         private Context parentContext;
         private Context serverContext;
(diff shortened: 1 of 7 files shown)


tjquinno:
Code looks fine for logging a warning.

This would be a little more complicated, but the user experience might be clearer if we rejected multiple servers unless the developer explicitly set a config value to allow them. That means the user would have taken a specific action to enable multiple server instances.

Users have been known to ignore even loud warning messages.

tomas-langer:
@tjquinno seems like a sound advice, I will refactor this.

tomas-langer:
I have modified the code. Now you must explicitly enable all MP Servers to run in parallel, otherwise an exception is thrown.

PR #910, opened 2019-08-16
gRPC microprofile support

thegridman:
Adding gRPC microprofile support


docs/src/main/docs/grpc/32_mp_server_side_services.adoc
```diff
@@ -0,0 +1,297 @@
...
+=== Annotate the Service Implementation
+When the gRPC MP server is starting it will discover all CDI beans of type `io.grpc.BindableService`. Service sub-classes
+implemented the traditional way with code generation are instances of `BindableService` so by annotating the implementation class
+with the `@ApplicationScoped` annotation they become discoverable and will be deployed into the gRPC server.
+
+[source,java]
+----
+@ApplicationScoped
+public class StringService 
+    extends extends StringServiceGrpc.StringServiceImplBase {
```
romain-grecourt: `extend extends`

docs/src/main/docs/grpc/33_mp_clients.adoc
```diff
@@ -0,0 +1,249 @@
...
+associated to more traditional approaches to writing gRPC Java clients. At it simplest a gRPC Java client can be written using
+nothing more than a suitably annotated interface.
+
+
+== Maven Coordinates
+
+The <<about/04_managing-dependencies.adoc, Managing Dependencies>> page describes how you
+should declare dependency management for Helidon applications. Then declare the following dependency in your project:
+
+[source,xml,subs="verbatim,attributes"]
```
romain-grecourt: the `subs` shouldn't be needed.

grpc/core/pom.xml
```diff
@@ -18,102 +18,142 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <extensions>
+            <extension>
+                <groupId>kr.motd.maven</groupId>
+                <artifactId>os-maven-plugin</artifactId>
+                <version>1.5.0.Final</version>
```
romain-grecourt: This should be managed in the top-level parent pom so that the version is not hard-coded down here.

grpc/core/pom.xml
```diff
@@ -18,102 +18,142 @@
...
+            </extension>
+        </extensions>
+
+        <plugins>
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>0.5.1</version>
```
romain-grecourt: Same thing, this should managed in the top-level parent pom.

grpc/core/src/test/java/io/helidon/grpc/core/GrpcTlsDescriptorTest.java
```diff
@@ -41,7 +41,7 @@ public void testDefaultSslDescriptor() {
 
     @Test
     public void testSslDescriptorWithBuilder() {
```
tomas-langer: Please have a look at `io.helidon.common.configurable.Resource` - that can be used to load a resource from various locations (classpath, file system, configuration value etc.) - also supports configuration.
This is used to locate such resources in other modules. Please replace your usage. 
We have an issue to improve the behavior of this class, we plan that for 2.0
This note belongs to `tlsCaCert`, `tlsCert` ans `tlsKey`

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/GrpcClientHelper.java
```diff
@@ -0,0 +1,64 @@
...
+     * {@link io.helidon.microprofile.grpc.core.RpcMethod} annotations
+     * so that the proxy can properly route calls to the server.
+     *
+     * @param channel  the {@link Channel} to connect to the server
+     * @param type     the service type
+     * @param <T>      the service type
+     * @return a dynamic proxy that makes calls to the gRPC service.
+     */
+    public static <T> T proxy(Channel channel, Class<T> type) {
+        ClientServiceDescriptor descriptor = DESCRIPTORS.computeIfAbsent(type, GrpcClientHelper::createDesriptor);
```
tomas-langer: Is this really intended for end users? Can this be package private?

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/GrpcProxyProducer.java
```diff
@@ -0,0 +1,97 @@
...
+     * will connect to the server using the default {@link Channel}.
+     * <p>
+     * This is not a real producer method but is used as a stub by the gRPC client
+     * CDI extension to create real producers as injection points are discovered.
+     *
+     * @param injectionPoint the injection point where the client proxy is to be injected
+     * @return a gRPC client proxy
+     */
+    @GrpcServiceProxy
+    public static Object proxyUsingDefaultChannel(InjectionPoint injectionPoint, ChannelProducer producer) {
```
tomas-langer: Class is package private, methods `public` - is this needed to correctly process the `GrpcServiceProxy` annotation, or can we decrease the visibility?

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/model/ClientServiceModeller.java
```diff
@@ -0,0 +1,193 @@
...
+import io.helidon.microprofile.grpc.core.model.AnnotatedMethod;
+import io.helidon.microprofile.grpc.core.model.AnnotatedMethodList;
+import io.helidon.microprofile.grpc.core.model.Instance;
+import io.helidon.microprofile.grpc.core.model.ModelHelper;
+
+/**
+ * Utility class for constructing a {@link ClientServiceDescriptor.Builder}
+ * from an annotated POJO.
+ */
+public class ClientServiceModeller
```
tomas-langer: Is this similar to MP REST Client? If so, then please have a look at the naming used in REST client and use similar approach.

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/model/ClientServiceModeller.java
```diff
@@ -0,0 +1,193 @@
...
+    private static final Logger LOGGER = Logger.getLogger(ClientServiceModeller.class.getName());
+
+    /**
+     * Create a new introspection modeller for a given gRPC service.
+     *
+     * @param service the service to call gRPC handler methods on
+     * @throws NullPointerException if the service is null
+     */
+    public ClientServiceModeller(Object service) {
+        this(service.getClass(), Instance.singleton(service));
```
tomas-langer: Please never use public contructors anywhere in Helidon (except for very specific cases - see dev guidelines).
If the instance created is the final instance needed, then this should be handled by static factory methods:
`static create(Object)`
`static create(Class)`
`static create(Class, Supplier)`
If these are needed to create a builder, then these should be parameters to a static builder method:
`static builder(Object)`
`static builder(Class)`
`static builder(Class, Supplier)`

microprofile/grpc/core/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-service-loader</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.microprofile.config</groupId>
+            <artifactId>helidon-microprofile-config</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
```
tomas-langer: Why do you depend on Weld? This should not be required.

microprofile/grpc/core/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+            <groupId>org.jboss.weld.se</groupId>
+            <artifactId>weld-se-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
```
tomas-langer: Do not handle slf4j or any other logging framework - we leave this to our users.
We should never directly depend on these libraries (`slf4j`, `log4j`, `logback` and any utilitiies)

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/GrpcCdiExtension.java
```diff
@@ -0,0 +1,53 @@
...
+     * This is required so that we can support the use-case where an interface has been
+     * annotated with {@link RpcService} but the implementation class has not but the
+     * implementation class is annotated with a bean discovering annotation such as
+     * {@link javax.enterprise.context.ApplicationScoped}. We need to make sure that the
+     * gRPC server can locate beans so we add the {@link RpcService} from the interface to
+     * the bean.
+     *
+     * @param event the {@link ProcessAnnotatedType} event
+     */
+    public void beforeBean(@Observes ProcessAnnotatedType<?> event) {
```
tomas-langer: I think that you can limit the event to be only triggered for specific annotation (`@AnnotatedWith` or something like that) - weld will print an Info message telling you what to do, if you listen for all annotated types.
This is a performance issue as well.

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/model/AbstractMethodHandlerSupplier.java
```diff
@@ -0,0 +1,297 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.microprofile.grpc.core.model;
+
```
tomas-langer: You should use a flat package structure in Helidon - e.g. each module (maven module) has a single package.
This is to enable proper use of package local visibility (and so we do not need `internal` packages. 
If really needed, move this to a separate module. If not, refactor to a single package.

microprofile/grpc/core/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,24 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<beans xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="http://xmlns.jcp.org/xml/ns/javaee"
```
tomas-langer: If you use a CDI extension, then this probably should not be marked as a bean archive - you should add the beans manually in the extension.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricsCdiExtension.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param pat  the {@link ProcessAnnotatedType} to observer
+     */
+    private void registerMetrics(@Observes
+                                 @WithAnnotations({Counted.class, Timed.class, Metered.class})
+                                 @Priority(Interceptor.Priority.APPLICATION)
+                                 ProcessAnnotatedType<?> pat) {
+
+        Class<?> javaClass = pat.getAnnotatedType().getJavaClass();
+        ServiceModeller modeller = new ServiceModeller(javaClass);
```
tomas-langer: This method would get called for any class annotated with these annotations, including JAX-RS resources and other CDI beans. Can you restrict the annotations with the Grpc specific annotation as well?
If `ServiceModeller` is public (and as a result part of Helidon API), then you should use factor methods or a builder, never a constructor.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/MetricsConfigurer.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * A {@link AnnotatedServiceConfigurer} that adds a
+ * {@link io.helidon.grpc.metrics.GrpcMetrics gRPC metrics interceptor}
+ * to an annotated gRPC service.
+ * <p>
+ * Metric interceptors are only added to services where the method on the
+ * service class is annotated with a metric annotation. Any metric annotations
+ * on super classes or interfaces will be ignored.
+ */
+public class MetricsConfigurer
```
tomas-langer: Must this class be public? Seems like locally used.

grpc/client/src/test/resources/test-client-config.yaml
```diff
@@ -22,19 +22,19 @@ grpc:
     default_port:
       host: "non_default_host.com"
     default_host_port_ssl_disabled:
-      ssl:
+      tls:
```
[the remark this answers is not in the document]
tomas-langer: There is another note further down - this should use `io.helidon.common.configurable.Resource` that uses a specific approach to configuration as well.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/MetricsConfigurer.java
thegridman: It is discovered in the gRPC MP server module using `ServiceLoader` so as far as I know has to be public.

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/model/ClientServiceModeller.java
thegridman: No I don't think that it is like the REST client  it is more like a builder. I have refactored it to be a Builder.

thegridman:
@tomas-langer I've addressed all of the comments that you made above.

grpc/core/src/main/java/io/helidon/grpc/core/ResponseHelper.java
```diff
@@ -0,0 +1,453 @@
...
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+import io.grpc.stub.StreamObserver;
+
+/**
+ * A number of helper methods to handle sending responses to a {@link StreamObserver}.
+ */
+public interface ResponseHelper {
```
tomas-langer: This interface is a utility class.
Please reimplement as a utility class with static methods.
We do not want to add multiple inheritance into Helidon. Also this adds a lot of methods to the class (and as soon as reflection or CDI is used to obtain an instance, it makes the whole thing a lot more complicated).

PR #912, opened 2019-08-16
Drain data chunks when there is no registered subscriber

romain-grecourt:
Upon response completion and after checking for data chunk leak, drain the data chunks from the publisher if there is no registered subscriber.


romain-grecourt:
This is the fix for PSR issue 2. CC @olotenko

Issue #913, opened 2019-08-16
Cannot use PersistenceContext without unitName


ljnelson:
It would be nice if we follow the spirit of "Java EE mode" JPA and permit a user to do this:
```
@PersistenceContext
private EntityManager em;
```
Right now she must do this:
```
@PersistenceContext(unitName = "someName")
private EntityManager em;
```
This is the most common way of injecting `EntityManager` instances in Java EE so it would be nice to support it here.

PR #914, opened 2019-08-17
Adds support for unitName-less PersistenceContext injection points

ljnelson:
Permits `@PersistenceContext`-annotated injection points where `unitName` is not specified and there is exactly one persistence unit in the environment, following the spirit of Java EE.


ljnelson:
See #913.

Issue #916, opened 2019-08-19
Incorrect assertion statement in Jedis extension


ljnelson:
In the Jedis CDI extension, there is an `assert` statement that incorrectly asserts a given `Type` is a `Class` when it is pretty clear it should be a `ParameterizedType`.

If the Jedis extension is run with assertions enabled and the application using it makes use of `Provider<JedisObjectGoesHere>` then an `AssertionError` will be thrown.

The fix is straightforward.

PR #917, opened 2019-08-19
Corrects erroneous assert statement in JedisExtension

ljnelson:


ljnelson:
Fixes issue #916.

PR #918, opened 2019-08-19
JPA guide

ljnelson:


spericas:
Looking good so far. The static weaving requirement is a bit unfortunate :( It reads like a guide to me!

ljnelson:
Right.  For the benefit of others watching this (public) pull request, the reason that static weaving is a requirement is, briefly, we're not an application server.  Since everything in a Helidon MP application shares the same classloader, to weave on the fly you would end up having to inspect every single class that is loaded, including CDI classes, Helidon classes, JDK classes, etc.  Static weaving of course also takes the hit at build time so startup time is much faster.

I encountered [this link](https://struberg.wordpress.com/2012/01/08/jpa-enhancement-done-right/) several years ago that does a relatively nice job of describing the tradeoffs.

PR #925, opened 2019-08-20
SE and MP Health Guides

pfmackin:
This is the first pass at a new SE and MP Health Guide


pfmackin:
I can do that but there was some discussion between Joe and Mark about refactoring the structure.  Anyway, I will do that for now.

romain-grecourt:
Thanks, at least the guide can fit nicely until we re-work the guides overview.

docs/src/main/docs/guides/07_health_se_guide.adoc
```diff
@@ -0,0 +1,677 @@
...
+    -DarchetypeArtifactId=helidon-quickstart-se \
+    -DarchetypeVersion={helidon-version} \
+    -DgroupId=io.helidon.examples \
+    -DartifactId=helidon-quickstart-se \
+    -Dpackage=io.helidon.examples.quickstart.se
+----
+
+=== Using the built-in health-checks
+
+Helidon has a set of built-in health checks can be optionally enabled to report various
```
barchetta: Missing "that" ("that can be")

docs/src/main/docs/guides/07_health_se_guide.adoc
```diff
@@ -0,0 +1,677 @@
...
+----
+
+import org.eclipse.microprofile.health.HealthCheckResponse;  // <1>
+
+...
+
+private static Routing createRouting(Config config) {
+
+    HealthSupport health = HealthSupport.builder()
+      .addLiveness(HealthChecks.healthChecks())  // <2>
```
barchetta: We should fix this in the quickstart example. I think it still uses the deprecated `add` method. You can do that in this PR or a separate one.

romain-grecourt:
Couple of thoughts.

The guide seems a little long, maybe we can cut the length by removing the JSON snippets for the expected outputs.

I think we should update the quickstart YAML file to add what's missing and documented in this guide (separate PR).

Note that I'm working on a bare archetype, when that's ready we can update this guide (separate) to use it.

markxnelson:
I don't think it is too long - I think some of the other existing ones are too short - they leave out vital information that a person who is not already familiar with Helidon (or MP) would actually need imho.

pfmackin:
I can do whatever you folks want regarding the JSON.  I personally like a guide to be something that a user can cut and paste into and IDE, build, run, and verify that the output is correct.

docs/src/main/docs/guides/07_health_se_guide.adoc
pfmackin: fixed

docs/src/main/docs/guides/07_health_se_guide.adoc
pfmackin: I can do in a separate PR unless Joe does it, I think he said he was going to rework the Quickstarts

romain-grecourt:
There aren't many other guides to compare the length to. Quickstart and meant to be "quick", the other ones are not really comparable.

The current angle for the guides is less to explain and more to exercise ; we should try to avoid overlapping with the "normal" docs.

m0mus:
Most my comments for MP guide are valid for SE guide as well. Please fix it in both guides.

docs/src/main/docs/guides/07_health_mp_guide.adoc
```diff
@@ -0,0 +1,524 @@
...
+
+[width=50%,role="flex, sm7"]
+|===
+|About 15 minutes
+|<<about/03_prerequisites.adoc,Helidon Prerequisites>>
+|===
+
+=== Create a sample MP project
+
+Generate the project sources using the Helidon MP Maven archetype.
```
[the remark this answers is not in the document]
pfmackin: I am not using the Quickstart since it adds code I don't want

docs/src/main/docs/guides/07_health_mp_guide.adoc
```diff
@@ -0,0 +1,524 @@
...
+* available heap memory
+
+This example will demonstrate how to use the built-in health checks.  These examples are all executed
+from the root directory of your project (helidon-quickstart-mp)
+
+
+[source,bash]
+.Build the application, skipping unit tests, then restart it
+----
+mvn package -DskipTests=true
```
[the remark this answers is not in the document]
pfmackin: This guide will cause the tests to break.  Also, it is quicker not to run the tests.

docs/src/main/docs/guides/07_health_mp_guide.adoc
```diff
@@ -0,0 +1,524 @@
...
+----
+<1> Include additional imports
+<2> Annotion specifies that this is a readiness health check
+<3> Build the HealthCheckResponse with status UP after 5 seconds else DOWN
+<4> Initialize the time at startup
+
+[source,bash]
+.Build and restart the application.  Issue the curl command with -v within 5 seconds and you see the application is not ready
+----
+curl -v  http://localhost:8080/health/ready
```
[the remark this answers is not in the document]
pfmackin: I tell them how to build and run at the start of the doc, I don't need to do that every time I tell them to run curl.

m0mus:
@romain-grecourt A guide should contain enough to get started with some specific topic *without* looking on `normal` docs. It may contain links to documentation if users want to go deeper in the topic.

romain-grecourt:
@m0mus I disagree, but there is no need to argue any further.

docs/src/main/docs/guides/07_health_mp_guide.adoc
```diff
@@ -0,0 +1,542 @@
...
+        "totalBytes": 284164096
+      }
+    }
+  ]
+}
+----
+
+NOTE: In MicroProfile Health 2.0 outcome and state were replaced by status in the JSON response wire format.
+Helidon currently provides both fields for backwards compatibility, but use of outcome and state is deprecated
+and will be removed in a future release. You should rely on state instead.
```
barchetta: "you should rely on **status** instead."
Also, might want to back tick outcome, state and status in this paragraph so they stand out as "code".

docs/src/main/docs/guides/07_health_se_guide.adoc
```diff
@@ -0,0 +1,624 @@
...
+.Verify the health endpoint in a new terminal window:
+----
+curl http://localhost:8080/health
+----
+
+[source,json]
+.JSON response:
+----
+{
+    "outcome": "UP",
```
barchetta: It looks like this was run with an older version of Helidon SE? When I run using the 1.2.1 archetype I get both `outcome` and `status` (and `state` and `status`) just like with MP Health. 

So a similar NOTE applies for SE as well ("you should rely on status")

docs/src/main/docs/guides/07_health_mp_guide.adoc
pfmackin: fixed

docs/src/main/docs/guides/07_health_se_guide.adoc
pfmackin: fixed

PR #934, opened 2019-08-21
Update graal docker images

romain-grecourt:
Update graal docker images to 19.2.0
Uptake latest available svm (19.1.1)


romain-grecourt:
@barchetta I've pushed the images to DockerHub.

PR #936, opened 2019-08-21
Fix use of deprecated add() method on HealthSupport in quickstart and bookstore test

barchetta:
Use `addLiveness()` instead of `add()` (which has been deprecated) in quickstart se and bookstore test app. Also fix README to use correct graalvm version.


romain-grecourt:
How about the health docs ?

barchetta:
> How about the health docs ?

Done

PR #972, opened 2019-08-28
Set useBaseVersion to false when copying dependencies 

barchetta:
Narayan has some dependencies that use pinned snapshots. An example

```
<version.microprofile.lra.api>1.0-20190801.113716-482</version.microprofile.lra.api>
```
We have to tell the `maven-dependency-plugin`'s `copy-dependencies` goal to use this in the jar filename, and not use the version in the base snapshot artifact. To do that we need to set useBaseVersion to false.

### DIFF
--- a/examples/quickstarts/helidon-quickstart-mp/pom.xml
+++ b/examples/quickstarts/helidon-quickstart-mp/pom.xml
@@ -114,6 +114,7 @@
                             <overWriteIfNewer>true</overWriteIfNewer>
                             <includeScope>runtime</includeScope>
                             <excludeScope>test</excludeScope>
+                            <useBaseVersion>false</useBaseVersion>
                         </configuration>
                     </execution>
                 </executions>
--- a/examples/quickstarts/helidon-quickstart-se/pom.xml
+++ b/examples/quickstarts/helidon-quickstart-se/pom.xml
@@ -114,6 +114,7 @@
                             <overWriteIfNewer>true</overWriteIfNewer>
                             <includeScope>runtime</includeScope>
                             <excludeScope>test</excludeScope>
+                            <useBaseVersion>false</useBaseVersion>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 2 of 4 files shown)


romain-grecourt:
I don't understand the need for useBaseVersion=false.

This assumes that there external SNAPSHOT dependencies in the dependency graph ; this should never be the case.
Also in case external SNAPSHOT dependencies are actually pulled down, having the timestamp information is valuable.

ljnelson:
For completeness, see: https://github.com/apache/maven-dependency-plugin/blob/1d2108b0cd7c5fae15edad8ba7f80df24f36297f/src/main/java/org/apache/maven/plugins/dependency/utils/DependencyUtil.java#L114-L121

romain-grecourt:
This change actually breaks the SNAPSHOT use-case. The maven-jar plugin does not add the timestamp the `class-path` MANIFEST.MF entry without settings `useUniqueVersion=true`.

ljnelson:
Are you [sure](https://stackoverflow.com/questions/14256291/maven-archiver-uses-unlocked-snapshots-in-classpath-but-copy-dependencies-copy)?  Maybe you/we are running with an old copy of the Maven bits involved?  Because `true` is the default value for this parameter, and we were seeing problems in that case.

My comment above shows where `useBaseVersion` affects things.  Maybe I've read it wrongly.

paulparkinson:
I saw the same behavior Laird speaks of

romain-grecourt:
The default value of `useUniqueVersion` does not apply since it's set explicitly to `false`, see https://github.com/oracle/helidon/blob/master/examples/quickstarts/helidon-quickstart-se/pom.xml#L85

`useBaseVersion=false` combined `useUniqueVersion=false` does not work with remote SNAPSHOT dependencies: the MANIFEST.MF files has a `class-path` entry without timestamps and the `libs` directory has SNAPSHOT dependencies files named with the timestamp.

It does work with local SNAPSHOT since they are not timestamped.

Note that narayana has dropped the timestamped SNAPSHOT dependency as of `5.9.7.Final`, see https://github.com/jbosstm/narayana/blob/5.9.7.Final/rts/lra/pom.xml

ljnelson:
@romain-grecourt Do you suggest eliminating both configuration settings?  Or…?

romain-grecourt:
Why not, the right values are the defaults ;) - I'm not sure there is any good reason to set them explicitly.

PR #986, opened 2019-09-04
Injection of SecurityContext using CDI.

### LINKED ISSUE #486: Allow injection of "Security" through CDI
The Security object is created in `SecurityMpService`. At the end of the `configure1` method, the Security object is "registered" in `MpServiceContext`, which seems to just add it to otherwise inaccessible Map.
It would be good to have it registered in CDI at this point to avoid parsing the Security config multiple times.
The Security object is normally accessible through `@Context` injection from Jersey server, but this is not available to us in some cases when we use microprofile (for CDI) without Jersey server (e.g. microservice consuming messages from message queue).

tomas-langer:

Resolves #486


microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityContextProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * A bean for injection of security context.
+ */
+class SecurityContextProvider {
+    private final AtomicReference<SecurityContext> currentReference = new AtomicReference<>();
+
+    SecurityContextProvider() {
+    }
+
+    @Produces
+    SecurityContext securityContext() {
```
ljnelson: Right now, this can return `null`.  It also produces objects in `Dependent` scope (which is the only scope for which `null` is a valid return value).  Is this intended?  Can this class be refactored such that:
* it simply creates or discovers a `SecurityContext` (i.e. the `currentReference` field goes away)
* it produces this `SecurityContext` in `ApplicationScoped` scope
…?  If not, then just recognize as written this currently sort of "tunnels" a singleton-like object "through" `Dependent` scope, which is a little odd.

microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityContextProvider.java
```diff
@@ -0,0 +1,41 @@
...
+        return currentReference.get();
+    }
+
+    void securityContext(SecurityContext toSet) {
+        this.currentReference.set(toSet);
```
ljnelson: Do I understand right that the only class that ever calls this method is a JAX-RS filter?

tomas-langer: yes

microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityContextProvider.java
tomas-langer: It produces in RequestScope, as that is how I register it in the extension

tomas-langer: For each request a new instance is created (of this producer) and that instance holds the request scoped security context as soon as it is ready (e.g. in Jersey request scope).

tomas-langer: Maybe I could just obtain the SecurityRequest from current context when it is requested, as that should work...

ljnelson: No, what you've done in the extension is registered the _provider class itself_ as a `RequestScoped` bean.  Then this producer method produces its return type in `Dependent` scope (so it is called any time a `SecurityContext` is called for elsewhere in CDI, and you'll get a `ContextNotActiveException` if for some reason a `SecurityContext` is called for when a request is not active).  Maybe that's what you want to do; maybe not.

tomas-langer: Understood. Refactored to work as expected (e.g. `Security` injected in `ApplicationScoped` and `SecurityContext` injected in `RequestScoped`)

microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityContextProvider.java
tomas-langer: Removed.

microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityCdiExtension.java
```diff
@@ -0,0 +1,31 @@
...
+import javax.enterprise.inject.spi.BeforeBeanDiscovery;
+import javax.enterprise.inject.spi.Extension;
+
+/**
+ * Extension to register bean {@link SecurityProducer}.
+ */
+public class SecurityCdiExtension implements Extension {
+    void registerBean(@Observes BeforeBeanDiscovery abd) {
+        abd.addAnnotatedType(SecurityProducer.class, "helidon-security-producer")
+                .add(ApplicationScoped.Literal.INSTANCE);
```
ljnelson: Just a comment, no action required: you can do this here, or you can just put `@ApplicationScoped` on `SecurityProducer`.

security/security/src/main/java/io/helidon/security/Security.java
```diff
@@ -69,7 +69,8 @@
  * @see #builder()
  * @see #create(Config)
  */
-public final class Security {
+// class cannot be final, so CDI can create a proxy for it
```
ljnelson: You may discover that you [may need a zero-argument constructor](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#unproxyable).  (It can be package-private, and does not have to fully initialize the class.)  Although note that as written this class will not be proxied because it is not annotated with a scope that requires a proxy.  (But you may be programmatically registering it in such a scope elsewhere.)

tomas-langer: Yes, this is registered by the `SecurityProducer` (part of this PR) in `AppicationScoped`.
The startup failed because CDI was not able to proxy this class. Once I removed the `final` keyword, Security is injected as expected.

microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityCdiExtension.java
tomas-langer: This is intentional - I am not annotating any class with a bean defining annotation. This is to help with uses that create a fat jar. In such a case, this class would be registered twice (once through extension, once due to annotation scanning).

Issue #939, opened 2019-08-22
Verify compatibility of new Metrics implementation with FT


spericas:
Latest Metrics specification has introduced a number of backward incompatible changes, including new semantics for counters and some name changes (in particular in the OpenMetrics/Prometheus format). 

Since FT depends on Metrics, we need to verify compatibility by at least re-running, and possibly upgrading to, the latest TCKs.

spericas:
Made part of PR #992.

PR #992, opened 2019-09-05
Support for MP Metrics 2.0

### LINKED ISSUE #833: MP Metrics 2.0
Upgrade metrics to 2.0.
This is a breaking change.

tjquinno:
Resolves #833 (MP Metrics 2.0) and #939 (verify new metrics with FT).

These changes add support for MP metrics 2.0, including a Helidon bundle for MP 3.0.

The following new modules support the new version of metrics: 
```
metrics2
    metrics2
    prometheus2
microprofile
   metrics2
```
Their original counterparts support the previous metrics version 1.1, as before:
```
metrics
   metrics
   prometheus
microprofile
   metrics
```
This PR updates examples, tests, and TCK drivers to use the newer metrics version with these exceptions:

- MP static content example
- MP security example
- MP IDCS security example
- zipkin integration test

Because MP Metrics 2.0 is incompatible with 1.1, Helidon allows developers with existing metrics-based apps to use the latest Helidon release with the _older_ MP metrics version by depending on the helidon-microprofile-2.2 bundle. 

Helidon SE apps that use metrics can continue to depend on the `metrics/metrics` artifact to use the older metrics implementation. Or, to use the newer SE metrics implementation, they can depend on `metrics2/metrics2`.

Helidon contains the following _internal_ metrics clients:
   fault tolerance
   grpc metrics

These have been changed so they can work with either version of metrics. 

To accomplish this a new metrics abstraction layer exists in the new module `common/metrics`, a separate module because the abstraction layer is used by more than one module. Therefore its key contents have to be `public` which is unfortunate. This abstraction layer _is not_ intended for developers using Helidon so its public classes and interfaces have `Internal` in their names to discourage developers from using them.


tjquinno:
BTW, the wercker/build failure is due to wercker timing out, not due to errors _in_ the build.

tomas-langer:
Currently the build error is because of failing TCK tests of metrics 2.0

tjquinno:
The TCK failures are known. I've been working on some things and have changed the title to WIP to reflect that.

tjquinno:
Removed WIP from the title. TCKs passed locally for me so they should in the pipeline now also.

spericas:
Just to clarify, the TCKs were passing, then an issue was found and some fixes cause the TCKs to fail temporarily. After Tim's latest commit all TCKs are passing again like at the time the PR was submitted.

metrics2/metrics2/src/main/java/io/helidon/metrics/ExponentiallyDecayingReservoir.java
```diff
@@ -0,0 +1,215 @@
+/*
```
tomas-langer: This class is exact duplicate of the class in `metrics`. Probably should be extracted to common metrics?

metrics2/metrics2/src/main/java/io/helidon/metrics/EWMA.java
```diff
@@ -0,0 +1,135 @@
+/*
```
tomas-langer: This class is exact duplicate of the one in `metrics` - maybe extract to common metrics?

metrics2/metrics2/src/main/java/io/helidon/metrics/InternalMetricIDImpl.java
```diff
@@ -0,0 +1,97 @@
...
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final InternalMetricIDImpl other = (InternalMetricIDImpl) obj;
+        if (!Objects.equals(this.delegate, other.delegate)) {
+            return false;
+        }
+        return true;
```
tomas-langer: Last four lines can be replaced with
`return Objects.equals(this.delegate, other.delegate)`

metrics2/prometheus2/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
```
tomas-langer: Why is there a Prometheus 2 - this does not depend on MP metrics at all, only on prometheus itself.
This module should not have been duplicated.

metrics/metrics/src/main/java/io/helidon/metrics/FinalRegistry.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
tjquinno: There were intervening changes along the way that turned out to be unnecessary. Leaving the copyright changed was an oversight. We'll revert this and other similar files.

metrics2/metrics2/src/main/java/io/helidon/metrics/ExponentiallyDecayingReservoir.java
tjquinno: Recall that the goal, eventually, is to _not_ have to support both versions of metrics (1.1 and 2.0) at the same time from a single copy of our internal metrics clients (accomplished in this PR using the bridge). If and when that day comes, we want to be able to remove easily the files tied to the older version -- the entire `metrics` and `microprofile/metrics` modules and even `common/metrics`, leaving only what are now called `metrics2` and `microprofile/metrics2`. 

For that reason, the PR has avoided optimizing the copy by refactoring identical copies into `common/metrics` as that would complicate the changes later to go back to a single copy of metrics and microprofile/metrics.

One cost is, of course, the possible need to update some code in two places if we find bugs. That is a price of making easier the later removal of the second copy.

metrics2/prometheus2/pom.xml
tjquinno: See the earlier comment about ease of removing the older `metrics` and `microprofile/metrics` modules later.

metrics2/metrics2/src/main/java/io/helidon/metrics/ExponentiallyDecayingReservoir.java
tomas-langer: Yes, you are right - we want to remove metrics 1.x and all related additional modules once we move to 2.0, so this makes sense

metrics2/pom.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
tomas-langer: This module should not exist.
Please move `metrics2` submodule to `$root/metrics`
The resulting structure should be:
metrics/metrics
metrics/metrics2
metrics/prometheus

The groupId is already following this, so the directory structure should as well.
Also Prometheus should not be touched by this change at all.

metrics/metrics/src/main/java/io/helidon/metrics/FinalRegistry.java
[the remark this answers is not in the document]
tjquinno: Working on this. Still haven't found a way to restore a version with the correct copyright that also passes the copyright checker.

metrics2/metrics2/src/main/java/io/helidon/metrics/InternalMetricIDImpl.java
tjquinno: Fixed in a pending commit.

metrics/metrics/src/main/java/io/helidon/metrics/FinalRegistry.java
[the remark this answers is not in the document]
tjquinno: I have heard from others who have wrestled with this same problem in other projects. They have also not found a way to solve this. So it seems we might have to live with copyrights that are later than the actual changes in the 8 (I think) affected files. This is unfortunate but not easily avoided at this point.

metrics/metrics/src/main/java/io/helidon/metrics/HelidonCounter.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
tjquinno: Seems we need to live with this.

metrics/metrics/src/main/java/io/helidon/metrics/HelidonGauge.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
tjquinno: Seems we need to live with this.

metrics/metrics/src/test/java/io/helidon/metrics/HelidonCounterTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
tjquinno: Seems we need to live with this.

metrics/metrics/src/test/java/io/helidon/metrics/HelidonHistogramTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
tjquinno: Seems we need to live with this. See the extensive notes on the first comment about copyrights above.

PR #997, opened 2019-09-06
Add support for using offlineLinks with javadocs. Add timestamps to ma…

barchetta:
…ven output

This changes the way we handle external cross reference links in our javadocs to use
`offlineLinks`. This should speed up the build. See the new README file for more
details.

Looks like this reduces build time by ~15 minutes.


etc/scripts/pipeline-env.sh
```diff
@@ -61,6 +61,10 @@ inject_credentials(){
 }
 
 if [ "${WERCKER}" = "true" ] ; then
-    export MAVEN_OPTS="-Dmaven.repo.local=${WERCKER_CACHE_DIR}/local_repository -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
+    MAVEN_OPTS="-Dmaven.repo.local=${WERCKER_CACHE_DIR}/local_repository"
+    MAVEN_OPTS="${MAVEN_OPTS} -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
+    MAVEN_OPTS="${MAVEN_OPTS} -Dorg.slf4j.simpleLogger.showDateTime=true"
+    MAVEN_OPTS="${MAVEN_OPTS} -Dorg.slf4j.simpleLogger.dateTimeFormat='HH:mm:ss,SSS'"
```
romain-grecourt: I looked at the build and saw the date format not printed properly, it shows `HH:mm:ss,SSS [INFO] ` instead of showing the actual time. This is caused by the simple quotes.

If you change this line to:
```bash
MAVEN_OPTS="${MAVEN_OPTS} -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss,SSS"
```
 then it will work properly ; sorry about that !

romain-grecourt:
Duration `36 minutes, 25 seconds`  ; this is it ! good job !

etc/scripts/pipeline-env.sh
barchetta: That fixed it.

PR #1002, opened 2019-09-06
Fix precedence for explicit sys prop and env var sources.

### LINKED ISSUE #977: Config.Builder ignores explicit ordering of systemProperties() and environmentVariables()
Helidon Version: 1.2.1 
Helidon MP 
JDK version: 1.8.0_211 
OS: MacOS 10.13.6
 Problem Description:
Config.Builder ignores explicit ordering of systemProperties() and environmentVariables(). 
When systemProperties() is before environmentVariables(), the env vars still take precedence.

Note: The Helidon Javadoc documents that the ordering of sources establishes precedence. See Config.Buider at https://helidon.io/docs/latest/apidocs/index.html?overview-summary.html

Steps to reproduce:
1) Create a a project using MP archetype to test this.

2) Change app.greeting to appGreeting in GreetingProvider.java and META-INF/microprofile-config.properties.

3) Change Main.java as follows:

    static Server startServer() {

        return Server.builder()
            .config(Config.builder()
                .sources(
                    systemProperties(),
                    environmentVariables(),
                    classpath("META-INF/microprofile-config.properties"))
                .build())
            .build()
            .start();

4) Build and run the application as follows:
export appGreeting=HelloFromEnvironment
mvn package -DskipTests=true
java -DappGreeting="HelloFromJavaProperty" -jar target/helidon-quickstart-mp.jar

5) Test the app

curl http://localhost:8080/greet
{"message":"HelloFromEnvironment World!"}

batsatt:
Fixes #977.

### DIFF
--- a/config/config/src/main/java/io/helidon/config/BuilderImpl.java
+++ b/config/config/src/main/java/io/helidon/config/BuilderImpl.java
@@ -39,7 +39,6 @@ import io.helidon.common.reactive.Flow;
 import io.helidon.config.ConfigMapperManager.MapperProviders;
 import io.helidon.config.internal.ConfigThreadFactory;
 import io.helidon.config.internal.ConfigUtils;
-import io.helidon.config.internal.MapConfigSource;
 import io.helidon.config.spi.ConfigContext;
 import io.helidon.config.spi.ConfigFilter;
 import io.helidon.config.spi.ConfigMapper;
@@ -48,7 +47,6 @@ import io.helidon.config.spi.ConfigParser;
 import io.helidon.config.spi.ConfigSource;
 import io.helidon.config.spi.OverrideSource;
 
-import static io.helidon.config.ConfigSources.ENV_VARS_NAME;
 import static io.helidon.config.ConfigSources.classpath;
 
 /**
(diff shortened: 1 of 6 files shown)


batsatt:
@tomas-langer Note the comments in the tests that say "Implicit sources always take precedence"; this strikes me as wrong/backwards, but I'm assuming fixing it would be a breaking change.

config/config/src/main/java/io/helidon/config/BuilderImpl.java
```diff
@@ -335,6 +342,19 @@ private ConfigSource targetConfigSource(ConfigContext context) {
         return targetConfigSource;
     }
 
+    private boolean hasMapSource(String name) {
+        if (sources != null) {
+            String qualifiedName = "[" + name + "]";
+            for (ConfigSource source : sources) {
+                if (source instanceof MapConfigSource
+                    && source.description().contains(qualifiedName)) {
```
tomas-langer: If we rely on a description, you should also update the place the description is created to make sure it is not changed.
I would prefer a more deterministic way of finding these sources out - maybe wrapping them in a specific class (`EnvVarConfigSource` and `SysPropConfigSource` or similar) and looking for these types.

batsatt: Yes, agreed, this is a little fragile; I'm just about to lose internet for a couple of days, so will deal with this next week.

PR #1003, opened 2019-09-07
Reactive follow up

tomas-langer:
Please at least merge the first commit (`System.out.println`).


common/reactive/src/main/java/io/helidon/common/reactive/OriginThreadPublisher.java
```diff
@@ -219,7 +219,7 @@ public void submit(U data) {
      */
     public void drain() {
         if (!hasSingleSubscriber.get() && !(completed && queue.isEmpty())) {
-            System.out.println("LEAK: no one registered to consume request");
+            LOGGER.fine(() -> "No one registered to consumer request");
```
romain-grecourt: Sorry about that, that change was merged with a PR with I had already merged it in the branch and forgot to clean it up like I did in the corresponding PR.

tomas-langer: No problem, happens.

Issue #882, opened 2019-08-08
MP Rest Client 1.3


tomas-langer:
Current implementation is 1.2.1

tomas-langer:
Needed steps:

- [ ] Jersey upgrade (see https://github.com/eclipse-ee4j/jersey/pull/4224)
- [ ] Jersey release
- [ ] Integration into Helidon

PR #1010, opened 2019-09-10
Upgrade to Jersey 2.29.1

tomas-langer:

### DIFF
--- a/examples/quickstarts/helidon-quickstart-mp/pom.xml
+++ b/examples/quickstarts/helidon-quickstart-mp/pom.xml
@@ -172,7 +172,7 @@
             <dependency>
                 <groupId>org.glassfish.jersey.media</groupId>
                 <artifactId>jersey-media-json-binding</artifactId>
-                <version>2.29</version>
+                <version>2.29.1</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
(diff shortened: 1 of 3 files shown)


tomas-langer:
Resolves #882

tomas-langer:
Latest version fixes:
- adds REST Client 1.3
- adds new version of hk2 with a performance bug fixed

romain-grecourt:
The pipeline build has the following error during the open-api tck:

```
Caused by: java.lang.NoSuchMethodError: 'java.lang.String org.eclipse.microprofile.rest.client.inject.RegisterRestClient.configKey()'
	at org.glassfish.jersey.microprofile.restclient.RestClientProducer.<init>(RestClientProducer.java:108)
	at org.glassfish.jersey.microprofile.restclient.RestClientExtension.lambda$restClientRegistration$0(RestClientExtension.java:81)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.glassfish.jersey.microprofile.restclient.RestClientExtension.restClientRegistration(RestClientExtension.java:81)
```

tomas-langer:
There are two problems
 1. wrong version of MP Rest Client (easy to fix)
 2. TCK for MP Rest Client do not pass in Helidon (need to investigate)

tomas-langer:
The problem is in loading resources from the classpath by integration with Arquillian.
There is a resource path "/META-INF/..." - it does not work in Helidon TCK tests (would work without the leading slash).
I consider this a problem with our integration for now: will look deeper into this tomorrow

tomas-langer:
Workaround to accept resource path starting with `/` - this works fine outside of Heildon Arquillian container, so fixed it so tests pass

Issue #1009, opened 2019-09-10
Harden some metrics tests susceptible to pipeline speed and timing


tjquinno:
Some (at least two) metrics tests fail intermittently due to timing issues in the pipeline. (We never see these same failures when we build and test on our local systems.)

The failures can greatly disrupt builds.

The two we have seen recently:

- `HelidonConcurrentGaugeTest`
- `HelidonHistogramTest.testPrometheus`

PR #1016, opened 2019-09-11
Add some resiliency to various metrics tests 

tjquinno:
Variability in the pipeline environment leads to intermittent test failures. (issue #1009)

These changes add some robustness to some of the metrics tests to be less susceptible to timing issues from slower-running pipeline nodes.

A new maven profile, `pipeline`, is added. The `build.sh` script now activates that profile and the `metrics2` pom is sensitive to that profile, setting two system properties that give more leeway to two of the tests. 

These changes also address some issues with potential failing asserts being hidden because they were inside `CompletableFutures`.

### DIFF
--- a/etc/scripts/build.sh
+++ b/etc/scripts/build.sh
@@ -48,6 +48,6 @@ inject_credentials
 mvn -f ${WS_DIR}/pom.xml \
     clean install -e \
     -B \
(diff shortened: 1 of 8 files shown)


tjquinno:
The latest commit makes the parallel changes in the older `metrics/metrics` histogram test, similar to the updates in the newer `metrics2/metrics2` test. There is no concurrent gauge test in the older metrics impl.

PR #1036, opened 2019-09-16
RSA encryption changed

Verdent:

### DIFF
--- a/config/encryption/src/main/java/io/helidon/config/encryption/EncryptionUtil.java
+++ b/config/encryption/src/main/java/io/helidon/config/encryption/EncryptionUtil.java
@@ -20,6 +20,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.security.Key;
 import java.security.PrivateKey;
+import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.spec.KeySpec;
 import java.util.Base64;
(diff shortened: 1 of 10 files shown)


config/encryption/src/main/java/io/helidon/config/encryption/EncryptionUtil.java
```diff
@@ -59,6 +60,32 @@ private EncryptionUtil() {
         throw new IllegalStateException("Utility class");
     }
 
+    /**
+     * Decrypt using RSA with OAEP.
+     * Expects message encrypted with the other key.
```
tomas-langer: modify the javadoc - no longer "other" key, but "public"

config/encryption/src/test/java/io/helidon/config/encryption/EncryptionUtilTest.java
```diff
@@ -64,10 +64,10 @@ public static void staticInit() {
         publicKey = kc.publicKey().orElseThrow(AssertionError::new);
     }
 
-    @Test
-    public void testEncryptAndDecryptRsaPrivate() {
-        testPki(privateKey, publicKey, false);
-    }
+//    @Test
```
tomas-langer: why is this commented out? If it no longer makes sense, remove the test.

docs/src/main/docs/microprofile/06_configuration.adoc
```diff
@@ -144,7 +144,7 @@ This approach is based on a pair of keys: a public key which is known to anybody
 - a value encrypted by a public key can only be decrypted by the private key
 - a value encrypted by a private key can only be decrypted by the public key
 
-When using the config encryption filter, you should encrypt the configuration values
+When using the config encryption filter, you have to encrypt the configuration values
```
tomas-langer: why? We provide recommendations - there is no enforcement to encrypt, unless the value is marked as a secret

docs/src/main/docs/microprofile/06_configuration.adoc
```diff
@@ -144,7 +144,7 @@ This approach is based on a pair of keys: a public key which is known to anybody
 - a value encrypted by a public key can only be decrypted by the private key
```
tomas-langer: Remove the part with encrypted by private key, as that is no longer supported.

docs/src/main/docs/microprofile/06_configuration.adoc
Verdent: I meant  it like that you cant select between public and private. That you have to use public for encryption.

PR #1040, opened 2019-09-17
Removes Oracle Maven Repository information

ljnelson:


ljnelson:
We'll see if the build goes through.  @romain-grecourt there may be things that need doing in the build scripts themselves; not sure.

ljnelson:
Looks like the `groupId` for `ucp` changed even though the `version` and `artifactId` elements are the same (!).

ljnelson:
From opening the Oracle Universal Connection Pool jar file, it looks like there's a class in there called `OracleShardingKeyBuilderFactory`.  From looking at the raw `.class` file, it appears to reference a class named `oracle.jdbc.OracleShardingKeyBuilder`.  This makes me think that the `ucp` artifact should require the Oracle JDBC drivers artifact, but it does not.  I'll see if adding that fixes the issue.

ljnelson:
It also appears that the `pom` file in the Oracle JDBC drivers jar requires the `ucp` artifact, but not vice versa.  I think this is exactly backwards, but there's not much we can do here.  I believe my latest commit will fix the build.

Issue #1046, opened 2019-09-20
Implementation for MP Metrics 1.1 does not enforce reusability


tjquinno:
## Environment Details
* Helidon Version: 1.3
* Helidon MP
* JDK version: 11
* OS: MacOS
* Docker version (if applicable):

----------

Using the MP metrics 1.1. support, application code and annotations are incorrectly permitted to register a non-reusable metric more than once. 

In 1.1, all metrics should by default be non-reusable according to the spec. (This changed in 2.0; metrics created programmatically rather than via annotation are reusable by default.) 

The current 1.1 implementation enforces only that the reusability setting of the existing metric registration and the new metric registration be the same. It does not go on to reject a re-registration of the metric if reusability is set to `false`. 

It is possible that developers have existing applications that work by inadvertently relying on this errant behavior but should not work as written. Therefore, we need to be thoughtful about whether to actually fix this or not.

PR #1048, opened 2019-09-20
Fully enforce reusability in MP Metrics 1.1. support

tjquinno:
The code that supports MP Metrics 1.1 does not fully enforce reusability. These changes correct that.

(This PR also addresses some unnecessary references to the internal bridging infrastructure which these classes did not need to use; they can and should depend directly on the MP metrics 1.1 artifacts.)

See issue #1046

### DIFF
--- a/metrics/metrics/src/main/java/io/helidon/metrics/InternalMetadataBuilderImpl.java
+++ b/metrics/metrics/src/main/java/io/helidon/metrics/InternalMetadataBuilderImpl.java
@@ -104,7 +104,7 @@ class InternalMetadataBuilderImpl implements InternalBridge.Metadata.MetadataBui
         if (name == null) {
             throw new IllegalStateException("name must be assigned");
         }
-        return new InternalMetadataImpl(name, displayName, description, type, unit);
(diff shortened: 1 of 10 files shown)


tjquinno:
The (tentative) changes to FT to make the automatically-created metrics reusable are because automatically-derived metric names come from the class name and method name of the annotated method. In the FT unit tests, the `CircuitBreakerBean` uses FT on two overloaded variants of `exerciseBreaker`. Both methods yield the same automatic metric name, so if the corresponding metric is not set up as reusable then, now that the registry code is enforcing the reusability, FT tries to reregister the same metric twice which fails.

tjquinno:
The preceding changes to `MetricsSupport` registers the vendor metrics as reusable. This is needed because the TCK starts the server multiple times, which causes the MP services to be loaded multiple times, but there is no life cycle support in the server for services to respond to server shutdown so they can clean up. With the fixes to reusability enforcement, and with reusability being `false` by default in metrics 1.1, this caused failures in the TCK as the vendor metrics were reregistered.

Issue #154, opened 2018-11-02
JAX-RS resources should have implicit Dependent annotation


tomas-langer:
Currently I must annotate a JAX-RS resource with @Path (for JAX-RS) and @RequestScoped or @ApplicationScoped (for CDI).

As these resources are by default request scoped, maybe this should be automatically used for injection.

If I do not annotate the class with neither of Application or Request Scoped annotations, injection is going to fail in HK2 

When debugging, I could see the CDI actually does inject the values correctly, yet HK2 then subsequently fails that it cannot inject them

## Environment Details
* Helidon Version: 0.10.5-SNAPSHOT
* Helidon MP
* JDK version: 8
* OS: Mac

----------

Stack trace:
```
WARNING: The following warnings have been detected: WARNING: Unknown HK2 failure detected:
MultiException stack 1 of 1
org.glassfish.hk2.api.UnsatisfiedDependencyException: There was no object available for injection at SystemInjecteeImpl(requiredType=JsonWebToken,parent=Runner$TheResource,qualifiers={},position=-1,optional=false,self=false,unqualified=null,942306357)
	at org.jvnet.hk2.internal.ThreeThirtyResolver.resolve(ThreeThirtyResolver.java:75)
	at org.jvnet.hk2.internal.Utilities.justInject(Utilities.java:1012)
	at org.jvnet.hk2.internal.ServiceLocatorImpl.inject(ServiceLocatorImpl.java:1008)
	at org.glassfish.jersey.inject.hk2.AbstractHk2InjectionManager.inject(AbstractHk2InjectionManager.java:231)
	at org.glassfish.jersey.inject.hk2.ImmediateHk2InjectionManager.inject(ImmediateHk2InjectionManager.java:54)
	at org.glassfish.jersey.ext.cdi1x.internal.AbstractCdiBeanSupplier$2.getInstance(AbstractCdiBeanSupplier.java:112)
	at org.glassfish.jersey.ext.cdi1x.internal.AbstractCdiBeanSupplier._provide(AbstractCdiBeanSupplier.java:127)
	at org.glassfish.jersey.ext.cdi1x.internal.GenericCdiBeanSupplier.get(GenericCdiBeanSupplier.java:66)
	at org.glassfish.jersey.inject.hk2.InstanceSupplierFactoryBridge.provide(InstanceSupplierFactoryBridge.java:77)
	at org.jvnet.hk2.internal.FactoryCreator.create(FactoryCreator.java:153)
[...]

PR #1050, opened 2019-09-24
JAX-RS resources now have CDI dependent scope by default.

tomas-langer:


tomas-langer:
Fixes #154

Issue #1055, opened 2019-09-25
Wrong datasource properties show up in JPA example


ljnelson:
Though the JPA guide is correct, there is at least one mostly bad property in the JPA example project.  PR shortly.

PR #1056, opened 2019-09-25
Fixes incorrect datasource property name

ljnelson:


ljnelson:
Fixes #1055.

Issue #716, opened 2019-05-21
Helidon MP Graal native-image


tomas-langer:
This is an issue to gather information about possibility to use Graal native-image with Helidon MP

tomas-langer:
I have created a "pure" CDI example that only starts Weld and uses a single `ApplicationScoped` bean.
The current blocker:
The `beans.xml` processing for beans discovery fails (as Weld tried to locate the file on the file system, where it cannot be).

Next step:
Disable runtime scanning
Use only Jandex for beans discovery (if possible)

tomas-langer:
- [x] Weld
- [x] Helidon WebServer
- [x] Jersey support for WebServer
- [x] Jersey
- [x] Jersey client
- [x] Jersey integration with CDI
- [x] Interceptor support (conceptual for Metrics)
- [ ] REST Client
- [ ] Metrics 2.0
- [ ] Healthchecks
- [ ] Tracing
- [ ] OpenAPI
- [ ] Fault Tolerance
- [ ] Security (MP JWT Auth + others)
- [ ] MP Config (conceptual)
- [ ] Override config from MP Server (Helidon config)
- [ ] Other media types (JSON-P, JSON-B etc.)
- [ ] JPA
- [ ] JTA

tomas-langer:
REST client - the implementation caches config instance that was used during CDI extension. As a result we could not use config passed to the MP Server.
Workaround: The MP configuration uses delegate pattern. We can replace the delegate before starting the CDI container fully. All components that hold the Config instance will get values from the new configuration.

tomas-langer:
REST client - uses method handles, cannot be compiled using native-image

tomas-langer:
Fault tolerance - Unsafe issue:
```
Warning: RecomputeFieldValue.ArrayIndexScale automatic substitution failed. The automatic substitution registration was attempted because a call to sun.misc.Unsafe.arrayIndexScale(Class) was detected in the static initializer of rx.internal.util.unsafe.ConcurrentCircularArrayQueue. Detailed failure reason(s): Could not determine the field where the value produced by the call to sun.misc.Unsafe.arrayIndexScale(Class) for the array index scale computation is stored. The call is not directly followed by a field store or by a sign extend node followed directly by a field store. 
[...]

PR #1063, opened 2019-09-26
Defer datasource configuration

ljnelson:
Adds the ability to push data source configuration until the last possible minute to help with Graal VM.


ljnelson:
In general if you run with `-Djpa.lateConfigurationBinding` and `-Djpa.dataSourceNames=test,foo,bar`, then beans will be set up eagerly anticipating that at some point in the future configuration will be available for `DataSource`s named `test`, `foo` and `bar`, but that configuration will not be actually loaded until `AfterDeploymentValidation` time.

A convenient side effect of this PR is validation of data source configuration whether configuration binding is early or late.

ljnelson:
See also: #716.

tomas-langer:
Closing this PR as there was no progress for a long time and the originating branch is probably deleted.

PR #1064, opened 2019-09-26
WIP: Adapt to changes in MP Metrics 2.1.0

tjquinno:
MP Metrics 2.1.0 brings a changes that affect us:
* Implementations need to be thread-safe, 
* a `Gauge`'s type must extend `Number`


tjquinno:
This PR is obsolete now, given the subsequent changes by MP to move to metrics 2.2.

Issue #1069, opened 2019-09-28
StaticContentSupport problem in helidon-se


piusalfred:
I have a helidon-se powered static website. It works fine except that whenever I pull the server logs there a lot of runtime exception I can not trace.

## Environment Details
* Helidon Version: 1.2.0
* Helidon SE
* JDK version: 11.0.4
* OS: Bionic Beaver 18.04

----------

## Problem Description
After keeping the Webserver running for a time, runtime exceptions are thrown 

## stack trace

```bash
Sep 28 08:46:09 tanganyikabeach systemd[1]: Started The Tanganyika Beach Resort Web Server.
Sep 28 08:46:09 tanganyikabeach java[26739]: [DEBUG] (main) Using Console logging
Sep 28 08:46:10 tanganyikabeach java[26739]: 2019.09.28 08:46:10 INFO io.helidon.webserver.NettyWebServer Thread[main,5,main]: Version: 1.2.0
Sep 28 08:46:10 tanganyikabeach java[26739]: 2019.09.28 08:46:10 INFO io.helidon.webserver.NettyWebServer Thread[nioEventLoopGroup-2-1,10,main]: Channel '@default' started: [id: 0x504b606f, L:/0:0:0:0:0:0:0:0:80
Sep 28 08:46:10 tanganyikabeach java[26739]: WEB server is up! http://localhost:8080/greet
Sep 28 08:49:10 tanganyikabeach java[26739]: 2019.09.28 08:49:10 WARNING io.helidon.webserver.RequestRouting Thread[nioEventLoopGroup-3-2,10,main]: Default error handler: Unhandled exception encountered.
Sep 28 08:49:10 tanganyikabeach java[26739]: java.util.concurrent.ExecutionException: Unhandled 'cause' of this exception encountered.
Sep 28 08:49:10 tanganyikabeach java[26739]:         at io.helidon.webserver.RequestRouting$RoutedRequest.defaultHandler(RequestRouting.java:365)
Sep 28 08:49:10 tanganyikabeach java[26739]:         at io.helidon.webserver.RequestRouting$RoutedRequest.nextNoCheck(RequestRouting.java:351)
Sep 28 08:49:10 tanganyikabeach java[26739]:         at io.helidon.webserver.RequestRouting$RoutedRequest.next(RequestRouting.java:305)
Sep 28 08:49:10 tanganyikabeach java[26739]:         at io.helidon.media.jsonp.server.JsonSupport.accept(JsonSupport.java:114)
[...]

PR #1072, opened 2019-09-29
Fixes NPE in RequestRouting#canonicalize()

ljnelson:

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
@@ -82,12 +82,16 @@ class RequestRouting implements Routing {
     }
 
     private static String canonicalize(String p) {
-        String result = p;
-        if (p.charAt(p.length() - 1) == '/') {
-            result = p.substring(0, p.length() - 1);
-        }
-        if (result.isEmpty()) {
+        String result;
+        if (p == null || p.isEmpty() || p.equals("/")) {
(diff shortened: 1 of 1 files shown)


ljnelson:
This PR makes the `RequestRouting#canonicalize(String)` method handle `null` and empty inputs in an effort to solve one of the issues brought up in #1069.

webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
```diff
@@ -82,12 +82,16 @@ public void route(BareRequest bareRequest, BareResponse bareResponse) {
     }
 
     private static String canonicalize(String p) {
-        String result = p;
-        if (p.charAt(p.length() - 1) == '/') {
-            result = p.substring(0, p.length() - 1);
-        }
-        if (result.isEmpty()) {
+        String result;
```
dansiviter: Two things:
* Should this not have a unit test to avoid issues going forward?
* Any reason why not a more fail-fast approach which is easier to read:
```
    private static String canonicalize(String p) {
        if (p == null || p.isEmpty() || p.equals("/")) {
            return "/";
        }
        int lastCharIndex = p.length() - 1;
        if (p.charAt(lastCharIndex) == '/') {
            return p.substring(0, lastCharIndex);
        }
        return p;
    }
```
I'm interested to know what the benefit is of indenting more or delaying the `return` of the method when you already know the result.

PR #1076, opened 2019-10-02
 Inform Jersey of broken pipe by throwing exception

spericas:
Ensure exceptions are propagated so that Jersey can detect problems writing to a broken pipe. This is very useful when the server streams data in chunks, and possibly over a long period, such as when streaming SSE events. A new test that verifies the fix has been added.

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/BaseProcessor.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/BaseProcessor.java
@@ -76,6 +76,8 @@ abstract class BaseProcessor<T, U> implements Processor<T, U>, Subscription {
             } catch (Throwable ex) {
                 onError(ex);
             }
+        } else {
+            throw new IllegalStateException("Subscriber is closed!");
         }
     }
 
--- a/common/reactive/src/main/java/io/helidon/common/reactive/BaseProcessor.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/BaseProcessor.java
@@ -70,14 +70,13 @@ abstract class BaseProcessor<T, U> implements Processor<T, U>, Subscription {
 
     @Override
     public final void onNext(T item) {
-        if (!subscriber.isClosed()) {
-            try {
-                hookOnNext(item);
-            } catch (Throwable ex) {
-                onError(ex);
-            }
-        } else {
+        if (subscriber.isClosed()) {
             throw new IllegalStateException("Subscriber is closed!");
+        }
+        try {
+            hookOnNext(item);
+        } catch (Throwable ex) {
+            onError(ex);
         }
     }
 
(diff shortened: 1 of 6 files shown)


common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamPublisher.java
```diff
@@ -144,6 +144,9 @@ private void publish(byte[] buffer, int offset, int length) throws IOException {
         } catch (ExecutionException e) {
             complete(e.getCause());
             throw new IOException(e.getCause());
+        } catch (IllegalStateException e) {
```
ljnelson: One minor comment.  Clearly your changes to `onNext()` now throw `IllegalStateException`.  Does anything else in this "wide" `try` block also throw `IllegalStateException`?  If so, is it OK to handle all such `IllegalStateExceptions` here in this manner?  Or should this `IllegalStateException` part of the `catch` block be "scoped tighter" around just the `onNext` call?  I don't know the answer.  I am just raising the point.

spericas: It's a good question. I haven't seen other places, but I will look a bit more closely.

spericas: I've now verified that just `onNext` can throw that exception.

spericas:
I have now removed the WIP label

common/reactive/src/main/java/io/helidon/common/reactive/BaseProcessor.java
```diff
@@ -76,6 +76,8 @@ public final void onNext(T item) {
             } catch (Throwable ex) {
                 onError(ex);
             }
+        } else {
```
tomas-langer: I would consider changing the if statement to remove the negation, e.g. (pseudo):
```
if (sub.closed()) {
  throw ISE;
} else {
  hookOnNext
}
```

As analyzing if statements with negation and else branch is a brain twister

common/reactive/src/test/java/io/helidon/common/reactive/BaseProcessorTest.java
```diff
@@ -80,7 +80,11 @@ public void testOnNextAfterOnComplete() {
         processor.subscribe(subscriber);
         subscriber.request1();
         processor.onComplete();
-        processor.onNext("foo");
+        try {
```
tomas-langer: You can use
```
import static org.junit.jupiter.api.Assertions.assertThrows;
//....
assertThrows(IllegalStateException.class, () -> processor.onNext("foo");
```

common/reactive/src/main/java/io/helidon/common/reactive/BaseProcessor.java
spericas: Will change, don't have a preference either way.

common/reactive/src/test/java/io/helidon/common/reactive/BaseProcessorTest.java
spericas: Sure, that would read better.

PR #1077, opened 2019-10-02
Restore helidon-metrics2 as default metrics library

spericas:
Restore helidon-metrics2 as default metrics library, where it applies. This includes normal dependencies, some tests and archetypes. Exclude `helidon-config-object-mapping` from its pom to ensure compatibility with GraalVM in SE apps.

### DIFF
--- a/metrics2/metrics2/pom.xml
+++ b/metrics2/metrics2/pom.xml
@@ -120,6 +120,12 @@
             <groupId>io.helidon.microprofile.config</groupId>
             <artifactId>helidon-microprofile-config</artifactId>
             <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.helidon.config</groupId>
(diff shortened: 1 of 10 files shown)


metrics2/metrics2/pom.xml
```diff
@@ -120,6 +120,12 @@
             <groupId>io.helidon.microprofile.config</groupId>
             <artifactId>helidon-microprofile-config</artifactId>
             <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
```
ljnelson: Minor comment: make sure to run `mvn dependency:tree -Dscope=runtime` to ensure that this artifact has not been pulled in transitively by some other dependency.  Getting exclusions right often means playing whack-a-mole.

spericas: If it somehow gets pulled, it would prevent SE apps (like quickstart-se) to compile under GraalVM. Those are compiling and running under GraalVM now.

PR #1086, opened 2019-10-08
Service name not needed when tracing is disabled.

tomas-langer:
Related to #1086


tomas-langer:
This PR now silences javadoc plugin, as I was not able to view log output (browser crashed).

PR #1088, opened 2019-10-08
Jackson 2.10.0

barchetta:
Upgrade Jackson to 2.10.0. This should reduce the stream of CVE patches. For more details see 

 https://medium.com/@cowtowncoder/jackson-2-10-features-cd880674d8a2

### DIFF
--- a/3RD-PARTY.txt
+++ b/3RD-PARTY.txt
@@ -2389,7 +2389,7 @@ Version 2.0, January 2004
 http://www.apache.org/licenses/
 
 ========================================================================
-jackson-databind 2.9.9.3
+jackson-databind 2.10.0
 ========================================================================
 COPYRIGHT: Copyright (c) 2007- Tatu Saloranta, tatu.saloranta@iki.fi
 LICENSE: Apache 2.0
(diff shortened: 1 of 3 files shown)


barchetta:
Hmmm...I've seen what looks to be a webserver crash both in my pipeline and now on my local machine with this change. Investigating...

barchetta:
I've reproduced a failure locally:
```
java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/TSFBuilder
        at io.smallrye.openapi.runtime.io.OpenApiSerializer.serialize(OpenApiSerializer.java:106)
        at io.helidon.openapi.OpenAPISupport.formatDocument(OpenAPISupport.java:224)
```
Looks like this class might not exist in 2.10.

ljnelson:
Hmm, no, `TSFBuilder` exists only in 2.10 and later (see http://fasterxml.github.io/jackson-core/javadoc/2.10/com/fasterxml/jackson/core/TSFBuilder.html).

barchetta:
OK, the problem is `microprofile/tests/tck/tck-opentracing` forces some jackson dependencies to 2.9.0 so we end up with a blend of 2.9.0 and 2.10.0 on the classpath.

PR #1092, opened 2019-10-10
Support for Health 2.1

spericas:
Support for MP Health 2.1: (1) New backward compatible flag 'health.backward-compatible' to enable/disable Metrics 1.X JSON support; by default it is set to true. New Health 2.1 TCKs enforce schema validation, so this flag is required. (2) Support to disable all built-in health checks using 'mp.health.disable-default-procedures' property; built-in checks are decorated with a new qualifier for the purpose of filtering when this property is set.

Will update docs before removing WIP label.

### DIFF
--- a/docs/src/main/docs/health/01_health.adoc
+++ b/docs/src/main/docs/health/01_health.adoc
@@ -248,7 +248,7 @@ Routing.builder()
 
 The JSON responses shown above contain properties `"status"` and `"outcome"` with the same
 values. Helidon reports both of these to maintain backward compatibility with older
-versions of the MicroProfile Metrics library. This behavior can be disabled by setting
+versions of MicroProfile Health. This behavior can be disabled by setting
 the property `health.backward-compatible` to `false`, in which case only `"status"`
-is reported. Future versions of Helidon will drop support for older versions of Metrics,
+is reported. Future versions of Helidon will drop support for older versions of Health,
 so it is recommended to rely on `"status"` instead of `"outcome"` in your applications.
--- a/docs/src/main/docs/health/01_health.adoc
+++ b/docs/src/main/docs/health/01_health.adoc
@@ -166,6 +166,7 @@ TIP: Balance collecting a lot of information with the need to avoid overloading
 ----
 {
     "outcome": "UP",
+    "status": "UP",
     "checks": [
         {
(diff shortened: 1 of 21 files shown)


docs/src/main/docs/health/01_health.adoc
```diff
@@ -241,3 +243,12 @@ Routing.builder()
     ]
 }
 ----
+
+=== Strict JSON Output
+
+The JSON responses shown above contain properties `"status"` and `"outcome"` with the same
+values. Helidon reports both of these to maintain backward compatibility with older
+versions of the MicroProfile Metrics library. This behavior can be disabled by setting
```
tjquinno: Should this refer to the Health library, not Metrics?

docs/src/main/docs/health/01_health.adoc
```diff
@@ -241,3 +243,12 @@ Routing.builder()
...
+the property `health.backward-compatible` to `false`, in which case only `"status"`
+is reported. Future versions of Helidon will drop support for older versions of Metrics,
```
tjquinno: Same: health, not metrics?

spericas: Yes, I don't know why I get the two mixed up all the time LOL

docs/src/main/docs/health/01_health.adoc
spericas: It didn't want that module to depend (even at runtime) on CDI. Originally, I put this new qualifier in `health-checks` but there were problems (can't depend on that module statically). Hence the creation of the new module.

spericas:
> Both the SE and MP health modules change to depend on the new `common/health` module. MP already depends on SE, so could `BuiltInHealthCheck` be added to the SE health module and thus be accessible to both while avoiding the creation of a new common module?

It didn't want that module to depend (even at runtime) on CDI. Originally, I put this new qualifier in health-checks but there were problems (can't depend on that module statically). Hence the creation of the new module.

docs/src/main/docs/health/01_health.adoc
spericas: Yeap

tjquinno:
Ah, got it (about the CDI dependency).

PR #1093, opened 2019-10-10
Add support for form parameters

### LINKED ISSUE #784: Add support for Form parameters in Helidon SE
We currently support path parameters and query parameters. It would be useful to add support for form parameters.
This may be either through direct support in `ServerRequest` or through entity supports, such as
`request.content().as(FormParams.class)` (if I remember correctly, form parameters are passed in entity).

tjquinno:
This PR adds support for form parameters in the entity, either as plain text or as URL-encoded text.

Developers will register the new `FormParamsSupport` with their `Routing` which will subsequently allow them to use `request.content().as(FormParams.class)` to get convenient access to the form parameters. `FormParams` extends `Parameters` for consistency with retrieving other types of parameters.

This is _not_ full form data support, which will probably need to wait for the multipart support.

Resolves #784

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/FormParams.java
+++ b/common/http/src/main/java/io/helidon/common/http/FormParams.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


tjquinno:
Thanks to Tomas for the review over zoom. The latest  commit incorporates his suggestions.

Most notably, the new `FormParamsImpl` class now extends `ReadonlyParameters` (instead of basically duplicating that class). Although there is really no behavioral difference between a `FormParamsImpl` instance and a `ReadonlyParameters` instance I decided to keep the separate implementation class because that's a good place for the static regex parsing of the entity content and preparing the map that's passed to the constructor. There's not as good an alternative for that logic.

PR #1096, opened 2019-10-10
Netty 4.1.42

barchetta:
I have verified this works with graalvm-ce-19.2.0 native image.


tomas-langer:
Should resolve #1103 as well

Issue #1099, opened 2019-10-14
Helidon server start-up message reports incorrect start-up timing


tjquinno:
Here are parts of a Helidon start-up log (thanks to Rudy DeBusscher):

```
2019.10.04 08:36:02 INFO io.helidon.microprofile.server.Main Thread[main,5,main]: Logging configured using classpath: /logging.properties
...
2019.10.04 08:36:04 INFO io.helidon.webserver.NettyWebServer Thread[nioEventLoopGroup-2-1,10,main]: Channel '@default' started: [id: 0xdcee65c4, L:/0:0:0:0:0:0:0:0:8080]
2019.10.04 08:36:04 INFO io.helidon.microprofile.server.ServerImpl Thread[nioEventLoopGroup-2-1,10,main]: Server started on http://localhost:8080 (and all other host addresses) in 94 milliseconds.
```
The reported start-up time in the last message does not match up with the delta in the log message timestamps.

PR #1100, opened 2019-10-14
Minimally-invasive refinement to start-up time calculation

tjquinno:
Addresses #1099 

The end-of-startup log message understated the actual start-up time for the MP server.

This PR simply captures the current system time at class-load-time of `ServerImpl` (which we now also trigger as early as possible in `Server.Builder`). This does not include all of the JVM and class loading time but it greatly improves the accuracy of the start-up log message. For example:

```
2019.10.14 12:43:08 INFO org.jboss.weld.Version Thread[main,5,main]: WELD-000900: 3.1.1 (Final)
...
2019.10.14 12:43:12 INFO io.helidon.microprofile.server.ServerImpl Thread[nioEventLoopGroup-2-1,10,main]: Server started on http://localhost:8080 (and all other host addresses) in 4284 milliseconds.
```
The reported start-up time in ms much more closely aligns with the log message timestamps. 

Another option (I looked into this and @batsatt also mentioned it in slack) would have been to use the `RuntimeMXBean` which reports the JVM start-up time. In some informal measurements I was seeing a 40-50 ms cost for that. 

Yet another option (thanks to @spericas) would have collected the start-up time in the built-in `Main` class, but users can provide their own `Main` which would skip that (as Santiago also mentioned).

The changes in the PR give a timing that matches up with the log message timestamps and also gives a reasonable approximation to the full start-up time. The `Server.Builder` and `ServerImpl` classes are loaded early during start-up.

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -148,7 +148,11 @@ public interface Server {
      * Builder to build {@link Server} instance.
      */
     final class Builder {
-        private static final long STARTUP_TIME = ServerImpl.STARTUP_TIME;
+
+        {
+            // Load the initialization start time as early as possible from non-public code.
+            ServerImpl.recordInitStart(System.nanoTime());
+        }
 
         // there should only be one
         private static final AtomicInteger MP_SERVER_COUNTER = new AtomicInteger(1);
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -148,6 +148,8 @@ public interface Server {
      * Builder to build {@link Server} instance.
      */
     final class Builder {
+        private static final long STARTUP_TIME = ServerImpl.STARTUP_TIME;
+
         // there should only be one
         private static final AtomicInteger MP_SERVER_COUNTER = new AtomicInteger(1);
         private static final Logger LOGGER = Logger.getLogger(Builder.class.getName());
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerImpl.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerImpl.java
@@ -484,11 +484,11 @@ public class ServerImpl implements Server {
 
                             if ("0.0.0.0".equals(host)) {
                                 // listening on all addresses
-                                LOGGER.info(() -> "Server initialized on http://localhost:" + port + " (and all other host addresses); "
-                                        + "in " + initializationElapsedTime + " milliseconds.");
+                                LOGGER.info(() -> "Server initialized on http://localhost:" + port + " (and all other host addresses)"
+                                        + " in " + initializationElapsedTime + " milliseconds.");
                             } else {
                                 LOGGER.info(() -> "Server initialized on http://" + host + ":" + port
-                                        + "in " + initializationElapsedTime + " milliseconds.");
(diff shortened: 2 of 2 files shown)


microprofile/server/src/main/java/io/helidon/microprofile/server/ServerImpl.java
```diff
@@ -446,14 +447,13 @@ public Server start() {
...
 
-        long beforeT = System.nanoTime();
         server.start()
                 .whenComplete((webServer, throwable) -> {
                     if (null != throwable) {
                         STARTUP_LOGGER.log(Level.FINEST, "Startup failed", throwable);
                         throwRef.set(throwable);
                     } else {
-                        long t = TimeUnit.MILLISECONDS.convert(System.nanoTime() - beforeT, TimeUnit.NANOSECONDS);
+                        long t = TimeUnit.MILLISECONDS.convert(System.nanoTime() - STARTUP_TIME, TimeUnit.NANOSECONDS);
```
spericas: Isn't it technically possible for a server to be started, stopped and started again? I think we may need to redefine what we call _startup time_ here. It should really be the elapsed time of calling `server.start()`, trying to calculate anything else in this method would be inaccurate; any such calculation needs to happen elsewhere IMO.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
```diff
@@ -148,6 +148,8 @@ static Builder builder() {
      * Builder to build {@link Server} instance.
      */
     final class Builder {
+        private static final long STARTUP_TIME = ServerImpl.STARTUP_TIME;
+
```
spericas: Where is this constant used?

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerImpl.java
tjquinno: I'd originally noticed that, after some code was added to the `start` method, the point where the start time was captured fell _after_ that added work. In the slack discussion  Tomas expressed the opinion that the reported time should include as much work as possible -- he mentioned CDI + the server start time.

Yes, it is conceivable that a server would be started, stopped, and restarted again although that's not the usual use case. Recording the system time at the very beginning of the `start` method results in a reported start-up time of 100 to 300 ms (in informal experiments). That's significantly under the log message timestamp delta, and that kind of mismatch is what prompted this issue originally.

The CDI initialization occurs in `Server.Builder.build()` (unless the caller to `Builder` passes a previously-init'd CDI container which is not our usual case). 

We could separately record the "initialization" time (which would include the CDI initialization) and the server start-up time. During the first start-up the log message could report the overall time and the two components; during a subsequent server restart it could report only the time in `ServerImpl.start` because the other initialization is not repeated in that case.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
tjquinno: The goal is to load the `ServerImpl` class as early as possible so the `STARTUP_TIME` constant there is set as early as possible. We could replace this constant in the `Builder` with a static initializer in the builder to accomplish the same thing.

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerImpl.java
tjquinno: Or, because the server start-up time itself seems to be a relatively small part of the overall initialization time, we could report the overall initialization time only once, at the end of the first invocation of `start`. Rereading what I wrote above it might be overly confusing to users to try to concisely convey the two separate times.

PR #1102, opened 2019-10-16
Config for Helidon 2.0

### LINKED ISSUE #1101: Config meta configuration refactoring
Config meta configuration should not use reflection. Refactor to use service loader. Modify MP examples to use meta conifguration

tomas-langer:
Resolves #1101 

Meta configuration now based on Java service loader
 
- `ConfigSourceProvider` - creates an instance of a config source based on a type (name -such as `file`) and meta configuration
- `PollingStrategyProvider` - creates an instance of a polling strategy function based on a type and meta configuration (the function gets the target type instance and provides a polling strategy, where target type may be `Path` and instance a specific file on a files system)
 - `RetryPolicyProvider` - creates an instance of a retry policy based on type and meta configuration

The `MetaConfig` class is responsible for locating meta configuration, creating config sources, retry policies and polling strategies from metaconfiguration


config/config/src/main/java/io/helidon/config/MetaProviders.java
```diff
@@ -0,0 +1,273 @@
...
+                .addService(new BuiltinOverrideSourceProvider())
+                .build()
+                .asList();
+
+        OVERRIDE_SOURCE_PROVIDERS.stream()
+                .map(OverrideSourceProvider::supported)
+                .forEach(SUPPORTED_OVERRIDE_SOURCES::addAll);
+    }
+
+    private MetaProviders() {
```
Verdent: In case of util classes there should also be an exception fired if this constructor is call by reflection or some other way.

config/config/src/main/java/io/helidon/config/internal/UrlOverrideSource.java
```diff
@@ -127,15 +147,26 @@ protected String uid() {
...
+         * @return updated builder instance
+         */
+        public UrlBuilder url(URL url) {
             this.url = url;
+            return this;
+        }
+
+        @Override
+        public UrlBuilder config(Config metaConfig) {
+            metaConfig.get("url").as(URL.class).ifPresent(this::url);
```
Verdent: I have to say that you are sometimes a bit inconsistent how you use constants. For example in case of KEY_KEY in PrefixedConfigSource you have separate constant for that property but here and on several other places you have the String value written directly to the method.

tomas-langer: The reason is simple - I use in place text to get stuff from config in general, as that is the only place it is used and the readability of code is actually better without a constant.
The `KEY_KEY` comes from original implementation.

config/config/src/main/java/io/helidon/config/MetaProviders.java
tomas-langer: Not really. It is an option. If we do it, we should have a test that validates that. So we do not do it.

config/config/src/main/java/io/helidon/config/MetaConfigFinder.java
```diff
@@ -0,0 +1,127 @@
...
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        return findSource(supportedMediaType, cl, CONFIG_PREFIX);
+    }
+
+    private static Optional<ConfigSource> findMetaConfigSource(Function<String, Boolean> supportedMediaType) {
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        Optional<ConfigSource> source;
+
+        // check if meta configuration is configured using system property
+        String property = System.getProperty("io.helidon.config.meta-config");
```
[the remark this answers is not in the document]
tomas-langer: that may be a good idea, will fix

config/config/src/main/java/io/helidon/config/internal/UrlOverrideSource.java
tomas-langer: Used a constant anyway to be consistent with the rest of the class.

tomas-langer:
Also resolves #1113 
Added MP Config implementation in Helidon SE, as this required the changes in previous commit.
This branch now becomes a 2.0 configuration branch.

config/config/src/main/java/io/helidon/config/UrlConfigSource.java
```diff
@@ -80,14 +75,23 @@
...
     public static UrlConfigSource create(Config metaConfig) throws ConfigMappingException, MissingValueException {
-        return (UrlConfigSource) new UrlBuilder(metaConfig.get(URL_KEY).as(URL.class).get())
-                .init(metaConfig)
+        return builder()
+                .config(metaConfig)
                 .build();
     }
 
+    /**
+     * A new fluent API builder.
```
ljnelson: "API"?

config/config/src/main/java/io/helidon/config/internal/FileOverrideSource.java
```diff
@@ -63,13 +63,29 @@ protected String uid() {
...
+     *
+     * @param metaConfig meta configuration containing the {@code path} and other configuration options
+     * @return a new file override source
+     */
+    public static FileOverrideSource create(Config metaConfig) {
+        return builder().config(metaConfig).build();
+    }
+
+    /**
+     * Create a new fluent API builder.
```
ljnelson: Cut and paste error?  What API are we building?

config/config/src/main/java/io/helidon/config/spi/AbstractMpSource.java
```diff
@@ -0,0 +1,173 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.helidon.common.CollectionsHelper;
+import io.helidon.common.reactive.Flow;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+/**
+ * MP Config source basis. Just extend this class to be both Helidon config source and an MP config source.
+ * @param <S> Type of the stamp of this config source
```
ljnelson: Can `S` be, say, `Object`?  Or is there some subtype we could use instead here?

config/config/src/main/java/io/helidon/config/spi/MetaConfigurableProvider.java
```diff
@@ -0,0 +1,49 @@
...
+ * Configurable object of config that can be loaded using a Java service loader.
+ * This is used by {@link io.helidon.config.MetaConfig}.
+ */
+interface MetaConfigurableProvider<T> {
+    /**
+     * Return true if this provider supports the type of meta-configurable object.
+     * @param type type that is supported (such as {@code file} for {@link io.helidon.config.spi.ConfigSource} meta configurable)
+     * @return {@code true} if this provider can create instances of the type
+     */
+    boolean supports(String type);
```
ljnelson: Is this really a `String` or is it an implied `enum`?

config/config/src/test/java/io/helidon/config/MpConfigTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertThat(values, arrayContaining("a", "b", "c"));
+    }
+
+    @Test
+    void testIntArray() {
+        Integer[] values = config.getValue("mp-list", Integer[].class);
+        assertThat(values, arrayContaining(1, 2, 3));
+    }
+
+    // TODO if I use app1.node1.value instead to test overriding, the override fails
```
ljnelson: Just noting the TODO

config/object-mapping/src/main/java/io/helidon/config/objectmapping/ObjectConfigMapperProvider.java
```diff
@@ -112,11 +113,21 @@
...
+                                                         methodName,
+                                                         String.class);
+
+        if (!method.isPresent()) {
+            method = findStaticMethod(type,
+                                      methodName,
+                                      CharSequence.class);
+        }
+
+        return method.map(handle -> new StringMethodHandleConfigMapper<>(
```
ljnelson: Maybe `new CharSequenceMethodHandleConfigMapper<>()`…?

common/service-loader/src/main/java/io/helidon/common/serviceloader/Priorities.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Find priority from class annotation, or return default if none found.
+     *
+     * @param aClass class to analyzed
+     * @param defaultPriority default priority of this class
+     * @return priority from {@link Priority}or the default provided
+     */
+    public static int find(Class<?> aClass, int defaultPriority) {
+        Priority priorityAnnot = aClass.getAnnotation(Priority.class);
```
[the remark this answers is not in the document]
tomas-langer: Can you send me how to handle CDI programmatic annotations correctly? In case we need that.

config/config/src/main/java/io/helidon/config/BuilderImpl.java
```diff
@@ -35,49 +39,86 @@
...
     private final List<ConfigParser> parsers;
     private boolean parserServicesEnabled;
+    /*
+     * Config filters
+     */
     private final List<Function<Config, ConfigFilter>> filterProviders;
     private boolean filterServicesEnabled;
-    private boolean cachingEnabled;
+    /*
+     * Changes (TODO to be removed)
```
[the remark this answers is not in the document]
tomas-langer: Removed from the code, marked as `@Deprecated` - there is another issue open for this.

config/config/src/main/java/io/helidon/config/BuilderImpl.java
```diff
@@ -546,4 +745,261 @@ private EmptyConfigHolder() {
...
+
+        private MpSourceWrapper(org.eclipse.microprofile.config.spi.ConfigSource delegate) {
+            this.delegate = delegate;
+        }
+
+        @Override
+        public int priority() {
+            String value = delegate.getValue(CONFIG_ORDINAL);
+            if (null != value) {
+                return 101 - Integer.parseInt(value);
```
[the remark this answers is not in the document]
tomas-langer: Added a comment

config/config/src/main/java/io/helidon/config/internal/FileOverrideSource.java
tomas-langer: Not sure. I consider "fluent API" to be the name of the design pattern using Builders and in general: classes, that return the same instance when something is updated.

config/config/src/main/java/io/helidon/config/spi/AbstractMpSource.java
tomas-langer: There is no supertype shared. I have checked the implementations, and we have `Instant`, `byte[]` (and maybe others).
I have not modified this type, it was there on `AbstractSource` since the beginning.

config/config/src/main/java/io/helidon/config/spi/MetaConfigurableProvider.java
tomas-langer: It is a string, as this is designed to be used by third parties to implement their own config sources, retry policies and polling strategies.
The `type` is a short text identifier of such a component.
Even within Helidon, we could use an enum for the "built-ins", but not for the sources that have a dedicated modules (such as `git`, `etcd` etc.)

config/config/src/test/java/io/helidon/config/MpConfigTest.java
tomas-langer: Yep, that is a bug. I will create an issue for it.

tomas-langer: #1182

config/object-mapping/src/main/java/io/helidon/config/objectmapping/ObjectConfigMapperProvider.java
tomas-langer: It is a method that expects `String` - even though it can now invoke methods with `CharSequence` as a parameter. The generic type is `String`, so I would keep the name

config/config/src/main/java/io/helidon/config/AbstractConfigImpl.java
```diff
@@ -218,12 +341,56 @@ ConfigFactory factory() {
...
+
+            if (it instanceof ConfigSources.EnvironmentVariablesConfigSource) {
+                // there is an env var config source
+                this.useEnvironmentVariables = true;
+                break;
+            }
+        }
+
+        // TODO this must be changed, as otherwise we would not get changes in MP
+        //       and why did it work when the MpConfig was a separate implementation?
```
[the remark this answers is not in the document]
tomas-langer: Submitted #1183 to fix this once change support is refactored

config/config/src/main/java/io/helidon/config/AbstractConfigImpl.java
```diff
@@ -172,6 +183,118 @@ void subscribe() {
...
+     * MicroProfile Config methods
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    public <T> T getValue(String propertyName, Class<T> propertyType) {
+        Config config = latestConfig.get();
+        try {
+            return mpFindValue(config, propertyName, propertyType);
+        } catch (MissingValueException e) {
+            throw new NoSuchElementException(e.getMessage());
```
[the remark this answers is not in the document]
ljnelson: For posterity: `throw (NoSuchElement) new NoSuchElementException(e.getMessage()).initCause(e);`

Issue #1140, opened 2019-11-07
Add new helidon-microprofile bundles and deprecate old bundles


barchetta:
## Environment Details
* Helidon Version: 1.3.1
* Helidon MP
----------

## Problem Description
Currently (Helidon 1.3.1) we handle bundles for MicroProfile this way:

We have a bundle for each MicroProfile release, including minor and micro releases. For example: `helidon-microprofile-3.0`.  We ship all previous bundles in subsequent Helidon releases, so we don't break customer's poms when they move to a new version of Helidon. The subcomponents (Metrics, Health, etc.) in those bundles do get updated. So the MicroProfile 1.2 bundle in Helidon 1.3 actually contains Health 2.0 (not 1.0). We consider this OK as long as the newer component version is compatible with older versions.

There are some problems with this:

* We have an accumulation of MicroProfile bundles.
* We need to create a new bundle and update the codebase to use it for every minor MicroProfile release even if that bundle creates no real value. And customers need to update their poms.
* To get a "slimmer" set of MicroProfile dependencies you need to include an older MicroProfile bundle. This works (because the components are up-to-date), but it is a bit funky.

## Proposal 

For Helidon 1.4 create two new bundles that do not have a version number in the artifact ID and deprecate the old bundles (which will be removed in Helidon 2.0). The new bundles will be: `helidon-microprofile`, `helidon-microprofile-core`. 

Exact contents of `helidon-microprofile-core` are TBD, but an example would be: CDI, JAX-RS, JSON-P, JSON-B, MicroProfile Config

When Helidon ships with a new version of MicroProfile, the component versions in these two bundles are updated. If the new version of MicroProfile is incompatible with the previous version (and we can't mitigate that), then we need to wait until we ship a major version of Helidon to include the new version of MicroProfile (and this is already the case).

See PR #1143

tjquinno:
The list of core artifacts described above matches up nicely with what the current `helidon-microprofile-server` artifact depends on.

Shall we just have the new `core` artifact depend on the existing `server`?

The full bundle can depend on `core` plus the other technologies.

tjquinno:
This should have been closed automatically earlier once the PR was merged. Closing it manually now.

Issue #1132, opened 2019-10-31
Create MP 3.2 bundle when MP releases 3.2


tjquinno:
MP has released metrics 2.2, and our PR #1117 addresses that.

MP has yet to release MP 3.2 (which will include metrics 2.2). When they do, we will want to create a new MP bundle for 3.2.

Because MP 3.1 is a flawed release, part of this issue is to _remove_ our 3.1 bundle.

tjquinno:
This morning MP published MP 3.2-RC1. Still watching...

tjquinno:
When we incorporate MP 3.2 for this issue we plan to adopt the new MP bundle naming scheme.

tjquinno:
This issue was resolved with PR #1143 

Note that after this issue was created, we decided to change the bundle naming scheme as described in issue #1140. As of 1.4.0, the numbered bundle names are deprecated and apps should use either `helidon-microprofile` (for the full set of artifacts) or `helidon-microprofile-core` for the very basic set and the app can add specific other dependencies as needed.

PR #1143, opened 2019-11-12
Support MP 3.2; add full and core MP bundles and deprecate existing numbered ones

tjquinno:
As part of support for MicroProfile 3.2 we are adding two new MP bundles: `helidon-microprofile-core` (basically what's needed for the server) and `helidon-microprofile` (everything).

We are also deprecating (but not yet removing) the bundles numbered after the MP version numbers.

The coding changes add the new bundles and also update all dependencies in Helidon that used to refer to the 3.1 MP bundle. The javadoc component had a dependency on the 1.2 bundle that has also been updated.

There are a few components (not included in this PR) which currently depend on some earlier (pre-3.1) MP bundle. We will need to decide for each what to do when we actually remove the numbered bundles in a future Helidon release:

- ./tests/integration/health/mp-disabled/pom.xml
- ./tests/integration/zipkin-mp-2.2/pom.xml
- ./examples/grpc/microprofile/metrics/pom.xml
- ./examples/grpc/microprofile/basic-client/pom.xml
- ./examples/grpc/microprofile/basic-server-implicit/pom.xml
- ./examples/security/attribute-based-access-control/pom.xml
- ./examples/security/idcs-login/pom.xml
- ./examples/security/oidc/pom.xml
- ./examples/microprofile/idcs/pom.xml
- ./examples/microprofile/mp1_1-static-content/pom.xml
- ./examples/microprofile/mp1_1-security/pom.xml
- ./examples/todo-app/backend/pom.xml
- ./microprofile/tests/tck/tck-fault-tolerance/pom.xml
- ./microprofile/tests/tck/tck-metrics/pom.xml

### DIFF
--- a/examples/microprofile/hello-world-explicit/pom.xml
+++ b/examples/microprofile/hello-world-explicit/pom.xml
@@ -41,7 +41,7 @@
     <dependencies>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
-            <artifactId>helidon-microprofile-3.1</artifactId>
+            <artifactId>helidon-microprofile-core</artifactId>
         </dependency>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
--- a/examples/microprofile/hello-world-explicit/pom.xml
+++ b/examples/microprofile/hello-world-explicit/pom.xml
@@ -41,7 +41,7 @@
     <dependencies>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
-            <artifactId>helidon-microprofile-core</artifactId>
+            <artifactId>helidon-microprofile</artifactId>
         </dependency>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
--- a/examples/microprofile/hello-world-implicit/pom.xml
+++ b/examples/microprofile/hello-world-implicit/pom.xml
@@ -38,7 +38,7 @@
     <dependencies>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
-            <artifactId>helidon-microprofile-core</artifactId>
+            <artifactId>helidon-microprofile</artifactId>
         </dependency>
         <dependency>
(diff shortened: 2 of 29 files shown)


barchetta:
For the record, this addresses #1140 and #1132

examples/microprofile/hello-world-explicit/pom.xml
```diff
@@ -41,7 +41,7 @@
     <dependencies>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
-            <artifactId>helidon-microprofile-3.1</artifactId>
+            <artifactId>helidon-microprofile-core</artifactId>
```
tomas-langer: This should be full microprofile as well.

examples/microprofile/hello-world-implicit/pom.xml
```diff
@@ -32,13 +32,13 @@
...
     <description>
-        Microprofile 2.2 example with implicit bootstrapping (server.Main(new String[0])
+        Microprofile example with implicit bootstrapping (server.Main(new String[0])
     </description>
 
     <dependencies>
         <dependency>
             <groupId>io.helidon.microprofile.bundles</groupId>
-            <artifactId>helidon-microprofile-3.1</artifactId>
+            <artifactId>helidon-microprofile-core</artifactId>
```
tomas-langer: This should be full microprofile

examples/microprofile/hello-world-explicit/pom.xml
tjquinno: One benefit of doing what you suggest is that users can then use any of the MP technologies if they use the example `pom`, without having to change which bundle they reference.

The cost is the possible confusion as to why the example - which actually only requires `core` - declares a dependency on the full bundle.

I assume you are OK with that trade-off.

tomas-langer: This is a basic example that shows how to start MP. I expect people use example as a first step to use Helidon - e.g. it is expected that additional MP specs will be used.

PR #1145, opened 2019-11-14
Changes default logging to write to System.out. 

### LINKED ISSUE #1144: Default logging should write to stdout, not stderr
`java.util.logging.ConsoleHandler` directs output to `stderr`, but server output would be better sent to `stdout`.

By fixing this with a new handler, we can also reduce the `logging.properties` boilerplate... 

1. Add a handler that writes to `System.out` and installs the `WebServerLogFormatter` formatter:
```
package io.helidon.common;

import java.util.logging.Level;
import java.util.logging.LogRecord;
import java.util.logging.StreamHandler;

public class HelidonConsoleHandler extends StreamHandler {

    public ConsoleHandler() {
        setOutputStream(System.out);
        setLevel(Level.ALL); // Handlers should not filter, loggers should
        setFormatter(new WebServerLogFormatter());
    }

    @Override
    public void publish(LogRecord record) {
        super.publish(record);
        flush();
    }

    @Override
    public void close() {
        flush();
    }
}
```
2. Change all example logging.properties to use this handler and remove the explicit formatter, reducing the boilerplate (sans comments) to:
```
.level=INFO
handlers=io.helidon.common.HelidonConsoleHandler
java.util.logging.SimpleFormatter.format=%1$tY.%1$tm.%1$td %1$tH:%1$tM:%1$tS %4$s %3$s !thread!: %5$s%6$s%n
```

batsatt:
Fixes #1144.

Note that by setting the level in the handler to `Level.ALL,` the only levels that matter are those on the loggers, which is what nearly everyone expects.

### DIFF
--- a/common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
+++ b/common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common;
+
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.SimpleFormatter;
+import java.util.logging.StreamHandler;
+import java.util.regex.Pattern;
+
+/**
+ * Logging handler that writes to {@link System#out} and uses a formatter that will replace {@code "!thread!"} substrings
+ * with the current thread.
+ */
+public class HelidonConsoleHandler extends StreamHandler {
+
+    /**
+     * Constructor.
+     */
+    public HelidonConsoleHandler() {
+        setOutputStream(System.out);
+        setLevel(Level.ALL); // Handlers should not filter, loggers should
+        setFormatter(new ThreadFormatter());
+    }
+
+    @Override
+    public void publish(LogRecord record) {
+        super.publish(record);
+        flush();
+    }
+
+    @Override
+    public void close() {
+        flush();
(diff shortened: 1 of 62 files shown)


common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.SimpleFormatter;
+import java.util.logging.StreamHandler;
+import java.util.regex.Pattern;
+
+/**
+ * Logging handler that writes to {@link System#out} and uses a formatter that will replace {@code "!thread!"} substrings
+ * with the current thread.
+ */
```
ljnelson: How about: "A {@link StreamHandler} that writes to {@linkplain System#out standard out} and uses a {@link ThreadFormatter} for formatting purposes."

And: "@see ThreadFormatter".

common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
```diff
@@ -0,0 +1,60 @@
...
+public class HelidonConsoleHandler extends StreamHandler {
+
+    /**
+     * Constructor.
```
ljnelson: How about: "Creates a new {@link HelidonConsoleHandler}."

common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public HelidonConsoleHandler() {
+        setOutputStream(System.out);
+        setLevel(Level.ALL); // Handlers should not filter, loggers should
+        setFormatter(new ThreadFormatter());
+    }
+
+    @Override
```
ljnelson: Probably worth documenting what you do in this method.

common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    public void publish(LogRecord record) {
+        super.publish(record);
+        flush();
+    }
+
+    @Override
```
ljnelson: Same as above.

common/common/src/main/java/io/helidon/common/HelidonConsoleHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    public void close() {
+        flush();
+    }
+
+    private static class ThreadFormatter extends SimpleFormatter {
```
ljnelson: This seems like a useful class in general; perhaps make it `public`?

PR #1148, opened 2019-11-17
WIP: MP Reactive Messaging POC

danielkec:
Implementation of https://github.com/oracle/helidon/issues/1206

- MicroProfile Reactive Streams Operators Specification 1.0.1 impl
- MicroProfile Reactive Messaging Specification 1.0 impl
- Kafka connector POC

eclipse/microprofile-reactive-messaging/issues/81
eclipse/microprofile-reactive-messaging/issues/79
eclipse/microprofile-reactive-messaging/issues/76
eclipse/microprofile-reactive-streams-operators/issues/129
eclipse/microprofile-reactive-streams-operators/issues/131


romain-grecourt:
Quick comments:
- Make sure you add/update the unit test for common/reactive so that the code coverage remain high (`mvn test -Pcoverage`)
- Some classes can be kept package protected: LimitProcessor, PeekProcessor, FilterProcessor, MultiMappingProcessor, BaseProcessor. Though BaseProcessor can useful to end-users, one could argue it should actually be public.

microprofile/reactive-streams/src/main/java/io/helidon/microprofile/reactive/MultiStagesCollector.java
```diff
@@ -0,0 +1,231 @@
...
+                processorList.add(new PeekProcessor<Object>(peekConsumer));
+
+            } else if (stage instanceof Stage.Limit) {
+                Stage.Limit limitStage = (Stage.Limit) stage;
+                processorList.add(new LimitProcessor(limitStage.getLimit()));
+
+            } else if (stage instanceof Stage.FlatMap) {
+                Stage.FlatMap flatMapStage = (Stage.FlatMap) stage;
+                Function<?, Graph> mapper = flatMapStage.getMapper();
+                processorList.add(new FlatMapProcessor(mapper));
```
danielkec: @romain-grecourt I am afraid I need Multi processors for backing the MicroProfile Reactive Streams Operators, only way to keep them package protected would be reusing package `io.helidon.common.reactive` in different project, namely `microprofile/reactive-streams` which is possible, it just doesn't seems right

danielkec: There is also a possibility to move MP Reactive Streams operators impl to `common-reactive` but its MP api it seems appropriate to place it among others MP modules

romain-grecourt: OK, I searched for usages of it in this diff when I wrote my comment but didn't find any. If you need that to be used in other modules the only way is to make it public. Modules must have different packages.

romain-grecourt:
Some of your updates to common/reactive are great. Could you submit them with a separate PR so that we can review the changes apart from the microprofile implementation ?

danielkec:
> Some of your updates to common/reactive are great. Could you submit them with a separate PR so that we can review the changes apart from the microprofile implementation ?

Sure oracle/helidon/pull/1260

PR #1159, opened 2019-11-20
Upgrade config libraries 

batsatt:


barchetta:
@batsatt Can you change the jgit version to v4.11.9.201909030838-r ?

PR #1186, opened 2019-12-05
Support for FT version 2.0.2

spericas:
Some new tests and clarifications in the spec. Changes in our implementation related to fallback, async and circuit breaker metrics. A few more tests have been created.


microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceCommand.java
```diff
@@ -422,6 +425,7 @@ private void updateMetricsAfter(Throwable throwable, boolean wasBreakerOpen, boo
             getCounter(method, BREAKER_CALLS_FAILED_TOTAL).inc();
             // If it will open, increment counter
             if (breakerWillOpen) {
+                System.out.println("### incrementing breaker");
```
tjquinno: Do we want to remove this or change it to use a logger?

spericas: We do, it slipped through the cracks. Change pushed now.

PR #1200, opened 2019-12-10
Compile package-info.java always, so rebuild does not trigger recompile.

tomas-langer:


tomas-langer:
I have found the correct maven issue reference and replaced it in the `pom.xml`

Issue #1220, opened 2019-12-12
Update to java11 (Flow, Optional etc.)


tomas-langer:
We have a set of APIs built in Helidon to add features that were missing in Java 8.
Now we have migrated to java 11, we can remove these.

- [x] `Flow` API
- [ ] `OptionalHelper`
- [ ] `CollectionsHelper`
- [ ] `InputStreamHelper`
- [ ] `StackWalker`

tomas-langer:
Hit a bug with Spotbugs/Java11 combination. Java 11 produces different bytecode for "try with resources" that does an additional null check. This triggers a redundant null check rule in Spotbugs.
See https://github.com/spotbugs/spotbugs/issues/756 for details.

Removing this check in Helidon

PR #1222, opened 2019-12-12
Using Flow API from Java, removed Flow from reactive module.

tomas-langer:
See #1220


tomas-langer:
Javadoc no longer fails the build. See #1223 as a follow up issue.

Issue #1218, opened 2019-12-12
Unexpected behavior while processing non-standard requests


spericas:
Reported by @Holubow. 

The following requests that don't start with '/' and resulted in a 500 response match up to "Caused by: java.net.URISyntaxException: Relative path in absolute URI...". Examples:

```
205.219.85.235 - [04/Dec/2019:06:49:26 +0000] "GET CMS HTTP/1.1" 500 0
205.219.85.235 - [04/Dec/2019:06:49:26 +0000] "GET bad397 HTTP/1.1" 500 0
```
My guess is that somehow without the leading slash, the first character is added to the port?

```
205.219.85.235 - [04/Dec/2019:06:49:26 +0000] "GET CMS HTTP/1.1" 500 0
Caused by: java.net.URISyntaxException: Relative path in absolute URI: http://192.168.18.2:8080C/
```

To reproduce you need latest `curl`:

```
curl --HTTP1.1 localhost:8080 --request-target test
```

spericas:
Another example is:

```
curl -H "Content-Type: %" "http://localhost:8080"
```

In this case, the server returns an empty response.

spericas:
Some improvements in #1233.

PR #1233, opened 2019-12-16
Better recovery for invalid URLs and content types

spericas:
See Issue #1218. Return HTTP 400 when appropriate; there are some cases now where we either return an empty response or a 500 error. Exceptions are still logged for further analysis.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
@@ -138,7 +138,13 @@ public class ForwardingHandler extends SimpleChannelInboundHandler<Object> {
                 send100Continue(ctx);
             }
 
-            routing.route(bareRequest, bareResponse);
+            // If a problem during routing, return 400 response
+            try {
+                routing.route(bareRequest, bareResponse);
+            } catch (IllegalArgumentException e) {
+                send400BadRequest(ctx, e.getMessage());
+                return;
+            }
         }
 
         if (msg instanceof HttpContent) {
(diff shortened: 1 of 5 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/RequestRouting.java
```diff
@@ -380,6 +380,10 @@ private void defaultHandler(Throwable t) {
 
                     response.status(Http.Status.INTERNAL_SERVER_ERROR_500);
                 }
+                String message = t.getMessage();
+                if (message != null) {
+                    response.send(message);
```
romain-grecourt: How does that play with the response.send() below ?

spericas: Forgot to add WIP to the title. But that needs to be fixed.

romain-grecourt: Oh ok, no problem.

webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
```diff
@@ -138,7 +138,13 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) {
                 send100Continue(ctx);
             }
 
-            routing.route(bareRequest, bareResponse);
+            // If a problem during routing, return 400 response
+            try {
+                routing.route(bareRequest, bareResponse);
+            } catch (IllegalArgumentException e) {
```
tjquinno: Just wondering... Are there other exceptions besides `IllegalArgumentException` that might be thrown that should also trigger the bad request response? I see that the checking at line 111 is the same.

spericas: I haven't encounter other exceptions, but perhaps. This is intended to be an improvement over what we have now, but perhaps not the last one of these.

PR #1253, opened 2019-12-19
Update exit.on.started value in 1.4

batsatt:
Replace use of unicode character with `'!'` since it doesn't work on some systems.


applications/mp/pom.xml
```diff
@@ -23,7 +23,7 @@
     <parent>
         <groupId>io.helidon.applications</groupId>
         <artifactId>helidon-applications-project</artifactId>
-        <version>1.4.1-SNAPSHOT</version>
+        <version>1.4.1</version>
```
barchetta: Danger! Danger!

batsatt: Huh, not sure what happened here. Will merge.

batsatt: Done.

batsatt: I don't like having that merge here, going to redo since it is such a simple change.

Issue #868, opened 2019-07-30
Automate smoke testing we do on staged release


barchetta:
After doing a release build, we run some tests against the staged repository (https://oss.sonatype.org/content/groups/staging/). It would be good to automate this, at least in the form of a shell script. 

The follow tests are run against the staged bits. So you must configure maven to use the staging repository and the staged (non SNAPSHOT) release of Helidon. 

1. Run both quickstart archetypes, build the resulting projects, run unit tests
2. Start quickstart apps (using jars) and verify they come up and respond
3. Build examples using staged bits
4. Build tests using staged bits

barchetta:
See PR #1257

PR #1257, opened 2019-12-20
Add smoketest script

barchetta:
This adds a script that can be used to smoketest a release that is staged. Fixes issue #868

### DIFF
--- a/etc/scripts/smoketest.sh
+++ b/etc/scripts/smoketest.sh
@@ -1,0 +1,250 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


etc/scripts/smoketest.sh
```diff
@@ -0,0 +1,250 @@
...
+fi
+
+full(){
+  echo "===== Full Test ====="
+  cd ${SCRATCH}
+  quick
+  cd ${SCRATCH}
+
+  echo "===== Cloning Workspace ====="
+  git clone https://github.com/oracle/helidon.git
```
romain-grecourt: Can you get this from the current workspace ? This way if running this from a fork or a different repository this script is still usable.

E.g. `origin_url=$(git remote get-url origin)`

barchetta: Updated script to have a command line option, and if that isn't provide to default to the script's workspace origin.

PR #1274, opened 2020-01-07
Hikari CP metrics

Tomas-Kraus:
+ some fixes and code cleanup


Tomas-Kraus:
Rebased to current master/581-helidon-db

PR #1277, opened 2020-01-08
gRPC minor enhancements

thegridman:
Some minor enhancements to the gRPC support based on feedback whilst using Helidon gRPC to build the Coherence gRPC server and clients.

### DIFF
--- a/grpc/client/src/main/java/io/helidon/grpc/client/GrpcChannelDescriptor.java
+++ b/grpc/client/src/main/java/io/helidon/grpc/client/GrpcChannelDescriptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


grpc/client/src/main/java/io/helidon/grpc/client/GrpcChannelDescriptor.java
```diff
@@ -70,6 +78,33 @@ public int port() {
         return port;
     }
 
+    /**
+     * Get the optional target string to use to resolve channel addresses.
+     *
+     * @return the optional target string to use to resolve channel addresses
+     */
+    public String target() {
```
tomas-langer: If this is optional, than you should return `Optional<String>` - Helidon APIs should not return (or accept) `null`.

grpc/client/src/main/java/io/helidon/grpc/client/GrpcChannelDescriptor.java
```diff
@@ -70,6 +78,33 @@ public int port() {
...
+     * @return the default load balancer policy to use
+     */
+    public String loadBalancerPolicy() {
+        return loadBalancerPolicy;
+    }
+
+    /**
+     * Get the default load balancer policy to use.
+     *
+     * @return the default load balancer policy to use
```
tomas-langer: copy paste from `loadBalancerPolicy`

PR #1280, opened 2020-01-08
Integration of WebSockets POC into Helidon 2.0

spericas:
Support for WebSockets in SE and MP.


webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusSupport.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    /**
+     * Creates executor service for Websocket in MP. No executor for SE.
+     *
+     * @return Executor service or {@code null}.
+     */
+    private static ExecutorService createExecutorService() {
+        if (HelidonFeatures.flavor() == HelidonFlavor.MP && DEFAULT_THREAD_POOL.get() == null) {
```
tomas-langer: I do not think the `.flavor()` method is good. We do not control the time when the flavor is configured and this is not the intention (`HelidonFeatures` task is to list features of Helidon, not to handle the flavor of Helidon).
Also the approach for thread pools is to use one of the existing `ThreadPoolSupplier` classes, that already handle lazy initiialization.
If you do not need an executor in SE, than why is this method in this class? It should be in the CDI extension for WebSockets, and the `TyrusSupport` class should only provider features to handle the SE model (with capability to build the MP model on top of it).

webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusSupport.java
```diff
@@ -0,0 +1,300 @@
...
+         * Process a server request/response.
+         *
+         * @param req an HTTP server request.
+         * @param res an HTTP server response.
+         */
+        @Override
+        public void accept(ServerRequest req, ServerResponse res) {
+            // Skip this handler if not an upgrade request
+            Optional<String> secWebSocketKey = req.headers().value(HandshakeRequest.SEC_WEBSOCKET_KEY);
+            if (!secWebSocketKey.isPresent()) {
```
tomas-langer: Please use `Optional.isEmpty()`

webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusSupport.java
```diff
@@ -0,0 +1,300 @@
...
+                return;
+            }
+
+            LOGGER.fine("Initiating WebSocket handshake ...");
+
+            // Create Tyrus request context and copy request headers
+            RequestContext requestContext = RequestContext.Builder.create()
+                    .requestURI(URI.create(req.path().toString()))      // excludes context path
+                    .build();
+            req.headers().toMap().entrySet().forEach(e ->
```
tomas-langer: use `req.headers().toMap().forEach((key, value) -> ...)`

webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusSupport.java
```diff
@@ -0,0 +1,300 @@
...
+
+            // Setup the WebSocket connection and internally the ReaderHandler
+            Connection connection;
+            if (executorService != null) {
+                try {
+                    // Set up connection and call @onOpen
+                    Future<Connection> future = executorService.submit(() ->
+                            upgradeInfo.createConnection(publisherWriter,
+                                    closeReason -> LOGGER.fine(() -> "Connection closed: " + closeReason)));
+                    connection = future.get();      // Need to sync here
```
tomas-langer: I am not sure I understand this - why do you use an executor service, when you call `get` on the result? You are blocking the current thread anyways, why not just call the statement directly?
Also this seems to be an issue for our WebServer, as you should never block

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
```diff
@@ -69,6 +71,20 @@ static Server create(Class<? extends Application>... applicationClasses) throws
...
 
+    /**
+     * Create a server instance using a Websocket application class.
+     *
+     * @param applicationClass websocket application class
+     * @return server instance to be started
+     * @throws MpException in case the server fails to be created
+     * @see #builder()
+     */
+    static Server create(Class<? extends ServerApplicationConfig> applicationClass) throws MpException {
```
tomas-langer: WebSocket support should not be part of Helidon MicroProfile server.
This adds an unwanted dependency on Tyrus for MP implementation.
Please extract WebSocket support into a separate module.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
```diff
@@ -366,6 +383,20 @@ public Builder addApplication(Application application) {
             return this;
         }
 
+        /**
+         * Registers a WebSocket application in the server. At most one application can be registered.
+         *
+         * @param wsApplication websocket application
+         * @return modified builder
+         */
+        public Builder websocketApplication(Class<? extends ServerApplicationConfig> wsApplication) {
```
tomas-langer: Remove from Server. WebSocket should be handled in a separate module.

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -117,6 +126,9 @@ private void startServer(@Observes @Priority(PLATFORM_AFTER + 100) @Initialized(
         // register static content if configured
         registerStaticContent();
 
+        // register websocket endpoints
```
tomas-langer: Extract into a separate module, use `WebSocketCdiExtension` in that module to setup everything

microprofile/server/src/main/java/module-info.java
```diff
@@ -38,9 +36,15 @@
     // there is now a hardcoded dependency on Weld, to configure additional bean defining annotation
     requires java.management;
 
+    requires io.helidon.microprofile.tyrus;
```
tomas-langer: There should be no dependency on tyrus or websocket in Helidon MP Server.

microprofile/tyrus/src/main/java/io/helidon/microprofile/tyrus/WebSocketCdiExtension.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private WebSocketApplication.Builder appBuilder = WebSocketApplication.builder();
+
+    /**
+     * Collect application class extending {@code ServerApplicationConfig}.
+     *
+     * @param applicationClass Application class.
+     */
+    private void applicationClass(@Observes ProcessAnnotatedType<? extends ServerApplicationConfig> applicationClass) {
+        LOGGER.info(() -> "Application class found " + applicationClass.getAnnotatedType().getJavaClass());
```
tomas-langer: Do not use `info` level, unless it is an information needed for every user on every startup.
This seems to be debug information.
You can list Tyrus applications in info - but only when actually registering them (e.g. by the time we have the complete list).

microprofile/tyrus/src/main/java/io/helidon/microprofile/tyrus/WebSocketCdiExtension.java
```diff
@@ -0,0 +1,92 @@
...
+        appBuilder.applicationClass(applicationClass.getAnnotatedType().getJavaClass());
+    }
+
+    /**
+     * Overrides a websocket application class.
+     *
+     * @param applicationClass Application class.
+     */
+    public void applicationClass(Class<? extends ServerApplicationConfig> applicationClass) {
+        LOGGER.info(() -> "Using manually set application class  " + applicationClass);
```
tomas-langer: see above - do not use info level

microprofile/tyrus/src/main/java/io/helidon/microprofile/tyrus/WebSocketCdiExtension.java
```diff
@@ -0,0 +1,92 @@
...
+        appBuilder.updateApplicationClass(applicationClass);
+    }
+
+    /**
+     * Collect annotated endpoints.
+     *
+     * @param endpoint The endpoint.
+     */
+    private void endpointClasses(@Observes @WithAnnotations(ServerEndpoint.class) ProcessAnnotatedType<?> endpoint) {
+        LOGGER.info(() -> "Annotated endpoint found " + endpoint.getAnnotatedType().getJavaClass());
```
tomas-langer: see above - do not use info level

microprofile/tyrus/src/main/java/io/helidon/microprofile/tyrus/WebSocketCdiExtension.java
```diff
@@ -0,0 +1,92 @@
...
+        appBuilder.annotatedEndpoint(endpoint.getAnnotatedType().getJavaClass());
+    }
+
+    /**
+     * Collects programmatic endpoints .
+     *
+     * @param endpoint The endpoint.
+     */
+    private void endpointConfig(@Observes ProcessAnnotatedType<? extends Endpoint> endpoint) {
+        LOGGER.info(() -> "Programmatic endpoint found " + endpoint.getAnnotatedType().getJavaClass());
```
tomas-langer: see above - do not use info level

webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
```diff
@@ -178,14 +198,49 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) {
...
             }
         }
+
+        // We receive a raw bytebuf if connection was upgraded to WebSockets
+        if (msg instanceof ByteBuf) {
+            if (!isWebSocketUpgrade) {
+                throw new IllegalStateException("Received ByteBuf without upgrading to WebSockets");
+            }
+            // Simply forward raw bytebuf to Tyrus for processing
+            LOGGER.fine("Received ByteBuf of WebSockets connection" + msg);
```
tomas-langer: This looks like a candidate for `finest`, not `fine`.
Also when constructing a message in log levels lower than `info`, please use lambda:
`LOGGER.finest(() -> "..."  + msg);` to avoid string concatenation on every single byte buffer received

microprofile/tyrus/src/main/java/io/helidon/microprofile/tyrus/WebSocketCdiExtension.java
spericas: Yes, these were all misuses of logging for development.

PR #1260, opened 2019-12-27
Make Multi streams compatible with Reactive Streams

danielkec:
- Changes to common reactive module needed for oracle/helidon/issues/1206

PR #1282, opened 2020-01-09
Reactive Streams impl

danielkec:
## [Reactive Streams Operators](https://github.com/eclipse/microprofile-reactive-streams-operators)
[Reactive Streams](https://www.reactive-streams.org/) implementation based on existing [Helidon Common Reactive Library](https://github.com/oracle/helidon/tree/master/common/reactive) compliant with 
[Reactive streams for JVM](https://github.com/reactive-streams/reactive-streams-jvm).

Part of the implementation of oracle/helidon/issues/1206


danielkec:
image

common/reactive/src/main/java/io/helidon/common/reactive/MultiFlatMapProcessor.java
```diff
@@ -137,6 +139,7 @@ public void onError(Throwable t) {
 
     @Override
     public void onComplete() {
+        System.out.println("outer onComplete");
```
spericas: println?

danielkec: :man_facepalming: Thx

danielkec:
Rebased on @tomas-langer 's native image changes in the master, sorry for force-push

danielkec:
@olotenko 
> Why is this declared `R` if `onNext` expects it to be `X`? (I know, this change is just adapting old to new, but still)

I can't see to which file is the comment pointing, but I guess this is about FlatMapProcessor, its a fix of the previous PR #1260 , its different/inner subscriber
```java
        public InnerSubscriber<? super X> executeMapper(U item) {
         ...
      }

    private class InnerSubscriber<R> implements Flow.Subscriber<R> {
        ...
        public void onNext(R o) {
            Objects.requireNonNull(o);
            MultiFlatMapProcessor.this.subscriber.onNext((X) o);
            ...
```

olotenko:
Right, that's the place. Note that `X` is still available in `onNext` of the `InnerSubscriber<R>`, so that `onNext` knows what `X` is, and expects the item to be of type `X`. So it seems only suitable to declare it implementing `Flow.Subscriber<X>`, not introducing a new type parameter `R`.

Mind you, I am not blocking the merge, just suggesting what may be a good idea to review as a future improvement.

danielkec:
@olotenko Thx a lot, Hybrids with default methods are great idea, why I didn't think of it! Also thanks for generics in the flatMap. 
You are right about over-usage of Atomic* but I wouldn't get rid of volatiles in the processors so fast, its true that by the spec some of the signals must be executed [serially](https://github.com/reactive-streams/reactive-streams-jvm#1.3), but reactive streams can incorporate to the stream 3rd party publishers/subscribers/processors and those can be implemented in "various" ways. It seems to me better to be defensive in this case.

danielkec:
Apology for force push, needed rebase on shrinkwrap upgrade in master

danielkec:
> This is not what has been discussed by email.
> 
> This is a nonsensical implementation of that rule in that spec.

It's just a quick fix passing tck tests so we are able to move forward, not the final solution.

danielkec:
> The implementation is not what has been discussed.
> 
> Please, split into smaller commits: one for changes to `BaseProcessor`, others for other changes.
> 
> [60739a8#diff-ca92013a99c748b99a9e08660b9dea17R99](https://github.com/oracle/helidon/commit/60739a8b611d862b2dc5ba44b7ffa05b6b94d66f#diff-ca92013a99c748b99a9e08660b9dea17R99) - racy. This should have been called only when `this.subscription` has been set.

it is 
```java
    private void tryOnSubscribe() {
        if (Objects.nonNull(subscription) && subscriber.tryOnSubscribe(this)) {
            if (done) {
                tryComplete();
            }
        }
    }
```

tomas-langer:
Hi Alex, is there any blocking issue that would prevent us from merging this pull request, or can we create a follow up issue (or issues) to fix some of your comments?
We need to move forward with the MP messaging work that depends on these changes.

akarnokd:
Why did you implement operators as `Flow.Processor`s? Implementing one implies you either want to multicast events to any number of downstream consumers or requires you to ensure there is at most one downstream consumer that can subscribe.

danielkec:
> Why did you implement operators as `Flow.Processor`s? Implementing one implies you either want to multicast events to any number of downstream consumers or requires you to ensure there is at most one downstream consumer that can subscribe.

Hi @akarnokd , sure second subscribe signals onError with `IllegalStateException` to second subscriber which is initialized with empty subscription

akarnokd:
Yes, I saw those. 

Still, I don't see why implement operators as `Flow.Processor`s. Do you intend to drive such chains via `onNext` calls imperatively somewhere? 

With a processor chain, you trigger a subscribe() storm while you are still assembling `map`s and `filter`s, which could trigger side-effects, resource utilization or even failure before you are ready to consume the whole sequence with a `Flow.Subscriber`. This is similar to how composing `CompletionStage` operators actually can race with the values in each of those stages. In addition, you are not allowed to drive individual `Flow.Processor`s in a chain externally because that could violate the serial requirement of `onXXX`s.

You can have simply a chain of `Flow.Publisher`s as operators and if necessary, have the upmost source be a `Flow.Processor` for imperative item emission.

akarnokd:
If you need one example why a chain of `Flow.Processor`s don't work: repeating or retrying a chain is not possible because the intermediate `Flow.Processor`s will be in terminal state. You'd have to recreate the entire chain from scratch with fresh `Flow.Processor`s to allow a new run. 

With `Flow.Publisher`s, they hold the blueprint of the operators thus a retry due to an error is straightforward and requires no manual chain recreation.

olotenko:
@akarnokd we can get there in stages. Essentially, there are two types of `Processor`: given `A`, produces `B` and given `Publisher<A>` produces `A`. Because the `Processor` is a structure-preserving map, we can replace a chain of mapping processors with a processor with a chain of functions for a map.

akarnokd:
> Because the Processor is a structure-preserving map, we can replace a chain of mapping processors with a processor with a chain of functions for a map.

Do you mean, `source.map(f1).map(f2)` -> `source.map(v -> f2.apply(f1.apply(v)))`?

I don't see how with `Processor`s because when you call `map(f1)`, it will immediately subscribe a processor to `source` at which point it is very difficult to replace that connection.

olotenko:
@akarnokd we'll get there. The focus of this work here is to get it conform to the spec better, less racy, and better performance for core operations. Refactoring how chaining is done is a separate concern.

akarnokd:
> conform to the spec better, less racy, and better performance for core operations

I'm certain half of the trouble is due to this processor-chaining design where you have to expect races.

I can contribute such spec-conforming, race-free, high performance implementations. I just wanted to clear the situation with the design philosophy behind the current processor-chaining code. I.e., it is not set in stone due to some higher level requirements from stakeholders or something.

danielkec:
Sorry for force push to anyone tracking this branch, it's rebase on the master

akarnokd:
How is it not emitted serially?

olotenko:
`request` can be during `onNext` and can be concurrent. Emitting `onError` concurrently breaks the guarantee that all `on*` invocations are in total order.

akarnokd:
Wrong. The atomic state transition of [addRequest](https://github.com/oracle/helidon/commit/3f610172aefaefb890af50bf8a84a7b555977003#diff-6e58085854ee1341e169599585b70a21R179) with requiring the [transition from zero](https://github.com/oracle/helidon/commit/3f610172aefaefb890af50bf8a84a7b555977003#diff-6e58085854ee1341e169599585b70a21R87) ensures there is only one thread [enters the emission section](https://github.com/oracle/helidon/commit/3f610172aefaefb890af50bf8a84a7b555977003#diff-6e58085854ee1341e169599585b70a21R90-R155).

olotenko:
Well, it's more subtle than that, because it also interacts with `canceled`, but ok, yes, it is serial then.

akarnokd:
Yes, the TCK expects a clear failure for non-positive request amounts which means the `Publisher` should not fail or complete on its own or get cancelled before that. This is one of the pain points of the TCK.

olotenko:
That requirement ("should not fail or complete on its own or get cancelled before that") is non-enforceable, and the code you submitted does not guarantee that either.

Eg `cancel` followed synchronously by `request(-1)` - `request` wins, and will produce `onError`. Whereas the spec requires a cancelled `Subscription` to behave as no-op. I treat this as not a bug, because there is no requirement for `cancel` and `request` to be observed in a total order.

But the part where you say "should not complete or error before that" is just not enforceable. If the `Publisher` has passed the branch where it determined `!hasNext`, then a `request(-1)` will not produce `onError`. I treat this as not a bug, just the nature of concurrent systems. But this means you can't enforce that claim.

All the `try-catch` can, and possibly should, be fused into one. The checks for `canceled` are unnecessary for the above reasons. The spec allows `cancel` and `request(...)` to be observed "eventually". There is no requirement to observe them immediately. So if it races against other `on*`, there is no requirement to prefer issuing error produced by a bad value in `request(...)`.

akarnokd:
> Eg cancel followed synchronously by request(-1) - request wins, and will produce onError. Whereas the spec requires a cancelled Subscription to behave as no-op.

Yes and such bad request amounts should not go unnoticed because it means there is a bug somewhere in the chain. Since there is currently no established approach in the module to not lose exceptions, this is the best it can be done while other operators are being rewritten. We are discussing the possibility of a global error consumer on Slack.

> All the try-catch can, and possibly should, be fused into one. The checks for canceled are unnecessary for the above reasons.

With a general `Iterator`, it could take an arbitrary time for `hasNext` and `next` to return, which if the sequence get canceled due to timeout, would enter the other method and wait even more. This setup bails out eagerly.

olotenko:
> Yes and such bad request amounts should not go unnoticed

Yes, but you can't guarantee that after `cancel` has been fired (the spec requires no-op behaviour, and the same is true with all implementations - after `canceled` has been observed, negative `request` values will not be notified via `onError`), or `Publisher` committed to fire `onError` / `onComplete`. In other cases it will be noticed. The time to execute `hasNext` / `next` is immaterial for correctness. Given these premises, there is no proof eager notification is any better than lazy.

olotenko:
"global error consumer" is a "unhandled exception handler" and should be at the root of any thread.

akarnokd:
The reactive foundation of Helidon is currently incomplete and is not yet prepared for all corner cases within and outside the spec and the TCK. This implementation passes the TCK and enables the development of more operators.

> The time to execute hasNext / next is immaterial for correctness.

They pose practical considerations.

> "global error consumer" is a "unhandled exception handler" and should be at the root of any thread

If you control all the threads that may come into contact with the reactive operators.

olotenko:
> They pose practical considerations.

:) what can be more practical than fencing off problematic implementations instead of punishing everyone? Eg `EagerCancellationIterablePublisher`, or `TimedIterableWrapper` to be used with `Iterables` that are problematic.

akarnokd:
I welcome you to benchmark implementations with and without those volatile reads.

olotenko:
:) simpler linear code with fewer things to think about, rather. It's not just a volatile read here, too.

danielkec:
Continuation of alternative cancelation strategy moved to oracle/helidon/issues/1441

PR #1287, opened 2020-01-10
MP Reactive Messaging impl

danielkec:
Part of the implementation of oracle/helidon/issues/1206
Depends on #1282


danielkec:
image

microprofile/messaging/src/main/java/io/helidon/microprofile/messaging/MessagingCdiExtension.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private void deploymentValidation(@Observes AfterDeploymentValidation event) {
+        Errors.Collector errors = channelRouter.getErrors();
+        boolean hasFatal = errors.hasFatal();
+        Errors errorMessages = errors.collect();
+        if (hasFatal) {
+            throw new DeploymentException(errorMessages.toString());
+        } else {
+            errorMessages.log(LOGGER);
+        }
```
danielkec: @tomas-langer Error collection for signature resolving and validation

microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AbstractCDITest.java
```diff
@@ -0,0 +1,195 @@
...
+    @AfterEach
+    public void tearDown() {
+        try {
+            singleServerReference.stop();
+            cdiContainer.close();
+        } catch (Throwable t) {
+            //emergency cleanup see #1446
+            stopCdiContainer();
+        }
+    }
```
danielkec: @tomas-langer Tear down is double checked to avoid cross test pollution related to #1446

microprofile/messaging/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    <name>Helidon MicroProfile Reactive Messaging</name>
+    <description>
+        Helidon MicroProfile Reactive Messaging
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.eclipse.microprofile.reactive.messaging</groupId>
+            <artifactId>microprofile-reactive-messaging-api</artifactId>
+            <version>1.0</version>
```
tomas-langer: Versions MUST be managed in `dependencies/pom.xml`, mark PR with tag `dependencies` and please add Joe Di Pol as a reviewer.

tomas-langer: This must also be fixed in `microprofile/reactive-streams/pom.xml`

microprofile/messaging/pom.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
```
tomas-langer: Remove ` All rights reserved.` from all copyrights that are updated.

microprofile/messaging/src/main/java/module-info.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Microprofile metrics implementation.
```
tomas-langer: Definitely not MP Metrics ;)

microprofile/messaging/src/main/java/module-info.java
```diff
@@ -0,0 +1,35 @@
...
+module io.helidon.microprofile.messaging {
+    requires java.logging;
+
+    requires static cdi.api;
+    requires static javax.inject;
+    requires static java.activation;
+    requires javax.interceptor.api;
+    requires io.helidon.config;
+    requires org.reactivestreams;
+    requires microprofile.reactive.messaging.api;
```
tomas-langer: All APIs that should be visible for the user should be `transitive` - that will make them imported to the using module as well.

microprofile/messaging/src/main/java/io/helidon/microprofile/messaging/AbstractMessagingMethod.java
```diff
@@ -0,0 +1,126 @@
...
+        Optional.ofNullable(method.getAnnotation(Acknowledgment.class))
+                .map(Acknowledgment::value)
+                .filter(s -> !type.getSupportedAckStrategies().contains(s))
+                .ifPresent(strategy -> {
+                    errors.fatal(String.format("Acknowledgment strategy %s is not supported for method signature: %s",
+                            strategy, type));
+                });
+    }
+
+    public void init(BeanManager beanManager, Config config) {
```
tomas-langer: Methods should not be declared `public` in a package private class unless implementing methods from iface.

microprofile/messaging/src/main/java/io/helidon/microprofile/messaging/CompletableQueueOverflowException.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ *
+ */
+
+package io.helidon.microprofile.messaging;
+
+/**
+ * Raised when {@link io.helidon.microprofile.messaging.CompletableQueue} max size is reached.
+ */
+public class CompletableQueueOverflowException extends RuntimeException {
```
tomas-langer: Do we really need so many types of exceptions? 
I think we do not expect anybody to catch these exceptions specifically, so we should be fine with a `MessagingException` and if needed a `MessagingDeploymentException` and describe the problem in `message`, rather then creating a typed exception for each case.

microprofile/messaging/src/main/java/io/helidon/microprofile/messaging/AbstractMessagingMethod.java
danielkec: fixed

dependencies/pom.xml
```diff
@@ -485,6 +488,21 @@
...
+            <dependency>
+                <groupId>org.eclipse.microprofile.reactive-streams-operators</groupId>
+                <artifactId>microprofile-reactive-streams-operators-core</artifactId>
+                <version>${version.lib.microprofile-reactive-streams-operators-core}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.eclipse.microprofile.reactive.messaging</groupId>
+                <artifactId>microprofile-reactive-messaging-api</artifactId>
+                <version>${version.lib.microprofile-reactive-messaging-api}</version>
+            </dependency>
```
danielkec: FYI @barchetta

microprofile/tests/tck/tck-messaging/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+            <artifactId>smallrye-reactive-streams-operators</artifactId>
+            <version>1.0.10</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.reactivex.rxjava2</groupId>
+            <artifactId>rxjava</artifactId>
+            <version>2.2.13</version>
+            <scope>test</scope>
+        </dependency>
```
danielkec: @tomas-langer Is it possible to not moving those test dependencies to dependency/pom.xml? Using SmallRye operators for messaging tck as workaround for eclipse/microprofile-reactive-messaging/issues/95 , SmallRye operators depends on different version of rxjava, than Helidon depend on already.

microprofile/messaging/src/main/java/io/helidon/microprofile/messaging/CompletableQueueOverflowException.java
danielkec: fixed

microprofile/tests/tck/tck-messaging/pom.xml
tomas-langer: I would agree to keep these dependencies explicit in this file only, as this is a workaround for a TCK issue (that is already reported). 
@barchetta do you agree?

barchetta: I agree

PR #1295, opened 2020-01-13
JWT-Auth for native-image.

tomas-langer:


security/security/src/main/java/io/helidon/security/Security.java
```diff
@@ -1100,10 +1108,47 @@ private boolean notReservedProviderKey(Config config) {
...
+            }
+            if (providerClass.equals(OutboundSecurityProvider.class)) {
+                return outboundProviders.isEmpty();
+            }
+            if (providerClass.equals(AuditProvider.class)) {
+                return auditProviders.isEmpty();
+            }
+            if (providerClass.equals(SubjectMappingProvider.class)) {
+                return subjectMappingProvider == null;
+            }
```
danielkec: `subjectMappingProvider.isEmpty()`

tomas-langer: It is not `Optional` in `Builder`, only in `Security`. This method is on `Builder`

danielkec: My bad sorry

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthCdiExtension.java
```diff
@@ -139,13 +150,47 @@ public void registerClaimProducers(@Observes AfterBeanDiscovery abd, BeanManager
...
+                .map(LoginConfig::authMethod)
+                .noneMatch("MP-JWT"::equals);
+
+        if (notNeeded) {
+            return;
+        }
+
+        if (security.securityBuilder().hasProvider("mp-jwt-auth")) {
+            return;
+        }
```
Verdent: I would say, that this check should be right after obtaining security. That mean somewhere around line 171.

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthCdiExtension.java
```diff
@@ -139,13 +150,47 @@ public void registerClaimProducers(@Observes AfterBeanDiscovery abd, BeanManager
...
+
+        security.securityBuilder()
+                .addProvider(JwtAuthProvider.create(config), "mp-jwt-auth");
```
Verdent: "mp-jwt-auth" should be as constant.

PR #1303, opened 2020-01-15
Work around shrinkwrap resolver issue seen in pipeline builds

barchetta:
Just seeing if this change caused the recent build problems.


barchetta:
This was used for debugging. Replaced by #1305

Issue #1302, opened 2020-01-15
OIDC cookie configuration to prevent infinite redirects


tomas-langer:
The OIDC example and documentation should be updated.
The following configuration should be default when doing redirects with OIDC servers:
`cookie-same-site=Lax`

If not configured (current state), the browser ignores `Set-Cookie` from Helidon and redirects without it. That causes the first request to fail with infinite redirects.
Once the `Set-Cookie` is configured

FWiesner:
Also, one more observation: with the default IDCS client configuration, Helidon logs
```Jan. 15, 2020 3:11:58 NACHM. io.helidon.security.providers.idcs.mapper.IdcsRoleMapperProviderBase processServerResponse
WARNING: Cannot read groups for user "florian.wiesner@oracle.com". Response code: 401, entity: {"schemas":["urn:ietf:params:scim:api:messages:2.0:Error","urn:ietf:params:scim:api:oracle:idcs:extension:messages:Error"],"detail":"Resource-Server-Demo-Florian is not allowed to CREATE /Asserter.","status":"401","urn:ietf:params:scim:api:oracle:idcs:extension:messages:Error":{"messageId":"error.common.common.accessDenied"}}
Jan. 15, 2020 3:11:58 NACHM. io.helidon.security.DefaultAuditProvider lambda$logEvent$1
[...]

PR #1313, opened 2020-01-17
Fixed issues with IDCS and OIDC provider.

tomas-langer:
Cookie now configured to prevent infinite redirects.
WebServer security for scopes works as it should.
MP integration with WebServer security works as it should.
Examples are clearly separated between SE and MP.

Solves #1302 for `master` branch


examples/microprofile/idcs/src/main/java/io/helidon/examples/microprofile/security/idcs/IdcsMain.java
```diff
@@ -35,13 +34,16 @@ private IdcsMain() {
...
 
         System.out.println("Endpoints:");
         System.out.println("Login");
         System.out.println("  http://localhost:7987/rest/login");
         System.out.println("Full security with scopes and roles (see IdcsResource.java)");
         System.out.println("  http://localhost:7987/rest/scopes");
+        System.out.println("A protected reactive service (see application.yaml - security.web-server)");
+        System.out.println("  http://localhost:7987/reactive");
+        System.out.println("A protected static resource (see application.yaml - security.web-server");
+        System.out.println("  http://localhost:7987/web/resource.html");
```
FWiesner: shouldn't the string be rendered from the frontend URI in application.yaml?

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -260,7 +262,7 @@
     static final String DEFAULT_COOKIE_PATH = "/";
     static final boolean DEFAULT_COOKIE_HTTP_ONLY = true;
     static final boolean DEFAULT_COOKIE_SECURE = false;
-    static final String DEFAULT_COOKIE_SAME_SITE = "Strict";
+    static final String DEFAULT_COOKIE_SAME_SITE = "Lax";
```
FWiesner: I think the change in the defaults makes sense with Oracle-apps that use IDCS. Nevertheless I would add a textual warning on startup that the cookie policy is now Lax instead of Strict. The developer might then want to turn off the warning or change to `Strict`, if possible. I wouldn't do this as undocumented/unalerted change. Certainly this has to go into the release notes as well

examples/microprofile/idcs/src/main/java/io/helidon/examples/microprofile/security/idcs/IdcsMain.java
tomas-langer: That is a possibility, though these printouts are really for the sake of the example. In a real application sitting in a container, such messages are useless. So I expect them to be removed.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
tomas-langer: You are right

tomas-langer: Although this problem is not specific to IDCS, it happens with any OIDC provider.

tomas-langer: For version 2.0 I will document this in changelog. For version 1.x I will add a warning message.

FWiesner: > Although this problem is not specific to IDCS, it happens with any OIDC provider.

any OIDC provider that is on a separate domain

tomas-langer: True. Maybe I should be more clever about it and just compare frontend host with the host of the OIDC provider?

tomas-langer: Done and logging an info in case this is triggered.

PR #1314, opened 2020-01-17
Fixed issues with IDCS and OIDC provider - 1.x

tomas-langer:
Cookie now configured to prevent infinite redirects.
WebServer security for scopes works as it should.
MP integration with WebServer security works as it should.
Examples are clearly separated between SE and MP.

Signed-off-by: Tomas Langer tomas.langer@oracle.com

Solves #1302 for `helidon-1.x` branch


examples/microprofile/idcs/src/main/java/io/helidon/examples/microprofile/security/idcs/IdcsMain.java
```diff
@@ -35,13 +38,29 @@ private IdcsMain() {
...
+    }
+
+    private static Config buildConfig() {
+        return Config.builder()
+                .sources(
+                        // you can use this file to override the defaults that are built-in
+                        file(System.getProperty("user.home") + "/helidon/conf/examples.yaml").optional(),
+                        // in jar file (see src/main/resources/application.yaml)
+                        classpath("application.yaml"))
+                .build();
```
FWiesner: I think this is a very important feature, especially when the app will run from Kubernetes and might get an override config seeded. Probably you should feature this more prominently in other places as well and you might also change the default from a user home example to a more Kubernetes-ish one.

examples/microprofile/idcs/src/main/resources/application.yaml
```diff
@@ -34,18 +34,27 @@ security:
...
-      client-id: "${ALIAS=security.properties.idcs-client-id}"
-      client-secret: "${ALIAS=security.properties.idcs-client-secret}"
-      identity-uri: "${ALIAS=security.properties.idcs-uri}"
-      # A prefix used for custom scopes
-      scope-audience: "http://localhost:7987/test-application"
-      proxy-host: "${ALIAS=security.properties.proxy-host}"
-      frontend-uri: "${ALIAS=security.properties.frontend-uri}"
-      # Retrieve roles
-      idcs-roles: true
+    - abac:
```
FWiesner: does the empty abac key do something? If so, please add a comment. Otherwise probably remove it

examples/security/idcs-login/src/main/resources/application.yaml
```diff
@@ -46,11 +46,13 @@ security:
...
         identity-uri: "${ALIAS=security.properties.idcs-uri}"
-      cache-config:
-        cache-timeout-millis: 100000
   web-server:
     # protected paths on the web server - do not include paths served by Jersey, as those are protected directly
     paths:
-    - path: "/rest/profile"
-      methods: ["get"]
-      authenticate: true
+      - path: "/rest/profile"
```
FWiesner: no idea whether this is defined by MicroProfile or native Helidon, but why don't you accept an array of paths here to use the same configuration multiple times?

security/providers/abac/src/main/java/io/helidon/security/providers/abac/AbacProvider.java
```diff
@@ -224,39 +234,53 @@ private void validateCustom(EndpointConfig epConfig, Errors.Collector collector)
...
-                            List<String> unsupportedKeys = new LinkedList<>();
-
-                            for (String key : theMap.keySet()) {
-                                attributes++;
-                                if (!supportedConfigKeys.contains(key)) {
-                                    unsupported++;
-                                    unsupportedKeys.add(key);
-                                }
-                            }
+        config.config(CONFIG_KEY)
```
FWiesner: same as on the other PR... I wonder whether there should be an orElse to warn the developer

examples/microprofile/idcs/src/main/java/io/helidon/examples/microprofile/security/idcs/IdcsMain.java
tomas-langer: We do plan to update our example to use meta configuration with description of how to override.
This is really for our developers - so we do not include our security setup specific URLs to the source base. 
There are examples specifically for configuration that show the capabilities. 
To use this in a k8s with config maps, the change is simple - just replace the path in the code with the file path the config map is mounted to...

examples/microprofile/idcs/src/main/resources/application.yaml
tomas-langer: yes it does - it adds the `ABAC` provider to the list of security providers used

tomas-langer: I have added a comment to each of these used with OIDC/IDCS in examples

examples/security/idcs-login/src/main/resources/application.yaml
tomas-langer: That is not implemented (and it is Helidon SE security integration).
If you feel this is useful, you can submit an enhancement issue.

security/providers/abac/src/main/java/io/helidon/security/providers/abac/AbacProvider.java
tomas-langer: I am not sure I understand - the ABAC provider does not require any configuration, so there is not much to do here.

examples/security/idcs-login/src/main/resources/application.yaml
FWiesner: will do

security/providers/abac/src/main/java/io/helidon/security/providers/abac/AbacProvider.java
FWiesner: well, say some developer has the same thought about the empty `abac` key in the `application.yaml`. Normally an empty key - especially when you come from Spring Boot - is having the connotation of a no-op. So, if you remove the `abac` key it would be helpful to get at least on FINE or FINER a warning that tells you that ABAC support is turned off

tomas-langer: It works slightly differently.
If you add the `abac` key to the configuration, security would pick the security provider up and configure it.
If remove the `abac` key from configuration, security will never try to initialize the ABAC provider.

FWiesner: yes. But many people will expect that

```yaml
   abac:
```
would be a no-op.

With Spring the pattern normally is
```yaml
  abac:
    enabled: true
```
which is more explicit

PR #1332, opened 2020-01-29
Config now uses common media type module instead of FileTypeDetector.

### LINKED ISSUE #1213: Config: replace use of FileTypeDetector
Use `media-type` module instead.

tomas-langer:
Resolves #1213


config/config/src/main/java/io/helidon/config/UrlConfigSource.java
```diff
@@ -162,25 +160,8 @@ private String mediaType(String responseMediaType) throws URISyntaxException {
...
-                break;
-            case "file":
-                path = Paths.get(uri);
-                break;
-            default:
-                path = Paths.get(url.getPath());
-        }
-        return ConfigHelper.detectContentType(path);
+    private String probeContentType() {
+        return MediaTypes.detectType(url).orElse(null);
```
tjquinno: Very minor point. The `probeContentType` methods in the other classes return `Optional<String>`. This one returns `String` and uses `orElse`. After a quick look I did not see any null checks on the returned `String` from this method. Just for consistency would it make sense (without excessive disruption of the other sites which invoke `UrlConfigSource#probeContentType`) to  return the `Optional` just as `probeContentType` in the other classes do?

tomas-langer: I wanted to simply replace the existing usage. In other classes, `Optional` was removed. This class just returned the value (which could have been `null` with previous implementation).
I will have a look if this is not a too complicated change to do

tomas-langer: This seems to be a deeper problem in the implementation, as `ConfigParser.Content.mediaType()` is allowed to return null. I will create a follow up issue to clean this up

tomas-langer: Follow up: https://github.com/oracle/helidon/issues/1333

PR #1343, opened 2020-02-03
BaseProcessor back-pressure fix backport from #1282

danielkec:
There is a problem with BaseProcessor request count:
For example if map operator is requested like this:
```java
request(2);
request(2);
request(2);
```
his publisher will receive:
```java
request(2);
request(4);
request(6);
```
That can lead to flooding subscriber if proper back-pressure is applied


danielkec:
@romain-grecourt Can I ask for review pls, it seems to be severe bug enough to be backported

PR #1359, opened 2020-02-06
2.x - PermitAll overridden by JWT

### LINKED ISSUE #1272: PermitAll on class level is ignored by security
When a class is annotated with `@PermitAll` and the application requires authentication (e.g. when using `@LoginConfig`, the class level is ignored and the resource methods require authentication.

This is caused by the processing of AnnotationAnalyzer responses.
The `JwtAuthAnnotationAnalyzer` marks method as required authentication (which should not be done, as only the application class has authentication required). This overrides the `RolesAnnotationAnalyzer` that marks the resource class as optional authentication (due to `@PermitAll`)

- [x] Fixed in 1.x branch
- [x] Fixed in `master`

Verdent:

### DIFF
--- a/microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthAnnotationAnalyzer.java
+++ b/microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthAnnotationAnalyzer.java
@@ -102,30 +102,15 @@ public class JwtAuthAnnotationAnalyzer implements AnnotationAnalyzer {
     // resource class analysis
     @Override
     public AnalyzerResponse analyze(Class<?> maybeAnnotated, AnalyzerResponse previousResponse) {
-        return performAnalysis(maybeAnnotated, previousResponse);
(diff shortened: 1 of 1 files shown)


microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthAnnotationAnalyzer.java
```diff
@@ -102,30 +102,18 @@ public AnalyzerResponse analyze(Class<?> maybeAnnotated) {
     // resource class analysis
     @Override
     public AnalyzerResponse analyze(Class<?> maybeAnnotated, AnalyzerResponse previousResponse) {
-        return performAnalysis(maybeAnnotated, previousResponse);
+        return performAnalysis(previousResponse);
```
tomas-langer: Please remove the method `performAnalysis` as it is now useless and confusing.

Verdent: Done

PR #1360, opened 2020-02-06
1.x - PermitAll overridden by JWT

### LINKED ISSUE #1272: PermitAll on class level is ignored by security

Verdent:


romain-grecourt:
Pull 1.x in your branch again, @tomas-langer has merged a change that bumps the copyright plugin to 2.3 it should fix the copyright error you are seeing.

Issue #878, opened 2019-08-07
Add exception writer to Helidon WebServer


tomas-langer:
It would be nice to have a feature that would allow the following:

```
try {
  doSomething(response);
} catch (Exception e) {
  response.send(e);
}
```
or
```
future.thenApply(response::send)
  .exceptionally(response::send);
```

The signature would be:
```
public Void send(Throwable e);
```

This method should have two modes:

1. Default mode - send a 500, log exception to a log and (maybe) send an entity with a (configurable) message
2. Development mode - send a 500, send an entity with formatted exception stacktrace

The switch to development mode would be done via configuration (and of course possible through a builder).

romain-grecourt:
This would replace `ServerRequest.next(Throwable)`, right ?

I like the idea of using writers, this would allow for media supports to handle their own exceptions.

The signature should probably not be coupled to `Void`, so that it can be used with other types, e.g.  `CompletionStage<U>`

```java
public <T> T send(Throwable e)
```

ljnelson:
Any ideas for configuration item name and location?  I'm all ears.

PR #1378, opened 2020-02-12
Adds a default send(Throwable) method to ServerResponse.java as the first step in providing an easy facility for reporting exceptions during HTTP processing

ljnelson:
This PR, when ready, will add the ability to conveniently send a `Throwable` during HTTP processing.

### DIFF
--- a/media/common/src/main/java/io/helidon/media/common/ContentWriters.java
+++ b/media/common/src/main/java/io/helidon/media/common/ContentWriters.java
@@ -188,17 +188,4 @@ public final class ContentWriters {
         return byteChannelWriter(null);
     }
 
-    /**
-     * Returns a writer function for a {@link Throwable}'s stack trace.
-     *
-     * @param context a {@link MessageBodyWriterContext}
-     * @param setContentLength whether {@link MessageBodyWriterContext#contentLength(long)} should be called
-     * @return a {@link Throwable} writer
-     * @see #writeStackTrace(Throwable, MessageBodyWriterContext, boolean)
-     */
-    public static Function<Throwable, Publisher<DataChunk>> stackTraceWriter(MessageBodyWriterContext context,
-                                                                             boolean setContentLength) {
-        return throwable -> writeStackTrace(throwable, context, setContentLength);
-    }
-
 }
--- a/media/common/src/main/java/io/helidon/media/common/ContentWriters.java
+++ b/media/common/src/main/java/io/helidon/media/common/ContentWriters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 
 package io.helidon.media.common;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.nio.ByteBuffer;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.Charset;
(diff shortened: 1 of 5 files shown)


ljnelson:
See issue #878.

ljnelson:
* [x] Add `default` `void send(Throwable)` method to `ServerResponse`
* [x] Add `ThrowableBodyWriter` class that handles cases where stack traces should be output or not
* [ ] Figure out how to choose whether stack traces should be output or not, following the directive that such a choice should come from some kind of configuration
* [ ] Add `ThrowableBodyWriter`, properly configured, to the tail of `MediaSupport`'s list of default `MessageBodyWriter` implementations

media/common/src/main/java/io/helidon/media/common/ContentWriters.java
```diff
@@ -147,4 +187,18 @@ private ContentWriters() {
...
+
+    /**
+     * Returns a writer function for a {@link Throwable}'s stack trace.
+     *
+     * @param context a {@link MessageBodyWriterContext}
+     * @param setContentLength whether {@link MessageBodyWriterContext#contentLength(long)} should be called
+     * @return a {@link Throwable} writer
+     * @see #writeStackTrace(Throwable, MessageBodyWriterContext, boolean)
+     */
+    public static Function<Throwable, Publisher<DataChunk>> stackTraceWriter(MessageBodyWriterContext context,
```
romain-grecourt: `Function<T, Publisher<DataChunk>` is the old style of writer that is deprecated, we don't need that.

media/common/src/main/java/io/helidon/media/common/ContentWriters.java
```diff
@@ -81,6 +83,44 @@ private ContentWriters() {
...
+     * Create a a publisher {@link DataChunk} with the given
+     * {@link Throwable} / {@link Charset} and return a {@link Single}.
+     *
+     * @param throwable the {@link Throwable}
+     * @param context a {@link MessageBodyWriterContext}
+     * @param setContentLength whether {@link MessageBodyWriterContext#contentLength(long)} should be called
+     * @return Single
+     */
+    public static Single<DataChunk> writeStackTrace(Throwable throwable,
+                                                    MessageBodyWriterContext context,
```
romain-grecourt: With the addition the context as parameter, this methods differs from all the other ones in `ContentWriters`.
Lets not over do it and stick with just `ThrowableBodyWriter`.

webserver/webserver/src/main/java/io/helidon/webserver/ServerResponse.java
```diff
@@ -100,6 +100,27 @@ default ServerResponse status(int statusCode) throws AlreadyCompletedException {
...
+    /**
+     * Send a {@link Throwable} and close the response.
+     *
+     * @param content the {@link Throwable} to send
+     * @return {@code null} when invoked
+     * @throws IllegalArgumentException if there is no registered writer for a given type
+     * @throws IllegalStateException if any {@code send(...)} method was already called
+     * @see #send(Object)
+     */
+    default Void send(Throwable content) {
```
romain-grecourt: Lets not use a default method here. 

The API and implementation come together, so there isn't much benefit to having a default method implementation right ?

ljnelson: OK; will change to an interface here and then will move its contents to {checks notes} `Response.java`.

media/common/src/main/java/io/helidon/media/common/ContentWriters.java
ljnelson: OK; note that the content length issue is one that will be faced by the `CharSequence`-related methods as well, so if it's OK to remove content length concerns then I can accept what they do.  WDYT?

media/common/src/main/java/io/helidon/media/common/ContentWriters.java
ljnelson: OK; will remove these.  (The other ones: should they be `@Deprecated`?)

ljnelson:
Yes, the configuration will be a separate follow-on issue, as Romain and I decided that getting configuration into `MediaSupport` etc. was worth its own issue.  You'll notice in this PR as it stands that stack trace writing is disabled.

ljnelson:
Thanks; I'll merge this soon and then start work on #1392.

tomas-langer:
Thanks for explanation, approved.

Issue #1223, opened 2019-12-12
Fix javadoc for java11


tomas-langer:
Javadoc fails with errors such as:
```
org.apache.maven.reporting.MavenReportException:
Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in http://www.reactive-streams.org/reactive-streams-1.0.2-javadoc/ are in the unnamed module.
```

The documentation is actually generated.

For the time being, javadoc will not fail the build. This must be fixed before release.

barchetta:
Is this related to:

https://bugs.openjdk.java.net/browse/JDK-8212233

if so, a fix might be getting backported to 11. Although the issues implies it was fixed in 11.0.3, but maybe I'm not reading it right.

barchetta:
This has been addressed by #1386. To fix this required:

1. Adding/updating module-info files
2. Using the maven-javadoc-plugin aggregate goal (instead of the role-our-own approach we had before)
3. Upgrading to version 3.2.0 of maven-javadoc-plugin
4. Commenting out cross reference links
5. Generate aggregated javadocs as part of the `site` lifecycle
6. Comment out cross reference links (which we still need to fix).

To build the aggregated javadocs you now run:

`mvn site -Dversion.plugin.javadoc=3.2.0 -DskipTests -Pjavadoc`

It will generate the aggregated javadocs and the documentation. This currently requires you to build version 3.2.0 of the `maven-javadoc-plugin` yourself. We expect it to be released any day now.

PR #1386, opened 2020-02-13
Java 11 javadoc fixes. Turn on failOnError

barchetta:
This PR does gets javadocs building on Java 11 (see issue #1223). To do so I broke cross reference links to third party javadocs. We will need to fix that later. But this gets javadocs building again. Here is a summary of the changes:

1. Comment out the configuration for links (`offlineLinks`). Because we are generating Java 11 javadocs there are various issues with `class-path` and `module-path` hygiene (plus the infamous Java 11 javadoc bug). So for now, we avoid that rat's nest. This means those cross-reference links are broken, but at least we have javadocs.

2. Fix  and add `module-info` files as needed so the `maven-javadoc-plugin` passes a correct module-path to javadoc. This usually meant adding `require` statements.

3. Switch to use the javadoc aggregate goal as part of the site lifecycle to build aggregated javadocs

4. Fix basic javadoc errors

Still working on an issue, where the aggregate goal does not work well unless the maven artifacts have already been built (it gets the module-path wrong). So for now we build the aggregated javadocs after the main build.


config/etcd/src/main/java/module-info.java
```diff
@@ -28,7 +28,8 @@
     requires com.google.common;
     requires io.helidon.common;
     requires io.helidon.common.media.type;
-    requires grpc.api;
+    //requires grpc.api;
```
romain-grecourt: left-over comment.

config/pom.xml
```diff
@@ -61,7 +61,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-javadoc-plugin</artifactId>
                 <configuration>
-                    <excludePackageNames>*.internal,io.helidon.config.examples,io.helidon.config.tests</excludePackageNames>
+                    <excludePackageNames>*.internal.*,io.helidon.config.examples,io.helidon.config.tests</excludePackageNames>
```
romain-grecourt: there are package named '.internal' under config:
- `io.helidon.config.etcd.internal`
- `io.helidon.config.yaml.internal`

examples/quickstarts/helidon-standalone-quickstart-mp/pom.xml
```diff
@@ -179,6 +180,14 @@
                         <mainClass>${mainClass}</mainClass>
                     </configuration>
                 </plugin>
+                <plugin>
```
romain-grecourt: How about just defining the property `<maven.site.skip>true</maven.site.skip>`

examples/quickstarts/helidon-standalone-quickstart-se/pom.xml
```diff
@@ -179,6 +180,14 @@
                     <artifactId>helidon-maven-plugin</artifactId>
                     <version>${version.plugin.helidon}</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
```
romain-grecourt: How about just defining the property `<maven.site.skip>true</maven.site.skip>`

integrations/cdi/jpa-cdi/pom.xml
```diff
@@ -274,7 +270,7 @@
                 <dependency>
                     <groupId>javax.activation</groupId>
                     <artifactId>javax.activation-api</artifactId>
-                    <scope>runtime</scope>
+                    <scope>compile</scope>
```
romain-grecourt: Is the scope change needed ?

jersey/jsonp/pom.xml
```diff
@@ -28,6 +28,10 @@
     <artifactId>helidon-jersey-media-jsonp</artifactId>
     <name>Helidon Jersey Media JSON-P</name>
 
+    <properties>
+        <maven.javadoc.skip>true</maven.javadoc.skip>
```
romain-grecourt: This should probably not be skipped.

jersey/server/pom.xml
```diff
@@ -28,6 +28,10 @@
     <artifactId>helidon-jersey-server</artifactId>
     <name>Helidon Jersey Server</name>
 
+    <properties>
+        <maven.javadoc.skip>true</maven.javadoc.skip>
```
romain-grecourt: This should probably not be skipped.

parent/pom.xml
```diff
@@ -148,6 +149,31 @@
...
+                        <execution>
+                            <id>site-jar</id>
+                            <goals>
+                                <goal>jar</goal>
+                            </goals>
+                            <phase>site</phase>
+                            <configuration>
+                                <archiveExcludes>
+                                    <exclude>.asciidoctor/**</exclude>
+                                    <exclude>css/maven-*.css</exclude>
```
romain-grecourt: The css excludes are not needed anymore, I've added code in the site renderer to exclude these.

config/pom.xml
barchetta: Fixed

examples/quickstarts/helidon-standalone-quickstart-mp/pom.xml
barchetta: Fixed

examples/quickstarts/helidon-standalone-quickstart-se/pom.xml
barchetta: Fixed

parent/pom.xml
barchetta: Fixed

integrations/cdi/jpa-cdi/pom.xml
barchetta: java.xml.bind requires java.activation, so java.activation needs to be on the module-path. To get the compiler plugin to do that I needed to change to compile scope.  I have updated the jpa-cdi module-info and removed the requires java.activation since this module (jpa-cdi) does not require java.activation.

jersey/jsonp/pom.xml
barchetta: These modules have  no source to document. They already create empty javadoc jars. They are just bundle modules. I have verified javadoc won't run with just module-info. So this skip excludes them from the aggregated javadoc.

microprofile/reactive-streams/src/main/java/module-info.java
```diff
@@ -27,5 +27,6 @@
     requires transitive microprofile.reactive.streams.operators.api;
 
     exports io.helidon.microprofile.reactive;
+    exports io.helidon.microprofile.reactive.hybrid;
```
danielkec: Sorry for that :+1:

Issue #1390, opened 2020-02-14
Cannot use OpenAPI U/Is to change the greeting message in the MP quickstart


tjquinno:
## Environment Details
* Helidon Version: 1.x, 2.x
* Helidon SE or Helidon MP: MP
* JDK version: n/a
* OS: n/a
* Docker version (if applicable): n/a

----------

## Problem Description
In working with the openapi-ui MP extension (and this is not unique to that extension), the user cannot execute the endpoint to update the greeting correctly because there is no way to provide the new greeting input parameter.

## Steps to reproduce
For example, to the 2.x quickstart MP example:
* Comment out the `GreetApplication#getClasses` method. (See also #1380)
* Add this dependency:
```
<dependency>
    <groupId>org.microprofile-ext.openapi-ext</groupId>
    <artifactId>openapi-ui</artifactId>
    <version>1.1.2</version>
    <scope>runtime</scope>
</dependency>
```
* `mvn package`
* `java -jar target/helidon-quickstart-mp.jar`
* From a browser access `http://localhost:8080/openapi-ui`
* Click on the PUT button for `/greet/greeting`. There is no way to enter the request body JSON for the new greeting message. Clicking `Execute` yields a server error because the expected request body is missing.

tjquinno:
Adding this code (with the corresponding `import`s) to the `GreetResource#updateGreeting` method resolves the problem:
```
@RequestBody(name = "greeting",
            required = true,
            content = @Content(mediaType = "application/json",
                    schema = @Schema(type = SchemaType.STRING)))
```

tjquinno:
Note that we could add a large number of other OpenAPI annotations to all endpoints in the MP quickstart example so the OpenAPI info is more extensive (descriptions, examples, etc.). But I like the idea of _not_ doing so, keeping the clutter to a minimum. That said, the `@RequestBody` change above is a necessary bug fix.

PR #1393, opened 2020-02-14
1.x Add OpenAPI annotations to MP quickstart for input and responses on updateGreeting

tjquinno:
Addresses #1390 for 1.x

Adding these annotations causes OpenAPI to correctly understand the input and the possible responses for `updateGreeting`.

Note that the defaults are OK for the other endpoints.


examples/quickstarts/helidon-quickstart-mp/src/main/java/io/helidon/examples/quickstart/mp/GreetResource.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
```
tomas-langer: `All rights reserved. ` should be removed from copyright when updating it.

tjquinno: Fixed

Issue #1380, opened 2020-02-13
Update helidon-quickstart-mp to show proper JAX-RS application discovery


barchetta:
## Environment Details
* Helidon Version: 2.0.0-M1
* Helidon MP
* JDK version: 11
----------

## Problem Description
helidon-quickstart-mp still uses `javax.ws.rs.core.Application.getClasses()` for discovery. In light of the new JAX-RS scanning improvements, should we change this (remove it)?

spericas:
I think so, it's easier to integrate with other libraries if scanning is enabled and not overridden.

tomas-langer:
Why do we have the application class at all? I am not confident this change is good.

spericas:
> Why do we have the application class at all? I am not confident this change is good.

Why aren't you confident?

PR #1395, opened 2020-02-14
Remove MP quickstart GreetApplication#getClasses method for auto-discovery

tjquinno:
Addresses #1380 

Removing the `getClasses` method from an app class allows auto-discovery to find resources automatically. This change showcases this feature in the quickstart (and stand-alone quickstart) MP apps.


tomas-langer:
I don't think this change is good. 
1. Application class is present to show how to list resources that belong to the app
2. When you remove `getClasses`, the Application class does not serve any purpose in this example, so you could just remove it

I liked the way the quickstart was explicit - you create an app that has resources (and you can add an additional app, that has different resources).
Now you have a set of classes that "automagically" end up in the application you define by that class? I know this is the way people from app servers expect this to work, but it is not nice (IMHO).

Please let's talk about it before merging.

spericas:
@tomas-langer Adding multiple apps is not the norm. Developers with experience in JAX-RS would not normally do that. Now adding libraries with resources and expecting those to be part of your app is more common, in fact this is how this whole discussion started, from a developer question. Other MP implementations seem to favor scanning as well.

tjquinno:
One concrete use case that came up in the slack channel is the MP extension openapi-ui. For this to work if `getClasses` is present, the developer has to modify that method _after_ inspecting the source code for openapi-ui to find out what the service classes are.

I do not think we do want to force developers to do dive into the internals of external JARs to make them work with our example code.

tjquinno:
In light of this discussion, I propose to remove the `GreetApplication` class from the MP quickstart and MP stand-alone quickstart.

@tomas-langer Are you OK with this approach?

tjquinno:
Tomas reports he is OK with removing the application class.

PR #1403, opened 2020-02-18
Adds Config into MediaSupport#builder() method

ljnelson:

### DIFF
--- a/media/common/src/main/java/io/helidon/media/common/MediaSupport.java
+++ b/media/common/src/main/java/io/helidon/media/common/MediaSupport.java
@@ -15,6 +15,8 @@
  */
 package io.helidon.media.common;
 
+import io.helidon.config.Config;
+
 /**
  * Media support.
  */
@@ -62,6 +64,17 @@ public final class MediaSupport {
     }
 
     /**
+     * Create a new instance with the default readers and writers registered on
+     * the contexts.
+     * @param config a {@link Config} that will be passed to {@link
+     * Builder#registerDefaults(Config)}
+     * @return MediaSupport
+     */
+    public static MediaSupport createWithDefaults(Config config) {
+        return builder().registerDefaults(config).build();
+    }
+
+    /**
      * Create a new {@link Builder} instance.
      * @return Builder
      */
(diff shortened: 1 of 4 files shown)


media/common/src/main/java/io/helidon/media/common/MediaSupport.java
```diff
@@ -61,27 +90,62 @@ public static MediaSupport createWithDefaults() {
...
     public static Builder builder() {
         return new Builder();
     }
 
+    /**
+     * Create a new {@link Builder} instance.
+     * @param config a {@link Config}
+     * @return a new {@link Builder}
+     */
+    public static Builder builder(Config config) {
```
tomas-langer: In other places where we use config, we usually create the following:
- static factory method `builder()` - already in this class
- method `Builder.config(Config)` - already in builder
- static factory method `create(Config)` 

this method is not that useful 
- if you want to create `MediaSupport` from config only (if that is even feasible), add the `create(Config)` method
- if you want to update the builder from config and then do more with the builder, just use `builder().config(Config)`

This shortcut does not seem useful.

media/common/src/main/java/io/helidon/media/common/MediaSupport.java
```diff
@@ -61,27 +90,62 @@ public static MediaSupport createWithDefaults() {
...
             writerContext = MessageBodyWriterContext.create();
         }
 
+        /**
+         * Configures this {@link Builder} from the supplied {@link Config}.
+         * @param config a {@link Config}
+         * @return this {@link Builder}
+         */
+        public Builder config(Config config) {
+            if (config == null) {
```
tomas-langer: config can never be null
Our APIs never accept `null` as a valid parameter - this should throw a `NPE`

media/common/src/main/java/io/helidon/media/common/MediaSupport.java
```diff
@@ -61,27 +90,62 @@ public static MediaSupport createWithDefaults() {
...
+                serverErrorsIncludeStackTraces = false;
+            } else {
+                serverErrorsIncludeStackTraces = config.get("serverErrorsIncludeStackTraces").as(Boolean.TYPE).orElse(false);
```
tomas-langer: Please use
`config("server-errors-include-stack-traces").asBoolean().ifPresent(this::serverErrorsIncludeStackTraces)`

The problem with your approach:
- if the `serverErrorsIncludeStackTraces` is configured by "hand" and then you update it with config, you override the value even though config does not include that configuration key
- the configuration keys should be dash separated, not camel case
- there are shortcuts for primitives and string on config (e.g. use `asBoolean()` rather than `as(Boolean.class)` or `as(Boolean.TYPE)`)
- I do not see a setter for the `serverErrorsIncludeStackTraces` field in builder, add `public Builder serverErrorsIncludeStackTraces(boolean)` method - everything configurable through config should be configurable through a builder.

romain-grecourt:
@ljnelson Can you add a config key for registering the defaults ?

ljnelson:
Done

PR #1410, opened 2020-02-19
Improve SingleJust + TCK

akarnokd:
Improve the implementation of `io.helidon.common.reactive.SingleJust` to conform the Reactive Streams specification and add a TCK test for it as well.


common/reactive/src/test/java/io/helidon/common/reactive/SingleTest.java
```diff
@@ -88,7 +88,7 @@ public void onSubscribe(Subscription subscription) {
         };
         Single.<String>just("foo").subscribe(subscriber);
         assertThat(subscriber.isComplete(), is(equalTo(false)));
-        assertThat(subscriber.getLastError(), is(nullValue()));
+        assertThat(subscriber.getLastError(), instanceOf(IllegalArgumentException.class));
```
akarnokd: Although not verified by the TCK, ignoring non-positive request amounts after cancellation can hide subtle bugs in operators.

akarnokd:
On a related note, `mvn validate -Pcheckstyle` keeps failing locally because the class [`Ascii.java`](https://github.com/oracle/helidon/blob/master/common/http/src/main/java/io/helidon/common/http/Ascii.java#L31) contains `@author` tags. I'm on Windows with IntelliJ 2019.3.x btw.

common/reactive/src/test/java/io/helidon/common/reactive/SingleTest.java
danielkec: Nice thx!

PR #1417, opened 2020-02-20
Config change support refactoring

### LINKED ISSUE #578: Config: Support for "lazy" config sources
Some config sources do not support loading the whole configuration tree (due to size or performance reasons).
Such as:
 - a remote source that does not have the capability to list all nodes
 - a database source with millions of records

For these sources, we should be at least able to obtain a specific node (identified by its key).

This is currently (I think) not possible

This task is to check our implementation if such a source could be implemented right now and provide an example.
If this is not possible with current SPI/API, we should modify these to support it (and in such a case this is a 2.0 feature).

### LINKED ISSUE #579: Config: refactor change support
Refactor current change support - remove deprecated elements, make sure we handle all desired use cases, remove dependency on `Flow` API and, if possible, on project reactor.

### LINKED ISSUE #1183: Config 2.0 follow up - change support commented out
In `AbstractConfigImpl`, the lines 386-391 are commented out, as the change support as implemented right now fails.
As we plan to refactor this (#579), we can delay fixing this issue after the refactoring is done.

tomas-langer:
Fixes #579 
Fixes #1183 
Fixes #578


tomas-langer:
Follow up: https://github.com/oracle/helidon/issues/1420

config/config/src/main/java/io/helidon/config/ConfigSourcesRuntime.java
```diff
@@ -0,0 +1,223 @@
...
+            throw new IllegalArgumentException("Unsupported node type: " + changeNode.nodeType());
+        }
+    }
+
+    synchronized Optional<ObjectNode> latest() {
+        List<ObjectNode> objectNodes = loadedData.stream()
+                .map(RuntimeWithData::data)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .collect(Collectors.toList());
```
danielkec:
```java
List<ObjectNode> objectNodes = loadedData.stream()
                .map(RuntimeWithData::data)
                .flatMap(Optional::stream)
                .collect(Collectors.toList());
```

config/config/src/main/java/io/helidon/config/ConfigSourcesRuntime.java
```diff
@@ -0,0 +1,223 @@
...
+                        .map(objectNode -> objectNode.initDescription(source.description()))));
+            }
+        }
+
+        Set<String> allKeys = loadedData.stream()
+                .map(RuntimeWithData::data)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .flatMap(this::streamKeys)
+                .collect(Collectors.toSet());
```
danielkec: Optional is such ... :disappointed: 
```java
.flatMap(Optional::stream)
.flatMap(this::streamKeys)
```

config/config/src/main/java/io/helidon/config/ConfigSourcesRuntime.java
```diff
@@ -0,0 +1,223 @@
...
+        for (RuntimeWithData data : loadedData) {
+            if (data.runtime().isLazy()) {
+                data.data(loadLazy(data.runtime(), allKeys));
+            }
+        }
+
+        List<ObjectNode> objectNodes = loadedData.stream()
+                .map(RuntimeWithData::data)
+                .filter(Optional::isPresent)
+                .map(Optional::get)
```
danielkec: same

config/config/src/main/java/io/helidon/config/ListNodeBuilderImpl.java
```diff
@@ -97,6 +97,13 @@ public ListNodeBuilderImpl value(String value) {
         return this;
     }
 
+    // this is a shortcut method to keep current fluent code
+    // even though value is now optional
+    ListNodeBuilderImpl value(Optional<String> value) {
+        value.ifPresent(this::value);
+        return this;
+    }
```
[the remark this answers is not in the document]
tomas-langer: This method is called as fluent from multiple places that have different methods called.

config/config/src/main/java/io/helidon/config/ConfigSourcesRuntime.java
tomas-langer: Fixing.

config/config/src/main/java/io/helidon/config/ConfigSourcesRuntime.java
tomas-langer: Fixing

tomas-langer: Also fixing in all other classes this was used.

PR #1425, opened 2020-02-21
Upgrade netty to 4.1.45

tomas-langer:
Upgrade Netty to same version we have in `master`

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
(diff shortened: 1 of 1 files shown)


aseovic:
You should also make sure that we don't pull in multiple versions of Netty via gRPC dependency. Basically, we need to upgrade to `grpc-java` version that depends on Netty 4.1.45 at the same time, which means bumping `grpc-java` to 1.27.1 (the latest) in both 1.x and 2.0 branches.

tomas-langer:
Just read @aseovic comment...

tomas-langer:
Will create another PR.

PR #1427, opened 2020-02-21
Prometheus format problems

spericas:
Do not repeat LINE and HELP lines in Prometheus format as this is a violation that can be flagged by tools with validation enabled. This can happen for metrics with same name and different tags. We now keep track of TYPE and HELP lines added to format to avoid repetition.

### DIFF
--- a/metrics2/metrics2/src/main/java/io/helidon/metrics/MetricsSupport.java
+++ b/metrics2/metrics2/src/main/java/io/helidon/metrics/MetricsSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/metrics2/metrics2/src/main/java/io/helidon/metrics/MetricsSupport.java
+++ b/metrics2/metrics2/src/main/java/io/helidon/metrics/MetricsSupport.java
@@ -22,10 +22,12 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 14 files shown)


metrics2/metrics2/src/main/java/io/helidon/metrics/MetricsSupport.java
```diff
@@ -182,29 +184,40 @@ private void optionsAll(ServerRequest req, ServerResponse res, Registry registry
...
-                .collect(StringBuilder::new,
-                        (sb, entry) -> toPrometheusData(sb, entry.getKey(), entry.getValue()),
-                        StringBuilder::append)
-                .toString();
+        StringBuilder builder = new StringBuilder();
+        Set<String> serialized = new HashSet<>();
+        List<Map.Entry<MetricID, HelidonMetric>> metrics = registry.stream()
+                .sorted(Map.Entry.comparingByKey())
+                .collect(Collectors.toList());
+        for (Map.Entry<MetricID, HelidonMetric> entry : metrics) {
```
tjquinno: The `List` `metrics` is used only in the immediately next `for` loop. Maybe instead of collecting into the list use `.forEach` after the `.sorted`?

metrics2/metrics2/src/test/java/io/helidon/metrics/MetricsSupportTest.java
```diff
@@ -144,4 +151,21 @@ void testBaseMetricsDisabled() {
...
+        Set<String> found = new HashSet<>();
+        String data = MetricsSupport.toPrometheusData(app, base);
+        try (BufferedReader reader = new BufferedReader(new StringReader(data))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                String[] tokens = line.split(" ");
+                if (tokens.length > 3 && tokens[1].equals("TYPE")) {
+                    String metric = tokens[2];
+                    assertFalse(found.contains(metric));
+                    found.add(metric);
```
tjquinno: How about combining these into `assertTrue(found.add(metric))`?

Because performance is _so_ important inside the test!

metrics2/metrics2/src/main/java/io/helidon/metrics/MetricsSupport.java
```diff
@@ -182,29 +184,40 @@ private void optionsAll(ServerRequest req, ServerResponse res, Registry registry
...
+            String name = entry.getKey().getName();
+            if (!serialized.contains(name)) {
+                toPrometheusData(builder, entry.getKey(), entry.getValue(), true);
+                serialized.add(name);
+            } else {
+                toPrometheusData(builder, entry.getKey(), entry.getValue(), false);
```
tjquinno: The stream is sorted by key (`MetricID`), and therefore primarily by name according to the  `MetricID#compareTo` implementation. So, instead of storing each processed metric name in  a `Set` and checking it each time, we could save the most-recently-processed name in a String (initialized to null) and if the name being processed is different then (and only then) pass `withHelpType` as `true`. And assigning the current name to the variable holding the previously-processed one (even it's the same value) is pretty cheap, so the code in the `forEach` could be something like:
```
// inside the forEach
String name = entry.getKey().getName();
toPrometheusData(builder, entry.getKey(), entry.getValue(), !name.equals(previousName);
previousName = name;
```
which is a bit simpler and quite a bit faster. Granted, this does depend on the ordering provided by `MetricID#compareTo` always giving preference to the name before the tag (or other stuff that might be added to `MetricID` in the future) so that, when sorted, `MetricID`s are grouped by name. But that seems fairly reliable.

metrics2/metrics2/src/test/java/io/helidon/metrics/MetricsSupportTest.java
spericas: Makes it less readable IMO

PR #1430, opened 2020-02-21
Reimplement Multi.filter + TCK test

akarnokd:
Reimplement `Multi.filter` and add TCK tests. The signature has been adjusted to allow contra-variant predicates too (predicates working on the supertype of `T`).

Requesting one replenishment may seem inefficient, but it is actually better than batch requesting. The reason is that in case of batch requesting (prefetching), the operator may run with a different speed than the downstream thus trampolining is required to sync up the two. Consequently, trampolining requires two atomic operations per item at best (trampoline enter/exit). Requesting one is at best requires one atomic operation (see the combined trampoline-request accounting gate in the improved [Multi.from(Iterable)](https://github.com/oracle/helidon/pull/1429/files#diff-6e58085854ee1341e169599585b70a21R87).


akarnokd:
Superseeded by #1442

Issue #1205, opened 2019-12-12
Web Client


m0mus:
*Web Client* is an HTTP client for Helidon SE based on Netty.

### SSL

- [x] Enabling SSL when connecting to https
- [x] Trust all certificates
- [x] Adding truststore
- [x] Client authentication
- [x] Client keystore
- [x]  Hostname verification
- [x]  Netty SslContext from java.net.SSLContext

### Proxy

- [x] Loading proxy from config
- [x] Adding proxy via builder
- [x] No proxy hosts

### Requests

- [x] GET request
- [x] Other request types
- [x] Headers added to request
- [x] Default headers from configuration
- [x] Query parameters
- [x] Redirection based on 30x responses
- [x] Max number of redirection
- [x] Custom writer registration
- [x] Metrics propagation
- [x] Security propagation
- [x] Tracing propagation - test creation in progress

### Response

- [x] Backpressure support
- [x]  Custom reader registration
- [x] JSON-P response type resolver

### Headers
- [x]  Add special methods for commonly used headers
- [x]  Headers with special format

### Cookies

- [x] Cookies propagation
- [x] Default cookies

Verdent:
Implemented.

PR #1431, opened 2020-02-21
Webclient implementation (#1205)

Verdent:
More tests and updated examples will be added after the merge probably.


webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+
+    ClientServiceRequest clientServiceRequest() {
+        return clientServiceRequest;
+    }
+
+    List<ClientService> services() {
+        return services;
+    }
+
+    Set<MessageBodyReader<?>> requestReaders() {
```
romain-grecourt: Thinking-out-loud.

I don't know if having a configuration for request/response specific readers/writers is really useful.
Instead you could expose the context from request/response and let the user add these manually for each request/response.

If we think there is a need to have HTTP client specific readers/writers for all client request/response, then maybe we should consider updating `MediaSupport`, `MessageBodyReaderContext` and `MessageBodyWriterContext` to support creating a parented `MediaSupport`.

This way you would only need to pass-in a `MediaSupport` instance to `RequestConfiguration`.

webclient/webclient/src/main/java/io/helidon/webclient/ClientException.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.Optional;
+
+/**
+ * An exception that caused this client request to fail. If the exception is based on status or content of response from
+ * server, the {@link ClientResponse} is available to obtain any needed details.
+ * If the exception was caused before data was sent, or due to timeouts, socket exceptions etc., the {@link ClientResponse} is
+ * not present and exception handling must be based on the wrapped exception.
+ */
+public class ClientException extends RuntimeException {
```
romain-grecourt: Should that be named `HttpClientException` or `WebClientException` instead ?

webclient/webclient/src/main/java/io/helidon/webclient/ClientConfiguration.java
```diff
@@ -0,0 +1,539 @@
...
+import io.netty.handler.ssl.JdkSslContext;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
+import io.netty.handler.ssl.SslProvider;
+import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
+
+/**
+ * Configuration of the Helidon web client.
+ */
+class ClientConfiguration {
```
romain-grecourt: Should that be named `HttpClientConfiguration` or `WebClientConfiguration` instead ?

webclient/tracing/src/main/java/io/helidon/webclient/tracing/ClientTracing.java
```diff
@@ -0,0 +1,80 @@
...
+import io.opentracing.SpanContext;
+import io.opentracing.Tracer;
+import io.opentracing.propagation.Format;
+import io.opentracing.propagation.TextMapAdapter;
+import io.opentracing.util.GlobalTracer;
+
+/**
+ * Client service for tracing propagation.
+ */
+public class ClientTracing implements ClientService {
```
romain-grecourt: Should that be named `HttpClientTracing` or `WebClientTracing` instead ?

webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
Verdent: Ah this is great you are mentioning this. I have a note to create issue for that. I think that having builder on `MediaSupport` which allows you to insert explicit parent, would be great for my usecase. I didn't added yet, because I wanted to talk to you about it before.

In case of the exposing of context, I do not think that is what we should do. 

The reason why I opted for specific readers/writers for each request/response is, that it allows you to have control over entity handling. Basically you can have two or more ways how to handle specific entity type in your code and this helps you to achieve this without interfering with other request where it is not desirable.

webclient/webclient/src/main/java/io/helidon/webclient/ClientRequestHeadersImpl.java
```diff
@@ -0,0 +1,302 @@
...
+import io.helidon.common.http.MediaType;
+import io.helidon.common.http.Parameters;
+
+import io.netty.handler.codec.http.cookie.ClientCookieEncoder;
+import io.netty.handler.codec.http.cookie.DefaultCookie;
+
+/**
+ * Client request header implementation.
+ */
+class ClientRequestHeadersImpl implements ClientRequestHeaders {
```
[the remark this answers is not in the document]
Verdent: Because we needed full control over the map/object which stores those headers to be able to easily clear them if needed.

examples/webclient/standalone/src/main/java/io/helidon/rest/client/example/basic/StandaloneClientExample.java
```diff
@@ -0,0 +1,113 @@
...
+         */
+        WebClient client = WebClient.builder()
+                .config(config.get("client"))
+                .context(context)
+                .build();
+
+        /*
+         * Each request is created using a builder like fluent api
+         */
+        //        CompletionStage<ClientResponse> response = client.put()
```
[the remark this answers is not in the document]
Verdent: Yes, thank you.

examples/webclient/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.helidon.examples</groupId>
+        <artifactId>helidon-examples-project</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+    </parent>
+
+    <groupId>io.helidon.examples.webclient</groupId>
+    <artifactId>helidon-examples-webclient</artifactId>
```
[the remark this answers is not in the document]
Verdent: Thank you.

webclient/webclient/src/main/java/io/helidon/webclient/ClientException.java
Verdent: Yes in fact I do agree. I didn't think about it when I was writing it, because it was obvious for me where does this belong, but I are correct that this is probably better name.

webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
romain-grecourt: There should be a parented context dedicated to each request, that would be symmetrical to how it's done in the webserver currently. We can talk more about this when you have some time.

Issue #1432, opened 2020-02-22
Add support for CDI discovery and registration of JAX-RS providers


aseovic:
## Environment Details
* Helidon Version: 1.4.1 and 2.0
* Helidon MP
* JDK version: any
* OS: any
* Docker version (if applicable):

----------

## Problem Description

At the moment JAX-RS providers, such as request/response filters and message body readers/writers are not discovered and registered automatically by the CDI, unlike JAX-RS resources.

This is problematic, as it requires one of several approaches:

1. Creating `Application` class and explicitly specifying all resources and providers via `getClasses` method
2. Implementing `MpService`, which is going away in Helidon 2.0 and seems like an overkill anyway
3. Implementing `AutoDiscoverable`, which is internal and fairly low level.

Neither of these approaches seems reasonable for such a common use case. If the provider is implemented as a proper CDI managed bean, we should be able to discover it and register it automatically.
 
## Example

The following filter should be automatically discovered and registered with JAX-RS application, just like the resources currently are:

```java
/**
 * Simple CORS filter.
 */
@ApplicationScoped
@Provider
@ConstrainedTo(RuntimeType.SERVER)
public class CorsResponseFilter
        implements ContainerResponseFilter {
    @Override
    public void filter(ContainerRequestContext requestContext, ContainerResponseContext responseContext) {
        MultivaluedMap<String, Object> headers = responseContext.getHeaders();

        headers.add("Access-Control-Allow-Origin", "*");
        headers.add("Access-Control-Allow-Methods", "GET, POST, DELETE, PUT, PATCH, OPTIONS, HEAD");
        headers.add("Access-Control-Allow-Headers", "Content-Type, *");
    }
}
```

ljnelson:
…but it would seem to be critical IMHO to respect https://eclipse-ee4j.github.io/jersey.github.io/apidocs/latest/jersey/org/glassfish/jersey/server/ServerProperties.html#FEATURE_AUTO_DISCOVERY_DISABLE if it is present in the application's configuration in any way, shape or form, yes?

PR #1435, opened 2020-02-24
Support for @Provider discovery in MP server (#1432)

aseovic:
Adds support for manual registration of providers (via Builder), as well as automatic discovery and registration of CDI beans with `@Provider` annotation

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2020 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -320,7 +320,7 @@ public interface Server {
         private SeContainer createContainer(ClassLoader classLoader) {
             // not in CDI
             Weld initializer = new Weld();
-            initializer.addBeanDefiningAnnotations(Path.class, Provider.class);
+            initializer.addBeanDefiningAnnotations(Path.class);
             initializer.setClassLoader(classLoader);
             Map<String, Object> props = new HashMap<>(config.helidonConfig()
                                                               .get("cdi")
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018-2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,6 +36,7 @@ import javax.enterprise.inject.se.SeContainer;
 import javax.enterprise.inject.spi.CDI;
 import javax.ws.rs.Path;
 import javax.ws.rs.core.Application;
+import javax.ws.rs.ext.Provider;
 
 import io.helidon.common.CollectionsHelper;
 import io.helidon.common.configurable.ServerThreadPoolSupplier;
@@ -160,6 +161,7 @@ public interface Server {
         private static final Logger STARTUP_LOGGER = Logger.getLogger("io.helidon.microprofile.startup.builder");
 
         private final List<Class<?>> resourceClasses = new LinkedList<>();
+        private final List<Class<?>> providerClasses = new LinkedList<>();
         private final List<JaxRsApplication> applications = new LinkedList<>();
         private HelidonServiceLoader.Builder<MpService> extensionBuilder;
         private ResourceConfig resourceConfig;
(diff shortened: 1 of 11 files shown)


microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018-2020 Oracle and/or its affiliates.
```
tomas-langer: Copyright years should be comma separated in Helido.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
```diff
@@ -315,7 +320,7 @@ private void resourcesFromContainer() {
         private SeContainer createContainer(ClassLoader classLoader) {
             // not in CDI
             Weld initializer = new Weld();
-            initializer.addBeanDefiningAnnotations(Path.class);
+            initializer.addBeanDefiningAnnotations(Path.class, Provider.class);
```
tomas-langer: As you said in chat, we want to load only providers that are annotated using a bean defining annotation. This change modifies that behavior by making all of the having a bean defining annotation.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
aseovic: Hmm... I'm pretty sure I've seen dash used somewhere, so we are definitely not consistent.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
aseovic: I just wanted to make sure it works exactly the way resources work. It is my understanding that this code ensures that any resources (and now providers) that have been explicitly registered via Builder are also registered with the CDI container. 

I think that makes sense, and is separate from CDI discovery, which is done by the extension and expects providers to be proper managed beans.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
aseovic: Fixed.

microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
tomas-langer: The problem is that this is a backward incompatible change and may have unexpected behavior.
If you add `@ApplicationScoped` to a class annotated as `@Provider`, it will be automatically picked up by your changes.
Adding `Provider` as a bean defining annotation is a step too far, as any provider in the code base would get added automatically, and that may not be desired. I would keep the original agreed approach that only beans are added automatically

spericas: Perhaps this is something we can do in 2.0 where the scanning code is already incompatible

aseovic: But that's the thing -- it doesn't, just like the resources annotated with`@Path` do not get automatically picked **unless** they are also annotated with `@ApplicationScoped` or `@RequestScoped`.

The only thing the code above does is **allow** you to register classes that do not have any standard bean defining annotations with the bean manager, which happens a little bit later in the code:

```
            // add resource classes explicitly configured without CDI annotations
            this.resourceClasses.stream()
                    .filter(this::notACdiBean)
                    .forEach(initializer::addBeanClasses);

            // add provider classes explicitly configured without CDI annotations
            this.providerClasses.stream()
                    .filter(this::notACdiBean)
                    .forEach(initializer::addBeanClasses);
```
I've tried running my demo app using this implementation, and my filter is only discovered and registered automatically if I add `@ApplicationScoped` annotation as well. None of the other `@Provider` classes get picked up automatically, even though I depend on MP bundle and am bringing in a ton of dependencies.

The bottom line is that because Jersey treats resources and providers the same (they are all returned from the same `Application.getClasses` method), we should treat them the same as well and allow for one what we already allow for the other (in this case, registration of explicitly added resources and providers that are **not** proper managed beans with bean manager). I thought about not even separating them out and using a single list to collect all resource and provider classes, but decided against it as we have a few places where we log warning if there are no resources defined for the app, so having them separate is helpful.

tomas-langer: I just realized this is the 1.4 branch, where we handle resources differently from 2.0. On `master` some of these assumptions are no longer valid. Sorry about that.
We probably can keep the same function for 2.0 when it comes to `@Provider`, though it does work differently for `@Path`

aseovic:
I've added the tests for new functionality, as we discussed, and hit the bug with MBR/MBW generic type argument processing when Weld proxies are used. I'll open a bug against Jersey for that, so we can track it.

Apart from that, I've changed some of the server semantics a bit in order to be able to mix explicit applications with auto-discovered resources and providers (implicit application).

Basically, you can now register one or more `Application` classes explicitly, just like before, but if you have explicitly registered or auto-discovered resources and providers in addition to those, they will be packaged into an implicit/default application and added to the `applications` list. That way it's not "either/or" as it is at the moment -- you can have both at the same time.

The main reason I did it was so I can have both in the test project, but I do think it makes sense in general -- it allows you to implement your own app/service using auto-discoverable resources and providers, but also add explicit apps defined elsewhere.

I've also changed `ServerCdiExtension` to support discovery of `@RequestScoped` root resources (which is an issue I've hit before but didn't know why they are not being picked up automatically), and changed it discovers everything that's discoverable, not either applications or resources/providers.

Anyway, let's discuss tomorrow and decide how to proceed.

aseovic:
Opened https://github.com/eclipse-ee4j/jersey/issues/4400 to track the issue with CDI-managed MessageBodyReaders/MessageBodyWriters.

tomas-langer:
The combination of discovered resource classes and application is intentionally not possible. You can add an auto-discovered application that explicitly returns resource and provider classes and combine it with registered applications
You cannot have registered application classes and combine them with discovered resources, as that may create unexpected endpoints.

tomas-langer:
Do not merge this PR please, see #1448

PR #1452, opened 2020-02-27
Reimplement ConcatPublisher + TCK tests

akarnokd:
Reimplement `io.helidon.common.reactive.ConcatPublisher` so it works as a properly reusable `Flow.Publisher`.

Fresh start due to master changes.


akarnokd:
Spotbugs doesn't like if I extend `AtomicInteger` because it is marked as Serializable and will cascade it everywhere even though Flow is not meant to be serialized at all.

romain-grecourt:
You can add an exclusion, either with an xml file, or using `SuppressFBWarnings`

PR #1458, opened 2020-03-02
Remove dependency on Jackson via SmallRye -2.x

### LINKED ISSUE #1109: Jackson dependency brought in by OpenApi
The Small Rye OpenAPI brings in dependency on Jackson:
```
[INFO] |  +- com.fasterxml.jackson.core:jackson-core:jar:2.9.8:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.9.8:compile
[INFO] |  |  \- com.fasterxml.jackson.core:jackson-annotations:jar:2.9.0:compile
[INFO] |  +- com.fasterxml.jackson.dataformat:jackson-dataformat-yaml:jar:2.9.8:compile
```

This is not a desired dependency for Helidon and we should check if this can be replaced with Yasson (which is our preferred JSON-B implementation). If not, we should rebuild the OpenAPI with JSON-P or Yasson dependencies only.

tjquinno:
Not ready for approval yet -- need to update the `helidon-build-tools` version in the top-level Helidon pom once we have a new `helidon-build-tools` release.

# Overview - removing our dependency on Jackson

(fixes #1109)

Our OpenAPI implementation layers on SmallRye’s OpenAPI module, which on its own uses Jackson to parse the OpenAPI JSON or YAML and to parse the JSON in OpenAPI `@Extension` annotations. The SmallRye API has for some time allowed the calling code to — and Helidon now does — completely bypass the main parsing (doing so itself before invoking SmallRye), and a recent SmallRye enhancement allows the caller to provide its own way to do the `@Extension` parsing. 

Removing the Jackson dependency involves three main areas of change:
1. creating the `helidon-build-tools` SnakeYAML plug-in, (merged separately)
2. changing the Helidon OpenAPI pom to drive that plug-in, and
3. changing the Helidon OpenAPI code to
enhance the SnakeYAML-created object model for OpenAPI in important ways,
to parse YAML and JSON using SnakeYAML (rather than leaving that to SmallRye), and
driving the SmallRye API slightly differently to accomplish the above.

## The plug-in
SnakeYAML does a good job identifying properties from the MicroProfile OpenAPI interfaces, but SnakeYAML cannot (nor can any tool that looks only at compiled byte code) build a complete map between the JSON/YAML serialization and the object model. The SnakeYAML API allows callers to provide this additional information.

The plug-in analyzes the MP OpenAPI interfaces and the SmallRye implementation classes and generates a class that includes this helper logic for SnakeYAML.

## The Helidon OpenAPI module component `pom.xml`
The pom:
1. enhances the dependency on SmallRye to exclude Jackson,
2. invokes the `dependency` plug-in to extract the source code for the MicroProfile OpenAPI interfaces and the SmallRye implementations, and
3. invokes the new `helidon-build-tools` SnakeYAML plug-in to generate the helper class.

## The Helidon OpenAPI changes
The new plug-in is independent of OpenAPI, yet there are some aspects of the OpenAPI interfaces and the SmallRye implementations that need special attention to correctly map between JSON/YAML and the object model. These adjustments to the model occur in the `OpenAPISupport#adjustTypeDescriptions` method.

In MP OpenAPI 1.1.2, several interfaces extend `Map<X, Y>`. This “confuses” SnakeYAML slightly, in that when deserializing a document it instantiates an `Object` for the children instead of the actual implementation class for the interface `Y`. MP OpenAPI seems to be abandoning this approach in 2.0, but we still need to deal with it now. That happens in `CustomConstructor`.

The SnakeYAML API relies heavily on the `TypeDescription` class for conveying non-default information about mapping between documents and the object model. In OpenAPI several interfaces have the same unusual needs, and those are collected in the Helidon OpenAPI-specific `ExpandedTypeDescription` class. (The class which the plug-in generates takes a type parameter which extends `TypeDescription` and a factory function for creating new instances of the desired type.)

For parsing the OpenAPI `@Extension` annotation (always JSON), SmallRye accepts an `AnnotationScannerExtension` which, among other things, supports the `parseExtension` method. The Helidon version of this is based heavily on the source from SmallRye, the key difference being in how the JSON is parsed and handled. There might be ways of doing this with less code but patterning closely after the (working) SmallRye implementation that uses Jackson seemed wise.

Oru custom `Serializer` class mostly sets SnakeYAML formatting configuration before invoking SnakeYAML’s `Yaml#dump` method to serialize the object model. We need to tailor some of the processing beyond what the config settings support, hence the `CustomRepresenter` and `TagSuppressingWriter` inner classes. Note that one of those customizations is to work around a bug in the MP OpenAPI TCK (see the comments in the code).


openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
```diff
@@ -0,0 +1,303 @@
...
+     */
+    static class ExtensionProperty extends Property {
+
+        ExtensionProperty(String name) {
+            super(name, Object.class);
+        }
+
+        @Override
+        public Class<?>[] getActualTypeArguments() {
+            return new Class[0];
```
ljnelson: Just a quick check to make sure you have to do this, and/or it's the right thing to do.  It struck me as surprising (as a layman reviewer) that the `Property` superclass wouldn't default to this value.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
```diff
@@ -0,0 +1,303 @@
...
+        public List<Annotation> getAnnotations() {
+            return null;
+        }
+
+        @Override
+        public <A extends Annotation> A getAnnotation(Class<A> annotationType) {
+            return null;
+        }
+
+        private Extensible<?> asExt(Object object) {
```
ljnelson: Consider making `static`

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    boolean hasDefaultProperty() {
+        return getPropertyNoEx("defaultValue") != null;
+    }
+
+    private boolean setupExtensionType(String key, Node valueNode) {
+        if (isExtension(key)) {
+            switch (valueNode.getNodeId()) {
+                case sequence:
```
ljnelson: These `case` values surprised me; where are they coming from?  Obviously everything compiles so it's fine but maybe a quick one-liner comment somewhere explaining their provenance.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        @Override
+        public Object get(Object object) {
+            return asExt(object).getExtensions().get(getName());
+        }
+
+        @Override
+        public List<Annotation> getAnnotations() {
+            return null;
```
ljnelson: OK to return `null`, or should it be `Collections.emptyList()`?  Don't know the innards of SnakeYAML to know whether one is different from the other.  Obviously avoiding any object allocations is good if possible.

openapi/src/main/java/io/helidon/openapi/ImplTypeDescription.java
```diff
@@ -0,0 +1,168 @@
...
+import org.yaml.snakeyaml.introspector.PropertySubstitute;
+import org.yaml.snakeyaml.introspector.PropertyUtils;
+import org.yaml.snakeyaml.nodes.Node;
+import org.yaml.snakeyaml.nodes.Tag;
+
+/**
+ * Version of {@code TypeDescription} used for implementation classes (rather than the corresponding interfaces) which delegate
+ * most method invocations to the {@code TypeDescription} for the related interface.
+ */
+class ImplTypeDescription extends TypeDescription {
```
ljnelson: Layman reviewer note: surprised me to see `Impl` as a prefix; was expecting a suffix here.  This is true even when I read your class comment.  I'm not sure what would be a better name.  Maybe `DelegatingTypeDescription`?

openapi/src/main/java/io/helidon/openapi/OpenAPIParser.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.eclipse.microprofile.openapi.models.OpenAPI;
+import org.yaml.snakeyaml.TypeDescription;
+import org.yaml.snakeyaml.Yaml;
+import org.yaml.snakeyaml.constructor.Constructor;
+
+/**
+ * Abstraction for SnakeYAML parsing of JSON and YAML.
+ */
+class OpenAPIParser {
```
ljnelson: Could this be `final`?

openapi/src/main/java/io/helidon/openapi/OpenAPIParser.java
```diff
@@ -0,0 +1,56 @@
...
+    static OpenAPI parse(Map<Class<?>, ExpandedTypeDescription> types, Reader reader,
+            OpenAPISupport.OpenAPIMediaType openAPIMediaType) {
+        TypeDescription openAPITD = types.get(OpenAPI.class);
+        Constructor topConstructor = new CustomConstructor(openAPITD);
+
+        types.values()
+                .forEach(topConstructor::addTypeDescription);
+
+        Yaml yaml = new Yaml(topConstructor);
+        OpenAPI result = yaml.loadAs(reader, OpenAPI.class);
```
ljnelson: Do we know if the `Reader` is closed?  I assume `Yaml` does in fact consume the whole thing and closes it afterwards (a pet peeve of mine, but I digress).  A quick comment noting this would be good here.

openapi/src/main/java/io/helidon/openapi/CustomConstructor.java
```diff
@@ -0,0 +1,104 @@
...
+import org.eclipse.microprofile.openapi.models.servers.ServerVariables;
+import org.yaml.snakeyaml.TypeDescription;
+import org.yaml.snakeyaml.constructor.Constructor;
+import org.yaml.snakeyaml.nodes.MappingNode;
+import org.yaml.snakeyaml.nodes.Node;
+import org.yaml.snakeyaml.nodes.NodeId;
+import org.yaml.snakeyaml.nodes.SequenceNode;
+
+/**
+ * Specialized constructor for modifying {@code Node} objects for OpenAPI types that extend {@code Map} to adjust the type of
```
ljnelson: Nit: consider "Specialized SnakeYAML `Constructor`"

openapi/src/main/java/io/helidon/openapi/CustomConstructor.java
```diff
@@ -0,0 +1,104 @@
...
+ * This class customizes the preparation of the node tree in these situations by setting the types for the child nodes explicitly
+ * to the corresponding child type. In OpenAPI 1.1.2 there are two situations, depending on whether the mapped-to type is a
+ * {@code List} or not.
+ * </p>
+ * <p>
+ * The MicroProfile OpenAPI 2.0 versions of the interfaces no longer use this construct of an interface extending {@code Map}, so
+ * ideally we can remove this workaround when we adopt 2.0.
+ * </p>
+ */
+class CustomConstructor extends Constructor {
```
ljnelson: Reading along, perhaps you effectively answer this later, perhaps not: can this be `final`?

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
```diff
@@ -0,0 +1,303 @@
...
+ * </p>
+ * <p>
+ *     A subnode {@code $ref} maps to the {@code ref} property on the MP OpenAPI types. This type
+ *     description simplifies defining the {@code $ref} property to those types that support it.
+ * </p>
+ * <p>
+ *     We use this expanded version of {@code TypeDescription} with the generated SnakeYAMLParserHelper class.
+ * </p>
+ */
+class ExpandedTypeDescription extends TypeDescription {
```
ljnelson: Can this be `final`?

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -116,6 +149,88 @@ public void configureEndpoint(Routing.Rules rules) {
...
+    static synchronized SnakeYAMLParserHelper<ExpandedTypeDescription> helper() {
+        if (helper == null) {
+            helper = SnakeYAMLParserHelper.create(ExpandedTypeDescription::create);
+            adjustTypeDescriptions(helper.types());
+        }
+        return helper;
+    }
+
+    static Map<Class<?>, ExpandedTypeDescription> buildImplsToTypes(SnakeYAMLParserHelper<ExpandedTypeDescription> helper) {
+        return helper.entrySet().stream()
```
ljnelson: Consider wrapping this in `Collections.unmodifiableMap()`.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -116,6 +149,88 @@ public void configureEndpoint(Routing.Rules rules) {
...
+            ExpandedTypeDescription tdWithEnumeration = types.get(c);
+            tdWithEnumeration.substituteProperty("enum", List.class, "getEnumeration", "setEnumeration");
+            tdWithEnumeration.addPropertyParameters("enum", String.class);
+            tdWithEnumeration.addExcludes("enumeration");
+        });
+
+        /*
+         * SnakeYAML derives properties only from methods declared directly by each OpenAPI interface, not from methods defined
+         *  on other interfaces which the original one extends. Those we have to handle explicitly.
+         */
```
ljnelson: Just a note to say I appreciate the comments!!

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -132,8 +247,10 @@ private OpenAPI prepareModel(OpenApiConfig config, IndexView indexView, OpenApiS
                 OpenApiDocument.INSTANCE.reset();
                 OpenApiDocument.INSTANCE.config(config);
                 OpenApiDocument.INSTANCE.modelFromReader(OpenApiProcessor.modelFromReader(config, getContextClassLoader()));
-                OpenApiDocument.INSTANCE.modelFromStaticFile(OpenApiProcessor.modelFromStaticFile(staticFile));
-                if (isAnnotationProcessingEnabled(config)) {
+                if (staticFile != null) {
+                    OpenApiDocument.INSTANCE.modelFromStaticFile(OpenAPIParser.parse(helper().types(), staticFile.getContent(),
+                            OpenAPIMediaType.byFormat(staticFile.getFormat())));
+                }if (isAnnotationProcessingEnabled(config)) {
```
ljnelson: Whoops

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -182,7 +305,7 @@ private void prepareResponse(ServerRequest req, ServerResponse resp) {
             resp.status(Http.Status.OK_200);
             resp.headers().add(Http.Header.CONTENT_TYPE, resultMediaType.toString());
             resp.send(openAPIDocument);
-        } catch (IOException ex) {
+        } catch (Exception ex) {
```
ljnelson: The old version would let `RuntimeException`s be thrown here; not sure if you intended to catch them as well.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -252,6 +375,92 @@ private MediaType chooseResponseMediaType(ServerRequest req) {
...
+        public Object parseExtension(String key, String value) {
+
+            // Inspired by SmallRye's JsonUtil#parseValue method.
+            if (value == null) {
+                return null;
+            }
+
+            value = value.trim();
+
+            if ("true".equals(value) || "false".equals(value)) {
```
ljnelson: Maybe `equalsIgnoreCase()`?

openapi/src/main/java/io/helidon/openapi/Serializer.java
```diff
@@ -0,0 +1,349 @@
...
+            serialize(types, implsToTypes, openAPI, writer, YAML_DUMPER_OPTIONS, DumperOptions.ScalarStyle.PLAIN);
+        }
+    }
+
+    private static void serialize(Map<Class<?>, ExpandedTypeDescription> types,
+            Map<Class<?>, ExpandedTypeDescription> implsToTypes, OpenAPI openAPI, Writer writer,
+            DumperOptions dumperOptions,
+            DumperOptions.ScalarStyle stringStyle) {
+
+        Yaml yaml = new Yaml(new CustomRepresenter(types, implsToTypes, dumperOptions, stringStyle), dumperOptions);
```
ljnelson: Hazy old memories of making sure that SnakeYAML's `SafeConstructor` or something is in play; I'll try to dig for more details.

openapi/src/main/java/io/helidon/openapi/Serializer.java
```diff
@@ -0,0 +1,349 @@
...
+                }
+            }
+            return propertyValue;
+        }
+
+        @Override
+        protected MappingNode representJavaBean(Set<Property> properties, Object javaBean) {
+            MappingNode result = super.representJavaBean(properties, javaBean);
+            processExtensions(result, javaBean);
+            representedObjects.clear();
```
ljnelson: Maybe put in a `finally` block?

openapi/src/main/java/io/helidon/openapi/Serializer.java
```diff
@@ -0,0 +1,349 @@
...
+         * Some SmallRye implementation classes have properties not supported by the implemented interface, so ignore those or
+         * the serialized YAML will contain SmallRye-only properties.
+         *
+         * @param javaBean the bean being serialized
+         * @param property the property being serialized
+         * @return true if the property should be processes; false otherwise
+         */
+        private boolean okToProcess(Object javaBean, Property property) {
+            boolean reject = false;
+            reject |= Parameter.class.isAssignableFrom(javaBean.getClass()) && property.getName().equals("hidden");
```
ljnelson: Been a while since I've seen this operator.  Won't this always assign `false` to `reject`?  {Heads off to JLS page}

openapi/src/main/java/io/helidon/openapi/Serializer.java
```diff
@@ -0,0 +1,349 @@
...
+            return result;
+        }
+    }
+
+    private static class DelegatingProperty extends Property {
+
+        private final Property delegate;
+
+        private DelegatingProperty(Property delegate, String name) {
+            super(name, delegate.getType());
```
ljnelson: Nit: why reverse the order?

openapi/src/main/java/io/helidon/openapi/Serializer.java
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        @Override
+        public Class<?>[] getActualTypeArguments() {
+            return delegate.getActualTypeArguments();
+        }
+
+        @Override
+        public String getName() {
+            return super.getName();
```
ljnelson: Surprise: was expecting `delegate`.  Don't know if that's on purpose or not.

openapi/src/main/java/module-info.java
```diff
@@ -27,6 +27,9 @@
     requires jandex;
 
     requires smallrye.open.api;
+    requires java.json;
+    requires java.desktop; // for java.bean package
```
ljnelson: Nit: `java.beans`, yes?

openapi/src/test/java/io/helidon/openapi/TestUtil.java
```diff
@@ -221,6 +222,12 @@ public static JsonStructure jsonFromResponse(HttpURLConnection cnx) throws IOExc
         return result;
     }
 
+    static JsonStructure jsonFromReader(Reader reader) {
+        JsonReader jsonReader = JSON_READER_FACTORY.createReader(reader);
+        JsonStructure result = jsonReader.read();
```
ljnelson: Nit: any kind of `close` operation that needs to be performed here?

openapi/src/main/java/io/helidon/openapi/Serializer.java
ljnelson: Oh, never mind.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
tjquinno: In fact, this is correct. We know that the `ExtensionProperty` class will be used only for OpenAPI extensions, and we know those properties will have no type arguments, so this method can safely always return an empty array. I will insert a comment to this effect and will also change the method so it does not create a new array upon each invocation.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
tjquinno: As written it invokes the instance method `getName` for the error message.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
tjquinno: Comment added.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
tjquinno: Fixed.

openapi/src/main/java/io/helidon/openapi/ImplTypeDescription.java
tjquinno: I think an `Impl` suffix might be problematic because this class is a `TypeDescription` for implementation classes (from SmallRye) rather than for interfaces (from MP OpenAPI). Plus `TypeDescription` is itself a class, so I'm not sure about the expectation of seeing an "implementation" of another class.

The apparent clarity of `DelegatingTypeDescription` might actually be a little misleading, because, although it delegates every method to `delegate.something`, the `getType` method_does not_ delegate to `delegate.getType()` but to `delegate.impl()`. This is what makes this an `ImplTypeDescription`.

I will make this more explicit in the class comments.

openapi/src/main/java/io/helidon/openapi/OpenAPIParser.java
tjquinno: Yes. Changed.

openapi/src/main/java/io/helidon/openapi/OpenAPIParser.java
tjquinno: `Yaml` does not in fact close the `Reader`. Because the `Reader` is passed as an argument, I've changed the call sites to close their readers.

openapi/src/main/java/io/helidon/openapi/CustomConstructor.java
tjquinno: Fixed

openapi/src/main/java/io/helidon/openapi/CustomConstructor.java
tjquinno: Yes. Changed.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
tjquinno: This one cannot be `final`; the inner `SchemaTypeDescription` class extends it. But _that_ one can be `final`; I changed that.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Done

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Aim to please!

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Fixed

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: The intent of the change is to catch _any_ `Exception` (and convert it into a 500 response with a brief explanation sent back to the client).

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: I pretty much lifted this method from the corresponding SmallRye one, but I think your suggestion is a good one. Done.

openapi/src/main/java/io/helidon/openapi/Serializer.java
tjquinno: IIRC that's if you want to restrict the document-to-object mapping to objects that come with Java (i.e., not your own). (see https://bitbucket.org/asomov/snakeyaml/wiki/Documentation and search for SafeConstructor)

openapi/src/main/java/io/helidon/openapi/Serializer.java
tjquinno: Clearing `representedObjects` is an awkward but effective way to prevent SnakeYAML from using anchors and links in the output document (this is the least bad way, currently, BTW) because Jackson - which is used in the MP OpenAPI TCK - does not process those correctly. If some method invoked here throws an exception then all bets are off and we don't really care about the contents of `representedObjects` any more. I've added a comment explaining this.

openapi/src/main/java/io/helidon/openapi/Serializer.java
tjquinno: I'm adding a comment in the code explaining. If, over time, we have to add more cases to ignore then we can just add a new line similar to the existing `reject |= ...` one for each case to be skipped without perturbing the other lines in the method.

openapi/src/main/java/io/helidon/openapi/Serializer.java
tjquinno: No longer an issue. See below.

openapi/src/main/java/io/helidon/openapi/Serializer.java
tjquinno: The whole `DelegatingProperty` class (and the logic that instantiated it) was a holdover from an earlier approach that's no longer used. Removed.

openapi/src/main/java/module-info.java
tjquinno: Fixed

openapi/src/test/java/io/helidon/openapi/TestUtil.java
tjquinno: Yes. Done.

openapi/src/main/java/io/helidon/openapi/ExpandedTypeDescription.java
ljnelson: 🤦‍♂ Got it.

microprofile/tests/tck/tck-rest-client/pom.xml
```diff
@@ -37,8 +37,8 @@
             <exclusions>
                 <exclusion>
                     <!-- TCK requires Jackson, not JSON-B -->
-                    <groupId>org.glassfish.jersey.media</groupId>
-                    <artifactId>jersey-media-json-binding</artifactId>
+                    <groupId>jakarta.json.bind</groupId>
+                    <artifactId>jakarta.json.bind-api</artifactId>
```
ljnelson: Just a dumb comment: comment says Jackson, artifact is JSON-B, yes?

PR #1473, opened 2020-03-05
Resolves JTA/JPA transaction synchronization issues

ljnelson:

### DIFF
--- a/integrations/cdi/jpa-cdi/pom.xml
+++ b/integrations/cdi/jpa-cdi/pom.xml
@@ -76,6 +76,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>helidon-integrations-cdi-hibernate</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>io.helidon.microprofile.config</groupId>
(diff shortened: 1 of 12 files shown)


ljnelson:
This PR, at a high level, ensures that Surefire will run twice (in normal operation): once using Eclipselink, and once using Hibernate.  This ensures that our tests run against both major JPA providers we support.

To accomplish this, the `pom.xml` effectively copies classes into two "weaving areas"—areas where static bytecode enhancement can take place for each JPA provider without polluting the other's classes.  Surefire is then configured to run against each weaving area so that the "right" classes are used at test time.

In addition, the Maven default executions—`default-test` and `default-cli`—are configured so that if they are invoked they work properly, though in such cases only the Eclipselink tests will be run.

ljnelson:
This PR also exposed several unspecified behaviors.  In one, Eclipselink happily lets a once-managed-now-detached entity get re-persisted, whereas Hibernate throws an exception in such a use case.  In another, Eclipselink defers acquiring database-generated identifiers until flush time; Hibernate seems to somehow come up with them early.  Each of these cases is exactly the sort of thing that makes running our tests against both providers valuable.

ljnelson:
This PR also proves that transactional interceptor-governed methods work fine under Eclipselink and Hibernate.

Issue #1293, opened 2020-01-13
JPA and JTA in MP native-image


tomas-langer:
Add support for JTA and JPA with Hibernate in MP native-image

tomas-langer:
Related to #716

Tomas-Kraus:
Unassigned from myself, I'm no longer maintaining this module.

tomas-langer:
Closing this issue, as it is targetted and 2.x (no longer supported).

PR #1478, opened 2020-03-09
JPA and JTA for native image

tomas-langer:
Related to #1293 
This is the first experimental support for Hibernate and JTA in native image.
See integration test `tests/integration/native-image/mp-2` that shows what is supported.


integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/HibernateJtaPlatform.java
```diff
@@ -29,13 +29,13 @@
...
  * {@link UserTransaction} instances that are supplied to it at
- * {@linkplain #CDISEJtaPlatform(TransactionManager, UserTransaction)
+ * {@linkplain #HibernateJtaPlatform(TransactionManager, UserTransaction)
  * construction time}.
  *
  * @see AbstractJtaPlatform
  */
 @ApplicationScoped
-public class CDISEJtaPlatform extends AbstractJtaPlatform {
+public class HibernateJtaPlatform extends AbstractJtaPlatform {
```
ljnelson: `JtaPlatform` is inherently a Hibernate construct, so renaming this to `HibernateJtaPlatform` is redundant.

integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/HibernateJtaPlatformProvider.java
```diff
@@ -30,14 +30,13 @@
...
  *
- * @see CDISEJtaPlatform
+ * @see HibernateJtaPlatform
  */
-public final class CDISEJtaPlatformProvider implements JtaPlatformProvider {
-
+public final class HibernateJtaPlatformProvider implements JtaPlatformProvider {
   /**
-   * Creates a new {@link CDISEJtaPlatformProvider}.
+   * Creates a new {@link HibernateJtaPlatformProvider}.
```
ljnelson: `JtaPlatformProvider` is inherently a Hibernate construct so renaming this to `HibernateJtaPlatformProvider` is redundant.

integrations/cdi/hibernate-cdi/src/main/resources/META-INF/services/org.hibernate.engine.transaction.jta.platform.spi.JtaPlatformProvider
```diff
@@ -11,4 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-io.helidon.integrations.cdi.hibernate.CDISEJtaPlatformProvider
\ No newline at end of file
+
+io.helidon.integrations.cdi.hibernate.HibernateJtaPlatformProvider
```
ljnelson: As noted earlier in the review, renames like this should be reverted.

integrations/cdi/jta-cdi/src/main/java/io/helidon/integrations/jta/cdi/NarayanaExtension.java
```diff
@@ -54,13 +57,14 @@
  * 2.0 SE environment</a>.
  */
 public final class NarayanaExtension implements Extension {
-
+    static {
+        HelidonFeatures.register(HelidonFlavor.MP, "JTA");
```
ljnelson: I'd prefer keeping the `HelidonFeatures` stuff out of this extension as right now it is capable of being a generic CDI portable extension usable by any CDI implementation in any product.  Perhaps this could be located elsewhere?

tomas-langer: I am not sure how to handle this - we need to list the features on startup, and JTA, JPA and Hibernate/Eclipse link are kind of required here.
And I do not like the idea of creating a new module for each of these that would be Helidon specific.
This would just bring a dependency on `helidon-common` module and give us consistency. 
Also this is Helidon product, so I do not understand why we should not have a dependency on our common module.
It still would work with any CDI implementation, as it would just ignore this class...

integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/HibernateJtaPlatform.java
tomas-langer: I will change it back and refresh the PR

integrations/cdi/jta-cdi/src/main/java/io/helidon/integrations/jta/cdi/NarayanaExtension.java
ljnelson: Could you move class construction/instantiation for all features into the `HelidonFeatures` class directly, and read some sort of descriptor out of `META-INF`?

tomas-langer: Not really. I want the features listed only when used, not just when on classpath.

integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/HibernateJtaPlatformProvider.java
tomas-langer: Reverted

integrations/cdi/hibernate-cdi/src/main/resources/META-INF/services/org.hibernate.engine.transaction.jta.platform.spi.JtaPlatformProvider
tomas-langer: Reverted

integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/HibernateJtaPlatform.java
tomas-langer: Reverted

integrations/cdi/jta-cdi/src/main/java/io/helidon/integrations/jta/cdi/NarayanaExtension.java
tomas-langer: I am not sure about your concern. Is it that we add one library to dependencies?

ljnelson: Here are some of my concerns:
* What is a Helidon feature (particularly with respect to CDI)?  Is it a portable extension?  Is it many?  Is it a bean archive?  How are they defined?  What granularity do they have?  Why this extension?  What happens if I add another extension in this package?
* Why _must_ they be registered in a static singleton?  Why must they be pushed?  Why must this coupling be tight?  Suppose `HelidonFeatures` grows and becomes complicated and the static call in the static initializer fails and this class doesn't load.  Is it straightforward to debug?
* So far this is the only compile-time `io.helidon` dependency in all of the integrations we have on anything outside of the `integrations` area.  Everything else is pure MicroProfile dependencies.  These integrations have proven to be separate from the rest of Helidon because of course they are runtime components that can be deployed, or not, as people see fit, within any MicroProfile application.  I personally would like to keep it this way.
* CDI has an event mechanism for broadcasting things of interest that facilitates loose coupling (yay) rather than tight coupling (boo).  It seems to me that any Helidon features facility should be built on top of it (at least in the CDI/MicroProfile area) since the extension in this case doesn't care.  Perhaps extensions could simply do `event.fire(this)` on the initialization of the application scope and then some collector in the Helidon-specific container could sort out how to make features out of this.
Anyway those are some that I can think of right now.

romain-grecourt: The HelidonFeatures needs to be re-worked, but, this is a must have for all components. The CDI / Microprofile pureness (TM) is nice but somewhat irrelevant to the Helidon product.

ljnelson: You're welcome to approve this PR; I don't feel comfortable doing that.

tomas-langer: Helidon Features is a cross SE/MP feature, so it cannot utilize CDI at all.
We may add a way to read features from a descriptor file (such as `META-INF/helidon/feature.json` or similar). It must be aligned with our CLI work.
We can talk about it on our meeting.

Nevertheless the "pureness" of our integrations implementations is not a design goal of Helidon, so using Helidon specific APIs is not forbidden and in case it is easier, it is actually recommended.

tomas-langer: Removed from this PR, we will re-introduce features to integrations once we agree on how to hanle it.

PR #1479, opened 2020-03-09
Mp messaging example

danielkec:


danielkec:
@romain-grecourt Fancy is what I have been aiming for, is 140 kB really too much?

romain-grecourt:
It's trying to be fancy, but it's not. It's an example, keep it simple rather than lame, go to the point.
I really want to avoid duplicating assets from the website all over the place.

m0mus:
I personally like the design with all these animations. It makes it nice and dynamic. I don't see anything wrong with checking it in into repository. I agree it's better to put js files in the repository. 

@romain-grecourt Are there any other concerns except that you don't like the design?

romain-grecourt:
It's hard enough to maintain a website and keep things consistent. Let's not copy stuff from the website all over the place.

I don't agree with examples being fancy, they should be simple and illustrate a point. Demos can be fancy, is this a demo ?

At the very least get rid of the website background, it's not meant to be reused.

m0mus:
I think that one of the purposes of having the graphics is that we can use it in Helidon related assets such as are presentations, demos and samples. I don't agree that it must be kept on the web site only. Why the website background cannot be reused?

romain-grecourt:
The re-usable assets are here: https://github.com/oracle/helidon-site/tree/master/graphics.

The website background was specifically designed for the parallax effect, it's not meant to be re-used ; I'm guessing that background.png was built using the individual layers of the parallax.

examples/microprofile/messaging-sse/README.md
```diff
@@ -0,0 +1,18 @@
+# Helidon Reactive Messaging Example
+
+ Example showing
+ * [Microprofile Reactive Messaging](https://github.com/eclipse/microprofile-reactive-messaging) 
+ with [Microprofile Reactive Stream Operators](https://github.com/eclipse/microprofile-reactive-streams-operators) 
+ connected to [Server-Sent Events](https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/sse.html).
+
+## Build and run
+
+With JDK8+
```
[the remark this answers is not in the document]
danielkec: thx fixed

danielkec:
Sorry for rebase, background image is removed and snippet is generated locally 
Peek 2020-03-25 10-44

examples/microprofile/messaging-sse/src/main/java/io/helidon/microprofile/example/messaging/sse/Main.java
```diff
@@ -0,0 +1,52 @@
...
+    private Main() {
+    }
+
+    /**
+     * Starts server and initializes CDI container manually.
+     *
+     * @param args command line arguments (ignored)
+     */
+    public static void main(String[] args) {
+        Server server = Server.builder()
```
tomas-langer: Just do `Server.create().start()`

danielkec: Fixed thx

PR #1491, opened 2020-03-11
Fail fast when policy validation fails because of setup/syntax.

### LINKED ISSUE #1487: EL Policy statement parsing failure does not fail setup
When the EL Policy statement contains an invalid expression, such as backtick instead of single quote in `${inRole(user, 'admin')}`, the parsing fails, but is only represented as a `403` response code, instead of a proper parsing expression that fails the application.

tomas-langer:

Resolves #1487 for 2.x

### DIFF
--- a/security/abac/policy-el/src/main/java/io/helidon/security/abac/policy/el/JavaxElPolicyExecutor.java
+++ b/security/abac/policy-el/src/main/java/io/helidon/security/abac/policy/el/JavaxElPolicyExecutor.java
@@ -162,7 +162,7 @@ public final class JavaxElPolicyExecutor implements PolicyExecutor {
                     expressionFactory = ExpressionFactory.newInstance();
                 } catch (ELException e) {
                     throw new SecurityException("Failed to configure ABAC Policy support for Jakarta Expression Language,"
-                                                        + " no implementation found through service loader.");
+                                                        + " no implementation found through service loader.", e);
                 }
             }
             return new JavaxElPolicyExecutor(this);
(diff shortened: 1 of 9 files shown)


security/abac/policy-el/src/main/java/io/helidon/security/abac/policy/el/JavaxElPolicyExecutor.java
```diff
@@ -164,9 +161,8 @@ public JavaxElPolicyExecutor build() {
                 try {
                     expressionFactory = ExpressionFactory.newInstance();
                 } catch (ELException e) {
-                    LOGGER.fine(
-                            "Expression language factory not available. To support policy expressions, you need to add some "
-                                    + "implementation to class/module path");
+                    throw new SecurityException("Failed to configure ABAC Policy support for Jakarta Expression Language,"
```
ljnelson: Consider wrapping `e` so it isn't lost.

tomas-langer: Good catch, will fix

tomas-langer: Fixed.

security/abac/policy-el/src/main/java/io/helidon/security/abac/policy/el/JavaxElPolicyExecutor.java
```diff
@@ -116,11 +116,8 @@ public void executePolicy(String policyStatement, Errors.Collector collector, Pr
                 collector.fatal(this, "Policy statement \"" + policyStatement + "\" evaluated to false");
             }
         } catch (Exception e) {
-            collector.fatal(this,
-                            "Policy statement \"" + policyStatement + "\" evaluated to an exception " + e.getClass()
-                                    .getName() + " with message: " + e.getMessage());
-
             LOGGER.log(Level.FINEST, e, () -> "Statement " + policyStatement + " evaluation failed");
+            throw new SecurityException("Policy statement \"" + policyStatement + "\" evaluated to an exception", e);
```
danielkec: Is logging still needed here?

tomas-langer: The security exception is propagated outside of security (and outside of Helidon code), so yes, I think it is reasonable to log it (that is why it is in `FINEST` level)

PR #1493, opened 2020-03-11
Support multiple apps in OpenAPI document

### LINKED ISSUE #1288: OpenAPI incomplete with multiple JAX-RS applications
When I have two JAX-RS applications in my MP project, the OpenAPI is generated only for one of them (a random one is picked).

I will link a reproducer.

tjquinno:
Resolves #1288 

The SmallRye OpenAPI annotation scanning does not correctly handle multiple applications in the server. To deal with this, the Helidon MP layer:
* invokes the SmallRye OpenAPI scanner once for each app, restricting the scanner so it works only with that app’s classes, thus producing a temporary OpenAPI model for just that app; then
* merges the resulting models into one (using a SmallRye utility method) for use by SmallRye as the model resulting from annotation scanning.

(Because OpenAPI is also supported in SE this PR includes some changes in SE to enhance the contract of `OpenAPISupport` and its builder. Functionally, SE is unaffected by this change.)

To locate the apps, these revised Helidon MP OpenAPI layer uses the pre-existing JaxRsCdiExtension which harvests not only annotated apps but also ones registered programmatically. 

There are a few changes to the MP server component so code that creates a `JaxRsApplication` can indicate if it's a synthetic application being represented. This is useful because the OpenAPI enhancements do not need to handle a synthetic app.


microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/MPOpenAPIBuilder.java
```diff
@@ -30,15 +44,106 @@
...
+    private static boolean isNonSynthetic(JaxRsApplication jaxRsApp) {
+        return !jaxRsApp.synthetic();
+    }
+
+    private Optional<? extends Application> appInstance(JaxRsApplication jaxRsApp) {
+        Application preexistingApp = jaxRsApp.resourceConfig().getApplication();
+        if (preexistingApp != null) {
+            return Optional.of(preexistingApp);
+        }
+        return instantiate(jaxRsApp.applicationClass());
```
tomas-langer: Please replace with 
```
return jaxRsApp.applicationClass()
                .flatMap(MPOpenAPIBuilder::instantiate);
```
And remove the `Optional<Class>` parameter from `instantiate` methods. Using Optional as a parameter is generally not recommended.

microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/MPOpenAPIBuilder.java
```diff
@@ -48,11 +153,34 @@ public IndexView indexView() {
...
         return this;
     }
 
+    MPOpenAPIBuilder config(Config mpConfig) {
+        this.mpConfig = mpConfig;
+        openAPIConfig(new OpenApiConfigImpl(mpConfig));
+        return this;
+    }
+
+    private static Optional<? extends Application> instantiate(Optional<Class<? extends Application>> optionalAppClass) {
```
tomas-langer: To work with my above note, replace with 
```
    private static Optional<? extends Application> instantiate(Class<? extends Application> appClass) {
        try {
            return Optional.of(appClass.getConstructor().newInstance());
        } catch (InstantiationException | IllegalAccessException | InvocationTargetException e) {
        } catch (NoSuchMethodException e) {
            // Wrapper app does not have a no-args constructor so we canont instantiate it.
            return Optional.empty();
        }
    }
```

PR #1499, opened 2020-03-12
Fix completable queue and clean-up of messaging tests

danielkec:
* Hamcrestination
  * Flaky tests fix
  * Back-pressure aware completable queue 
  * Source acknowledgement 

### DIFF
--- a/microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AssertableTestBean.java
+++ b/microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AssertableTestBean.java
@@ -20,10 +20,57 @@ package io.helidon.microprofile.messaging;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import org.hamcrest.Matcher;
 
 public interface AssertableTestBean {
 
     Set<String> TEST_DATA = new HashSet<>(Arrays.asList("teST1", "TEst2", "tESt3"));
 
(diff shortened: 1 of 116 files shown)


microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AsyncTestBean.java
```diff
@@ -0,0 +1,40 @@
...
+package io.helidon.microprofile.messaging;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public interface AsyncTestBean {
+    ExecutorService executor = Executors.newSingleThreadExecutor();
```
tomas-langer: This is a constant - naming convention wrong, also the actual behavior may be different from what you expect
- as soon as you shutdown the executor, all subsequent requests to execute anything would fail, and method `awaitShutdown` will work for a single class implementing this interface

microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AssertableTestBean.java
```diff
@@ -20,10 +20,57 @@
...
 
 public interface AssertableTestBean {
 
     Set<String> TEST_DATA = new HashSet<>(Arrays.asList("teST1", "TEst2", "tESt3"));
 
     void assertValid();
+
+    default void await(String msg, CountDownLatch countDownLatch) {
+        try {
+            countDownLatch.await(500, TimeUnit.MILLISECONDS);
```
tomas-langer: `CountDownLatch.await` returns a boolean that tells you whether it succeeded, or timed out. That should be asserted to make sure the CDL completed.

danielkec: fixed

microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AsyncTestBean.java
danielkec: Fixed

PR #1505, opened 2020-03-12
Kafka connector

jbescos:
Find here the spec of this:
https://download.eclipse.org/microprofile/microprofile-reactive-messaging-1.0/microprofile-reactive-messaging-spec.pdf

It was mostly done by @danielkec months ago. I only took his changes and I did some small fixes and some tests.


bom/pom.xml
```diff
@@ -795,6 +795,16 @@
...
+                <groupId>io.helidon.microprofile</groupId>
+                <artifactId>helidon-microprofile-reactive-streams</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.microprofile</groupId>
+                <artifactId>helidon-microprofile-messaging</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
         </dependencies>
```
danielkec: It would be maybe better to rebase on top of current master, those two dependencies are already present

jbescos: Just I noticed I can get rid of everything except the new helidon/messaging. I'm going to make it simple in a new PR, with a clean history.

PR #1508, opened 2020-03-12
Ensures that JtaDataSource instances that are acquired when a transaction is already active have their Synchronizations registered appropriately

ljnelson:


ljnelson:
This backports #1497 to the `helidon-1.x` branch.

Issue #1500, opened 2020-03-12
More reactive operators (tracking issue)


akarnokd:
Proposed operators:

- [x] `Multi.defaultIfEmpty(T)` (#1520)
- [x] `Multi.defer(Supplier)` (#1503)
- [x] `Multi.from(CompletionStage)` (#1511)
- [x] `Multi.from(Stream)` (#1525)
- [x] `Multi.interval()` (#1526)
- [x] `Multi.observeOn` (#1546)
- [x] `Multi.reduce(BiFunction)` (#1504)
- [x] `Multi.reduce(Supplier, BiFunction)` (#1504)
- [x] `Multi.retry(long)` (#1541)
- [x] `Multi.retry(BiPredicate<Throwable, Long>)` (#1541)
- [x] `Multi.retryWhen(BiFunction<Throwable, Long, Publisher>)` (#1541)
- [x] `Multi.switchIfEmpty(Publisher)` (#1527)
- [x] ~~`Multi.takeUntil(CompletionStage)`~~ decided not to have a dedicated operator for this
- [x] `Multi.takeUntil(Publisher)` (#1502)
- [x] `Multi.timeout()` (#1532)
- [x] `Multi.timer()` (#1516)

---------------

- [x] `Single.defaultIfEmpty(T)` (#1520)
- [x] `Single.defer(Supplier)` (#1503)
- [x] `Single.flatMapIterable(Function)` (#1517)
- [x] `Single.from(CompletionStage)` (#1511)
- [x] `Single.observeOn` (#1546)
- [x] `Single.retry(long)` (#1541)
- [x] `Single.retry(BiPredicate<Throwable, Long>)` (#1541)
- [x] `Single.retryWhen(BiFunction<Throwable, Long, Publisher>)` (#1541)
- [x] `Single.switchIfEmpty(Single)` (#1527)
- [x] ~~`Single.takeUntil(CompletionStage)`~~ decided not to have a dedicated operator for this
- [x] `Single.takeUntil(Publisher)` (#1502)
- [x] `Single.timer()` (#1516)
- [x] `Single.timeout()` (#1532)

akarnokd:
Done.

PR #1511, opened 2020-03-13
[Reactive] Reimplement Microprofile-RS, fix operators, add from(CS)

akarnokd:
This PR reimplements the Microprofile Reactive Streams adapter of Helidon.

- Fix iterator failure not cancelling the flow in `Multi.flatMapIterable`.
- Fix `Multi.flatMap` not properly accounting items emitted.
- Add `from(CompletionStage)` implementation (Related: #1500) required by MPRS.
- Expose `onCancel`.
- Fix `OnErrorResumeNext` possibly trying to self-suppress.


akarnokd:
> [ERROR] /pipeline/source/microprofile/reactive-streams/src/main/java/io/helidon/microprofile/reactive/HelidonReactiveStreamsEngine.java:129:5: Method length is 352 lines (max allowed is 150). [MethodLength]

I don't know what to do. The method is basically a large switch-case because of the sheer number of Stage to support of MPRS.

akarnokd:
I try my best to make this work but the full project doesn't build on Windows so my only feedback is from the CI failures.

romain-grecourt:
@akarnokd some fixes for the windows build are underway, hopefully it's going to be fixed soon.

akarnokd:
@danielkec Could you look into that failure with the `InternalProcessor` overflowing some queue?

danielkec:
> @danielkec Could you look into that failure with the `InternalProcessor` overflowing some queue?

@akarnokd Sure fix is already prepared at #1499

Issue #1350, opened 2020-02-04
OpenAPI support should have customizable context root in MP


tomas-langer:
Currently we can customize OpenAPI context root in Helidon SE, but we do not have such an option in MP.
This should be configurable, such as:
```
openapi:
  web-context: "/manage/openapi"
```

PR #1521, opened 2020-03-16
Openapi custom context root 2.x

tjquinno:
Addresses #1350 for Helidon 2.x

The changes include a little refactoring to reuse logic from both SE and MP so both can use Helidon config to set an alternate, non-default context root.

### DIFF
--- a/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
+++ b/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
@@ -590,7 +590,7 @@ public class OpenAPISupport implements Service {
      * helidonConfig settings
      */
     public static OpenAPISupport create(Config config) {
-        return builderSE().helidonConfig(config).build();
+        return builderSE().config(config).build();
     }
 
     /**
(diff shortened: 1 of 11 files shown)


openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -620,6 +623,27 @@ public OpenAPISupport build() {
...
+         * The {@code Config} object can specify {@value #CONFIG_PREFIX}.web-context
+         * and {@value #CONFIG_PREFIX}.static-file in addition to settings
+         * supported by {@link OpenAPIConfigImpl.Builder}.
+         *
+         * @param config the {@code Config} object possibly containing settings
+         * @exception NullPointerException if the provided {@code Config} is null
+         * @return updated builder instance
+         */
+        public Builder helidonConfig(Config config) {
+            config.get(CONFIG_PREFIX + ".web-context")
```
tomas-langer: You should not use the  `CONFIG_PREFIX` - the config should be on the node of `openapi`. Microprofile component configuring this class should request that node before calling this method.

The method should be called `config` - that is the rule for any builder in any Helidon component.

openapi/src/main/java/io/helidon/openapi/SEOpenAPISupportBuilder.java
```diff
@@ -49,8 +48,7 @@
      * @return updated builder instance
      */
     public SEOpenAPISupportBuilder helidonConfig(Config config) {
-        config.get(CONFIG_PREFIX + ".web-context").asString().ifPresent(this::webContext);
-        config.get(CONFIG_PREFIX + ".static-file").asString().ifPresent(this::staticFile);
+        super.helidonConfig(config);
```
tomas-langer: Method should be called `config`

PR #1524, opened 2020-03-16
Openapi custom context root 1.x

### LINKED ISSUE #1350: OpenAPI support should have customizable context root in MP

tjquinno:
Resolves #1350 for 1.x. Very similar to the 2.x changes for the same issue.

### DIFF
--- a/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
+++ b/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
@@ -620,7 +620,7 @@ public class OpenAPISupport implements Service {
          * <p>
          * The {@code Config} object can specify {@value #CONFIG_PREFIX}.web-context
          * and {@value #CONFIG_PREFIX}.static-file in addition to settings
-         * supported by {@link OpenAPIConfigImpl.Builder}.
+         * supported by {@code OpenAPIConfigImpl.Builder}.
          *
          * @param config the {@code Config} object possibly containing settings
          * @exception NullPointerException if the provided {@code Config} is null
--- a/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
+++ b/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
@@ -604,7 +604,10 @@ public class OpenAPISupport implements Service {
      */
     public abstract static class Builder implements io.helidon.common.Builder<OpenAPISupport> {
 
-        static final String CONFIG_PREFIX = "openapi";
+        /**
+         * Config key for locating openapi settings within a Helidon config object.
+         */
+        public static final String CONFIG_KEY = "openapi";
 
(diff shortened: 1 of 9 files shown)


openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -613,6 +615,27 @@ public OpenAPISupport build() {
...
+         * The {@code Config} object can specify {@value #CONFIG_PREFIX}.web-context
+         * and {@value #CONFIG_PREFIX}.static-file in addition to settings
+         * supported by {@code OpenAPIConfigImpl.Builder}.
+         *
+         * @param config the {@code Config} object possibly containing settings
+         * @exception NullPointerException if the provided {@code Config} is null
+         * @return updated builder instance
+         */
+        public Builder helidonConfig(Config config) {
+            config.get(CONFIG_PREFIX + ".web-context")
```
tomas-langer: No prefix should be used, config should be on `openapi` node already.
Method should be called `config`

tjquinno: Changing the method name is no problem, as it is a new method. 

Are you suggesting that the `Config` object passed in should already be the node for "openapi" or that this method should do a `get` using the prefix and then the `get`s for the two properties would use the openapi node?

The released 1.x OpenAPI code was written to expect the top-level `Config` node as the argument, so changing that would be a backward-incompatible change.

tjquinno: This method, `helidonConfig`, has been present on `SEOpenAPISupportBuilder` in SE OpenAPI since we first created it, as has the expectation that the `Config` node passed in is the top-level one, within which `openapi` might exist.

Because we want to support Helidon config in MP also, this method is migrating to the superclass `OpenAPISupport`. So while it might appear that this is a new method, it is not a new part of the contract between SE OpenAPI and SE apps.

Changing either the method name or the expectation of which config node is passed - in 1.x anyway - would be a backward-incompatible change.

tjquinno: We can make the incompatible change in the 2.x PR to align OpenAPI with the other components that have similar config support.

tjquinno: For 1.x, we can _add_ the `config` method as you suggest -- with a `Config` argument that should be the `openapi` node -- while keeping the existing `helidonConfig` method to preserve backward compatibility and mark `helidonConfig` as deprecated.

tomas-langer: yes please

Issue #1559, opened 2020-03-20
helidon-microprofile-openapi dependency convergence and duplicate class problems


Holubow:
## Problem Description
My project uses a handful of Helidon MicroProfile 2.0.0-M2 dependencies and I run the maven-enforcer-plugin and the duplicate-finder-plugin and I have found that helidon-microprofile-openapi has dependency convergence and duplicate class problems with other Helidon modules.

It seems that helidon-microprofile-openapi is using newer versions of some dependencies shared with other Helidon modules.

This is an example of the exclusions I need to apply to helidon-microprofile-openapi in my Maven POM to address the problems for illustration:

	<exclusions>
		
		<exclusion>
			<groupId>org.jboss.logging</groupId>
			<artifactId>jboss-logging</artifactId>
		</exclusion>
		
		<exclusion>
			<groupId>org.jboss</groupId>
			<artifactId>jandex</artifactId>
		</exclusion>
		
		<exclusion>
			<groupId>org.eclipse.microprofile.config</groupId>
			<artifactId>microprofile-config-api</artifactId>
		</exclusion>
		
		<exclusion>
			<groupId>jakarta.json</groupId>
			<artifactId>jakarta.json-api</artifactId>
		</exclusion>
		
		<exclusion>
			<groupId>org.glassfish</groupId>
			<artifactId>jakarta.json</artifactId>
		</exclusion>
	</exclusions>

## Environment Details
* Helidon Version: 2.0.0-M2
* Helidon SE or Helidon MP: Helidon MP
* JDK version: jdk-11.0.5
* OS: win64
* Docker version (if applicable):

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

## Steps to reproduce
[//]: # "Step by step instructions to reproduce the problem"
[//]: # "Provide sample code/application if relevant"

Create a Helidon project with dependencies on helidon.microprofile-server and helidon-microprofile-openapi.
Look at the different versions of the dependencies listed above in helidon.microprofile-server and helidon-microprofile-openapi using Eclipse's Dependency Hierarchy view.
The maven-enforcer-plugin will complain about dependency convergence errors.
The duplicate-finder-maven-plugin will complain about duplicate classes.
Reviewing the underlying POM versions should also illustrate the difference in versions.

romain-grecourt:
[...]

PR #1567, opened 2020-03-23
[skip ci] WIP: Fix dependency convergence issues

tjquinno:
Addresses #1559 

Helidon has transitive (or explicit and transitive) dependencies on both JBoss logging and Jakarta JSON along different dependency paths and the transitive dependencies are pulling in conflicting versions.

In OpenAPI, our dependency on SmallRye's OpenAPI pulls in JBoss logging. And an artifact outside of Helidon -- `org.glassfish.jersey.media:jersey-media-json-binding`-- itself depends on jarkata.json and also depends on yasson which in turn depends on a newer version of jakarta.json.


tjquinno:
Tomas has taken on this work, so I'm closing this PR.

PR #1568, opened 2020-03-23
[Reactive] Workaround for an RS 1.0.3 TCK bug

akarnokd:
Cross cancellation from an `onError`/`onComplete` method is not correctly verified by the Reactive Streams TCK. There is a [patch waiting](https://github.com/reactive-streams/reactive-streams-jvm/pull/483) to be included but it could take an arbitrary amount of time to get it in a released form.

In addition, the original workaround of using the ForkJoinPool sometimes doesn't work because the FJP executes the task on the caller thread, thus triggering the TCK bug again.

This PR adds a workaround in the form of a configurable ExecutorService which is set to a single-threaded one for the duration of the TCK.

### DIFF
--- a/microprofile/reactive-streams/src/main/java/io/helidon/microprofile/reactive/HelidonReactiveStreamsEngine.java
+++ b/microprofile/reactive-streams/src/main/java/io/helidon/microprofile/reactive/HelidonReactiveStreamsEngine.java
@@ -20,6 +20,7 @@ package io.helidon.microprofile.reactive;
 import java.util.Iterator;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Flow;
 import java.util.concurrent.ForkJoinPool;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


microprofile/reactive-streams/src/main/java/io/helidon/microprofile/reactive/HelidonReactiveStreamsEngine.java
```diff
@@ -526,18 +527,32 @@ static Object build(Iterable<Stage> graph, Mode mode) throws UnsupportedStageExc
...
     static void fail(CompletableFuture<Object> cf, Throwable ex) {
-        ForkJoinPool.commonPool().submit(() -> {
+        coupledExecutor.execute(() -> {
             cf.completeExceptionally(ex);
-            return null;
         });
     }
+
+    // Workaround for a TCK bug when calling cancel() from any method named onComplete().
+    private static volatile ExecutorService coupledExecutor = ForkJoinPool.commonPool();
```
danielkec: For future reference TCK fix: https://github.com/reactive-streams/reactive-streams-jvm/pull/483

PR #1572, opened 2020-03-24
Reactive cleanup

danielkec:


common/reactive/src/main/java/io/helidon/common/reactive/Flows.java
```diff
@@ -1,66 +0,0 @@
...
- * limitations under the License.
- */
-package io.helidon.common.reactive;
-
-import java.util.concurrent.Flow;
-
-/**
- * Utilities for Flow API.
- */
-public final class Flows {
```
danielkec: @tomas-langer is it ok to remove this?

tomas-langer: Yes, if we have an alternative that does the same

akarnokd:
Since you reference the PR from the changelog, could you put in the description about what to use instead?

### Removal of processor-like operators

The formerly public `Flow.Processor` implementations performing common operations have been removed. Users should use the respective operators from `Single` and `Multi` instead:

```java
// before
// ------
Flow.Publisher<Integer> source = ...
MappingProcessor<Integer, String> mapper = new MappingProcessor<>(Integer::toString);
source.subscribe(mapper);
mapper.subscribe(subscriber);

// after
// -----
Flow.Publisher<Integer> source = ...

Multi.from(source)
     .map(Integer::toString)
     .subscribe(subscriber)
```

### Removal of `Flows`

The class was providing basic `Flow.Publisher` implementations (which were implemented incorrectly anyway). Users should pick one of the static methods of `Single` or `Multi` instead, which provide the additional benefits of having fluent operators available to them for assembling reactive flows conveniently:

```java
// before
// ------
Flow.Publisher<Integer> just = Flows.singletonPublisher(1);
Flow.Publisher<Object> empty = Flows.emptyPublisher();

// after
// -----
Multi<Integer> just1 = Multi.singleton(1);
Single<Integer> just2 = Single.just(1);

Multi<Object> empty1 = Multi.empty();
Single<Object> empty2 = Single.empty();
```
```

danielkec:
@akarnokd great thx! done

PR #1600, opened 2020-03-30
Dependency convergence + jakarta libraries.

tomas-langer:

Switched to `jakarta` group id for Jakarta EE standards (keeping `javax` package).
Fixed all related issues with imports, exclusions etc.
Added dependency convergence plugin to Helidon project.
Added sanity checks for MP - both classpath and module path to `build.sh`

Also resolved problems with MP JPMS - quickstart should now work with JPMS without issues.

Marking as API change, as the libraries we depend on have changed.

### DIFF
--- a/grpc/client/src/test/java/io/helidon/grpc/client/GrpcChannelsProviderIT.java
+++ b/grpc/client/src/test/java/io/helidon/grpc/client/GrpcChannelsProviderIT.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,7 +37,6 @@ import io.grpc.NameResolver;
 import io.grpc.StatusRuntimeException;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import services.TreeMapService;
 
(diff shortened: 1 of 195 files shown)


tomas-langer:
Merged latest master (conflict)

grpc/client/src/test/java/io/helidon/grpc/client/GrpcChannelsProviderIT.java
```diff
@@ -37,13 +37,15 @@
...
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import services.TreeMapService;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+@Disabled("Certificate expired on 27th March 2020")
```
romain-grecourt, line 48: The certificate has been updated right? If so this change should not be needed.

pom.xml
```diff
@@ -747,6 +766,22 @@
                 <artifactId>reactive-streams-tck-flow</artifactId>
                 <version>${version.lib.reactivestreams}</version>
             </dependency>
+            <dependency>
```
romain-grecourt, line 769: These dependency management for 4th party dependencies, they belong in helidon-dependencies

tomas-langer: These are only internal dependencies (`reactive-streams-tck-flow`, `wagon-provider-api`, `hamcrest-library`, `weld-se-core`)
I will move the others to `dependencies`

romain-grecourt: just to make sure we are on the same page: anything that manages the version of a dependency that may end-up under target/libs for an helidon app needs to be defined in helidon-dependencies.

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
```diff
@@ -610,6 +610,13 @@ private InputStream locateStream(String uri) throws IOException {
             InputStream is;
 
             URL url = Thread.currentThread().getContextClassLoader().getResource(uri);
+            if (url == null) {
```
romain-grecourt, line 613: Why is this change needed for this PR ?
Should this go in a separate PR ?

tomas-langer: Because this failed as part of the `module-path` based tests, and I needed to fix it. The build does not pass without it.

PR #1613, opened 2020-04-03
Restart support for CDI regardless of mode used.

### LINKED ISSUE #1554: CDI restart cause issues
When we start CDI using one approach, shut it down and start again using a different approach, we get an exception.

The following:
```
        SeContainer container = SeContainerInitializer.newInstance()
                .disableDiscovery()
                .addBeanClasses(BmProducers.class)
                .initialize();

        container.close();

        Server.create().start();
```
causes
```
Exception in thread "main" java.lang.IllegalStateException: Unable to access CDI
	at javax.enterprise.inject.spi.CDI.lambda$getCDIProvider$1(CDI.java:88)
	at java.base/java.util.Optional.orElseThrow(Optional.java:408)
	at javax.enterprise.inject.spi.CDI.getCDIProvider(CDI.java:88)
	at javax.enterprise.inject.spi.CDI.current(CDI.java:64)
	at io.helidon.microprofile.server.Server$Builder.doBuild(Server.java:203)
	at io.helidon.common.context.Contexts.runInContext(Contexts.java:118)
	at io.helidon.microprofile.server.Server$Builder.build(Server.java:188)
	at io.helidon.microprofile.server.Server.create(Server.java:81)
```

The following:
```
SeContainer container = SeContainerInitializer.newInstance()
                .disableDiscovery()
                .addBeanClasses(BmProducers.class)
                .initialize();

        ((SeContainer)CDI.current()).close();


        Server.create().start();
```

Causes
```
Exception in thread "main" java.lang.NullPointerException
	at io.helidon.microprofile.server.Server$Builder.build(Server.java:166)
	at io.helidon.microprofile.server.Server.create(Server.java:81)
```

tomas-langer:
Resolves #1554

### DIFF
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ContainerInstanceHolder.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ContainerInstanceHolder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,6 +30,7 @@ import java.util.concurrent.atomic.AtomicReference;
 final class ContainerInstanceHolder {
     private static final AtomicReference<HelidonContainer> CONTAINER = new AtomicReference<>();
     private static final List<Runnable> RESET_LISTENERS = new LinkedList<>();
+    private static boolean isReset = false;
 
     private ContainerInstanceHolder() {
     }
@@ -38,7 +39,12 @@ final class ContainerInstanceHolder {
         CONTAINER.set(container);
     }
 
-    static HelidonContainer get() {
+    // return true if the container was reset, indicating somebody started CDI by hand and then shut it down
+    static synchronized boolean isReset() {
+        return isReset;
+    }
+
+    static synchronized HelidonContainer get() {
         HelidonContainer helidonContainer = CONTAINER.get();
         if (null == helidonContainer) {
             helidonContainer = fromBuildTimeInitializer();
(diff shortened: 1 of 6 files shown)


microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ContainerInstanceHolder.java
```diff
@@ -38,7 +39,12 @@ static void set(HelidonContainer container) {
         CONTAINER.set(container);
     }
 
-    static HelidonContainer get() {
+    // return true if the container was reset, indicating somebody started CDI by hand and then shut it down
+    static synchronized boolean isReset() {
+        return isReset;
+    }
```
danielkec, line 45: Is synchronized needed here?

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/Main.java
```diff
@@ -58,13 +60,21 @@ private Main() {
...
-        CONTAINER.start();
+        if (ContainerInstanceHolder.isReset()) {
+            // in case somebody restarted the container, we need to get a new one
+            inUse = ContainerInstanceHolder.get();
+        } else {
+            // use the statically initialized one
+            inUse = CONTAINER;
+        }
+
+        inUse.start();
```
danielkec, line 71: I don't see the whole picture, but can someone call `reset()` on container holder between the `isReset()` check and `start()`?

tomas-langer: No. You cannot run two containers in parallel. You can run them sequentially. And this is to make sure that if you run a container, shut it down and then run the `Main` class, everything works fine (the `CONTAINER` field is there for `native-image` static initialization)

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ContainerInstanceHolder.java
tomas-langer: Yes, as the field is not atomic. Either you must use `synchronized` from all places that use the field, have it declared `volatile` or use `AtomicBoolean`. As I already used `synchronized` in the other methods and this class is used once during startup and once during shutdown, the overhead is negligible

PR #1615, opened 2020-04-03
Feature design

tomas-langer:
Design of features.
I have moved this from gist as I wanted to split the design into the example files.


docs-internal/features/features.yaml
```diff
@@ -0,0 +1,87 @@
...
+              - name: "Validators"
+                # each feature implementing this service is valid here
+                service: "io.helidon.security.providers.abac.spi.AbacValidatorService"
+  - name: "Tracing"
+    implementations:
+      service: "io.helidon.tracing.spi.TracerProvider"
+  - name: "DbClient"
+    drivers:
+      - name: "DbClient Type"
+        service: "io.helidon.dbclient.spi.DbClientProvider"
```
romain-grecourt, line 30: What is service indented for ?

docs-internal/features/features.yaml
```diff
@@ -0,0 +1,87 @@
...
+        # explicitly list JDBC, as it requires sub-drivers (mongo is loaded through service loader)
+        list:
+          - name: "JDBC"
+            drivers:
+              type: "external"
+              service: "java.sql.Driver"
+              # maybe better for external:
+              category: "JDBC Driver"
```
romain-grecourt, line 38: I think additional metadata such as category and description can also be mandated for every feature. Maybe that's part of the annotation.

docs-internal/features/features.yaml
```diff
@@ -0,0 +1,87 @@
...
+        service: "io.helidon.config.spi.ConfigParser"
+
+
+integrations:
+  # name is the feature path (may have more than one integration for same feature)
+  - group:
+    name: "Security"
+    integrations:
+      - name: "Security/Integration/WebServer"
+        when: ["Security","WebServer"]
```
romain-grecourt, line 64: "Security" is not a features, it's some sort of path prefix right ? I guess It means all features with a path that starts with "Security".

Why not use a wildcard expression here, e.g. `security/*`

Also since we have a path, we should have some rules for the segments, e.g. no space etc. Right now we are using camel case for the paths which feels wrong. We should perhaps define an ID which composes the path, a name which is human readable (e.g. can be camel cased, can have spaces etc).

The ID is meant to be used for referring to features, it's defined in paths like here, it's declared in the annotation, and might also be referenced in the archetype stuff.

docs-internal/features/package-info.java
```diff
@@ -0,0 +1,18 @@
+@Feature(value = "Security/Providers/ABAC/Validators/EL",
```
romain-grecourt, line 1: During the meeting we had this done on a module-info. Here it's done on a package-info, why ? Is it because this annotation is retained at runtime and we need a way to query the annotations regardless if the module-path is used or not ?

What's missing in this document is an example of how the features are queried for the startup list scenario. Also, for the `helidon features --list` this needs to be done statically without executing actual code, for this scenario it's easier to scan for resources files in the JAR files.

How about making this a compile time only annotation and generate a resource file during compilation with an annotation processor.

Also, I think the feature resource file should not be in a structured format that requires a library to be parsed. We need to be able to read such file without any additional library ; I guess a properties file is AOK.

docs-internal/features/package-info.java
```diff
@@ -0,0 +1,18 @@
...
+        description = "Expression language validator for ABAC security provider.");
+@HelidonFlavor("SE", "MP")
+package io.helidon.security.abac.el {
+
+}
+
+@Features(
+    @Feature(value = "Security/Providers/Digest-Auth",
```
romain-grecourt, line 9: The "path" is declared here and in the catalog as a YAML tree.
This feels like it's needed for the startup list since it would be an over-head to read the catalog just to render the tree structure.

I'd like to see if we can get rid of the path and instead have each feature be defined with a unique id and the feature tree be defined only in the catalog.

This would require the startup list to read the catalog to render the feature tree.

If the startup list is optional and a feature itself (i.e a separate modules, a drop-in JAR for MP, and a mini API for the main class in SE), then the feature-startup-list module would depend on the catalog and would have access to it (perhaps it bundles it).

However, reading the catalog would add a bit of overhead, especially if it's JSON. I guess we can optimize this usecase, maybe with code generation based on the YAML, or by generating a file with an optimized format.

docs-internal/features/feature-db-se.json
```diff
@@ -0,0 +1,99 @@
+{
```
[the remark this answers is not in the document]
tomas-langer: Effective computed catalog JSON for a single flavor (could be a single file for both flavors of course)

docs-internal/features/features.md
```diff
@@ -0,0 +1,289 @@
+# Features
+Helidon feature serve two main purposes:
+
+1. Startup list (and possibly tree) to show which features are used
+2. Mechanism for CLI/starter to correctly add features to a new/existing project
+3. Configuration metadata (to create configuration templates, support IDE etc.)
```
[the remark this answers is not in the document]
tomas-langer: This was to map the configuration update/file to be added to the project when you add a new feature.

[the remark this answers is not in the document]
romain-grecourt: Ok, I don't think we should include this in the feature model/metadata and instead have this logic in the archetypes.

docs-internal/features/features.yaml
tomas-langer: Service is a java service loader service

PR #1620, opened 2020-04-06
Fix for prioritized custom MP config sources in Helidon config.

tomas-langer:

Fixes #1615

### DIFF
--- a/microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfigBuilder.java
+++ b/microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfigBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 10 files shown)


microprofile/config/config/src/main/java/io/helidon/microprofile/config/MpConfigBuilder.java
```diff
@@ -376,8 +391,9 @@ public String description() {
             return configSource.changes();
         }
 
-        int getOrdinal() {
-            return ordinal;
+        @Override
+        public int priority() {
+            return 100 - ordinal;
```
Verdent, line 396: I am kind of confused why is it 100 - ordinal?

tomas-langer: This is an aribtrary constant used to get from ordinals to priority.
Sources with ordinals will in most cases (unless negative priority is used) win over sources with Priority.
This is undefined in any specification, as MP uses ordinals, and Helidon uses Priority.
Mixing Helidon config sources with MP config sources is not recommended and not documented as an option.

PR #1624, opened 2020-04-08
Changed the default context root for Websocket endpoints

spericas:
Changed the default context root for Websocket endpoints to be "/", just like for REST. Improved testing of endpoints with and without using the default context root. One additional test that mixes REST resources and Websocket endpoints in same application.

### DIFF
--- a/microprofile/websocket/src/test/java/io/helidon/microprofile/tyrus/WebSocketBaseTest.java
+++ b/microprofile/websocket/src/test/java/io/helidon/microprofile/tyrus/WebSocketBaseTest.java
@@ -21,6 +21,7 @@ import java.net.URI;
 import java.util.logging.Logger;
 
 import javax.enterprise.inject.se.SeContainer;
+import javax.enterprise.inject.spi.CDI;
 import javax.websocket.CloseReason;
 import javax.websocket.Endpoint;
 import javax.websocket.EndpointConfig;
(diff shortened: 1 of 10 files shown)


microprofile/websocket/src/test/java/io/helidon/microprofile/tyrus/WebSocketBaseTest.java
```diff
@@ -26,71 +30,33 @@
...
-        EchoClient echoClient = new EchoClient(echoUri);
-        echoClient.echo("hi", "how are you?");
-    }
+    public abstract String context();
 
     @Test
-    public void testEchoProg() throws Exception {
-        URI echoUri = URI.create("ws://localhost:" + server.port() + "/web/echoProg");
+    public void testEchoAnnot() throws Exception {
+        URI echoUri = URI.create("ws://localhost:" + DEFAULT_PORT + context() + "/echoAnnot");
```
tjquinno, line 55: I know there have been discussions about avoiding fixed ports because of conflicts on shared systems or parallel tests. Is this opening us up to that problem?

spericas: Possibly. I had to switch the startup code to use CDI so I can register beans directly and I don't have access to the server port anymore. The old test was always starting on the default port anyway. So it shouldn't be worse than before, but it does not mean is great.

Issue #1391, opened 2020-02-14
Simplify adding *-image profile configuration


batsatt:
Given that the profile declarations are in the `helidon-se` or `helidon-mp` parent poms, adding configuration in the application pom is either verbose or somewhat ugly.

The ugly solution is to use global properties with fully qualified names, e.g.
``` 
<native.image.buildShared>false<\native.image.buildShared>
```

The more natural solution is to use a configuration element:
```
<configuration>
    <buildShared>false<\buildShared>
<\configuration>
```
The only way I know to add that configuration and retain the profile behavior is by duplicating the parent `<profile>` declaration. Hopefully there is a less verbose way to do this.

romain-grecourt:
If you have a profile like this:
```xml
    <profiles>
        <profile>
            <id>native-image</id>
            <build>
                <plugins>
                    <plugin>
                        <groupId>io.helidon.build-tools</groupId>
                        <artifactId>helidon-maven-plugin</artifactId>
                        <executions>
                            <execution>
                                <id>native-image</id>
                                <goals>
                                    <goal>native-image</goal>
                                </goals>
                            </execution>
                        </executions>
                    </plugin>
                </plugins>
            </build>
            <dependencies>
                <dependency>
                    <groupId>io.helidon.integrations.graal</groupId>
                    <artifactId>helidon-graal-native-image-extension</artifactId>
                </dependency>
            </dependencies>
        </profile>
    </profiles>
```

You can override the configuration per execution like this:
```xml
    <build>
        <pluginManagement>
            <plugins>
                <plugin>
                    <groupId>io.helidon.build-tools</groupId>
                    <artifactId>helidon-maven-plugin</artifactId>
                    <executions>
                        <execution>
                            <id>native-image</id>
                            <configuration>
                                <finalName>foo</finalName>
                            </configuration>
                        </execution>
                    </executions>
                </plugin>
            </plugins>
        </pluginManagement>
    </build>
```

batsatt:
I see; yes, that works. Unfortunately, though, IntelliJ doesn't understand that idiom enough to provide any completions for the supported configuration options. I find those so useful that in my local project I would just copy the `<profile>`.

batsatt:
So with the current version of IntelliJ, even adding the goal doesn't cause completions to work. This is really frustrating, as it is vastly easier to add configuration when the IDE helps!

spericas:
[...]

PR #1639, opened 2020-04-14
Docs: Add custom runtime image guide, maven guide, gradle guide, refactored native image guide. 

barchetta:
1. Add a new guide for custom runtime image creation (MP and SE)
2. Add new Maven and Gradle build guides. A little sparse for now, but it's a start.
3. Refactor native image guide to document both SE and MP
4. Shares content between MP and SE document by putting common content under `common` and using include and properties to generate edition specific document.
5. Fixed some typos in WebServer javadocs.


docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,207 @@
...
+// :docker-multi-stage-url: https://docs.docker.com/develop/develop-images/multistage-build/
+
+This guide describes how to build a custom runtime image for your Helidon application
+using Helidon's support for the JDK's `jlink` tool.
+
+== Introduction
+
+JDK 9 introduced the https://docs.oracle.com/en/java/javase/11/tools/jlink.html[`jlink`]
+command that supports assembling a set of modules and their dependencies into a custom
+runtime image. The `helidon-mave-plugin` has support for easily creating a custom runtime image for your
```
batsatt, line 35: typo: mave -> maven

docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,207 @@
...
+Helidon application resulting in a smaller, better performing runtime.
+
+In this guide you will learn how to build a custom runtime image locally on your machine,
+as well as using Docker.
```
batsatt, line 39: maybe "as well as how to build it in a Docker image" ?

docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,207 @@
...
+|===
+|About 10 minutes
+| <<about/03_prerequisites.adoc,Helidon Prerequisites>>
+|===
+
+== Verify Java 11 or Newer
+
+[source,bash]
+----
+$JAVA_HOME/bin/jlink --version
```
batsatt, line 52: This might give the impression that we're going to use jlink directly; how about just `java -version`.

docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,207 @@
...
+Aslo included in the custom image is a Class Data Sharing (CDS) archive that
+improves your application's startup performance and in-memory footprint.
+You can learn more about Class Data Sharing in the
+https://docs.oracle.com/en/java/javase/11/vm/class-data-sharing.html[JDK documentation].
+
+The CDS archive increases your image size to get these performance optimizations.
+It can be of significant size (tens of MB). The size of the CDS archive is
+reported at the end of the build output.
+
+If you rather have a smaller image size (with a slightly increased startup time) you
```
batsatt, line 144: typo: you -> you'd

docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,207 @@
...
+will be re-downloaded.
+
+Start the application:
+
+[source,bash,subs="attributes+"]
+----
+docker run --rm -p 8080:8080 helidon-quickstart-{helidon-lc-edition}-jlink:latest
+----
+
+You can exercise the application's endpoints as before.
```
batsatt, line 180: I think it would also be worth showing passing script arguments, e.g. `--test` or `--help` just so it is clear how it is done, (and maybe that the port mappings isn't required).

docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,207 @@
...
+
+== How small is this?
+
+Let's take a look at the Docker image for a custom runtime that does
+not include the CDS archive:
```
batsatt, line 185: Maybe clarify that this isn't the same image just built above?

docs/common/guides/jlink_image.adoc
barchetta: fixed

docs/common/guides/jlink_image.adoc
barchetta: fixed

docs/common/guides/jlink_image.adoc
barchetta: fixed

docs/common/guides/jlink_image.adoc
barchetta: I've removed the size/timing info from both the graalvm guide and custom image guide since they won't age well. I'm thinking that information is best left for blogs.

docs/common/guides/jlink_image.adoc
barchetta: fixed

barchetta:
> We have an open issue (#1391) about the right way to add configuration to the pom for these plugins, say if you want to disable CDS or whatever. Yes, properties work, but they're a bit of a pain. And just pointing to the plugin docs is confusing, since they talk about adding the profile, which is already present now.
> 
> It would be best if we could resolve that issue and then show how to do it here.

I've add (an admittedly sparse) Maven Build Guide that shows examples of configuring the plugin and crossed referenced in the jlink guide.

docs/common/guides/jlink_image.adoc
```diff
@@ -0,0 +1,188 @@
...
+
+If you'd rather have a smaller image size (with a slightly increased startup time) you
+can skip the creation of the CDS archive by executing your build like this:
+
+[source,bash]
+----
+mvn package -Pjlink-image -Djlink.image.addClassDataSharingArchive=false
+----
+
+For more information see the
```
batsatt, line 151: Since those docs also say to add a profile, which would be confusing here since it is already present, maybe say "For more information on available configuration options see the"

barchetta: fixed

PR #1640, opened 2020-04-14
Update CONTRIBUTING.md

makbn:
+ slack join link changed: previous slack link was not valid anymore


romain-grecourt:
We actually have a DNS redirection in place at slack.helidon.io ; but forgot to update the CONTRIBUTING.md. Thanks for pointing this out, I will update the file in a separate PR.

PR #1659, opened 2020-04-17
WebClient Intro doc updated with changes

vanajaguru:
I have updated the examples for JSON Processing Media Support. Please review.


Verdent:
This is no longer valid. I have made some changes to this structure and these methods are no longer there.
```
Adding readers and writers to the request helps to handle the entity of the response and request.
register(MessageBodyReader<?> reader): Registers specific response entity reader.
register(MessageBodyWriter<?> writer): Registers specific response entity writer.
```
I might probably remove this part completely. I will provide you new piece of code to the chapter `Adding JSON Processing Media Support to the WebClient` about setting specific reader/writer to the request. I will do that in 2 days at most (I hope), because we are still discussing how it will look like.

Other parts should be correct from the code perspective.

Verdent:
Followup issue #1862

PR #1664, opened 2020-04-20
Multi await feature for intentional blocking

danielkec:
There was a request for blocking feature for `Multi` and `Single` 
```java
Multi.interval(20, TimeUnit.MILLISECONDS, Executors.newSingleThreadScheduledExecutor())
        .limit(5)
        .forEach(sum::addAndGet)
        .await();
```
```java
Multi.interval(20, TimeUnit.MILLISECONDS, Executors.newSingleThreadScheduledExecutor())
        .limit(5)
        .reduce(Long::sum)
        .await();
```

I guess this needs some broader discussion, it kind of feels like introducing an antipattern. But the truth is other impls has similar feature and users will be expecting it.


common/reactive/src/main/java/io/helidon/common/reactive/MultiCompletionStage.java
```diff
@@ -0,0 +1,66 @@
...
+     */
+    default void await(long timeout, TimeUnit unit) {
+        try {
+            this.toCompletableFuture().get(timeout, unit);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new CompletionException(e);
+        }
+    }
+
+    class MultiCompletableFuture extends CompletableFuture<Void> implements MultiCompletionStage {
```
akarnokd, line 68: `final`

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
```diff
@@ -843,9 +843,13 @@
      *
      * @param consumer consumer to be invoked for each item
      */
-    default void forEach(Consumer<? super T> consumer) {
-        FunctionalSubscriber<T> subscriber = new FunctionalSubscriber<>(consumer, null, null, null);
+    default MultiCompletionStage forEach(Consumer<? super T> consumer) {
```
akarnokd, line 847: I'd definitely add cancellation support to this. `AutoCloseable` perhaps.

common/reactive/src/main/java/io/helidon/common/reactive/MultiCompletionStage.java
```diff
@@ -0,0 +1,70 @@
...
+     * @param unit    the time unit of the timeout argument
+     * @throws java.util.concurrent.CancellationException if this future was cancelled
+     * @throws CompletionException                        if this future completed exceptionally,
+     *                                                    was interrupted while waiting or the wait timed out
+     */
+    default void await(long timeout, TimeUnit unit) {
+        try {
+            this.toCompletableFuture().get(timeout, unit);
+        } catch (InterruptedException | ExecutionException | TimeoutException e) {
+            throw new CompletionException(e);
```
akarnokd, line 64: I'd unwrap `ExecutionException` here.

common/reactive/src/main/java/io/helidon/common/reactive/MultiCompletionStage.java
danielkec: fixed

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
danielkec: That's pretty cool! :+1:

common/reactive/src/main/java/io/helidon/common/reactive/MultiCompletionStage.java
danielkec: thx

romain-grecourt:
Can you rebase against master ? you're missing one of the checks from the new infra.

danielkec:
@tomas-langer had proposed `Single` implementing `CompletionStage` which is seems to be quite nicer, so `Multi.forEach` can return`Single<Void>`, `await` methods are part of single on one place only and its possible to return Single on the apis where completables are expected

danielkec:
Can I ask @romain-grecourt and @tomas-langer for review? @Verdent is blocked by this with #1646 Question is if `Single implementing CompletionStage` is preferable or not, Imho I like it as it makes all apis returning Single more comfortable.

romain-grecourt:
@danielkec Yeah, Let's go ahead with Single implementing CompletionStage, we can leverage it in our APIs where we return a CompletionStage.

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
```diff
@@ -842,10 +855,23 @@
      * Terminal stage, invokes provided consumer for every item in the stream.
      *
      * @param consumer consumer to be invoked for each item
-     */
-    default void forEach(Consumer<? super T> consumer) {
-        FunctionalSubscriber<T> subscriber = new FunctionalSubscriber<>(consumer, null, null, null);
+     * @return MultiCompletionStage
```
tomas-langer, line 858: Wrong javadoc for return type

tomas-langer: Also the type is clearly visible from Javadoc already, this should describe behavior, not type

danielkec:
Awaits are now preserved when chaining `CompletionStage` methods:
```java
        testMulti()
                .forEach(sum::addAndGet)
                .whenComplete((aVoid, throwable) -> {})
                .thenRun(() -> {})
                .thenAccept(aVoid -> {})
                .await();
```

olotenko:
This seems like a massive breaking change just to get a wrapper for a timed `CompletableFuture.get`.

`onCancel` implementation is incompatible in its behaviour with superclass. `Multi.forEach` can miss cancellation indefinitely. Dangling `Subscribers` are possible.

danielkec:
@olotenko pls elaborate, is it order of?:
```java
                    subscription.request(Long.MAX_VALUE);
                    single.onCancel(subscription::cancel);
```

olotenko:
Yes, but don't you see the whole design of the change is rickety?

A plain static method for two flavours of `await` would serve the purpose. I really don't see the value of rewriting `CompletableFuture` and changing how all the reactive pipelines everywhere behave.

PR #1714, opened 2020-05-01
Fix CorsSupport default behavior

### LINKED ISSUE #1713: CORS default CorsSupport instance does not have expected wildcard behavior
## Environment Details
* Helidon Version: 2.x
* Helidon SE 
* JDK version: n/a
* OS: n/a
* Docker version (if applicable):n/a

----------

## Problem Description
Using `CorsSupport.create()` should give wildcard behavior, just as `CrossOriginConfig.builder().build()` does. But, alas, it gives no CORS behavior at all.

## Steps to reproduce
Use
```
Routing.Builder builder = Routing.builder()
    .any(CorsSupport.create())
    .options("/greet", (req, resp) -> resp.send);
```
Then send a CORS-triggering `OPTIONS` request:
```
curl -i -X OPTIONS -H "Origin: http://foo.com" -H "Host: bar.com" 
```
The response will not contain any `Access-Control-xxx` headers but it should.

tjquinno:
Resolves #1713 

There are two aspects to this.

A. If the developer creates a `CorsSupport` instance it will provide the default (wild carded) behavior if no specific settings were provided. In particular, passing a `missing` config node to `CorsSupport.create(Config)` creates a new `CorsSupport` instance with the default wildcarding behavior; CORS also logs an `INFO` message to this effect in case the developer actually expected the config node to contain something.

There is a new `CrossOriginConfig.create()` method which developers can use to include the default wild carded behavior as they build a `CorsSupport` instance.

(I refactored many methods from `Aggregator` to its new `Builder` inner class so the diffs look deceptively large.)

B. All cross-origin info added to a `CorsSupport` instance is now searched sequentially in order of addition to the builder. The `Aggregator` which collects the cross-origin information (basically as `CrossOriginConfig` instances) used to use a map, keyed by the path expression. This could lead to confusion if the developer provided multiple `CrossOriginConfig` objects with the same path (or with no path which is essentially the same thing) because later additions would overwrite earlier ones. 

This PR changes the internal data structure to a `List` and maintains the order in which the developer supplies the `CrossOriginConfig` information. When requests arrive, the Helidon CORS implementation scans the list in order-of-addition and uses the first that matches the request's path and HTTP method.

This gives the developer complete control by making sure the order in which the application adds `CrossOriginConfig`s matches the needs of the application, without the Helidon CORS code imposing a "one-setting-per-path" restriction.


microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsSupportMp.java
```diff
@@ -112,7 +112,7 @@ protected Builder secondaryLookupSupplier(
 
         @Override
         public String path() {
-            return requestContext.getUriInfo().getPath();
+            return "/" + requestContext.getUriInfo().getPath();
```
tomas-langer, line 115: Could not find in docs whether there is a possibility we would get the leading `/` in response. In such a case I would probably try to guard against it, e.g. 
```java
String path = requestContext.getUriInfo().getPath();
if (path.startsWith("/")) {
            return path;
        } else {
            return '/' + path;
        }
```

webserver/cors/src/main/java/io/helidon/webserver/cors/CorsSupportHelper.java
```diff
@@ -510,26 +524,16 @@ R processCorsPreFlightRequest(CrossOriginConfig crossOrigin,
...
-                    METHOD_NOT_IN_ALLOWED_LIST,
-                    () -> "header " + ACCESS_CONTROL_REQUEST_METHOD + " absent from request");
-        }
-
         // Check if method is allowed
-        String method = methodOpt.get();
         List<String> allowedMethods = Arrays.asList(crossOrigin.allowMethods());
         if (!allowedMethods.contains("*")
-                && !contains(method, allowedMethods, String::equals)) {
+                && !contains(requestedMethod, allowedMethods, String::equals)) {
```
tomas-langer, line 530: HTTP methods should be compared ignore case (unless you fix the requestedMethod and configured methods somewhere to be always upper/lower cased)

webserver/cors/src/main/java/io/helidon/webserver/cors/CrossOriginConfig.java
```diff
@@ -218,6 +228,21 @@ public long maxAgeSeconds() {
...
 
+    /**
+     * Reports whether the specified HTTP method name matches this {@code CrossOriginConfig}.
+     *
+     * @param method HTTP method name to check
+     * @return true if this {@code CrossOriginConfig} matches the specified method; false otherwise
+     */
+    public boolean matches(String method) {
+        for (String allowMethod : allowMethods) {
+            if (allowMethod.equals(method) || allowMethod.equals("*")) {
```
tomas-langer, line 239: HTTP methods should be compared ignore case (unless you fix the requestedMethod and configured methods somewhere to be always upper/lower cased)

PR #1729, opened 2020-05-05
Docs helidon se intro 1626

ljamen:
New intro topic for the now split mp/se sections. 
SE/Introduction/01_introduction.adoc


docs/se/introduction/01_introduction.adoc
```diff
@@ -38,37 +38,58 @@ of Netty and uses a straight forward request routing API and supports a
 number of additional Helidon features:
```
m0mus, line 38: It makes sense to add links to the corresponding docs for each feature in the table.

PR #1732, opened 2020-05-06
OutputStreamPublisher now handles close on its own.

tomas-langer:

This was discovered by Jersey team - when using `OutputStreamPublisher`, there was an infinite blocking problem with method `close`.

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamPublisher.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamPublisher.java
@@ -21,6 +21,7 @@ import java.io.OutputStream;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.Objects;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Flow;
@@ -59,6 +60,9 @@ public class OutputStreamPublisher extends OutputStream implements Flow.Publishe
                 @Override
                 public void cancel() {
                     subscriber.cancel();
+                    // when write is called, an exception is thrown as it is a cancelled subscriber
+                    // when close is called, we do not throw an exception, as that should be silent
+                    completionResult.complete(null);
                 }
             });
         }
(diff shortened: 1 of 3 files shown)


tomas-langer:
Fixed the cases identified based on @olotenko input. Created a follow up #1741 to fix related issues (and classes).

common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamPublisher.java
```diff
@@ -163,6 +195,7 @@ private void complete(Throwable t) {
         subscriber.close(sub -> {
             synchronized (invocationLock) {
                 sub.onError(t);
```
danielkec, line 197: There is a race condition against `requested.increment(n, t -> complete(t));` on line 57(not caused by this PR), can be solved by removing synchronization logic from this publisher and wrapping subscriber in `io.helidon.common.reactive.SequentialSubscriber`

danielkec, line 197: Lock doesn't help to prevent rule [§1.7](https://github.com/reactive-streams/reactive-streams-jvm#1.7) violation, we can remove synchronization logic from publisher and wrap subscriber in `io.helidon.common.reactive.SequentialSubscriber`

danielkec: Not a blocker for this PR, just noticing

olotenko: 1.7 violation can be seen to not be reachable in this instance.

danielkec: I'm thinking about not very probable use-case of request(-1) vs. exception in publish method

olotenko: You can see `subscriber.close` ensures completion signals will be signalled once and only once, as long as they always go through `complete` which calls `subscriber.close`.

danielkec: Missed that, thanks for clarification

olotenko: Pro tip: when designing routines with CPS, ensure a continuation is invoked _always_. (Not necessarily the same continuation for all outcomes, not necessarily immediately, but you've got to design routines with a guarantee that the continuation is invoked always) You want to not call any continuation only in very rare circumstances.

The problem with the current design of `SingleSubscriberHolder` is that it can omit calling the continuation, and the caller is not told that that is the case. It is the same behaviour as non-termination. The outcome is that `signalCloseComplete` is not guaranteed to be reachable, and the `completionResult.get` can wait indefinitely. The (rather long) timeout then only eventually reclaims the thread.

PR #1742, opened 2020-05-06
1.4: use release flag for compiler, not source/target

barchetta:
JDK 9 added the --release flag to the compiler which is the proper way to target a Java release. This will prevent us from accidentally introducing dependencies on system libraries in Java 9+.

With this change you can no longer build the helidon-1.x branch using Java 8.

For reference:

https://openjdk.java.net/jeps/247

https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -187,7 +187,7 @@
                     <artifactId>maven-compiler-plugin</artifactId>
                     <version>${version.plugin.compiler}</version>
                     <configuration>
-                        <release>1.8</release>
+                        <release>8</release>
                         <compilerArgs>
                             <arg>-Xlint:unchecked</arg>
                         </compilerArgs>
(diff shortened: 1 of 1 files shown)


tomas-langer:
I am actually using jdk8 to build Helidon.
Why do we want to drop support for building Helidon using java 8, when it is targeted for java 8?

barchetta:
Because when using just `target` and `source` you might not produce binaries that will run on JDK 8 -- even if you stick to JDK 8 APIs. For one example see: https://github.com/eclipse/jetty.project/issues/3244 (in particular stuart-marks comment).

barchetta:
We were doing this correctly already with the jdk9+ profile. Closing.

PR #1747, opened 2020-05-07
Cors MP example

### LINKED ISSUE #1726: CORS example for MP
This issue captures the work for creating a Helidon MP example.

tjquinno:
Resolves #1726

### DIFF
--- a/examples/microprofile/cors/src/main/java/io/helidon/microprofile/examples/cors/GreetResource.java
+++ b/examples/microprofile/cors/src/main/java/io/helidon/microprofile/examples/cors/GreetResource.java
@@ -152,13 +152,13 @@ public class GreetResource {
 
     @OPTIONS
     @CrossOrigin()
-    public void optionsForRetrievingUnnamedGreetings() {
+    public void optionsForRetrievingUnnamedGreeting() {
     }
 
     @OPTIONS
     @CrossOrigin()
     @Path("/{name}")
-    public void optionsForRetrievingNamedGreetings() {
+    public void optionsForRetrievingNamedGreeting() {
     }
 
     private JsonObject createResponse(String who) {
--- a/examples/microprofile/cors/src/main/java/io/helidon/microprofile/examples/cors/GreetResource.java
+++ b/examples/microprofile/cors/src/main/java/io/helidon/microprofile/examples/cors/GreetResource.java
@@ -73,7 +73,6 @@ public class GreetResource {
      *
      * @return {@link JsonObject}
      */
-    @SuppressWarnings("checkstyle:designforextension")
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     public JsonObject getDefaultMessage() {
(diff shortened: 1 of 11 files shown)


examples/microprofile/cors/src/main/java/io/helidon/microprofile/examples/cors/GreetResource.java
```diff
@@ -0,0 +1,156 @@
...
+    public GreetResource(GreetingProvider greetingConfig) {
+        this.greetingProvider = greetingConfig;
+    }
+
+    /**
+     * Return a worldly greeting message.
+     *
+     * @return {@link JsonObject}
+     */
+    @SuppressWarnings("checkstyle:designforextension")
```
tomas-langer, line 76: Please do not add these suppress warnings. 
If you use Idea, you can suppress warnings for methods/classes annotated with a specific annotation (such as `GET`, `PUT` etc.), so this is not marked as a problem. I hope other IDEs have similar way of getting around it.
Suppress warnings in source code should be used for compiler warnings (such as generics) and warnings that are significant (e.g. the warning is sound, but we have a good reason to ignore it).
This case is neither

examples/microprofile/cors/src/main/resources/META-INF/beans.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2018 Oracle and/or its affiliates. All rights reserved.
```
tomas-langer, line 4: The copyright is wrong.

examples/microprofile/cors/src/test/java/io/helidon/microprofile/examples/cors/TestCORS.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    @Order(1) // Make sure this runs before the greeting message changes so responses are deterministic.
+    @Test
+    public void testHelloWorld() throws Exception {
+
+        WebClientResponse r = getResponse("/greet");
+
+        assertEquals(200, r.status().code(), "HTTP response1");
```
tomas-langer, line 96: Please use Hamcrest assertions.

All Helidon tests should be written using Hamcrest (there are a few exceptions such as `fail` and `assertAll`) - see dev guidelines.

This should be
`assertThat("HTTP response1", r.status().code(), is(200));`
This is an overall problem in all the tests.

examples/microprofile/cors/src/main/java/io/helidon/microprofile/examples/cors/GreetResource.java
tjquinno: Removed. (These are present in other examples which I borrowed from. All should be cleaned up then.)

examples/microprofile/cors/src/main/resources/META-INF/beans.xml
tjquinno: Fixed

examples/microprofile/cors/src/test/java/io/helidon/microprofile/examples/cors/TestCORS.java
tjquinno: Done. (We should bring all examples into line with this so there are no misleading patterns around people might follow.)

PR #1760, opened 2020-05-09
Fixes StackOverflowError unearthed by new MicroProfile Config implementation

ljnelson:

### DIFF
--- a/integrations/cdi/oci-objectstorage-cdi/src/main/java/io/helidon/integrations/cdi/oci/objectstorage/OciConfigConfigSource.java
+++ b/integrations/cdi/oci-objectstorage-cdi/src/main/java/io/helidon/integrations/cdi/oci/objectstorage/OciConfigConfigSource.java
@@ -22,7 +22,7 @@ import java.util.Map;
 
 import com.oracle.bmc.auth.ConfigFileAuthenticationDetailsProvider;
 import org.eclipse.microprofile.config.Config;
-import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.config.spi.ConfigSource;
 
 /**
@@ -30,8 +30,6 @@ import org.eclipse.microprofile.config.spi.ConfigSource;
  * ConfigFileAuthenticationDetailsProvider}.
  */
 public final class OciConfigConfigSource implements ConfigSource {
-
-  private static final ThreadLocal<Map<String, String>> PROPERTIES = new ThreadLocal<>();
 
   private volatile Map<String, String> properties;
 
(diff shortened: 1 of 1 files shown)


integrations/cdi/oci-objectstorage-cdi/src/main/java/io/helidon/integrations/cdi/oci/objectstorage/OciConfigConfigSource.java
```diff
@@ -98,32 +100,45 @@ public int getOrdinal() {
...
         try {
-          if (configFilePath == null) {
-            temp = new ConfigFileAuthenticationDetailsProvider(profile);
-          } else {
-            temp = new ConfigFileAuthenticationDetailsProvider(configFilePath, profile);
+          // Reentrant call; i.e. this causes a new
+          // OciConfigConfigSource instance to be created and queried
+          // for various values.
+          final Config config = ConfigProvider.getConfig();
+          assert config != null;
```
tomas-langer, line 119: Please do not use `assert` in the code. I do not see an advantage checking for null not being returned by methods that claim to never return null (such as `Optional.orElse(T)`).
We are not using this approach anywhere in Helidon.
No need to remove existing `assert` statements from code, but please do not add new ones.

integrations/cdi/oci-objectstorage-cdi/src/main/java/io/helidon/integrations/cdi/oci/objectstorage/OciConfigConfigSource.java
```diff
@@ -98,32 +100,45 @@ public int getOrdinal() {
...
+        PROPERTIES.set(Collections.emptyMap());
         try {
-          if (configFilePath == null) {
-            temp = new ConfigFileAuthenticationDetailsProvider(profile);
-          } else {
-            temp = new ConfigFileAuthenticationDetailsProvider(configFilePath, profile);
+          // Reentrant call; i.e. this causes a new
+          // OciConfigConfigSource instance to be created and queried
+          // for various values.
+          final Config config = ConfigProvider.getConfig();
```
tomas-langer, line 118: This approach is broken by design.
Consider a user creating a custom configuration using a builder and registering config sources, including this one.
If at this point, you attempt to use `ConfigProvider.getConfig()`, another instance of config is created (using just the defaults of MicroProfile, with only `microprofile-config.properties`, system properties and environment variables)
This config may not contain any configuration value you expect to find there, because it is not provided at all (or may provide configuration built-into the application only valid for development environment).
As MP does not have the concept of "meta-configuration", you have the following options:
1. provide a builder approach to create an instance of this config source "by hand"
2. use a config-independent way of obtaining configuration of this config source - such as system properties and environment variables.

This is an unfortunate side effect of how the specification is done, yet there is no way to get hold of a `Config` instance while setting up a `ConfigSource` that is not broken.

ljnelson: I see.  I was following [this discussion](https://groups.google.com/forum/#!topic/microprofile/rkkZJGJpxMU) which implied that (careful) use of `ConfigProvider.getConfig()` from within a `ConfigSource` was possible and generally accepted.  Clearly I misunderstood the discussion.  What would you recommend instead?

tomas-langer: 'ConfigProvider.getConfig()' will give you either the default config, or a config previously configured by a user.
In case you expect that your config source's configuration in one of the default sources (system properties, env vars, 'microprofile-config.properties'), you can use the builder to build your own instance of the config using builder and just these three sources.
This way you never get to the situation, that your config source would be created as part of your own call to 'ConfigProvider.getConfig()'.
You would obviously need to document such thing clearly.

How to (sorry, not in front of a working IDE, typing from memory):
'Config configOfMySource = ConfigProviderResolver.instance().newBuilder().addDiscoveredConverters().addDefaultSources().build()'

I understand that the approach you have implemented works in a single scenario, but unfortunately there are other cases where it would not work as expected.
The scenarios I consider:
1. Use of default configuration - in this case you create an additional instance of configuration and register it as current, depending on implementation of ConfigProviderResolver to be fine with a parallel call to 'getConfig()'
2. Use of manual configuration with custom config sources - you would register a default instance of config as current, even though that is not desired by the user (as they are building a custom instance and want to register that one instead) - so event though possibly for a short period of time, the system would be misconfigured

tomas-langer: My proposal is similar to the one in the mentioned discussion - see the 'CloudantConfigSource' that uses:
'Config c = ConfigProviderResolver.instance().getBuilder().addDefaultSources().withSources(configSrc).build();' to prevent cycle

ljnelson: I see.  Doesn't `ConfigProvider.getConfig()` [also pick up `ConfigSource` instances on the classpath](https://github.com/eclipse/microprofile-config/blob/1.4/spec/src/main/asciidoc/configprovider.asciidoc#accessing-or-creating-a-certain-configuration) (i.e. it's not just the defaults)?  I think maybe that's where I must have gotten confused.

PR #1766, opened 2020-05-11
WebServer configuration changes

### LINKED ISSUE #453: Config property names inconsistencies
Config issue. The docs point to,

https://helidon.io/docs/latest/apidocs/index.html?io/helidon/webserver/ServerConfiguration.html

In order to override `workersCount`, the property `workers` is expected as shown here from the Helidon code:

           config.get("workers").asInt().ifPresent(this::workersCount);

There may be a few others like this one.

### LINKED ISSUE #575: Refactor ServerConfiguration to follow development guidelines
Refactor ServerConfiguration to follow Helidon development guidelines for Build/Config.

The config mapping for SocketConfiguration is done by ServerConfiguration, this should be done in SocketConfiguration instead.

The fix for #519 introduces a new configuration attribute called "enabled-protocols" to SocketConfiguration.

Instead this attribute should be added to SSLContextBuilder, so that all SSL configuration is nested under the same config object. 

Note that SSLContextBuilder builds `javax.net.ssl.SSLContextBuilder`, we will need to create a sub-class to store custom attributes. SocketConfiguration.Builder also accepts `javax.net.ssl.SSLContext` ; it will need to be changed to the new SSLContext class, and SSLContextBuilder will need to have a way to build instances of the new SSLContext class using instances of `javax.net.ssl.Context`.

The new SSLContext class should be something like `io.helidon.webserver.SSLContext`, with a inner statis class Builder. SSLContextBuilder should inherit `io.helidon.webserver.SSLContext.Builder` and be marked deprecated.

tomas-langer:
Resolves #575 
Resolves #453 

Need to update:
 - documentation
 - release notes

Ready for code review (except for the docs mentioned above)


examples/cors/src/test/java/io/helidon/examples/cors/MainTest.java
```diff
@@ -53,11 +51,16 @@
...
-
     @BeforeAll
     public static void start() throws Exception {
-        webServer = Main.startServer();
+        // the port is only available if the server started already!
+        // so we need to wait
+        webServer = Main.startServer()
+                        .start()
+                        .toCompletableFuture()
+                        .get();
```
danielkec, line 61: Timeout can be friendlier in case something goes wrong

PR #1769, opened 2020-05-12
Fix handling of generics when reading objects

graemerocher:
Generics information is not currently used when deserializing objects which leads to Jackson deserializing objects as a `List` of `Map` instances causing downstream class cast exceptions. This change ensures Jackson's `JavaType` API is used to correctly deserializing objects using generics.

### DIFF
--- a/media/jackson/common/src/main/java/io/helidon/media/jackson/common/JacksonBodyReader.java
+++ b/media/jackson/common/src/main/java/io/helidon/media/jackson/common/JacksonBodyReader.java
@@ -15,9 +15,9 @@
  */
 package io.helidon.media.jackson.common;
 
-import java.io.IOException;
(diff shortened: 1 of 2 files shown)


romain-grecourt:
@graemerocher Looks like you have some checkstyle errors.

jbescos:
I think the checkstyle doesn't like the order of imports.
```
+ ./etc/scripts/checkstyle.sh
[ERROR] /builds/workspace/ublic_helidon-main_PR-1763-merge/media/jackson/common/src/main/java/io/helidon/media/jackson/common/JacksonBodyReader.java:33: Wrong order for 'com.fasterxml.jackson.databind.JavaType' import. [ImportOrder]
CHECKSTYLE ERROR
```

Issue #1741, opened 2020-05-06
OutputStreamPublisher and related refactoring


tomas-langer:
## Environment Details
* Helidon Version: 2.0+
* Helidon SE
----------

## Problem Description
We have the following use cases and implementations:

*  `OutputStreamPublisher`: `OutputStream` and `Publisher` - we write to output stream, it is used as publisher to send entity
* `InputStreamPublisher`: `Publisher` constructed from `InputStream` - we have an input stream, we need a publisher 
* `DataChunkInputStream`: `InputStream` constructed from `Publisher` - we have a publisher, we need an input stream

We should move these classes to a single location (either media or reactive common)
We should revisit implementations to make sure these are correctly synchronized (esp. the `OutputStreamPublisher` had a few issues and maybe some are left -make sure it handles all states correctly (cancel, error etc.). The wait cycle may also be cleaned up (maybe we could get a notification when more data is requested by subscriber?).
We should maybe support a reader and writer for `InputStream` for WebClient and WebServer

danielkec:
@tomas-langer `maybe we could get a notification when more data is requested by subscriber?` we have onRequest callback now https://github.com/oracle/helidon/pull/1900/files#diff-972b28cea8694114d6cd5214b40bbe4dR73-R105

tomas-langer:
Closing this issue, as we no longer use reactive in Helidon webserver. If there is any issue related to reactive libraries, submit a new github issue.

PR #1770, opened 2020-05-12
Multi from InputStream

danielkec:
Alternative impl for `InputStreamPublisher` #1741


common/reactive/src/main/java/io/helidon/common/reactive/MultiFromInputStream.java
```diff
@@ -0,0 +1,145 @@
...
+                            downstream.onError(new IllegalArgumentException(
+                                    "Rule §3.9 violated: non-positive request amount is forbidden"));
+                        } else if (isCanceled == NORMAL_CANCEL) {
+                            try {
+                                toClose.close();
+                            } catch (IOException e) {
+                                // Rule §1.8 says eventually
+                                downstream.onError(e);
+                            }
+                        }
```
danielkec, line 98: Not sure if publisher should be responsible for closing the `inputstream` but variant with `onCancel` is kind of ugly because of the checked exception
```java
Multi.from(is, 5)
.onCancel(()->{
  try{
    is.close();
    ...
})...
```

common/reactive/src/main/java/io/helidon/common/reactive/MultiFromInputStream.java
```diff
@@ -0,0 +1,145 @@
...
+                        return;
+                    }
+
+                    ByteBuffer value;
+
+                    try {
+                        value = ByteBuffer.wrap(inputStream.readNBytes(bufferSize));
```
tomas-langer, line 105: AFAIK this method should not be blocking. If you block here, you block the subscriber trying to request data.
It feels the only way to implement this correctly is to use an executor service.
I think the existing implementations are not correct.

danielkec: Depends if you can trust InputStream to not block, then it would be excusable I guess, after all every method out of our control can block

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
```diff
@@ -178,6 +181,43 @@
...
+     * @param inputStream the Stream to publish
+     * @param buffSize size of the chunks to be sent to downstream in the form of {@link ByteBuffer}
+     * @param executor executor to use for waiting at {@link InputStream}'s blocking reads
+     * @return Multi
+     * @throws NullPointerException if {@code stream} is {@code null}
+     */
+    static Multi<ByteBuffer> from(InputStream inputStream, int buffSize, ExecutorService executor) {
+        Objects.requireNonNull(inputStream, "stream is null");
+        return new MultiFromInputStream(inputStream, buffSize, executor);
+    }
```
danielkec, line 201: @tomas-langer Added both variants so user can keep performance up in case he can trust `InputStream` to not block(for example `ByteArrayInputStream`) or utilize his own `ExecutorServic`e in case blocking of read operations is expected.

tomas-langer: Our internal rules do not allow `null` parameters - please check the executor is not null in this case

common/reactive/pom.xml
```diff
@@ -67,13 +67,8 @@
...
         </dependency>
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-all</artifactId>
+            <groupId>io.helidon.microprofile.bundles</groupId>
```
tomas-langer, line 70: Please do not use Helidon microprofile libraries in SE modules

common/reactive/src/main/java/io/helidon/common/reactive/MultiFromInputStream.java
```diff
@@ -0,0 +1,186 @@
...
+            if (n <= 0L) {
+                canceled = BAD_REQUEST;
+                n = 1; // for cleanup
+            }
+
+            if (SubscriptionHelper.addRequest(this, n) != 0L) {
+                return;
+            }
+
+            if (executorService == null) {
```
tomas-langer, line 153: This approach is 
1. error prone
2. hard to read

please either use two implementation classes, or use an executor service that executes in thread.

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
danielkec: fixed

common/reactive/pom.xml
danielkec: fixed

common/reactive/src/main/java/io/helidon/common/reactive/MultiFromInputStream.java
danielkec: fixed

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
```diff
@@ -178,6 +181,44 @@
...
+     * {@link Multi#from(java.io.InputStream, int, java.util.concurrent.ExecutorService)}.
+     * <p>
+     * The operator calls {@link InputStream#close()} when the cancel signal is intercepted.
+     *
+     * @param inputStream the Stream to publish
+     * @param buffSize    size of the chunks to be sent to downstream in the form of {@link ByteBuffer}
+     * @return Multi
+     * @throws NullPointerException if {@code stream} is {@code null}
+     */
+    static Multi<ByteBuffer> from(InputStream inputStream, int buffSize) {
```
tomas-langer, line 217: The `buffSize` in both methods seems a bit of an overhead.
I think the static method `from` should only accept `InputStream` (and `ExecutorService` for blocking variant).
If you want to have the buffSize, a builder may be the right way to go, using some reasonable default for buffer size otherwise.

danielkec: We can do something like this:
```java
        String s = Multi.from(is, executorService)
                .withByteBufferSize(4)
                .map(StandardCharsets.UTF_8::decode)
                ....
```

common/reactive/src/main/java/io/helidon/common/reactive/MultiFromInputStream.java
danielkec: removed

common/reactive/src/main/java/io/helidon/common/reactive/Multi.java
tomas-langer: Helidon builder pattern:
- A static factory method `builder` that creates a new builder instance
- Builder has only setters in Helidon style (e.g. no verb)
- There is a clear transition from builder to built instance

So your example:
```java
Multi.builder(is, executorService)
   .bufferSize(1024)
   .build()
   .map...
```

I think it is enough you are breaking the rule for static factory methods, which should always be called `create` and not `from`

danielkec:
```java
IoMulti.builder(is)
                .executor(Executors.newSingleThreadExecutor())
                .byteBufferSize(BUFFER_SIZE)
                .build();
```

PR #1796, opened 2020-05-14
JSON-P Streaming support

### LINKED ISSUE #1790: Create a WebServer media stream mapper for JSON-P
We need support for `Publisher<JsonObject>` in Helidon, so we can easily map result from database to a server response.

The tools are ready, we need to add a stream mapper that would support it.

Also remove the existing `DbResultSupport` from DbClient and refactor examples to use the new JSON-P streaming support.

This will be producing the usual JSON object (not using streaming media type)

### LINKED ISSUE #1793: Missing media mapper ends with unexpected results
When a media type mapper is missing from WebServer, you get:

```
io.netty.handler.codec.EncoderException: java.lang.IllegalStateException: unexpected message type: DefaultLastHttpContent
```
and the client receives empty response.

This should end up in a 500 error code, and a stack trace in the log.

tomas-langer:
Resolves #1790 
Resolves #1793


media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpBodyStreamWriter.java
```diff
@@ -0,0 +1,88 @@
...
+        context.contentType(contentType);
+
+        // we do not have join operator
+        AtomicBoolean first = new AtomicBoolean(true);
+
+        JsonStructureToChunks jsonToChunks = new JsonStructureToChunks(jsonWriterFactory,
+                                                                       context.charset());
+
+        // we also do not have an append operator
+        Multi<DataChunk> stream = Multi.concat(
```
romain-grecourt, line 68: @danielkec FYI. Can this be done better ? Is there something we can add in Multi to support this ?

media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpBodyWriter.java
```diff
@@ -50,15 +49,14 @@ public boolean accept(GenericType<?> type, MessageBodyWriterContext context) {
 
     @Override
     public Publisher<DataChunk> write(Single<JsonStructure> content, GenericType<? extends JsonStructure> type,
-            MessageBodyWriterContext context) {
+                                      MessageBodyWriterContext context) {
 
         MediaType contentType = context.findAccepted(MediaType.JSON_PREDICATE, MediaType.APPLICATION_JSON);
         context.contentType(contentType);
-        return content.flatMap(new JsonStructureToChunks(jsonWriterFactory, context.charset()));
+        return content.map(new JsonStructureToChunks(jsonWriterFactory, context.charset()));
```
romain-grecourt, line 56: I under what you are doing here, but if we do this we need to do this for all the places where there is similar code. i.e  jsonb, jackson.

media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpSupport.java
```diff
@@ -61,6 +62,19 @@ public JsonpBodyWriter newWriter() {
...
+    /**
+     * Create a new JSON-P stream writer.
+     * <p>
+     * This stream writer supports {@link java.util.concurrent.Flow.Publisher publishers}
+     * of {@link javax.json.JsonStructure} (such as {@link javax.json.JsonObject})
+     * , writing them as an array of JSONs.
+     *
+     * @return JSON processing stream writer.
+     */
+    public JsonpBodyStreamWriter newStreamWriter() {
```
romain-grecourt, line 74: The newXXX static methods have been removed. Let's remove it.

Verdent: Actually it is not removed in this version. I didn't commit these changes yet.

media/jsonb/common/src/main/java/io/helidon/media/jsonb/common/JsonbBodyWriter.java
```diff
@@ -48,6 +49,13 @@ private JsonbBodyWriter(Jsonb jsonb) {
     public boolean accept(GenericType<?> type,
             MessageBodyWriterContext context) {
 
+        // We are excluding the following types from support:
+        // 1. any char sequence
+        // 2. Flow.Publisher - that can only be supported by streaming media
+        if (Flow.Publisher.class.isAssignableFrom(type.rawType())) {
```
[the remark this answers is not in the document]
tomas-langer: Yes, because otherwise when you have JSON-B registered, you do not get an error, if you do  `response.send(Publisher<Pojo>)`
You should get appropriate error telling you there is no registered writer, but JSON-B just processes it

tomas-langer:
`examples/toto-app/frontend-package-lock.json` - I am fixing what I broke in webserver configuration changes.
This is the original state of the file from master.

media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpBodyWriter.java
tomas-langer: this is not a different functionality - it still works the same, I am just skipping going from an immediate value to publisher, as that created an unnecessary `flatMap` operation.
Also when I need to process records one by one, the intermediate publisher would cause trouble (as if it had more than one chunk per `JsonObject`, I would insert two commas there).

We should revisit it when implementing the stream writers for `JSON-B` and for `Jackson`

tomas-langer: will add a reference to those issues

tomas-langer:
Removed copyright changes and examples/toto-app/frontend-package-lock.json changes.

tomas-langer:
Split unrelated changes to #1804 and #1805

media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpBodyStreamWriter.java
danielkec: Now we do #1806

romain-grecourt: @tomas-langer Can you update the corresponding code to leverage #1806

media/jsonb/common/src/main/java/io/helidon/media/jsonb/common/JsonbBodyWriter.java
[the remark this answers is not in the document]
tomas-langer: I will look for the place we check for `byte[]` and update it

media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpBodyStreamWriter.java
```diff
@@ -0,0 +1,88 @@
...
+import io.helidon.common.reactive.Multi;
+import io.helidon.common.reactive.Single;
+import io.helidon.media.common.MessageBodyStreamWriter;
+import io.helidon.media.common.MessageBodyWriterContext;
+import io.helidon.media.jsonp.common.JsonpBodyWriter.JsonStructureToChunks;
+
+/**
+ * Message body writer for {@link javax.json.JsonStructure} sub-classes (JSON-P).
+ */
+public class JsonpBodyStreamWriter implements MessageBodyStreamWriter<JsonStructure> {
```
tomas-langer, line 37: Add a unit test.

media/jsonp/common/src/main/java/io/helidon/media/jsonp/common/JsonpBodyStreamWriter.java
tomas-langer: Done

tomas-langer:
Merging these changes as other changes depend on them.

PR #1799, opened 2020-05-14
Docs for MP Config

vanajaguru:
Added the updates in the MP docs. Please review.

### DIFF
--- a/docs/mp/config/01_introduction.adoc
+++ b/docs/mp/config/01_introduction.adoc
@@ -27,47 +27,58 @@
 
 
 == About {spec-name}
-Helidon MP Config is an implementation of https://github.com/eclipse/microprofile-config/[Eclipse MicroProfile Config] that now includes Helidon-specific meta-configuration and APIs. You can configure your applications using MicroProfile's config properties file and APIs and extend the configuration system with Helidon-specific meta-configuration and APIs. You can also write and register custom ConfigSources.
+Helidon MicroProfile Config is an implementation of https://github.com/eclipse/microprofile-config/[Eclipse MicroProfile Config] that includes Helidon-specific meta-configuration and APIs. You can configure your applications using MicroProfile's config properties file and APIs.
+
+You can also extend the configuration using MicroProfile SPI to add custom `ConfigSource` and `Converter` by implementing the  `org.eclipse.microprofile.config.spi.ConfigSource` and   `org.eclipse.microprofile.config.spi.Converter` interfaces respectively.
 
 === Helidon MicroProfile Config Features
 
 Helidon MicroProfile Config offers the following features:
 
 * *References* +
-You can use dollar, curly brackets, and reference to another key in a value to provide references and contruct the value when you use meta-configuration file for MP configuration. You can either reference the whole value or reference the value and add prefix or suffix, and the referenced value is replaced with the property.
+You can use dollar, curly brackets, and reference to another key in a value to provide references and contruct the value when you use meta-configuration file for MicroProfile configuration. You can either reference the whole value, or reference the value and add prefix or suffix, and the referenced value is replaced with the property.
 
 [source=java]
 .Example
 ----
 
-providers:
-    - abac:
-      # Adds ABAC Provider - it does not require any configuration
-    - oidc:
-        client-id: "${ALIAS=security.properties.idcs-client-id}"
-        client-secret: "${ALIAS=security.properties.idcs-client-secret}"
-        identity-uri: "${ALIAS=security.properties.idcs-uri}"
-        # A prefix used for custom scopes
-        scope-audience: "http://localhost:7987/test-application"
-        proxy-host: "${ALIAS=security.properties.proxy-host}"
-        frontend-uri: "${ALIAS=security.properties.frontend-uri}"
-    - idcs-role-mapper:
-        multitenant: false
-        oidc-config:
-          # Repeat the IDCS configuration, as in this case. IDCS serves both as open ID connect authenticator and as a role mapper. Use of minimal configuration here.
-          client-id: "${ALIAS=security.properties.idcs-client-id}"
-          client-secret: "${ALIAS=security.properties.idcs-client-secret}"
-          identity-uri: "${ALIAS=security.properties.idcs-uri}"
+- oidc:
+    client-id: "${security.properties.idcs-client-id}"
+    client-secret: "${security.properties.idcs-client-secret}"
+    identity-uri: "${security.properties.idcs-uri}"
+    // A prefix used for custom scopes
+    scope-audience: "http://localhost:7987/test-application"
+    proxy-host: "${security.properties.proxy-host}"
+    frontend-uri: "${security.properties.frontend-uri}"
+    // Support for non-public signature JWK (and maybe other IDCS specific handling)
+    server-type: "idcs"
+- idcs-role-mapper:
+    multitenant: false
+    oidc-config:
+      // In this case, repeat the IDCS configuration as IDCS serves both as open ID connect authenticator and as role mapper. So, using minimal configuration.
+      client-id: "${security.properties.idcs-client-id}"
+      client-secret: "${security.properties.idcs-client-secret}"
+      identity-uri: "${security.properties.idcs-uri}"
 ----
 
 
 * *Encryption* +
-You can encrypt configuration values in MicroProfile Config. The config encryption filter is enabled by default. For more information, see <<about/02_configuration-secrets.adoc,Configuration Secrets>>.
+The config encryption filter in MicroProfile Config is enabled by default.  For more information, see <<about/02_configuration-secrets.adoc,Configuration Secrets>>.
 
-//Please provide an example of configuration
+//Please review this example
 [source=java]
 .Example
-
+----
+Password can be stored as follows:
+    client_secret=${GCM=mYRkg+4Q4hua1kvpCCI2hg==}
+// Encrypted password using a master password, this can be used in production environments
+    service_password=${RSA=mYRkg+4Q4hua1kvpCCI2hg==} 
+// Encrypted password using a public key
+    another_password=${ALIAS=service_password}  
+// Reference to another property, that is encrypted
+    cleartext_password=${CLEAR=known_password}
+// Password in clear text, this can be used in development environments
+----
 
 == Guides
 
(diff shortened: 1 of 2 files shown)


docs/mp/config/02_MP_config_sources.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+=== Config Source Creation based on Environment Variables
+
+The Config Source replaces the properties with environment variables as defined in MicroProfile Config specification.
+
+[source,java]
+----
+ConfigSource environmentVariables() { <1>
+        return new MpEnvironmentVariablesSource();
+----
```
tomas-langer, line 157: This is not correct. Helidon classes do not have public constructors. You need to use the factory methods in `MpConfigSources`

docs/mp/config/02_MP_config_sources.adoc
```diff
@@ -0,0 +1,230 @@
...
+
+=== Config Source Creation based on a File on the File System
+
+The Config Source reads the file just once when the source is created and any other changes to the file are ignored.
+
+[source,java]
+----
+ConfigSource create(Path path) { <1>
```
tomas-langer, line 165: Such documentation is not useful, unless you show the class that contains the `create` method.

docs/mp/config/01_introduction.adoc
```diff
@@ -21,15 +22,52 @@
 :toc-placement: preamble
 :spec-name: MicroProfile Config
 :description: {spec-name} support in Helidon MP
-:keywords: helidon, mp, microprofile, config
+:keywords: helidon, mp, microprofile, config, encryption, reference
+
 
 
 == About {spec-name}
+Helidon MP Config is an implementation of https://github.com/eclipse/microprofile-config/[Eclipse MicroProfile Config] that now includes Helidon-specific meta-configuration and APIs. You can configure your applications using MicroProfile's config properties file and APIs and extend the configuration system with Helidon-specific meta-configuration and APIs. You can also write and register custom ConfigSources.
```
tomas-langer, line 30: This does not read very nice.

1. the `now` in the first sentence is not correct. We already had support for that in 1.4
2. I would use the Oxford comma: "You can configure your application using MicroProfile's config properties and APIs, and extend the configuration system with Helidon specific features."
3. The last sentence is redundant (or not correct) - custom ConfigSources are part of the MicroProfile specification. It is usually called an SPI. Maybe a new paragraph about extensibility? E.g. "You can extend configuration with MicroProfile Config SPI for `ConfigSource` and `Converter` functionality."

docs/mp/config/01_introduction.adoc
```diff
@@ -21,15 +22,52 @@
...
+
+[source=java]
+.Example
+----
+
+providers:
+    - abac:
+      # Adds ABAC Provider - it does not require any configuration
+    - oidc:
+        client-id: "${ALIAS=security.properties.idcs-client-id}"
```
tomas-langer, line 47: This example is wrong - the correct reference is by `${}`.
This is a deprecated feature from another component.
To fix it, just remove the `ALIAS=` from each value.

Correct example:
```yaml
- oidc:
      client-id: "${security.properties.idcs-client-id}"
```

docs/mp/config/01_introduction.adoc
```diff
@@ -21,15 +22,52 @@
...
+          client-id: "${ALIAS=security.properties.idcs-client-id}"
+          client-secret: "${ALIAS=security.properties.idcs-client-secret}"
+          identity-uri: "${ALIAS=security.properties.idcs-uri}"
+----
+
+
+* *Encryption* +
+You can encrypt configuration values in MicroProfile Config. The config encryption filter is enabled by default. For more information, see <<about/02_configuration-secrets.adoc,Configuration Secrets>>.
+
+//Please provide an example of configuration
```
tomas-langer, line 67: example is in configuration file - in java, it is used as any other property
```properties
# for development, you can define passwords in clear-text
service1.password="${CLEAR=known_password}"
# for other environments, you can provide encrypted passwords
service2.password="${GCM=bQPPs+e2i4Sry7xB3MJWUPCnGWzx69Rgs6kskDIq8LRSJFonZ1NDlLhVhZBhtwpVdNlDa9qQSw==}"
```

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
 --
 ====
 
-== Using {spec-name} in Helidon MP applications
 
-=== Configuration
+=== Using MicroProfile Config Sources and META-INF
```
tomas-langer, line 85: META-INF is just a directory, should not be mentioned here at all

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+
+Default configuration values can be retrieved from system properties, environment variables, files, directories, URLs or configuration files of the form `${CLASSPATH}/META-INF/microprofile-config.properties`.
```
tomas-langer, line 87: Not sure if "Default" is correct here - the default is only system properties, environment variables and `microprofile-config.properties`.
When not using the Oxford comma, it is very confusing whether the "files of the form" belong to just the configuration files, or the whole list before that.
I think this deserves a visual list to make it clear, such as:
Configuration can be retrieved from the following configuration sources:

- System properties - a mutable source using `System.getProperties()` to obtain its configuration values
- Environment variables - an immutable source using `System.env()` to obtain its configuration values, also resolves aliases as defined by the MicroProfile Config specification
- `${CLASSPATH}/META-INF/microprofile-config.properties`- properties configuration source as defined by the MP Config spec
- `${CLASSPATH)/application.yaml` - Helidon default configuration source
- File - you can use any properties file on the file system (see `MpConfigSources.create(Path)`)
- URL - you can use a properties source from an URL (see `MpConfigSources.create(URL)`)
- `Map<String, String>` - you can create a source from a Map (see `MpConfigSources.create(Map)`)
- `Properties` - you can create a source directly from properties (see `MpConfigSources.create(Properties)`)
- File on classpath - you can use any properties file on classpath (see `MpConfigSources.classpath(String)`)
- YAML - you can use YAML config sources, as supported by `YamlMpConfigSource`

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
 ====
 
-== Using {spec-name} in Helidon MP applications
 
-=== Configuration
+=== Using MicroProfile Config Sources and META-INF
+
+Default configuration values can be retrieved from system properties, environment variables, files, directories, URLs or configuration files of the form `${CLASSPATH}/META-INF/microprofile-config.properties`. 
+
+The example below shows how the MicroProfile configuration file `microprofile-config.properties` can be used to modify properties such as server port or host IP.
```
tomas-langer, line 89: not "host IP", but "server listen host"

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+# Microprofile server properties
+server.port=8080
+server.host=0.0.0.0
+----
+
+=== Using MicroProfile Config API
+
+You can use the default MicroProfile configuration by injecting configuration properties or use `ConfigProvider.getConfig()`.
+
+----
```
tomas-langer, line 104: would be nice to have an injection example as well, such as:

```java
@Inject
public GreetingProvider(@ConfigProperty(name = "app.greeting") String message) {
    this.message = message;
}
```

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+   org.eclipse.microprofile.config.Config config = ConfigProvider.getConfig();
```
tomas-langer, line 105: Add the actual retrieval of a property:
```java
this.message = config.getValue("app.greeting", String.class);
```

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+----
+
+=== Using Helidon Meta Configuration
+
+The meta configuration uses underlying implementation of Helidon config. In this case, you can either inject configuration properties or use `ConfigProvider.getConfig()`.
```
tomas-langer, line 110: "In this case"? Which is the same as above - so the second sentence is I think redundant.

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+  - type: "environment-variables" <2>
+  - type: "file"
+    properties:
+      optional: true
+      path: "mp-config.yaml" <3>
+  - type: "classpath"
+    properties:
+      optional: true
+      resource: application.yaml <4> 
+    type: "classpath"
```
tomas-langer, line 128: missing a `-` before `type`

tomas-langer: Also for this specific source add
`multi-source: true`
as a new line after `- type: "classpath"

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
-=== Use case 1
+=== Using Helidon Config APIs
 
-- Description
-- Small code snippet
+You can use the `MpConfigSources.create(helidonConfig)` to create a config source from Helidon config and then use it to create a MicroProfile instance.
 
-=== Use case 2
+----
+ConfigBuilder metaConfig(io.helidon.config.Config metaConfig) {
```
tomas-langer, line 145: this is implementation code of Helidon, not an example, please remove this whole block (the whole method `metaConfig`)

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+        io.helidon.config.Config helidonConfig = io.helidon.config.Config.builder()
+                .config(metaConfig)
+                .build();
+        this.sources.add(new OrdinalSource(MpConfigSources.create(helidonConfig))); <1>
+        return this;
+    }
 
-- Description
-- Small code snippet
+ConfigProviderResolver.instance()
```
tomas-langer, line 153: add the following lines before `ConfigProviderResolver.instance()`:
```
io.helidon.config.Config helidonConfig = io.helidon.config.Config.builder()
                .addSource(ConfigSources.create(Map.of("key", "value")))
                .build();
```

docs/mp/config/01_introduction.adoc
```diff
@@ -43,29 +81,88 @@ Step-by-step guide about using {spec-name} in your Helidon MP application.
...
+                    .getBuilder()
+                    .withSources(MpConfigSources.create(config)) <2>
```
tomas-langer, line 155: and use `helidonConfig` instead of `config` here

docs/mp/config/01_introduction.adoc
tomas-langer: After that, assign the result of `ConfigProviderResolver.instance()` block to a variable
```java
Config config = ConfigProviderResovler.instance()
....
```

docs/mp/config/02_MP_config_sources.adoc
```diff
@@ -0,0 +1,155 @@
...
+* `String getName()`
+* `Map<String, String> getProperties()`
+* `String getValue(String key)`
+* `getOrdinal()`
+
+=== Example of a Custom Config Source
+
+[source,java]
+----
+public class CustomConfigSource implements ConfigSource {
```
tomas-langer, line 53: I think the fields used to be constants - please change it back to constants, or rename the fields and reformat the source code.
If a fields is in all capitals, it is usually a constant (and must be `static final` - in this case `private static final`)
Fields that are not constants use camel case with initial lower case (so it would be `name`, `priority` and `properties`)

In this example, the constant is expected.

docs/mp/config/02_MP_config_sources.adoc
```diff
@@ -0,0 +1,155 @@
...
+== Creating MicroProfile Config Source from Helidon SE Config Instance
+
+You can create MicroProfile Config Source from Helidon SE Config to use advanced Helidon SE features in Helidon MP. The Config Source is mutable if the config uses either polling strategy and change watchers or polling strategy or change watchers.
+The latest config version is queried each time  `org.eclipse.microprofile.config.spi.ConfigSource#getValue(String)` is called.
+
+//Please provide example
+
+== Create Yaml MicroProfile Config Source Programatically
+
+You can create Yaml Microprofile Config Source from a path or a URL. When you create a MicroProfile instance from the builder, the `YamlMPConfigSource` allows you to edit the list of config sources and creates a custom YAML.
```
tomas-langer, line 144: It does not "allow you to edit"
It works exactly the same as the `MpConfigSources` above - you can use `YamlMpConfigSource` (btw there is a typo in your text in class name) to create a new config source to register it with the builder.

docs/mp/config/02_MP_config_sources.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+
+=== Example
+
+[source,java]
+.Create YamlMPConfigSource from a path
```
tomas-langer, line 150: Again, showing implementation method of Helidon, not an example. 
Example would be similar to example in main MP Config docs:
```java
ConfigProviderResolver.instance().newBuilder()
 .withSources(YamlMpConfigSource.create(path))
.build()
```

docs/mp/config/01_introduction.adoc
vanajaguru: I am not sure when you are referring to (see MpConfigSources.create(Path)). Please let me know what reference should be provided here.

tomas-langer:
Closing this PR, I will create a new one with updated docs.

PR #1803, opened 2020-05-14
Fix #1802 - Allow use of filters and mappers when converting MP to He…

### LINKED ISSUE #1802: Allow use of filters and mappers when converting MP to Helidon config

## Environment Details
* Helidon Version: 2.0.0-SNAPSHOT
* Helidon MP
* JDK version: 11.0.7
* OS: Mac OS 10.15.4

----------

## Problem Description

The issue is that gRPC MP and SE client share configuration, which means that they both ultimately use Helidon SE Config, and depend on object-mapping to read complex config structure into a POJO that describes gRPC channel.

However, because we start everything up from MP world, we are loading MpConfig and converting it to Config. In the process, mapper services get disabled and this conversion doesn’t work any longer. However, if we comment out the line below within MpConfig.toHelidonConfig() that disables them, everything works fine:

```java
io.helidon.config.Config.Builder builder = io.helidon.config.Config.builder()
                //.disableMapperServices()
                .disableCaching()
                .disableParserServices()
                .disableFilterServices(); 
```

Response from @tomas-langer:

> The intention of the change was to get the same configuration as we have in MP. But that can never be the case.
> So I think we can enable mappers and filters (to allow use of advanced Helidon SE features if developers want to)

rlubke:
…lidon config

Fix is simple -  remove the lines disabling filters and mappers when creating the new SE configuration.

Added test to verify the mappers would be available in this scenario.
Was uncertain as to the best approach for filters (or if the mapper test is the best way to verify for that matter), so recommendations are welcome if additional testing on this is desired.

### DIFF
--- a/config/config-mp/src/main/java/io/helidon/config/mp/MpConfig.java
+++ b/config/config-mp/src/main/java/io/helidon/config/mp/MpConfig.java
@@ -52,10 +52,8 @@ public final class MpConfig {
         io.helidon.config.Config.Builder builder = io.helidon.config.Config.builder()
                 .disableEnvironmentVariablesSource()
                 .disableSystemPropertiesSource()
-                .disableMapperServices()
                 .disableCaching()
-                .disableParserServices()
-                .disableFilterServices();
+                .disableParserServices();
 
         if (mpConfig instanceof MpConfigImpl) {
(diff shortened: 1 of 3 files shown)


tomas-langer:
Thank you for this contribution!

PR #1807, opened 2020-05-15
Change CORS config key from path-prefix to path-expr; method names also

### LINKED ISSUE #1767: CORS mapped config should use `path` or `path-expr` not `path-prefix`
## Environment Details
* Helidon Version: 2.x
* Helidon SE or Helidon MP: both
* JDK version:n/a
* OS:n/a
* Docker version (if applicable):n/a

----------
The mapped style of CORS configuration uses `path-prefix` for the path value. In fact, the value is used as a _path expression_ to match incoming request paths, as with the `PathMatcher` class. So we should change the config key and the doc to reflect how this value is actually used.

`path-expr` is one short option that conveys the usage. Just plain `path` is another, although that name does not reinforce the way the value is actually used.

tjquinno:
Resolves #1767 

The value is, in fact, used as a pattern (expression) and not a prefix.

### DIFF
--- a/examples/cors/src/main/resources/application.yaml
+++ b/examples/cors/src/main/resources/application.yaml
@@ -28,7 +28,7 @@ restrictive-cors:
 # The the example app uses the following for overriding other settings.
 #cors:
 #  paths:
-#    - path-expr: /greeting
+#    - path-pattern: /greeting
 #      allow-origins: ["http://foo.com", "http://there.com", "http://other.com"]
 #      allow-methods: ["PUT", "DELETE"]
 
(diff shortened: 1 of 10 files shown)


tomas-langer:
A note for consistency - in Security, we have the following configuration:
```yaml
security.web-server:
        paths:
            - path: "/noRoles"
              methods: ["get"]
            - path: "/user[/{*}]"
              methods: ["get"]
              roles-allowed: ["user"]
```

I use just `path`, even though it can be an expression (as in the example above).
I did not feel the `expr` was required here.
In routing, we call it `pathPattern` 

This is adding a third option.
Please consider this - does it make sense to change all of them to expression, or use one of the existing terms?

Options:
1. `path`
2. `path-pattern`
3. `path-expr`

I have approved this PR, as we may delay that decision for a future Helidon version...

tjquinno:
After a quick slack chat with Tomas, I am re-changing the config key name to `path-pattern` and associated method names, arguments, etc. to `pathPattern` to be consistent with similar usage elsewhere in `WebServer`. Update to the PR coming soon.

PR #1809, opened 2020-05-15
Health check fixed

tomas-langer:
For native image and for hotspot (used Set instead of List)

### DIFF
--- a/health/health/src/main/java/io/helidon/health/HealthSupport.java
+++ b/health/health/src/main/java/io/helidon/health/HealthSupport.java
@@ -20,7 +20,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


health/health/src/main/java/io/helidon/health/HealthSupport.java
```diff
@@ -260,9 +259,9 @@ public static HealthSupport create(Config config) {
      * Fluent API builder for {@link io.helidon.health.HealthSupport}.
      */
     public static final class Builder implements io.helidon.common.Builder<HealthSupport> {
-        private final Set<HealthCheck> allChecks = new LinkedHashSet<>();
-        private final Set<HealthCheck> livenessChecks = new LinkedHashSet<>();
-        private final Set<HealthCheck> readinessChecks = new LinkedHashSet<>();
+        private final List<HealthCheck> allChecks = new LinkedList<>();
```
ljnelson, line 262: Just double-checking.  Repeatable semantics here are OK?

tests/integration/native-image/mp-1/src/main/java/io/helidon/tests/integration/nativeimage/mp1/Mp1Main.java
```diff
@@ -427,8 +429,13 @@ private static void validateHealth(Errors.Collector collector, WebTarget target)
                 .get(JsonObject.class);
 
         JsonArray checks = health.getJsonArray("checks");
-        if (checks.size() < 1) {
-            collector.fatal("There should be at least one readiness healtcheck provided by this app");
+        if (checks.size() != 1) {
+            collector.fatal("There should be a readiness healtcheck provided by this app");
```
ljnelson, line 433: typo

health/health/src/main/java/io/helidon/health/HealthSupport.java
tomas-langer: It would be called twice. There was a problem that some of our health checks just disappeared (we ended up with 2 out of 6 in a test). Some quirk most likely caused by Weld proxies and their equals/hashCode methods.

tests/integration/native-image/mp-1/src/main/java/io/helidon/tests/integration/nativeimage/mp1/Mp1Main.java
tomas-langer: Fixed.

health/health/src/main/java/io/helidon/health/HealthSupport.java
tomas-langer: I did consider using an identity set to at least prevent the same instances. Yet in MP, we register this from selected beans (so it should be unique), and in SE the user registers checks by hand (so they have full control).

PR #1814, opened 2020-05-18
Single onCancel and OnComplete fixes

danielkec:
Thx @olotenko for finding the issue

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/CompletionSingle.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/CompletionSingle.java
@@ -48,7 +48,7 @@ public abstract class CompletionSingle<T> extends CompletionAwaitable<T> impleme
     @Override
     public Single<T> onCancel(final Runnable onCancel) {
         cancelFuture.thenRun(onCancel);
-        return this;
+        return Single.super.onCancel(onCancel);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


tomas-langer:
Why did we  rewrite `Single` as an implementation of `CompletionStage`:
1. `Single` is a promise of a single value, most of reactive implementations treat this as a specific `Flow.Publisher` case, yet it is a `CompletionStage` that can act as a publisher as well
2. Lot of implementation do not handle `Flow` (or it is too complicated to handle subscribers) for `Single` cases - a lot of times the API of `CompletionStage` is more suitable to implement desired functionality
3. Helidon APIs now have a single method, and it is a choice of user whether to use `Single`, `Flow.Publisher` or `CompletionStage` for a promise of a single unit. This makes the design and used of the API much easier, as before we needed to call additional methods

danielkec:
@olotenko There is no `onComplete` signal coming from publisher if you send `cancel` from `onNext`

olotenko:
So?

danielkec:
`Single.just("a").onComplete(()-> this is missed).await()` commit message mentions operator `onComplete` not the subscriber's method

olotenko:
You've just shown yet another problem with this design where both `CompletionStage` and `Subscriber` contracts overlap. Exactly what I told you offline.

Requesting cancellation is legitimate. If it breaks `await`, the design is wrong.

PR #1787, opened 2020-05-14
Multipart

romain-grecourt:
Fixes #189
Fixes #102

PR #1830, opened 2020-05-19
BufferedEmittingPublisher as replacement for OriginThreadPublisher

danielkec:
Simplified replacement for `OriginThreadPublisher` , performance is better and api is cleaner. 
```
Benchmark                            Mode  Cnt   Score   Error  Units
EmitterTest.testEmitterReqMillion   thrpt   20  21.261 ± 0.160  ops/s
EmitterTest.testEmitterReqOneByOne  thrpt   20  11.845 ± 0.357  ops/s
EmitterTest.testEmitterUnbounded    thrpt   20  44.591 ± 0.131  ops/s
EmitterTest.testOTPReqMillion       thrpt   20  14.702 ± 0.124  ops/s
EmitterTest.testOTPReqOneByOne      thrpt   20   7.707 ± 0.088  ops/s
EmitterTest.testOTPUnbounded        thrpt   20  14.902 ± 0.026  ops/s
```
[JMH test](https://gist.github.com/danielkec/828fd2afcc0841cb508a4b73f94e25b1)

Used in webclient, webserver, Kafka connector and Multipart #1787


romain-grecourt:
I noticed you are not removing the EmittingPublisher from the kafka module, is that intended ?

common/reactive/src/main/java/io/helidon/common/reactive/BiConsumerChain.java
```diff
@@ -0,0 +1,57 @@
...
+        implements BiConsumer<T, S> {
+
+    @Override
+    public void accept(T t, S s) {
+        for (BiConsumer<? super T, ? super S> inner : this) {
+            inner.accept(t, s);
+        }
+    }
+
+    public BiConsumerChain<T, S> combineWith(BiConsumer<? super T, ? super S> another) {
```
tomas-langer, line 33: Do not use `public` methods in package local classes

common/reactive/src/main/java/io/helidon/common/reactive/RunnableChain.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+class RunnableChain extends ArrayList<Runnable> implements Runnable {
+    @Override
+    public void run() {
+        for (Runnable inner : this) {
+            inner.run();
+        }
+    }
+
+    public RunnableChain combineWith(Runnable another) {
```
tomas-langer, line 32: Please remove `public` modifier from methods in package local classes

common/reactive/src/main/java/io/helidon/common/reactive/BiConsumerChain.java
danielkec: fixed

common/reactive/src/main/java/io/helidon/common/reactive/RunnableChain.java
danielkec: fixed

danielkec:
@romain-grecourt sorry for late response, Kafka connector switched to BEP in the last commit

common/reactive/src/main/java/io/helidon/common/reactive/BufferedEmittingPublisher.java
```diff
@@ -0,0 +1,299 @@
...
+     * <li><b>param</b> {@code result} the current total cumulative requested count, ranges between [0, {@link Long#MAX_VALUE}]
+     * where the max indicates that this publisher is unbounded.</li>
+     * </ul>
+     *
+     * @param requestCallback to be executed
+     */
+    public void onRequest(BiConsumer<Long, Long> requestCallback) {
+        this.requestCallback = BiConsumerChain.combine(this.requestCallback, requestCallback);
+    }
+
```
romain-grecourt, line 82: How about having support for onEmit callbacks ?

common/reactive/src/main/java/io/helidon/common/reactive/EmittingPublisher.java
```diff
@@ -0,0 +1,334 @@
...
+    private void signalOnComplete() {
+        if (state.compareAndSet(State.NOT_REQUESTED_YET, State.COMPLETED)
+                || state.compareAndSet(State.READY_TO_EMIT, State.COMPLETED)) {
+            for (;;) {
+                try {
+                    if (emitting.getAndSet(true)) {
+                        continue;
+                    }
+                    this.subscriber.onComplete();
+                    return;
```
romain-grecourt, line 158: The subscriber reference is not released when completed.
One way of fixing it would be to add `EmittingPublisher.this.subscriber = null;` here

danielkec: Uf thx for that, there was also possible NPE if emit started right before `READY_TO_EMIT` -> `CANCELLED` state transition, but invoked `onNext` after reference release

common/reactive/src/main/java/io/helidon/common/reactive/BufferedEmittingPublisher.java
danielkec: Would be sufficient to invoke such callback right after emit invocation? To keep request counting isolated in `EmittingPublisher` the `BufferedEmittingPublisher` never know in advance if emit will be successful

romain-grecourt: What you did seems sufficient to me.

PR #1836, opened 2020-05-20
Add JSONB support to gRPC

aseovic:
A very simple Marshaller implementation that allows us to support JSON serialization between gRPC clients and servers OOB, using JSONB.


grpc/core/src/main/java/io/helidon/grpc/core/JsonbMarshaller.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final Jsonb JSONB = JsonbBuilder.create();
+
+    private final Class<T> clazz;
+
+    /**
+     * Construct {@code JsonbMarshaller} instance.
+     *
+     * @param clazz  the type of object to marshall
+     */
+    public JsonbMarshaller(Class<T> clazz) {
```
tomas-langer, line 48: Is the usage of public constructor for a marshaller a standard? 
If not, please change to `create()` factory method.

grpc/core/pom.xml
```diff
@@ -116,6 +116,14 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
+        <dependency>
```
tomas-langer, line 119: This dependency should never be in SE. Please change it to `provided` scope.

grpc/core/src/main/java/io/helidon/grpc/core/JsonbMarshaller.java
```diff
@@ -0,0 +1,74 @@
...
+        return JSONB.fromJson(in, clazz);
+    }
+
+    /**
+     * A {@link MarshallerSupplier} implementation that supplies
+     * instances of {@link JsonbMarshaller}.
+     */
+    @Named("jsonb")
+    @ApplicationScoped
+    public static class Supplier implements MarshallerSupplier {
```
tomas-langer, line 68: The name `Supplier` conflicts with JDK's supplier. I think this should be named differently.

grpc/core/src/main/java/io/helidon/grpc/core/JsonbMarshaller.java
aseovic: I made it package private instead

grpc/core/src/main/java/io/helidon/grpc/core/JsonbMarshaller.java
aseovic: It's an inner class, it's never used directly (only via `ServiceLoader`), and we already have a precedent in `JavaMarshaller.Supplier`, so it's probably best to leave it as is. Nobody will ever import it...

PR #1851, opened 2020-05-22
gRPC client API improvements (2.0)

aseovic:
This PR has two related things
1. The same set of client API improvements that make it easier to use as in #1850
2. Renamed `@RpcService`, `@RpcMethod`, and `@GrpcServiceProxy` annotations to `@Grpc`, `@GrpcMethod` and `@GrpcProxy`, to make them consistent with other annotations and/or less verbose.

 Current annotations just don't look right when combined:
```java
       @RpcService(name = "FooService")
       @GrpcChannel(name = "foo")
       public interface FooServiceClient {
           ...
       };
```
This PR changes the above to:
```java
       @Grpc(name = "FooService")
       @GrpcChannel(name = "foo")
       public interface FooServiceClient {
           ...
       };
```
which reads better and makes a lot more sense long term.

The new name also better describes what the thing is -- it is a gRPC service, not a generic RPC service.

This is a breaking change from 1.x, but considering that gRPC is experimental in 1.x and that 2.0 is a major release, this seems like the right time to make this change.


tomas-langer:
Merged changelog.md manually.

PR #1861, opened 2020-05-25
Service loader added to MediaContext

Verdent:


media/common/src/main/java/io/helidon/media/common/MediaContext.java
```diff
@@ -112,105 +140,202 @@ private Builder() {
...
-            writerContext.registerWriter(writer);
+        public Builder includeStackTraces(boolean includeStackTraces) {
+            servicesConfig.computeIfAbsent(DEFAULTS_NAME, k -> new HashMap<>())
+                    .put(DEFAULTS_INCLUDE_STACK_TRACES, Boolean.toString(includeStackTraces));
             return this;
         }
 
         /**
-         * Whether defaults should be included.
+         * Whether system loader should be used to load {@link MediaSupportProvider} from classpath.
```
tomas-langer, line 253: please do not use `from classpath` anywhere. Use from `Java Service Loader`

media/common/src/main/java/io/helidon/media/common/MediaContext.java
```diff
@@ -112,105 +140,202 @@ private Builder() {
...
          *
-         * @param registerDefaults register defaults
+         * @param discoverServices use system loader
```
tomas-langer, line 255: Similar - it is not "system loader", it is "Java Service Loader"

media/common/src/main/java/io/helidon/media/common/MediaContext.java
Verdent: updated

media/common/src/main/java/io/helidon/media/common/MediaContext.java
Verdent: updated

PR #1877, opened 2020-05-27
DataChunk ByteBuffer array

### LINKED ISSUE #1876: DataChunk as iterable of ByteBuffer
Make `DataChunk` an `Iterable<ByteBuffer>` in order to allow composite buffer implementations, similar to https://netty.io/4.0/api/io/netty/buffer/CompositeByteBuf.html

I.e we replace `.data()` with the `Iterable` implementation.

romain-grecourt:
Update DataChunk to hold an array of ByteBuffer instead of a single ByteBuffer.

Fixes #1876


media/common/src/main/java/io/helidon/media/common/DataChunkInputStream.java
```diff
@@ -132,32 +132,38 @@ public int read(byte[] buf, int off, int len) throws IOException {
...
+                off += currentLen;
+                count += currentLen;
+
+                // Chunk is consumed entirely - release the chunk, and prefetch a new chunk; do not
+                // wait for it to arrive - the next read may have to wait less.
+                //
+                // Assert: it is safe to request new chunks eagerly - there is no mechanism
+                // to push back unconsumed data, so we can assume we own all the chunks,
+                // consumed and unconsumed.
+                if (i == currentBuffers.length - 1 && currentLen == rem) {
```
spericas, line 160: What would happen with this condition if the last `ByteBuffer` in the chunk returns 0 and continues in line 144?

romain-grecourt: Good point. I will make another pass at this and add a unit test.

romain-grecourt: @spericas, I've made changes, please take a look and let me know.

spericas: @romain-grecourt I'll take a closer look. At first glance, the code is still a bit odd; it seems that the `DataChunk` abstraction needs to be improved so we can call `get(byte[], int off, int len)` and so that it can iterate over its `ByteBuffer`'s, copy the data and also remember which buffer it is reading from. 
Your code is always starting from the first buffer on each call to `read(...)`, even though buffers may have been consumed in an earlier call. If the `DataChunk` abstraction is enhanced, then`DataChunkInputStream` shouldn't need to change too much.

PR #1883, opened 2020-05-27
ResponseCloser now supports Single

Verdent:


webclient/webclient/src/main/java/io/helidon/webclient/NettyClientHandler.java
```diff
@@ -273,27 +279,44 @@ public int emit(final ByteBuf buf) {
...
 
-        ChannelFuture close() {
-            if (closed.compareAndSet(false, true)) {
+        /**
+         * Asynchronous close method.
+         *
+         * @return single of the closing process
+         */
+        Single<Void> close() {
+            if (singleClosed.compareAndSet(null, Single.from(cf, true))) {
```
tomas-langer, line 302: This approach is not correct.
1. you will create a new `Single` every time method `close` is called
2. you should create a new single for each response
3. isClosed may use the `cf.isDone` rather than this strange construct

PR #1888, opened 2020-05-27
Deprecated Multi#from

danielkec:


romain-grecourt:
I think you should go ahead and remove these API, RC1 is really close ; we shouldn't be deprecating APIs at this stage ; @tomas-langer WDYT ?

danielkec:
@romain-grecourt @tomas-langer If we want to align Multi with the Helidon fluent api, we should be careful not to discourage users coming in with previous knowledge of other reactive libraries. I would like to leave a breadcrumbs for them in the shape of intentionally deprecated `from` methods, so they can find they way to our `create` attitude more easily. Does it make sense to have such deprecated methods without actually planning to remove them?

tomas-langer:
I think this approach is OK. And if we have big enough traction, we can eventually remove them (in 3.0, or 4.0...)

PR #1890, opened 2020-05-28
Kafka specific message

danielkec:
Kafka specific message instead of wrapping ConsumerRecord as payload

- support for Kafka headers

### DIFF
--- a/messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
+++ b/messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
@@ -16,6 +16,7 @@
 
 package io.helidon.messaging.connectors.kafka;
 
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
@@ -47,6 +48,7 @@ class KafkaConsumerMessage<K, V> implements KafkaMessage<K, V> {
      *                             the commit in Kafka. Applies only if autoCommit is false.
      */
     KafkaConsumerMessage(ConsumerRecord<K, V> consumerRecord, CompletableFuture<Void> kafkaCommit, long millisWaitingTimeout) {
+        Objects.requireNonNull(consumerRecord);
         this.consumerRecord = consumerRecord;
         this.kafkaCommit = kafkaCommit;
         this.millisWaitingTimeout = millisWaitingTimeout;
@@ -74,7 +76,7 @@ class KafkaConsumerMessage<K, V> implements KafkaMessage<K, V> {
 
     @Override
     public Optional<ConsumerRecord<K, V>> getConsumerRecord() {
-        return Optional.ofNullable(this.consumerRecord);
+        return Optional.of(this.consumerRecord);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public Optional<Long> getOffset() {
+        return getConsumerRecord().map(ConsumerRecord::offset);
+    }
+
+    @Override
+    public Headers getHeaders() {
+        return this.consumerRecord.headers();
```
tomas-langer, line 72: as other methods show that `consumerRecord` is nullable, this would end with a NPE

messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Override
+    public Optional<K> getKey() {
+        return getConsumerRecord().map(ConsumerRecord::key);
+    }
+
+    @Override
+    public V getPayload() {
+        return this.consumerRecord.value();
```
tomas-langer, line 87: dtto - nullable `consumerRecord`

messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
```diff
@@ -0,0 +1,118 @@
...
+        return kafkaCommit.orTimeout(millisWaitingTimeout, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public <C> C unwrap(Class<C> unwrapType) {
+        if (consumerRecord.getClass().isAssignableFrom(unwrapType)) {
+            return (C) consumerRecord;
+        } else {
+            throw new IllegalArgumentException("Can't unwrap to " + unwrapType.getName());
```
tomas-langer, line 102: Maybe the exception should contain the `consumerRecord.getClass().getName()` to simplify troubleshooting

messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
danielkec: `KafkaConsumerMessage` can be created only by us from `ConsumerRecord`, optional on the interface is ment for `KafkaProducerMessage` which can be created by user when he uses `KafkaMessage#of`

danielkec: added null check to constructor, `this.consumerRecord` is final

messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
danielkec: fixed

messaging/kafka/src/main/java/io/helidon/messaging/connectors/kafka/KafkaConsumerMessage.java
danielkec: fixed

PR #1892, opened 2020-05-28
Removed deprecations

### LINKED ISSUE #1873: Remove deprecations
Remove code deprecated in 1.x

tomas-langer:
Each commit is a single task, except for the last one that updates multiple messages (+ resolves a few missed things).

Resolves #1873


webclient/webclient/src/main/java/io/helidon/webclient/WebClient.java
```diff
@@ -283,7 +283,7 @@ public Builder config(Config config) {
          *
          * @param amount amount of time
          * @param unit   time unit
-         * @deprecated see {@link WebClient.Builder#connectTimeout(long, TimeUnit)}
+         * @deprecated use {@link WebClient.Builder#connectTimeout(long, TimeUnit)}, this method will be removed
          * @return updated builder instance
          */
         @Deprecated
```
ljnelson, line 289: Maybe since the method will be removed you could use `@Deprecated(forRemoval = true)`?

tomas-langer: I will create a follow up for that.

tomas-langer: #1901

PR #1900, opened 2020-05-29
Use EmittingPublisher in OutputStreamPublisher to remove busy waiting

danielkec:
Partially solving #1741, fixing race-condition in `EmittingPublisher`

Added to `IoMulti` api
```java
        StringBuilder result = new StringBuilder();

        MultiFromOutputStream osMulti = IoMulti.createOutputStream();

        Single<Void> multiFuture = osMulti
                .map(ByteBuffer::array)
                .map(String::new)
                .forEach(result::append);

        PrintWriter printer = new PrintWriter(osMulti);
        printer.println("test1");
        printer.println("test2");
        printer.println("test3");
        printer.close();

        multiFuture.await();
        assertThat(result.toString(), is(equalTo("test1\ntest2\ntest3\n")));
```

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/BufferedEmittingPublisher.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/BufferedEmittingPublisher.java
@@ -237,12 +237,12 @@ public class BufferedEmittingPublisher<T> implements Flow.Publisher<T> {
 
     private int emitOrBuffer(T item) {
         for (;;) {
-            try {
                 if (emitting.getAndSet(true)) {
                     // race against parallel emits
                     // only those can add to buffer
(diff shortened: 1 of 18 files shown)


danielkec:
There was a lock in `write` method but I would expect `java.io.OutputStream#write(byte[], int, int)` should be always presumed as not thread safe(rule 1.3 should be assured by `EmittingPublisher` anyway)
```java
          synchronized (invocationLock) {
                if (subscriber.isClosed()) {
                    throw new IOException("Output stream already closed.");
                }

                sub.onNext(createBuffer(buffer, offset, length));
            }
```

olotenko:
The changes to `BufferedEmittingPublisher` are racy.

PR #1905, opened 2020-05-29
Media support methods

Verdent:


media/jackson/src/main/java/io/helidon/media/jackson/JacksonSupport.java
```diff
@@ -33,94 +34,114 @@
...
-     * @return a new {@link JacksonSupport}
-     *
-     * @exception NullPointerException if {@code objectMapper}
-     * is {@code null}
+     * @return Jackson reader instance
      */
-    public static JacksonSupport create(ObjectMapper objectMapper) {
-        Objects.requireNonNull(objectMapper);
-        return new JacksonSupport(objectMapper);
+    public MessageBodyReader<Object> readerInstance() {
```
tomas-langer, line 124: Inconsistency:
`readerInstance()` against `newWriter()` methods

The `readerInstance()` method is worse, as it prescribes implementation - it is expected to return the same instance every time.

The `newWriter()` method clearly says that it may be a different instance each time (and yet if implementation wants, it can return the same one).

I would stay with the `newXXX` methods

Verdent: Actually this method is not described by any interface, so it is basically up to every support class how it chooses to name it or even decides to have them. Since we are returning every time the same instance of the reader/writer, I think `readerInstance()` is actually better name here.

Verdent:
Fixes #1907

PR #1908, opened 2020-05-31
EmittingPublisher race condition fix

danielkec:
Separated bug fix from #1900


danielkec:
Fixed in #1900

PR #1912, opened 2020-06-01
Container startup issue fixed.

### LINKED ISSUE #1446: Container starts even if exception is thrown during initialization
# Helidon container starts even if exception is thrown during initialization

## Environment Details
* Helidon Version: master 2.0.0-SNAPSHOT
* Helidon MP
* JDK version: 11
* OS: Ubuntu

----------

## Problem Description
Helidon container starts even if exception is thrown during initialization(after deployment)

## Steps to reproduce
```java
void init(@Observes @Priority(PLATFORM_AFTER + 101) @Initialized(ApplicationScoped.class) Object e, BeanManager bm) {
   throw new RuntimeException("Survive this!");
}
```
## Reproducer

```java
@Test
    void testCleanDeath() {
        assertThrows(IllegalStateException.class, CDI::current);
        assertThrows(RuntimeException.class, () -> startCdiContainer(Map.of(), Set.of(TestBean.class)));
        assertThrows(IllegalStateException.class, CDI::current);
    }

    private static SeContainer startCdiContainer(Map<String, String> p, Set<Class<?>> beanClasses) {
        Config config = ConfigProviderResolver.instance().getBuilder()
                .withSources(MpConfigSources.create(p),
                        MpConfigSources.create(Map.of("mp.initializer.allow", "true")))
                .build();

        ConfigProviderResolver.instance()
                .registerConfig(config, Thread.currentThread().getContextClassLoader());

        final SeContainerInitializer initializer = SeContainerInitializer.newInstance();
        initializer.addBeanClasses(beanClasses.toArray(new Class<?>[0]));
        return initializer.initialize();
    }

    @ApplicationScoped
    static class TestBean {
        void init(@Observes @Priority(PLATFORM_AFTER + 101) @Initialized(ApplicationScoped.class) Object e, BeanManager bm) {
            throw new RuntimeException("Survive this!");
        }
    }
```

tomas-langer:

Resolves #1446

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
@@ -28,6 +28,7 @@ import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -49,6 +50,7 @@ import io.helidon.common.Prioritized;
 import io.helidon.common.configurable.ServerThreadPoolSupplier;
 import io.helidon.common.http.Http;
 import io.helidon.config.Config;
+import io.helidon.microprofile.cdi.BuildTimeStart;
 import io.helidon.microprofile.cdi.RuntimeStart;
 import io.helidon.webserver.Routing;
 import io.helidon.webserver.Service;
@@ -70,6 +72,7 @@ public class ServerCdiExtension implements Extension {
     }
 
     private static final Logger LOGGER = Logger.getLogger(ServerCdiExtension.class.getName());
+    private static final AtomicBoolean IN_PROGRESS_OR_RUNNING = new AtomicBoolean();
 
     // build time
     private WebServer.Builder serverBuilder = WebServer.builder()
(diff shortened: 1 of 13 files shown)


microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -216,6 +228,17 @@ private void registerClasspathStaticContent(Config config) {
     }
 
     private void stopServer(@Observes @Priority(PLATFORM_BEFORE) @BeforeDestroyed(ApplicationScoped.class) Object event) {
+        boolean isRunning = IN_PROGRESS_OR_RUNNING.get();
+        try {
+            doStop(event);
+        } finally {
+            if (isRunning) {
+                IN_PROGRESS_OR_RUNNING.set(false);
+            }
```
tjquinno, line 237: This logic sets `IN_PROGRESS_OR_RUNNING` to false only if it was true at the start of the method. Could there be a race condition here in which an (incorrect) second attempt to start the server concurrently occurred after `isRunning` is assigned `false` (and that parallel start-up would set `IN_PROGRESS_OR_RUNNING` to `true`), the `doStop` actually does the shutdown, but because `isRunning` is not true then `IN_PROGRESS_OR_RUNNING` is not set to `false`? So the server would be stopped but `IN_PROGRESS_OR_RUNNING` incorrectly says `true`. 

Maybe this cannot happen in practice because of other reasons?

tomas-langer: There was a problem with this. Fixing.

tomas-langer: Fixing this uncovered other issues with correct shutdown. Fixed them as well.

PR #1916, opened 2020-06-01
Tracing performance optimization.

### LINKED ISSUE #1915: Tracing causes performance regression even when not enabled
The tracing framework for webserver is used even when disabled. It should not be triggered at all when tracing is disabled.

tomas-langer:

Resolves #1915 

Tracing now skips creation of spans when the tracer is a Noop.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
```diff
@@ -224,13 +226,24 @@ public void accept(ServerRequest req, ServerResponse res) {
...
+        private final AtomicBoolean checkedIfShouldTrace = new AtomicBoolean();
+        private volatile boolean shouldTrace = true;
 
         RequestSpanHandler() {
         }
 
         @Override
         public void accept(ServerRequest req, ServerResponse res) {
-            doAccept(req, res);
+            if (checkedIfShouldTrace.compareAndSet(false, true)) {
```
danielkec, line 237: In case there is no transition of `shouldTrace` back to `true` we can save few ops by 
```java
if (shouldTrace && checkedIfShouldTrace.compareAndSet(false, true)) {
```

danielkec: But its almost un-measurable

tomas-langer: Fixed

PR #1921, opened 2020-06-02
Media lazy accepted types parsing

danielkec:


common/common/src/main/java/io/helidon/common/LazyValue.java
```diff
@@ -47,7 +47,7 @@
      * @return a lazy value that will always return the value provided
      */
     static <T> LazyValue<T> create(T value) {
-        return () -> value;
+        return new LazyValueImpl<>(value);
```
tomas-langer, line 50: This constructor is good.

common/common/src/main/java/io/helidon/common/LazyValueImpl.java
```diff
@@ -16,42 +16,28 @@
 
 package io.helidon.common;
 
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.atomic.AtomicReference;
```
tomas-langer, line 19: Please revert the state to original implementation + add the new constructor that sets the `value` and sets `loaded` to true.

common/common/src/main/java/io/helidon/common/LazyValue.java
danielkec: Deffered to #1924

common/common/src/main/java/io/helidon/common/LazyValueImpl.java
danielkec: reverted

Issue #1588, opened 2020-03-26
Update built-in handlers to not be sensitive to media support setup


romain-grecourt:
The use of media operators (readers and writers) is sensitive to their registration order.
Our built-in services such as static content support, metrics or health-checks should not depend on the users setup of the media support.

We need to track all the built-in handlers and update them to not be sensitive to the media support ordering setup by the user:

- [x] static content
- [x] metrics
- [x] health checks
- [ ] others?

We should also come-up with the proper way of doing this.

E.g. for a handler that needs to writer a `Path` to a response:

```java
response.send(response
    .writerContext()
    .marshall(Single.just(path), PathBodyWriter.class, GenericType.create(Path.class), null));
```

The above (what we have in 2.0.0-M2) is rather ugly, we should update it something like this:

```java
response.send((ctx) -> ctx.marshall(Single.just(path), PathBodyWriter.class, GenericType.create(Path.class)));
```

This approach is not order dependent but still relies on the PathBodyWriter to be registered in the media support. This is registered by default when the media support is not provided explicitly to the webserver.

However one can setup the media support without the default operators and pass it to the webserver, in this case the code above would not work:

```java
WebServer.builder(Routing.builder().register(StaticContentSupport.create("/WEB")))
    .mediaSupport(MediaSupport.builder().build())
    .build();
```

We need to decide if this is acceptable or if our built-in handlers should be completely standalone WRT to media support.

Instead of using the underlying operator context to select the operator by class name, we can produce the Publisher manually.

E.g. for the same handle that needs to send a `Path` to a response:

```java
FileChannel fc = FileChannel.open(path, StandardOpenOption.READ);
response.send(new ReadableByteChannelPublisher(fc, RetrySchema.linear(0, 10, 250)));
``` 

One of the drawback of not using an operator is that the underlying publisher implementation needs to be public like `ReadableByteChannelPublisher` above, and that the headers related work performed by the operator has to be done by hand.

One can also use the operator manually:

```java
PathBodyWriter writer = PathBodyWriter.create(); // can be static
response.send(writer.write(Single.just(path), GenericType.create(path), response.writerContext()));
```

This approach seems best as it does not rely on the `MediaSupport` setup and prevent code duplication and also let the operator do the header work if needed.

romain-grecourt:
@Verdent @ljnelson @tomas-langer You guys might be familiar with this issue, let me know if you have comments / thoughts about this.

tomas-langer:
[...]

PR #1922, opened 2020-06-02
Static content handlers now use explicit writers.

tomas-langer:

Related to #1588


tomas-langer:
@romain-grecourt I think that static content, metrics and health do not need to mark the used content writer in tracing (as long as we know the endpoint was handled by metrics, health, or static content), as these are hardcoded (and cannot be changed by user).
So for the built-in handlers, we can safely merge the current solution.

PR #1924, opened 2020-06-02
ReadableByteChannelPublisher executor leak

### LINKED ISSUE #1835: ReadableByteChannelPublisher keeps executor service forever (and scheduled)
1. The executor service is not daemon, so it keeps the JVM running (line 194)
2. The executor service is not shut down, so it stays around forever - this is a memory leak

We should:
 - provide option to use an external executor service
 - provide a thread factory so the threads are daemon
 - if our own instance is used, shut it down once the last bytes are written

tomas-langer:
And related fixes.

Resolves #1835

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/IoMulti.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/IoMulti.java
@@ -20,12 +20,16 @@ package io.helidon.common.reactive;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.ByteBuffer;
+import java.nio.channels.ReadableByteChannel;
 import java.util.Objects;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 
 import io.helidon.common.Builder;
+import io.helidon.common.LazyValue;
 
 /**
  * Create reactive stream from standard IO resources.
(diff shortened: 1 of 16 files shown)


common/reactive/src/main/java/io/helidon/common/reactive/IoMulti.java
```diff
@@ -67,23 +103,175 @@ static OutputStreamMultiBuilder builderOutputStream() {
...
+     * @param byteChannel readable byte channel with data
+     * @return fluent API builder to configure additional details
+     */
+    static MultiFromByteChannelBuilder multiFromByteChannelBuilder(ReadableByteChannel byteChannel) {
+        return new MultiFromByteChannelBuilder(Objects.requireNonNull(byteChannel));
+    }
+
+    /**
+     * Fluent API builder for creating a {@link io.helidon.common.reactive.Multi} from an
+     * {@link java.io.InputStream}.
```
danielkec, line 179: {@link ReadableByteChannel}.

tomas-langer: Not sure I understand the comment? 
In general it is preferable to use fully qualified class names in Javadoc, so you do not import a class only referenced in Javadoc

danielkec: Its just its on `MultiFromByteChannelBuilder` so it should be from `ReadableByteChannel` not from `InputStream`

tomas-langer: Fixed.

PR #1928, opened 2020-06-04
BufferedEmittingPublisher draining race condition fix

danielkec:
Fix race condition found by @olotenko


danielkec:
Both race conditions should be fixed. CPU burning by TAS locks has been addressed by using TTAS with exponential backoff, as the assumption is that reactive streams shouldn't block or obstruct [§2.2](https://github.com/reactive-streams/reactive-streams-jvm#2.2) Lost performance has been compensated by unbounded fast-path in  `BufferedEmittingPublisher`:
```

Benchmark                       Mode  Cnt   Score   Error  Units
EPTest.testEmitterReqMillion   thrpt   20  34.236 ± 0.235  ops/s
EPTest.testEmitterReqOneByOne  thrpt   20  17.118 ± 0.329  ops/s
EPTest.testEmitterUnbounded    thrpt   20  47.536 ± 0.123  ops/s
EPTest.testOTPReqMillion       thrpt   20  26.610 ± 1.026  ops/s
EPTest.testOTPReqOneByOne      thrpt   20  14.107 ± 0.251  ops/s
EPTest.testOTPUnbounded        thrpt   20  26.338 ± 0.051  ops/s
```

PR #1932, opened 2020-06-04
Jersey connector to Helidon WebClient

jansupol:
Signed-off-by: Jan Supol <jan.supol@oracle.com>

### DIFF
--- a/jersey/connector/pom.xml
+++ b/jersey/connector/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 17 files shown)


jersey/connector/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.github.tomakehurst</groupId>
+            <artifactId>wiremock</artifactId>
+            <version>2.21.0</version>
+            <type>pom</type>
```
romain-grecourt, line 40: - Make sure all versions are declared in one of the appropriate parent poms.
- Also, why the pom type dependency ?
- You may want to consider not making this PR as draft, otherwise you won't have feedback from the pipelines.

jansupol: Fixed

PR #1934, opened 2020-06-05
WebSocket Extensions

dansiviter:
Few notes:
* Extensions (such as `org.glassfish.tyrus.ext.extension.deflate.PerMessageDeflateExtension`) that are in archives that are not 'bean archives' are not added as CDI ignores them. Is there a mechanism to override this or an acceptable limitation forcing workarounds? I thought about using synthetic bean archives, but the classpath would still need to be scanned to do this and not aware of any mechanism for this.
* ~The only build available with the fixes required in Tyrus is 2.0.0+ which also mandates `jakarta.websocket.*` packages which means quite a few more changes.~ No longer an issue,
* ~The both Tyrus and WebSocket API are not at release versions yet. I'm assuming this would prevent merging.~ No longer an issue.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -109,7 +109,7 @@
         <version.lib.snakeyaml>1.24</version.lib.snakeyaml>
         <version.lib.transaction-api>1.3.3</version.lib.transaction-api>
         <version.lib.typesafe-config>1.4.0</version.lib.typesafe-config>
-        <version.lib.tyrus>1.15</version.lib.tyrus>
+        <version.lib.tyrus>1.17</version.lib.tyrus>
         <version.lib.ucp>${version.lib.ojdbc8}</version.lib.ucp>
         <version.lib.validation-api>2.0.2</version.lib.validation-api>
         <version.lib.websockets-api>1.1.2</version.lib.websockets-api>
--- a/microprofile/websocket/src/main/java/io/helidon/microprofile/tyrus/WebSocketApplication.java
+++ b/microprofile/websocket/src/main/java/io/helidon/microprofile/tyrus/WebSocketApplication.java
@@ -22,6 +22,7 @@ import java.util.Set;
 import java.util.logging.Logger;
 
 import javax.websocket.Endpoint;
+import javax.websocket.Extension;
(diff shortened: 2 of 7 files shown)


spericas:
> * The both Tyrus and WebSocket API are not at release versions yet. I'm assuming this would prevent merging.

 This is a problem. What do you think @barchetta?

barchetta:
> > * The both Tyrus and WebSocket API are not at release versions yet. I'm assuming this would prevent merging.
> 
> This is a problem. What do you think @barchetta?

I agree that this PR should stay in WIP for now because:
1. It moves to `jakarta.websocket`. Moving to `jakarta` packages should be a coordinated decision across the project.
2. We prefer not to depend on milestone builds if at all possible.

So this PR might need to stay in the wings until we make some decisions concerning moving to Jakarta EE 9.

romain-grecourt:
Good question. This is debatable :D ; I'd say it should be before `io.helidon`

dansiviter:
It appears 1.17 has both the changes I submitted and still using `javax.websocket` so I've updated the MR. I guess this won't be in time for v2 release though.

dansiviter:
Any idea how to re-trigger CI?

dansiviter:
I'm hoping this will be merged soon but the CI never seems to trigger when I push. I tried using `/trigger` but that didn't seem to work.

romain-grecourt:
Basically accounts that do not have access to the repository are not "trusted" and their pull requests need to  be manually triggered by "trusted" accounts.

romain-grecourt:
you have checkstyle errors, you can run the check locally: `mvn validate -Pcheckstyle` from `./webserver/tyrus/`

dansiviter:
Thanks @romain-grecourt for the info. I've updated the MR with the checkstyle updates.

spericas (approved):
LGTM

romain-grecourt:
@dansiviter we need to do some checks for the Tyrus version change before we merge this.

barchetta (approved):

barchetta:
@dansiviter thanks for the contribution and apologies this took so long to merge!

PR #1940, opened 2020-06-06
Lazy list for even lazier acceptedTypes parsing

danielkec:
Fixes https://gitlab-odx.oracledx.com/helidon/helidon/-/issues/14

### DIFF
--- a/media/common/src/main/java/io/helidon/media/common/MessageBodyWriterContext.java
+++ b/media/common/src/main/java/io/helidon/media/common/MessageBodyWriterContext.java
@@ -27,7 +27,6 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 
 import io.helidon.common.GenericType;
-import io.helidon.common.LazyValue;
 import io.helidon.common.http.DataChunk;
 import io.helidon.common.http.Http;
 import io.helidon.common.http.MediaType;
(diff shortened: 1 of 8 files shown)


media/common/src/main/java/io/helidon/media/common/MessageBodyWriterContext.java
```diff
@@ -118,7 +117,7 @@ private MessageBodyWriterContext(MessageBodyWriterContext writerContext, Paramet
         this.headers = headers;
         this.writers = new MessageBodyOperators<>(writerContext.writers);
         this.swriters = new MessageBodyOperators<>(writerContext.swriters);
-        this.acceptedTypes = writerContext.acceptedTypes;
+        this.acceptedTypes = List.copyOf(writerContext.acceptedTypes);
```
tomas-langer, line 120: As far as I can see, `List.copyOf` actually gets all the elements in the list. So the lazyness would be gone in this case.

danielkec: I am not sure if we do need the copy, it's  unmodifiable list all the time, but I could have missed something WDYT @romain-grecourt ?

romain-grecourt: Right, this list shouldn't need to be copied.

danielkec: fixed

PR #1951, opened 2020-06-09
Minor gRPC fixes (2.0)

aseovic:
Minor gRPC metrics cleanup and fix for Channel injection


grpc/metrics/src/main/java/io/helidon/grpc/metrics/GrpcMetrics.java
```diff
@@ -70,6 +66,19 @@
...
+        HelidonFeatures.register("gRPC Server", "Metrics");
+        HelidonFeatures.register("gRPC Client", "Metrics");
+
+        VENDOR_REGISTRY.meter(org.eclipse.microprofile.metrics.Metadata.builder()
+                .withName("grpc.requests.meter")
+                .withDisplayName("Meter for overall gRPC requests")
+                .withDescription("Each gRPC request will mark the meter to see overall throughput")
+                .withType(MetricType.METERED)
+                .withUnit(MetricUnits.NONE)
+                .build());
```
tjquinno, line 79: I see that this takes the place of the meter that used to be registered in `MetricsSupport`. There is no replacement here for the counter that `MetricsSupport` used to register. Is that intentional?

aseovic: Yes -- meter already has a counter, so having a separate counter is a bit pointless.

PR #1982, opened 2020-06-11
New DB archetype for SE

spericas:
Based on Pokemons and similar to the MP one but using dbclient and JDBC.


archetypes/database-se/src/main/resources/src/main/java/__pkg__/Main.java.mustache
```diff
@@ -0,0 +1,96 @@
...
+                .build();
+
+        // Start the server and print some info
+        server.start()
+                .thenAccept(ws -> System.out.println("WEB server is up! http://localhost:" + ws.port() + "/"));
+
+        // Server threads are not daemon
+        server.whenShutdown()
+                .thenRun(() -> System.out.println("WEB server is DOWN. Good bye!"));
+
```
tjquinno, line 66: Other archetypes seem to use a different pattern:
https://github.com/oracle/helidon/blob/6e4e3aa3e749c9280eea06f927004cb7e4cfd940/archetypes/bare-se/src/main/resources/src/main/java/__pkg__/Main.java.mustache#L55-L68

We should probably be consistent unless it makes sense otherwise.

PR #1988, opened 2020-06-11
MP native image no longer misses bean types for proxies.

### LINKED ISSUE #1983: MP Native image error
`WeldFeature` sometimes ignores bean types, if a bean class defines more than one bean type.
Manifested as an intermittent native image runtime failure.

tomas-langer:

Fixes #1983

### DIFF
--- a/integrations/graal/mp-native-image-extension/src/main/java/io/helidon/integrations/graal/mp/nativeimage/extension/WeldFeature.java
+++ b/integrations/graal/mp-native-image-extension/src/main/java/io/helidon/integrations/graal/mp/nativeimage/extension/WeldFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


integrations/graal/mp-native-image-extension/src/main/java/io/helidon/integrations/graal/mp/nativeimage/extension/WeldFeature.java
```diff
@@ -168,11 +168,15 @@ private void initializeProxy(DuringSetupAccess access,
...
                               ClientProxyProvider cpp,
-                              Set<Class<?>> processed,
+                              Set<BeanId> processed,
                               Collection<Bean<?>> beans) {
         for (Bean<?> bean : beans) {
-            Class<?> clazz = bean.getBeanClass();
-            if (!processed.add(clazz)) {
+            Set<Type> beanTypes = bean.getTypes();
+
+            BeanId id = new BeanId(bean.getBeanClass(), beanTypes);
```
ljnelson, line 176: Depending on how proper you want to be, you may want to see if the `bean` is `PassivationCapable` and if so use the return value of its `getId()` method.

tomas-langer: I think we can live with a certain amount of improper, unless this would fail for some cases.

PR #2009, opened 2020-06-15
1.4: Include NOTICE file in all jar files

barchetta:


NOTICE.txt
```diff
@@ -2,7 +2,7 @@
                            Project Helidon
                            ===============
 
-Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
```
tjquinno, line 5: We can (are supposed to?) remove "All rights reserved" from the copyright notices now.

barchetta: Fixed.

PR #2015, opened 2020-06-16
DbClient example: wrong table reference fix

tomas-langer:
Renamed table types as it collides with h2 system table
Added h2 connection options (commented out)


tomas-langer:
Will add to db client update PR (native image changes)

PR #2019, opened 2020-06-16
Using SafeConstructor with YAML parsing.

tomas-langer:

Even though the safe constructor was used implicitly (when map was used as a type), we were getting some false static analysis failures.

This makes the use explicit.


JLLeitschuh:
Associated with CVE-2022-21404:
https://www.websec.ca/publication/Blog/CVE-2022-21404-Another-story-of-developers-fixing-vulnerabilities-unknowingly-because-of-CodeQL

PR #2032, opened 2020-06-17
SE Messaging connector bad prefix within MP config

danielkec:
Align with MP config


messaging/messaging/src/main/java/io/helidon/messaging/ConnectorConfigHelper.java
```diff
@@ -34,16 +34,14 @@ private ConnectorConfigHelper() {
     }
 
     static ConfigValue<String> getIncomingConnectorName(Config config, String channelName) {
-        //Looks suspicious but incoming connector configured for outgoing channel is ok
-        return config.get(ConnectorFactory.OUTGOING_PREFIX)
+        return config.get(ConnectorFactory.INCOMING_PREFIX)
                 .get(channelName)
```
danielkec, line 38: Looked suspicious and was!

PR #2034, opened 2020-06-17
Add trademark notices for Pokemon

barchetta:
This addresses example readmes and archetype source files.


barchetta:
> "trademarks" might need to be "registered trademarks"; IANAL

Me neither! I copied the text from pokemon.com, so I'll leave it as is.

PR #2041, opened 2020-06-18
add empty javadoc and empty sources to integrations/db/hk2

romain-grecourt:


ljnelson:
(H2, right? not HK2?)

PR #2060, opened 2020-06-19
SE Config can be created from MP Config 

### LINKED ISSUE #2059: MP to SE Config issue
When I cast MP Config to SE Config, environment variables are not correctly mapped to key.

For example `TRACING_HOST` is configured as an environment variable.
When I call `Config.value("tracing.host", String.class)` on MP config, it works as expected
When I call `((io.helidon.config.Config)config).get("tracing.host").asString().get()` it does not.

This is because SE config uses MP Config as a single config source in this case, and does not see environment variables.

tomas-langer:
Environment variables were not honored correctly - now we use MP config as the implementation and we have an SE config implementation fronting it when creating SE config from MP Config.

Resolves #2059

### DIFF
--- a/config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
+++ b/config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
@@ -30,6 +30,7 @@ import io.helidon.config.spi.ConfigMapper;
 import io.helidon.config.spi.ConfigMapperProvider;
 
 import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.config.spi.ConfigSource;
 import org.junit.jupiter.api.BeforeAll;
@@ -272,6 +273,30 @@ public class MpConfigTest {
 
         public static void reset() {
             counter.set(0);
+        }
+    }
+
+    @Test
+    void testEnvVar() {
+        ConfigProviderResolver instance = ConfigProviderResolver.instance();
+        ClassLoader myCl = Thread.currentThread().getContextClassLoader();
+        Config current = ConfigProvider.getConfig(myCl);
+
+        try {
+            instance.registerConfig(instance.getBuilder()
+                                            .withSources(MpConfigSources.environmentVariables())
+                                            .build(),
+                                    myCl);
+            Config myConfig = instance.getConfig(myCl);
+            // this must not throw an exception - path should be on any environment
+            // and the MP env var processing should make it available
+            String path = myConfig.getValue("path", String.class);
+
+            io.helidon.config.Config helidonConfig = (io.helidon.config.Config) myConfig;
+            // should work if we use it as SE as well
+            helidonConfig.get("path").asString().get();
+        } finally {
+            instance.registerConfig(current, myCl);
         }
     }
 }
--- a/config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
+++ b/config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
@@ -290,11 +290,13 @@ public class MpConfigTest {
             Config myConfig = instance.getConfig(myCl);
             // this must not throw an exception - path should be on any environment
             // and the MP env var processing should make it available
-            String path = myConfig.getValue("path", String.class);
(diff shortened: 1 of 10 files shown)


aseovic:
Looks like there are a few test failures and a SpotBugs issue to address.

As for the implementation, I understand and agree with the overall approach, but can't really provide much feedback on the actual implementation.

config/config/src/main/java/io/helidon/config/BuilderImpl.java
```diff
@@ -605,6 +584,7 @@ private EmptyConfigHolder() {
                 .disableSystemPropertiesSource()
                 .disableParserServices()
                 .disableFilterServices()
+                .changesExecutor(command -> {})
```
danielkec, line 587: Maybe we could add some `disableMutableSources` or something similar in the future

config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
```diff
@@ -274,5 +275,29 @@ public static void reset() {
...
+                                            .build(),
+                                    myCl);
+            Config myConfig = instance.getConfig(myCl);
+            // this must not throw an exception - path should be on any environment
+            // and the MP env var processing should make it available
+            String path = myConfig.getValue("path", String.class);
+
+            io.helidon.config.Config helidonConfig = (io.helidon.config.Config) myConfig;
+            // should work if we use it as SE as well
+            helidonConfig.get("path").asString().get();
```
danielkec, line 297: Can we assert those resolved PATH values equals with `System.getenv("PATH")`?

config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
```diff
@@ -274,5 +275,29 @@ public static void reset() {
...
+
+        try {
+            instance.registerConfig(instance.getBuilder()
+                                            .withSources(MpConfigSources.environmentVariables())
+                                            .build(),
+                                    myCl);
+            Config myConfig = instance.getConfig(myCl);
+            // this must not throw an exception - path should be on any environment
+            // and the MP env var processing should make it available
+            String path = myConfig.getValue("path", String.class);
```
danielkec, line 293: Failing on windows `java.util.NoSuchElementException: Property "path" is not available in configuration`

danielkec: Because its `Path`

microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AdHocConfigBuilderTest.java
```diff
@@ -76,6 +132,38 @@ void customValueOverride() {
...
+
+        resolver.registerConfig(resolver.getBuilder()
+                                        .withSources(MpConfigSources.create(propMap))
+                                        .build(), null);
+
+        Config config = (Config) resolver.getConfig();
+
+        resolver.registerConfig(resolver.getBuilder()
+                                        .withSources(MpConfigSources.create(propMap2))
+                                        .build(), null);
```
danielkec, line 154: `io.helidon.config.mp.MpConfigProviderResolver#registerConfig` has no defense against `null` and registered config is stored in `io.helidon.config.mp.MpConfigProviderResolver#CONFIGS` map with `null` key, but that Config is not acessible any more because `getConfig` has defense against `null`
```java
    public Config getConfig(ClassLoader loader) {
        if (null == loader) {
            loader = ClassLoader.getSystemClassLoader();
        }

```

config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
tomas-langer: Using an env var configured in pom.xml now.

microprofile/messaging/src/test/java/io/helidon/microprofile/messaging/AdHocConfigBuilderTest.java
tomas-langer: Fixed all instances to work the same - if loader is `null`, use context class loader.

config/config-mp/src/test/java/io/helidon/config/mp/MpConfigTest.java
tomas-langer: Using custom env var, validating content.

PR #2104, opened 2020-06-27
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


romain-grecourt:
Thanks for the suggestion. However, we do already have an internal process for managing dependencies.

PR #2121, opened 2020-07-02
Single add forSingle and exceptionallyAccept

danielkec:
Add terminal operation `forEach` equivalent to Single:
```java
Single.just("test")
                .forSingle(s -> System.out.println(">>>" + s));
>>>test
``` 

Add `exceptionallyAccept` which takes consumer as parameter to `CompletionAwaitable` so there is no need to return something all the time, so instead of:
```java
        Single.error(testException)
                .exceptionally(t -> {
                    t.printStackTrace();
                    return null;
                })
```
Is possible:
```java
        Single.error(testException)
                .exceptionallyAccept(Throwable::printStackTrace);
``` 

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/CompletionAwaitable.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/CompletionAwaitable.java
@@ -295,7 +295,15 @@ public class CompletionAwaitable<T> implements CompletionStage<T>, Awaitable<T> 
     }
 
     /**
-     * {@inheritDoc}
+     * Returns a new CompletionAwaitable that, when this stage completes
+     * exceptionally, is executed with this stage's exception as the
+     * argument to the supplied consumer. Otherwise, if this stage
+     * completes normally, then the returned stage also completes
+     * normally with the same value.
+     *
+     * @param consumer the consumer to invoke if this CompletionAwaitable completed
+     *                 exceptionally
+     * @return the new CompletionAwaitable
      */
     public CompletionAwaitable<T> exceptionallyAccept(final Consumer<Throwable> consumer) {
         return this.handle((item, t) -> {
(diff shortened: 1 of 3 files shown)


common/reactive/src/main/java/io/helidon/common/reactive/CompletionAwaitable.java
```diff
@@ -294,6 +294,18 @@ void setOriginalStage(final Supplier<CompletionStage<T>> originalStage) {
         return new CompletionAwaitable<T>(() -> completionStage, this);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public CompletionAwaitable<T> exceptionallyAccept(final Consumer<Throwable> consumer) {
```
romain-grecourt, line 300: Is this missing a `@Override` annotation ?

common/reactive/src/main/java/io/helidon/common/reactive/Single.java
```diff
@@ -793,4 +803,17 @@ default T get(long timeout, TimeUnit unit) throws InterruptedException, Executio
 
     @Override
     CompletionAwaitable<T> exceptionally(Function<Throwable, ? extends T> fn);
+
+    /**
+     * Returns a new CompletionStage that, when this stage completes
```
tomas-langer, line 808: Seems a copy paste error - probably not a CompletionStage (here, in param and in return tags as well)

common/reactive/src/main/java/io/helidon/common/reactive/CompletionAwaitable.java
danielkec: No this one is ours, but the inherit docs was wrong thx

common/reactive/src/main/java/io/helidon/common/reactive/Single.java
danielkec: Fixed thx

Issue #2107, opened 2020-06-29
DbClient and Webserver - wrong result on JSON-P when library missing


tomas-langer:
## Environment Details
* Helidon Version:  2.0.0
* Helidon SE 
----------

## Problem Description
When the following dependency is not on the classpath
```
<dependency>
        <groupId>io.helidon.dbclient</groupId>
        <artifactId>helidon-dbclient-jsonp</artifactId>
</dependency>
```

and Json processing for WebServer is on the classpath, calling

```
Multi<JsonObject> dbMulti = dbClient.execute(exec -> exec.namedQuery("select-all-pokemons"))
                .map(it -> it.as(JsonObject.class);
response.send(dbMulti, JsonObject.class);
```
Fails and only returns `[`, instead of a usual exception.
This should fail, as DbClient cannot find the mapper to `JsonObject` for rows.

## Reproducer
Go to Pokemon example in dbclient and remove the dependency from pom.xml. Then try to get list of Pokemon.

romain-grecourt:
I'm getting `[` even when the dependency with `helidon-dbclient-jsonp` in the classpath.

romain-grecourt:
`JsonpBodyStreamWriter` is implemented like this:

```java
Single.just(DataChunk.create(ARRAY_JSON_BEGIN_BYTES))
    .onCompleteResumeWith(Multi.create(publisher)
        .map(jsonToChunks)
        .flatMap(it -> {
            if (first.getAndSet(false)) {
                // first record, do not prepend a comma
                return Single.just(it);
            } else {
                // any subsequent record starts with a comma
                return Multi.just(DataChunk.create(COMMA_BYTES), it);
            }
        }))
    .onCompleteResume(DataChunk.create(ARRAY_JSON_END_BYTES));
```

If the upstream publisher holds an error, it is not propagated and it produces a stream that contains just `[`.

```java
Single.just("[")
    .onCompleteResumeWith(Single.error(new Exception("booh!")))
    .onCompleteResume("]")
     .forEach(System.out::println)
// prints '['
```

romain-grecourt:
The example requires a database to be running, if the database is not running the publisher holds an exception:

```java
io.helidon.dbclient.DbClientException: Failed to create a connection to jdbc:oracle:thin:@localhost:1521
```

If the `helidon-dbclient-jsonp` dependency is missing, the publisher holds an exception:

```java
[...]

PR #2129, opened 2020-07-03
Flush streamed datachunk one by one

danielkec:
So items are propagated to the client one by one:
recording


danielkec:
@tomas-langer Not sure what is the performance impact but otherwise is whole stream buffered and delivered whole at once upon completion

danielkec:
Fixes #2107

PR #2137, opened 2020-07-06
Create cli-data for snapshot builds

batsatt:


archetypes/catalog/pom.xml
```diff
@@ -67,4 +67,95 @@
...
+                                </directory>
+                            </directories>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>com.soebes.maven.plugins</groupId>
+                <artifactId>echo-maven-plugin</artifactId>
+                <version>0.4.0</version>
```
romain-grecourt, line 124: This plugin version needs to be managed in the top level pom.

archetypes/catalog/pom.xml
```diff
@@ -67,4 +67,95 @@
...
+                        <echo> |</echo>
+                        <echo> |                      /')    </echo>
+                        <echo> |               ////  /' )'   </echo>
+                        <echo> |              @   \/'  )'    </echo>
+                        <echo> |  { note }   &lt; (  (_...)' </echo>
+                        <echo> |               \      )      </echo>
+                        <echo> |                \,,,,/       </echo>
+                        <echo> |                 _|_         </echo>
+                        <echo> |</echo>
+                        <echo> |  You must add the following arguments to the CLI to use these snapshot archetypes:</echo>
```
romain-grecourt, line 144: There is nothing to do unless one wants to test the archetypes. Can you change the beginning of this sentence to provide some context ? E.g. `You can test the archetypes locally with the Helidon CLI using the following arguments:`

Also, the version is not always snapshot. E.g. if you build a tag.

archetypes/catalog/pom.xml
```diff
@@ -67,4 +67,95 @@
...
+                                                                 version="${project.version}"
+                                                                 type="zip"
+                                                                 target="."/>
+                                            </unpack-artifacts>
+                                            <templates>
+                                                <template source="cli-metadata.properties.mustache"
+                                                          target="metadata.properties">
+                                                    <variables>
+                                                        <variable name="buildtoolsVersion"
+                                                                  value="${build.tools.version}"/>
```
romain-grecourt, line 104: Can you use the same name of in helidon-site ? I.e. `cli.maven.plugin.version`
I know we have `build.tools.version` in `metadata.properties`, but IMO this is too confusing to expose as property.
E.g. The line below has `${version.plugin.helidon-build-tools}`

I would like to use `cli.maven.plugin.version` onwards and look at renaming the property inside `metadata.properties` (even if that means having code that support both properties to maintain backward compatibility).

Also this property needs to have a default value since this stager plugin execution is added by default.
You should define the property with a default value in the top level pom:

```
<cli.maven.plugin.version>${version.plugin.helidon-build-tools}</cli.maven.plugin.version>
```

archetypes/catalog/pom.xml
```diff
@@ -67,4 +67,95 @@
...
+                        <echo> |</echo>
+                        <echo> |     --reset --url file://${project.build.directory}/cli-data</echo>
+                        <echo> |</echo>
+                        <echo> |  For example:</echo>
+                        <echo> |</echo>
+                        <echo> |     helidon init --batch --reset --url file://${project.build.directory}/cli-data</echo>
+                        <echo> |</echo>
+                        <echo> |  To use a build tools snapshot, specify the version when building this project, e.g.:</echo>
+                        <echo> |</echo>
+                        <echo> |     mvn clean install -Dbuild.tools.version=2.0.3-SNAPSHOT</echo>
```
romain-grecourt, line 154: This is a continuation of my previous comment. 
I think it would be more explicit using `cli.maven.plugin.version` as the property name:

```
mvn clean install -Dcli.maven.plugin.version=2.0.3-SNAPSHOT
```

archetypes/catalog/pom.xml
```diff
@@ -67,4 +67,95 @@
...
+                        <phase>install</phase>
+                        <goals>
+                            <goal>echo</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <echos>
+                        <echo> |</echo>
+                        <echo> |                      /')    </echo>
```
romain-grecourt, line 136: Not really a comment, but what's the ascii art ?

archetypes/catalog/pom.xml
batsatt: `cli.maven.plugin.version` bothers me first because it doesn't following the normal naming convention `version.plugin.*` and second because there is no actual corresponding `cli` artifact; I didn't change it to better match the metadata property name. (Also, I still think we should be using this value when we generate from the archetypes, and then this name would be accurate.)

But I will change it just to make you happy ;-).

batsatt: image

archetypes/catalog/pom.xml
batsatt: see above

archetypes/catalog/pom.xml
romain-grecourt: My main issue with `build-tools.version` is that it's too generic. It does not follow the convention so you could use `version.plugin.helidon.cli`, or since as you point out there is no such maven artifact, you can use the existing property used to drive the helidon maven plugin: `version.plugin.helidon-build-tools`.

batsatt: The changes are already in to switch to `cli.maven.plugin.version`.

romain-grecourt: Right, but as you noted there it does not follow the current convention and does not really point to a maven artifact.

batsatt: We agreed on `cli.build.tools.version` and that change has been pushed.

PR #2163, opened 2020-07-14
Update microprofile examples to use microprofile-config.properties

barchetta:
Changed these two examples to use microprofile-config.properties (and not application.yaml).

This required removing the integer array in config (supported by yaml, not properties).

### DIFF
--- a/examples/microprofile/hello-world-explicit/README.md
+++ b/examples/microprofile/hello-world-explicit/README.md
@@ -13,5 +13,8 @@ Then try the endpoints:
 
 ```
 curl -X GET http://localhost:7001/helloworld
-curl -X GET http://localhost:7001/helloworld/another
+curl -X GET http://localhost:7001/helloworld/earth
 ```
+
+By default the server will use a dynamic port, see the messages displayed
+when the application starts.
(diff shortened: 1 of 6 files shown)


ljnelson:
(Full disclaimer: haven't looked in detail at this.)  MicroProfile Config [does support arrays of things natively](https://github.com/eclipse/microprofile-config/blob/1.4/spec/src/main/asciidoc/converters.asciidoc#array-converters).  For example, if this is in a `ConfigSource` somewhere:
```
foo.bar = 2,4,6,8
```
…or:
```
foo.bar = 2, 4, 6, 8
```
…and you include it like this:
```
@Inject
@ConfigProperty("foo.bar")
private int[] ints;
```
…or you get it like this:
```
final int[] ints = Config.getValue("foo.bar", int[].class);
```
…I believe it is supposed to be handled automatically by all conformant implementations (such as ours), though I'm typing off the cuff here so be kind in terms of syntax 😄 .  Were you not seeing this?

barchetta:
Put back int list in the example that used them.

examples/microprofile/hello-world-explicit/README.md
```diff
@@ -13,5 +13,8 @@ Then try the endpoints:
 
 ```
 curl -X GET http://localhost:7001/helloworld
-curl -X GET http://localhost:7001/helloworld/another
+curl -X GET http://localhost:7001/helloworld/earth
 ```
+
+By default the server will use a dynamic port, see the messages displayed
+when the application starts.
```
ljnelson, line 20: Haven't looked deeply but down below you set the port to 7001, so will a dynamic port truly be used?

barchetta: In the explicit example it uses a dynamic port. In the implicit example it uses 7001.

PR #2193, opened 2020-07-20
Multipart decoder rework

romain-grecourt:
Incorporate the changes to MultiPartDecoder from @olotenko.

### DIFF
--- a/media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
+++ b/media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
@@ -18,16 +18,17 @@ package io.helidon.media.multipart;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedList;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Flow.Processor;
+import java.util.concurrent.Flow.Publisher;
 import java.util.concurrent.Flow.Subscriber;
 import java.util.concurrent.Flow.Subscription;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 
 import io.helidon.common.http.DataChunk;
-import io.helidon.common.reactive.BufferedEmittingPublisher;
+import io.helidon.common.reactive.Multi;
 import io.helidon.common.reactive.SubscriptionHelper;
 import io.helidon.media.common.MessageBodyReadableContent;
 import io.helidon.media.common.MessageBodyReaderContext;
@@ -35,20 +36,28 @@ import io.helidon.media.multipart.VirtualBuffer.BufferEntry;
 
 /**
  * Reactive processor that decodes HTTP payload as a stream of {@link BodyPart}.
+ *
+ * This implementation is documented here {@code /docs-internal/multipartdecoder.md}.
  */
 public class MultiPartDecoder implements Processor<DataChunk, ReadableBodyPart> {
 
+    private static final int DOWNSTREAM_INIT = Integer.MIN_VALUE >>> 1;
+    private static final int UPSTREAM_INIT = Integer.MIN_VALUE >>> 2;
+    private static final int SUBSCRIPTION_LOCK = Integer.MIN_VALUE >>> 3;
+    private static final Iterator<MimeParser.ParserEvent> EMPTY_PARSER_ITERATOR = new EmptyIterator<>();
+
     private Subscription upstream;
     private Subscriber<? super ReadableBodyPart> downstream;
-    private BufferedEmittingPublisher<ReadableBodyPart> emitter;
     private ReadableBodyPart.Builder bodyPartBuilder;
     private ReadableBodyPartHeaders.Builder bodyPartHeaderBuilder;
-    private BufferedEmittingPublisher<DataChunk> bodyPartPublisher;
-    private final CompletableFuture<BufferedEmittingPublisher<ReadableBodyPart>> initFuture;
-    private final LinkedList<ReadableBodyPart> bodyParts;
+    private DataChunkPublisher bodyPartPublisher;
+    private Iterator<MimeParser.ParserEvent> parserIterator = EMPTY_PARSER_ITERATOR;
+    private volatile Throwable error;
+    private boolean cancelled;
+    private AtomicInteger contenders = new AtomicInteger(Integer.MIN_VALUE);
+    private AtomicLong partsRequested = new AtomicLong();
     private final HashMap<Integer, DataChunk> chunksByIds;
     private final MimeParser parser;
-    private final ParserEventProcessor parserEventProcessor;
     private final MessageBodyReaderContext context;
 
     /**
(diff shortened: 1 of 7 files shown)


media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -102,11 +103,10 @@ public void subscribe(Subscriber<? super ReadableBodyPart> subscriber) {
...
             public void request(long n) {
-                long curr = n <= 0 ?
-                        partsRequested.getAndSet(-1) :
-                        partsRequested.getAndUpdate(v -> Long.MAX_VALUE - v > n ? v + n
-                                : v < 0 ? v
-                                : Long.MAX_VALUE);
+                long curr = n <= 0
+                        ? partsRequested.getAndSet(-1)
+                        : partsRequested.getAndUpdate(v -> Long.MAX_VALUE - v > n
+                        ? v + n : v < 0 ? v : Long.MAX_VALUE);
```
olotenko, line 109: This is a ternary operator inside a lambda, so you may want to indent it a bit more

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -426,11 +425,10 @@ public void subscribe(Subscriber<? super DataChunk> sub) {
...
                     // entire bufferEntryIterator, and signal onError
-                    long curr = n <= 0 ? chunksRequested.getAndSet(-1) :
-                            chunksRequested.getAndUpdate(v -> Long.MAX_VALUE - v > n ? v + n
-                                    : v < 0 ? v == Long.MIN_VALUE ? n
-                                    : v
-                                    : Long.MAX_VALUE);
+                    long curr = n <= 0
+                            ? chunksRequested.getAndSet(-1)
+                            : chunksRequested.getAndUpdate(v -> Long.MAX_VALUE - v > n
+                            ? v + n : v < 0 ? v == Long.MIN_VALUE ? n : v : Long.MAX_VALUE);
```
olotenko, line 431: Ditto: maybe indent this ternary operator a bit more

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -106,120 +142,229 @@ public void onNext(DataChunk chunk) {
...
+                        if (!bodyPartPublisher.drain()) {
+                            // the body was not fully drained, exit the parser iterator loop
+                            // the parser events processing will resume upon inner Subscriber demand
+                            return;
+                        }
+                        break;
+                    case END_PART:
+                        bodyPartPublisher.complete(null);
+                        bodyPartHeaderBuilder = null;
+                        bodyPartBuilder = null;
```
olotenko, line 303: I think these must be moved to line `289`: the cleanup must occur asap. Also, add nulling out to `cleanup`. Although the objects are small, let's be very clean about the promise to not retain objects for longer than necessary.

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -106,120 +142,229 @@ public void onNext(DataChunk chunk) {
...
+                downstream.onError(error);
+                cleanup();
+                return;
             }
 
-            @Override
-            public void onComplete() {
+            // parserIterator is drained, drop the reference to it, but keep it safe for any later invocations
+            parserIterator = EMPTY_PARSER_ITERATOR;
+            if (upstream == SubscriptionHelper.CANCELED) {
```
olotenko, line 340: There is a subtle race condition here. I'll propose a fix later.

(In essence, Dekker idiom is needed - `onError` concurrent with `drainBoth` executed by, say, `request` - may set `CANCELED` before `error`, so `drainBoth` may see `error` is `null`, but after reaching this line also see `CANCELED`, and deliver `onComplete`)

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -247,85 +392,148 @@ private BodyPartChunk createPartChunk(BufferEntry entry) {
...
+                return;
+            }
+            cancelled = true;
+
+            // bufferEntryIterator is drained because complete() is invoked only by drain() which proceeds past
+            // state == BODY only when drain() returned true
+            if (th != null) {
+                subscriber.onError(th);
+            } else if (chunksRequested.get() < 0) {
+                subscriber.onError(new IllegalArgumentException("Expecting only positive requests"));
```
olotenko, line 482: Move to line `471` and change the condition to:

```
if (chunksRequested.get() < 0 && !cancelled && th == null) {
   th = new IllegalArgumentException("Expecting only positive requests for content");
}
```

(Because currently a `cancel` concurrent with `complete` will not be distinguishable from bad request)

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -106,120 +142,229 @@ public void onNext(DataChunk chunk) {
...
-            upstream.request(1);
+            drain(ex);
         }
     }
 
     @Override
     public void onError(Throwable throwable) {
         Objects.requireNonNull(throwable);
-        initFuture.whenComplete((e, t) -> e.fail(throwable));
+        if (upstream != SubscriptionHelper.CANCELED) {
```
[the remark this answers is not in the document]
olotenko: Actually, this is what needs doing: remove the check and the following line setting `upstream = CANCELED;`. Then the subtle race condition discussed in the comment to `drainBoth` will disappear: then observing `upstream == CANCELED` is possible only if `onComplete` has been observed, which is mutually exclusive with `onError`.

Add `upstream = CANCELED;` to `cleanup`.

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
olotenko: (With the change so that `onError` does not set `upstream = CANCELED;` the race condition disappears)

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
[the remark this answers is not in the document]
olotenko: Well, that won't work. The spec is very particular about not allowing synchronous cancellation, even though concurrent cancellation should be supported by `upstream`. And "they" will tell you "just add a check". It isn't about adding a check. It is about synchronizing on two variables instead of just one, for a dubious purpose.

media/multipart/src/main/java/io/helidon/media/multipart/MultiPartDecoder.java
```diff
@@ -106,120 +142,231 @@ public void onNext(DataChunk chunk) {
...
+        // Otherwise returns a positive number, if serving inner Subscriber, or actual partsRequested.
+        return bodyPartPublisher != null ? 1 : partsRequested.get();
     }
 
-    private void onPartCancel() {
-        emitter.clearBuffer(this::drainPart);
+    private void cleanup() {
+        // drop the reference to parserIterator, but keep it safe for any later invocation of parserIterator
+        parserIterator = EMPTY_PARSER_ITERATOR;
+        error = null;
```
olotenko, line 203: I apologise I didn't make the comment 
> Add upstream = CANCELED; to cleanup

clearer and forgot to add it in the `onError` synchronization fixup.

Please, add `upstream = CANCELED` somewhere here to drop references to upstream.

PR #2216, opened 2020-07-27
Migration guide update

barchetta:
Document that some accessors have changed to return `Optional<T>`


barchetta:
Bad PR. Please see #2217

Issue #1412, opened 2020-02-19
Prototype / Early Access use of virtual threads (loom, fibers)


barchetta:
Prototype support for using virtual threads in Helidon. Ideally we would release early access modules that when used with an early access Loom release allows experimentation with virtual threads.

Some ideas:

- [x] Helidon MP: Supply an UnboundedExecutor for the JAX-RS executor #3164 
- [x] Common: Supply an UnboundedExecutor via common.configurable.ThreadPoolSupplier #3164 
- [x] Helidon SE: Support passing ExecutorService when registering Service #1327
- [x] Distribution: explore ways to make above early access features available to Helidon customers.
- [ ] Netty: investigate/experiment Netty support for virtual threads
- [x] Other #549

PR #2224, opened 2020-07-31
Initial revision of Loom experiments

ljnelson:
This is just the config stuff to trigger it; more commits to follow.  The intent is that you can check this PR out and it will still build and run if you have JDK 11 but will be capable of being run under a Loom early build.

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
@@ -111,7 +111,7 @@ public class ServerCdiExtension implements Extension {
         if (null == jaxRsExecutorService) {
             Config serverConfig = config.get("server");
             final java.lang.reflect.Method m;
-            if (serverConfig.get("virtual-threads").asBoolean().isPresent()) {
+            if (serverConfig.get("virtual-threads").asBoolean().orElse(false)) {
                 java.lang.reflect.Method temp = null;
                 try {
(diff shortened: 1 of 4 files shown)


ljnelson:
This (currently draft) PR is related to #1412.

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -108,10 +109,34 @@ private void startServer(@Observes @Priority(PLATFORM_AFTER + 100) @Initialized(
 
         // make sure all configuration is in place
         if (null == jaxRsExecutorService) {
-            jaxRsExecutorService = ServerThreadPoolSupplier.builder()
+            Config serverConfig = config.get("server");
+            final java.lang.reflect.Method m;
+            if (serverConfig.get("virtual-threads").asBoolean().isPresent()) {
```
tomas-langer, line 114: `asBoolean().orElse(false)` - what if `virtual-threads: false` - you would execute as well

ljnelson: Yep; just a dumb typo in this draft PR

ljnelson:
I removed my fork at one point so I'm going to close this and re-roll it with substantially the same changes.

ljnelson:
Closing; watch for the new one!

ljnelson:
See https://github.com/oracle/helidon/pull/2417

PR #2229, opened 2020-08-03
Make sure request scope is propagated to newly created threads in FT

spericas:
After discussing the solution with @jansupol, I believe this is currently the simplest way to support this. We do need to create a dummy CDI bean with an injection point for Jersey to make its `RequestScope` type available via CDI. Using that and some other CDI classes, we ensure all injections (`@Context` or `@Inject` ones) are available in newly created threads.

One of the functional tests has been updated to verify this fix.

### DIFF
--- a/tests/functional/context-propagation/src/test/java/io/helidon/tests/functional/context/hello/HelloTest.java
+++ b/tests/functional/context-propagation/src/test/java/io/helidon/tests/functional/context/hello/HelloTest.java
@@ -73,7 +73,7 @@ class HelloTest {
     }
 
     @Test
-    void testRemoteAAddress() {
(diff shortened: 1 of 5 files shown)


tests/functional/context-propagation/src/test/java/io/helidon/tests/functional/context/hello/HelloTest.java
```diff
@@ -72,6 +72,12 @@ void testHelloAsync() {
         assertOk(target.request().get(), "Hello World");
     }
 
+    @Test
+    void testRemoteAAddress() {
```
ljnelson, line 76: Nit: typo

PR #2241, opened 2020-08-07
Support for bean producers in different package than beans that have …

### LINKED ISSUE #2223: Package private methods broken in some CDI proxies
## Environment Details
* Helidon Version: 2.0.1
* Helidon MP
* JDK version: 11.0.8
* OS: Linux

----------

## Problem Description
When a proxied object is created through a producer (in a different package than the object class, which might be from a library), it's package-private methods will be broken, returning default field values instead of actual bean values.

## Steps to reproduce
See attached example. SampleProcessor internal logic is broken because of invalid proxy.
[example.zip](https://github.com/oracle/helidon/files/5002788/example.zip)

tomas-langer:
…package local methods.

Resolves #2223 

### DIFF
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
@@ -23,6 +23,7 @@ import java.security.ProtectionDomain;
 import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.Map;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import io.helidon.common.NativeImageHelper;
@@ -75,8 +76,17 @@ class HelidonProxyServices implements ProxyServices {
             // classloader, preventing it from seeing these fields/methods
             return defineClassSamePackage(originalClass, className, classBytes, off, len);
         } else {
-            // use a custom classloader to define classes in a new package
-            return wrapCl(originalClass.getClassLoader()).doDefineClass(originalClass, className, classBytes, off, len);
+            // try to use same package approach (maybe in same module?) to support package local methods
+            try {
+                return defineClassSamePackage(originalClass, className, classBytes, off, len);
+            } catch (Exception e) {
+                LOGGER.log(Level.FINEST,
+                           "Failed to create class " + className + " in same classloader. Will use a different one",
+                           e);
+                // use a custom classloader to define classes in a new package
+                return wrapCl(originalClass.getClassLoader())
+                        .doDefineClass(originalClass, className, classBytes, off, len);
+            }
         }
     }
 
@@ -91,8 +101,17 @@ class HelidonProxyServices implements ProxyServices {
         if (samePackage(originalClass, className)) {
             return defineClassSamePackage(originalClass, className, classBytes, off, len);
         } else {
-            return wrapCl(originalClass.getClassLoader())
-                    .doDefineClass(originalClass, className, classBytes, off, len, protectionDomain);
+            // try to use same package approach (maybe in same module?) to support package local methods
+            try {
+                return defineClassSamePackage(originalClass, className, classBytes, off, len);
+            } catch (Exception e) {
+                LOGGER.log(Level.FINEST,
+                           "Failed to create class " + className + " in same classloader. Will use a different one",
+                           e);
+
+                return wrapCl(originalClass.getClassLoader())
+                        .doDefineClass(originalClass, className, classBytes, off, len, protectionDomain);
+            }
         }
     }
 
@@ -117,8 +136,20 @@ class HelidonProxyServices implements ProxyServices {
                 // it also needs to open the package we are doing the lookup in
                 myModule.addReads(classModule);
             }
+
+            // I would like to create a private lookup in the same package as the proxied class, so let's
+            // try to load it
+            String proxiedClassName = className.substring(0, className.indexOf('$'));
+            Class<?> lookupClass;
+            try {
+                lookupClass = originalClass.getClassLoader().loadClass(proxiedClassName);
+            } catch (Throwable e) {
+                LOGGER.log(Level.FINEST, "Cannot load proxied class: " + proxiedClassName, e);
+                lookupClass = originalClass;
+            }
+
             // next line would fail if the module does not open its package, with a very meaningful error message
-            MethodHandles.Lookup lookup = MethodHandles.privateLookupIn(originalClass, MethodHandles.lookup());
+            MethodHandles.Lookup lookup = MethodHandles.privateLookupIn(lookupClass, MethodHandles.lookup());
             if (classBytes.length == len) {
                 return lookup.defineClass(classBytes);
             } else {
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
@@ -35,8 +35,8 @@ class HelidonProxyServices implements ProxyServices {
     private static final Logger LOGGER = Logger.getLogger(HelidonProxyServices.class.getName());
     private static final String CLIENT_PROXY_SUFFIX = "$$Proxy$_$$_WeldClientProxy";
     private static final String PROXY_SUFFIX = "$$Proxy$_$$_Weld$Proxy$";
-    private static final String WELDX_PACKAGE_PREFIX = "org.jboss.weldx";
-    private static final String WELD_LANG_PACKAGE_PREFIX = "org.jboss.weld.lang";
+    private static final String WELD_JAVAX_PREFIX = "org.jboss.weldx.";
+    private static final String WELD_JAVA_PREFIX = "org.jboss.weld.";
 
     // a cache of all classloaders (this should be empty in most cases, as we use a single class loader in Helidon)
     private final Map<ClassLoader, ClassDefiningCl> classLoaders = Collections.synchronizedMap(new IdentityHashMap<>());
@@ -123,7 +123,7 @@ class HelidonProxyServices implements ProxyServices {
     }
 
     private boolean weldInternalProxyClassName(String className) {
-        return className.startsWith(WELDX_PACKAGE_PREFIX) || className.startsWith(WELD_LANG_PACKAGE_PREFIX);
+        return className.startsWith(WELD_JAVAX_PREFIX) || className.startsWith(WELD_JAVA_PREFIX);
     }
 
     private Class<?> defineClassPrivateLookup(Class<?> originalClass, String className, byte[] classBytes, int off, int len) {
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
@@ -141,13 +141,15 @@ class HelidonProxyServices implements ProxyServices {
             }
 
             // I would like to create a private lookup in the same package as the proxied class, so let's
-            // try to load it
-            String proxiedClassName = className.substring(0, className.indexOf('$'));
+            // try to load it - I load the enclosing class (or the proxied class if not inner) to have
+            // a lookup in the correct package/class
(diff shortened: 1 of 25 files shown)


microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -75,8 +76,17 @@ public boolean supportsClassDefining() {
             // classloader, preventing it from seeing these fields/methods
             return defineClassSamePackage(originalClass, className, classBytes, off, len);
         } else {
-            // use a custom classloader to define classes in a new package
-            return wrapCl(originalClass.getClassLoader()).doDefineClass(originalClass, className, classBytes, off, len);
+            // try to use same package approach (maybe in same module?) to support package local methods
+            try {
+                return defineClassSamePackage(originalClass, className, classBytes, off, len);
```
ljnelson, line 81: Maybe refactor so this statement happens only once instead of twice, so something like:
```
try {
    return defineClassSamePackage()...
} catch (Exception e) {
    if (samePackage(originalClass, className)) {
        throw e or wrapped e;
    } else {
        // carry on
    }
}
```

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -91,8 +101,17 @@ public boolean supportsClassDefining() {
         if (samePackage(originalClass, className)) {
             return defineClassSamePackage(originalClass, className, classBytes, off, len);
         } else {
-            return wrapCl(originalClass.getClassLoader())
-                    .doDefineClass(originalClass, className, classBytes, off, len, protectionDomain);
+            // try to use same package approach (maybe in same module?) to support package local methods
+            try {
+                return defineClassSamePackage(originalClass, className, classBytes, off, len);
```
ljnelson, line 106: Same as above

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
tomas-langer: Fixed

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
tomas-langer: Fixed

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -117,8 +136,20 @@ public boolean supportsClassDefining() {
                 // it also needs to open the package we are doing the lookup in
                 myModule.addReads(classModule);
             }
+
+            // I would like to create a private lookup in the same package as the proxied class, so let's
+            // try to load it
+            String proxiedClassName = className.substring(0, className.indexOf('$'));
```
ljnelson, line 145: Thinking about it, this worries me.  If I have a nested class, won't this chop off the nested class name?  e.g. `com.foo.Bar$Baz$various$proxy$related$bits`: you'll conclude here that the proxied class name is `com.foo.Bar`, which is incorrect.  Or maybe this is intentional in some way, as the outer class (`Bar`) will of course be in the same package as the nested (and proxied) class (`Bar$Baz`) so it doesn't matter in practice?  Maybe a comment to further clarify here what you're doing?

tomas-langer: This is really to get a private lookup - so either you get the actual class, or the enclosing class - in both cases the private lookup should work

tomas-langer: Problem is that we only have a class that produces the bean, which is absolutely useless when in a different package.
I need to get a class that allows me to obtain the private lookup - this seemed to be the easiest approach, as it should work both for class itself or inner class.

ljnelson: Then maybe the name is at fault?  Shouldn't it be `privateLookupClassName` or something?

tomas-langer: And if the class is not valid (e.g. we cannot load it), I fall back to the producer class name

tomas-langer: Also the method name is now `defineClassPrivateLookup` which is exactly what it does - it defines a class using a private lookup. 
The name is related to how the lookup is used, not how it is obtained.

ljnelson: But isn't `proxiedClassName` the name of the lookup class, not the name of the class being proxied?  See line 148.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -88,11 +98,23 @@ public boolean supportsClassDefining() {
...
+                throw e;
+            } else {
+                LOGGER.log(Level.FINEST,
+                           "Failed to create class " + className + " in same classloader. Will use a different one",
+                           e);
+
+                // other cases (except for a few edge cases, such as producer in a different package and usage
+                // of bean in the same package) we can live with a different classloader to hold the proxy class
+                return wrapCl(originalClass.getClassLoader())
+                        .doDefineClass(originalClass, className, classBytes, off, len, protectionDomain);
```
ljnelson, line 116: I don't know enough about this to have a strong opinion, but let's say the private lookup fails.  Shouldn't we still endeavor to use [`MethodHandles.Lookup#defineClass(byte[])`](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html#defineClass(byte[])) to define the class instead of widening the access of `ClassLoader#defineClass(byte[])`?  Maybe using a public lookup rather than a private one?

tomas-langer: Will try

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
tomas-langer: Refactored the whole thing, closing this discussion. I hope the new approach is correctly named (and documented).

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
tomas-langer: `PublicLookup` is not useful, as it does not have the `PACKAGE` right, so we cannot define classes using it.
I have changed the code to work as follows:
- Weld specific packages (weldx, weld.lang) use the original classloader based approach
- all other beans use private lookup

If the private lookup fails, the bean class creation fails as well with appropriate message.
There may be some edge cases not covered by this approach, though our repository does not contain any such case.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -110,39 +135,91 @@ public boolean supportsClassDefining() {
...
-            Module classModule = originalClass.getModule();
-            if (!myModule.canRead(classModule)) {
+            // lookup class name based on full class name (e.g. for inner classes)
+            String lookupClassName;
+            // lookup class name based on the first class in the name
+            String fallbackLookupClassName;
+
+            if (className.contains("$")) {
+                // fallback is the package + first name in the compound proxy class name
+                fallbackLookupClassName = className.substring(0, className.indexOf('$'));
```
ljnelson, line 148: Is there any reason not to just go with this strategy?  That is, is anything gained by first trying `lookupClassName`?  My understanding is that a lookup class with respect to `MethodHandles.Lookup` is used only for its package and module—and an enclosing or top level class should be fine here.

tomas-langer: You are right, this is too complicated and most likely useless. Simplified.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -110,39 +133,75 @@ public boolean supportsClassDefining() {
...
+
         try {
-            Module classModule = originalClass.getModule();
-            if (!myModule.canRead(classModule)) {
+            // lookup class name "guessed" from the class name of the proxy
+            String lookupClassName;
+
+            if (className.contains("$")) {
+                // package + first name in the compound proxy class name
+                lookupClassName = className.substring(0, className.indexOf('$'));
```
ljnelson, line 144: `contains` effectively does an `indexOf`; no need to do `indexOf` again.  I'd rewrite this block something like this:
```
final int index = className.indexOf('$');
lookupClassName = index < 0 ? className : className.substring(0, index);
```

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -110,39 +133,75 @@ public boolean supportsClassDefining() {
...
+
+            if (null == lookupClass) {
+                // try to load the full class name
+                lookupClass = tryLoading(cl, lookupClassName);
+            }
+
+            if (null == lookupClass) {
+                // and if that fails, just use the bean producer class
+                lookupClass = originalClass;
+            }
```
ljnelson, line 163: Similarly, I think this can all be simplified based on whether `index < 0`.  So:
```
final Class<?> lookupClass = index < 0 ? originalClass : tryLoading(cl, lookupClassName);
```

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
tomas-langer: Changed.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
tomas-langer: Changed.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -110,39 +133,65 @@ public boolean supportsClassDefining() {
...
+
+            Class<?> lookupClass;
+            if (index < 0) {
+                LOGGER.finest(() -> "Attempt to define a proxy class without a $ in its name. Class name: " + className + ","
+                                      + " original class name: " + originalClass.getName());
+                lookupClass = originalClass;
+            } else {
+                // I would like to create a private lookup in the same package as the proxied class, so let's do it
+                // use the "extracted" lookup class name, if that fails, use the original class
+                lookupClass = tryLoading(originalClass, className.substring(0, className.indexOf('$')));
```
ljnelson, line 151: tiny nit: you already stored `index`; just reuse it instead of `className.indexOf('$')` again

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonProxyServices.java
```diff
@@ -110,39 +133,65 @@ public boolean supportsClassDefining() {
...
     }
 
-    private boolean samePackage(Class<?> originalClass, String className) {
-        String origPackage = originalClass.getPackageName();
-        String newPackage = packageName(className);
-        return newPackage.equals(origPackage);
+    private Class<?> tryLoading(Class<?> originalClass, String className) {
+        try {
+            ClassLoader cl = originalClass.getClassLoader();
+            return cl.loadClass(className);
```
ljnelson, line 173: Nit: just do `return originalClass.getClassLoader().loadClass(className)`; no need to define a local variable.

PR #2257, opened 2020-08-13
WebClient request id added and logging messages updated

### LINKED ISSUE #2130: WebClient request identification
Add some sort of unique request identification to easily identify which logging messages are bound to which request.

Verdent:
Fixes #2130

### DIFF
--- a/webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
+++ b/webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
@@ -52,7 +52,7 @@ class RequestConfiguration extends WebClientConfiguration {
         return services;
     }
 
-    public long requestId() {
+    long requestId() {
         return requestId;
     }
 
--- a/webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
+++ b/webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
@@ -28,6 +28,7 @@ import io.helidon.webclient.spi.WebClientService;
 class RequestConfiguration extends WebClientConfiguration {
 
     private final URI requestURI;
+    private final long requestId;
     private final WebClientServiceRequest clientServiceRequest;
     private final List<WebClientService> services;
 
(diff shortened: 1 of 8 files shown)


webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
```diff
@@ -50,13 +52,18 @@ WebClientServiceRequest clientServiceRequest() {
         return services;
     }
 
+    public long requestId() {
```
tomas-langer, line 55: public method in a non-public class, also no javadoc

webclient/webclient/src/main/java/io/helidon/webclient/WebClientRequestBuilderImpl.java
```diff
@@ -147,7 +149,8 @@ private WebClientRequestBuilderImpl(LazyValue<NioEventLoopGroup> eventGroup,
         this.services = configuration.clientServices();
         this.readerContext = MessageBodyReaderContext.create(configuration.readerContext());
         this.writerContext = MessageBodyWriterContext.create(configuration.writerContext(), headers);
-        Context.Builder contextBuilder = Context.builder().id("webclient-" + REQUEST_NUMBER.incrementAndGet());
+        this.requestId = REQUEST_NUMBER.incrementAndGet();
```
tomas-langer, line 152: If we want to have request id, it should be configurable (esp. when using tracing, we may want to bind it to the trace id).
Also this is usually handled in logging through MDC (Mapped diagnostic context) in frameworks that support such a feature.
I am not sure the chosen approach is aligned with overall logging in Helidon (e.g. using (..) and message).

tomas-langer:
Is this related to an issue?

webclient/webclient/src/main/java/io/helidon/webclient/RequestConfiguration.java
Verdent: ehm... yes.. This should be package private

webclient/webclient/src/main/java/io/helidon/webclient/WebClientRequestBuilderImpl.java
Verdent: fixed

PR #2269, opened 2020-08-18
Introduces an ExecutorServices implementation to Helidon's repackaged Weld implementation to ensure context classloaders are never null

ljnelson:


ljnelson:
@tomas-langer @romain-grecourt I am not sure if I have put this `ExecutorServices` implementation in the right place.  While this PR is draft, could you at least tell me if it is an OK place for this file to go?  I have verified that it is recognized and loaded and its `wrap` method is called.  I've done nothing else yet.

tomas-langer:
This module is only for repackaged code (e.g. stuff we need to modify in Weld).
If you need to update something through SPI or API of Weld, please add it to `microprofile/cdi` module (`HelidonProxyServices` is also located here).

Regarding the bug you are fixing - we need to make sure, that the components always get the same instance of MP Config, regardless of the lifecycle of CDI - otherwise we would still get issues when using a customized config instance.

tomas-langer:
Looks good.

tomas-langer (approved):

PR #2283, opened 2020-08-24
Fix issue with gRPC clients and services where method signatures have types with nested generics

thegridman:


tomas-langer (approved):
LGTM

thegridman:
Someone with permissions to the repo will need to merge this as I cannot do it.

PR #2285, opened 2020-08-25
BEP refactor

danielkec:

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/BiConsumerChain.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/BiConsumerChain.java
@@ -30,13 +30,6 @@ class BiConsumerChain<T, S>
         }
     }
 
-    BiConsumerChain<T, S> combineWith(BiConsumer<? super T, ? super S> another) {
-        BiConsumerChain<T, S> newChain = new BiConsumerChain<>();
-        newChain.addAll(this);
-        newChain.add(another);
-        return newChain;
-    }
-
     static <T, S> BiConsumer<T, S> combine(
             BiConsumer<T, S> current,
             BiConsumer<T, S> another) {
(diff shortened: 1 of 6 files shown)


olotenko (commented):
This looks like the proof of concept verbatim, so looks ok to me :). The questions raised in the comments need addressing. (eg ensure `clearBuffer` is unusable - I propose to just delete the method, but maybe you have different requirements to "ease in" the change - it is broken in the old implementation, and cannot be made to work correctly in the proposed - in fact, in any concurrent setting; the statement about the meaning of `isCanceled` must be verified across the usage sites - the users of the return value must be prepared to deal with `isCanceled` potentially switching between `true` and `false`; what to do with `IllegalStateException` potentially being thrown from `request`, if the victim manages to enter `drain`, and encounters an `Exception` due to `onNext`)

danielkec:
> This looks like the proof of concept verbatim, so looks ok to me :). The questions raised in the comments need addressing. (eg ensure `clearBuffer` is unusable - I propose to just delete the method, but maybe you have different requirements to "ease in" the change - it is broken in the old implementation, and cannot be made to work correctly in the proposed - in fact, in any concurrent setting; the statement about the meaning of `isCanceled` must be verified across the usage sites - the users of the return value must be prepared to deal with `isCanceled` potentially switching between `true` and `false`; what to do with `IllegalStateException` potentially being thrown from `request`, if the victim manages to enter `drain`, and encounters an `Exception` due to `onNext`)

- `clearBuffer` is intended only for cleanup, javadoc explains that pretty well, we need it for releasing unconsumed Netty ByteBufs where GC wont help
- `isCanceled`we will need it for [MP RM Emitter](https://download.eclipse.org/microprofile/microprofile-reactive-messaging-2.0/apidocs/org/eclipse/microprofile/reactive/messaging/Emitter.html#isCancelled--) unlike the spec we covered it's volatility in the javadoc 
- Exception during `onNext` gets translated to onError or did I miss something?

olotenko:
> * `clearBuffer` is intended only for cleanup, javadoc explains that pretty well, we need it for releasing unconsumed Netty ByteBufs where GC wont help

The point is that `clearBuffer` is an inherently unsafe method in a concurrent setting. You should never call `clearBuffer`. You should instead embrace continuation passing paradigm - that is, set up a call back for the purpose of managing lifecycle of buffered objects, as in the proposed implementation.

> * Exception during `onNext` gets translated to onError or did I miss something?

Correct. This is the new behaviour of the proposal. The old implementation does not do that, which can be perceived as a change in behaviour - hence the request to review such changes in behaviour.

tomas-langer (dismissed):
LGTM

tomas-langer (approved):

PR #2288, opened 2020-08-26
Allow cancellation of Future's wrapped in Single's and vice versa

spericas:
Allow cancellation of Future's wrapped in Single's and vice versa. This form of cancellations are necessary to implement FT MP. New test.

### DIFF
--- a/common/reactive/src/test/java/io/helidon/common/reactive/CancellationTest.java
+++ b/common/reactive/src/test/java/io/helidon/common/reactive/CancellationTest.java
@@ -57,11 +57,7 @@ public class CancellationTest {
         AtomicBoolean cancelled = new AtomicBoolean(false);
         Single<Object> single = Single.create(new CompletableFuture<>());
         CompletableFuture<Object> future = single.toStage().toCompletableFuture();
-        single.whenComplete((o, t) -> {
-            if (t instanceof CancellationException) {
-                cancelled.set(true);
-            }
-        });
+        single.onCancel(() -> cancelled.set(true));
         single.cancel();
         future.cancel(true);        // should cancel single
         assertThat(cancelled.get(), is(true));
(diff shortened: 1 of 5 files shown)


common/reactive/src/test/java/io/helidon/common/reactive/CancellationTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    public void testSingleCancel() {
+        AtomicBoolean cancelled = new AtomicBoolean(false);
+        Single<Object> single = Single.create(new CompletableFuture<>());
+        CompletableFuture<Object> future = single.toStage().toCompletableFuture();
+        single.whenComplete((o, t) -> {
+            if (t instanceof CancellationException) {
+                cancelled.set(true);
+            }
+        });
```
danielkec, line 64: Just for the clarity could there be `single.onCancel(() -> cancelled.set(true));` ?

spericas: Sure

common/reactive/src/test/java/io/helidon/common/reactive/CancellationTest.java
```diff
@@ -0,0 +1,65 @@
...
+     * a {@code CompletableFuture}.
+     */
+    @Test
+    public void testSingleCancel() {
+        AtomicBoolean cancelled = new AtomicBoolean(false);
+        Single<Object> single = Single.create(new CompletableFuture<>());
+        CompletableFuture<Object> future = single.toStage().toCompletableFuture();
+        single.onCancel(() -> cancelled.set(true));
+        single.cancel();
+        future.cancel(true);        // should cancel single
```
danielkec, line 62: Sorry I didn't notice first time, `future.cancel(true);` should be enough `single.cancel();` makes the test inconclusive

spericas: Good catch, had to do some manual merging :(

danielkec (approved):

danielkec (approved):

PR #2290, opened 2020-08-26
Set junit4 version where used. Fixes #2289

### LINKED ISSUE #2289: junit version in microprofile/tests/arquillian/pom.xml is not managed by Helidon
Reported by David Delabassee:
```
Btw, I wanted to test Helidon with a snapshot build of Jackson (they just merged Records support).
I updated the Jackson deps but the build fail

[ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for junit:junit:jar is missing. @ io.helidon.microprofile.tests:helidon-arquillian:[unknown-version], /Users/delabassee/Desktop/helidon/microprofile/tests/arquillian/pom.xml, line 64, column 21
```

It looks like we don't manage the version of junit 4 and we get the version transitively from Jackson. When I generate the effective pom for the pom referenced in the error above I see that the dependencyManagement from  Jackson manages the junit 4 version. I assume in David's case the new Jackson version no longer manages the junit 4 version.

barchetta:
Fixes #2289 

I did not put the junit 4 version in dependency management since Helidon uses Junit 5. Instead the version is specified in the one place where it is needed.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -78,6 +78,7 @@
         <version.lib.jsonp-api>1.1.6</version.lib.jsonp-api>
         <version.lib.jsonp-impl>1.1.6</version.lib.jsonp-impl>
         <version.lib.junit>5.6.2</version.lib.junit>
+        <version.lib.junit4>4.12</version.lib.junit4>
         <version.lib.maven-wagon>2.10</version.lib.maven-wagon>
         <version.lib.microprofile-config>1.4</version.lib.microprofile-config>
         <version.lib.microprofile-health>2.2</version.lib.microprofile-health>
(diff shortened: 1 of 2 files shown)


barchetta:
`mvn org.apache.maven.plugins:maven-help-plugin:3.2.0:effective-pom -Dverbose` shows:

```
  <dependencyManagement>
    <dependencies>   
. . .
      <dependency>
        <groupId>junit</groupId>  <!-- com.fasterxml.jackson:jackson-parent:2.11, line 76 -->
        <artifactId>junit</artifactId>  <!-- com.fasterxml.jackson:jackson-parent:2.11, line 77 -->
        <version>4.12</version>  <!-- com.fasterxml.jackson:jackson-parent:2.11, line 78 -->
      </dependency>
    </dependencies>
  </dependencyManagement>
```

So this seems to indicate the junit version is coming from Jackson.

PR #2298, opened 2020-08-28
Security now has higer priority

### LINKED ISSUE #2296: Security MP service not triggered for metrics
## Environment Details
* Helidon Version: 1.4.7
* Helidon MP
* JDK version: 1.8
----------

## Problem Description

- when I configure security to protect metrics endpoint (using web server security), it may not get triggered, as security MP service has same priority as metrics MP service

Example configuration:
```yaml
security:
  providers:
    - abac:
    - http-basic-auth:
        realm: "helidon"
        users:
          - login: "success"
            password: "password"
            roles: ["admin"]
          - login: "fail"
            password: "password"
  web-server:
    paths:
      - path: "/metrics[/{*}]"
        authenticate: true
        authorize: true
        abac:
          roles-allowed:
            user:
              - admin
```

Please also validate this works as it should in 2.0

### LINKED ISSUE #2297: Abac time fails even when not configured
Follow up of #2296  (same version, environment, and configuration)

When I access the endpoint, I get the following exception:
```
Caused by: java.util.concurrent.CompletionException: io.helidon.config.MissingValueException: Requested value for configuration key 'security.web-server.paths.0.abac.time.time-of-day' is not present in the configuration.
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
```

We should never get this exception, as `time` validator is not configured and should not be triggered

Validate this works as it should in 2.0

tomas-langer:
Abac now correctly handles validator config

Security MP service must have higher priority than services exposing endpoints (such as metrics, health), so we can protect these endpoints.

In addition when an endpoint was protected using abac, each validator tried to read the configuration and time validator threw an exception.

Resolves #2296 
Resolves #2297 
For 1.x

### DIFF
--- a/microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityMpService.java
+++ b/microprofile/security/src/main/java/io/helidon/microprofile/security/SecurityMpService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


Verdent (approved):
LGTM

PR #2300, opened 2020-08-31
Security provider registration added

### LINKED ISSUE #2293: Example "security/outbound-override" is broken
## Environment Details
* Helidon Version: 2.x
* Helidon SE
----------

## Problem Description
Seems the change to webclient broke this example. There is no test to validate it, so we never found out.
When I run the example, I get `Anonymous` from the backend service, but identity should be propagated (or explicit).

Verdent:
Fixes #2293


tomas-langer (approved):
LGTM

PR #2308, opened 2020-09-01
Update MediaType, lazy non known type, read-only parameters map

romain-grecourt:
Update MediaType.create
 - return a known type only if there is no parameters
 - change MediaType.KNOWN_TYPES from Map<MediaType,MediaType> to Map<String, MediaType>
 - build a key as a String of the form type/subtype,
   this is a minor optimization for known types that avoids creating a MediaType instance as a key
- replace use of Optional with plain if conditions

Update MediaType static fields initialization
- replace createContentType with a dedicated constructor to avoid the calls to `createContentType/create/builder` and the creation of an instance of `Builder`
- use a static initializer to populate the `KNOWN_TYPE` map

Make parameters a read-only map
 - Optimize for empty parameters and use Map.of()
 - otherwise wrap it with Collections.unmodifiableMap()

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/MediaType.java
+++ b/common/http/src/main/java/io/helidon/common/http/MediaType.java
@@ -337,9 +337,7 @@ public final class MediaType implements AcceptPredicate<MediaType> {
         return CHARSET_ATTRIBUTE.equals(attribute) ? Ascii.toLowerCase(value) : value;
     }
 
-    private static MediaType create(String type, String subtype,
-            Map<String, String> parameters) {
-
+    private static MediaType create(String type, String subtype, Map<String, String> parameters) {
         Objects.requireNonNull(type, "Parameter 'type' is null!");
         Objects.requireNonNull(subtype, "Parameter 'subtype' is null!");
         Objects.requireNonNull(parameters, "Parameter 'parameters' is null!");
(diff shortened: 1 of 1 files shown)


common/http/src/main/java/io/helidon/common/http/MediaType.java
```diff
@@ -268,21 +351,26 @@ private static MediaType create(String type, String subtype,
...
-                .parameters(builder)
-                .build();
+        MediaType mediaType = null;
 
         // Return one of the constants if the media type is a known type.
-        //TODO or else get?
-        return Optional.ofNullable(KNOWN_TYPES.get(mediaType)).orElse(mediaType);
+        if (normalizedParameters.isEmpty()) {
+            mediaType = KNOWN_TYPES.get(normalizedType + '/' + normalizedSubtype);
+        }
```
ljnelson, line 365: You can avoid the `HashMap` creation in probably the vast majority of cases by testing `parameters.isEmpty()` first.  Then, you can move line 363 up as well and simplify.  e.g.:
```
if (parameters.isEmpty()) {
  mediaType = KNOWN_TYPES.get(normalizedType + '/' + normalizedSubtype);
} else {
  Map<String, String> normalizedParameters = new HashMap<>();
  for (Map.Entry<String, String> entry : parameters.entrySet()) {
    String attribute = Tokenizer.normalize(TOKEN_MATCHER, entry.getKey());
    normalizedParameters.put(attribute, normalizeParameterValue(attribute, entry.getValue()));
  }
}
if (mediaType == null) {
  mediaType = MediaType.builder()
    .type(normalizedType)
    .subtype(normalizedSubtype)
    .parameters(normalizedParameters)
    .build();
}
```

romain-grecourt: fixed.

PR #2309, opened 2020-09-01
Webclient uses relative uri in request instead of full uri

### LINKED ISSUE #2302: WebClient uses absolute request URI in requests
## Environment Details
* Helidon Version: 2.0.1
* Helidon SE
* JDK version: 11.0.4
* OS: any

----------

## Problem Description
Helidon WebClient uses an absolute request URI rather than a relative one when submitting HTTP requests. Most of times this is not an issue, but there are some HTTP server implementation that are not able to deal with it.

For example, due to this issue using Helidon WebClient I'm not able to get an OAuth Token from Oracle Identity Cloud because my requests fail with a 404.

The requests issued by Helidon WebClient look like the following:
```
POST <idcs-host>/oauth2/v1/token HTTP/1.1
Accept: text/plain
Accept: application/json
Content-Type: application/x-www-form-urlencoded
host: <idcs-host>
user-agent: Helidon/2.0.1 (java 11.0.4+10-LTS)
```
While instead the expected format should be:
```
POST /oauth2/v1/token HTTP/1.1
Accept: text/plain
Accept: application/json
Content-Type: application/x-www-form-urlencoded
host: <idcs-host>
user-agent: Helidon/2.0.1 (java 11.0.4+10-LTS)
```

[Section 5 of RFC 2116](https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html) is a bit ambiguous about the Request-URI format responsibilities:
> To allow for transition to absoluteURIs in all requests in future versions of HTTP, all HTTP/1.1 servers MUST accept the absoluteURI form in requests, even though HTTP/1.1 clients will only generate them in requests to proxies.

To avoid such kind of issues, the Helidon WebClient should used absolute URI only when effectively required, or make this behavior configurable.

## Steps to reproduce
To verify the Request-URI format used by the Helidon Web Client, just enable the logging at finest level to see the request issued towards the HTTP server

Verdent:
Fixes #2302

### DIFF
--- a/tests/integration/webclient/src/test/java/io/helidon/tests/integration/webclient/TracingPropagationTest.java
+++ b/tests/integration/webclient/src/test/java/io/helidon/tests/integration/webclient/TracingPropagationTest.java
@@ -81,7 +81,7 @@ class TracingPropagationTest {
         assertThat(wsSpan.operationName(), is("HTTP Request"));
         tags = wsSpan.tags();
         assertThat(tags.get(Tags.HTTP_METHOD.getKey()), is("GET"));
-        assertThat(tags.get(Tags.HTTP_URL.getKey()), is(uri));
+        assertThat(tags.get(Tags.HTTP_URL.getKey()), is("/greet"));
         assertThat(tags.get(Tags.HTTP_STATUS.getKey()), is(200));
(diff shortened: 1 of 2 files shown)


tomas-langer (approved):
LGTM

Verdent:
Follow-up issue #2318

PR #2311, opened 2020-09-01
Helidon MP OIDC Security provider guide

tvallin:

### DIFF
--- a/docs/mp/guides/05_security-oidc.adoc
+++ b/docs/mp/guides/05_security-oidc.adoc
@@ -1,0 +1,117 @@
+///////////////////////////////////////////////////////////////////////////////
+
+    Copyright (c) 2020 Oracle and/or its affiliates.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+///////////////////////////////////////////////////////////////////////////////
+
+= Helidon MP OIDC Security Provider
+:h1Prefix: MP
+:description: Helidon OIDC Security Provider guide
+:keywords: helidon, security, guide, oidc, provider
+
+This guide describes how to set up Keycloak and Helidon
+to secure an application with OIDC security provider.
+
+== What you need
+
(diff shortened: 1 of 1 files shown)


docs/mp/guides/05_security-oidc.adoc
```diff
@@ -0,0 +1,517 @@
...
+Use the Helidon MP Maven archetype to create a simple project. It will be used as an example
+to show how to set up Helidon.
+
+[source,bash,subs="attributes+"]
+.Run the Maven archetype
+----
+mvn -U archetype:generate -DinteractiveMode=false \
+    -DarchetypeGroupId=io.helidon.archetypes \
+    -DarchetypeArtifactId=helidon-quickstart-mp \
+    -DarchetypeVersion={helidon-version} \
```
Verdent, line 208: I think you should mention here, that they have to replace {helidon-version} with the version of Helidon they want to use.

docs/mp/guides/05_security-oidc.adoc
```diff
@@ -0,0 +1,517 @@
...
+----
+<dependency>
+    <groupId>io.helidon.microprofile</groupId>
+    <artifactId>helidon-microprofile-oidc</artifactId>
+</dependency>
+----
+
+==== Add OIDC security properties
+
+The OIDC security provider properties can be joined to helidon property file.
```
Verdent, line 235: I think that calling it configuration instead of properties is better here since you are adding OIDC security provider configuration and you are adding those to helidon configuration file. And I might also suggest to add here some note where this application.yaml will be located.

docs/mp/guides/05_security-oidc.adoc
```diff
@@ -0,0 +1,517 @@
...
+----
+security:
+  providers:
+    - abac:
+      # Adds ABAC Provider - it does not require any configuration
+    - oidc:
+        redirect-uri: "/oidc/redirect/*"
+        audience: "account"
+        client-id: "myClientID"   // <1>
+        client-secret: "Client secret generated into Keycloak client credential"
```
Verdent, line 249: I might suggest to add some note as in client-id etc. that user should insert that generated hash here.

Verdent (changes requested):

docs/mp/guides/05_security-oidc.adoc
```diff
@@ -0,0 +1,517 @@
...
+----
+
+The server has now one security provider, basic authentication configured.
+Next step is to modify the test to check that the application is correctly protected.
+
+[source,java]
+.Replace the first webclient call by this one:
+----
+Response response = client
+        .target(getConnectionString("/greet"))
```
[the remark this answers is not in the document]
tvallin: It is a private method from MainTest class to return application endpoint and append a path to it.

Verdent (approved):

PR #2314, opened 2020-09-02
Integrate build-tools 2.1.1

romain-grecourt:


archetypes/catalog/cli-metadata.properties.mustache
```diff
@@ -13,8 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-build-tools.version={{buildtoolsVersion}}
+build-tools.version=2.0.2
```
tomas-langer, line 16: Seems to be the wrong version here, shouldn't this be the placeholder?

archetypes/catalog/cli-metadata.properties.mustache
```diff
@@ -13,8 +13,9 @@
...
+cli.latest.plugin.version={{cliPluginVersion}}
 cli.version={{cliVersion}}
 {{#cliUpdateMessages}}
-cli.{{version}}.message={{message}}
+    cli.{{version}}.message={{message}}
```
tomas-langer, line 20: Is the whitespace intentional?

pom.xml
```diff
@@ -120,6 +120,8 @@
         <version.plugin.wiremock>2.14.0</version.plugin.wiremock>
         <version.plugin.echo>0.4.0</version.plugin.echo>
 
+        <version.helidon-cli>2.1.0</version.helidon-cli>
```
tomas-langer, line 123: Is the `cli` released separately? If not, this should be the same version as `build-tools`

tomas-langer (changes requested):

tomas-langer:
This PR is needed for `2.0.2` release, as otherwise the Helidon maven plugin fails to generate native image (at least on my machine for `mp-1` integration test).

archetypes/catalog/cli-metadata.properties.mustache
barchetta: Romain said this was intentional as it was "the version of the helidon-maven-plugin that contains the goal “dev” before we split it into helidon-cli-maven-plugin". And that this is used only for testing. We might want to follow-up on this, but I don't think it should block the PR.

pom.xml
barchetta: No change was made to the CLI plugin, so he left this version the same.

archetypes/catalog/cli-metadata.properties.mustache
romain-grecourt: This property is correct. It corresponds to the older version of `helidon-maven-plugin` that contained the goal `dev`, it is used for backward compatibility. Also note that this file is only used for testing.

archetypes/catalog/cli-metadata.properties.mustache
romain-grecourt: Yes, this is a mustache loop ; thus the indentation.

pom.xml
romain-grecourt: The CLI wasn't released as part of 2.1.1 because no changes were made to the CLI ; thus the difference in versions.

PR #2322, opened 2020-09-03
Use Config.getPropertyNames() instead of retrieving getPropertyNames() from ConfigSources

ljnelson:


tjquinno (approved):
After quite an enlightening discussion with Laird over slack, LGTM.

PR #2335, opened 2020-09-08
Support for methods in outbound targets.

### LINKED ISSUE #2292: Security outbound targets should honor methods as well
Currently we can configure different outbound configuration based on transport, host, and path.
We should also support HTTP method, so we can have a different configuration based on the outbound HTTP method (we may have a public endpoint to get data, but protected to PUT/POST/DELETE data).

tomas-langer:

Resolves #2292


Verdent (approved):
LGTM

PR #2348, opened 2020-09-09
Implementation of MP FT 2.1.1 using FT SE

spericas:
Replacing FailSafe and Hystrix by our own implementation of FT primitives. Some minor changes to our first version of these primitive operations was necessary to be fully compatible with MP and pass all the TCKs.


fault-tolerance/src/main/java/io/helidon/faulttolerance/AsyncImpl.java
```diff
@@ -35,14 +36,17 @@
...
+            taskFuture = executor.get().submit(task);
         } catch (Throwable e) {
             // rejected execution and other executor related issues
             return Single.error(e);
         }
 
-        return Single.create(future);
+        Single<T> single = Single.create(future, true);
+        single.onCancel(() -> taskFuture.cancel(false));     // cancel task
+        return single;
```
danielkec, line 49: This looks like duplication of `CompletableFuture.supplyAsync` internals, is there some `AsyncTask`'s feature I missed @tomas-langer, or we can simplify?
```java
    public <T> Single<T> invoke(Supplier<T> supplier) {
        try {
            return Single.create(CompletableFuture.supplyAsync(task, executor.get()), true);
        } catch (Throwable e) {
            // rejected execution and other executor related issues
            return Single.error(e);
        }
    }
```

fault-tolerance/src/main/java/io/helidon/faulttolerance/ErrorChecker.java
```diff
@@ -22,25 +22,21 @@
...
-                                                           + "This cannot be correctly handled; skipOn: " + skipOn
-                                                           + " applyOn: " + applyOn);
-            }
+        return throwable -> containsThrowable(skipOnSet, throwable)
+                || !applyOnSet.isEmpty() && !containsThrowable(applyOnSet, throwable);
+    }
 
-        }
+    private static boolean containsThrowable(Set<Class<? extends Throwable>> set, Throwable throwable) {
+        return set.stream().filter(t -> t.isAssignableFrom(throwable.getClass())).count() > 0;
```
danielkec, line 40: `anyMatch` could save few cycles

fault-tolerance/src/main/java/io/helidon/faulttolerance/TimeoutImpl.java
```diff
@@ -16,34 +16,117 @@
...
+                        monitorStarted.complete(null);
+                        return new TimeoutSingleNever();        // new instance
+                    })
+                    .exceptionally(it -> {
+                        if (callReturned.compareAndSet(false, true)) {
+                            future.completeExceptionally(new TimeoutException("Method interrupted by timeout"));
+                            thisThread.interrupt();
+                        }
+                        return null;
+                    });
```
danielkec, line 83:
```java
                    .onError(it -> {
                        if (callReturned.compareAndSet(false, true)) {
                            future.completeExceptionally(new TimeoutException("Method interrupted by timeout"));
                            thisThread.interrupt();
                        }
                    });
```

fault-tolerance/src/main/java/io/helidon/faulttolerance/ErrorChecker.java
spericas: Good point

fault-tolerance/src/main/java/io/helidon/faulttolerance/AsyncImpl.java
spericas: Maybe, but not sure it is more readable tbh

fault-tolerance/src/main/java/io/helidon/faulttolerance/AsyncImpl.java
```diff
@@ -35,14 +36,17 @@
...
-            executor.get().submit(task);
+            taskFuture = executor.get().submit(task);
         } catch (Throwable e) {
             // rejected execution and other executor related issues
             return Single.error(e);
         }
 
-        return Single.create(future);
+        Single<T> single = Single.create(future, true);
+        single.onCancel(() -> taskFuture.cancel(false));     // cancel task
```
akarnokd, line 48: This doesn't look right. I think `Single` operators are still immutable and create a new instance, thus returing `single` will not have the cancel operation. You'd probably want `return single.onCancel(() -> taskFuture.cancel(true));`.

spericas: I discussed this with @danielkec. Code works because of how we interact with `Future` but returning the `Single` from `onCancel` is the correct way. Fixed.

fault-tolerance/src/main/java/io/helidon/faulttolerance/AsyncImpl.java
spericas: Code has changed now, see below

fault-tolerance/src/main/java/io/helidon/faulttolerance/ErrorChecker.java
```diff
@@ -22,25 +22,21 @@
...
-            }
-        } else {
-            if (applyOn.isEmpty()) {
-                return throwable -> skipOn.contains(throwable.getClass());
-            } else {
-                throw new IllegalArgumentException("You have defined both skip and apply set of exception classes. "
-                                                           + "This cannot be correctly handled; skipOn: " + skipOn
-                                                           + " applyOn: " + applyOn);
-            }
+        return throwable -> containsThrowable(skipOnSet, throwable)
```
tomas-langer, line 35: There is a reason for copying the sets - this method is called from a builder, which has mutable properties. This should work only on the data at the point the method was called and should ignore future mutation of these sets.

spericas: Make sense, will fix

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/TimeoutTest.java
```diff
@@ -79,7 +94,7 @@ public void testForceTimeoutLoop() {
         try {
             bean.forceTimeoutLoop();        // cannot interrupt
         } catch (TimeoutException e) {
-            assertThat(System.currentTimeMillis() - start, is(greaterThan(2000L)));
+            assertThat(System.currentTimeMillis() - start, is(greaterThanOrEqualTo(2000L)));
```
danielkec, line 97: What a chance 🙂

danielkec: Got me curious is there some rounding involved?

spericas: No, but this is special case where the thread continues to run and cannot be interrupted (note that timeout is 1000). It hit 2000 at least once :)

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
```diff
@@ -260,6 +264,19 @@ void registerFaultToleranceMetrics(@Observes AfterDeploymentValidation validatio
                 }
             });
         }
+
+        // Initialize executors for MP FT - default size of 16
+        io.helidon.config.Config config = io.helidon.config.Config.create();
+        FaultTolerance.scheduledExecutor(ScheduledThreadPoolSupplier.builder()
+                .threadNamePrefix("ft-mp-schedule-")
+                .corePoolSize(16)
```
danielkec, line 272: `16 is the default right?

spericas: I think so. But defaults can change and a smaller number will likely result in FT TCK failures.

fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
```diff
@@ -266,7 +276,17 @@ private FtHandlerTypedImpl(List<FtHandlerTyped<T>> validFts) {
...
+            }
+
+            @Override
+            public String toString() {
+                StringBuilder sb = new StringBuilder();
+                for (int i = validFts.size() - 1; i >= 0; i--) {
+                    sb.append(validFts.get(i).toString());
+                    sb.append("\n");
+                }
+                return sb.toString();
```
danielkec, line 289: `validFts` is always LinkedList, using `descendingIterator` would be nice

danielkec (approved):

PR #2352, opened 2020-09-11
Messaging health check

danielkec:
Add messaging health check so it is possible detect failed channels asap, if any channel intercept **onError** or **cancel** signals, whole messaging gets marked as DOWN
```json
{
  "outcome": "UP",
  "status": "UP",
  "checks": [
    {
      "name": "messaging",
      "state": "UP",
      "status": "UP",
      "data": {
        "test-channel-1": "UP",
        "test-channel-2": "UP"
      }
    }
  ]
}
```


microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/UniversalChannel.java
```diff
@@ -67,6 +69,10 @@ void setOutgoing(OutgoingMethod outgoingMethod) {
         this.outgoingMethod = outgoingMethod;
     }
 
+    AtomicBoolean isUp() {
+        return up;
```
tomas-langer, line 73: Is there a possibility `isUp` is called before `connect`? If so, this would return null and most likely cause a NPE

tomas-langer: Also `up` is not volatile. Even if the field is AtomicBoolean, access to it must be correctly synchronized if the field may be created in a different thread.
Maybe the AtomicBoolean should be final and true and sent to `Probe.connect` as a parameter

danielkec: oj, thx!

microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/UniversalChannel.java
```diff
@@ -69,8 +70,12 @@ void setOutgoing(OutgoingMethod outgoingMethod) {
         this.outgoingMethod = outgoingMethod;
     }
 
-    AtomicBoolean isUp() {
-        return up;
+    AtomicBoolean isLive() {
```
tomas-langer, line 73: This is leaking internal state - this method may just return a `boolean`

tomas-langer (approved):
Make sure the atomic boolean may be updated only where it makes sense.

microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/UniversalChannel.java
danielkec: package private, but true, I've got changes in here from another PR, will fix it there

PR #2356, opened 2020-09-14
Fix issue with gRPC clients and services where method signatures have types with nested generics

thegridman:
This is a back-port of fix applied to 2.x in PR #2283


tomas-langer (approved):
LGTM

PR #2365, opened 2020-09-16
Cloud Functions (Google, Azure, OCI and AWS)

jbescos:
Related to this issue https://github.com/oracle/helidon/issues/2342

This PR includes support for Google Cloud Functions, Azure Functions and AWS Lambda

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -883,6 +883,37 @@
                 <artifactId>helidon-microprofile-tests-junit5</artifactId>
                 <version>${helidon.version}</version>
             </dependency>
+            <!-- Cloud -->
+            <dependency>
+                <groupId>io.helidon.microprofile.cloud</groupId>
+                <artifactId>helidon-microprofile-cloud-common</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.microprofile.cloud</groupId>
+                <artifactId>helidon-microprofile-cloud-google-cloud-http-function</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.microprofile.cloud</groupId>
+                <artifactId>helidon-microprofile-cloud-google-cloud-background-function</artifactId>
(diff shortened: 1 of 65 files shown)


jbescos:
This should be ready for review. But I still need to deploy this in real cloud environments and fix the check styles.

jbescos:
This is deployed in Google Cloud. It will upper case the JSON:
curl --header "Content-Type: application/json" --request POST --data '{"username":"jorge"}' https://us-central1-botlogic-140309.cloudfunctions.net/http

jbescos:
With Azure and AWS there is one problem, it seems we need to be able to deliver one unique jar file.
With maven-shade-plugin I was not able to make it work.

jbescos:
For Azure we need to create also an unique jar and then deploy it with azure-functions-maven-plugin.

The plugin fails with:
[ERROR] Failed to execute goal com.microsoft.azure:azure-functions-maven-plugin:1.9.0:deploy (default-cli) on project helidon-microprofile-cloud-azure-functions: Invalid signature file digest for Manifest main attributes

Previous error happens if maven-shade-plugin is added. Otherwise it doesn't fail, but the jar is missing classes, so it doesn't work.

jbescos:
In OCI it is working too:
`curl --noproxy "*" -d 'test' -X POST https://oupcn7bdm7kzjtun6dpse2mcpe.apigateway.us-phoenix-1.oci.customer-oci.com/uppercase/uppercase`

Although it is **REALLY** difficult to configure it in OCI (I spent around 3 days). Some notes to make the path easier for anyone who wants to try it:
- Create a dockerfile and upload the image in OCI. (Example of dockerfile attached)
- Create a function from that docker image.
- Create the VCN with public subnet and with internet access (there are some security rules to allow the access, by default you cannot access it).
- Create a gateway API to use the function.
- Add the policies to allow the gateway to use the function.
[helidon-playground.zip](https://github.com/oracle/helidon/files/5464005/helidon-playground.zip)

jbescos:
Closing and opening a new one because of a copy-right plugin issue:
https://github.com/oracle/helidon/pull/2747

PR #2379, opened 2020-09-21
Enable support for HTTP compression in the webserver

spericas:
Enable support for HTTP compression in the webserver. Netty provides support for gzip and deflate encodings. Some new tests for SE and MP.


webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -751,6 +751,17 @@ public Builder workersCount(int workers) {
             return this;
         }
 
+        /**
+         * Configure whether to enable content negotiation for compression.
```
tomas-langer, line 755: This should maybe say "for default socket", as this is socket specific, if I understood this PR correctly

tomas-langer (changes requested):

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -751,6 +751,17 @@ public Builder workersCount(int workers) {
...
+         *
+         * @param value compression flag
+         * @return updated builder instance
+         */
+        public Builder enableCompression(boolean value) {
```
tomas-langer, line 760: This method should be annotated with `@Override` and the javadoc is not necessary. If the javadoc is there, please add "negotiation for compression for the default socket"

tomas-langer (changes requested):

tomas-langer (approved):
LGTM

PR #2388, opened 2020-09-23
Update gRPC Version

thegridman:
The gRPC Java libraries have a dependency on Netty. This PR is to update to the latest gRPC Java version to match with the version of Netty being used by the rest of Helidon.


barchetta:
@thegridman you might want to do this on the helidon-1.x branch as well (Netty was upgraded there as well).

tomas-langer (approved):

tomas-langer:
@barchetta could you please have a look at the dependency version and merge if OK? Thanks!

Issue #2371, opened 2020-09-16
Intermittent java.lang.IllegalStateException: DataChunkInputStream needs to be handled in separate thread to prevent deadlock


jansupol:
A single [test]( https://github.com/jansupol/helidon.client) runs 1000x and intermittently fails.

jansupol:
The stack trace:

>java.lang.IllegalStateException: DataChunkInputStream needs to be handled in separate thread to prevent deadlock.
        at io.helidon.media.common.DataChunkInputStream.validate(DataChunkInputStream.java:190)
        at io.helidon.media.common.DataChunkInputStream.read(DataChunkInputStream.java:123)
        at java.base/java.io.FilterInputStream.read(FilterInputStream.java:133)
        at org.glassfish.jersey.message.internal.EntityInputStream.read(EntityInputStream.java:79)
        at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$UnCloseableInputStream.read(ReaderInterceptorExecutor.java:273)
        at java.base/sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
        at java.base/sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
        at java.base/sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
        at java.base/java.io.InputStreamReader.read(InputStreamReader.java:185)
        at java.base/java.io.Reader.read(Reader.java:229)
        at org.glassfish.jersey.message.internal.ReaderWriter.readFromAsString(ReaderWriter.java:152)
        at org.glassfish.jersey.message.internal.ReaderWriter.readFromAsString(ReaderWriter.java:137)
        at org.glassfish.jersey.message.internal.AbstractMessageReaderWriterProvider.readFromAsString(AbstractMessageReaderWriterProvider.java:94)
        at org.glassfish.jersey.message.internal.StringMessageProvider.readFrom(StringMessageProvider.java:54)
        at org.glassfish.jersey.message.internal.StringMessageProvider.readFrom(StringMessageProvider.java:36)
        at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$TerminalReaderInterceptor.invokeReadFrom(ReaderInterceptorExecutor.java:233)
        at org.glassfish.jersey.message.internal.ReaderInterceptorExecutor$TerminalReaderInterceptor.aroundReadFrom(ReaderInterceptorExecutor.java:212)
[...]

PR #2401, opened 2020-09-30
DataChunkedInputStream deadlock protection removed

Verdent:
#2371


tomas-langer (approved):
LGTM
Would be good to update javadoc

danielkec (dismissed):
We were able to reproduce the deadlock even with reactive stream supplied to callback, 
image

It should be highlighted in the doc, that reactive callbacks should be never blocked

danielkec (approved):

tomas-langer (approved):

PR #2404, opened 2020-09-30
Name support for all those operations that implement FtHandler

spericas:
Name support for all those operations that implement FtHandler. These names can be used for debugging, error reporting and future config support. Fixed some warnings too.


tomas-langer (approved):
LGTM

PR #2417, opened 2020-10-06
Initial stab at incorporating some Loom features into Helidon in a way that will work under JDK 11 or 16-loom-ea

ljnelson:


barchetta (approved):
LGTM. I tried this with native-image and had no issues.

Issue #2418, opened 2020-10-07
Helidon attempts to load vetoed JAX-RS application


mhorejsi:
# Environment Details
* Helidon Version: 2.0.2
* Helidon MP
* JDK version: 11.0.8
* OS: OS X El Capitan

----------

## Problem Description
If I write a custom CDI extension that `@Observes ProcessAnnotatedType<T>` with [lower priority](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#observer_ordering) than Helidon's `JaxRsCdiExtension` and `veto()` my JAX-RS application in the observer method, Helidon ignores the veto and attempts to load it. This attempt fails because my JAX-RS application injects other vetoed CDI beans.

This bug wasn't present in Helidon 1.4.x

spericas:
Would make sense as there were changes in that area in 2.x. Will investigate.

spericas:
See PR #2429

PR #2429, opened 2020-10-08
Allow user-provided CDI extensions to veto JAX-RS classes 

spericas:
Allow user-provided CDI extensions to veto JAX-RS classes so that they are not included as part of an application during discovery. Changed JaxRsCdiExtension to use a different event to process classes. Updated the bookstore-mp test to verify the veto process and also modernized its tests.


spericas:
See issue #2418

ljnelson (approved):
Very nicely done.

PR #2438, opened 2020-10-12
CDI.current().getBeanManager() now available during shutdown

### LINKED ISSUE #2421: CDI container is not available during shutdown
When the `container.close()` is called, we may get an exception from `HelidonContainerImpl` that `Container not running`.

This is because we set the "not running" before finalizing the container work. One of the things it does is that it triggers events (such as Application scoped before destroyed event) that may in turn instantiate beans. If such a bean uses `CDI.current()`, it breaks.

To reproduce:

- create a bean that is not used by the application at all
- add the following method
  ```java 
  public void method(@Observes @BeforeDestroyed(ApplicationScoped.class) Object event) {
    CDI.current().getBeanManager()
  }
  ```
 - start the container
 - stop the container

The "stop" part should throw an exception as described above.
Correct behavior:
The `CDI.current()` should work (maybe we could issue a single warning with stacktrace, such as "container is shutting down, yet CDI was requested)

tomas-langer:
Resolves #2421 

### DIFF
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
@@ -402,9 +402,12 @@ final class HelidonContainerImpl extends Weld implements HelidonContainer {
             if (!isRunning.get()) {
                 LOGGER.warning("BeanManager requested during container shutdown. This may be caused by observer methods "
                                        + "that use CDI.current(). Switch to finest logging to see stack trace.");
-                LOGGER.log(Level.FINEST,
-                           "Invocation of container method during shutdown",
-                           new IllegalStateException("Container not running"));
+                if (LOGGER.isLoggable(Level.FINEST)) {
+                    // this should not be common, but guarding, so we do not fill stack trace unless necessary
+                    LOGGER.log(Level.FINEST,
+                               "Invocation of container method during shutdown",
+                               new IllegalStateException("Container not running"));
+                }
             }
 
             return new BeanManagerProxy(beanManager());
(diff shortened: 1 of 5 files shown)


microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
```diff
@@ -399,10 +399,15 @@ public boolean isRunning() {
...
         @Override
         public BeanManager getBeanManager() {
-            if (isRunning.get()) {
-                return new BeanManagerProxy(beanManager());
+            if (!isRunning.get()) {
+                LOGGER.warning("BeanManager requested during container shutdown. This may be caused by observer methods "
+                                       + "that use CDI.current(). Switch to finest logging to see stack trace.");
+                LOGGER.log(Level.FINEST,
+                           "Invocation of container method during shutdown",
+                           new IllegalStateException("Container not running"));
```
danielkec, line 407: Probably not worth it but can be performing better when avoiding `Throwable.fillInStackTrace()`
```java
    if (LOGGER.isLoggable(Level.FINEST)) {
        LOGGER.log(Level.FINEST, "Invocation of container method during shutdown",
                new IllegalStateException("Container not running"));
    }
```

danielkec (approved):

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
tomas-langer: Fixed

danielkec (approved):

Verdent (approved):

PR #2439, opened 2020-10-12
Attempt to start a stopped server will fail with IllegalStateException

### LINKED ISSUE #377: question about restart webserver
Currently it look's like WebServer cannot perform "restart" ( sequence `startup()->shoutdown()->startup()` ).
It is intended behaviour or something else?

In case that this behaviour is intended "Helidon" should throw exception when second time `startup()` method is called or maybe `hide`  shoutdown() method from API because`WebServer.whenShoutdown()` is enough?

tomas-langer:
Resolves #377 


tomas-langer:
Original behavior was to silently fail at starting the server.

danielkec (approved):

PR #2453, opened 2020-10-15
Messaging discrepancies

### LINKED ISSUE #2326: Messaging method signatures discrepancies
- [x] NPE with signature 
```java
@Incoming("fromJms")
@Acknowledgment(Acknowledgment.Strategy.MANUAL)
public void wrapSseEvent(org.eclipse.microprofile.reactive.messaging.Message<String> msg) throws JMSException, SQLException {
```

- missing NPE check [AbstractMessagingMethod.java:61](https://github.com/oracle/helidon/blob/master/microprofile/messaging/src/main/java/io/helidon/microprofile/messaging/AbstractMessagingMethod.java#L61)

- Signature `void method(Message<I> msg)` not supported by spec

- [x] Unwrapped payload leaks in engine(and to connector) with signature:
```java
@Incoming("multiplyVariants")
@Outgoing("toJms")
public ProcessorBuilder<String, String> multiply() {
     return ReactiveStreams.<String>builder();
}
```
- [x] Should throw DeploymentException for unsupported signature 
```java
@Incoming("fromJms")
@Outgoing("toJms")
@LRA(value = LRA.Type.REQUIRES_NEW)
public ProcessorBuilder<String, Message<String>> placeOrder(String msg)
```

- Actually shouldn't, it should and resolves as signature `O method(I payload)` which is totally legal

danielkec:
Fixes #2326

### DIFF
--- a/microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/AbstractMessagingMethod.java
+++ b/microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/AbstractMessagingMethod.java
@@ -17,11 +17,13 @@
 
 package io.helidon.microprofile.messaging;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Optional;
 
 import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.BeanManager;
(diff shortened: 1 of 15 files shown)


tomas-langer (approved):
LGTM

PR #2472, opened 2020-10-21
Add dependency-check-maven plugin

barchetta:
Configure owasp dependency-check-maven plugin in project to make it easier to run dependency scanning.


barchetta:
Artifact exclusions don't seem to be working correctly.

pom.xml
```diff
@@ -491,13 +493,46 @@
                         <xmlOutput>true</xmlOutput>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.owasp</groupId>
+                    <artifactId>dependency-check-maven</artifactId>
+                    <version>${version.plugin.dependency-check}</version>
+                    <configuration>
+                        <skip>${dependency-check.skip}</skip>
+                        <skipProvidedScope>true</skipProvidedScope>
```
tomas-langer, line 502: I think both provided and runtime scopes are valid, as these are still helidon dependencies.

etc/scripts/dependency-check.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash -e
```
tomas-langer, line 1: Could we use a different name? `dependency` sounds like general dependency checking (such as to make sure we do not have conflicting versions). Could this be `security-check.sh` or maybe even `owasp-check.sh`?

etc/dependency-check-suppression.xml
```diff
@@ -0,0 +1,131 @@
...
+   <notes><![CDATA[
+   file name: jsonp-jaxrs-1.1.6.jar
+   ]]></notes>
+   <packageUrl regex="true">^pkg:maven/org\.glassfish/jsonp\-jaxrs@.*$</packageUrl>
+   <cpe>cpe:/a:processing:processing</cpe>
+</suppress>
+
+<!-- We use snakeyaml for reading local configuration, so this CVE
+     does not apply to our use case. For more information:
+     https://bitbucket.org/asomov/snakeyaml/issues/377/allow-configuration-for-preventing-billion
```
tomas-langer, line 24: We can upgrade to 1.26 that should not be affected at all (not sure if the scan knows about this). From what I see in the commit, the default behavior is not to allow recursive keys: https://bitbucket.org/asomov/snakeyaml/commits/da11ddbd91c1f8392ea932b37fa48110fa54ed8c

pom.xml
barchetta: Fixed.

etc/scripts/dependency-check.sh
barchetta: Fixed. Now `owasp-dependency-check`

etc/dependency-check-suppression.xml
barchetta: Yes. See issue https://github.com/oracle/helidon/issues/2475

romain-grecourt (approved):

PR #2491, opened 2020-10-30
Support for a max payload limit on client requests

spericas:
New `max-payload-size` property in server's configuration to limit size of a client request's payload, using content length or chunked encoding. Default value for property is -1, no limit. Unit for size is bytes.


romain-grecourt (approved):

ljnelson (approved):
This looks fine.  One thing that does stand out is: the header talks about "content" (`Content-Length`), the error code mentions "payload" and various other things mention "entity".  We might want to think about normalizing our nouns in this area.

spericas:
> This looks fine. One thing that does stand out is: the header talks about "content" (`Content-Length`), the error code mentions "payload" and various other things mention "entity". We might want to think about normalizing our nouns in this area.

It's a good point and I struggled with that too. If you search for HTTP error 413, it will be described as both "payload too large" and "entity too large", so they are partly to blame as well.

barchetta (approved):

PR #2493, opened 2020-11-02
Use ProcessManagedBean (not ProcessAnnotatedType) to create metrics - honors vetoes

### LINKED ISSUE #2492: Use ProcessManagedBeans in metrics CDI extension to support vetoing
See #2431 

The metrics CDI portable extension does not honor vetoing of beans by other extensions, because it defines metrics based on annotations via observers of `ProcessAnnotatedType`. It should wait to actually create the metrics until `ProcessBean` time to honor other extensions' vetoes of beans.

tjquinno:
There are two overall changes so that we honor other CDI extensions' vetoes of beans.

1. Previously, we defined metrics for annotated elements in the `ProcessAnnotatedType` observer. This incorrectly included beans which other extension might have vetoed.

    Instead,  now the `ProcessAnnotatedType` observer records the Java class for each annotated type identified by CDI as being annotated with a metrics annotation.
 
    Later, a new `ProcessManagedBean` observer ignores any beans for which the Java class was not previously recorded.

    We do this in two steps because CDI invokes the `ProcessBean` observer for _every_ bean. We have to check each constructor and method in every bean we process to look for metrics annotations, and this two-step approach allows us to avoid scanning constructors and methods of beans that CDI did not previously identify has having metrics annotations.

1. A relatively recent change to MP metrics includes automatically adding (if configured) a `SimpleTimer` metric to every JAX-RS endpoint method. Previously, we did this by:
    1. Using a `ProcessAnnotatedType` observer on the HTTP operation annotations (`@GET`, `@PUT`, etc.) which would store each `Method` that had one of the HTTP method annotations and add a synthetic annotation `@SyntheticSimplyTimed` to the CDI `AnnotatedMethod` for the method.
    1. Using an `AfterBeanValidation` observer to go through each such saved `Method` and create the corresponding `SimpleTimer` metric.  

    With this PR, the `ProcessAnnotatedType` observer only adds the synthetic annotation to the relevant methods in the `AnnotatedType`.


tjquinno:
Obsolete. These changes have been incorporated into the metrics (and several other) CDI extensions already.

PR #2495, opened 2020-11-03
Request scope propagation for non-async calls in FT

spericas:
Use a wrapped supplier to propagate request scope even for synchronous calls as some FT primitives (like Retry) use executors and may run code in a different thread. Updated new test that fails without this change and passes with it.


tjquinno (approved):
LGTM.

tests/functional/request-scope/src/main/java/io/helidon/tests/functional/requestscope/RequestTestQualifier.java
```diff
@@ -0,0 +1,36 @@
...
+@RequestScoped
+@TestQualifier
+public class RequestTestQualifier {
+
+    @Inject
+    TenantContext tenantContext;
+
+    public String test() throws Exception {
+        String tenantId = tenantContext.getTenantId();
+        System.out.println("Tenant Context: " + tenantId);
```
ljnelson, line 30: Intentional?

ljnelson (approved):

PR #2505, opened 2020-11-05
Add a shutdown hook to executor service in JerseySupport

spericas:
Add a shutdown hook to executor service in JerseySupport. This should prevent threads in this executor to attempt reads/writes after Netty's executor is also down.


spericas:
This fix seems to help but does not completely eliminate the problem of stack traces in log by Netty. There always will be _a window_ perhaps in which those exceptions can be reported. We should decide whether we want to keep this change or not.

tjquinno (approved):
LGTM

PR #2507, opened 2020-11-05
Add vetoed-bean support to metrics CDI extension

### LINKED ISSUE #2492: Use ProcessManagedBeans in metrics CDI extension to support vetoing

tjquinno:
Resolves #2492 

There are two overall changes so that we honor other CDI extensions' vetoes of beans.

1. Previously, we defined metrics for annotated elements in the `ProcessAnnotatedType` observer. This incorrectly included beans which other extension might have vetoed.

    Instead,  now the `ProcessAnnotatedType` observer records the Java class for each annotated type identified by CDI as being annotated with a metrics annotation.
 
    Later, CDI invokes a new `ProcessManagedBean` observer only for beans which were not vetoed. This by itself takes care of honoring bean vetoes by other extensions.

    But this new observer runs for _every_ bean, whether decorated with a metrics annotation or not. We have to check each constructor and method in every bean we process to look for metrics annotations.

    This new observer will do that processing only if the type in the `ProcessManagedBean` appears in the collection of metrics-annotated classes prepared by the `ProcessAnnotatedType` observer. This observer also maintains a collection of types that _have_ been processes. An `AfterDeploymentValidation` observer method uses the collections to log (FINE) any vetoed beans that the extension decided to _not_ process.

1. A relatively recent change to MP metrics includes automatically adding (if configured) a `SimpleTimer` metric to every JAX-RS endpoint method. Previously, we did this by:

    1. Using a `ProcessAnnotatedType` observer on the HTTP operation annotations (`@GET`, `@PUT`, etc.) which would store each `Method` that had one of the HTTP method annotations and add a synthetic annotation `@SyntheticSimplyTimed` to the CDI `AnnotatedMethod` for the method.

    1. Using an `AfterBeanValidation` observer to go through each such saved `Method` and create the corresponding `SimpleTimer` metric.

    But that would create metrics for methods on vetoed beans.

    With this PR, the `ProcessAnnotatedType` observer:

    1. Adds the synthetic annotation to the relevant methods in the `AnnotatedType`, and
    1. Records a map entry with key = the Java class and value = collection of `Method`s with a JAX-RS operation annotation. 

    A new `ProcessManagedBean` observer checks the Java class of the bean against the map, and if found registers the synthetic `SimpleTimer` metrics for each method stored as the map entry's value.

The net result of these pairs of observers is that the extension _efficiently_ creates the metrics for metrics-annotated and JAX-RS-annotated methods but only for non-vetoed beans.

### DIFF
--- a/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
+++ b/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
@@ -70,8 +70,6 @@ import io.helidon.common.Errors;
 import io.helidon.config.Config;
 import io.helidon.config.ConfigValue;
 import io.helidon.metrics.MetricsSupport;
-import io.helidon.metrics.Registry;
-import io.helidon.metrics.RegistryFactory;
 import io.helidon.microprofile.server.ServerCdiExtension;
 import io.helidon.webserver.Routing;
 
@@ -92,7 +90,6 @@ import org.eclipse.microprofile.metrics.annotation.Counted;
 import org.eclipse.microprofile.metrics.annotation.Gauge;
 import org.eclipse.microprofile.metrics.annotation.Metered;
 import org.eclipse.microprofile.metrics.annotation.Metric;
-import org.eclipse.microprofile.metrics.annotation.RegistryType;
 import org.eclipse.microprofile.metrics.annotation.SimplyTimed;
 import org.eclipse.microprofile.metrics.annotation.Timed;
 
--- a/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
+++ b/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
@@ -64,7 +64,6 @@ import javax.interceptor.Interceptor;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.HEAD;
-import javax.ws.rs.HttpMethod;
 import javax.ws.rs.OPTIONS;
 import javax.ws.rs.PATCH;
 import javax.ws.rs.POST;
@@ -350,7 +349,7 @@ public class MetricsCdiExtension implements Extension {
         AnnotatedType<?> annotatedType = pat.getAnnotatedType();
         Class<?> clazz = annotatedType.getJavaClass();
 
-        // If abstract class, then handled by concrete subclasses
+        // Abstract classes are handled when we deal with a concrete subclass. Also, ignore if @Interceptor is present.
         if (annotatedType.isAnnotationPresent(Interceptor.class)
                 || Modifier.isAbstract(clazz.getModifiers())) {
             LOGGER.log(Level.FINER, () -> "Ignoring " + clazz.getName()
@@ -388,6 +387,12 @@ public class MetricsCdiExtension implements Extension {
         AnnotatedType<?> type =  pmb.getAnnotatedBeanClass();
         Class<?> clazz = type.getJavaClass();
         if (!metricsAnnotatedClasses.contains(clazz)) {
+            return;
+        }
+        // Recheck for Interceptor.
+        if (type.isAnnotationPresent(Interceptor.class)) {
+            LOGGER.log(Level.FINE, "Ignoring metrics defined on type " + clazz.getName()
+                    + " because a CDI portable extension added @Interceptor to it dynamically");
             return;
         }
         metricsAnnotatedClassesProcessed.add(clazz);
@@ -445,7 +450,8 @@ public class MetricsCdiExtension implements Extension {
      * @param pat the {@code ProcessAnnotatedType} for the type containing the JAX-RS annotated methods
      */
     private void recordSimplyTimedForRestResources(@Observes
-                                                   @WithAnnotations({HttpMethod.class})
+                                                   @WithAnnotations({GET.class, PUT.class, POST.class, HEAD.class, OPTIONS.class,
+                                                           DELETE.class, PATCH.class})
                                                            ProcessAnnotatedType<?> pat) {
 
         /// Ignore abstract classes or interceptors. Make sure synthetic SimpleTimer creation is enabled, and if so record the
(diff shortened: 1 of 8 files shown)


microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
```diff
@@ -297,61 +310,124 @@ void before(@Observes BeforeBeanDiscovery discovery) {
...
+     * Checks to make sure the annotated type is not abstract and is not an interceptor.
+     *
+     * @param pat {@code ProcessAnnotatedType} event
+     * @return true if the annotated type should be kept for potential processing later; false otherwise
+     */
+    private boolean checkCandidateMetricClass(ProcessAnnotatedType<?> pat) {
+        AnnotatedType<?> annotatedType = pat.getAnnotatedType();
+        Class<?> clazz = annotatedType.getJavaClass();
 
         // If abstract class, then handled by concrete subclasses
```
ljnelson, line 353: Need to adjust this comment to match the code below it.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
```diff
@@ -297,61 +310,124 @@ void before(@Observes BeforeBeanDiscovery discovery) {
...
-        if (Modifier.isAbstract(clazz.getModifiers())) {
+        if (annotatedType.isAnnotationPresent(Interceptor.class)
```
ljnelson, line 354: Strictly speaking this is not reliable as other observers can add the `@Interceptor` annotation programmatically later.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
```diff
@@ -364,65 +440,55 @@ private void processInjectionPoints(@Observes ProcessInjectionPoint<?, ?> pip) {
...
     /**
-     * Records the need to add a {@code SyntheticSimplyTimed} annotation to each JAX-RS endpoint method.
+     * Adds a {@code SyntheticSimplyTimed} annotation to each JAX-RS endpoint method.
      *
      * @param pat the {@code ProcessAnnotatedType} for the type containing the JAX-RS annotated methods
      */
     private void recordSimplyTimedForRestResources(@Observes
-                                                   @WithAnnotations({GET.class, PUT.class, POST.class, HEAD.class, OPTIONS.class,
-                                                                            DELETE.class, PATCH.class})
+                                                   @WithAnnotations({HttpMethod.class})
```
ljnelson, line 448: Oh, interesting; I guess Jandex takes meta-annotations into account?

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
```diff
@@ -643,53 +726,51 @@ private static boolean chooseRestEndpointsSetting(Config metricsConfig) {
...
-                                       + "this warning.");
+        // Process @Gauge methods keeping non-private declared on this class
+        for (AnnotatedMethod method : type.getMethods()) {
+            Method javaMethod = method.getJavaMember();
+            if (!javaMethod.getDeclaringClass().equals(clazz)
+                    || Modifier.isPrivate(javaMethod.getModifiers())
+                    || !method.isAnnotationPresent(Gauge.class)) {
+                continue;
             }
+            Annotation requestScopedAnnotation = type.getAnnotation(RequestScoped.class);
```
ljnelson, line 750: Better to use `pmb.getBean().getScope()` unless I'm missing something.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
```diff
@@ -643,53 +726,51 @@ private static boolean chooseRestEndpointsSetting(Config metricsConfig) {
...
+            if (requestScopedAnnotation != null) {
+                errors.fatal(clazz, "Cannot configure @Gauge on a request scoped bean");
+                return;
+            }
+            if (type.getAnnotation(ApplicationScoped.class) == null && type.getAnnotation(Singleton.class) == null) {
```
ljnelson, line 755: See above

ljnelson (changes requested):
Just a few minor changes.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
tjquinno: Done.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
tjquinno: I've left that but added a second check in the `ProcessManagedBean` observer, just in case.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
tjquinno: I've switched back just in case there is some oddness that hasn't surfaced in the testing so far.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
tjquinno: Done.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
tjquinno: Done.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
```diff
@@ -445,7 +450,8 @@ private void processInjectionPoints(@Observes ProcessInjectionPoint<?, ?> pip) {
      * @param pat the {@code ProcessAnnotatedType} for the type containing the JAX-RS annotated methods
      */
     private void recordSimplyTimedForRestResources(@Observes
-                                                   @WithAnnotations({HttpMethod.class})
+                                                   @WithAnnotations({GET.class, PUT.class, POST.class, HEAD.class, OPTIONS.class,
```
ljnelson, line 453: (This is fine; when I commented on this previously, I didn't mean necessarily to remove the `HttpMethod` annotation and replace it with the list of options—it was more just that I was surprised that a meta-annotation like that would be processed by `@WithAnnotations`.)

ljnelson (approved):

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricsCdiExtension.java
tjquinno: Yeah, the original code had the list of individual annotations and that worked fine. I happened to see `@HttpMethod` in the def for `Get` (I think) and thought I'd try it and it seemed to work. But that is not related to the functional change at the heart of this PR so it's probably best to not make that change...at least now.

PR #2508, opened 2020-11-08
Concat array enhancement

danielkec:
- More effective cancellation handling
- Better handling of negative requests

[JMH test](https://github.com/danielkec/helidon-jmh/blob/ca-refactor/src/main/java/io/helidon/reactive/jmh/multi/ConcatArrayComparison.java) results:
```
Benchmark                                     Mode  Cnt        Score        Error  Units
ConcatArrayComparison.reqMaxNewCA            thrpt   10   742789.861 ±  25826.167  ops/s
ConcatArrayComparison.reqMaxOldCA            thrpt   10   716121.440 ±  28539.035  ops/s
ConcatArrayComparison.variableRequestsNewCA  thrpt   10  6538445.756 ± 629438.611  ops/s
ConcatArrayComparison.variableRequestsOldCA  thrpt   10  5267535.840 ± 650941.922  ops/s
```

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/MultiConcatArray.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/MultiConcatArray.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/reactive/src/main/java/io/helidon/common/reactive/MultiConcatArray.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/MultiConcatArray.java
@@ -17,8 +17,9 @@
 
 package io.helidon.common.reactive;
 
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.VarHandle;
 import java.util.concurrent.Flow;
-import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Relay items in order from subsequent Flow.Publishers as a single Multi source.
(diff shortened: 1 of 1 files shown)


olotenko (approved):

akarnokd:
The code looks quite unconventional to me and the state transitions are somewhat hard to follow. In general, `concat` can break when there is a switchover from source to source at the same time the downstream requests thus the outstanding request amount is not properly forwarded to them.

Here is a test that passes on master with the `SubscriptionArbiter` but fails on this implementation because some `request(1)`s are not handed over properly:

```java

    static void race(Runnable r1, Runnable r2, ExecutorService exec) {

        AtomicInteger sync = new AtomicInteger(2);
        CountDownLatch cdl = new CountDownLatch(1);

        exec.submit(() -> {
            if (sync.decrementAndGet() != 0) {
                while (sync.get() != 0) ;
            }

            try {
                r1.run();
            } finally {
                cdl.countDown();
            }
        });

        if (sync.decrementAndGet() != 0) {
            while (sync.get() != 0) ;
        }
        r2.run();

        try {
            Assert.assertTrue(cdl.await(5, TimeUnit.SECONDS));
        } catch (InterruptedException ex) {
            Assert.fail("", ex);
        }
    }


    @Test
    public void switchOverRequestRace() {
        var exec = Executors.newSingleThreadExecutor();
        try {
            for (int i = 0; i < 10_000; i++) {
                System.out.printf("Round %d%n", i + 1);
                TestSubscriber<Object> ts = new TestSubscriber<>(0L);
                AtomicReference<Flow.Subscriber<? super Integer>> sref = new AtomicReference<>();

                Multi.concatArray(new Flow.Publisher<Integer>() {

                    @Override
                    public void subscribe(Flow.Subscriber<? super Integer> subscriber) {
                        subscriber.onSubscribe(EmptySubscription.INSTANCE);
                        sref.set(subscriber);
                    }
                }, Multi.range(1, 100)).subscribe(ts);

                race(() -> {
                    sref.get().onComplete();
                }, () -> {
                    var s = ts.getSubcription();
                    for (int j = 0; j < 100; j++) {
                        s.request(1);
                    }
                }, exec);

                ts.awaitDone(5, TimeUnit.SECONDS);
                ts.assertItemCount(100);
            }
        } finally {
            exec.shutdown();
        }
    }
```

olotenko:
Thanks.

The problem with the original code is the lack of atomicity in state transitions around the same place, which can also be shown to lead to problems (not conformant to the specification). Consequently, an attempt to capture all state transitions in a single atomic counter.

common/reactive/src/main/java/io/helidon/common/reactive/MultiConcatArray.java
```diff
@@ -35,31 +36,89 @@
...
+            long p0 = pending;
+            if (p0 == CANCEL) {
+               subscription.cancel();
+               return;
+            }
+
+            produced++; // assert: matching request(1) has been done by nextSource()
+            this.subscription = subscription;
+            // assert: requested == SEE_OTHER
+            REQUESTED.setOpaque(this, p0); // assert: p0 is guaranteed to be a value of requested never seen before
```
olotenko, line 97: Two things need doing:
`long oldProduced = produced;` before this line, then use `oldProduced` to the end of this method.

This addresses the case raised by @akarnok - concurrent `request` used to update `produced`.

and

`REQUESTED.setVolatile(this, p0)` to replace `setOpaque` - opaque doesn't guarantee ordering with other stores.

olotenko (changes requested):

olotenko:
> The code looks quite unconventional to me and the state transitions are somewhat hard to follow. In general, `concat` can break when there is a switchover from source to source at the same time the downstream requests thus the outstanding request amount is not properly forwarded to them.
> 
> Here is a test that passes on master with the `SubscriptionArbiter` but fails on this implementation because some `request(1)`s are not handed over properly:

I even remember considering the race condition this test shows.

It really only shows `Multi.range` does not conform to the spec - it is able to deliver `onNext` during `onSubscribe`. The spec requires these signals to be delivered serially, and there is a TCK capturing this condition (which fails for several `Publishers`).

akarnokd:
103 tests for concurrent access violation which `range` does not do. Also [§2.1](https://github.com/reactive-streams/reactive-streams-jvm#2.1) and [§3.2](https://github.com/reactive-streams/reactive-streams-jvm#3.2) cover reentrance and [§3.3](https://github.com/reactive-streams/reactive-streams-jvm#3.3) covers recursion which `range` conforms as well.

Of course, you can ensure no request is delivered until the `subscriber.onSubscribe` invocation has returned via the [deferred subscription approach](https://github.com/oracle/helidon/blob/54ad2e65fbedc5925a7ae11285102adb82ac64d6/common/reactive/src/main/java/io/helidon/common/reactive/DeferredSubscription.java#L29-L33). However, that incurs extra allocation and full-barrier atomics thus I'd expect no clear performance benefit over the current implementation.

olotenko:
That is not the proof of the absence of the race in `range`. The proof is the existence of this failing condition, which can be shown to occur because `onNext` is delivered concurrently with `onSubscribe` by `range`.

We have a prototype that works around this issue with `range`, @danielkec is preparing a merge.

common/reactive/src/main/java/io/helidon/common/reactive/MultiConcatArray.java
```diff
@@ -70,38 +141,215 @@ public void onNext(T item) {
...
-                this.produced = 0L;
-                super.produced(produced);
+            Thread current = Thread.currentThread();
+            if (LASTTHREADCOMPLETING.getOpaque(this) == current) {
+                redo = true;
+                return;
+            }
+
+            LASTTHREADCOMPLETING.setOpaque(this, current);
+            VarHandle.storeStoreFence();
```
olotenko, line 157: Reading this again, strictly speaking this fence is not needed, because we have stronger fences between the stores targeted by this fence. But no need to change the code as the cost of this fence on the target platforms is zero.

olotenko (approved):
Appears to have only cosmetic changes compared to the internal proposal.

olotenko (approved):

spericas (approved):

PR #2511, opened 2020-11-09
Support for overriding configuration of security providers

### LINKED ISSUE #2455: Add possibility to override lists in config
Consider following two sources:
`main`:
```
security.providers:
  - abac:
 - oidc:
    client-id: "abcd" 
```

`test`:
```
security.providers:
 - abac:
 - http-basic-auth:
    realm: "helidon"
```

What I want to achieve:
 I want to be able to use HTTP Basic authentication for unit tests, but Open ID Connect provider for production.

What happens:
Security fails to start, as there is more than one security provider specified on second position

Why?
Because the merging happens on property name level. And as arrays are stored as indexed properties, we end up with:
```
security.providers.0.abac=
security.providers.1.oidc.client-id=abcd
security.providers.1.http-basic-auth.realm=helidon
```

When security iterates over the providers, it identifies that on position `1` we have two keys (`oidc` and `http-basic-auth`) and fails.

What should happen:
We need to find a way to nicely override such configuration - may require a different way to configure things

tomas-langer:
using simple properties.

Resolves #2455 

### DIFF
--- a/tests/integration/security/gh2455/src/test/java/io/helidon/tests/integration/security/gh2455/TestProviderOverrides.java
+++ b/tests/integration/security/gh2455/src/test/java/io/helidon/tests/integration/security/gh2455/TestProviderOverrides.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


tests/integration/security/gh2455/src/test/java/io/helidon/tests/integration/security/gh2455/TestProviderOverrides.java
```diff
@@ -0,0 +1,43 @@
...
+        );
+        Config config = Config.builder()
+                .addSource(ConfigSources.create(map))
+                .addSource(ConfigSources.classpath("application.yaml"))
+                .disableEnvironmentVariablesSource()
+                .disableSystemPropertiesSource()
+                .build();
+
+        Security security = Security.create(config.get("security"));
+    }
```
danielkec, line 42: Assert something?

tomas-langer: Asserting that `Security.create` does not throw an exception, which it did before the fix

danielkec: make sense thx

danielkec (approved):

Verdent (approved):

PR #2517, opened 2020-11-10
Address intermittent test failures

tjquinno:
The tests which exercise synthetic `@SimplyTimed` annotation handling when beans are vetoed show intermittent failures.

The problem turned out to be in the tests. The base registry (where, by spec, the synthetic `SimpleTimer` metrics go) could be populated with synthetic `SimpleTimer`s by earlier tests. Then, when the veto tests would run, and the `MetricsCdiExtension` code would _correctly_ ignore vetoed beans and not add the synthetic metrics to the registry again, but the registry entries were already present from earlier tests run in the same JVM. The tests, finding those entries, would report a failure.

The difference between the tests passing locally and failing in the pipeline was due to differences in the order in which surefire ran the tests.

I've added a `@BeforeAll` that clears out the `REST.request` `SimpleTimer`s to remove stale registrations, so the tests of the veto support find no synthetic `SimpleTimer`s from previous tests.


tjquinno:
The previous change resulted in clean builds. I've triggered them again manually to check again.

ljnelson (approved):

PR #2526, opened 2020-11-13
POC: test all packaging options

tomas-langer:


tomas-langer:
@romain-grecourt @Tomas-Kraus - I have added support for JLink images (both module path and classpath), now using a direct java call, as the `start` script does not honor current directory (https://github.com/oracle/helidon-build-tools/issues/334)

Issue #2415, opened 2020-10-06
mismatch versions of ojdbc artifacts


paulparkinson:
There are mismatch versions of ojdbc artifacts. See dependencies pom.xml...
       <version.lib.oracle.ojdbc10>19.3.0.0</version.lib.oracle.ojdbc10>
          [...]
        <version.lib.ojdbc8>19.6.0.0</version.lib.ojdbc8>

## Environment Details
* Helidon Version: 2.0.1
* Helidon SE and MP
* JDK version: any
* OS: any
* Docker version (if applicable):

----------

## Problem Description
All references to ojdbc (and ucp) should be com.oracle.database.jdbc ojdbc8 v 19.7 which is not currently the case, however, we may as well wait until November to make this change as v20 of these artifacts will be out then with support for native image in UCP and wallet which is the only major issue this mismatch causes.  We can pick up ojdbc11 and v20 at that time. 
We should also consider making production rather than "regular" the default artifacts as it has become the more common requirement/use case due to ATP database.

## Steps to reproduce
Attempt to build native image using ojdbc, ucp, and jpa

paulparkinson:
correction, will be v21 not v20

arjav-desai:
With above PR, we have removed ojdbc10 and ucp and now keeping only ojdc8 (v 19.8.0.0 and picking up all production version).

Keeping this issue open since we want to pick v21 which has native image support and will be available early december.

arjav-desai:
We have uptake 21.3 ojdbc jars and verified that native image works with UCP for accessing ATP.

PR #2530, opened 2020-11-16
Update ojdbc deps

arjav-desai:
Addresses #2415
We will have only one version of ojdbc (i.e. 8 and removing 10). Also, using "production" pom which bundles all required dep files (for now and future ATP integrations as well) rather than importing individual.


paulparkinson (approved):
As discussed we will need to update this soon and update/test the workshop accordingly so let's merge but keep the issue open until that time.

PR #2533, opened 2020-11-18
Examples of logging with Helidon

tomas-langer:
Also added support for native image to log4j, slf4j and JUL modules of the new logging implementation.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -116,7 +116,7 @@
         <version.lib.reactive-streams-tck>1.0.3</version.lib.reactive-streams-tck>
         <version.lib.reactivestreams>1.0.3</version.lib.reactivestreams>
         <version.lib.scala>2.12.10</version.lib.scala>
-        <version.lib.slf4j>1.7.26</version.lib.slf4j>
+        <version.lib.slf4j>1.7.30</version.lib.slf4j>
         <version.lib.smallrye-openapi>1.2.0</version.lib.smallrye-openapi>
         <version.lib.snakeyaml>1.27</version.lib.snakeyaml>
         <version.lib.transaction-api>1.3.3</version.lib.transaction-api>
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -766,6 +766,16 @@
                 <version>${version.lib.log4j}</version>
             </dependency>
             <dependency>
+                <groupId>org.apache.logging.log4j</groupId>
+                <artifactId>log4j-core</artifactId>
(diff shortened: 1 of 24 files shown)


dependencies/pom.xml
```diff
@@ -84,6 +84,7 @@
         <version.lib.kafka-junit5>3.2.1</version.lib.kafka-junit5>
         <version.lib.kafka>2.5.0</version.lib.kafka>
         <version.lib.log4j>2.13.3</version.lib.log4j>
+        <version.lib.logback>1.2.3</version.lib.logback>
```
barchetta, line 87: Can we bump SLF4J to 1.7.30? It's about time to do that.

barchetta (commented):
Looks good, but wondering if we could bump SLF4J to 1.7.30

examples/logging/jul/README.md
```diff
@@ -0,0 +1,7 @@
+JUL Example
+---
+
+This example shows how to use Java Util Logging with MDC
+ using Helidon API.
+ 
+The example can be built using GraalVM native image as well.
```
Verdent, line 7: I was thinking if you don't want to add exact commands how to build this example as we do have in our other examples. And possibly also command for native-image build since you are explicitly mentioning it.

Verdent (approved):
Looks good. I had only one idea there but I will leave that up to you.

examples/logging/log4j/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+            <artifactId>helidon-logging-log4j</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-core</artifactId>
+            <version>2.14.0</version>
```
barchetta, line 58: Shouldn't this match <version.lib.log4j>2.13.3</version.lib.log4j>? Maybe use that property here? Or dependency manage the version of these artifacts?

tomas-langer: yes, missed this one.

examples/logging/jul/README.md
tomas-langer: Fixed

examples/logging/log4j/pom.xml
tomas-langer: Fixed

Verdent (approved):

barchetta (approved):

PR #2549, opened 2020-11-25
Global context and its use in Security

### LINKED ISSUE #2548: Support outbound security in MP when outside of secured scope
Helidon 2.x

When a request to a REST client or JAX-RS client is invoked from a non-secured context (such as when using unsecured messaging), outbound security is ignored, even when explicitly configured (such as an explicit username/password for basic authentication outbound target).

Example outbound config that should work regardless of current subject:

```
security:
  providers:
  - http-basic-auth:
      users:
      - login: "john"
        password: "password"
        roles: ["admin"]
      outbound:
      - name: "internal"
        hosts: ["localhost"]
        username: "internal-user"
        password: "my-private-password"
```

In MP, we should use the security configured in CDI to do outbound for unsecured requests (or invocations that do not have a Helidon `Context` available).

tomas-langer:
Resolves #2548 

The `Contexts.globalContext()` can be used to register components that are application wide (e.g. `Singleton`) - such as `Security`, metrics `RegistryFactory` and `Tracer` in Helidon MP (in Helidon SE, this is at the discretion of developer, what get registered, if anything).

Updated JAX-RS Client and REST client support for security to use this global registry for cases when we invoke an outbound call without `SecurityContext` or `Context`. This is the case for example in messaging flows.

### DIFF
--- a/common/context/src/main/java/io/helidon/common/context/Context.java
+++ b/common/context/src/main/java/io/helidon/common/context/Context.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spericas (approved):
LGTM

PR #2559, opened 2020-12-02
Improve doc for built-in health check thresholds, etc.

### LINKED ISSUE #2558: Config values for health thresholds are described as fractions but interpreted as percents; plus no doc describing thresholds and how to override
## Environment Details
* Helidon Version: 1.x and 2.x
* Helidon SE or Helidon MP: both
* JDK version: n/a
* OS: n/a

----------

## Problem Description
Doc does not describe that config values which override built-in health checks are interpreted as percents. In fact, the JavaDoc for the built-in health checks _says_ to use fractions which contradicts what the code actually does.

## Steps to reproduce
Generate a quick-start project (MP in the original user's case but the problem is the same in both SE and MP). Edit the config file to add these:

```
helidon.health.heapMemory.thresholdPercent=.90
helidon.health.diskSpace.thresholdPercent=.90
```
which conforms to the JavaDoc. 

Then run `mvn package` and the unit test fails because the code interprets the values as percents and not as fractions.
Change the values to `90` and `mvn package` works.

tjquinno:
Resolves #2558 

Changes JavaDoc for the classes which provide the built-in health checks.

Also adds to the doc site documentation for SE health checks.

### DIFF
--- a/docs/se/health/01_health.adoc
+++ b/docs/se/health/01_health.adoc
@@ -181,25 +181,30 @@ You can use Helidon-provided health checks to report various
  common health check statuses:
 
 [[built-in-health-checks-table]]
-[cols="1,1,3,20"]
+[cols="1,1,3,15,3"]
 |=======
-|Built-in health check |Health check name |JavaDoc |Config properties [default value]
+|Built-in health check |Health check name |JavaDoc |Config properties |Default config value
 
 |deadlock detection
 |`deadlock`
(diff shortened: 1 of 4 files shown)


tomas-langer (dismissed):
LGTM

ljamen (approved):
changes from 1.x have been ported to 2.x. Other content looks good.

Issue #2341, opened 2020-09-08
Blocking APIs for SE


m0mus:
- [ ] Webserver #2563 
- [ ] DB Client #2796 
- [ ] Web Client #2698 
- [ ] Messaging #2795

tomas-langer:
This epic is obsolete. Loom should bring some new features!

PR #2563, opened 2020-12-02
SE blocking API for WebServer

tomas-langer:
Related to #2341 

This is a brand new feature to support blocking APIs that work nicely with project Loom.
Support for virtual threads is incorporated into the `ThreadPoolSupplier`, so it can be used in any part of Helidon simply by configuring `virtual-threads: true` for that executor.

This PR contains the API, implementation, and example. It is tested with native image.

This is first (experimental) approach, for discussion. Main features:

1. Blocking APIs do not introduce new routing abstraction
2. Main access point is `BlockingHandler` that allows you to create blocking routing methods and register them as usual with WebServer routing
3. All blocking calls are invoked in a custom executor service, that is by default using virtual threads when running on Loom enabled JVM
4. I have moved Loom support into `ThreadPoolSupplier`, so we can use it anywhere in Helidon (also refactored existing usage)

### DIFF
--- a/examples/webserver/blocking/src/main/java/io/helidon/webserver/examples/blocking/Main.java
+++ b/examples/webserver/blocking/src/main/java/io/helidon/webserver/examples/blocking/Main.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2021 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.webserver.examples.blocking;
+
+import java.net.http.HttpClient;
+
+import io.helidon.common.LogConfig;
+import io.helidon.common.configurable.ThreadPoolSupplier;
+import io.helidon.config.Config;
+import io.helidon.webclient.WebClient;
+import io.helidon.webserver.blocking.BlockingHandler;
+
+/**
+ * Main class of the example, starts the server.
+ */
+public final class Main {
+    private Main() {
+    }
+
+    /**
+     * Application main entry point.
+     * @param args command line arguments.
+     */
+    public static void main(final String[] args) {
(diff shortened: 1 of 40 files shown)


romain-grecourt (commented):
- How about a new type of routing, it would make the example a little more elegant so that one does to explicitly use BlockingHandler.

- How about adding this the webserver module directly?

- There is a some duplication of the interfaces (BlockingRequest, BlockingResponse), we could probably extract the common stuff to compose the interfaces.

tomas-langer:
I had a version with a dedicated routing type, but that could not be combined with the existing `Routing`. I like this approach, as I can chose for each route whether it is reactive or blocking.
Also the `BlockingHandler` was needed still, as there are too many possible method signatures - with reactive, it is just one `(ServerRequest, ServerResponse)`, with blocking it is:

1. `void (T entity)` for accepting an entity (PUT/POST)
2. `T (T entity)` for returning an entity with entity request
3. `T (BlockingRequest request)` for returning an entity and using request (e.g. to obtain headers, path params etc.)
4. `BlockingResponse (BlockingRequest)` for full control
5 `T ()` for a simple response

I may have missed some combinations here as well...

tomas-langer:
Adding to webserver - yes, that may be useful, as the API is quite small. If we would end up with custom routing, I would rather have a separate module

Duplication - yes, that can be reduced (even though I did an implementation that is quite complete, this is still very much a POC and a topic for discussion)

romain-grecourt:
Let's discuss this offline, I think having custom handle and routing are not exclusive.

examples/webserver/blocking/src/main/java/io/helidon/webserver/examples/blocking/Main.java
```diff
@@ -0,0 +1,152 @@
...
+                .config(config.get("server"))
+                .addMediaSupport(JsonpSupport.create())
+                .build();
+
+        server.start()
+                .thenAccept(ws -> {
+                    int backendPort;
+                    if (clientPort == -1) {
+                        backendPort = ws.port();
+                    } else {
```
danielkec, line 86: I would expect something like 
```java
        server.start();
        System.out.println("WEB server is up! http://localhost:" + server.port());
        server.whenShutdown().thenRun(() -> {
            System.out.println("WEB server is DOWN. Good bye!");
        });
```

danielkec: Maybe `server.startAndBlock()`

tomas-langer: You can use `server.start().await()` - I think blocking API is mostly needed for routing, not so much for starting/stopping the webserver.

tomas-langer:
Closing for now. Will revisit later

PR #2570, opened 2020-12-04
Updating JTA and JPA guides

ljnelson:


barchetta (commented):
Just ran through the JPA guide. Two minor things:
1. The log file (line 504) already exists. So you're adding the last 5 lines to an existing file.
2. The URL in the curl command on line 544 should be http://localhost:8080/greet/response/Marco

barchetta (approved):

PR #2591, opened 2020-12-09
Support for HTTP pipelining

spericas:
This patch ensures that responses are ordered correctly when HTTP pipelining is used. If multiple requests are received over the same connection, the corresponding responses need to be ordered according to the spec. It also affects pipelining over HTTP/2 connections. The crux of the fix is to wait for the _previous_ response to complete before writing data for the _current_ one. Thanks to @olotenko.

### DIFF
--- a/webserver/webserver/pom.xml
+++ b/webserver/webserver/pom.xml
@@ -164,7 +164,6 @@
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
-            <version>${version.lib.okhttp3}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/webserver/webserver/pom.xml
+++ b/webserver/webserver/pom.xml
@@ -161,5 +161,11 @@
             <artifactId>helidon-config-testing</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>${version.lib.okhttp3}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
@@ -78,19 +78,19 @@ class BareResponseImpl implements BareResponse {
     private volatile DefaultHttpResponse response;
     private volatile boolean lengthOptimization;
     private volatile boolean isWebSocketUpgrade = false;
-    private CompletableFuture<?> prev;
+    private CompletableFuture<?> previousRequestFuture;
 
     /**
      * @param ctx the channel handler context
      * @param request the request
      * @param requestContentConsumed whether the request content is consumed
-     * @param prev Future that represents previous request completion for HTTP pipelining
+     * @param previousRequestFuture Future that represents previous request completion for HTTP pipelining
      * @param requestId the correlation ID that is added to the log statements
      */
     BareResponseImpl(ChannelHandlerContext ctx,
                      HttpRequest request,
                      BooleanSupplier requestContentConsumed,
-                     CompletableFuture<?> prev,
+                     CompletableFuture<?> previousRequestFuture,
                      long requestId) {
         this.requestContentConsumed = requestContentConsumed;
         this.responseFuture = new CompletableFuture<>();
@@ -99,7 +99,7 @@ class BareResponseImpl implements BareResponse {
         this.requestId = requestId;
         this.keepAlive = HttpUtil.isKeepAlive(request);
         this.requestHeaders = request.headers();
-        this.prev = prev;
+        this.previousRequestFuture = previousRequestFuture;
 
         // We need to keep this listener so we can remove it when this response completes. If we don't, we leak
         // while the channel remains open since each response adds a new listener that references 'this'.
@@ -114,6 +114,12 @@ class BareResponseImpl implements BareResponse {
         responseFuture.whenComplete(this::responseComplete);
     }
 
+    /**
+     * Steps required for the completion of this response.
+     *
+     * @param self this instance
+     * @param throwable a throwable indicating unsuccessful completion
+     */
     private void responseComplete(BareResponse self, Throwable throwable) {
         if (throwable == null) {
             headersFuture.complete(this);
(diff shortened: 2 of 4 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -79,27 +78,28 @@
...
      * @param requestContentConsumed whether the request content is consumed
-     * @param thread the outbound event loop thread which will be used to write the response
+     * @param prev Future that represents previous request completion for HTTP pipelining
      * @param requestId the correlation ID that is added to the log statements
      */
     BareResponseImpl(ChannelHandlerContext ctx,
                      HttpRequest request,
                      BooleanSupplier requestContentConsumed,
-                     Thread thread,
+                     CompletableFuture<?> prev,
```
ljnelson, line 93: Readability nit: `prev` should probably be something like `previousRequestFuture` to "look like", e.g., `responseFuture`, `headersFuture`, etc.

ljnelson (dismissed):
Nicely done.  Only one suggested change from me.

webserver/webserver/src/test/java/io/helidon/webserver/HttpPipelineTest.java
```diff
@@ -0,0 +1,204 @@
...
+                            int n = counter.getAndIncrement();
+                            int delay = (n % 2 == 0) ? 1000 : 0;    // alternate delay 1 second and no delay
+                            executor.schedule(() -> res.status(Http.Status.OK_200).send("Response " + n + "\n"),
+                                    delay, TimeUnit.MILLISECONDS);
+                        })
+                        .build())
+                .build()
+                .start()
+                .toCompletableFuture()
+                .get(10, TimeUnit.SECONDS);
```
danielkec, line 118: `await(10, TimeUnit.SECONDS)`

webserver/webserver/src/test/java/io/helidon/webserver/HttpPipelineTest.java
```diff
@@ -0,0 +1,204 @@
...
+            }
+
+            @Override
+            public void onResponse(Call call, Response r0) throws IOException {
+                LOGGER.info("Received r0");
+                if (cf1.isDone()) {
+                    LOGGER.info("Expected r0 before r1");
+                    cf0.completeExceptionally(new RuntimeException("Expected r0 before r1"));
+                } else {
+                    assertThat(r0.body().string(), containsString("Response 0"));
```
danielkec, line 177: Will be the assertion error thrown on correct thread?

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
spericas: Good point

webserver/webserver/pom.xml
```diff
@@ -161,5 +161,11 @@
             <artifactId>helidon-config-testing</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>${version.lib.okhttp3}</version>
```
romain-grecourt, line 167: Why do you have to specify the version, isn't that managed by the helidon-project parent pom ?

webserver/webserver/src/test/java/io/helidon/webserver/HttpPipelineTest.java
spericas: Copy/paste, yes too long.

webserver/webserver/src/test/java/io/helidon/webserver/HttpPipelineTest.java
spericas: No, will fix this.

webserver/webserver/pom.xml
spericas: Will drop version.

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -78,19 +78,19 @@
     private volatile DefaultHttpResponse response;
     private volatile boolean lengthOptimization;
     private volatile boolean isWebSocketUpgrade = false;
-    private CompletableFuture<?> previousRequestFuture;
+    private CompletableFuture<?> prevRequestFuture;
```
ljnelson, line 81: I am unsure of the threading implications of who can modify this field, but I see it being modified in several places.  Does it need to be `volatile`?

spericas: Reading and writing this variable as a volatile as we do in `onNext` is odd and typically flagged as a problem (non-atomic update). Method `onNext` cannot be called in parallel, but it can be called sequentially in different threads. Perhaps this should be an `AtomicReference` instead.

ljnelson: …although `AtomicReference` is really just `volatile` memory semantics on steroids, right?  Up to you; just saw it being assigned in various places and wanted to double-check.

olotenko: `onNext`/`onComplete` from which these are invoked are totally ordered. The inter-thread semantics are ensured by the `Publisher` that calls these.

On the other hand, if you have a non-compliant `Publisher` that may issue `onNext` concurrently, the chaining logic is going to be broken even if `prev` is volatile.

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -78,19 +78,19 @@
     private volatile DefaultHttpResponse response;
     private volatile boolean lengthOptimization;
     private volatile boolean isWebSocketUpgrade = false;
-    private CompletableFuture<?> prev;
+    private CompletableFuture<?> previousRequestFuture;
```
olotenko: Since we are into renaming, I'd like to make sure the meaning is documented to be not only previous *request*, but really any chunk.

(N.B.: `onNext` may push several content chunks for the same request, so `prev` would reference the future for those chunks)

romain-grecourt (approved):

webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
```diff
@@ -144,7 +156,9 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) {
             // Queue, context and publisher creation
             ReferenceHoldingQueue<DataChunk> queue = new ReferenceHoldingQueue<>();
             queues.add(queue);
-            requestContext = new RequestContext(new HttpRequestScopedPublisher(ctx, queue), request);
+            RequestContext requestContext = new RequestContext(new HttpRequestScopedPublisher(ctx, queue), request);
+            this.requestContext = requestContext;
+
```
danielkec, line 161: Why declare local var?

danielkec: aha makes sense

danielkec (approved):

PR #2593, opened 2020-12-10
fix broken links and general doc updates

ljamen:
I had requests to fix broken links and other small doc issues. I removed a broken link from the config doc but I didn't update anything else as Tomas stated that the config doc will be changing.


ljamen:
Will create new PR to track updates to this PR

PR #2595, opened 2020-12-10
Exclude jakarta.ejb-api

barchetta:
While reviewing the dependencies in the MP quickstart I noticed jakarta.ejb-api (and jakarta.transaction-api). Those are not needed for the quickstart and are being dragged in by jakarta.interceptor-api 1.2.5 that has a dependency on jakarta.ejb-api. That is a bug. See https://github.com/eclipse-ee4j/interceptor-api/issues/31

So we exclude jakarta.ejb-api. Not sure why we didn't do this earlier.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -257,6 +257,13 @@
                 <groupId>jakarta.interceptor</groupId>
                 <artifactId>jakarta.interceptor-api</artifactId>
                 <version>${version.lib.interceptor-api}</version>
+                <exclusions>
+                  <!-- Exclude EJB. See https://github.com/eclipse-ee4j/interceptor-api/issues/31 -->
+                  <exclusion>
+                      <groupId>jakarta.ejb</groupId>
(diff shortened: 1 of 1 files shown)


ljnelson (approved):
Looks good.  Will CDI pull this in too?  If so you may need to add an exclusion there as well.

romain-grecourt (approved):

barchetta:
> Looks good. Will CDI pull this in too? If so you may need to add an exclusion there as well. (See [eclipse-ee4j/cdi#449](https://github.com/eclipse-ee4j/cdi/pull/449) for example.)

CDI 2.0.2 looks OK: https://repo1.maven.org/maven2/jakarta/enterprise/jakarta.enterprise.cdi-api/2.0.2/jakarta.enterprise.cdi-api-2.0.2.pom

I also verified the quickstart looks clean after excluding just from interceptor-api. But I do see something in CDI 3.x so we will want to keep an eye out for this.

PR #2604, opened 2020-12-14
Added support for @Singleton as a bean defining annotation.

tomas-langer:

By default beans with `@Singleton` are ignored by Weld. I have added this annotation as a bean defining annotation, so it is picked when discovery mode is `annotated` (it works now with `all` mode).

### DIFF
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
@@ -38,6 +38,7 @@ import javax.enterprise.inject.se.SeContainer;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.CDI;
 import javax.enterprise.inject.spi.Extension;
+import javax.inject.Singleton;
 
 import io.helidon.common.HelidonFeatures;
 import io.helidon.common.HelidonFlavor;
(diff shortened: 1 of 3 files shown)


ljnelson (commented):
My opinion: I vote against this since it reduces the portability of MP applications.  Also, Weld [has recently added the ability to add these things as properties](https://github.com/weld/core/pull/2167).

tomas-langer:
@ljnelson this was actually reported as a portability issue, migrating from an existing solution. Helidon did not discover singletons, where original implementation did...

ljnelson:
I can't comment on that.  As you know, [`@Singleton` is not a bean-defining annotation](https://jakarta.ee/specifications/cdi/3.0/jakarta-cdi-spec-3.0.html#bean_defining_annotations) so making it one would be surprising to anyone who is familiar with CDI bean discovery.  I know that CDI recently (and over the years) has mulled defining it to be a bean-defining annotation, but they haven't yet.  Hence I am still not in favor of baking this into Helidon.

tomas-langer:
OK. I will close this PR.
The correct way to handle this - there are two options - one seems to be as Laird mentioned to configure it using properties, once released and integrated into Helidon.

Second option is to create a stereotype, such as:
```java
@javax.enterprise.inject.Stereotype
@javax.inject.Singleton
@Target(TYPE)
@Retention(RUNTIME)
public @interface CdiSingleton {

}
```

PR #2605, opened 2020-12-14
2592 Chunk leak

### LINKED ISSUE #2592: Request chunks emitted after response is completed are not released
Request chunks emitted in the content publisher after the response is completed are not released, this triggers LEAK warning messages.

Any endpoint that returns a response instantly would reproduce this behavior.
Note that this is a regression introduced by https://github.com/oracle/helidon/pull/1830, and was originally fixed in https://github.com/oracle/helidon/pull/912

## Reproduce

Any endpoint that returns a response without consuming its payload.
The example below invokes `System.gc()` to expose the LEAK warning message.

```java
WebServer.builder(
        Routing.builder()
               .get((req, res) -> {
                   System.gc();
                   res.send("OK");
               })
               .build())
         .port(8080)
         .build()
         .start();
```

The chunks are emitted after routing, so this requires two invocation to trigger the LEAK warning message.

Invoke the endpoint twice:
```bash
curl -sSf http://localhost:8080/health -X GET -H "Content-Type: application/json" -d {}
curl -sSf http://localhost:8080/health -X GET -H "Content-Type: application/json" -d {}
```

Any HTTP method can be used, `GET` is only used as an example, but `PUT` or `POST` would work as well.

In the logs you should see:

```
WARNING: LEAK: RequestChunk.release() was not called before it was garbage collected. While the Reactive WebServer is designed to automatically release all the RequestChunks, it still comes with a considerable performance penalty and a demand for a large memory space (depending on expected throughput, it might require even more than 2GB). As such the users are strongly advised to release all the RequestChunk instances explicitly when they're not needed.
```

## Fix

The following changes fix the issue:

```java
diff --git a/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java b/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
index e72bf391a..6e967f745 100644
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
@@ -21,6 +21,7 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Queue;
+import java.util.concurrent.Flow;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Logger;
@@ -196,7 +197,25 @@ public class ForwardingHandler extends SimpleChannelInboundHandler<Object> {
                             if (queue.release()) {
                                 queues.remove(queue);
                             }
-                            publisherRef.clearBuffer(DataChunk::release);
+                            publisherRef.subscribe(new Flow.Subscriber<DataChunk>() {
+                                @Override
+                                public void onSubscribe(Flow.Subscription subscription) {
+                                    subscription.request(Long.MAX_VALUE);
+                                }
+
+                                @Override
+                                public void onNext(DataChunk item) {
+                                    item.release();
+                                }
+
+                                @Override
+                                public void onError(Throwable throwable) {
+                                }
+
+                                @Override
+                                public void onComplete() {
+                                }
+                            });
                             // Enable auto-read only after response has been completed
                             // to avoid a race condition with the next response
```

danielkec:
Fixes #2592

### DIFF
--- a/webserver/webserver/src/test/java/io/helidon/webserver/DataChunkReleaseTest.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/DataChunkReleaseTest.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright (c) 2020 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.webserver;
+
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+
+import io.helidon.common.http.Http;
+import io.helidon.media.common.DefaultMediaSupport;
+import io.helidon.webserver.utils.SocketHttpClient;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.endsWith;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.util.ResourceLeakDetector;
+import io.netty.util.internal.StringUtil;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class DataChunkReleaseTest {
+
+    static Logger dataChunkLogger = Logger.getLogger(ByteBufRequestChunk.class.getName());
+    static Logger leakDetectorLogger = Logger.getLogger(io.netty.util.ResourceLeakDetector.class.getName());
+
+    static volatile boolean leakIntercepted = false;
+
+    private static String originalLeakDetectionLevel;
+    private static String originalLeakDetectionSamplingInterval;
+    private static final Handler testHandler = new Handler() {
+        @Override
+        public void publish(final LogRecord record) {
+            if (record.getLevel() == Level.WARNING &&
+                    record.getMessage()
+                            .startsWith("LEAK: RequestChunk.release() was not called before it was garbage collected.")) {
+                leakIntercepted = true;
+            }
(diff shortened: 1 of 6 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/DataChunkReleaseTest.java
```diff
@@ -0,0 +1,171 @@
...
+        @Override
+        public void publish(final LogRecord record) {
+            if (record.getLevel() == Level.WARNING &&
+                    record.getMessage()
+                            .startsWith("LEAK: RequestChunk.release() was not called before it was garbage collected.")) {
+                leakIntercepted = true;
+            }
+            if (record.getLevel() == Level.SEVERE &&
+                    record.getMessage()
+                            .startsWith("LEAK: ByteBuf.release() was not called before it's garbage-collected.")) {
```
ljnelson, line 64: "`ByteBuf.release()` was not called before it was garbage collected"

ljnelson (changes requested):
One minor message change; otherwise LGTM

ljnelson (dismissed):
LGTM

webserver/webserver/src/test/java/io/helidon/webserver/DataChunkReleaseTest.java
romain-grecourt: This is an assertion for a log message emitted by Netty, see https://github.com/netty/netty/blob/44f85bba5f47df885dbbe5243d008220bfbab5ca/common/src/main/java/io/netty/util/ResourceLeakDetector.java#L320

ljnelson: Oh! I see; @danielkec I suppose you should revert the change then.

danielkec: Need for better commentary empirically verified 😄

danielkec: Added better description, the test is more complicated than I would like it to be, I just didn't found any nicer way to do it without heavy mocking.

romain-grecourt (approved):

webserver/webserver/src/test/java/io/helidon/webserver/DataChunkReleaseTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+    private static String originalLeakDetectionLevel;
+    private static String originalLeakDetectionSamplingInterval;
+    private static final Handler testHandler = new Handler() {
+        @Override
+        public void publish(final LogRecord record) {
+            // look for ByteBufRequestChunk's leak detection records
+            if (record.getLevel() == Level.WARNING &&
+                    record.getMessage()
+                            .startsWith("LEAK: RequestChunk.release() was not called before it was garbage collected.")) {
```
ljnelson, line 60: This one I assume needs to be the same as line 66?

danielkec: No its different check and it has its own message, They just happens to have very similar but not exactly same messages. I am tracking both to be extra safe

romain-grecourt:
@danielkec the new test seems to be failing intermittently, can you take a look ?
https://builds.helidon.io/api/88262DFBE59FE5DBF638CCD60E619737/tests/13/io.helidon.webserver.DataChunkReleaseTest/3/output

Issue #2647, opened 2021-01-11
ObjectMappingTest in microprofile/config can affect other unit tests' data


ljnelson:
This is mostly a tracking bug mainly so it doesn't get lost.

While investigating a user's report about environment variable overrides, I wrote a simple unit test in `microprofile/config`.  When this unit test is run and `ObjectMappingTest` is also run, my test's environment does not contain what it should.  When I exclude `ObjectMappingTest` from the test run but include all others, its environment does contain what it should.

My first uninformed guess is that something in registering `Config` instances in our `ConfigProviderResolver` is not fully resetting state.  But that is a guess.

The easy workaround to this issue is almost certainly going to be to set Surefire's `reuseForks` to `false`.

ljnelson:
Indeed, setting `reuseForks` to `false` allows my environment-variable-reading test to pass.  There is probably a larger issue buried in here.  Of note, editing `ObjectMappingTest` to explicitly release the `Config` it registers does not change the behavior.  Cleary a `static` reference of some kind somewhere is sticking around such that `ConfigProvider#getConfig(ClassLoader)` is still returning a `Config` instance put there by the `ObjectMappingTest`.

PR #2648, opened 2021-01-11
Proves that environment variable overrides work with our MicroProfile Config implementation

ljnelson:
A user is having trouble overriding YAML `camelCase` properties using Helidon MP.  This unit test proves that such overrides work properly.


ljnelson:
Please also see #2647.

romain-grecourt (dismissed):

romain-grecourt (approved):

PR #2656, opened 2021-01-13
Updated HealthCheck API and modified examples to use current HealthCheck API properly.

Tomas-Kraus:
API changes are just about adding @Deprecated and log a warning when deprecated code path is used.

### DIFF
--- a/examples/dbclient/jdbc/src/main/java/io/helidon/examples/dbclient/jdbc/JdbcExampleMain.java
+++ b/examples/dbclient/jdbc/src/main/java/io/helidon/examples/dbclient/jdbc/JdbcExampleMain.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


Tomas-Kraus:
This resolves Issue 2433

examples/dbclient/jdbc/src/main/java/io/helidon/examples/dbclient/jdbc/JdbcExampleMain.java
```diff
@@ -96,8 +96,13 @@ private static Routing createRouting(Config config) {
         DbClient dbClient = DbClient.builder(dbConfig)
                 .build();
 
+        // Some relational databases do not support DML statement as ping so using query which works for all of them
         HealthSupport health = HealthSupport.builder()
-                .addLiveness(DbClientHealthCheck.create(dbClient))
+                .addLiveness(
```
tomas-langer, line 101: Is the `query` required? It would be good to have a configuration based approach that does not require any additional code.
When I change between databases (and one supports Query and the other DML statement for health) I do not want to change code.
It would be good to have `DbClientHealthCheck.create(dbConfig)` or similar, to be able to define whether it is a `query`  or `dml` without coding.

Tomas-Kraus: Default is DML, but not all databases support DML statement as ping. Query works for everything including Mongo.
I can solve it by changing default to query.

Tomas-Kraus: Pushed modified DbClientHealthCheck API. Will modify this example to use setup from config.

Tomas-Kraus: Examples updated and tested that health check still works.

Tomas-Kraus:
Rebased on top of master after 2.2.1 release. Squashed all commits.

tomas-langer (approved):
LGTM.

PR #2686, opened 2021-01-24
Add fallbackPath to StaticContentSupport, fixes #2631

### LINKED ISSUE #2631: Improve support for serving SPAs with Client-Side Routing
## Environment Details
* Helidon Version: 2.2.0
* Helidon SE
* JDK version: openjdk 15.0.1 2020-10-20
* OS: debian

----------

## Problem Description

Single Page Applications (SPAs) often implement routing (e.g. clicks on links) by manipulating the URL path client side through browser APIs without sending requests to the server. The first request (or page refreshes) will go through to the server however, so a request to `/todos/12` should return the entry point into the application. See for example the [React docs](https://create-react-app.dev/docs/deployment/#serving-apps-with-client-side-routing) for this.

`StaticContentSupport` makes this a little awkward: The entry point for a React SPA is `index.html`. The Helidon app has to return `index.html` for every requested path, but those are not known in advance (e.g. invalid paths will be handled by the SPA), so duplicating `index.html` doesn't work. `StaticContentSupport` will next the request, leaving the developer to do the file handling, not benefitting from the caching and other optimizations in  `StaticContentSupport` / `ClassPathContentHandler`.

One straightforward solution would be to allow setting a fallback file, that gets sent when there was no matching file for the requested path:

```java
var staticContent = StaticContentSupport.builder("META-INF/frontend", ClassLoader.getSystemClassLoader())
    .welcomeFileName("index.html")
    // If the requested file wasn't found, use this one instead of nexting the request
    .fallbackFile("/index.html")
    .build();
```

Another solution would be to allow server side redirects like [`RequestDispatcher::forward`](https://javadoc.io/doc/jakarta.servlet/jakarta.servlet-api/latest/jakarta/servlet/RequestDispatcher.html#forward-jakarta.servlet.ServletRequest-jakarta.servlet.ServletResponse-), but that's overkill here, IMO.

zimmi:
This implements a variation of the feature proposed in #2631:

```java
var staticContent = StaticContentSupport.builder("META-INF/frontend", ClassLoader.getSystemClassLoader())
    .welcomeFileName("index.html")
    // If the requested file wasn't found, try this path instead of nexting the request.
    // Since the fallback path identifies a directory, the welcome file will be tried against it,
    // resulting in META-INF/frontend/index.html being served.
    .fallbackPath("/")
    .build();
```

I signed the OCA.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ClassPathContentHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ClassPathContentHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ClassPathContentHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ClassPathContentHandler.java
@@ -52,11 +52,12 @@ class ClassPathContentHandler extends StaticContentHandler {
     private final Path tempDir;
 
     ClassPathContentHandler(String welcomeFilename,
+                            String fallbackPath,
                             ContentTypeSelector contentTypeSelector,
                             String root,
                             Path tempDir,
                             ClassLoader classLoader) {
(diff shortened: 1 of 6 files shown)


tomas-langer:
I have updated the issue - please let me know if this would be really required, as I think the same functionality can be already achieved with WebServer routing features and without "nesting" of content locations in a single static content support

zimmi:
Yeah, turns out it's not required, see working solution here: https://github.com/oracle/helidon/issues/2631#issuecomment-771195941

Issue #2672, opened 2021-01-18
jackson-databind-2.10.0 Security Issue


tmb867:
## Environment Details
* Helidon Version: 1.4.7
* Helidon MP
* JDK version: 1.8
* OS: 
* Docker version (if applicable):

----------

Hi Team,

In the recent OWASP report for helidon-1.4.7, we see that jackson-databind-2.10.0 dependency of helidon has a high security issue. We confirmed the same by checking it in the following URL
https://nvd.nist.gov/vuln/detail/CVE-2020-25649.
Please let us know if there is any patch for 1.4.7 with the issue fixed or if there is a plan for upgrade.

Regards,
Thammaiah MB

barchetta:
We will upgrade this in Helidon 1.4.8. For 1.4 we should probably stay on 2.10, so the versions will be:
jackson-databind: 2.10.5.1
jackson-core and others: 2.10.5

@tmb867 As a work-around you can explicitly override Jackson in your project to the above versions.

barchetta:
Fixed. Upgraded jackson-databind in
1.4.8: 2.10.5.1
2.2.1: 2.12.1

tmb867:
Hi Team, 
I was trying to uptake 1.4.8 and I cannot find it in https://github.com/oracle/helidon/releases. Can you please let me know if this is released. If not can you please share the timeline for the same.
Regards,
Thammaiah MB

PR #2690, opened 2021-01-25
Update Jackson to 2.12.1

barchetta:
junit5 upgrade is needed to get fix for https://github.com/junit-team/junit5/issues/2198

Also address #2695


tomas-langer:
After upgrading Jackson, the following happened:
- in grpc-server, the transitive dependency of `zipkin-junit` changed Junit version from 4.12 to 4.13.1, which was not supported by the version of Junit Jupiter engine
- I have upgraded Junit 5 to latest version, which fixed this

I have no idea why a transitive dependency version would change when changing a version of Jackson, which is not part of dependencies of grpc-server, nor does it version manage junit. This may be interesting to investigate... @romain-grecourt any idea why this could happen?

romain-grecourt (approved):

barchetta:
Also see #2672

Issue #2599, opened 2020-12-13
JdbcStatement named parameters should be usable more than once in a statement


joev0000:
(or, it would be nice if named parameters in a statement didn't have to be unique)

I am trying to build a statement that uses the same value in multiple positions:

    SELECT ID FROM MY_TABLE WHERE (FOO = :foo) OR (FOO > :foo AND BAR = :bar)

When executed with a map containing `{ foo=1, bar=2 }`, this causes an exception

    Caused by: io.helidon.dbclient.DbClientException: Query parameters missing in Map:
        at io.helidon.dbclient.jdbc@2.1.0/io.helidon.dbclient.jdbc.JdbcStatement.prepareNamedStatement(JdbcStatement.java:121)

There is a check in that method that makes sure the number of entries in the parameter map is the same as the number of parameter names as they appear in the query:

    // SQL statement and provided parameters integrity check
    if (namesOrder.size() > parameters.size()) {
        throw new DbClientException(namedStatementErrorMessage(namesOrder, parameters));
    }

The debugger here tells me that the `namesOrder` contains `["foo", "foo", "bar"]` but `parameters` contains `{foo=1, bar=2}`.  and so the exception is raised.  

## Environment Details
* Helidon Version: 2.1.0
* Helidon SE
* JDK version: 11.0.8
* OS: Oracle Linux 7.8
* Docker version (if applicable): N/A

----------

### Expected Behavior
JdbcStatement allows the use of a named parameter more than once in a statement.
### Current Behavior
The exception is raised: `DbClientException: Query parameters missing in Map`

    Caused by: io.helidon.dbclient.DbClientException: Query parameters missing in Map:
            at io.helidon.dbclient.jdbc@2.1.0/io.helidon.dbclient.jdbc.JdbcStatement.prepareNamedStatement(JdbcStatement.java:121)
            at io.helidon.dbclient.jdbc@2.1.0/io.helidon.dbclient.jdbc.JdbcStatement.lambda$build$3(JdbcStatement.java:76)
            at java.base/java.util.Optional.map(Optional.java:265)
            at io.helidon.dbclient.jdbc@2.1.0/io.helidon.dbclient.jdbc.JdbcStatement.build(JdbcStatement.java:76)
            at io.helidon.dbclient.jdbc@2.1.0/io.helidon.dbclient.jdbc.JdbcStatementQuery.lambda$doExecute$2(JdbcStatementQuery.java:95)
            ... 7 more

## Steps to reproduce
[...]

PR #2693, opened 2021-01-26
Removed query parameter check that is too strict.

Tomas-Kraus:
Fixes issue #2599


tomas-langer (changes requested):
I think this deserves a test that uses more than one named parameter, to make sure this issue is fixed.

Tomas-Kraus:
Looks like we have ci/pr-head — test failures. I'll rebase it on the top of the master.

tomas-langer (approved):

Issue #2697, opened 2021-01-27
Db examples should use reactive approach (including template for CLI)


tomas-langer:
Current code:
```
try {
            List<PokemonType> pokemonTypes =
                    dbClient.execute(exec -> exec.namedQuery("select-all-types"))
                            .map(row -> row.as(PokemonType.class)).collectList().get();
            response.send(pokemonTypes);
        } catch (InterruptedException | ExecutionException e) {
            throw new RuntimeException(e);
        }
```

Correct code:
```
dbClient.execute(exec -> exec.namedQuery("select-all-types"))
                .map(row -> row.as(PokemonType.class))
                .collectList()
                .forSingle(response::send);
```

This is valid for other methods as well - we should never ever call `.get` on a completable future or `await` on `Single` or `Multi` in a routing method of web server

PR #2701, opened 2021-01-28
PokemonService template fixed in SE Database Archetype.

Tomas-Kraus:
Fixes Issue #2697.

### DIFF
--- a/archetypes/database-se/src/main/resources/src/main/java/__pkg__/PokemonService.java.mustache
+++ b/archetypes/database-se/src/main/resources/src/main/java/__pkg__/PokemonService.java.mustache
@@ -52,25 +52,17 @@ public class PokemonService implements Service {
     }
 
     private void listTypes(ServerRequest request, ServerResponse response) {
-        try {
-            List<PokemonType> pokemonTypes =
-                    dbClient.execute(exec -> exec.namedQuery("select-all-types"))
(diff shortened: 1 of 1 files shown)


archetypes/database-se/src/main/resources/src/main/java/__pkg__/PokemonService.java.mustache
```diff
@@ -52,25 +52,17 @@ public class PokemonService implements Service {
...
     private void listTypes(ServerRequest request, ServerResponse response) {
-        try {
-            List<PokemonType> pokemonTypes =
-                    dbClient.execute(exec -> exec.namedQuery("select-all-types"))
-                            .map(row -> row.as(PokemonType.class)).collectList().get();
-            response.send(pokemonTypes);
-        } catch (InterruptedException | ExecutionException e) {
-            throw new RuntimeException(e);
-        }
+        dbClient.execute(exec -> exec.namedQuery("select-all-types"))
```
tomas-langer, line 55: This approach will collect all records into memory.
We can use:
```
response.send(dbClient.execute(exec -> exec.namedQuery("select-all-types"))
                .map(row -> row.as(PokemonType.class)), PokemonType.class);
```
To use JSON-Binding streaming support in Helidon to send records using publisher directly.

tomas-langer (approved):

PR #2704, opened 2021-01-29
Performance improvements to queue(s) management in Webserver

spericas:
The Webserver keeps track of a queue of queues to release buffers back to Netty. Every new request that comes in creates a queue to track the Netty buffers. The cost of removing one of these queues when no longer needed is O(N) where N is the number of active connections. When N is large (e.g. 16K) this housekeeping operation can be costly. 

This PR uses a different approach that avoids the O(N) removal operation. It uses phantom references and lets the GC gather the queues when the associated publisher becomes ready for collection. A `clearQueues` method is called periodically to clean up the queue of queues. Thanks to @olotenko.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,13 +22,15 @@ import java.security.cert.X509Certificate;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.logging.Logger;
+import java.lang.ref.Reference;
+import java.lang.ref.ReferenceQueue;
 
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLPeerUnverifiedException;
 
-import io.helidon.common.http.DataChunk;
 import io.helidon.webserver.HelidonConnectionHandler.HelidonHttp2ConnectionHandlerBuilder;
 
+import io.helidon.webserver.ReferenceHoldingQueue.IndirectReference;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
@@ -60,7 +62,8 @@ class HttpInitializer extends ChannelInitializer<SocketChannel> {
     private final NettyWebServer webServer;
     private final SocketConfiguration soConfig;
     private final Routing routing;
-    private final Queue<ReferenceHoldingQueue<DataChunk>> queues = new ConcurrentLinkedQueue<>();
+    private final ReferenceQueue<Object> refQ = new ReferenceQueue<>();
+    private final Queue<ReferenceHoldingQueue<?>> queues = new ConcurrentLinkedQueue<>();
 
     HttpInitializer(SocketConfiguration soConfig,
                     SslContext sslContext,
@@ -73,10 +76,23 @@ class HttpInitializer extends ChannelInitializer<SocketChannel> {
     }
 
     private void clearQueues() {
+        for(Reference<?> r = refQ.poll(); r != null; r = refQ.poll()) {
+           if (!(r instanceof IndirectReference<?>)) {
+              continue;
+           }
+           ReferenceHoldingQueue<?> q = ((IndirectReference<ReferenceHoldingQueue<?>>) r).acquire();
+           if (q == null) {
+              continue;
+           }
+           if (!q.release()) {
+              queues.add(q);
+           }
+        }
         queues.removeIf(ReferenceHoldingQueue::release);
     }
 
     void queuesShutdown() {
+        clearQueues();
         queues.removeIf(queue -> {
             queue.shutdown();
             return true;
@@ -132,7 +148,7 @@ class HttpInitializer extends ChannelInitializer<SocketChannel> {
         }
 
         // Helidon's forwarding handler
-        p.addLast(new ForwardingHandler(routing, webServer, sslEngine, queues,
+        p.addLast(new ForwardingHandler(routing, webServer, sslEngine, refQ,
                                         requestDecoder, soConfig.maxPayloadSize()));
 
         // Cleanup queues as part of event loop
--- a/webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
@@ -16,20 +16,18 @@
 
 package io.helidon.webserver;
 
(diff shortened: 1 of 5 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/ReferenceHoldingQueue.java
```diff
@@ -122,6 +119,40 @@ private void unlink(ReleasableReference<T> reference) {
...
+     *
+     * @param <T> type of referent object
+     * @param <R> type of the other object
+     */
+    static class IndirectReference<T, R> extends PhantomReference<T> {
+        private final AtomicReference<R> otherRef = new AtomicReference<>();
+
+        IndirectReference(T referent, ReferenceQueue<? super T> q, R otherRef) {
+            super(referent, q);
+            this.otherRef.lazySet(otherRef);
```
ljnelson, line 138: Why `lazySet`?

olotenko: then it is a normal `store` (no expensive full barrier)

(think `mov mm, r` vs `lock: mov mm, r`)

olotenko: Actually, I am not sure that the atomicity of the reference is even required here.

All accesses are meant to be single-threaded in the intended use. It is atomic only for the purpose of potential other uses.

olotenko (dismissed):
I understand all apart from the removal of `ChannelHandlerContext ctx` from `HttpRequestScopedPublisher`.

Also, I would do the handling of `failPublisher` for 4xx errors as a separate commit, but separating that is not essential.

As far as the reference queue handling is concerned, it is good to ship.

spericas:
> I understand all apart from the removal of `ChannelHandlerContext ctx` from `HttpRequestScopedPublisher`.

 That context is no longer used in the publisher.

> 
> Also, I would do the handling of `failPublisher` for 4xx errors as a separate commit, but separating that is not essential.
> 
> As far as the reference queue handling is concerned, it is good to ship.

 Cool, thx.

webserver/webserver/src/main/java/io/helidon/webserver/ReferenceHoldingQueue.java
spericas: Maybe not needed, but seems like the safer option.

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
```diff
@@ -60,7 +63,21 @@
...
-    private final Queue<ReferenceHoldingQueue<DataChunk>> queues = new ConcurrentLinkedQueue<>();
+    private final AtomicBoolean clearLock = new AtomicBoolean();
+
+    /**
+     * Reference queue that collects ReferenceHoldingQueue's when they become
+     * ready for garbage collection. ReferenceHoldingQueue's extracted from
+     * this collection that cannot be fully released (some buffers still in
+     * use) will be added to {@code unreleasedQueues} for later retries.
+     */
+    private final ReferenceQueue<Object> queues = new ReferenceQueue<>();
```
ljnelson, line 74: Just being careful: is `Object` the type argument you want to pass here?

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
```diff
@@ -72,17 +89,53 @@
...
+                if (q == null) {
+                    continue;       // no longer referenced
+                }
+                if (!q.release()) {
+                    unreleasedQueues.add(q);
+                }
+            }
+            unreleasedQueues.removeIf(ReferenceHoldingQueue::release);
+        } finally {
+            clearLock.lazySet(false);
```
ljnelson, line 119: Just being careful: `lazySet` is designed for very specialized use cases.  Are you sure `set` isn't the right choice here?

ljnelson (approved):

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
olotenko: Correct. That type-parameter is silly. (Reads: I don't get it) It is meant to represent the type of element returned by `References` enqueued in `ReferenceQueue`. But by the time a reference is enqueued the element is no longer reachable, as determined by GC. So...

What would make more sense as a type parameter for the `ReferenceQueue`, is the class of Reference that is going to be returned by `poll`.

ljnelson: Well, I don't know about silly.  Would `? super Reference<?>` or something like that work better here?

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
olotenko: Yes, this is one case where it makes sense.

We only have a `lock` that can either be taken immediately, or the contender goes away:
https://github.com/oracle/helidon/pull/2704/files/06c459f90972ba6a14a500bfaea7cd4b91beb3cb#diff-a0cac19f24ae85fbc99afd8c0b4e8375d3c036a0ad949c68befe56fb7346461dR100 - that yellow line

In this case we only need to ensure the `get()` and `compareAndSet` on that line synchronize-with this `lazySet`, which they do.

We don't need to "publish" any other changes, as queue poll and removeIf are atomic in their own right.

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
olotenko: No.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/ReferenceQueue.html#poll() - `ReferenceQueue<T>.poll()` returns `Reference<? extends T>`. I am not interested in what `T` is, because that value is no longer accessible. I am only interested in the type of `Reference`. If it is something that has a method to release whatever resource it is associated with (ie `instanceof IndirectReference` in this case), I am going to call it - end of story.

I'd rather have a guarantee that I will not encounter other type of `Reference` here, so that I didn't have to type-cast. :)

olotenko: Think of it this way: `IndirectReference` is about two things: some owner of type `T` that actually does not matter to a resource `R` that is temporarily handed off to the owner of type `T`. The owner has the obligation to always terminate in a predictable path where it hands off the resource `R` back to some resource pool.

`IndirectReference` then is a safety net catching the cases when the owner `T` fails to fulfil the obligation, and dies before returning resource `R`. In this case `IndirectReference` ends up in the `ReferenceQueue`, and the queue processing routine is able to return resource `R` back to the pool. Note only the owner of type `T` is referenced phantomly. The resource `R` remains strongly reachable, at least through `IndirectReference` - unless, of course, the owner does fulfil the obligation, and returns the resource, first reclaiming it through `IndirectReference.acquire()`.

olotenko: Having explained all this, I think more work is needed.

The focus has been to reduce the cost of maintaining resources for well-terminating responses. But need to consider what happens to strong references when the channel is closed (`channelInactive` fired) and there are unfinished responses.

PR #2705, opened 2021-01-29
Static content support module.

### LINKED ISSUE #1236: Refactor media type selector in WebServer.
`io.helidon.webserver.ContentTypeSelector` maps file extensions to `MediaType` and can be useful for application developers.

We should consider making it public for 2.0, and perhaps move it to `common/http`.

### LINKED ISSUE #1869: WebServer: static content module
Follow up from #1217 

- Move the static content support to a separate module and Implement directory listing (maps to a feature as well)

### LINKED ISSUE #2631: Improve support for serving SPAs with Client-Side Routing

### LINKED ISSUE #2764: Query parameters lost on redirect in static content.
When redirecting because a directory was requested without trailing `/` we lose query parameters.

tomas-langer:
Resolves #1869 
Resolves #1236 
Resolves #2631 
Resolves #2764

There is a new module: `helidon-webserver-static-content`.
Entry point is `io.helidon.webserver.staticcontent.StaticContentSupport.builder(..)`

To achieve a fallback for static content, you can use the following when setting up routing:
```
Routing.builder()
  .register("/static", StaticContentSupport.builder("web")
    .welcomeFileName("index.txt")
    .build())
  .register("/static", StaticContentSupport.builder("fallback")
    .pathMapper(path -> "index.txt")
    .build())
```

This will resolve any request to `/static` as follows:

1. Attempt to find a resource on the classpath under `web`, using `index.txt` as the welcome file for directory requests
2. Use `fallback/index.txt` for any request that does not find a resource in the first step

The default `pathMapper` does not modify the path (so for `static/css/company.css` it returns `static/css/company.css`).

There is a known duplicate use of temporary directory - each instance of static content support based on classpath would extract the files separately. If we used the same classpath location to read the "fallback" file, it would be extracted and cached twice.
As this use-case is based on the need to have a single backup file, the impact is minimal (extracting one extra file). If this is a major performance issue for your application, please submit a new issue with your use case.


romain-grecourt:
Shouldn't this PR be targeted for 3.x ? It breaks compatibility right ?

romain-grecourt (commented):
Shouldn't this PR be targeted for 3.x ? It breaks compatibility right ?

tomas-langer:
I have only created a new module and deprecated the old classes. Everything still works as it used to.

tomas-langer:
Will need to incorporate changes from #2686 if we decide to merge them

tomas-langer:
Check #2631 

We should support a fallback to a single file from static content.

Request to `/static/webserver/index.html` - which does not exist, should return `/static/index.html`.

This would allow for routing like this:
```
Routing.builder()
  .register("/static", StaticContentSupport.create("/static")
 .register("/static", SingleFileContent.create("/index.html")
```

tomas-langer:
Refactored implementation to have nicer builders, added integration tests to validate implementation.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/FileBasedContentHandler.java
```diff
@@ -0,0 +1,181 @@
...
+        // we know the file exists, though it may be a directory
+        //First doHandle a directory case
+        if (Files.isDirectory(path)) {
+            String rawFullPath = request.uri().getRawPath();
+            if (rawFullPath.endsWith("/")) {
+                // Try to found welcome file
+                path = resolveWelcomeFile(path, welcomePage);
+            } else {
+                // Or redirect to slash ended
+                redirect(response, rawFullPath + "/");
```
zimmi, line 149: This drops query params on redirect (same for `ClassPathContentHandler`). A request to `/fallback/existing?a=b` should redirect to `/fallback/existing/?a=b`. It would further be nice to make it configurable if the canonical URL that is redirected to has a trailing slash or not.

BTW, thanks for doing this @tomas-langer ! :+1:

tests/integration/webserver/gh2631/src/main/java/io/helidon/tests/integration/webserver/gh2631/Gh2631.java
```diff
@@ -0,0 +1,60 @@
...
+                .welcomeFileName("index.txt")
+                .build();
+        StaticContentSupport file = StaticContentSupport.builder(Paths.get("src/main/resources/web"))
+                .welcomeFileName("index.txt")
+                .build();
+
+        return Routing.builder()
+                .register("/simple", classpath)
+                .register("/fallback", classpath)
+                .register("/fallback", StaticContentSupport.builder("fallback")
```
zimmi, line 50: Maybe a minor thing, but `fallback/index.html` will be extracted twice to the tmp directory when using two separate instances.

tomas-langer: I do not think this is the case. There is one based on classpath, one based on path.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/FileBasedContentHandler.java
romain-grecourt: This code is copied over, and should be fixed a separate issue / PR.

tests/integration/webserver/gh2631/src/main/java/io/helidon/tests/integration/webserver/gh2631/Gh2631.java
romain-grecourt: If both `StaticContentSupport` were using the classpath then yes the file would be extracted twice ; this is a side-effect of the current implementation of the classpath based `StaticContentSupport`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/FileBasedContentHandler.java
tomas-langer: I have fixed it for this PR, so we have this working ;)

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -230,12 +230,21 @@ static void throwNotFoundIf(boolean condition) {
...
      * @param location a location to redirect
      */
-    static void redirect(ServerResponse response, String location) {
+    static void redirect(ServerRequest request, ServerResponse response, String location) {
+        String query = request.query();
+        String locationWithQuery;
+        if (query == null) {
+            locationWithQuery = location;
+        } else {
+            locationWithQuery = location + "?" + request.query();
```
zimmi, line 243: Coulde use `locationWithQuery = location + "?" + query;`.

romain-grecourt (approved):

PR #2708, opened 2021-02-01
Fix TODO application:

### LINKED ISSUE #2707: TODO application does not work
The TODO example application is not working (examples/todo-app), here is a description of the issues:

- [ ] The authentication configuration of the frontend is not used
- [ ] The regular expression for the paths to be secured is  `/api/[{*}]`, this is not recursive so the DELETE calls are failing
- [ ] In the past, all headers were propagated but it now requires explicit code or configuration, thus the `Authorization` header is not passed from the frontend calls to the backend.
- [ ] The current version of the Cassandra driver uses guava and the version it needs conflict with the version of guava used by the google login lib
- [ ] The K8S YAML files are using old version of Kubernetes APIs and do not work with the latest version available with docker for desktop.

romain-grecourt:
- WebSecurity needs to be passed config.get("security") to take the "security.web-server" configuration
 - Added outbound configuration for the google login
 - Upgraded cassandra driver to fix issues with old guava dependencies
 - Removed metrics to avoid issues with cassandra driver.

Fixes #2707

### DIFF
--- a/examples/todo-app/backend/pom.xml
+++ b/examples/todo-app/backend/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2018, 2020 Oracle and/or its affiliates. All rights reserved.
+    Copyright (c) 2018, 2021 Oracle and/or its affiliates. All rights reserved.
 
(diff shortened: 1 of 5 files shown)


tomas-langer (approved):
LGTM

PR #2714, opened 2021-02-01
Fixed interfering tests in MP config.

### LINKED ISSUE #2647: ObjectMappingTest in microprofile/config can affect other unit tests' data

tomas-langer:
Resolves #2647 

 - The method `releaseConfig` did not correctly take into account usage of `ConfigDelegate`.
 - reset the reuseForks back to default
 - META-INF/services for YAML config source is not needed, it is part of the library
 - The `ObjectMappingTest` did not correctly handle release of configuration.
 - Removed extensive `final` declarations in test


ljnelson (approved):

config/config-mp/src/main/java/io/helidon/config/mp/MpConfigProviderResolver.java
```diff
@@ -173,13 +173,23 @@ public void releaseConfig(Config config) {
...
+                        break;
+                    }
+                }
+            } else {
+                for (Map.Entry<ClassLoader, ConfigDelegate> entry : CONFIGS.entrySet()) {
+                    Config configFromRef = entry.getValue().delegate();
+                    if (config == configFromRef) {
+                        cl.set(entry.getKey());
+                        break;
+                    }
```
danielkec, line 192: Maybe something like this could avoid code duplication
```java
Predicate<Map.Entry<ClassLoader, ConfigDelegate>> entryPredicate;

if(config instanceof ConfigDelegate){
    entryPredicate = entry -> config == entry.getValue();
}else {
    entryPredicate = entry -> config == entry.getValue().delegate();
}

CONFIGS.entrySet()
        .stream()
        .filter(entryPredicate)
        .findFirst()
        .map(Map.Entry::getKey)
        .ifPresent(cl::set);
```

danielkec: On the other hand it's not shorter nor more readable and definitely not faster

danielkec (approved):

PR #2715, opened 2021-02-01
Properly release underlying buffer before passing it to WebSocket handler

spericas:
Underlying buffers are reference counted and need to be manually released back to the pool

### DIFF
--- a/webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusReaderSubscriber.java
+++ b/webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusReaderSubscriber.java
@@ -63,33 +63,55 @@ public class TyrusReaderSubscriber implements Flow.Subscriber<DataChunk> {
 
     @Override
     public void onNext(DataChunk item) {
-        // Copy underlying buffer into a ByteBuffer and release DataChunk
-        ByteBuffer byteBuffer = ByteBuffer.wrap(item.bytes());
-        item.release();
-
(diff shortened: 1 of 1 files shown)


webserver/tyrus/src/main/java/io/helidon/webserver/tyrus/TyrusReaderSubscriber.java
```diff
@@ -63,16 +63,15 @@ public void onSubscribe(Flow.Subscription subscription) {
 
     @Override
     public void onNext(DataChunk item) {
+        // Copy underlying buffer into a ByteBuffer and release DataChunk
+        ByteBuffer byteBuffer = ByteBuffer.wrap(item.bytes());
```
romain-grecourt, line 67: Isn't there a way to know that the WebSocket handler is done and that the corresponding DataChunk can be released ?
Copying the data may have a performance cost right ?

spericas: Yes, there is cost. Actually, I think there's a better way to do this ... Will try.

spericas: Latest version avoids copying and is more draconian. I think there was a problem with subscriptions that was likely introduced when we switched to multiple arrays in a `DataChunk` --which, of course, would not be easily testable under normal conditions.

romain-grecourt (approved):

PR #2717, opened 2021-02-01
Support for custom default config file suffixes

### LINKED ISSUE #2171: Default configuration files
Currently the default configuration files are hardcoded (`application.yaml`, `application.properties`)

These should be defined by appropriate `ConfigParser` or `ConfigSource` implementation, so we only look for the sources currently supported (and also we may add new default configuration files without changing Helidon config code).

tomas-langer:
Resolves #2171 

Config now allows `ConfigParser`s to configure a new suffix to use when searching for default config source.
YAML parser now added `yml` suffix.
The search is ordered - so if there is `application.yaml` and `application.yml`, we still use `application.yaml`.

### DIFF
--- a/config/config/src/main/java/io/helidon/config/BuilderImpl.java
+++ b/config/config/src/main/java/io/helidon/config/BuilderImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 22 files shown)


tjquinno:
Because the order behavior of the suffixes is documented, maybe add a test verifying that ordering of two sources works as expected? Otherwise LGTM.

tjquinno (dismissed):
Ideally, add a test to verify that the documented order (.yaml, then .yml) is followed. But I guess that's not a blocker. LGTM

tjquinno (approved):

PR #2733, opened 2021-02-04
gRPC native image support in SE

### LINKED ISSUE #1294: gRPC support in native-image
Check gRPC support in native image both for MP and SE and (possibly) create follow up issues to add support for it.

tomas-langer:
Partially resolves #1294 - for SE

Support for gRPC for GraalVM native image in Helidon SE.

Supports both server and client, including standalone client.
Tested with GraalVM CE 21.0.0


common/common/src/main/java/io/helidon/common/FeatureCatalog.java
```diff
@@ -152,6 +154,10 @@
                     .path("WebServer", "Websocket")
                     .nativeSupported(true)
                     .nativeDescription("Server only"));
+        addSe("io.helidon.faulttolerance",
+              "Fault Tolerance",
+              "Fault tolerance support",
+              "FT");
```
danielkec, line 160: intentional?

danielkec: If so its missing in the [docs/se/aot/01_introduction.adoc](https://github.com/tomas-langer/helidon/blob/1294-grpc-native/docs/se/aot/01_introduction.adoc)

tomas-langer: Removed

danielkec (approved):

PR #2747, opened 2021-02-08
Cloud Functions (Google, Azure, OCI and AWS)

jbescos:
This is a copy of https://github.com/oracle/helidon/pull/2365 but changing the dates to 2021

The maven copyright plugins is not working properly because my changes belong to 2020. After some rebases some files were updated to 2021, but my commit is on top of them with the date 2020. That made the plugin to fail.

### DIFF
--- a/microprofile/cloud/common/src/main/java/io/helidon/microprofile/cloud/common/CommonCloudFunction.java
+++ b/microprofile/cloud/common/src/main/java/io/helidon/microprofile/cloud/common/CommonCloudFunction.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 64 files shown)


microprofile/cloud/common/src/main/java/io/helidon/microprofile/cloud/common/CommonCloudFunction.java
```diff
@@ -0,0 +1,66 @@
...
+    @SuppressWarnings("unchecked")
+    protected T delegate() {
+        return (T) LazyHelidonInitializer.DELEGATE;
+    }
+
+    /**
+     * Avoids Helidon is initialized in case CommonCloudFunction is never used.
+     *
+     */
+    private static class LazyHelidonInitializer {
```
danielkec, line 47: Mabe [io.helidon.common.LazyValue](https://github.com/oracle/helidon/blob/785ce38ca06b268d16e387fb6498aaaa890695cc/common/common/src/main/java/io/helidon/common/LazyValue.java) could help here

jbescos: I think current one performs a little bit better.

LazyValueImpl has some MethodHandles.lookup().findVarHandle and some lines of code related to get the value in a thread safety manner.

This 'MethodHandles.lookup().findVarHandle' is going to be executed every time there is a cold start. In lambdas it will happen quite often.

I think LazyValueImpl is better for some java processes that are running for longer time and for general usage.

romain-grecourt:
Closing for inactivity.

PR #2748, opened 2021-02-08
Do not attempt to access the request context in Fallback callback

spericas:
Do not attempt to access the request context in Fallback callback. If used together with Retry, it is possible for the fallback to be called in a fresh thread for which there is no current request scope. Instead just use the original value obtained in this class' constructor. Updated functional test (with some class renaming) to cover this use case.


tjquinno (approved):
LGMT after a nice Zoom walkthrough.

Issue #2211, opened 2020-07-23
JAX-RS ContainerRequestContext contract is not adequately fulfilled


pbull-oracle:
## Environment Details
* Helidon Version: 2.0.0-M3
* Helidon SE or Helidon MP: MP
* JDK version: jdk-11u6-linux-x64
* OS: Linux RHEL 7.4
* Docker version (if applicable): N/A

----------

## Problem Description
In our Helidon MP app, `javax.ws.rs.container.ContainerRequestContext` throws unexpected exceptions for removing properties.

**Expected behavior is clearly outlined in Javadocs, with no mention of deprecation or optional support:**
https://github.com/eclipse-ee4j/jaxrs-api/blob/2.1.6/jaxrs-api/src/main/java/javax/ws/rs/container/ContainerRequestContext.java

1. `ContainerRequestContext.removeProperty(name)`
Actual behavior: 
```
java.lang.UnsupportedOperationException: property removal from the context is not allowed
    at io.helidon.webserver.jersey.JerseySupport$WebServerPropertiesDelegate.removeProperty(JerseySupport.java:353)
    at org.glassfish.jersey.message.internal.TracingAwarePropertiesDelegate.removeProperty(TracingAwarePropertiesDelegate.java:50)
    at org.glassfish.jersey.server.ContainerRequest.removeProperty(ContainerRequest.java:294)
```

2. `ContainerRequestContext.setProperty(name, null)`
Actual behavior:
```
java.lang.NullPointerException: Parameter 'instance' is null!
    at java.util.Objects.requireNonNull(Objects:246)
    at io.helidon.common.context.ListContext.ClassifiedRegistry.register(ListContext:146)
    at io.helidon.common.context.ListContext.register(ListContext:81)
    at io.helidon.common.http.ListContextualRegistry.register(ListContextualRegistry:73)
    at io.helidon.webserver.jersey.JerseySupport$WebServerPropertiesDelegate.setProperty(JerseySupport.java:347)
    at org.glassfish.jersey.message.internal.TracingAwarePropertiesDelegate.setProperty(TracingAwarePropertiesDelegate.java:58)
    at org.glassfish.jersey.server.ContainerRequest.setProperty(ContainerRequest.java:289)
```

## Steps to reproduce

[...]

PR #2754, opened 2021-02-10
Updated implementation of PropertiesDelegate 

spericas:
Updated implementation of PropertiesDelegate based on Context with support for all methods. This class is used by Jersey to implement a ContainerRequestContext. See issue #2211.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


tjquinno (approved):

common/context/src/main/java/io/helidon/common/context/Context.java
```diff
@@ -162,6 +163,20 @@ static Builder builder() {
      */
     String id();
 
+    /**
+     * Unmodifiable collection of all classifiers registered in this context.
+     *
+     * @return collection of classifiers
+     */
+    Collection<Object> classifiers();
```
tomas-langer, line 171: The context is not considering classifiers to be property names - there is no semantic restriction.
Allowing the list of classifiers to escape is actually not desired - as these may be package private classes.

If you want to use `Context` as a way to propagate Jersey properties across threads, please use a custom object.

```
context.register(MyClass.class, new Properties());
then you can call add/list/remove on the properties object, rather then on context.
```

tomas-langer (changes requested):

common/context/src/main/java/io/helidon/common/context/Context.java
spericas: OK, that sounds reasonable.

PR #2768, opened 2021-02-15
Download Kafka binaries from archive if not present on mirror

danielkec:
Only few latest versions of Kafka binaries are accessible over Apache mirror, download the latest one to keep the docker file from getting obsolete.


tomas-langer (changes requested):
I think we should always use an explicit version of dependencies, even in examples.
You can add this to `README.md` as a way to get the latest version.
We should never have code that can use unknown versions.

tomas-langer (approved):
LGTM

PR #2783, opened 2021-02-19
Remove the Helidon specific info from application effective pom

romain-grecourt:


ljnelson (approved):
Oh, nice approach.  I wouldn't have thought of that.

tomas-langer (approved):

PR #2784, opened 2021-02-20
Cover todo app with tests

tvallin:


examples/todo-app/backend/src/test/resources/test-application.yaml
```diff
@@ -0,0 +1,36 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+server:
+  port: 8854
```
tomas-langer, line 18: We should not use hardcoded ports for testing. 
The server should be opened on a random port (using 0). 
Not sure how to handle this with cassandra (and if it is possible to use an explicit port), but this is not feasible, as it may conflict with existing tools running on the machine.

examples/todo-app/frontend/src/test/java/io/helidon/demo/todos/frontend/FrontendTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+    @AfterAll
+    public static void stopServers() {
+        serverBackend.shutdown();
+        serverFrontend.shutdown();
+    }
+
+    private static void startBackendServer() {
+        serverBackend = WebServer.builder(createRouting())
+                .port(8854)
```
tomas-langer, line 121: Same as above - we cannot use a hardcoded port for tests.

tomas-langer (changes requested):

tomas-langer:
@romain-grecourt could you please confirm if it is OK to add versions for example dependencies, or if these should be version managed in `dependencies/pom.xml`.

romain-grecourt:
> @romain-grecourt could you please confirm if it is OK to add versions for example dependencies, or if these should be version managed in `dependencies/pom.xml`.

Need to double check with Joe, we should add them to dependencies/pom.xml only if we have an existing approval for them. At the very least we should declare properties for these versions to make it look cleaner.

barchetta (approved):

tomas-langer (approved):

tomas-langer (approved):

PR #2800, opened 2021-02-25
Native image test: JPA with MySQL and PgSQL

Tomas-Kraus:
Signed-off-by: Tomas Kraus <Tomas.Kraus@oracle.com>


Tomas-Kraus:
Trying to enable build stages with native image tests. Those tasks are consuming around 8G RAM so I have to make sure it won't break the build.

PR #2805, opened 2021-02-25
Implementation of ResponseWriter using pooled Netty buffers

spericas:
Changes in this PR:

- Some new methods in DataChunk
- New DataChunk implementation based on ByteBuf (Netty)
- Changes in ResponseWriter to use the new DataChunk implementation (ByteBufDataChunk)
- Changes in BareResponseImpl to consume ByteBufDataChunk's
- Allocation of underlying ByteBuf's in ByteBufDataChunk using Netty's pool

With all these changes, the DataChunk's passed from Jersey/Helidon's layer are now backed by pooled Netty's ByteBufs. This should reduced the amount of time spent running GC's in MP applications.


webserver/jersey/src/main/java/io/helidon/webserver/jersey/ResponseWriter.java
```diff
@@ -105,19 +65,14 @@ public void flush() throws IOException {
...
         if (context.getStatus() == 404 && contentLength == 0) {
             whenHandleFinishes.thenRun(() -> {
                 LOGGER.finer("Skipping the handling and forwarding to downstream WebServer filters.");
-
                 req.next();
             });
             return new OutputStream() {
                 @Override
-                public void write(int b) throws IOException {
+                public void write(int b) {
```
ljnelson, line 75: (This could be a static class.)

webserver/jersey/src/main/java/io/helidon/webserver/jersey/ResponseWriter.java
```diff
@@ -178,18 +139,185 @@ public boolean enableResponseBuffering() {
...
+            oneByteArray[0] = (byte) b;
+            write(oneByteArray, 0, 1);
+        }
+
+        @Override
+        public void write(byte[] b, int off, int len) throws IOException {
+            while (len > 0) {
+                if (byteBuf == null) {
+                    awaitRequest();
+                    byteBuf = PooledByteBufAllocator.DEFAULT.buffer(BYTEBUF_DEFAULT_SIZE);
```
ljnelson, line 179: @spericas This was what I was asking about earlier today.  This will not make use of direct buffers, and perhaps that is just fine.  On the other hand, if you use `ByteBufAllocator.DEFAULT` you'll get [some perhaps useful behavior](https://github.com/netty/netty/blob/1529ef1794e0a6654fa4334fd979b769d6940e61/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L77-L93).

ljnelson: Sorry; I got slightly confused here.  `PooledByteBufAllocator` will indeed use direct buffers.  I guess my comment is more: if you simply use `ByteBufAllocator.DEFAULT` you pick up whatever default behavior Netty wants in the future.

webserver/webserver/src/main/java/io/helidon/webserver/ByteBufDataChunk.java
```diff
@@ -0,0 +1,155 @@
...
+    private CompletableFuture<DataChunk> writeFuture;
+
+    /**
+     * Creates an instance given an array of {@code ByteBuf}'s.
+     *
+     * @param flush a signal that this chunk should be written and flushed from any cache if possible
+     * @param readOnly marks this buffer as read only
+     * @param byteBufs the data for this chunk. Should not be reused until {@code releaseCallback} is used
+     */
+    public ByteBufDataChunk(boolean flush, boolean readOnly, ByteBuf... byteBufs) {
```
romain-grecourt, line 49: This does not comply with the dev [guidelines](https://github.com/oracle/helidon/blob/master/DEV-GUIDELINES.md#builders), (no public constructors on public classes)

webserver/jersey/src/main/java/io/helidon/webserver/jersey/ResponseWriter.java
spericas: But If we use `ByteBufAllocator` today, are we guaranteed to get a _pooled_ buffer now? After all, that is the whole point of this PR.

webserver/webserver/src/main/java/io/helidon/webserver/ByteBufDataChunk.java
spericas: Do you think it would be acceptable to just have a static `create` method like `DataChunk`?

romain-grecourt: Yes.

romain-grecourt (commented):
How about merging ByteBufDataChunk and ByteBufRequestChunk ?

spericas:
> How about merging ByteBufDataChunk and ByteBufRequestChunk ?

That class includes the reference queue's logic. Perhaps we can connect the two at some point, but that's a task for a different day in my view.

webserver/jersey/src/main/java/io/helidon/webserver/jersey/ResponseWriter.java
ljnelson: If you see https://github.com/netty/netty/blob/1529ef1794e0a6654fa4334fd979b769d6940e61/buffer/src/main/java/io/netty/buffer/ByteBufUtil.java#L77-L93 you can see what it does.  In terms of guarantees, there is very little documentation so effectively no contract as to what you can expect or depend upon.

romain-grecourt (approved):

webserver/webserver/src/main/java/io/helidon/webserver/ByteBufDataChunk.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Override
+    public Optional<CompletableFuture<DataChunk>> writeFuture() {
+        return Optional.ofNullable(writeFuture);
+    }
+
+    @Override
+    public int remaining() {
+        int remaining = 0;
+        for (ByteBuf byteBuf : byteBufs) {
```
ljnelson, line 142: If you're going to do stuff with multiple `ByteBuf` instances, consider Netty's built-in `CompositeByteBuf`.  It may or may not suit your needs.

ljnelson (approved):
LGTM; I made a note about Netty's built-in `CompositeByteBuf`.

webserver/jersey/src/main/java/io/helidon/webserver/jersey/ResponseWriter.java
danielkec: This could be `OutputStream.nullOutputStream();`

PR #2811, opened 2021-02-26
Fix for Jersey native image.

tomas-langer:
Moved a fix from `helidon-jersey-client` to `helidon-webserver-jersey`, and made dependency on `helidon-jersey-client` explicit, as Jersey server always comes with the client.

Now if you used Helidon MP Server, or Helidon SE Webserver with Jersey, the `helidon-jersey-client` module would not be on the classpath, and our fixes for some Jersey features were missing.


Tomas-Kraus (approved):
Looks good.

tomas-langer:
Rebased and fixed compilation issue caused by defining Jersey client in test scope when it was already transitively added by MP Server (fails now because we have an explicit JPMS module dependency on it)

danielkec (approved):

danielkec (approved):

PR #2817, opened 2021-02-27
Fix JSON conversion error during metrics formatting after idle period 1.x

### LINKED ISSUE #464: Request to endpoint /metrics fails occasionally

tjquinno:
Resolves #464 for Helidon 1.x

See PR #2812 for full details.

### DIFF
--- a/metrics2/metrics2/src/test/java/io/helidon/metrics/HelidonTimerTest.java
+++ b/metrics2/metrics2/src/test/java/io/helidon/metrics/HelidonTimerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018,2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


metrics2/metrics2/src/test/java/io/helidon/metrics/HelidonTimerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2018,2021 Oracle and/or its affiliates.
```
tomas-langer, line 2: Please use spaces between comma and the second year.

tjquinno: Yeah, I know. I had to kick the pipeline to run the job again and was delayed in submitting the second, reverting commit. Sigh.

spericas (approved):

PR #2818, opened 2021-02-27
Add missing metrics coverage and ensure proper coverage of future metrics in gRPC

### LINKED ISSUE #2694: gRPC metrics code replicates logic in metrics component and is possibly out-of-date
## Environment Details
* Helidon Version: 2.x
* Helidon SE or Helidon MP SE
* JDK version: n/a
* OS: n/a

----------
In gRPC, the `MetricsConfigurer` class contains an `accept` method that specifically acts on metrics annotations `@Counted`, `@Timed`, and `@Metered`. That method then delegates to `addMetric` which checks for those (and other, but not all possible) metrics annotations and then runs some logic that replicates code in the metrics component to create metrics.

I'm checking with the gRPC folks whether the code should pay attention to all metrics which SE metrics supports or only some, and in that case which ones.

Further, we should avoid the duplicated code in `MetricsConfigurer` and have it instead delegate to the metrics SE code. If needed we can expose a new method from SE metrics which gRPC can invoke.

tjquinno:
Resolves #2694 

This PR adds support for some metrics annotations that are not currently handled in gRPC. It also reworks some of the gRPC code related to metrics to be data-driven as much as possible to simplify adding proper support for future metrics (should they ever appear).

It also adds tests to help verify that the gRPC code handles all the metrics annotations corresponding to the MP metric types (except those we explicitly exclude because we don't support them).

A new `README.md` describes the steps needed to add support for a new metric in gRPC.

I've done my best to make these changes consistent with what gRPC was already doing. I'll rely on the reviewers to make sure the changes are correct and complete.


aseovic (approved):
Looks great! Really like the refactoring to use functions for metric attributes, removes a lot of `instance of` nonsense that we had 👍

spericas (approved):

PR #2822, opened 2021-03-02
Add support for gRPC reflection

aseovic:


grpc/server/src/main/java/io/helidon/grpc/server/BindableServiceImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
```
tomas-langer, line 2: This is wrong copyright format. Please use `2019, 2021`.

grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates.
```
tomas-langer, line 2: Wrong copyright format

grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2021, Oracle and/or its affiliates. All rights reserved.
```
tomas-langer, line 2: Wrong copyright format.

grpc/server/src/main/java/io/helidon/grpc/server/ProtoReflectionService.java
```diff
@@ -0,0 +1,579 @@
...
+        return new ProtoReflectionService();
+    }
+
+    /**
+     * Retrieves the index for services of the server that dispatches the current call. Computes
+     * one if not exist. The index is updated if any changes to the server's mutable services are
+     * detected. A change is any addition or removal in the set of file descriptors attached to the
+     * mutable services or a change in the service names.
+     */
+    private ServerReflectionIndex getRefreshedIndex() {
```
tomas-langer, line 94: How often is this called? Hard synchronization is expensive... 
Do we need to do it as pull approach (e.g. do we need to query the server for all services, as opposed to getting events about mutable services being changed)?

tomas-langer (changes requested):

grpc/server/src/main/java/io/helidon/grpc/server/ProtoReflectionService.java
aseovic: In reality, only once, so it shouldn't be a big deal.

Services don't really change at runtime, so there shouldn't be any changes that need refreshing.

aseovic:
@tomas-langer Any chance we can get this merged?

tomas-langer (approved):
LGTM

PR #2829, opened 2021-03-04
Remove need to instantiate Application classes twice for OpenAPI support

### LINKED ISSUE #2799: @PostConstruct in RestApplication invoked twice when using openapi
## Environment Details
* Helidon Version: 2.2 (2.2.1 has the issue to)
* Helidon MP
* JDK version: 11
* OS: Windows
----------

## Problem Description
Seem that using the implementation (implementation 'io.helidon.microprofile.openapi:helidon-microprofile-openapi') triggers Helidon to invoke the @PostConstruct in the RestApplication class twice.

## Steps to reproduce
I've included a sample project for reproduction of the issue.
Running the sample project will print "-----> PostConstruct has been invoked <-----" in the console everytime it invokes the @PostConstruct function

[sample-project.zip](https://github.com/oracle/helidon/files/6043361/sample-project.zip)

tjquinno:
Resolves #2799 

If multiple `Application` classes exist in the user's app, we build separate OpenAPI models for each one. As part of that, we need to invoke each `Application`'s `getClasses` and `getSingletons` methods. The OpenAPI CDI extension needs to act before the server's CDI extension actually creates the `Application` instances because the OpenAPI extension needs to adjust the routing to include the OpenAPI endpoint. But at that time, the server had not created the `Application` instances. So to invoke `getClasses` and `getSingletons` the OpenAPI CDI extension would use CDI to create _unmanaged_ instances of each `Application`, invoke those methods, and then dispose of the unmanaged instances. 

This caused `@PostConstruct` methods to be invoked twice, once for the unmanaged instances and again when the server created the "live" instances. Developers found this disconcerting.

Avoiding the double-instantiation involves splitting the OpenAPI CDI extension's work into two steps: first, creating the `OpenAPISupport` object so its routing can be plugged in correctly, and second later on when the `Application` instances are available for use in building the in-memory OpenAPI model.

To do _that_ involved refactoring the model-building logic from the `Builder` class to the `OpenAPISupport` class. 

Allowing the MP CDI extension to prepare the model during start-up (rather than lazily upon first request for the OpenAPI document), but _without_ exposing a `prepareModel()` method as `public` involved creating SE and MP variants of the `OpenAPISupport` class. The now-abstract superclass `OpenAPISupport` exposes a `protected prepareModel()` method which the MP variant overrides, also as `protected`, so the MP CDI extension in the same package can invoke it.

And doing _that_ in a type-safe way requires using type parameters on the `OpenAPISupport` class and its builder, which brought along another set of changes.

As a result, there is a seemingly large number of changes required for a conceptually simple improvement in user experience, but doing it this way avoids expanding the public surface area more than we want.

### DIFF
--- a/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
+++ b/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
@@ -144,7 +144,7 @@ public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends O
     private final OpenApiStaticFile openApiStaticFile;
     private final Supplier<List<? extends IndexView>> indexViewsSupplier;
 
-    protected OpenAPISupport(Builder builder) {
+    protected OpenAPISupport(Builder<T, B> builder) {
         adjustTypeDescriptions(helper().types());
         implsToTypes = buildImplsToTypes(helper());
         webContext = builder.webContext();
@@ -614,7 +614,7 @@ public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends O
      *
      * @return new Builder
      */
-    public static Builder builder() {
+    public static SEOpenAPISupportBuilder builder() {
         return builderSE();
     }
 
@@ -623,7 +623,7 @@ public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends O
      *
      * @return new OpenAPISUpport
      */
-    public static OpenAPISupport create() {
+    public static OpenAPISupport<?, ?> create() {
         return builderSE().build();
     }
 
@@ -636,7 +636,7 @@ public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends O
      * @return new {@code OpenAPISupport} instance created using the
      * helidonConfig settings
      */
-    public static OpenAPISupport create(Config config) {
+    public static OpenAPISupport<?, ?> create(Config config) {
         return builderSE().config(config).build();
     }
 
@@ -661,7 +661,7 @@ public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends O
      * @param <B> concrete subclass of OpenAPISupport.Builder
      */
     public abstract static class Builder<T extends OpenAPISupport<T, B>, B extends Builder<T, B>>
-            implements io.helidon.common.Builder<OpenAPISupport> {
+            implements io.helidon.common.Builder<T> {
 
         /**
          * Config key to select the openapi node from Helidon config.
@@ -681,9 +681,6 @@ public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends O
         protected B me() {
             return builderClass.cast(this);
         }
-
-        @Override
-        public abstract T build();
 
         /**
          * Set various builder attributes from the specified {@code Config} object.
--- a/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
+++ b/openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
@@ -95,10 +95,8 @@ import static io.helidon.webserver.cors.CorsEnabledServiceHelper.CORS_CONFIG_KEY
  * {@code openapi} file, then the {@code /openapi} endpoint responds with a
  * nearly-empty OpenAPI document.
  *
- * @param <T> concrete subclass which extends OpenAPISupport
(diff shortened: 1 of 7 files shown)


openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -126,17 +130,28 @@
...
+    /*
+     * To handle the MP case, we must defer constructing the OpenAPI in-memory model until after the server has instantiated
+     * the Application instances. By then the builder has already been used to build the OpenAPISupport object. So save the
+     * following raw materials so we can construct the model at that later time.
+     */
+    private final OpenApiConfig openApiConfig;
+    private final OpenApiStaticFile openApiStaticFile;
+    private final Supplier<List<? extends IndexView>> indexViewsSupplier;
+
+    protected OpenAPISupport(Builder builder) {
```
tomas-langer, line 147: Should be `Builder<?, ?>` so we do not have a `Raw` use of class with type parameters

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -630,24 +656,35 @@ static SEOpenAPISupportBuilder builderSE() {
      * Helidon SE apps and once for use from the Helidon MP-provided OpenAPI
      * service. This lets us constrain what use cases are possible from each
      * (for example, no anno processing from SE).
+     *
+     * @param <T> concrete subclass of OpenAPISupport
+     * @param <B> concrete subclass of OpenAPISupport.Builder
      */
-    public abstract static class Builder implements io.helidon.common.Builder<OpenAPISupport> {
+    public abstract static class Builder<T extends OpenAPISupport<T, B>, B extends Builder<T, B>>
+            implements io.helidon.common.Builder<OpenAPISupport> {
```
tomas-langer, line 664: Should be `implements io.helidon.common.Builder<T>`

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -630,24 +656,35 @@ static SEOpenAPISupportBuilder builderSE() {
...
+        protected Builder(Class<B> builderClass) {
+            this.builderClass = builderClass;
+        }
+
+        protected B me() {
+            return builderClass.cast(this);
         }
 
+        @Override
+        public abstract T build();
```
tomas-langer, line 686: This can be removed (already implied by `implements io.helidon.common.Builder<T>`)

tomas-langer (changes requested):
A few type parameter related issues.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -93,8 +94,11 @@
  * the server uses none of these builder methods and does not provide a static
  * {@code openapi} file, then the {@code /openapi} endpoint responds with a
  * nearly-empty OpenAPI document.
+ *
+ * @param <T> concrete subclass which extends OpenAPISupport
+ * @param <B> concrete builder subclass which extends OpenAPISupport.Builder
  */
-public class OpenAPISupport implements Service {
+public abstract class OpenAPISupport<T extends OpenAPISupport<T, B>, B extends OpenAPISupport.Builder<T, B>> implements Service {
```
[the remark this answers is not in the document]
tjquinno: The type `B` is used as the return type of `Builder#me()`. Otherwise there is no way for the builder methods that are implemented on the abstract class to return `this` correctly-typed to the builder subclass. Or at least I've not seen one. If there is I would love to do away with `me()`.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Yes, this should not use the raw type. For the moment I've changed it to `Builder<T, B>` to strongly type the parameter.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Fixed.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tjquinno: Fixed.

tomas-langer (approved):
LGTM

PR #2830, opened 2021-03-04
Multipart examples update

romain-grecourt:
Add a unit test that uses WebClient in the multipart example for SE.
Add a new example under microprofile that demonstrates how to use `MultiPartFeature`.

### DIFF
--- a/examples/media/multipart/src/main/java/io/helidon/examples/media/multipart/FileService.java
+++ b/examples/media/multipart/src/main/java/io/helidon/examples/media/multipart/FileService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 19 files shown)


examples/media/multipart/src/main/java/io/helidon/examples/media/multipart/FileService.java
```diff
@@ -125,41 +107,22 @@ private void streamUpload(ServerRequest req, ServerResponse res) {
...
-        } catch (IOException ex) {
-            throw new RuntimeException(ex);
-        }
-    }
-
-    private static void writeBytes(Path storage, String fname, byte[] bytes) {
+    private static void writeBytes(Path file, byte[] bytes) {
         try {
-            Files.write(storage.resolve(fname), bytes,
+            Files.write(file, bytes,
```
tomas-langer, line 120: This is a blocking operation that is not allowed within a reactive server processing.

tomas-langer (changes requested):

examples/media/multipart/src/main/java/io/helidon/examples/media/multipart/FileService.java
romain-grecourt: The only alternative is to create a dedicated thread-pool, shall I do that  ?

tomas-langer: Please create a follow up issue - we need to resolve this for other examples as well.

romain-grecourt: Yes and there is also the blocking start of the webserver in the unit test, that you pointed out in the new multiport PR. I copied that from the quickstarts, so I will create another issue to address that.

romain-grecourt: created #2844 and #2845

PR #2834, opened 2021-03-05
Examples: multiple ports

### LINKED ISSUE #2815: Examples: multiple ports
## Environment Details
* Helidon Version: 2.2.1
* Helidon SE and Helidon MP
----------

## Problem Description
We often get questions concerning running a server on multiple ports. It would be good to have examples for both MP and SE that shows configuring a server to run (for example):

* Public application rest endpoints on 8080
* Private application rest endpoints on 8081
* Admin endpoints (like health and metrics) on 8082

barchetta:
Fixes #2815

Also adds `int port(String name)` to `ServerCdiExtension`


examples/microprofile/multiport/src/main/java/io/helidon/examples/microprofile/multiport/Main.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package io.helidon.examples.microprofile.multiport;
+
+import io.helidon.config.Config;
+import io.helidon.microprofile.server.Server;
+
+/**
+ * Runs helidon on two ports - one for admin, one for business logic.
+ */
+public final class Main {
```
romain-grecourt, line 24: Is the explicit main class required ?

examples/webserver/multiport/src/main/resources/application.yaml
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+#
+server:
+  port: 8080
+  host: 0.0.0.0
+  sockets:
+    private:
+      port: 8081
+    admin:
+      port: 8082
```
tjquinno, line 23: Same note about the newer syntax.

examples/microprofile/multiport/src/main/resources/application.yaml
```diff
@@ -0,0 +1,32 @@
...
+server:
+  port: 8080
+  host: "localhost"
+  sockets:
+    private:
+      port: 8081
+      bind-address: "localhost"
+    admin:
+      port: 8082
+      bind-address: "localhost"
```
tjquinno, line 25: This syntax  is deprecated. There's a warning in the server start-up output to that effect. The newer syntax is like this, using an array/list with a `name` attribute:
```
server:
  port: 8080
  host: "localhost"
  sockets:
    - name: "private"
      port: 8081
      bind-address: "localhost"
    - name: "admin"
      port: 8082
      bind-address: "localhost"
```

tjquinno (changes requested):

examples/microprofile/multiport/src/main/java/io/helidon/examples/microprofile/multiport/PublicResource.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.enterprise.context.RequestScoped;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+
+/**
+ * Simple resource.
+ */
+@RequestScoped
+@Path("/public")
```
tjquinno, line 26: I don't think we want to have `/public` prefixed to every public path, do we? It does emphasize that they are public paths, but that's not how an actual app would be set up.

tjquinno:
I came at this by considering how a user might convert our classic greeting app to use multiple routing names.

## First choice: add `@RoutingName` to the `PUT` method for changing the greeting
Ideally, there would be the obvious change to config and then _one line_ of change in the code: add `@RoutingName("private")` to the `PUT` method on the existing resource. The public port would be available to outside requests but not the private one so only internal users could change the greeting.

This will not even compile because the `@RoutingName` annotation is defined to apply only to types. Ideally, it would work on methods as well. 

## Second choice: add `@RoutingName` to the resource classes
The next best would be to create two resources, one for the public endpoints with `@Path("/greet")` and the second annotated with `@RoutingName("private") @Path("/greet/greeting")` for the private ones. That builds but does not work as desired; the app endpoints (`/greet` and `/greet/greeting`) seem available on both the default routing and the `private` one so it looks as if the `@RoutingName` on resources is ignored.

## Last choice: split the existing resource and create two `Application` classes (as you've had to do)
So, are we really telling developers that to use this feature they need take the quick-start example (or their own app), create two `Application` classes so they can specify the `@RoutingName("private")` on one and then write their respective `getClasses` methods, and also divide the resource class into two different resource classes so each application can report its own resource? 

Ugh.

Even with the first and second choices not possible, it would be good in our example to at least preserve the external API of the greeting app. I tried binding the "public" resource to the path `/greet` and the "private" one to `/greet/greeting` and that seems to work. Trying to `PUT` to the public port returns a `405 Method Not Allowed` which makes sense.

examples/microprofile/multiport/src/main/java/io/helidon/examples/microprofile/multiport/Main.java
tjquinno: In my experimenting I did not need seem to need one.

examples/microprofile/multiport/src/main/resources/application.yaml
barchetta: Thanks for the catch. Looks like we need to update our docs. I was also going to see if I could do this via microprofile config, but the new syntax does not seem compatible. I'll experiment.

examples/microprofile/multiport/src/main/java/io/helidon/examples/microprofile/multiport/PublicResource.java
barchetta: Yes that was done for emphasis with the assumptions folks know they can easily change this.

barchetta:
@tjquinno I experimented with those first two options as well with the same results.

tjquinno:
@barchetta It might be worth seeing if the `@RoutingName` processing could be changed so one of the first two approaches would work.

examples/microprofile/multiport/src/main/java/io/helidon/examples/microprofile/multiport/Main.java
barchetta: Fixed. Removed Main class.

examples/webserver/multiport/src/main/resources/application.yaml
barchetta: Fixed. Used new syntax.

examples/microprofile/multiport/src/main/resources/application.yaml
barchetta: Fixed. Used new syntax.

examples/microprofile/multiport/src/main/java/io/helidon/examples/microprofile/multiport/PublicResource.java
barchetta: Fixed. Public endpoints now default (no `/public` prefix).

examples/microprofile/multiport/src/test/java/io/helidon/examples/microprofile/multiport/MainTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/**
+ * Unit test for {@link Main}.
+ */
+@HelidonTest
+class MainTest {
```
tomas-langer, line 36: Testing should never use set ports, this should use random ports. 
Right now, you can only use `HelidonTest` to inject a client pointing to the main port (which is randomized by the test itself).
To be able to use additional ports, we would need to either update the `HelidonTest` implementation to support additional clients, or add injection support for `WebServer` in Helidon MP.

examples/webserver/multiport/src/test/java/io/helidon/examples/webserver/multiport/MainTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+public class MainTest {
+
+    private static WebServer webServer;
+    private static WebClient webClient;
+    private static final int PUBLIC_PORT = 8080;
```
tomas-langer, line 33: Do not use explicit ports. You have access to webserver, so for testing use random ports and retrieve them using `WebServer.port()` and `WebServer.port(String)`

examples/webserver/multiport/src/test/java/io/helidon/examples/webserver/multiport/MainTest.java
```diff
@@ -0,0 +1,123 @@
...
+    private static final int PRIVATE_PORT = 8081;
+    private static final int ADMIN_PORT = 8082;
+
+    @BeforeAll
+    public static void startTheServer() throws Exception {
+        webServer = Main.startServer();
+
+        long timeout = 2000; // 2 seconds should be enough to start the server
```
tomas-langer, line 41: Active blocking is not the right thing to do.
Refactor the method `Main.startServer()` to return a `Single<WebServer>` so you can use `await` on it, or just call `start()` again, as you would get a new Single waiting on the same future

tomas-langer (changes requested):

examples/webserver/multiport/src/test/java/io/helidon/examples/webserver/multiport/MainTest.java
barchetta: Fixed.

examples/webserver/multiport/src/test/java/io/helidon/examples/webserver/multiport/MainTest.java
barchetta: Fixed.

examples/microprofile/multiport/src/test/java/io/helidon/examples/microprofile/multiport/MainTest.java
barchetta: Fixed.
For now I explicitly set the config to dynamically generate ports and inject `ServerCdiExtension` to access the port numbers of the running server (this required adding` ServerCdiExtension.getPort(String)` which was missing).

examples/microprofile/multiport/src/test/java/io/helidon/examples/microprofile/multiport/MainTest.java
```diff
@@ -0,0 +1,117 @@
...
+        // Probe PRIVATE port
+        baseTarget = client.target(base + privatePort);
+        response = baseTarget.path("/private/hello")
+                .request()
+                .get();
+        assertThat("port " + privatePort  + " should be serving private resource",
+                response.getStatusInfo().toEnum(), is(Response.Status.OK));
+        assertThat("port " + privatePort  + " should return private data",
+                response.readEntity(String.class), is("Private Hello World!!"));
+
```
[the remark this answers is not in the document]
barchetta: Fixed. I've added a parameterized test to do this.

examples/microprofile/multiport/src/test/java/io/helidon/examples/microprofile/multiport/MainTest.java
```diff
@@ -0,0 +1,117 @@
...
+                .get();
+        assertThat("port " + adminPort  + " should be serving health",
+                response.getStatusInfo().toEnum(), is(Response.Status.OK));
+
+        response = baseTarget.path("/metrics")
+                .request()
+                .get();
+        assertThat("port " + adminPort  + " should be serving metrics",
+                response.getStatusInfo().toEnum(), is(Response.Status.OK));
+    }
```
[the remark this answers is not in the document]
barchetta: Fixed. I've added a parameterized test to do this.

tjquinno (approved):

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -502,6 +502,17 @@ public int port() {
         return port;
     }
 
+    /**
```
tomas-langer, line 505: Adding a public API as part of an unrelated example. For future, this would deserve its own PR. Also please create a follow up to add tests for this new feature.

tomas-langer (approved):
LGTM, please see the comment

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
barchetta: I agree. Will add the test on a separate PR.

Issue #2838, opened 2021-03-08
Context is not available throughout request lifecycle


FWiesner:
* Helidon Version: 2.2.0
* Helidon SE
* JDK version: GraalVM 20.3.0
* OS: various

----------

## Problem Description
We try accessing the request context in a custom deserializer method for Jackson as deserialization in our case is based on the REST request. `JacksonSupport` initializes our custom Deserializer two times. Once for the check of whether a type is assignable and a second time for the actual deserialization. In a debugging session with @tomas-langer, he figured out that the deserialization step has no access to the Context, which we would need.

## Steps to reproduce
Implement a custom Jackson deserializer and register it with a Jackson `ObjectMapper`, which in turn you use to initialize `JacksonSupport`. Calls to `Contexts.context()` will not return you the request context.

tomas-langer:
Hi Florian,
 we have discovered that this problem should be fixed at a different place - we need to run in context for any operation that handles the publisher operations. This will automatically fix the problem with media supports as well.

tomas-langer:
The class to fix this in:
`ForwardingHandler` + the chain from there - the `Context` must be created in Forwarding handler, sent over to request...

PR #2839, opened 2021-03-08
Fix for #2838 - Update to ForwardingHandler and Request to propagate Context to MediaSupport

FWiesner:
~JacksonBodyReader#read now propagates the current Context~


FWiesner:
The `earlyOutFuture` is there because I didn't dare refactoring `channelRead0` and cut it into chunks. I bet you had done a lot of tuning there and I don't want to affect it too much. The `runInContext` in case of handling a `HttpRequest` is also certainly too greedy in it's span, but there I didn't know what different sorts of dispatches you have. So, I've covered everything after the `reset`

romain-grecourt:
Done with https://github.com/oracle/helidon/pull/2898

Issue #2632, opened 2021-01-04
Request Scope not active in Fault Tolerance handling


bunto76:
## Environment Details
* Helidon Version: 2.2.0
* Helidon SE or Helidon MP
* JDK version: 11.0.6
* OS: macOS 10.15.7

----------

## Problem Description
This is an intermittent problem that I can only reproduce locally about 10% of the time and only by constantly re-running the tests. It is causing periodic fails in project jenkins build, which has highlighted the issue. During Fault Tolerant handling, Retry and Fallback handling, the Request Scope is not active or not available during the retry or fallback handling. I can't say if it is exclusive to those scenarios but those are the scenarios used in this code.
```
[ApplicationContainer] 	at com.oracle.cxcom.image.cache.ImageCacheAccessor.getCachedSourceImage(ImageCacheAccessor.java:134)
[ApplicationContainer] 	at com.oracle.cxcom.image.cache.ImageCacheAccessor$Proxy$_$$_WeldSubclass.getCachedSourceImage$$super(Unknown Source)
[ApplicationContainer] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ApplicationContainer] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[ApplicationContainer] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ApplicationContainer] 	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[ApplicationContainer] 	at org.jboss.weld.interceptor.proxy.TerminalAroundInvokeInvocationContext.proceedInternal(TerminalAroundInvokeInvocationContext.java:51)
[ApplicationContainer] 	at org.jboss.weld.interceptor.proxy.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:78)
[ApplicationContainer] 	at io.helidon.microprofile.faulttolerance.MethodInvoker.lambda$requestContextSupplier$11(MethodInvoker.java:484)
[ApplicationContainer] 	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
[ApplicationContainer] 	at org.glassfish.jersey.internal.Errors.process(Errors.java:219)
[...]

PR #2856, opened 2021-03-11
New RequestScopeHelper class to track request scopes across threads

spericas:
New class in FT that isolates code related to saving and restoring request scopes. The new class `RequestScopeHelper` provides support for CDI, HK2 and multiple JAX-RS applications; it includes additional code to track injection managers across threads. New test that verifies issue reported in #2632. For more information, see https://github.com/oracle/helidon/issues/2632#issuecomment-796831904


microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
```diff
@@ -315,15 +298,8 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
-            requestController = CDI.current().select(RequestContextController.class).get();
-            requestScope = CDI.current().select(RequestScope.class).get();
-            requestContext = requestScope.referenceCurrent();
-        } catch (Exception e) {
-            requestScope = null;
-            LOGGER.fine(() -> "Request context not active for method " + method
-                    + " on thread " + Thread.currentThread().getName());
-        }
+        requestScopeHelper = new RequestScopeHelper();
+        requestScopeHelper.saveScope();
```
tjquinno, line 302: Would we ever legitimately instantiate a helper without also saving the scope right away? 

If not, maybe fold the save work into the constructor? Or make the constructor private and use a static factory method that instantiates the helper and then invokes its `saveScope` method?

That would not work if there are ever cases where a single helper instance is created and then reused through several `save` and `clear` cycles.

tjquinno (commented):
Looks good with one small question.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
spericas: I don't have a use case for saving/clearing/saving yet, so it is a bit more generic than necessary. Having said that, I think it easier to read with an explicit call to `saveScope`.

tjquinno: My concern was that correct usage requires me to remember to both instantiate the helper and invoke `saveScope` explicitly. This is internal so I'm not overly worried.

romain-grecourt:
Can you rename this PR to something meaningful and use `Fixes #2632`. Otherwise it shows up in the git log and does mean much after a while.

tjquinno (approved):

PR #2859, opened 2021-03-12
New implementation of WorkQueue that allows concurrent enqueuing of tasks

spericas:
New implementation of `WorkQueue` that allows concurrent enqueuing of tasks. Previous implementation was based on `LinkedBlockingQueue` which uses a lock to prevent concurrent additions to the queue. Under high concurrency, this effectively introduces a "chicane" that results in a lot of threads waiting and a poor CPU utilization. The new implementation uses a `ConcurrentLinkedQueue` and two semaphores to enable concurrent enqueuing of tasks. The poor CPU utilization was discovered while running DBClient tests that used a ThreadPool to execute blocking calls. Credit to @olotenko.

### DIFF
--- a/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
+++ b/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
@@ -508,8 +508,11 @@ public class ThreadPool extends ThreadPoolExecutor {
             if (!semaphoreWrite.tryAcquire()) {
                return false;
             }
-            enqueue(task);
-            semaphoreRead.release();
+            try {
+                enqueue(task);
+            } finally {
+                semaphoreRead.release();
+            }
             return true;
         }
 
@@ -527,8 +530,11 @@ public class ThreadPool extends ThreadPoolExecutor {
            if (!semaphoreWrite.tryAcquire(timeout, tu)) {
               return false;
            }
-           enqueue(task);
-           semaphoreRead.release();
+           try {
+               enqueue(task);
+           } finally {
+               semaphoreRead.release();
+           }
            return true;
         }
 
@@ -544,8 +550,11 @@ public class ThreadPool extends ThreadPoolExecutor {
         @Override
         public void put(Runnable task) throws InterruptedException {
            semaphoreWrite.acquire();
-           enqueue(task);
-           semaphoreRead.release();
+           try {
+               enqueue(task);
+           } finally {
+               semaphoreRead.release();
+           }
         }
 
         @Override
--- a/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
+++ b/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,14 +28,17 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.Semaphore;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.ThreadPoolExecutor;
@@ -54,7 +57,7 @@ import io.helidon.common.context.ContextAwareExecutorService;
 /**
  * A {@link ThreadPoolExecutor} with an extensible growth policy and queue state accessors.
  */
-public class    ThreadPool extends ThreadPoolExecutor {
+public class ThreadPool extends ThreadPoolExecutor {
     private static final Logger LOGGER = Logger.getLogger(ThreadPool.class.getName());
     private static final int MAX_GROWTH_RATE = 100;
 
@@ -397,7 +400,7 @@ public class    ThreadPool extends ThreadPoolExecutor {
             // Just add it to the queue if there is capacity
 
             final WorkQueue queue = ((ThreadPool) executor).getQueue();
-            if (!queue.enqueue(task)) {
+            if (!queue.offer(task)) {
 
                 // No capacity, so reject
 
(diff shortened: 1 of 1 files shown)


common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
```diff
@@ -498,31 +505,76 @@ void setPool(ThreadPool pool) {
 
         @Override
         public boolean offer(Runnable task) {
-            return enqueue(task);
+            if (!semaphoreWrite.tryAcquire()) {
+               return false;
+            }
+            enqueue(task);
```
ljnelson, line 511: Naïve reviewer comment: if this throws a (runtime) exception, `semaphoreRead` will remain unreleased, right?  Can `enqueue` throw an exception?

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
```diff
@@ -468,20 +471,24 @@ private static WorkQueue createQueue(int capacity, int corePoolSize, int maxPool
...
+         * Constructor. Initially {@code capacity} writes (enqueues) and 0 reads (dequeues)
+         * are available.
          *
          * @param capacity The queue capacity.
          */
         WorkQueue(int capacity) {
-            super(capacity);
             this.capacity = capacity;
+            this.semaphoreRead = new Semaphore(0);
+            this.semaphoreWrite = new Semaphore(capacity);
```
ljnelson, line 491: Is there a reason (I'm sure there is) that a `ReadWriteLock` isn't used instead?

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
spericas: I asked myself that question too. Since the new queue is really unbounded, only limited by the semaphores, maybe only an NPE can be thrown. However, perhaps using a finally block is more elegant.

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
spericas: That's just a pair of locks, right? I don't think that would enable the type of concurrency that is sought after here unless I'm missing something.

olotenko: Two things:

1. `Semaphore` allows up to `capacity` concurrent writers to make progress. (A lock showed contention)
2. `Semaphore` allows a release by a thread that doesn't own the lock. (readers unblock writers, writers unblock readers)

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
spericas: Please ignore latest commit, did not intent to push. Fixing now.

olotenko: A `finally` block can give you unwanted results, if indeed anything is thrown. Ie a `release` will result in reader/writer making progress, and polling `null` from the queue.

It makes sense to check the `task` is non-null (before acquiring `Semaphore`). Then we will rely on `j.u.c.ConcurrentLinkedQueue` not throwing.

Alternatively, in `enqueue` catch exceptions and release `semaw` (not `semar`) upon exceptions or if `super.offer` returns `false` (should never do that for `ConcurrentLinkedQueue`).

olotenko: eg 
```
private boolean enqueue(Runnable task) {
   boolean r=false;
   try {
             r = super.offer(task);
   } finally {
      if (!r) {
         semaw.release(); // let another offer/put try their luck
      }
   }
```

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
ljnelson: The second point was what I wasn't remembering; thanks!

ljnelson (approved):

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
spericas: I accidentally pushed those changes from Intellij. Please check final version.

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
```diff
@@ -498,31 +505,91 @@ void setPool(ThreadPool pool) {
...
+        }
+
+        @Override
+        public void put(Runnable task) throws InterruptedException {
+           semaphoreWrite.acquire();
+           try {
+               enqueue(task);
+           } catch (Exception e) {
+               semaphoreWrite.release();
+               throw e;
```
olotenko, line 561: This probably doesn't compile.

I would prefer a `try...finally` inside `enqueue`. The reason is that `enqueue` knows where the underlying queue has been touched successfully, and where it wasn't (anything happened after `offer` returned successfully). So if it was touched successfully, should `semaphoreRead.release`, otherwise should `semaphoreWrite.release`.

```
private boolean enqueue(...) {
   boolean offered = false;
   try {
      offered = super.offer(...);
      ...
   } finally {
      (offered? semaphoreRead: semaphoreWrite).release();
   }
}
```

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
```diff
@@ -556,6 +623,48 @@ public float getAverageSize() {
...
+        }
+
+        @Override
+        public int drainTo(Collection<? super Runnable> c, int m) {
+            if (m <= 0) {
+                return 0;
+            }
+
+            Runnable r = poll();
+            int i = 1;
```
olotenko, line 643: I think there is an off-by-one bug here. It should be safer to increment `i` just before adding `r`:

```
public int drainTo(Collectio<? super Runnable> c, int m) {
   if (m <= 0) {
      return 0;
   }
   Runnable r = poll();
   int i = 0;
   m--;
   while (m > 0 && r != null) {
      i++;
      c.add(r);
      m--;
      r = poll();
   }
   if (r != null) {
      i++;
      c.add(r);
   }
   return i;
}
```

olotenko (changes requested):

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
spericas: > This probably doesn't compile.

 This is fine. If it can be thrown by the try block without a throws clause in the method, it can be _safely_ re-thrown in the catch block.

> 
> I would prefer a `try...finally` inside `enqueue`. The reason is that `enqueue` knows where the underlying queue has been touched successfully, and where it wasn't (anything happened after `offer` returned successfully). So if it was touched successfully, should `semaphoreRead.release`, otherwise should `semaphoreWrite.release`.
> 
> ```
> private boolean enqueue(...) {
>    boolean offered = false;
>    try {
>       offered = super.offer(...);
>       ...
>    } finally {
>       (offered? semaphoreRead: semaphoreWrite).release();
>    }
> }
> ```

 Either way works for me, let me try this.

olotenko (approved):

ljnelson (approved):

PR #2871, opened 2021-03-23
Hash equals

### LINKED ISSUE #2877: Fix hash comparison
## Problem Description

When comparing message digests it is a best practice to do a full comparison of every byte and not short circuit comparisons when the first difference is found.

We should fix the comparison here:  https://github.com/barchetta/helidon/blob/ac8a3ebb0cb4aa604452db1568ee2df477a00ac3/security/providers/http-auth/src/main/java/io/helidon/security/providers/httpauth/DigestToken.java#L241

barchetta:
When comparing message digests it is a best practice to do a full comparison of every byte and not short circuit comparisons when the first difference is found.

This change replaces string comparison in http auth with `MessageDigest.isEqual()` when comparing message digests.

Fixes #2877


tomas-langer (approved):
LGTM

PR #2882, opened 2021-03-25
WebClient keep-alive minor improvements

### LINKED ISSUE #2881: WebClient should check Connection header when enabling keep-alive
WebClient should reflect whether Connection header has been set by the user and reflect whether keep-alive should be used or not.

Verdent:
Fixes #2881


tomas-langer (approved):
LGTM

PR #2900, opened 2021-03-31
Reloadable WebServer TLS during runtime

### LINKED ISSUE #2863: Need support for hot SSL certificate reload
We need to be able to reload SSL keys while the server is running. 

## Environment Details
* Helidon Version: 1.4.8
* Helidon SE 
* JDK version: 1.8
* OS: Linux OEL7

----------

## Problem Description
Currently, if we modify the certificate the application will have to be restarted. We have a requirement where there will be multiple clients connecting to the application and each will have their own certificate. In this situation, restarting the application is not feasible as that will require coordination between multiple clients. Would it be possible for Helidon to allow SSL certificates to be reloaded at runtime.

## Steps to reproduce
Nothing to reproduce. This is an enhancement request.

Verdent:
Fixes #2863


webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
```diff
@@ -131,6 +133,23 @@ void queuesShutdown() {
         });
     }
 
+    void updateSslContext(String socketName, SslContext context) {
+        if (sslContext == null) {
+            if (context != null) {
+                throw new IllegalStateException("TLS was not configured on the socket " + socketName + " before. Could not be "
```
romain-grecourt, line 139: This sentence is a bit weird. Maybe just "Disabling TLS context at runtime is not allowed". Disable implies that it is currently enabled, I don't think we need to mention the socket name.

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
```diff
@@ -131,6 +133,23 @@ void queuesShutdown() {
...
+                                                        + "enabled at runtime.");
+            }
+            //Do not set anything
+        } else if (context == null) {
+            //Disabling TLS context is not possible at runtime
```
romain-grecourt, line 144: Here you mean "TLS was **not** configured" right ?
Maybe just mirror the message from the previous branch: "Enabling TLS context at runtime is not allowed". Enable implies that it is currently disabled.

webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java
```diff
@@ -420,6 +426,19 @@ public int port(String name) {
...
 
+    @Override
+    public void updateTls(WebServerTls tls) {
+        updateTls(tls, WebServer.DEFAULT_SOCKET_NAME);
+    }
+
+    @Override
+    public void updateTls(WebServerTls tls, String socketName) {
+        SslContext context = createSslContext(tls.sslContext(), new HashSet<>(tls.enabledTlsProtocols()), tls.clientAuth());
+        Optional.ofNullable(initializers.get(socketName))
```
romain-grecourt, line 437: I know you've refactored that code but the use of optional is not needed. A good old if-null-throw-else would be more appropriate.

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -143,6 +143,27 @@ default int port() {
      */
     int port(String socketName);
 
+    /**
+     * Update TLS configuration of the default socket {@link WebServer#DEFAULT_SOCKET_NAME}.
```
romain-grecourt, line 147: Update **the** TLS configuration.

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -143,6 +143,27 @@ default int port() {
...
+     *
+     * If the socket has not been https and now the TLS is being configured, an exception will be thrown.
```
romain-grecourt, line 149: I think it might be better to document exceptions with`\@throws` event if they are runtime exceptions, and add them to the signature.

How about simplifying the errors?:
- require a non null returned from `tls.sslContext()`
- if the current sslContext is null throw an `IllegalStateException` with "TLS context is not set, update not allowed"

The exceptions from HttpInitializer could remain, but this would be part of the WebServer contract and would be implemented by `NettyWebServer`.

romain-grecourt (commented):
LGTM, only a few minor comments. How does that work for request being processed ? Do they still work off the previously configured TSL context ?

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
Verdent: I have used tls.enabled() instead of tls.sslContext() since sslContext() is set to null only if the TLS is not enabled. Otherwise it makes sense.

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
Verdent: Reworked

webserver/webserver/src/main/java/io/helidon/webserver/HttpInitializer.java
Verdent: Reworked

Verdent:
Yes, those already connected are still working fine, but the new ones are validated against the new TLS.

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -143,6 +143,23 @@ default int port() {
...
+    void updateTls(WebServerTls tls);
+
+    /**
+     * Update the TLS configuration of the named socket.
+     *
+     * @param tls new TLS configuration
+     * @param socketName specific named socket name
+     * @throws IllegalStateException if the new tls is not enabled or currently set tls is null
+     */
+    void updateTls(WebServerTls tls, String socketName);
```
romain-grecourt, line 161: Should we update the prototype to also throw the exception ? I think checkstyle might complain.

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
```diff
@@ -143,6 +143,23 @@ default int port() {
...
+     * @throws IllegalStateException if the new tls is not enabled or currently set tls is null
+     */
+    void updateTls(WebServerTls tls);
+
+    /**
+     * Update the TLS configuration of the named socket.
+     *
+     * @param tls new TLS configuration
+     * @param socketName specific named socket name
+     * @throws IllegalStateException if the new tls is not enabled or currently set tls is null
```
romain-grecourt, line 159: Looking at the constructor for [SockerConfig](https://github.com/oracle/helidon/blob/b7cbd81c10b2fada78399f655ae1980c5ce6b8d1/webserver/webserver/src/main/java/io/helidon/webserver/ServerBasicConfig.java#L211), we don't keep the `WebServerTls` so it's not symetrical :(

How about `if {@link WebServerTls#enabled()} returns {@code false} or if {@link SocketConfiguration#ssl()} returns {@code null}.

We should probably update the ssl configuration for 3.0 @tomas-langer thoughts ?

webserver/webserver/src/main/java/io/helidon/webserver/WebServer.java
Verdent: Hmmm I dont think it will. I have checked the other methods in this interface and they have it the same way.

Issue #2913, opened 2021-04-05
MP ReST Client & Graal Static Native results in ClassCastException


dansiviter:
## Environment Details
* Helidon Version: 2.2.2
* Helidon SE or Helidon MP: MP
* JDK version: 11
* OS: N/A
* Docker version (if applicable): Docker for Windows

----------

## Problem Description
Creating a Graal static native image using Microprofile ReST Client results in ClassCastException:

```
#11 22.27 [WARNING] WARNING: Feature 'REST Client' for path 'RESTClient' has limited support in native image: Does not support execution of default methods on interfaces.
#11 23.84 [WARNING] Fatal error:java.lang.ClassCastException
#11 23.84 [WARNING]     at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
#11 23.85 [WARNING]     at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
#11 23.85 [WARNING]     at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
#11 23.85 [WARNING]     at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
#11 23.85 [WARNING]     at java.base/java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:603)
#11 23.85 [WARNING]     at java.base/java.util.concurrent.ForkJoinTask.get(ForkJoinTask.java:1006)
#11 23.85 [WARNING]     at com.oracle.svm.hosted.NativeImageGenerator.run(NativeImageGenerator.java:488)
#11 23.85 [WARNING]     at com.oracle.svm.hosted.NativeImageGeneratorRunner.buildImage(NativeImageGeneratorRunner.java:370)
#11 23.85 [WARNING]     at com.oracle.svm.hosted.NativeImageGeneratorRunner.build(NativeImageGeneratorRunner.java:529)
#11 23.85 [WARNING]     at com.oracle.svm.hosted.NativeImageGeneratorRunner.main(NativeImageGeneratorRunner.java:119)
#11 23.85 [WARNING]     at com.oracle.svm.hosted.NativeImageGeneratorRunner$JDK9Plus.main(NativeImageGeneratorRunner.java:561)
[...]

PR #2917, opened 2021-04-06
ClassCastException during Graal native image and ReST Client

dansiviter:
I've had issues testing this locally. I can replicate the issue and the build gets further with these changes, but ultimately still fails but for another reason.

### DIFF
--- a/tests/integration/native-image/mp-1/src/main/java/io/helidon/tests/integration/nativeimage/mp1/Mp1Main.java
+++ b/tests/integration/native-image/mp-1/src/main/java/io/helidon/tests/integration/nativeimage/mp1/Mp1Main.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2021 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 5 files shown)


dansiviter:
I've updated the manual integration tests to trigger a `int` primitive `@PathParam` on the ReST Client.

tests/integration/native-image/mp-1/src/main/java/io/helidon/tests/integration/nativeimage/mp1/Mp1Main.java
```diff
@@ -191,8 +191,9 @@ private static void testBean(int port, String jwtToken) {
         // + JSON-B
         invoke(collector, "Rest client JSON-B", "json-b", aBean::restClientJsonB);
         // + query params
-        invoke(collector, "Rest client Query param long", "1020", () -> aBean.restClientQuery(1020L));
```
dansiviter, line 194: I updated the message to reflect they're targeting the wrapper Objects and not the primitive type.

m0mus:
Fixes #2913

dansiviter:
Not sure why it's stating test failures. When I click on it it says `BUILD SUCCESS` in the log.

tomas-langer (approved):
LGTM

Issue #2587, opened 2020-12-08
HTTP/2 stream error with high number of concurrent request


spericas:
Environment Details
·        Helidon Version: 2.1.0
·        Helidon MP
·        JDK version: 11
·        OS: Windows or linux
·        Docker version (if applicable): NA
Problem Description

·        Helidon MP application has simple GET /great api configured with http2 protocol, which we are trying to hit with 100+ user using JavaScript client. Out of 100 request very few request are getting failed and  Failure rate varies every run.

·        We have tested Concurrent user request with Gatling , JMeter, Chrome, Firefox. concurrency working with Gatling and JMeter . but when its tested with browser very few request getting errored out giving HTTP2_server_refused_stream or ERR_CONNECTION_CLOSED on browser. this issue occurring only with http2, not with http1.

State if the problem is easily reproducible or happens intermittently

·        Reproducible

Steps to reproduce

·        download the code from Attachment.

·        unzip helidon-quickstart-mp.zip

·        cd helidon-quickstart-mp/target

·        java -jar helidon-quickstart-mp.jar

·        run sync_client_test_1.html in browser

·        open network tab and select XHR from the filter

·        refresh the page again. will start seeing out of 100 request few request be shown as failed with error "connection closed" or "HTTP2 server refused stream" or ERR:NET_FAILED.

·        Note: in case you are not able to find any failure reload the page several time.
or else open sync_client_test_1.js with notepad, edit below number from 100 to 200 or more and reload the page
var arr_list = new Array(100);

 

Include stack traces or command outputs

```
java.util.concurrent.CompletionException: java.lang.IllegalStateException: An exception occurred when writing last http content.
at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:346)
at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:632)
at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
at io.helidon.common.reactive.SingleToFuture.onError(SingleToFuture.java:74)
at io.helidon.common.reactive.DeferredScalarSubscription.error(DeferredScalarSubscription.java:132)
[...]

PR #2920, opened 2021-04-09
Fixes problems with HTTP/2 connections 

spericas:
Summary of changes:

1. Proper handling of RST_STREAM event
2. Fixed write sequencing between responses
3. Improved logging that includes handler, response and stream ID (http/2) information

Issue #2587.


webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
```diff
@@ -355,6 +366,15 @@ protected void channelRead0(ChannelHandlerContext ctx, Object msg) {
      */
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+        LOGGER.fine(() -> log("Exception caught: %s", ctx, cause.toString()));
+
+        // We ignore stream resets (RST_STREAM) from HTTP/2
+        if (cause instanceof Http2Exception.StreamException
+                && ((Http2Exception.StreamException) cause).error() == Http2Error.CANCEL) {
+            return;     // no action
```
ljnelson, line 374: no action, including `ctx.close()`?

ljnelson (approved):
LGTM

webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
spericas: Right we do not want close the connection in this case

PR #2933, opened 2021-04-14
Handle startups with request scopes available from CDI

spericas:
Guard for the case in which a request scope object is already available from CDI but a request context is not. New test.


tjquinno (approved):
LGTM

PR #2934, opened 2021-04-14
Allow use of config to set up built-in disk space and heap memory health checks

### LINKED ISSUE #2885: HealthChecks should be configurable in Helidon SE
There is no way to configure health checks using `Config` in Helidon SE.

There should be a method `HealthChecks.healthChecks(Config)` and for each health check a method on `Builder` `Builder config(Config)` 

Then I should be able to have `application.yaml`:

```yaml
helidon.health.diskSpace:
  thresholdPercent: 99.9
```

And code:
```java
Config config = Config.create();
HealthSupport.builder()
 .addLiveness(HealthChecks.healthChecks(config.get("helidon.health")))
...
```

tjquinno:
Resolves #2885

### DIFF
--- a/health/health-checks/src/main/java/io/helidon/health/checks/HeapMemoryHealthCheck.java
+++ b/health/health-checks/src/main/java/io/helidon/health/checks/HeapMemoryHealthCheck.java
@@ -155,7 +155,7 @@ public final class HeapMemoryHealthCheck implements HealthCheck {
         /**
          * Set up the heap space health check via config key, if present.
          *
-         * @param config root {@code Config} node
+         * @param config {@code Config} node for heap memory
          * @return updated builder instance
          */
         public Builder config(Config config) {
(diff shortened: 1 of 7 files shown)


health/health-checks/src/main/java/io/helidon/health/checks/HeapMemoryHealthCheck.java
```diff
@@ -144,5 +151,19 @@ public Builder thresholdPercent(double threshold) {
             this.threshold = threshold;
             return this;
         }
+
+        /**
+         * Set up the heap space health check via config key, if present.
```
tomas-langer, line 156: The javadoc should contain a table of config keys used, description and default value. 
Example:
```
Configuration options:
             * <table class="config">
             * <caption>Secret configuration</caption>
             * <tr>
             *     <th>key</th>
             *     <th>default value</th>
             *     <th>description</th>
             *     <th>builder method</th>
             * </tr>
             * <tr>
             *     <td>{@value CONFIG_KEY_THRESHOLD_PERCENT_SUFFIX}</td>
             *     <td>{@code 99.9}</td>
             *     <td>Percentage of heap memory that needs to be full for this health check to fail</td>
             *     <td>{@link #tresholdPercent(double}</td>
             * </tr>
             * </table>
```

tomas-langer (changes requested):

health/health-checks/src/main/java/io/helidon/health/checks/HeapMemoryHealthCheck.java
tjquinno: Added for both classes.

tomas-langer (approved):
LGTM

PR #2938, opened 2021-04-15
Fix delay computation in Retry policies

### LINKED ISSUE #2937: Retry.DelayingRetryPolicy does not work with delays
## Environment Details
* Helidon Version: 2.2.2
* Helidon SE or Helidon MP
* JDK version: 11
* OS: Linux

----------

## Problem Description
Retry.DelayingRetryPolicy is always repeating without a delay. 

## Steps to reproduce
Create a DelayingRetryPolicy with delay(Duration.ofSeconds(5)) and delayFactor(1.0) - it won't do any delays.

martin-sladecek:
Resolves #2937

### DIFF
--- a/fault-tolerance/src/main/java/io/helidon/faulttolerance/RetryImpl.java
+++ b/fault-tolerance/src/main/java/io/helidon/faulttolerance/RetryImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2021 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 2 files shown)


spericas (changes requested):
LGTM. So the `lastDelay` was not set in context, basically. I see that `PartialPublisher` is dead code in the test. Could we remove it? I know you didn't add it, but just noticed that. Thanks.

martin-sladecek:
> LGTM. So the `lastDelay` was not set in context, basically. I see that `PartialPublisher` is dead code in the test. Could we remove it? I know you didn't add it, but just noticed that. Thanks.

Done

spericas (approved):

PR #2939, opened 2021-04-15
WIP - Revise and expand key performance indicators metrics beyond requests counter and meter

### LINKED ISSUE #2689: Expose a metric / healthcheck with the number of inflight requests
(via @olotenko) In order to perform efficient monitoring of microservices, need to expose the number of inflight requests, to be able to determine overload conditions, or stuck requests.

tjquinno:
# Overview
A key performance indicator (KPI) can be useful not only for measuring the performance of a server but also for building application-specific health checks using the KPI.

## Existing ("basic") KPI metrics and implementation

Previously, in any app including `metrics/metrics`, Helidon has published two KPI metrics whenever the app includes the module `metrics/metrics`:

1.  `counter` - `Counter` of the total number of requests received by the server
2. `meter` - `Meter` (rate) for above count.

The full metric names are the routing name for the routing (if any) plus `request.` so typically users see `requests.count` and `requests.meter`.

Recall that a `Meter` reports  an overall count, mean rate, and rates over the previous 1, 5, and 15-minute intervals.

These were created in `MetricsSupport#configureVendorMetrics` and a handler lambda declared in that same method would update the metrics before invoking `request.next()`.

## New ("extended") KPI metrics
The extended metrics are disabled by default, enabled by configuration or builder methods.

1.  `inFlight` - `ConcurrentGauge` of requests actively being _processed_ at a given moment
2. `load` - `Meter` of `inFlight`
3. `longRunning` -`Meter` -  requests which take longer than a configurable threshold to complete their processing
4. `deferred` - `Meter` - received requests were not processed immediately

Some requests (for example but not limited to Jersey) might be _deferred_--received by Helidon but then delayed for some reason (e.g, queued due to a shortage of available threads) before actually being processed. This means there can be a difference between when Helidon _receives_ a request and when it starts _processing_ the request.

Similarly, some requests (again, for example but not limited to Jersey) might complete their work asynchronously, _after_ the user code has invoked `request.next()` and, therefore, separate from when the `MetricsSupport` handler's call to `request.next()` returns.

Some of the new KPI metrics require action _after_ as well as _before_ each request is handled and processed.

These complexities in behavior affect how Helidon should manage the KPI metrics, with the result that the code changes are a little more more extensive and complicated than might be expected at first glance.

# Design
Introduce two abstractions:
* A KPI metrics-related `Context` which tracks the key events in a request's life cycle that are relevant to KPI metrics
* A KPI `Metrics` interface which encapsulates whether the basic or extended KPI metrics are set up

Make some small changes in the current request workflow:
* The existing `microprofile/server` `ServerCdiExtension` class places an instance of the KPI `Context` into the request's context store.
* The existing `webserver/jersey` `JerseySupport` class,  just before it submits the request to Jersey for processing, retrieves the KPI `Context` and, if found, reports to it that request processing is starting.
* The existing `MetricsSupport` class, in its existing KPI-related handler:
    * retrieves the KPI `Context`, or, if there is none present, creates one;
    * obtains the correct impl of the `Metrics` abstraction mentioned above;
    * conveys the `Metrics` object to the KPI `Context`;
    * invokes methods on the KPI `Context` to report the handling and processing of the request.

The basic or extended KPI metrics are registered in the vendor registry when `MetricsSupport` instantiates the `Metrics` abstraction.

# Changed components
## `webserver/webserver` 
* New interface `KeyPerformanceIndicatorSupport` - defines the KPI behavior through nested interfaces
    * `Metrics`- abstracts the creation and collection of original (simple) and new (extended) KPI metrics
    * `Context` - abstracts the stages in a request's life cycle that are relevant to KPI metrics
        * Methods invoked when a request is received, work started, work completed
        * Static factory method
    *  `DeferrableRequestContext` - sub-interface of `Context` 
        * Adds a method invoked when processing actually begins on a deferrable request
        * Static factory method
* New package-private class `KeyPerformanceIndicatorContextImpls` - impls for the above `Context` interfaces.

The specific type of `Context` needed for a request varies from request to request. 
Callers of the `Context` factory methods always know which type they need.

## `webserver/jersey`
* Existing `JerseySupport` class now, just before starting the processing of the request, retrieves the KPI `DeferrableRequestContext` from the request `Context` and invokes `requestProcessingStarted`.

`JerseySupport` always invokes the `DeferrableRequestContext` if it's present. If `metrics/metrics` is not on the class or module path, the `DeferrableRequestContext#requestProcessingStarted` invocation done here acts as a no-op.

## `metrics/metrics`
* New interface `KeyPerformanceIndicatorMetricsSettings` 
   A POJO which holds all the configurable settings related to KPI metrics, currently:
    * whether the extended KPI metrics should be used, and
    * what threshold defines a "long-running" request

    A buildable object which encapsulates the KPI metrics settings (whether extended metrics are enabled and the threshold for long-running requests). The builder is used as a "sub-builder" for `MetricsSupport`.
* `KeyPerformanceIndicatorMetricsSettingsImpl` - impl for above
* `KeyPerformanceIndicatorMetricsImpls`
    * Basic and extended implementations of the KPI `Metrics` abstraction 
    * A factory method for the above
* `MetricsSupport`
    * Enhance `Builder` and constructor to incorporate KPI metrics config settings
    * In `configureVendorMetrics`
        * Gets the correct instance of `Metrics` via the `KeyPerformanceIndicatorMetricsImpls` factory method
        * In the existing metrics-updating handler in `configureVendorMetrics`:
            * Retrieves the KPI `Context` from the request Context or creates a suitable one
            * Invokes `requestHandlingStarted`, `requestProcessingCompleted`, and `requestHandlingCompleted` from appropriate points. This handler cannot invoke `requestProcessingStarted` because the request's processing might be deferred by a handler later in the chain.

## `microprofile/metrics`
Revise and add tests to check the extended KPI metrics.

## `microprofile/server`
`ServerCdiExtension` now adds a new handler to the handler chain. The handler adds a new `DeferrableRequestContext` to each app's handler chain in a position earlier than the handler added by `MetricsSupport`. 

This means:
* The handler added by `MetricsSupport` will find that `Context` and use it, rather than creating and using a non-extended one. 
* `JerseySupport` will also find that `DeferrableRequestContext` and use it to track when the processing of the request begins.

## `examples/metrics/kpi`
 New example app showing config and explicit set-up of extended KPI metrics.

## `docs`
Updates to the SE and MP metrics guides, involving refactoring duplicated text into a new common file.

# Other notes
## Non-Jersey apps with deferred requests 
Users can write their own non-Jersey apps which defer processing of requests. To work correctly with the enhanced KPI support, such apps would need to build the handler chains for the relevant routes as `JerseySupport` does:
* Inserting an early handler which adds a `DeferrableRequestContext` to the request's `Context` (so `MetricsSupport` will use that KPI `Context` when it updates the KPI metrics), and
* During request processing, retrieving that context and invoking its `requestProcessingStarted()` method at the appropriate moment.

Note that if such an app does not take these steps, nothing fails. The KPI metrics will not accurately reflect deferred requests; it will seem as if all requests are processed immediately upon receipt.

## SE apps using Jersey
Although we do not promote it (and coming changes in other technologies might invalidate this use case), developers can currently use `JerseySupport` in SE apps. Such apps can benefit from the KPI metrics improvements. 

To do so, such apps, like `ServerCdiExtension`, must add a handler to each routing that should participate in extended KPI metrics. That handler must add a new `DeferrableRequestContext` to the request's context store. The application must add that handler earlier in the handler chain than it registers the `MetricsSupport` instance as a service.


tjquinno:
Closed because another PR supersedes this one.

PR #2970, opened 2021-04-23
DB Client unwrap support for JDBC and MongoDB internals.

Tomas-Kraus:
Added support to unwrap to internals (java.sql.Connection ,com.mongodb.reactivestreams.client.MongoClient and com.mongodb.reactivestreams.client.MongoDatabase) from DbClient and DbExecute.


dbclient/dbclient/src/main/java/io/helidon/dbclient/DbClient.java
```diff
@@ -63,6 +64,19 @@
      */
     String dbType();
 
+    /**
+     * Unwrap database client internals.
+     * Only database connection is supported. Any operations based on this connection are <b>blocking</b>.
+     * Reactive support must be implemented in user code.
+     * Mongo database driver internals support {@code MongoClient} and {@code MongoDatabase} classes.
```
romain-grecourt, line 71: Probably better to avoid the reference to impls since they are not links.

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbExecute.java
```diff
@@ -354,4 +354,23 @@ default DbStatementDml createNamedDelete(String statementName, String statement)
...
+    /*
+     * Unwrap support
+     */
+
+    /**
+     * Unwrap database executor internals.
+     * Only database connection is supported. Any operations based on this connection are <b>blocking</b>.
+     * Reactive support must be implemented in user code. This connection instance is being used to execute
+     * all statements in current database executor context.
+     * Mongo database driver internals support {@code MongoDatabase} class.
```
romain-grecourt, line 367: Same comment regarding the impl reference.

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbClient.java
Tomas-Kraus: OK, I'm removing it. But API user won't know what classes are supported as unwrap targets now.

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbExecute.java
Tomas-Kraus: Removed impl reference.

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbClient.java
romain-grecourt: In that case you could create a class like `DbClients` and have constants for class names.

romain-grecourt (approved):

Issue #2967, opened 2021-04-22
Integrate Jersey 2.34


tomas-langer:
- [x] Upgrade version
- [x] Exclude MP Config 2.0 from dependencies (REST Client)
- [x] Validate MP REST Client TCKs pass (Jersey implements newer major version)
- [x] Validate native image

spericas:
PR #2971

PR #2971, opened 2021-04-23
Updating Jersey and MP Rest Client versions

spericas:
This PR includes:

- Update to Jersey 2.34 and REST Client API 2.0
- New exclude list for REST Client TCK
- New exclusions in our dependencies pom file due to changes in Jersey
- Passes native integration tests
- Verified a couple of MP native samples by hand as well
- Fixed `microprofile-rest-client-api` dependency


barchetta:
I notice that `microprofile/rest-client` does not have a direct dependency on `microprofile-rest-client-api`. That seems wrong. If you concur could you add that?

barchetta (approved):

barchetta:
See #2967

PR #3000, opened 2021-05-07
Publish internal Jaeger tracing metrics via Helidon's metrics system

### LINKED ISSUE #2947: Provide  integration for for exposing Jaeger client's internal metrics through /metrics endpoint
The Jaeger  client's   have a number of  internal metrics  for   key data points on sampler and reporter statistics.   These include data points such as counts of spans reported,  spans  dropped ,  current queue length of the reporter,  sampling strategy changes in case of remote sampling etc. 

 Details of such metrics  are available at  - https://github.com/jaegertracing/jaeger-client-java/blob/7544006e1aaf42c35d1e9dbc086608f6f6964e56/jaeger-core/src/main/java/io/jaegertracing/internal/metrics/Metrics.java#L138

Some of these metrics are very useful  for  diagnosis  of issues  where we see that traces are loosing spans,  or sampling not working as expected. 

Jaeger client provides  SPI  "io.jaegertracing.spi.MetricsFactory"  that is loaded via the Java Service loader  mechanism.   An implementation has to provide  the mechanism to create  the  metrics such as Counter, Timer & Gauge that is specific to the  platform.   

An example  of such  a implementation already exists for Micrometer and is  available at  
https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-micrometer

We need to have an implementation for Helidon that will work for both SE as well as MP.    

## Environment Details
* Helidon Version:
* Helidon SE or Helidon MP  - Both 
* JDK version:   11
* OS:

tjquinno:
Resolves #2947 

Monitor Jaeger's internal metrics and publish them as Helidon `vendor` metrics.

Creates a new module `metrics/jaeger` which implements a Jaeger metric factory interface; Jaeger instantiates it via a Java service loader.

I added doc pages to `docs/se/tracing` and `docs/mp/tracing`, because users considering Jaeger would be browsing in the tracing topics anyway. (The new pages nearly completely include from a single, common page.)


ljamen (commented):
I noticed one "no newline" error in the pom.xml file but I understand that won't cause an issue. Just FYI.

ljamen (commented):
I noted some minor questions (no edits)>

spericas (approved):

PR #3004, opened 2021-05-08
Webclient query ecoding final

### LINKED ISSUE #2959: Unexpected URI Decoding in WebClient
## Environment Details
* Helidon Version: 2.2.2
* Helidon SE
* JDK version: GraalVM 20.3.0
* OS: Windows 10
* Docker version: N/A

----------

## Problem Description

I am having an issue using WebClient with already encoded query parameters. In my case the query parameters are coming from a server request and my requirement is to send the same parameters using WebClient without any modification.
For example:

final var query = serverRequest.query();
webClient.get()
		.uri(endpointUri + "?" + query)
		.skipUriEncoding()
		.request()

Note that ServerRequest::query returns the raw/encoded query. For the above example I would expect to see WebClient send the query parameters exactly as given. I am seeing that in fact the query parameters are being modified by WebClient. I have traced the cause to WebClientRequestBuilderImpl.java. It has code such as on line 688:

queries = uri.getQuery();

Note that URI::getQuery does not return the raw query but actually a decoded query.

Note also that WebClientRequestBuilderImpl decodes the query in a number of different places, making it a challenge to work around the issue.

I am not seeing any good reason for WebClient to decode the query parameters given by the user in any scenario. The user might provide unencoded query parameters in which case there is no need to decode them. If the user provides encoded query parameters and uses skipUriEncoding then the intention is clearly not to have the parameters decoded before being sent.

It seems like a bug in WebClient. I believe that WebClient should NOT call URI::getQuery that decodes the query parameters, rather it should be using URI::getRawQuery.

## Steps to reproduce

Note that in the original code we were not using skipUriEncoding. In my case I received from the serverRequest something like "a+b" that first got decoded as "a b" and then re-encoded as "a%20b".

In my opinion the API provided by Helidon is fine. It encodes the URI by default and I am now using skipUriEncoding to override the default. My point is that whether it is encoding or not encoding it should never first decode the user provided string.

Verdent:
Fixes #2959

### DIFF
--- a/tests/integration/webclient/src/main/java/io/helidon/tests/integration/webclient/GreetService.java
+++ b/tests/integration/webclient/src/main/java/io/helidon/tests/integration/webclient/GreetService.java
@@ -87,6 +87,8 @@ public class GreetService implements Service {
                 .get("/secure/basic", this::basicAuth)
                 .get("/secure/basic/outbound", this::basicAuthOutbound)
                 .get("/valuesPropagated", this::valuesPropagated)
+                .get("/obtainedQuery", this::obtainedQuery)
+                .get("/pattern with space", this::getDefaultMessageHandler)
                 .put("/greeting", this::updateGreetingHandler);
     }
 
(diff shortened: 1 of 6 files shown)


tomas-langer (approved):
LGTM

PR #3028, opened 2021-05-19
Host IPv6 matching is now cached

### LINKED ISSUE #2870: Optimize RE usage in io.helidon.webclient.Proxy
## Environment Details
* Helidon Version: 2.2.1
* Helidon SE

----------

## Problem Description

A couple of the regular expressions used for Proxy handling by WebClient have been identified as expensive to evaluate. Specifically:

 https://github.com/oracle/helidon/blob/3fa4242eb4e5ab2515af7d7e566f333b24cf2b9e/webclient/webclient/src/main/java/io/helidon/webclient/Proxy.java#L58

https://github.com/oracle/helidon/blob/3fa4242eb4e5ab2515af7d7e566f333b24cf2b9e/webclient/webclient/src/main/java/io/helidon/webclient/Proxy.java#L61

If these are evaluated per request then we should investigate if it is worth optimizing this code (by caching for example).

Verdent:
Fixes #2870

### DIFF
--- a/webclient/webclient/src/main/java/io/helidon/webclient/Proxy.java
+++ b/webclient/webclient/src/main/java/io/helidon/webclient/Proxy.java
@@ -61,8 +61,12 @@ public class Proxy {
     private static final Pattern IP_V6_HEX_HOST = Pattern
             .compile("^((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)$");
 
-    private static final LruCache<String, Boolean> IVP6_HOST_MATCH_RESULTS = LruCache.create();
-    private static final LruCache<String, Boolean> IVP6_IDENTIFIER_MATCH_RESULTS = LruCache.create();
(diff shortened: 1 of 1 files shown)


webclient/webclient/src/main/java/io/helidon/webclient/Proxy.java
```diff
@@ -59,6 +60,10 @@
     private static final Pattern IP_V6_HOST = Pattern.compile("^(?:[0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}$");
     private static final Pattern IP_V6_HEX_HOST = Pattern
             .compile("^((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)::((?:[0-9A-Fa-f]{1,4}(?::[0-9A-Fa-f]{1,4})*)?)$");
+
+    private static final LruCache<String, Boolean> IVP6_HOST_MATCH_RESULTS = LruCache.create();
```
barchetta, line 64: Default capacity for LruCache is 10,000 which seems much larger than is needed?

Verdent: Good point. I will change that.

Verdent: fixed

barchetta (approved):

Issue #3030, opened 2021-05-19
Create a convenience method to send a Clear Cookie directive on ServerResponse


CbxDunderklumpen:
It would be useful to have a easy way to set a HTTP Header which would send a directive to clear an existing cookie.

This can be done by setting a Set-Cookie header where:
* the name and path match the cookie that needs to be cleared.
* the expires and max age attributes are in the past.
* the value should be empty

A convenience method (for example ServerResponse.headers().cookies().clear(....); ) would make this much easier.

## Environment Details
* Helidon Version: 2.2.0
* Helidon SE
* JDK version: 11.0
* OS: (all)
* Docker version (if applicable): n/a

----------

## Problem Description

This is not a problem, but a request for a qualirty of life improvement to the Heldion API

## Steps to reproduce

No repro (see above)

spericas:
PR #3037

PR #3037, opened 2021-05-21
Adds convenience method to clear/invalidate a cookie

spericas:
Adds convenience method to clear/invalidate a cookie. Looks up a cookie by name and sets its expiration in the past.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/HashResponseHeaders.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/HashResponseHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/HashResponseHeaders.java
```diff
@@ -203,6 +203,24 @@ public void addCookie(SetCookie cookie) {
         add(Http.Header.SET_COOKIE, cookie.toString());
     }
 
+    @Override
+    public void clearCookie(String name) {
+        SetCookie expiredCookie = SetCookie.builder(name, "deleted")
+                .path("/")
+                .expires(ZonedDateTime.of(1970, 1, 1, 0, 0, 0, 0, ZoneId.of("GMT+0")))
```
romain-grecourt, line 210: The date instance could be created as a static lazy value

spericas: Sure, don't think this would be a popular method, but we can cache it

romain-grecourt (approved):

spericas:
Issue #3030

Issue #2978, opened 2021-04-27
Third party upgrades


barchetta:
Not urgent, but we should look into upgrading the following:

- [ ] com.h2database:h2
- [ ] org.eclipse.jgit:org.eclipse.jgit
- [ ] org.jboss.narayana.jta:cdi
- [ ] org.jboss:jandex
- [ ] org.jboss:jboss-transaction-spi
- [ ] org.mousio:etcd4j
- [ ] redis.clients:jedis

PR #3040, opened 2021-05-21
Upgrade jgit to 5.11.1.202105131744-r

barchetta:
See #2978

New version of jgit core no longer has dependency on jsch so no longer need to manage its version.

Moved management of org.eclipse.jgit from project pom to `dependencies/pom.xml`. I think this was an oversight.


romain-grecourt (approved):

tjquinno (dismissed):
RG beat me to it, but LGTM!

romain-grecourt (approved):

Issue #3041, opened 2021-05-21
HttpBasicAuthProvider.java Fails To Parse Password Containing a Colon (:)


RichardExley:
## Environment Details
* Helidon Version: 2.2.0
* Helidon MP
* JDK version: 11
* OS: Linux

----------

## Problem Description
Authentication fails if the password contains a colon because the regular expression pattern is incorrect:

`    private static final Pattern CREDENTIAL_PATTERN = Pattern.compile("(.*):(.*)");`

Should be:

`    private static final Pattern CREDENTIAL_PATTERN = Pattern.compile("(.*?):(.*)");`

spericas:
PR #3045

PR #3045, opened 2021-05-24
Use reluctant instead of greedy to match username and password

spericas:
Use reluctant instead of greedy to match username and password. This is necessary to support colons in passwords.


ljnelson (approved):

romain-grecourt (approved):

spericas:
PR #3041

PR #3050, opened 2021-05-26
Reactive routing via cdi provider

### LINKED ISSUE #3031: @RoutingName should be applicable to method
Helidon MP provides a way to define reactive routing with SE api:
 https://helidon.io/docs/v2/#/mp/jaxrs/10_reactive-routing 

By marking `@ApplicationScoped` bean which implements `Service` interface.

But such bean can't be created with:
```java
    @Produces
    @ApplicationScoped
    @RoutingName("@default")
    public CoolFilteringService connectionFactory() {
        return new CoolFilteringService();
    }
```
As `io.helidon.microprofile.server.RoutingName` is not applicable on method.

danielkec:
Fixes #3031

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
@@ -17,10 +17,13 @@
 package io.helidon.microprofile.server;
 
 import java.lang.management.ManagementFactory;
-import java.lang.reflect.Member;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.IdentityHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -41,14 +44,13 @@ import javax.enterprise.context.BeforeDestroyed;
 import javax.enterprise.context.Initialized;
 import javax.enterprise.context.spi.CreationalContext;
 import javax.enterprise.event.Observes;
-import javax.enterprise.inject.spi.AnnotatedMember;
 import javax.enterprise.inject.spi.Bean;
 import javax.enterprise.inject.spi.BeanManager;
 import javax.enterprise.inject.spi.DeploymentException;
 import javax.enterprise.inject.spi.Extension;
-import javax.enterprise.inject.spi.InjectionPoint;
-import javax.enterprise.inject.spi.ProcessProducer;
-import javax.enterprise.inject.spi.Producer;
+import javax.enterprise.inject.spi.ProcessManagedBean;
+import javax.enterprise.inject.spi.ProcessProducerField;
+import javax.enterprise.inject.spi.ProcessProducerMethod;
 
 import io.helidon.common.Prioritized;
 import io.helidon.common.configurable.ServerThreadPoolSupplier;
@@ -100,6 +102,9 @@ public class ServerCdiExtension implements Extension {
     private volatile boolean started;
     private final List<JerseySupport> jerseySupports = new LinkedList<>();
 
+    private final Map<Bean<?>, RoutingConfiguration> serviceBeans
+            = Collections.synchronizedMap(new IdentityHashMap<>());
+
     private void buildTime(@Observes @BuildTimeStart Object event) {
         // update the status of server, as we may have been started without a builder being used
         // such as when cdi.Main or SeContainerInitializer are used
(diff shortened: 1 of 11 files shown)


microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -385,6 +390,33 @@ private void addApplication(JaxRsCdiExtension jaxRs, JaxRsApplication applicatio
...
+
+    private void registerServiceProducers(@Observes ProcessProducer<?, Service> pp) {
+        Producer<Service> producer = pp.getProducer();
+        AnnotatedMember<?> annotatedMember = pp.getAnnotatedMember();
+
+        pp.setProducer(new Producer<>() {
+            @Override
+            public Service produce(final CreationalContext<Service> ctx) {
+                Service service = producer.produce(ctx);
+                producedServices.put(service, annotatedMember);
```
ljnelson, line 403: Conceivably this could be called on any thread (whenever someone injects a `Service`, wherever they do so), so `producedServices` will need to be thread safe.  Also I don't see any matching deletion?

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -416,31 +448,40 @@ private static int priority(Class<?> aClass) {
     private void registerWebServerService(Class<?> serviceClass,
```
ljnelson, line 448: Is this method actually called?  I see that it is referenced in `registerWebServerServices`, but that's `private` and I don't see _it_ being called anywhere.  Maybe I'm missing something?

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
ljnelson: This also relies on `Service` equality to be unchanging.  I don't think `Service` has an `equals`/`hashCode` contract, does it?

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -416,31 +448,40 @@ private static int priority(Class<?> aClass) {
...
                                           Service service) {
 
-        RoutingPath rp = serviceClass.getAnnotation(RoutingPath.class);
-        RoutingName rn = serviceClass.getAnnotation(RoutingName.class);
+        RoutingPath rp;
+        RoutingName rn;
+        Config serviceConfig;
+
+        if (producedServices.containsKey(service)) {
+            // created with producer
```
ljnelson, line 456: One way to test "producerness" without having to drag a map around is to do something like this:
```
final Bean<?> bean = // get a bean
final Class<?> beanClass = bean.getBeanClass(); // returns the class hosting the means of production
if (!bean.getTypes().contains(beanClass)) {
  // it's a producer
} else {
  // it's a managed bean
}
```

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
```diff
@@ -416,31 +448,40 @@ private static int priority(Class<?> aClass) {
...
+            AnnotatedMember<?> producerAnnotatedMember = producedServices.get(service);
+            rp = producerAnnotatedMember.getAnnotation(RoutingPath.class);
+            rn = producerAnnotatedMember.getAnnotation(RoutingName.class);
+            Member member = producerAnnotatedMember.getJavaMember();
+            serviceConfig = config.get(member.getDeclaringClass().getName() + "." + member.getName());
+        } else {
+            // standalone bean
+            rp = serviceClass.getAnnotation(RoutingPath.class);
```
ljnelson, line 464: This isn't guaranteed to work in the presence of portable extensions.  You must always use an `Annotated` to get annotations, since extensions may add or remove extensions.

If you have a "regular" `Class`, you can always "get" an `AnnotatedClass` representing it by using `BeanManager::createAnnotatedType`.

ljnelson (changes requested):

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
danielkec: Thx! Fixed with synchronized IdentityHashMap

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
danielkec: Sure it is, github manages to hide it but ist called from `startServer` method https://github.com/oracle/helidon/blob/320be29cc2c9779f53a10a8de7bd64e61fb6cfc4/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java#L167-L211

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
danielkec: Now its abstracted out the way cdi does the decision https://github.com/oracle/helidon/blob/320be29cc2c9779f53a10a8de7bd64e61fb6cfc4/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java#L132-L147

microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
danielkec: Thx, annotations age grabbed over `getAnnotated` now
https://github.com/oracle/helidon/blob/320be29cc2c9779f53a10a8de7bd64e61fb6cfc4/microprofile/server/src/main/java/io/helidon/microprofile/server/RoutingConfiguration.java#L29-L33

ljnelson (approved):

Issue #3067, opened 2021-06-03
AsciiDoctor error for site generation


NandX00r:
## Environment Details

* JDK version: 16
* OS: macos

## Problem Description
1. I pulled the project src code and build it with maven. it compiles.
2. I want to generate the same documentation like [helidon documentation](https://helidon.io/docs/v1/#/about/01_overview)
3. 
$ mvn site
....
...

[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.8.2:site (default-site) on project helidon-project: Error generating maven-javadoc-plugin:3.2.0:aggregate-no-fork report: 
[ERROR] Exit code: 1 - warning: unknown enum constant Format.MESSAGE_FORMAT
[ERROR]   reason: class file for org.jboss.logging.annotations.Message$Format not found
.....
...

4. what I'm missing, any guidance will be appreciated.

romain-grecourt:
Does it work with JDK11 ?

barchetta:
I have reproduced with JDK 16. As @romain-grecourt says, this should work with JDK 11 where this error was considered non-fatal by javadoc. Looks like we can fix the  Format.MESSAGE_FORMAT errors by adding the following to the aggregated javadoc config in the root pom.xml (not sure why we need to do this, but javadoc is full of mysteries):

```
                            <additionalDependencies>
                                <additionalDependency>
                                    <groupId>org.jboss.logging</groupId>
                                    <artifactId>jboss-logging-annotations</artifactId>
                                    <version>${version.lib.jboss-logging-annotations}</version>
                                </additionalDependency>
                            </additionalDependencies>
```

Once this is fixed we then get other errors related to javadoc formatting which appears to be stricter in JDK 16.

NandX00r:
Thanks @barchetta 

I tried to generate [documentation ](https://helidon.io/docs/v2/#/about/01_overview) and failed.

- MacOs
- JDk11&16:   $ mvn install produce build success
- when I try to execute $ mvn site I get:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-site-plugin:3.8.2:site (default-site) on project helidon-project: Error generating maven-javadoc-plugin:3.2.0:aggregate-no-fork report: 
[ERROR] Exit code: 1 - warning: unknown enum constant Format.MESSAGE_FORMAT
[ERROR]   reason: class file for org.jboss.logging.annotations.Message$Format not found
[...]

PR #3075, opened 2021-06-04
Upgrade javadoc plugin to 3.3.0. Silence javadoc error org.jboss.logging.annotations.Message$Format

barchetta:
* Upgrade javadoc plugin to 3.3.0. 
* Silence javadoc error org.jboss.logging.annotations.Message$Format
* Some javadoc cleanup

See #3067

### DIFF
--- a/service-common/rest-cdi/src/main/java/io/helidon/servicecommon/restcdi/HelidonRestCdiExtension.java
+++ b/service-common/rest-cdi/src/main/java/io/helidon/servicecommon/restcdi/HelidonRestCdiExtension.java
@@ -194,6 +194,12 @@ public abstract class HelidonRestCdiExtension<T extends HelidonRestServiceSuppor
                     .getJavaClass());
     }
 
+    /**
+     * Check if OwnProducerOrNonDefaultQualified.
+     * @param bean to check
+     * @param ownProducerClass producer class
+     * @return true if OwnProducerOrNonDefaultQualified
+     */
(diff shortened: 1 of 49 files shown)


service-common/rest-cdi/src/main/java/io/helidon/servicecommon/restcdi/HelidonRestCdiExtension.java
```diff
@@ -219,6 +225,10 @@ protected void recordProducerMethod(ProcessProducerMethod<?, ?> ppm) {
         recordProducerMember("recordProducerMethod", ppm.getAnnotatedProducerMethod(), ppm.getBean());
     }
 
+    /**
+     * Get producers.
```
tomas-langer, line 229: In general, the verb `get` should not be used in getter javadocs. So in this case it should be something relevant, e.g.

// I am not sure the actual text is correct, just to demonstrate how javadoc should be written
```
Map of bean instances to members producing them.
```

tomas-langer (changes requested):
The content of javadocs is not correct. This should be fixed by authors of original code.
We have three options here:
1. Merge as is
2. Fix as part of this PR with help from others
3. Create follow up issues to fix javadoc content

I would support 2. or 3.

barchetta:
Closing this and replacing it with #3077 that only updates the javadoc-plugin version and configuration. I'll file a new issue for the doclint errors.

Issue #3057, opened 2021-05-28
The "Helidon MP Config guide" provides confusing code samples 


javierdiegof:
## Environment Details
* Helidon Version: 2.3.0
* Helidon MP
* JDK version: OpenJDK 11
* OS: Oracle Linux 8
----------
## Problem Description
It is a minor problem, but I think it makes the guide hard to follow. 
In the **Default configuration** section of the [Helidon MP Config Guide](https://helidon.io/docs/latest/#/mp/guides/03_config) a code snippet with a Main class is provided:
```java
package io.helidon.examples.quickstart.mp;

import io.helidon.microprofile.server.Server;
import java.io.IOException;

public final class Main {

    private Main() { } 

    public static void main(final String[] args) throws IOException {
        Server server = startServer();
        System.out.println("http://localhost:" + server.port() + "/greet");
    }

    static Server startServer() {
        return Server.create().start(); 
    }

}
```
The class is supposed to be added to the archetype template code generated earlier in the tutorial, in the section **Create a sample Helidon MP project**:
``` bash
mvn -U archetype:generate -DinteractiveMode=false \
    -DarchetypeGroupId=io.helidon.archetypes \
    -DarchetypeArtifactId=helidon-quickstart-mp \
    -DarchetypeVersion=2.3.0 \
    -DgroupId=io.helidon.examples \
    -DartifactId=helidon-quickstart-mp \
    -Dpackage=io.helidon.examples.quickstart.mp
```
That archetype project does not provide a `Main.java` file, just a `GreetingProvider.java` and a `GreetResource.java`. That means that something besides those two classes is providing the 'main entry point'. Looking at the `MANIFEST.MF` of the jar file, it appears that the main class is being provided by the Helidon framework:
```
Main-Class: io.helidon.microprofile.cdi.Main
```
Which is OK. The problem is that adding the `Main.java` file in the tutorial does not change that, the `io.helidon.microprofile.cdi.Main` is still the entry point. In practical terms, this means that the `Main.java` file that the user is writing as he/she goes through the tutorial is being completely ignored when packaging and running the project, which makes the tutorial very hard to follow since the code does not do what it is supposed to do.
## Steps to reproduce
1. Create the project using archetype in the code snippet above. 
2. Add the `Main.java` to the project according to the snippet above.
3. Package the application: `mvn package`.
4. Run the packaged application: `java -jar target target/java_jar_file.jar`.
5. See how the code provided in the `Main.java` file is ignored.

[...]

PR #3083, opened 2021-06-05
Update Config Guide to describe how to change pom for custom main class

barchetta:
See #3057

### DIFF
--- a/docs/mp/guides/03_config.adoc
+++ b/docs/mp/guides/03_config.adoc
@@ -85,8 +85,10 @@ For example, if you specify a custom server port in `META-INF/microprofile-confi
 then your server will listen on that port.
 
 A main class is also required to start up the server and run the
-application. If you don't use Helidon's built-in main class you can
-define your own:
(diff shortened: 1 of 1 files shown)


docs/mp/guides/03_config.adoc
```diff
@@ -85,8 +85,10 @@ For example, if you specify a custom server port in `META-INF/microprofile-confi
 then your server will listen on that port.
 
 A main class is also required to start up the server and run the
-application. If you don't use Helidon's built-in main class you can
-define your own:
+application. By default the Quickstart sample project uses the built-in
+Helidon main. In this guide you want to use your own main so you can have
```
tjquinno, line 89: Minor point. If we say "main class," then using normal, lower-case "main" seems fine. But if we use "main" by itself--as in "uses the built-in Helidon main" or "use your own main," I feel as if it should be `Main` (capitalized and treated as code).

barchetta: Fixed.

tjquinno (approved):

Issue #3008, opened 2021-05-11
JPA Enhancements


ljnelson:
This epic is to track various JPA-related fixes and improvements.  The sub-tasks in this epic are in no particular order and are subject to change, deletion or augmentation without notice.

* [ ] Inflate container-managed `EntityManagerFactory` client proxies early so that such inflation does not occur during a JTA-managed transaction
* [ ] Maven plugin for auto-including entity classes in a `persistence.xml`
* [ ] Look at doing our own Eclipselink weaving plugin so we can avoid depending on [the one we're using](https://github.com/ethlo/eclipselink-maven-plugin) at the moment
* [ ] Make it easy to make `PersistenceUnitInfo` objects (they're already picked up if present as CDI beans but this is sort of a hidden feature at the moment)

ljnelson:
Closed per request

PR #3089, opened 2021-06-09
Adds integrations for pseudo-XA datasources

ljnelson:


ljnelson:
In general, this PR moves some package-private classes out of the JPA integration area and into their own projects, since these constructs have nothing to do with JPA.  Some of these constructs look to be extremely useful for unit testing, which is the focus of #3008 with which this PR is associated.

I haven't changed any logic along the way (although obviously look out for boneheaded mistakes).  I've taken advantage of the making-public by adding documentation and removing crufty methods and such.

PR #3105, opened 2021-06-11
Honor bind address and host in configuration

### LINKED ISSUE #3103: `host` vs `bind-address` configuration issues
## Environment Details
* Helidon Version: 2.3.0
* Helidon SE and Helidon MP

----------

## Problem Description

While investigating a questions from a user I've uncovered issues with configuring the bind-address for the default socket.

1. In Helidon SE the `host` configuration property for the default socket appears to be ignored. We use this property in many of our examples including the quickstarts. It looks like these should be changed to `bind-address`.  For named sockets it looks like we do always use `bind-address` (but see related issue #3104).
2. The above is also true for Helidon MP examples where `server.host` is usually ignored but `server.bind-address` works. I think this is because the MP configuration is converted to SE config and passed to WebServer and it hits 1 above.
3. But beware in MP Server because there is code that assumes `server.host` when the Server builder is explicitly used : https://github.com/oracle/helidon/blob/69812d43c53e4f777cf872c176a4a0221ee83366/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java#L259

## Steps to reproduce
1. Generate the Helidon SE quickstart application. Build and run it. It works!
2. Change the `host` attribute to your local IP address like:
    ```
    server:
      port: 8080
      host: 192.1.2.3
    ```
    Build and run the app and you see it still binds to the default address (0.0.0.0).
3. Change `host:` to `bind-address:` and build and run again and voila! It binds to the IP address as it should.
4. You can do something similar with the Helidon MP quickstart.

### LINKED ISSUE #3104: Named sockets ignore `bind-address`
## Environment Details
* Helidon Version: 2.3.0
* Helidon SE and Helidon MP

----------

## Problem Description
If you configure your Helidon application with named sockets on different IP address, it appears as though the `bind-address` property on the name socket is ignored and the `bind-address` property on the default socket is used instead.

## Steps to reproduce
1. Build `examples/webserver/mutliport` and run it. This defines multiple sockets on different ports, but all bound to `localhost`. It works.
2. In `application.yaml` change the `bind-address` for socket `private` from localhost to your local IP. Build and run and when the server starts you'll see the private channel still binds to 0.0.0.0. 
3. In `application.yaml` change the `private` bind-address back to localhost. 
3. In `application.yaml` change the default server `host: "localhost"` to `bind-address: <your local IP>` (yes you need to change `host` to `bind-address` see issue #3103 ).
4. Now when you build and run you'll see *all* sockets bind to the IP

So the `bind-address` configuration on the named sockets is ignored and the one on the default server is always used.

Looks like the problem is here:
https://github.com/oracle/helidon/blob/69812d43c53e4f777cf872c176a4a0221ee83366/webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java#L245

This code is executed for each named socket. The `port` is correctly pulled from the named socket configuration, but the bindAddress is always pulled from the server configuration.

tomas-langer:
Resolves #3104 
Resolves #3103 

### DIFF
--- a/examples/microprofile/multiport/src/main/resources/application.yaml
+++ b/examples/microprofile/multiport/src/main/resources/application.yaml
@@ -19,9 +19,9 @@ server:
   sockets:
     - name: "private"
       port: 8081
-      bind-address: "localhost"
     - name: "admin"
       port: 8082
+      # bind address is optional, if not defined, server host will be used)
       bind-address: "localhost"
 
 # Metrics and health run on admin port
--- a/webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java
@@ -17,6 +17,7 @@
 package io.helidon.webserver;
 
 import java.net.BindException;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.HashMap;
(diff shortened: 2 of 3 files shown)


spericas (approved):
LGTM

barchetta (approved):

PR #3108, opened 2021-06-14
Introduce backpressure to webserver

### LINKED ISSUE #3092: Memory leak in the webserver
## Environment Details
* Helidon Version: 2.2.1
* Helidon SE or Helidon MP:  Helidon SE
* JDK version: 15
* OS: mac 11.3.1
* Docker version (if applicable): 20.10.6, build 370c289

----------

## Problem Description
I use the following libraries 
import  io.helidon.webserver.*;
import io.helidon.common.http.DataChunk;
import io.helidon.common.reactive.IoMulti;
import io.helidon.common.reactive.Multi;
to send a large file over http, and got a java.lang.OutOfMemoryError: Java heap space error:

        | java.lang.OutOfMemoryError: Java heap space
        |        at java.base/java.io.InputStream.readNBytes(Unknown Source)
        |        at io.helidon.common.reactive.MultiFromInputStream$InputStreamSubscription.submit(MultiFromInputStream.java:97)
        |        at io.helidon.common.reactive.MultiFromInputStream$InputStreamSubscription.trySubmit(MultiFromInputStream.java:146)
        |        at io.helidon.common.reactive.MultiFromInputStream$InputStreamSubscription.request(MultiFromInputStream.java:142)
        |        at io.helidon.common.reactive.MultiMapperPublisher$MapperSubscriber.request(MultiMapperPublisher.java:104)
        |        at io.helidon.common.reactive.MultiTappedPublisher$MultiTappedSubscriber.request(MultiTappedPublisher.java:276)
        |        at io.helidon.webserver.BareResponseImpl.onSubscribe(BareResponseImpl.java:328)
        |        at io.helidon.media.common.MessageBodyContext$EventingSubscriber.onSubscribe(MessageBodyContext.java:324)
        |        at io.helidon.webserver.accesslog.SizeLogEntry$ByteCountingSubscriber.onSubscribe(SizeLogEntry.java:112)
        |        at io.helidon.common.reactive.MultiTappedPublisher$MultiTappedSubscriber.onSubscribe(MultiTappedPublisher.java:214)
        |        at io.helidon.common.reactive.MultiMapperPublisher$MapperSubscriber.onSubscribe(MultiMapperPublisher.java:62)
        |        at io.helidon.common.reactive.MultiFromInputStream.subscribe(MultiFromInputStream.java:51)
        |        at io.helidon.common.reactive.MultiMapperPublisher.subscribe(MultiMapperPublisher.java:42)
        |        at io.helidon.common.reactive.MultiTappedPublisher.subscribe(MultiTappedPublisher.java:86)
        |        at io.helidon.webserver.accesslog.SizeLogEntry$ByteCountingPublisher.subscribe(SizeLogEntry.java:96)
        |        at io.helidon.media.common.MessageBodyContext$EventingPublisher.subscribe(MessageBodyContext.java:293)
        |        at io.helidon.webserver.Response.lambda$send$2(Response.java:206)
        |        at io.helidon.webserver.Response$$Lambda$722/0x0000000100373ab8.run(Unknown Source)
        |        at io.helidon.webserver.Response$SendLockSupport.execute(Response.java:375)
        |        at io.helidon.webserver.Response.send(Response.java:204)
        |        at io.helidon.webserver.Response.send(Response.java:192)

## Steps to reproduce

code snippet for sending a large file over http:
BufferedInputStream catStdOut = new BufferedInputStream(process.getInputStream());
```
Multi<DataChunk> ioMulti = IoMulti.multiFromStreamBuilder(catStdOut).byteBufferSize(8 * 1024 * 1024).build()
                                .map(byteBuffer -> DataChunk.create(true, byteBuffer))
​
                                .onTerminate(() -> {
                                    CompletableFuture.runAsync(() -> {
                                       
                                       ...          
                                    });
                                });
​
                        res.status(200);

                        res.send(ioMulti)
                                .onTerminate(() -> {
                                    try {
                                        catStdOut.close();
                                    } catch (IOException e) {
                   ...
                                    }
                                });
```

danielkec:
There are basically 2 possibilities how to solve the problem with upstream being faster than the Netty event loop. 
Both solutions are introducing back-pressure.

* Request one-by-one - dead simple solution which can be less performant with publishers designed to produce data in batches(file reading, db access), that can be compensated with buffering on the publisher side on as needed basis(reactive operator for buffering)
* Requesting by batches depending on the state of newly introduced webserver buffer - this solution would make webserver responsible for additional buffering, solution woul have to be highly configurable for use-cases where buffering is not desirable

This PR fixes #3092 by requesting one-by-one item.

Introduction of the symmetrical back-pressure is expected to bring some small performance penalty:
[JMH comparison](https://github.com/danielkec/helidon-jmh/tree/webserver-backpressure/src/main/java/io/helidon/reactive/jmh/webserver)
```
# 1M100x means a stream of 100 DataChunks backed by 1MB ByteBuffer

Benchmark                Mode  Cnt    Score    Error  Units
OneByOne1M100x.test     thrpt    5    0.069 ±  0.002  ops/s
OneByOne1k10000x.test   thrpt    5    5.179 ±  0.224  ops/s
OneByOne1k100x.test     thrpt    5  502.839 ± 19.513  ops/s
Unbounded1M100x.test    thrpt    5    0.066 ±  0.002  ops/s
Unbounded1k10000x.test  thrpt    5    5.137 ±  0.346  ops/s
Unbounded1k100x.test    thrpt    5  494.400 ± 57.133  ops/s

  Unbounded1M100x.test  0.066
   OneByOne1M100x.test  0.069
Unbounded1k10000x.test █ 5.137
 OneByOne1k10000x.test █ 5.179
  Unbounded1k100x.test ██████████████████████████████████████████████████████████ 494.400
   OneByOne1k100x.test ████████████████████████████████████████████████████████████ 502.839
```
Compensation is possible by buffering on the publisher's size, this could be achieved by introducing buffering reactive operator.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
@@ -334,7 +334,7 @@ class BareResponseImpl implements BareResponse {
     @Override
     public void onSubscribe(Flow.Subscription subscription) {
         this.subscription = subscription;
-        subscription.request(Long.MAX_VALUE);
+        subscription.request(2); // Because first chunk gets cached
     }
 
     @Override
@@ -349,6 +349,7 @@ class BareResponseImpl implements BareResponse {
                 } else {
                    prevRequestChunk = prevRequestChunk.thenRun(ctx::flush);
                 }
+                subscription.request(1);
                 return;
             }
 
@@ -431,6 +432,7 @@ class BareResponseImpl implements BareResponse {
             channelFuture = ctx.writeAndFlush(httpContent);
         } else {
             channelFuture = ctx.write(httpContent);
+            subscription.request(1);
         }
 
         return channelFuture
@@ -444,6 +446,9 @@ class BareResponseImpl implements BareResponse {
                         }
                     });
                     data.release();
+                    if (data.flush()) {
+                        subscription.request(1);
+                    }
                     LOGGER.finest(() -> log("Data chunk sent with result: %s", future.isSuccess()));
                 })
                 .addListener(completeOnFailureListener("Failure when sending a content!"))
--- a/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
@@ -334,12 +334,13 @@ class BareResponseImpl implements BareResponse {
     @Override
     public void onSubscribe(Flow.Subscription subscription) {
         this.subscription = subscription;
-        subscription.request(2); // Because first chunk gets cached
+        subscription.request(1);
     }
 
     @Override
     public void onNext(DataChunk data) {
         if (internallyClosed.get()) {
+            subscription.request(1);
             throw new IllegalStateException("Response is already closed!");
         }
         if (data != null) {
@@ -355,6 +356,7 @@ class BareResponseImpl implements BareResponse {
 
             if (lengthOptimization && firstChunk == null) {
                 firstChunk = data.isReadOnly() ? data : data.duplicate();      // cache first chunk
+                subscription.request(1);
                 return;
             }
 
@@ -445,8 +447,9 @@ class BareResponseImpl implements BareResponse {
                             writeFuture.completeExceptionally(future.cause());
                         }
                     });
+                    boolean flush = data.flush();
                     data.release();
-                    if (data.flush()) {
+                    if (flush) {
                         subscription.request(1);
                     }
                     LOGGER.finest(() -> log("Data chunk sent with result: %s", future.isSuccess()));
--- a/webserver/webserver/src/test/java/io/helidon/webserver/BackpressureTest.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/BackpressureTest.java
@@ -49,13 +49,14 @@ import static org.hamcrest.Matchers.is;
 public class BackpressureTest {
 
     private static final Logger LOGGER = Logger.getLogger(BackpressureTest.class.getName());
-    static final long TIMEOUT_SEC = 30;
+    static final long TIMEOUT_SEC = 40;
+    // 5 MB buffer size should be enough to cause incomplete write in Netty
+    static final int BUFFER_SIZE = 5 * 1024 * 1024;
 
     @Test
     void overloadEventLoopWithIoMulti() {
         Multi<DataChunk> pub = IoMulti.multiFromStreamBuilder(randomEndlessIs())
-                // 10 MB buffer size should be enough to cause incomplete write in Netty 
-                .byteBufferSize(10 * 1024 * 1024)
+                .byteBufferSize(BUFFER_SIZE)
                 .build()
                 .map(byteBuffer -> DataChunk.create(true, byteBuffer));
         overloadEventLoop(pub);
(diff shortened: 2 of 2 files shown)


spericas:
@danielkec We should probably create another issue to run the perf tests after this change.

danielkec:
I have ran TechEmpower benchmark locally to compare snapshot version with the change against the 2.2.1

**2.2.1**
image

**2.3.1-SNAPSHOT** - with this PR
image

spericas (dismissed):
LGTM

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -444,6 +446,9 @@ private ChannelFuture sendData(DataChunk data) {
                         }
                     });
                     data.release();
+                    if (data.flush()) {
```
olotenko, line 449: I think the order of `data.release` and `data.flush` may be important?

I didn't look into the ownership of `data`, but typically you are not allowed to access pooled objects after you declared you don't need them (after `data.release()`).

olotenko (changes requested):
I am not too sure what this change achieves.

Ie how is the backpressure introduced here. The old version effectively said: "push as much data as you want, we'll push it to `channel` as soon as we can". The proposed version effectively says: "push at most two items, and we'll tell you when to push more, and we do that as soon as we pushed it to `channel` as soon as we could".

I would understand if `subscription.request(1)` were invoked after `channel.write` completed - that is definitely allowing more data to arrive after it has been committed to the channel buffers - presumably, Netty won't complete the related Future, if write buffers are full.

Another thing that makes sense - count the bytes that go through here, and allow up to a given amount to be buffered. Eg `onNext` counts how many bytes were in the buffers it has seen, and requests one more chunk if it has seen below watermark. Then `channel.write` completion Future subtracts the bytes it has just written. If this results in crossing the "watermark", then request 1. (because `onNext` stopped requesting more)

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
danielkec: Its a final flag set during construction of DataChunk

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -334,7 +334,7 @@ private void writeLastContent(final Throwable throwable, final ChannelFutureList
     @Override
     public void onSubscribe(Flow.Subscription subscription) {
         this.subscription = subscription;
-        subscription.request(Long.MAX_VALUE);
+        subscription.request(2); // Because first chunk gets cached
```
olotenko, line 337: This is an incorrect assumption. I am afraid it is not correct in several cases, perhaps even in the case you are optimising for.

The first chunk gets cached conditionally. A safer approach is to request 1, and ensure 1 is requested for every invocation of `onNext` (ie guard all exits from that routine, except when deferred to `socket.write`).

olotenko (changes requested):

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
danielkec: fixed

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
danielkec: fixed

danielkec:
We trying to solve the issue #3092 with this PR eg. sending long streams of huge immediately flushed chunks.

Watermarking for not-immediately flushed chunks seems pretty cool and would make webserver truly backpressure aware, lets move it to #3136

olotenko (approved):
`try...finally` is a safer pattern for "always request in `onNext`, unless explicitly delegated to `orderedWrite`", but ok, we can live with this implementation.

Thank you for taking care of my comments. Let's implement better backpressure as a separate ticket that you created.

spericas (approved):

PR #3111, opened 2021-06-15
Honor binding address (host) when binding multiple ports.

### LINKED ISSUE #3104: Named sockets ignore `bind-address`

tomas-langer:
Resolves #3104 

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerConfiguration.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -549,6 +549,10 @@ public interface ServerConfiguration extends SocketConfiguration {
         private SocketConfiguration.Builder configureSocket(Config config, SocketConfiguration.Builder soConfigBuilder) {
 
             config.get("port").asInt().ifPresent(soConfigBuilder::port);
+            // we use `host` in every MP example and it should be supported in SE as well
+            config.get("host").asString()
+                    .map(this::string2InetAddress)
+                    .ifPresent(soConfigBuilder::bindAddress);
             config.get("bind-address")
                     .asString()
                     .map(this::string2InetAddress)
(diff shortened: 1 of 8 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/ServerConfiguration.java
```diff
@@ -549,6 +549,10 @@ public Builder config(Config config) {
         private SocketConfiguration.Builder configureSocket(Config config, SocketConfiguration.Builder soConfigBuilder) {
 
             config.get("port").asInt().ifPresent(soConfigBuilder::port);
+            // we use `host` in every MP example and it should be supported in SE as well
+            config.get("host").asString()
+                    .map(this::string2InetAddress)
+                    .ifPresent(soConfigBuilder::bindAddress);
             config.get("bind-address")
                     .asString()
                     .map(this::string2InetAddress)
```
danielkec, line 558: Maybe format `.asString()` to new line so its consistent with to bind-address fetching

danielkec (approved):

webserver/webserver/src/main/java/io/helidon/webserver/ServerConfiguration.java
tomas-langer: Fixed

danielkec (approved):

barchetta (approved):

PR #3127, opened 2021-06-17
Crypto common module integrated into config and vault

### LINKED ISSUE #3094: Change Helidon modules with crypto operations
Change following modules to use new common crypto module instead of their own cryptography implementation.

- [x] Encryption config
- [x] Vault config

Verdent:
Fixes #3094


tomas-langer (dismissed):
Tested OCI Vault with native image. LGTM

tomas-langer (approved):

PR #3134, opened 2021-06-18
Improved handling of Jaeger spans/scopes across threads

spericas:
Improved handling of Jaeger spans/scopes across threads. Fixes problems with MP async calls that resulted in memory leaks of `ThreadLocalScope` objects due to the failed attempt to close a scope from a different thread. 

New implementation uses a `DataPropagationProvider` to close a scope on the primary thread and re-open it on the secondary thread --and eventually close it there as well. Minor change to the `DataPropagationProvider` spi.


tomas-langer (commented):
A backward incompatible change. 
This feature was introduces quite recently, yet I am not sure we can simply modify the SPI. Could we achieve the same with backward compatible default methods that would be deprecated?

tomas-langer (approved):
LGTM

PR #3156, opened 2021-06-25
Audit does not format message if not loggable.

### LINKED ISSUE #3143: DefaultAuditProvider spends significant time on formatting the message even for cases for which events are not logged
Problem Description
--------------------
The helidon's default AuditProvider is used when no custom  audit provider is used.   Based on what we are observing  this seems to be  common case.  

In the implementation of  [DefaultAuditProvider.audit()](https://github.com/oracle/helidon/blob/master/security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java#L47)  doesn't check the logger level  prior to invoking the logEvent() methods.   For example events for SUCCESS, INFO etc are logged  at FINEST level.   In most cases production environments will not have  FINEST logging level.  

 The code ends up spending significant time formatting complex and large messages which eventually do not get logged.  In reviewing theJFR  from one of the services we do see  that   over 50% of the time spent on the Forkjoin common  pool threads is on the below stack

java.util.Arrays.copyOf(byte[], int)	                             
java.lang.AbstractStringBuilder.ensureCapacityInternal(int)	
java.lang.AbstractStringBuilder.append(String)	
java.lang.StringBuilder.append(String)	
javax.security.auth.Subject.toString()	
io.helidon.security.Subject.toString()	
java.util.Formatter$FormatSpecifier.printString(Object, Locale)	
java.util.Formatter$FormatSpecifier.print(Object, Locale)	
java.util.Formatter.format(Locale, String, Object[])	
java.util.Formatter.format(String, Object[])	
java.lang.String.format(String, Object[])	
java.util.Optional.toString()	
java.util.Formatter$FormatSpecifier.printString(Object, Locale)	
java.util.Formatter$FormatSpecifier.print(Object, Locale)	
java.util.Formatter.format(Locale, String, Object[])	
java.util.Formatter.format(String, Object[])	
java.lang.String.format(String, Object[])	
io.helidon.security.DefaultAuditProvider.formatMessage(AuditEvent)	
io.helidon.security.DefaultAuditProvider.logEvent(String, AuditProvider$TracedAuditEvent, Level)	
io.helidon.security.DefaultAuditProvider.audit(AuditProvider$TracedAuditEvent)	
io.helidon.security.DefaultAuditProvider$$Lambda$1291.1693004675.accept(Object)	
io.helidon.security.Security.audit(String, AuditEvent)	
io.helidon.security.SecurityContextImpl.audit(AuditEvent)	
io.helidon.security.AuthorizationClientImpl.lambda$submit$0(AuthorizationProvider, AuthorizationResponse
io.helidon.security.AuthorizationClientImpl$$Lambda$2382.1521780927.apply(Object)	
java.util.concurrent.CompletableFuture$UniApply.tryFire(int)	
java.util.concurrent.CompletableFuture.postComplete()	
java.util.concurrent.CompletableFuture$AsyncSupply.run()	
java.util.concurrent.CompletableFuture$AsyncSupply.exec()	
java.util.concurrent.ForkJoinTask.doExec()
java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinTask, ForkJoinPool$WorkQueue, int)
java.util.concurrent.ForkJoinPool.scan(ForkJoinPool$WorkQueue, int)	
java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool$WorkQueue)	
java.util.concurrent.ForkJoinWorkerThread.run()

  The ask here is to  have check based on the logger's level and avoid processing associated with logEvent if the event is not going to be  logged based on the log levels.   

## Environment Details
* Helidon Version:   2.2.0
* Helidon  MO 
* JDK version:  11
* OS:   Linux  x64

----------

tomas-langer:
Resolves #3143 

Configurable log levels.

Fixed configuration for audit - the configuration node is now just `audit`, so we can configure it as follows:

```
security:
  audit:
    defaultProvider:
      logger: "helidon.audit"
      level:
        failure: "INFO"
```

This is the intended way of configuring default audit provider (before you would need to do `security.security.audit`.

Added support for configuring the log levels for various audit severities (as seen above in example).
Audit severities (and their default log levels):

- failure (`FINEST`)
- success (`FINEST`)
- info (`FINEST`)
- warn (`WARNING`)
- error (`SEVERE`)
- audit-failure (`SEVERE`)

### DIFF
--- a/security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
+++ b/security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
@@ -40,6 +40,7 @@ final class DefaultAuditProvider implements AuditProvider {
     private DefaultAuditProvider(Config config) {
         // config node is already located on the security node
         this.auditLogger = Logger.getLogger(config.get("audit.defaultProvider.logger").asString().orElse("AUDIT"));
+
         this.failureLevel = level(config, "failure", Level.FINEST);
         this.successLevel = level(config, "success", Level.FINEST);
         this.infoLevel = level(config, "info", Level.FINEST);
@@ -49,7 +50,7 @@ final class DefaultAuditProvider implements AuditProvider {
     }
 
     private Level level(Config config, String auditSeverity, Level defaultLevel) {
-        return config.get("audit.defaultProvider.level.failure")
+        return config.get("audit.defaultProvider.level." + auditSeverity)
                 .asString()
                 .map(Level::parse)
                 .orElse(defaultLevel);
@@ -65,7 +66,6 @@ final class DefaultAuditProvider implements AuditProvider {
     }
 
     private void audit(TracedAuditEvent event) {
-        String tracingId = event.tracingId();
         Level level;
 
         switch (event.severity()) {
@@ -91,15 +91,16 @@ final class DefaultAuditProvider implements AuditProvider {
             break;
         }
 
-        logEvent(tracingId, event, level);
+        logEvent(event, level);
     }
 
-    private void logEvent(String tracingId, TracedAuditEvent event, Level level) {
+    private void logEvent(TracedAuditEvent event, Level level) {
         if (!auditLogger.isLoggable(level)) {
             // no need to create the message when the message would not be logged
             return;
         }
 
+        String tracingId = event.tracingId();
         AuditSource auditSource = event.auditSource();
 
         StringBuilder locationInfo = new StringBuilder();
--- a/security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
+++ b/security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
```diff
@@ -30,13 +30,33 @@
...
+        this.infoLevel = level(config, "info", Level.FINEST);
+        this.warnLevel = level(config, "warn", Level.WARNING);
+        this.errorLevel = level(config, "error", Level.SEVERE);
+        this.auditFailureLevel = level(config, "audit-failure", Level.SEVERE);
+    }
 
-    private DefaultAuditProvider(String loggerName) {
-        this.auditLogger = Logger.getLogger(loggerName);
+    private Level level(Config config, String auditSeverity, Level defaultLevel) {
+        return config.get("audit.defaultProvider.level.failure")
```
barchetta, line 52: Should this be `"audit.defaultProvider.level." + auditSeverity` ?

security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
```diff
@@ -46,30 +66,40 @@ public static DefaultAuditProvider create(Config config) {
 
     private void audit(TracedAuditEvent event) {
         String tracingId = event.tracingId();
+        Level level;
```
ljnelson, line 69: (No need to allocate the `tracingId` variable.)

security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
```diff
@@ -46,30 +66,40 @@ public static DefaultAuditProvider create(Config config) {
...
+            level = errorLevel;
             break;
         case AUDIT_FAILURE:
         default:
             //audit failure - something wrong with auditing mechanism...
-            logEvent(tracingId, event, Level.SEVERE);
+            level = auditFailureLevel;
             break;
         }
+
```
ljnelson, line 93: Assuming you remove the allocation of the `tracingId` variable above, then you can perform the `if (!auditLogger.isLoggable(level))` check here instead of (or in addition to) in the `logEvent` method.

security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
tomas-langer: Fixed

security/security/src/main/java/io/helidon/security/DefaultAuditProvider.java
tomas-langer: Moved to `logEvent` after checking if loggable.

barchetta (approved):

PR #3157, opened 2021-06-25
Health check for OCI ObjectStorage that works in SE and MP

spericas:
This is WIP.

Liveness health check for OCI ObjectStorage that works in SE and MP, like our other built-in health checks. Verifies that a bucket is accessible (HTTP GET) using the OCI client API. In addition to the user-specific OCI credentials, it needs the namespace and bucket name. MP and SE examples have been updated to use the new health check.

Current output for health check is shown below: `GET http://.../health/live

```
{
  "outcome": "UP",
  "status": "UP",
  "checks": [
    {
      "name": "objectStorage",
      "state": "UP",
      "status": "UP",
      "data": {
        "bucket": "main",
        "namespace": "idv5cxr1uqqk"
      }
    }
  ]
}
```


integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+    private static final int DEFAULT_TIMEOUT_SECONDS = 10;
+
+    private String bucket;
+    private String namespace;
+    private int timeout;
+    private OciObjectStorageRx ociObjectStorage;
+
+    private OciObjectStorageHealthCheck() {
+        this.timeout = DEFAULT_TIMEOUT_SECONDS;
+        Config ociConfig = Config.create().get("oci");
```
tomas-langer, line 56: I guess this is intended to be used by CDI?
If that is the case, you can just use `@Inject` on the constructor, and use `@ConfigProperty` to inject all configuration and get rid of the system properties.
Using `Config.create()` outside of user code is not recommended and may produce unexpected results.

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+        this.ociObjectStorage = OciObjectStorageRx.create(ociConfig);
+    }
+
+    private OciObjectStorageHealthCheck(Builder builder) {
+        this.timeout = builder.timeout;
+        this.bucket = builder.bucket;
+        this.namespace = builder.namespace;
+        this.ociObjectStorage = builder.ociObjectStorage;
+
+        Config ociConfig = builder.config;
```
tomas-langer, line 69: This should be done in the builder, not here.

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Create an instance.
+     *
+     * @return an instance.
+     */
+    public static OciObjectStorageHealthCheck create() {
```
tomas-langer, line 121: This method should not exist, as there are some required properties.
method `create(Config config)` would make sense here.

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * Fluent API builder for {@link OciObjectStorageHealthCheck}.
+     */
+    public static final class Builder implements io.helidon.common.Builder<OciObjectStorageHealthCheck> {
+
+        private Config config;
+        private String bucket;
+        private String namespace;
+        private int timeout = -1;
```
tomas-langer, line 165: The timeout should be kept as Duration.

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        /**
+         * Set up this builder using config.
+         *
+         * @param config the config.
+         * @return the builder.
+         */
+        public Builder config(Config config) {
+            this.config = config;
```
tomas-langer, line 183: do not store config instance, read everything in this method

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+            Config timeout = config.get("objectstorage.healthcheck.timeout");
```
tomas-langer, line 184: config key for durations should contain the unit - such as
`timeout-millis` `timeout-seconds` so it is clear when setting it in a configuration file.

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+            return this;
+        }
+
+        /**
+         * Set timeout in millis.
+         *
+         * @param timeout timeout in millis.
+         * @return the builder.
+         */
+        public Builder timeout(int timeout) {
```
tomas-langer, line 195: Public builder method should be
```
public Builder timeout(Duration timeout) {
            this.timeout = timeout;
            return this;
}
```
A helper private method can be created that accepts configured value, such as `timeoutMillis(long)` or `timeoutSeconds(long)` that will call this method:
```
private timeoutSeconds(long seconds) {
  timeout(Duration.ofSeconds(seconds));
}
```

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
tomas-langer: bucket, namespace etc. should be read here. The builder should throw an exception if either is undefined.

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageRxImpl.java
```diff
@@ -84,6 +85,18 @@
         return restApi.post(apiPath, request, RenameObject.Response.builder());
     }
 
+    @Override
+    public Single<ApiOptionalResponse<GetObjectRx.Response>> getBucket(GetObject.Request request) {
```
tomas-langer, line 89: getBucket should probably not have `GetObject.Request` request, as that is for a streaming response. Bucket is probably a different type. If you want to have `getBucket`, please create a new type `GetBucket` with `Request` and `Response` inner classes that have proper fields.

tomas-langer: see https://docs.oracle.com/en-us/iaas/api/#/en/objectstorage/20160918/Bucket/GetBucket

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,233 @@
...
+import static io.helidon.common.http.Http.Status.OK_200;
+
+/**
+ * Liveness check for an OCI's ObjectStorage bucket. Reads configuration to
+ * obtain bucket name as well as OCI properties from '~/.oci/config'.
+ */
+@Liveness
+@ApplicationScoped // this will be ignored if not within CDI
+@BuiltInHealthCheck
+public final class OciObjectStorageHealthCheck implements HealthCheck {
```
tomas-langer, line 44: If this is part of the object storage module and it is run in CDI, this bean will not work unless correctly configured.
What if the user does not want to have this health check at all? 
I think this should either be in a separate module, or not a bean (so the user would have to create a bean of their own and use the health check as a delegate).

integrations/oci/objectstorage/src/main/java/io/helidon/integrations/oci/objectstorage/OciObjectStorageImpl.java
```diff
@@ -52,4 +52,10 @@
         return delegate.renameObject(request)
                 .await();
     }
+
+    @Override
+    public ApiOptionalResponse<GetObject.Response> getBucket(GetObject.Request request) {
```
tomas-langer, line 57: The request and response types are wrong (link in a different comment regarding the same).

tomas-langer (changes requested):

examples/integrations/vault/hcp-reactive/src/main/resources/application.yaml
```diff
@@ -50,6 +50,10 @@ vault:
       k8s:
         enabled: false
       app-role:
+        # this is not needed, as we use a builder,
```
tomas-langer, line 53: Not sure why this change (and related changes in code) is shown as part of your PR.

integrations/oci/objectstorage-health/src/main/java/io/helidon/integrations/oci/objectstorage/health/OciObjectStorageHealthCheck.java
```diff
@@ -0,0 +1,260 @@
...
+    @ConfigProperty(name = "oci.objectstorage.namespace")
+    private String namespace;
+
+    @Inject
+    @ConfigProperty(name = "oci.objectstorage.health.timeout-millis", defaultValue = "10000")
+    private long timeoutMillis;
+
+    private Duration timeout;
+
+    private OciObjectStorageRx ociObjectStorage;
```
tomas-langer, line 70: All fields should be final. You can use OciObjectStoraget both for SE and for MP (as you can create an instance of it in se using a factory method).
In MP this instance should be injectable.

integrations/oci/objectstorage-health/src/test/resources/application.yaml
```diff
@@ -0,0 +1,24 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+server:
+  port: 8080
+
+oci:
+  objectstorage:
+    namespace: "idv5cxr1uqqk"
```
tomas-langer, line 22: This is probably not a valid value. Maybe it should not be defined?

tomas-langer (approved):

PR #3192, opened 2021-07-15
Caching - Initial design proposal

tomas-langer:
This is just an initial design of API/SPI.
We may not implement this feature at all if we decide so!


examples/caching/cdi/src/main/java/io/helidon/examples/caching/cdi/CachedBean.java
```diff
@@ -0,0 +1,64 @@
...
+        System.out.println(cache.get(79).await());
+    }
+
+    @CacheGet
+    public String getIt(@CacheKey Integer id) {
+        return "produced-" + id + "-" + counter.incrementAndGet();
+    }
+
+    @CacheInit
+    public String createIt(@CacheKey Integer id) {
```
aseovic, line 51: How is `@CacheInit` different from `@CacheGet`? Is the idea to initialize the cache with a response, but always execute the method, where `@CacheGet` would only execute the method if the value is not cached?

examples/caching/cdi/src/main/java/io/helidon/examples/caching/cdi/CachedBean.java
```diff
@@ -0,0 +1,64 @@
...
+import io.helidon.caching.annotation.CacheGet;
+import io.helidon.caching.annotation.CacheInit;
+import io.helidon.caching.annotation.CacheKey;
+import io.helidon.caching.annotation.CacheName;
+import io.helidon.caching.annotation.CachePut;
+import io.helidon.caching.annotation.CacheRemove;
+import io.helidon.caching.annotation.CacheValue;
+
+@ApplicationScoped
+@CacheName("simple-cache")
```
aseovic, line 34: I'm assuming this is just for convenience, and can be specified/overriden at the method level as well, in case you have `getFoo`, `getBar`, etc?

aseovic (approved):
I like it.

For a moment I wasn't sure about all the annotations, and thought they may be an overkill, but after looking at the example I realized that they map quite well to different things you may want to accomplish, and allow you to do more than what either Micronaut or Spring do, for example.

One thing I'm not sure about is the `CacheLoader`. It seems unnecessary, as the method annotated with `@CacheGet` or `@CacheInit` effectively accomplishes the same thing (ie. we execute it and cache its result). Also, many cache providers, including Coherence, will have their own cache loading support, which is probably better suited if you really want to avoid cache miss. For example, Coherence cache loaders will lock the key that is being loaded and only load the value from a backend data store once, regardless of how many requests for the same key are made concurrently. I may be wrong, but I don't think current implementation provides the same optimization, which as I said earlier, makes it a bit redundant.

spericas:
The semantics of `@CacheInit` are not very clear to me. The name suggests a more primordial operation that what it appears to be. Your example seems to suggest that `@CacheCreate` (or maybe `@CacheAdd`) would be a better name for the annotation.

examples/caching/cdi/src/main/java/io/helidon/examples/caching/cdi/CachedBean.java
tomas-langer: Yes, that is the distinction

examples/caching/cdi/src/main/java/io/helidon/examples/caching/cdi/CachedBean.java
tomas-langer: Yes, the cache name is taken from method, then from class

tomas-langer:
`CacheLoader` is now propagated to cache integration layer - so it can be used to initialize the cache itself (if supported by the provider), or used in the manner of  `LoaderBackedCache` which delegates all other methods to the cache implementation.
If this is not feasible even for Coherence, I can remove that

tomas-langer:
@spericas - yes, `@CacheAdd` may be a better name.

spericas (approved):
LGTM

tomas-langer:
Closing this PR, as we never decided to continue with the feature.

Issue #3200, opened 2021-07-19
Disable deserialization by default


tomas-langer:

tomas-langer:
The default action on missing config should be `CONFIGURE` instead of `WARN`.
The default action on wrong config should be `FAIL` instead of `WARN`

See if we can support `CONFIGURE` action for wrong config through JEP-415 https://openjdk.java.net/jeps/415

PR #3201, opened 2021-07-19
JEP-290, SerialConfig support

### LINKED ISSUE #3158: JEP 290 - deserialization filters
Although Helidon does not use any Java serialization, we should provide a robust approach to JEP 290

This may include:

- [ ] #3200
- [x] Allow custom filters overriding this default
- [x] (maybe) only do whitelisting with blacklist being everything else
- [x] Blog article

tomas-langer:
Helidon does not use Java serialization, though we want to be as secure as possible. This PR provides ground work to customize configuration that only allows whitelisting (everything should be blacklisted by default).
We cannot change the defaults in this version, as this would create a backward incompatible behavior.
The defaults will change in 3.0.0

Resolves #3158

### DIFF
--- a/common/common/src/main/java/io/helidon/common/SerializationConfig.java
+++ b/common/common/src/main/java/io/helidon/common/SerializationConfig.java
@@ -19,181 +19,603 @@ package io.helidon.common;
 import java.io.IOException;
 import java.io.ObjectInputFilter;
 import java.net.URL;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 /**
  * Support for JEP 290 - deserialization filtering.
- * Helidon is implemented to support whitelists, automatically blacklisting
+ * Configuration options mentioned below will differ in Helidon 3.0.0, the following table lists the options:
+ * <p>
+ * <table class="config">
+ * <caption>Configuration Options</caption>
+ * <tr>
+ *     <th>system property</th>
+ *     <th>2.x default</th>
+ *     <th>3.x default</th>
+ *     <th>description</th>
+ * </tr>
+ * <tr>
+ *     <td>{@link #PROP_WRONG_CONFIG_ACTION}</td>
+ *     <td>{@code warn} - {@link io.helidon.common.SerializationConfig.Action#WARN}</td>
+ *     <td>{@code fail} - {@link io.helidon.common.SerializationConfig.Action#FAIL}</td>
+ *     <td>What to do if an existing global deserialization filter exists without a global blacklist.</td>
+ * </tr>
+ * <tr>
+ *     <td>{@link #PROP_NO_CONFIG_ACTION}</td>
+ *     <td>{@code warn} - {@link io.helidon.common.SerializationConfig.Action#WARN}</td>
+ *     <td>{@code configure} - {@link io.helidon.common.SerializationConfig.Action#CONFIGURE}</td>
+ *     <td>What to do if there is no global deserialization filter.</td>
+ * </tr>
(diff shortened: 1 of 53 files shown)


tomas-langer:
Follow up issue #3200

common/common/src/main/java/io/helidon/common/SerializationConfig.java
```diff
@@ -0,0 +1,624 @@
...
+ *     <th>system property</th>
+ *     <th>2.x default</th>
+ *     <th>3.x default</th>
+ *     <th>description</th>
+ * </tr>
+ * <tr>
+ *     <td>{@link #PROP_WRONG_CONFIG_ACTION}</td>
+ *     <td>{@code warn} - {@link io.helidon.common.SerializationConfig.Action#WARN}</td>
+ *     <td>{@code fail} - {@link io.helidon.common.SerializationConfig.Action#FAIL}</td>
+ *     <td>What to do if an existing global deserialization filter exists without a global blacklist.</td>
```
barchetta, line 53: Maybe use allow-list and reject-list (and allow and reject) to align with https://docs.oracle.com/en/java/javase/11/core/serialization-filtering1.html

common/common/src/main/java/io/helidon/common/SerializationConfig.java
```diff
@@ -0,0 +1,624 @@
...
+            return this;
+        }
+
+        /**
+         * Filter pattern to use.
+         *
+         * @param filterPattern filter pattern
+         * @return updated builder
+         */
+        public Builder filterPattern(String filterPattern) {
```
barchetta, line 449: Is `!*` always appended to this pattern? If so then I think it's more of an allow-list then a general filter. So maybe this method should be `allowPattern` or `allowListPattern`.

common/common/src/main/java/io/helidon/common/SerializationConfig.java
```diff
@@ -0,0 +1,624 @@
...
+ * <ul>
+ *     <li>{@code none} to disable tracing (this is the default)</li>
+ *     <li>{@code basic} to enable basic tracing (only traces requests for class deserialization)</li>
+ *     <li>{@code full} to enable full tracing (including sizes, depth etc.)</li>
+ * </ul>
+ */
+public final class SerializationConfig {
+    static final String PROP_WRONG_CONFIG_ACTION = "helidon.serialFilter.failure.action";
+    static final String PROP_NO_CONFIG_ACTION = "helidon.serialFilter.missing.action";
+    static final String PROP_PATTERN = "helidon.serialFilter.pattern";
```
barchetta, line 93: If `!*` is always appended to the value of this property, then maybe the property name should be something like `helidon.serialFilter.allowPattern`

common/common/src/main/java/io/helidon/common/SerializationConfig.java
```diff
@@ -0,0 +1,624 @@
...
+    static final String PROP_TRACE = "helidon.serialFilter.trace";
```
barchetta, line 94: I wonder if this could be triggered off of log levels instead of inventing a new property. For example:
helidon.serialFilter.trace=basic -> io.helidon.serialFilter.level=INFO 
helidon.serialFilter.trace=full -> io.helidon.serialFilter.level=FINE

Maybe that complicates the implementation too much.

tomas-langer: Problem with logging is that we may not be using Java Util logging at runtime - it may be log4j, logback etc. And I am not sure we could even reach this information from code

tomas-langer: the reason it is done in advance is for performance - I do not have tracing code on code path at all, unless you enable it

common/common/src/main/java/io/helidon/common/SerializationConfig.java
tomas-langer: This can get either a full pattern or allow list. This is not limited to allow...

tomas-langer: I have updated the documentation and code to support patterns with reject list

common/common/src/main/java/io/helidon/common/SerializationConfig.java
tomas-langer: Updated docs and code

barchetta (approved):
LGTM

PR #3206, opened 2021-07-20
Fix missing CORS headers if response has 404 status

### LINKED ISSUE #3188: Investigate and fix if needed the absence of CORS headers on unsuccessful (4xx or 5xx) replies
## Environment Details
* Helidon Version: 2.3.2 (reported in 2.2.2)
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
An internal user reported problems in a Jet U/I because responses from a Helidon MP app correctly had 4xx or 5xx statuses but the CORS headers were absent, causing the explanatory returned information in the response to not reach the Jet U/I.  CORS is enabled and the CORS headers are present on successful responses.

## Steps to reproduce
One way: Modify the Helidon MP CORS `CrossOriginTest` class as follows: (inspired by the internal user's fact-finding)
```
@@ -354,10 +354,10 @@ class CrossOriginTest {
         Response res = target.path("/notfound")
                 .request()
                 .header(ORIGIN, "http://foo.bar")
-                .header(ACCESS_CONTROL_REQUEST_METHOD, "PUT")
-                .put(Entity.entity("", MediaType.TEXT_PLAIN_TYPE));
-        assertThat(res.getStatusInfo(), is(Response.Status.NOT_FOUND));
-        assertThat(res.getHeaders().containsKey(ACCESS_CONTROL_ALLOW_ORIGIN), is(false));
+                .header(ACCESS_CONTROL_REQUEST_METHOD, "GET")
+                .get();
+        assertThat(res.getStatusInfo(), is(Response.Status.INTERNAL_SERVER_ERROR));
+        assertThat(res.getHeaders().containsKey(ACCESS_CONTROL_ALLOW_ORIGIN), is(true));
     }
 
     @Test
```

One key question is whether, in fact, CORS headers _should_ be present on unsuccessful responses. The CORS spec https://www.w3.org/TR/2020/SPSD-cors-20200602/ does not seem to mandate that the server suppress the CORS headers on a response if the HTTP status is 4xx or 5xx. Some random web search hits on the topic reveal this suppression in other projects being treated as a bug and fixed as such. 

`CorsSupportHelper#prepareResponse` contains logic that explicitly omits the CORS headers from such responses.

tjquinno:
Resolves #3188 

## Overview

Some code had been added a while ago to the CORS logic to explicitly avoid setting CORS headers for 404 responses. This PR removes that code.

An earlier unrelated change in Helidon's Jersey support was suppressing headers added to a response if the status was 404. This PR removes that header suppression.

## Two very slight behavioral changes
These are both truly bug fixes.
1. Previously, if the Helidon Jersey `ResponseWriter` detected a response with status 404 and no content, it would ignore any headers already set and immediately delegate to the next handler in the chain (if any). With these changes, any headers previously set on the `ContainerResponse` are _always_ copied to the `ServerResponse`. 
2. Previously, in MP, even if the `cars.enabled` config key was set to `false`, CORS processing would still occur for JAX-RS endpoints (as indicated by JAX-RS annotations). This PR includes changes so that config setting controls JAX-RS CORS behavior as well.

## Details of changes

* `webserver/jersey` 
    * `ResponseWriter#writeReponseStatusAndHeaders` - An earlier change deferred the transmission of the response to a later handler if the response status is 404 and the content length is 0. Change to return response (not defer) if there are any headers. This handles the case where CORS is in play and has added headers to the response even for the 404 case.
* `webserver/cors` 
    * `CorsSupportHelper#prepareResponse` - If response is 404, return valid CORS headers tailored to the (failed) request instead of suppressing CORS headers.
* `CrossOriginConfig` - add constant used for catch-all headers in 404 responses
* Added and modified some tests.

### DIFF
--- a/examples/microprofile/cors/src/test/java/io/helidon/microprofile/examples/cors/TestCORS.java
+++ b/examples/microprofile/cors/src/test/java/io/helidon/microprofile/examples/cors/TestCORS.java
@@ -125,7 +125,7 @@ public class TestCORS {
         assertThat("HTTP response payload", payload, is("Hola World!"));
         headers = r.headers();
         Optional<String> allowOrigin = headers.value(CrossOriginConfig.ACCESS_CONTROL_ALLOW_ORIGIN);
-        assertThat("Expected CORS header " + CrossOriginConfig.ACCESS_CONTROL_ALLOW_ORIGIN + " is absent",
+        assertThat("Expected CORS header " + CrossOriginConfig.ACCESS_CONTROL_ALLOW_ORIGIN + " is present",
                 allowOrigin.isPresent(), is(true));
         assertThat("CORS header " + CrossOriginConfig.ACCESS_CONTROL_ALLOW_ORIGIN, allowOrigin.get(), is("*"));
     }
--- a/microprofile/cors/src/test/java/io/helidon/microprofile/cors/CorsDisabledTest.java
+++ b/microprofile/cors/src/test/java/io/helidon/microprofile/cors/CorsDisabledTest.java
@@ -18,6 +18,7 @@ package io.helidon.microprofile.cors;
 import io.helidon.microprofile.tests.junit5.AddBean;
 import io.helidon.microprofile.tests.junit5.AddConfig;
 import io.helidon.microprofile.tests.junit5.HelidonTest;
+import org.junit.jupiter.api.Test;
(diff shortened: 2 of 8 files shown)


spericas (approved):
LGTM

Issue #3199, opened 2021-07-19
Memory leak with Request Scope and Fault Tolerance


gmpatter:
## Environment Details
* Helidon Version: 2.3.0
* Helidon MP
* JDK version: 11.0.9
* OS: MacOS 11.4

----------

## Problem Description
With Helidon MP 2.3.0 I’m seeing a potential memory leak in our application. In a heap dump from one load test I could see 595,000 instances of org.jboss.weld.contexts.SerializableContextualInstanceImpl with reference to an instance of Hk2RequestScope being hung on to by HelidonCdi instanceCache.

It seems to happen when injecting ServerRequest into a resource to use in an endpoint, while also using a Fault Tolerance annotation on the endpoint. Without the FT annotation I’m not seeing any instance of org.jboss.weld.contexts.SerializableContextualInstanceImpl created with reference to Hk2RequestScope.

## Steps to reproduce
Create a new helidon-quickstart-mp app.

Inject ServerRequest and update the GET endpoint to use ServerRequest. Also add @Timeout annotation to the GET endpoint:
```
@Context
ServerRequest mServerRequest;

@Path("/{name}")
@GET
@Produces(MediaType.APPLICATION_JSON)
@Timeout
public JsonObject getMessage(@PathParam("name") String name) {
  return createResponse(name + " tenant " + mServerRequest.headers().first("X-CCTenantId").get());
}
```
Make n number of requests to the endpoint:
```
for ((i=1;i<=1000;i++)); do   curl -v --header "X-CCTenantID: tenant1" "localhost:8080/greet/gareth"; done
```
Take a heap dump, you can see n number of SerializableContextualInstanceImpl with instance=org.glassfish.jersey.inject.hk2.Hk2RequestScope. Eclipse MAT OQL query:
```
SELECT instance AS instance FROM org.jboss.weld.contexts.SerializableContextualInstanceImpl s WHERE instance.toString().contains("org.glassfish.jersey.inject.hk2.Hk2RequestScope")
```
Force GC, take another heap dump and still see n number of SerializableContextualInstanceImpl.

gmpatter:
Although in the simple reproducer I see one instance per request, in our app, with multiple CDI beans using FT annotations and injecting ServerRequest, we see 9 instances created per request. Just adding in case it adds any extra layer of complication.

spericas:
@gmpatter Do you have a sense as to whether this issue was present in earlier versions of Helidon? Or was 2.3.0 the first version you tried this with?

[...]

PR #3214, opened 2021-07-22
Explicitly destroy bean instance obtained from CDI object

spericas:
Explicitly destroy bean instance obtained from CDI object that is not in a normal scope. Not manually destroying these instances resulted in the context' map monotonically growing in size. See analysis and description in Issue #3199.


tjquinno (approved):
LGTM.

PR #3225, opened 2021-07-27
Fix for issue 3174 - typo

### LINKED ISSUE #3174: Need to fix Helidon MP/SE doc - Metrics Support for Exemplars (typo)
## Environment Details
* Helidon Version: 2.3.1
* Helidon MP and SE

----------

ref: 
- [https://github.com/oracle/helidon/blob/master/docs/shared/metrics/prometheus-exemplar-support.adoc](https://github.com/oracle/helidon/blob/master/docs/shared/metrics/prometheus-exemplar-support.adoc)

now: If a metric necessarily corresponds to a specific sample—​for example a minimum or maximum—​Helidon associates a sample that has that exact value as the exemmplar for the metric.

shoud be: If a metric necessarily corresponds to a specific sample—​for example a minimum or maximum—​Helidon associates a sample that has that exact value as the exemplar for the metric.

ljamen:
Fixes #3174 
Fix was made in shared file and appears in both MP and SE versions of the document.

### DIFF
--- a/docs/mp/oci/03_vault.adoc
+++ b/docs/mp/oci/03_vault.adoc
@@ -70,7 +70,7 @@ oci.vault.cryptographic-endpoint: "<...>"
 
 The OCIDs can be set up and found in OCI under Security tab.
 
-image::oci/ocivault.png[OCI Vault, align="center"]
+image::oci/vaultkey.png[OCI Vault, align="center"]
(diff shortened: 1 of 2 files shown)


ljamen:
deleting PR to fix extra file issue

PR #3226, opened 2021-07-27
Doc fix to address adding long running code in hooks

ljamen:
There could be multiple places where this is applicable. Starting with Metrics. Need to identify other areas.

### DIFF
--- a/docs/se/metrics/01_metrics.adoc
+++ b/docs/se/metrics/01_metrics.adoc
@@ -30,7 +30,8 @@ Helidon SE provides the following to support metrics:
 2. A base set of metrics, available at `/metrics/base`, as specified by the MicroProfile Metrics specification.
 3. A set of Helidon-specific metrics, available at `/metrics/vendor`
 
-NOTE: Helidon-provided endpoints for `/metrics` do their work synchronously, using the same thread on which the request arrived via Netty. To avoid performance degradation, avoid including long-running code that can be invoked by these handlers while Helidon is responding to the metric.
+
+
 
 include::{common-deps-page-prefix-inc}[tag=maven-dependency]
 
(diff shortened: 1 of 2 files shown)


docs/se/metrics/01_metrics.adoc
```diff
@@ -30,6 +30,8 @@ Helidon SE provides the following to support metrics:
 2. A base set of metrics, available at `/metrics/base`, as specified by the MicroProfile Metrics specification.
 3. A set of Helidon-specific metrics, available at `/metrics/vendor`
 
+NOTE: Helidon-provided endpoints for `/metrics` do their work synchronously, using the same thread on which the request arrived via Netty. To avoid performance degradation, avoid including long-running code that can be invoked by these handlers while Helidon is responding to the metric.
```
tjquinno, line 33: I should have been clearer in the issue. 

Users will rarely encounter this situation. Perhaps add this after the new text you've already inserted to clarify when they need to be on-guard.

> For example, if you implement your own application-specific metrics, you will write logic to format the JSON and OpenMetrics output for that metric type. Helidon invokes this formatting logic whenever a client accesses the `/metrics` endpoints, so make that formatting code as efficient as possible.

tjquinno (changes requested):
Sorry for the change. I think the addition will help.

docs/se/metrics/01_metrics.adoc
ljamen: I moved the note a bit further down since its not likely to be applicable to most users (if I understood that correctly?) I also added the extra info.

tjquinno (approved):

tomas-langer:
This PR is approved and opened for a long time. Merging

PR #3233, opened 2021-07-29
Upgrade jaeger-client to 1.6.0

barchetta:
1. Upgrade jaeger-client to 1.6.0. This results in an upgrade of libthrift to 0.14.1 since it is a transitive dependency of jaeger-client
2. Move version management of okhttp3 and okio to `dependencies/pom.xml`
3. Exclude bad dependency that was introduced in libthrift 0.14.0. 

I have successfully built native-image for `tests/integration/native-image/mp-1` and run it.


tomas-langer (approved):
LGTM

PR #3234, opened 2021-07-29
Fix micrometer test

### LINKED ISSUE #3212: Micrometer test in examples fails
Follow up of #3210 

The test in io.helidon.examples.micrometer.se.MainTest does not have `.await()` in the last section that validates Micrometer.
The #3210 will separate the tests into methods and disable it.

The original test would just hang, as the failures are not correctly propagated. The test actually fails with:
```
org.opentest4j.AssertionFailedError: Unable to find expected all-gets timer count 2.0 ==> 
Expected :true
Actual   :false
```

tjquinno:
Resolves #3212


spericas (approved):
LGTM. 

My only comment is that startup 2 seconds and shutdown 10 seconds seems a bit odd. On a really slow pipeline, I would think 2 seconds may be a bit tight. That was there before, though.

spericas (approved):

Issue #3166, opened 2021-06-30
JAXRS filters invoked when not registered with JAXRS Application


gmpatter:
## Environment Details
* Helidon Version: 2.3.1
* Helidon MP
* JDK version:11.0.9
* OS: macOS 10.15.7

----------

## Problem Description
With Helidon MP 2.3.1 we are seeing JAXRS filters being executed for JAXRS Application's that do not register the filter, and when the filter does not use Provider annotation for auto discovery.
The problem seems to arise when we have more than one JAXRS Applications on the classpath. A filter registered with one JAXRS Application is invoked for all JAXRS Applications even if the filter is not registered with all JAXRS Applications.

## Steps to reproduce
Create a new filter that does some logging
```
public class LoggingFilter implements ContainerRequestFilter {
  protected static final Logger sLogger = LoggerFactory.getLogger(LoggingFilter.class);
  @Override
  public void filter(ContainerRequestContext containerRequestContext) throws IOException {
    sLogger.info("Filter Logging.");
  }
}
```
Create a JAXRS Application that registers the filter
```
@ApplicationScoped
@ApplicationPath("/alternate")
public class BlogsAlternateApplication extends Application {
  @Override
  public Set<Class<?>> getClasses() {
    return Set.of(BlogsAlternateApplication.class, LoggingFilter.class);
  }
}
```
And create a JAXRS Application that does not register the filter
```
@ApplicationScoped
@ApplicationPath("/")
public class BlogsApplication extends Application {
  @Override
  public Set<Class<?>> getClasses() {
    return Set.of(BlogsRestResource.class);
  }
}
```
Hit the endpoint of the rest resource in the JAXRS Application that does not register the filter, but still see the filter executed and logging in the console
```
2021-06-30 16:11:11.166 [helidon-1] INFO  com.oracle.cxcom.blogs.LoggingFilter - Filter logging.
2021-06-30 16:11:11.173 [helidon-1] INFO  com.oracle.cxcom.blogs.BlogsRestResource - Endpoint logging from app without filter registered.
```

spericas:
PR #3245

PR #3245, opened 2021-08-03
New injection manager for Helidon

spericas:
New implementation of a Jersey injection manager to handle separation of JAX-RS applications. The new injection manager uses two underlying managers: one that is shared among all applications and one to register classes and singletons returned by the application. For a Helidon application that defines N JAX-RS applications, there will be N+1 of these _core_ managers. 

Lookups and injections are carried out by first consulting the shared manager and then the one for the application. This may potentially add some runtime overhead that would need to be tested using our micro-benchmarks.

PR #3166


tjquinno (approved):
We had a thorough walkthrough of the bulk of the changes, and the single-case optimization is straight-forward.

ljnelson:
The only (naïve) question I have is: shouldn't you check the application-specific `InjectionManager` first and then fall back on the shared one?

spericas:
> The only (naïve) question I have is: shouldn't you check the application-specific `InjectionManager` first and then fall back on the shared one?

There's no overlap between them, and most of the beans will be in the shared one, so checking it in the reverse order will be more costly.

Issue #2670, opened 2021-01-15
io_uring support is used when available


dansiviter:
Although early days on [Netty `io_uring` support](https://netty.io/news/2020/11/16/io_uring-0-0-1-Final.html) it has some pretty positive performance improvements over `epoll`. As an extension to the work for that (#2444) would be great to get this too where it's available.

ljnelson:
Closing; Helidon 4.x is no longer based on Netty

Issue #2444, opened 2020-10-12
Ensure epoll support is used when available


ljnelson:
Netty offers the ability to [use epoll when it is available](https://netty.io/wiki/native-transports.html#using-the-linux-native-transport).  We should have this, silently, in our server layer.

dansiviter:
This will also help with Unix Domain Socket connections to K8s sidecars for a mild performance improvement.

ljnelson:
Closing this issue because #2732 introduces an experimental epoll `Transport` implementation as a first step. Selecting it as default when available is not necessarily what we want to do, but for now at least it is available.

PR #3247, opened 2021-08-03
Default to Epoll if available

dansiviter:
Open for feedback:
* As per the issue, defaults to Epoll if available,
* Unit test is conditional based on if Epoll is available so not to break on Mac or Windows,
* An `abstract` `NettyTransport` which gives scope to add a `io_uring` (#2670) implementation at a later date with few changes.

### DIFF
--- a/webserver/webserver/pom.xml
+++ b/webserver/webserver/pom.xml
@@ -160,6 +160,11 @@
             <artifactId>netty-handler</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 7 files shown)


ljnelson:
Note that #2444 exists; I need to rebase it and get it rolling.

dansiviter:
@ljnelson Apologies, I didn't notice your PR. I suppose it depends if Epoll is something that it to defaults to automatically or opt-in to which is the difference between the two approaches.

ljnelson:
No problem. I think that this should be opt-in: someone building on Windows doesn't want to drag around the dependency for epoll. My PR just checks to see, effectively, if the dependency is on the classpath, and, if so, uses it; otherwise nothing changes. Once I crawl out from my post-vacation inbox I'll rebase the PR and we'll see where it gets to.

dansiviter:
Sure. This was a precursor to trailing io_uring so I'll base that on your pattern.

PR #3250, opened 2021-08-05
Configuration annotation processor.

tomas-langer:

Documentation is now here: https://github.com/oracle/helidon/pull/3250/files#diff-fb1435883a8ac748d3e2ac8dd8e416d4ff9b32912b3595631479d5f2e3d90f19

To see what can be done - please check the new example. Just add a dependency that has metadata on the classpath, and run it (I would recommend `helidon-microprofile` bundle, as that contains most of the types that were created as part of this PR.

Annotations are in module `helidon-config-metadata`
The annotation processor is in module `helidon-config-metadata-processor`.

There are a few new annotations:

1. `io.helidon.config.metadata.Configured` - marks a Builder or a class as configurable through configuration
2. `io.helidon.config.metadata.ConfiguredOption` (repeatable) - defines a single configuration option
3. `io.helidon.config.metadata.ConfiguredValue` - possible values of an option (such as enum values)

`Configured` has the following properties:
- `root()` - root configured classes can be configured in configuration (such as `application.yaml`) as standalone components. Example: `server`, `security`, `tracing`
- `prefix()` - the configuration prefix used for this class (such as `server` for web server, `abac` for ABAC security provider), empty for most nested configurations, as they use the key defined on owning type
- `ignoreBuildMethod()` - treat this type as an abstract class/interface and not as a builder type (for cases where the builder is used as a parameter to another builder that builds the target type. This is not according to our guidelines, but used for example in `KeyConfig` for PEM and Keystore builders 
- `provides()` - interfaces/abstract classes implemented by this type that are used as providers to another configuration. Example: `SecurityProvider` SPI in Security, where each provider has its own configuration

`ConfigureOption` has the following properties:
- `key()` - defines key in configuration. Guessed from method name when on builder method (failOnError -> `fail-on-error`)
- `type()` - type of the configuration value. May be another `@Configured` type. In such a case that type can be used as a subtree of the key of this property. Guessed from method parameter when on builder method (primitive types are boxed)
- `description()` - description, created from Javadoc when on builder method
- `required()` - set to `true` for required options, false by default
- `value()` - what is the default value of this property, optional (if not defined and not required, default is defined in documentation)
- `experimental()` - marks an experimental property
- `kind()` - can be configured to `LIST`, `MAP` or `VALUE`, `VALUE` is the default
- `provider()` - set to true if this is provided by a provider that may be in another module
- `allowedValues()` - values with description for possible options of a property. Guessed from enum and javadoc for enumerations.
- `deprecated()` - deprecated option - description should describe how to change this
- `mergeWithParent()` - whether to ignore the key of this property and use all keys in the nested object instead (only for nested types, defaults to `false`)

Annotations are processed and a `META-INF/helidon/config-metadata-json` is created for each module
See the readme mentioned above that describes the JSON structure. If you build helidon from this PR, a few metadata files get generated (see `security/security`, `webserver/webserver` and some security providers (`oidc`, `http-basic-auth`, `abac`).

Required follow up - I will create these issues once this PR is merged, as working on them would not be benefitial now:
1. Add metadata to all Helidon modules
2. Use generated metadata in documentation
3. Use generated metadata in javadoc (maybe just a reference to a generated document from javadoc, or some more complex solutin)
4. Consider creating a validator plugin that would check a configuration file and report wrong keys/values based on current classpath
5. Consider creating a plugin that would generate the sample `application.yaml` from current project (as the example in this PR does), or that would generate a document with the same information (HTML?)
6. Consider generating an XML schema for the current classpath + a config source capable of processing configuration generated from that schema


common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
```diff
@@ -233,6 +240,7 @@ public Builder queueCapacity(int queueCapacity) {
          * @param daemon whether the threads are daemon threads
          * @return updated builder instance
          */
+        @ConfiguredOption(value = "is-daemon", defaultValue = "true")
```
danielkec, line 243: `value` containing  config key name looks little confusing especially next to `defaultValue`

security/providers/abac/src/main/java/io/helidon/security/providers/abac/AbacProvider.java
```diff
@@ -155,24 +158,24 @@ protected AuthorizationResponse syncAuthorize(ProviderRequest providerRequest) {
...
+                                                                                 EndpointConfig.AnnotationScope.values());
+                                                     annotationConfig.addAll(list);
+                                                 }
+                                             }
+
+                                             if (!annotationConfig.isEmpty()) {
+                                                 attributes.add(new RuntimeAttribute(validator,
+                                                                                     validator.fromAnnotations(epConfig)));
+                                             }
+                                         });
```
danielkec, line 178: just reformated

danielkec:
Awesome! :clap:

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
tomas-langer: Yes, I know. On the other hand this allows simple `@ConfiguredOption("key")`. I will check how often this would be used. If a lot, it makes sense to call it `value`, otherwise I will rename. it

tomas-langer: @ConfiguredOption("..") - 18 times
@ConfiguredOption(defaultValue="..") - 42 times
@ConfiguredOption(value = "", ...) - a lot

I will rename `value` to `key`.
We may also want to rename `defaultValue` to `value` as that is used much more often

security/providers/abac/src/main/java/io/helidon/security/providers/abac/AbacProvider.java
tomas-langer: Resolved

danielkec (approved):

config/metadata-processor/src/test/java/io/helidon/config/metadata/processor/ConfigMetadataProcessorTest.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package io.helidon.config.metadata.processor;
+
+import org.junit.jupiter.api.Test;
+
+class ConfigMetadataProcessorTest {
+    @Test
+    void testToConfigKey() {
+//        ConfigMetadataProcessor p = new ConfigMetadataProcessor();
```
romain-grecourt, line 24: What's the reason for this to be commented-out ?
You could use the `org.junit.jupiter.api.Disabled` annotation with some text to specify the reason.

docs-internal/config-metadata.md
```diff
@@ -0,0 +1,404 @@
...
+merge: if set to `true` (default is `false`) this option's key is ignored and all its nested keys are inserted
+     directly into the parent option
+experimental: experimental options may change without warning even between minor releases (such as support for Loom in Java)
+required: if set to `true` (default is `false`) this option must be present in configuration if parent option is present
+provider: this option expects a type with matching `provides` - such as security providers configuration
+deprecated: this option should no longer be used, there is another option allowing the same (usually when renaming options,
+     fixing inconsistencies etc.); description should provide more details
+defaultValue: for simple values, this defines the string default value
+
+Example of a simple optional element with defa# Configuration metadata
```
tjquinno, line 150: I think everything up to here is a near-duplicate of what follows. Extra paste?

docs-internal/config-metadata.md
```diff
@@ -0,0 +1,404 @@
...
+}
+```
+
+prefix - configuration prefix, if standalone, this is the root configuration key (such as `server`, `tracing`),
+if nested, this is a prefix added before all options of this type (such as each Security provider)
+standalone - if set to true, this type is configurable in the root of configuration (such as webserver, security, tracing, metrics),
+otherwise it is a nested type in another configuration (such as WebServerTls, OidcConfig)
+default is `false` (if not defined in document)
+producers (methods in general) - format is "fully qualified type name#method name (method parameter types)", example:
+`io.helidon.security.providers.common.OutboundTarget.Builder#build()`
```
tjquinno, line 253: This renders poorly. The lines all run together.

docs-internal/config-metadata.md
```diff
@@ -0,0 +1,404 @@
...
+kind VALUE: either a simple value (String, Long etc.) or a nested object (this is the default)
+method: method to be called to configure this option (may be a static factory method, such as `create(Config)`, or a builder method)
+merge: if set to `true` (default is `false`) this option's key is ignored and all its nested keys are inserted
+directly into the parent option
+experimental: experimental options may change without warning even between minor releases (such as support for Loom in Java)
+required: if set to `true` (default is `false`) this option must be present in configuration if parent option is present
+provider: this option expects a type with matching `provides` - such as security providers configuration
+deprecated: this option should no longer be used, there is another option allowing the same (usually when renaming options,
+fixing inconsistencies etc.); description should provide more details
+defaultValue: for simple values, this defines the string default value
```
tjquinno, line 297: Similar issue with rendered lines running together.

docs-internal/config-metadata.md
```diff
@@ -0,0 +1,404 @@
...
+                    <path>
+                        <groupId>io.helidon.config</groupId>
+                        <artifactId>helidon-config-metadata-processor</artifactId>
+                        <version>${helidon.version}</version>
+                    </path>
+                </annotationProcessorPaths>
+            </configuration>
+        </plugin>
+    </plugins>
+</build>
```
tjquinno, line 189: This is a lot of boilerplate to get right in multiple components' `pom.xml` files, especially if it ever has to change for some reason. 

Is the annotation processor fast enough so we could set it on by default for all components? If it finds nothing configurable it would not output anything.

If speed is an issue, maybe the plug-in could first check to see if the module depends on the metadata artifact and, if so, then do its full processing?

There might be other, better techniques to accomplish this. It would just be very nice if the build did the right thing without me (as a component developer) having to insert this identical boilerplate in all affected modules.

config/metadata/src/main/java/io/helidon/config/metadata/Configured.java
```diff
@@ -0,0 +1,84 @@
...
+ * There are two options for a supported type - either it is an interface/class
+ * with a {@code public static X create(Config)}, or a builder class with
+ * a method {@code public Builder config(Config)}.
+ *
+ * This annotation is used to provide IDE autocompletion, and to build documentation metadata
+ * using an annotation processor.
+ */
+@Target(ElementType.TYPE)
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
```
tjquinno, line 36: The annotations are not examined at runtime, are they? Why `RUNTIME` retention?

tjquinno (changes requested):
I've made some detailed comments attached to lines.

Would the `build-tools` repo, rather than the main repo, be a better fit for the new modules? All the interesting work--or _all_ the work?--happens at build-time.

romain-grecourt:
@tjquinno  The processor depends on the new metadata module, It has a high coupling with config overall. Having it co-located with config is a nice convenience. If it were located in build-tools it would make things quite complicated.

tjquinno:
@romain-grecourt Neither `config/metadata/pom.xml` nor `config/metadata-processor/pom.xml` contains any explicit or transitive dependencies on actual config, do they? 

For sure the processor is tightly coupled to the metadata, but I am not seeing the high coupling you are from either of those to config itself.

And now that I look at the poms again, `config/config/pom.xml` adds a provided optional dependency on the new metadata module but I do not see other files in the change list from `config/config` that would trigger that added dependency.

docs-internal/config-metadata.md
romain-grecourt: @tjquinno 

(not doing a quote reply as this is the very next comment)

`config/metadata` is a new config module and is used through-out the repository to annotate config related code.
One may have to update the metadata and processor when making certain kind of changes to the config API.

If you look at the processor implementation you can see it visits config code. E.g.
```java
configType = elementUtils.getTypeElement("io.helidon.config.Config").asType();
```

romain-grecourt:
@tjquinno 

(not doing a quote reply as this is the very next comment)

`config/metadata` is a new config module and is used through-out the repository to annotate config related code.
One may have to update the metadata and processor when making certain kind of changes to the config API.

If you look at the processor implementation you can see it visits config code. E.g.
```java
configType = elementUtils.getTypeElement("io.helidon.config.Config").asType();
```

tjquinno:
I did get that `config/metadata` is new and used from lots of modules. That by itself doesn't imply the metadata module should be in the main repo.

On the other hand, `configType` and `builderType` in `ConfigMetadataProcessor` do create the binding between `metadata` and `config/config` you alluded to earlier. It looks as if the `configType` is used in only a few places and maybe those spots could have used `String` comparisons instead of the (admittedly nicer) type comparisons. But `buildType` is referenced from many places involving erasures so that would have been more difficult to avoid.

I do get how messy it can get if a module in one repo depends on one from another repo and each have there own release cadence, etc. and I was never arguing for that. It just seems a shame that build-time stuff drifts into the main repo. Always trade-offs.

config/metadata-processor/src/test/java/io/helidon/config/metadata/processor/ConfigMetadataProcessorTest.java
tomas-langer: Forgot to uncomment

etc/scripts/test-integ-vault.sh
```diff
@@ -26,10 +26,22 @@ error_trap_setup
 
 mvn ${MAVEN_ARGS} --version
 
+# Workaround
```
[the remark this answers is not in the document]
tomas-langer: This should be resolved now as we have a dependency on it from Config.

tomas-langer:
> I've made some detailed comments attached to lines.
> 
> Would the `build-tools` repo, rather than the main repo, be a better fit for the new modules? All the interesting work--or _all_ the work?--happens at build-time.

I think the annotations are part of Helidon public API, so these should be part of Helidon itself. The annotation processor could be part of build tools, but then we would have a cyclic dependency (even if a "soft" one). If we did any change in the annotations, we would need to release new build tools, then upgrade to them in Helidon to be able to update the annotations themself.

docs-internal/config-metadata.md
tomas-langer: Removed

docs-internal/config-metadata.md
tomas-langer: added bullet points

docs-internal/config-metadata.md
tomas-langer: Added bullet points

docs-internal/config-metadata.md
tomas-langer: Yes, I will need to dive a bit deeper into this if we can find a better way of handling it.

config/metadata/src/main/java/io/helidon/config/metadata/Configured.java
tomas-langer: Using `SOURCE` as that is sufficient for annotation processing.

docs-internal/config-metadata.md
tomas-langer: I have found a way to do this. I have removed all dependencies from the annotation processor. Now it can be simply added as an optional provided dependency. 
This required creating a few classes  to write JSON output, but I think worth it to reduce dependencies...

romain-grecourt (approved):

danielkec (approved):

tomas-langer:
Follow up issue: #3463

PR #3251, opened 2021-08-05
Improve some interceptor code paths in metrics

tjquinno:
Optimize some metrics interceptor code paths.

The various commit messages describe the improvements.

### DIFF
--- a/metrics/metrics/src/main/java/io/helidon/metrics/ExponentiallyDecayingReservoir.java
+++ b/metrics/metrics/src/main/java/io/helidon/metrics/ExponentiallyDecayingReservoir.java
@@ -17,6 +17,7 @@
 package io.helidon.metrics;
 
 import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -24,6 +25,8 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import static java.lang.Math.exp;
 import static java.lang.Math.min;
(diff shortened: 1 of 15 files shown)


spericas (approved):
LGTM. This should help but we need to verify this with some benchmark numbers.

tjquinno:
I did some local benchmarking before, during, and after I made these changes. Locally I see distinct improvement, but the results varied enough from run to run on my system that I have asked the PSR team to rerun their own metrics benchmark in their environment--using the revised code--to assess the improvement.

tjquinno:
The PSR team reran the metrics workload with the changes. The overall throughput has improved but is not all the way back to the baseline of 2.2.2. PSR helpfully provided the JRF file from the run, as with the original JRF from 2.3.1. The 2.3.1 JRF showed at least two distinct metrics-related hotspots, both of which the coding changes in this PR addressed. The JRF from my own local testing as well as from PSR show those hotspots are gone and I see no other metrics-related hotspots. 

It's plausible that other, non-metrics performance regressions have crept in. Given that the metrics workload includes processing HTTP requests to trigger the metrics updates in the server (and, therefore, transmitting the HTTP responses) there might be other places to look for causes of the remaining regression.

metrics/metrics/src/main/java/io/helidon/metrics/Registry.java
```diff
@@ -225,12 +236,14 @@ public SimpleTimer simpleTimer(Metadata metadata, Tag... tags) {
      */
     @Override
     public synchronized boolean remove(String name) {
-        final List<MetricID> metricIDs = allMetricIDsByName.get(name);
-        if (metricIDs == null) {
+        final List<Map.Entry<MetricID, HelidonMetric>> doomedMetrics = getMetricsByName(name);
```
ljnelson, line 239: Just a note: this seems like a large block of work to use `synchronized` on. I haven't been following all the locking semantics in here but maybe this could be chunked up a little more? Would that help with performance, or is this not in the hot path?

ljnelson (approved):

metrics/metrics/src/main/java/io/helidon/metrics/Registry.java
tjquinno: The good news is that `remove` happens very rarely, certainly not during typical workloads.

metrics/metrics/src/main/java/io/helidon/metrics/ExponentiallyDecayingReservoir.java
```diff
@@ -49,6 +54,49 @@
...
+        CURRENT_TIME_UPDATER_EXECUTOR_SERVICE.shutdown();
+        try {
+            boolean stoppedNormally =
+                    CURRENT_TIME_UPDATER_EXECUTOR_SERVICE.awaitTermination(CURRENT_TIME_IN_SECONDS_UPDATE_INTERVAL_MS * 10,
+                            TimeUnit.MILLISECONDS);
+            if (!stoppedNormally) {
+                LOGGER.log(Level.WARNING, "Shutdown of current time updater timed out; continuing");
+            }
+        } catch (InterruptedException e) {
+            LOGGER.log(Level.WARNING, "InterruptedException caught while stopping the current time updater; continuing");
```
ljnelson, line 88: Nit: if you "eat" an `InterruptedException` you should do `Thread.currentThread().interrupt()` to preserve its interrupted status. Obviously here stuff is shutting down so it probably doesn't matter.

ljnelson (approved):

PR #3252, opened 2021-08-05
Upgrade brave-opentracing to 1.0.0

barchetta:
Upgrades brave-opentracing to 1.0.0.

I have successfully built native-image for `tests/integration/native-image/mp-1` and run it.


tomas-langer (approved):
This change is introducing a small backward incompatibility due to module name change. Can only be part of 2.4.0

Issue #3249, opened 2021-08-05
Make sure JAX-RS client responses are closed in examples


tomas-langer:
Validate that all examples (and other usages) of JAX-RS clients correctly close responses (and clients if relevant)

jansupol:
The archetypes ([for instance quickstart](https://github.com/oracle/helidon/blob/master/archetypes/quickstart-mp/src/main/resources/src/test/java/__pkg__/MainTest.java.mustache)) would be good to be closed as well. Autoclosable can be used:
```
try (Response r=...) {
}
```

spericas:
Clearly we should do this in server environments and for long-lived VM's. But do we really want to update the code in all relevant tests even though those are short-lived VM's? Clearly correct but also a bit academic in a testing environment.

spericas:
PR #3259

PR #3259, opened 2021-08-10
Use try-with-resources with Response's in examples and quickstarts

spericas:
See Issue #3249


examples/microprofile/websocket/src/test/java/io/helidon/microprofile/example/websocket/MessageBoardTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021 Oracle and/or its affiliates. All rights reserved.
```
tjquinno, line 2: Should be "2020, 2021" I think?

tjquinno (commented):
One tiny copyright correction. Otherwise LGTM.

examples/microprofile/websocket/src/test/java/io/helidon/microprofile/example/websocket/MessageBoardTest.java
spericas: Yeap

tjquinno (dismissed):

tjquinno (approved):

PR #3268, opened 2021-08-13
Direct dependency for reactive streams api

### LINKED ISSUE #3169: Add direct dependency for reactive-streams
We have code that uses `org.reactivestreams` but relies on `reactive-streams` being pulled in transitively. In general we should not rely on transitive dependencies to resolve dependencies that we use directly. Impacted modules:

```
dbclient/mongodb
microprofile/reactive-streams
microprofile/messaging/core
messaging/connectors/jms
messaging/kafka
messaging/messaging

examples/microprofile/messaging-sse
examples/messaging/oracle-aq-websocket-mp
examples/messaging/jms-websocket-mp
examples/messaging/kafka-websocket-mp
```

These modules should declare direct dependencies on:

```
            <dependency>
                <groupId>org.reactivestreams</groupId>
                <artifactId>reactive-streams</artifactId>
            </dependency>
```

Also in `dependencies/pom.xml` the dependency management for `reactive-streams` should be moved from Section 3 to Section 1.

danielkec:
Fixes #3169

### DIFF
--- a/dbclient/mongodb/pom.xml
+++ b/dbclient/mongodb/pom.xml
@@ -54,6 +54,10 @@
             <artifactId>mongodb-driver-reactivestreams</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.reactivestreams</groupId>
+            <artifactId>reactive-streams</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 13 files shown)


barchetta (approved):
LGTM

Issue #3286, opened 2021-08-18
Micrometer support uses wrong configuration


tomas-langer:
Part 1 - see PR #3290 
There is a code in `MeterRegistryProducer`:
`Config micrometerConfig = Config.create().get(CONFIG_KEY);`
Should be: `Config micrometerConfig = MpConfig.toHelidonConfig(ConfigProvider.getConfig()).get(CONFIG_KEY);`

Part 2
Configuration should only be used at runtime (from native image point of view). The current implementation of the CDI extension and MeterRegistryProducer reads the configuration once (in constructor) due to call to `meterRegistry = MeterRegistryProducer.getMeterRegistry();`

As the instance is created eagerly during initialization of extension, it would use build time config and not runtime config.

PR #3290, opened 2021-08-18
Use MP config instead of Config.create() in MP components.

### LINKED ISSUE #3285: We should always use MP config in CDI extension
FaultToleranceCdiExtension uses `Config.create()` which should not be done. This should use the current MP config (and convert it to SE config if desired)

To replicate - create an invalid `meta-config.yaml` on classpath, with for example the following content:
```yaml
sources:
  - type: "wrong"
```
Create a custom main class and configure MP config as follows:
```java
ConfigProviderResolver configProvider = ConfigProviderResolver.instance();

configProvider.registerConfig(configProvider.getBuilder()
                                                // system properties, env vars and META-INF/microprofile-config.properties
                                              .addDefaultSources()
                                              .build(),
                                      Thread.currentThread().getContextClassLoader());

io.helidon.microprofile.cdi.Main.main(args);
```

This should make sure the registered config is used by all components. It fails now with
```
Caused by: java.lang.IllegalArgumentException: Config source of type wrong is not supported. Supported types: [file, prefixed, classpath, system-properties, directory, url, environment-variables]
...
at io.helidon.microprofile.faulttolerance.FaultToleranceExtension.registerMetricsAndInitExecutors(FaultToleranceExtension.java:273)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:97)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	
```

tomas-langer:
Resolves #3285 
Related to #3286

### DIFF
--- a/integrations/micrometer/cdi/src/main/java/io/helidon/integrations/micrometer/cdi/MeterRegistryProducer.java
+++ b/integrations/micrometer/cdi/src/main/java/io/helidon/integrations/micrometer/cdi/MeterRegistryProducer.java
@@ -21,10 +21,12 @@ import javax.enterprise.inject.Produces;
 
 import io.helidon.common.LazyValue;
 import io.helidon.config.Config;
+import io.helidon.config.mp.MpConfig;
 import io.helidon.integrations.micrometer.MeterRegistryFactory;
 import io.helidon.integrations.micrometer.MicrometerSupport;
 
 import io.micrometer.core.instrument.MeterRegistry;
+import org.eclipse.microprofile.config.ConfigProvider;
 
 @ApplicationScoped
 class MeterRegistryProducer {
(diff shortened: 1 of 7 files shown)


tjquinno:
Should we make a similar sort of change in `service-common/rest-cdi` `HelidonRestCdiExtension#registerService`? That code does the casting of MP `Config` to Helidon's `Config` itself, but it could instead use the utility method as in these changes.

tomas-langer:
> Should we make a similar sort of change in `service-common/rest-cdi` `HelidonRestCdiExtension#registerService`? That code does the casting of MP `Config` to Helidon's `Config` itself, but it could instead use the utility method as in these changes.

It is probably a good way to go, as that would work even with a third party implementation of MP Config.

tjquinno (approved):
LGTM. I looked at all the changes, paying particular attention to the revisions in Micrometer integration.

PR #3292, opened 2021-08-18
Config now registered before container initialization.

### LINKED ISSUE #3289: Config not used correctly when configured through Server.builder()
When I use `Server.builder().config(mpConfig).build().start()` the configuration that I provide is not used by all components in CDI.
This is most likely related to our static initialization of CDI (for native image). As a result, some components may be incorrectly configured (as they just use the default config, instead of the one explicitly provided).

This can be easily tested with JPA, if the datasoruce configuration is not available on default MP sources and only in an explicitly configured one.

We should either fix this problem (for example by doing the static initialization only in `build` method), or deprecate the `config` methods on server and require that configuration is registered with the `ConfigProviderResolver` before we build server.

tomas-langer:
Resolves #3289 

When config was set on `Server.builder()`, the container was initialized first, and only then was the config set as the MP wide config instance. This PR changes the ordering, so the MP configuration is first set, and only then the container gets initialized.

Note that when building native image the container is initialized immediately (as the build method is never called) as part of our native image process. This means that configuration at build time would differ.

It is still recommended to use default configuration or MP meta configuration when using Helidon MP with native image!

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
@@ -128,10 +128,6 @@ public interface Server {
      * Builder to build {@link Server} instance.
      */
     final class Builder {
-        private static final Logger LOGGER = Logger.getLogger(Builder.class.getName());
-
-        // this constant is to ensure we initialize Helidon CDI at build time
-        private static final HelidonContainer CONTAINER = HelidonContainer.instance();
         private static final Logger STARTUP_LOGGER = Logger.getLogger("io.helidon.microprofile.startup.builder");
 
         private final List<Class<?>> resourceClasses = new LinkedList<>();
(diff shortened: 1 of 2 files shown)


microprofile/server/src/main/java/io/helidon/microprofile/server/Server.java
```diff
@@ -177,6 +171,18 @@ private Builder() {
...
+            // configuration must be initialize before we start the container
+            ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+
+            if (null == config) {
+                this.config = ConfigProviderResolver.instance().getConfig(classLoader);
+            }
+
+            // make sure the config is available to application
+            ConfigProviderResolver.instance().registerConfig(config, classLoader);
+
```
danielkec, line 185: // configuration must be initializeED

danielkec (approved):

tjquinno (dismissed):
I like Daniel's correction of the typo. Other than that, LGTM.

danielkec (approved):

PR #3295, opened 2021-08-18
Fix NPE in outbound of JWT provider.

### LINKED ISSUE #3246: JWT fails when calling outbound using JAX-RS client
When JWT provider is configured (using JWK to verify inbound JWT), and a call is done using JAX-RS client, a NPE is throw from `JwtProvider` line 283 (version 2.3.0).

This can be fixed by denying propagation explicitly using `propagate:false` in `jwt` provider configuration.

tomas-langer:
Added tests to verify.

Resolves #3246 

### DIFF
--- a/security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
+++ b/security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
@@ -89,7 +89,7 @@ public final class JwtProvider extends SynchronousProvider implements Authentica
     private JwtProvider(Builder builder) {
         this.optional = builder.optional;
         this.authenticate = builder.authenticate;
-        this.propagate = builder.propagate;
+        this.propagate = builder.propagate && builder.outboundConfig.targets().size() > 0;
         this.allowImpersonation = builder.allowImpersonation;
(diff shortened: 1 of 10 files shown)


security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
```diff
@@ -603,7 +604,7 @@ public Builder validitySeconds(long validitySeconds) {
                 .tokenHeader("Authorization")
                 .tokenPrefix("bearer ")
                 .build();
-        private OutboundConfig outboundConfig;
+        private OutboundConfig outboundConfig = OutboundConfig.builder().build();
```
danielkec, line 607: Other defaults are defined in the builder, wouldn't it be a better place for this one? 
https://github.com/oracle/helidon/blob/ce28e044a5e13428a6d16f946b2fe10709fd6ed4/security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java#L594-L611

danielkec: It is, sorry my bad

danielkec (approved):

PR #3299, opened 2021-08-19
WIP: Health check for OCI vaults

spericas:


integrations/oci/vault-health/src/main/java/io/helidon/integrations/oci/vault/health/OciVaultHealthCheck.java
```diff
@@ -0,0 +1,165 @@
...
+import org.eclipse.microprofile.health.Liveness;
+
+import static io.helidon.common.http.Http.Status.OK_200;
+
+/**
+ * Liveness check for an OCI's Vault. Reads OCI properties from '~/.oci/config'.
+ */
+@Liveness
+@ApplicationScoped  // this will be ignored if not within CDI
+@BuiltInHealthCheck
```
tomas-langer, line 45: Built in health checks should only be used for those that should be disabled by `mp.health.disable-default-procedures` - I do not think this belongs to the list.
On the other hand you did the `BuiltInHealthCheck` annotation, so maybe my expectation is wrong here...

spericas: That's interesting. I have thought that annotation would apply to any health check that isn't application defined, but I can see why we may want to separate these. If dropped here, it should probably be dropped for object storage as well.

examples/integrations/oci/vault-cdi/src/main/resources/application.yaml
```diff
@@ -34,3 +34,6 @@ oci:
     vault:
       # named OCI configuration
       cryptographic-endpoint: "${oci.properties.cryptographic-endpoint}"
+  vault:
+    vault-ocid: "${oci.properties.vault-ocid}"    # Vault healthcheck property
```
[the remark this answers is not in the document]
spericas: We have used the `oci` prefix for the framework-defined properties elsewhere. The application in this case has opted to use the `app`, that's totally up to it.

PR #3311, opened 2021-08-24
Change Micrometer integration to use MP config and use runtime, not build-time, config

### LINKED ISSUE #3286: Micrometer support uses wrong configuration

tjquinno:
Resolves #3286 

The commit messages convey pretty much the nature of the changes.

This PR will _not_ build successfully until after #3290 merges; the Micrometer example will not work correctly.

### DIFF
--- a/examples/integrations/micrometer/mp/src/main/resources/META-INF/microprofile-config.properties
+++ b/examples/integrations/micrometer/mp/src/main/resources/META-INF/microprofile-config.properties
@@ -20,3 +20,6 @@ app.greeting=Hello
 # Microprofile server properties
 server.port=8080
 server.host=0.0.0.0
+
+micrometer.builtin-registries.0.type=prometheus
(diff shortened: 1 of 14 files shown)


tomas-langer (approved):
LGTM

PR #3329, opened 2021-08-27
Single.onCompleteResumeWithSingle

### LINKED ISSUE #3300: Issue with onCompleteResumeWith on Single
There are two problems:

- [ ]  `onCompleteResumeWith(T)` on `Single` returns `Multi`, where it should return a `Single`

- It looks like a bug at first look, but on the second it makes sense to return `Multi`:
```java
        Single.just(1)
                .onCompleteResume(2)
                .collectList()
                .forSingle(System.out::println);
// returns [1, 2]
```
```java
        Single.empty()
                .onCompleteResume(2)
                .collectList()
                .forSingle(System.out::println);
// returns [2]
```

- [ ] method `onCompleteResumeWithSingle(Function<T, Single<T>>)` is missing, yet it is available on `onErrorResume...`, for the sake of consistency, this method should exist as well
- Fixed by adding the operator `onCompleteResumeWithSingle(Function<Optional<T>, ? extends Single<? extends T>> onComplete)` in #3329

- [ ] method `onErrorResume()` expects a function (mapping throwable to a value), the same method on Single should be a function of result of the last Single mapping to a value, and on Multi it should be a `Supplier<T>`, as we may want to supply the value only at the time the stream finished, not right now
- Not sure I follow, its almost impossible for error to happen on `Single` between `onNext` and `onComplete`(only exception is RTE in onComplete body)

danielkec:
Fixes #3300

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/Single.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/Single.java
@@ -457,6 +457,17 @@ public interface Single<T> extends Subscribable<T>, CompletionStage<T>, Awaitabl
         return new MultiOnCompleteResumeWith<>(Multi.create(this), publisher);
     }
 
+
+    /**
+     * Resume stream with supplied single if onComplete signal is intercepted.
+     *
+     * @param onComplete function for supplying new Single with optional upstream value as a parameter
+     * @return Single
+     */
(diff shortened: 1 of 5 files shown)


tomas-langer (approved):
LGTM

Issue #3319, opened 2021-08-25
RequestScope not propagated to @Asynchronous method


mikmoila:
## Environment Details
* Helidon Version: 2.3.2
* Helidon MP
* JDK version:
openjdk version "18-loom" 2022-03-15
OpenJDK Runtime Environment (build 18-loom+2-74)
OpenJDK 64-Bit Server VM (build 18-loom+2-74, mixed mode, sharing)
* OS: Windows

----------

## Problem Description
It looks like request scope is not propagated to a bean having @Asynchronous method (Microprofile FT).

## Steps to reproduce
1) Add the following two classes to Helidon MP Quickstart ( https://helidon.io/docs/v2/#/mp/guides/02_quickstart ) example:

```
package io.helidon.examples.quickstart.mp;

import java.util.UUID;
import javax.annotation.PostConstruct;
import javax.enterprise.context.RequestScoped;

@RequestScoped
public class SharedBean {
  private String shared;

  @PostConstruct
  public void init() {
    this.shared = UUID.randomUUID().toString();
  }

  @Override
  public String toString() {
    return "SharedBean [shared=" + shared + "]";
  }
}

package io.helidon.examples.quickstart.mp;

import java.util.concurrent.CompletableFuture;
import java.util.concurrent.Future;

import javax.enterprise.context.RequestScoped;
import javax.inject.Inject;

import org.eclipse.microprofile.faulttolerance.Asynchronous;

@RequestScoped
public class AsyncWorker {
  @Inject
  private SharedBean shared;

  @Asynchronous
  public Future<String> asyncOp() {
    return CompletableFuture.completedFuture("tid=" + Thread.currentThread().getId() + " sees shared=" + shared);
  }
}

```
2) Replace getDefaultMessage() method of GreetResource and add two injections:

```
  @Inject
  private AsyncWorker worker;

  @Inject
  private SharedBean shared;

  @GET
  @Produces(MediaType.APPLICATION_JSON)
  public JsonObject getDefaultMessage() {
    try {
      var f = worker.asyncOp();
      System.out.println("tid=" + Thread.currentThread().getId() + " sees shared=" + shared + " " + f.get());
    } catch (Exception ee) {
      throw new RuntimeException(ee);
    }
    return createResponse("World");
  }

```
3) Call JAX-RS endpoint http://localhost:8080/greet - console logging shows that request scope'd bean instance is not shared e.g : 
[...]

PR #3332, opened 2021-08-27
Migration of CDI contextual beans into new threads

spericas:
Additional code is provided to not only active the request scope in the new thread but also migrate the beans in request scope. This is done according to the Weld documentation on context migration. 

A new test shows how a request-scoped bean created in thread A is the _same bean_ available in thread B using FT's `@Asynchronous` annotation. Note that the request-scoped bean in thread A _needs to be accessed_ for its proxy to be created and migration into thread B to be successful. 

Bean migration is one directional: from thread A (original request thread) to thread B and not the other way around. Application developers are responsible for properly synchronizing access to these beans that can potentially be accessed from multiple threads.

Issue #3319

### DIFF
--- a/microprofile/fault-tolerance/pom.xml
+++ b/microprofile/fault-tolerance/pom.xml
@@ -93,6 +93,14 @@
             <groupId>org.glassfish.jersey.ext.cdi</groupId>
             <artifactId>jersey-weld2-se</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.weld</groupId>
+            <artifactId>weld-spi</artifactId>
+        </dependency>
         <!-- To keep HK2 from complaining when run with Java9+ -->
         <dependency>
             <groupId>jakarta.activation</groupId>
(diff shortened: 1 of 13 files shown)


ljnelson:
Haven't even looked at this but I know that [Weld does context propagation](https://docs.jboss.org/weld/reference/latest/en-US/html/contexts.html#_propagating_built_in_contexts); you may want to double-check to see if one of its methods can help here so future bugs end up living where they're supposed to.  (These methods may not be available in the version of Weld that Helidon uses; not sure.  I just didn't want to lose the thread, so to speak.)

spericas:
@ljnelson That's exactly where I got my Inspiration, that link. I've adapted their code for the migration process, excluding the other scopes they show there. Interestingly, Jersey has similar code in `WeldRequestScope`, however, they don't actually call `clearAndSet` ever, which sees very strange. For that reason, I've decided to do this _by hand_ in Helidon instead of leveraging that Jersey class.

Could you think of a better way to do this migration than what is shown in this PR?

tomas-langer:
@spericas is this something specific to Fault tolerance? Could we use the same code to have a general way of enabling request context in any asynchronous code when using CDI?

tjquinno (approved):
Go!

PR #3335, opened 2021-08-30
Await with duration

### LINKED ISSUE #3303: Add await methods with Duration as a parameter.
We are using `Duration` in Helidon code to define timeouts more and more often.
It would be good to have support for it in reactive code method `await()`.

danielkec:
Fixes #3303


tomas-langer (approved):
LGTM

PR #3340, opened 2021-08-31
Avoid double caching by calling Application methods in ResourceConfig

spericas:
Also prevents calling `getClasses` and `getSingletons` multiple times. Calling `getSingleton` more than once can be problematic.


tjquinno (approved):
LGTM assuming existing tests would fail if caching does not happen exactly once.

Issue #1789, opened 2020-05-14
Align mappers (Config, DbClient, WebServer)


tomas-langer:
In WebServer, we now support mapping based on types that has a few possible outcomes:
- exact match
- match
- no match
using a `PredicateResult`

Similar approach would be useful in Config and in DbClient, when we search for a mapper.

Consider a scenario, where we have a mapper for `ZonedDateTime` registered, and also a general purpose mapper such as `JSON-B`.

In case JSON-B is registered first, it would be used to map also `ZonedDateTime`

If we use the predicate result, the mapper for `ZonedDateTime` would return `exact match`, while JSON-B would return `match`.

This way we could first look for exactly matching mappers, and only when none is found, we would use the first "matching" mapper.

It would also make sense to move the `PredicateResult` to common mapper (and maybe find other common classes)

tomas-langer:
The current implementation in Helidon 4.x honors weight of each media support (same for mappers), and we do have a distinction between "supported" and "compatible" - to avoid the case mentioned above.
Closing this issue as I think it was fixed as a side effect of other changes we have done for Helidon

PR #3344, opened 2021-08-31
Mappable values

### LINKED ISSUE #1897: mappable query params and path params
E.g.

```
req.path.param("foo").asInt().ifPresent()
req.queryParams().first("foo").asInt().ifPresent()
```

Related to #1789 and #1867

tomas-langer:
Resolves #1897 (mappable query params and path params)
Related to #1789 (align mappers)

Added `Value` and `ValueProvider` to `mapper` module.
It is now used in various HTTP classes (Headers, query parameters etc.) and in Config (`ConfigValue`).

Added support for a shared `MapperManager` that can be custom registered by user.
Mapping to basic types is built-in (String -> a lot of types), including date types. Format can be overridden through context if desired (see `MapperManager.FORMAT_CLASSIFIER`).

The changes are fully backward compatible - we have to decide how to approach this (e.g. do we want to keep all the methods, or do we want to change signatures in the future?).

Created as a draft until we agree on approach.

### DIFF
--- a/common/mapper/src/main/java/io/helidon/common/mapper/BackedValue.java
+++ b/common/mapper/src/main/java/io/helidon/common/mapper/BackedValue.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2021 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common.mapper;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+class BackedValue<T> extends EmptyValue<T> {
+    private final T value;
+
+    BackedValue(String name, T value) {
+        super(name);
+        this.value = value;
+    }
+
+    @Override
+    public Optional<T> asOptional() throws MapperException {
+        return Optional.of(value);
+    }
+
+    @Override
+    public <N> Value<N> as(Function<T, N> mapper) {
+        N result = mapper.apply(value);
+        if (result == null) {
+            return new EmptyValue<>(name());
+        }
+        return new BackedValue<>(name(), result);
+    }
+}
--- a/common/mapper/src/main/java/io/helidon/common/mapper/BackedValue.java
+++ b/common/mapper/src/main/java/io/helidon/common/mapper/BackedValue.java
@@ -16,6 +16,7 @@
 
 package io.helidon.common.mapper;
 
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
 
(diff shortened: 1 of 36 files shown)


romain-grecourt:
It would be easier to have a markdown document going over the proposal.
We should describe all the various use-cases with snippets:
 - config
 - dbclient
 - media support
 - webserver query params

Highlight the complex type usages with GenericType, document the special layering of media-support.

We should also have an example that demonstrates the value of this API when combining dbclient and webserver/webclient. E.g. Show how to support a custom entity all the way from a request payload to the database.

common/mapper/src/main/java/io/helidon/common/mapper/BackedValue.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+
+package io.helidon.common.mapper;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+
+class BackedValue<T> extends EmptyValue<T> {
```
danielkec, line 23: Why extend the EmptyValue? Only common func seems to be name accessor

common/mapper/src/main/java/io/helidon/common/mapper/MapperManagerImpl.java
```diff
@@ -81,6 +123,10 @@ private RuntimeException createMapperException(Object source,
                     .orElseGet(() -> {
                         GenericType<SOURCE> sourceGenericType = GenericType.create(sourceType);
                         GenericType<TARGET> targetGenericType = GenericType.create(targetType);
+                        if (targetType.isAssignableFrom(sourceType)) {
+                            // the same type
+                            return it -> it;
```
danielkec, line 128: Would be nice to have something like `Fuction.identity()` maybe `Mapper.noop()`?

common/mapper/src/main/java/io/helidon/common/mapper/SharedManagerHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common.mapper;
+
+import java.util.concurrent.atomic.AtomicReference;
+
+final class SharedManagerHandler {
+    private static final AtomicReference<MapperManager> SHARED = new AtomicReference<>(MapperManager.builder()
```
danielkec, line 22: Could be just volatile

common/mapper/src/main/java/io/helidon/common/mapper/Value.java
```diff
@@ -0,0 +1,296 @@
...
+     * @param mapper a mapping function to apply to the value, if present
+     *               the mapping function
+     * @return the result of applying an {@code Optional}-bearing mapping
+     * function to the value of this {@code Optional}, if a value is present,
+     * otherwise an empty {@code Optional}
+     * @throws NullPointerException if the mapping function is null or returns
+     *                              a null result
+     * @see Optional#flatMap(Function)
+     */
+    default <U> Optional<U> flatMap(Function<? super T, Optional<U>> mapper) {
```
danielkec, line 232: Why not `Function<? super T, Optional<? extends U>> mapper` ?

common/mapper/src/main/java/io/helidon/common/mapper/BackedValue.java
tomas-langer: yes, that is the only common thing. On the other hand, why implement it twice? Something with value is a logical extension of something without a value...

common/mapper/src/main/java/io/helidon/common/mapper/Value.java
tomas-langer: fixed

common/mapper/src/main/java/io/helidon/common/mapper/SharedManagerHandler.java
tomas-langer: Fixed

common/mapper/src/main/java/io/helidon/common/mapper/MapperManagerImpl.java
tomas-langer: Fixed

tomas-langer:
I am adding a `docs-internal/generic-mapping-2.md`

PR #3352, opened 2021-09-01
OCI Rest API fix security service in webclient

### LINKED ISSUE #3350: OCI API calls are returning 401
While developing OCI ATP integration, I noticed that after uptaking latest master, the calls to OCI APIs are failing with 401. I noticed that the required authorization headers are not added as part of request.

This has started happening with change https://github.com/oracle/helidon/pull/3277. 

Basically, the webClientSecurity service is not becoming available as part of webClientConfiguration. This may be happening as we now copy the services list instead of referring to same list with the above mentioned change in file webclient/webclient/src/main/java/io/helidon/webclient/WebClientConfiguration.java i.e. https://github.com/arjav-desai/helidon/commit/99eeccd9f6b43abbe41b595b60187e671f7976fd#diff-f7238458dd565852e7673f0396a22014eb6a75b08e152ae4d847815dc4b9eefd

Fwiw, reverted that one file and its working as before.

tomas-langer:
Resolves #3350 

### DIFF
--- a/integrations/oci/connect/src/main/java/io/helidon/integrations/oci/connect/OciRestApi.java
+++ b/integrations/oci/connect/src/main/java/io/helidon/integrations/oci/connect/OciRestApi.java
@@ -485,7 +485,6 @@ public class OciRestApi extends RestApiBase {
 
         @Override
         protected void preBuild() {
-            super.preBuild();
             if (ociConfigProvider == null) {
                 LOGGER.finest("Config provider is not configured explicitly. Config type: " + configType);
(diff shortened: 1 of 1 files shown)


arjav-desai (approved):
LGTM.

Have run OCI ATP test with this change locally and it works now.

spericas (approved):

spericas:
@tomas-langer Thanks, I was debugging this and modifying code in that class. Glad this fixes it.

PR #3358, opened 2021-09-02
Automated discovery of JAX-RS method entities.

### LINKED ISSUE #3090: Should automatically provide GraalVM settings when endpoint accepts or returns a POJO
## Environment Details
* Helidon Version: 2.3.0
* Helidon SE or Helidon MP MP
* JDK version: n/a
* OS: n/a

----------

## Problem Description
A JAX-RS endpoint method which returns a POJO works fine when launched using `java -jar` but returns empty responses when run as a native image.

## Steps to reproduce
1. Use `helidon init` to create an MP project based on the quick-start example.
2. Add a POJO `Message` class with the `String` attribute `message`:
    ```
    public class Message {
        private String message;
    
        public Message(String msg) {
            message = msg;
        }
    
        public String getMessage() {
            return message;
        }

        public void setMessage(String message) {
            this.message = message;
        }
    }
    ```
3. Modify one of the generated endpoints, or add a new one at a new path, which instantiates and returns a `Message` object instead of `JsonObject`:
    ```
        @GET
        @Path("/pojo")
        public Message getDefaultMessagePojo() {
            return new Message(String.format("%s %s!", greetingProvider.getMessage(), "World"));
        } 
    ```
4. Build for GraalVM but run _without_ GraalVM:
    ```
    mvn package -Pnative-image
    ```
    Access the endpoint: `curl http://localhost:8080/greet/pojo` The response is as expected.
5. Run _with_ GraalVM (`-Pnative-image`). Access the endpoint. The response is empty.

## To work around this:
1. Add 
    ```
    [
      {
        "name": "me.tjquinn.mp.quickstart.Message",
        "allPublicConstructors" : true,
        "allPublicMethods" : true
      }
    ]
    ```
    to the empty generated `META-INF/native-image/reflect-config.json` file.
2. Rebuild (with `-Pnative-image`) and rerun. The response is now correct.

## Change:
Ideally, Helidon would automatically generate the `reflect-config` data for any POJO that the developer uses as a JAX-RS endpoint parameter or return value.

tomas-langer:
Resolves #3090 

Added analysis of response types of JAX-RS resource methods, and entity parameter types (types with no annotation).
These types are added for reflection automatically now.

### DIFF
--- a/integrations/graal/native-image-extension/src/main/java/io/helidon/integrations/graal/nativeimage/extension/HelidonReflectionFeature.java
+++ b/integrations/graal/native-image-extension/src/main/java/io/helidon/integrations/graal/nativeimage/extension/HelidonReflectionFeature.java
@@ -116,6 +116,9 @@ public class HelidonReflectionFeature implements Feature {
         // all classes, fields and methods annotated with @Reflected
         addAnnotatedWithReflected(context);
 
+        // all classes used as return types and parameters in JAX-RS resources
+        processJaxRsTypes(context);
+
         // JAX-RS types required for headers, query params etc.
         addJaxRsConversions(context);
 
@@ -220,6 +223,17 @@ public class HelidonReflectionFeature implements Feature {
                 }
             }
         }
+    }
+
+    private void processJaxRsTypes(BeforeAnalysisContext context) {
+        tracer.parsing(() -> "Looking up JAX-RS resource methods.");
+
+        new JaxRsMethodAnalyzer(context, util)
+                .find()
+                .forEach(it -> {
+                    tracer.parsing(() -> " class " + it);
+                    context.register(it).addAll();
+                });
     }
 
     private void addAnnotatedWithReflected(BeforeAnalysisContext context) {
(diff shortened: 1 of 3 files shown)


integrations/graal/native-image-extension/src/main/java/io/helidon/integrations/graal/nativeimage/extension/HelidonReflectionFeature.java
```diff
@@ -222,6 +225,17 @@ private void addJaxRsConversions(BeforeAnalysisContext context, String annotatio
         }
     }
 
+    private void processJaxRsTypes(BeforeAnalysisContext context) {
+        tracer.parsing(() -> "Looking up JAX-RS resource methods.");
+
+        new JaxRsMethodAnalyzer(context, util)
```
spericas, line 231: Looks a bit weird not being a static method or using a local var.

spericas (approved):

integrations/graal/native-image-extension/src/main/java/io/helidon/integrations/graal/nativeimage/extension/HelidonReflectionFeature.java
tomas-langer: I have merged the PR even with this comment, as I do not think of it as a technical debt. The implementation was easier using an instance.

Issue #3354, opened 2021-09-01
MayPayloadSizeTest fails intermittently


tomas-langer:
`io.helidon.webserver.MaxPayloadSizeTest.testActualLengthExceededWithPayload`

```
java.util.concurrent.CompletionException: java.lang.IllegalStateException: (client reqID: 9) Connection reset by the host
	at io.helidon.common.reactive.Awaitable.await(Awaitable.java:71)
	at io.helidon.webserver.MaxPayloadSizeTest.testActualLengthExceededWithPayload(MaxPayloadSizeTest.java:152)
```

stdout
```
2021.09.01 15:20:02 INFO io.helidon.webserver.NettyWebServer Thread[nioEventLoopGroup-10-1,10,main]: Channel '@default' started: [id: 0x1c59e88a, L:/0.0.0.0:34659]
2021.09.01 15:20:02 INFO io.helidon.webserver.MaxPayloadSizeTest Thread[main,5,main]: Started server at: https://localhost:34659
2021.09.01 15:20:02 INFO io.helidon.webserver.NettyWebServer Thread[nioEventLoopGroup-10-1,10,main]: Channel '@default' closed: [id: 0x1c59e88a, L:/0.0.0.0:34659]
```

cause
```
Caused by: java.io.IOException: Broken pipe
	at java.base/sun.nio.ch.FileDispatcherImpl.write0(Native Method)
	at java.base/sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:47)
	at java.base/sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:113)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:58)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:50)
	at java.base/sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:466)
	at io.netty.channel.socket.nio.NioSocketChannel.doWrite(NioSocketChannel.java:408)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush0(AbstractChannel.java:953)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.flush0(AbstractNioChannel.java:354)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.flush(AbstractChannel.java:917)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.flush(DefaultChannelPipeline.java:1372)
[...]

PR #3360, opened 2021-09-02
Handle potential connection reset conditions in test

spericas:
Handle potential connection reset conditions when waiting for a 413 while attempting to write more data to the server. This appears to be a problem in slower systems causing testActualLengthExceededWithPayload() to fail.

Issue #3354


tjquinno (approved):
LGTM

Issue #3072, opened 2021-06-03
OCI Integration with ATP DB


arjav-desai:
This task is to document how to integrate helidon apps with ATP DB running in OCI e.g. how wallet file generated can be stored in vault and used with helidon applications etc.

aalmiray:
FWIW Micronaut uses the OCI REST API to request the wallet on the fly and load it in memory thus avoiding the need to store the wallet in a secure location as an intermediate step. You might want to support this use case as an additional option.

arjav-desai:
Thanks @aalmiray for the feedback. We will look into similar feature/API support in Helidon.

PR #3369, opened 2021-09-03
OCI ATP Integration Part 2

arjav-desai:
This PR is for reminder of work for #3072 where wallet is stored in memory and example shows how to access it.


integrations/oci/atp/src/main/java/io/helidon/integrations/oci/atp/GenerateAutonomousDatabaseWallet.java
```diff
@@ -85,15 +87,15 @@ public WalletArchive walletArchive() {
...
 
         /**
          * Set wallet data.
          *
          * @param content
          */
-        public WalletArchive(final byte[] content) {
-            this.content = content;
+        public WalletArchive(byte[] content) {
+            this.content = Arrays.copyOf(content,content.length);
```
ljnelson, line 98: Or `this.content = content.clone()`

integrations/oci/atp/src/main/java/io/helidon/integrations/oci/atp/GenerateAutonomousDatabaseWallet.java
```diff
@@ -102,7 +104,7 @@ public WalletArchive(final byte[] content) {
          * @return bytes[]
          */
         public byte[] getContent() {
-            return content;
+            return Arrays.copyOf(content,content.length);
```
ljnelson, line 107: Or `return content.clone()`

integrations/oci/atp/src/main/java/io/helidon/integrations/oci/atp/GenerateAutonomousDatabaseWallet.java
danielkec: I think the performance is the same, but using `clone()` was rather discouraged in  the Effective Java book

arjav-desai: IIRC, its more about custom object implementing clone method in book. I don't have preference over one vs other as, like Daniel said, performance is almost the same.

arjav-desai: Fixed.

integrations/oci/atp/src/main/java/io/helidon/integrations/oci/atp/GenerateAutonomousDatabaseWallet.java
arjav-desai: Fixed.

examples/integrations/oci/atp-cdi/src/main/java/io/helidon/examples/integrations/oci/atp/cdi/AtpResource.java
```diff
@@ -69,21 +73,29 @@ public Response generateWallet() {
...
-                .header("request-id", ociResponse.requestId());
-
-        ociResponse.headers()
-                .first(Http.Header.CONTENT_TYPE)
-                .ifPresent(ok::type);
-
-        ociResponse.headers()
-                .first(Http.Header.CONTENT_LENGTH)
-                .ifPresent(it -> ok.header(Http.Header.CONTENT_LENGTH, it));
+        try {
```
ljnelson, line 76: Consider try-with-resources instead so that the `close()` is effectively put into a `finally` block for you

arjav-desai: Changing examples a lot in upcoming PR to add how to use wallet and do DB operations using it. Will take this up with that.

danielkec (approved):

PR #3380, opened 2021-09-07
Close connection with ServerRequest

### LINKED ISSUE #1278: ServerRequest should have API method for closing connection
Because closing by cancelling stream has been removed by #1260

@romain-grecourt :
IIRC this is because `HttpRequestScopedPublisher` closes the connection when the subscription is canceled (see `hookOnCancel`).

This is not compliant with the reactive spec because I did not want to change this behavior and instead worked around the issue. This test simulates that context to make sure that we do not end-up closing connections (the test subscription represents the subscription to HttpRequestScopedPublisher).

I do not know the original reason behind this behavior, I'm guessing it was by design so that a user can close the underlying connection by canceling the subscription.

I'd say it's fine to tackle this issue for 2.0, i.e change `HttpRequestScopedPublisher.hookOnCancel` to not close the connection and remove the non compliant test + base processor implementation.

We can also discuss the need for a proper API to close the connection, maybe on `ServerRequest`.

Whole discussion can be found [here](https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_oracle_helidon_pull_1260-23discussion-5Fr364211092&d=DwMCaQ&c=RoP1YumCXCgaWHvlZYR8PZh8Bv7qIrMUB65eapI_JnE&r=U5cu6Y7WrL4EsqRsc4Tgk6Zsa2RXQd5T6QzK-AXSugw&m=ZYPWkBWtlok3sn1BcDdHUCHbANJwQbZJeRLWFDEBimc&s=MchFIYbDfoZ_F8KOlbV_cR7Q86ushgG-YdciU9ztCDE&e=)

danielkec:
Fixes #1278

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerRequest.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import java.util.Optional;
 import io.helidon.common.context.Context;
 import io.helidon.common.http.Http;
 import io.helidon.common.http.HttpRequest;
+import io.helidon.common.reactive.Single;
 import io.helidon.media.common.MessageBodyReadableContent;
 
 import io.opentracing.SpanContext;
@@ -141,6 +142,13 @@ public interface ServerRequest extends HttpRequest {
      * @return the tracer associated, or {@link io.opentracing.util.GlobalTracer#get()}
      */
     Tracer tracer();
+
+    /**
+     * Closes the connection.
+     *
+     * @return Single completed when channel is closed.
+     */
+    Single<Void> close();
 
     /**
      * Absolute URI of the incoming request, including query parameters and fragment.
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerRequest.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerRequest.java
@@ -144,11 +144,12 @@ public interface ServerRequest extends HttpRequest {
     Tracer tracer();
 
     /**
-     * Closes the connection.
+     * Request to close the connection and report success or failure asynchronously with returned single.
+     * After connection is closed it is not possible to use it again.
      *
-     * @return Single completed when channel is closed.
+     * @return Single completed when connection is closed.
      */
-    Single<Void> close();
+    Single<Void> closeConnection();
 
     /**
      * Absolute URI of the incoming request, including query parameters and fragment.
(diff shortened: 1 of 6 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/CloseConnectionTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Test
+    void closeManually() throws Exception {
+        try (SocketHttpClient c = new SocketHttpClient(webServer)) {
+            c.request(Http.Method.GET);
+            String result = c.receive();
+            SocketHttpClient.assertConnectionIsClosed(c);
+            assertThat(result, Matchers.containsString("item0"));
+            assertThat(result, not(Matchers.containsString("item9")));
```
Verdent, line 85: Since you are having not() statically imported, I think having containsString() the same way would be better :-)

danielkec: fixed

webserver/webserver/src/test/java/io/helidon/webserver/CloseConnectionTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private WebServer webServer;
+    private final ScheduledExecutorService exec = Executors.newSingleThreadScheduledExecutor();
+
+    @BeforeEach
+    void setUp() throws UnknownHostException {
+        InetAddress localHost = InetAddress.getLocalHost();
+        webServer = WebServer
+                .builder()
+                .port(5555)
```
romain-grecourt, line 48: Can this be a random available port instead of being hard-coded ?

Verdent (dismissed):

webserver/webserver/src/test/java/io/helidon/webserver/CloseConnectionTest.java
danielkec: Fixed

Verdent (dismissed):

webserver/webserver/src/main/java/io/helidon/webserver/ServerRequest.java
```diff
@@ -142,6 +143,13 @@
      */
     Tracer tracer();
 
+    /**
+     * Closes the connection.
+     *
+     * @return Single completed when channel is closed.
+     */
+    Single<Void> close();
```
tomas-langer, line 151: I think this method name is not good. This would mean we are closing the request, where in fact it would close the underlying connection. I think `closeConnection` may be a better name, as it clearly states what is the intention of it.
This is the same for all the classes this was added to

tomas-langer: Also the documentation should clearly state what does this actually mean - is it immediately closing the connection? Is it adding `Connection: close` header? Will it close the connection after response is sent?
All mean "close connection", each in a very different way

tomas-langer (changes requested):

webserver/webserver/src/main/java/io/helidon/webserver/ServerRequest.java
danielkec: fixed

tomas-langer (approved):
LGTM

Issue #1426, opened 2020-02-21
EtcdConfigSourceBuilder should accept more than one URL


ottuzzi:
## Environment Details
* Helidon Version: 1.4.1
* Helidon SE or **Helidon MP**
* JDK version: 11
* OS: Whatever
* Docker version (if applicable):

----------

## Problem Description
As ETCD is a distributed cluster by design I would expect the etcd integration built in Helidon would accept many URLs (one per cluster node): if that ETCD node is down we loose etcd connection.

## Steps to reproduce
Both documentation and API do not support passing more than ONE URLs

spericas:
@ottuzzi Is this still a problem in your setup? Which version of Etcd are your using? We are trying to prioritize this work (yes, it's been a while) and would like to understand the use case more. Any additional info that you can provide about _your use case_ would be of great help. Thanks.

ottuzzi:
Hi,

   etcd is distributed by nature so it is expected that in a production setup you have at least 3 nodes or more.
Let's call those nodes with hostname:

- etcd1.mydomain.com
- etcd2.mydomain.com
- etcd3.mydomain.com

Actually in helidon API to connect to an etcd backend you can set only **one** name-host-URL and let's say I use etcd1.mydomain.com: what happens if, for any reason, etcd1.mydomain.com becomes unavailable? 
As etcd is distributed I think that helidon API should accept a list of etcd nodes (for fault tolerance) and not a single one like it is today: also in the etcd4j library you are using for integration this is an expected behavior (see the following code from their own github page[1]).
```
try(EtcdClient etcd = new EtcdClient(
    URI.create("http://123.45.67.89:8001"),
    URI.create("http://123.45.67.90:8001"))){
  // Logs etcd version
  System.out.println(etcd.getVersion());
}
```

I hope this explanation is clearer than original one... if in doubt please ask :)

*Ciao
Piero
[1]https://github.com/jurmous/etcd4j#setup

spericas:
@ottuzzi Thanks for the clarification. I'm I correct in assuming that you're using Etcd V2?

ottuzzi:
Hi,

   yes we use ETCD V2 Protocol.

spericas:
Fixed by #3381. Should be in 2.4.0.

PR #3381, opened 2021-09-07
Added support for multiple URIs to Config Etcd API

spericas:
Implementation is currently supported only for V2 which is based on Etcd4j. Attempting to pass more than one URI to a V3 factory will result in an exception. This change is source backward compatible since it only replaces `URI` for `URI...`. See #1426.


barchetta (approved):

tomas-langer (approved):
Could you please create a follow up issue to look into etcd v3 and see if there is some way to support it as well (maybe using etcd client library as we do for v2, or using our own load balancing/failover)

PR #3386, opened 2021-09-07
Add test-native-image.sh

barchetta:


romain-grecourt (commented):
LGTM

romain-grecourt (approved):

tomas-langer:
This requires build tools 2.2.4 - there is a problem with a race condition in the `ProcessMonitor` that fails if the process finishes too fast, which happens quite often when showing the `-help` output of the start command of JLink.

Issue #3382, opened 2021-09-07
Additional reactive operators


danielkec:
- [ ]      Duration support
- [x]     `flatMapOptional(Function<? super T, Optional<? extends U>> mapper)`
#3387
- [x]     `flatMapCompletionStage(Function<? super T, CompletionStage<? extends U>> mapper)` 
#3339
- [x]     `ifEmpty(Runnable r)`
#3470
- [x]     `defaultIfEmpty(Supplier<? extends T> supplier)`
#3592
- [ ]     `resumeIfEmpty(Supplier<? extends Flow.Publisher<? extends T>> supplier)`
We already have `switchIfEmpty(Flow.Publisher<T> other)`
- [ ]     `<T extends Throwable> mapError(Class<T> clazz, Function<T,  ? extends Throwable> mapper)`
- [ ]     `mapError(Function<? extends Throwable,  ? extends Throwable> mapper)`
- [x]     `log()` - log operator on Single(already present on Multi)
#3544

Examples:
```java
// Item "test" is inlined, same as flatMapping Single.just("test")
Single.just(Optional.of("test"))
          .flatMapOptional(Function.identity())        
          .await();
> test

// No item is inlined from empty optional, same as flatmapping Single.empty()
Single.just(Optional.<String>empty())
          .flatMapOptional(Function.identity())
          .defaultIfEmpty(() -> "Nothing here!")
          .await();
> Nothing here!

// Single extends CompletableStage, so flatMapCompletionStage gets quite versatile
Single.just("test")
          .flatMapCompletionStage(CompletableFuture::completed)
          .await();
> test

// MapError is basically much faster version of onErrorResumeWith(t -> Single.error(t.getCause()))
Single.error(new IllegalStateException("custom err", new NullPointerException("cause npe")))
          .mapError(IllegalStateException.class, ise -> ise.getCause())
          .onErrorResume(Throwable::getMessage)
          .await();
> cause npe

// ifEmpty in the typical webserver usecase of empty stream
rules.get("/test", (req, res) ->
                DbClient.create(config)
                        .execute(exec -> exec.query("SELECT ID FROM SOME_TABLE"))
                        .map(dbRow -> dbRow.column(1).as(String.class))
                        .first()
                        .ifEmpty(() -> res.status(404).send())
                        .onError(res::send)
                        .forSingle(res::send));
```

danielkec:
Not relevant any more

PR #3387, opened 2021-09-08
Multi flatMapOptional

danielkec:
```java
// Item "test" is inlined, same as flatMapping Single.just("test")
Single.just(Optional.of("test"))
          .flatMapOptional(Function.identity())        
          .await();
> test

// No item is inlined from empty optional, same as flatmapping Single.empty()
Single.just(Optional.<String>empty())
          .flatMapOptional(Function.identity())
          .defaultIfEmpty(() -> "Nothing here!")
          .await();
> Nothing here!
```
See #3382 for more info


common/reactive/src/test/java/io/helidon/common/reactive/MultiFlatMapOptionalTest.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ */
+package io.helidon.common.reactive;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+import org.junit.jupiter.api.Test;
+
+public class MultiFlatMapOptionalTest {
```
tomas-langer, line 24: I think this should also have tests when the originating stream is not a stream of optionals, such as:
```
     Multi.range(1, 5)
                .flatMapOptional(i -> (i % 2) == 0 ? Optional.of(i) : Optional.<Integer>empty()))
                .subscribe(subscriber);
```
Which should work as well (and should cut the middle man of converting the actual value to optional).
This is valid for all tests I think

danielkec: Not sure I follow, you mean something like this?
```java
        Multi.range(1, 5)
                .flatMapOptional(Optional::of)
                .subscribe(subscriber);
```

tomas-langer: your `flatMapOptional(Function.identity())` is redundant

danielkec: fixed

tomas-langer (dismissed):
LGTM

tomas-langer (approved):
LGTM

PR #3399, opened 2021-09-09
Fluent headers.

### LINKED ISSUE #3309: Make ResponseHeaders fluent
We should consider making the methods of ResponseHeaders that return void return ResponseHeaders instead.

Setting headers and status for a response requires two statements.

```
        response.headers()
                .add(Http.Header.LOCATION, "/foo");
        response.status(Http.Status.FOUND_302)
                .send();
```

This would be better:

```
        response
                .status(Http.Status.FOUND_302)
                .headers()
                .add(Http.Header.LOCATION, "/foo")
                .send();
```

romain-grecourt:
Make the add/put methods fluent by return the parameters instance.
Add default methods on ServerResponse to add headers.

Fixes #3309

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/HashParameters.java
+++ b/common/http/src/main/java/io/helidon/common/http/HashParameters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


danielkec (approved):

tomas-langer (dismissed):
LGTM, we should also add methods `addHeader` to `WebClientRequestBuilder` - OK to create a follow up issue.

romain-grecourt:
I guess it's better to update the WebClient API here as well. I'll take a look.

tomas-langer (approved):

danielkec (approved):

Verdent (approved):

PR #3403, opened 2021-09-13
Using Jersey context rather then Helidon context.

### LINKED ISSUE #3402: Tracing fails Jersey async - regression
When a JAX-RS resource returns a `CompletionStage` obtained from a third party library (or not using executor service supporting Helidon Context), the response fails.

Reason:
- there is an attempt to obtain `Context` in response filter, which fails in such a scenario

Fix:
- register the `Scope` with `RequestContext` and with Helidon Context, use the instance from `ReqeustContext` when closing it

tomas-langer:
Resolves #3402 

### DIFF
--- a/tracing/jersey/src/main/java/io/helidon/tracing/jersey/AbstractTracingFilter.java
+++ b/tracing/jersey/src/main/java/io/helidon/tracing/jersey/AbstractTracingFilter.java
@@ -167,12 +167,8 @@ public abstract class AbstractTracingFilter implements ContainerRequestFilter, C
         requestContext.setProperty(SPAN_FINISHED_PROPERTY, true);
         span.finish();
 
-        // using the helidon context is not supported here, as we may be executing in a completion stage future returned
+        // using the helidon context is not supported here, as we may be executing in a completion stage returned
         // from a third party component
-        // Context context = Contexts.context().orElseThrow(() ->
-        //      new IllegalStateException("Context must be available in Jersey"));
-        // context.get(Scope.class).ifPresent(Scope::close);
-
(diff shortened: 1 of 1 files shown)


tracing/jersey/src/main/java/io/helidon/tracing/jersey/AbstractTracingFilter.java
```diff
@@ -166,8 +167,16 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
         requestContext.setProperty(SPAN_FINISHED_PROPERTY, true);
         span.finish();
 
-        Context context = Contexts.context().orElseThrow(() -> new IllegalStateException("Context must be available in Jersey"));
-        context.get(Scope.class).ifPresent(Scope::close);
+        // using the helidon context is not supported here, as we may be executing in a completion stage future returned
+        // from a third party component
+        // Context context = Contexts.context().orElseThrow(() ->
```
spericas, line 172: Do we want to leave the old commented out like this?

tomas-langer: I wanted to have something there, so we do not refactor it to Helidon context. Not sure if all is needed.

spericas: Leave the comment without the old code maybe? "// using the Helidon context Context.context() is not supported here ..."

tomas-langer: ok

spericas (approved):

Issue #3411, opened 2021-09-14
Review dependencies in module helidon-webserver-tyrus


spericas:
Pulling some unwanted dependencies:

```
INFO] |  |  \- org.glassfish.tyrus:tyrus-server:jar:1.17:compile
[INFO] |  |     \- org.glassfish.tyrus:tyrus-container-grizzly-client:jar:1.17:compile
[INFO] |  |        +- org.glassfish.grizzly:grizzly-framework:jar:2.3.22:compile
[INFO] |  |        \- org.glassfish.grizzly:grizzly-http-server:jar:2.3.22:compile
[INFO] |  |           \- org.glassfish.grizzly:grizzly-http:jar:2.3.22:compile
```

spericas:
PR #3412

PR #3412, opened 2021-09-14
Moved some dependencies into test scope only in helidon-webserver-tyrus

spericas:
Issue #3411.


spericas:
```
[INFO] +- org.glassfish.tyrus:tyrus-client:jar:1.17:test
[INFO] +- org.glassfish.tyrus:tyrus-container-grizzly-client:jar:1.17:test
[INFO] |  +- org.glassfish.grizzly:grizzly-framework:jar:2.3.22:test
[INFO] |  \- org.glassfish.grizzly:grizzly-http-server:jar:2.3.22:test
[INFO] |     \- org.glassfish.grizzly:grizzly-http:jar:2.3.22:test
```

dependencies/pom.xml
```diff
@@ -510,10 +510,25 @@
                 <artifactId>tyrus-client</artifactId>
                 <version>${version.lib.tyrus}</version>
             </dependency>
+            <dependency>
+                <groupId>org.glassfish.tyrus</groupId>
+                <artifactId>tyrus-container-grizzly-client</artifactId>
```
romain-grecourt, line 515: Is this dependency meant to be used by end-users ? If not they can be moved to the top level pom.

spericas: They would use it only for testing WS, just like we do now. WDYT?

romain-grecourt: In that case, that is the right place for this.

romain-grecourt (approved):

PR #3420, opened 2021-09-16
Race condition in Microstream example.

tomas-langer:
Race condition was causing requests against uninitialized Microstream root.


Verdent (approved):

spericas:
So this now works because everything is correctly initialized before the constructor returns?

tomas-langer:
yes

PR #3425, opened 2021-09-17
 Fix race condition when analysing whether entity is fully read when …

tomas-langer:
…using Jersey


spericas (approved):
LGTM

romain-grecourt (approved):

Issue #3377, opened 2021-09-06
WebClient gets stuck when connection is closed between sent datachunks


danielkec:
## Environment Details
* Helidon Version: snapshot from master
* Helidon SE
* JDK version: 21.0.0.2.r11-grl, 21.1.0.r16-grl
* OS: Ubuntu

----------

## Problem Description
WebClient should throw `WebClientException: Connection reset by the host` when channel gets closed. 
It does when connection gets closed before chunks are flushed, eg. comment out forcing flush
from reproducer.
```java
// force flush after each chunk
// .flatMap(bb -> Multi.just(bb, DataChunk.create(true)))
```

But it gets stuck when connection is killed between chunks.

## Steps to reproduce
```java
    @Test
    void testCutConnection() throws ExecutionException, InterruptedException, TimeoutException {
        AtomicReference<WebServer> server = new AtomicReference<>();
        server.set(WebServer
                .builder()
                .port(0)
                .routing(Routing.builder()
                        .get((req, res) -> {
                            res.send(Multi
                                    .interval(100, 500, TimeUnit.MILLISECONDS,
                                            Executors.newSingleThreadScheduledExecutor())
                                    .map(i -> "item" + i)
                                    .limit(15)
                                    .peek(s -> {
                                        if ("item2".equals(s)) {
                                            server.get().shutdown();
                                        }
                                    })
                                    .map(String::getBytes)
                                    .map(DataChunk::create)
                                    // force flush after each chunk
                                    .flatMap(bb -> Multi.just(bb, DataChunk.create(true)))
                            );
                        })
                        .build())
                .build()
                .start()
                .await());

        CompletableFuture<Throwable> actualErrorCf = new CompletableFuture<>();

        // Expecting WebClientException: Connection reset by the host
        WebClient.builder()
                .baseUri("http://localhost:" + server.get().port())
                .build()
                .get()
                .request()
                .flatMap(WebClientResponse::content)
                .map(DataChunk::bytes)
                .map(String::new)
                .log()
                .onError(actualErrorCf::complete)
                .ignoreElements();

        Throwable actual = actualErrorCf.get(10, TimeUnit.SECONDS);

        assertThat(actual.getCause(), Matchers.instanceOf(WebClientException.class));
    }
```

spericas:
I'm able to reproduce the problem locally. Analyzing it further as it may be related to #3361.

Verdent:
[...]

PR #3427, opened 2021-09-20
WebClient now throws WebClientException when channel gets closed

spericas:
WebClient now throws WebClientException (Connection reset by the host) when channel gets closed after some data has already been flushed by the server. Integration tests updated to verify the new logic. New test needs to run in its own VM. See issue #3377.


tests/integration/webclient/src/main/java/io/helidon/tests/integration/webclient/GreetService.java
```diff
@@ -201,6 +205,24 @@ private void updateGreetingHandler(ServerRequest request,
...
 
+    private void connectionClose(ServerRequest request, ServerResponse response) {
+        response.send(Multi
+                .interval(100, 500, TimeUnit.MILLISECONDS,
+                        Executors.newSingleThreadScheduledExecutor())
+                .map(i -> "item" + i)
+                .limit(15)
+                .peek(s -> {
+                    if ("item2".equals(s)) {
+                        Main.webServer.shutdown();        // abrupt shutdown to detect connection close
```
danielkec, line 216: Test can be made little more elegant with `ServerRequest#closeConnection` #3380

spericas: I think I like the abruptness of the shutdown to simulate a server going down unexpectedly. I will try the close too anyway.

tests/integration/webclient/pom.xml
```diff
@@ -93,4 +93,49 @@
         </dependency>
     </dependencies>
 
+    <build>
```
Verdent, line 96: I think, it should be safe to remove this. I have tested to run the tests without it and it runs as expected. Could you please double-check?

spericas: It depends in the order in which tests are executed, which can vary. Was ConnectionClose executed at the end of your run? Otherwise, it can shut down the server midstream.

Verdent (approved):
LGTM

PR #3439, opened 2021-09-28
OCI ATP Integration Part 3

arjav-desai:
This PR is for reminder of work for #3072 where wallet has utility methods and example shows how to access DB using the wallet info.


examples/integrations/oci/atp-cdi/src/main/java/io/helidon/examples/integrations/oci/atp/cdi/AtpResource.java
```diff
@@ -50,10 +46,15 @@
     private static final Logger LOGGER = Logger.getLogger(AtpResource.class.getName());
 
     private final OciAutonomousDb autonomousDb;
+    private final PoolDataSource atpDataSource;
+    private final String atpServiceName;
 
     @Inject
-    AtpResource(OciAutonomousDb autonomousDb) {
+    AtpResource(OciAutonomousDb autonomousDb, @Named("atp") PoolDataSource atpDataSource,
+                @ConfigProperty(name = "oracle.ucp.jdbc.PoolDataSource.atp.serviceName") String atpServiceName) {
```
ljnelson, line 54: I'm reading this maybe without context, but now your datasource name _has_ to be `atp` in configuration or this won't work.  In JPA (which I know is not what you're doing), there's an extra layer of indirection: the `persistence.xml` names the datasource, and then that name is used to look up all the JDBC configuration.  You might consider something similar here: maybe have a (meta-, really) configuration item named something like `dataSourceName` whose default value is `atp`, and then use MicroProfile Config programmatically to look up what you need using the value of that property as necessary.

examples/integrations/oci/atp-cdi/src/main/java/io/helidon/examples/integrations/oci/atp/cdi/AtpResource.java
```diff
@@ -68,34 +69,28 @@ public Response generateWallet() {
...
-                //while ((bytesRead = zipStream.read(byteBuff)) != -1) {
-                //    out.write(byteBuff, 0, bytesRead);
-                //}
-                //out.close();
-                zipStream.closeEntry();
+            this.atpDataSource.setSSLContext(walletArchive.getSSLContext());
+            this.atpDataSource.setURL(walletArchive.getJdbcUrl(this.atpServiceName));
+            try(Connection connection = this.atpDataSource.getConnection();
+            PreparedStatement ps = connection.prepareStatement("SELECT 'Hello world!!' FROM DUAL");
+            ResultSet rs = ps.executeQuery()) {
```
ljnelson, line 84: Nit: the formatting here threw me off.

examples/integrations/oci/atp-cdi/src/main/resources/application.yaml
```diff
@@ -21,7 +21,17 @@
 server:
   port: 8080
 
+oracle:
+  ucp:
+    jdbc:
+      PoolDataSource:
+        atp:
+          connectionFactoryClassName: oracle.jdbc.pool.OracleDataSource
+          serviceName: "${atp.db.serviceName}"
```
ljnelson, line 30: Aha; regarding my comment above, I see it gets tricky here (here's the context I hadn't yet run across).  I see that you're using `atp` as a fairly important configuration signifier.  Nevertheless, it remains the case that your data source currently needs to be "named" `atp` (the `atp` "part" in `oracle.ucp.jdbc.PoolDataSource.atp.something`).  I'll stop short of recommending a change, since it is extraordinarily unlikely that someone will have a data source with that name already, but I think my prior comment is still worth considering.

examples/integrations/oci/atp-cdi/src/main/java/io/helidon/examples/integrations/oci/atp/cdi/AtpResource.java
arjav-desai: User can name it to their liking. I just used "atp" in this example's config so using that.

arjav-desai: This is example code/resource, we don't rely on "atp" datasource config in our code.

examples/integrations/oci/atp-cdi/src/main/resources/application.yaml
arjav-desai: This is example code/resource, we don't rely on "atp" datasource config in our code.

examples/integrations/oci/atp-cdi/src/main/java/io/helidon/examples/integrations/oci/atp/cdi/AtpResource.java
arjav-desai: Fixed.

examples/integrations/oci/atp-reactive/pom.xml
```diff
@@ -43,13 +43,30 @@
...
+            <artifactId>helidon-dbclient-jdbc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.oracle.database.jdbc</groupId>
+            <artifactId>ojdbc8-production</artifactId>
+            <type>pom</type>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-reactive</artifactId>
```
danielkec, line 60: Common reactive probably already is on the classpath

examples/integrations/oci/atp-reactive/src/main/java/io/helidon/examples/integrations/oci/atp/reactive/AtpService.java
```diff
@@ -52,36 +57,53 @@ public void update(Routing.Rules rules) {
...
+        return new JdbcDbClientProvider().builder()
+                .connectionPool(() -> {
+                    try {
+                        return pds.getConnection();
+                    } catch (SQLException e) {
+                        throw new IllegalStateException("Error while setting up new connection", e);
                     }
                 })
-                .exceptionally(res::send);
+                .build();
```
danielkec, line 107: Returning Single from `createDbClient` can help with those ugly checked exceptions
```java
Single<DbClient> createDbClient(...
try{
...
} catch (SQLException e) {
  LOGGER...
  return Single.error(e);
}
return Single.just(new JdbcDbClientProvider().builder()..connectionPool(() -> {...}).build());
```
Eg.:
```java
.flatMap(maybeArchive ->
    maybeArchive.map(archive -> 
          createDbClient(archive).execute(exec -> exec.query("SELECT 'Hello world!!' FROM DUAL"))
                    ).orElseGet(() -> {
                            res.status(404).send();
                            return Multi.empty();
                    })
        )
```

danielkec: that 404 is exactly the usecase for `ifEmpty` #3382, we will have it soon

examples/integrations/oci/atp-reactive/pom.xml
arjav-desai: Removed this dependency.

examples/integrations/oci/atp-reactive/src/main/java/io/helidon/examples/integrations/oci/atp/reactive/AtpService.java
arjav-desai: Thanks, updated code to be more "reactive".

danielkec (approved):

PR #3440, opened 2021-09-28
Remove uses of SecurityManager and AccessController dependencies 

spericas:
Remove SecurityManager and AccessController dependencies in preparation for their deprecation in JDK 17. Changes in GRPC core package left out because they require changes to public classes/methods. Will revisit this for 3.0.0.

### DIFF
--- a/microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/ModelHelper.java
+++ b/microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/ModelHelper.java
@@ -23,7 +23,6 @@ import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
-import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Objects;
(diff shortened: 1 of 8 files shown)


microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/ModelHelper.java
```diff
@@ -90,56 +89,45 @@
...
+     * Get collection of methods declared on given class.
      *
      * @param clazz class for which to get the declared methods.
-     * @return privileged action to obtain methods declared on the {@code clazz} class.
-     * @see java.security.AccessController#doPrivileged(java.security.PrivilegedAction)
+     * @return methods declared on the {@code clazz} class.
      */
-    public static PrivilegedAction<Collection<? extends Method>> getDeclaredMethodsPA(final Class<?> clazz) {
-        return () -> Arrays.asList(clazz.getDeclaredMethods());
+    public static Collection<? extends Method> getDeclaredMethods(final Class<?> clazz) {
```
tomas-langer, line 97: This seems to be a public class and public method. I think it was intended for our use, but not sure. 
As this is a backward incompatible change - we should either document (and explain) then change as an intended break that must be fixed between minor version, or create separate methods without `PrivilegedAction` and mark these as deprecated, or mark these methods as deprecated and change the signature in 3.0.0

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/ModelHelper.java
```diff
@@ -90,56 +89,45 @@
...
-                return cls.getMethod(m.getName(), m.getParameterTypes());
-            } catch (final NoSuchMethodException e) {
-                for (final Method method : cls.getMethods()) {
-                    if (method.getName().equals(m.getName())
-                            && method.getParameterTypes().length == m.getParameterTypes().length) {
-                        if (compareParameterTypes(m.getGenericParameterTypes(),
-                                method.getGenericParameterTypes())) {
-                            return method;
-                        }
+    public static Method findMethodOnClass(final Class<?> cls, final Method m) {
```
tomas-langer, line 116: Same as above - backward incompatible change

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/ModelHelper.java
spericas: I think they are just public for use outside of `core` in other parts of the GRPC implementation. My preference would be to document and change rather than go through the whole deprecation dance for no practical reason.

tomas-langer (approved):
LGTM

PR #3441, opened 2021-09-28
Allow metrics-capable components to work in absence of full-featured metrics

### LINKED ISSUE #2180: Metrics support in other modules
We need to add metrics for various components in Helidon.
Currently we use two approaches:
1. We simply depend on `metrics` module (example: `helidon-microprofile-fault-tolerance`)
2. We have an integration module that depends both on the module and on metrics, this requires SPI for the new module being integrated (example: `helidon-webclient`, `helidon-dbclient`)

It would be useful to come up with a small footprint module with no-op implementation of metrics that could be used by any module that supports metrics. This module would use a service loader SPI to discover actual metrics implementation (currently `helidon-metrics`. If none is found, it would use the no-op approach.

As part of this change, we should add configurable way to disable specific metrics (similar to what MP has for base metrics)

tjquinno:
Resolves #2180 

This PR creates a new component, `helidon-metrics-api`, which abstracts the behavior of registry factories, metrics registries, and metrics implementations to use "no-op" metrics if:
- `helidon-metrics` is not on the runtime path
- overall metrics is disabled via `MetricsSupport.Builder` or config

Further, if a _metrics-capable_ component's metrics usage is disabled via its own builder or config, that component uses no-op metrics.

See the `README.md` in the new component for more details how metrics-capable components can be revised in small ways to take advantage of this new feature.

Here is a summary of the changes by component:

### New component `helidon-metrics-api`

- New `RegistryFactory` interface and related service loading interface `RegistryFactoryProvider` interface and loader class.
- New `BaseMetric` interface defining common behavior of full-featured and no-op metrics.
- New ‘AbstractRegistry` implements most of `MetricRegistry` with abstract methods for creating metrics instances.
- New interfaces `MetricsSettings` (POJO analog to overall metrics config) and `ComponentMetricsSettings` (POJO analog for metrics-capable component metrics settings) and implementations for both.
- “No-op” implementations for metrics, metric registry, and registry factory.
- New home for existing `KeyPerformanceIndicatorMetricsSettings` interface and impl.
- Internally-focused `README.md` describing the approach and how to write or revise components to use the new approach.

### `helidon-metrics`
- Existing interfaces and classes shrink now that they implement and extend new elements in `helidon-metrics-api`.
- Record `MetricsSettings` instead of the corresponding `Config` (but still support assigning those settings via `Config`)
- Previous `RegistryFactory` static factory methods for obtaining registry factory instances now delegate to `helidon-metrics-api` for backward compatibility and to enable migration over time of existing metrics-capable components.

### `helidon-service-common-rest`
- Add support in `HelidonRestServiceSupport` and its `Builder` for `ComponentMetricsSettings` and those settings in `Config`.

### `helidon-microprofile-metrics`
- Enhanced tests

### `helidon-examples-metrics`
- Update `import`


metrics/api/README.md
```diff
@@ -0,0 +1,206 @@
...
+* `ComponentMetricsSettings` interface and its `Builder` interface
+  
+   Attributes of metrics for 
+  metrics-capable components (currently, 
+  simply whether metrics are to be enabled for the component or not). 
+
+## Converting a metrics-dependent component to a metrics-enabled one
+1. Change `pom.xml` to depend on `helidon-metrics-api` instead of `helidon-metrics`.
+2. Either:
+   1. Change the service to extend `HelidonRestServiceSuport` if it does not already, or
```
ljnelson, line 68: Typo `Suport` -> `Support`

metrics/api/README.md
```diff
@@ -0,0 +1,206 @@
...
+   
+   This method returns a `RegistryFactory` which automatically accounts for:
+   * whether the metrics implementation is available at runtime,
+   * whether the overall metrics configuration has enabled or disabled metrics in general, and
+   * whether the component's metrics settings indicate to use the full-featured metrics
+  implementation (assuming it is available).
+
+Aside from that, leave the metrics-related code in metrics-capable components as-is.
+
+For example, suppose the `helidon-xxx` component creates or updates some metrics. You have two
```
ljnelson, line 83: If this is for application developers then maybe `yourapp-xxx` instead of `helidon-xxx`?

metrics/api/README.md
```diff
@@ -0,0 +1,206 @@
...
+    
+    Metrics are enabled by default. Configuring `metrics.enabled` to `false` disables all 
+   metrics. In that case, even if metrics-capable components request a full-featured metrics 
+   `RegistryFactory` 
+   they will get a no-op one.
+3. Enable or disable metrics for individual components.
+   
+   Metrics-capable components are enabled to use metrics by default. Configuring `xxx.metrics.
+   enabled` to `false` disables metrics for the `XXX` component.
+
```
ljnelson, line 160: Very clear

metrics/api/src/main/java/io/helidon/metrics/api/AbstractMetric.java
```diff
@@ -0,0 +1,56 @@
...
+ * Common reusable implementation for any category of metric implementation (full-featured, no-op).
+ * <p>
+ * Helidon relies on this additional behavior beyond that provided by MP {@code Metric}.
+ * </p>
+ */
+public abstract class AbstractMetric implements BaseMetric {
+
+    private final String registryType;
+    private final Metadata metadata;
+    private boolean isDeleted;
```
ljnelson, line 30: On my context-free checklist of things that I pause on: should this be `volatile`?

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
```diff
@@ -0,0 +1,791 @@
...
+
+    /**
+     * Get internal map entry given a metric name. Synchronized for atomic access of more than
+     * one internal map.
+     *
+     * @param metricName The metric name.
+     * @return Optional map entry..
+     */
+    public synchronized Optional<Map.Entry<? extends Metric, List<MetricID>>> getOptionalMetricWithIDsEntry(String metricName) {
+        final List<MetricID> metricIDs = allMetricIDsByName.get(metricName);
```
ljnelson, line 422: `allMetricIdsByName` is a concurrent structure; maybe use `computeIfAbsent` and remove `synchronized` from the method body (with care obviously)

ljnelson: (Oh, I see the multiple accesses; hmm)

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
```diff
@@ -0,0 +1,808 @@
...
+    public void removeMatching(MetricFilter filter) {
+        allMetrics.entrySet().stream()
+                .filter(entry -> filter.matches(entry.getKey(), entry.getValue()))
+                .map(entry -> remove(entry.getKey()))
+                .reduce((a, b) -> a || b);
+    }
+
+    @Override
+    public SortedSet<String> getNames() {
+        return allMetrics.keySet().stream()
```
ljnelson, line 291: Would `allMetricIDsByName.keySet()` work here?

metrics/metrics/src/main/java/io/helidon/metrics/Registry.java
```diff
@@ -82,763 +51,82 @@ public static Registry create(Type type) {
...
+    protected Map<Class<? extends HelidonMetric>, MetricType> prepareMetricToTypeMap() {
+        final Map<Class<? extends HelidonMetric>, MetricType> result = new HashMap<>();
+        result.put(HelidonConcurrentGauge.class, MetricType.CONCURRENT_GAUGE);
+        result.put(HelidonCounter.class, MetricType.COUNTER);
+        result.put(HelidonGauge.class, MetricType.GAUGE);
+        result.put(HelidonHistogram.class, MetricType.HISTOGRAM);
+        result.put(HelidonMeter.class, MetricType.METERED);
+        result.put(HelidonTimer.class, MetricType.TIMER);
+        result.put(HelidonSimpleTimer.class, MetricType.SIMPLE_TIMER);
+        return result;
```
ljnelson, line 94: (Can't remember. Should this be `Collections.unmodifiableMap(result)`?

metrics/metrics/src/main/java/io/helidon/metrics/Registry.java
```diff
@@ -82,763 +51,82 @@ public static Registry create(Type type) {
...
+                MetricType.METERED, HelidonMeter::create,
+                MetricType.TIMER, HelidonTimer::create,
+                MetricType.SIMPLE_TIMER, HelidonSimpleTimer::create,
+                MetricType.CONCURRENT_GAUGE, HelidonConcurrentGauge::create);
     }
 
-    // For testing
-    static Map<Class<? extends HelidonMetric>, MetricType> metricToTypeMap() {
-        return METRIC_TO_TYPE_MAP;
+    // -- declarations which let us keep the methods in the superclass protected; we do not want them public.
```
ljnelson, line 107: Should they therefore be `final`, since otherwise I can simply subclass _this_ class and make them `public`?

ljnelson (approved):
A couple of comments.

metrics/api/README.md
tjquinno: Gone due to other edits.

metrics/api/README.md
tjquinno: Gone do to other edits.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractMetric.java
tjquinno: Fixed.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
tjquinno: Yes, accesses to multiple data structures, unfortunately. Luckily, registrations of new metrics and removals of existing ones are relatively rare, compared to look-ups, so the odds for contention are low.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
```diff
@@ -0,0 +1,808 @@
...
+        final List<Map.Entry<MetricID, M>> doomedMetrics = getMetricsByName(name);
+        if (doomedMetrics.isEmpty()) {
+            return false;
+        }
+
+        final boolean result = doomedMetrics.stream()
+                .peek(entry -> entry.getValue().markAsDeleted())
+                .map(entry -> allMetrics.remove(entry.getKey()) != null)
+                .reduce((a, b) -> a || b)
+                .orElse(false);
```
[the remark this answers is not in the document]
tjquinno: Replacing with
```
        boolean result = false;
        for (Map.Entry<MetricID, M> doomedMetric : doomedMetrics) {
            doomedMetric.getValue().markAsDeleted();
            result |= allMetrics.remove(doomedMetric.getKey()) != null;
        }
```
Perhaps things here could have been handled a little more cleanly overall, but as it stands the method `getMetricsByName` which supplies the entries in `doomedMetrics` returns a `List`, the `Map.Entry<MetricID, M>` elements of which are not tied to any original map. They are just convenient pairings of a metric ID with its metric which happens to use `Map.Entry` to put them together rather than some private static class with two fields. So `Iterator.remove` is not an option here, at least not as currently implemented.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
```diff
@@ -0,0 +1,808 @@
...
+        M doomedMetric = allMetrics.remove(metricID);
+        if (doomedMetric != null) {
+            doomedMetric.markAsDeleted();
+        }
+        return doomedMetric != null;
+    }
+
+    @Override
+    public void removeMatching(MetricFilter filter) {
+        allMetrics.entrySet().stream()
```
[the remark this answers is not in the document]
tjquinno: Good catch. Fixed.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
tjquinno: Not quite. The keys in the `allMetrics` `Map` are not sorted as the return value must be. 

But I will change it to
```
return new TreeSet<>(allMetricsIDsByName.keySet());
```
which lets the constructor decide how it wants to build the `TreeSet` instead of our code choosing how.

ljnelson: Don't forget to wrap it in `Collections.unmodifiableSortedSet()`.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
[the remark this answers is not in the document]
ljnelson: Oh, OK

metrics/metrics/src/main/java/io/helidon/metrics/Registry.java
tjquinno: Does making them `final` really help?

Yes, it would prevent a user developer from overriding those _exact_ methods with `public` implementations in the subclass, but a developer so determined to expose the information returned from the protected methods on `Registry` could just implement in the subclass some parallel methods which delegate to the protected `Registry` methods, `final` or not.

The goal here was to minimize the declared-as-`public` methods exposed from the new `helidon-metrics-api` module, just to help reduce the _blatantly_ public surface area that might confuse developers _using_ the API `Registry` interface. These methods are used (all of them, I think) internally within the package from `MetricsSupport` or tests in `helidon-metrics`.

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
tjquinno: Hmm. The contract for `getNames` in the MP Metrics `MetricRegistry` interface doesn't indicate that the result will be immutable. Plus, our prior implementation returns a mutable set, so although you might have a good point returning an immutable instance would be a potentially backward-incompatible change.

metrics/api/src/main/java/module-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+module io.helidon.metrics.api {
+
+    requires java.logging;
+
+    requires io.helidon.common.serviceloader;
+    requires io.helidon.config;
+
+    requires microprofile.metrics.api;
```
tomas-langer, line 23: this dependency should be transitive (as it is part of its public API)

metrics/api/src/main/java/module-info.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+module io.helidon.metrics.api {
+
+    requires java.logging;
+
+    requires io.helidon.common.serviceloader;
+    requires io.helidon.config;
```
tomas-langer, line 21: this dependency should be transitive (as it is part of public API)

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
```diff
@@ -0,0 +1,808 @@
...
+
+    protected Type registryType() {
+        return type;
+    }
+
+    protected List<MetricID> metricIDsForName(String metricName) {
+        return allMetricIDsByName.get(metricName);
+    }
+
+    static <T extends Metadata, U extends Metadata> boolean  metadataMatches(T a, U b) {
```
tomas-langer, line 489: double space after return type

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
```diff
@@ -0,0 +1,808 @@
...
+        if (a == b) {
+            return true;
+        }
+        if (a == null || b == null) {
+            return false;
+        }
+        return a.getName().equals(b.getName())
+                && a.getTypeRaw().equals(b.getTypeRaw())
+                && (a.getDisplayName().equals(b.getDisplayName())
+                && Objects.equals(a.getDescription(), b.getDescription())
```
tomas-langer, line 499: there is a misleading end bracket or misleading formatting.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsSettings.java
```diff
@@ -0,0 +1,139 @@
...
+     */
+    static Builder builder() {
+        return new MetricsSettingsImpl.Builder();
+    }
+
+    /**
+     *
+     * @return whether metrics are enabled according to the settings
+     */
+     boolean isEnabled();
```
tomas-langer, line 65: There is an additional space before `boolean`

metrics/api/src/main/java/io/helidon/metrics/api/MetricsSettingsImpl.java
```diff
@@ -0,0 +1,126 @@
...
+            return this;
+        }
+
+        @Override
+        public MetricsSettings.Builder config(Config config) {
+            config.get(MetricsSettings.ENABLED_CONFIG_KEY)
+                    .asBoolean()
+                    .ifPresent(this::enable);
+            Config baseConfig = config.get(MetricsSettings.BASE_CONFIG_KEY);
+                    baseConfig.get(MetricsSettings.ENABLED_CONFIG_KEY)
```
tomas-langer, line 100: wrong formatting. `baseConfig` should be aligned with `Config` above (a new statement)

tomas-langer (changes requested):
Formatting issues. There are a few additional (such as alignment of method parameters) that I did not mark.
I can provide formatting for Idea (XML) or `.editorconfig`

metrics/api/src/main/java/io/helidon/metrics/api/NoOpRegistryFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.EnumMap;
+import java.util.stream.Stream;
+
+import org.eclipse.microprofile.metrics.MetricRegistry;
+import org.eclipse.microprofile.metrics.MetricRegistry.Type;
+
+/**
+ * No-op implementation of {@link RegistryFactory}.
+ */
+public class NoOpRegistryFactory implements RegistryFactory {
```
tomas-langer, line 27: Do we need this class to be public?

metrics/api/src/main/java/module-info.java
tjquinno: fixed

metrics/api/src/main/java/module-info.java
tjquinno: fixed

metrics/api/src/main/java/io/helidon/metrics/api/AbstractRegistry.java
tjquinno: fixed

metrics/api/src/main/java/io/helidon/metrics/api/MetricsSettingsImpl.java
tjquinno: Noted; I've changed this code in the meantime already.

PR #3447, opened 2021-09-28
Optimize concurrent gauges to use a shared current-time-in-minutes for all instances which use the system clock

### LINKED ISSUE #3398: Change ConcurrentGauge to use new PeriodicExecutor
## Environment Details
* Helidon Version: 2.x
* Helidon SE or Helidon MP SE
* JDK version:
* OS:

----------

## Problem Description
`ConcurrentGauge` depends on statistics reflecting the previous clock minute. The implementation currently computes the current minute from the current system time in ms on every update to the metric. This  can be costly under heavy load.

As part of PR #3385 which addresses issue #3389, a new class is added to `metrics/metrics` which `ConcurrentGauge` should use to optimize the calculation of current minute.

tjquinno:
Resolves #3398 

Earlier changes added a `PeriodicExecutor` which is used in metrics to optimize updates to current-time-in-seconds values.

This PR adds the use of `PeriodicExecutor` by the `ConcurrentGauge`  implementation which relies on a current-time-in-minutes value.

In real apps, virtually all `ConcurrentGauges` will use the system clock which means they can all share the same current-time-in-minutes value, which further means we can enroll only a single updater with `PeriodicExecutor`. 

This PR adds that optimization to `ConcurrentGauge` while preserving the existing behavior (which updates the current-time-in-minutes on each metric update) if a given instance _does not_ use the system clock.

### DIFF
--- a/metrics/metrics/src/main/java/io/helidon/metrics/HelidonConcurrentGauge.java
+++ b/metrics/metrics/src/main/java/io/helidon/metrics/HelidonConcurrentGauge.java
@@ -16,6 +16,7 @@
 
 package io.helidon.metrics;
 
+import java.time.Duration;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicLong;
 
@@ -124,14 +125,38 @@ final class HelidonConcurrentGauge extends MetricImpl implements ConcurrentGauge
         super.prometheusType(sb, nameWithUnits, PROMETHEUS_TYPE);
     }
 
+    // For testing
+    AtomicLong currentMinuteRef() {
+        return (delegate instanceof ConcurrentGaugeImpl)
+                ? ((ConcurrentGaugeImpl) delegate).currentMinute
+                : null;
+    }
+
     static class ConcurrentGaugeImpl implements ConcurrentGauge {
+
+        private static final Duration CURRENT_MINUTE_UPDATE_INTERVAL = Duration.ofMillis(500);
+
+        private static final Clock SYSTEM_CLOCK = Clock.system();
+        private static final AtomicLong SYSTEM_CURRENT_MINUTE = new AtomicLong(currentTimeMinute(SYSTEM_CLOCK));
+
+        /*
+         * Most concurrent gauges will use the system clock, and all of them can share the same current-time-in-minutes value.
+         * To update that (likely) common value, we enroll an updater for the static, shared value.
+         * Any concurrent gauge that uses a non-system clock will compute the current-time-in-minutes value on every update.
+         */
+        static {
+            PeriodicExecutor.enroll(ConcurrentGaugeImpl::updateSystemCurrentMinute, CURRENT_MINUTE_UPDATE_INTERVAL);
+        }
+
         private final AtomicLong count;
         private final AtomicLong lastMax;
         private final AtomicLong lastMin;
         private final AtomicLong currentMax;
         private final AtomicLong currentMin;
         private final AtomicLong lastMinute;
+        private final AtomicLong currentMinute;
         private final Clock clock;
+        private final boolean usingSystemClock;
 
         ConcurrentGaugeImpl(Clock clock) {
             this.clock = clock;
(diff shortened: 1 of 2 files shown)


metrics/metrics/src/main/java/io/helidon/metrics/HelidonConcurrentGauge.java
```diff
@@ -183,17 +210,31 @@ public synchronized void dec() {
         }
 
         public synchronized void updateState() {
-            long currentMinute = currentTimeMinute();
-            long diff = currentMinute - lastMinute.get();
+            if (!usingSystemClock) {
+                updateCurrentMinute();
+            }
+            long diff = currentMinute.get() - lastMinute.get();
```
ljnelson, line 216: Not sure this has to be in the `synchronized` block, right?

metrics/metrics/src/main/java/io/helidon/metrics/HelidonConcurrentGauge.java
```diff
@@ -183,17 +210,31 @@ public synchronized void dec() {
...
             }
         }
 
         private long currentTimeMinute() {
-            return clock.milliTime() / 1000 / 60;
+            return currentTimeMinute(clock);
+        }
+
+        private static long currentTimeMinute(Clock clock) {
+            return clock.milliTime() / 1000 / 10;
```
olotenko, line 229: Is this deliberately in 10-seconds, not minutes?

metrics/metrics/src/main/java/io/helidon/metrics/HelidonConcurrentGauge.java
```diff
@@ -183,17 +210,31 @@ public synchronized void dec() {
...
+        }
+
+        private long updateCurrentMinute() {
+            return currentMinute.getAndSet(currentTimeMinute());
```
olotenko, line 233: It seems no one consumes the value returned by this `getAndSet`. Use `set`. In this context `lazySet` is also fine.

metrics/metrics/src/main/java/io/helidon/metrics/HelidonConcurrentGauge.java
```diff
@@ -183,17 +210,31 @@ public synchronized void dec() {
         }
 
         public synchronized void updateState() {
```
olotenko, line 212: I do not quite get the point of `synchronized` here and in both `inc` and `dec`.

`synchronized` `inc` makes sense only if all accesses to `count` and `currentMax` are also `synchronized`, including reads, in which case we don't need them to be `AtomicLong`. But I presume they are `AtomicLong` to enable atomic updates that are cheaper than `synchronized`. In which case we don't need `synchronized`.

Atomic updates should be like so:

```
public void inc() {
   long count = count.incrementAndGet();
   if (count > currentMax.get()) currentMax.getAndUpdate(v -> count > v ? count: v);
}
```

`updateState` should probably be done only by the minute-updating routine.

And to check the intention: the `currentMin` and `currentMax` don't get reset after every minute intentionally?

olotenko (changes requested):

tjquinno:
Thanks to the reviewers, but...

I am closing this PR without merging, at least for now.

A test in the TCK schedules some updates to a `ConcurrentGauge` at precisely the top of a minute. The scheme in this PR for updating the shared current-time-in-minutes does not run (most likely) until after those updates to the metric, so the update logic incorrectly sees the update as happening in the previous minute, not the current one.

IMO, ideally the TCK test would schedule its work for 1 ms _after_ the start of the new minute. That way, I could revise the update scheme to schedule its update for exactly the start of each minute. For use cases other than the edge case where an update is scheduled at precisely the top of a minute, that would seem sufficient. 

But that would require a change in the TCK test, and MP might be heading towards adopting Micrometer for its metrics solution going forward so it seems fruitless to file an issue against the TCK under the circumstances.

Therefore, for now at least we'll just have to tolerate the performance hit from computing the current-time-in-minutes on each update.

PR #3449, opened 2021-09-29
Propagate supplier exceptions even on a timeout in FT

spericas:
Propagate supplier exceptions even on a timeout. They are recorded as a cause to the TimeoutException reported.


tomas-langer (approved):
LGTM

PR #3450, opened 2021-09-29
Adds support for the limited customization and configuration permitted by the UCP

ljnelson:

### DIFF
--- a/integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java
+++ b/integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java
@@ -20,6 +20,7 @@ import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Method;
+import java.sql.SQLException;
 import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
@@ -102,7 +103,7 @@ public class UCPBackedDataSourceExtension extends AbstractDataSourceExtension {
             .produceWith(instance -> {
                     try {
                         return createDataSource(instance, dataSourceName, dataSourceProperties);
-                    } catch (final IntrospectionException | ReflectiveOperationException exception) {
+                    } catch (final IntrospectionException | ReflectiveOperationException | SQLException exception) {
                         throw new CreationException(exception.getMessage(), exception);
                     }
                 })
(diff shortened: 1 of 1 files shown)


ljnelson:
This PR does what it can to get as much configuration "down" to the underlying `DataSource` wrapped by the `PoolDataSource` implementation (the `PoolDataSource` calls this underlying data source a "connection factory").

Connection factory customization can only be accomplished through the `PoolDataSource` facilities, since there is no way to pass a fully configured connection factory to a `PoolDataSource`.  Unfortunately those facilities are very limited.  This PR works as much as possible within their constraints.

One bit of behavior introduced by this PR as a side effect of working within those constraints is that properties destined for the connection factory that are unknown to it will cause pool initialization to fail.  In general, this is what you would want, but strictly speaking before all this you could set whatever properties you wanted and they would ALL be ignored.  Now only those properties accepted by either (a) the `PoolDataSource` itself or (b) the underlying connection factory will be valid.

Finally, I had to special case the `serviceName` `PoolDataSource` property, which is a property used internally by the `PoolDataSource` implementation itself (it is passed to an instance of the [`UCPConnectionBuilder`](https://javadoc.io/static/com.oracle.database.jdbc/ucp/21.3.0.0/oracle/ucp/jdbc/UCPConnectionBuilder.html) regardless of whether the underlying database driver is Oracle-based or not with every `getConnection` request).  The only way to cause this property to be initialized otherwise appears to be via some sort of undocumented XML configuration that is probably relevant only in certain application server scenarios that do not concern Helidon, so this special casing should be harmless.

integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java
```diff
@@ -119,42 +121,152 @@ protected final void addBean(final BeanConfigurator<DataSource> beanConfigurator
...
+                    Method m = returnValue.getClass().getDeclaredMethod("setPdbRoles", Properties.class);
+                    if (m.trySetAccessible()) {
+                        m.invoke(returnValue, pdbRoles);
+                    }
+                } catch (final NoSuchMethodException ignoreOnPurpose) {
+
+                }
+            }
+        }
+        final Instance<SSLContext> sslContextInstance = instance.select(SSLContext.class);
```
arjav-desai, line 264: Can we scope/restrict this to somehow related to datasource being produced here?

ljnelson: Yeah, that's a good idea.

PR #3457, opened 2021-09-30
OIDC - redirect URL frontend-uri built from req if left out of config

jmntn2000:
In a multi-tenant environment or any instance of multiple host names resolving to same IP the frontend-uri can not be hard coded in the config file for OIDC redirect.  If this is left out of the config then it is now build from the request.


tomas-langer:
Hello @jmntn2000, we have a parallel PR in progress (#3456) that needs a similar change. If you would be OK with it, I would take over your work and finish it once that PR is merged. I will keep your commit in to preserve the contribution.

jmntn2000:
> 
> 
> Hello @jmntn2000, we have a parallel PR in progress (#3456) that needs a similar change. If you would be OK with it, I would take over your work and finish it once that PR is merged. I will keep your commit in to preserve the contribution.

That is fine with me.  Thank you.

PR #3460, opened 2021-10-02
add io_uring support

duplexsystem:
pretty much ctrl+c, ctrl+v, ctrl+r (epoll -> iouring) of the epoll transport impl. let me know if it needs any changes. io ring might have some quirks that mean the worker count needs to be restricted but I see that as unlikely

### DIFF
--- a/webserver/transport/netty/iouring/pom.xml
+++ b/webserver/transport/netty/iouring/pom.xml
@@ -1,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2021 Oracle and/or its affiliates.
+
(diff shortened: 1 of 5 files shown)


duplexsystem:
Also closes #2670

duplexsystem:
Also the OCA link is not found so if there is another place to sign it let me know

romain-grecourt:
We need to update the link in the README, see https://oca.opensource.oracle.com/

duplexsystem:
> We need to update the link in the README, see https://oca.opensource.oracle.com/
signed

romain-grecourt:
Something was off.. Things should be better now but I'll keep monitoring just in case.

PR #3596, opened 2021-10-28
Guide: JBatch with Helidon

ljamen:
Let's check with Dmitry re: the location of the sample code currently located here:  https://github.com/dalexandrov/helidon-jbatch

Should examples be moved to Helidon repo?


dalexandrov (changes requested):

ljamen:
See #3895

PR #3607, opened 2021-10-31
Redesign the per-application OpenAPI processing

### LINKED ISSUE #3587: Incorrect OpenAPI output in MP app with multiple Application classes
## Environment Details
* Helidon Version: 2.x
* JDK version:
* OS:

----------

## Problem Description
If an MP app contains more than one `Application` class, and at least one JAX-RS endpoint uses a POJO class for a parameter, then the OpenAPI output does not correctly include the component definition for the POJO class or the reference to it elsewhere in the document.

If the app contains zero or one `Application` then this error does not occur.

## Steps to reproduce
* Generate the Helidon MP QuickStart app.
* Add a new `Greeting` POJO class with a `greeting` private `String` field and a public getter and setter.
* Add these annotations to the `getDefaultMessage` and `getMessage` methods of the `GreetingResource` class:
   ```
   @Operation(description = "Returns a generic greeting")
   @APIResponse(responseCode = "200",
              content = @Content(schema = @Schema(implementation = Greeting.class)))
   ```
* Create an `Application` class for the greeting resource.
  ```
  @ApplicationScoped
  public class GreetApp extends Application {
  
      @Override
          return Set.of(GreetResource.class);
      }
  }
  ```
* Build and run the app and access the `/openapi` endpoint:
  ```
  mvn clean package
  ```
  In another window:
  ```
  curl http://localhost:8080/openapi
  ```
  and observe that the output includes:
  * a components section for the `Greeting` POJO:
    ```
    components: 
      schemas:
        Greeting: 
          properties:
            greeting: 
              type: string
          type: object
    ```
  * a reference to that component from the two `GET` endpoints:
    ```
      /greet: 
        get: 
          description: Returns a generic greeting
          responses:
            '200': 
              content:
                application/json: 
                  schema: 
                    $ref: '#/components/schemas/Greeting'
              description: OK
    ```
* Now, create another resource class:
   ```
  public class HelloWorldResource {
        private static final JsonBuilderFactory JSON = Json.createBuilderFactory(Collections.emptyMap());
   
      /**
       * Return a worldly greeting message.
       *
       * @return {@link JsonObject}
       */
      @GET
      @Operation(description = "Returns a generic greeting")
      @APIResponse(responseCode = "200",
                   content = @Content(schema = @Schema(implementation = Greeting.class)))
      public JsonObject getDefaultMessage() {
          return createResponse("World");
      }
  
      private JsonObject createResponse(String who) {
          String msg = String.format("Hello Everyone!");
   
           return JSON.createObjectBuilder()
                   .add("message", msg)
                   .build();
      }
  }
  ```
* Create an `Application` class for the new resource:
  ```
  @ApplicationScoped
  public class HelloWorldApp extends Application {
  
      @Override
          return Set.of(HelloWorldResource.class);
     }
  }
  ```
* Build and run the app:
  ```
  mvn clean package
  ```
  The output contains no `components` section, and the information for the `GET` methods looks like this:
  ```
    /greet: 
      get: 
        description: Returns a generic greeting
        responses:
          '200': 
            content:
              application/json: 
                schema: 
                  type: object
            description: OK
  ```
  which is an accurate description of the endpoint but lacks the full information about the `Greeting` POJO which it should have.

tjquinno:
Resolves #3587 

Instead of choosing which classes to _include_ in the annotation scanning by SmallRye for each individual JAX-RS application (which previously overlooked POJO classes used as input or output parameter types to the endpoints) , this new approach chooses which classes to _exclude_ instead. 

The logic is described in the updated JavaDoc in the changed file.

This PR also removes the `beans.xml` from `helidon-microprofile-server`. It is not needed, and it was causing a rogue synthetic JAX-RS `Application` to be incorrectly instantiated early in the prep for jlink/native-image. 

Further, an apparent flaw in the MP OpenAPI TCK tests expects that an `Application` which returns a non-empty set from `getSingletons` but an empty set from `getClasses` will also expose endpoints from other resource classes _not_ returned by `getSingletons`. This requires the OpenAPI document to contradict the actual server behavior (which follows the JAX-RS spec https://jakarta.ee/specifications/restful-ws/3.0/jakarta-restful-ws-spec-3.0.html#servlet). This PR introduces a config property which controls the JAX-RS semantics for OpenAPI processing (defaults to true) and sets it to false for the TCK run.


tjquinno:
Closing, replacing with a PR that does not have the rogue copyright change.

Issue #3577, opened 2021-10-22
Helidon Webclient fails to properly send HTTP PATCH request with Proxy (Getting Illegal Request URI Error)


anand-osvc:
I am using Helidon Webclient class to trigger a PATCH request with an endpoint. The Webclient is configured with a proxy to perform the request. I am getting a 500, Illegal Request URI from the remote server. I have tried replacing the Helidon Webclient with 1) Apache HTTP Client, 2) JDK 11 Http Client Both worked well.

## Environment Details
* Helidon Version: 2.3.4
* Helidon SE
* JDK version: 11
* OS: Oracle Linux/ Mac Big Sur
* Docker version (if applicable): NA

----------

## Problem Description
Helidon web client fails to complete PATCH requests when the client is configured with a proxy. One finding from the debug logs was that the helidon webclient tries to send the absolute uri in the request payload where as apache client was sending only the relative url.

2021.10.13 11:07:46 FINEST io.helidon.webclient.NettyClientInitializer$ClientNettyLog Thread[helidon-client-0,10,main]: [id: 0xa7c46a33] CONNECT: aiapps-dp-fn.dq.lan/10.89.161.71:443

2021.10.13 11:07:46 FINEST io.helidon.webclient.WebClientRequestBuilderImpl Thread[helidon-client-0,10,main]: (client reqID: 5) Channel hashcode -> -1691259188

2021.10.13 11:07:46 FINEST io.helidon.webclient.RequestContentSubscriber Thread[helidon-client-0,10,main]: (client reqID: 5) Message body contains only one data chunk. Setting chunked encoding to false.

2021.10.13 11:07:46 FINEST io.helidon.webclient.NettyClientInitializer$ClientNettyLog Thread[helidon-client-0,10,main]: [id: 0xa7c46a33, L:/10.245.3.101:41006 - R:/100.111.88.146:3128] WRITE: 328B

         -------------------------------------------------

         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |

------------------------------------------------------++---------------

00000000	50 41 54 43 48 20 68 74 74 70 73 3a 2f 2f 61 69	*PATCH https://ai*
00000010	61 70 70 73 2d 64 70 2d 66 6e 2e 64 71 2e 6c 61	*apps-dp-fn.dq.la*
00000020	6e 3a 34 34 33 2f 73 65 72 76 69 63 65 73 2f 72	*n:443/services/r*
[...]

PR #3614, opened 2021-11-03
New flag to force the use of relative URIs in WebClient 

spericas:
New flag to force the use of relative URIs (paths) on all requests, even if proxies are configured. Default is false. See #3577 for a use case.


Verdent (approved):
LGTM

PR #3629, opened 2021-11-08
Should resolve #3627.  Issue is resolve and relative paths.  Default …

### LINKED ISSUE #3627: OIDC loading meta data fails due to path being stripped from identity-uri
## Environment Details
* Helidon Version: 2.4.0
* Helidon SE
* JDK version: JDK-17
* OS: any

----------

## Problem Description
An error is thrown due to failed lookup from .well-known/openid-configuration not existing.  The issue is this path should be appended to the full identity-uri from the config but currently its being appended to just the scheme and host portion.

## Steps to reproduce
oidc:
 ...
 identity-uri:  https://my-oidc.someplace.com/opidc/realm

Throws an exception with the following:
https://my-oidc.someplace.com/.well-known/openid-configuration can't be loaded. 

The correct path should be:
https://my-oidc.someplace.com/opidc/realm/.well-known/openid-configuration

I believe this started with 2.4.0

jmntn2000:
default wellknown path starts with / which means resolve will make it relative to host.  Needs to be current path to identity uri + wellknown path.

See issue #3627

### DIFF
--- a/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcMetadata.java
+++ b/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcMetadata.java
@@ -141,7 +141,7 @@ final class OidcMetadata {
         }
 
         private void load() {
-            URI wellKnown = identityUri.resolve(DEFAULT_OIDC_METADATA_URI);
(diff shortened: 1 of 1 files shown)


romain-grecourt:
Closing as this issue was fixed in a different PR.

PR #3630, opened 2021-11-09
Implement metrics for thread pool suppliers

### LINKED ISSUE #947: Produce metrics for ThreadPoolExecutors
At minimum, create metrics for pool sizes and task counts, maybe active threads (this is more expensive). Additional metrics can be created for known subtypes.

Executors should not be modified, rather metrics should be pulled from them so that any type can be supported. This implies some form of registration (e.g. at the point of installation via a builder, etc.).

tjquinno:
Resolves #947 for 2.x

## Overview
Helidon implements three categories of thread pool supplier, depending mostly on the type of executor service they create,  implemented in `helidon-common-configurable`:
* scheduled (`ScheduleThreadPoolSupplier`)
* server (`ServerThreadPoolSupplier`)
* general (`ThreadPoolSupplier` with no overriding category from the other suppliers)

Beginning with this PR, for each `ExecutorService` created from these thread pool suppliers Helidon will register `Gauge` metrics in the vendor registry. Gauges wrap values which other objects expose. Only when the gauge values are retrieved (either by code retrieving the metric from the registry and querying its value or by an HTTP client accessing the `/metrics` endpoint) are the underlying values read, so the performance drain is minimal.

### Metrics exposed
The classic `ExecutorService` implementations created by Helidon's thread pool suppliers are `ThreadPoolExecutor` subclasses. `ThreadPoolExecutor` exposes several methods which reveal dynamic information about the executor, and Helidon will create these metrics for each one :

| Method                     | Metric name              |
|----------------------------|--------------------------|
| `getActiveCount`             | `active-count`             |
| `getCompletedTaskCount`      | `completed-task-count`     |
| `getPoolSize`                | `pool-size`                |
| `getLargestPoolSize`         | `largest-pool-size`        |
| `getTaskCount`               | `task-count`               |
| `getQueue.remainingCapacity` | `queue.remaining-capacity` |
| `getQueue.size`              | `queue.size`               |

The Loom-based `ThreadPerTaskExecutor` exposes much less:

| Method                     | Metric name              |
|----------------------------|--------------------------|
| `threadCount`             | `thread-count`             |

and so does the new metrics support for these executors.

### Metric identity
Helidon identifies the new executor service metrics by:
* name (`executor-service.xxx` where `xxx` is the metric name from above) and,
* tags:
   * `supplierCategory` - either `scheduled`, `server`, some other user-supplied name if user-created, or `helidon-thread-pool-i` where `i` is an increasing integer (if the supplier name is otherwise unspecified)
   * `supplierIndex` - index of the supplier within a given category
   * `poolIndex` - index of the thread pool within the supplier 

### `helidon-common-configurable` 
The several thread pool suppliers in `helidon-common-configurable` now locate observers, using the service loading mechanism, which implement a new SPI interface. When the suppliers are constructed, and when they create new thread pools (as `ExecutorService`s), they notify the observers.

### `helidon-metrics`
The full-featured metrics component adds an implementation of the new observer interface which registers metrics for each new thread pool created from the thread pool suppliers.

Because the `ThreadPerTaskExecutor` (Loom) is not present in all JDK versions we support, the mechanism for getting the metrics values is different. The existing code in `helidon-common-configurable` uses reflection to obtain the thread-per-task executors, and the metrics support also uses reflection also to create a gauge to expose the thread count for those executors.

### `helidon-microprofile-lra`
In the course of working on this I found a bean that declared an observer on all app-scoped beans being destroyed--including itself which meant CDI created the bean again just to notify it of the shutdown. I changed the method to `@PreDestroy`.

### docs
The metrics guides (SE and MP) already contained a very brief mention of "built-in metrics" in Helidon. They now discuss them a bit more (the predefined base metrics for the JVM, the KPI metrics, and now the thread pool ones).

## Example metrics output

```
"executor-service.active-count;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 0,
    "executor-service.completed-task-count;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 0,
    "executor-service.largest-pool-size;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 5,
    "executor-service.pool-size;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 5,
    "executor-service.queue.remaining-capacity;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 10000,
    "executor-service.queue.size;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 0,
    "executor-service.task-count;poolIndex=0;supplierCategory=helidon-thread-pool-2;supplierIndex=0": 0,
```


common/configurable/src/main/java/io/helidon/common/configurable/ObserverManager.java
```diff
@@ -0,0 +1,260 @@
...
+            .create(ObserverManager::loadObservers);
+
+    private static final Map<Supplier<? extends ExecutorService>, SupplierInfo> SUPPLIERS = new ConcurrentHashMap<>();
+
+    // A given supplier category can have multiple suppliers, so keep track of the next available index by category.
+    private static final Map<String, AtomicInteger> SUPPLIER_CATEGORY_NEXT_INDEX_VALUES = new ConcurrentHashMap<>();
+
+    private static final Map<ExecutorService, SupplierInfo> EXECUTOR_SERVICES = new ConcurrentHashMap<>();
+
+    // Defer building list until use to avoid loading problems if this JDK does not support ThreadPerTasExecutor.
```
danielkec, line 61: ThreadPerTaskExecutor

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
```diff
@@ -383,6 +383,7 @@ protected void afterExecute(Runnable r, Throwable t) {
     @Override
     public void shutdown() {
         Event.write();
+        ObserverManager.unregisterExecutorService(this);
         super.shutdown();
     }
```
danielkec, line 388: What happens when `shutdownNow()` is called?

common/configurable/src/main/java/io/helidon/common/configurable/spi/ExecutorServiceSupplierObserver.java
```diff
@@ -0,0 +1,133 @@
...
+     * @param methodInvocations method invocation information for retrieving interesting information from the supplier's
+     *                          executor services
+     *
+     * @return the {@code SupplierObserverContext} for the supplier
+     */
+    default SupplierObserverContext registerSupplier(Supplier<? extends ExecutorService> supplier,
+                                                     int supplierIndex,
+                                                     String supplierCategory,
+                                                     List<MethodInvocation> methodInvocations) {
+        return null;
```
danielkec, line 73: Is this intentional?

common/configurable/src/main/java/io/helidon/common/configurable/VirtualExecutorUtil.java
```diff
@@ -98,4 +98,5 @@ private static Method findMethod() throws ReflectiveOperationException {
         }
         throw previous;
     }
+
```
danielkec, line 101: orphaned whitespace

common/configurable/src/main/java/io/helidon/common/configurable/ObserverManager.java
tjquinno: Fixed.

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPool.java
tjquinno: Good point. I've added an override of `shutdownNow`.

common/configurable/src/main/java/io/helidon/common/configurable/spi/ExecutorServiceSupplierObserver.java
tjquinno: It was intentional, originally, but there's not really a good reason to have it that way. This method is used only for the virtual threads environment, but any implementation of the SPI should implement this method as well.

Fixed.

common/configurable/src/main/java/io/helidon/common/configurable/VirtualExecutorUtil.java
tjquinno: Fixed.

danielkec (approved):

Issue #3608, opened 2021-10-31
Helidon maven plugin fails to generate custom JRE on openjdk 17


AlirezaDastyar:
I'm facing error while building docker image with `Docerfile.jlink` using openjdk 17.
## Environment Details
* Helidon Version: 2.3.4
* Helidon MP
* JDK version:  17.0.1
* Maven version: 3.8.3
* OS:  Ubuntu 20.04 (Host OS)
* Docker version (if applicable): 20.10.10
* Docker buildkit: enabled
----------

## Problem Description
While trying to create a docker image with provided Dockerfile.jlink in MP quickstart 
maven plugin fails to generate custom JRE on openjdk 17.
Maven error log:
> #13 78.96 [ERROR] Failed to execute goal io.helidon.build-tools:helidon-maven-plugin:2.3.0:jlink-image (jlink-image) on project helidon-quickstart-mp: Image generation error: io.helidon.build.util.ProcessMonitor$ProcessFailedException: Creating Class Data Sharing archive for helidon-quickstart-mp.jar failed with exit code 134 -> [Help 1]

## Steps to reproduce
1. Create the quick start maven project:
> mvn -U archetype:generate -DinteractiveMode=false \
    -DarchetypeGroupId=io.helidon.archetypes \
    -DarchetypeArtifactId=helidon-quickstart-mp \
    -DarchetypeVersion=2.3.4 \
    -DgroupId=io.helidon.examples \
    -DartifactId=helidon-quickstart-mp \
    -Dpackage=io.helidon.examples.quickstart.mp

2. Replace the base build image in `Dockerfile.jlink` from `maven:3.6.3-jdk-11-slim` to `maven:3.8.3-openjdk-17-slim`
3. Run the docker build command: `docker build -t helidon:test -f Dockerfile.jlink .`

barchetta:
See https://github.com/oracle/helidon-build-tools/issues/537

A work-around is to disable generation of the CDS archive.  See the `addClassDataSharingArchive` parameter for the helidon-maven-plugin: https://github.com/oracle/helidon-build-tools/blob/2.3.0/helidon-maven-plugin/README.md

barchetta:
I have confirmed this is also fixed in an early build of 18  (build 18-ea+21-1341)

PR #3633, opened 2021-11-09
3.0 preparation: change sources to Java 17

tomas-langer:
Source and target are now `17`.
All graalvm related artifacts changed to 17.
(docker images for graalvm are pushed).
Required javadoc fixes.


etc/scripts/includes/pipeline-env.sh
```diff
@@ -67,10 +67,12 @@ if [ -z "${__PIPELINE_ENV_INCLUDED__}" ]; then
 
     if [ -n "${JENKINS_HOME}" ] ; then
         export PIPELINE="true"
-        export JAVA_HOME="/tools/jdk-11.0.12"
+        export JAVA_HOME="/tools/jdk-17"
```
barchetta, line 70: Reminder: we will want to upgrade to `jdk-17.0.3` when it is released. It should have the CDS archive fix for issue #3608 (which apparently our jlink package test is not hitting).

barchetta (approved):

barchetta (approved):

Issue #3590, opened 2021-10-27
OIDC - maybe allow id token for authentication


tomas-langer:
Similar to what was proposed in https://github.com/oracle/helidon/pull/3457

This requires some analysis whether such usage is the right way to go.

jmntn2000:
See pull request #3637 

Tried to build off of the updates from @tomas-langer and add the following functionality to OIDC:

1) Added server side refresh.  If you are using cookies and server side OIDC provider you run into issues when the access token expires.  All javascript related calls from the client will fail due to expired token and js calls for SPA will not redirect to lean on OIDC Provider for new tokens.  To solve this I saved the refresh token encrypted in a separate cookie and when access token used for the request is expired I refresh it and update cookies so there is no disruption to the client (assuming refresh token is still valid).

2) Change the build subject to utilize the id token for principal if available.

jmntn2000:
@tomas-langer Could you please review these OIDC changes so we can look at getting them merged?  Let me know if you see anything you want changed or added and I can get it done. Thanks!

PR #3637, opened 2021-11-12
OIDC updates

jmntn2000:
See issue #3590


tomas-langer:
I will create new PRs for changes mentioned in this PR. I will separate these into two: required https for redirects and support for refresh cookie.

tomas-langer:
see #4222

romain-grecourt:
Closing as this has been fixed as part of #4222

fssouza:
@romain-grecourt https://github.com/helidon-io/helidon/pull/4222 fixed the "https for redirects" but not the "support for refresh cookie". 
I think the support for refresh token is being discussed here https://github.com/helidon-io/helidon/issues/3590 

This are all the items included in this PR:
https://github.com/helidon-io/helidon/issues/3590#issuecomment-966732557

I believe the code in here might still be relevant (maybe a starting point) for this item:

> 1. Added server side refresh. If you are using cookies and server side OIDC provider you run into issues when the access token expires. All javascript related calls from the client will fail due to expired token and js calls for SPA will not redirect to lean on OIDC Provider for new tokens. To solve this I saved the refresh token encrypted in a separate cookie and when access token used for the request is expired I refresh it and update cookies so there is no disruption to the client (assuming refresh token is still valid).

PR #3642, opened 2021-11-15
Checkstyle upgrade

tomas-langer:
Upgrade checkstyle to support `record`.

This upgrade also (finally) correctly enforces javadocs on static methods on interfaces, and public methods on inner public classes. I have fixed all such issues in Helidon sources.


barchetta (approved):
LGTM

PR #3652, opened 2021-11-17
Implement metrics for thread pool suppliers (3.0)

### LINKED ISSUE #3639: Metrics for thread pools (3.x)
3.x tracker for #947

tjquinno:
Resolves #3639 

For description see PR #3630 (2.x)

### DIFF
--- a/common/configurable/src/test/java/io/helidon/common/configurable/ScheduledThreadPoolSupplierTest.java
+++ b/common/configurable/src/test/java/io/helidon/common/configurable/ScheduledThreadPoolSupplierTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/configurable/src/test/java/io/helidon/common/configurable/ScheduledThreadPoolSupplierTest.java
+++ b/common/configurable/src/test/java/io/helidon/common/configurable/ScheduledThreadPoolSupplierTest.java
@@ -25,10 +25,13 @@ import io.helidon.config.Config;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
+import static org.hamcrest.CoreMatchers.hasItem;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.CoreMatchers.startsWith;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasKey;
+import static org.hamcrest.Matchers.not;
 
 /**
  * Unit test for {@link ScheduledThreadPoolSupplier}.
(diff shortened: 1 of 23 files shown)


microprofile/lra/jax-rs/src/main/java/io/helidon/microprofile/lra/NonJaxRsResource.java
```diff
@@ -202,7 +200,8 @@ private void sendResponse(ServerResponse res, Response response) {
         }
     }
 
-    void terminate(@Observes @BeforeDestroyed(ApplicationScoped.class) Object event) {
+   @PreDestroy
+    void terminate() {
```
tomas-langer, line 204: Why is this changed?

tjquinno: In my local tests, I was seeing the class instantiated twice. From the stack traces in the debugger is seemed that CDI was instantiating it the second time in order to notify it of the before-destroyed event. Because the logic in the method was simply cleaning up resources allocated by the bean itself (and not really taking action related to the shut-down of other beans), using `@PreDestroy` not only resolved the problem I was seeing but also more closely aligned with the work the method was doing.

tjquinno: @tomas-langer Are you OK with this? Was this your only concern? I would like to merge this PR.

metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
```diff
@@ -153,4 +154,51 @@ void checkKPIDisabledByDefault() {
...
+                .accept(MediaType.TEXT_PLAIN)
+                .path("greet/slow")
+                .submit()
+                .await();
+
+        assertThat("Slow greet access response status", slowGreetResponse.status().code(), is(200));
+
+        WebClientResponse secondMetricsResponse = metricsRequestBuilder
+                .submit()
+                .await();
```
danielkec, line 194: There is some internal timeout in the webclient but it is always better to use `await` with some duration

tjquinno: Fixed.

common/configurable/src/test/java/io/helidon/common/configurable/ScheduledThreadPoolSupplierTest.java
```diff
@@ -38,20 +41,28 @@
...
+    private static ScheduledThreadPoolSupplier configuredSupplier;
+    private static ScheduledThreadPoolSupplier defaultSupplier;
+
+
     @BeforeAll
     static void initClass() {
-        defaultInstance = ScheduledThreadPoolSupplier.create().getThreadPool();
+        ObserverForTesting.clear();
+        defaultSupplier = ScheduledThreadPoolSupplier.create();
+        defaultInstance = defaultSupplier.getThreadPool();
```
danielkec, line 53: This test doesn't seem to be shutting down the executors

danielkec (changes requested):

common/configurable/src/test/java/io/helidon/common/configurable/ScheduledThreadPoolSupplierTest.java
tjquinno: Fixed.

danielkec (approved):

PR #3659, opened 2021-11-18
Prepare RegistryFactory lazily to use the most-recently-assigned MetricsSettings

### LINKED ISSUE #3653: Disabling metrics with full-featured metrics on runtime path does not work
## Environment Details
* Helidon Version: 2.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Disabling metrics (either via config with `metrics.enabled=false` or programmatically via settings) should cause a minimal metrics implementation to be used. Because the `RegistryFactoryManager#INSTANCE` is statically initialized using no config or settings, the default (metrics enabled) is used.

Later, when the instance is updated using `RegistryFactoryManager#getInstance(MetricsSettings)` the settings are indeed updated but on the previously-created full-featured registry. The instance itself is not replaced with a minimal implementation.

## Steps to reproduce
1. Build a QuickStart MP app.
2. Add `metrics.enabled = false` to `microprofile-config.properties`.
3. Start the server and ping the `/metrics` endpoint. You get metrics when you should get a 404.

tjquinno:
Resolves #3653 

The prior  implementation used a static initializer to create a `RegistryFactory` using default `MetricsSettings` (which would imply that metrics would be enabled). Even though subsequent `RegistryFactory#getInstance(MetricsSettings)` calls would correctly communicate if metrics had been disabled, by this time the full-featured registry had already been created and saved as the singleton instance. That would mean we would lose some of the efficiency of the minimal metrics implementation when metrics is configured as disabled.

This PR changes the code to lazily create the singleton on first retrieval, using the latest `MetricsSettings` passed to `RegistryFactory#getInstance()`.

### DIFF
--- a/metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactoryManager.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactoryManager.java
@@ -37,8 +37,13 @@ import io.helidon.metrics.api.spi.RegistryFactoryProvider;
  *     <ul>
  *         <li>The {@code create} methods create a new {@link RegistryFactory} according to the specified settings (or defaults)
  *         but keep no reference to such registry factories.</li>
- *         <li>The {@code getInstance} methods reuse a single {@code RegistryFactory}, initialized using the default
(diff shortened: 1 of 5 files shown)


tomas-langer (approved):
LGTM

Issue #3650, opened 2021-11-16
Retry cannot be cancelled


martin-sladecek:
There is no way to cancel Retry FT handler.
 
## Environment Details
* Helidon Version: 2.4.0
* Helidon SE or Helidon MP
* JDK version: 15
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description
The `Single` instance that comes from `io.helidon.faulttolerance.Retry` `invoke` method has a cancel() method. Calling this method should stop Retry handler, but it just cancels the internal CompletableFuture and keeps on with the retries.

## Steps to reproduce
```

    public static ScheduledExecutorService EXECUTOR_SERVICE = Executors.newScheduledThreadPool(2);

    public static void main(String[] args) throws Exception{
        Single<Void> invoke = Retry.builder()
                .retryPolicy(Retry.JitterRetryPolicy.builder().calls(100).delay(Duration.ofSeconds(1))
                        .jitter(Duration.ZERO).build())
                .overallTimeout(Duration.ofSeconds(2000)).build().invoke(
                        () -> CompletableFuture.runAsync(() -> {
                            try {
                                System.out.println("Executing task");
                                Thread.sleep(1000);
                            } catch (Throwable e) {
                                e.printStackTrace();
                            }
                            throw new RuntimeException("Retry!");
                        }, EXECUTOR_SERVICE));
        invoke.exceptionallyAccept(Throwable::printStackTrace);
;
        Thread.sleep(2000);
        System.out.println("Calling cancel()");
        invoke.cancel();
    }
```

martin-sladecek:
Also, shouldn't all FtHandler implementations actually cancel the provided CompletionStage (if it is a CompletableFuture)?

This is what SingleFromCompletionStage is currently doing for the "source" CompletionStage, however, none of the FtHandler implementations actually call it.

There are at least 2 groups:
1. FtHandlers that use DelayedTask, which actually uses its internal "resultFuture" CompletableFuture as a source for the SingleFromCompletionStage, which means the original CompletionStage will not be cancelled. 
2. TimeoutImpl is actually using the provided CompletionStage as a source
```
    @Override
    public <T> Single<T> invoke(Supplier<? extends CompletionStage<T>> supplier) {
        if (!currentThread) {
            return Single.create(supplier.get(), true)
                    .timeout(timeoutMillis, TimeUnit.MILLISECONDS, executor.get());
```
so I would expect cancel() to be called on that stage in this case, but it is not.

[...]

PR #3680, opened 2021-11-30
Improved support for cancellation of FT handlers

spericas:
See description in Issue #3650.

### DIFF
--- a/fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
+++ b/fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
@@ -18,7 +18,6 @@ package io.helidon.faulttolerance;
 
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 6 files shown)


fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
```diff
@@ -145,6 +147,32 @@ static Throwable cause(Throwable throwable) {
...
+                dependent.completeExceptionally(t);
+            } else {
+                dependent.complete(o);
+            }
+        });
+        dependent.whenComplete((o, t) -> {
+            if (t instanceof CancellationException) {
+                source.toCompletableFuture().cancel(true);
+            }
+        });
```
danielkec, line 172: Maybe using `CompletableFuture#isCancelled` would be safer, it's quite improbable but what if someone does:
```java
CompletionStage<Void> source = new CompletableFuture<>();
CompletableFuture<Void> dependant = new CompletableFuture<>();
createDependency(source, dependant);
dependent.completeExceptionally(new CancellationException());
```

danielkec (approved):

PR #3692, opened 2021-12-02
Support MP OpenAPI 3.0

### LINKED ISSUE #2638: Open API 3.0
- [ ] Open API 3.0 support
- [ ] Documentation
- [ ] Samples
- [ ] Guides

MicroProfile 4.0 support epic: #2636

### LINKED ISSUE #3638: OpenAPI redundant additional properties (3.x)
3.x tracker for #3624

tjquinno:
Resolves #2638 
Resolves #3638 

Adopt MP OpenAPI 3.0.

The changes are segregated by commit with explanatory comments with each.

### DIFF
--- a/openapi/src/test/java/io/helidon/openapi/TestUtil.java
+++ b/openapi/src/test/java/io/helidon/openapi/TestUtil.java
@@ -116,6 +116,22 @@ public class TestUtil {
         return actualMT;
     }
 
+    static MediaType connectAndConsumePayload(
+            int port, String path, String queryParameter, MediaType expectedMediaType) throws Exception {
+        HttpURLConnection cnx = getURLConnection(port, "GET", path, queryParameter);
+        MediaType actualMT = validateResponseMediaType(cnx, expectedMediaType);
+        if (actualMT.test(MediaType.APPLICATION_OPENAPI_YAML) || actualMT.test(MediaType.APPLICATION_YAML)) {
+            yamlFromResponse(cnx);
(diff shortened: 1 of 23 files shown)


openapi/src/test/java/io/helidon/openapi/TestUtil.java
```diff
@@ -293,6 +309,18 @@ public static HttpURLConnection getURLConnection(
...
 
+    static HttpURLConnection getURLConnection(
+            int port,
+            String method,
+            String path,
+            String queryParameter) throws Exception {
+        URL url = new URL("http://localhost:" + port + path + "?" + queryParameter);
+        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+        conn.setRequestMethod(method);
+        System.out.println("Connectiong: " + method + " " + url);
```
tomas-langer, line 320: Please remove the system out. Also do we need to use HttpURLConnection?

tomas-langer (changes requested):

openapi/src/test/java/io/helidon/openapi/TestUtil.java
tjquinno: I have removed the `System.out`. 

The new `HttpURLConnection` code follows the same pattern as the code that was already in the test; I used the same approach in the new test so it would be very similar to what is already there for ease of understanding by readers.

I know we have more recent approaches for creating client connections in new tests. I avoided rewriting the test as part of this PR to avoid changes not related to the Open API spec version update.

tomas-langer (approved):

Issue #3648, opened 2021-11-15
Provide config override of servicename for OCI ATP Service Integration with Wallet Management


paulparkinson:
Currently, the OCI ATP Service Integration with Wallet Management feature automatically uses the [servicename]"_high" service and overriding this serviceAlias must be done programmatically. Would be convenient to be able to provide that in config instead.

## Environment Details
* Helidon Version: 2.4.0
* Helidon MP
* JDK version: GraalVM 20.2.0
* OS:
* Docker version (if applicable): ghcr.io/graalvm/graalvm-ce:21.3.0

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

## Steps to reproduce
Use https://github.com/oracle/helidon/tree/master/examples/integrations/oci/atp-cdi and attempt to connect to  [servicename]"_tp" or any service that is not "_high" . Services provided by ADB are high, low, medium, tp, and tpurgent.

ljnelson:
Hi, @paulparkinson; you may be [in luck](https://github.com/oracle/helidon/blob/c51cf34df1c7aa4cebe165c13525bcd492912288/integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java#L182-L192).  If I'm reading you right, that's [in 2.4.0 today](https://github.com/oracle/helidon/blob/c51cf34df1c7aa4cebe165c13525bcd492912288/integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java#L232-L247).

arjav-desai:
We have removed adding "_high" by default to provided service name. Now, User provides the full service name instead.

PR #3711, opened 2021-12-08
Fix serviceName usage for OCI ATP integration

arjav-desai:
#3648


integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java
```diff
@@ -223,6 +223,10 @@ private static PoolDataSourceImpl createDataSource(final Instance<Object> instan
             }
             final Object serviceName = connectionFactoryProperties.remove("serviceName");
             final Object pdbRoles = connectionFactoryProperties.remove("pdbRoles");
+            // Used for OCI ATP Integration
+            // Removing this so that it is not set on connectionFactoryProperties,
+            // Else we get exception with getConnection using this DS, if its set.
+            connectionFactoryProperties.remove("tnsNetServiceName");
```
ljnelson, line 229: Let me think about a better way to do this so there is a minimum of "magic" properties in the generic extension.

ljnelson: Looks like there's not given that `tnsNetServiceName` is not a Java Bean property recognized by the underlying data source and that it (I guess) throws an `IllegalArgumentException` in the presence of bad properties here, which normally would be a feature, not a bug, but if we were previously relying on this…anyway, carry on.

ljnelson (approved):

integrations/cdi/datasource-ucp/src/main/java/io/helidon/integrations/datasource/ucp/cdi/UCPBackedDataSourceExtension.java
arjav-desai: Thanks @ljnelson for the review. Will look to see if adding observer for PoolDataSource in ATP integration can remove this special handling in extension.

PR #3722, opened 2021-12-10
SignedJwt's parseToken() expects characters from base64 instead of ba…

klustria:


klustria:
The goal of this change is to resolve issue reported in https://github.com/oracle/helidon/issues/3675

tomas-langer (approved):

PR #3740, opened 2021-12-14
SignedJwt's parseToken() expects characters from base64 instead of ba64URL encoding

klustria:


klustria:
This is a back port to helidon-2x

arjav-desai (approved):
This is a port of already approved change. LGTM.

PR #3744, opened 2021-12-16
Upgrades H2 to 2.0.206.

ljnelson:


ljnelson:
Note: https://github.com/h2database/h2database/issues/2886#issuecomment-693578610 and the surrounding issue.

barchetta (approved):

ljnelson:
Heads up that this "passes" only because tests are currently disabled on `master` in this area due to the switch to Jakarta namespace: 

https://github.com/oracle/helidon/blob/00fd2b41cfa158b4449312baa9841c5c6e1362ff/integrations/cdi/jpa-cdi/pom.xml#L35-L36

PR #3755, opened 2021-12-16
Handle creation of InjectionManager when parent is a HelidonInjectionManager

spericas:
Handle creation of InjectionManager when parent is a HelidonInjectionManager which can happen with multiple application subclasses and dynamic features.

### DIFF
--- a/webserver/jersey/src/main/java/io/helidon/webserver/jersey/HelidonHK2InjectionManagerFactory.java
+++ b/webserver/jersey/src/main/java/io/helidon/webserver/jersey/HelidonHK2InjectionManagerFactory.java
@@ -60,16 +60,16 @@ public class HelidonHK2InjectionManagerFactory extends Hk2InjectionManagerFactor
             LOGGER.finest(() -> "Creating injection manager " + result);
         } else if (parent instanceof ImmediateHk2InjectionManager) {        // single JAX-RS app
             result = (InjectionManager) parent;
-            LOGGER.finest(() -> "Using injection manager " + result);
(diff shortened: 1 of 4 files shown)


webserver/jersey/src/main/java/io/helidon/webserver/jersey/HelidonHK2InjectionManagerFactory.java
```diff
@@ -67,6 +67,9 @@ public InjectionManager create(Object parent) {
             result = new HelidonInjectionManager(forApplication, wrapper.injectionManager, wrapper.application);
             LOGGER.finest(() -> "Creating injection manager " + forApplication + " with shared "
                     + wrapper.injectionManager);
+        } else if (parent instanceof HelidonInjectionManager) {
+            result = (InjectionManager) parent;
+            LOGGER.finest(() -> "Using injection manager " + result);
```
ljnelson, line 72: Duplicates lines 62–63 exactly; was that your intent? If so, maybe add this `if` condition as an `||` on line 61?

spericas: No intent, was just being lazy. Will fix.

spericas: Actually, I will update the log messages line and keep them separate as it separates the cases nicely

tjquinno (approved):

tjquinno (approved):

PR #3756, opened 2021-12-16
Propagate any existing server context into a Webclient reactive code (helidon-2.x)

spericas:
This PR is about Helidon's context propagation to non-request threads (typically Netty threads). It makes the current context available to any reactive code handler in WebClient and also ensures context is propagated if `req.next()` is called inside Helidon in a different thread. Customer has verified that this PR fixes their application. See Issue #3734 for a discussion.


webclient/webclient/src/main/java/io/helidon/webclient/SingleWithContext.java
```diff
@@ -0,0 +1,546 @@
...
+import java.util.function.Supplier;
+import java.util.logging.Level;
+
+import io.helidon.common.context.Context;
+import io.helidon.common.context.Contexts;
+import io.helidon.common.reactive.CompletionAwaitable;
+import io.helidon.common.reactive.Multi;
+import io.helidon.common.reactive.Single;
+
+class SingleWithContext<T> implements Single<T> {
```
danielkec, line 44: Shorter version could look like this:
```java
    public <T> Single<T> singleWithContext(Single<T> single, Context context) {
        return Single.create(subscriber -> single.subscribe(new Flow.Subscriber<T>() {
            @Override
            public void onSubscribe(Flow.Subscription subscription) {
                Contexts.runInContext(context, () -> subscriber.onSubscribe(subscription));
            }

            @Override
            public void onNext(T item) {
                Contexts.runInContext(context, () -> subscriber.onNext(item));
            }

            @Override
            public void onError(Throwable throwable) {
                Contexts.runInContext(context, () -> subscriber.onError(throwable));
            }

            @Override
            public void onComplete() {
                Contexts.runInContext(context, subscriber::onComplete);
            }
        }));
    }
```

danielkec: Could be tweaked for propagating also to the upstream over the backpressure, but that could lead to unexpected/unwanted effects on different kinds of publishers

danielkec (approved):

PR #3775, opened 2021-12-21
Fix MP archetypes

tvallin:

### DIFF
--- a/archetypes-v2/common/files/pom.xml.mustache
+++ b/archetypes-v2/common/files/pom.xml.mustache
@@ -15,7 +15,9 @@
     <name>{{name}}</name>
 
     <properties>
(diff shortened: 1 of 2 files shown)


romain-grecourt (approved):

romain-grecourt:
I merged your PR so that you are unblocked. However the snippet below is weird.

```
<map key="mainClass">
    <value key="package" template="mustache">{{package}}</value>
</map>
```

"package" is always defined as a top-level model, It seems that you doing this here to do some kind of conditional.

If you simply defined the "mainClass" model under se, then `{{#mainClass}}<mainClass>{{.}}</mainClass>{{/mainClass}}` can be used to implement a conditional.

Issue #3784, opened 2022-01-05
All ThreadPools should customize name


barchetta:
We are pretty good about adding custom thread prefixes for the ThreadPools we create, but we are not good about setting a custom pool name. This matters when setting `io.helidon.common.configurable.ThreadPool.level=FINE` and looking at logs. For example:

```
2022.01.05 12:24:22 FINE io.helidon.common.configurable.ThreadPool Thread[helidon-1,5,server]: ThreadPool 'helidon-thread-pool-2' {corePoolSize=10, maxPoolSize=50, queueCapacity=10000, growthThreshold=1000, growthRate=0%, averageQueueSize=0.00, peakQueueSize=0, averageActiveThreads=0.00, peakPoolSize=0, currentPoolSize=0, completedTasks=0, failedTasks=0, rejectedTasks=0}
```

WTF is helidon-thread-pool-2 ?

I did a quick scan and here are some classes where we are not setting a custom name. There might be others:

```
FaulToleranceExtension
FaulTolerance
MessagingImpl
AqConnector
JmsConnector
TyrusSupportMp
JaxRsClient
AsyncExecutorProvider
```

Also, here are a few places that set the thread pool name, but all set it to "server". And do not set a thread prefix. I assume only one of these pools are ever created, but it might still be beneficial to use unique names, and have thread prefixes:

```
JerseySupport
ServerCdiExtension
Server
```

spericas:
I didn't know about the importance of the thread pool name for logging. Seems to me there should be some correlation between the thread pool name and the thread name prefix. One should be derived from the other if both are not specified. Is there really a use case for these names to be completely unrelated?

spericas:
A simple one-line improvement would be this:

```
--- a/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
+++ b/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
@@ -170,7 +170,7 @@ public final class ThreadPoolSupplier implements Supplier<ExecutorService> {
         @Override
         public ThreadPoolSupplier build() {
             if (name == null) {
-                name = DEFAULT_POOL_NAME_PREFIX + DEFAULT_NAME_COUNTER.incrementAndGet();
+                name = threadNamePrefix + "thread-pool-" + DEFAULT_NAME_COUNTER.incrementAndGet();
             }
             if (rejectionHandler == null) {
[...]

PR #3789, opened 2022-01-07
Ensure all thread pools created by Helidon are named

spericas:
This is necessary for debugging and performance analysis of logs. All pools are now named and default names are improved for application-created pools where only a thread name prefix is provided. A couple of methods to create thread pools without naming them are now deprecated. See #3784.

### DIFF
--- a/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
+++ b/common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolSupplier.java
```diff
@@ -171,7 +198,8 @@ private Builder() {
         @Override
         public ThreadPoolSupplier build() {
             if (name == null) {
-                name = DEFAULT_POOL_NAME_PREFIX + DEFAULT_NAME_COUNTER.incrementAndGet();
+                LOGGER.warning("Neither a thread name prefix nor a pool name specified");
```
barchetta, line 201: Is this message correct? If I specify a thread name prefix but no pool name won't this message still be logged?

spericas: Good catch, looks like I removed an if statement by accident.

barchetta (approved):

PR #3790, opened 2022-01-10
Upgrades H2 to 2.0.206 (helidon-2.x backport)

ljnelson:

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -61,7 +61,7 @@
         <version.lib.graphql-java.extended.scalars>15.0.0</version.lib.graphql-java.extended.scalars>
         <version.lib.grpc>1.35.0</version.lib.grpc>
         <version.lib.guava>30.0-jre</version.lib.guava>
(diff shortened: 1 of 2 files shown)


ljnelson:
`helidon-2.x` backport of #3744.

ljnelson:
This is currently failing because H2 changed its grammar (!) and the underlying JPA providers issue now syntactically invalid statements.  This is true on `master` as well but was not caught in #3744 because tests are disabled in this area due to the Jakarta package migration issues.  This may be an involved fix since both Hibernate and Eclipselink providers will either need to be upgraded (assuming that they too have run into this problem) or database platforms for each provider will have to be written.

ljnelson:
See also:

https://github.com/hibernate/hibernate-orm/blob/7d30b57f15617f679a20aa1389c9385433e45b2c/hibernate-core/src/main/java/org/hibernate/dialect/identity/H2IdentityColumnSupport.java#L18-L27

ljnelson:
Closing in favor of #4298.

PR #3824, opened 2022-01-19
1.x: Upgrade grpc-java to 1.41.2

barchetta:
In 1.42.0 they removed `CallCredentials2` which we use. So we stick with 1.41 for now.


romain-grecourt (approved):

thegridman (commented):
This looks ok. I don't have permissions to approve a PR, but I would have approved it.

PR #3827, opened 2022-01-20
1.x: native-image: remove initialize-at-run-time for Slf4JLoggerFactory

barchetta:


ljnelson (approved):
LGTM; it would be nice in the abstract to know why, since this line was added in response to GraalVM's saying, at some point, that it needed to be added, this line must now be removed

PR #3830, opened 2022-01-21
Fix wrong connection close

### LINKED ISSUE #3829: Invalid connection close on post requests spanning multiple chunks
When post request contains more than two chunks over the network, we assume it is wrong and close the connection (it works, but we send `Connection: close`)

tomas-langer:
Resolves #3829 

Entity is consumed for any request that completed and did not subscribe (including JAX-RS).
We were consuming the entity anyway at a later stage (when we already did a connection close). For connection:close cases, entity is ignored if not consumed.

This keeps original handling and improved it even for requests with more than one chunk in request entity, while simplifying the code.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -56,7 +56,6 @@ import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
  * The BareResponseImpl.
  */
 class BareResponseImpl implements BareResponse {
-
     private static final Logger LOGGER = Logger.getLogger(BareResponseImpl.class.getName());
 
     // See HttpConversionUtil.ExtensionHeaderNames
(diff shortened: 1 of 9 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -195,23 +194,24 @@ public void writeStatusAndHeaders(Http.ResponseStatus status, Map<String, List<S
...
+                        entityRequested = false;
+                    }
+                }
+                if (entityRequested) {
+                    HttpRequest request = requestContext.request();
+                    LOGGER.warning("Entity was requested and not fully consumed before a response is sent. "
+                                           + "This is not supported. Connection will be closed. Please fix your route for "
+                            + request.method() + " " + request.uri());
+                }
+                if (!isWebSocketUpgrade) {
```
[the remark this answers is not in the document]
danielkec: Maybe moving the expression to the `RequestContext` would be best:
```java
    boolean partiallyDrainedStream() {
        return !requestCompleted()
                && !requestCancelled()
                && (hasRequests() || hasEmitted());
    }
```

danielkec (changes requested):

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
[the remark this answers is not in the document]
tomas-langer: Fixed

danielkec (approved):

webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java
```diff
@@ -220,8 +239,8 @@ public void writeStatusAndHeaders(Http.ResponseStatus status, Map<String, List<S
 
         // Content length optimization attempt
         if (!lengthOptimization) {
-            LOGGER.fine(() -> log("Writing headers %s", status));
             requestEntityAnalyzed = requestEntityAnalyzed.thenApply(listener -> {
+                LOGGER.fine(() -> log("Writing headers %s", status));
```
spericas, line 243: Given the assignment in the line above, shouldn't we renamed `requestEntityAnalyzed`? Its name is too close to the new one added, yet its completion reflects more than just the request entity being processed?

tomas-langer: Not sure I understand - it is the same future. We just need to keep the original instance and complete on it, as the "last" isntance in the field will not reflect `thenApply` and other methods registered on previous instances.

spericas: That's exactly the point, the last instance on this field is related to the writing of the response and no longer the analysis of the request entity (as its name suggests). My comment is just about naming, not semantics. Certainly not a stopper.

spericas (approved):

PR #3831, opened 2022-01-24
Follow up to fix for issue 3801

spericas:
Follow up to fix for issue 3801. New version ensures properties are correctly in native image as well.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java
@@ -163,8 +163,10 @@ class NettyWebServer implements WebServer {
             bootstraps.put(name, bootstrap);
         }
 
-        // Log entry that also initializes NettyInitializer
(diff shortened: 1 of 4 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/NettyWebServer.java
```diff
@@ -171,6 +162,9 @@
 
             bootstraps.put(name, bootstrap);
         }
+
+        // Log entry that also initializes NettyInitializer
+        LOGGER.fine(() -> NettyInitializer.getMaxOrderProperty() + " set to " + NettyInitializer.getMaxOrderValue());
```
wojtask9, line 167: Please don't do that. 
JVM loads classes when bytecode is reached (sometimes eager but this depends)

Be explicit and not depending on "log level"

barchetta (dismissed):

barchetta (approved):

dalexandrov:
Confirmed: works in native image!

PR #3849, opened 2022-01-28
Fix dbclient threading issues when DML operations are executed multiple times in a tight loop

klustria:
Goals of the change:
1. Make ConnectionPool.connection() run on the same thread as the execution of the DML statement to avoid threadPool and connectionPool exhaustion to fix an issue reported in https://github.com/oracle/helidon/issues/2484.
2. Add unit test that runs each of the DML operations (insert, update, delete)  with and without transaction, in a tight loop of 100 iterations while  mocking connectionPool to attempt to reproduce the problem.

### DIFF
--- a/dbclient/jdbc/src/test/java/io/helidon/dbclient/jdbc/JdbcClientMultipleDMLOperationTest.java
+++ b/dbclient/jdbc/src/test/java/io/helidon/dbclient/jdbc/JdbcClientMultipleDMLOperationTest.java
@@ -16,18 +16,18 @@
  */
 package io.helidon.dbclient.jdbc;
 
-import java.io.InputStream;
-import java.io.Reader;
-import java.math.BigDecimal;
-import java.net.URL;
 import java.sql.*;
-import java.sql.Date;
 import java.util.*;
-import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.logging.Logger;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.anyString;
 
 import io.helidon.dbclient.DbClientException;
+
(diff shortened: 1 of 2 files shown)


dbclient/jdbc/src/test/java/io/helidon/dbclient/jdbc/JdbcClientMultipleDMLOperationTest.java
```diff
@@ -0,0 +1,1006 @@
...
+                                    String.format("Unable to acquire connection after %d sec", timer.getTimeout()));
+                        }
+                        Thread.sleep(100);
+                    } catch (InterruptedException e) {
+                        throw new DbClientException(
+                                String.format("Unable to acquire connection due to InterruptedException: %s", e.getMessage()));
+                    }
+                }
+                // Create a mocked up Connection that returns a PreparedStatement
+                conn = new Connection() {
```
tomas-langer, line 221: This seems like using `Mockito` would be good, as you only need to implement a few methods. It also allows you to validate what was called.

tomas-langer (changes requested):

dbclient/jdbc/src/test/java/io/helidon/dbclient/jdbc/JdbcClientMultipleDMLOperationTest.java
klustria: I initially wanted to use Mockito, but the few implemented methods need to do more than just return a value. For example, Connection.close() would be doing this:

```
                    @Override
                    public void close() throws SQLException {
                        if (usedConnections.remove(this)) {
                            connectionPool.add(this);
                        }
                        LOGGER.fine(() ->
                                String.format("After Connection.close() metrics: connectionPool=%d, usedConnections=%d",
                                connectionPool.size(), usedConnections.size()));
                    }
```

I thought Mockito can't do such? But I will look at it some more to see if it is possible.

tomas-langer (approved):

Issue #3850, opened 2022-01-28
WebClient memory leak in long-lived connections


lambdaupb:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 2.4.1
* Helidon SE
* JDK version: 17
* OS: macOS / linux

----------

## Problem Description
WebClient instances leak memory on long lived connections, as frequently seen with rest api clients.

Note the 50k instances:
```
 num     #instances         #bytes  class name (module)
-------------------------------------------------------
   2:         50000         800000  io.helidon.webclient.NettyClientHandler$$Lambda$402/0x0000000800d9a648
```

```
Jan 28, 2022 6:14:47 PM io.helidon.common.HelidonFeatures features
INFO: Helidon SE 2.4.1 features: [Config, WebServer]
Jan 28, 2022 6:14:47 PM io.helidon.webserver.NettyWebServer lambda$start$7
INFO: Channel '@default' started: [id: 0xb8a1d02d, L:/127.0.0.1:57239]
0
Jan 28, 2022 6:14:48 PM io.netty.util.ResourceLeakDetector reportTracedLeak
SEVERE: LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:402)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:179)
	io.netty.buffer.AbstractByteBufAllocator.ioBuffer(AbstractByteBufAllocator.java:140)
	io.netty.channel.DefaultMaxMessagesRecvByteBufAllocator$MaxMessageHandle.allocate(DefaultMaxMessagesRecvByteBufAllocator.java:120)
	io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:150)
	io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:722)
	io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:658)
	io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:584)
[...]

PR #3853, opened 2022-01-31
Do not create close listener handlers for every new request

spericas:
Do not create listener handlers for every new request resulting in a memory leak. Use using channelInactive() instead. Issue #3850.

### DIFF
--- a/webclient/webclient/src/main/java/io/helidon/webclient/NettyClientHandler.java
+++ b/webclient/webclient/src/main/java/io/helidon/webclient/NettyClientHandler.java
@@ -231,7 +231,7 @@ class NettyClientHandler extends SimpleChannelInboundHandler<HttpObject> {
         super.channelInactive(ctx);
         // Connection closed without last HTTP content received. Some server problem
         // so we need to fail the publisher and report an exception.
-        if (!responseCloser.isClosed()) {
+        if (publisher != null && !responseCloser.isClosed()) {
             WebClientException exception = new WebClientException("Connection reset by the host");
             publisher.fail(exception);
         }
--- a/webclient/webclient/src/main/java/io/helidon/webclient/NettyClientHandler.java
+++ b/webclient/webclient/src/main/java/io/helidon/webclient/NettyClientHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lambdaupb:
Wouldn't that have the same issue with connections that have a steady rate of requests, like polling a service once per second?

spericas:
> Wouldn't that have the same issue with connections that have a steady rate of requests, like polling a service once per second?

Despite its name, that method is called only when a channel is closed.

lambdaupb:
> > Wouldn't that have the same issue with connections that have a steady rate of requests, like polling a service once per second?
> 
> Despite its name, that method is called only when a channel is closed.

Exactly.

if I make one million (and more) request at a rate of 1/s the there will be 1mio listeners (and growing).

or is this a per-request sub channel?

spericas:
> > > Wouldn't that have the same issue with connections that have a steady rate of requests, like polling a service once per second?
> > 
> > 
> > Despite its name, that method is called only when a channel is closed.
> 
> Exactly.
> 
> if I make one million (and more) request at a rate of 1/s the there will be 1mio listeners (and growing).
> 
> or is this a per-request sub channel?

Not sure I follow. The listeners that were there before have been removed by this PR; the close event is now handled by the same class as the other connection events.

lambdaupb:
A connection with keep-alive is long lived and the potential requests processed over it is practically unbounded.

Adding a close listener to the long lived connection without removing it upon request completion will lead to unbounded listeners being registered.

The problem wasn't that listeners kept around after a connection was closed (after the keep-alive timeout they went away), but that listeners of long completed futures stayed registered, keeping the corresponding futures from being GCed.

As far as I can tell, I'm not well versed in the async intricacies.

spericas:
> A connection with keep-alive is long lived and the potential requests processed over it is practically unbounded.
> 
> Adding a close listener to the long lived connection without removing it upon request completion will lead to unbounded listeners being registered.

 This is correct. This is why this PR *does not* add those listeners anymore. That code is being removed by this PR.
 
> The problem wasn't that listeners kept around after a connection was closed (after the keep-alive timeout they went away), but that listeners of long completed futures stayed registered, keeping the corresponding futures from being GCed.

 That's right. Again, no more listeners now.

lambdaupb:
Aaah, I misunderstood then. Sorry for the noise.

Thanks for the quick fix.

webclient/webclient/src/main/java/io/helidon/webclient/NettyClientHandler.java
```diff
@@ -236,6 +226,17 @@ protected void channelRead0(ChannelHandlerContext ctx, HttpObject msg) throws IO
...
 
+    @Override
+    public void channelInactive(ChannelHandlerContext ctx) throws Exception {
+        super.channelInactive(ctx);
+        // Connection closed without last HTTP content received. Some server problem
+        // so we need to fail the publisher and report an exception.
+        if (!responseCloser.isClosed()) {
+            WebClientException exception = new WebClientException("Connection reset by the host");
+            publisher.fail(exception);
+        }
```
danielkec, line 237: There is null check in the `channelReadComplete` I would expect same is needed here, previous version captured it from the `channelRead0` context.
```suggestion
        if (publisher != null && !responseCloser.isClosed()) {
            WebClientException exception = new WebClientException("Connection reset by the host");
            publisher.fail(exception);
        }
```

danielkec (approved):

PR #3875, opened 2022-02-08
Native image support for MP.

tomas-langer:
Reintroducing support for native image to 3.x


danielkec (approved):

tjquinno:
Obviously it would be preferable to keep the reduced visibility of the `protected` methods that are becoming `public` just to keep the public surface area of Helidon slightly smaller. But I assume there is no way to give native image more directives to avoid this. 

AFAIK the `protected` versions worked in earlier releases of Helidon. Has something changed recently in the native image infrastructure to require this?

PR #3935, opened 2022-03-04
Add HEAD support to health endpoints (2.x)

### LINKED ISSUE #3716: Add support for head method in Health
We should support also `HEAD` in addition to `GET` - as it is sometimes not really required to return an entity (just the response code)

tjquinno:
Resolves #3716 

Adds `HEAD` support to the health endpoints. Responses to `HEAD` include only the status; no payload describing detailed health results.


spericas (approved):
LGTM

PR #3960, opened 2022-03-13
Issue 3926 - Hocon #includes not handled properly

### LINKED ISSUE #3926: Hocon configs does not support include
## Environment Details
* Helidon Version:  2.4.1
* JDK version: 17
* OS: mac os Montery

----------

## Problem Description
hocon cofigs directive "include" does not work.

repro: create two config files 1.conf and 2.conf. Add include 1 in 2.

trentjeff:
discoveries:
1. Hocon include is actually enabled by default in Helidon - we thought it was turned off by default but it's not.
2. The config context root directory (our Config.builder()) seems to be causing the includes not to be resolved properly - and in fact will cause surprising errant behavior for the customer.

This PR Fixes #3926 for discoveries above.

@tomas-langer will need to discuss this with you since I went much deeper than I was hoping to go.  I've made notes in the code for specific things you should pay particular attention to.  I'm guessing there are more appropriate ways to handle, but was not sure where to look.  See slack thread for more background.


trentjeff:
replaced with a newer PR

PR #3970, opened 2022-03-16
Disable paths such as /metrics and /health from tracing. 

spericas:
This change applies to SE and MP, and it is mandated by the MP OpenTracing API. It will not collect/report by default any traces for "/health", "/health/{+}", "/metrics", "/metrics/{+}" and "/openapi". Any of these can be manually enabled via config.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
@@ -165,6 +165,10 @@ public abstract class WebTracingConfig {
                     .tracingConfig(TracingConfig.DISABLED)
                     .build());
             addPathConfig(PathTracingConfig.builder()
+                    .path("/health/{+}")
+                    .tracingConfig(TracingConfig.DISABLED)
+                    .build());
+            addPathConfig(PathTracingConfig.builder()
                     .path("/openapi")
                     .tracingConfig(TracingConfig.DISABLED)
                     .build());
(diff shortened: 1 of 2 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
```diff
@@ -141,6 +141,35 @@ Service service() {
...
+                    .path("/metrics")
+                    .tracingConfig(TracingConfig.DISABLED)
+                    .build());
+            // Simplified matching for base, vendor and application paths
+            addPathConfig(PathTracingConfig.builder()
+                    .path("/metrics/{+}")
+                    .tracingConfig(TracingConfig.DISABLED)
+                    .build());
+            addPathConfig(PathTracingConfig.builder()
+                    .path("/health")
```
tjquinno, line 164: Even if the tracing spec does not mention them, there are subpaths for `/health` as well as for `/metrics`: `/health/ready`, `/health/live`, and (in the latest spec but not the one which applies to Helidon 2.x) `/health/started`. (https://download.eclipse.org/microprofile/microprofile-health-4.0/microprofile-health-spec-4.0.html#_appendix_a_rest_interfaces_specifications)

These should probably be excluded by default as well.

tjquinno (commented):
Possibly expand the default list of excluded paths.

webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
spericas: Good point, I forgot about those. Will exclude them as well.

tjquinno (approved):

PR #3981, opened 2022-03-18
Follow up to PR 3921 that adds support for base interfaces as well

spericas:
Follow up to PR 3921 that adds support for base interfaces as well. Integration test was updated.


klustria (approved):
Change is perfect. I have tried it from my POC app and it is working. Hence, I'm approving this.

Issue #3929, opened 2022-03-02
Grant.builder() should extend Grant.Builder<Builder>


martin-sladecek:
## Environment Details
* Helidon Version: 3
* Helidon MP
* JDK version: 11
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description
Grant.builder() should return some class that extends Grant.Builder<Builder> instead of the wildcarded Builder (i.e. there will need to be a special subclass for pure Grant).

Currently, it is not possible to use the Grant.Builder and Role.Builder as a parameter of the same methods (without ugly code).

## Sample method that should be able to accept all Grant builders, including a builder for Grant itself:
```
private static <B extends Grant.Builder<B>> B populateCommonFields(B builder)
```

martin-sladecek:
Ideally, the testAccept should be able to work with the builder normally, like this:
```
    private static <B extends Grant.Builder<B>> B deserializeCommonFields(B builder) {
        return builder.origin("my-origin");
    }
```

spericas:
I'm not sure the type system can handle that without the wildcard.

martin-sladecek:
The way I had to workaround it was to create a special class `GrantBuilder extends Grant.Builder<GrantBuilder>`, would it be possible to do the same?

spericas:
And have `Role.Builder extend GrantBuilder`? The problem with that is that it will fail to create a `Role` (not backward compatible) since `Role.builder().name("xxx").build()` will be of type `Grant` and not `Role` --- the return value of `name` will be `GrantBuilder`.

martin-sladecek:
No, I thought Role.Builder would also have to extend Grant.Builder.

martin-sladecek:
You are right, this doesn't have any nice solution. I guess your solution is the best possible.

spericas:
@martin-sladecek I believe the solution that I proposed does not really improve the status quo (it should compile without any warnings as is). Closing the issue for now, we can revisit later.

PR #3983, opened 2022-03-21
Improve return type of static builder method to avoid compilation warnings

spericas:
Improve return type of static builder method to avoid compilation warnings. Type erasure is backward compatible. Issue #3929.

### DIFF
--- a/security/security/src/main/java/io/helidon/security/Grant.java
+++ b/security/security/src/main/java/io/helidon/security/Grant.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -59,9 +59,11 @@ public class Grant implements AbacSupport, Principal {
      * Creates a fluent API builder to build new instances of this class.
      *
      * @return builder instance
+     * @param <T> type of builder
      */
-    public static Builder<?> builder() {
-        return new Builder<>();
+    @SuppressWarnings("unchecked")
+    public static <T extends Builder<T>> T builder() {
+        return (T) new Builder<T>();
     }
 
     @Override
--- a/security/security/src/test/java/io/helidon/security/BuilderTest.java
+++ b/security/security/src/test/java/io/helidon/security/BuilderTest.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


security/security/src/main/java/io/helidon/security/Grant.java
```diff
@@ -59,9 +59,11 @@ protected Grant(Builder<?> builder) {
      * Creates a fluent API builder to build new instances of this class.
      *
      * @return builder instance
+     * @param <T> type of builder
      */
-    public static Builder<?> builder() {
-        return new Builder<>();
+    @SuppressWarnings("unchecked")
+    public static <T extends Builder<T>> T builder() {
+        return (T) new Builder<T>();
```
tomas-langer, line 66: This cannot be done, as the `builder` method always returns the same instance, which may not fit into the return type

tomas-langer: We should change this to have:
`AbstractBuilder` class that has type definition (`AbstractBuilder<B extends AbstractBuilder<B>`) and the `Builder` class that `extends AbstractBuilder<Builder>` and this method should return `Builder` explicitly.
Then other builders extends `AbstractBuilder` as well.

Not sure if this could be done in a backward compatible way here

tomas-langer: We should change this to have:
`AbstractBuilder` class that has type definition (`AbstractBuilder<B extends AbstractBuilder<B>`) and the `Builder` class that `extends AbstractBuilder<Builder>` and this method should return `Builder` explicitly.
Then other builders extends `AbstractBuilder` as well.

Not sure if this could be done in a backward compatible way here

tomas-langer (changes requested):

security/security/src/main/java/io/helidon/security/Grant.java
spericas: > This cannot be done, as the `builder` method always returns the same instance, which may not fit into the return type

You mean the implementation was incorrect before?

spericas: > We should change this to have: `AbstractBuilder` class that has type definition (`AbstractBuilder<B extends AbstractBuilder<B>`) and the `Builder` class that `extends AbstractBuilder<Builder>` and this method should return `Builder` explicitly. Then other builders extends `AbstractBuilder` as well.
> 
> Not sure if this could be done in a backward compatible way here

Please see the discussion in the issue. The static `builder` method in a subclass cannot return a type that isn't related to that returned in the base class. So if Grant.Builder <: AbstractBuilder and Role.Builder <: AbstractBuilder, then the `builder` method cannot return the corresponding subtypes.

spericas:
@martin-sladecek After reviewing this one more time, I don't think this PR helps at all. Perhaps you can provide another example, but based on our discussion in the issue, the Java type system just doesn't help here even if we change the return type of `build()` to use existential types.

PR #3993, opened 2022-03-22
3.x: Fix #3966 set collectorUri() with URL with no port number adds port 

### LINKED ISSUE #3966: Tracing: set collectorUri() with URL with no port number adds port 9411
## Environment Details
* Helidon Version: 2.4.2

If you create a zipkin TracerBuilder and set the collector uri to a url with no port ("http://localhost/path" for example), the endpoint URL for the Zipkin collector will use port 9411. This is incorrect. If I configure the builder  with a URL that explicitly does not have a port then the endpoint URL should do as I say and not use a port.

This should be the equivalent to setting the collector port number to -1 which means no port number (and which appears to work correctly).

Why would you ever want to force the use of no port number? Why wouldn't you just use 80, or 443, or whatever the default port is for the protocol? Because we have run into a service that rejects HTTP requests in this case -- maybe because the service was not expecting a port number in the `Host:` header. The only way to get the service to accept the request was to ensure no port number in the URL.

To reproduce add this test to `ZipkinTracerBuilderTest` and it fails (because port() returns 9411:
```
    @Test
    void testConfigDefaultsSuppressPortUri() {
        TracerBuilder<?> builder = TracerBuilder.create("unit-test-suppress-port-uri")
                .collectorUri(URI.create("https://localhost/path"));
        ZipkinTracerBuilder zBuilder = (ZipkinTracerBuilder) builder;
        assertThat(zBuilder.port(), is(-1));
    }
```

This looks to be the questionable code: https://github.com/oracle/helidon/blob/799ebd22ceae7019b7a903b482c35635264615e3/tracing/tracing/src/main/java/io/helidon/tracing/TracerBuilder.java#L163

barchetta:
See #3966

### DIFF
--- a/tracing/tracing/src/main/java/io/helidon/tracing/TracerBuilder.java
+++ b/tracing/tracing/src/main/java/io/helidon/tracing/TracerBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


arjav-desai (approved):
LGTM

PR #3999, opened 2022-03-23
Update MediaType parser to handle parameter without value

### LINKED ISSUE #3956: MediaType.parse doesn't handle parameter without value
```java
MediaType.parse("text/html; charset=");
```

throws the following Exception:
```java
java.lang.IllegalStateException: No more elements!
	at io.helidon.common.http.Tokenizer.checkState(Tokenizer.java:172)
	at io.helidon.common.http.Tokenizer.checkState(Tokenizer.java:157)
	at io.helidon.common.http.Tokenizer.previewChar(Tokenizer.java:119)
	at io.helidon.common.http.MediaType.parse(MediaType.java:328)
	... 1 more
```

romain-grecourt:
Use consumeTokenIfPresent instead of consumeToken
Update consumeTokenIfPresent to return null instead of an empty string when matcher is not matched
Add unit test

Fixes #3956


spericas (approved):
LGTM

PR #4011, opened 2022-03-24
Quickstart cleanup, using @HelidonTest in MP.

tomas-langer:
Resolves #4007 for 2.x

Signed-off-by: Tomas Langer <tomas.langer@oracle.com>

Issue #4007, opened 2022-03-24
Remove reactive endpoint from quickstart example


tomas-langer:
The reactive endpoint in quickstart is confusing and should be removed.

tomas-langer:
Merged both into master and 2.x

PR #4014, opened 2022-03-25
Quickstart cleanup, using @HelidonTest in MP. (#4011)

tomas-langer:
* Quickstart cleanup, using @HelidonTest in MP.
* Archetype update to align with quickstart examples.

Forward port of #4007  for 3.x


archetypes/quickstart-se/src/main/resources/src/test/resources/config-profile.yaml
```diff
@@ -0,0 +1,11 @@
+# Use a profile (without a name) to use a random port in tests
```
romain-grecourt, line 1: `archetypes/quickstart-se` does not exist anymore, Templates and files are exclusively located under `archetypes/helidon/src/main/archetype`.

tomas-langer: Removed changes in archetypes

PR #4029, opened 2022-03-28
Moves universal OCI CDI extension to new location

ljnelson:

### DIFF
--- a/examples/integrations/cdi/oci-objectstorage/pom.xml
+++ b/examples/integrations/cdi/oci-objectstorage/pom.xml
@@ -41,7 +41,6 @@
             <groupId>com.oracle.oci.sdk</groupId>
             <artifactId>oci-java-sdk-objectstorage</artifactId>
             <scope>compile</scope>
-            <type>pom</type>
         </dependency>
         <dependency>
             <groupId>org.eclipse.microprofile.config</groupId>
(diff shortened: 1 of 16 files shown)


examples/integrations/cdi/oci-objectstorage/pom.xml
```diff
@@ -51,8 +51,8 @@
```
barchetta, line 51: Can you remove `<type>pom</type>` from above?

ljnelson: At line 44?  I can.  That's been in there forever; it was originally how OCI wanted you to consume the two artifacts it aggregates.  Maybe they've changed that recommendation.

ljnelson: Ah; I see now they've made it an [empty jar file that does the same thing](https://github.com/oracle/oci-java-sdk/tree/master/bmc-objectstorage/bmc-objectstorage-combined).  Let me dig in here; both of these situations seem like perhaps bad ideas.

ljnelson: Looks like the empty jar project (`oci-objectstorage-combined`) has an `artifactId` of `oci-objectstorage` and a type of `jar`, and all it does is pull in transitive dependencies (`oci-objectstorage-extensions`, `oci-objectstorage-generated`).  Nevertheless it does indeed appear to be how OCI wants you to depend on Object Storage artifacts, so I'll remove the `<type>pom</type>` as suggested, and in `dependencies/pom.xml` as well.  Addressed in [75add80](https://github.com/oracle/helidon/pull/4029/commits/75add80219c55d498007c299e46da4680c478ba1).

barchetta (approved):

PR #4090, opened 2022-03-31
New security response mapper mechanism for MP

spericas:
New security response mapper mechanism that enables MP applications to provide custom responses to users when a security problem (e.g., authentication error) is encountered. New integration test.


security/integration/jersey/src/main/java/io/helidon/security/integration/jersey/SecurityFilterCommon.java
```diff
@@ -344,7 +345,11 @@ protected void abortRequest(SecurityFilter.FilterContext context,
             updateHeaders(responseHeaders, responseBuilder);
         }
 
-        if (featureConfig.isDebug()) {
+        // Run security response mappers if available, or revert to old logic for compatibility
+        ServiceLoader<SecurityResponseMapper> loader = ServiceLoader.load(SecurityResponseMapper.class);
```
tomas-langer, line 349: The service loader should be used only  once, and in a static section.
Also please use `HelidonServiceLoader` as it supports prioritization and some basic control through system properties.

tests/integration/security/security-response-mapper/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    <description>
+        Integration test for security response mappers
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.microprofile.bundles</groupId>
+            <artifactId>helidon-microprofile</artifactId>
+        </dependency>
+        <dependency>
```
tomas-langer, line 40: Why is there a dependency on EL implementation? That should not be required by this module at all.

tests/integration/security/security-response-mapper/src/test/java/io/helidon/tests/integration/security/mapper/RestrictedResourceTest.java
```diff
@@ -0,0 +1,64 @@
...
+import io.helidon.microprofile.server.ServerCdiExtension;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+class RestrictedResourceTest {
```
tomas-langer, line 35: If you use `@HelidonTest`, you do not need beans.xml (you can add required beans using `@AddBean`), you also do not need to manually start/stop the container and create a web target (`WebTarget` can be injected into constructor, with the URI of the server prepared)

tomas-langer (changes requested):

security/integration/jersey/src/main/java/io/helidon/security/integration/jersey/SecurityFilterCommon.java
spericas: Yes, good point

tests/integration/security/security-response-mapper/pom.xml
spericas: Copy/paste from the other test. Will fix.

tests/integration/security/security-response-mapper/src/test/java/io/helidon/tests/integration/security/mapper/RestrictedResourceTest.java
spericas: Copy/paste from the other test. Will fix.

tomas-langer (approved):
LGTM

PR #4093, opened 2022-03-31
New security response mapper mechanism (master)

spericas:
New security response mapper mechanism that enables applications to provide custom responses to users when a security problem (e.g., authentication error) is encountered. New integration test.


tomas-langer (approved):
LGTM

PR #4100, opened 2022-04-05
Fixed dbclient integration tests build issues.

Tomas-Kraus:
Fixed build issues to make tests buildable again.

Java VM mode tests are passing for mysql and postgres.


tomas-langer (approved):

tomas-langer:
It would be good if the code was built as part of the pipeline (even if not tested)

PR #4103, opened 2022-04-06
Remove buffered multipart

romain-grecourt:
### API change notes

`WriteableMultiPart` does not implement `MultiPart` since it is being removed and thus the default methods inherited from `MultiPart` are also removed:
- `default Optional<T> field(String name)`
- `default List<T> fields(String name)`
- `Map<String, List<T>> fields()`


tomas-langer (approved):
LGTM

PR #4139, opened 2022-04-19
2.x: possible fix for HttpPipelineTest hang

barchetta:
This is a possible fix for #3301.

The test class `SocketHttpClient` was creating a new `BufferedReader` every time it's `receive()` method was called. This could potentially loose bytes since an undetermined number of bytes could be read and buffered off of the socket, only to be discarded when the `BufferedReader` was tossed on the garbage heap.

Or at least that is my theory. The test seems more stable after this fix.

### DIFF
--- a/webserver/webserver/src/test/java/io/helidon/webserver/utils/SocketHttpClient.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/utils/SocketHttpClient.java
@@ -55,6 +55,7 @@ public class SocketHttpClient implements AutoCloseable {
     private static final Pattern FIRST_LINE_PATTERN = Pattern.compile("HTTP/\\d+\\.\\d+ (\\d\\d\\d) (.*)");
 
     private final Socket socket;
+    private final BufferedReader socketReader;
 
     /**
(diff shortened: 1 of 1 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/utils/SocketHttpClient.java
```diff
@@ -277,7 +278,7 @@ public String receive() throws IOException {
 
             if ("".equalsIgnoreCase(t) && contentLength >= 0) {
                 char[] content = new char[contentLength];
-                br.read(content);
+                socketReader.read(content);
```
ljnelson, line 281: Not entirely paying attention but since this PR crossed my desk, and knowing not very much about the overall context: whenever you call `Reader#read(char[])`, it returns an `int` representing the bytes actually read.  This invocation&mdash;again, no context, just Java developer spidey senses triggering&mdash;throws that number away, so this read may have not completed entirely.  Maybe that's OK and maybe it's not.

spericas (approved):
LGTM

PR #4146, opened 2022-04-20
2.x: Uppgrade grpc-java to 1.45.1

barchetta:


aseovic (approved):
Does Netty version need to be updated as well?

Issue #4155, opened 2022-04-25
Fallback.createMulti does not call correct method in builder


spericas:
There's a copy/paste error in `Fallback<T> createMulti(Function<Throwable, ? extends CompletionStage<T>> fallback)`. We also seem to lack proper testing for this method.

spericas:
PR #4157
PR #4176

PR #4157, opened 2022-04-26
Fixed a few problems with Fallback and Multi's in SE

spericas:
Fixed a few problems with Fallback and Multi's in SE. Updated Fallback implementation to mimic exception nesting/structure of Single for Multi. Issue #4155.

Note that the signature of the `createMulti` method has changed, but seemed incorrect before.


tomas-langer (approved):
LGTM. Indeed the signatures seems wrong, and a test was missing.

Issue #4152, opened 2022-04-22
Support HOCON include without extension


janotav:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 2.5.0
* Helidon SE 
* JDK version: 14
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description

According to [spec](https://github.com/lightbend/config/blob/master/HOCON.md#include-semantics-file-formats-and-extensions), extension should not be mandatory when processing includes. Following should work:

* base.conf
* cloud.conf: include "base"

Currently only this works:

* base.conf
* cloud.conf: include "base.conf"

spericas:
PR #4152

PR #4162, opened 2022-04-28
Support for Hocon inclusion of files without an extension

spericas:
Support for Hocon inclusion of files without an extensions. Uses default of '.conf' as stated in spec. Issue #4152.


barchetta (approved):
LGTM

PR #4167, opened 2022-04-29
Turn off reference substitution in Hocon parser level for lazy resolution of references

spericas:
Turn off reference resolution/substitution at parser level and leave reference in tree for later resolution in config module. Unfortunately, to avoid depending on parser impl packages, we need to catch an exception for each parsed reference at this time.

### DIFF
--- a/config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
+++ b/config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
@@ -112,18 +112,6 @@ public class HoconConfigParser implements ConfigParser {
     }
 
     /**
-     * Create a new instance of HOCON config parser using default configuration.
-     *
-     * @param resolvingEnabled whether to resolve references at the parser level
-     * @return a new instance of parser
-     * @see #builder()
-     */
-    public static HoconConfigParser create(boolean resolvingEnabled) {
-        return resolvingEnabled ? builder().enableResolving().build()
-                : builder().disableResolving().build();
-    }
-
-    /**
      * Create a new fluent API builder for a HOCON config parser.
      *
      * @return a new builder instance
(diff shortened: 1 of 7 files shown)


config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
```diff
@@ -111,6 +111,18 @@ public static HoconConfigParser create() {
...
     }
 
+    /**
+     * Create a new instance of HOCON config parser using default configuration.
+     *
+     * @param resolvingEnabled whether to resolve references at the parser level
+     * @return a new instance of parser
+     * @see #builder()
+     */
+    public static HoconConfigParser create(boolean resolvingEnabled) {
```
tomas-langer, line 121: Not sure we should create a shortcut method like this. Is the `resolvingEnabled` really the most commonly used configuration option?

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParserBuilder.java
```diff
@@ -56,6 +56,16 @@ public HoconConfigParserBuilder disableResolving() {
         return this;
     }
 
+    /**
+     * Enables HOCON resolving substitutions support. Default is {@code false}.
+     *
+     * @return modified builder instance
+     */
+    public HoconConfigParserBuilder enableResolving() {
```
tomas-langer, line 64: Please use method `resolvingEnabled(boolean enabled)` or similar. Having two methods to toggle one option seem excessive.
Maybe deprecate the existing one (if disableResolving exists and is in a public release)

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
spericas: Moved to test

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParserBuilder.java
spericas: Fixed

tomas-langer (approved):
LGTM, thanks!

PR #4172, opened 2022-05-04
RoleContainer support added

### LINKED ISSUE #4158: Problems with RoleValidator.Roles annotation at type level
Added role validation to my REST endpoint using the [Helidon Security Provider Role Validator] (https://helidon.io/docs/v2/#/mp/security/02_providers#_role_validator) example. It only works for me if I add the annotation to the endpoint method, if I put the annotation on the class, as per the documentation example, it does not work. 

This is the annotation I added to my endpoint method.
```
@RoleValidator.Roles(value = MY_ROLE, subjectType = SubjectType.USER)
@Authenticated
```

Using IDCS (abac,oidc,idcs-role-mapper providers). Reported by Glen Shearlaw.

Verdent:
Fixes #4158

### DIFF
--- a/security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
+++ b/security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
+++ b/security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
@@ -116,8 +116,7 @@ public final class RoleValidator implements AbacValidator<RoleValidator.RoleConf
                         }
                         builder.permitAll(false);
                         builder.denyAll(false);
-                    } else if (annotation instanceof RolesContainer) {
-                        RolesContainer container = (RolesContainer) annotation;
+                    } else if (annotation instanceof RolesContainer container) {
                         for (Roles role : container.value()) {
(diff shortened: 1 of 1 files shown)


security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
```diff
@@ -116,6 +116,17 @@ public RoleConfig fromAnnotations(EndpointConfig endpointConfig) {
...
+                    } else if (annotation instanceof RolesContainer) {
+                        RolesContainer container = (RolesContainer) annotation;
+                        for (Roles role : container.value()) {
+                            if (role.subjectType() == SubjectType.USER) {
+                                roles.addAll(Arrays.asList(role.value()));
+                            } else {
+                                serviceRoles.addAll(Arrays.asList(role.value()));
+                            }
+                        }
+                        builder.permitAll(false);
```
tomas-langer, line 128: we should do this only if the roles container has at least one value (I think it can be declared as empty).

tomas-langer (changes requested):

security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
```diff
@@ -116,6 +116,19 @@ public RoleConfig fromAnnotations(EndpointConfig endpointConfig) {
                         }
                         builder.permitAll(false);
                         builder.denyAll(false);
+                    } else if (annotation instanceof RolesContainer) {
+                        RolesContainer container = (RolesContainer) annotation;
```
danielkec, line 120:
```suggestion
                    } else if (annotation instanceof RolesContainer container) {
```

danielkec (changes requested):

danielkec (approved):

security/abac/role/src/main/java/io/helidon/security/abac/role/RoleValidator.java
Verdent: done

Verdent:
Cannot merge since Tomas requested changes, but is currently out

PR #4185, opened 2022-05-09
Added X509 certificate context key when client certificate is present and pem trust store configuration

### LINKED ISSUE #2752: MTLS client certificate propagation
When MTLS is performed it would be useful for users to have access to the used client certificate over the context.

### LINKED ISSUE #3279: Use PEM file as certificate trust store
## Environment Details
* Helidon Version: 2.3.2
* Helidon SE
* JDK version: 11

----------

## Problem Description
It would be useful for the Web Client to be able to use PEM format certificates as key stores AND trust stores.

## Steps to reproduce
The below code compiles, but the connection fails.
```java
// ca.crt is just a single certificate "-----BEGIN CERTIFICATE-----"

KeyConfig keyConfig = KeyConfig.pemBuilder()
		.certChain(Resource.create(Path.of("/tmp/ca.crt")))
		.build();

WebClient webClient = WebClient.builder()
		.baseUri("https://myurl")
		.tls(WebClientTls.builder()
				.certificateTrustStore(keyConfig)
				.build())
		.build();

webClient.get().submit(); // Fails with "unable to find valid certification path to requested target"
```

The equivalent code using a JKS works fine.
```java
// Create new JKS
// keytool -noprompt -importcert -file /tmp/ca.crt -keystore /tmp/ca.jks -storepass password

KeyConfig keyConfig2 = KeyConfig.keystoreBuilder()
		.trustStore()
		.keystore(Resource.create(Path.of("/tmp/ca.jks")))
		.keystorePassphrase("password")
		.build();

WebClient webClient2 = WebClient.builder()
		.baseUri("https://myurl")
		.tls(WebClientTls.builder()
				.certificateTrustStore(keyConfig2)
				.build())
		.build();

webClient2.get().submit(); // Success
```

astromechza:
This PR adds some additional support for mTLS based workloads:

1. Adds PEM configuration of trust store certificates.
2. Exposes a context key for the X509Certificate of the client.

A new unittest class was added to add rapid testing of MTLS and (1) was done to support the testing of (2) using the self-signed certificate already present in the test resources.

Resolves #2752 and resolves #3279

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
@@ -63,6 +63,7 @@ import io.netty.handler.codec.http.websocketx.WebSocketServerProtocolHandler;
 import io.netty.handler.codec.http2.Http2Error;
 import io.netty.handler.codec.http2.Http2Exception;
 
+import static io.helidon.webserver.HttpInitializer.CLIENT_CERTIFICATE;
 import static io.helidon.webserver.HttpInitializer.CLIENT_CERTIFICATE_NAME;
 import static io.netty.handler.codec.http.HttpVersion.HTTP_1_1;
 
@@ -310,6 +311,10 @@ public class ForwardingHandler extends SimpleChannelInboundHandler<Object> {
         request.headers().remove(Http.Header.X_HELIDON_CN);
         Optional.ofNullable(ctx.channel().attr(CLIENT_CERTIFICATE_NAME).get())
                 .ifPresent(name -> request.headers().set(Http.Header.X_HELIDON_CN, name));
+        // If the client x509 certificate is present on the channel, add it to the context scope of the ongoing
+        // request so that helidon handlers can inspect and react to this.
+        Optional.ofNullable(ctx.channel().attr(CLIENT_CERTIFICATE).get())
+                .ifPresent(cert -> requestScope.register(WebServerTls.CLIENT_X509_CERTIFICATE, cert));
 
         // Context, publisher and DataChunk queue for this request/response
         DataChunkHoldingQueue queue = new DataChunkHoldingQueue();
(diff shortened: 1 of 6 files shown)


astromechza:
Fixing checkstyle.. pr ammended

webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
```diff
@@ -310,6 +311,10 @@ private boolean channelReadHttpRequest(ChannelHandlerContext ctx, Context reques
         request.headers().remove(Http.Header.X_HELIDON_CN);
         Optional.ofNullable(ctx.channel().attr(CLIENT_CERTIFICATE_NAME).get())
                 .ifPresent(name -> request.headers().set(Http.Header.X_HELIDON_CN, name));
+        // If the client x509 certificate is present on the channel, add it to the context scope of the ongoing
+        // request so that helidon handlers can inspect and react to this.
+        Optional.ofNullable(ctx.channel().attr(CLIENT_CERTIFICATE).get())
```
tomas-langer, line 316: We have chosen an approach with header (as can be seen right above with `X_HELIDON_CN`). I think that to be consistent, the certificate should be added as a header as well. @Verdent please comment as well

astromechza: currently it's a java.security X509Certificate

Verdent: The problem here is that the cert is currently java.security.X509Certificate. If we would use headers to store it, that would mean, we need to create String out of it first and when if anyone wants to use that, one would need to create X509Cert out of the String again.

There is header method which supports Object value, but when I went through the code, I saw it is actually converted to the String later on in Netty. (It is done by simply calling toString method). Because of that, I might say it is more convenient to do it as @astromechza did that.

astromechza: Yeah serializing back and forth to string is going to be fairly expensive and painful compared to having the object there in memory.

astromechza: one note on the existing X_HELIDON_CN header is that it is not documented anywhere and only used in some test cases.

tomas-langer: OK. in that case we should add the CN in a similar way, we may create a follow up issue.

tomas-langer (approved):
LGTM

Verdent (approved):

tomas-langer:
@astromechza Thank you for your contribution!

PR #4189, opened 2022-05-09
Fixing repackaged io.grpc module-info and MP gRPC client and server module-info files

thegridman:
Helidon gRPC is currently broken if running using Java modules.

- The module-info file in the repackaged `io.grpc` module has missing `requires` and missing `uses` statements.
- The module-info file in `microprofile/grpc/client` and  `microprofile/grpc/server` have need to be opened to Weld and CDI server.
- The module-info file in  `microprofile/grpc/server` has a missing `uses` statement.


tomas-langer (commented):
A few files have wrong copyright, but considering these are dummy files, I think it is OK (have 2022, should have 2019, 2022).
Need confirmation from @barchetta that all the dependencies check out

barchetta (approved):

PR #4216, opened 2022-05-13
New AnnotationFinder to handle transitive annotations in FT 

spericas:
New AnnotationFinder to handle transitive annotations in FT used by interceptor bindings. This allows users to define new annotations that are themselves annotated with FT annotations to be used. See new tests and Issue 4171.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AnnotationFinder.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AnnotationFinder.java
@@ -17,13 +17,13 @@
 package io.helidon.microprofile.faulttolerance;
 
 import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import javax.enterprise.inject.spi.BeanManager;
 import javax.interceptor.InterceptorBinding;
 
 /**
@@ -65,28 +65,8 @@ public class AnnotationFinder {
         return new AnnotationFinder(pkg);
     }
 
-    /**
-     * Searches for transitive annotations starting with a class.
-     *
-     * @param clazz the class
-     * @return set of transitive annotations found
-     */
-    Set<Annotation> findAnnotations(Class<?> clazz) {
-        return findAnnotations(Set.of(clazz.getAnnotations()));
-    }
-
-    /**
-     * Searches for transitive annotations starting with a method.
-     *
-     * @param method the method
-     * @return set of transitive annotations found
-     */
-    Set<Annotation> findAnnotations(Method method) {
-        return findAnnotations(Set.of(method.getAnnotations()));
-    }
-
-    Set<Annotation> findAnnotations(Set<Annotation> set) {
-        return findAnnotations(set, new HashSet<>(), new HashSet<>(), pkg);
+    Set<Annotation> findAnnotations(Set<Annotation> set, BeanManager bm) {
+        return findAnnotations(set, new HashSet<>(), new HashSet<>(), pkg, bm);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


ljnelson (approved):
Rubber stamp approval for now due to urgency; this will not work in many edge scenarios involving other extensions redefining interceptor bindings, designating other annotations as interceptor bindings, etc.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
```diff
@@ -79,24 +79,24 @@ public class FaultToleranceExtension implements Extension {
...
-            this.beanClass = beanClass;
-            this.method = method;
+        BeanMethod(AnnotatedType<?> annotatedType, AnnotatedMethod<?> annotatedMethod) {
+            this.annotatedType = annotatedType;
+            this.annotatedMethod = annotatedMethod;
         }
 
-        Class<?> beanClass() {
-            return beanClass;
+        AnnotatedType<?> annotatedType() {
```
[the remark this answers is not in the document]
spericas: Hmm, good point. I don't recall the reason for the pairing (annotation inheritance?). I removed it but will verify any odd cases in the TCKs.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AnnotationFinder.java
```diff
@@ -0,0 +1,127 @@
...
+                seen.add(a1);
+                Set<Annotation> a1Set = Set.of(a1Type.getAnnotations());
+                findAnnotations(a1Set, result, seen, pkg);
+            }
+        }
+        return result;
+    }
+
+    private static boolean isOfInterest(Annotation a) {
+        if (a.annotationType().isAnnotationPresent(InterceptorBinding.class)) {
```
ljnelson, line 119: Since you've now brought in `AnnotatedType` and `AnnotatedMethod` and so on, and are hence playing mostly nicely by the CDI rules, I suggest somehow getting this method to be more like:
```java
private static boolean isOfInterest(Annotation a, Predicate<? super Class<? extends Annotation>> p) {
  if (p.test(a.annotationType())) {
    // ...  
  }
  return false;
}
```
Then you can supply, for example, [`BeanManager#isInterceptorBinding(Class)`](https://jakarta.ee/specifications/cdi/2.0/apidocs/javax/enterprise/inject/spi/beanmanager#isInterceptorBinding-java.lang.Class-) for the `Predicate`, which is the correct test to use in CDI.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AsynchronousAntn.java
```diff
@@ -16,26 +16,25 @@
...
      *
-     * @param beanClass Bean class.
-     * @param method The method.
+     * @param annotatedType The annotated type.
+     * @param annotatedMethod The annotated method.
      */
-    AsynchronousAntn(Class<?> beanClass, Method method) {
-        super(beanClass, method);
+    AsynchronousAntn(AnnotatedType<?> annotatedType, AnnotatedMethod<?> annotatedMethod) {
+        super(annotatedType, annotatedMethod);
```
ljnelson, line 37: As with an earlier comment of mine, I don't think you need the explicit pairing since `annotatedMethod` can get you its declaring type.  There are other classes in this PR where you can also make this simplification; for brevity, I won't comment on all of them.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AnnotationFinder.java
spericas: Refactored

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AsynchronousAntn.java
spericas: I would prefer to come back to this after reviewing some TCK cases.

ljnelson (approved):

Issue #4183, opened 2022-05-06
DelayingRetryPolicy does not appear to work in Helidon SE


asantaga:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 2.5.0
* Helidon SE 
* JDK version: 11
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description

DelayingRetryPolicy does not appear to work, regardless of the settings it simply repeats the call immediately . Same code works perfectly if we replace DelayingRetryPolicy with jitterRetryPolicy

## Steps to reproduce
``` Sample snippet
  private void getDefaultMessageHandler(ServerRequest request, ServerResponse response) {

        Retry delayingRetry =  Retry.builder().retryPolicy(Retry.DelayingRetryPolicy.builder()
                        .calls(10)
                        .delay(Duration.ofMillis(5000L))
                        .delayFactor(2.5)
                        .build())
                .overallTimeout(Duration.ofSeconds(100000L))
                .build();
        delayingRetry.invoke(() -> {
                    //Pretend to Call my Service
                    System.out.println("Pretending to call  dummy service "+LocalDateTime.now());
                    // Oh it pretend failed, return failure
                    return Single.error(new RuntimeException("Failed to call service"));
                }).thenAccept(response::send)
                .exceptionally(response::send);


    }

```

Sample code uploaded, simply run and execute ```curl http://127.0.0.1:8080/greet```

[angelo-helidon-faulttolerence.zip](https://github.com/oracle/helidon/files/8641443/angelo-helidon-faulttolerence.zip)

spericas:
PR #4217 
PR #4221

PR #4217, opened 2022-05-13
Fixed problem in DelayRetryPolicyTest that would cause all delays to be zero

spericas:
Fixed problem in DelayRetryPolicyTest that would cause all delays to be zero when using the last delay returned. Fixed tests as well to reflect proper semantics. See issue #4183.


tomas-langer (approved):
LGTM

Issue #4195, opened 2022-05-11
Helidon Config should not depend on Reactive


tomas-langer:
Currently we never use reactive in production code of Config.
We should remove this dependency.
Unfortunatelly it does a `requires transitive` on `io.helidon.common.reactive`, so removing this will most likely cause failures in other modules.

PR #4225, opened 2022-05-16
2.x: Remove dependency on helidon-common-reactive from config

barchetta:
Removes the dependency on `helidon-common-reactive` from `confg` and adjust other modules as needed.

See #4195

### DIFF
--- a/integrations/vault/secrets/database/src/main/java/module-info.java
+++ b/integrations/vault/secrets/database/src/main/java/module-info.java
@@ -20,7 +20,7 @@
 module io.helidon.integrations.vault.secrets.database {
     requires java.logging;
 
-    requires io.helidon.common.reactive;
+    requires transitive io.helidon.common.reactive;
     requires io.helidon.integrations.vault;
     requires io.helidon.integrations.common.rest;
 
--- a/integrations/vault/secrets/database/src/main/java/module-info.java
+++ b/integrations/vault/secrets/database/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2021, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


integrations/vault/secrets/database/src/main/java/module-info.java
```diff
@@ -20,6 +20,7 @@
 module io.helidon.integrations.vault.secrets.database {
     requires java.logging;
 
+    requires io.helidon.common.reactive;
```
romain-grecourt, line 23: This module exposes an API `DbSecretsRx` that uses `Single`, we should use `requires transitive`. CC @tomas-langer

integrations/vault/sys/sys/src/main/java/module-info.java
```diff
@@ -23,6 +23,7 @@
 module io.helidon.integrations.vault.sys {
     requires java.logging;
 
+    requires io.helidon.common.reactive;
```
romain-grecourt, line 26: This module exposes a public API `SysRx` that uses `Single`. We should use requires transitive. CC @tomas-langer

messaging/messaging/src/main/java/module-info.java
```diff
@@ -22,6 +22,7 @@
 
     requires io.helidon.common.context;
     requires io.helidon.common.configurable;
+    requires io.helidon.common.reactive;
```
romain-grecourt, line 25: This module exposes `io.helidon.common.reactive` as part of its API, we should use `requires transitive`. CC @danielkec

integrations/vault/secrets/database/src/main/java/module-info.java
barchetta: Fixed

messaging/messaging/src/main/java/module-info.java
barchetta: Fixed

integrations/vault/sys/sys/src/main/java/module-info.java
barchetta: Fixed

PR #4226, opened 2022-05-17
[backport] Added X509 certificate context key when client certificate is present and pem trust store configuration

astromechza:
Backporting https://github.com/oracle/helidon/pull/4185#event-6613756686 to helidon 2.x branch with the intention to release it in 2.5.1.

### DIFF
--- a/webserver/webserver/src/test/java/io/helidon/webserver/MtlsTest.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/MtlsTest.java
@@ -1,0 +1,151 @@
+/*
+ * Copyright (c) 2017, 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.webserver;
+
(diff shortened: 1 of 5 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/MtlsTest.java
```diff
@@ -0,0 +1,151 @@
...
+        rawConfig.put("private-key.pem.cert-chain.resource.resource-path", "ssl/certificate.pem");
+
+        webServer = WebServer.builder(
+                Routing.builder()
+                        .any((req, res) -> {
+                            // This is annoyingly complex to pull just the CN out of an x509 cert, but it's generally easier if the caller
+                            // has access to other libraries like bouncy castle.
+                            Optional<X509Certificate> cert = req.context().get(WebServerTls.CLIENT_X509_CERTIFICATE, X509Certificate.class);
+                            res.send(cert.map(X509Certificate::getSubjectX500Principal).map(X500Principal::getName)
+                                .map(name -> Pattern.compile("(?:^|,\s?)(?:CN=(?<val>\"(?:[^\"]|\"\")+\"|[^,]+))").matcher(name))
```
tomas-langer, line 77: The pattern seems to be wrong - if you want to have \s as whitespace, it must be double escaped in Java (e.g. `\\s`). Probably true for other escapes in this regexp (and same in `master` in the original PR)

tomas-langer (changes requested):
See comment in `MtlsTest.java` - this fails the build

webserver/webserver/src/test/java/io/helidon/webserver/MtlsTest.java
astromechza: Looking into this now. Have been a bit busy today. It's interesting that this causes a compilation failure on the 2.x branch (java 11?) but not on the master branch (java 17?). That's why I didn't catch it there 🤔

astromechza: This is fixed now.

Verdent (approved):

PR #4229, opened 2022-05-17
3.x: Remove dependency on helidon-common-reactive from config (#4225)

### LINKED ISSUE #4195: Helidon Config should not depend on Reactive

barchetta:
* Remove dependency on helidon-common-reactive from config
Fixes #4195

### DIFF
--- a/fault-tolerance/src/main/java/module-info.java
+++ b/fault-tolerance/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


fault-tolerance/src/main/java/module-info.java
```diff
@@ -20,7 +20,8 @@
 module io.helidon.faulttolerance {
     requires io.helidon.config;
     requires io.helidon.common.configurable;
+    requires io.helidon.common.reactive;
```
tomas-langer, line 23: This should be a transitive dependency, as Single and Multi are part of public API of this module.

tomas-langer (changes requested):

fault-tolerance/src/main/java/module-info.java
barchetta: Fixed

tomas-langer (approved):

PR #4255, opened 2022-05-23
Port of a few Fault Tolerance related PRs to master

spericas:
This PR includes forward ports of 2.x PR's as well as TCK verifications:

1. PR #4216 
2. PR #4239
3. Verified all FT 4.0 TCKs are passing

Due to changes in master, these PR's had to be merged manually. Hence, there may be some differences when compared to the original ones in helidon-2.x.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AsynchronousAntn.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AsynchronousAntn.java
@@ -20,8 +20,6 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Future;
 
 import jakarta.enterprise.inject.spi.AnnotatedMethod;
-import jakarta.enterprise.inject.spi.AnnotatedType;
-
 import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException;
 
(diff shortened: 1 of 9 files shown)


barchetta (approved):

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AnnotationFinder.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param pkg the package
+     * @return the result set of annotations
+     */
+    private Set<Annotation> findAnnotations(Set<Annotation> set, Set<Annotation> result,
+                                            Set<Annotation> seen, Package pkg, BeanManager bm) {
+        for (Annotation a1 : set) {
+            Class<? extends Annotation> a1Type = a1.annotationType();
+            if (a1Type.getPackage().equals(pkg)) {
+                result.add(a1);
+            } else if (!seen.contains(a1) && isOfInterest(a1, bm)) {
```
tjquinno, line 89: I haven't thought about this enough, but for performance reasons is there any point in adding _every_ `a1` --even uninteresting ones--to `seen` to short-circuit the `else if` clause in later iterations? Or does it make sense to maintain a separate `uninterestingSeen` collection and check that also as part of the `else if` test?

spericas: Yes, perhaps, it is usual balance between using more memory or a few more cycles. The `seen` logic is to avoid entering infinite loops, even though I found that Weld will blow the stack if you define such a loop.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/AnnotationFinder.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Collects a set of transitive annotations in a package. Follows any
+     * annotation that has not been already seen (to avoid infinite loops)
+     * and is of interest.
+     *
+     * @param set set of annotations to start with
+     * @param result set of annotations returned
+     * @param seen set of annotations already processed
+     * @param pkg the package
+     * @return the result set of annotations
```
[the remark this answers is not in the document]
spericas: We can fix it if there's another commit to this I guess

PR #4334, opened 2022-06-03
Deserialization disabled by default.

tomas-langer:
Added serial config where required.
MP Config uses known class for converters (to avoid lambdas and complex deserialization)


barchetta (approved):
LGTM

PR #4337, opened 2022-06-03
Added config support for bulkheads, breakers, timeouts and retries

spericas:
Added config support for bulkheads, breakers, timeouts and retries. New tests that show how to configure handlers using YAML.

### DIFF
--- a/fault-tolerance/src/main/java/io/helidon/faulttolerance/Timeout.java
+++ b/fault-tolerance/src/main/java/io/helidon/faulttolerance/Timeout.java
@@ -20,6 +20,7 @@ import java.time.Duration;
 import java.util.concurrent.ScheduledExecutorService;
 
 import io.helidon.common.LazyValue;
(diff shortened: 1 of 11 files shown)


fault-tolerance/src/main/java/io/helidon/faulttolerance/Timeout.java
```diff
@@ -120,6 +121,50 @@ public Builder cancelSource(boolean cancelSource) {
...
+         *     <td>cancel-source</td>
+         *     <td>true</td>
+         *     <td>Cancel task source if task is cancelled</td>
+         * </tr>
+         * </table>
+         *
+         * @param config the config node to use
+         * @return updated builder instance
+         */
+        public Builder config(Config config) {
```
tomas-langer, line 160: Now that config is supported, we should also add the configuration metadata annotations + annotation processor

tomas-langer (approved):
LGTM

PR #4382, opened 2022-06-13
3.x - MP Config - CDI TCK Synthetic Observers Fix

MattGill98:
The CDI TCK will fail against any container with the ConfigCdiExtension registered as synthetic observers don't contain any annotated method, which causes a NPE during the 'ProcessObserverMethod'.

These synthetic events can be safely ignored, as it isn't a valid use of MP Config.

Note that this was discovered when using the MP Config implementation in GlassFish.


ljnelson (approved):
LGTM

arjav-desai:
@MattGill98 Can you please sign the OCA https://github.com/oracle/helidon/blob/master/CONTRIBUTING.md#code, as its required for "Contributing to Project Helidon"?

PR #4394, opened 2022-06-15
Add helidon-config-yaml-mp as a dependency in helidon-microprofile-config. (#4379)

klustria:
helidon-config-yaml-mp was recently decoupled from helidon-config-mp which is part of helidon-microprofile-config and as a consequence, is not a part of that config bundle anymore. This work was done to add it back and avoid any effects of incompatibility.

(cherry picked from commit aca9bf76f046fd94f292c8a6d61ae79c65f72031)

### DIFF
--- a/microprofile/config/pom.xml
+++ b/microprofile/config/pom.xml
@@ -41,6 +41,10 @@
         </dependency>
         <dependency>
             <groupId>io.helidon.config</groupId>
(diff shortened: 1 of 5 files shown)


ljnelson (approved):

klustria:
This is a fix for https://github.com/oracle/helidon/issues/4375 in v2

Issue #4365, opened 2022-06-09
Deprecations: Metrics


tomas-langer:
- MetricsSupport - 3 methods, replacing Conifg with metrics settings (no info)
- KeyPerformanceIndicatorMetricsSettings - New class in metrics API, this is for backward compatibility only (no info)
- RegistryFactory - New class in metrics API, this is for backward compatibility only (no info)

Tasks:
- [x] #4397 remove the deprecated code
- [x] #4396 make sure the deprecation is correctly documented in 2.x (since, forRemoval on @Deprecated annotation, use @deprecated javadoc tag to link to the correct new implementation, or describe what to do if a simple replacement does not exist)
- [x] #4396 validate documentation and examples are correct with regard to the deprecated element

PR #4396, opened 2022-06-15
Fix deprecation annotations in metrics; update doc; revise examples (2.x)

tjquinno:
See #4365 

This PR:
1. Adjusts `@Deprecated` annotations as needed.
2. Updates a few files in docs to refer to the non-deprecated alternatives.
3. Update many examples to use the non-deprecated alternatives.

Recall that metrics is now split into these components:
* `helidon-metrics-api` - metrics-related interfaces (`RegistryFactory`, `MetricsSettings`, etc.) and a no-op implementation
* `helidon-metrics-service-api` - `MetricsSupport` interface and no-op impl
* `helidon-metrics` - full-featured metrics impl

In the many places where examples used deprecated classes or methods this PR updates the code to use the non-deprecated alternatives. 

Note that this involves changes to poms. Our doc tells users how to build and package _metrics-capable_ apps which work regardless of whether the full metrics impl is present at runtime.  This involves declaring build-time dependencies on `helidon-metrics-api` and/or `helidon-metrics-service-api` (as needed) and then deciding if and how to make the full-featured impl available at runtime.

I have changed the examples accordingly (I should have done this when I created the new components), including adding a runtime-scope dependency on the full-featured metrics impl. This way the examples serve as metrics-capable apps, and  they also exhibit live metrics as they have before.

### DIFF
--- a/metrics/metrics/src/main/java/io/helidon/metrics/MetricsSupportProviderImpl.java
+++ b/metrics/metrics/src/main/java/io/helidon/metrics/MetricsSupportProviderImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2021, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,7 +26,7 @@ public class MetricsSupportProviderImpl implements MetricsSupportProvider<Metric
 
     @Override
     public MetricsSupport.Builder builder() {
-        return MetricsSupport.builder();
+        return new MetricsSupport.Builder();
     }
 
     @Override
--- a/docs/se/cors/04_support-in-builtin-services.adoc
+++ b/docs/se/cors/04_support-in-builtin-services.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2020 Oracle and/or its affiliates.
+    Copyright (c) 2020, 2022 Oracle and/or its affiliates.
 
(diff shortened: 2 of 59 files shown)


ljnelson (approved):

ljnelson (dismissed):

metrics/metrics/src/main/java/io/helidon/metrics/MetricsSupportProviderImpl.java
```diff
@@ -26,7 +26,7 @@ public class MetricsSupportProviderImpl implements MetricsSupportProvider<Metric
 
     @Override
     public MetricsSupport.Builder builder() {
-        return MetricsSupport.builder();
+        return new MetricsSupport.Builder();
```
tomas-langer, line 29: This should call `MetricsSupport.builder()` and not create a new instance explicitly. I am not sure why this change was made....

tomas-langer (changes requested):

ljamen (commented):
Doc looks good to me. I also double-checked the copyright years.

metrics/metrics/src/main/java/io/helidon/metrics/MetricsSupportProviderImpl.java
tjquinno: Changed back.

tomas-langer: Maybe the method should be changed to package local? Builder should not be instantiated in any other place except for the class that "owns" it.

tomas-langer (commented):
I have reopened one comment, as we need to be clear on the builder method, otherwise this seems OK

tomas-langer (approved):

PR #4397, opened 2022-06-15
Deal with deprecations in metrics (3.x) 

tjquinno:
See #4365 

This PR is largely a forward-port of the corresponding 2.x changes  in #4396 for 3.x (code, doc, and examples).

Note: There are number of methods marked as deprecated in 2.x that we should keep in 3.x (although still marked as deprecated). This way  we do not force developers with existing applications to do work to fix applications that would break. 

Although it's true that a major release such as 3.0 allows us to make breaking changes, the upcoming rev of the MicroProfile Metrics spec will introduce breaking changes on its own that our developers will have to deal with along with other potential metrics changes in SE. 

At a very low cost--the deprecated methods carried forward into 3.0 are typically one-line delegations to their non-deprecated alternatives--we can save our developers successive conversion efforts: now (to 3.0) and again to adopt our next major release.

### DIFF
--- a/docs/se/cors/04_support-in-builtin-services.adoc
+++ b/docs/se/cors/04_support-in-builtin-services.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2020 Oracle and/or its affiliates.
+    Copyright (c) 2020, 2022 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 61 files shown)


docs/se/cors/04_support-in-builtin-services.adoc
```diff
@@ -87,14 +88,16 @@ private static Routing createRouting(Config config) {
...
+                .register(metrics)                  // Metrics at "/metrics" // <4>
                 .register("/greet", greetService)
                 .build();
     }
 ----
-<1> Create the `CrossOriginConfig` for metrics, limiting sharing to `\http://foo.com`.
-<2> Use the `CrossOriginConfig` instance in constructing the `MetricsSupport` service.
-<3> Use the `MetricsSupport` object in creating the routing rules.
+<1> Create the `CrossOriginConfig.Builder` for metrics, limiting sharing to `\http://foo.com`.
+<2> Use the `CrossOriginConfig.Builder` instance in constructing
```
ljamen, line 97: in constructing [....] or is this the end of the statement in constructing[.]

ljamen (commented):
Looks good - just one comment (question).

tomas-langer (approved):
LGTM

PR #4408, opened 2022-06-17
Add default value constants and use in config doc

### LINKED ISSUE #4359: Config metadata error in health-check
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
@Verdent reported that adding the metadata-processor dependency to health-check causes errors due to some annotations being incorrect.

tjquinno:
Resolves #4359 

Added constants for default configured values and used them in assigning initial values and in config doc annotations.

### DIFF
--- a/docs/shared/config/io.helidon.health.HealthSupport.adoc
+++ b/docs/shared/config/io.helidon.health.HealthSupport.adoc
@@ -41,7 +41,7 @@ Optional configuration options:
 
 |`exclude-classes` |Class<?>[&#93; |{nbsp} |A class may be excluded from invoking health checks on it.
  This allows configurable approach to disabling broken health-checks.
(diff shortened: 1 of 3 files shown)


tjquinno:
@arjav-desai @spericas @ljnelson I have added you as reviewers to this for a style question Tomas raised.

I added a bunch of constants to the code to support default values. This creates some near-duplication (i.e., bloat) because annotations do not accept "virtually constant" final values that are computed so in some cases there's both a string and a non-string setting. 

So, the style question is whether the proliferation of constants helps (because related constants are right next to each other so it's more likely someone changing a default will be visually reminded to change both) or hurts (because there's so many pairs of constants which serve the same function).

tomas-langer (approved):
LGTM

PR #4410, opened 2022-06-20
Media support for Quickstart and Custom SE

tvallin:
jsonb - jsonp - jackson


romain-grecourt (approved):

romain-grecourt:
You can comment out the pipelines in Jenkinsfile for now, since we use a snapshot version of build-tools.

PR #4412, opened 2022-06-20
Re-enable Packaging Tests

### LINKED ISSUE #4341: re-enable native image tests in pipeline on master

tomas-langer:
Resolves #4341 for jar, jlink and native image for SE, MP will be fixed in a follow up PR

As part of this change, we had to switch to Parsson JSON-P implementation. This requires multiple changes in dependencies.


microprofile/cdi/src/main/java/module-info.java
```diff
@@ -23,6 +23,8 @@
     // needed for Unsafe used from Weld
     requires jdk.unsupported;
     requires java.logging;
+    // weld requires java.sql.Data and we fail if not on classpath
```
ljnelson, line 26: Hmm; no such class. Do you mean `java.sql.Date`?

barchetta (approved):
Approve dependency changes.

microprofile/cdi/src/main/java/module-info.java
tomas-langer: yes

tomas-langer:
Merging even though there is a failing test - the test failure is intermittent and known

Issue #4406, opened 2022-06-17
RequestAdapterMp.path() doesn't return full path in URI for the below application usage.


bvkrish83:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

With a simple application like below, RequestAdapterMp.path() method returns /segment2 instead of /segment1/segment2.
Due to this behaviour we are not able to properly add CORS path-params in our original application.

`
@ApplicationScoped
@ApplicationPath("/segment1")
public class SampleApplication extends Application {
    @Override
    public Set<Class<?>> getClasses() {
        return Set.of(SampleResource.class);
    }
}

@Path("/")
public class SampleResource {
    @GET
    @Path("/segment2")
    public Response method() {
        return Response.ok().build();
    }
}
`

Shouldn't RequestAdapterMp.path be like below to read the proper and complete path information???

`
        @Override
        public String path() {
            String path = requestContext.getRequestUri().getPath();
            return path.startsWith("/") ? path : '/' + path;
        }
`

## Environment Details
* Helidon Version: 2.5.1
* Helidon MP
* JDK version: JDK 17
* OS: Windows 11

PR #4416, opened 2022-06-21
Updating RequestAdapterMp.path to return correct path.

bvkrish83:
Updating RequestAdapterMp.path to return correct path.

To fix issue mentioned in #4406

### DIFF
--- a/microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsSupportMp.java
+++ b/microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsSupportMp.java
@@ -111,7 +111,7 @@ class CorsSupportMp extends CorsSupportBase<ContainerRequestContext, Response, C
 
         @Override
         public String path() {
-            String path = requestContext.getUriInfo().getPath();
+            String path = requestContext.getUriInfo().getRequestUri().getPath();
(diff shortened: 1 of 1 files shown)


tjquinno:
@bvkrish83

tjquinno:
@bvkrish83 

1. The system seems to think you have not completed the contributor agreement. Have you?
2. If you want to complete the agreement and then merge this PR, please add a test that triggers the error without the fix and passes with the fix.
3. We will also need a PR with the same fix and test based on the `helidon-2.x` branch.

Instead, if you prefer, I can create new PRs with the fix and test. Let me know how you want to proceed.

Thanks.

bvkrish83:
@tjquinno

I have started the process on Contributor Agreement, but it is taking more time than expected to get this completed and done.

I am fine with either of the approaches. With the fact that OCA is taking more time than expected, it might be better for you  (or someone else in Helidon Team) take the ownership of this change and get it done.

tjquinno:
@bvkrish83 It's unfortunate the OCA process is not faster, because naturally we welcome and encourage contributions.

That said, I will get to work on PRs myself to fix this in 2.x and 3.x as that might be faster.

tjquinno:
Closing this. See PRs #4434 (2.x) and #4435 (3.x) for the change and added tests.

PR #4418, opened 2022-06-21
Upgrade to Tyrus 2.0.2 

spericas:
Upgrade to Tyrus 2.0.2 that now uses the JPMS.
Note that Tyrus server module depends on Tyrus client, hence the change of scope in pom.


romain-grecourt (approved):

webserver/tyrus/pom.xml
```diff
@@ -58,7 +58,6 @@
         <dependency>
             <groupId>org.glassfish.tyrus</groupId>
             <artifactId>tyrus-client</artifactId>
-            <scope>test</scope>
```
danielkec, line 61: @spericas Why the client?

danielkec: I don't think this is going to work without `requires org.glassfish.tyrus.container.grizzly.client;`

tomas-langer: Please no. Let's talk to Tyrus to get a module that does not explicitly depend on a client

spericas:
https://github.com/oracle/helidon/pull/4429

Issue #4314, opened 2022-06-01
[Doc:] Helidon SE OpenAPI documentation


dalexandrov:
Helidon SE OpenAPI documentation (according to the new documentation structure).

Path: `/HelidonSE/Calling Other Services`

- [x] Overview (general description should be created)
- [x] Maven Dependency (may be reused from current doc)
- [x] Usage (may be reused from current doc)
- [x] API (should be created)
- [x] Configuration (may be extended from the current docs)
- [x] Reference (generated)
- [x] Examples (reference to `Greeting` application can be used)
- [x] Additional Information (`OpenAPI-Generator` reference and usage may be included here)

ljamen:
Comments on 4558

Issue #4209, opened 2022-05-13
[Doc:] Helidon MP OpenAPI 


dalexandrov:
Helidon MP OpenAPI documentation (according to the new documentation structure).

Path: `/HelidonMP/Calling Other Services`

Currently mostly reference MP OpenAPI spec.

- [x] Overview (general description should be created)
- [x] Maven Dependency (may be reused from current doc)
- [x] Usage (may be reused from current doc)
- [x] API (should be created)
- [x] Configuration (may be extended from the current docs)
- [x] Reference (generated)
- [x] Examples (reference to `Greeting` application can be used)
- [x] Additional Information (`OpenAPI-Generator` reference and usage may be included here)

No `Jandex` should be included.

PR #4419, opened 2022-06-21
Fix config doc annotations for OpenAPI builder

tjquinno:
Part of #4209 (and part of #4314)

Reviewers: Please pay particular attention to the default value for the `static-file` config setting as proposed. Our code looks for a static OpenAPI document file at `META-INF/openapi.` and any of `yml`, `yaml`, or `json` for the file type.

I've expressed the default value for config doc purposes as `META-INF/openapi.{json,yaml,yml}` to try to capture this succinctly, although obviously that's not a value that a user would be able to copy and paste into an actual config file and have it work. 

I think that's OK. In fact, expressing this in some other way would be more verbose and no more clear, I think.


tjquinno:
After I created this PR, I discovered that @dalexandrov has also been working on some of the same files but had not created a PR yet (although he has subsequently: #4421). He and I will be reconciling our changes. I have marked this as a draft until he and I figure that out.

tjquinno:
NOTE: I made a small local change to the config doc generator tool to get the proper output in the `docs/shared/config/io.helidon.openapi.SEOpenAPISupport.adoc` file in the latest push. I am in touch with David and Tomas to see if I was doing something wrong with the annotations or if there is, in fact, a change needed in the tool.

tjquinno:
In a semi-private chat, Tomas pointed out that in the merging config options I was incorrectly specifying the impl builder instead of the interface as `type`. The latest push fixes that, and when I ran the tool locally, the merging seems to be happening as expected.

PR #4423, opened 2022-06-22
Removed support for JavaMarshaller as default marshaller

spericas:
Changes in this PR:

- Removed support for `JavaMarshaller` as default marshaller
- Moved `JavaMarshaller` implementation to test sources for IT testing
- New default marshaller supplier will throw an exception if the code falls to where the `JavaMarshaller` was returned before to inform developer of the change
- Updated several tests to manually provide a marshaller supplier in order to run
 
This is a backward incompatible change.


grpc/core/src/main/java/io/helidon/grpc/core/MarshallerSupplier.java
```diff
@@ -71,7 +71,10 @@ public <T> MethodDescriptor.Marshaller<T> get(Class<T> clazz) {
             if (MessageLite.class.isAssignableFrom(clazz)) {
                 return proto.get(clazz);
             }
-            return JavaMarshaller.instance();
+            String msg = String.format(
+                    "Attempting to use class %s, a Java serialization marshaller is no longer supported by default",
+                    clazz.getName());
+            throw new IllegalArgumentException(msg);
```
aseovic, line 77: Throwing exception here makes sense, but I wouldn't even mention `JavaMarshaller` in the error message.

Assuming that `proto` is now the default marshaller (which it pretty much is), I would change the message to:

"Class %s must be a valid ProtoBuf message, or a custom marshaller for it must be specified explicitly."

I think that is much more appropriate, as it provides a hint to user on what to do.

microprofile/grpc/client/src/test/java/io/helidon/microprofile/grpc/client/GrpcClientBuilderTest.java
```diff
@@ -80,8 +78,6 @@ public void assertServiceOne(GrpcClientBuilder builder) {
         assertThat(grpcDescriptor, is(notNullValue()));
         assertThat(grpcDescriptor.getFullMethodName(), is("ServiceOne/foo/unary"));
         assertThat(grpcDescriptor.getType(), is(io.grpc.MethodDescriptor.MethodType.UNARY));
-        assertThat(grpcDescriptor.getRequestMarshaller(), is(instanceOf(JavaMarshaller.class)));
-        assertThat(grpcDescriptor.getResponseMarshaller(), is(instanceOf(JavaMarshaller.class)));
```
aseovic, line 84: Shouldn't we still check for "stub" marshaller, just to make sure the service has been configured correctly?

microprofile/grpc/server/src/test/java/io/helidon/microprofile/grpc/server/GrpcServiceBuilderTest.java
```diff
@@ -118,8 +116,6 @@ public void assertServiceOne(GrpcServiceBuilder builder) {
         assertThat(grpcDescriptor, is(notNullValue()));
         assertThat(grpcDescriptor.getFullMethodName(), is("ServiceOne/foo/unary"));
         assertThat(grpcDescriptor.getType(), is(io.grpc.MethodDescriptor.MethodType.UNARY));
-        assertThat(grpcDescriptor.getRequestMarshaller(), is(instanceOf(JavaMarshaller.class)));
-        assertThat(grpcDescriptor.getResponseMarshaller(), is(instanceOf(JavaMarshaller.class)));
```
aseovic, line 122: Ditto here. Why not test for "stub" marshaller?

aseovic (approved):
Couple minor comments, but overall it looks good.

Replicating `JavaMarshaller` multiple times is probably not ideal, but it makes more sense than introducing a new module that different test modules would depend on.

klustria (approved):
Looks good

klustria:
This resolves the removal of JavaMarshaller as requested in https://github.com/oracle/helidon/issues/4367

PR #4429, opened 2022-06-23
Restoring dependency with Tyrus 2.0.1

spericas:
Restoring dependency with Tyrus 2.0.1. The lastest version 2.0.2 introduces a strange dependency with the client module that is problematic for us.


tomas-langer (approved):
Thank you, let's revisit this later.

PR #4434, opened 2022-06-24
Correct return of path from MP CORS request adapter; add test

### LINKED ISSUE #4406: RequestAdapterMp.path() doesn't return full path in URI for the below application usage.

tjquinno:
Resolves #4406 for MP

See also PR #4416 from @bvkrish83 

The fix is straightforward.

Also added tests to exercise the fix. 

Doing that triggered a slight refactoring of tests. That's because we have to assign a `sun.xxx` property to allow setting restricted headers (which we must to to truly exercise the CORS processing) in outbound requests. 

Previously only one test class needed to do this, but the new test also must so I abstracted the common behavior into a common base class. The new test and the existing `CrossOriginTest` now extend that base class.


spericas (commented):
LGTM

tjquinno:
@spericas Thanks for the comment but GitHub didn't record an approval, if you intended to do so!

spericas (approved):

Issue #4276, opened 2022-05-26
Implement all templates for all archetype options


romain-grecourt:
Setup a branch with the updated 3.x archetypes that includes all the options (see https://dev.helidon.io/starter) and implement the templates for all options.

- [x] MP Media Support
- [x] MP Observability
- [x] MP Security
- [x] MP Extras
- [x] MP Packaging
- [x] SE Media Support
- [x] SE Observability
- [x] SE Security
- [x] SE Extras
- [x] SE Packaging
- [x] SE Database
- [x] MP Database

PR #4467, opened 2022-07-04
database MP templates

aserkes:
Database MP templates

see #4276 

Signed-off-by: aserkes <andrii.serkes@oracle.com>


archetypes/helidon/src/main/archetype/mp/database/files/src/test/resources/META-INF/microprofile-config.properties
```diff
@@ -0,0 +1,16 @@
+# Microprofile server properties
```
romain-grecourt, line 1: Shouldn't this file be removed since you have a template for it?

aserkes: This file contains a config for tests with values that will not change, so I think this file should not be deleted.

PR #4476, opened 2022-07-06
Revert back to using javax.validation as jakarta.validation is not supported by micronaut

klustria:
Note: Use jakarta.validation-api version 2.0.2 instead of version 3.0.0 to be able to use javax.validation instead of jakarta.validation.


barchetta (approved):
Approving dependency upgrades.

klustria:
Tomas has approved this based on this slack discussion:
```
[Tomas Langer][2:03 AM](https://proddev-paas-fmw.slack.com/archives/D02H83WT8LW/p1657184618305489)
That change is OK. As I have seen you are changing some versions of dependencies, so I have added Joe + added the label
[Tomas Langer][8:51 AM](https://proddev-paas-fmw.slack.com/archives/D02H83WT8LW/p1657209114331679)
go ahead and merge
```
... so I will merge once the pipeline completes successfully.

PR #4488, opened 2022-07-08
Introduces an idempotent shell function to download and install the O…

ljnelson:
…CI shaded full jar into the local Maven repository.

Signed-off-by: Laird Nelson <laird.nelson@oracle.com>

PR #4498, opened 2022-07-08
Makes OCI CDI extension work with OCI's shaded full jar

ljnelson:


ljnelson:
This is a by-hand port of `OCIExtension` from the `helidon-2.x` branch, with only changes that are necessary to depend on OCI's shaded full jar.

This PR depends on #4488 and must not be merged until that PR is merged.

spericas (approved):
LGTM

ljnelson:
JPMS strikes again. The `maven-javadoc-plugin` sets the `--module-path` which ends up using both "normal" OCI jar files and the shaded full jar file.  I will need to figure out how to work around this with profile adjustments in the relevant projects.  This will take some time.

PR #4500, opened 2022-07-08
Address deprecations in MP metrics for 3.x

### LINKED ISSUE #4371: Deprecations: MP Metrics
MetricUtil, MetricsCdiExtension methods
Marked as to be removed, for migration of grpc metrics, not done…

Tasks:
- [ ] remove the deprecated code
- [ ] make sure the deprecation is correctly documented in 2.x (since, forRemoval on @Deprecated annotation, use @deprecated javadoc tag to link to the correct new implementation, or describe what to do if a simple replacement does not exist)
- [ ] validate documentation and examples are correct with regard to the deprecated element

tjquinno:
Resolves #4371

# Background

The gRPC MP metrics component performs special handling for methods which are effectively annotated with `@GrpcMethod` and also bear a metrics annotation:
1. The normal metrics interception binding should not apply. gRPC methods are async and the interceptors are sync.
2. Under some gRPC-specific conditions, overridden methods _should_ and in other conditions they _should not_ inherit metrics annotation behavior from a superclass. When they should not, the system must should not register the metric which would correspond to that endpoint.

# In previous releases
During `ProcessAnnotatedType` processing, the gRPC MP metrics CDI extension _removed_ the metrics annotation from methods with both `@GrpcMethod` and a metrics annotation to prevent the normal interceptor binding. This not only prevented the normal metrics interceptor behavior (which is what gRPC really wanted), but it also burdened gRPC with registering the correct metrics with the correct metadata and name because the metrics CDI extension no longer saw the metrics annotations on those gRPC methods.

To do this, gRPC MP metrics invoked several methods in MP metrics that should not be exposed, methods that we subsequently marked as deprecated. It also depended too much on other metrics implementation internals.

# Redesign in 3.0.0

Note that these changes are intended for--and tuned to--usage from our own gRPC code. Although they are not intended for use from general user code we cannot prevent that. As a result there are no doc updates (other than JavaDoc) describing the new public interfaces.

## Remove the hard binding of our metrics interceptors to the MicroProfile Metrics annotations

Helidon MP metrics has always had an interceptor for each different MP metric annotation. Previously, these were bound directly, by declaration, to the corresponding MP metric annotation.

This PR introduces separate annotations for the interceptor bindings.

The metrics CDI extension code now _programmatically_ adds the correct interceptor binding annotation to each constructor and method if all observers agree that using the normal interceptor is OK. (See below for more about "all observers.")

## Clean up the contract between gRPC and metrics
* Remove (or reduce visibility of) much of the previously `public` and `@Deprecated` code.
* Expose new cleaner public interfaces:

1. New interface in `io.helidon.microprofile.metrics`: `MetricAnnotationDiscovery`
   
   Exposes information about the discovery of a metrics annotation as applied to an executable (constructor or method). Also allows an observer (see below) to:
      1. Suppress the normal metrics interceptor behavior for the specific combination of metric annotation/executable represented by the discovery object.
      2. Request that the metrics system completely disregard this discovery, preventing even the registration of a metric for it.
2. In the new package `io.helidon.microprofile.metrics.spi`:
   1. `MetricAnnotationDiscoveryObserver` - implementations are invoked with a `MetricAnnotationDiscovery` object for each discovery of a metrics annotation which applies to an executable.
   2. `MetricRegistrationObserver` - implementations are notified as each metric inspired by an annotation is registered. (There is no special "event" object  here; the SPI method's parameters  are the associated discovery and the `Metadata`, `MetricID`, and `Metric` objects for the just-registered metric.)
   3. Factory interfaces for the two observer interfaces. Metrics finds factory implementations using service loading and uses them to create the observers.

## Change the MP metrics CDI extension

The metrics CDI extension now:
1. Uses service loading to find and instantiate the two types of factories and the observers.
2. Invokes all discovery observers as the metrics annotations are found. If no observer deactivated the discovery or suppressed the default interceptor behavior, the extension programmatically adds the appropriate interceptor binding annotation to the annotation site.
3. During metrics registration (later in the CDI lifecycle), checks the corresponding discovery to make sure no discovery observer deactivated it. Then registers the metric and notifies the registration observers.

## Modify the gRPC code

1. Implement the discovery interface to record discoveries and suppress the normal interceptor binding if appropriate, or deactivate the discovery if needed. These decisions are governed by gRPC rules that I've carried over from the old code.
2. Implement the registration interface to record metrics information gRPC needs for its own interceptor.
3. Remove the gRPC metrics CDI extension. It is no longer needed.
4.  Update an important test to use `@HelidonTest` in place of the direct use of the Weld API.5. 
    
    Because the metrics CDI extension now creates all metrics, even gRPC ones, and notifies enrolled observers, the test needs to have the CDI container and the metrics CDI extension running. Just starting the server using `@HelidonTest` caused failures in the gRPC server extension because of the way some test classes are designed to exercise specific behavior. I got things to work using `@HelidonTest` and `@DisableDiscovery` and explicitly adding the CDI extensions required for the test--which does not include the gRPC server CDI extension.
    
    Getting to this point was a bit challenging and revealed either:   
    
    * a bug in the gRPC code (not in this PR), 
    * a test that exercises an unsupported case, or
    * a gap in my understanding. 
      
    The two gRPC experts are both out of the office for several days, so the updated `MetricsConfigurerTest` class in this PR is not as robust as it might be, but it works correctly. We might revisit this after the gRPC team weighs in. Any changes would not change the existing public interface or behavior of Helidon so we could improve this in a future dot release.

# Misc. Notes

I tried to purturb the general structure of the code as little as possible. There might have been ways to consolidate some of the data structures which `MetricsCdiExtension` manages internally, but that would have been considerably more work without much payoff. We are likely to undergo a significant rewrite of metrics for MP Metrics 5.0 and time was short enough as it was.


examples/quickstarts/helidon-quickstart-mp/pom.xml
```diff
@@ -42,6 +42,10 @@
             <scope>runtime</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
```
tomas-langer, line 45: grpc metrics should not be part of Quickstart MP

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricAnnotationDiscoveryObserverImpl.java
```diff
@@ -0,0 +1,107 @@
...
+import jakarta.enterprise.inject.spi.configurator.AnnotatedMethodConfigurator;
+
+/**
+ * The gRPC implementation of {@link io.helidon.microprofile.metrics.spi.MetricAnnotationDiscoveryObserver}.
+ *
+ * <p>
+ *     This implementation
+ * </p>
+ */
+public class GrpcMetricAnnotationDiscoveryObserverImpl implements MetricAnnotationDiscoveryObserver {
```
tomas-langer, line 37: Please do not use the `Impl` in this name - the class name is already specific enough.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricAnnotationDiscoveryObserverImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private static GrpcMetricAnnotationDiscoveryObserverImpl instance;
+
+    static GrpcMetricAnnotationDiscoveryObserverImpl instance() {
```
tomas-langer, line 41: Please fields first, methods later.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricAnnotationDiscoveryObserverImpl.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * The gRPC implementation of {@link io.helidon.microprofile.metrics.spi.MetricAnnotationDiscoveryObserver}.
+ *
+ * <p>
+ *     This implementation
+ * </p>
+ */
+public class GrpcMetricAnnotationDiscoveryObserverImpl implements MetricAnnotationDiscoveryObserver {
+
+    private static GrpcMetricAnnotationDiscoveryObserverImpl instance;
```
tomas-langer, line 39: This is a strange construct. This should at least be volatile.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricAnnotationDiscoveryObserverImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private final Map<Method,
+            Map<Class<? extends Annotation>, MetricAnnotationDiscovery.OfMethod>> discoveriesByMethod =
+            new HashMap<>();
+
+    /**
+     * Creates a new instance.
+     */
+    public GrpcMetricAnnotationDiscoveryObserverImpl() {
+        instance(this);
```
tomas-langer, line 57: This escape of instance to a static field is dangerous. Couldn't this be solved by using static fields to store the information, rather then storing the whole instance to a static field?

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricRegistrationObserverImpl.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.microprofile.metrics.MetricID;
+
+/**
+ * The gRPC implementation of {@link io.helidon.microprofile.metrics.spi.MetricRegistrationObserver} with a static factory method.
+ */
+public class GrpcMetricRegistrationObserverImpl implements MetricRegistrationObserver {
+
+    private static GrpcMetricRegistrationObserverImpl instance;
+
+    static GrpcMetricRegistrationObserverImpl instance() {
```
tomas-langer, line 35: Same as with previous type - static instance that is stored from constuctor is unusual and I am not sure what it is trying to achieve

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/MetricsConfigurer.java
```diff
@@ -258,4 +166,17 @@ public String createName(ServiceDescriptor service, String methodName, MetricTyp
             return name;
         }
     }
+
+    private boolean isDiscovered(Method method) {
+        return GrpcMetricAnnotationDiscoveryObserverImpl.instance().isDiscovered(method);
```
tomas-langer, line 171: this method should be defined on the GrpcMetricAnnotationDiscoveryObserverImpl, and use static fields.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricRegistrationObserverImpl.java
```diff
@@ -0,0 +1,63 @@
...
+import io.helidon.microprofile.metrics.spi.MetricRegistrationObserver;
+
+import org.eclipse.microprofile.metrics.Metadata;
+import org.eclipse.microprofile.metrics.Metric;
+import org.eclipse.microprofile.metrics.MetricID;
+
+/**
+ * The gRPC implementation of {@link io.helidon.microprofile.metrics.spi.MetricRegistrationObserver} with a static factory method.
+ */
+public class GrpcMetricRegistrationObserverImpl implements MetricRegistrationObserver {
```
tomas-langer, line 31: Also please remove Impl from name

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/MetricsConfigurer.java
```diff
@@ -258,4 +166,17 @@ public String createName(ServiceDescriptor service, String methodName, MetricTyp
...
+    }
+
+    private Metadata metadata(MetricAnnotationDiscovery discovery) {
+        return GrpcMetricRegistrationObserverImpl.instance().metadata(discovery);
```
tomas-langer, line 175: dtto

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/MetricsConfigurer.java
```diff
@@ -258,4 +166,17 @@ public String createName(ServiceDescriptor service, String methodName, MetricTyp
...
+    }
+
+    private Map<Class<? extends Annotation>,
+            MetricAnnotationDiscovery.OfMethod> discoveries(Method method) {
+        return GrpcMetricAnnotationDiscoveryObserverImpl.instance().discoveries(method);
```
tomas-langer, line 180: dtto

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/InterceptorSimplyTimed.java
```diff
@@ -29,11 +38,15 @@
  *     method, and the separate {@link InterceptorSyntheticRestRequest} interceptor deals with those.
  * </p>
  */
-@SimplyTimed
+@InterceptorSimplyTimed.Binding
 @Interceptor
 @Priority(Interceptor.Priority.PLATFORM_BEFORE + 10)
 final class InterceptorSimplyTimed extends InterceptorTimedBase<SimpleTimer> {
 
+    static Binding.Literal binding() {
```
tomas-langer, line 46: Static methods should be after constructors. This is valid for all interceptors.

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/InterceptorSimplyTimed.java
```diff
@@ -43,4 +56,24 @@ void postComplete(SimpleTimer metric) {
...
+    @Retention(RetentionPolicy.RUNTIME)
+    @Metered
+    @interface Binding {
+        class Literal extends AnnotationLiteral<Binding> implements Binding {
+
+            private static final long serialVersionUID = 1L;
+
+            private static final Literal INSTANCE = new Literal();
+
+            static Literal getInstance() {
```
tomas-langer, line 71: please rename to `instance` (in all interceptors)

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricAnnotationDiscoveryImpl.java
```diff
@@ -0,0 +1,160 @@
...
+ * <p>
+ *     The {@link jakarta.enterprise.inject.spi.configurator.AnnotatedConstructorConfigurator} and
+ *     {@link jakarta.enterprise.inject.spi.configurator.AnnotatedMethodConfigurator} interfaces share no common ancestor, so
+ *     we have two subtypes of discovery, one for each:
+ *     {@link io.helidon.microprofile.metrics.MetricAnnotationDiscoveryImpl.OfConstructor OfConstructor} and
+ *     {@link MetricAnnotationDiscovery.OfMethod ofMethod}.
+ * </p>
+ *
+ */
+abstract class MetricAnnotationDiscoveryImpl implements MetricAnnotationDiscovery {
```
tomas-langer, line 38: This is not an Impl but an abstract class - please either use `AbstractMetricAnnotationDiscovery` or `MetricAnnotationDiscoveryBase`. The current name is misleading

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/MetricAnnotationDiscoveryImpl.java
```diff
@@ -0,0 +1,160 @@
...
+                                (AnnotatedMethodConfigurator<?>) annotatedCallableConfigurator,
+                                annotation);
+        } else {
+            throw new IllegalArgumentException(String.format("annotatedCallableConfigurator must be of type %s or %s",
+                                                             AnnotatedConstructorConfigurator.class.getName(),
+                                                             AnnotatedMethodConfigurator.class.getName()));
+        }
+    }
+
+    private final AnnotatedTypeConfigurator<?> annotatedTypeConfigurator;
```
tomas-langer, line 59: Fields should be before methods.
static fields
fields
constructor(s)
static methods
public methods
...

tests/integration/mp-grpc/src/test/java/io/helidon/microprofile/grpc/server/InterceptorsTest.java
```diff
@@ -49,22 +48,16 @@
...
-                                                 ServerInterceptorTwo.class,
-                                                 ServerInterceptorThree.class,
-                                                 ServerInterceptorFour.class,
-                                                 InterceptedServiceOne.class,
-                                                 InterceptedServiceTwo.class,
-                                                 InterceptedServiceThree.class,
-                                                 InterceptedServiceFour.class,
-                                                 InterceptedServiceFive.class,
-                                                 InterceptedServiceSix.class);
+    private CDI weld;
```
tomas-langer, line 55: This deserves to be renamed to `cdi` now

tomas-langer (changes requested):

examples/quickstarts/helidon-quickstart-mp/pom.xml
tjquinno: Sigh. Leftover from an ad hoc test. Fixed.

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricRegistrationObserverImpl.java
tjquinno: Both of these types of observers need to be singletons. The metrics CDI extension uses service loading to obtain each one and store them for later use. Later, gRPC needs a way to find the already-created singletons. 

Storing the instance statically makes it easy for gRPC to find the singletons.

Another option I considered: The metrics CDI could expose methods which let gRPC explicitly enroll its observers during `BeforeBeanDiscovery`. The gRPC extension would instantiate its observers however it wanted, use CDI to look up the metrics extension, and pass the observers to the metrics extension `enroll` methods. I wanted to avoid exposing yet another public method on the metrics extension. 

Would that approach resolve your concerns?

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/InterceptorSimplyTimed.java
tjquinno: Changed

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/InterceptorSimplyTimed.java
tjquinno: Changed

tests/integration/mp-grpc/src/test/java/io/helidon/microprofile/grpc/server/InterceptorsTest.java
tjquinno: Done

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricAnnotationDiscoveryObserverImpl.java
tjquinno: Done

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/GrpcMetricAnnotationDiscoveryObserverImpl.java
tjquinno: Done

microprofile/grpc/metrics/src/main/java/io/helidon/microprofile/grpc/metrics/MetricsConfigurer.java
tjquinno: Done

tomas-langer (approved):
LGTM

Issue #4480, opened 2022-07-07
Deprecations: OCI Integration


arjav-desai:
With new OCI SDK integration approach, we have deprecated 

- our Gen 1/CDI extension OCI Objectstorage integration, in https://github.com/oracle/helidon/tree/master/integrations/cdi/oci-objectstorage-cdi
- our Gen 2/Native OCI integration, mainly in https://github.com/oracle/helidon/tree/master/integrations/oci

Tasks:

- [ ] remove the deprecated code
- [ ] make sure the deprecation is correctly documented in 2.x (since, forRemoval on @deprecated annotation, use @deprecated javadoc tag to link to the correct new implementation, or describe what to do if a simple replacement does not exist)
- [ ] validate documentation and examples are correct with regard to the deprecated element

PR #4502, opened 2022-07-10
OCI Integration Examples Update 2.x

arjav-desai:
Changes for #4480 in 2.x


examples/integrations/oci/atp-cdi/src/main/java/module-info.java
```diff
@@ -22,16 +22,19 @@
     requires java.naming;
     requires java.sql;
     requires java.ws.rs;
+
     requires jakarta.inject.api;
+    requires jakarta.enterprise.cdi.api;
```
ljnelson, line 27: This is the `helidon-2.x` branch, so I think all references to `jakarta.*` must be replaced with `javax.*` instead.

examples/integrations/oci/atp-reactive/pom.xml
```diff
@@ -60,8 +60,14 @@
             <artifactId>helidon-config-yaml</artifactId>
         </dependency>
         <dependency>
-            <groupId>io.helidon.integrations.oci</groupId>
-            <artifactId>helidon-integrations-oci-atp</artifactId>
+            <groupId>com.oracle.oci.sdk</groupId>
+            <artifactId>oci-java-sdk-database</artifactId>
+        </dependency>
+        <dependency>
+            <!-- required by SDK -->
```
ljnelson, line 67: I'm not sure this is right.  It should be pulled in transitively.  My guess is this is something added to make `jakarta.*` imports work; they shouldn't be here either (see above).

examples/integrations/oci/metrics-reactive/pom.xml
```diff
@@ -42,14 +42,20 @@
...
         <dependency>
             <groupId>io.helidon.config</groupId>
             <artifactId>helidon-config-yaml</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.oracle.oci.sdk</groupId>
+            <artifactId>oci-java-sdk-monitoring</artifactId>
+        </dependency>
+        <dependency>
+            <!-- required by SDK -->
```
ljnelson, line 54: Check this; see above

examples/integrations/oci/objectstorage-cdi/src/main/java/module-info.java
```diff
@@ -18,18 +18,21 @@
  * Example of integration with OCI object storage in a CDI application.
  */
 module io.helidon.examples.integrations.oci.objectstorage.cdi {
+    requires java.logging;
     requires java.ws.rs;
     requires java.json.bind;
+
     requires jakarta.inject.api;
+    requires jakarta.enterprise.cdi.api;
```
ljnelson, line 26: I admit that module names have changed during the torturous path toward Jakartafication but carefully check to see whether on the `helidon-2.x` line these shouldn't be `javax.xyz` rather than `jakarta.xyz`.  The important thing, I think, is that whichever modules they are, they have `javax.*` packages, not `jakarta.*` packages.

examples/integrations/oci/objectstorage-reactive/pom.xml
```diff
@@ -43,16 +43,18 @@
...
         <dependency>
-            <groupId>io.helidon.integrations.oci</groupId>
-            <artifactId>helidon-integrations-oci-objectstorage-health</artifactId>
+            <groupId>com.oracle.oci.sdk</groupId>
+            <artifactId>oci-java-sdk-objectstorage</artifactId>
         </dependency>
         <dependency>
-            <groupId>io.helidon.config</groupId>
-            <artifactId>helidon-config-yaml</artifactId>
+            <!-- required by SDK -->
```
ljnelson, line 54: Check this; see above

ljnelson (changes requested):

examples/integrations/oci/objectstorage-cdi/src/main/java/module-info.java
arjav-desai: Have removed it.

examples/integrations/oci/objectstorage-reactive/pom.xml
arjav-desai: Addressed comment on atp-reactive/pom.xml above

examples/integrations/oci/metrics-reactive/pom.xml
arjav-desai: Addressed comment on atp-reactive/pom.xml above

examples/integrations/oci/atp-cdi/src/main/java/module-info.java
arjav-desai: Have removed it.

examples/integrations/oci/atp-reactive/pom.xml
arjav-desai: As you said in another comment, jakarta.inject still have classes defined with javax.inject package name. OCI SDK required this to be available at runtime else it fails with `Caused by: java.lang.ClassNotFoundException: javax.inject.Singleton`

Since jakarta.inject is version managed in helidon's dependency pom, using it here. None of the test class is importing/using any jakarta.* classes.

ljnelson (approved):
LGTM

PR #4522, opened 2022-07-12
Adds slf4j-api as a transitive runtime dependency of the OCI shaded f…

ljnelson:
…ull jar in helidon-integrations-oci-sdk-cdi


ljnelson:
Because the OCI shaded full jar must be installed into the user's local Maven repository (cache) by hand, no transitive dependency information is available for it.  The OCI shaded full jar depends on `slf4j-api` so we need to make sure there's a `runtime`-scoped dependency for it.

arjav-desai (approved):
LGTM

PR #4530, opened 2022-07-13
Introduction and MP Introduction rework

m0mus:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
romain-grecourt, line 1: Perhaps this should be added to .gitignore

PR #4541, opened 2022-07-14
New config docs

tomas-langer:
Refactored config documentation:

    * move to docs/config
    * fixed wrong annotations
    * added section with all generated docs
    * use _ instead of . in file names (could not render menu)
    * sorted options

There is now a new section "Config reference" on same level as "Javadocs" - this section contains every single generated config documentation, so we can reference them from other configuration docs. This is needed due to a limitation of our current toolchain, where links can only lead to visible pages (pages in sitegen.yaml).


m0mus:
Why do we need another root menu "Config docs"? Configuration should be a part of the document it belongs to. Alos, we are trying to separate documentation of each flavor as much as possible.  Bringing another combined section works against it.

PR #4556, opened 2022-07-15
3.0 archetypes minor issues : Multiple fix to templates

tvallin:
See https://github.com/oracle/helidon/issues/4547

Fixing : 
  - Hanging Database SE tests
  - Media support when none is selected for bare SE/MP
  - Readme.md for bare SE/MP with wrong endpoints `/greet` instead of `/simple-greet`
  - Readme.md for database wrong endpoint `/greet` instead of `/simple-greet`


tvallin:
I pushed a commit with tests enabled to check if they are passing on pipeline, locally it works well.

romain-grecourt (approved):

Issue #4576, opened 2022-07-20
GRPC examples are failing


klustria:
## Environment Details
* Helidon Version: 3.0.0 and 2.5.x
* Helidon SE and Helidon MP

----------

## Problem Description
gRPC examples under `examples/grpc` are not running because:
1. JavaMarshaller was deprecated (affects version 3 only)
1. Microprofile examples has obsolete pom.xml (affects versions 2.5.x and 3)

## Steps to reproduce
Follow the steps in the readme of each of the gRPC feature examples.

klustria:
Should be resolved via PR: https://github.com/oracle/helidon/pull/4585 which is currently in progress.

klustria:
Still need to backport part of the fixes to v2.5

klustria:
Backported to 2.5.2 via https://github.com/oracle/helidon/pull/4635 hence closing.

PR #4585, opened 2022-07-21
Fix gRPC examples that are failing.

klustria:
1. Use built-in JsobBMarshaller for the gRPC request/response that needs it by setting it as the custom marshaller.
2. Update pom.xml for microservices example because the contents are already obsolete.
3. Enhance the READMEs across all the examples.
4. Update config for security-outbound example to add outbound entry so that the security context can properly be propagated from gRPC to Rest and vice-versa.
5. Remove grpc.marshaller.java.enabled from the config as it is not used anymore.
6. Replace policy with policy-validator as the former is an invalid ABAC attribute.


klustria:
This should fix [GRPC examples are failing #4576](https://github.com/oracle/helidon/issues/4576)

examples/grpc/common/src/main/java/io/helidon/grpc/examples/common/JavaMarshaller.java
```diff
@@ -0,0 +1,78 @@
...
+import io.grpc.MethodDescriptor;
+import jakarta.inject.Named;
+
+/**
+ * An implementation of a gRPC {@link MethodDescriptor.Marshaller} that
+ * uses Java serialization for testing.
+ *
+ * @param <T>  the type of value to be marshalled
+ */
+public class JavaMarshaller<T>
```
tomas-langer, line 42: Do we really need to use a Java marshaller? We actually wanted to remove any serialization from Helidon, and this is just re-introducing it into examples. 
We should use a different serialization approach in examples (proto, json) instead

tomas-langer (changes requested):

examples/grpc/common/src/main/java/io/helidon/grpc/examples/common/JavaMarshaller.java
klustria: Removed and replaced with built-in JsonBMarshaller

arjav-desai (approved):
LGTM

PR #1721, opened 2020-05-04
MP Config fixes

tomas-langer:
MP Configuration is now compliant with the specification (mutability of config).
Please see #1698 for details.

Resolves #1698

PR #4600, opened 2022-07-22
MP Config Map injection behavior restored to original

### LINKED ISSUE #4414: Bug: ConfigProperty annotation on Map<String, String> field doesn't filter the configuration values, but populates all configuration entries in Map
For a map configuration field like below, i expect the map to be populated with all configuration entries under configKey section, but the map is populated with all possible configuration entries - without any filtering.

```java
    @Inject
    @ConfigProperty(name = "configKey")
    Map<String, String> configKeyMap;
```

Is this behaviour expected??? Is there a provision to filter out the configuration entries that just match to the name mentioned in the annotation?

If i understand correctly, filtration happens properly for List and Set field types (at-least as per the code for those branches). Why the behaviour is different for Map alone???

## Environment Details
* Helidon Version: 2.5.1
* Helidon MP
* JDK version: JDK 17
* OS: Windows 11
* Docker version (if applicable): NA

tomas-langer:
Resolves #4414 

Shoudl behave similar to original before #1721 was applied

Injecting without explicit key will inject all properties, injecting with a key will inject properties where keys have the prefix removed (and only containing keys that are prefixed with the key).


tjquinno (approved):

tjquinno (approved):

tomas-langer:
Closing this PR, it is broken. Will reopen (cannot build using our pipelines)

PR #4612, opened 2022-07-25
Context propagation across HTTP.

### LINKED ISSUE #4256: support for transport level context propagation
Between client-side and server-side, we should be able to provide a means for developers to provide an extensible mechanism to have certain context attributes propagated (presumably over headers) to the remote server over a Helidon any Http Transport.  @tomas-langer provided an example of this in an example he provided to some internal teams.  The enhancement request here is to turn that example into a 1st class, supported framework mechanism in Helidon.

Relates tangentially to #4016

tomas-langer:
Resolves #4256 for 2.x

### DIFF
--- a/tests/integration/webserver/context-propagation/src/test/java/io/helidon/tests/integration/webserver/context/propagation/ContextPropagationTest.java
+++ b/tests/integration/webserver/context-propagation/src/test/java/io/helidon/tests/integration/webserver/context/propagation/ContextPropagationTest.java
@@ -70,6 +70,7 @@ class ContextPropagationTest {
         context.register(ContextPropagationMain.CLASSIFIER_VALUES, new String[] {"second", "third"});
         context.register(ContextPropagationMain.CLASSIFIER_DEFAULT, "fourth");
         context.register(ContextPropagationMain.CLASSIFIER_DEFAULTS, new String[] {"fifth", "sixth"});
+        context.register(ContextPropagationMain.NOT_PROPAGATED, "value");
 
         DataDto dto = client.get()
                 .context(context)
@@ -82,5 +83,6 @@ class ContextPropagationTest {
         assertThat(dto.getDefaultValue(), is("fourth"));
         assertThat(dto.getDefaultValues(), is(arrayContaining("fifth",
                                                               "sixth")));
+        assertThat(dto.getNotPropagated(), is(nullValue()));
     }
 }
--- a/tests/integration/webserver/context-propagation/src/test/java/io/helidon/tests/integration/webserver/context/propagation/ContextPropagationTest.java
+++ b/tests/integration/webserver/context-propagation/src/test/java/io/helidon/tests/integration/webserver/context/propagation/ContextPropagationTest.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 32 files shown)


tests/integration/webserver/context-propagation/src/test/java/io/helidon/tests/integration/webserver/context/propagation/ContextPropagationTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertThat(dto.getValue(), nullValue());
+        assertThat(dto.getValues(), nullValue());
+        assertThat(dto.getDefaultValue(), is("default-value"));
+        assertThat(dto.getDefaultValues(), is(arrayContaining("default-value1",
+                                                              "default-value2",
+                                                              "default-value3")));
+    }
+
+    @Test
+    void testAllCustom() {
```
cpotter1, line 67: would be nice to see a negative test case, where a particular value in the context is NOT propagated across.

webclient/context-propagation/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-context</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.webclient</groupId>
+            <artifactId>helidon-webclient</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
```
cpotter1, line 43: for my own understanding, why isn't this a normal compile time dep?

webserver/context-propagation/src/main/java/io/helidon/webserver/context/propagation/PropagationRecord.java
```diff
@@ -0,0 +1,149 @@
...
+                classifier = headerName;
+            } else if (headerName == null) {
+                headerName = classifier;
+            }
+            if (headerName == null) {
+                throw new IllegalArgumentException("Either header name or classifier must be configured");
+            }
+
+            if (isArray) {
+                return new ArrayRecord(classifier, headerName, defaultValue);
```
cpotter1, line 72: any considerations for record size limit / thresholds?

cpotter1 (approved):

webclient/context-propagation/pom.xml
tomas-langer: Because this is source only annotation that is processed by annotation processor.

trentjeff (approved):

webserver/context-propagation/src/main/java/io/helidon/webserver/context/propagation/PropagationRecord.java
tomas-langer: we already have a limit on maximal size of headers, so this should not be needed. We may want to add it on client side (so we do not have too big headers).

tomas-langer: I will leave this for our future selves - if there is a need to limit the size (or length of an array), let's get an issue and fix it

tests/integration/webserver/context-propagation/src/test/java/io/helidon/tests/integration/webserver/context/propagation/ContextPropagationTest.java
tomas-langer: Added a negative test (part of the custom test method)

trentjeff (approved):

Verdent (approved):

PR #4635, opened 2022-07-29
Fix gRPC examples that are failing. (#4585)

klustria:
1. Update pom.xml for microservices example because the contents are already obsolete.
2. Enhance the READMEs across all the examples.
3. Update config for security-outbound example to add outbound entry so that the security context can properly be propagated from gRPC to Rest and vice-versa.
4. Replace policy with policy-validator as the former is an invalid ABAC attribute.


barchetta (approved):

klustria:
This PR was created to fix https://github.com/oracle/helidon/issues/4576

Issue #4355, opened 2022-06-09
Retry Annotation fails with NullPointerException


Aditya-Jaitly:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 2.5.0
* Helidon MP
* JDK version: 11
* OS: Oracle Linux
* Docker version (if applicable): NA

----------

## Problem Description
Retry Annotation when applied on Rest client throws NullPointerException even when retry shouldnt be triggered. Expected Behaviour: Retry when Rest call fails.
Easily Reproducible
Stack Trace:
```
java.lang.NullPointerException
        at io.helidon.microprofile.faulttolerance.MethodInvoker.updateCounter(MethodInvoker.java:762)
        at io.helidon.microprofile.faulttolerance.MethodInvoker.updateMetricsAfter(MethodInvoker.java:655)
        at io.helidon.microprofile.faulttolerance.MethodInvoker.get(MethodInvoker.java:436)
        at io.helidon.microprofile.faulttolerance.CommandInterceptor.interceptCommand(CommandInterceptor.java:54)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.jboss.weld.interceptor.reader.SimpleInterceptorInvocation$SimpleMethodInvocation.invoke(SimpleInterceptorInvocation.java:73)
        at org.jboss.weld.interceptor.proxy.WeldInvocationContextImpl.invokeNext(WeldInvocationContextImpl.java:92)
        at org.jboss.weld.interceptor.proxy.WeldInvocationContextImpl.proceed(WeldInvocationContextImpl.java:124)
        at org.jboss.weld.bean.InterceptorImpl.intercept(InterceptorImpl.java:105)
        at org.glassfish.jersey.microprofile.restclient.InterceptorInvocationContext$InvocationInterceptor.intercept(InterceptorInvocationContext.java:136)
        at org.glassfish.jersey.microprofile.restclient.InterceptorInvocationContext.proceed(InterceptorInvocationContext.java:108)
[...]

PR #4641, opened 2022-07-29
Changes to FT implementation to support interception of proxy methods

spericas:
Changes to FT implementation to support interception of proxy methods created from RestClient interfaces. New test that shows integration between RestClient and FT both using injected proxies and proxies created from builders. Issue #4355.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodIntrospector.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodIntrospector.java
@@ -65,7 +65,7 @@ class MethodIntrospector {
                         .stream()
                         .filter(am -> am.getJavaMember().equals(method))
                         .findFirst();
-        this.annotatedMethod = annotatedMethodOptional.orElseThrow();
+        this.annotatedMethod = annotatedMethodOptional.orElse(new FtAnnotatedMethod(method));
 
         this.retry = isAnnotationEnabled(Retry.class) ? new RetryAntn(annotatedMethod) : null;
         this.circuitBreaker = isAnnotationEnabled(CircuitBreaker.class)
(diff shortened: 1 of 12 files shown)


microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodIntrospector.java
```diff
@@ -65,7 +65,7 @@ class MethodIntrospector {
                         .stream()
                         .filter(am -> am.getJavaMember().equals(method))
                         .findFirst();
-        this.annotatedMethod = annotatedMethodOptional.orElseThrow();
+        this.annotatedMethod = annotatedMethodOptional.orElse(new FtAnnotatedMethod(method));
```
tjquinno, line 68: Is the `AnnotatedType` derived from the `beanClass` valid for the `FtAnnotatedMethod` case as well? If so, would it make sense to pass that into the `FtAnnotatedMethod` constructor instead of having the constructor creating its own impl of `AnnotatedType` for the wrapper method with several "should not be called" methods?

tjquinno (dismissed):
Looks good to me. I posted one question that's probably not very important.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodIntrospector.java
spericas: Good question, no in the proxy case (something new to me) the type of the intercepted `beanClass` is `org.glassfish.jersey.microprofile.restclient.MethodModel` and that obviously has no annotations for us to directly inspect. Hence, the annotated type is derived from the Java method's declaring type. I believe digging into `MethodModel` we could find the real type, but that's deep into Jersey's implementation, which isn't ideal.

tjquinno (approved):

Issue #4512, opened 2022-07-12
cookie-encryption-password configuration parameter take only a single character


satya-24:
cookie-encryption-password configuration parameter take only a single character, it should take a string.

## Environment Details
* Helidon Version: 2.5.1 
* Helidon SE or Helidon MP: Helidon MP
* JDK version: 1.8
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description
cookie-encryption-password configuration parameter take only a single character, it should take a string.
Configuration - security.providers.oidc.cookie-encryption-password: “masterpassword”
Stack trace 
```
Exception in thread "main" java.lang.IllegalArgumentException: Failed to convert property "security.providers.1.oidc.cookie-encryption-password" from its value "masterpassword" to char
        at io.helidon.config.mp.MpConfigImpl.convert(MpConfigImpl.java:308)
        at io.helidon.config.mp.MpConfigImpl.toArray(MpConfigImpl.java:346)
        at io.helidon.config.mp.MpConfigImpl.optionalValue(MpConfigImpl.java:135)
        at io.helidon.config.mp.MpConfigImpl.getOptionalValue(MpConfigImpl.java:120)
        at io.helidon.config.mp.SeConfig.as(SeConfig.java:199)
        at io.helidon.security.providers.oidc.common.OidcConfig$Builder.config(OidcConfig.java:1248)
        at io.helidon.security.providers.oidc.common.OidcConfig.create(OidcConfig.java:463)
        at io.helidon.security.providers.oidc.OidcProvider$Builder.config(OidcProvider.java:699)
        at io.helidon.security.providers.oidc.OidcProvider.create(OidcProvider.java:207)
        at io.helidon.security.providers.oidc.OidcProviderService.providerInstance(OidcProviderService.java:45)
        at io.helidon.security.Security$Builder.providerFromConfig(Security.java:1336)
        at io.helidon.security.Security$Builder.lambda$fromConfig$1(Security.java:1191)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at io.helidon.security.Security$Builder.lambda$fromConfig$2(Security.java:1191)
        at java.base/java.util.Optional.ifPresent(Optional.java:183)
        at io.helidon.config.ConfigValue.ifPresent(ConfigValue.java:227)
        at io.helidon.security.Security$Builder.fromConfig(Security.java:1190)
        at io.helidon.security.Security$Builder.config(Security.java:1046)
[...]

PR #4657, opened 2022-08-02
Configuration parameter 'cookie-encryption-password' takes only a single character rather than a string (#4512)

klustria:
…

### DIFF
--- a/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
+++ b/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
@@ -1283,7 +1283,8 @@ public final class OidcConfig {
             config.get("header-token").as(TokenHandler.class).ifPresent(this::headerTokenHandler);
             // encryption of cookies
             config.get("cookie-encryption-enabled").asBoolean().ifPresent(this::cookieEncryptionEnabled);
-            config.get("cookie-encryption-password").as(Builder::toCookieEncryptionPassword)
+            config.get("cookie-encryption-password").as(String.class)
+                    .map(String::toCharArray)
                     .ifPresent(this::cookieEncryptionPassword);
             config.get("cookie-encryption-name").asString().ifPresent(this::cookieEncryptionName);
 
@@ -2054,14 +2055,6 @@ public final class OidcConfig {
 
         private void clientTimeoutMillis(long millis) {
             this.clientTimeout(Duration.ofMillis(millis));
-        }
-
-        private static char[] toCookieEncryptionPassword(Config value) {
-            String cookieEncryptionPasswordValue = value.asString().get();
-            if (cookieEncryptionPasswordValue == null || cookieEncryptionPasswordValue.trim().isEmpty()) {
-                throw new IllegalArgumentException("Required value for cookieEncryptionPassword is missing");
-            }
-            return cookieEncryptionPasswordValue.toCharArray();
         }
     }
 }
--- a/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
+++ b/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
@@ -1283,7 +1283,8 @@ public final class OidcConfig {
             config.get("header-token").as(TokenHandler.class).ifPresent(this::headerTokenHandler);
             // encryption of cookies
             config.get("cookie-encryption-enabled").asBoolean().ifPresent(this::cookieEncryptionEnabled);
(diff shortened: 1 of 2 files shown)


security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -1283,7 +1283,8 @@ public Builder config(Config config) {
             config.get("header-token").as(TokenHandler.class).ifPresent(this::headerTokenHandler);
             // encryption of cookies
             config.get("cookie-encryption-enabled").asBoolean().ifPresent(this::cookieEncryptionEnabled);
-            config.get("cookie-encryption-password").as(char[].class).ifPresent(this::cookieEncryptionPassword);
+            config.get("cookie-encryption-password").as(Builder::toCookieEncryptionPassword)
```
tomas-langer, line 1286: This would be easier:
```java
config.get("cookie-encryption-password").as(String.class)
                    .map(String::toCharArray)
                    .ifPresent(this::cookieEncryptionPassword);
```

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -2054,5 +2055,13 @@ public Builder clientTimeout(Duration duration) {
         private void clientTimeoutMillis(long millis) {
             this.clientTimeout(Duration.ofMillis(millis));
         }
+
+        private static char[] toCookieEncryptionPassword(Config value) {
+            String cookieEncryptionPasswordValue = value.asString().get();
```
tomas-langer, line 2060: the value can never be null - method `get()` would throw an exception if the value is missing

tomas-langer: empty password is a valid value

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcConfigFromBuilderTest.java
```diff
@@ -43,4 +50,27 @@ class OidcConfigFromBuilderTest extends OidcConfigAbstractTest {
     OidcConfig getConfig() {
         return oidcConfig;
     }
+
+    @Test
+    void testCookieEncryptionPasswordFromBuilderConfig() {
```
tomas-langer, line 55: I do not think this is correct. Empty password is OK if the user desires so.
Can you please add a test to verify that this issue is fixed (e.g. to make sure we get the password from config)?

tomas-langer (changes requested):

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
klustria: Ok, will not block empty password then. Or will just remove the condition for empty value.

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcConfigFromBuilderTest.java
klustria: I've been looking at validating the value but it is added to private vars. Will dig some more to see  if I can extract it from somewhere.

tomas-langer: no problem to add a package local method to retrieve the password from the builder

klustria:
goal of this PR is to fix https://github.com/oracle/helidon/issues/4512

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcConfigFromBuilderTest.java
klustria: OK, NW. I think I found a way to stub the config builder class so the cookie-encryption-password value can be retrieved and validated. Working on it.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
klustria: Yep, thought about this initially but also was thinking that empty string is not allowed hence why I created the custom method. But as you mentioned below, empty password is valid so now I have the opportunity to simplify it this way.

tomas-langer (approved):
LGTM

PR #4659, opened 2022-08-02
Injection of Map from configuration now honors prefix (backport)

### LINKED ISSUE #4414: Bug: ConfigProperty annotation on Map<String, String> field doesn't filter the configuration values, but populates all configuration entries in Map

tomas-langer:
Backport of #4653 for 2.x
Resolves #4414 

This is a refactoring (not a cherry-pick) due to quite big differences of sources between 3 and 2.

### DIFF
--- a/microprofile/config/src/main/java/io/helidon/microprofile/config/ConfigCdiExtension.java
+++ b/microprofile/config/src/main/java/io/helidon/microprofile/config/ConfigCdiExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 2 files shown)


tjquinno (approved):
LGTM

PR #4675, opened 2022-08-04
Configuration parameter 'cookie-encryption-password' takes only a single character rather than a string

klustria:


klustria:
goal of this PR is to fix https://github.com/oracle/helidon/issues/4512 in Master branch.

tomas-langer (approved):

PR #4678, opened 2022-08-05
Unescape the keys when config is returned as a map

klustria:


tomas-langer (approved):

klustria:
Fixes https://github.com/oracle/helidon/issues/4398

PR #4679, opened 2022-08-05
[Doc Update] Helidon Version update

mkhalid-s:
version update from 3.0.0-SNAPSHOT to 3.0.0, Since it started giving issues while pulling the artifacts.
Below issue
Screenshot 2022-08-05 at 18 55 42


barchetta:
@mkhalid-s Thanks for contributing this, but it turns out the root cause of the problem is in our release script that was not updated after we refactored our documentation. This is fixed by PR #4689. So I'm going to close this in deference to that PR.

PR #4693, opened 2022-08-10
Updates to Helidon Common, Part 1

tomas-langer:
- Use `System.Logger` in common (except for reactive)
- Move `HelidonServiceLoader` to `helidon-common`
- `GenericType.STRING` constant
- Add `Weight`, `Weighted`, `Weights` to `helidon-common`, do not use `@Priority` from Jakarta in SE


common/common/src/main/java/io/helidon/common/Weights.java
```diff
@@ -0,0 +1,153 @@
...
+    public static <S> Comparator<S> weightComparator() {
+        return (o1, o2) -> {
+            if (o1 == null) {
+                return o2 == null ? 0 : 1;
+            }
+            if (o2 == null) {
+                return -1;
+            }
+
+            double firstWeight = find(o1, -101);
```
danielkec, line 120: Some constant for magic number would be nicer

tomas-langer: Using a constant, added a comment

PR #4716, opened 2022-08-10
Unescape the keys when config is returned as a map in Main branch (#4678)

klustria:
* Unescape the keys when config is returned as a map

* Update copyright year

* Using as(Map.class).get() and as(new GenericType<Map>(){}).get() should have keys unescaped


tomas-langer (approved):

klustria:
Fixes https://github.com/oracle/helidon/issues/4398 in Main

PR #4627, opened 2022-07-27
Adopt review comments on doc updates

tjquinno:
Resolves #4626

Issue #4626, opened 2022-07-27
Revise various doc updates (metrics, OpenAPI) based on review comments


tjquinno:
## Environment Details
* Helidon Version: 3.0.0
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
Adopt review comments on the doc updates.

PR #4721, opened 2022-08-11
Adopt review comments on doc updates (#4627)

tjquinno:
Forward-port of doc comments #4626 from Lisa previously merged in 3.x into main.


ljamen (approved):
Changes look good to me, thank you.

Issue #4643, opened 2022-07-29
OOM with JAX-RS streaming output 


danielkec:
As reported by Tapash Paul, using StreamingOutput in Helidon with very large files leads to frequent OOMs

## Environment Details
* Helidon Version: 1.x, 2.x, 3.x
* Helidon MP

----------

## PR's
- [x] 4x #5063
- [x] 3x #4724
- [x] 2x #5062

## Problem Description
When streaming large data over JAX-RS StreamingOutput, ResponseWriter forces Netty to allocate too much ByteBuffs ending with OOM

## Steps to reproduce
```java

@HelidonTest
@DisableDiscovery
@AddBean(LeakTest.DownloadResource.class)
@AddExtension(ServerCdiExtension.class)
@AddExtension(JaxRsCdiExtension.class)
@AddExtension(CdiComponentProvider.class)
public class LeakTest {

    @Test
    void name(WebTarget target) throws IOException {

        InputStream is = target.path("/download")
                .request()
                .get(InputStream.class);
        while (0 != is.available()) {
            is.read();
        }
        is.close();
    }

    @Path("/download")
    public static class DownloadResource {

        private static final byte[] DATA_10MB = new byte[10485760];

        static {
            Random r = new Random();
            r.nextBytes(DATA_10MB);
        }

        @GET
        @Produces(MediaType.MULTIPART_FORM_DATA)
        public Response getPayload(
                @NotNull @QueryParam("fileName") String fileName) {
            StreamingOutput fileStream = output -> {

                // 5gb
                for (int i = 0; i < 500; i++) {
                    output.write(DATA_10MB);
                    output.flush();
                }

            };
            return Response
                    .ok(fileStream, MediaType.MULTIPART_FORM_DATA)
                    .build();
        }
    }
}
```

danielkec:
Simplest possible workaround is slowing down the writes so Netty can keep up sending the data and properly reusing ByteBuffs

```java
for (int i = 0; i < 500; i++) {
    if (PooledByteBufAllocator.DEFAULT.pinnedHeapMemory() > MEMORY_LIMIT_MB) {
        Thread.sleep(1000);
    }
    if (PooledByteBufAllocator.DEFAULT.pinnedDirectMemory() > MEMORY_LIMIT_MB) {
        Thread.sleep(1000);
    }
    output.write(DATA_10MB);
    output.flush();
}
```

spericas:
[...]

PR #4724, opened 2022-08-11
Watermarked response backpressure 3x

### LINKED ISSUE #3136: Watermark based backpressure for webserver
Investigate use of backpressure for not immediately flushed datachunks as suggested by @olotenko :

> Count the bytes that go through here, and allow up to a given amount to be buffered. Eg `onNext` counts how many bytes were in the buffers it has seen, and requests one more chunk if it has seen below watermark. Then `channel.write` completion Future subtracts the bytes it has just written. If this results in crossing the "watermark", then request 1. (because `onNext` stopped requesting more)
> 
> Watermarking approach is more predictable. Just deferring to `channel.write` completion would work poorly, if in the common case socket write completion does not happen during `onNext`, but will work better than watermarking, if there is this guarantee. (Ie that `onNext` attempts to call `channel.write`, through a few calls of intermediate routines, and that Netty completes the future before `onNext` returns, if the write fully fits in socket send buffers)

https://github.com/oracle/helidon/pull/3108#pullrequestreview-686454578

## PR's
- [ ] 4x #5063
- [ ] 3x #4724
- [ ] 2x #5062

danielkec:
Fixes #3136 and #4643


microprofile/server/src/test/java/io/helidon/microprofile/server/StreamingOutputLeakTest.java
```diff
@@ -0,0 +1,106 @@
...
+                .request()
+                .get(InputStream.class);
+        long size = 0;
+        while (is.read() != -1) {
+            size++;
+        }
+        is.close();
+
+        // Make sure all data has been read
+        assertThat(size, is(NUMBER_OF_BUFS * SIZE));       // 5 GB
```
spericas, line 82: // 2 GB ?

docs/config/io_helidon_webserver_WebServer.adoc
```diff
@@ -71,6 +71,22 @@ This is a standalone configuration type, prefix from configuration root: `server
...
 
  Default is `64*1024`
+|`backpressure-buffer-size` |long |`5242880` |Set a maximum length of the unflushed response data sending buffer can keep without applying backpressure.
+Depends on `backpressure-policy` what happens if max buffer size is reached.
+
+ Default is `5*1024*1024` - 5Mb
+|`backpressure-policy` | String | `LINEAR` |Sets the strategy for applying backpressure to the reactive stream
+of response data.
+
+* LINEAR - Data chunks are requested one-by-one after previous data chunk has been written to Netty's buffer, when
```
spericas, line 81: What are the use cases to support all these strategies? May be useful to include this info the doc as well.

webserver/webserver/src/main/java/io/helidon/webserver/BackpressureStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.helidon.webserver;
+
+/**
+ * Strategy for applying backpressure to the reactive stream of response data.
+ */
+public enum BackpressureStrategy {
+    /**
+     * Data chunks are requested one-by-one after previous data chunk has been written to Netty’s buffer,
```
spericas, line 24: Is it "written to Netty's buffer" or "given to Netty for writing"?

microprofile/server/src/test/java/io/helidon/microprofile/server/StreamingOutputLeakTest.java
```diff
@@ -0,0 +1,106 @@
...
+import static org.hamcrest.CoreMatchers.is;
+
+@HelidonTest
+@DisableDiscovery
+@AddBean(StreamingOutputLeakTest.DownloadResource.class)
+@AddExtension(ServerCdiExtension.class)
+@AddExtension(JaxRsCdiExtension.class)
+@AddExtension(CdiComponentProvider.class)
+@AddConfig(key = "server.backpressure-buffer-size", value = "20971520")//20Mb
+public class StreamingOutputLeakTest {
```
tomas-langer, line 53: should be package private

microprofile/server/src/test/java/io/helidon/microprofile/server/StreamingOutputLeakTest.java
```diff
@@ -0,0 +1,106 @@
...
+        Random r = new Random();
+        r.nextBytes(DATA_10MB);
+    }
+
+    /**
+     * Reproducer for issue #4643
+     */
+    @Test
+    @Disabled("Takes too long (around 2 minutes), root-cause covered by WaterMarkedBackpressureTest")
+    void streamingOutput(WebTarget target) throws IOException {
```
tomas-langer, line 70: The test is disabled and there is no other test in this class, so this class is practically useless.
This should not be included into Helidon.
Either we should come up with an integration tests profile for "longer" tests, or change the test to be usable.

webserver/webserver/src/test/java/io/helidon/webserver/WaterMarkedBackpressureTest.java
```diff
@@ -0,0 +1,173 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static io.helidon.webserver.BackpressureStrategy.AUTO_FLUSH;
+import static io.helidon.webserver.BackpressureStrategy.LINEAR;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+public class WaterMarkedBackpressureTest {
```
tomas-langer, line 44: Class should be package local.

webserver/webserver/src/test/java/io/helidon/webserver/WaterMarkedBackpressureTest.java
```diff
@@ -0,0 +1,173 @@
...
+                    .flatMap(WebClientResponse::content)
+                    .forEach(chunk -> {
+                        byte[] bytes = chunk.bytes();
+                        receivedSize.addAndGet(bytes.length);
+                        chunk.release();
+                    })
+                    .onError(Throwable::printStackTrace);
+
+            try {
+                Single.create(firstBatchSent, true).await(Duration.ofSeconds(5));
```
tomas-langer, line 103: If an exception is expected, use `assertThrows` instead.

webserver/webserver/src/test/java/io/helidon/webserver/WaterMarkedBackpressureTest.java
tomas-langer: Also this looks like an integration test. So it should be in integration tests, or marked as an IT within this module.

tomas-langer (changes requested):

docs/config/io_helidon_webserver_WebServer.adoc
danielkec: Speed vs memory as always, I wouldn't go that far with explaining use-cases, users who know how to work with backpressure on reactive pipeline have enough information, others shouldn't really touch it. 95% use-cases should fit in to the default setting.

webserver/webserver/src/test/java/io/helidon/webserver/WaterMarkedBackpressureTest.java
danielkec: its not expected

danielkec:
Upper one is the one with this PR changes, lower one is 3.0.1 release 
image

tomas-langer (approved):
LGTM

PR #4745, opened 2022-08-16
Make Application subclasses available via our context during Feature executions

spericas:
A global/shared Feature is executed for each Application subclass. This change makes the Application subclass available to the Feature during its execution. Just like for requests, the availability is through the Helidon context. Updated functional tests for multiple apps.

### DIFF
--- a/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
+++ b/microprofile/server/src/main/java/io/helidon/microprofile/server/ServerCdiExtension.java
@@ -58,6 +58,8 @@ import javax.ws.rs.ext.ParamConverterProvider;
 
 import io.helidon.common.Prioritized;
 import io.helidon.common.configurable.ServerThreadPoolSupplier;
+import io.helidon.common.context.Context;
+import io.helidon.common.context.Contexts;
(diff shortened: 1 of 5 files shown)


tjquinno:
IIUC the developer's code (if it cares) needs to check which app the feature corresponds to. I see that the changes to the main (non-test) Helidon code now _set_ the app subclass in the context but only the tests (i.e., the user code) retrieve it. 

Or put another way, there is no other Helidon code that directly uses features and therefore might need to look at the app subclass?

spericas:
> IIUC the developer's code (if it cares) needs to check which app the feature corresponds to. 

 Yes, a Feature is executed for each app for which it is part of.

>I see that the changes to the main (non-test) Helidon code now _set_ the app subclass in the context but only the tests (i.e., the user code) retrieve it.

 Yes, this is mostly for user's logic. We already make it available during execution (e.g. in a request filter) and also as part of the Helidon context. 

> 
> Or put another way, there is no other Helidon code that directly uses features and therefore might need to look at the app subclass?

 Right, I don't believe we use Features in Helidon (we use AutoDiscoverable from Jersey). This is mostly for app developers.

tjquinno (approved):

Issue #4725, opened 2022-08-11
CORS does not deal with `Forwarded` or `X-Forwarded-xxx` headers


tjquinno:
## Environment Details
* Helidon Version: 2.x, 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
Helidon's CORS support relies on the `Host` header (among other things) to make its decisions. 

If intermediaries--load balancers, etc.--lie between the real client and the service, then when the request reaches the Helidon service the `Host` value no longer reflects the actual host.

HTTP specifies the `Forwarded` header which conveys `by`, `for`, `host` (which might include the port), and `protocol` as specified by the originating client.

Non-standard but widespread headers `X-Forwarded-Host`, `X-Forwarded-For`, `X-Forwarded-Proto`, and `X-Forwarded-Port` express the same information.

Helidon's CORS processing can be enhanced to prefer `Forwarded` if present, then `X-Forwarded-xxx` if present, and then `Host` for deriving the "effective host" to use in CORS decision-making.

Depends on #5824

tomas-langer:
My recommended design (for discussion):

- add a configuration option to use non-Host headers (defaults to false)
- add a configuration option to provide a list of headers that can be used to determine "frontend host", will be used in order of definition; if any of them is `Forwarded`, other headers would be ignored if `Forwarded` is present (as it should cover all cases); e.g. the options should be "forwarded, x-forwarded-host, x-forwarded-port, x-forwarded-proto" etc., so we always have a complete list of headers we should support (security measure)
- add a method `ServerRequest.usedAuthority()` (or `originalAuthority` or similar) to obtain the authority as constructed from configured HTTP headers (may be same as `HOST` header in the simple case of no proxying or default config)
- add a method `ServerRequest.usedProtocol()` to obtain the original protocol (such as `http` or `https`)
We may add additional methods, or we may add a single method that just returns `URI` - this would be a bit more expensive, and would have to be done lazily (as should any access to request headers anyway)

tjquinno:
1. Adding `usedXXX` methods which take the headers into account implies that most parts of Helidon should use the raw values, ignoring the headers, and those parts would continue to invoke `authority()` and `protocol` as they do now. 
   
   Is that true? Could the argument be made that most parts of Helidon _should_ consider the headers, and any that do not should invoke `rawAuthority` and `rawProtocol` explicitly?
[...]

PR #4767, opened 2022-08-22
Support for Forwarded, X-Forwarded-Host, X-Forwarded-Proto headers

tomas-langer:
to get used authority and protocol.

This should be used for redirections and CORS.

Related to #4725

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
+++ b/common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
@@ -93,6 +93,8 @@ enum HeaderEnum implements Http.HeaderName {
     X_HELIDON_CN("X-HELIDON-CN"),
     X_FORWARDED_FOR("X-Forwarded-For"),
     X_FORWARDED_HOST("X-Forwarded-Host"),
+    X_FORWARDED_PORT("X-Forwarded-Port"),
+    X_FORWARDED_PREFIX("X-Forwarded-Prefix"),
     X_FORWARDED_PROTO("X-Forwarded-Proto");
 
     private static final Map<String, Http.HeaderName> BY_NAME;
--- a/common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
+++ b/common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
@@ -90,7 +90,10 @@ enum HeaderEnum implements Http.HeaderName {
     VARY("Vary"),
     WARNING("Warning"),
     WWW_AUTHENTICATE("WWW-Authenticate"),
-    X_HELIDON_CN("X-HELIDON-CN");
+    X_HELIDON_CN("X-HELIDON-CN"),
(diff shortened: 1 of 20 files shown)


common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
```diff
@@ -90,7 +90,10 @@ enum HeaderEnum implements Http.HeaderName {
     VARY("Vary"),
     WARNING("Warning"),
     WWW_AUTHENTICATE("WWW-Authenticate"),
-    X_HELIDON_CN("X-HELIDON-CN");
+    X_HELIDON_CN("X-HELIDON-CN"),
+    X_FORWARDED_FOR("X-Forwarded-For"),
+    X_FORWARDED_HOST("X-Forwarded-Host"),
+    X_FORWARDED_PROTO("X-Forwarded-Proto");
```
tjquinno, line 96: `X-Forwarded-Port` also seems to be in use in the wild, in particular by OCI LBs. Here is a sanitized excerpt from what a Helidon server in OCI receives:

```
, x-forwarded-for=[w.x.y.z]
, x-forwarded-host=[a.b.c.com]
, x-forwarded-port=[443]
, x-forwarded-prefix=[/some/path]
, x-forwarded-proto=[https]
```

We probably need to declare that here and handle it in all the right places. (I won't tag each one with PR comments.)

tjquinno (changes requested):

common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
tomas-langer: If I get it, we should not need x-forwarded-for, as that is the client (and we do not really care about the client).
We should care for the rest. If I get it, then a redirect should be:
```
{proto}://{host}:{port}/?{prefix}
```
(slash only inserted if prefix does not contain it). If any is missing, use the actual information from the request.
Is that correct?

tomas-langer: If the name prefix is correct, then it probably should be
`{proto}://{host}:{port}/?{prefix}/?{path}`

tomas-langer: I have added the missing headers to the constants. Not using them yet in the method to resolve original authority

tjquinno: I did not say `port` and `prefix` were part of `Forwarded`. They are part of the non-standard `X-Forwarded-*` family. If `Forwarded` is not present but the `X-Forwarded-*` headers are then Helidon should use the ones that are present and that apply: proto, host, port, prefix.

`X-Forwarded-For` is in the example output only because it is part of what the OCI LBs seem to include. I did not mean to suggest that Helidon should use it for computing the authority or anything else we need for redirection or CORS decision-making.

PR #4769, opened 2022-08-23
RestClient proxy test exclusion removed

Verdent:
Jersey dependency version bump to 3.0.8 was required to pass the RestClient proxy test


tomas-langer (approved):
LGTM

Issue #4764, opened 2022-08-22
refactor helidon config for helidon 4


trentjeff:
# Introduce pico-config (pico/config)
* Extract parts of helidon-config and promote into pico-config
  * Config (only parts dealing with immutable/get)
  * ConfigValue
 * Exception classes
* Introduce pico-api (Contract, ExternalContract)

trentjeff:
Merged.

PR #4776, opened 2022-08-23
Refactor helidon config 4.0

trentjeff:
Fix for #4764

### DIFF
--- a/common/common/pom.xml
+++ b/common/common/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2022 Oracle and/or its affiliates.
+    Copyright (c) 2017, 2022 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/common/config/pom.xml
+++ b/common/config/pom.xml
@@ -1,19 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-  ~ Copyright (c) 2017, 2022 Oracle and/or its affiliates.
-  ~
-  ~ Licensed under the Apache License, Version 2.0 (the "License");
-  ~ you may not use this file except in compliance with the License.
-  ~ You may obtain a copy of the License at
-  ~
-  ~     http://www.apache.org/licenses/LICENSE-2.0
-  ~
-  ~ Unless required by applicable law or agreed to in writing, software
-  ~ distributed under the License is distributed on an "AS IS" BASIS,
-  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-  ~ See the License for the specific language governing permissions and
-  ~ limitations under the License.
-  -->
+
+    Copyright (c) 2022 Oracle and/or its affiliates.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
 
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
--- a/common/config/pom.xml
+++ b/common/config/pom.xml
@@ -1,21 +1,19 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-
-    Copyright (c) 2022 Oracle and/or its affiliates.
-
-    Licensed under the Apache License, Version 2.0 (the "License");
-    you may not use this file except in compliance with the License.
-    You may obtain a copy of the License at
-
-        http://www.apache.org/licenses/LICENSE-2.0
-
-    Unless required by applicable law or agreed to in writing, software
-    distributed under the License is distributed on an "AS IS" BASIS,
-    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-    See the License for the specific language governing permissions and
-    limitations under the License.
-
--->
+  ~ Copyright (c) 2017, 2022 Oracle and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
 
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
@@ -25,8 +23,8 @@
         <artifactId>helidon-common-project</artifactId>
         <version>4.0.0-SNAPSHOT</version>
     </parent>
-    <artifactId>helidon-common</artifactId>
-    <name>Helidon Common</name>
+    <artifactId>helidon-common-config</artifactId>
+    <name>Helidon Common config</name>
 
     <properties>
         <!-- Helidon common should be backward compatible with Java 11 -->
--- a/common/config/pom.xml
+++ b/common/config/pom.xml
@@ -26,67 +26,15 @@
         <version>4.0.0-SNAPSHOT</version>
     </parent>
     <artifactId>helidon-common-config</artifactId>
-    <name>Helidon Common config</name>
+    <name>Helidon Common Config</name>
 
     <properties>
         <!-- Helidon common should be backward compatible with Java 11 -->
         <version.java>11</version.java>
     </properties>
 
-    <dependencies>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-all</artifactId>
-            <scope>test</scope>
-        </dependency>
-    </dependencies>
-
     <build>
         <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-resources-plugin</artifactId>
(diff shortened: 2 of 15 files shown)


romain-grecourt:
Can we separate pico into its own dedicated PR ?

pico/LICENSE.txt
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
romain-grecourt, line 1: There is already a LICENSE file at the root of the repo. Is this intended ?

config/config/pom.xml
```diff
@@ -42,6 +42,10 @@
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
+        <dependency>
```
romain-grecourt, line 45: This feels backwards.

The declarative API should be on layered on top of the programmatic API, right ?
As it stands `helidon-config` pulls `pico-config` even if pico isn't used.

pico/config/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+    <groupId>io.helidon.pico.config</groupId>
+    <artifactId>helidon-pico-config-project</artifactId>
+    <name>Helidon Pico Config</name>
+    <packaging>pom</packaging>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+
+        <!--        <maven.deploy.skip>true</maven.deploy.skip>-->
```
romain-grecourt, line 40: Commented-out.

pico/config/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+        <!--        <maven.sources.skip>true</maven.sources.skip>-->
+        <!--        <maven.javadoc.skip>true</maven.javadoc.skip>-->
+        <!--        <spotbugs.skip>true</spotbugs.skip>-->
+        <!--        <dependency-check.skip>true</dependency-check.skip>-->
+        <!--        <helidon.services.skip>true</helidon.services.skip>-->
+    </properties>
+
+    <modules>
+<!--        <module>api</module>-->
```
romain-grecourt, line 49: Commented-out.

pico/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+    <groupId>io.helidon.pico</groupId>
+    <artifactId>helidon-pico-project</artifactId>
+    <name>Helidon Pico Project</name>
+
+    <packaging>pom</packaging>
+
+    <properties>
+        <version.java>11</version.java>
+
+        <!--        &lt;!&ndash; javadoc fails on Java 19 &ndash;&gt;-->
```
romain-grecourt, line 41: Commented-out.

pico/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <!--        <maven.javadoc.skip>true</maven.javadoc.skip>-->
+
+<!--        <javax.injection.version>1</javax.injection.version>-->
```
romain-grecourt, line 44: Commented-out.

integrations/cdi/datasource/pom.xml
```diff
@@ -47,6 +47,11 @@
             <artifactId>jakarta.annotation-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
romain-grecourt, line 50: Not sure that belongs in this PR.

trentjeff: it was required to build since some of the scopes changed from provided to runtime as part of this change.

pico/LICENSE.txt
trentjeff: There was a point in time where pico was going to be a top-level module. Will discuss it with @tomas-langer - currently it is a child of helidon (obviously), but now that we are close to having our our org on gitHub this might move.

romain-grecourt: Since you are adding pico as a module as part of this PR, the license file should be removed.
You can add the license file back If/when pico moves into its own repository.

config/config/pom.xml
tomas-langer: There is a chicken-egg problem with config. It can be solved two ways:
1. create a config API that is implemented by `helidon-config` and backward compatible
2. make `helidon-config` the config API and have a separate implementation module (not backward compatible)

The first option seems to have the least impact on Helidon

pico/api/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.inject</groupId>
+            <artifactId>jakarta.inject-api</artifactId>
+            <scope>compile</scope>
```
wojtask9, line 41: I don't understand idea of this.
Starting with 4.x jakarta.* annotations will be required for SE ?

trentjeff: The jakarta.inject module is needed (as compile scope) for the definition of the Provider class, sadly in the same package with other annotations that would otherwise only be needed as provided scope.

common/config/pom.xml
```diff
@@ -0,0 +1,100 @@
...
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
```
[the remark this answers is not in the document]
trentjeff: resolved & removed/pushed

common/common/pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2017, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2022 Oracle and/or its affiliates.
```
tomas-langer, line 4: This is wrong copyright change. It must be `2017, 2022`. Please revert this change

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+ * limitations under the License.
+ */
+
+package io.helidon.common.config;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * <h2>Configuration</h2>
```
tomas-langer, line 23: Adding headers to javadocs is tricky and not really needed here, when we have a single sentence description

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+     * <p>
+     * The fully-qualified key is a sequence of tokens derived from the name of
+     * each node along the path from the config root to the current node. Tokens
+     * are separated by {@code .} (the dot character). See {@link #name()} for
+     * more information on the format of each token.
+     *
+     * @return current config node key
+     * @see #name()
+     */
+    BaseKey key();
```
tomas-langer, line 41: BaseKey? This should return `Key`. Our API should be `Key`, `BaseKey` I would expect to be an abstract class I can extend, not API

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+    /**
+     * Returns {@code true} if this node exists and is a leaf node (has no
+     * children).
+     * <p>
+     * A leaf node has no nested configuration subtree and has a single value.
+     *
+     * @return {@code true} if the node is existing leaf node, {@code false}
+     *         otherwise.
+     */
+    boolean isLeaf();
```
tomas-langer, line 164: Not really needed. This is basically `!isObject() && !isKey() && hasValue()`.
Also method `hasValue()` is missing

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+
+    /**
+     * Returns list of specified type.
+     *
+     * @param type type class
+     * @param <T>  type of list elements
+     * @return a typed list with values
+     * @throws io.helidon.common.config.ConfigException in case of problem to map property value.
+     */
+    default <T> ConfigValue<List<T>> asList(Class<T> type) throws ConfigException {
```
tomas-langer, line 208: This should not have a default implementation. Esp. not an implementation like this

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+        throw new ConfigException("unsupported; do you have a valid config module in the classpath?");
+    }
+
+    /**
+     * Transform all leaf nodes (values) into Map instance.
+     *
+     * @return new Map instance that contains all config leaf node values
+     * @throws io.helidon.common.config.ConfigException in case the node is Type#MISSING.
+     */
+    default ConfigValue<Map<String, String>> asMap() throws ConfigException {
```
tomas-langer, line 218: dtto

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+     *  key-token = *( unescaped / escaped )
+     *  unescaped = %x00-2D / %x2F-7D / %x7F-10FFFF
+     *            ; %x2E ('.') and %x7E ('~') are excluded from 'unescaped'
+     *    escaped = "~" ( "0" / "1" )
+     *            ; representing '~' and '.', respectively
+     * }</pre>
+     *
+     * @see Config#key()
+     */
+    interface BaseKey extends Comparable<BaseKey> {
```
tomas-langer, line 246: This should be named `Key`. BaseKey would mean an abstract class. Also this is one of the main APIs of Config, so not prefixes should be used.

common/config/src/main/java/io/helidon/common/config/Config.java
```diff
@@ -0,0 +1,346 @@
...
+         * Returns instance of Key that represents key of parent config node.
+         * <p>
+         * If the key represents root config node it throws an exception.
+         *
+         * @return key that represents key of parent config node.
+         * @see #isRoot()
+         * @throws IllegalStateException in case you attempt to call this method on a root node
+         * @throws io.helidon.common.config.ConfigException if not defined
+         */
+        default BaseKey parent() throws ConfigException {
```
tomas-langer, line 292: Do not add default implementations like this

common/config/src/main/java/io/helidon/common/config/package-info.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2017, 2022 Oracle and/or its affiliates.
```
tomas-langer, line 2: Copyright should be just `2022`, this file did not exist in 2017

common/pom.xml
```diff
@@ -32,8 +32,9 @@
 
     <modules>
         <module>common</module>
-        <module>reactive</module>
+        <module>config</module>
         <module>configurable</module>
+        <module>reactive</module>
```
tomas-langer, line 37: why is this reordered? Please simplify the change

config/config/etc/spotbugs/exclude.xml
```diff
@@ -83,5 +83,20 @@
         <Method name="parse"/>
         <Bug pattern="RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE"/>
     </Match>
+    <Match>
+        <!-- Path shadows common/config -->
```
tomas-langer, line 87: what does `Path` mean in these descriptions? I do not understand this explanation

tomas-langer (changes requested):

common/config/src/main/java/io/helidon/common/config/Config.java
trentjeff: There is no isKey() on either common/config/Config or config/config.  I will add hasValue() decl on common/config.

tomas-langer: sorry, meant `isList()` not `isKey()`

tomas-langer (approved):
LGTM

PR #4798, opened 2022-08-25
Round Robin added as DNS resolver option

### LINKED ISSUE #4744: Allow custom DNS resolver in Webclient
## Environment Details
* Helidon Version: 2.5.x
* Helidon SE
* JDK version: Java 11
* OS:

----------

When using Helidon Webclient, there is no way how to override (Netty) default DNS resolver. It should be possible to provide custom resolver implementation.

Our specific scenario:

We are implementing reverse proxy in Helidon for a service whose DNS resolves to multiple IPs (for load-balancing purposes). Without ability to override the DNS resolver, we cannot spread the load among the target IPs.

Verdent:
Fixes #4744

### DIFF
--- a/webclient/webclient/src/main/java/io/helidon/webclient/WebClientConfiguration.java
+++ b/webclient/webclient/src/main/java/io/helidon/webclient/WebClientConfiguration.java
@@ -83,6 +83,7 @@ class WebClientConfiguration {
     private final URI uri;
     private final boolean validateHeaders;
     private final boolean relativeUris;
+    private final DnsResolverType dnsResolverType;
 
     /**
      * Creates a new instance of client configuration.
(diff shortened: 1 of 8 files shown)


webclient/webclient/src/main/java/io/helidon/webclient/WebClientConfiguration.java
```diff
@@ -316,6 +322,7 @@ boolean relativeUris() {
         private MessageBodyWriterContext writerContext;
         private boolean validateHeaders;
         private boolean relativeUris;
+        public DnsResolverType dnsResolverType;
```
tomas-langer, line 325: Should be private

tomas-langer (changes requested):
This brings a new dependency.
What is the size?

Verdent:
New dependency is `netty-resolver-dns` which brings along `netty-codec-dns`.

Complete size is: 217 kb

tomas-langer (approved):

barchetta (approved):

PR #4804, opened 2022-08-27
Http exception refactoring

tomas-langer:
HttpException moved to common http module.
RequestException (new thing) moved to common http module.

Document describing how exceptions should be handled in HTTP added to repository (internal docs).

`ErrorHandlers` added to Nima (to be fully implemented later - already on task list).


nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http/ErrorHandlers.java
```diff
@@ -0,0 +1,107 @@
...
+            if (hasErrorHandler(e.getCause())) {
+                handleError(ctx, request, response, e.getCause());
+            } else {
+                handleError(ctx, request, response, e);
+            }
+        } catch (HttpException e) {
+            handleError(ctx, request, response, e);
+        } catch (RuntimeException e) {
+            handleError(ctx, request, response, e);
+        }
```
danielkec, line 74: I would rather catch Throwable to be sure

danielkec (dismissed):

common/http/src/main/java/io/helidon/common/http/ForbiddenException.java
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common.http;
+
+/**
+ * A runtime exception indicating a {@link io.helidon.common.http.Http.Status#NOT_FOUND_404 not found}.
```
ljnelson, line 20: Did you mean 403?

nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http/ErrorHandlers.java
tomas-langer: I don't think we want to handle "Throwable" - that would contain things like `OutOfMemoryError` and that should not be a case for exception handling by users. We can add it once we add error handling to Nima...

common/http/src/main/java/io/helidon/common/http/ForbiddenException.java
tomas-langer: Fixed

barchetta (approved):

PR #4809, opened 2022-08-29
Nima: a few optimizations

tomas-langer:
- compare strings if headers have single value
- support for blocking writes when queue is disabled


danielkec (approved):

danielkec:
image

PR #4819, opened 2022-08-31
Shutdown executors while stopping the server

spericas:
Attempt to shutdown executors gracefully while the server is being stopped. After shutdown, no new `ConnectionHandler` will be accepted for execution (submit() call on executor will fail).


tomas-langer (approved):
LGTM

PR #4820, opened 2022-08-31
3.x upgrade reactive streams

barchetta:
Upgrade reactive-streams to 1.0.4 which changes license from CC-0 to MIT-0
Remove duplicate entry for netty-resolver-dns from dependencyManagement


barchetta:
Did this upgrade in #5045, will do license update in another PR.

PR #4823, opened 2022-09-01
Port and cleanup of old reactive tests for Nima bulkheads

spericas:
Port and cleanup of old reactive tests for Nima bulkheads. Restored Async class for convenience, now implemented using virtual threads.

### DIFF
--- a/nima/fault-tolerance/src/main/java/io/helidon/nima/faulttolerance/Async.java
+++ b/nima/fault-tolerance/src/main/java/io/helidon/nima/faulttolerance/Async.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


nima/fault-tolerance/src/main/java/io/helidon/nima/faulttolerance/Async.java
```diff
@@ -0,0 +1,58 @@
...
+package io.helidon.nima.faulttolerance;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Supplier;
+
+/**
+ * Runs synchronous suppliers asynchronously using virtual threads. Includes
+ * convenient static method to avoid creating instances of this class.
+ */
+public interface Async {
```
tomas-langer, line 26: I have excluded async intentionally, as it does not really make sens - you can execute anything as a blocking operation, and if you need to run multiple things in parallel, you are better off using an executor service directly.
If you feel strongly we should keep this in for symmetry with SE and MP, then OK

tomas-langer (approved):
Please see comment before merging

nima/fault-tolerance/src/main/java/io/helidon/nima/faulttolerance/Async.java
spericas: I thought it would be useful to use virtual threads for running concurrent actions, to further test the infrastructure. Using `supplyAsync` created real threads AFAICT.

tomas-langer: In that case we should also support supplying a custom executor service (in case the user wants to choose whether to use platform or virtual threads)

spericas: Yes, I agree. I will add that in a follow-up PR.

PR #4824, opened 2022-09-01
4749 LRA fixes backport

### LINKED ISSUE #4749: LRA - hardcoded coordinator call timeout
Already fixed in the 3.x and main, backport fix is missing

danielkec:
Fixes #4749


Verdent (approved):

lra/coordinator/server/src/main/resources/application.yaml
```diff
@@ -28,7 +28,7 @@ helidon.lra.coordinator:
     connection:
       url: "jdbc:h2:file:./lra-coordinator"
       username: "h2"
-      password: ""
+      password: ${EMPTY}
```
tomas-langer, line 31: This is not correct for version 2.x, we do not understand the ${EMPTY} replacement

microprofile/tests/arquillian/src/main/java/io/helidon/microprofile/arquillian/HelidonDeployableContainer.java
```diff
@@ -422,6 +455,22 @@ public void undeploy(Descriptor descriptor) {
         // No-Op
     }
 
+    /**
+     * Injects the base metric registry and cleans up all metrics in preparation to run another
+     * Arquillian test in the same VM. Without this cleanup, metrics added by a previous test
+     * would be available and may cause failures.
+     */
+    private void cleanupBaseMetrics() {
```
tomas-langer, line 463: This change seems to be unrelated to LRA. Is it really needed?

tomas-langer (changes requested):

microprofile/tests/arquillian/src/main/java/io/helidon/microprofile/arquillian/HelidonDeployableContainer.java
danielkec: I have also backported chunk of changes to arqillian deployer to avoid h2 to be generated outside of the target folder, metrics cleanup invocation was part of those changes with aborted deployment cleanup and etc, didn't want to backport just a part of it. 
These changes greatly improve test isolation which is mostly visible only when some tck test fails - only test related to particular change fail, not all the other broken by polluted deployment.

lra/coordinator/server/src/main/resources/application.yaml
danielkec: fixed

tomas-langer (approved):

PR #4827, opened 2022-09-01
Upgrades OCI to 2.45.0 on helidon-3.x

ljnelson:


ljnelson:
@barchetta If you could take a look I'd appreciate it.  Although the OCI BOM contents will not (should not) influence anything on this branch, I feel that they need to be there to ensure backwards compatibility.

barchetta:
@ljnelson I suppose this is OK (keeping in the ugly BOM workaround even though we don't need it). Can you update to latest OCI SDK and take out of draft?

ljnelson:
Roger

barchetta (approved):

PR #4871, opened 2022-09-13
Remove virtual executor util and use common config in common modules.

### LINKED ISSUE #4869: Remove VirtualExecutorUtil
As Helidon 4.x is fully based on Virtual threads and JDK supporting it, we can remove this utility.

This should be fixed once virtual threads are no longer preview feature, as this would make a lot of modules require to use `--enable-preview`

tomas-langer:
Resolves #4869 

The original attempted fix is too intrusive (it requires `--enable-preview` for any module that depends on `common-configurable`, which is not desired). I will keep the code as is, we should remove this once virtual threads are no longer preview feature.

Requires --enable-preview to be added to a few test/compiler configs.
Moved most commonly used shortcut methods to common config (`asInt()` etc), as we do not want a dependency on full config from common modules, and this simplifies the code (and leaves it untouched).
Added method `map(Function<common.config.Config, T> mapper)` to common config, as it cannot be compatible with `as(Function<Config, T> mapper)` in config. This makes it much easier to migrate. We can eventually deprecate the `as(mapper)` method if we decide to.


common/configurable/src/main/java/io/helidon/common/configurable/WrappedVirtualExecutor.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
+
+class WrappedVirtualExecutor implements ExecutorService {
```
trentjeff, line 31: fundamentally, why do we need executors in a "common configurable" module?  Seems like this should be a higher level function...  Needs javadoc so maybe I'd understand it better after it was written.

common/configurable/src/main/java/io/helidon/common/configurable/spi/ExecutorServiceSupplierObserver.java
```diff
@@ -15,38 +15,35 @@
...
- *         <li>the index of this supplier among suppliers in the same category.</li>
- *     </ul>
- *     Further, executor services furnished by the suppliers are identified to observers using:
- *     <ul>
- *         <li>the executor service itself, and</li>
- *         <li>the index of the executor service among those from the same supplier.</li>
- *     </ul>
- *     The consuming observers can use this identifying information however makes sense for them.
- *
+ * This component identifies suppliers to observers using:
```
trentjeff, line 25: can we keep this in config/config instead of in common?

tomas-langer: I don't understand your comment. This class has been here before. It has nothing to do with config - it is support for observability of executors (for metrics).

common/configurable/src/main/java/io/helidon/common/configurable/WrappedVirtualExecutor.java
tomas-langer: This class does not need javadoc, as it is not generated for it (it is a package local class).
This is a refactoring of original approach that used reflection. The new approach is reflection free.
Common configurable module was created for classes that are used by multiple other modules that can be configured through config. Executor services are used for example by Fault tolerance, MP. 
Theoretically we could extract executor services into a new module, such as `common-executors` if we wanted to

tomas-langer:
Closing this PR, as the change is too intrusive. Will create a new one with switch to common config in common configurable.

PR #4876, opened 2022-09-14
4.x - Custom DNS resolvers for Nima

### LINKED ISSUE #4896: Custom DNS resolvers for Nima
Create similar functionality available in Netty to support custom DNS resolvers.

Round-robin impl needs to be provided as well.

Verdent:
Added custom DNS resolving support to Nima client
Added Round Robin implementation

Fixes #4896

### DIFF
--- a/nima/http2/webclient/src/main/java/io/helidon/nima/http2/webclient/ClientRequestImpl.java
+++ b/nima/http2/webclient/src/main/java/io/helidon/nima/http2/webclient/ClientRequestImpl.java
@@ -219,7 +219,6 @@ class ClientRequestImpl implements Http2ClientRequest {
     private Http2ClientStream reserveStream() {
         if (explicitConnection == null) {
             ConnectionKey connectionKey = new ConnectionKey(uri.scheme(),
-                                                            uri.authority(),
                                                             uri.host(),
                                                             uri.port(),
                                                             tls,
--- a/nima/http2/webclient/src/main/java/io/helidon/nima/http2/webclient/ClientRequestImpl.java
+++ b/nima/http2/webclient/src/main/java/io/helidon/nima/http2/webclient/ClientRequestImpl.java
@@ -36,6 +36,7 @@ import io.helidon.nima.common.tls.Tls;
 import io.helidon.nima.http2.Http2Headers;
 import io.helidon.nima.webclient.ClientConnection;
 import io.helidon.nima.webclient.ClientRequest;
+import io.helidon.nima.webclient.ConnectionKey;
 import io.helidon.nima.webclient.UriHelper;
 
 class ClientRequestImpl implements Http2ClientRequest {
@@ -217,7 +218,13 @@ class ClientRequestImpl implements Http2ClientRequest {
 
     private Http2ClientStream reserveStream() {
         if (explicitConnection == null) {
-            ConnectionKey connectionKey = new ConnectionKey(uri.scheme(), uri.authority(), uri.host(), uri.port(), tls);
+            ConnectionKey connectionKey = new ConnectionKey(uri.scheme(),
+                                                            uri.authority(),
+                                                            uri.host(),
+                                                            uri.port(),
+                                                            tls,
+                                                            client.dnsResolver(),
+                                                            client.dnsAddressLookup());
 
             // this statement locks all threads - must not do anything complicated (just create a new instance)
             return CHANNEL_CACHE.computeIfAbsent(connectionKey,
--- a/nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/DefaultDnsResolverProvider.java
+++ b/nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/DefaultDnsResolverProvider.java
@@ -14,12 +14,23 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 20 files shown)


nima/http2/webclient/src/main/java/io/helidon/nima/http2/webclient/ClientRequestImpl.java
```diff
@@ -217,7 +218,13 @@ private Http2Headers prepareHeaders(WritableHeaders<?> headers) {
 
     private Http2ClientStream reserveStream() {
         if (explicitConnection == null) {
-            ConnectionKey connectionKey = new ConnectionKey(uri.scheme(), uri.authority(), uri.host(), uri.port(), tls);
+            ConnectionKey connectionKey = new ConnectionKey(uri.scheme(),
+                                                            uri.authority(),
```
tomas-langer, line 222: `authority` is a combination of `host` and `port`, we should not use both at the same time.

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/DefaultDnsResolverProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.helidon.nima.webclient;
+
+import io.helidon.nima.webclient.spi.DnsResolver;
+import io.helidon.nima.webclient.spi.DnsResolverProvider;
+
+/**
+ * Provider of the {@link DefaultDnsResolver} instance.
+ */
+public class DefaultDnsResolverProvider implements DnsResolverProvider {
```
tomas-langer, line 25: If this is an implementation of service loader, please add explicit default constructor with javadoc describing that it should only be used by service loader.
If we expect users to create instances, add `public static Defa... create()`

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/NoDnsResolverProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.helidon.nima.webclient;
+
+import io.helidon.nima.webclient.spi.DnsResolver;
+import io.helidon.nima.webclient.spi.DnsResolverProvider;
+
+/**
+ * Provider of the {@link NoDnsResolver} instance.
+ */
+public class NoDnsResolverProvider implements DnsResolverProvider {
```
tomas-langer, line 25: Same as above - if service loader, please add public constructor with javadocs.
Also this should contain a `@Weight` or implement `Weighted`, to have lower weight than other resolvers, so when multiple are on classpath, this one does not get used

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/RoundRobinDnsResolverProvider.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.helidon.nima.webclient;
+
+import io.helidon.nima.webclient.spi.DnsResolver;
+import io.helidon.nima.webclient.spi.DnsResolverProvider;
+
+/**
+ * Provider of the {@link RoundRobinDnsResolver} instance.
+ */
+public final class RoundRobinDnsResolverProvider implements DnsResolverProvider {
```
tomas-langer, line 25: As above for other service loader implementations. Also add weights (if we have any preferred order of loading)

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/spi/DnsResolverProvider.java
```diff
@@ -14,12 +14,25 @@
...
  */
 
-package io.helidon.nima.http2.webclient;
+package io.helidon.nima.webclient.spi;
 
-import io.helidon.nima.common.tls.Tls;
+/**
+ * Provider interface for custom DNS resolvers.
+ */
+public interface DnsResolverProvider {
```
tomas-langer, line 22: This is an SPI and it is called `Provider`, yet I do not see any `uses` statement in `module-info.java`. 
If this is an omission, please fix.
If this is by design (and users are expected to provide an instance), we need to use factory methods for implementations and not default constructors.

tomas-langer (changes requested):

nima/http2/webclient/src/main/java/io/helidon/nima/http2/webclient/ClientRequestImpl.java
Verdent: fixed

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/DefaultDnsResolverProvider.java
Verdent: fixed

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/NoDnsResolverProvider.java
Verdent: fixed

tomas-langer (approved):

PR #4895, opened 2022-09-15
Remove module-info files from examples

arjav-desai:
Porting https://github.com/oracle/helidon/pull/4894 Addresses https://github.com/oracle/helidon/issues/4847


bundles/config/pom.xml
```diff
@@ -19,6 +19,18 @@
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <build>
```
barchetta, line 22: This is already handled in the project pom, right? So these pom changes should not be needed.

barchetta (changes requested):

bundles/config/pom.xml
arjav-desai: Yes, seems rebase on my branch/fork is off so fixing that.

arjav-desai: Reverted the change causing this diff.

barchetta (approved):

PR #4897, opened 2022-09-15
Fixed race condition in the OCI Metrics integration test between retrieval of metrics from registry and asserting that from expected results

klustria:
To fix the issue, here are the list of changes made:

1. Used CountDownLatches to signal when to start testing, for example, test only after results has been retrieved.
2. Make OciMetricsCdiExtension Priority higher than MetricsCdiExtension so that it will only start after MetricsCdiExtension has completed.

### DIFF
--- a/integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
+++ b/integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
@@ -55,9 +55,8 @@ public class OciMetricsSupportTest {
 
     private static volatile Double[] testMetricUpdateCounterValue = new Double[2];
     private static volatile int testMetricCount = 0;
-    // Use countDownLatches to signal when to start testing, for example, test only after results has been retrieved.
+    // Use CountDownLatch to signal when to start testing, for example, test only after results has been retrieved.
     private static CountDownLatch countDownLatch1;
-    private static CountDownLatch countDownLatch2;
     private static int noOfExecutions;
 
     private final RegistryFactory rf = RegistryFactory.getInstance();
@@ -81,8 +80,9 @@ public class OciMetricsSupportTest {
 
     @Test
     public void testMetricUpdate() throws InterruptedException {
+        Counter counter = baseMetricRegistry.counter("DummyCounter");
+
         countDownLatch1 = new CountDownLatch(1);
-        countDownLatch2 = new CountDownLatch(1);
         noOfExecutions = 0;
 
         doAnswer(invocationOnMock -> {
@@ -96,18 +96,15 @@ public class OciMetricsSupportTest {
             testMetricUpdateCounterValue[noOfExecutions == 1 ? 0 : 1] =
                     allMetricDataDetails.get(0).getDatapoints().get(0).getValue();
             if (noOfExecutions == 1) {
-                // Give signal that 1st metric update call has been completed
-                countDownLatch1.countDown();
+                counter.inc();
             } else {
-                // Give signal that 2nd metric update call has been completed
-                countDownLatch2.countDown();
+                // Give signal that multiple metric updates have been triggered
+                countDownLatch1.countDown();
             }
             return PostMetricDataResponse.builder()
                     .__httpStatusCode__(200)
                     .build();
         }).when(monitoringClient).postMetricData(any());
-
-        Counter counter = baseMetricRegistry.counter("DummyCounter");
 
         OciMetricsSupport.Builder ociMetricsSupportBuilder = OciMetricsSupport.builder()
                 .compartmentId("compartmentId")
@@ -116,7 +113,6 @@ public class OciMetricsSupportTest {
                 .initialDelay(1L)
                 .delay(2L)
                 .descriptionEnabled(false)
-
                 .monitoringClient(monitoringClient);
 
         Routing routing = createRouting(ociMetricsSupportBuilder);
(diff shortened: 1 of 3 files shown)


integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
```diff
@@ -76,14 +79,21 @@ public boolean matches(MetricID metricID, Metric metric) {
...
+    public void testMetricUpdate() throws InterruptedException {
+        countDownLatch1 = new CountDownLatch(1);
+        countDownLatch2 = new CountDownLatch(1);
         doAnswer(invocationOnMock -> {
+            if (countDownLatch1.getCount() != 0) {
+                // Wait for 1st signal before we can proceed
+                countDownLatch1.await(10, java.util.concurrent.TimeUnit.SECONDS);
+            } else {
+                // Give signal that 2nd call has been completed
+                countDownLatch2.countDown();
```
tjquinno, line 91: Should latch 2  be counted down _after_ `testMetricUpdateCounterValue` has been updated below? 

As written, If immediately after counting down latch 2 this thread yields to the test thread, then the test code can proceed and might apply the assertion to the not-yet-updated value of  `testMetricUpdateCounterValue` leading to a false failure.

integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
```diff
@@ -76,14 +79,21 @@ public boolean matches(MetricID metricID, Metric metric) {
...
     @Test
-    public void testMetricUpdate() {
-        // mock monitoringClient.postMetricData()
+    public void testMetricUpdate() throws InterruptedException {
+        countDownLatch1 = new CountDownLatch(1);
+        countDownLatch2 = new CountDownLatch(1);
         doAnswer(invocationOnMock -> {
+            if (countDownLatch1.getCount() != 0) {
+                // Wait for 1st signal before we can proceed
+                countDownLatch1.await(10, java.util.concurrent.TimeUnit.SECONDS);
```
tjquinno, line 88: We can simplify this to just
```java
countDownLatch1.await(...);
```
The `await` method will return immediately if the latch has already gone to 0; there is no need to check its value first.

integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
```diff
@@ -262,45 +270,24 @@ private void validateMetricCount(String[] scopes, int expectedMetricCount) {
...
-        int getTimeout() {
-            return this.timeOut;
-        }
     }
 
     private static void delay(long millis) {
         try {
             Thread.sleep(millis);
         } catch (InterruptedException ignore) {
-            Thread.currentThread().interrupt();
```
tjquinno, line 303: Should this method at least do _something_ if it is interrupted? Maybe invoke `fail` to fail whatever test is in progress?

tjquinno (changes requested):

integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
klustria: Good point, making the change

integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
klustria: Adding fail rather than ignoring.

integrations/oci/metrics/metrics/src/test/java/io/helidon/integrations/oci/metrics/OciMetricsSupportTest.java
klustria: The goal really is for countDownLatch2.countdown() to be executed only when **countDownLatch1.getCount() == 0** which is the else part and should happen after the second call of this mocked method. If do something like this
```
countDownLatch1.await(10, java.util.concurrent.TimeUnit.SECONDS);
if (countDownLatch1.getCount == 0) {
    countDownLatch2.countDown();
}
```
then  `countDownLatch2.countDown()` will be executed right after  countDownLatch1.await(..) proceeds in the same call rather than it be executed on the next call?

klustria: Modified to explicitly test 1st and 2nd metric update.

tjquinno:
I think there is still a possible race condition (although probably unlikely).

What if the mocking code in `doAnswer` invokes `countDown` on the first latch, but then, _before_ the main test code resumes the mocking code runs again? The counter will not have been incremented between the two runs of the mocking code. When the main test thread does resume the scheduled mocking code will have stored the value `1` into both slots in the array and the second assertion will fail. This is probably _very_ unlikely given the delay between runs of the mocking code, but it’s _conceivable_.

The real goal of the test is to make sure that the value stored in the counter is used correctly in preparing the OCI metrics data, specifically that the mocking code captures `1` and then `2` on successive executions. It does not matter which thread updates the counter's value as long as the counter is incremented after the mocking code captures the value the first time but before it does so the second time.

We can accomplish that goal by:
1. moving the second `counter.inc()` from the main test thread to inside the `then` part of the `if (noOfExecutions == 1)` in the mocking code. This makes absolutely sure that the second `counter.inc()` invocation occurs _after_ the mocking code has captured the metric data the first time but _before_ it captures it the second time. (`ScheduledExecutorService` will not allow the next periodic run to begin until after the preceding one completes.)
2. moving the `countDownLatch1.countDown()` invocation from the `then` part to the `else` part of that same `if`. Now the latch means that both data points have been captured and placed into the results array.
3. removing the second countdown latch.

After the main test thread starts the web server, it `await`s on the countdown latch as before. Once the latch allows that thread to pass (which _now_ happens only after the mocking code has run _twice_ and fully loaded the array of results) then the main test thread just does the two assertions on the two array elements which are now both set.

tjquinno (changes requested):
Sorry, more changes (but at least they simplify things). See the other longer comment.

klustria:
I agree with the feedback and hence implemented it. Tested it again for around 800 times and issue is not reproduced.

klustria:
The goal of this PR is to resolve the issue reported in https://github.com/oracle/helidon/issues/4813

tjquinno (approved):
LGTM

Issue #1749, opened 2020-05-07
Use Hamcrest assertions instead of JUnit


tomas-langer:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 2.0.0
* Whole code base
----------

## Problem Description
Currently we have a few tests that still use JUnit assertions, such as 
`assertEquals(actual, expected, "Test a and b are equal")`
all of these should be replaced with Hamcrest, such as:
`assertThat("Test a and b are equal", actual, is(expected))`)

There are a few assertions we can use:
`fail("")`
`assertAll()`
`assertThrows`

romain-grecourt:
Two items have been fixed in helidon-3.x, and will need to be backported in 2.x:
- https://github.com/oracle/helidon/pull/4886
- https://github.com/oracle/helidon/pull/4898

Captain1653:
@romain-grecourt  Can any contributor (not a member of helidon) do backport? I can try to do it. Are there any instructions for backport process in the documentation?

romain-grecourt:
You can do a backport. There is no special process at the moment (though we've been working on one recently).

A `git cherry-pick` should do the trick.

Captain1653:
@romain-grecourt  Should I backport all these PR into helidon-4.x ?

romain-grecourt:
> @romain-grecourt Should I backport all these PR into helidon-4.x ?

Yes, please !

PR #4898, opened 2022-09-15
Use Hamcrest assertions instead of JUnit in common/http (#1749)

Captain1653:
Part of #1749

I have already signed OCA. It's an automation problem.


Captain1653:
@romain-grecourt  Issue #1749 has label 2.x . My PR will be merged into 3.x . Is it ok?

romain-grecourt:
Yes as long the this PR does not auto close the issue.

PR #4908, opened 2022-09-17
Use Hamcrest assertions instead of JUnit in common/reactive (#1749)

Captain1653:
Part of #1749 

I'll do backport into helidon-2.x after merge.

I've already signed OCA. It's an automation problem.

I followed [these instructions](https://github.com/oracle/helidon/pull/1765#discussion_r423690868) 

- [x] [Backport into helidon-2.x](https://github.com/helidon-io/helidon/pull/4946)

### DIFF
--- a/common/reactive/src/test/java/io/helidon/common/reactive/SequentialSubscriberTest.java
+++ b/common/reactive/src/test/java/io/helidon/common/reactive/SequentialSubscriberTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


common/reactive/src/test/java/io/helidon/common/reactive/SequentialSubscriberTest.java
```diff
@@ -98,7 +99,7 @@ public void onComplete() {
         } catch (InterruptedException e) {
             executorService.shutdownNow();
         }
-        assertFalse(errorFound.isPresent(), () -> errorFound.get());
+        assertThat(errorFound.orElse(""), errorFound.isPresent(), is(false));
```
Captain1653, line 102: I don't like this approach with `Optional`, but if you need the error message...

tomas-langer: For Helidon 4 (`main` branch), we do have an optional matcher. For this version, we can keep this.

Alternative:
```java
assertThat(errorFound, is(Optional.empty()));
```

tomas-langer (approved):

PR #4932, opened 2022-09-21
4.x Update dependency check suppressions for false postives

barchetta:
Also update version of dependency check plugin

### DIFF
--- a/etc/dependency-check-suppression.xml
+++ b/etc/dependency-check-suppression.xml
@@ -291,5 +291,51 @@
    <cve>CVE-2022-2393</cve>
 </suppress>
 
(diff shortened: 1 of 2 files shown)


romain-grecourt (approved):

etc/dependency-check-suppression.xml
```diff
@@ -291,5 +291,51 @@
    <cve>CVE-2022-2393</cve>
 </suppress>
 
+<!-- False Positive. This is a CVE again Payara. This is generating a number of false positives.
```
tomas-langer, line 294: against not again (just a typo)

PR #4690, opened 2022-08-09
Fix intermittent failure - archetype build (FT timeout)

tvallin:
fix https://github.com/oracle/helidon/issues/4688

Signed-off-by: tvallin <thibault.vallin@oracle.com>

PR #4934, opened 2022-09-22
Fix FT intermittent failure - archetype build (backport)

tvallin:
backport fix #4690

### DIFF
--- a/archetypes/helidon/src/main/archetype/se/custom/files/src/test/java/__pkg__/FtServiceTest.java.mustache
+++ b/archetypes/helidon/src/main/archetype/se/custom/files/src/test/java/__pkg__/FtServiceTest.java.mustache
@@ -1,5 +1,6 @@
 package {{package}};
 
+import java.time.Duration;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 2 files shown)


archetypes/helidon/src/main/archetype/se/custom/files/src/test/java/__pkg__/FtServiceTest.java.mustache
```diff
@@ -172,9 +173,9 @@ class FtServiceTest {
         assertThat(response, is("Slept for 50 ms"));
 
         WebClientResponse clientResponse = client.get()
-                .path("/timeout/105")
+                .path("/timeout/10000")
```
tomas-langer, line 176: This is I think not correct - you have increased both the time the service sleeps and the timeout. I think the sleep time should stay the same (105), the timeout  should be updated to 11 seconds. This way the difference is still one second as it used to be

tomas-langer (changes requested):

archetypes/helidon/src/main/archetype/se/custom/files/src/test/java/__pkg__/FtServiceTest.java.mustache
tvallin: I am not sure to understand, I set the timeout to 5 seconds in`FtService`, so the difference between the sleep and timeout is actually 5 seconds, right ?
I thought after 5 seconds of sleep, timeout is triggered and response sent, so having client waiting for 11 seconds was reasonable.

tomas-langer: OK, I see it now. Thanks.

tomas-langer (approved):

romain-grecourt (approved):

Issue #4691, opened 2022-08-09
ConfigProperties - Cannot convert "axe,sword,wand" to type java.util.List


danielkec:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 3.0.0
* Helidon MP
* JDK version: 17.0.2+8-86
* OS: Ubuntu

----------

## Problem Description
MP Configuration `@ConfigProperties` doesn't properly convert to List, array works as expected.

## Steps to reproduce
```java
@ApplicationScoped
@ConfigProperties(prefix="sorcerer")
public class SorcererProperties {
    public @ConfigProperty(name="orcSlayingPotions") int potions;
    public boolean cloak;
    public List<String> weapons;
}
```
```
java.lang.IllegalArgumentException: Failed to convert property "sorcerer.weapons" from its value "axe,sword,wand" to java.util.List
        at io.helidon.config.mp.MpConfigImpl.convert(MpConfigImpl.java:312)
        at io.helidon.config.mp.MpConfigImpl.lambda$optionalValue$7(MpConfigImpl.java:193)
        at java.base/java.util.Optional.map(Optional.java:260)
        at io.helidon.config.mp.MpConfigImpl.optionalValue(MpConfigImpl.java:193)
        at io.helidon.config.mp.MpConfigImpl.getOptionalValue(MpConfigImpl.java:131)
        at io.helidon.config.mp.MpConfigProviderResolver$ConfigDelegate.getOptionalValue(MpConfigProviderResolver.java:356)
        at io.helidon.microprofile.config.FieldSetter.set(FieldSetter.java:61)
        at io.helidon.microprofile.config.ConfigBeanDescriptor.produce(ConfigBeanDescriptor.java:144)
        at io.helidon.microprofile.config.ConfigBeanDescriptor.produce(ConfigBeanDescriptor.java:132)
        at io.helidon.microprofile.config.ConfigCdiExtension.lambda$registerConfigProducer$5(ConfigCdiExtension.java:211)
        at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$CreateCallback.create(BeanConfiguratorImpl.java:368)
        at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$CreateCallback.access$700(BeanConfiguratorImpl.java:338)
[...]

PR #4959, opened 2022-09-26
fix for issue in handling ConfigProperties using List

trentjeff:
fix for #4691 .

### DIFF
--- a/microprofile/config/src/main/java/io/helidon/microprofile/config/FieldSetter.java
+++ b/microprofile/config/src/main/java/io/helidon/microprofile/config/FieldSetter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2021, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package io.helidon.microprofile.config;
 
 import java.lang.reflect.Field;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -35,18 +36,19 @@ class FieldSetter {
     private final Class<?> fieldType;
     private final String defaultValue;
     private final boolean isOptional;
+    private final boolean isList;
 
     FieldSetter(Field field, String configKey, String defaultValue) {
         this.field = field;
         this.configKey = configKey;
         this.defaultValue = defaultValue;
         FieldTypes fieldTypes = FieldTypes.create(field.getGenericType());
-        if (fieldTypes.field0().rawType().equals(Optional.class)) {
+        this.isOptional = fieldTypes.field0().rawType().equals(Optional.class);
+        this.isList = fieldTypes.field0().rawType().equals(List.class);
+        if (isOptional || isList) {
             fieldType = fieldTypes.field1().rawType();
-            isOptional = true;
         } else {
             fieldType = field.getType();
-            isOptional = false;
         }
         field.setAccessible(true);
     }
@@ -58,18 +60,37 @@ class FieldSetter {
                 setValue(prefixedKey, field, instance, valueWithDefault(config, prefixedKey));
             } else {
                 Object currentValue = field.get(instance);
-                Optional<?> value = config.getOptionalValue(prefixedKey, fieldType);
-                if (currentValue == null) {
-                    // no current value, need configuration
-                    if (value.isPresent()) {
-                        setValue(prefixedKey, field, instance, value.get());
+
+                if (isList) {
+                    Optional<? extends List<?>> value = config.getOptionalValues(prefixedKey, fieldType);
+                    if (currentValue == null) {
+                        // no current value, need configuration
+                        if (value.isPresent()) {
+                            setValue(prefixedKey, field, instance, value.get());
+                        } else {
+                            setValue(prefixedKey, field, instance, null);
+                        }
                     } else {
-                        setValue(prefixedKey, field, instance, null);
+                        // there is existing value, only override if value in config
+                        if (value.isPresent()) {
+                            setValue(prefixedKey, field, instance, value.get());
+                        }
                     }
+
                 } else {
-                    // there is existing value, only override if value in config
-                    if (value.isPresent()) {
-                        setValue(prefixedKey, field, instance, value.get());
+                    Optional<?> value = config.getOptionalValue(prefixedKey, fieldType);
+                    if (currentValue == null) {
+                        // no current value, need configuration
+                        if (value.isPresent()) {
+                            setValue(prefixedKey, field, instance, value.get());
+                        } else {
+                            setValue(prefixedKey, field, instance, null);
+                        }
+                    } else {
+                        // there is existing value, only override if value in config
+                        if (value.isPresent()) {
+                            setValue(prefixedKey, field, instance, value.get());
+                        }
                     }
                 }
             }
--- a/microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
+++ b/microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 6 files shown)


microprofile/config/src/main/java/io/helidon/microprofile/config/FieldSetter.java
```diff
@@ -44,9 +47,15 @@ class FieldSetter {
         if (fieldTypes.field0().rawType().equals(Optional.class)) {
             fieldType = fieldTypes.field1().rawType();
             isOptional = true;
+            isList = false;
+        } else if (Objects.nonNull(fieldTypes.field1()) && List.class.isAssignableFrom(fieldTypes.field0().rawType())) {
```
tomas-langer, line 51: 1. must be compared to `List` (same as `Optional` above). If implemented this ways, `LinkedList` would be OK, and that would fail with class cast exception
2. the `field1()` is used above without null check. If the type is null, this should fail (maybe not with NPE, but we do not support untyped `List` or `Optional`)

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Lower level testing for {@link io.helidon.microprofile.config.ConfigBeanDescriptor}.
+ */
+public class ConfigBeanDescriptorTest {
```
tomas-langer, line 44: We use package local test classes and test methods, not `public`

tomas-langer: I am not sure what is this test testing.

tomas-langer (changes requested):

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
trentjeff: // the target of the test starts here:

        Object instance = descriptor.produce(strListInjectionPoint, config);
        assertNotNull(instance);
        assertThat(((TestConfiguredBean) instance).strList, Matchers.is(List.of("a", "b", "c")));

        InjectionPoint intListInjectionPoint = mock(InjectionPoint.class);
        when(intListInjectionPoint.getType()).thenReturn(intListType);
        instance = descriptor.produce(intListInjectionPoint, config);
        assertNotNull(instance);
        assertThat(((TestConfiguredBean) instance).intList, Matchers.is(List.of(1, 2, 3)));

trentjeff: Will push the change to use package privates soon.

microprofile/config/src/main/java/io/helidon/microprofile/config/FieldSetter.java
trentjeff: The Optional code was there already, I only indented it.  No other logic changed with that section.

The List check is using Class.isAssignable() and there is a null check on like 51.

tomas-langer: I was saying you should implement it the same way as the `Optional` is implemented.

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        ConfigBeanDescriptor descriptor = ConfigBeanDescriptor.create(annotatedType, configProperties);
+        assertNotNull(descriptor);
+
+        InjectionPoint strListInjectionPoint = mock(InjectionPoint.class);
+        when(strListInjectionPoint.getType()).thenReturn(strListType);
+
+        // the target of the test is here:
+        Object instance = descriptor.produce(strListInjectionPoint, config);
+        assertNotNull(instance);
```
tomas-langer, line 87: Please use Hamcrest (always) for assertions
Do static imports for Hamcrest methods.
Add assertion for instanceOf if it may produce a different type.

```
assertThat(o, instanceOf(TestConfiguredBean.class));
assertThat(o, notNullValue());
```

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,106 @@
...
+        assertThat(((TestConfiguredBean) instance).strList, Matchers.is(List.of("a", "b", "c")));
+
+        InjectionPoint intListInjectionPoint = mock(InjectionPoint.class);
+        when(intListInjectionPoint.getType()).thenReturn(intListType);
+        instance = descriptor.produce(intListInjectionPoint, config);
+        assertNotNull(instance);
+        assertThat(((TestConfiguredBean) instance).intList, Matchers.is(List.of(1, 2, 3)));
```
tomas-langer, line 94: assertThat(intList, contains(1, 2, 3));

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        // the target of the test is here:
+        Object instance = descriptor.produce(strListInjectionPoint, config);
+        assertNotNull(instance);
+        assertThat(((TestConfiguredBean) instance).strList, Matchers.is(List.of("a", "b", "c")));
+
+        InjectionPoint intListInjectionPoint = mock(InjectionPoint.class);
+        when(intListInjectionPoint.getType()).thenReturn(intListType);
+        instance = descriptor.produce(intListInjectionPoint, config);
+        assertNotNull(instance);
```
tomas-langer, line 93: use hamcrest

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,106 @@
...
+        ConfigBeanDescriptor descriptor = ConfigBeanDescriptor.create(annotatedType, configProperties);
+        assertNotNull(descriptor);
+
+        InjectionPoint strListInjectionPoint = mock(InjectionPoint.class);
+        when(strListInjectionPoint.getType()).thenReturn(strListType);
+
+        // the target of the test is here:
+        Object instance = descriptor.produce(strListInjectionPoint, config);
+        assertNotNull(instance);
+        assertThat(((TestConfiguredBean) instance).strList, Matchers.is(List.of("a", "b", "c")));
```
tomas-langer, line 88: assertThat(strList, contains("a", "b", "c"));

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,106 @@
...
+import jakarta.enterprise.inject.spi.InjectionPoint;
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.inject.ConfigProperties;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
+import org.hamcrest.Matchers;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
```
tomas-langer, line 37: Do not use this assertion, use Hamcrest `import static org.hamcrest.CoreMatchers.notNullValue;`

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
```diff
@@ -0,0 +1,106 @@
...
+        configProvider.registerConfig(config, Thread.currentThread().getContextClassLoader());
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    void testProduceListFields() throws Exception {
+        AnnotatedType<?> annotatedType = mock(AnnotatedType.class);
+        when(annotatedType.getJavaClass()).thenReturn((Class) TestConfiguredBean.class);
+
+        Field strListField = Objects.requireNonNull(TestConfiguredBean.class.getField("strList"));
```
tomas-langer, line 71: `Objects.requireNonNull` is not the right choice here. If you want to check for null, use an assertion.

tomas-langer (changes requested):

microprofile/config/src/main/java/io/helidon/microprofile/config/FieldSetter.java
trentjeff: pushed

microprofile/config/src/test/java/io/helidon/microprofile/config/ConfigBeanDescriptorTest.java
trentjeff: pushed

tomas-langer (dismissed):

PR #4964, opened 2022-09-26
Fix identification of parallel startup of CDI

### LINKED ISSUE #4963: [4.x] Make sure that container is shut down when tests fail
Backport of #4940 for Helidon 4.x

tomas-langer:
that build time initialization has failed.

Fixes #4963


microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/HelidonContainerImpl.java
```diff
@@ -209,14 +209,23 @@ public Collection<URL> getResources(String name) {
...
-        cdi = new HelidonCdi(id, bootstrap, deployment);
-        HelidonCdiProvider.setCdi(cdi);
+            cdi = new HelidonCdi(id, bootstrap, deployment);
+            HelidonCdiProvider.setCdi(cdi);
 
-        beanManager.getEvent().select(BuildTimeEnd.Literal.INSTANCE).fire(id);
+            beanManager.getEvent().select(BuildTimeEnd.Literal.INSTANCE).fire(id);
+        } catch (Throwable e) {
+            try {
+                beanManager.getEvent().select(InitializationFailed.Literal.INSTANCE).fire(e);
```
ljnelson, line 223: What operations are permitted at this state transition?

tomas-langer: This was added for cleanup operations - to allow the extension to cleanup static resources related to startup (e.g. designed specifically to our server cdi extension which makes sure you do not attempt to start server twice)

ljnelson: I guess my point is: extensions themselves don't really have a well-defined lifecycle (they're created, then they get certain container lifecycle events and then they just stick around and [behave like any other CDI bean](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#type_bean_discovery_full:~:text=For%20each%20service%20provider%2C%20the%20container%20must%20provide%20a%20bean%20of%20scope%20%40ApplicationScoped%20and%20qualifier%20%40Default%2C%20supporting%20injection%20of%20a%20reference%20to%20the%20service%20provider%20instance)).

So normally, a portable extension wouldn't be doing anything that needs to be "undone", because they basically just set up raw materials for use by the container, which is really where lifecycles "start".  For example, normally a regular bean/observer of, say, `@Initialized(ApplicationScoped.class)` would "do" startup-related things, and then because of the pre-existing well-defined CDI lifecycles would be notified according to fairly strict rules when those things need to be "undone".  (Some of those rules include, for example, that [contexts must have already been destroyed](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0.html#before_shutdown), what state the `BeanManager` is in, what you can and cannot expect to work—stuff like that.  Rules.)

If, despite all this, we decide to introduce another kind of lifecycle here that looks like a version of the existing lifecycles, what are its rules?  That's all I'm asking.

My view would be to look and see if there's any way to avoid starting a webserver (or whatever), for example, during portable extension initialization, since that is a task normally performed by a bean.  If that can be located in the proper spot, then the existing CDI lifecycles and their well-defined rules just take over, notify all the players already, and we don't have to invent our own.

tomas-langer: We are not starting a webserver. There is a check that we only have a single instance of the extension within a VM (preventing two containers to be started within the same VM practically). 
I am using method `stopServer` just because it also clears the "I am attempting to do something with webserver setup in CDI" flag.

I will try to check if we can do this later in lifecycle (such as when application scope is initialized), or we need it as early as it is implemented right now

tomas-langer: I have refactored the approach to use CDI information we already have in place. The parallel startup tests are passing.
As this is using state management we already have in place, it should be a better fix than the custom event.

ljnelson (approved):
Much simpler.

PR #5013, opened 2022-09-29
[3.x] - Tracing header case

### LINKED ISSUE #4956: B3 headers are not considered when there are case-miss matches
Are B3 header names case-sensitive? B3 propagation is not happening when cases mismatch. Incoming B3 headers are not considered for tracing when cases miss matches.

## Environment Details
* Helidon Version: 3.0.1
* Helidon SE or Helidon MP: Helidon SE
* JDK version: Java 17
* OS: Windows 11

----------

## Problem Description
Incoming B3 headers are not considered for tracing when cases miss matches. If B3 headers exactly match those below, then it's working; otherwise, it's not.
Working:
`X-B3-TraceId, X-B3-SpanId, X-B3-ParentSpanId, X-B3-Sampled`
Not Working:
`x-b3-traceid, x-b3-spanid, x-b3-parentspanid, x-b3-sampled`
`X-B3-TraceID, X-B3-SpANId, X-B3-PARENTSPANID, x-b3-SAMPLED`
and so on...

## Steps to reproduce
Create a Helidon SE 3.0.1 application and configure Jaeger tracing with B3 propagation.
```
tracing:
  service: helidon-quick-start-1
  protocol: http
  port: 14250
  path: /api/traces
  tags:
    env: development
  enabled: true
  sampler-type: "const"
  sampler-param: 1
  log-spans: true
  propagation: b3
```
Add B3 headers to the request, all in small case.
image
Incoming B3 header trace ID should be used, and the same should be propagated to downstream applications, but a new trace ID is generated and propagated to the downstream application.
image
image

Headers need to be case insensitive.

tjquinno:
Resolves #4956 

In two places related to `HeaderProvider`, instead of storing the input headers `Map` and returning a copy of it from `headers()` the code now builds a new `Map` using the lower-cased keys (and unchanged values) from the input `Map`.

The changes are in `HeaderProvider.create` and `WebTracingConfig.TracingHeaderProvider`.

Two questions for reviewers:

1. There are other implementations of `HeaderProvider` listed below. It is not clear (to me) whether they should also change. I have left them alone so far so as to perturb as little code as possible. Do they need to change also to address this issue?
2. I looked at creating an integration test that would create a server with two endpoints:
   * `/hello`
   * `/spans`
      * `POST` would act as a tracing server, accepting tracing data and storing it internally
  
The test would ping `/hello` with preset B3 headers (but with varying case). That `GET` would trigger a tracing `POST` to `/spans`.
Then (after the tracing data had been collected as indicated by a countdown latch) the test would check the gathered tracing data to make sure the headers had propagated correctly despite the intentionally odd case of the header keys.

This could get complicated because the `POST` requests to `/spans` could themselves trigger traces, there could be threading issues with the server sending trace updates to itself, etc. which could mean we'd need two separate servers running.

I can pursue this if the unit tests included here do not seem sufficient.

----

helidon-grpc-client
  io.helidon.grpc.client
    ClientTracingInterceptor.ClientTracingListener
        private class MetadataHeaderConsumer implements HeaderConsumer {

helidon-grpc-server
  io.helidon.grpc.server
    GrpcTracing
      private static class MapHeaderProvider implements HeaderProvider {

helidon-tracing
  io.helidon.tracing
    HeaderConsumer.java
    MapHeaderConsumer.java

helidon-webclient-tracing
  io.helidon.webclient.tracing
    WebClientTracing
      private static class ClientHeaderConsumer implements HeaderConsumer {

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
@@ -15,8 +15,10 @@
  */
 package io.helidon.webserver;
 
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -339,12 +341,13 @@ public abstract class WebTracingConfig {
                     });
         }
 
-        private static class TracingHeaderProvider implements HeaderProvider {
+        static class TracingHeaderProvider implements HeaderProvider {
 
-            private final Map<String, List<String>> headers;
+            private final Map<String, List<String>> headers = new HashMap<>();
 
             TracingHeaderProvider(Map<String, List<String>> headers) {
-                this.headers = headers;
+                // Normalize to lower-case keys, particularly so B3 headers match correctly.
+                headers.forEach((key, values) -> this.headers.put(key.toLowerCase(Locale.ROOT), values));
             }
 
             @Override
--- a/tracing/tracing/src/main/java/io/helidon/tracing/HeaderProvider.java
+++ b/tracing/tracing/src/main/java/io/helidon/tracing/HeaderProvider.java
@@ -15,7 +15,10 @@
  */
 package io.helidon.tracing;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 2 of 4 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/WebTracingConfig.java
```diff
@@ -339,12 +341,13 @@ private void doAccept(ServerRequest req, ServerResponse res) {
...
-        private static class TracingHeaderProvider implements HeaderProvider {
+        static class TracingHeaderProvider implements HeaderProvider {
 
-            private final Map<String, List<String>> headers;
+            private final Map<String, List<String>> headers = new HashMap<>();
 
             TracingHeaderProvider(Map<String, List<String>> headers) {
-                this.headers = headers;
+                // Normalize to lower-case keys, particularly so B3 headers match correctly.
+                headers.forEach((key, values) -> this.headers.put(key.toLowerCase(Locale.ROOT), values));
```
tomas-langer, line 350: We should not normalize on any case - we must use what we get, but treat it correctly as case insensitive

tomas-langer (changes requested):

tjquinno:
Now obsolete due to more general header and parameter fixes.

PR #5034, opened 2022-09-30
Sec provider 4810

### LINKED ISSUE #4810: Documentation bug in Security Providers
## Environment Details
* Helidon Version: 2.5.x
* JDK version: 17
* OS: OS X
----------

## Problem Description
The MP security provider documentation uses clear text passwords in configuration. This however fails if I copy and paste the configuration into the quickstart project it fails with an obscure error ".. is a clear text password, yet encryption is required".

The example requires config.require-encryption=false, which is an undocumented configuration in Helido.io/docs, so the security providers example needs to mention this fact upfront to avoid having to google around for the failure text. 

## Steps to reproduce
[//]: # Follow along the tutorial: https://helidon.io/docs/v2/#/mp/security/02_providers

Geetha-Savithriamma:
Fixes #4810.

### DIFF
--- a/docs/mp/security/02_providers.adoc
+++ b/docs/mp/security/02_providers.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2018, 2020 Oracle and/or its affiliates.
+    Copyright (c) 2018, 2022 Oracle and/or its affiliates.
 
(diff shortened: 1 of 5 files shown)


ljamen (changes requested):
Builds are failing due to copyright year.

romain-grecourt (approved):

Issue #4984, opened 2022-09-28
2.x: Intermittent HelloWorldAsyncResponseTest.testAsyncWithArg


barchetta:
```
java.lang.AssertionError: 
Synthetic SimpleTimer count
Expected: is <3L>
     but: was <2L>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at io.helidon.microprofile.metrics.HelloWorldAsyncResponseTest.testAsyncWithArg(HelloWorldAsyncResponseTest.java:142)
	at io.helidon.microprofile.metrics.HelloWorldAsyncResponseTest$Proxy$_$$_WeldClientProxy.testAsyncWithArg(Unknown Source)
```

```
Standard Output

2022.09.28 00:01:39 INFO org.jboss.weld.Bootstrap Thread[main,5,main]: WELD-ENV-000014: Falling back to Java Reflection for bean-discovery-mode="annotated" discovery. Add org.jboss:jandex to the classpath to speed-up startup.
2022.09.28 00:01:39 INFO org.jboss.weld.Bootstrap Thread[main,5,main]: WELD-000101: Transactional services not available. Injection of @Inject UserTransaction not available. Transactional observers will be invoked synchronously.
2022.09.28 00:01:39 INFO org.jboss.weld.Event Thread[main,5,main]: WELD-000411: Observer method [BackedAnnotatedMethod] public org.glassfish.jersey.ext.cdi1x.internal.ProcessAllAnnotatedTypes.processAnnotatedType(@Observes ProcessAnnotatedType<?>, BeanManager) receives events for all annotated types. Consider restricting events using @WithAnnotations or a generic type with bounds.
2022.09.28 00:01:39 WARNING io.helidon.microprofile.metrics.MetricUtil Thread[main,5,main]: Attribute 'absolute=true' in metric annotation ignored at class level
2022.09.28 00:01:39 WARNING io.helidon.microprofile.metrics.MetricUtil Thread[main,5,main]: Attribute 'absolute=true' in metric annotation ignored at class level
2022.09.28 00:01:39 INFO io.helidon.microprofile.server.ServerCdiExtension Thread[main,5,main]: Registering JAX-RS Application: HelloWorldApp
2022.09.28 00:01:39 INFO io.helidon.webserver.NettyWebServer Thread[nioEventLoopGroup-4-1,10,main]: Channel '@default' started: [id: 0x922626cc, L:/0.0.0.0:41493]
[...]

PR #5047, opened 2022-10-03
[3.x] Add robustness to some of the timing-susceptible metrics tests; add util matcher with retry

### LINKED ISSUE #5029: [3.x] Intermittent HelloWorldAsyncResponseTest.testAsyncWithArg
Backport of #4984 for Helidon 3.x

tjquinno:
Resolves #5029 

(back port of #4984)

See overview in #5032 

### DIFF
--- a/config/testing/src/main/java/io/helidon/config/testing/MatcherWithRetry.java
+++ b/config/testing/src/main/java/io/helidon/config/testing/MatcherWithRetry.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


config/testing/src/main/java/io/helidon/config/testing/MatcherWithRetry.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public static void retry(Callable<Boolean> work) throws Exception {
+        for (int i = 0; i < RETRY_COUNT; i++) {
+            if (work.call()) {
+                return;
+            }
+            Thread.sleep(RETRY_DELAY_MS);
+        }
+    }
+}
```
danielkec, line 82: @tjquinno I don't think this actually retries anything since only usage always returns true and this thing doesn't catch exceptions. We do have really fancy [fault tolerance Retry](https://helidon.io/docs/v3/#/se/fault-tolerance) which works really well.

tjquinno: You're right. I'd intended for the `retry` method to repeat if the work returned `false` (and you are correct, so far the only use returns true) _or_ if the work threw an exception (such as the `asserts` in the lambda passed to `retry` in its only use).

PR #5091, opened 2022-10-05
3.x: using jdk JDK 17.0.3 or newer in Dockerfile.jlink

joedayz:
In JDK 17.0.2 and earlier there was a bug where the JVM would crash when generating the CDS archive (see [helidon-io/helidon-build-tools#537](https://github.com/helidon-io/helidon-build-tools/issues/537)). So you need to use JDK 17.0.3 or newer. The Dockerfile-jlink in Helidon 3.0.1 uses maven:3.8.4-openjdk-17-slim as the base image. That uses JDK 17.0.2, so you get the failure.

### DIFF
--- a/archetypes/helidon/src/main/archetype/common/files/Dockerfile.jlink.mustache
+++ b/archetypes/helidon/src/main/archetype/common/files/Dockerfile.jlink.mustache
@@ -1,6 +1,6 @@
 
 # 1st stage, build the app
-FROM maven:3.8.4-openjdk-17-slim as build
+FROM maven:3.8.6-eclipse-temurin-17 as build
 
 WORKDIR /helidon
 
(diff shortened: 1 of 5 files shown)


barchetta:
Thanks for your contribution! Have you signed the OCA? https://oca.opensource.oracle.com/

joedayz:
Screen Shot 2022-10-05 at 19 08 46
Done.

barchetta:
@joedayz we are working on a plan to move off of openjdk docker images. Until that plan is determined I'd like to work-around this issue by changing the line that builds the jlink image to:

```
# Don't generate CDS archive to work around JVM bug https://bugs.openjdk.org/browse/JDK-8274944
RUN mvn -Ddocker.build=true package -Pjlink-image -DskipTests -Djlink.image.addClassDataSharingArchive=false
```

This adds `-Djlink.image.addClassDataSharingArchive=false` to prevent the generation of the CDS archive.

barchetta:
We appreciate the contribution, but we are working on other plans to address the openjdk issue.

PR #5103, opened 2022-10-05
FT retry checkMetricsForExecutorService

danielkec:

### DIFF
--- a/metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
+++ b/metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
@@ -17,9 +17,6 @@ package io.helidon.metrics;
 
 import java.time.Duration;
 import java.util.Optional;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -39,7 +36,6 @@ import org.eclipse.microprofile.metrics.ConcurrentGauge;
 import org.eclipse.microprofile.metrics.MetricRegistry;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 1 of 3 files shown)


metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
```diff
@@ -204,25 +212,16 @@ void checkMetricsForExecutorService() throws Exception {
...
+        retry.invoke(() -> metricsRequestBuilder
+                        .submit()
+                        .peek(res -> assertThat("Second access to metrics", res.status().code(), is(200)))
+                        .flatMapSingle(res -> res.content().as(JsonObject.class))
+                        .peek(json -> assertThat("JSON metrics results after accessing slow endpoint",
+                                json, hasKey(jsonKeyForCompleteTaskCountInThreadPool)))
+                        .map(json -> json.getInt(jsonKeyForCompleteTaskCountInThreadPool))
+                        .forSingle(secCompletedTaskCnt ->
+                                assertThat("Completed task count after accessing slow endpoint", secCompletedTaskCnt, is(1))))
+                .await(CLIENT_TIMEOUT);
```
tjquinno, line 224: It is appealing to reuse the existing FT impl vs. a retry loop.

The JavaDoc for `peek` does say it's intended mainly for debugging and warns that the stream implementation might optimize away calls to `peek` for some elements of the stream. That probably cannot happen here (there are no `findFirst` or other optimizable steps in handling this stream).

metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
```diff
@@ -64,7 +65,7 @@ public class TestServer {
     private static MetricsSupport metricsSupport;
 
     private static final Duration CLIENT_TIMEOUT = Duration.ofSeconds(5);
```
tjquinno, line 67: I had included property look-up of `RETRY_ COUNT` and `RETRY_DELAY_MS` so we would have an easy way to adjust for slow pipelines or action runners.

I know `CLIENT_TIMEOUT` was not originally set up that way, but do you think it is worth doing a similar property look-up for it for the same reason as part of this PR?

tjquinno (dismissed):
Looks good. See comments for a suggestion you might want to include.

metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
danielkec: This is our reactive peek, we don't omit side-effects like Java streams does

metrics/metrics/src/test/java/io/helidon/metrics/TestServer.java
danielkec: Right

tjquinno (approved):
LGTM

PR #5141, opened 2022-10-10
Introduce Helidon Pico

trentjeff:
This PR introduces Pico, specifically the API and Types modules.


.gitignore
```diff
@@ -23,6 +23,7 @@
 hs_err_pid*
 # Java Flight Recorder
 *.jfr
+javac.*.args
```
tomas-langer, line 26: This file should be generated in target directory, which is already excluded, why add it again?

pico/.gitignore
```diff
@@ -0,0 +1,16 @@
+README.md
```
tomas-langer, line 1: This file does not belong here. There is only one git ignore for the repo, in the root

pico/api/src/main/java/module-info.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Pico API module.
```
tomas-langer, line 18: This should explain what Pico is and where to start

pico/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+
+    <packaging>pom</packaging>
+
+    <properties>
+        <!-- all pico modules should be backward compatible with 11 -->
+        <version.java>11</version.java>
+
+        <!-- pico supports legacy javax.inject and javax.annotations in tooling -->
+        <javax.injection.version>1</javax.injection.version>
+<!--        <javax.annotations.version>1.3.2</javax.annotations.version>-->
```
tomas-langer, line 44: No commented out lines should be in our production code

pico/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+                    <configuration>
+                        <!-- This module is Java 11, we must remove enable-preview from parent project -->
+                        <additionalOptions combine.self="override"/>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
+
+    <developers>
```
tomas-langer, line 83: This only belongs to root pom file of Helidon, we do not have any other developer attributions (git is the source of truth)

pico/types/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+    <parent>
+        <groupId>io.helidon.pico</groupId>
+        <artifactId>helidon-pico-project</artifactId>
+        <version>4.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>helidon-pico-types</artifactId>
+    <name>Helidon Pico (runtime) Types</name>
```
tomas-langer, line 31: Please use the convention of other modules - should be artifact id in english (in this case "Helidon Pico Types".
Add additional details to description if you want to

tomas-langer (changes requested):
This seems not to contain any code. This should not work with our pipelines (Javadoc should fail on this). Also hard to approve module names when we do not know what is the content.

.gitignore
trentjeff: I added it so that I can hide the pico modules locally that are not yet ready to merge down.

pico/.gitignore
trentjeff: As I mentioned previously, I want to hide files not ready to merge down.

pico/api/src/main/java/io/helidon/pico/api/ExternalContracts.java
```diff
@@ -0,0 +1,49 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Placed on the implementation of a service as an alternative to using a {@link Contract}.
+ *
+ * Use this annotation when it is impossible to place an annotation on the interface itself - for instance of the interface comes
+ * from a 3rd party library/provider.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
```
tomas-langer, line 31: We should not need runtime retention, as we plan to do everything at compile time (this is valid for all the annotations).

pico/api/src/main/java/module-info.java
trentjeff: The readme will explain that - which will be in a future pr.

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
```diff
@@ -0,0 +1,74 @@
...
+     * @return the annotation type name
+     */
+    TypeName getTypeName();
+
+    /**
+     * The value.
+     *
+     * @return The string value of (all attributes of) the annotation (expressed as toString()), or null if no value is present
+     */
+    String getValue();
```
tomas-langer, line 39: Naming of methods is not according to Helidon code style. We use "record" style (valid for all methods)

tomas-langer (changes requested):

.gitignore
tomas-langer: I was talking about `javac.*.args` - that exclusion should not be needed. Which files match it that you need to add it to exclusion?

pico/.gitignore
tomas-langer: Please remove this file. If you "un-add" your files from git, they will not be added to your commits and pushes.

pico/api/src/main/java/module-info.java
tomas-langer: `module-info.java` and `package-info.java` are the only things to be seen when you access Javadoc. So it is good to add some basic documentation to them (at least where to start).

pico/api/src/main/java/io/helidon/pico/api/ExternalContracts.java
trentjeff: I've already advised the developer to not include this module into compile scope; my thinking is to leave it here for anyone that still wants to use reflection, etc.

pico/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+    <name>Helidon Pico Project</name>
+
+    <packaging>pom</packaging>
+
+    <properties>
+        <!-- all pico modules should be backward compatible with 11 -->
+        <version.java>11</version.java>
+
+        <!-- pico supports legacy javax.inject and javax.annotations in tooling -->
+        <javax.injection.version>1</javax.injection.version>
```
tomas-langer, line 43: not used

pico/types/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>helidon-pico-types</artifactId>
+    <name>Helidon Pico Types</name>
+    <description>Provides the runtime types (a subset of SPI) useful for Pico</description>
+
+    <dependencies>
+        <dependency>
```
tomas-langer, line 35: these two dependencies are not used

tomas-langer (changes requested):

pico/api/src/main/java/io/helidon/pico/api/ExternalContracts.java
tomas-langer: I think we should not do that - it will make the class file bigger and we want to process it compile time always.

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
```diff
@@ -0,0 +1,74 @@
...
+    default boolean hasValue() {
+        return hasValue(getValue());
+    }
+
+    /**
+     * Helper method to determine if the value is present (i.e., non-null and non-blank).
+     *
+     * @return true if the value provided is non-null and non-blank.
+     */
+    static boolean hasValue(String val) {
```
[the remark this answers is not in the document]
trentjeff: For optional string attributes on annotations, most people as a convention will use 'default ""'. That does not technically count as having a value.  There are also many places in code coming that will need the "not null and not blank" API like this.  We can think about renaming it, but taking it out is not an option.

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    Map<String, String> getValues();
+
+    /**
+     * Determines whether the value is set to a non-null and non-blank value.
+     *
+     * @return true if the value provided is non-null and non-blank.
+     */
+    default boolean hasValue() {
+        return hasValue(getValue());
```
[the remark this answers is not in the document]
trentjeff: will make the javadoc clearer

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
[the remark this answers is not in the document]
tomas-langer: If you look at our own annotations, we do use empty strings a lot of time (when feasible), but sometimes we had to implement a custom default value, to allow usage of empty string as a valid value.
So yes, in a lot of cases empty string means not having a value. But is some cases it does not.

pico/types/src/main/java/io/helidon/pico/types/TypeName.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public interface TypeName extends Comparable<TypeName> {
+
+    /**
+     * Functions the same as {@link Class#getPackageName()}.
+     *
+     * @return the package name
+     */
+    @JsonIgnore
+    String getPackageName();
```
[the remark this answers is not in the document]
tomas-langer: Such an annotation would only make sense if you decide to send `TypeName` through Jackson. As that is not the case, the annotations are useless for target teams. 
This type should only be used by Helidon team (or team extending Pico).

.gitignore
trentjeff: took that out and pushed.

pico/.gitignore
trentjeff: removed and pushed.

pico/api/src/main/java/module-info.java
trentjeff: pushed updates to package-info; please review again.

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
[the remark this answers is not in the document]
trentjeff: I fixed the javadoc.

pico/types/src/main/java/io/helidon/pico/types/TypeName.java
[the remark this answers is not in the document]
ljnelson: (Neutral on all this; just here to note the existence of [Jackson mixins](https://medium.com/@shankar.ganesh.1234/jackson-mixin-a-simple-guide-to-a-powerful-feature-d984341dc9e2), which is the normal way someone can "put" a Jackson annotation on a class that does not or cannot have Jackson annotations on it.)

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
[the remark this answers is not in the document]
trentjeff: First, I need a method that will do this check in its implementation somewhere in the API. If not here then I'll need to pick a new spot for that impl. This avoids having to write "val != null && !val.isBlank()" in many, many places that will soon be coming.  I can conceptually use "empty" instead of "blank" with no adverse effects, but I thought blank might be better to use in this case. This is point 1.

Second, are you suggesting we need to distinguish *at compile time* a need to differentiate the default value of "" vs intended/ascribed value of "" by the developer? If so then that simply doesn't exist, nor do I believe have any need for it since the processing will behave the same regardless of how it was originated using defaults or via an override.  If it's blank (or empty) it doesn't have a value - plain and simple.

So the realistic options I see are:
0) Use javadoc to clearly describe how "" is considered to not have any value from our opinionated PoV on this class at leasy.
1) use "empty" instead of "blank" for hasValue()
2) use a better description: hasValue() -> hasNonBlankValue() and/or hasNonEmptyValue()

pico/types/src/main/java/io/helidon/pico/types/TypeName.java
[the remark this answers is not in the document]
trentjeff: Chances are better than not that consumers of Helidon are using Jackson and Lombok.

And in such cases then we are implicitly hurting them by not including JsonIgnore.  I therefore would maintain my point that it doesn't harm anyone (internally or externally) by us having a provided/static module dependency here, but it can in fact hurt our customers that may be using it in these scenarios where they are using Jackson, Lombok, etc.

See https://medium.com/@mrkrijgsman/flexible-immutability-a475346d9518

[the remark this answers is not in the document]
m0mus: What about defining this field as `transient`?

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
spericas: Right, these should be called `typeName()` and `value()` to be consistent with Helidon conventions.

pico/types/src/main/java/io/helidon/pico/types/TypeName.java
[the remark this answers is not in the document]
spericas: I agree, it's odd to use Jackson annotations here. But perhaps the PR is too narrow focused to understand the bigger picture.

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
trentjeff: I still think this is the wrong decision, just for the record. These APIs were meant to mirror TypeName -> https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Type.html and AnnotationAndValue -> https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/AnnotationValue.html

But it seems like consistency to the rest of Helidon for the API contracts is paramount, so I ended up pushing the suggested changes.

pico/types/src/main/java/io/helidon/pico/types/TypeName.java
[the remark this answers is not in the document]
trentjeff: i took it out of the contracts for this PR, but we will see these again when we get to the impl modules :-)

pico/api/src/main/java/io/helidon/pico/api/Contract.java
```diff
@@ -0,0 +1,40 @@
...
+package io.helidon.pico.api;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The Contract annotation is used to relay significance to the type. While remaining optional in its use, it is typically placed
+ * on an interface definition to signify that the given type can be used for lookup in the service registry, or otherwise be
```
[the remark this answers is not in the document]
trentjeff: i thought it was 130?  we have a checksyle failure for 130.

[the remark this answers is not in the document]
spericas: Yes, max is 130, but I thought 120 was recommended. Interestingly, `RunLevel.java` seems to be over but is passing the check.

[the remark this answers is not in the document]
trentjeff: i will address if i have to make any more changes, otherwise will leave as-is thanks!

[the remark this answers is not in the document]
tomas-langer: Let's leave this to the checkstyle setup. 
We allo longer lines if the line is a link (and a few other)

tomas-langer (changes requested):

pico/types/src/main/java/io/helidon/pico/types/AnnotationAndValue.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    String value();
+
+    /**
+     * Get a value of an annotation property.
+     *
+     * @param name name of the annotation property
+     * @return string value of the property
+     */
+    String valueOf(String name);
```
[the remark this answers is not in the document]
trentjeff: renamed to value() and pushed.

pico/types/src/main/java/io/helidon/pico/types/TypeName.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * @return the fully qualified name which includes the use of generics/parameterized types, arrays, etc.
+     */
+    String fqName();
+
+    /**
+     * @return the base type name given the set package and class name, but not including the generics/parameterized types.
+     */
+    String getName();
```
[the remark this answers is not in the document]
trentjeff: renamed to name() and pushed.

tomas-langer (approved):
LGTM

spericas (approved):

Issue #4567, opened 2022-07-18
Grpc component does not handle "package" directive in proto files.


dgnemo:
## Environment Details
* Helidon Version: 2.x
* Helidon SE
* JDK version: 11
* OS: linux

## Problem Description

ServiceDescriptor class doesn't handle proto definition that use the `package` directive.

When Helidon compose the class name of the input/output types [1], it uses `type.getFullName()`, but that method returns `the type's fully-qualified name, within the proto language's namespace.` not just the Java classname relative to the java Outerclass.

[1] https://github.com/oracle/helidon/blob/6cf9874809e29ab3ed3f025cea5d5255124e731d/grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java#L705

## Steps to reproduce

1. Change `/home/diego/github/helidon/grpc/server/src/test/proto/echo.proto` to include: `package helidon.tests;`
2. Execute `mvn test`

```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running io.helidon.grpc.server.ServiceDescriptorTest
[ERROR] Tests run: 26, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.332 s <<< FAILURE! - in io.helidon.grpc.server.ServiceDescriptorTest
[ERROR] io.helidon.grpc.server.ServiceDescriptorTest.shouldBuildFromProtoFile  Time elapsed: 0.128 s  <<< ERROR!
java.lang.RuntimeException: java.lang.ClassNotFoundException: io.helidon.grpc.server.test.Echo$helidon$tests$EchoRequest
        at io.helidon.grpc.server.ServiceDescriptor$Builder.getTypeFromMethodDescriptor(ServiceDescriptor.java:713)
        at io.helidon.grpc.server.ServiceDescriptor$Builder.createMethodDescriptor(ServiceDescriptor.java:668)
        at io.helidon.grpc.server.ServiceDescriptor$Builder.unary(ServiceDescriptor.java:531)
        at io.helidon.grpc.server.ServiceDescriptor$Builder.unary(ServiceDescriptor.java:524)
        at io.helidon.grpc.server.ServiceDescriptorTest.shouldBuildFromProtoFile(ServiceDescriptorTest.java:542)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #5150, opened 2022-10-10
2.x: Issue 4567 - Grpc component Does not handle "package" directive in proto files.

klustria:
Problems:
1. `ServiceDescriptor$Builder.getTypeFromMethodDescriptor` throws `ClassNotFoundException`. This is the original issue reported by the filer. 
   Example: ` java.lang.RuntimeException: java.lang.ClassNotFoundException: io.helidon.grpc.server.test.Echo$helidon$tests$EchoRequest`
2. Request throws this exception: `io.grpc.StatusRuntimeException: UNIMPLEMENTED: Method not found`. This issue was only exposed when `mvn verify` was used which triggered 3 additional integration test: `ContextIT`, `TracingIT` and `SslIT` that all encountered the same problem. 
    Example: ```[ERROR] io.helidon.grpc.server.ContextIT.shouldObtainValueFromContextForRequest  Time elapsed: 0.192 s  <<< ERROR!
         io.grpc.StatusRuntimeException: UNIMPLEMENTED: Method not found: helidon.test.EchoService/Echo
                 at io.helidon.grpc.server.ContextIT.shouldObtainValueFromContextForRequest(ContextIT.java:90)```

Solutions:
1. For the 1st Problem, change `className` generation in `ServiceDescriptor$Builder.getTypeFromMethodDescriptor` of using `type.getFullName().replace('.', '$')` to instead use `type.getName()`.
4. For 2nd Problem, There are 2 fixes that needs to be done:
   a. `io.grpc.MethodDescriptor()` needs to have the service part of the fullMethodName to be prefixed with the package directive name, example it should be `helidon.test.EchoService/Echo` instead of `EchoService/Echo` if the package directive name was set to `helidon.test`. Not doing this will cause `registry.lookupMethod(methodName)` call in `ServerImpl.runInternal()` in the `gprc-java-core` library to throw `UNIMPLEMENTED: Method not found` as it won't be able to find it from the `io.grpc.HandlerRegistry` without the service name prefixed with the package name. This will be fixed in `ServiceDescriptor$Builder.build()` and `ServiceDescriptor$Builder.createMethodDescriptor()` by passing in the fullMethodName (that prefixes it with the package name) through the `fullName()` method in `io.helidon.grpc.server.MethodDescriptor$Builder` which in turn sets the `io.grpc.MethodDescriptor()` that it encapsulates.
   b. Even if above is fixed, `io.grpc.ServiceDescriptor()` also needs to have its name be prefixed with the package directive name. This is because when `io.grpc.ServiceDescriptor()` is instantiated, it validates its name against the service Name of all `io.grpc.MethodDescriptor()` objects that it holds and will throw an error like this: `java.lang.IllegalArgumentException: service names helidon.test.EchoService != EchoService` if they don't match. To fix this issue, `BindableServiceImpl.bindService()` needs to use the fullServiceName (i.e. the name of the service prefixed with the package name) in `io.grpc.ServiceDescriptor.newBuilder()`.

### DIFF
--- a/grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
+++ b/grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
@@ -54,6 +54,7 @@ public class ServiceDescriptor {
 
     private final String name;
     private final String fullName;
+    private final String protoPackage;
     private final Map<String, MethodDescriptor> methods;
     private final PriorityBag<ServerInterceptor> interceptors;
     private final Map<Context.Key<?>, Object> context;
@@ -61,19 +62,28 @@ public class ServiceDescriptor {
     private final Descriptors.FileDescriptor proto;
 
     private ServiceDescriptor(String name,
-                              String fullName,
                               Map<String, MethodDescriptor> methods,
                               PriorityBag<ServerInterceptor> interceptors,
                               Map<Context.Key<?>, Object> context,
                               HealthCheck healthCheck,
                               Descriptors.FileDescriptor proto) {
-        this.name = Objects.requireNonNull(name);
-        this.fullName = fullName;
+        String assignedName = Objects.requireNonNull(name);
         this.methods = methods;
         this.context = Collections.unmodifiableMap(context);
         this.healthCheck = healthCheck;
         this.interceptors = interceptors.copyMe();
         this.proto = proto;
+
+        protoPackage = proto == null ? "" : proto.getPackage();
+        String servicePrefix = protoPackage + (!protoPackage.isEmpty() ? "." : "");
+        if (!servicePrefix.isEmpty() && assignedName.startsWith(servicePrefix)) {
+            // If assignedName is already prefixed with package name, remove the package name part
+            this.name = assignedName.replace(servicePrefix, "");
+            this.fullName = this.name;
+        } else {
+            this.name = assignedName;
+            this.fullName = servicePrefix + assignedName;
+        }
     }
 
     /**
@@ -82,6 +92,22 @@ public class ServiceDescriptor {
      */
     public String name() {
         return name;
+    }
+
+    /**
+     * Returns the service name prefixed with package directive if one exists.
+     * @return service name prefixed with package directive if one exists.
+     */
+    public String getFullName() {
+        return fullName;
+    }
+
+    /**
+     * Returns package name from proto file.
+     * @return package name from proto file
+     */
+    public String getPackage() {
+        return protoPackage;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


klustria:
This proposed merge was created to fix https://github.com/helidon-io/helidon/issues/4567

thegridman (approved):

grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
```diff
@@ -131,6 +131,22 @@ public Descriptors.FileDescriptor proto() {
...
     }
 
+    /**
+     * Returns the service name prefixed with package directive if one exists.
+     */
+    String getFullName() {
+        return getPackagedName(proto, name);
+    }
+
+    static String getPackagedName(Descriptors.FileDescriptor proto, String name) {
```
aseovic, line 141: Why is this a static method? It looks to me like this logic really belongs to the `getFullName` method above (which should probably be public as well), and there is no need for a separate, static method at all.

grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
```diff
@@ -131,6 +131,22 @@ public Descriptors.FileDescriptor proto() {
         return proto;
     }
 
+    /**
+     * Returns the service name prefixed with package directive if one exists.
+     */
+    String getFullName() {
```
aseovic, line 137: This method should be public.

grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
```diff
@@ -719,6 +738,13 @@ private String getPackageName() {
             return "".equals(pkg) ? proto.getPackage() : pkg;
         }
 
+        /**
+         * Returns the service name prefixed with package directive if one exists.
+         */
+        private String getFullName() {
```
aseovic, line 744: I see, you need it within the `Builder` as well...

Maybe that's the only place you should have this code, and the outer `ServiceDescriptor` should simply have a `fullName` field that will be initialized by the `Builder`, along with other fields?

grpc/server/src/test/proto/echopackage.proto
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+syntax = "proto3";
+
+package helidon.test;
+
+option java_package = "io.helidon.grpc.server.test";
```
aseovic, line 21: So this will only test that the `java_package` works as expected, right?

Do we already have the tests that ensure that the `package` is used instead when `java_package` is missing? If not, we should add those as well.

aseovic (changes requested):
Overall it looks good, but I think it can be cleaned up a bit.

grpc/server/src/test/proto/echopackage.proto
klustria: No we don't. I can check that out.

grpc/server/src/main/java/io/helidon/grpc/server/ServiceDescriptor.java
klustria: Can you please check if the new change I pushed is the approach you are envisioning? I maintained ServiceDescriptor.getFullName() method as `protected` rather than `public` is it is only used within the package.

grpc/server/src/test/proto/echopackage.proto
klustria: Added

aseovic (approved):
LGTM

Issue #4936, opened 2022-09-22
OIDC multitenant support


Verdent:
We should add OIDC multitenant support.

edbratt:
How does this relate to https://github.com/helidon-io/helidon/issues/378?

Issue #4383, opened 2022-06-13
OIDC lazy loading of metadata


tomas-langer:
Currently when Helidon starts, it loads Open ID Connect (OIDC) metadata from the "well known" endpoint by default.
To have better startup speed, and the capability to boot without the OIDC available, we should load this lazily on first request (with configurable "fail fast" flag to keep current behavior when wanted/needed).

tomas-langer:
This change should be combined with support for multitenancy in OIDC. 
@Verdent please contact me for a prototype

Issue #2075, opened 2020-06-22
Issuer validator should accept List of String as possible issuers


abhishek-mvr:
Issuer validator should be able to validate against multiple possible jwt issuers. This would benefit when integrating to multiple services.

https://github.com/oracle/helidon/blob/c0f4a86acde11d33e593bb88c8d5f0ddcf99d82f/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java#L408

barchetta:
Also, see this issue from `java-jwt`: https://github.com/auth0/java-jwt/issues/246

tomas-langer:
Issuer validator is validating a token against a single issuer (the one that our configuration is created for) - an issuer is mapped to either an endpoint (metadata), or `jwk`  to validate  a signature (or to decrypt).

So what I understand from this is that we want to choose an OIDC provider (JWT configuration) based on the issuer in the JWT.
This could be achieved through our new multitenancy support in OIDC, where the tenant finder would choose tenant id from the JWT issuer (would cause double parsing of the token).
We may want to consider this use case when selecting the correct OIDC configuration for validation.

@Verdent FYI

tomas-langer:
We are currently investigating this issue. 
This is basically a request for federation. We are looking into the possibilities of achieving this with third party tools (such as Keycloak) and integration with Helidon through existing means (OIDC). 
We have a workaround (using multiple providers in sequence, until one that can authenticate is found) - this is not the most efficient way of solving this problem, but it should work (as issuer would fail validation very quickly, so there will be no network traffic involved for providers that do not match it).

After our investigation, we will decide how to treat this issue (enhancement request, guide, workaround).

tomas-langer:
The referenced issue (really) mentions support for an array of issuers as a claim in JWT (as @Verdent noticed).
This fix would be quite easy, nevertheless I do have issues with it:
If a JWT defines multiple issuers, what does it mean?
- Are these aliases for the same issuer?
- Are these issuers sharing private keys (please do not do this) and provide replaceable tokens?

Also the JWT specification clearly states that `iss` claim is a single string, not an array:
https://www.rfc-editor.org/rfc/rfc7519#section-4.1.1

tomas-langer:
@barchetta could you please go through my comments and provide feedback?

barchetta:
[...]

PR #5168, opened 2022-10-13
2.x: OIDC multi-tenant and lazy loading implementation

Verdent:
Fixes https://github.com/helidon-io/helidon/issues/4936
Fixes https://github.com/helidon-io/helidon/issues/4383


tests/integration/oidc/src/test/java/io/helidon/tests/integration/oidc/IntegrationTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.junit.jupiter.api.Test;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+
+@Testcontainers
+public class IntegrationTest {
```
tomas-langer, line 31: The naming convention should be `...IT` for integration tests, where the name is a usual test name (such as `OidcMultitenantIT`)

tests/integration/oidc/src/test/java/io/helidon/tests/integration/oidc/IntegrationTest.java
```diff
@@ -0,0 +1,133 @@
...
+            .property(HelidonProperties.CONFIG,
+                      Config.builder()
+                              .disableSystemPropertiesSource()
+                              .disableEnvironmentVariablesSource()
+                              .build()
+                              .get("client"));
+    private static final Client CLIENT = ClientBuilder.newClient(CONFIG);
+
+    @Container
+    public static KeycloakContainer keycloakContainer = new KeycloakContainer()
```
tomas-langer, line 45: must this be public?

tests/integration/oidc/src/test/java/io/helidon/tests/integration/oidc/IntegrationTest.java
```diff
@@ -0,0 +1,133 @@
...
+            .withRealmImportFiles("/test-realm.json", "/test2-realm.json")
+            // this enables KeycloakContainer reuse across tests
+            .withReuse(true);
+
+    private static final int PORT = 7777;
+    private static Server server;
+
+    @BeforeAll
+    public static void beforeAll() {
```
tomas-langer, line 54: These methods should be package local, not public (all JUnit methods)

tests/integration/oidc/src/test/java/io/helidon/tests/integration/oidc/IntegrationTest.java
```diff
@@ -0,0 +1,133 @@
...
+        System.setProperty("security.providers.1.oidc.identity-uri",
+                           keycloakContainer.getAuthServerUrl() + "realms/test/");
+        System.setProperty("security.providers.1.oidc.tenants.0.identity-uri",
+                           keycloakContainer.getAuthServerUrl() + "realms/test2/");
+        System.setProperty("security.providers.1.oidc.frontend-uri", "http://localhost:" + PORT);
+        server = Server.builder()
+                .port(PORT)
```
tomas-langer, line 61: Hardcoded port is not good for testing, is there a way to use a random port and set the properties post server start?

webclient/webclient/src/main/java/io/helidon/webclient/WebClientConfiguration.java
```diff
@@ -578,7 +578,7 @@ public B addMediaSupport(MediaSupport mediaSupport) {
             return me;
         }
 
-        private B enableAutomaticCookieStore(Boolean enableAutomaticCookieStore) {
+        public B enableAutomaticCookieStore(Boolean enableAutomaticCookieStore) {
```
tomas-langer, line 581: If this is public, then it either misses `@Override` or javadoc

tomas-langer (changes requested):

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcProvider.java
```diff
@@ -233,298 +131,23 @@ public Collection<Class<? extends Annotation>> supportedAnnotations() {
...
-            if (LOGGER.isLoggable(Level.FINEST)) {
-                // only log errors when details requested
-                errors.log(LOGGER);
-                validationErrors.log(LOGGER);
-            }
-            return errorResponse(providerRequest, Http.Status.UNAUTHORIZED_401, "invalid_token", "Token not valid");
-        }
+        String tenantId = tenantIdFinders.stream()
+                .map(tenantIdFinder -> tenantIdFinder.tenantId(providerRequest))
+                .filter(Optional::isPresent)
```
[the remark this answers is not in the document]
Verdent: good point.

tests/integration/oidc/pom.xml
```diff
@@ -0,0 +1,89 @@
...
+        <artifactId>helidon-tests-integration</artifactId>
+        <groupId>io.helidon.tests.integration</groupId>
+        <version>2.5.4-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>oidc</artifactId>
+
+    <properties>
+        <maven.compiler.source>19</maven.compiler.source>
```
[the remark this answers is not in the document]
Verdent: IDEA... I missed that

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/Tenant.java
```diff
@@ -0,0 +1,281 @@
...
+import io.helidon.security.providers.httpauth.HttpBasicOutboundConfig;
+import io.helidon.webclient.WebClient;
+import io.helidon.webclient.security.WebClientSecurity;
+
+import org.glassfish.jersey.client.authentication.HttpAuthenticationFeature;
+
+/**
+ * Holder of the tenant configuration resolved in runtime. Used for OIDC lazy loading.
+ */
+public final class Tenant {
```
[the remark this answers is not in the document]
trentjeff: +1 on TenantConfig being a iface - this will also fit nicely in the future as we adopt more pico config-driven.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -641,38 +505,43 @@ public String paramName() {
...
+     * Tenant query parameter name.
      *
-     * @return if cookie should be used
-     * @see Builder#useCookie(Boolean)
+     * @return name of the tenant query parameter to use
+     * @see Builder#paramNameTenant(String)
      */
-    public boolean useCookie() {
-        return useCookie;
+    public String paramNameTenant() {
```
tomas-langer, line 513: Use approach similar to `cookie` - e.g. this method should be named `tenantParamName`

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcConfigAbstractTest.java
```diff
@@ -73,7 +73,7 @@ void testDefaultValues() {
                   () -> assertThat("Issuer", config.issuer(), nullValue()),
                   () -> assertThat("Client without authentication", config.generalClient(), notNullValue()),
                   () -> assertThat("Client with authentication", config.appClient(), notNullValue()),
-                  () -> assertThat("JWK Keys", config.signJwk(), notNullValue())
+                  () -> assertThat("JWK Keys", config.signJwk(), nullValue())
```
tomas-langer, line 76: We do not return `null` from APIs. The original assertion was correct

tomas-langer: If there is a chance this is really null, we should throw an exception instead (such as `IllegalStateException("OIDC Configuration is not yet loaded")` or whatever is relevant in this case

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcConfigFromBuilderTest.java
```diff
@@ -99,7 +100,7 @@ void testDefaultValues() {
                 () -> assertThat("Issuer", config.issuer(), nullValue()),
                 () -> assertThat("Client without authentication", config.generalClient(), notNullValue()),
                 () -> assertThat("Client with authentication", config.appClient(), notNullValue()),
-                () -> assertThat("JWK Keys", config.signJwk(), notNullValue())
+                () -> assertThat("JWK Keys", config.signJwk(), nullValue())
```
tomas-langer, line 103: Same as above

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/spi/TenantIdProvider.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package io.helidon.security.providers.oidc.common.spi;
+
+import io.helidon.config.Config;
+
+/**
+ * Java {@link java.util.ServiceLoader} service interface for multitenancy support.
+ */
+public interface TenantIdProvider {
```
trentjeff, line 24: why do we have two/three levels of abstraction here? Provider -> Finder -> Optional<String> ?

1. I can see the need for Provider finding the 1st Finder that is capable of returning a non-optional tenant-id, in weighted order.  Is that the intent?  If so then we should provide methods on Provider that do sort of thing.
2. If that is not the intent then why not just use serviceloader directly on the Finder; no need to go to provider first.

pom.xml
```diff
@@ -60,6 +60,7 @@
         <version.lib.groovy-all>2.4.14</version.lib.groovy-all>
         <!-- Silence javadoc error org.jboss.logging.annotations.Message$Format not found -->
         <version.lib.jboss-logging-annotations>2.2.1.Final</version.lib.jboss-logging-annotations>
+        <version.lib.jsoup>1.15.3</version.lib.jsoup>
```
trentjeff, line 63: Using this location to just add a general comment.

My main recommendation for this change is to add more documentation, a README.md would be ideal that showcases the problem statement, approach taken, and finally usage pattern.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcSupport.java
```diff
@@ -208,6 +219,13 @@ public void update(Routing.Rules rules) {
     }
 
     private void processLogout(ServerRequest req, ServerResponse res) {
+        String tenantName = req.queryParams().first(oidcConfig.paramNameTenant()).orElse(TenantConfigFinder.DEFAULT_TENANT_ID);
```
tomas-langer, line 222: as far as I can see, we can either use a parameter, or a cookie. This is only using parameter.

tomas-langer (changes requested):

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcConfigAbstractTest.java
Verdent: I have updated JWK value assignment to be more aligned with the previous approach. Now it never returns null.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
Verdent: renamed

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/spi/TenantIdProvider.java
Verdent: With having TenantIdProvider it is possible to create new TenantIdFinder with the config. (user can specify their own config structure how to create finder)
Finder itself is there for finding the tenant id and if it is missing (tenant id), we do not want to return null, but rather Optional. That is the reason for structure like this.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcSupport.java
```diff
@@ -260,6 +264,44 @@ private void processLogout(ServerRequest req, ServerResponse res) {
...
+                    .findFirst();
+
+            if (tenantId.isEmpty()) {
+                missingLocations.add("cookie");
+            }
+        }
+        if (tenantId.isPresent()) {
+            return tenantId.get();
+        } else {
+            LOGGER.finest(() -> "Missing tenant id, could not find in either of: " + missingLocations);
```
tomas-langer, line 291: 1. This should be guarded by `if (LOGGER.isEnabledFor(FINEST)` (pseudocode)
2. Use only a single log statement - does not make sense to call logger twice for a single message.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcSupport.java
```diff
@@ -260,6 +264,44 @@ private void processLogout(ServerRequest req, ServerResponse res) {
...
+            LOGGER.finest(() -> "Falling back to the default tenant id: " + DEFAULT_TENANT_ID);
+            return DEFAULT_TENANT_ID;
+        }
+    }
+
+    private String getValueFromSingle(Single<String> cookieSingle) {
+        try {
+            return cookieSingle.get();
```
tomas-langer, line 299: This is wrong - this will make us blocking

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
```diff
@@ -294,7 +294,7 @@ private AuthenticationResponse errorResponse(ProviderRequest providerRequest,
             String authorizationEndpoint = tenant.authorizationEndpointUri();
             String nonce = UUID.randomUUID().toString();
             String redirectUri =
-                    encode(redirectUri(providerRequest.env()) + "?" + oidcConfig.paramNameTenant() + "=" + tenantId);
+                    encode(redirectUri(providerRequest.env()) + "?" + oidcConfig.tenantParamName() + "=" + tenantId);
```
tomas-langer, line 297: are you sure this whole thing should be encoded? This will encode even the question mark and equals sign.

tomas-langer (changes requested):

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
Verdent: Yes, this is query parameter. This whole value.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcSupport.java
Verdent: OK, will fix that

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcProvider.java
```diff
@@ -165,30 +170,22 @@ private String findTenantIdFromRedirects(ProviderRequest providerRequest) {
...
-        }
-    }
-
-    private String getValueFromSingle(Single<String> cookieSingle) {
-        try {
-            return cookieSingle.get();
-        } catch (Exception e) {
-            throw new RuntimeException(e);
+            if (LOGGER.isLoggable(Level.FINEST)) {
+                LOGGER.finest(() -> "Missing tenant id, could not find in either of: " + missingLocations + "\n"
```
tomas-langer, line 185: when guarded by `isLoggable`, do not use a lambda

tomas-langer: Also do not add new lines to log output, just keep it as a single line, single message

tomas-langer (changes requested):
A very small change requested

tomas-langer (approved):
LGTM
May be related to #2075

PR #5195, opened 2022-10-14
4.x: Introduces Pico Builder

trentjeff:


pico/types/src/main/java/io/helidon/pico/types/DefaultTypedElementName.java
```diff
@@ -0,0 +1,283 @@
...
+        return toDeclaration();
+    }
+
+    /**
+     * Provides a description for this instance.
+     *
+     * @return provides the {typeName}{space}{elementName}
+     */
+    public String toDeclaration() {
+//        if (Objects.nonNull(typeName()) && Objects.nonNull(elementName())) {
```
spericas, line 110: Can we remove all this commented out code?

trentjeff: removed and pushed.

pico/pom.xml
```diff
@@ -48,7 +48,7 @@
         <module>api</module>
 <!--        <module>spi</module>-->
         <module>types</module>
-<!--        <module>builder</module>-->
+        <module>builder</module>
 <!--        <module>config</module>-->
```
spericas, line 52: I'm assuming all these other modules will come in a future PR, right?

trentjeff: Correct.

spericas (approved):
Technically, this PR is perfectly sound and clear, and that is why I'm approving it. Architecturally, I have some concerns about adding (mostly independent and potentially "sold" separately) "Pico" modules into the Helidon repo. This is also emphasized by the current naming that uses "Helidon Pico" as opposed to just "Helidon" for naming modules.  

It would be nice to get others to review this PR as well.

PR #5220, opened 2022-10-20
Broad changes to rework headers and parameters

### LINKED ISSUE #5149: All implementations of headers need to be case-insensitive
## Environment Details
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Helidon contains several implementations of headers. Not all of them conform to HTTP's requirement that headers be case-insensitive.

We have a separate issue https://github.com/helidon-io/helidon/issues/4956 as well as internal Slack channel reports:
> I believe there's a bug in io.helidon.common.http.Parameters.toMap() in Helidon 2.5.1
It seems, 2 of the 5 implementations I could find don't preserve the case insensitivity of the keys.
This leads to not finding headers if they aren't cased by the client as we expect. This case-insensitivity is part of the http spec.
Here are the 5 places I could find:
> * HashParameters, ReadOnlyParameters, and UnmodifiableParameters in the common.http package
> * WebClientQueryParams, and WebClientRequestHeadersImpl in the webclient package
The two are: ReadOnlyParameters and WebClientRequestHeadersImpl

The afflicted implementations percolate into other, more directly visible places (potentially request headers, response headers, etc.). We need to review all.

## Steps to reproduce
For one, see the referenced issue about B3 headers in tracing.

tjquinno:
Resolves #5149

HTTP header names should be case-insensitive; HTTP parameters (query parameter, path params) should be case-sensitive.

This PR fixes problems in both areas.

I tried to be as minimally-invasive as possible while still addressing the problems. Still, there are quite a few changes needed to fix this situation.

## Some highlights:

Key changes are in `helidon-common-http` but other components are affected as well.

* The new class `HashHeaders` extends `HashParameters`. Each now creates the correct type of `Map` (case-sensitive or -insensitive) in separate protected factory methods. Various writeable headers-related classes now extend `HashHeaders` (instead of extending `HashParameters` as they did before) and so now inherit the correct case-insensitive behavior.
* The new class `ReadOnlyHeaders` extends `ReadOnlyParameters`. Each now creates the correct type of `Map` in protected factory methods for various purposes. Various read-only headers-related classes now extend `ReadOnlyHeaders` to inherit the correct case-insensitive behavior.
* Several other classes are changed to use `Headers` or the various implementations rather than `Parameters` so the data is handled in a case-insensitive manner.
* Tests:
   *Lots of new tests check for the appropriate behavior in many classes related to headers and parameters.
   * A new tracing-specific test checks for proper B3 propagation from client to server.

## Some lower-level details:

`Parameters` now extends `Iterable<Map.Entry<String, List<String>>>` and provides a default implementation of `iterator`. This reduces creation of temporary data structures in other code. Our classes which implement `Parameters` override iterator with more efficient code than the default.

There is opportunity to refactor significantly to reduce some code duplication. But it would be quite a bit more far-reaching than the proposed changes here. Plus, header handling is much different in 3.x so any refactoring work in 3.x would not carry forward and is likely not worth the investment.

### DIFF
--- a/common/http/pom.xml
+++ b/common/http/pom.xml
@@ -42,6 +42,11 @@
             <artifactId>helidon-common-context</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
--- a/common/http/src/main/java/io/helidon/common/http/HashHeaders.java
+++ b/common/http/src/main/java/io/helidon/common/http/HashHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package io.helidon.common.http;
 
 import java.util.ArrayList;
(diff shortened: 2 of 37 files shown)


tjquinno:
Ugh. Created the PR from the wrong working branch. I'll create a corrected PR.

PR #5225, opened 2022-10-20
JAXRS doc updates for 3611

### LINKED ISSUE #3611: Add documentation that Helidon can invoke JAX-RS Application.getSingletons multiple times
## Environment Details
* Helidon Version: 2.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
The JAX-RS `Application` class exposes the `getSingletons` method which returns instances of resources, providers, or features. Currently Helidon can invoke this method multiple times. The JAX-RS spec does not mention whether developers can expect multiple invocations or always at most one.

Our doc should make clear that developers should expect multiple invocations of the method. If they want to return the same instances on all calls, they should cache the values themselves.

ljamen:
fixes #3611
https://github.com/helidon-io/helidon/issues/3611

### DIFF
--- a/docs/mp/jaxrs/jaxrs-applications.adoc
+++ b/docs/mp/jaxrs/jaxrs-applications.adoc
@@ -25,8 +25,8 @@ include::{rootdir}/includes/mp.adoc[]
 
 == JAX-RS Applications
 
-NOTE: In this section we shall distinguish the notion of a JAX-RS `Application` subclass
-from a Helidon application. As we shall learn shortly, the latter may include zero or more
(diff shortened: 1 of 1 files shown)


tjquinno (approved):
Looks good.

It's possible that Helidon invokes the `Application` `getClasses` method (exposed like `getSingletons`) multiple times, although I'm not sure about that. Probably developers should not assume single invocations of either method. Not sure this warrants further changes so I'm approving this. Maybe Santiago would know for sure?

PR #5250, opened 2022-10-25
4269 openapi tools examples

### LINKED ISSUE #4269: OpenAPITool Helidon Examples
- [x] Server MP
- [x] Server SE
- [x] Client MP
- [x] Client SE

aserkes:
Helidon examples for OpenApi Tools

resolves #4269


examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,352 @@
...
+    private static Routing createRouting(Config config) {
+
+        return Routing.builder()
+                      .register("/", new MessageService())
+                      .build();
+    }
+}
+```
+
+4) Create `resource` folder and put a `application.yaml` file inside.
```
tjquinno, line 252: Change to 

Create the directory `src/main/resources/`. Create `application.yaml` in that directory with the following content:

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,352 @@
...
+  port: 8081
+  host: localhost
+```
+
+To run the application :
+
+With JDK17+
+```bash
+mvn package
+java -jar target/openapi-se-client.jar
```
tjquinno, line 264: The build creates `openapi-java-client.jar` for me.

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,352 @@
...
+curl -X GET http://localhost:8081/greet
+{"message":"Lisa","greeting":"Hola"}
+
+curl -X GET http://localhost:8081/greet/Joe
+{"message":"Joe","greeting":"Hola"}
+```
+
+## Update applications
+
+To keep the server and the client up to date according to the OpenApi document, we can use the maven plugin.
```
tjquinno, line 281: Please add instructions showing how users can run the profile to run the plug-in to generate the code. Maybe something like the next line:

To run the generator during your build, invoke the profile: `mvn clean package -P openapi`

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,352 @@
...
+
+Add these lines to the `pom.xml` of our server :
```
tjquinno, line 283: This did not work for me without adding 

```
<version.openapi.generator.maven.plugin>6.2.1</version.openapi.generator.maven.plugin>
```
to the `<properties>` section.

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,352 @@
...
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+```
+
+For the client application :
+1) Copy the OpenApi document `path-to-openapi-doc/quickstart.yaml` to the folder `resources` and rename it to `openapi.yaml`.
+2) Add these lines to the `pom.xml` of our client :
```
tjquinno, line 319: Same problem here with  `<version.openapi.generator.maven.plugin>6.2.1</version.openapi.generator.maven.plugin>`

examples/openapi-tools/quickstart-mp/README.md
```diff
@@ -0,0 +1,258 @@
...
+        defaultMessage.set(message);
+    }
+```
+4) Replace implementation of the method `public Message getDefaultMessage()` by this:
+```java
+        return defaultMessage.get();
+```
+5) Replace implementation of the method `public Message getMessage(@PathParam("name") String name)` by this:
+```java
+        defaultMessage.get().setMessage(name);
```
tjquinno, line 46: In both servers this code changes the name in the default greeting. This is different from what the normal QuickStart server does, which is to change the name only in the _returned_ greeting, not the default one used for later requests.

It will be less confusing if this app behaves the same way. I suggest that `getMessage` create its own local instance of `Message` based on the default one, then change the name of the local instance and return the local instance, leaving the default unchanged:

```
public void getMessage(ServerRequest request, ServerResponse response) {
        String name = request.path().param("name");
        Message msg = new Message();
        msg.setGreeting(defaultMessage.get().getGreeting());
        msg.setMessage(name);
        response.send(msg);
    }
```

examples/openapi-tools/quickstart-mp/README.md
```diff
@@ -0,0 +1,258 @@
...
+    private final AtomicReference<Message> defaultMessage = new AtomicReference<>();
+```
+3) Add default constructor to the class :
+```java
+    public MessageServiceImpl() {
+        Message message = new Message();
+        message.setMessage("World");
+        message.setGreeting("Hello");
+        defaultMessage.set(message);
+    }
```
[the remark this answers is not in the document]
aserkes: The described scenario works and I do not need to change anything because I used `@ApplicationScoped` for `MessageServiceImpl` class, but I have noticed that the method `public Message getMessage(@PathParam("name") String name)` does not work as expected so I am going to change it.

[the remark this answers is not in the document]
tjquinno: Oh, yes. I missed that step in the instructions.

I did not notice a problem with the `getMessage` method but I look forward to seeing your changes.

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,383 @@
+# Helidon OpenAPI Generator example for Helidon SE server and client
+
+The goal of this example is to show how a user can easily create a Helidon SE server or client from an OpenAPI document using OpenAPI Generator.  
+
+Here we will show the steps that a user has to do to create Helidon SE server and client using OpenAPI Generator and what has to be done to make the generated server and client fully functional.
+
+For generation of our projects we will use `openapi-generator-cli.jar` that can be downloaded form the maven repository (instructions and other options can be found [here](https://openapi-generator.tech/docs/installation)) and OpenAPI document `quickstart.yaml` that can be found next to this `README.md`.
```
tjquinno, line 7: "form" -> "from"

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,383 @@
...
+        <groupId>io.helidon.webserver</groupId>
+        <artifactId>helidon-webserver</artifactId>
+    </dependency>
+    <dependency>
+        <groupId>io.helidon.config</groupId>
+        <artifactId>helidon-config-yaml</artifactId>
+    </dependency>
+```
+
+2) Let's add a class `MessageService` that will use `MessageApi` and `ApiClient` to interact with the server :
```
tjquinno, line 129: We should either:

- add "to the `api` package" after "add a class `MessageService`", or
-  include the following line at the beginning of the code example: `package org.openapitools.client;` 

so it's clear where we want readers to add the file.

examples/openapi-tools/quickstart-se/README.md
```diff
@@ -0,0 +1,383 @@
...
+    }
+```
+4) Replace implementation of the method `public void getDefaultMessage(ServerRequest request, ServerResponse response)` by this:
+```java
+        response.send(defaultMessage.get());
+```
+5) Replace implementation of the method `public void getMessage(ServerRequest request, ServerResponse response)` by this:
+```java
+        String name = request.path().param("name");
+        defaultMessage.get().setMessage(name);
```
tjquinno, line 46: In my  comment in my previous review on the `quickstart-mp/README.md` I mentioned that _both_ servers had the same problem; they both changed the default greeting's name in the `getMessage` method which should not happen. 

I probably should have repeated that comment here at that time.

The latest changes to the PR fix the MP example but not the SE example here. This SE `getMessage` code still changes the default name.

tjquinno (changes requested):
Just a few more changes.

examples/openapi-tools/quickstart-se/se-server/src/main/java/org/openapitools/server/api/MessageServiceImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        defaultMessage.set(message);
+    }
+
+    public void getDefaultMessage(ServerRequest request, ServerResponse response) {
+        response.send(defaultMessage.get());
+    }
+
+    public void getMessage(ServerRequest request, ServerResponse response) {
+        String name = request.path().param("name");
+        Message result = new Message();
```
tjquinno, line 51: Here you used the suggested fix I mentioned in my previous review, but I like the fix you did in the MP example better:

```
        Message result = new Message();
        return result.message(defaultMessage.get().getMessage()).greeting(name);
```
Just make sure that the instructions in the `README.md` and the code in this example file are the same.

tjquinno (changes requested):

examples/openapi-tools/quickstart-se/README.md
aserkes: Sorry, I forgot to update README.md

examples/openapi-tools/quickstart-se/se-server/src/main/java/org/openapitools/server/api/MessageServiceImpl.java
aserkes: Unfortunately, it is not possible because the class `Message` contains only standard getters and setters.

tjquinno: Good point! (Our generators should probably all share the same model template but are clearly not.)

tjquinno (approved):

Issue #4722, opened 2022-08-11
Metrics TCK runs twice with different test counts


tjquinno:
## Environment Details
* Helidon Version: 2.x, 3.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
Our driver project which runs the MP metrics TCK triggers two TCK runs which seem to duplicate some of the tests.

## Steps to reproduce
```
cd helidon/microprofile/tests/tck/tck-metrics
mvn test
```
Output shows two runs:
```
16:57:39,659 [INFO] Results:
16:57:39,660 [INFO] 
16:57:39,660 [INFO] Tests run: 233, Failures: 0, Errors: 0, Skipped: 0
16:57:39,660 [INFO] 
16:57:39,703 [INFO] 
16:57:39,703 [INFO] -------------------------------------------------------
16:57:39,703 [INFO]  T E S T S
16:57:39,703 [INFO] -------------------------------------------------------
16:57:40,213 [INFO] Running TestSuite
17:04:38,533 [INFO] Tests run: 213, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 418.3 s - in TestSuite
17:04:38,579 [INFO] 
17:04:38,579 [INFO] Results:
17:04:38,579 [INFO] 
17:04:38,579 [INFO] Tests run: 213, Failures: 0, Errors: 0, Skipped: 0
17:04:38,579 [INFO] 
```
There seems to be no inherent need to run any tests twice.

Captain1653:
I think I've found the reason. It depends on the configuration of the `maven-surefire-plugin`. Root project `tck-project` has `surefire-testng`, but `tck-metrics` ingerited this configuration and has `surefire-junit`. 


<img width="1004" alt="Screenshot 2023-03-03 at 23 38 38" src="https://user-images.githubusercontent.com/4740207/222834518-c75bb562-a55c-4e5c-86a6-7e9eead556fa.png">

### The first run - junit, the second run - testng.

<img width="1681" alt="Screenshot 2023-03-03 at 23 30 03" src="https://user-images.githubusercontent.com/4740207/222835252-01f4d8b6-0e7e-46f1-8c48-45b9fa7387d4.png">


I changed all to `surefire-junit` and these tests run once. Should I create PR? I'm not sure. 

Some modules should use `surefire-testng`, but  some`surefire-junit`. Should every module configure its own dependency for `maven-surefire-plugin`?

[...]

PR #5252, opened 2022-10-25
Use Hamcrest assertions instead of JUnit in  integrations/cdi/jpa-cdi- backport 2.x (#1749)

Captain1653:
Part of #1749 

[Original PR](https://github.com/helidon-io/helidon/pull/5251)


romain-grecourt (dismissed):

Captain1653:
@romain-grecourt Is it some pipeline error?
Screenshot 2022-10-26 at 17 32 09

romain-grecourt:
I can reproduce the issue locally.

Captain1653:
Me too. I will try to investigate.

Captain1653:
@romain-grecourt I've found the reason. It depends on `eclipselink-maven-plugin`. Build will be green If I comment it. I'll try to understand why it breaks the build.  Some note: tests run twice then. It can be connected to https://github.com/helidon-io/helidon/issues/4722 , but I'm not sure.

Screenshot 2022-11-03 at 20 23 55

Captain1653:
I tried to fix it a few days ago, but I couldn't. It depends on `eclipselink-maven-plugin`, but I don't know how :( .

ethlo:
Looks like the introspected files are not files that should be really handled by the plugin. Could you share the configuration you run the plugin with?

ethlo:
So from what I can see, you have a dependency on JUnit when running the plugin. 
Now that you have changed the tests to use Hamcrest for assertions, you need to have them available as well.

Captain1653:
I've added `hamcrest-all` for dependencies in `eclipselink-maven-plugin`. Tests are green, but they run twice.

romain-grecourt (approved):

PR #5267, opened 2022-10-27
Add relativeUris flag in OidcConfig to allow Oidc webclient to use relative path on the request URI

klustria:


spericas (dismissed):
LGTM

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcUtil.java
```diff
@@ -69,12 +72,14 @@ static ClientBuilder clientBaseBuilder(String proxyProtocol, String proxyHost, i
...
                                                   int proxyPort,
+                                                  boolean relativeUris,
                                                   Duration clientTimeout) {
         WebClient.Builder webClientBuilder = WebClient.builder()
                 .addService(WebClientTracing.create())
                 .addMediaSupport(JsonpSupport.create())
                 .connectTimeout(clientTimeout.toMillis(), TimeUnit.MILLISECONDS)
-                .readTimeout(clientTimeout.toMillis(), TimeUnit.MILLISECONDS);
+                .readTimeout(clientTimeout.toMillis(), TimeUnit.MILLISECONDS)
+                .config(Config.create(ConfigSources.create(Map.of("relative-uris", String.valueOf(relativeUris)))));
```
Verdent, line 82: I am thinking that more correct approach would be to create `relativeUris(boolean relativeUris)` method on the WebClient.Builder, rather than creating new config just to set this.

Verdent (changes requested):
LGTM with one exception.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -1951,6 +1974,22 @@ public Builder proxyPort(int proxyPort) {
...
+         * regardless of the presence or absence of proxies or no-proxy lists. By default,
+         * requests that use the Proxy will have absolute URIs. Set this flag to {@code true}
+         * if the host is unable to accept absolute URIs.
+         * Defaults to {@value #DEFAULT_RELATIVE_URIS}.
+         *
+         * @param  relativeUris relative URIs flag
+         * @return updated builder instance
+         */
+        @ConfiguredOption("false")
+        public Builder relativeUris(Boolean relativeUris) {
```
Verdent, line 1988: use primitive type boolean

Verdent (changes requested):

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
spericas: +1

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcUtil.java
klustria: Implemented and added unit test for the change.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
klustria: Changed to primitive type boolean

Verdent (approved):

PR #5284, opened 2022-10-29
Issue 4567 - Grpc component Does not handle "package" directive in proto files.

klustria:
Problems:
1. ServiceDescriptor$Builder.getTypeFromMethodDescriptor throws ClassNotFoundException. This is the original issue reported by the filer. example java.lang.RuntimeException: java.lang.ClassNotFoundException: io.helidon.grpc.server.test.Echo$helidon$tests$EchoRequest
2. Request throws this exception: "io.grpc.StatusRuntimeException: UNIMPLEMENTED: Method not found". This issue was only exposed when "mvn verify" was used which triggered 3 additional integration tests, ContextIT, TracingIT and SslIT that all encountered the same problem. example: [ERROR] io.helidon.grpc.server.ContextIT.shouldObtainValueFromContextForRequest  Time elapsed: 0.192 s  <<< ERROR!
         io.grpc.StatusRuntimeException: UNIMPLEMENTED: Method not found: helidon.test.EchoService/Echo
                 at io.helidon.grpc.server.ContextIT.shouldObtainValueFromContextForRequest(ContextIT.java:90)

Solutions:
1. For the 1st Problem, change className generation in ServiceDescriptor$Builder.getTypeFromMethodDescriptor of using "type.getFullName().replace('.', '$')" to instead use "type.getName()".
2. For 2nd Problem, There are 2 fixes that needs to be done: a. io.grpc.MethodDescriptor() needs to have the service part of the fullMethodName to be prefixed with the package directive name, example it should be "helidon.test.EchoService/Echo" instead of "EchoService/Echo" if the package directive name is "helidon.test". Not doing this will cause "registry.lookupMethod(methodName)" call in ServerImpl.runInternal() on gprc-java-core library to throw "UNIMPLEMENTED: Method not found" as it won't be able to find it without the service name prefixed with the package name. This is fixed in ServiceDescriptor$Builder.build() and ServiceDescriptor$Builder.createMethodDescriptor() by passing in the fullMethodName (that prefixes it with the package name) though the fullName() method in io.helidon.grpc.server.MethodDescriptor$Builder which in turn sets the io.grpc.MethodDescriptor() that it encapsulates. b. Even if above is fixed, io.grpc.ServiceDescriptor() also needs to have its name be prefixed with the package directive name. This is because when io.grpc.ServiceDescriptor() is instantiated, it validates its name against the service Name of all io.grpc.MethodDescriptor() objects that it holds and will throw an error like this: "java.lang.IllegalArgumentException: service names helidon.test.EchoService != EchoService". To fix this issue, BindableServiceImpl.bindService() needs to use the fullServiceName (i.e. the name of the service prefixed with the package name) when creating a new Builder for io.grpc.ServiceDescriptor.
3. getFullName logic is moved inside ServiceDescriptor builder and will set ServiceDescriptor fullName field when build is triggered to create a new instance.
4. More refactoring and additional unit test to cover no java package in proto file scenario
5. Rename methods for retrieving fullName and packageNane in ServiceDescriptor and add shouldBuildFromBindableService test in GrpcProtoPackageTest and GrpcProtoPackageNoJavaPackageTest to test scenario where name is already prefixed with package.

### DIFF
--- a/grpc/server/src/main/java/io/helidon/grpc/server/BindableServiceImpl.java
+++ b/grpc/server/src/main/java/io/helidon/grpc/server/BindableServiceImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -73,7 +73,8 @@ class BindableServiceImpl implements BindableService {
     @SuppressWarnings("unchecked")
     @Override
     public ServerServiceDefinition bindService() {
-        io.grpc.ServiceDescriptor.Builder serviceDescriptorBuilder = io.grpc.ServiceDescriptor.newBuilder(descriptor.name());
+        io.grpc.ServiceDescriptor.Builder serviceDescriptorBuilder =
+                io.grpc.ServiceDescriptor.newBuilder(descriptor.fullName());
         if (descriptor.proto() != null) {
             serviceDescriptorBuilder.setSchemaDescriptor((ProtoFileDescriptorSupplier) descriptor::proto);
         }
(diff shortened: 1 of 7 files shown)


aseovic (approved):
LGTM

Issue #5174, opened 2022-10-13
No AccessLog output with http response 413


klustria:
## Environment Details
* Helidon Version: 2.5.x
* Helidon SE

----------

## Problem Description
AcccessLog does not output anything when `HTTP/1.1 413 Request Entity Too Large` is returned due to maxPayLoadSize violation. 

## Steps to reproduce
1. Use the helidon-quickstart-se app
2. Register AccessLogSupport:
```
routingBuilder.register(AccessLogSupport.create(config.get("server.access-log")));
```
3. Set access-log configuration in application.yaml like below:
```
server:
  port: 8080
  host: 0.0.0.0
  access-log: 
    enabled: true
    format: "%h %l %u %t %r %s %b %{Referer}i"
```
5. Add maxPayloadSize method in the WebServer builder and it to a a very low value to force error 413 ` Request Entity Too Large`
```
 WebServer server = WebServer.builder(createRouting(config))
                .config(config.get("server"))
                .addMediaSupport(JsonpSupport.create())
                . maxPayloadSize(10L)
                .build();
``` 
6. Build and run the app
7. Issue a PUT request: curl -X PUT -H "Content-Type: application/json" -d '{"greeting" : "Hola"}' http://localhost:8080/greet/greeting
8. `HTTP/1.1 413 Request Entity Too Large` should be received from that request, but no access log will be shown from the console.

PR #5295, opened 2022-11-01
Log an entry in warning level for a 400 or 413 response

spericas:
Very simple enhancement for issue #5174 that logs an entry in warning level for a 400 or 413 response returned by the ForwardingHandler. Note that the access log is not reached if an error is reported at this stage, so these log entries should help diagnose problems.

This seems a lot simpler than creating a new log file or attempting to use the access log (which may not be enabled) just for this purpose. The log entry that includes the error is very visible as shown below:

```
2022.11.01 11:45:03 INFO io.helidon.common.HelidonFeatures Thread[features-thread,5,main]: Helidon SE 3.0.3-SNAPSHOT features: [Config, Tracing, Web Client, WebServer]
2022.11.01 11:45:03 INFO io.helidon.webserver.NettyWebServer Thread[nioEventLoopGroup-2-1,10,main]: Channel '@default' started: [id: 0xa063c0ce, L:/127.0.0.1:64280]
2022.11.01 11:45:03 INFO io.helidon.webserver.MaxPayloadSizeTest Thread[main,5,main]: Started server at: https://localhost:64280
2022.11.01 11:45:04 WARNING io.helidon.webserver.ForwardingHandler Thread[nioEventLoopGroup-3-1,10,main]: 413: Payload is too large
java.lang.Error: 413: Payload is too large
	at io.helidon.webserver/io.helidon.webserver.ForwardingHandler.send413PayloadTooLarge(ForwardingHandler.java:551)
	at io.helidon.webserver/io.helidon.webserver.ForwardingHandler.channelReadHttpRequest(ForwardingHandler.java:379)
	at io.helidon.webserver/io.helidon.webserver.ForwardingHandler.lambda$channelRead0$3(ForwardingHandler.java:156)
	at io.helidon.common.context/io.helidon.common.context.Contexts.runInContext(Contexts.java:137)
        ...
```

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
@@ -524,7 +524,9 @@ public class ForwardingHandler extends SimpleChannelInboundHandler<Object> {
         ctx.writeAndFlush(response)
                 .addListener(future -> ctx.close());
 
-        failPublisher(new Error("400: Bad request"));
+        Error error = new Error("400: Bad request");
+        LOGGER.log(Level.WARNING, error, error::getMessage);
+        failPublisher(error);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


klustria (approved):
I think this is fine compromise, hence approving.

PR #5298, opened 2022-11-01
Log an entry in warning level for a 400 or 413 response 

spericas:
Log an entry in warning level for a 400 or 413 response returned by the ForwardingHandler. Note that the access log is not reached if an error is reported at this stage, so these log entries should help diagnose problems. (#5295)

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ForwardingHandler.java
@@ -564,7 +564,7 @@ public class ForwardingHandler extends SimpleChannelInboundHandler<Object> {
      * @param request Netty HTTP request
      * @param t associated throwable
      */
-    private void  send400BadRequest(ChannelHandlerContext ctx, HttpRequest request, Throwable t, String message) {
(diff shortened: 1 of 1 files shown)


klustria (commented):
LGTM

tomas-langer (commented):
I am in general against logging warnings based on user requests that processed correctly (e.g. in this case, if you send invalid request, it is not server failure, same for request entity too large).
I think this opens space for logs that are very simply initiated by a user that has malicious intents. That is the reason all such logs (I think) should be in `FINE` level or lower.
If you need to troubleshoot this kind of problems, you should enable that log level. Maybe we should just use a known logger for this, so you could enable it easily "on purpose" if you have safe clients (and 400 or 413 is a problem you need to solve)

spericas:
@tomas-langer The problem is that dev or admins won't set that log level in production and simply won't know at this time that they are rejecting requests (as it was reported by the user in the issue). Perhaps we can separate 400 from 413, but not informing the admin of these conditions that don't make into the access log seems bad. As for the malicious intents, those can (and likely will) exist regardless of whether we log warnings or not, and worse if we don't log, they simply won't know anything about it --again as reported by the user. If the concern is the number of lines we log now, we can reduce that to one line.

tomas-langer (approved):

Issue #5281, opened 2022-10-28
Create Backport Issues Action will not create issues if subject has double quotes in it


klustria:
## Problem Description
If issue subject has double quotes like
```
Grpc component does not handle "package" directive in proto files
```
running workflow for [Create Backport issues](https://github.com/helidon-io/helidon/actions/workflows/create-backport-issues.yml) will complete but not able to create the issues.
```
Run etc/scripts/actions/create-backport-issues.sh $GITHUB_REPOSITORY 4567 2.x
Created issue for version 3.x, issue number: null, url: null
Created issue for version 4.x, issue number: null, url: null
```

## Steps to reproduce
1. An issue must exist with subject that has double quotes in it
2.  Run [Create Backport issues](https://github.com/helidon-io/helidon/actions/workflows/create-backport-issues.yml) 
3. Check the details and observe that `issue number` and `url` will be null.

klustria:
Root cause analysis:

After running `etc/scripts/actions/create-backport-issues.sh` from local machine, I found that it is returning:  
```
"message": "Problems parsing JSON",
```
and the JSON payload is:
```
"{"title":"[4.x] Grpc component does not handle "package" directive in proto files.","body":"Backport of #4567 for Helidon 4.x","assignees":["klustria"],"labels":["4.x","backport","bug","grpc","P2"]}"
```
As shown above, subject contains double quotes for the word package which will mess up the format of the title value.

klustria:
Goals to 
1. Replace double quotes `"` with  single quote `'` to avoid Github JSON parsing issue
2. Print out the Github API Server response if unable to parse the issue number. Also display the json payload that was sent so it can be inspected for problems if such issue occur. This will help in other unknown situations where the request is failing. Example output:
```
Encountered an error while attempting to create an issue: {
  "message": "Problems parsing JSON",
  "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue"
}
Json payload: {"title":"[3.x] Grpc component does not handle "package" directive in proto files.","body":"Backport of #4567 for Helidon 3.x","assignees":["klustria"],"labels":["3.x","backport","bug","grpc","P2"]}
```

klustria:
A potential fix in PR https://github.com/helidon-io/helidon/pull/5311 is in the works.

klustria:
[...]

PR #5311, opened 2022-11-03
Fix problem when double quotes exist in the subject title during Create backport issues action

klustria:
**Goals of the change:**
1. Replace double quotes `"` with single quote `'` in the issues subject title to avoid Github JSON parsing issue.
2. Print out the Github API Server response if unable to parse the issue number. Also display the json payload that was sent so it can be inspected for problems if such issue occur. This will help in other unknown situations where the request is failing. Example output: 
   ```
   Encountered an error while attempting to create an issue: {
     "message": "Problems parsing JSON",
     "documentation_url": "https://docs.github.com/rest/reference/issues#create-an-issue"
   }
   Json payload: {"title":"[3.x] Grpc component does not handle "package" directive in proto files.","body":"Backport of #4567 for Helidon 3.x","assignees":["klustria"],"labels":["3.x","backport","bug","grpc","P2"]}
   ```

_Note: Potential fix was tested in local machine using a temporary github token_

### DIFF
--- a/etc/scripts/actions/create-backport-issues.sh
+++ b/etc/scripts/actions/create-backport-issues.sh
@@ -70,7 +70,7 @@ readonly ISSUE=$(curl -s -X GET \
   "$GET_ISSUE_URL")
 
 # Get issue information
-ISSUE_TITLE=$(echo "$ISSUE" | jq -r ".title")
+issue_title=$(echo "$ISSUE" | jq -r ".title")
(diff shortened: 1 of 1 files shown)


klustria:
Goal is to resolve issue https://github.com/helidon-io/helidon/issues/5281

tomas-langer (approved):

tomas-langer (approved):

PR #5341, opened 2022-11-06
OpenAPI U/I integration

### LINKED ISSUE #4883: Create simple wrapper around SmallRye OpenAPI U/I
## Environment Details
* JDK version:
* OS:

----------

## Problem Description
The OpenAPI U/I (now managed by SmallRye, analogous to the Swagger U/I) has evolved to use static content. This is very easy to configure in a Helidon app.

But, if the Helidon service also uses static content then the developer needs to make sure that the OpenAPI U/I content and her service's content are handled properly. 

Instead, we can create a small module which wraps the OpenAPI U/I in a way that does not rely on Helidon's static content configuration. This frees the developer to use the static content configuration for the service's static content and not have to worry about the U/I's static content.

Creating this as an integration might help emphasize that Helidon does not support the OpenAPI U/I although we do try to facilitate its use for our users.

tjquinno:
Resolves #4883 

SmallRye offers a mostly-static content solution for presenting essentially the Swagger U/I.

This PR creates two new integrations--`helidon-integrations-openapi-ui` (for SE apps) and `helidon-integrations-openapi-ui-cdi` (for MP apps)--which layer on the SmallRye code, customizing the main page dynamically with some Helidon-specific content (logo, title, links) and serving the U/I at  at `/openapi-ui` (by default).

SE developers add a build-time dependency and then use a builder (or simple factory methods) to create an `OpenApiUiSupport` instance and add it to their routing. They can use config for a few settings as well.

MP developers add a runtime dependency and use config as needed to tailor the behavior.

Changes:
* The two new integrations, one for SE and one for MP
* Very small changes to OpenAPI and to the MP QuickStart examples
* Docs


tjquinno:
Hmm. A bunch of messed up copyrights. I might just create a new PR. Stand by.

tjquinno:
Superseded by #5343

PR #5344, opened 2022-11-07
Update Microstream version to 07.01.00-MS-GA

hg-ms:
Update Microstream version from 05.00.02-MS-GA to 07.01.00-MS-GA


hg-ms:
The failure when running the javadoc-plugin seem to be caused by missing exports in the java module definitions of Microstream. This must be fixed in the Microstream libraries. I’ll withdraw the pull request until a Microstream fix is available.

PR #5355, opened 2022-11-08
Log simple message for a 400 or a 413 and more under FINE

spericas:
Log simple message for a 400 or a 413 and a more detailed message if log FINE level is enabled.

Follow up to PR #5295.


klustria (approved):
LGTM

Issue #5022, opened 2022-09-30
Add null check to MP Server.Builder.config()


tomas-langer:
Make sure we throw NPE early in `io.helidon.microprofile.server.Server.Builder.config(Config)` - both for SE config and MP config
Same in `MpConfigSources.create(Config)`

tomas-langer:
Related to issue reported by customer, where we get a NPE at runtime from internals with the following stack trace:

```
Exception in thread "main" java.lang.NullPointerException
        at io.helidon.config.mp.MpHelidonConfigSource.getValue(MpHelidonConfigSource.java:48)
        at org.eclipse.microprofile.config.spi.ConfigSource.getOrdinal(ConfigSource.java:152)
        at io.helidon.config.mp.MpConfigBuilder$OrdinalSource.findOrdinal(MpConfigBuilder.java:493)
        at io.helidon.config.mp.MpConfigBuilder$OrdinalSource.<init>(MpConfigBuilder.java:484)
        at io.helidon.config.mp.MpConfigBuilder.withSources(MpConfigBuilder.java:202)
        at io.helidon.microprofile.server.Server$Builder.config(Server.java:323)
```

PR #5363, opened 2022-11-10
Add null check to MP Server.Builder.config()

klustria:


config/config-mp/src/test/java/io/helidon/config/mp/MpConfigSourcesTest.java
```diff
@@ -99,6 +100,15 @@ void testHelidonLazy() {
         assertThat("exists called exactly once", lazy.exists.get(), is(1));
     }
 
+    @Test
+    void testMpConfigSourcesNullConfig() {
+        try {
```
tomas-langer, line 105: Please use `assertThrows(NullPointerException.class, () -> MpConfigSources.create((Config) null));` instead.
It returns the thrown exception, so you can assert the message as well.
Same for the second test in this PR.

(import: `import static org.junit.jupiter.api.Assertions.assertThrows;`)

tomas-langer (changes requested):

config/config-mp/src/test/java/io/helidon/config/mp/MpConfigSourcesTest.java
klustria: Done

tomas-langer (approved):

klustria:
This change was created specific for https://github.com/helidon-io/helidon/issues/5022.

PR #5372, opened 2022-11-10
Add null check to MP Server.Builder.config() (#5363)

klustria:
* Add null check to MP Server.Builder.config()

* Improve the NullPointerException assertion test


tomas-langer (approved):

klustria:
This fix is for https://github.com/helidon-io/helidon/issues/5370

Issue #4836, opened 2022-09-02
Upgrade GraalVM to v22


edbratt:
Tracking issue to coordinate work necessary for upgrading to GraalVM v22.

tomas-langer:
Started working on this issue.
I will use GraalVM 22.3.0 dev builds, as since this version we can access required APIs in published module (and we no longer need to use the internal `svm` module).
Until 22.3 is released, we cannot finish this task.

tomas-langer:
Fixed in 3.x and 4.x branches. I am not working on backport to 2.x. Is this something we should do as well?

tomas-langer:
Closing this issue. If there is a specific problem, or this needs to be done for 2.x, please kindly create a new issue.

PR #5378, opened 2022-11-11
4836 graalvm 22 4.x

tomas-langer:
Upgrade GraalVM to 22
Fix native image build for Níma and SE 
Helidon MP native image still does not work

Related to #4836


arjav-desai (approved):
I have built helidon locally with this and ran native-image on quickstart-se archetype which works fine.

LGTM

PR #5441, opened 2022-11-15
Fix handling of optional whitespace at the beginning of headers.

tomas-langer:


common/buffers/src/test/java/io/helidon/common/buffers/LazyStringTest.java
```diff
@@ -0,0 +1,74 @@
...
+        public String toString() {
+            StringBuilder result = new StringBuilder();
+
+            for (char c : string().toString().toCharArray()) {
+                switch (c) {
+                case '\t' -> result.append("\\t");
+                case '\r' -> result.append("\\r");
+                default -> result.append(c);
+                }
+            }
```
danielkec, line 69: Formatting seems little off
```suggestion
            for (char c : string().toString().toCharArray()) {
                switch (c) {
                    case '\t' -> result.append("\\t");
                    case '\r' -> result.append("\\r");
                    default -> result.append(c);
                }
            }
```

danielkec (changes requested):

common/buffers/src/test/java/io/helidon/common/buffers/LazyStringTest.java
tomas-langer: This is the formatting I get with the Helidon format for Idea

danielkec (approved):

spericas (approved):
LGTM. Will test it more.

PR #5468, opened 2022-11-17
Allow @CrossOrigin to appear at the class level (#5260)

### LINKED ISSUE #5260: Allow `@CrossOrigin` to appear at the class level
## Environment Details
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
Currently, the `@CrossOrigin` annotation applies only to methods.

If the developer want the same CORS behavior for all resources in a given resource class, it would be convenient to annotate the class once rather than apply the same method-level annotation to multiple `@Options` methods.

Captain1653:
Fixes #5260

[Mini spec](https://github.com/helidon-io/helidon/issues/5260#issuecomment-1307550103)

Done:
* Main logic
* Some tests (to be discussed below)

Not Done:
* `examples/microprofile/cors`  (I need some feedback about this draft)

Some questions:
1. How can I check logging? I can't see it in my tests. Test `CrossOriginTest#testPreFlightMethodLevelAnnotationTakesPrecedenceOverClassLevel` should log. Did I miss something?

2. Should I add some additional tests? For example `testPreFlightConfigFileTakesPrecedenceOverMethodLevelAnnotation`? I've done it only for class-level annotation. I'm not sure.

3. I can't run tests from my IDE. I get error
Screenshot 2022-11-17 at 19 52 16
But I can run tests via maven. Should I do some specific actions with my IDE? It uses Java 17.

4. Error messages can be changed. Sorry for my English :) .

@tjquinno Any advice will be appreciated. Don't hesitate to write comments. It's my first important PR.

### DIFF
--- a/microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
+++ b/microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
@@ -19,10 +19,14 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
 import io.helidon.config.Config;
 import io.helidon.config.mp.MpConfig;
@@ -32,6 +36,8 @@ import jakarta.annotation.Priority;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.Initialized;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.spi.AnnotatedMethod;
+import jakarta.enterprise.inject.spi.AnnotatedType;
 import jakarta.enterprise.inject.spi.Extension;
 import jakarta.enterprise.inject.spi.ProcessManagedBean;
 import jakarta.ws.rs.OPTIONS;
@@ -49,6 +55,8 @@ import static jakarta.interceptor.Interceptor.Priority.LIBRARY_BEFORE;
  */
 public class CorsCdiExtension implements Extension {
 
+    private static final Logger LOGGER = Logger.getLogger(CorsCdiExtension.class.getName());
+
     /**
      * Key used for retrieving CORS-related configuration from MP configuration.
      */
@@ -60,9 +68,29 @@ public class CorsCdiExtension implements Extension {
 
     private final Set<Method> methodsWithCrossOriginIncorrectlyUsed = new HashSet<>();
     private final Map<Method, CrossOriginConfig> corsConfigs = new HashMap<>();
+    private Class<?> clazzWithClassLevelAnnotationCrossOriginWithoutOptionMethods = null;
 
     void processManagedBean(@Observes ProcessManagedBean<?> pmb) {
-        pmb.getAnnotatedBeanClass().getMethods().forEach(am -> {
+        AnnotatedType<?> annotatedBeanClass = pmb.getAnnotatedBeanClass();
+
+        if (annotatedBeanClass.isAnnotationPresent(CrossOrigin.class)) {
+            List<AnnotatedMethod<?>> optionMethods = getOptionMethods(pmb);
+            Class<?> resourceJavaClass = annotatedBeanClass.getJavaClass();
+
+            if (optionMethods.isEmpty()) {
+                clazzWithClassLevelAnnotationCrossOriginWithoutOptionMethods = resourceJavaClass;
+                return;
+            } else if (isEveryMethodHasOwnCrossOriginAnnotation(optionMethods)) {
+                LOGGER.log(Level.FINE, String.format("%s has unnecessary class level annotation @%s. "
+                                + "Every @%s method has own @%s. Method-level annotation takes precedence.",
+                        resourceJavaClass.getSimpleName(),
+                        CrossOrigin.class.getSimpleName(),
+                        OPTIONS.class.getSimpleName(),
+                        CrossOrigin.class.getSimpleName()));
+            }
+        }
+
+        annotatedBeanClass.getMethods().forEach(am -> {
             Method method = am.getJavaMember();
             if (am.isAnnotationPresent(CrossOrigin.class) && !am.isAnnotationPresent(OPTIONS.class)) {
                 methodsWithCrossOriginIncorrectlyUsed.add(method);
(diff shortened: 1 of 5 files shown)


tjquinno:
Hi @Captain1653 

I will take a closer look at the changes soon.

About your questions...

1. Logging - Do you want the test to check the logging automatically, or are you just wanting to look at the logging yourself manually?
   1. For automatic, take a look at `webserver/cors` `MissingConfigTest`. That test contains some code to capture log output so the test can look at it. The code in that test only looks for a single, hard-coded string but it could easily be generalized by, for example, adding a parameter to `checkForLogMessage` for the string to be searched for and using the parameter in the `assertThat` invocation.
   2. For manual, look for `target/surefire-reports. There will be one or two output files for each test. Look for the ones with names ending in `-output.txt` which will contain logging messages. You might also need to change or create `test/resources/logging.properties` in order to enable whatever logging you want to see or check for, depending on what logging level the messages have.
2. Additional tests - I do not remember if we check that configuration overrides annotations. It would be good to have a test like that if we do not already.
3. Testing from the IDE - I see the same error. I will try some things to resolve that and let you know if I find a solution.

microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
```diff
@@ -87,6 +115,14 @@ void ready(@Observes @Priority(LIBRARY_BEFORE + 10) @Initialized(ApplicationScop
...
                             methodsWithCrossOriginIncorrectlyUsed));
         }
+        if (clazzWithClassLevelAnnotationCrossOriginWithoutOptionMethods != null) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "%s has class-level annotation @%s and should have at least one @%s method",
+                            clazzWithClassLevelAnnotationCrossOriginWithoutOptionMethods.getSimpleName(),
+                            CrossOrigin.class.getSimpleName(),
+                            OPTIONS.class.getSimpleName()));
+        }
```
tjquinno, line 125: See earlier comment. I think this checking, if it happens, needs to be inside (or invoked from) `processManagedBean`.

microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
```diff
@@ -60,9 +68,29 @@ public class CorsCdiExtension implements Extension {
...
+                return;
+            } else if (isEveryMethodHasOwnCrossOriginAnnotation(optionMethods)) {
+                LOGGER.log(Level.FINE, String.format("%s has unnecessary class level annotation @%s. "
+                                + "Every @%s method has own @%s. Method-level annotation takes precedence.",
+                        resourceJavaClass.getSimpleName(),
+                        CrossOrigin.class.getSimpleName(),
+                        OPTIONS.class.getSimpleName(),
+                        CrossOrigin.class.getSimpleName()));
+            }
+        }
```
tjquinno, line 91: A few thoughts...

1. IIUC, this block's purpose is largely (although not exclusively) to log a message if all `@OPTIONS` methods have a method-level `@CrossOrigin` annotation _and_ the class also has a class-level annotation which will therefore be ignored.
   
   For readability and efficiency please move this checking into a separate method and invoke that method from here only if `FINE` is loggable. That way only users who enable `FINE`-level logging pay the performance cost of the computation required to do the checking.
   
   That also means removing the "return-fast" logic if there are no `@OPTIONS` methods but I think that's OK.

2. I don't think it works to have the instance field `clazzWithClassLevelAnnotationCrossOriginWithoutOptionMethods`. CDI invokes `processManagedBean` multiple times, once for each managed bean it processes. There could be multiple of those which have a class-level `@CrossOrigin` annotation and no `@OPTIONS` methods with that annotation. That means the instance field could be overwritten multiple times before the `ready` method--which CDI invokes only once, after all beans have been processed--refers to that field, and the `ready` method would see only the last value assigned. 
   
   That checking would also need to be done during a single invocation of `processManagedBean` so _each_ offending class could  be reported.

3. At new line 98 below, we know exactly what the `CrossOriginConfig` should be for the method being analyzed...it's either the method-level info from the corresponding `@OPTIONS` method or it's the class-level info (if any). I'd suggest revising this code slightly so we `put` whichever `CrossOriginConfig` applies--method level or class level--into the map's entry for this method. 
   `crossOriginConfigFromAnnotationOnAssociatedMethod` could go back to how it was originally (and go back to doing what it's name says!), and we wouldn't need the new `getCrossOriginAnnotation` method. 
   
   I'm thinking something like this: At the top of the method:
   ```
   Optional<CrossOriginConfig> classLevelCrossOriginConfig =  Optional.ofNullable(pmb.getAnnotatedBeanClass()
                                                                                       .getAnnotation(CrossOrigin.class))
                   .map(CorsCdiExtension::annotationToConfig);
   ```
   
   and then new lines 98-100 become
   
   ```
                   crossOriginConfigFromAnnotationOnAssociatedMethod(method)
                           .or(() -> classLevelCrossOriginConfig)
                           .ifPresent(crossOriginConfig -> corsConfigs.put(method,
                                                                           crossOriginConfig));
   ```
   
   This reduces the amount of other code elsewhere in the class that has to change and also associates the CORS information with the method as soon as we know it.

microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
tjquinno, line 125: See earlier comment. 

I think this checking, if it happens, needs to be inside (or invoked from) processManagedBean.

tjquinno (changes requested):
Sorry for the delay in providing this feedback.

I've requested some changes and written (lengthy) explanations for them. I have not looked closely at the tests yet but will do so tomorrow.

microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
```diff
@@ -105,6 +141,19 @@ CorsSupportMp corsSupportMp() {
         return corsSupportMp;
     }
 
+    private boolean isEveryMethodHasOwnCrossOriginAnnotation(List<AnnotatedMethod<?>> methods) {
+        return methods
+                .stream()
+                .allMatch(am -> am.isAnnotationPresent(CrossOrigin.class));
+    }
```
tjquinno, line 148: Should this look only at `@OPTIONS` methods, not all methods?

microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
Captain1653: Thank you for your explanations. I'll fix it on Saturday or Sunday

PR #5472, opened 2022-11-18
NullPointerException when there is an illegal character in the request (#5470)

klustria:
Move requestEntityAnalyzed CompletableFuture instantiation way before it gets called channelReadHttpContent as a LastHttpContent


klustria:
Resolves https://github.com/helidon-io/helidon/issues/5474

spericas (approved):

PR #5521, opened 2022-11-23
[3.x] Jwt scope handling extended over array support

### LINKED ISSUE #5365: OIDC provider with OAM as OAuth server - Helidon treats `scopes` as string but is actually returned as Json array
## Environment Details
* Helidon Version: 3.0.1
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description

We are using OIDC Provider of Helidon(3.0.1) with OAM (12.2.1.4) as OAuth server. OAM returns token that contain 'scopes' as array, where as Helidon is looking for string. We facing issues with Helidon's way of getting scopes from token in `JWTUtils.toScopes(JsonObject json)`, as it always considers value as string.
Changing the implementation as below would resolve the issue:
```
return (json.get("scope") instanceof JsonArray) ? 
    		      getStrings(json, "scope") : 
    		          getString(json, "scope").<List<String>>map(it -> Arrays.asList(it.split(" ")));
```

Verdent:
Fixes #5365

### DIFF
--- a/security/jwt/src/main/java/io/helidon/security/jwt/JwtUtil.java
+++ b/security/jwt/src/main/java/io/helidon/security/jwt/JwtUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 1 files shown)


tomas-langer (approved):
LGTM

PR #5543, opened 2022-11-28
Níma: Static content handling rework

tomas-langer:
Static content now uses cache of computed values (not cache of bytes).
Support for in-memory caching of bytes added as well, must be explicitly configured using configuration.
For classpath, each file must be listed.
For file system, directories can be listed as well (only first level depths is done).

The caching uses the same path that is requested by a user

### DIFF
--- a/nima/webserver/static-content/src/main/java/io/helidon/nima/webserver/staticcontent/CachedHandlerJar.java
+++ b/nima/webserver/static-content/src/main/java/io/helidon/nima/webserver/staticcontent/CachedHandlerJar.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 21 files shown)


nima/webserver/static-content/src/main/java/io/helidon/nima/webserver/staticcontent/CachedHandlerJar.java
```diff
@@ -0,0 +1,79 @@
...
+                          ServerRequest request,
+                          ServerResponse response,
+                          String requestedResource) throws IOException {
+
+        // check if file still exists (the tmp may have been removed, file may have been removed
+        // there is still a race change, but we do not want to keep cached records for invalid files
+        if (!Files.exists(path)) {
+            cache.remove(requestedResource);
+            return false;
+        }
```
danielkec, line 56: would file watcher help here to offload the check from handler?

danielkec (approved):

nima/webserver/static-content/src/main/java/io/helidon/nima/webserver/staticcontent/CachedHandlerJar.java
tomas-langer: The check needs to touch the file system, nevertheless the only case the file may get removed is when somebody (or the OS) cleans up the temporary directory we use to extract files. I think file watcher is not a good match here

Verdent (approved):
LGTM

Issue #5159, opened 2022-10-11
Yaml config profiles support


botsteve:
## Environment Details
* Helidon MP Version: 3.0.1

I have a question regarding the profile support for yaml config sources. I know there is profile support for microprofile-config-${profile}.properties files but does this support extends also to application-${profile}.yaml files? 

I know that something like this can be done using mp-meta-config file and specify the profile property and an higher ordinal value for application-${profile}  files, thus providing the same functionality as mp.config.profile system property. 

I think that there should be a similar profile support for yaml files using the same system property as for the properties files.

If indeed there is no support for this and it is wanted, i can further pursue this.

romain-grecourt:
See https://helidon.io/docs/v3/#/se/config/config-profiles

klustria:
Can you please review MP meta-config feature: https://helidon.io/docs/v3/#/mp/config/advanced-configuration#_creating_microprofile_config_sources_from_meta_config and see if that will meet your needs?

botsteve:
@klustria yea, this can work as I mentioned above, it is not a problem of making it work using meta-config but more of having a consistent support across different config extensions(.properties, .yaml etc...). Also I think that by having the same system property "mp.config.profile" working using different types of configs will make use of the convention over configuration model, which is desired.

klustria:
The problem is that the config source implementation is MP spec based, and that does not have profile anywhere in the SPI. So we do not know which profile to use. The implementation of the [PR](https://github.com/helidon-io/helidon/pull/5163) you created is depending on a system property, which is not the only way to define a profile (it can be defined in another config source). So unless the SPI contains information about which profile to use, we cannot modify it. There might be a way to do this but it may be a little complex than what you PR presents but this may need to be looked at further by the team. In the meantime, your only workaround is to use meta-config.

botsteve:
yea, I agree, the problem is the ConfigSourceProvider contract from the MP spec which doesn't provide a profile support, and indeed this doesn't cover the ConfigSource usecase, but I think the general usecase is to cover the system property use.

botsteve:
@klustria updated the PR to use reflection to set the profiles property since the MP spec doesn't cover this, this should cover the configuration from other sources besides the system as well.

tomas-langer:
Hello @botsteve, this is a limitation of MicroProfile specification and we cannot do much about it. 
[...]

PR #5565, opened 2022-11-29
Provide MP config profile support for application.yaml

klustria:
1. Create a new interface MpConfigSourceProviderWithProfile that extends ConfigSourceProvider and adds profile support.

### DIFF
--- a/config/config-mp/src/main/java/io/helidon/config/mp/spi/MpConfigSourceProviderWithProfile.java
+++ b/config/config-mp/src/main/java/io/helidon/config/mp/spi/MpConfigSourceProviderWithProfile.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.config.mp.spi;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
+
+/**
+ * Java Service loader interface for MP ConfigSource Providers that adds configuration profile support.
+ */
+public interface MpConfigSourceProviderWithProfile extends ConfigSourceProvider {
+    /**
+     * Returns a list of configuration sources.
+     *
+     * @param classLoader classloader where resource will be retrieved from (will use the
+     *                    current thread's classloader if value is {@code null})
+     * @param profile configuration profile to use
+     *
+     * @return list of config sources
+     */
+    Iterable<ConfigSource> getConfigSources(ClassLoader classLoader, String profile);
+}
--- a/config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
+++ b/config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
@@ -21,9 +21,11 @@ import java.net.URL;
 import java.util.Enumeration;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 
 import io.helidon.config.ConfigException;
-import io.helidon.config.mp.spi.MpConfigSourceProviderWithProfile;
+import io.helidon.config.mp.spi.MpConfigSourceProvider;
 
 import org.eclipse.microprofile.config.spi.ConfigSource;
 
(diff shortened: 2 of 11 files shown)


config/config-mp/src/main/java/io/helidon/config/mp/spi/MpConfigSourceProviderWithProfile.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Returns a list of configuration sources.
+     *
+     * @param classLoader classloader where resource will be retrieved from (will use the
+     *                    current thread's classloader if value is {@code null})
+     * @param profile configuration profile to use
+     *
+     * @return list of config sources
+     */
+    Iterable<ConfigSource> getConfigSources(ClassLoader classLoader, String profile);
```
tomas-langer, line 35: As with any other API in Helidon, the profile MUST NOT be null.

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlMpConfigSource.java
```diff
@@ -196,10 +196,25 @@ public static List<ConfigSource> classPath(String resource) {
...
+    /**
+     * Create from YAML file(s) on classpath from specified classloader with profile support.
+     *
+     * @param resource resource name to locate on classpath (looks for all instances)
+     * @param profile configuration profile to use
+     * @param classLoader classloader where resource will be retrieved from (will use the
+     *                    current thread's classloader if value is {@code null})
+     * @return list of config sources discovered (may be zero length)
+     */
+    public static List<ConfigSource> classPath(String resource, String profile, ClassLoader classLoader) {
```
tomas-langer, line 211: Once again - we cannot have a public API method that accepts nulls

config/config-mp/src/main/java/io/helidon/config/mp/spi/MpConfigSourceProviderWithProfile.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.helidon.config.mp.spi;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
+
+/**
+ * Java Service loader interface for MP ConfigSource Providers that adds configuration profile support.
+ */
+public interface MpConfigSourceProviderWithProfile extends ConfigSourceProvider {
```
tomas-langer, line 25: I think the name can be just `MpConfigSourceProvider`, as it is sufficiently distinct from the MP SPI.
Also I prefer to have the word `Provider` as the last (to make it clear this is a service provider interface).

tomas-langer (changes requested):
A few changes on API (mostly related to null parameters). Otherwise this looks good (and allows extensibility by our customers).

config/config-mp/src/main/java/io/helidon/config/mp/spi/MpConfigSourceProviderWithProfile.java
klustria: implementation now validates profile  must not be null

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlMpConfigSource.java
klustria: implementation now validates profile and classloader must not be null

config/config-mp/src/main/java/io/helidon/config/mp/spi/MpConfigSourceProviderWithProfile.java
klustria: good idea and SPI is renamed as suggested

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlMpConfigSource.java
```diff
@@ -192,14 +192,26 @@ public static List<ConfigSource> classPath(String resource) {
      * Create from YAML file(s) on classpath with profile support.
      *
      * @param resource resource name to locate on classpath (looks for all instances)
-     * @param profile name of the profile to use
+     * @param profile configuration profile to use, must not be null
      * @return list of config sources discovered (may be zero length)
      */
     public static List<ConfigSource> classPath(String resource, String profile) {
+        return classPath(resource, profile, Thread.currentThread().getContextClassLoader());
```
tomas-langer, line 199: if I remember correctly, `getContextClassLoader()` may return `null`
It would be better to have a private method that accepts class loader and checks for null (and does something reasonable when it is null, such as using classloader of the current class)

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
```diff
@@ -21,30 +21,47 @@
...
+        return getConfigSources(classLoader, Optional.empty());
+    }
+
+    @Override
+    public Iterable<ConfigSource> getConfigSources(ClassLoader classLoader, String profile) {
+        Objects.requireNonNull(profile, "Profile must not be null");
+        return getConfigSources(classLoader, Optional.of(profile));
+    }
 
+    private Iterable<ConfigSource> getConfigSources(ClassLoader classLoader, Optional<String> profile) {
```
tomas-langer, line 50: For private methods, you can use `null`, the restriction is for public API.
Also optional parameters are not good.

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
```diff
@@ -21,30 +21,47 @@
...
         List<ConfigSource> result = new LinkedList<>();
-        while (resources.hasMoreElements()) {
-            URL url = resources.nextElement();
-            result.add(YamlMpConfigSource.create(url));
+        if (profile.isEmpty()) {
```
tomas-langer, line 52: When using a profile, we need to load the source without a profile, and then also load all profile sources and combine them so that a config source with profile overrides config source without a profile.

tomas-langer (changes requested):

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
klustria: This is the behavior we have now in YamlMpConfigSource.classPath() and new unit tests were added to validate when a config property is overridden (config property is added in config profile files) and not overridden (config property is only on the default config and not added in config profile files).

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
klustria: Will take note of avoiding optional from hereon. Private methods were refactored to allow for null parameter.

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlMpConfigSource.java
klustria: Added a private method to allow null classloader.

tomas-langer (approved):

klustria:
This will resolve enhancement request in https://github.com/helidon-io/helidon/issues/5159.

PR #5568, opened 2022-11-30
Add OpenAPI U/I integration [3.x]

tjquinno:
Resolves https://github.com/helidon-io/helidon/issues/4883

SmallRye offers a mostly-static content solution for presenting essentially the Swagger U/I.

This PR creates a new integration--`helidon-integrations-openapi-ui`--which layers on the SmallRye code, customizing the main page dynamically with some Helidon-specific content (logo, title, links) and serving the U/I at at `/openapi/ui` (by default) and, using content negotiation, also at `/openapi`.

SE developers add a build-time dependency and then create an `OpenApiUi.Builder` instance and add it to the `OpenAPISupport.Builder` if they want to customize the U/I behavior. Or, they can just let `OpenAPISupport.Builder` automatically create a U/I using default settings. They can use config for a few settings as well.

MP developers add a runtime dependency and use config as needed to tailor the behavior.

## Changes to `helidon-openapi`
* Defines some new U/I-related interfaces (`OpenApiUiFactory` and `OpenApiUi`) and contains no-op implementations. 
* `OpenAPISupport`:
   * Uses service loading to obtain the correct factory and get the correct `OpenApiUi.Builder` impl
   * Invokes the `OpenApiUi` `update` method so it can add its own routing. (none for the no-op impl) The developer's code  does not explicitly register the U/I service.
   * delegates to the `OpenApiUi` instance upon receiving a request at `/openapi` so it can handle requests for text/* content.
* Some new tests

## New `helidon-integrations-openapi-ui`
* Provides its `OpenApiUiFactory` impl and contains a full-featured `OpenApiUi` impl which:
   * Redirects HTML requests to `/openapi` and `/openapi/ui` to return a dynamically-computed `index.html` (from SmallRye) which contains JavaScript which retrieves the OpenAPI document and then runs the browser U/I)
   * Adds routing for static content, some from Helidon, some from SmallRye, used by the U/I.
* Some Helidon-specific static content for the U/I page.
* Tests

## MP QuickStart examples
* Added an MP OpenAPI annotation to add some information so the U/I is nicer if someone were to add the Helidon OpenAPI U/I dependency to the MP QuickStart pom manually.

## Docs
* Extensive additions to describe adding the U/I dependency and using the builder or config to tailor the behavior.
* Removed `se/openapi.adoc` which was moved to `se/openapi/openapi.adoc` in a previous PR but I accidentally left the old file in place.
* Fixed an unrelated typo in the OpenAPI generator doc.


docs/config/io_helidon_openapi_OpenApiUi.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+
+ifndef::rootdir[:rootdir: {docdir}/..]
+:description: Configuration of io.helidon.openapi.OpenApiUi
+:keywords: helidon, config, io.helidon.openapi.OpenApiUi
+:basic-table-intro: The table below lists the configuration keys that configure io.helidon.openapi.OpenApiUi
+include::{rootdir}/includes/attributes.adoc[]
+
+= OpenApiUi (openapi) Configuration
```
ljamen, line 25: OpenAPI UI

docs/config/io_helidon_openapi_OpenApiUi.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+[source,text]
+.Config key
+----
+ui
+----
+
+
+
+== Configuration options
```
ljamen, line 41: Configuration Options (if this were being used as a standalone doc)

docs/includes/openapi/openapi-ui.adoc
```diff
@@ -0,0 +1,200 @@
...
+- <<Additional Information, Additional Information>>
+
+// end::intro[]
+
+// tag::overview[]
+== Overview
+
+SmallRye offers an link:{smallrye-openapi-ui-base-url}[OpenAPI user interface component] which displays a web page based on your application's OpenAPI document.
+Through that U/I, users can invoke the operations declared in the document.
+While not generally suitable for end-users, the OpenAPI U/I can be useful for demonstrating and "test driving" your service's endpoints.
```
ljamen, line 49: OpenAPI UI  --- this should be a global change to the docs. UI is the preferred abbreviation of user interface (not U/I).

docs/includes/openapi/openapi-ui.adoc
```diff
@@ -0,0 +1,200 @@
...
+ifdef::se-flavor[your configuration file, such as `application.yaml`.]
+
+include::{rootdir}/config/io_helidon_openapi_OpenApiUi.adoc[tag=config,leveloffset=+1]
+The default U/I `web-context` value is the web context for your `OpenAPISupport` service with the added suffix `/ui`. If you use the default web context for both `OpenAPISupport` and the U/I, the U/I responds at `/openapi/ui`.
+
+// end::config-intro[]
+
+// tag::config-details[]
+
+You can use configuration to affect the U/I path in two ways.
```
ljamen, line 134: in two ways:

docs/se/openapi/openapi-ui.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+///////////////////////////////////////////////////////////////////////////////
+
+= OpenAPI U/I
```
ljamen, line 19: UI

ljamen (commented):
I didn't call them out individually, but perhaps we should stick to UI (vs U/I) throughout the docs? Is there a way to find/replace? The only other issue I had was with the logo.svg file. It wasn't displaying for me -- or at least it didn't look like it was?

docs/config/io_helidon_openapi_OpenApiUi.adoc
tjquinno: I have globally replaced "U/I" with "UI". 

There are some places where interface or class names (such as `OpenApiUi`) will remain but such cases should already be enclosed in single tick marks. 

This particular file is generated by a tool. Probably the tool should enclose the class/interface name in single ticks as it generates the file.

docs/config/io_helidon_openapi_OpenApiUi.adoc
tjquinno: This is in a generated file. The doc generator tool should be changed to emit the correct case.

docs/includes/openapi/openapi-ui.adoc
tjquinno: Done.

docs/se/openapi/openapi-ui.adoc
tjquinno: Done

barchetta (approved):

PR #5585, opened 2022-12-02
Provide MP config profile support for application.yaml (#5565)

klustria:
1. Create a new interface MpConfigSourceProvider that extends ConfigSourceProvider and adds profile support.

### DIFF
--- a/config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
+++ b/config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2021, 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


klustria:
Resolves backport request https://github.com/helidon-io/helidon/issues/5582

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
```diff
@@ -21,30 +21,46 @@
...
+        Objects.requireNonNull(profile, "Profile must not be null");
+        return configSources(classLoader, profile);
+    }
 
+    private Iterable<ConfigSource> configSources(ClassLoader classLoader, String profile) {
         List<ConfigSource> result = new LinkedList<>();
-        while (resources.hasMoreElements()) {
-            URL url = resources.nextElement();
-            result.add(YamlMpConfigSource.create(url));
+        if (profile == null) {
```
botsteve, line 51: Why not reuse YamlMpConfigSource.classPath(String resource) and write more concise?

maybe something like this? 

return Objects.isNull(profile)? YamlMpConfigSource.classPath(RESOURCE_NAME) : YamlMpConfigSource.classPath(RESOURCE_NAME, profile, classLoader)

klustria: Because `YamlMpConfigSource.classPath(RESOURCE_NAME)` does not accept classLoader as an argument.

tomas-langer (approved):

config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlConfigSourceProvider.java
botsteve: yea, but why do you need to pass it? it already uses the same Thread.currentThread().getContextClassLoader() as classPathConfigSources(resource, profile, ...);

klustria: Because getConfigSources() which was overriden from `io.helidon.config.mp.spi.MpConfigSourceProvider` and in turn inherited from `org.eclipse.microprofile.config.spi.ConfigSourceProvider` has classLoader in its API. It is possible that the implementation can be used in some other scenario where the classPath will not be `Thread.currentThread().getContextClassLoader()`.

PR #5586, opened 2022-12-02
Provide MP config profile support for application.yaml (#5565)

klustria:
1. Create a new interface MpConfigSourceProvider that extends ConfigSourceProvider and adds profile support.


klustria:
Will resolve backport request https://github.com/helidon-io/helidon/issues/5583

tomas-langer (approved):

Issue #5440, opened 2022-11-15
4.x: Pico and builder follow up


tomas-langer:
Follow up issues for builders and Pico API:

- [x] #5447
- [x] Current `@Annotated` allows only marker annotations, not annotations with values. Some annotations may only support fields, some methods - do we check this?
- [ ] #6231
- [x] #5578
- [x] Version ID recorded in builder-generated meta attributes (extensible for config-builder also); used for backwards interop
- [x] #5609
- [x] #5608
- [x] Use of assert() - should they stayor should they go? - If we decide to remove asserts(), then followup task is to remove them from #5598
- [x] #5712
- [x] #5978

Split off some tasks to https://github.com/helidon-io/helidon/issues/6221

trentjeff:
Addressed in https://github.com/helidon-io/helidon/pull/5750

PR #5598, opened 2022-12-06
Pico Tools - Part 1 of N

trentjeff:


pico/tools/src/main/java/io/helidon/pico/tools/Item.java
```diff
@@ -0,0 +1,200 @@
...
+import io.helidon.builder.Singular;
+
+/**
+ * These are the individual items that compose the {@link ModuleInfoDescriptor} builder.
+ * <p>
+ * Note that generally speaking the majority of use cases can use methods from {@link ModuleInfoDescriptor}
+ * instead of needing to use anything provided here.
+ */
+@Builder
+public interface Item {
```
tomas-langer, line 35: The name seems to be very general. If this is related to module info, maybe it should be `ModuleInfoItem`?

pico/tools/src/main/java/io/helidon/pico/tools/Item.java
```diff
@@ -0,0 +1,200 @@
...
+    /**
+     * Provides the content of the description item appropriate to write out.
+     *
+     * @return The contents (source code body) for this descriptor item.
+     */
+    default String contents() {
+        StringBuilder builder = new StringBuilder();
+        boolean handled = false;
+        if (uses()) {
+            assert (!requires());
```
tomas-langer, line 140: If this should be an actual runtime check, do not use `assert`
The assertions make this less readable and I do not see a benefit in adding them.

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
```diff
@@ -0,0 +1,633 @@
...
+    }
+
+    /**
+     * Loads the {@code module-info} descriptor given its source file location.
+     *
+     * @param file the source file location
+     * @return the module-info descriptor
+     * @throws io.helidon.pico.tools.ToolsException if there is any exception encountered
+     */
+    static ModuleInfoDescriptor uncheckedLoad(File file) {
```
tomas-langer, line 167: 1. we use `Path`, not `File` in Helidon
2. we try not to have checked exceptions in public APIs (except where required by implemented interface), so only use unchecked exceptions. There is an `UncheckedIoException` available in Java you can use here
3. static factory methods should be named `create`

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
```diff
@@ -0,0 +1,633 @@
...
+     *
+     * @param moduleInfo    the source
+     * @param ordering      the ordering to apply
+     * @return the descriptor
+     * @throws java.io.IOException if there is any IO exceptions encountered
+     * @throws io.helidon.pico.tools.ToolsException if there is any other exception encountered
+     */
+    static ModuleInfoDescriptor load(String moduleInfo,
+                                     Ordering ordering) throws IOException {
+        final DefaultModuleInfoDescriptor.Builder descriptor = DefaultModuleInfoDescriptor.builder();
```
tomas-langer, line 249: Do not use `final` except for fields and classes (and methods if it makes sense)

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
```diff
@@ -0,0 +1,633 @@
...
+import io.helidon.pico.types.DefaultTypeName;
+import io.helidon.pico.types.TypeName;
+
+/**
+ * Provides the basic formation for {@code module-info.java} creation and manipulation.
+ *
+ * @see java.lang.module.ModuleDescriptor
+ */
+@Builder
+public interface ModuleInfoDescriptor {
```
tomas-langer, line 49: Is this a parser of source of `module-info.java`?
There is class `java.lang.module.ModuleDescriptor` that supports reading the binary representation of module info - that can be used after compilation.
If you need to read module-info during compilation in annotation processor, I think there is a tool to obtain it correctly using annotation processor APIs.

This class seems redundant, and in general something that we should not be doing at all

pico/tools/src/main/java/io/helidon/pico/tools/TemplateHelper.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    private static String applySubstitutions(String target,
+                                             Map<String, Object> props,
+                                             boolean logErr,
+                                             boolean throwOnMissingArgs,
+                                             Set<String> missingArgs,
+                                             String delimStart,
+                                             String delimEnd) throws IOException {
+        if (null == target) {
```
tomas-langer, line 194: please use `(target == null)` instead (always)

tomas-langer (changes requested):

etc/copyright-exclude.txt
```diff
@@ -66,3 +66,5 @@ persistence_3_0.xjb
 persistence_3_0.xsd
 # excluded as this is a test file and we need to validate its content
 src/test/resources/static/classpath/index.html
+.hbs
+._java_
```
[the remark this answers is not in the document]
trentjeff: just decided on that name - feel free to suggest an alternative.  It can't be ".java" because it confuses IDEA and other tooling into thinking it's a real source file in the project.  This is used as a resource for unit testing - comparing what is actually code-generated to what was expected to be code-generated. So I could use the extension ".java.expected" or ".java-expected", etc.

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
trentjeff: 1. will change File -> Path.
2. I am using ToolsException (which is unchecked) for all forms of exceptions here including IO related.  If you prefer I can separate out a ToolsException (e.g., badly formatted file, or unhandled parsing scenario involving inner comments, etc) from a pure IO related exception.  LMK..
3. 3. will change to create.

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
trentjeff: next push

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
trentjeff: A few things to note on this one.

The java.lang.module.ModuleDescriptor is intended to work with the binary form of the module-info, not the source form. See https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleDescriptor.html.  Same is true for its builder.

And fwiw, this is intended to be internally facing for tooling related to pico, at least that is the case initially. I can deprecate it perhaps to avoid common use perhaps?  I do agree it is unfortunate that I had to parse it but I found no other good alternatives considering there is no way to see or build the stringified version of the source.

pico/tools/src/main/java/io/helidon/pico/tools/TemplateHelper.java
trentjeff: next commit

trentjeff:
@barchetta FYI this is adding a new dep on Handlebars for production use w/in Helidon.

pico/tools/src/main/java/io/helidon/pico/tools/Item.java
trentjeff: done; next commit

pico/pico/src/main/java/io/helidon/pico/PicoServicesHolder.java
```diff
@@ -41,6 +45,21 @@ private static Optional<PicoServices> load() {
...
+            throw new IllegalStateException("bootstrap already set");
+        }
+    }
+
+    static Optional<Bootstrap> bootstrap(boolean assignIfNeeded) {
+        if (assignIfNeeded) {
+            BOOTSTRAP.compareAndSet(null, DefaultBootstrap.builder().build());
+        }
+
+        return Optional.ofNullable(BOOTSTRAP.get());
```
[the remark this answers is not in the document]
trentjeff: next commit

etc/copyright-exclude.txt
```diff
@@ -66,3 +66,5 @@ persistence_3_0.xjb
 persistence_3_0.xsd
 # excluded as this is a test file and we need to validate its content
 src/test/resources/static/classpath/index.html
+.hbs
```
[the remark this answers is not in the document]
trentjeff: I am a bit surprised to hear you say that we support it since I added a new dependency to JKnack handlebars, and it seems I am the first to use anything in 3.x or 4.x for .hbs files in any other cases.

Fwiw, I tried to use the convention suggested and it failed the checkstyle check.  I will put it in my commit anyway for you to review.

etc/copyright-exclude.txt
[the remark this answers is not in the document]
trentjeff: just for testing

etc/copyright-exclude.txt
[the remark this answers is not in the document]
trentjeff: Follow-up will file bug on buildtools, and create a follow-up task here to remove .hbs exclusion.

pico/tools/src/main/java/io/helidon/pico/tools/Item.java
trentjeff: Follow-up task updated https://github.com/helidon-io/helidon/issues/5440

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptor.java
trentjeff: Follow-up task logged: https://github.com/helidon-io/helidon/issues/5697

etc/copyright-exclude.txt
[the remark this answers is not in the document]
trentjeff: Logged https://github.com/helidon-io/helidon-build-tools/issues/829

tomas-langer (approved):

PR #5611, opened 2022-12-06
Various metrics test improvements to avoid intermittent failures

### LINKED ISSUE #5345: Review all metrics tests for possible race conditions
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP 
* JDK version:
* OS:

----------

## Problem Description
Santiago saw this intermittent test failure:
```
[INFO] Running io.helidon.integrations.micrometer.cdi.HelloWorldTest
Error:  Tests run: 6, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 10.517 s <<< FAILURE! - in io.helidon.integrations.micrometer.cdi.HelloWorldTest
Error:  io.helidon.integrations.micrometer.cdi.HelloWorldTest.testSlowTimer  Time elapsed: 6.014 s  <<< FAILURE!
java.lang.AssertionError: 
Slow message timer count
Expected: is <3L>
     but: was <2L>
	at io.helidon.integrations.micrometer.cdi.HelloWorldTest.testSlowTimer(HelloWorldTest.java:98)
	at io.helidon.integrations.micrometer.cdi.HelloWorldTest$Proxy$_$$_WeldClientProxy.testSlowTimer(Unknown Source)
```
with 4.x. 

It's similar to some other intermittent failures we've seen. The underlying cause is that, for metrics updates which occur after the measured code completes (such as timers), the server updates the values using `response.whenSent`. It's possible, due to thread scheduling, that the server has sent the response to the test client and the test client has checked the metric that it expects to be updated _before_ (due to the async nature of `whenSent`) the server has updated the metric value. The test then reports a failure.

Other metrics tests might have the same potential problem.

This issue represents an overall review of all metrics tests that check a metric value that is updated after the measured code runs.

tjquinno:
Resolves #5345 (at least partially)

I reviewed all the metrics tests. I found some places to improve, but I did not identify specific causes for two of the failures Tomas described in the issue. They are certainly not simply "retry" problems.

I will continue looking into those, but I did not want to delay these improvements.

### DIFF
--- a/integrations/micrometer/cdi/pom.xml
+++ b/integrations/micrometer/cdi/pom.xml
@@ -115,6 +115,11 @@
             <artifactId>helidon-microprofile-tests-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.helidon.common.testing</groupId>
+            <artifactId>helidon-common-testing-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/integrations/micrometer/cdi/src/test/java/io/helidon/integrations/micrometer/cdi/HelloWorldTest.java
+++ b/integrations/micrometer/cdi/src/test/java/io/helidon/integrations/micrometer/cdi/HelloWorldTest.java
@@ -31,8 +31,11 @@ import jakarta.ws.rs.client.WebTarget;
 import jakarta.ws.rs.core.MediaType;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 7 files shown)


tomas-langer (commented):
Could you maybe change the retry matcher not to throw an interrupted exception, but to call `fail(interruptedException)` or throw a runtime exception?
I don't think we need to handle InterruptedException in any specific way, so it may be easier to have no checked exception at all (and in the end less code in our tests)

tjquinno:
Changed `MatcherWithRetry` as Tomas described and adjusted call sites.

tomas-langer (dismissed):

tomas-langer (approved):

PR #5667, opened 2022-12-13
Declare `h1-prefix` as early as possible so it is used for the title prefix

### LINKED ISSUE #5665: Typo (SE instead of MP) in MP -> Metrics -> Metrics-Capable Modules page header at top of page
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Doc page mentioned in title has the wrong upper-case flavor at the top line.

## Steps to reproduce
Browse to helidon.io/docs.
MP -> Metrics -> Metrics-Capable Modules

Top line on the page says SE instead of MP.

tjquinno:
Fixes #5665 

Even though `includes/mp.adoc` declares `h1-prefix` as `MP`, and `mp/metrics/metrics-capable-components.adoc` `includes` `mp.adoc[]`, the only way I found to get the prefix on the rendered page to work was to declare `h1-prefix` really early in the `mp/metrics/metrics-capable-components.adoc` file.

### DIFF
--- a/docs/mp/metrics/metrics-capable-components.adoc
+++ b/docs/mp/metrics/metrics-capable-components.adoc
@@ -17,6 +17,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 = Metrics-Capable Modules
+:h1-prefix: MP
 :description: Helidon MP metrics-capable modules
 :keywords: helidon, metrics, metrics-capable, microprofile, guide
 :intro-project-name: {h1-prefix}


docs/mp/metrics/metrics-capable-components.adoc
```diff
@@ -17,6 +17,7 @@
 ///////////////////////////////////////////////////////////////////////////////
 
 = Metrics-Capable Modules
+:h1-prefix: MP
 :description: Helidon MP metrics-capable modules
 :keywords: helidon, metrics, metrics-capable, microprofile, guide
 :intro-project-name: {h1-prefix}
```
romain-grecourt, line 23: How about moving `:intro-project-name: {h1-prefix}` after the include of `mp.adoc` ; does that work ?

romain-grecourt: Another way, which I think is better, is to use conditionals in the common content directly.

This makes the common docs "polymorphic" (big word :D) ; and sort of standalone (as it depends on the flavor include).

It does simplify the "contract" of these kind of common documents.
It can also help the GitHub rendering and set default values for the cases where the required attributes are not set.

PR #5668, opened 2022-12-13
Add allow-list handling to requested URI behavior

tjquinno:
This builds on your earlier changes in your fork.

1. Moved `Whitelist` and tests to `helidon-common-configurable` and renamed to `AllowList`
2. Added `trustedProxies` settings to relevant classes and builders.
3. Added whitelist processing to `Request#createRequestedUri`.

Note that if the allow-list processing is enabled it also makes sure that `remoteAddress()` is trustworthy--as well as any more remote proxies reported in the headers--or else we should not believe the headers we get at all.

I want to add more tests but I wanted you to have a chance to look at these changes before I go further.

Also needs doc. 

(Also, maybe I'm looking at this wrong but even though I started with your changes it seems that the combined changes are based on `helidon-3.x` not `main`. Is that what you intended? If not, I can straighten that out later after you have had a look at the current changes.)


dansiviter:
Does Helidon team have an inclusive wording policy? https://inclusivenaming.org/

tjquinno:
Dan, I'll be renaming the interface and implementing class in an upcoming push.

webserver/webserver/src/main/java/io/helidon/webserver/Request.java
```diff
@@ -219,40 +220,80 @@ private UriInfo createRequestedUri() {
...
-            switch (type) {
-            case FORWARDED -> {
-                List<Forwarded> forwardedList = Forwarded.create(headers);
-                if (!forwardedList.isEmpty()) {
-                    Forwarded f = forwardedList.get(0);
-                    scheme = f.proto().orElse(null);
-                    authority = f.host().orElse(null);
-                    discovered = true;
+        Whitelist trustedProxies = bareRequest.socketConfiguration().trustedProxies();
+        if (trustedProxies != null && trustedProxies.test(remoteAddress())) {
```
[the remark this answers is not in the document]
tjquinno: Fixed.

webserver/webserver/src/main/java/io/helidon/webserver/ServerBasicConfig.java
```diff
@@ -210,6 +218,7 @@ static class SocketConfig implements SocketConfiguration {
         private final BackpressureStrategy backpressureStrategy;
         private final int maxUpgradeContentLength;
         private List<RequestedUriDiscoveryType> requestedUriDiscoveryTypes;
+        private Whitelist trustedProxies;
```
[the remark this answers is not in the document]
tjquinno: @tomas-langer I wanted to clarify part of your comment.

What config setting did you have in mind for selecting "trust all" or "trust only allow-listed" behavior? 

If the user configures `server.requested-uri-discovery` at all then by default the discovery behavior ought to be  enabled and governed by whatever other settings are assigned within that config section. (It can be disabled using `requested-uri-discovery.enabled=false` or by removing the `requested-uri-discovery` config section entirely.)

The user can already select "trust all" like this:
```
server:
  requested-uri-discovery:
    types: forwarded
    trusted-proxies:
      allow:
        all: true
```
If the user has a `requested-uri-discovery` setting then isn't that by itself selecting `trust only allow-listed`?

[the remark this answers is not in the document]
tjquinno: I think with our subsequent discussions we are in agreement on what combinations of settings are valid. Here is what the code as written (I'm about to push the latest) does and is a candidate for what we could add to the doc for the webserver to describe this feature: (I've written in terms of config but the same applies to settings assigned using the builder):

In `SocketConfiguration.Builder` (which means immediately under `server` or under one of the `server.sockets` entries):

* `requested-uri-discovery` is optional.
* If `requested-uri-discovery` is disabled (either because it is absent or it is present and `requested-uri-discovery.enabled` is set to `false`) Helidon:
   *  ignores any `Forwarded` and `X-Forwarded` headers and adopts the `HOST` discovery type, meaning the `Host` header is used for host info and the request URI's scheme and port are used;
   * ignores any discovery `types` or `trusted-proxies` settings.
* If `requested-uri-discovery` is enabled (either because `enabled` is explicitly set to `true` or if either `types` or `trustedProxies` has been set), then Helidon performs a simple validity check: If `types` is specified and includes either `FORWARDED` or `X-FORWARDED`, then `trusted-proxies` must also be set explicitly. No `trusted-proxies` setting is needed if the user selects only the `HOST` discovery type.
* If discovery is enabled (explicitly or implicitly) and no discovery type is set, then Helidon defaults the type to `FORWARDED` and, therefore, requires at least an explicit `trusted-proxies` setting.

common/configurable/src/main/java/io/helidon/common/configurable/AllowList.java
```diff
@@ -0,0 +1,440 @@
...
+        return builder().config(config).build();
+    }
+
+    /**
+     * Test whether a value can be permitted.
+     *
+     * @param value value to test against
+     * @return {@code true} if the value is allowed, {@code false} if it is not allowed or it is explicitly denied
+     */
+    boolean test(String value);
```
tomas-langer, line 64: This should have `@Override` annotation

webserver/webserver/src/main/java/io/helidon/webserver/SocketConfiguration.java
```diff
@@ -259,6 +262,13 @@ default BackpressureStrategy backpressureStrategy() {
      */
     List<RequestedUriDiscoveryType> requestedUriDiscoveryTypes();
 
+    /**
+     * The allow list for trusted proxies.
+     *
+     * @return allow list for trusted proxies
+     */
+    AllowList trustedProxies();
```
tomas-langer, line 270: This must never return null. Either change to `Optional<AllowList>`, or return some default instance (by default this should be an allow list that denies all)

webserver/webserver/src/main/java/io/helidon/webserver/UnsafeRequestedUriSettings.java
```diff
@@ -0,0 +1,47 @@
...
+package io.helidon.webserver;
+
+/**
+ * Indicates unsafe settings for a socket's request URI discovery.
+ * <p>
+ *     This exception typically results when the user has enabled requested URI discovery or selected a discovery type
+ *     but has not assigned the trusted proxy {@link io.helidon.common.configurable.AllowList}.
+ * </p>
+ */
+public class UnsafeRequestedUriSettings extends RuntimeException {
```
tomas-langer, line 25: An exception class should always have `Exception` suffix in the name

webserver/webserver/src/test/java/io/helidon/webserver/BaseServerTest.java
```diff
@@ -47,16 +48,22 @@ protected static WebClient webClient() {
     }
 
     protected static void startServer(int port, Routing routing) throws Exception {
-        webServer = WebServer.builder()
+        startServer(port, routing, null);
```
tomas-langer, line 51: Please use `Config.empty()` instead of null - you will not need a null check and behavior will be the same

webserver/webserver/src/main/java/io/helidon/webserver/Request.java
```diff
@@ -218,41 +219,41 @@ private UriInfo createRequestedUri() {
...
-            switch (type) {
-            case FORWARDED -> {
-                List<Forwarded> forwardedList = Forwarded.create(headers);
-                if (!forwardedList.isEmpty()) {
-                    Forwarded f = forwardedList.get(0);
-                    scheme = f.proto().orElse(null);
-                    authority = f.host().orElse(null);
-                    discovered = true;
+        // Trusted proxies could be null if the discovery type is Host, either explicitly or by default.
+        AllowList trustedProxies = bareRequest.socketConfiguration().trustedProxies();
```
tomas-langer, line 223: `trustedProxies` can never be null (also commented on socket configuration interface).

tomas-langer (changes requested):
A few comments in code + please update copyrights of all changed files to 2023.

common/configurable/src/main/java/io/helidon/common/configurable/AllowList.java
tjquinno: fixed

tjquinno: fixed

webserver/webserver/src/main/java/io/helidon/webserver/SocketConfiguration.java
tjquinno: If the application does not specify any trusted proxies (config or builder setting), the `build` method now creates a default one with a "deny" predicate that is always `true` which denies all.

webserver/webserver/src/main/java/io/helidon/webserver/UnsafeRequestedUriSettings.java
tjquinno: changed

webserver/webserver/src/test/java/io/helidon/webserver/BaseServerTest.java
tjquinno: changed

webserver/webserver/src/main/java/io/helidon/webserver/Request.java
tjquinno: See above change.

tjquinno:
Marked as draft; need to update doc.

webserver/webserver/src/main/java/io/helidon/webserver/Request.java
```diff
@@ -218,41 +219,41 @@ private UriInfo createRequestedUri() {
...
+        if (trustedProxies != null && trustedProxies.test(hostPart(remoteAddress()))) {
```
tomas-langer, line 224: The null check should be removed. The `trustedProxies` cannot be null.

webserver/webserver/src/main/java/io/helidon/webserver/ServerBasicConfig.java
```diff
@@ -210,6 +218,7 @@ static class SocketConfig implements SocketConfiguration {
         private final BackpressureStrategy backpressureStrategy;
         private final int maxUpgradeContentLength;
         private List<RequestedUriDiscoveryType> requestedUriDiscoveryTypes;
+        private AllowList trustedProxies;
```
tomas-langer, line 221: Both fields should be final

tomas-langer (changes requested):

docs/config/io_helidon_common_configurable_AllowList.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+// tag::config[]
+
+
+Type: link:{javadoc-base-url}/io.helidon.common.configurable/io/helidon/common/configurable/AllowList.html[io.helidon.common.configurable.AllowList]
+
+
+
+
+== Configuration options
```
ljamen, line 35: my preference is title case but I see sentence case used throughout. H2 Configuration Options H3/H4 Configuration options. I'm ok with either as long as we're consistent (and we aren't across the board).

ljamen (commented):
Completed review. Added one comment that applies to all. No changes required.
Noted 2023 year updates in copyright.

docs/config/io_helidon_common_configurable_AllowList.adoc
tjquinno: These `.adoc` files are generated by a separate tool Tomas wrote. The tool could emit whatever we agree makes sense.

webserver/webserver/src/main/java/io/helidon/webserver/Request.java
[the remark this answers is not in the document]
tjquinno: fixed

webserver/webserver/src/main/java/io/helidon/webserver/Request.java
tjquinno: fixed

tomas-langer (approved):

PR #5683, opened 2022-12-15
[nima][performance] http1 server response BlockingOutputStream has necessary byte[] copy

rbygrave:
For BlockingOutputStream in Http1ServerResponse for the first buffer it currently has:

  // if somebody re-uses the byte buffer sent to us, we must copy it
  firstBuffer = buffer.copy();

 but the buffer here is always created locally via the 3
 BlockingOutputStream.write() methods so this buffer instance
 can never be re-used and the copy() is unnecessary performance
 overhead


rbygrave:
Ah no, ResponseOrderingTest fails with the use case of:

```java
.any("/stream", (req, res) -> {
    try (InputStream in = req.content().inputStream();
            OutputStream out = res.outputStream()) {
        in.transferTo(out);
    } catch (IOException e) {
        throw new UncheckedIOException(e);
    }
});
```

So the copy() is needed for this case. I need to think if there is some way to enable a "zero copy" write of content / bytes ... hmm.

I'll close this PR though.

Issue #5685, opened 2022-12-15
[nima][se] IllegalStateException("OutputStream already obtained"), when ErrorHandler uses ServerResponse.outputStream()


rbygrave:
## Environment Details
* Helidon Version:  4.0.0-SNAPSHOT / main
* Helidon SE 
* JDK version:  19.0.1
* OS: Linux

----------

## Problem Description
1. A Handler calls `ServerResponse.outputStream()`
2. Some exception occurs before any content has been sent, there are 0 bytes written 
3. A registered ErrorHandler uses `ServerResponse.outputStream()` in an attempt to write error response content
4. This fails with `IllegalStateException("OutputStream already obtained");`

Thrown at: https://github.com/helidon-io/helidon/blob/main/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1ServerResponse.java#L149-L151

### Option A: 
In ErrorHandlers.handleRequestException() when isSent() is false the "reset" the internal state of Http1ServerResponse such that it is ready for the error handler to set new headers and response content.  Set streamingEntity to false, set outputStream to null, clear the response headers?

https://github.com/helidon-io/helidon/blob/main/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http/ErrorHandlers.java#L142-L160
... if isSent() is false **_reset_** the ServerResponse 

### Option B:
Allow multiple calls to `ServerResponse.outputStream()` as long as isSent is false.

### Background: The old world `ServletResponse.reset()`

In the old world `ServletResponse` there are methods: 
- `isCommitted()` to indicate if some response content has been sent which is pretty similar to `isSent()`
- `reset()` to reset headers and any response buffers such that new content can be send (typically an error message).  reset() is only allowed to be called when isCommitted() is false.

Currently, `io.helidon.nima.webserver.http1.Http1ServerResponse` does not have an equivalent to the `ServletResponse.reset()` method.

## Steps to reproduce

I will look to produce an example test case that reproduces this issue and update this issue when I've done that. 
```
git clone git@github.com:rbygrave/nima-error-handler.git
cd nima-error-handler
mvn clean test

```

https://github.com/rbygrave/nima-error-handler/blob/main/src/test/java/org/example/MainTest.java

```java
package org.example;

import io.avaje.http.client.HttpClientContext;
import io.helidon.common.http.Http;
import io.helidon.nima.webserver.WebServer;
import io.helidon.nima.webserver.http.HttpRouting;
[...]

PR #5694, opened 2022-12-15
#5685 - Add ServerResponse.reset() to support ErrorHandlers

rbygrave:
- Fix for #5685 - IllegalStateException("OutputStream already obtained"), when ErrorHandler uses ServerResponse.outputStream()
- Adds ServerResponse.reset() to reset response headers and buffers
- Only allowed to reset() when isSent is false

Fixes the test at: https://github.com/rbygrave/nima-error-handler/blob/main/src/test/java/org/example/MainTest.java

### DIFF
--- a/nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2ServerResponse.java
+++ b/nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2ServerResponse.java
@@ -119,6 +119,17 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
     }
 
     @Override
+    public boolean reset() {
+        if (isSent) {
+            return false;
+        }
+        headers.clear();
+        streamingEntity = false;
+        outputStream = null;
+        return true;
+    }
+
+    @Override
     public OutputStream outputStream() {
         if (isSent) {
             throw new IllegalStateException("Response already sent");
--- a/nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2ServerResponse.java
+++ b/nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2ServerResponse.java
@@ -119,17 +119,6 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
     }
 
     @Override
-    public boolean reset() {
-        if (isSent) {
-            return false;
-        }
-        headers.clear();
-        streamingEntity = false;
-        outputStream = null;
-        return true;
-    }
-
-    @Override
     public OutputStream outputStream() {
         if (isSent) {
             throw new IllegalStateException("Response already sent");
@@ -164,6 +153,24 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
     @Override
     public boolean hasEntity() {
         return isSent || streamingEntity;
+    }
+
+    @Override
+    public boolean reset() {
+        if (isSent || outputStream != null && outputStream.bytesWritten > 0) {
+            return false;
+        }
+        headers.clear();
+        streamingEntity = false;
+        outputStream = null;
+        return true;
+    }
+
+    @Override
+    public void commit() {
+        if (outputStream != null) {
+            outputStream.commit();
+        }
     }
 
     private static class BlockingOutputStream extends OutputStream {
@@ -211,7 +218,18 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         }
 
         @Override
+        public void flush() throws IOException {
+            if (firstByte && firstBuffer != null) {
+                write(BufferData.empty());
+            }
+        }
+
+        @Override
         public void close() {
+            // does nothing, we expect commit(), so we can reset response when no bytes were written to response
+        }
+
+        void commit() {
             if (closed) {
                 return;
             }
@@ -292,6 +310,7 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
             headers.setIfAbsent(Header.create(Header.DATE, true, false, Http.DateTime.rfc1123String()));
 
             Http2Headers http2Headers = Http2Headers.create(headers);
+            http2Headers.status(status);
             http2Headers.validateResponse();
             int written = writer.writeHeaders(http2Headers,
                                               streamId,
--- a/nima/tests/integration/webserver/webserver/src/test/java/io/helidon/nima/tests/integration/server/ErrorHandlingWithOutputStreamTest.java
+++ b/nima/tests/integration/webserver/webserver/src/test/java/io/helidon/nima/tests/integration/server/ErrorHandlingWithOutputStreamTest.java
@@ -16,8 +16,10 @@
 
 package io.helidon.nima.tests.integration.server;
 
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+
 import io.helidon.common.buffers.DataReader;
-import io.helidon.common.http.Headers;
 import io.helidon.common.http.Http;
 import io.helidon.nima.testing.junit5.webserver.ServerTest;
 import io.helidon.nima.testing.junit5.webserver.SetUpRoute;
@@ -27,10 +29,8 @@ import io.helidon.nima.webserver.http.ErrorHandler;
 import io.helidon.nima.webserver.http.HttpRouting;
 import io.helidon.nima.webserver.http.ServerRequest;
 import io.helidon.nima.webserver.http.ServerResponse;
-import org.junit.jupiter.api.Test;
 
-import java.io.OutputStream;
-import java.nio.charset.StandardCharsets;
+import org.junit.jupiter.api.Test;
 
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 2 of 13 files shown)


rbygrave:
BTW: I have actually signed the OCA earlier this evening so not sure why that OCA Required is still showing up. Maybe it takes a while to sync that information.

tomas-langer:
Seems to me the following could happen:
- user obtains `outputStream`
- user writes data to output stream
- (this causes all headers to be written to the output stream)
- code throws exception (without closing the stream)
- error handler calls reset
- error handler sends response

The state is now very bad - we have already sent status and headers and  a part of entity, yet now error handler thinks it can write response status, headers and entity.
I think this would end in an inconsistent response data and a broken connection.

To work around this, we would need to keep track of the information if anything was written at all, and only allow `reset` if nothing was written (e.g. the response has not yet sent a single byte over the network).

Also this would need a test to validate such edge cases...

rbygrave:
> error handler calls reset

A call to `reset()` would not succeed if isSent is true.  I can add a test for this case easily enough so I'll do that but there isn't an issue wrt not closing the OutputStream _until_ it's actually sent content and isSent is then true.  reset() should not be called (and will not succed) if isSent() is true. 

> user writes data to output stream (this causes all headers to be written to the output stream)

Just to say it does not actually send for the **_first_** call to os.write() [which is instead stored as firstBuffer] so if there has been only 1 call to os.write() it doesn't actually send anything yet. The second and subsequent calls to os.write() do though.

That is, the ServerResponse isSent() is false until ... a second call to os.write() or os.flush() or os.close() - until then it is isSent false and able to be reset (which means firstBuffer can be set back to null).

rbygrave:
> To work around this, we would need to keep track of the information if anything was written at all

Yes. As I see it that exists now as `outputStream.totalBytesWritten()` 

> , and only allow reset if nothing was written (e.g. the response has not yet sent a single byte over the network).

Yes. I get the test below to pass when `isSent()` is adjusted to be:

```java
    @Override
    public boolean isSent() {
        return isSent || outputStream.totalBytesWritten() > 0;
    }
```

The test as I have it is - https://github.com/rbygrave/nima-error-handler/blob/main/src/test/java/org/example/MainTest.java#L59-L67
... and this test passes with the change to isSent() and use of reset() at https://github.com/helidon-io/helidon/pull/5694/files#diff-b91d11cb7be7bc7310211e9669285fed86ad298e284c44f32fbcead79e9799ddR195-R197

rbygrave:
> Also this would need a test to validate such edge cases...

Ah, I found the integration tests in `tests / integration / webserver / webserver` ... which are like the tests I created (IntelliJ failed me for a bit there) - I can look to add tests there that are like [the ones I created .](https://github.com/rbygrave/nima-error-handler/blob/main/src/test/java/org/example/MainTest.java)

nima/tests/integration/webserver/webserver/src/test/java/io/helidon/nima/tests/integration/server/ErrorHandlingWithOutputStreamTest.java
```diff
@@ -0,0 +1,145 @@
...
+        try (Http1ClientResponse response = client.method(Http.Method.GET)
+                .uri("/get-outputStream-writeFlushThenError")
+                .request()) {
+
+            // Getting (because Http1ServerResponse BlockingOutputStream.flush() is not overwritten
+            // and doesn't write the chunked firstBuffer - looks like a bug here for BlockingOutputStream.flush()
+            assertThat(response.status(), is(Http.Status.I_AM_A_TEAPOT_418));
+            assertThat(response.entity().as(String.class), is("CustomErrorContent"));
+
+            // FIXME: Expecting the below, looks like bug with flush() in Http1ServerResponse BlockingOutputStream
```
rbygrave, line 125: To me this looks like a separate bug here with BlockingOutputStream.flush() - when there is only one write() and firstBuffer is set but not sent ... then flush() really should send that firstBuffer content but it doesn't

nima/tests/integration/webserver/webserver/src/test/java/io/helidon/nima/tests/integration/server/ErrorHandlingWithOutputStreamTest.java
```diff
@@ -0,0 +1,145 @@
...
+        Headers headers;
+        try (Http1ClientResponse response = client.method(Http.Method.GET)
+                .uri("/get-outputStream-writeTwiceThenError")
+                .request()) {
+            assertThat(response.status(), is(Http.Status.OK_200));
+            assertThrows(DataReader.InsufficientDataAvailableException.class, () -> response.entity().as(String.class));
+            headers = response.headers();
+        }
+        // FIXME: It this supposed to be a (trailing) header for connection close ??
+        // assertThat(headers, hasHeader(Http.HeaderValues.CONNECTION_CLOSE));
```
rbygrave, line 110: As I see it this FIXME should be removed. We should not expect the Close Connection header here in this case. We expect the chunked data to not be terminated correctly and hence the InsufficientDataAvailableException and that should be correct and sufficient.

I should remove this FIXME comment from here.

nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2ServerResponse.java
```diff
@@ -118,6 +118,16 @@ public boolean isSent() {
...
+    @Override
+    public void reset() {
+        if (isSent) {
+            throw new IllegalStateException("Response already sent");
+        }
+        headers.clear();
+        streamingEntity = false;
+        outputStream = null;
+    }
+
```
rbygrave, line 130: We need the same tests for HTTP 2 (as we have for HTTP 1)

nima/tests/integration/webserver/webserver/src/test/java/io/helidon/nima/tests/integration/server/ErrorHandlingWithOutputStreamTest.java
```diff
@@ -0,0 +1,145 @@
...
+        assertThat(response.entity().as(String.class), is("ok"));
+    }
+
+    @Test
+    void testGetOutputStreamThenError_expect_CustomErrorHandlerMessage() {
+        var response = client.get("/get-outputStream").request();
+
+        assertThat(response.status(), is(Http.Status.I_AM_A_TEAPOT_418));
+        assertThat(response.entity().as(String.class), is("CustomErrorContent"));
+    }
```
rbygrave, line 89: I should improve the tests by asserting that we should see headers from the "main" handler and only headers from the error handler

rbygrave (commented):
Needs a few adjustments here

tomas-langer:
I think there should be a difference between `isSent()` and `bytesWrittenOverNetwork()`
The first (isSent()) is used as a very important check (see `Filters.FilterChainImpl.proceed()`, line 117) - we MUST complete the whole request/response interaction within the current virtual thread. So I really need to know that the response was fully sent, and no asynchronous handling will happen on the output after this point.
So modifying `isSent()` semantics will break some checks we have in place.

I can see two options how to solve this:
1. keep all check within the `reset` method, and return a `boolean` (e.g. success resetting is `true`, failure is `false`) In such a case, we do not need to change the API, other code can attempt to reset, and gracefully handle the case where it is just too late
2. add another public method such as `dataSent()` which would expose the information outside of response

I would prefer option 1 in this case, as it keeps the check internal to each specific implementation of response (HTTP/1 and HTTP/2) and does not expose information that is more of an implementation detail.

Please let me know if you want to continue on this feature yourself, or if you would prefer some other arrangement. If you want, you can contact me over our public slack channel to take care of the details

Thank you very much for your efforts on this!

rbygrave:
> I would prefer option 1

That makes good sense to me.

> if you want to continue on this feature yourself

I can make that change for option 1. I'm not sure about the http2 tests and implementation and maybe someone needs to take that over.

I have no issue if someone wants to take over this PR or completely redo it. All good either way. I'll look to make the change to option 1 but all fine it someone wants to start again fresh on this.

rbygrave:
I have pushed the change for "Option 1" to revert `isSent()` to original behavior and `reset()` now returns a boolean

> let me know if you want to continue on this feature yourself

I've probably hit my knowledge limit here. In ErrorHandlers the code in this PR throws `CloseConnectionException` when it can't reset before handing over to a custom ErrorHandler which seems reasonable and gives the behavior I expect (incorrectly terminated chunked response) for Http1 but might not be correct especially for Http2 (there might be better desired behavior). With Http2 I have 2 tests in Http2ErrorHandlingWithOutputStreamTest that hang which I have commented out. I don't know how to move those forward.

I'm happy if someone wants to start afresh with a new PR or take this PR forward or suggest changes.

Cheers, Rob.

tomas-langer:
Before I take over your code, could you kindly squash your commits into a single one (as I want to keep your commit in the history of Helidon)? 
I will then check your PR and work on top of it to make sure it works both for HTTP/1 and HTTP/2. 

Thanks!

rbygrave:
> squash your commits into a single one

Done.

> check your PR and work on top of it to make sure it works both for HTTP/1 and HTTP/2

That would be awesome.

Cheers, Rob.

tomas-langer:
@rbygrave I have rebased on latest `main` (some conflicting changes in `ErrorHandlers`). Could you please review my changes as well? 
Thanks for your contribution!

rbygrave (commented):
Looks good. As a side comment, I prefer AssertJ :)

tomas-langer (approved):

tomas-langer:

Issue #5706, opened 2022-12-16
Cannot debug in Eclipse IDE Version: 2022-09 (4.25.0)


jbescos:
```
!MESSAGE Errors running builder 'Java Builder' on project 'helidon-common-testing-junit5'.
!STACK 0
java.lang.NullPointerException: Cannot invoke "javax.lang.model.element.TypeElement.asType()" because the return value of "javax.lang.model.util.Elements.getTypeElement(java.lang.CharSequence)" is null
	at io.helidon.config.metadata.processor.ConfigMetadataHandler.init(ConfigMetadataHandler.java:112)
	at io.helidon.config.metadata.processor.ConfigMetadataProcessor.init(ConfigMetadataProcessor.java:61)
	at org.eclipse.jdt.internal.apt.pluggable.core.dispatch.IdeAnnotationProcessorManager.discoverNextProcessor(IdeAnnotationProcessorManager.java:97)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.round(RoundDispatcher.java:119)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.BaseAnnotationProcessorManager.processAnnotations(BaseAnnotationProcessorManager.java:172)
	at org.eclipse.jdt.internal.apt.pluggable.core.dispatch.IdeAnnotationProcessorManager.processAnnotations(IdeAnnotationProcessorManager.java:138)
	at org.eclipse.jdt.internal.compiler.Compiler.processAnnotations(Compiler.java:953)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:450)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:426)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(AbstractImageBuilder.java:379)
	at org.eclipse.jdt.internal.core.builder.BatchImageBuilder.compile(BatchImageBuilder.java:214)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(AbstractImageBuilder.java:311)
	at org.eclipse.jdt.internal.core.builder.BatchImageBuilder.build(BatchImageBuilder.java:79)
	at org.eclipse.jdt.internal.core.builder.JavaBuilder.buildAll(JavaBuilder.java:273)
	at org.eclipse.jdt.internal.core.builder.JavaBuilder.build(JavaBuilder.java:193)
[...]

PR #5708, opened 2022-12-16
Cannot debug in Eclipse IDE Version: 2022-09 (4.25.0) #5706

jbescos:
Solves: https://github.com/helidon-io/helidon/issues/5706


config/metadata-processor/src/main/java/io/helidon/config/metadata/processor/ConfigMetadataHandler.java
```diff
@@ -109,7 +109,10 @@ synchronized void init(ProcessingEnvironment processingEnv) {
 
         // get the types
         configuredElement = elementUtils.getTypeElement(CONFIGURED_CLASS);
-        builderType = elementUtils.getTypeElement("io.helidon.common.Builder").asType();
+        TypeElement builderTypeElement = elementUtils.getTypeElement("io.helidon.common.Builder");
```
tomas-langer, line 112: Any element annotated with metadata must have `io.helidon.common.Builder` on classpath, as otherwise the annotation processor does not make sense.

tomas-langer (commented):
I do not see why this change should be made. Is Eclipse adding the annotation processor to modules that should not have it?

config/metadata-processor/src/main/java/io/helidon/config/metadata/processor/ConfigMetadataHandler.java
jbescos: Eclipse IDE does not work very well with Maven (this happens more when there are plugins doing particular things). Possibly if you tune the module-path a bit you could end up fixing this issue.

But then, you are making it more complicated for any dev that is using Eclipse IDE and wants to contribute in Helidon.

This change does not have side effects, and it will help users to contribute in Helidon. For this reason I think we should include it.

We can add a WARNING if you want.

tomas-langer (approved):

Issue #5717, opened 2022-12-19
Nima WebServer.stop() is slow / not timely when there are no active requests


rbygrave:
## Environment Details
* Helidon Version: Nima 4.0.0-SNAPSHOT
* Helidon SE 
* JDK version: 19.0.1
* OS: Linus

----------

## Problem Description
1) `WebServer#stop()` when there have been NO requests takes ~500ms (value of ServerListener.EXECUTOR_SHUTDOWN_MILLIS)
2) `WebServer#stop()` on an Idle server that has no active requests takes ~500ms 

## Background wrt Reative Netty server
https://github.com/helidon-io/helidon/issues/2087

## Steps to reproduce
PR -  https://github.com/helidon-io/helidon/pull/5748 with failing tests for 1) and 2) Edit: This PR now also includes 2 ways of addressing the issue via modified jdk ThreadPerTaskExecutor and a application code approach (which I don't think is very nice).

Edit: Old PR was https://github.com/helidon-io/helidon/pull/5718

## Motivation

Fast graceful shutdown as I see it is critical to the performance of Kubernetes rollout. That is, K8s rollout is generally only as fast as the **startup time + shutdown time** so if server shutdown is slow then rollout is slow. This is more impactful when a K8s service has a lot of pods/replicas and fast graceful shutdown is very important.

rbygrave:
Note that the WebServerStopOnlyTest will pass when `executor.shutdown();` is added to `ServerListener.shutdownExecutor()` prior to the call to `executor.awaitTermination(...)`.  I expect this to also fix the idle case when the client requests are made as http 1.0 or no keepalive.

rbygrave:
For the `WebServerStopIdleTest`, as I see it the readerExecutor has a `ConnectionHandler` and for this test its `connection` is a `Http1Connection`.

At the time of `webServer.stop()` the thread running the `Http1Connection` is state WAITING while it's readPrologue() - it's waiting for the next request to come. At: https://github.com/helidon-io/helidon/blob/main/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1Connection.java#L115

A **_sort-of-hack-that-fixes-this_** is to add a stop() method to `Http1Connection` like below and invoke it before the `ServerListener` does `shutdownExecutor(readerExecutor);` at: https://github.com/helidon-io/helidon/blob/main/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/ServerListener.java#L134

```java
[...]

PR #5718, opened 2022-12-19
#5717 - Failing tests for - Nima WebServer.stop() is slow / not timel…

rbygrave:
…y when there are no active requests


rbygrave:
Closing this PR which is replaced by https://github.com/helidon-io/helidon/pull/5748

PR #5732, opened 2022-12-20
Add built-in support for converting config strings to enum values

### LINKED ISSUE #5719: Add built-in support for enums to config mapping
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Currently, SE config does not provide built-in support for enums, although we have some code that is written as if such support exists.

Proposed enhancement:

The developer's code could use something like this:

```java
public class MyClass {

    enum Type {A, B, SPECIAL_B};

    public class Builder... {

        Builder config(Config config) {
            ...
            config.get("type").as(MyClass.Type.class).ifPresent(this::type};
            ...
        }

        Builder type(MyClass.Type type) {
            ...
        }
    }
}
 ```

We would want to support some convenient automatic conversion from a string in the config source to the corresponding enum value. The following conventions would be useful and convenient to developers and users without being overly restrictive:

* The conversion from the string in the config source to the enum value is case-insensitive.
* Helidon converts hyphens (`-`) in the config source to underscores for matching to enum value names.
* Helidon throws a new `ConfigEnumMappingException` for any string that fails to convert to one of the enum's value names using the above rules. The exception reports the offending string and the valid values.

For example, the following would be valid config settings with the obvious mappings to the enum values:
```
myClass:
    type: a
    type: A
    type: Special-B
```

The only real restriction these rules would impose is that the enum values as declared in the Java code could not differ in only case because then the mapping would be ambiguous. In such cases, Helidon would throw a `ConfigEnumMappingException` exception indicating the ambiguous mapping. I think this is a reasonable restriction--it'd be potentially quite confusing in the java code itself to have enum values with the same name except for changes in case.

### PR DESCRIPTION
**Pull request by tjquinno**
Resolves #5719 

Adds a new `enum` mapper and automatically adds it to the config mappers (see the change in `ConfigMapperManager`).

I've attached a PDF of the updated doc (see the new section on the enum mapping) to simplify reviewing the doc changes.

[property-mapping.pdf](https://github.com/helidon-io/helidon/files/10271045/property-mapping.pdf)

### CODE REVIEW

File: config/config/src/main/java/io/helidon/config/ConfigMapperManager.java [hunk 1]
```diff
@@ -161,6 +161,7 @@ static final class MapperProviders {
                 Optional<? extends BiFunction<Config, ConfigMapper, ?>>>> providers = new LinkedList<>();
 
         private MapperProviders() {
+            add(new EnumMapperProvider());
```
ljnelson, line 164: This struck me as odd; is there something that makes this more primordial than other providers? (I am coming to this without a deep knowledge of the config system so there very well may be such a thing.)

config/config/src/test/java/io/helidon/config/ConfigMappersTest.java
```diff
@@ -256,4 +350,26 @@ private Config createConfig() {
                 .build();
     }
 
+    public enum MyType {A, B, SPECIAL_B};
+
+    public enum ConfusingType {RED, Red, BLUE};
+
+    private Config createEnumConfig() {
+        return Config.just(ConfigSources.create(Map.of(
```
trentjeff, line 358: "type", "a"
 - or -
"type", "SPECIAL_B"

^^^
This is the prime use case IMHO.

"type" is of type MyType, and the value will be an enum value of MyType.

config/config/src/main/java/io/helidon/config/ConfigMapperManager.java
trentjeff: Yea, basically I am suggesting the same thing in one of my comments as well.

tjquinno: I think it might strike you as odd only because this is the first built-in conversion that needs to be a mapper _provider_ and not simply a mapper.

All the other built-in conversions can be (and are) implemented as a single mapper for each type conversion, Concretely, we need only a single mapper for `Integer`, for example, because all integers obey the same conversion rules. 

Not so for `enum`s because each `enum` type needs its own mapper that knows which values are legal for _that_ `enum`.

Those other converters--being mappers--are plugged into the config infrastructure in one way. Being a mapper provider, this new one get plugged in differently.

ljnelson: Ah, makes sense. Feel free to resolve the comment.

config/config/src/main/java/io/helidon/config/ConfigEnumMappingException.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package io.helidon.config;
+
+import java.util.Arrays;
+
+/**
+ * Indicates an error attempting to map a string config value to a specified enum type.
+ */
+public class ConfigEnumMappingException extends ConfigMappingException {
```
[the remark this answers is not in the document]
tjquinno: Easily changed. Thoughts on this from other reviewers?

config/config/src/main/java/io/helidon/config/ConfigMapperManager.java
trentjeff: Actually it is analogous to Integer.
		Mytype.valueOf(<configValStr>)

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,107 @@
...
+
+import io.helidon.config.spi.ConfigMapperProvider;
+
+/**
+ * Built-in mapper for {@code enum}s.
+ *
+ * <p>
+ *     This mapper attempts to match strings in the config source to enum values as follows:
+ *     <ul>
+ *         <li>The mapper treats hyphens ('-') in config strings as underscores when comparing to enum value names.</li>
```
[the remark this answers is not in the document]
tjquinno: Here was my thinking on including hyphens.

In config, hyphens are often used stylistically instead of underscores as for readability _within_ names (`buffer-size`). So it seemed natural to support an equivalence between hyphens for in-name readability style in config sources and underscores as in-name readability style in `enum` value names.

I (at least) do not see dots playing the same kind of "in-name" readability role in config. Rather, they (again, to me) suggest hierarchy rather than readability enhancements.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,107 @@
...
+ *     </ul>
+ *     These conversions are intended to maximize ease-of-use for authors of config sources so the values need not be
+ *     upper-cased nor punctuated with underscores rather than the more conventional (in config at least) hyphen.
+ * </p>
+ * <p>
+ *     The only hardship this imposes is if a confusingly-designed enum has values which differ only in case <em>and</em> the
+ *     string in the config source does not exactly match one of the enum value names. In such cases
+ *     the mapper will be unable to choose which enum value matches an ambiguous string. A developer faced with this
+ *     problem can simply provide her own explicit config mapping for that enum, for instance as a function parameter to
+ *     {@code Config.as}.
```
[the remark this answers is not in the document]
tjquinno: I usually do. Will fix.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,107 @@
...
+ * </p>
+ *
+ */
+class EnumMapperProvider implements ConfigMapperProvider {
```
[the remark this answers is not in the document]
tjquinno: You've raised two issues.

1. Why a map? For efficiency so we don't instantiate multiple identical function objects for the same `enum` type if user code requests conversion of the same `enum` more than once.

2. (by "a manager" do you mean "a mapper provider"?) 

If you know of a way to do this without a mapper provider I'm all ears.

From what I found in the config code (and the inner workings of config are rather new to me so I might have missed things), the single-mapper approach does not work for `enum` support. Each mapper works with a given type _all instances of which obey the same mapping rules_ (such as the mapper for `Integer`, the mapper for `Long`, etc.). Mappers are therefore passed only the `Config` node at conversation time...nothing about the specific type being converted to because the config infrastructure assumes the mapper knows which type it is dealing with and so knows what conversion rules to apply.

In contrast, with the `enum` case each `enum` is a _different_ Java type, not a different instance of the same Java type. So a single mapper for all `enum` types does not work because a single mapper is not given enough information at mapping time to know _which specific_ `enum` type a value needs to be mapped to.

config/config/src/test/java/io/helidon/config/ConfigMappersTest.java
```diff
@@ -242,6 +246,96 @@ public void testBuiltinMappersSubNodeProperties() throws MalformedURLException {
...
+        assertThat("Lower-case mapping", type, is(MyType.A));
+
+        type = manager.map(config.get("type-A"), MyType.class);
+        assertThat("Upper-case mapping", type, is(MyType.A));
+
+        type = manager.map(config.get("type-special-b"), MyType.class);
+        assertThat("Mixed-clase and hyphen mapping", type, is(MyType.SPECIAL_B));
+
+        type = manager.map(config.get("type-special_b"), MyType.class); // <-- That's an underscore, not a hyphen.
+        assertThat("Mixed-clase and no hyphen mapping", type, is(MyType.SPECIAL_B));
```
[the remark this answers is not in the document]
tjquinno: I think am missing the point of this comment.

Are you suggesting that the config _key_ for a node is somehow linked to the `enum` value names?

The point of `enum` support is to map config _values_ to `enum` values. I chose the config keys in the tests to help me keep track of which key was paired with which values so I could write the test code to check the right conditions.

There is intended to be _no_ correspondence between config _keys_ and the `enum` values.

config/config/src/test/java/io/helidon/config/ConfigMappersTest.java
tjquinno: Again, I'm sorry I seem to be missing your point.

In particular, I'm not getting what role you see the string "type" playing in the conversion/mapping.

Let's use the color example (slightly expanded but not essentially different) from the doc instead of the artificial and (apparently) confusing config keys from the test.

```java
enum Color {RED, YELLOW, BLUE_GREEN};
```

```yaml
myapp:
    house:
      size: 2500
      tint: "blue-green"
    car:
      color: "Red"
      make: "Honda"
      model: "Pilot"
    warning: "YELLOW"
```
As with any config value of any type, there's nothing in the config key or value that conveys the type to which the config value might be mapped. That's up entirely to the application and its use of `config.asInt()` or `config.as(SomeType.class)` or the equivalent. 

To convert the config value with key `myapp.house.tint` to `Color` the app could use
```
   Color houseTint = config.get("myapp.house.tint").as(Color.class).get();
```

Can you recast your comment using this example so I can better understand what you're getting at?

config/config/src/test/java/io/helidon/config/ConfigMappersTest.java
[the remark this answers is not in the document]
trentjeff: The simpler solution could have been achieved in ~ 3 lines of code in the base config as method:

if (type.isEnum()) {
  Enum.valueOf(type, asString());
}

Obviously this is not capturing the naming heuristics, and not gracefully dealing with exceptions, etc - but that does the heart of what I think we are looking for.

config/config/src/test/java/io/helidon/config/ConfigMappersTest.java
trentjeff: sorry, this comment was more of a continuation of the previous one - we should just be testing the simpler prototypical usage where the ConfigValue (e.g., "A" or "SPECIAL_B") is being carried in a config key named "type", rather than mangled to use camel case or hyphens, etc.

config/config/src/test/java/io/helidon/config/ConfigMappersTest.java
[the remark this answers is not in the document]
tjquinno: 1. Please clarify the role you see the config key playing in all this. There is no such role and I did not intend to convey that there is one. If I have suggested that somehow I need to make sure the doc does not do so.
2. Re: a 3-line solution

A quick look at `as` from `ConfigExistingImpl` (_one_ of the implementations of `Config`) and a look at what happens downstream confirms otherwise:
https://github.com/helidon-io/helidon/blob/f4ff8438667b00c43494d9716feec5e15f21b292/config/config/src/main/java/io/helidon/config/ConfigExistingImpl.java#L72-L75

This kicks off the whole config mapping mechanism which includes priorities among mappers, runtime look-ups of additional mapping code, and--importantly--user-provided mappers and mapper providers which can override any of Helidon's built-in ones.

config/config/src/main/java/io/helidon/config/ConfigEnumMappingException.java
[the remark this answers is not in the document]
tjquinno: Removed.

config/config/src/main/java/io/helidon/config/ConfigMapperManager.java
tjquinno: Not at all analogous in the very general config mapping world where this is all happening.

Yes, both `Integer` and `MyEnum` have a `valueOf` method. But that's not enough to allow us to handle them the same way in value mapping.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private final Map<Class<?>, Function<Config, ?>> mappers = new HashMap<>();
+
+    @Override
+    public Map<Class<?>, Function<Config, ?>> mappers() {
+        return mappers;
+    }
+
+    @Override
+    public <T> Optional<Function<Config, T>> mapper(Class<T> type) {
```
ljnelson, line 62: I wonder if this shouldn't be:
```
public <S, T extends Enum<S>> Optional<Function<Config, T>> mapper(Class<T> type) {
```
I'm typing off the cuff, but I'm trying to indicate that `T` should both be an enum and one of a specific kind. Probably the compiler would hate me if I put this in for real but hopefully you get the idea.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,107 @@
...
+        if (!type.isEnum()) {
```
ljnelson, line 63: …and then this wouldn't be needed.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,107 @@
...
+            return Optional.empty();
+        }
+
+        Function<Config, T> result = (Function<Config, T>) mappers.computeIfAbsent(type, this::enumMapper);
+        return Optional.of(result);
+    }
+
+    private <T> Function<Config, T> enumMapper(Class<T> type) {
```
ljnelson, line 71: (See my earlier comments. I'm sure they'll have ripple effects. This would be one of probably many such places.)

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,113 @@
...
+ * </p>
+ *
+ */
+class EnumMapperProvider implements ConfigMapperProvider {
+
+    private final Map<Class<?>, Function<Config, ?>> mappers = new HashMap<>();
+
+    @Override
+    public Map<Class<?>, Function<Config, ?>> mappers() {
+        return mappers;
```
tomas-langer, line 58: This should not be returned, just return `Map.of()`, as this method is used for providers that have a finite list of known classes it supports.
You do not need to cache anything, it is done by the config mapper manager.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tomas-langer: You cannot do that, that would violate the SPI.
You cannot restrict the signature, as the config mapper manager calls this method with an arbitrary class (one that somebody requested).

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tomas-langer: And it would still be needed, as the invoker of this method does not have the tools to only invoke us with a valid type

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tomas-langer: Here we can change the signature, as we already validated that it is an enum, so we can cast the type

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,113 @@
...
+        return config -> {
+            // type.isEnum() must be true in the caller, so we know the type is Class<Enum<?>>.
+            Class<Enum<?>> enumType = (Class<Enum<?>>) type;
+            if (!config.hasValue() || !config.exists()) {
+                throw MissingValueException.create(config.key());
+            }
+            if (!config.isLeaf()) {
+                throw new ConfigMappingException(config.key(), enumType, "config node must be a leaf but is not");
+            }
+            String value = config.asString().get();
```
tomas-langer, line 81: why did you decide to do the conversion rules? 
I though this would be a simple use of the value (I could also understand the case insensitiveness). 
I am not much in favor of doing tricks like this to try to "handle all cases". If they want "MY_VALUE", they can specify "MY_VALUE" or "my_value"... no need to support dashes and/or underscores (you could come up with lot of rules then, which is not very helpful)

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
```diff
@@ -0,0 +1,113 @@
...
+            String convertedValue = value.replace('-', '_');
+            List<Enum<?>> caseInsensitiveMatches = new ArrayList<>();
+            for (Enum<?> candidate : enumType.getEnumConstants()) {
+                // Check for an exact match first, with or without hyphen conversion.
+                if (candidate.name().equals(convertedValue) || candidate.name().equals(value)) {
+                    return (T) candidate;
+                }
+                if (candidate.name().compareToIgnoreCase(convertedValue) == 0
```
tomas-langer, line 89: why do you use `compareToIgnoreCase` when there is `equalsIgnoreCase`?

tomas-langer (changes requested):
Please make sure this works in native image as well (without adding the enum types for reflection).

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tjquinno: No particular reason. Will change.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tjquinno: I will remove the caching and change `mappers()`.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tjquinno: I described my thinking in the JavaDoc for `EnumMapperProvider`:

 > These conversions are intended to maximize ease-of-use for authors of config sources so the values need not be   upper-cased nor punctuated with underscores rather than the more conventional (in config at least) hyphen.

I generally agree about not trying to be overly clever in code trying to guess what a user intends, particularly where that intent is ambiguous. Here, though, the purpose is definitely _not_ to "handle all cases" but to anticipate two specific mistakes I expect users will commonly make where their intent is quite clear:
* The hyphen is not a legal character in a Java `enum` value declaration. Given the appearance of a hyphen in a config value where an `enum` value name is expected, the user intent seems clear.
* Case insensitivity seemed obvious (and we agree on that).

I suggested these two specific rules--not "lot of rules"--to help users avoid two clear sources of errors and frustration (which seem likely) and help us avoid subsequent incoming support traffic.

Most importantly for users, I think the conversion rules improve ease-of-use without really adding confusion or ambiguity in how Helidon will behave.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tjquinno: Laird, I tried to find a way around this, but, as Tomas mentioned, it cannot be done.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tjquinno: Changed.

config/config/src/main/java/io/helidon/config/EnumMapperProvider.java
tomas-langer: OK, let's keep it in, as hyphen is not legal in enum value declarations.

config/config/src/main/java/io/helidon/config/ConfigMapperManager.java
tomas-langer, line 164: I would actually prefer to add the `EnumMapperProvider` through service loader as any other, with a very low priority (e.g. Priority 10000 in Helidon 3.x and 2.x, Weight 0 in Helidon 4).
This will give it the same treatment as any user created provider (and an option to use Helidon service loader features for it)

tomas-langer: There are two way to do this:
1. Add `provides` to `module-info.java` (needs to be public)
2. Add it manually in `BuilderImpl.loadMapperServices(List)` to the `HelidonServiceLoader.Builder` (can be package local)

tomas-langer (changes requested):

config/config/src/main/java/io/helidon/config/ConfigMapperManager.java
tjquinno: I can change this easily and, yes, I have supplied service loader impls in several places before 😉 

My original thinking was that this is _not_ a user-provided mapper provider but the first Helidon-provided one (unless I missed something) and therefore it made sense to include it directly as we do with so many of the built-in mappers.

Either way is fine with me based on your preference. If I change it to be loaded as a service I'll not use the `provides` approach. No need for the provider to be public.

tjquinno: Changed

trentjeff: I'm missing the basic premise as to why we feel we need a pluggable manager for dealing with Enums.  It can be implemented much more simpler in the base Config classes using the code snippet I provided.  Can someone explain why we'd want to have a manager for dealing with Enum type A that is different from a manager for dealing with Enum type B?  Or why these need to be overridable?  I still feel we should treat Enums as a primitive type, and built into the core config library.

tjquinno: If the config system were not already written the way it is, there might not be a compelling reason to _create_ a way for the enum conversion to be pluggable and overridable. But given the code as it stands, creating a mapper provider for enum conversion is the right approach (see below). Tomas's suggestion for how to incorporate our enum mapper provider reuses existing code which happens to make it pluggable and overridable. Developers might never write their own but doing it this way follows the existing patterns and conventions in the config subsystem for providing conversion rather than inventing some new mechanism.

There is no single "base `Config#as`" method, as your earlier suggestion assumed, where your three lines might be inserted. I pointed before to what the `as` method looks like in one implementation of the `Config` interface.
If you look at the logic which starts there, you see that the code there does _not_  perform the conversion. 

Instead--and I'm simplifying a bit--that code creates a `ConfigValue` instance and stores with it a function which, given the type passed to `as`, will choose the mapper to apply at the _later_ point when conversion _does_ occur.

Only later, when the converted value is actually needed (e.g., when `ifPresent` or `get` are invoked), does config invoke the saved function to choose the mapper and then invoke the mapper to perform the conversion.

That's why we need a way--that fits in with the way config is designed--of delivering enum mappers.

I've explained before why we need different mappers for different specific enum types, and therefore why we need a mapper provider rather than a single mapper for all enums, but to restate: 

Config's mapper selection is by the type that was stored with the `ConfigValue` instance created by `as`. So when the developer invokes `config.as(Color.class)` then `Color` is the type--not `Enum`--config will later use to find a mapper. So we'd better be prepared to deliver config a mapper for `Color` when config asks for one.

The new enum mapper provider does exactly that--it creates a mapper on-demand that converts a string to `Color` or `Shape` or whatever enum type might arise.

tests/integration/native-image/se-1/src/main/java/io/helidon/tests/integration/nativeimage/se1/ColorService.java
```diff
@@ -0,0 +1,59 @@
...
+        configuredColor = config.get("color.tint").as(Color.class).get();
+    }
+
+    @Override
+    public void update(Routing.Rules rules) {
+        rules.get("/", this::reportColor);
+    }
+
+    private void reportColor(ServerRequest request, ServerResponse response) {
+        if (configuredColor == null) {
```
tomas-langer, line 53: This should never be true - either the constructor throws an exception, or we get a valid color.

tomas-langer (changes requested):

tomas-langer (approved):

PR #5733, opened 2022-12-20
Introduces LocalXAResource and a few support classes in jta/jdbc

ljnelson:

### DIFF
--- a/integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
+++ b/integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
@@ -1,0 +1,1031 @@
+/*
+ * Copyright (c) 2022 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.integrations.jta.jdbc;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.EnumSet;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.UnaryOperator;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.transaction.xa.XAException;
+import javax.transaction.xa.XAResource;
+import javax.transaction.xa.Xid;
+
+import io.helidon.integrations.jdbc.SQLRunnable;
+import io.helidon.integrations.jdbc.UncheckedSQLException;
+import io.helidon.integrations.jta.jdbc.ExceptionConverter.XARoutine;
+
+import static javax.transaction.xa.XAException.XAER_DUPID;
+import static javax.transaction.xa.XAException.XAER_INVAL;
+import static javax.transaction.xa.XAException.XAER_NOTA;
+import static javax.transaction.xa.XAException.XAER_PROTO;
+import static javax.transaction.xa.XAException.XAER_RMERR;
+import static javax.transaction.xa.XAException.XAER_RMFAIL;
+import static javax.transaction.xa.XAException.XA_RBROLLBACK;
+import static javax.transaction.xa.XAResource.TMENDRSCAN;
+import static javax.transaction.xa.XAResource.TMFAIL;
+import static javax.transaction.xa.XAResource.TMJOIN;
+import static javax.transaction.xa.XAResource.TMNOFLAGS;
+import static javax.transaction.xa.XAResource.TMONEPHASE;
+import static javax.transaction.xa.XAResource.TMRESUME;
+import static javax.transaction.xa.XAResource.TMSTARTRSCAN;
+import static javax.transaction.xa.XAResource.TMSUCCESS;
+import static javax.transaction.xa.XAResource.TMSUSPEND;
+import static javax.transaction.xa.XAResource.XA_OK;
+import static javax.transaction.xa.XAResource.XA_RDONLY;
+
+/**
+ * An {@link XAResource} that adapts an ordinary arbitrary {@link Connection} as much as possible to the {@link
+ * XAResource} contract.
+ *
+ * <p><strong>Note:</strong> instances of this class are lossless in the presence of one-phase commit operations and
+ * potentially lossy in the presence of two-phase commit operations.</p>
+ *
+ * <p>Instances of this class are safe for concurrent use by multiple threads.</p>
+ */
+final class LocalXAResource implements XAResource {
+
+
+    /*
+     * Static fields.
+     */
+
+
+    private static final Logger LOGGER = Logger.getLogger(LocalXAResource.class.getName());
+
+    private static final Xid[] EMPTY_XID_ARRAY = new Xid[0];
+
+    // package-protected for testing only.
+    static final ConcurrentMap<Xid, Association> ASSOCIATIONS = new ConcurrentHashMap<>();
+
+
+    /*
+     * Instance fields.
+     */
+
+
+    private final Function<? super Xid, ? extends Connection> connectionFunction;
+
+    private final ExceptionConverter exceptionConverter;
+
+
+    /*
+     * Constructors.
+     */
+
+
+    /**
+     * Creates a new {@link LocalXAResource}.
+     *
+     * @param connectionFunction a {@link Function} that accepts a {@link Xid} (supplied by the {@link #start(Xid, int)}
+     * method) and returns a {@link Connection} to associate with the global transaction; must not be {@code null}; must
+     * never return {@code null}; must be safe for concurrent use by multiple threads; will never be invoked with a
+     * {@code null} {@link Xid}
+     *
+     * @param exceptionConverter a {@link ExceptionConverter} that accepts a {@link XARoutine} and a {@link
+     * SQLException} and converts the {@link SQLException} to an <em>appropriate</em> {@link XAException} following the
+     * rules defined by the <a href="https://pubs.opengroup.org/onlinepubs/009680699/toc.pdf">XA Specification</a> as
+     * interpreted by the specification of the {@code javax.transaction.xa} package and its classes; may be {@code null}
+     * in which case a default implementation will be used instead
+     *
+     * @see #start(Xid, int)
+     */
+    LocalXAResource(Function<? super Xid, ? extends Connection> connectionFunction, ExceptionConverter exceptionConverter) {
+        super();
+        this.connectionFunction = Objects.requireNonNull(connectionFunction, "connectionFunction");
+        this.exceptionConverter = exceptionConverter == null ? LocalXAResource::convert0 : exceptionConverter;
+    }
+
+
+    /*
+     * Instance methods.
+     */
+
+
+    @Override // XAResource
+    public void start(Xid xid, int flags) throws XAException {
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.entering(this.getClass().getName(), "start", new Object[] {xid, flagsToString(flags)});
+        }
+        requireNonNullXid(xid);
+
+        BiFunction<Xid, Association, Association> remappingFunction;
+        switch (flags) {
+        case TMJOIN:
+            remappingFunction = LocalXAResource::join;
+            break;
+        case TMNOFLAGS:
+            remappingFunction = this::start;
+            break;
+        case TMRESUME:
+            remappingFunction = LocalXAResource::resume;
+            break;
+        default:
+            // Bad flags.
+            throw (XAException) new XAException(XAER_INVAL)
+                .initCause(new IllegalArgumentException("xid: " + xid + "; flags: " + flagsToString(flags)));
+        }
+
+        this.computeAssociation(XARoutine.START, xid, remappingFunction);
+
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "start");
+        }
+    }
+
+    // (Remapping BiFunction, used in start() above and supplied to computeAssociation() below.)
+    private Association start(Xid x, Association a) {
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.entering(this.getClass().getName(), "start", new Object[] {x, a});
+        }
+        assert x != null; // x has already been vetted and is known to be non-null
+        if (a != null) {
+            throw new UncheckedXAException((XAException) new XAException(XAER_DUPID)
+                                           .initCause(new IllegalArgumentException("xid: " + x + "; association: " + a)));
+        }
+        Connection c;
+        try {
+            c = this.connectionFunction.apply(x);
+        } catch (RuntimeException e) {
+            // Weirdly, XAER_RMERR seems to be the "please retry" error code in this one case, not XAER_RMFAIL:
+            // https://github.com/jbosstm/narayana/blob/8ccaf0f85c7a76c227941d26cc3aa3fa9f05b160/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L682-L689.
+            throw new UncheckedXAException((XAException) new XAException(XAER_RMERR).initCause(e));
+        }
+        if (c == null) {
+            // Weirdly, XAER_RMERR seems to be the "please retry" error code in this one case, not XAER_RMFAIL:
+            // https://github.com/jbosstm/narayana/blob/8ccaf0f85c7a76c227941d26cc3aa3fa9f05b160/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L682-L689.
+            throw new UncheckedXAException((XAException) new XAException(XAER_RMERR)
+                                           .initCause(new NullPointerException("connectionFunction.apply(" + x + ")")));
+        }
+        a = new Association(Association.BranchState.ACTIVE, x, c);
+        if (LOGGER.isLoggable(Level.FINE)) {
+            LOGGER.logp(Level.FINE, this.getClass().getName(), "start",
+                        "Created new Association ({0}) for connection ({1}) in state ACTIVE", new Object[] {a, c});
+        }
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "start", a);
+        }
+        return a;
+    }
+
+    @Override // XAResource
+    public void end(Xid xid, int flags) throws XAException {
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.entering(this.getClass().getName(), "end", new Object[] {xid, flagsToString(flags)});
+        }
+        requireNonNullXid(xid);
+
+        BiFunction<Xid, Association, Association> remappingFunction;
+        switch (flags) {
+        case TMFAIL:
+        case TMSUCCESS:
+            remappingFunction = LocalXAResource::activeToIdle;
+            break;
+        case TMSUSPEND:
+            remappingFunction = LocalXAResource::suspend;
+            break;
+        default:
+            // Bad flags.
+            throw (XAException) new XAException(XAER_INVAL)
+                .initCause(new IllegalArgumentException("xid: " + xid + "; flags: " + flagsToString(flags)));
+        }
+
+        // Any XAException thrown can have any error code. The transaction will be marked as rollback only. See
+        // https://github.com/jbosstm/narayana/blob/8ccaf0f85c7a76c227941d26cc3aa3fa9f05b160/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L978-L992.
+        this.computeAssociation(XARoutine.END, xid, remappingFunction);
+
+        if (LOGGER.isLoggable(Level.FINER)) {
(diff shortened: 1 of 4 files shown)


ljnelson:
Another little grab bag pulled from [my private branch work](https://github.com/helidon-io/helidon/compare/helidon-3.x...ljnelson:helidon:nested-transactions-helidon-3.x).

This introduces the `LocalXAResource` class, which is an `XAResource` implementation that adapts a non-JTA-aware `Connection` to the XA protocol.

The other classes introduced by this PR are supporting in nature and fairly simple.

More PRs in this area to follow.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+/**
+ * An {@link XAResource} that adapts an ordinary arbitrary {@link Connection} as much as possible to the {@link
+ * XAResource} contract.
+ *
+ * <p><strong>Note:</strong> instances of this class are lossless in the presence of one-phase commit operations and
+ * potentially lossy in the presence of two-phase commit operations.</p>
+ *
+ * <p>Instances of this class are safe for concurrent use by multiple threads.</p>
+ */
+final class LocalXAResource implements XAResource {
```
ljnelson, line 66: Hello, reviewers. I'll annotate this PR myself with comments like this to help you navigate the XA world.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+    }
+
+
+    /*
+     * Instance methods.
+     */
+
+
+    @Override // XAResource
+    public void start(Xid xid, int flags) throws XAException {
```
ljnelson, line 126: The XA protocol—the permitted sequence of invocations on an `XAResource`—begins with `start`.  `start`'s job is to mark the beginning of the association of operations on the current thread against the "resource manager" (the database connection in this case) with the global transaction identified by the supplied `Xid`.  (The ending of this thread association is accomplished by the `end` method.)

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        case TMRESUME:
+            remappingFunction = LocalXAResource::resume;
+            break;
+        default:
+            // Bad flags.
+            throw (XAException) new XAException(XAER_INVAL)
+                .initCause(new IllegalArgumentException("xid: " + xid + "; flags: " + flagsToString(flags)));
+        }
+
+        this.computeAssociation(XARoutine.START, xid, remappingFunction);
```
ljnelson, line 149: We select the appropriate `BiFunction` to pass in to the `computeAssociation` method.  As hopefully you can see above, it's determined by what switches the transaction manager has passed when it called this `start` method.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "start");
+        }
+    }
+
+    // (Remapping BiFunction, used in start() above and supplied to computeAssociation() below.)
+    private Association start(Xid x, Association a) {
```
ljnelson, line 157: In the `start` method above, if we were called with the right set of flags, then we are truly beginning a new transaction.  The `Association` object is an inner class (see down at the end of this file) representing the association of a given `Connection` (in this `XAResource` implementation's case) with the current thread.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        }
+        assert x != null; // x has already been vetted and is known to be non-null
+        if (a != null) {
+            throw new UncheckedXAException((XAException) new XAException(XAER_DUPID)
+                                           .initCause(new IllegalArgumentException("xid: " + x + "; association: " + a)));
+        }
+        Connection c;
+        try {
+            c = this.connectionFunction.apply(x);
+        } catch (RuntimeException e) {
```
ljnelson, line 169: Figuring out whether to throw an `XAException` with `XAER_RMERR` or `XAER_RMFAIL` is a problem that has resulted in dozens of bugs by all major `XAResource` implementors over a couple of decades. It is far from straightforward and ends up being a _de facto_ decision: you have to do what Narayana expects, regardless of whether it implements the specification properly or not, because it is the only real robust JTA implementation out there and has encountered just about everything in its 30+ years of existence.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+                        "Created new Association ({0}) for connection ({1}) in state ACTIVE", new Object[] {a, c});
+        }
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "start", a);
+        }
+        return a;
+    }
+
+    @Override // XAResource
+    public void end(Xid xid, int flags) throws XAException {
```
ljnelson, line 192: When the transaction manager decides that whatever has gone on on the current thread should stop being associated with the global JTA transaction, it will call this method.  Once again the flags tell you what sort of operation is really being invoked: a disassociation-from-the-current-thread due to an ordinary transaction boundary, a problem or a suspension.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+            remappingFunction = LocalXAResource::suspend;
+            break;
+        default:
+            // Bad flags.
+            throw (XAException) new XAException(XAER_INVAL)
+                .initCause(new IllegalArgumentException("xid: " + xid + "; flags: " + flagsToString(flags)));
+        }
+
+        // Any XAException thrown can have any error code. The transaction will be marked as rollback only. See
+        // https://github.com/jbosstm/narayana/blob/8ccaf0f85c7a76c227941d26cc3aa3fa9f05b160/ArjunaJTA/jta/classes/com/arjuna/ats/internal/jta/transaction/arjunacore/TransactionImple.java#L978-L992.
```
ljnelson, line 214: Once again, figuring out what kind of error code to throw when is trial and error in the real world.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        this.computeAssociation(XARoutine.END, xid, remappingFunction);
+
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "end");
+        }
+    }
+
+    @Override // XAResource
+    public int prepare(Xid xid) throws XAException {
```
ljnelson, line 223: This method is called after the `start`/`end` sequence has occurred.  It signifies the start of the two-phase completion process.  It is only called in the case of a true two-phase situation, i.e. where multiple resource managers are involved.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        int returnValue = association == null ? XA_RDONLY : XA_OK;
+
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "prepare", returnValue);
+        }
+        return returnValue;
+    }
+
+    @Override // XAResource
+    public void commit(Xid xid, boolean onePhase) throws XAException {
```
ljnelson, line 247: This method is called by the transaction manager during the two-phase completion stage, after the `start`/`end` sequence has occurred, and possibly after a `prepare` invocation (if `onePhase` is false).  At this point the resource manager (represented by the `Connection` in our case) must commit its work, or the XA protocol will be deemed to have been violated.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+                                a -> commitAndReset(a, onePhase),
+                                true); // do remove association on error
+
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "commit");
+        }
+    }
+
+    @Override // XAResource
+    public void rollback(Xid xid) throws XAException {
```
ljnelson, line 279: This method is invoked as part of the two-phase completion process (after the `start`/`end` cycle, and possibly after a `prepare` invocation).  Strictly speaking this may be invoked on any thread at any time after the `start`/`end` cycle.  The resource manager must roll its work back.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+                                LocalXAResource::forgetAndReset,
+                                false); // don't remove association on error
+
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.exiting(this.getClass().getName(), "forget");
+        }
+    }
+
+    @Override // XAResource
+    public Xid[] recover(int flags) throws XAException {
```
ljnelson, line 321: I believe this method may be invoked by any thread at any time at the discretion of the transaction manager to recover from a failure with the two-phase completion process (so something going wrong after the `start`/`end` cycle).  It is sort of irrelevant as this "local" `XAResource` implementation can't recover things anyway.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+    public int getTransactionTimeout() {
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.entering(this.getClass().getName(), "getTransactionTimeout");
+            LOGGER.exiting(this.getClass().getName(), "getTransactionTimeout", 0);
+        }
+        return 0;
+    }
+
+    @Override // XAResource
+    public boolean setTransactionTimeout(int transactionTimeoutInSeconds) {
```
ljnelson, line 362: Interesting point of note: this is the first method that is called by the transaction manager (at least in Narayana's implementation), even before `start`.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+            return ASSOCIATIONS.compute(xid, (x, a) -> remap(x, a, legalBranchStates, f));
+        } catch (RuntimeException e) {
+            if (removeAssociationOnError) {
+                ASSOCIATIONS.remove(xid);
+            }
+            throw this.convert(xaRoutine, e);
+        }
+    }
+
+    private XAException convert(XARoutine xaRoutine, Throwable e) {
```
ljnelson, line 399: After reading many bug reports from different decades it's clear that no one, not even the specification authors, has made the error requirements clear to implementors, so this method plus the `ExceptionConverter` interface should allow an escape hatch for us if we end up getting the default behavior wrong.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+
+    private static void requireNonNullXid(Xid xid) throws XAException {
+        if (xid == null) {
+            throw (XAException) new XAException(XAER_INVAL).initCause(new NullPointerException("xid"));
+        }
+    }
+
+    // (Used via method reference only when an exceptionConverter was not supplied at construction time. This is the
+    // default implementation.)
+    private static XAException convert0(XARoutine xaRoutine, Exception e) {
```
ljnelson, line 442: This represents the default XA exception conversion logic.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+                    sqlException = s;
+                } else {
+                    sqlException = null;
+                }
+                if (sqlException != null) {
+                    String sqlState = sqlException.getSQLState();
+                    if (sqlState != null
+                        && (sqlState.startsWith("080")
+                            || sqlState.equalsIgnoreCase("08S01") // ("ess" not "five")
+                            || sqlState.equalsIgnoreCase("JZ006"))) {
```
ljnelson, line 466: SQLStates are extremely specific: `08` indicates a class of error that designates some kind of a connection error, and the other specific codes cited here are common issues that places like StackOverflow reveal are problematic in the real world.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+                                           .initCause(new NullPointerException("xid: " + xid + "; association: null")));
+        } else if (!legalBranchStates.contains(a.branchState())) {
+            throw new UncheckedXAException((XAException) new XAException(XAER_PROTO)
+                                           .initCause(new IllegalStateException("xid: " + xid + "; association: " + a)));
+        }
+        return remapOperator.apply(a);
+    }
+
+    // (Remapping BiFunction. Used in end() above.)
+    private static Association activeToIdle(Xid x, Association a) {
```
ljnelson, line 502: For reviewers reading sequentially: you'll find all this "active" and "idle" talk and so on confusing unless you scroll to the bottom of this file and read a bit about the arcane branch association states defined by the specification.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+            LOGGER.exiting(this.getClass().getName(), "start");
+        }
+    }
+
+    // (Remapping BiFunction, used in start() above and supplied to computeAssociation() below.)
+    private Association start(Xid x, Association a) {
+        if (LOGGER.isLoggable(Level.FINER)) {
+            LOGGER.entering(this.getClass().getName(), "start", new Object[] {x, a});
+        }
+        assert x != null; // x has already been vetted and is known to be non-null
```
ljnelson, line 161: I use the `assert` statement in this file carefully. Please let me keep it. Please.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        }
+    }
+
+
+    /*
+     * Inner and nested classes.
+     */
+
+
+    static record Association(BranchState branchState,
```
ljnelson, line 706: Holds a `Connection` together with a branch association state and information needed to restore its initial state.  Associations are immutable and logically move through a state machine defined by the XA specification.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        public boolean suspended() {
+            assert this.suspended ? this.branchState() == BranchState.IDLE : true;
+            return this.suspended;
+        }
+
+        private Association activeToIdle() {
+            if (!this.suspended()) {
+                switch (this.branchState()) {
+                case ACTIVE:
+                    // OK; end(*) was called and didn't fail with an XAER_RB* code and we are not suspended
```
ljnelson, line 772: You'll see notation like `end(*)`: that's shorthand for `end(<any flags>)`, i.e. `end` was called with any possible legal flag, i.e. `TMSUCCESS` or what have you.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+                    // Active     -> Idle       (S1 -> S2)
+                    if (LOGGER.isLoggable(Level.FINE)) {
+                        LOGGER.logp(Level.FINE, Association.class.getName(), "activeToIdle",
+                                    "Transitioning Association ({0}) from state ACTIVE to state IDLE", this);
+                    }
+                    return new Association(BranchState.IDLE,
+                                           this.xid(),
+                                           false,
+                                           this.connection(),
+                                           this.priorAutoCommit());
```
ljnelson, line 784: (We transition an association by creating a new one in the new state.)

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+        }
+
+        private Association activeToIdle() {
+            if (!this.suspended()) {
+                switch (this.branchState()) {
+                case ACTIVE:
+                    // OK; end(*) was called and didn't fail with an XAER_RB* code and we are not suspended
+                    //
+                    // Associated -> Associated (T1 -> T1; unchanged)
+                    // Active     -> Idle       (S1 -> S2)
```
ljnelson, line 775: These rules are drawn from the XA specification's state machine tables.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+     * Static fields.
+     */
+
+
+    private static final Logger LOGGER = Logger.getLogger(LocalXAResource.class.getName());
+
+    private static final Xid[] EMPTY_XID_ARRAY = new Xid[0];
+
+    // package-protected for testing only.
+    static final ConcurrentMap<Xid, Association> ASSOCIATIONS = new ConcurrentHashMap<>();
```
ljnelson, line 79: If you're reading sequentially, this may not make sense, but you can come back to it later.

This `XAResource` implementation properly supports suspension and resuming of `Connection`s' participation in global transactions. Because these operations can occur on any transaction manager thread, this `Map` is a `ConcurrentMap`.  Because they do not in fact involve any state at all connected with the `LocalXAResource` instance itself, this `Map` can be `static`.

The `Association` class is an inner record found at the bottom of this file.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
```diff
@@ -0,0 +1,1031 @@
...
+
+    /*
+     * Constructors.
+     */
+
+
+    /**
+     * Creates a new {@link LocalXAResource}.
+     *
+     * @param connectionFunction a {@link Function} that accepts a {@link Xid} (supplied by the {@link #start(Xid, int)}
```
ljnelson, line 100: Technically a connection doesn't have to be supplied at construction time, because it won't be used unless `start` is called.  So this implementation allows you to pass a `Function` that will be called by the `start` method, if and when it is called.

tjquinno (commented):
A few questions/thoughts.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
ljnelson: "Remapping" in the sense documented here: https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/concurrent/ConcurrentMap.html#compute(K,java.util.function.BiFunction)

I use [`ConcurrentMap#compute(Object, BiFunction)`](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/util/concurrent/ConcurrentMap.html#compute(K,java.util.function.BiFunction)) because for some invocations of, say, `start` the `Association` (the value) will not exist, and will need to be created, and in other invocations the `Association` will exist (i.e. in the case of joining).

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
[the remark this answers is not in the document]
ljnelson: Indeed. Most of the history of this extremely bizarre interface are lost to the mists of time but it is very clear it arose from the (C-centric) XA specification, where `flags` is indeed, well, flags: you do bitwise operations against them, they _are_ a group of orthoganal indicators of state, etc. The exceedingly strange C-ish-sort-of-ported-to-Java documentation says: "flags - One of TMNOFLAGS, TMJOIN, or TMRESUME".  Maybe they meant "after performing bitwise operations on the supplied `int`"? but they don't say that; in general the "specification" of these `javax.transaction.xa.*` classes is poor and full of mistakes.

So they are `int` constants.  There should, of course, have instead been a `start` method, a `join` method, a `suspend` method, a `resume` method, and so on and so on, but sadly this C-sort-of-ported-to-Java signature is what we have to work with.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/LocalXAResource.java
tjquinno: Ah, OK. I was trying to make "remapping" fit the vocabulary of XA rather than of the data structure API you're using.

tjquinno (approved):
The notes were a nice guided tour. LGTM.

PR #5748, opened 2022-12-21
#5717 - Nima WebServer.stop() is slow / not timely when there are no active requests

rbygrave:
This branch current has 2 tests and 2 approaches to solving this issue.  

The latest commit solves the issue using a modified/hacked version of `ThreadPerTaskExecutor` and a new interface `InterruptibleTask` (and this approaches points to trying to get a similar change to the JDK). 

The prior solution used application code to maintain the active `Http1Connection` which duplicates functionality of `ThreadPerTaskExecutor` and as such doesn't appeal as an elegant solution.


tomas-langer:
This should be fixed by #5876, I think this PR no longer makes sense.

Issue #5396, opened 2022-11-13
Nima: MicroMeter now contains Nima and SE integration


tomas-langer:
MicroMeter integration now has both Reactive and Níma implementation in the same module - needs to be refactored

PR #5768, opened 2023-01-02
Nima: MicroMeter now contains Nima and SE integration

jbescos:
https://github.com/helidon-io/helidon/issues/5396


tjquinno (changes requested):
We need to avoid duplicated code.
   
1. For example, instead of splitting the original `MeterRegistryFactory` into two nearly-identical classes, one for `Nima` and one for `Reactive`, we should be able to keep the original but perhaps parameterize it with the handler type and request type. 
2. Similarly, at a quick glance it looks as if the only code in the original `MicrometerPrometheusRegistrySupport` that is different for Nima and reactive are the two `requestToHandler` functions. Maybe that class can be parameterized and when it is instantiated it can be passed a `Function<MeterRegistry, H>` where `H` is the Nima or reactive `Handler` type.

Let's start with those--and any other similar opportunities you see--to reduce the copied code and see where that leads.

jbescos:
> We need to avoid duplicated code.
> 
>     1. For example, instead of splitting the original `MeterRegistryFactory` into two nearly-identical classes, one for `Nima` and one for `Reactive`, we should be able to keep the original but perhaps parameterize it with the handler type and request type.
> 
>     2. Similarly, at a quick glance it looks as if the only code in the original `MicrometerPrometheusRegistrySupport` that is different for Nima and reactive are the two `requestToHandler` functions. Maybe that class can be parameterized and when it is instantiated it can be passed a `Function<MeterRegistry, H>` where `H` is the Nima or reactive `Handler` type.
> 
> 
> Let's start with those--and any other similar opportunities you see--to reduce the copied code and see where that leads.

The problem of these factories is that the rely in static methods, and that makes difficult to have an upper abstract factory with common functionality. I will review it again to see if we can, at least, have some common functionality by composition.

jbescos:
> We need to avoid duplicated code.
> 
>     1. For example, instead of splitting the original `MeterRegistryFactory` into two nearly-identical classes, one for `Nima` and one for `Reactive`, we should be able to keep the original but perhaps parameterize it with the handler type and request type.
> 
>     2. Similarly, at a quick glance it looks as if the only code in the original `MicrometerPrometheusRegistrySupport` that is different for Nima and reactive are the two `requestToHandler` functions. Maybe that class can be parameterized and when it is instantiated it can be passed a `Function<MeterRegistry, H>` where `H` is the Nima or reactive `Handler` type.
> 
> 
> Let's start with those--and any other similar opportunities you see--to reduce the copied code and see where that leads.

I did it for `MicrometerPrometheusRegistrySupport`, but it is not that simple for the factories in a first view, I will try it further. However the factories have been refactored, so there is no copy paste (except for statics).

tjquinno:
After looking at your updates and thinking more about the Micrometer integration as a whole I think we an clean things up quite a bit by approaching things a bit differently.

## Some background

The class-level JavaDoc for `MeterRegistryFactory` explains why the original Micrometer integration needed to accept _enrollments_ which associate handlers with meter registries: Different registry types have different ways of producing output when an HTTP request arrives at `/micrometer`, so each Micrometer meter registry added at runtime had to have a handler factory function associated with it. Each function would accept the incoming request and return an `Optional<Handler>` containing that function's handler if it was able to respond to that request or `empty` if that function's handler could not process that request.

It might have been better if the original integration code recorded these enrollments on the `MicrometerBuiltInRegistrySupport` object rather than the `MeterRegistryFactory` object. It didn't make much difference originally, but it is complicating things now that we have to deal with Nima and reactive.

## Next step for this PR

Given where we are now, I suggest refactoring the code in this PR following the above approach.

Specifically:

1. `MeterRegistryFactory` would _not_ be parameterized for the `REQ`, `RESP`, or `HAND` types. It would still accept enrollments of registries, but not along with the handlers. Because it would deal only with meter registries, not with requests, responses, handlers, or functions returning handlers, it would not need type parameters. There would be no Nima- or reactive-specific subclasses of `MeterRegistryFactory`.
2. `MicrometerBuiltInRegistrySupport` would continue to be parameterized (perhaps adding `RESP` to the list of type parameters it already has). This class, instead of the registry factory, would accept enrollments of `MeterRegistry`/`HAND` pairs. It would enroll the meter registries (not with the `HAND`) with `MeterRegistryFactory` (because that class will still need to know about all meter registries that are enrolled). And this class would manage the list of `Enrollment` (the `MeterRegistry`/`HAND` pairs).
3. The code that currently enrolls registries and handlers with the meter registry factory would instead enroll them with the appropriate concrete subclass of `MicrometerBuiltinRegistrySupport` (either `MicrometerFeature` for Nima or `MicrometerSupport` for reactive).

jbescos:
> After looking at your updates and thinking more about the Micrometer integration as a whole I think we an clean things up quite a bit by approaching things a bit differently.
> ## Some background
> 
> The class-level JavaDoc for `MeterRegistryFactory` explains why the original Micrometer integration needed to accept _enrollments_ which associate handlers with meter registries: Different registry types have different ways of producing output when an HTTP request arrives at `/micrometer`, so each Micrometer meter registry added at runtime had to have a handler factory function associated with it. Each function would accept the incoming request and return an `Optional<Handler>` containing that function's handler if it was able to respond to that request or `empty` if that function's handler could not process that request.
> 
> It might have been better if the original integration code recorded these enrollments on the `MicrometerBuiltInRegistrySupport` object rather than the `MeterRegistryFactory` object. It didn't make much difference originally, but it is complicating things now that we have to deal with Nima and reactive.
> ## Next step for this PR
> 
> Given where we are now, I suggest refactoring the code in this PR following the above approach.
> 
> Specifically:
> 
>     1. `MeterRegistryFactory` would _not_ be parameterized for the `REQ`, `RESP`, or `HAND` types. It would still accept enrollments of registries, but not along with the handlers. Because it would deal only with meter registries, not with requests, responses, handlers, or functions returning handlers, it would not need type parameters. There would be no Nima- or reactive-specific subclasses of `MeterRegistryFactory`.
> 
>     2. `MicrometerBuiltInRegistrySupport` would continue to be parameterized (perhaps adding `RESP` to the list of type parameters it already has). This class, instead of the registry factory, would accept enrollments of `MeterRegistry`/`HAND` pairs. It would enroll the meter registries (not with the `HAND`) with `MeterRegistryFactory` (because that class will still need to know about all meter registries that are enrolled). And this class would manage the list of `Enrollment` (the `MeterRegistry`/`HAND` pairs).
> 
>     3. The code that currently enrolls registries and handlers with the meter registry factory would instead enroll them with the appropriate concrete subclass of `MicrometerBuiltinRegistrySupport` (either `MicrometerFeature` for Nima or `MicrometerSupport` for reactive).

I am with this, but I found one design blocker, related to the Builder and having a default instanced factory.

When you make an instance of `MeterRegistryFactory` using an empty builder, it adds the default `MicrometerPrometheusRegistrySupport`. This is a problem because `MicrometerPrometheusRegistrySupport` has knowledge about REQ and HAND, so the builder requires few new methods to set a function that is able to give instances of this type.

This cannot be set if the factory has a default instance, unless you read what is in modulepath to load that one. This is not clean, because there could be both implementations (Nima and Reactive). What do you load first?.

Having 2 different factories (one for Nima and other for Reactive) gets rid of this issue.

integrations/micrometer/nima/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+    <description>
+        Helidon Micrometer Integration Nima
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.integrations.micrometer</groupId>
+            <artifactId>helidon-integrations-micrometer</artifactId>
+        </dependency>
+                <dependency>
```
tomas-langer, line 44: wrong formatting

integrations/micrometer/nima/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata</artifactId>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata-processor</artifactId>
```
tomas-langer, line 71: Annotation processors should be added through maven compiler plugin annotation processors

tomas-langer (changes requested):

integrations/micrometer/nima/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+                        <arg>-Xlint:-requires-automatic</arg>
+                    </compilerArgs>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
```
tomas-langer, line 87: As you are adding `config-metadata`, I guess you also need the `config-metadata-processor` here, as otherwise the annotations are not handles and the configuration metadata file is not generated

tomas-langer (changes requested):

integrations/micrometer/nima/pom.xml
jbescos: Done

jbescos:
I am back to this.

To have an unique factory, I am going to check the classpath/modulepath to load one strategy or other (nima or reactive). As @tjquinno said, the developer should include the correct dependency.

jbescos:
Closing this because there is another PR: https://github.com/helidon-io/helidon/pull/7090

PR #5790, opened 2023-01-06
Adds connection unwrapping abilities to CDISEPlatform.java

ljnelson:

### DIFF
--- a/integrations/cdi/eclipselink-cdi/pom.xml
+++ b/integrations/cdi/eclipselink-cdi/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2019, 2022 Oracle and/or its affiliates.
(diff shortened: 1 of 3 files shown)


ljnelson:
This pull request augments the `CDISEPlatform` class to unwrap connections.

Eclipselink has a bunch of scaffolding to perform various unwrapping-type operations on connections to work around the Oracle driver's blind casting of a `Connection` to an `OracleConnection` in certain date-and-time-related scenarios.  The existing default behavior that Eclipselink implements is itself flawed: it "unwraps" a `Connection` by calling `connection.getMetaData().getConnection()` and assumes that the `Connection` so returned will be different from the `connection` whose `getMetaData()` method was called.  This is actually ruled out by the JDBC contract.

This PR simply checks to see if the incoming `Connection` that should be "unwrapped" is a `DelegatingConnection`, and, if so, calls `super.unwrapConnection(d)` on its delegate `d`.

barchetta (approved):

PR #5793, opened 2023-01-06
[3.x] Adapt to SnakeYAML 2.0 changes

### LINKED ISSUE #5792: [3.x] Prepare for SnakeYAML 2.0
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
SnakeYAML 2.0 (not yet released but coming) contains backward-incompatible changes we need to accommodate.

See also #5823 for an analysis of Helidon's usage of SnakeYAML.

## Steps to reproduce
Try to build `helidon-config-yaml`, `helidon-config-yaml-mp`, or `helidon-openapi`.

tjquinno:
Resolves #5792 

SnakeYAML 2.0 is now released.

This PR updates Helidon's own code to work with either 2.0 or 1.32.

It includes changes to config and OpenAPI, as well as an integration test verifying that config and OpenAPI work with 1.32.

A change in SnakeYAML from 1.32 to 2.0 now logs many `WARNING` messages (they were `FINE` in 1.32). To my mind, these are due to a shortcoming in SnakeYAML, not in how our code uses it. I have asked about this in the Google group for SnakeYAML but have not heard back yet. In the meantime, I've included code in this PR to change the logging level for the relevant SnakeYAML logger to `SEVERE` and added a config setting with which a user could disable our warning suppression to see those and other more important warnings if necessary.

Most of the other changes are so our code uses more secure APIs in SnakeYAML and works with either 1.32 or 2.0.


config/yaml-mp/src/main/java/io/helidon/config/yaml/mp/YamlMpConfigSource.java
```diff
@@ -380,7 +381,7 @@ public String getName() {
     static Map toMap(Reader reader) {
         // the default of Snake YAML is a Map, safe constructor makes sure we never deserialize into anything
         // harmful
-        Yaml yaml = new Yaml(new SafeConstructor());
+        Yaml yaml = new Yaml(new SafeConstructor(new LoaderOptions()));
```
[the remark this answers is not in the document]
tjquinno: Yup. I checked before making the change. 

The two key security-related aspect of `LoaderOptions` are:
* the `tagInspector` which defaults to an `UntrustedTagInspector; it prohibits global tags (the source of a vulnerability). Letting the default `UntrustedTagInspector` prevails is the safer choice.
* the `allowRecursiveKeys` setting which defaults to `false`.

See [src/main/java/org/yaml/snakeyaml/LoaderOptions.java](https://bitbucket.org/snakeyaml/snakeyaml/src/master/src/main/java/org/yaml/snakeyaml/LoaderOptions.java)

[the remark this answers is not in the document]
barchetta: Roger!

barchetta:
We definitely want to do this upgrade for 4.x. But the SafeConstructor() change seems like an incompatibility that might not be appropriate for Helidon 2.x and 3.x if a customer is using SnakeYaml in  their application code (and quite possibly using `new SafeConstructor()`).  For 2.x/3.x we could continue to use 1.X and re-verify our use of SafeConstructor() is secure by default. Something to discuss.

openapi/src/main/java/io/helidon/openapi/Serializer.java
```diff
@@ -111,6 +111,7 @@ static class CustomRepresenter extends Representer {
         CustomRepresenter(Map<Class<?>, ExpandedTypeDescription> types,
                 Map<Class<?>, ExpandedTypeDescription> implsToTypes, DumperOptions dumperOptions,
                 DumperOptions.ScalarStyle stringStyle) {
+            super(dumperOptions);
             this.implsToTypes = implsToTypes;
             this.dumperOptions = dumperOptions;
```
ljnelson, line 116: (Do you still need this?)

tjquinno: Yes. The no-args constructor in the superclass was removed.

ljnelson: Right, that's what I meant (the `private` field).

barchetta (commented):
This seems to be missing the upgrade to SnakeYaml 2.0?

And when I run the test app with SnakeYaml 2.0 (skipping tests, just running testapp) I get a lot of warnings like:
```
2023.03.07 13:00:46 WARNING org.yaml.snakeyaml.introspector Thread[main,5,main]: Failed to find field for org.eclipse.microprofile.openapi.models.media.Schema.enum
2023.03.07 13:00:46 WARNING org.yaml.snakeyaml.introspector Thread[main,5,main]: Failed to find [getExtensions(0 args)] for org.eclipse.microprofile.openapi.models.media.XML.extensions
```

tjquinno:
I accidentally omitted the change to `dependencies/pom.xml` from my commit.

I've reproduced the warnings and am investigating.

tjquinno:
The warnings are appearing now from the SnakeYAML `PropertySubstitute` class because the logging level for these messages was changed from `FINE` in 1.x to `WARNING` in 2.0 as part of converting from `java.util.logging` to SnakeYAML's own logging facility. 

Similar warnings now appear when, for example, building our QuickStart app (which has OpenAPI support in it). 

The good news is that the internal model is being built correctly, despite these warnings, but the messages are alarming. I'm in the process of looking for exactly why these messages are appearing despite the model being built properly. Once we know that, we can look for a way to avoid the messages.

tjquinno:
I have found the immediate cause of the messages. It's not yet clear what the best workaround will be. 

See https://groups.google.com/g/snakeyaml-core/c/JOzKbNLorHs/m/K2XO2on7AgAJ where I explained the cause and have asked about a solution.

tjquinno:
The MP OpenAPI TCK fails because it uses Jackson dataformat which in turn uses SnakeYAML, in particular at least one 1.x constructor removed in 2.0:
```
java.lang.NoSuchMethodError: 'void org.yaml.snakeyaml.parser.ParserImpl.<init>(org.yaml.snakeyaml.reader.StreamReader)'
	at com.fasterxml.jackson.dataformat.yaml.YAMLParser.<init>(YAMLParser.java:178)
```

Jackson has updated to use SnakeYAML 2.0  (https://github.com/FasterXML/jackson-dataformats-text/pull/390) but has not yet issued a release containing the fix. It seems targeted for 2.15 which is forecast to have a release candidate in March 2023. https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.15

spericas (approved):

barchetta (approved):

Sineaggi:
> A change in SnakeYAML from 1.32 to 2.0 now logs many WARNING messages (they were FINE in 1.32).

The mapping should be the same. SnakeYaml 2 is now an MR jar with a module-info and a custom logging implementation on java 9+ (switches from JUL to the java platform logger).

 I based the change off of FINE (in JUL) having a severity of 500 and DEBUG (in JPL) having a severity of 500.

PR #5812, opened 2023-01-09
[3.x] Fix improper handling of metrics global tags

### LINKED ISSUE #5791: Metrics incorrectly uses top-level config, not `metrics`-level, in setting global tags
MetricsCdiExtension fails when `tags` env. property is present.

## Environment Details
* Helidon Version: 3.1.0
* Helidon MP
* JDK version: 17
* OS: Linux

----------

## Problem Description
We have `tags` environment in our pipeline, which leads to a following error in tests:
```
java.lang.IllegalArgumentException: Error(s) in global tag expression: [Missing '=': ]
	at io.helidon.metrics.api.MetricsSettingsImpl$Builder.globalTagsExpressionToMap(MetricsSettingsImpl.java:219)
	at io.helidon.config.ConfigValues.lambda$create$4(ConfigValues.java:146)
	at io.helidon.config.ConfigValues$GenericConfigValueImpl.asOptional(ConfigValues.java:233)
	at io.helidon.config.ConfigValue.ifPresent(ConfigValue.java:227)
	at io.helidon.metrics.api.MetricsSettingsImpl$Builder.config(MetricsSettingsImpl.java:161)
	at io.helidon.metrics.api.MetricsSettingsImpl$Builder.config(MetricsSettingsImpl.java:95)
	at io.helidon.metrics.api.MetricsSettings.create(MetricsSettings.java:53)
	at io.helidon.metrics.serviceapi.MetricsSupport.create(MetricsSupport.java:89)
	at io.helidon.servicecommon.restcdi.HelidonRestCdiExtension.registerService(HelidonRestCdiExtension.java:245)
	at io.helidon.microprofile.metrics.MetricsCdiExtension.registerService(MetricsCdiExtension.java:736)
``` 

The reason is that `MetricsCdiExtension.seComponentConfig` is constructing a Config that also contains system properties and env variables, so our `tags` env variable is interpreted as `metrics.tags` configuration property.

## Steps to reproduce
Run a microservice with empty tags env. property.

tjquinno:
Resolves #5791 

MP Metrics requires that the config setting`mp.metrics.tags` (format `tag1=value1,tag2=value2")` should cause MP metrics to assign the specified  "global tags" to decorate every `MetricID` during output. 

In our SE implementation, we follow the same approach except with `metrics.tags` as the config key.

Investigating the issue revealed one error and one not-great practice.

1. Error: `MetricsSettings`, (in SE), when using the default config to create an instance, used `Config.create()` without also using `.get("metrics")` to obtain the metrics-specific assignments. As a result, a top-level config key of "tags" (the issue described using an environment variable but a setting at the top level of any config source would show the problem) caused `MetricsSettings` to try to use _that_ as the global metrics tag settings. The value was not formatted as the code expected so it threw an exception.
2. Not-great practice: In `MetricsCdiExtension`, we have to merge settings from `mp.metrics` and `metrics` if both are present to create the metrics settings we use in preparing the `MetricsSupport` object. The code does this by creating a temporary `Config` object with config sources from the two config nodes at `mp.metrics` and `metrics`. But it did so by using `Config.builder()...` without disabling the built-in env. var. and system property sources. This should not have affected anything, but it was confusing because the `mp.metrics` and `metrics` sections of config used to create the temporary `Config` object would already have accounted for env. vars and properties. 

The code contains fixes in these two spots plus new tests that run in separate JVMs to avoid cross-contaminating the test configs with the tag settings. (I probably could have used custom `Config` objects in the tests to simplify the test executions but I wanted to reproduce the conditions reported in the issue, which used an env. var.)

### DIFF
--- a/metrics/api/src/main/java/io/helidon/metrics/api/MetricsSettings.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/MetricsSettings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2021, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,12 +33,12 @@ import org.eclipse.microprofile.metrics.MetricRegistry;
 public interface MetricsSettings {
 
     /**
-     * Returns default metrics settings based on default config.
+     * Returns default metrics settings based on the {@code metrics} section of the default config.
      *
-     * @return new settings reflecting the default config
+     * @return new settings reflecting the default metrics config
      */
     static MetricsSettings create() {
-        return create(Config.create());
+        return create(Config.create().get(Builder.METRICS_CONFIG_KEY));
     }
 
     /**
--- a/metrics/api/pom.xml
+++ b/metrics/api/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    Copyright (c) 2021, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2021, 2023 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


metrics/api/src/main/java/io/helidon/metrics/api/MetricsSettings.java
```diff
@@ -33,12 +33,12 @@
...
 
     /**
-     * Returns default metrics settings based on default config.
+     * Returns default metrics settings based on the {@code metrics} section of the default config.
      *
-     * @return new settings reflecting the default config
+     * @return new settings reflecting the default metrics config
      */
     static MetricsSettings create() {
-        return create(Config.create());
```
barchetta, line 41: How incompatible is this change? I'm concerned that users will have defined a top-level config key of "tags" to get metrics global tags to work -- or maybe that is unlikely.

tomas-langer (approved):

PR #4003, opened 2022-03-24
Add Helidon Metrics integration with OCI

klustria:
The goal of this feature is to be able to harvest Helidon generated metrics and send to OCI monitoring service.This can be enabled in Helidon MP by adding the dependency (io.helidon.integrations.oci:helidon-integrations-oci-metrics-cdi) and setting a similar configuration in application.yaml like below:
```yaml
ocimetrics:
  # OCI compartment ID where the metrics will be sent
  compartmentId: ocid1.compartment.oc1..aaaaaaaalc3tlb3xcqpsbyvp72po6ybqopvycf7wdvl6gjpa7mry77pdkgza
  # OCI metric namespace
  namespace: helidon_pegasus
  # Delay in seconds before the 1st metrics transmission to OCI takes place.  Defaults to 1 second if not specified.
  initialDelay: 30
  # Interval in seconds between metrics transmission to OCI. Defaults to 60 seconds if not specified.
  delay: 180
  # Filter only the scopes that will be sent to OCI. This is optional and will default to all scopes if not specified.
  scopes: [base, vendor, application]
  # Enable or disable metric transmission to OCI. Defaults to true if not specified.
  enabled: true
```

PR #5829, opened 2023-01-11
Add Helidon Metrics integration with OCI

klustria:
The change includes the following:
1. Port of PR https://github.com/helidon-io/helidon/pull/4003 that adds the Helidon Metrics to OCI integration
2. Port of PR https://github.com/helidon-io/helidon/pull/4897 that fixes race condition in the unit test
3. Adjust code to deal with MP metrics API changes
4. Change endpoint to the ingestion endpoint when posting the metrics as this is not handled anymore by the OCI SDK integration due to changes in the OCI Java SDK v3.
5. Change MonitoringClient.class to Monitoring.class for mocking using Mockito in the unit test as the OCI Java SDK v3 converted some of the methods in MonitoringClient as Final making them difficult to mock.
6. Trim the OCI Metadata value which contains the metric description if the value exceeds 256 characters, otherwise it will fail.
7. OCI Monitoring service only allows a maximum of 50 metrics per posting, hence additional configuration parameters were added to control sending metrics in batches. The configuration parameters are: a. batchSize - Maximum no. of metrics to send in a batch. Defaults to 50 which is what OCI allows b. batchDelay - Interval between batch posting For example if there are 51 metrics and batchSize is set to 25 and batchDelay to 5 seconds, the Helidon metric integration module will divide the posting to 3 batches sending 25 metrics on the 1st and 2nd batches and 1 metric on the 3rd batch with 5 seconds interval between batch posting.
8. Refactor OciMetricsCdiExtension to add a new bean (OciMetricsBean) to handle the Observer method which will inject Monitoring. Previous code of OciMetricsCdiExtension cannot independently handle instantiation of Monitoring client via CDI.
9. Add unit tests to verify batch posting feature and the use of ingestion endpoint.


integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+// This bean is added to handle injection on the ObserverMethod as it does not work on an Extension class.
+@Singleton
+public class OciMetricsBean {
+
+    // Make Priority higher than MetricsCdiExtension so this will only start after MetricsCdiExtension has completed.
+    void registerOciMetrics(@Observes @Priority(LIBRARY_BEFORE + 20) @Initialized(ApplicationScoped.class) Object ignore,
+                            Monitoring monitoringClient) {
+        Config helidonConfig = MpConfig.toHelidonConfig(ConfigProvider.getConfig()).get("ocimetrics");
```
ljnelson, line 44: Consider just injecting the root Helidon `Config`, so:
```
void registerOciMetrics(@Observes @Priority(LIBRARY_BEFORE + 20) @Initialized(ApplicationScoped.class) Object event, Monitoring monitoringClient, io.helidon.config.Config helidonConfig) {
    helidonConfig = helidonConfig.get("ocimetrics");
    if (helidonConfig.exists()) {
        // ...
    }
}
```

klustria: I did try this before when you first suggested it, but gave me this error:
```
WELD-001408: Unsatisfied dependencies for type Config with qualifiers @Default
  at injection point [UnbackedAnnotatedParameter] Parameter 3 of [UnbackedAnnotatedMethod] io.helidon.integrations.oci.metrics.cdi.OciMetricsBean.registerOciMetrics(@Initialized @Observes @Priority Object, Monitoring, Config)
```

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
```diff
@@ -0,0 +1,527 @@
...
+                .build();
+
+        PostMetricDataRequest postMetricDataRequest = PostMetricDataRequest.builder()
+                .postMetricDataDetails(postMetricDataDetails)
+                .build();
+
+        LOGGER.finest(String.format("Pushing %d metrics to OCI", metricDataDetailsList.size()));
+        if (LOGGER.isLoggable(Level.FINEST)) {
+            metricDataDetailsList
+                    .stream()
```
tjquinno, line 215: `List` has its own `forEach`. No stream is needed here.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
```diff
@@ -0,0 +1,527 @@
...
+                                m.getDatapoints().get(0).getTimestamp(),
+                                m.getDatapoints().get(0).getValue(),
+                                m.getMetadata()));
+                    });
+        }
+        String originalMonitoringEndpoint = this.monitoringClient.getEndpoint();
+        try {
+            // Use the ingestion endpoint for posting
+            this.monitoringClient.setEndpoint(
+                    monitoringClient.getEndpoint().replaceFirst("telemetry\\.", "telemetry-ingestion."));
```
tjquinno, line 232: Are the paths always these particular strings? Is there any time they might be different and, therefore, need to be configurable on the Helidon integration side?

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
```diff
@@ -0,0 +1,527 @@
...
+            LOGGER.warning(String.format("Unable to send metrics to OCI: %s", e.getMessage()));
+        } finally {
+            // restore original endpoint
+            this.monitoringClient.setEndpoint(originalMonitoringEndpoint);
+        }
+    }
+
+    @Override
+    public void update(Routing.Rules rules) {
+        rules.onNewWebServer(this::prepareShutdown);
```
tjquinno, line 245: Does it make sense to move this to after line 255? If we don't need the executor (because the feature is disabled or there are no metrics to report) then we won't need the shutdown processing.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
```diff
@@ -0,0 +1,527 @@
...
+        }
+
+        /**
+         * Sets the initial delay before metrics are sent to OCI
+         * (defaults to {@value #DEFAULT_SCHEDULER_INITIAL_DELAY}).
+         *
+         * @param value initial delay, expressed in time units set by {@link #schedulingTimeUnit(TimeUnit)}
+         * @return updated builder
+         */
+        @ConfiguredOption
```
tjquinno, line 320: For this and other configurable items, if there is a default value consider adding the `value` attribute (which is used to convey the default value) to the annotation so the default value is highlighted as such in the generated config documentation.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
```diff
@@ -0,0 +1,527 @@
...
+        public Builder initialDelay(long value) {
```
tjquinno, line 321: The developer could invoke this and other builder methods directly, so they might need to have something like  `Objects.requireNonNull` invocations to check the parameter value before proceeding.

tjquinno (changes requested):
A few notes.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
klustria: Implementing

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
klustria: Implemented on methods with non-primitive type parameters

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
klustria: Implementing.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
klustria: implementing

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsSupport.java
klustria: Yeah, that is a fix value: https://docs.oracle.com/en-us/iaas/api/#/en/monitoring/20180401/. We use to do the same magically via the filter of the [OCI client's sdk](https://github.com/helidon-io/helidon/blob/helidon-2.x/integrations/oci/sdk/cdi/src/main/java/io/helidon/integrations/oci/sdk/cdi/OciExtension.java#L1378), but that will work anymore in OCI Java SDK v3.

integrations/oci/metrics/cdi/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.integrations.oci.metrics</groupId>
+            <artifactId>helidon-integrations-oci-metrics</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
klustria: Good catch, this may be a result of copying from `integrations/oci/metrics/metrics/pom.xml`where it is needed. Removing did not affect the build.

integrations/oci/metrics/metrics/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+            <artifactId>helidon-webserver</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.oracle.oci.sdk</groupId>
+            <artifactId>oci-java-sdk-monitoring</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
klustria: implemented.

PR #5849, opened 2023-01-12
Use Hamcrest assertions instead of JUnit in examples/metrics/examplar(#1749)

Captain1653:
Part of #1749

I didn't touch [org.junit.jupiter.api.Assertions.assertLinesMatch](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertLinesMatch-java.util.List-java.util.List-). It should be there because of `otherwise treat expected as a regular expression and check via`. Am I wrong?

[Backport 2.x](https://github.com/helidon-io/helidon/pull/5863)

### DIFF
--- a/examples/metrics/exemplar/pom.xml
+++ b/examples/metrics/exemplar/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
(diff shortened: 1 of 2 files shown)


romain-grecourt (approved):
Yes, let's keep assertLinesMatch.

PR #5856, opened 2023-01-13
Oidc tenant name now properly escaped 

### LINKED ISSUE #5855: Oidc fails with 'Not a valid authorization code' error
When we try to access OIDC identity server, without providing any tenant we want to use (such as using it as a single tenant),
Helidon chooses default tenant configuration. This tenant has name `@default`. This name is now propagated as the tenant to be used. It is part of the redirection URI we receive back from the OIDC identity server (in my case KeyCloak when tested)

In TenantAuthenticationHandler#297 is the following code:
```java
String redirectUri =
        encode(redirectUri(providerRequest.env()) + "?" + oidcConfig.tenantParamName() + "=" + tenantId);
```

This redirectUri is encoded to be part of the query params for OIDC identity server during initial request, but it its tenant name stops being encoded when sent back to the client. That means, there is suddenly  `@default` as a query parameter even though it should be encoded.

This can be fixed by the following line:
```java
String redirectUri =
        encode(redirectUri(providerRequest.env()) + "?" + encode(oidcConfig.tenantParamName()) + "=" + encode(tenantId));
```

Verdent:
Fixes #5855

### DIFF
--- a/security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
+++ b/security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
@@ -294,7 +294,8 @@ class TenantAuthenticationHandler {
             String authorizationEndpoint = tenant.authorizationEndpointUri();
             String nonce = UUID.randomUUID().toString();
             String redirectUri =
-                    encode(redirectUri(providerRequest.env()) + "?" + encode(oidcConfig.tenantParamName()) + "=" + encode(tenantId));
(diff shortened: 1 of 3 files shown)


danielkec (dismissed):

danielkec (approved):

klustria:
I have built this change and tested both locally, remotely and remotely with proxy and validated that this fixed the OIDC issue that we observed in v2.5.5 as reported in https://github.com/helidon-io/helidon/issues/5855.

klustria (approved):
LGTM too

PR #5920, opened 2023-01-20
Archetype generates wrong Jaeger configuration (SE)

### LINKED ISSUE #5902: 3.x archetype generates wrong Jaeger configuration
Helidon SE 3.1.0 started

The following is generated:
```yaml
tracing:
    service: "helidon-full-http"
    protocol: "https"     # JAEGER_ENDPOINT (if not udp, http is expected and endpoint is filled)
    host: "192.168.1.3"   # JAEGER_ENDPOINT
    port: 14240           # JAEGER_ENDPOINT
    path: "/api/traces/mine"   # JAEGER_ENDPOINT
    token: "token"        # JAEGER_AUTH_TOKEN
    # Either token or username/password
    #username:  "user"     # JAEGER_USER
    #password: "pass"      # JAEGER_PASSWORD
    propagation: "jaeger" # JAEGER_PROPAGATION either "jaeger" or "b3"
    log-spans: false      # JAEGER_REPORTER_LOG_SPANS
    max-queue-size: 42    # JAEGER_REPORTER_MAX_QUEUE_SIZE
    flush-interval-ms: 10001 # JAEGER_REPORTER_FLUSH_INTERVAL
    sampler-type: "remote"# JAEGER_SAMPLER_TYPE (https://www.jaegertracing.io/docs/latest/sampling/#client-sampling-configuration)
    sampler-param: 0.5    # JAEGER_SAMPLER_PARAM (number)
    sampler-manager: "localhost:47877" # JAEGER_SAMPLER_MANAGER_HOST_PORT
    tags:
      tag1: "tag1-value"  # JAEGER_TAGS
      tag2: "tag2-value"  # JAEGER_TAGS
    boolean-tags:
      tag3: true          # JAEGER_TAGS
      tag4: false         # JAEGER_TAGS
    int-tags:
      tag5: 145           # JAEGER_TAGS
      tag6: 741           # JAEGER_TAGS
```

Supported options: https://helidon.io/docs/v3/#/config/io_helidon_tracing_jaeger_JaegerTracerBuilder

For example: the example contains `sampler-type: "remote"` which is no longer supported by Jaeger (use either `constant` or `ratio`)

Also the archetype does not register the tracer, so the configuration would be ignored

tvallin:
fix #5902

### DIFF
--- a/archetypes/helidon/src/main/archetype/se/custom/observability.xml
+++ b/archetypes/helidon/src/main/archetype/se/custom/observability.xml
@@ -31,21 +31,11 @@
                 <value if="${tracing.provider} == 'jaeger'"><![CDATA[
 tracing:
     service: "helidon-full-http"
-    protocol: "https"     # JAEGER_ENDPOINT (if not udp, http is expected and endpoint is filled)
-    host: "192.168.1.3"   # JAEGER_ENDPOINT
-    port: 14240           # JAEGER_ENDPOINT
-    path: "/api/traces/mine"   # JAEGER_ENDPOINT
+    host: "localhost"     # JAEGER_ENDPOINT
+    port: 14250           # JAEGER_ENDPOINT
     sampler-type: "ratio" # JAEGER_SAMPLER_TYPE (https://www.jaegertracing.io/docs/latest/sampling/#client-sampling-configuration)
     sampler-param: 0.5    # JAEGER_SAMPLER_PARAM (number)
-    tags:
-      tag1: "tag1-value"  # JAEGER_TAGS
-      tag2: "tag2-value"  # JAEGER_TAGS
-    boolean-tags:
-      tag3: true          # JAEGER_TAGS
-      tag4: false         # JAEGER_TAGS
-    int-tags:
-      tag5: 145           # JAEGER_TAGS
-      tag6: 741           # JAEGER_TAGS]]></value>
+]]></value>
                 <value if="${tracing.provider} == 'zipkin'"><![CDATA[
 tracing:
   zipkin:
--- a/archetypes/helidon/src/main/archetype/se/custom/observability.xml
+++ b/archetypes/helidon/src/main/archetype/se/custom/observability.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2022 Oracle and/or its affiliates.
+    Copyright (c) 2022, 2023 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


archetypes/helidon/src/main/archetype/se/custom/observability.xml
```diff
@@ -37,7 +43,7 @@ tracing:
     log-spans: false      # JAEGER_REPORTER_LOG_SPANS
     max-queue-size: 42    # JAEGER_REPORTER_MAX_QUEUE_SIZE
     flush-interval-ms: 10001 # JAEGER_REPORTER_FLUSH_INTERVAL
-    sampler-type: "remote"# JAEGER_SAMPLER_TYPE (https://www.jaegertracing.io/docs/latest/sampling/#client-sampling-configuration)
+    sampler-type: "constant"# JAEGER_SAMPLER_TYPE (https://www.jaegertracing.io/docs/latest/sampling/#client-sampling-configuration)
```
tomas-langer, line 46: Please go through the linked documentation (from issue). Some of the options are no longer available (this was just an example - there are others that are not supported, such as `sampler-manager` etc.)

archetypes/helidon/src/main/archetype/se/custom/observability.xml
```diff
@@ -21,6 +21,12 @@
                   xsi:schemaLocation="https://helidon.io/archetype/2.0 https://helidon.io/xsd/archetype-2.0.xsd">
     <output>
         <model if="${tracing}">
+            <list key="Main-helidon-imports">
+                <value>io.helidon.webserver.WebTracingConfig</value>
+            </list>
+            <list key="Main-routingBuilder">
```
tomas-langer, line 27: This is not sufficient - the tracer must be created and (ideally) registered with the server:
```
WebServer.builder()
  ...
  .tracer(TracerBuilder.create(config.get("tracing"))
                    .build())
```

tomas-langer (changes requested):

archetypes/helidon/src/main/archetype/se/custom/observability.xml
```diff
@@ -29,17 +35,8 @@ tracing:
...
-    token: "token"        # JAEGER_AUTH_TOKEN
-    # Either token or username/password
-    #username:  "user"     # JAEGER_USER
-    #password: "pass"      # JAEGER_PASSWORD
-    propagation: "jaeger" # JAEGER_PROPAGATION either "jaeger" or "b3"
-    log-spans: false      # JAEGER_REPORTER_LOG_SPANS
-    max-queue-size: 42    # JAEGER_REPORTER_MAX_QUEUE_SIZE
-    flush-interval-ms: 10001 # JAEGER_REPORTER_FLUSH_INTERVAL
-    sampler-type: "remote"# JAEGER_SAMPLER_TYPE (https://www.jaegertracing.io/docs/latest/sampling/#client-sampling-configuration)
+    sampler-type: "ratio" # JAEGER_SAMPLER_TYPE (https://www.jaegertracing.io/docs/latest/sampling/#client-sampling-configuration)
```
tomas-langer, line 38: I am not really sure what we want to accomplish here. If this is to make it easy for users to set up Jager, then we should ONLY have `host` and `port` (and maybe `sampler-type` and `sampler-param`) defined,. and the host should be `localhost` - as this will work out of the box with local Jaeger in docker.
The port should be set to `14250` (default for Jaeger)
Everything else is kind of confusing for first time user.

tomas-langer (approved):

PR #5930, opened 2023-01-20
Remove incorrect override of UriQuery.isEmpty in the impl class

### LINKED ISSUE #5929: `UriQueryImpl` incorrectly reports `isEmpty` as false always
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
The `UriQuery` interface defines a default impl of `isEmpty()` as returning `size() == 0`.

But the only implementation, `UriQueryImpl`, overrides this and always returns `false`, even if the `UriQuery` is created with a null or empty query string.

tjquinno:
Resolves #5929 

### DIFF
--- a/common/uri/src/test/java/io/helidon/common/uri/UriQueryTest.java
+++ b/common/uri/src/test/java/io/helidon/common/uri/UriQueryTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


danielkec (approved):

common/uri/src/test/java/io/helidon/common/uri/UriQueryTest.java
```diff
@@ -54,4 +54,13 @@ void testEncodedOtherChars() {
         assertThat(uriQuery.value("h"), is("xc#e<"));
         assertThat(uriQuery.value("a"), is("b&c=d"));
     }
+    
+    @Test
+    void testEmptyQueryString() {
+        UriQuery uriQuery = UriQuery.create(null);
```
tomas-langer, line 60: `UriQuery.create(null)` should not be supported and should throw a NPE.

tomas-langer (changes requested):

common/uri/src/test/java/io/helidon/common/uri/UriQueryTest.java
tjquinno: If you feel this strongly that null should not be allowed as the value for `UriQuery` then let's approve this very focused bug fix as-is and create a separate issue for null handling.

There are several places in `UriQueryImpl` (unchanged by this PR) that check for null and take relevant action, so the original author (!) seems to have considered this a valid use case and there could be code elsewhere that depends on this behavior. 

I would rather not expand the scope of _this_ bug fix to address other aspects of `UriQuery` that seem to be working as designed.

tomas-langer: The original author understands.

tomas-langer (approved):

PR #5945, opened 2023-01-23
helidon metrics to oci integration

klustria:
The change includes the following:
1. Port of v3 PR https://github.com/helidon-io/helidon/pull/5829 that includes additional changes and fixes.
2. Changes related to OciMetricsSupport use of Nima instead of SE Webserver dependency.

### DIFF
--- a/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
+++ b/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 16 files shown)


integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
```diff
@@ -0,0 +1,53 @@
...
+    // Make Priority higher than MetricsCdiExtension so this will only start after MetricsCdiExtension has completed.
+    void registerOciMetrics(@Observes @Priority(LIBRARY_BEFORE + 20) @Initialized(ApplicationScoped.class) Object ignore,
+                            Config config, Monitoring monitoringClient) {
+        Config helidonConfig =  config.get("ocimetrics");
+        if (helidonConfig.exists()) {
+            OciMetricsSupport.Builder builder = OciMetricsSupport.builder()
+                    .config(helidonConfig)
+                    .monitoringClient(monitoringClient);
+
+            RoutingBuilders.create(helidonConfig)
```
tjquinno, line 48: Probably very minor. 

Because the `ocimetrics` config can be present but `enabled` set to false, maybe add `public boolean enabled()` to `OciMetricsSupport.Builder` (if it's not there already) and have this code check that value (after creating the builder) and set up the routing only if enabled is `true`. It might not seem like much but avoiding adding an unneeded routing could conceivable help improve performance a little bit on every request.

tjquinno (commented):
One possible slight optimization suggestion.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
klustria: As discussed, I will deal with this in https://github.com/helidon-io/helidon/issues/5946

tjquinno (approved):

klustria:
https://github.com/helidon-io/helidon/pull/5945 is completed hence closing.

PR #5950, opened 2023-01-24
UriQuery should not support null parameters.

tomas-langer:
Fixed TyrusUpgrader which was not thread safe.


tjquinno (approved):
Certainly the `UriQuery` changes look good to me. I am less familiar with the other affected files.

spericas (approved):
Good catch

PR #5956, opened 2023-01-24
4.x: Upgrade OCI SDK to 3.2.1

barchetta:


romain-grecourt (approved):

klustria:
already merged but I tested with OCI Monitoring service anyway and works fine.

Issue #5946, opened 2023-01-24
Register service only when OCI metric enable flag is set to true


klustria:
## Environment Details
* Helidon Version: 2.x/3.x
* Helidon MP

----------

## Problem Description
For Helidon Metrics to OCI feature, when ocimetrics.enable is not set to true, the OciMetricsSupport service is still registered in the default routing, but not running. It would be more optimal if the service is not registered at all, so that the service won't be loaded from the start.

klustria:
PR https://github.com/helidon-io/helidon/issues/5946 is the potential fix

klustria:
PR https://github.com/helidon-io/helidon/pull/6032 has been completed.

PR #6032, opened 2023-02-01
Register OciMetricsSupport service only when enable flag is set to true

klustria:
Changes include:
1. Add a getter enabled method in OciMetricsSupport.Builder to provide boolean value indicating whether OciMetricsSupport service will be activated or not.
2. In OciMetricsBean, use the new enabled method to determine whether OciMetricsSupport service will be registered in the default routing or not.
3. Refactor routing registration of OciMetricsSupport, so it can be properly unit tested to verify if it is skipped when disabled or invoked when enabled.
4. Add unit test validation when enable flag is set to true, false, or not set (which defaults to true).


klustria:
Will deal with issue reported in https://github.com/helidon-io/helidon/issues/5946

tjquinno (approved):

Issue #5187, opened 2022-10-14
Doc for Jaeger tracing integration in 3.x: shows incorrect port as of Helidon 3.0 and maybe should discuss removed config settings


tjquinno:
## Environment Details
* Helidon Version: 3.x, (4.x?)
* Helidon SE or Helidon MP SE and MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
1. Helidon 3.x uses the Jaeger OpenTelemetry client to send information to the Jaeger server. That client uses port 14250 (correctly defaulted in the `JaegerTracerBuilder` class.
   
   But our doc for Jaeger tracing integration uses port 14240, not 14250 (or leaving it unset to use the correct default). See the config example snippets just  below the config tables at https://helidon.io/docs/v3/#/se/tracing#_configuration_options_3 and https://helidon.io/docs/v3/#/mp/tracing#_configuration_options_3.
   
   We could simply remove the port setting. The default in the code works correctly and the config doc shows `port` as a config key so readers will know they can override the default if they need to.

2. Several config keys that we supported in 2.x we no longer support because 3.x uses the OpenTelemetry tracing client rather than the native Jaeger one (as 2.x did). Should the doc contain a brief note at least listing these and/or explaining workarounds if any? For example, `propagation` is no longer supported, but `propagation: b3` is no longer needed because the Helidon's `JaegerTracerBuilder` class always includes the OpenTelemetry `B3Propagator`. 

   The class-level JavaDoc for `JaegerTracerBuilder` has changed accordingly, but we mention nothing in the doc on our website. See the `JaegerTracerBuilder` source diffs to see which keys are no longer supported. Here is a copy of the config code that shows which keys were supported in 2.x but not in 3.x:
   ```
        config.get("token").asString().ifPresent(this::token);
        config.get("username").asString().ifPresent(this::username);
        config.get("password").asString().ifPresent(this::password);
        config.get("propagation").asList(String.class).ifPresent(propagations -> {
            propagations.stream()
                    .map(String::toUpperCase)
                    .map(Configuration.Propagation::valueOf)
                    .forEach(this::addPropagation);

        });
        config.get("log-spans").asBoolean().ifPresent(this::logSpans);
        config.get("max-queue-size").asInt().ifPresent(this::maxQueueSize);
        config.get("flush-interval-ms").asLong().ifPresent(this::flushIntervalMs);
   ...
        config.get("sampler-manager").asString().ifPresent(this::samplerManager);
[...]

PR #6060, opened 2023-02-02
Doc updates for Jaeger tracing port value

ljamen:
Doc for issue #5187 and #6094

### DIFF
--- a/docs/config/io_helidon_tracing_jaeger_JaegerTracerBuilder.adoc
+++ b/docs/config/io_helidon_tracing_jaeger_JaegerTracerBuilder.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2023 Oracle and/or its affiliates.
+    Copyright (c) 2022 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -36,7 +36,7 @@ This is a standalone configuration type, prefix from configuration root: `tracin
 
 
 
-== Configuration Options
+== Configuration options
 
 
 
--- a/docs/config/io_helidon_tracing_jaeger_JaegerTracerBuilder.adoc
+++ b/docs/config/io_helidon_tracing_jaeger_JaegerTracerBuilder.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2022 Oracle and/or its affiliates.
+    Copyright (c) 2023 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


ljamen:
Hi @tjquinno  - There are a couple of ways to address this doc bug. I believe you were asking the team how to update. Did you get your answer?

Should the doc contain a brief note at least listing these and/or explaining workarounds if any? For example, propagation is no longer supported, but propagation: b3 is no longer needed because the Helidon's JaegerTracerBuilder class always includes the OpenTelemetry B3Propagator.

tjquinno:
TBH I am not sure what workarounds, if any, are available for the settings that Jaeger no longer supports. Our config properties, then and now, simply mirror whatever Helidon's code can set on Jaeger's span exporter builder. Jaeger changed and so the settings we can expose changed as well.

Given that, it's probably clearer to _not_ list the specific settings we _no longer_ support (given that, at least at the moment, I don't know what, if any, workaround there might be). Maybe, just before the `include` of the generated config `.adoc` file, we could state something like this:

> Because Jaeger changed its client implementation some Jaeger settings exposed by earlier releases of Helidon are no longer available. Please note the currently-supported settings in the table below.

or something to that effect.

docs/config/io_helidon_tracing_jaeger_JaegerTracerBuilder.adoc
```diff
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2022 Oracle and/or its affiliates.
+    Copyright (c) 2023 Oracle and/or its affiliates.
```
tjquinno, line 3: Do not change this file (or any in the `docs/config` directory) manually.

They are all generated.

docs/includes/tracing/tracer-jaeger.adoc
```diff
@@ -47,14 +47,16 @@ include::{rootdir}/includes/dependencies.adoc[]
```
tjquinno, line 47: See my comment in the PR discussion for a suggested brief note to insert here.

docs/se/tracing.adoc
```diff
@@ -203,9 +203,9 @@ tracing:
 routing.register(WebTracingConfig.create(config.get("tracing")));
 ----
 
-==== Path based configuration in Helidon Web Server
+==== Path-based configuration in Helidon Web Server
 
-For Web Server we have a path based support for configuring tracing, in addition
+For Web Server we have a path-based support for configuring tracing, in addition
```
tjquinno, line 208: Very small. Maybe remove the "a" from before "path-based" here?

tjquinno (changes requested):

al3xandru:
This commit doesn't include a fix for https://helidon.io/docs/v3/#/se/guides/tracing (the snippet using private void getDefaultMessageHandler refers to and uese the io.opentracing instead of io.helidon.tracing API)

ljamen:
Hi @al3xandru I'm working with Tim on that. This is still in draft mode until I've completed all of the updates. They will be included in this PR.

ljamen:
Closing this PR and will include files in a new PR that does not include the generated tracing doc that was accidentally updated.

Issue #6106, opened 2023-02-07
LRA TCK failing randomly


jbescos:
## Environment Details
* Helidon Version: 4.0.0-SNAPSHOT
* Helidon SE or Helidon MP: Helidon MP
* JDK version: 19.0.2
* OS: Linux Mint 21

----------

## Problem Description
org.eclipse.microprofile.lra.tck.TckRecoveryTests was failing intermitently and it was excluded in this PR:
https://github.com/helidon-io/helidon/pull/5830

We have to re-enable it. I am trying to reproduce the issue and now it looks it is always working, so I am not able to copy-paste the stacktrace of the issue, but I remember it was related to a timeout. One test was waiting to something to happen.

## Steps to reproduce
```
cd microprofile/tests/tck/tck-lra
mvn clean install
```

PR #6107, opened 2023-02-07
LRA TCK failing randomly #6106

jbescos:
Relates to https://github.com/helidon-io/helidon/issues/6106

I am not able to make it to fail anymore. This issue was happening more or less the 50% of the time.

Then, I am enabling the test again.

In case it fails, we will take the stacktrace of the error and we can analyze it.

### DIFF
--- a/microprofile/tests/tck/tck-lra/pom.xml
+++ b/microprofile/tests/tck/tck-lra/pom.xml
@@ -108,10 +108,6 @@
                     <includes>
                         <include>**/*Test*.java</include>
                     </includes>
-                    <excludes>
-                        <!-- It fails randomly -->
-                        <exclude>org.eclipse.microprofile.lra.tck.TckRecoveryTests.java</exclude>
-                    </excludes>
                 </configuration>
                 <dependencies>
                     <dependency>


microprofile/tests/tck/tck-lra/pom.xml
```diff
@@ -108,10 +108,6 @@
                     <includes>
                         <include>**/*Test*.java</include>
                     </includes>
-                    <excludes>
-                        <!-- It fails randomly -->
-                        <exclude>org.eclipse.microprofile.lra.tck.TckRecoveryTests.java</exclude>
-                    </excludes>
```
danielkec, line 114: :eyes: This is LRA not a Fault Tolerance and it was fixed https://github.com/helidon-io/helidon/commit/8a041bc063ff39f9b75a665f89663021f5eb2c76 a week ago, why it was excluded in the first place?

danielkec (changes requested):
Please rename PR according to actual changes it makes

microprofile/tests/tck/tck-lra/pom.xml
jbescos: I have updated it the PR and the original issue. It is LRA.

It was excluded because test was failing very often locally and when it was executed by the github actions. I don't know what has changed but now it is consistently working.

danielkec (approved):

Issue #6094, opened 2023-02-06
3.x: Incorrect code example in Helidon SE Tracing doc


al3xandru:
In the Helidon SE Tracing doc https://helidon.io/docs/v3/#/se/tracing, the custom span creation code snippet is not correct for Helidon 3.x API:

```
Span span = request.tracer()
        .buildSpan("my-operation")
        .asChildOf(request.spanContext())
                         .start();
```

`io.helidon.webserver.Request#tracer()` returns an `io.helidon.tracing.Tracer` object which has a different API than the one used above.

al3xandru:
Similar issue in https://helidon.io/docs/v3/#/se/guides/tracing

ljamen:
@al3xandru I created a new PR for all of the Tracing updates. It is now #6113

ljamen:
fixed by #6113 (approved / merged)

PR #6113, opened 2023-02-07
DOC: changes for tracing and tracing guide

### LINKED ISSUE #5187: Doc for Jaeger tracing integration in 3.x: shows incorrect port as of Helidon 3.0 and maybe should discuss removed config settings

ljamen:
Fixes #5187 and #6094

### DIFF
--- a/docs/se/guides/tracing.adoc
+++ b/docs/se/guides/tracing.adoc
@@ -162,7 +162,7 @@ WebServer server = WebServer.builder(createRouting(config))
 <2> Build and register a `Tracer` object using the tracing configuration.
 
 [source,java]
-.Update the `GreetService` class; 1) Add a new import and 2) Replace the `getDefaultMessageHandler` method:
+.Update the `GreetService` class; Replace the `getDefaultMessageHandler` method:
 ----
     private void getDefaultMessageHandler(ServerRequest request,
                                    ServerResponse response) {
@@ -382,7 +382,7 @@ WebServer server = WebServer.builder(createRouting(config))
 ----
 
 [source,java]
-.Update the `GreetService` class; 1) Add a new import and 2) Replace the `getDefaultMessageHandler` method:
+.Update the `GreetService` class; Replace the `getDefaultMessageHandler` method:
 ----
     private void getDefaultMessageHandler(ServerRequest request,
                                    ServerResponse response) {
@@ -463,7 +463,6 @@ import io.helidon.webserver.Routing;
 import io.helidon.webserver.ServerRequest;
 import io.helidon.webserver.ServerResponse;
 import io.helidon.webserver.Service;
-import io.opentracing.Span;
 import java.util.Collections;
 import java.util.concurrent.atomic.AtomicReference;
 import jakarta.json.Json;
--- a/docs/se/guides/tracing.adoc
+++ b/docs/se/guides/tracing.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2019, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2019, 2023 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


docs/se/guides/tracing.adoc
```diff
@@ -163,30 +164,28 @@ WebServer server = WebServer.builder(createRouting(config))
 [source,java]
 .Update the `GreetService` class; 1) Add a new import and 2) Replace the `getDefaultMessageHandler` method:
```
tjquinno, line 165: I missed this earlier. We no longer add a new import; we just replace the method.

docs/se/guides/tracing.adoc
```diff
@@ -383,25 +382,30 @@ WebServer server = WebServer.builder(createRouting(config))
 ----
 
 [source,java]
-.Update the `GreetService` class; 1) Add new import and 2) Replace the `getDefaultMessageHandler` method:
+.Update the `GreetService` class; 1) Add a new import and 2) Replace the `getDefaultMessageHandler` method:
```
tjquinno, line 385: Again, no new import any more.

docs/se/guides/tracing.adoc
```diff
@@ -492,17 +496,19 @@ public class GreetService implements Service {
         .put("/greeting", this::updateGreetingHandler);
   }
 
-  private void getDefaultMessageHandler(ServerRequest request, ServerResponse response) {
+  private void getDefaultMessageHandler(ServerRequest request,
```
tjquinno, line 499: At line 466 remove the now-obsolete `import io.opentracing.Span`

tjquinno (changes requested):

Issue #6112, opened 2023-02-07
OCI metrics intermittent failure(s)


tomas-langer:
```
Error:  Failures: 
Error:    OciMetricsSupportTest.testEndpoint:196 
Expected: is "https://telemetry.dummyendpoint.com/"
     but: was "https://telemetry-ingestion.dummyendpoint.com/"
Error:    OciMetricsSupportTest.testMetricScope:320->validateMetricCount:423->validateMetricCount:439 
Expected: is <1>
     but: was <6>
```

I have seen the second issue at least 3 times

klustria:
Been running from 100 to 200 continuous retries of the OciMetricsSupportTest like below:
```bash
 for i in {1..200}; do mvn clean test -Dtest=OciMetricsSupportTest; done 
```
and so far have not reproduced this issue. Will continue to attempt to reproduce this.

klustria:
The use of `@RepeatedTest(4000)` on testEndpoint helped reproduce the problem and I think I have now an understanding of the behavior:

1. The failure in `OciMetricsSupportTest.testEndpoint` is a race condition wherein the restore of the endpoint in the OciMetricsSupport has not been executed yet before the test assert call is made to validate it. I think that the test should wait for a certain period of time (maybe a few seconds even though it probably just missed it by a few milliseconds) to validate that the endPoint has been restored.
2. The failure in `OciMetricsSupportTest.testMetricScope` is a chain reaction effect of the above problem. Because it failed, the webServer was not stopped and so the ociMetricsSupport executor thread will continue to be running and at a certain interval, will be retriggered, and will call `countDownLatch1.countDown()` which will affect one of the validations in `OciMetricsSupportTest.testMetricScope`. Because `countDownLatch1` is a global static variable, the one used by the failed test will be reused by the new test causing this failure. We can ignore this and fix the above issue and this will not be a problem anymore or make `countDownLatch1` as test method local variable.

klustria:
Problem can be reproduced consistently by

1.  Replace `@Test` with `@RepeatedTest(4000)` in `OciMetricsSupportTest.testEndpoint`
2. Execute the mvn test commands executing `OciMetricsSupportTest.testEndpoint` and `OciMetricsSupportTest.testMetricScope`  5 times or until it fails using this one liner bash command:
```bash
for i in {1..5}; do mvn clean test -Dtest=OciMetricsSupportTest#testEndpoint+testMetricScope || break && echo "Finished test #$i"; done;
```

[...]

PR #6151, opened 2023-02-10
Fix intermittent issue on OciMetricsSupportTest

klustria:
- RCA is described here: https://github.com/helidon-io/helidon/issues/6112#issuecomment-1426213497
- Steps to reproduce is described here: https://github.com/helidon-io/helidon/issues/6112#issuecomment-1426278866

Changes include the following:
1. In OciMetricsSupportTest.testEndpoint, extend the amount of validation time to 3 seconds for checking that the metric endpoint has been restored. Intermittently, a race condition exist where the validation happens before the endpoint is restored.
2. Modify all countdownLatch to be locally defined in the test methods rather than being a static variable, which is causing chain reaction failure to other tests if a previous test fails because they share the same countdownLatch.
3. Always check that countDownLatch.await() is verified to have completed or otherwise, assert a failure.
4. Remove the use of fixed port when starting a WebServer.
5. Reset postingEndPoint to its original value before each test, so @RepeatedTest can be used in the future for debugging purposes.
6. Apply Helidon Code Style on both OciMetricsSupportTest and OciMetricsCdiExtensionTest. This would include making the tests's class and methods package local  rather than public, rearranging variable fields order based on whether they are static, final, etc.
7. Note that OciMetricsCdiExtensionTest only involves Code Style change and removal of delay method which is never used, so logic in that test class will be the same as before. Only OciMetricsSupportTest contain significant change to resolve the issue reported.

### DIFF
--- a/integrations/oci/metrics/cdi/src/test/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsCdiExtensionTest.java
+++ b/integrations/oci/metrics/cdi/src/test/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsCdiExtensionTest.java
@@ -71,6 +71,7 @@ import org.junit.jupiter.api.Test;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.fail;
 
 @HelidonTest(resetPerTest = true)
 @AddBean(OciMetricsCdiExtensionTest.MockMonitoring.class)
@@ -93,7 +94,7 @@ import static org.hamcrest.Matchers.is;
 @AddConfig(key = "ocimetrics.delay", value = "2")
 class OciMetricsCdiExtensionTest {
     private static volatile int testMetricCount = 0;
-    private static CountDownLatch countDownLatch1 = new CountDownLatch(1);
+    private static CountDownLatch countDownLatch = new CountDownLatch(1);
     private static PostMetricDataDetails postMetricDataDetails;
     private static boolean activateOciMetricsSupportIsInvoked;
     private final RegistryFactory rf = RegistryFactory.getInstance();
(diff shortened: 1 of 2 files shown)


klustria:
Validate / build (ubuntu-20.04) (pull_request)  failed with (which is not related to the change):
```
Warning:  Tests run: 6, Failures: 0, Errors: 0, Skipped: 1, Time elapsed: 0.814 s - in io.helidon.nima.tests.integration.http2.webserver.Http2ServerTest
[28988](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28989)
[INFO] 
[28989](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28990)
[INFO] Results:
[28990](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28991)
[INFO] 
[28991](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28992)
Error:  Failures: 
[28992](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28993)
Error:    Http2WebServerStopIdleTest.stopWhenIdleExpectTimelyStopHttp2:82 
[28993](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28994)
Expected: is a value less than <500>
[28994](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28995)
     but: <501> was greater than <500>
[28995](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28996)
[INFO] 
[28996](https://github.com/helidon-io/helidon/actions/runs/4147930397/jobs/7175380333#step:4:28997)
Error:  Tests run: 22, Failures: 1, Errors: 0, Skipped: 1
```

klustria:
`Validate / examples (macos-latest) (pull_request)`on the other hand was cancelled. Will restart both validation.

tjquinno (approved):

PR #6163, opened 2023-02-13
netty 4.1.90 final

sullis:
https://netty.io/news/index.html


sullis:
WDYT? @barchetta

barchetta:
> WDYT? @barchetta

Much thanks for the PR @sullis !  Due to project policies we generally require third party dependency updates  come from a project member. So what will likely happen is that I will turn my attention to this and  create a new PR based on your work. Apologies this is taking longer than it should. Stay tuned!

PR #6203, opened 2023-02-14
Register injected metrics during start-up

### LINKED ISSUE #6122: `@Inject @Metric` should register the metric during startup but does so later
## Environment Details
* Helidon Version: 2.5.5 and later including 3.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
A declaration like this in a resource class
```
@Inject
@Metric(...) // contains metadata information
private Counter myCounter;
```
should cause Helidon to register that metric during start-up, so that accessing the `/metrics` endpoint even before accessing the business endpoints should show the counter.

Instead, Helidon registers the counter only later, once the REST client accesses a business endpoint in that resource class.

## Steps to reproduce
1. In the MP QuickStart example app, add a declaration like the one above to `GreetResource`.
2. Access `http://localhost:8080/metrics/application` and note that the output is empty.
3. Access `http://localhost:8080/greet`
4. Access `http://localhost:8080/metrics/application` and notice that the output now includes the metric.

tjquinno:
Resolves #6122 

# Background
MP Metrics permits developers to use annotations to refer to metrics in two ways:

## "Metric-declaring" annotations
`@Counted`, `@Timed`, etc. _declare_ metrics. In 2.x, there can be only one such metric-declaring annotation for a given metric _unless_ that metric's metadata allows reuse. In that case, we permit multiple registrations of the same metric provided that the metadata (description, display name, units, reusability) supplied by all the metric's metric-declaring annotations are consistent. 

In addition, developers can declare `@Produces` methods and fields to serve as sources for injecting metrics elsewhere. Such producers can have the `@Metric` annotation to provide additional metadata information for the produced metric.

## `@Inject` sites
The developer can `@Inject` a metric as well, either as fields or as parameters to methods and constructors. Again, such references can use the `@Metric` annotation to specify metadata information for the metric to be injected. 
Helidon's MP implementation injects the correct metrics either using developer-provided producers (if any) or, failing that, its own built-in producers for the various kind of metrics.

If the referenced metric has not yet been registered, the producer (either the developers or ours) creates and registers the metric. If the metric _has_ been previously registered, the producer simply returns that one to CDI for injection.

The `@Metric` annotation on an injection site _can_ only identify the metric to be injected: give the name and any relevant tags. The parameter's or field's Java type determines the metric type to be injected. The ID and type are enough for Helidon to locate and inject the correct metric. If the metric has not already been registered, Helidon registers it automatically before injecting it. 

If the `@Metric` annotation appears on multiple injection sites for the same metric, the metadata declared in the annotations must be consistent across all appearances.

# What behavior does this PR change?
## Former behavior
Helidon at start-up would automatically register only those metrics triggered by metric-_declaring_ annotations and producers. Metrics that appeared only at injection sites would be registered just-in-time, when they were first needed. For a JAX-RS resource, for example, that translates to the first access to an endpoint exposed by the resource. So, after start-up but before any endpoint accesses, the output from `/metrics` would _not_ include metrics which appear only at injection sites in beans which CDI had not yet had to create.

This was odd and, potentially, contrary to the MicroProfile Metrics spec (although no TCK tests exercise this).

## New behavior
Users can configure Helidon to register, during _start-up_, all metrics implied by metric-declaring sites, producers, _and injection sites_.

To work, though, in 2.x these changes must bypass any user-provided producers for metrics. That is why this feature is off by default. Users must explicitly enable it:
```properties
mp.metrics.startup-registration.enabled = true
```

This PR requires injected meter parameters to have a `@Metric` annotation with at least the `name` specified. (If tags are important to identify the metric they should be supplied as well but we have no way of knowing if we should require tags.) At runtime the Java parameter names are not always available, in which case they are named `arg<n>` based on the position in the parameter list. What should be the same injected metric could appear in different places in two different method signatures. It is much safer to require the developer to be explicit about the name (and, if relevant, tags).

## How the MP metrics CDI extension works (unchanged at a high level)
The MP metrics CDI extension follows this high-level life cycle:
1. Observes `ProcessAnnotatedType` for any type bearing any metric-declaring or metric-referring annotation and records each type.
2. Observes `ProcessManagedBean`. Ignores any bean not previously recorded by the `PAT` observer. For each declared or referenced metric in the type, creates and saves a deferred registration object. We cannot register the metrics at this point because runtime config can affect the behavior of the metric registry we will use, so instead we save the registration work to be done later.
3. Observes `@Initialized(ApplicationScoped.class)`. At this point the runtime config is available so we can obtain the correct metric registry and perform the deferred registration work saved during `PMB`.

# Overview of the changes
Two main categories:
1. Because we now potentially attempt to register the same metric from multiple sites (declaring and referring), we need to make sure the metadata we use from all those sites is consistent. In particular, we now normalize the description and display name to null if the original value is an empty string (as can happen easily from a `@Metric` annotation).
2. During start-up we now have to gather information about every metric injection site (formerly just the metric-declaring ones) and record the deferred registration work to register the metrics for those sites. This resulted in quite a few changes.

## `MetricAnnotationInfo`
The metric-declaring annotations have identical elements (`name`, `tags`, `absolute`, `displayName`, `description`, `units`, `reusable`). 
Because annotations cannot extend a common superclass, the metric-declaring annotations have to be handled individually despite the fact they have common methods. The `MetricAnnotationInfo` class abstracts all those details to simplify code elsewhere in MP metrics. This package-private class now supports some new functionality useful from other changed code. It also used to expose the maps it populates; now it exposes look-up methods and the maps are private. 

## `RegistrationPrep` class
This class represents deferred registrations of metrics. It used to be an inner class of `MetricAnnotationInfo` but it became complicated enough to warrant being a separate class on its own. It now has two subtypes, one for intercepted sites (metric-declaring annotations) and one for injected sites. The rules for consistency checking and the info needed to register metrics are a little different for the two.

## `MetricUtil`
Some refactoring of utility methods among classes.

## `MetricProducer`
Acts as a CDI producer of metrics. No substantive changes; removed unused parameters and code and rearranged some code.

## `MetricsCdiExtension`
Previously it recorded deferred registrations only for metric-_defining_ annotations (intercept sites). Now it also records deferred registrations for injection sites. We want to register the metrics based on referring sites last, so the extension maintains two separate lists for deferred registrations of the two types.

The extension has CDI produce the injected metrics, thus ensuring that developer-provided producers, if any, are properly invoked.

## Tests
Added tests to make sure injected metrics are registered with the correct metadataand before the system starts.

### DIFF
--- a/docs/mp/guides/05_metrics.adoc
+++ b/docs/mp/guides/05_metrics.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2019, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2019, 2023 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -77,6 +77,7 @@ include::{metrics-common}[tag=built-in-metrics-discussion]
 
 include::{metrics-common}[tag=controlling-intro-part-1]
 * <<controlling-rest-request-metrics,Control `REST.request` metrics.>>
+* <<controlling-registration-of-injected-metrics,Control Registration of Injected Metrics>>
 include::{metrics-common}[tag=controlling-intro-part-2]
 [[controlling-rest-request-metrics]]
 
@@ -95,6 +96,32 @@ Enable it by editing your application configuration to set `metrics.rest-request
 Note that the applications you generate using the full Helidon archetype _do_ enable this feature in the
 generated config file.
 You can see the results in the sample output shown in earlier example runs.
+
+[[controlling-registration-of-injected-metrics]]
+==== Controlling Registration of Injected Metrics
+Normally, the Helidon MP metrics subsystem relies on CDI to trigger the registration of metrics which are injected into beans.
+For example, your JAX-RS resource class might inject a metric into a field or method parameter declared by that class.
+Because you typically declare a JAX-RS resource classe as `@Dependent` scoped, CDI does not create an instance of the JAX-RS resource class until a client accesses an endpoint served by that resource class.
+At that time CDI also processes any injections into the resource class, including metrics.
+If your code only injects a metric and does not otherwise declare it, then only when a client first accesses that resource will CDI resolve the injection and thereby trigger the registration of the injected metric.
+
+One side effect of this behavior is that, after your service starts but before any client accesses a resource, the output from the `/metrics` endpoint will _not_ include metrics which are only injected, even though Helidon does register such metrics correctly--just in time--and updates them appropriately.
+
+[CAUTION]
+====
+_Do not_ use this feature if your application includes any CDI `@Produces` fields or methods for any of the injected metrics. To register injected-only metrics during start-up, Helidon MP 2 must bypass CDI producers and register the metrics directly itself.
+====
+You can configure the Helidon MP metrics system so it _does_ register injected-only metrics during start-up as follows:
+
+[source,properties]
+.Configuration properties file to select start-up (instead of just-in-time) injected-only metrics registration
+----
+mp.metrics.startup-registration.enabled = true
+----
+
+
+
+
 
 include::{metrics-common}[tags=controlling-by-component;controlling-by-registry]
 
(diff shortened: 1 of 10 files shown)


spericas:
@tjquinno Should this PR be closed?

tjquinno:
Closing, as we will probably not implement these changes because the spec and TCK does not require them.

PR #6213, opened 2023-02-15
[3.x] Register injected metrics during start-up

### LINKED ISSUE #6209: [3.x] `@Inject @Metric` should register the metric during startup but does so later
Backport of #6122 for Helidon 3.x

tjquinno:
Resolves #6209 

See PR #6203 for details. The same discussion applies here with very minor changes.


tjquinno:
We will most likely not change this behavior. The current implementation complies with the MP Metrics spec and TCK.

PR #6244, opened 2023-02-17
Nested transactions helidon 3.x

ljnelson:
This pull request should complete the JPA/JTA overhaul.

Most of its changes are due to extraordinarily annoying and complicated rebasing and/or checkstyle formatting and/or making the copyright checker happy.

One thing to watch out for is git has decided at various points that `JtaConnection.java` should be `JTAConnection.java`, which is incorrect. If you see this case sensitivity crop up, I will try to fix it again (I've already used `git mv` with single quotes to resolve it but it keeps coming back for reasons I don't care to investigate unless I have to).


ljnelson:
Converted back to draft status since integration tests are failing.

ljnelson:
(For myself and posterity: may have found the? a? problem in `JtaConnection.java` line 1127. That will "expand" to `!(this.isClosePending() || this.delegate().isClosed())` but that's not right.)

ljnelson:
Hibernate's default behavior is to aggressively close connections (normally releasing them back to a connection pool) after every statement creation. This assumes that a `DataSource` will always return a connection pinned to the current thread when there is an active, non-suspended JTA transaction. There is of course no requirement that a `DataSource` should do this, only that the transaction semantics should be respected. I can get the integration tests to pass by specifying that Hibernate should not do this. See https://docs.jboss.org/hibernate/orm/current/userguide/html_single/Hibernate_User_Guide.html#database-connection-handling for more details.

ljnelson:
Closing because #7719 and related were merged.

PR #6251, opened 2023-02-18
[4.x] 6121 100 continue request reset fix

### LINKED ISSUE #6121: Using Expect-100 fails on Transfer Encoding Chunk
## Environment Details
* Helidon Version: 4
* Helidon NIMA
* 
----------

## Problem Description
When using Expect-100 to send Transfer Encoding Chunk, no response is received. When using this handler, the entity is empty: 

```
private static void customHandler(ServerRequest req, ServerResponse res) {
    Headers headers = req.headers();

    requestChunked = headers.contains(Http.HeaderValues.TRANSFER_ENCODING_CHUNKED);
    requestContentLength = headers.contentLength().orElse(NO_CONTENT_LENGTH);
    if (headers.contains(Http.HeaderValues.EXPECT_100)) {
        expects100 = true;
    }

    try (InputStream inputStream = req.content().inputStream();
            OutputStream outputStream = res.outputStream()) {
        // inputStream.transferTo(outputStream);
        byte[] entity = inputStream.readAllBytes();
        System.out.println("Entity is '" + new String(entity) + "'");
        new ByteArrayInputStream(entity).transferTo(outputStream);
    } catch (Exception e) {
        res.status(Http.Status.INTERNAL_SERVER_ERROR_500).send("failed: " + e.getMessage());
    }
}
```

danielkec:
Fixes #6121

### DIFF
--- a/common/testing/http-junit5/src/main/java/io/helidon/common/testing/http/junit5/SocketHttpClient.java
+++ b/common/testing/http-junit5/src/main/java/io/helidon/common/testing/http/junit5/SocketHttpClient.java
@@ -534,6 +534,7 @@ public class SocketHttpClient implements AutoCloseable {
 
     /**
      * Continue sending more to text to socket.
+     *
      * @param payload text to be sent
      * @return this http client
      * @throws IOException
@@ -545,6 +546,18 @@ public class SocketHttpClient implements AutoCloseable {
             pw.print(payload);
         }
         pw.flush();
+        return this;
+    }
+
+    /**
+     * Send single chunk as defined by RFC 9112 §7.1.
+     *
+     * @param payload of the chunk
+     * @return this http client
+     * @throws IOException
+     */
+    public SocketHttpClient sendChunk(String payload) throws IOException {
+        continuePayload(Integer.toHexString(payload.length()) + EOL + payload + EOL);
         return this;
     }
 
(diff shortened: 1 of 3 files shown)


klustria:
Change looks good. However, just one question which I also posted in https://github.com/helidon-io/helidon/issues/6121

I was encountering issue reading the http status after a continue-100 response and found that the main reason for this is because of this: https://github.com/helidon-io/helidon/blob/main/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1Connection.java#L58 where there is extra /r/n at the end of Continue 100 response. Prior to the expect-continue-100, there is only 1 /r/n but in the recent code there are 2. Is this a requirement to have 2x CRLF?

I can of course skip/ignore the remaining lines, but just wondering if this is a requirement.

klustria (approved):
I will approve this anyway, because I can deal with the extra CRLF, but would just like to verify if that is really needed for a 100-Continue response. Thanks for the quick turnaround, even on your weekend.

danielkec:
Hi @klustria afaik double CRLF is mandatory, see [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html)
```
Response      = Status-Line *(( general-header | response-header | entity-header ) CRLF)  
                       CRLF        ; <---- Second
                       [ message-body ]          

Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF ;<----- First
```

klustria:
Thanks @danielkec!

PR #6268, opened 2023-02-22
JWT-Auth upgrade to 2.1 version

### LINKED ISSUE #5819: 4.x: Upgrade MP JWT-Auth to 2.1

Verdent:
Fixes #5819

### DIFF
--- a/microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
+++ b/microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
@@ -32,6 +32,7 @@ import java.security.interfaces.ECPrivateKey;
 import java.security.interfaces.ECPublicKey;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Base64;
@@ -116,6 +117,8 @@ public class JwtAuthProvider extends SynchronousProvider implements Authenticati
      * Configuration key for expected audiences of incoming tokens. Used for validation of JWT.
      */
     public static final String CONFIG_EXPECTED_AUDIENCES = "mp.jwt.verify.audiences";
+    private static final String CONFIG_EXPECTED_MAX_TOKEN_AGE = "mp.jwt.verify.token.age";
+    private static final String CONFIG_CLOCK_SKEW = "mp.jwt.verify.clock.skew";
     /**
      * Configuration of Cookie property name which contains JWT token.
      *
@@ -147,7 +150,10 @@ public class JwtAuthProvider extends SynchronousProvider implements Authenticati
     private final Map<OutboundTarget, JwtOutboundTarget> targetToJwtConfig = new IdentityHashMap<>();
     private final String expectedIssuer;
     private final String cookiePrefix;
+    private final String decryptionKeyAlgorithm;
     private final boolean useCookie;
+    private final Duration expectedMaxTokenAge;
+    private final Duration clockSkew;
 
     private JwtAuthProvider(Builder builder) {
         this.optional = builder.optional;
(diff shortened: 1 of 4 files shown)


microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
```diff
@@ -116,6 +117,8 @@ public class JwtAuthProvider extends SynchronousProvider implements Authenticati
      * Configuration key for expected audiences of incoming tokens. Used for validation of JWT.
      */
     public static final String CONFIG_EXPECTED_AUDIENCES = "mp.jwt.verify.audiences";
+    private static final String CONFIG_EXPECTED_MAX_TOKEN_AGE = "mp.jwt.verify.token.age";
```
spericas, line 120: Addition is good, but it would be nice to order public before private static fields.

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
```diff
@@ -1205,6 +1245,17 @@ public Builder decryptKeyLocation(String decryptKeyLocation) {
...
         }
 
+        /**
+         * Required decryption key algorithm.
+         *
+         * @param decryptionKeyAlgorithm required decryption key algorithm
+         * @return updated builder instance
+         */
+        public Builder decryptKeyAlgorithm(String decryptionKeyAlgorithm) {
+            this.decryptionKeyAlgorithm = decryptionKeyAlgorithm;
```
spericas, line 1255: Required param non-null?

security/jwt/src/main/java/io/helidon/security/jwt/EncryptedJwt.java
```diff
@@ -193,6 +194,32 @@ public static EncryptedJwt parseToken(JwtHeaders header, String token) {
...
+     *
+     * @param validators collection of validators
+     * @param expectedKekAlg   audience key encryption key algorithm
+     * @param mandatory  whether the alg field is mandatory in the token
+     */
+    public static void addKekValidator(Collection<Validator<EncryptedJwt>> validators, String expectedKekAlg, boolean mandatory) {
+        validators.add((encryptedJwt, collector) -> {
+            Optional<String> kekAlgorithm = encryptedJwt.header.algorithm();
+            if (kekAlgorithm.isPresent()) {
+                if (expectedKekAlg == null) {
```
spericas, line 208: Could this be an OR conditional?

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
Verdent: Good point, I will change that. :-)

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
Verdent: This can be null. I think better would be probably to change the wording to not say required, but expected :-) This required could be a bit misleading.

security/jwt/src/main/java/io/helidon/security/jwt/EncryptedJwt.java
Verdent: Yep, make sense.

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
Verdent: Hmmm I took a look into this and right now and public static fields are before the private ones. You are right, there is LOGGER and JSON defined before that, but those should not be moved from my point of view. I would very much rather keep it the way it is now, since public property names are before the private ones already. Does it make sense?

danielkec:
Dependency change needs to be reviewed by release manager @barchetta

security/jwt/src/main/java/io/helidon/security/jwt/EncryptedJwt.java
```diff
@@ -173,7 +174,7 @@ public static EncryptedJwt parseToken(String token) {
      * </ul>
      *
      * @param header parsed JWT header
-     * @param token String with the token
+     * @param token  String with the token
```
danielkec, line 177: :space_invader:

microprofile/jwt-auth/src/main/java/io/helidon/microprofile/jwt/auth/JwtAuthProvider.java
Verdent: changed.

danielkec (approved):
LGTM please wait for @barchetta review

PR #6296, opened 2023-02-25
Escape the key when copying a config node

### LINKED ISSUE #6295: Copied config node uses unescaped key if the key contains a dot
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
When a config node is copied, if the key contains a dot the copied key is not escaped so the resulting copy has the wrong key.

## Steps to reproduce
Add the following test to `ConfigCopyTest` and run it. It fails because the copied key is _not_ escaped when the new copy of the node is created:

```java
   @Test
    void testDottedKey() {
        var withDottedName = ConfigNode.ObjectNode.builder()
                .addValue(Config.Key.escapeName("first.one"), "firstValue")
                .addValue("second", "secondValue")
                .build();

        var rootNode = ConfigNode.ObjectNode.builder()
                .addObject("object", withDottedName)
                .build();

        var originalConfig = Config.builder()
                .addSource(io.helidon.config.ConfigSources.create(rootNode))
                .build();

        assertThat("Dotted name entry in original",
                   originalConfig.get("object." + Config.Key.escapeName("first.one")).asString().get(),
                   is("firstValue"));

        var copyConfig = Config.builder()
                .addSource(ConfigSources.create(originalConfig))
                .build();

        var firstOneConfigValue = copyConfig.get("object." + Config.Key.escapeName("first.one")).asString();
        assertThat("Dotted name key in copy is present",
                   firstOneConfigValue.isPresent(),
                   is(true));
        assertThat("Dotted name value in copy",
                   firstOneConfigValue.get(),
                   is("firstValue"));
    }
```

tjquinno:
Resolves #6295

### DIFF
--- a/config/config/src/main/java/io/helidon/config/ObjectNodeImpl.java
+++ b/config/config/src/main/java/io/helidon/config/ObjectNodeImpl.java
@@ -116,19 +116,20 @@ public class ObjectNodeImpl extends AbstractMap<String, ConfigNode> implements O
     private static void addObjectNode(Builder parentBuilder, Config parent) {
         parent.asNodeList().ifPresent(it -> {
             for (Config child : it) {
+                String escapedName = Config.Key.escapeName(child.name());
                 switch (child.type()) {
                 case OBJECT -> {
                     Builder childBuilder = ObjectNode.builder();
                     addObjectNode(childBuilder, child);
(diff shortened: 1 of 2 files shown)


spericas (approved):
LGTM

PR #6321, opened 2023-02-28
Fixes erroneous closing behavior in JtaConnection.java

ljnelson:
Consider the use case where a method is annotated with `jakarta.transaction.Transactional` and `close()` is called on a `Connection` obtained therein. `close()` in this case should not actually close the `Connection` at that very moment, but closing _should_ happen after the completion of the transaction. The prior version of this class was OK on the first point, and butchered the second point, so these changes fix the behavior. Note that the `JtaConnection` class is not currently used in Helidon, but will be soon.

### DIFF
--- a/integrations/jta/jdbc/etc/spotbugs/exclude.xml
+++ b/integrations/jta/jdbc/etc/spotbugs/exclude.xml
@@ -29,6 +29,10 @@
         <Bug pattern="SQL_INJECTION_JDBC"/>
     </Match>
     <Match>
+        <Class name="io.helidon.integrations.jta.jdbc.JtaConnection$Enlistment"/>
+        <Bug pattern="IP_PARAMETER_IS_DEAD_BUT_OVERWRITTEN"/>
+    </Match>
+    <Match>
         <Class name="io.helidon.integrations.jta.jdbc.LocalXAResource$Association"/>
         <Bug pattern="IP_PARAMETER_IS_DEAD_BUT_OVERWRITTEN"/>
     </Match>
--- a/integrations/jta/jdbc/etc/spotbugs/exclude.xml
+++ b/integrations/jta/jdbc/etc/spotbugs/exclude.xml
@@ -29,10 +29,6 @@
         <Bug pattern="SQL_INJECTION_JDBC"/>
     </Match>
     <Match>
-        <Class name="io.helidon.integrations.jta.jdbc.JtaConnection$Enlistment"/>
-        <Bug pattern="IP_PARAMETER_IS_DEAD_BUT_OVERWRITTEN"/>
-    </Match>
-    <Match>
         <Class name="io.helidon.integrations.jta.jdbc.LocalXAResource$Association"/>
         <Bug pattern="IP_PARAMETER_IS_DEAD_BUT_OVERWRITTEN"/>
     </Match>
--- a/integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
+++ b/integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
@@ -1124,27 +1124,18 @@ class JtaConnection extends ConditionallyCloseableConnection {
     private void transactionCompleted(int commitedOrRolledBack) {
         this.enlistment = null; // volatile write
         try {
-            if (!super.isClosed()) {
-                // Did someone call close() while we were enlisted?  That's permitted by section 4.2 of the JTA
-                // specification, but it shouldn't actually close the connection because the prepare/commit cycle
-                // wouldn't have started.
-                boolean closeWasPending = this.isClosePending(); // volatile read
-
-                // If they did, then we must be non-closeable. If they didn't, we could be either closeable or not
-                // closeable.
-                assert !closeWasPending || !this.isCloseable();
-
-                // Now the global transaction is over, so blindly set our closeable status to true. (It may already be
-                // true.)  This resets the closePending status, per spec.
-
-                this.setCloseable(true);
-                assert this.isCloseable();
+            boolean closeWasPending = this.isClosePending();
+            this.setCloseable(true);
+            assert this.isCloseable();
+            assert !this.isClosePending();
+            if (closeWasPending) {
+                assert !this.isClosed();
+                assert !this.delegate().isClosed();
+                this.close();
+                assert this.isClosed();
+                assert this.delegate().isClosed();
+                assert !this.isCloseable();
                 assert !this.isClosePending();
-
-                // If there WAS a close() attempt, now it CAN work, so let it work.
-                if (closeWasPending) {
-                    this.close();
-                }
             }
         } catch (SQLException e) {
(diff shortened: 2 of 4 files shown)


integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
```diff
@@ -1124,27 +1124,18 @@ private Connection connectionFunction(Xid xid) {
...
-                // If they did, then we must be non-closeable. If they didn't, we could be either closeable or not
-                // closeable.
-                assert !closeWasPending || !this.isCloseable();
-
-                // Now the global transaction is over, so blindly set our closeable status to true. (It may already be
-                // true.)  This resets the closePending status, per spec.
-
-                this.setCloseable(true);
-                assert this.isCloseable();
+            boolean closeWasPending = this.isClosePending();
```
tjquinno, line 1127: An informational question... (Maybe we discussed this in an earlier round of reviews on related PRs and I've forgotten. Also, this is not directly related to this latest set of changes.) 

The group of flags (`isCloseable`, `isClosePending`, `isClosed`) suggest that there's a state model lurking here. I wonder if being explicit about the states and the transitions among them might clarify things or would, in fact, actually muddy the waters.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
```diff
@@ -1124,27 +1124,18 @@ private Connection connectionFunction(Xid xid) {
...
+            this.setCloseable(true);
+            assert this.isCloseable();
+            assert !this.isClosePending();
+            if (closeWasPending) {
+                assert !this.isClosed();
+                assert !this.delegate().isClosed();
+                this.close();
```
tjquinno, line 1134: Probably inconsequential, but the `close` method in this class (I did not look at `super.close()` does not check `isCloseable()`. In fact, in this class only the `assert`s here check it. Is invoking `setCloseable(true)` before invoking `close()` necessary? Redundant? Immaterial? (I guess I'm mostly just grasping for the underlying state model represented by the various flags.)

ljnelson: Fair questions all.
1. `super.close()` (`ConditionallyCloseableConnection#close()`) does in fact check `isCloseable()`, and takes different actions depending on the outcome. If the connection is closeable, then the delegate it wraps gets its `close()` method called. If the connection is not closeable (because it is enrolled in a transaction, for example), then the intent to close it is recorded and action is taken at transaction completion time to actualize the request.
2. Calling `setCloseable(true)` here is critical. Basically, someone called `close()` on the `JtaConnection`, and, now that the transaction is definitely completed, that operation that the caller _thought_ was immediate but was not should _now_ take effect. It can only take effect if the `JtaConnection` is, in fact, closeable.

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
ljnelson: Yes. The trick is that the state model has to play nice with the state model of `java.sql.Connection`, which exposes `close()` and `isClosed()`.

(I also can't change the signatures of the public classes, like `ConditionallyCloseableConnection`, that already exist.)

The state model is described in javadoc; see https://helidon.io/docs/v3/apidocs/io.helidon.integrations.jdbc/io/helidon/integrations/jdbc/ConditionallyCloseableConnection.html.

In general, if the `closeable` property is `false`, then attempts to close the connection will record that a close request happened.  `isClosed`, in order to honor the `java.sql.Connection` contract, will always return `true` once `close()` has been invoked. Clear as mud? The spirit of this contract, while not reified anywhere other than these Helidon classes, is pervasive throughout the JDBC specification.

ljnelson:
@tjquinno I've incorporated your suggestion to bake an explicit state model into `ConditionallyCloseableConnection` without changing existing signatures etc. and without disturbing existing test behavior. While the change was a bit hairy for me to implement I greatly prefer it, so thank you for the suggestion. I'll resolve the other discussions above.

danielkec:
@ljnelson Please add link to corresponding issue

integrations/jta/jdbc/etc/spotbugs/exclude.xml
```diff
@@ -28,6 +28,10 @@
         <Class name="io.helidon.integrations.jta.jdbc.JtaConnection"/>
         <Bug pattern="SQL_INJECTION_JDBC"/>
     </Match>
+    <Match>
+        <Class name="io.helidon.integrations.jta.jdbc.JtaConnection$Enlistment"/>
+        <Bug pattern="IP_PARAMETER_IS_DEAD_BUT_OVERWRITTEN"/>
+    </Match>
```
danielkec, line 34: Is this needed because of :point_down:  ?
```java
        private Enlistment {
            threadId = Thread.currentThread().getId();
```

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
```diff
@@ -1205,6 +1196,7 @@ private Enlistment(Transaction transaction, XAResource xaResource) {
         }
 
         private Enlistment {
+            threadId = Thread.currentThread().getId();
             Objects.requireNonNull(transaction, "transaction");
             Objects.requireNonNull(xaResource, "xaResource");
         }
```
danielkec, line 1202: Seems like its used on one place only, why not set the threadId there?

danielkec (changes requested):
Can't we encapsulate this in AtomicReference for the sake of readability?
```java
Enlistment enlistment = new Enlistment(t, xar);
if (!ENLISTMENT.compareAndSet(this, null, enlistment)) { // atomic volatile write
    // Setting this.enlistment could conceivably fail if another thread already enlisted this JtaConnection.
    // That would be bad.

```

integrations/jta/jdbc/src/main/java/io/helidon/integrations/jta/jdbc/JtaConnection.java
ljnelson: The thought was:
* `Enlistment` is a record
* `threadId` needs to be part of its equality and hash code contract
* I _could_ override both instead in which case now the "recordness" is kind of pointless
* Meh, it's a `private` internal class anyway so let me just ensure that the thread id is always set correctly

ljnelson: In my latest change I went ahead and just simplified all this so you have to specify the thread ID.

tjquinno (approved):
I know it was work to use explicit states, but I think it's much clearer that way.

LGTM

danielkec (approved):

PR #6322, opened 2023-02-28
Capture and propagate the CCL in ThreadPerTaskExecutor 

spericas:
Capture the CCL when ThreadPerTaskExecutor is created and set it on every thread it creates. This is required by our Arquillian integration code and generally by MP config. Issue #6325.


danielkec (approved):
Didn't observe any significant performance regression

danielkec:
@spericas Please link corresponding issue before merging

PR #6333, opened 2023-03-02
Change default exemplar behavior to conform to OpenMetrics spec; allow users to choose former non-standard behavior

### LINKED ISSUE #6318: Prometheus unable to parse Helidon MP Exemplars

## Environment Details
* Helidon Version: 2.5.6
* Helidon SE or Helidon MP : MP
* JDK version: 17
* OS: Oracle Linux 9

----------

## Problem Description
Exemplars are automatically added to metrics of type 'gauge' for JAX-RS annotated Services.  This results in Prometheus ServiceMonitor failing to parse with the following error:

'metric name base_REST_request_elapsedTime_seconds does not support exemplars'

snip from /metrics EP:

```
# TYPE base_REST_request_elapsedTime_seconds gauge
base_REST_request_elapsedTime_seconds{class="com.oracle.oal.wlp.wbs.HTMLService",method="getContent_java.lang.String_java.lang.String_java.lang.String_int_java.lang.String"} 49.343265107 # {trace_id="b17d175d1bfbcd8f"} 49343265107 1677598231.371000
```

According to the following link, gauges are not supposed to be supported and suffix needs to be either _total or _bucket: https://github.com/prometheus/prometheus/issues/9738 

Reading the specs, it is unclear if this should be the case.  Would like to know a workaround/solution to this

tjquinno:
Resolves #6318 

The original exemplar implementation was too aggressive in adding exemplars, at least when compared to the OpenMetrics spec (only counters and buckets may have exemplars).

This PR changes the default behavior to conform to the spec, while letting users elect the older behavior if they depend on it.

As before, users still control whether Helidon adds exemplars at all by adding (or not) the `helidon-metrics-trace-exemplar` component to their projects. Beyond that, this PR allows a user to use configuration or programmatic builders to control, for each of the three registry types (application, base, and vendor), whether each should use strict exemplar support (the new default) or not. 

I added this flag to the registry settings, rather than to the overall metrics settings, because in the code each metric instance expresses itself in Prometheus format (which is where exemplars might appear) and each metric instance already has a reference to the registry in which it appears and therefore the registry settings. Each metric _does not_ already have a way to find the overall metric settings. Adding that ability would have required changing quite a few more classes and public interfaces. 

We expect that few users will need to control this behavior and Helidon 4.x will probably use a different approach for managing exemplars altogether so we do not want to add complexity to the current implementation that will be (relatively) short-lived.

The changed-file count looks large, but many of the files are the individual metric impl classes, each of which changes slightly (to accept and, as needed, act on the new flag, basically), and their corresponding test classes. Many others are the `XXXSettings` classes which deal directly with the new flag.


tjquinno:
I've marked this as a draft PR; I need to add doc changes and I need to review and maybe update the exemplar example.

tjquinno:
I converted this to draft, at least temporarily, because a new release of Prometheus due next week (see the linked Helidon issue) might resolve the problem without any changes from Helidon.

I have pinged the issue filer to see if upgrading the Prometheus server will be an option. If so, we might not need these changes at all.

etc/checkstyle-suppressions.xml
```diff
@@ -57,6 +57,10 @@
     <suppress checks="IllegalImport"
             files="DerUtils\.java"/>
 
+    <!-- Metrics Prometheus output generation depends on numerous settings -->
+    <suppress checks="ParameterNumber"
+              files="MetricImpl\.java"/>
+
```
danielkec, line 63: Checkstyle exclusion needs to be confirmed by release manager @barchetta

tjquinno:
The original filer of the issue does not control the release of Prometheus that is in play. As a result, relying on the upcoming Prom. release which relaxes some of these restrictions is not an option. That's why I changed this from a draft PR back to an active one.

etc/checkstyle-suppressions.xml
barchetta: LGTM

danielkec (approved):

PR #6342, opened 2023-03-02
[3.x] OIDC original uri resolving leaving out query params

### LINKED ISSUE #6341: OIDC removes query parameters from original URI
URI query parameters are removed when OIDC resolves original URI

Verdent:
Fixes #6341

### DIFF
--- a/security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
+++ b/security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
@@ -395,7 +395,13 @@ class TenantAuthenticationHandler {
                 .getOrDefault(Security.HEADER_ORIG_URI, List.of());
 
         if (origUri.isEmpty()) {
-            origUri = List.of(providerRequest.env().targetUri().getPath());
+            String query = providerRequest.env().targetUri().getQuery();
+            String path = providerRequest.env().targetUri().getPath();
+            if ((null == query) || query.isEmpty()) {
+                origUri = List.of(path);
+            } else {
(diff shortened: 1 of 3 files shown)


security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
```diff
@@ -395,7 +395,13 @@ private String origUri(ProviderRequest providerRequest) {
...
 
         if (origUri.isEmpty()) {
-            origUri = List.of(providerRequest.env().targetUri().getPath());
+            String query = providerRequest.env().targetUri().getQuery();
+            String path = providerRequest.env().targetUri().getPath();
+            if ((null == query) || query.isEmpty()) {
+                origUri = List.of(path);
+            } else {
+                origUri = List.of(path + "?" + query);
+            }
```
danielkec, line 404: Could it be?
```suggestion
            URI uri = providerRequest.env().targetUri();
            String query = uri.getQuery();
            String path = uri.getPath();

            if ((null != query) && !query.isEmpty()) {
                origUri = List.of(path + "?" + query);
            } else {
                origUri = List.of(path);
            }
```

danielkec (changes requested):

danielkec (dismissed):

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
spericas: `(null != query)` -> query != null

danielkec (approved):

PR #6350, opened 2023-03-02
[2.x]: Add @Target(ElementType.METHOD) for annotation @MPTest

Captain1653:
Javadoc for annotation `@MPTest` tells `Test method annotation`, but it can be in any place: parameter or class.

I think it's better to add some specific target.


Tomas-Kraus (approved):
Looks good.

PR #6384, opened 2023-03-08
[2.x] Adopt SnakeYAML 2.0

### LINKED ISSUE #6372: [2.x] Prepare for SnakeYAML 2.0
Backport of #5792 for Helidon 2.x

tjquinno:
Resolves #6372 

SnakeYAML 2.0 is now released.

This PR updates Helidon's own code to work with either 2.0 or 1.32.

It includes changes to config and OpenAPI, as well as an integration test verifying that config and OpenAPI work with 1.32.

A change in SnakeYAML from 1.32 to 2.0 now logs many `WARNING` messages (they were `FINE` in 1.32). To my mind, these are due to a shortcoming in SnakeYAML, not in how our code uses it. I have asked about this in the Google group for SnakeYAML but have not heard back yet. In the meantime, I've included code in this PR to change the logging level for the relevant SnakeYAML logger to `SEVERE` and added a config setting with which a user could disable our warning suppression to see those and other more important warnings if necessary.

Most of the other changes are so our code uses more secure APIs in SnakeYAML and works with either 1.32 or 2.0.


tests/integration/gh-5792/src/main/java/io/helidon/tests/integration/yamlparsing/GreetService.java
```diff
@@ -0,0 +1,151 @@
...
+        JsonObject returnObject = JSON.createObjectBuilder()
+                .add("message", msg)
+                .build();
+        response.send(returnObject);
+    }
+
+    private static <T> T processErrors(Throwable ex, ServerRequest request, ServerResponse response) {
+
+         if (ex.getCause() instanceof JsonException){
+
```
spericas, line 108: - "{" -> " {"
- Are the empty lines at the beginning of each if-else block intentional?

tjquinno: Inadvertent, all around. Fixed in latest push.

tests/integration/gh-5792/pom.xml
```diff
@@ -52,12 +52,6 @@
         </dependency>
         <!-- Verify that the app builds and that config and OpenAPI document parsing work with the older SnakeYAML release. -->
         <dependency>
-            <exclusions>
```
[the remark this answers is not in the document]
tjquinno: 1. It turns out it was the explicit dependency on snakeyaml:1.32 in the test app's pom that was forcing the version to what we wanted; the exclusion I had included in that dep. was a no-op.
2. I have since pushed a commit which adds a dep. mgt section to the test app's pom to accomplish this instead of using a direct dep. We really do want any transitive dependency that uses SnakeYAML to use the older release (which the dep. mgt does nicely) and this project does not itself need a direct dep. on SnakeYAML.
3. Because both config-yaml and openapi depend on snakeyaml it would have been fragile to add the exclusion to this project's dep. on only one of them. Adding to both would have been "safer" perhaps but not totally safe; if some additional component added a dep. on SnakeYAML in the future that might have messed up this test project.

spericas (dismissed):

barchetta (approved):

Issue #6278, opened 2023-02-23
4.x: programmatically control media providers with Nima WebServer


barchetta:
While working on migrating the Bookstore test application to Nima WebServer I ran into a gap (or maybe it is ignorance on my part).

In 3.x I can add specific media support programmatically:

```
webserverBuilder.addMediaSupport(JsonpSupport.create());
```

I can't find a way to do that with Nima WebServer. Maybe I'm missing something but it seems to rely only on service loading of `MediaSupportProvider`s.  

Note that this is different than Features which have `useSystemServices()` to turn on/off auto loading of services and there are ways to programmatically add a feature using `addFeature`.

Tomas-Kraus:
I'll have a look at this one.

PR #6412, opened 2023-03-15
Issue 6278: Programmatically control media providers with Nima WebServer

Tomas-Kraus:
Fixes issue #6278.

### DIFF
--- a/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
+++ b/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
@@ -166,7 +166,8 @@ public interface WebServer {
                 = HelidonServiceLoader.builder(ServiceLoader.load(ServerConnectionProvider.class));
 
         private Config providersConfig = Config.empty();
-        private MediaContext mediaContext = MediaContext.create();
+        // Config processing is always called so value will be initialized there.
+        private MediaContext mediaContext;
         private ContentEncodingContext contentEncodingContext = ContentEncodingContext.create();
 
         private boolean shutdownHook = true;
@@ -254,6 +255,11 @@ public interface WebServer {
             config.get("content-encoding")
                     .as(ContentEncodingContext::create)
                     .ifPresent(this::contentEncodingContext);
+            // Configure media support
+            config.get("media-support")
+                    .as(MediaContext::create)
+                    // MediaContext always needs to be refreshed after config change
+                    .ifPresentOrElse(this::mediaContext, () -> mediaContext(MediaContext.create()));
             // Store providers config node for later usage.
             providersConfig = config.get("connection-providers");
             return this;
--- a/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
+++ b/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
@@ -166,7 +166,7 @@ public interface WebServer {
                 = HelidonServiceLoader.builder(ServiceLoader.load(ServerConnectionProvider.class));
 
         private Config providersConfig = Config.empty();
-        // Config processing is always called so value will be initialized there.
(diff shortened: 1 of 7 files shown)


nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
barchetta, line 1: I understand having the implementation of this API driven through `MediaContext`. Makes sense, but what do you think about adding convenience APIs on `WebServer.Builder()`. My understanding is the new API is:
```
        WebServer.Builder wsBuilder = WebServer.builder()
                .config(config.get("server"))
                .mediaContext(MediaContext.builder()
                        .discoverServices(false)
                        .addMediaSupportProvider(new JsonpMediaSupportProvider())
                        .build());
```

I would like to see some convenience methods on `WebServer.Builder` for this that is more consistent with what we had in 3.x and also with 4.x `addFeature` (which is on `HttpRouting.builder`). For example something like:

```
JsonpMediaSupport jsonpMediaSupport = JsonpMediaSupport.builder().useSystemServices(false);  //or JsonpMediaSupport.create()

WebServer.Builder wsBuilder = WebServer.builder()
                .addMediaSupport(jsonpMediaSupport).build();
```

I have a new class here `JsonpMediaSupport` instead of `JsonpMediaSupportProvider` because the later is an implementation of an SPI (loaded by service loader) and maybe we don't want to force user's to know about `Provider` classes?

Tomas-Kraus: `JsonpMediaSupport jsonpMediaSupport = JsonpMediaSupport.builder().useSystemServices(false);`
Think that `useSystemServices(false)` should not be here. It's MediaContext (service loader) related option, not JSON-P media support specific settings/feature..

But I have another strong reason to use SPI: Nima `WebServer.Builder` is using the same API pattern to build `ContentEncodingContext` and also to add `ServerConnectionProvider` instances. We shall keep this builder API consistent.
So if we want to make changes, we have to change all SPI usages, not just media support.

nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
```diff
@@ -254,6 +255,11 @@ public Builder config(Config config) {
             config.get("content-encoding")
                     .as(ContentEncodingContext::create)
                     .ifPresent(this::contentEncodingContext);
+            // Configure media support
+            config.get("media-support")
+                    .as(MediaContext::create)
+                    // MediaContext always needs to be refreshed after config change
+                    .ifPresentOrElse(this::mediaContext, () -> mediaContext(MediaContext.create()));
```
tomas-langer, line 262: this should not use `ifPresentOrElse`, just `ifPresent`.
If there is no media-support in config, it should use whatever is configured by user explicitly. The current approach would always replace mediacontext with a default one, even if already configured.
So set it to `null` by default, and then in `build` method check if null, and if so, use default `MediaContext.create()`

nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
tomas-langer: Please create a follow up issue. We may want to separate this into an additional interface (to keep the API similar to what we have in reactive), but we should make sure it is aligned correctly. I think this PR brings something we need even without the shortcuts.

tomas-langer (changes requested):

nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/WebServer.java
Tomas-Kraus: done

Tomas-Kraus:
Added 2 new issues:

- #6420 to clean up the Builder code
- #6421 to implement reactive like builder API

tomas-langer (approved):

PR #6415, opened 2023-03-16
Error handling removed from the filter chain

### LINKED ISSUE #5394: Nima: Filters error handling
Filters - error handling is probably done for each filter separately, check if this could be simplified (see stack trace in a handler).

There is a chance that this cannot be changed, but it does create quite a deep stack trace...

Verdent:
Fixes #5394

### DIFF
--- a/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http/Filters.java
+++ b/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http/Filters.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http/Filters.java
```diff
@@ -111,7 +116,7 @@ public void proceed() {
                 return;
             }
             if (filters.hasNext()) {
-                errorHandlers.runWithErrorHandling(ctx, request, response, this::runNextFilter);
+                filters.next().filter(this, request, response);
```
tomas-langer, line 119: The problem I see with this new approach is that if an error handler is executed for routing, it would get executed again for the whole `executeFilters` once again.

I think that if the same exception that was thrown from routing would enter the second error handler (for filtering), the error handler should not be executed (same instance of exception).

tomas-langer (changes requested):

tomas-langer (approved):

PR #6428, opened 2023-03-17
[Nima] Set running flag to false in shutdownhook

wojtask9:
Currently shutdownhook doesn't set flag `running` to `false`.
Setting to `false` is useful because we can test that `WebServer.isRunning()`  in user `custom` hook


spericas:
@wojtask9 Not sure I follow the use case you have in mind. Could you include a unit test? Also shutdown hooks are executed in an undefined order by the JDK.

wojtask9:
@spericas 
Not particular use case. 
As you said `shoutdown hooks` are executed in an undefined order so there is no guarantee that this be useful but at least shows `truth` about current `webserver` status.

Also this align with method `stopIt()` where such flag is correctly assigned

https://github.com/helidon-io/helidon/blob/main/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/LoomServer.java#L204

spericas (approved):

PR #6440, opened 2023-03-21
Convert to SnakeYAML 2.0; add test

tjquinno:
See description here: #6384


tests/integration/gh-5792/.dockerignore
```diff
@@ -0,0 +1 @@
+target/*
```
barchetta, line 1: Is this file needed?

tests/integration/gh-5792/app.yaml
```diff
@@ -0,0 +1,50 @@
+#
```
barchetta, line 1: Is this file needed?

tests/integration/gh-5792/Dockerfile
```diff
@@ -0,0 +1,45 @@
+#
```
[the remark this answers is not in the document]
tjquinno: I just added all the files that the archetype generated. Given that the test does not use docker or native image, we don't need them. I'll remove them.

barchetta (approved):
LGTM

PR #6474, opened 2023-03-22
Upgrade Java to 20 GA

### LINKED ISSUE #6222: Adopt Java 20
Move from Java 19 to Java 20 when it becomes available.

tomas-langer:
Resolves #6222

- ~~GraalVM is not yet available - tested on dev version of Graal VM (Níma observability example with native image)~~
- ~~distribution may not be yet available~~


barchetta (dismissed):
This LGTM and the dependency changes LGTM too.

PR #6482, opened 2023-03-23
[3.x] Unauthenticated status code fix 

### LINKED ISSUE #6172: gRPC Security Handler returns PERMISSION_DENIED not UNAUTHENTICATED for failing authn
## Environment Details
* Helidon Version: 3.1.0
* Helidon SE
* JDK version: 17
----------

## Problem Description

I'm trying to enable authentication only (not authorization) for the gRPC security handler and it looks like it combines both the authn and authz checks together and sets PERMISSION_DENIED status if they fail.

My understanding is that if authn fails, a gRPC UNAUTHENTICATED status should be returned.

https://github.com/helidon-io/helidon/blob/e560ba08237bc5af4e6be2ef8cc121e779d551a8/security/integration/grpc/src/main/java/io/helidon/security/integration/grpc/GrpcSecurityHandler.java#L403

(It's kinda funny this is on line 403)

Verdent:
Fixes #6172


tomas-langer (approved):
Please create follow up issues for 2.x and 4.x

Issue #6481, opened 2023-03-23
Use GitHub Action for helidon-3.x branch


tvallin:
Move from Jenkins to GitHub Action

PR #6500, opened 2023-03-28
Use GitHub Action for helidon-3.x branch

tvallin:
fix https://github.com/helidon-io/helidon/issues/6481


.github/workflows/validate.yaml
```diff
@@ -0,0 +1,134 @@
...
+        uses: actions/setup-java@v3.6.0
+        with:
+          distribution: ${{ env.JAVA_DISTRO }}
+          java-version: ${{ env.JAVA_VERSION }}
+          cache: maven
+      - name: Build Helidon
+        run: etc/scripts/github-compile.sh
+      - name: JAR packaging
+        run: etc/scripts/test-packaging-jar.sh
+      - name: JLink packaging
```
tomas-langer, line 133: This is missing native image packaging tests

tomas-langer (changes requested):
The failing tests are ok in 4.x, so probably some changes were not backported to 3.x

tvallin:
Copyright is failing because changes made to a file were reverted. If you are okay with the final changes, I will squash the commit and copyright check should pass.

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - uses: actions/checkout@v3
+      - name: Set up JDK ${{ env.JAVA_VERSION }}
+        uses: actions/setup-java@v3.6.0
+        with:
+          distribution: ${{ env.JAVA_DISTRO }}
+          java-version: ${{ env.JAVA_VERSION }}
+          cache: maven
+      - name: Checkstyle
+        run: etc/scripts/checkstyle.sh
+  build:
```
tomas-langer, line 48: I am missing spotbugs

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,143 @@
...
+      - uses: actions/checkout@v3
+      - name: Set up JDK ${{ env.JAVA_VERSION }}
+        uses: actions/setup-java@v3.6.0
+        with:
+          distribution: ${{ env.JAVA_DISTRO }}
+          java-version: ${{ env.JAVA_VERSION }}
+          cache: maven
+      - name: Maven build
+        run: etc/scripts/github-build.sh
+  tck:
```
tomas-langer, line 64: is this running an TCK other than MP TCK? If not, which is it in a different place and called differently than in `main`?

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,143 @@
...
+        run: etc/scripts/test-archetypes.sh
+  packaging:
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        os: [ ubuntu-20.04, macos-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: graalvm/setup-graalvm@v1
```
tomas-langer, line 123: This sets up both graalvm and Java. This is confusing, and I am not sure which JDK would be used

.github/workflows/validate.yaml
tomas-langer, line 48: Also missing docs build

.github/workflows/validate.yaml
tomas-langer: OK, so it is only runnin gthe reactive streams TCK, but not MP TCK. MP TCK is a must during build as well

tomas-langer (changes requested):

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,143 @@
...
+          java-version: ${{ env.JAVA_VERSION }}
+          cache: maven
+      - name: Build Helidon
+        run: etc/scripts/github-compile.sh
+      - name: JAR packaging
+        run: etc/scripts/test-packaging-jar.sh
+      - name: JLink packaging
+        run: etc/scripts/test-packaging-jlink.sh
+      - name: Native-Image packaging
+        run: etc/scripts/test-packaging-native.sh
```
tomas-langer, line 143: All these scripts do a priming build, which should be done just once (through `githbug-compile.sh`), it can be removed from all of them.

tomas-langer (changes requested):

tomas-langer (dismissed):

romain-grecourt (approved):

Issue #6421, opened 2023-03-16
Níma: shortcut methods for media support(s)


Tomas-Kraus:
Issue #6278 follow up.

In reactive, we can configure media support directly on web server builder. This would require a new interface (e.g. `MediaSupport`) to be created by the provider.
So users can register `JsonSupport.create()` instead of creating a new instance of a service loader provider implementation

tomas-langer:
The new approach should support:
- customization of underlying features (custom ObjectMapper for Jackson, JSON-B and JSON-P as well)
- using config to set up the underlying features
- customizing supported media types

The config based approach would ideally work for the service providers as well, if feasible (e.g. if I construct my `MediaContext` from `Config`)

PR #6507, opened 2023-03-29
Issue #6421 - Nima media support

Tomas-Kraus:
Fixes issue #6421


nima/http/media/jsonp/src/main/java/io/helidon/nima/http/media/jsonp/JsonpMediaSupportProvider.java
```diff
@@ -16,127 +16,38 @@
...
 
     @Override
-    public <T> WriterResponse<T> writer(GenericType<T> type, WritableHeaders<?> requestHeaders) {
-        // if the type is JsonStructure, we support it and do not care about content type
-        // you provide json, you get json...
-        if (isSupportedType(type)) {
-            return new WriterResponse<>(SupportLevel.SUPPORTED, this::writer);
-        }
-        return WriterResponse.unsupported();
+    public double weight() {
```
tomas-langer, line 48: JSON-P has to have higher than both JSON-B and Jackson, but it does not need to be low, as it supports very specific types (`JsonObject`, `JsonArray` etc.)

nima/webserver/webserver/src/test/java/io/helidon/nima/webserver/WebServerConfigTest.java
```diff
@@ -148,7 +147,7 @@ void testMediaSupportManualConfigJson() throws IOException {
                 .config(config.get("server"))
                 .mediaContext(MediaContext.builder()
                         .discoverServices(false)
-                        .addMediaSupportProvider(new JsonpMediaSupportProvider())
+                        .addMediaSupport(new JsonpSupport())
```
tomas-langer, line 150: All MediaSupport implementation must have a private/package local/protected constructor only (whatever is the lowest one we need to use).
So you cannot create an instance like this (this should fail compilation). 
Please fix it for all supports, and use `create()` or `create(Config)` methods instead

tomas-langer (changes requested):

common/config/pom.xml
```diff
@@ -28,4 +28,39 @@
     <artifactId>helidon-common-config</artifactId>
     <name>Helidon Common Config</name>
 
+    <properties>
```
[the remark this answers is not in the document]
Tomas-Kraus: Removed

common/config/pom.xml
```diff
@@ -28,4 +28,39 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
```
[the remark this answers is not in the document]
Tomas-Kraus: Removed.

nima/webserver/webserver/src/test/java/io/helidon/nima/webserver/WebServerConfigTest.java
Tomas-Kraus: All except String support have private constructors + create factory method.
StringSupport class is being extended in tests so constructor has protected constructor.

nima/http/media/jsonp/src/main/java/io/helidon/nima/http/media/jsonp/JsonpMediaSupportProvider.java
Tomas-Kraus: OK. As I understand all the notes here and in the code, weights shall be JSON-P > Jackson > JsonB
and JSON-P shall be high, while Jackson and JsonB very low.
It's set to:

- JsonbMediaSupportProvider - 10
- JacksonMediaSupportProvider - 15
- JsonpMediaSupportProvider - 1000

tomas-langer (approved):

Issue #4928, opened 2022-09-21
JPA integration has trouble with nested transactions


ljnelson:
Container-managed JPA has issues with JTA transactions started within other JTA transactions.

ljnelson:
This issue is very, very hard (for me, anyway) to solve and has taken and will take a _significant_ amount of time to address properly.

Non-normative notes to myself mostly from a few weeks of false starts and dark alleys (do not rely on these notes as an indication of what I'm definitely going to fix, what I might fix or what I might not fix, nor as any commitment whatsoever):

- [ ] Start with pseudo-JTA layer; ensure transaction-specific connections are stored in a TSR and not by thread ID, since there can be multiple transactions per thread, not just one (but only one active one per thread)
  - [ ] Ensure this can be done in a backwards-compatible way; some classes involved are `public`
  - This goal can be completed on its own, ideally, and the existing code should Just Work™.
- [ ] Possibly change pseudo-JTA connections such that once they are associated with a suspended transaction they throw `IllegalStateException`s. It's not clear whether this is the proper behavior, or whether JDBC operations should be permitted (so long as autocommit is off, as it will be if a JTA transaction is suspended).
  - Not clear on how to get notified of transaction suspension since `Synchronization` does not offer these hooks and transaction suspension is not associated with CDI scope events denoting the beginning and end of the scope's lifecycle
- [ ] (Derivative goal) `JtaDataSource#getConnection()` should Just Work™, so that means that once the TSR stuff described above is in place ideally `getConnection()` will acquire a new connection from the delegate since transaction affinity will be different

I think that takes care of the JDBC/JTA part.  The key insight is multiple transactions per thread of which only one may be active at any point.

Next up, `CDITransactionScopedEntityManager`:

- [ ] Currently has a single delegate; needs a stack of them
- [ ] Pushing and popping governed by transaction beginning and ending. Something still feels off here but I can't yet put my finger on it
  - Need to preserve semantics that transaction "scope" is owned wholesale by Narayana, but I think if I'm understanding it all right that `@TransactionScoped`'s lifespan is that of the outermost non-suspended transaction.
  - Might need to fake a kind of "delegate" transaction scope "inside" the "real" one by having the `CDITransactionScopedEntityManager` observe the initialization of the scope; this assumes that Narayana blindly fires the scope initialization event whenever `tm.begin()` is called, which strictly speaking is not correct but may be quite useful

This comment is subject to editing or deletion at any time for any reason.

[...]

PR #6508, opened 2023-03-29
Improving JPA pom.xml as part of overall JPA refactoring

ljnelson:
This PR improves the JPA `pom.xml` to support further upcoming refactorings, as well as to permit testing using Eclipselink, Hibernate or both based on a user-supplied Maven property. It also corrects some erroneous dependency information and adds documentation.


ljnelson:
Tangentially related to #4928.

tjquinno (approved):
I bow to your diagnosis as represented in the useful comments in the pom! Yikes.

ljnelson:
> I bow to your diagnosis as represented in the useful comments in the pom! Yikes.

Yikes is right. Thanks for the approval.

PR #6511, opened 2023-03-29
3.x upgrade netty to 4.1.90.Final

barchetta:
Upgrades Netty to 4.1.90.Final
Uses `netty-bom` for dependency management instead of individual dependencies


ljnelson (approved):
LGTM

PR #6512, opened 2023-03-29
Adds an enabled flag to JpaExtension to permit subsequent refactoring and replacement

ljnelson:
This PR allows the `JpaExtension` to be disabled so that eventually it can coexist alongside its replacement. That way if anything goes wrong with the future machinery a user in the field can roll back to this implementation without waiting on a new release from Helidon or having to avoid upgrading.

This PR also brings the class into conformance with our development guidelines. Less stuffily, `final`s have been removed and comments have been formatted for easier reading.

### DIFF
--- a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
+++ b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
@@ -104,21 +104,20 @@ import static jakarta.interceptor.Interceptor.Priority.LIBRARY_AFTER;
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_BEFORE;
 
 /**
- * A {@linkplain Extension portable extension} normally instantiated
- * by the Java {@linkplain java.util.ServiceLoader service provider
- * infrastructure} that integrates the provider-independent parts of
- * <a href="https://javaee.github.io/tutorial/partpersist.html#BNBPY"
- * target="_parent">JPA</a> into CDI.
+ * A {@linkplain Extension portable extension} normally instantiated by the Java {@linkplain java.util.ServiceLoader
+ * service provider infrastructure} that integrates the provider-independent parts of <a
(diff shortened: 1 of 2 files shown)


ljnelson:
Tangentially related to #4928.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
```diff
@@ -337,31 +297,31 @@ public class JpaExtension implements Extension {
...
-        final String cn = JpaExtension.class.getName();
-        final String mn = "<init>";
+        String cn = JpaExtension.class.getName();
+        String mn = "<init>";
         if (LOGGER.isLoggable(Level.FINER)) {
             LOGGER.entering(cn, mn);
         }
+        this.enabled = Boolean.parseBoolean(System.getProperty(this.getClass().getName() + ".enabled", "true"));
+        if (LOGGER.isLoggable(Level.FINE) && !this.enabled) {
+            LOGGER.logp(Level.FINE, cn, mn, "{0} is not enabled", this.getClass().getName());
```
tjquinno, line 314: Tiny question... By default, this extension is enabled. Is there any point to changing "is not enabled" to "has been disabled" or something else like that (maybe even citing the property name the user has set??) to reinforce that it's disabled because the user did something?

tjquinno (commented):
Looks good. One possible tiny change to consider.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
ljnelson: I like it. Stand by.

ljnelson: I've made the suggested change. When the new machinery goes in, I'll probably change the message again since the extension will then be disabled by default.

tjquinno (approved):

Issue #6506, opened 2023-03-28
2.x, 3.x Java 20 support


barchetta:
ASM 9.4.0 supports Java 20
ASM 9.5.0 supports Java 21
https://asm.ow2.io/versions.html

# Helidon 2.x

EclipseLink 2.7.12 (depends on ASM 9.4.0)
Hibernate 5.6.15 (depends on ByteBuddy 1.12.18 depends on ASM 9.4.0

# Helidon 3.x

Uses EclipseLink 3.0.3. No newer version of 3.x available -- so upgrade ASM directly?
Hibernate 6.1.7.Final (depends on ByteBuddy 1.12.18 depends on ASM 9.4.0l
plugin.hibernate.enhance 6.1.7.Final

ljnelson:
As of this writing, Hibernate is not ready yet; the mainline depends on Byte Buddy 1.12.18 which uses ASM 9.4: https://github.com/raphw/byte-buddy/blob/byte-buddy-1.12.18/pom.xml#L71

ljnelson:
As of this writing, Eclipselink is ready for Java 20 but not ready yet for Java 21; the mainline depends on ASM 9.4: https://github.com/eclipse-ee4j/eclipselink/blob/8ac1935575d461d3a51672c92a16062645e11bb3/pom.xml#L188

PR #6514, opened 2023-03-29
3.x Upgrade hibernate to 6.1.7.Final and eclipselink asm to 9.4.0

barchetta:
Upgrade eclipselink asm to 9.4.0 and hibernate to 6.1.7

Note that there does not appear to be a 3.x eclipselink release with ASM 9.4.0 -- so we force upgrade of eclipselink asm instead. Once eclipselink 3.0.4 (or whatever) comes out we can remove the forced eclipselink ASM upgrade.

This enables use of Java 20 due to the upgrade of ASM to 9.4.0

See https://github.com/helidon-io/helidon/issues/6506

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -45,7 +45,8 @@
         <version.lib.cron-utils>9.1.6</version.lib.cron-utils>
         <version.lib.dropwizard.metrics>4.1.2</version.lib.dropwizard.metrics>
         <version.lib.eclipselink>3.0.3</version.lib.eclipselink>
-        <!-- Needed only to force upgrade to ASM 9.4.0. Once eclipselink is upgraded this can be removed -->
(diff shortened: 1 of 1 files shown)


dependencies/pom.xml
```diff
@@ -881,6 +883,12 @@
                 <artifactId>org.eclipse.persistence.jpa</artifactId>
                 <version>${version.lib.eclipselink}</version>
             </dependency>
+            <!-- Needed only to force upgrade to ASM 9.4.0. Once eclipselink is upgraded this can be removed -->
```
ljnelson, line 886: Maybe indicate _why_ we would want to upgrade ASM to 9.4.0 and why an upgrade of Eclipselink to some other version would let us remove this. I know it, you know it, but maybe a casual reader doesn't know it.

Issue #6338, opened 2023-03-02
Nima: Validate headers in ClientRequestImpl


klustria:
## Environment Details
* Helidon NIMA

----------

## Problem Description
This is spawned off of issue https://github.com/helidon-io/helidon/issues/5386. Explore adding header validation (configurable) for Client Request similar to Helidon SE WebClient.

klustria:
This will be resolved by PR: https://github.com/helidon-io/helidon/pull/6515

danielkec:
Follow up issue for HTTP/2 client #6547

klustria:
PR https://github.com/helidon-io/helidon/pull/6515 is completed, hence closing this issue.

PR #6515, opened 2023-03-30
Validate http request and response headers when using the webclient

klustria:
1. The main method for validating header value now exists in HTTP HeaderValue, so it can be reused in other places outside of WebClient, if needed.
2. HttpCallChainBase.writeHeaders will validate both the header name and values for the HTTP request using  Http.HeaderValue.validate().
3. Http1HeadersParser.readHeaders used for parsing response headers now includes validation of header name and values using Http.HeaderValue.validate().
4. Extensive unit test for each of the new features added.


klustria:
This PR is a potential fix for https://github.com/helidon-io/helidon/issues/6338

Verdent:
I feel like it is OK, but I would like to know also your opinion @tomas-langer

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/http1/ClientRequestImpl.java
```diff
@@ -356,6 +317,75 @@ ClientConnection getConnection(boolean keepAlive) {
         return connection;
     }
 
+    // See https://www.rfc-editor.org/rfc/rfc7230#section-3.2
+    private static void validateHeaderValues(HeaderValue header) throws IllegalArgumentException {
```
tomas-langer, line 321: I think that this method may be better suited to `HeaderValue` directly, as it represents an HTTP Header value, and the same rules apply to any version of HTTP I think.
Then we could use the same validation logic regardless of server/client

tomas-langer (changes requested):

klustria:
Please check the description section of this issue to see the list of items that are included in the change.

tomas-langer (commented):
We should not do the change in `LazyString`
I think validation should be done once for all headers after they are read from the stream (both on server and on client when enabled).
We could add `HeaderValue.validate(Validator)` method - and make it work nicely with multi-valued headers in the right way, rather then attempting to target `LazyString` specifically

klustria:
> We should not do the change in `LazyString`
> I think validation should be done once for all headers after they are read from the stream (both on server and on client when enabled).
> We could add `HeaderValue.validate(Validator)` method - and make it work nicely with multi-valued headers in the right way, rather then attempting to target `LazyString` specifically

With the new change, removed response header validation from LazyString and placed it in Http1HeaderParser.readHeaders after a header's name and value are parsed and using the same Http.HeaderValue.validate() that HttpCallChainBase.writeHeaders() uses to validate both the header name/values for the request Headers. Also remove name token validation in Http1HeaderParser.readHeaderName() as the name is already validated in Http.HeaderValue.validate().

tomas-langer (approved):

PR #6526, opened 2023-03-30
[3.x] Proxy now properly selects proxy settings from system properties

### LINKED ISSUE #6446: Unable to create a Proxy of ProxyType.SYSTEM
## Environment Details
* Helidon Version: 2.5.0
* Helidon MP
* JDK version: 17.0.3
* OS: MacOS

----------

## Problem Description
Unable to create a Proxy of ProxyType.SYSTEM. This is how I was able to workaround but it is not ideal since I'm not using the SystemSelector.
```
    static Optional<Proxy> createProxyFromSystemProperties() {
        String httpsProxy = System.getProperty("https.proxyHost");
        String host = httpsProxy == null ? System.getProperty("http.proxyHost") : httpsProxy;
        String noProxy = System.getProperty("http.nonProxyHosts");
        String proxyPort = System.getProperty("http.proxyPort");
        if (host != null) {
            Proxy.Builder proxyBuilder = Proxy.builder()
                    .type(Proxy.ProxyType.HTTP)
                    .host(host);

            if (noProxy != null) {
                Arrays.stream(noProxy.split(Pattern.quote("|")))
                        // Helidon NoProxy only works if the * is not provided
                        .map(s -> s.startsWith("*.") ? s.substring(1) : s)
                        .forEach(proxyBuilder::addNoProxy);
            }

            if (proxyPort != null) {
                proxyBuilder.port(Integer.parseInt(proxyPort));
            }
            return Optional.ofNullable(proxyBuilder.build());
        }
        return Optional.empty();
    }
```

## Steps to reproduce
I've created the following test classe in project generated by the helidon-quickstart-mp archetype to highlight the issue:
```
package io.helidon.examples.quickstart.mp;

import io.helidon.webclient.Proxy;

import java.lang.reflect.Field;

import static org.junit.jupiter.api.Assertions.assertEquals;
public class SystemProxyTest {
    @Test
    void staticMethodCreationTest() throws NoSuchFieldException, IllegalAccessException {
        // The javadoc says:
        // Create from environment and system properties.
        // Returns: a proxy instance configured based on this system settings
        Proxy proxy = Proxy.create();
        Field field = proxy.getClass().getDeclaredField("type");
        field.setAccessible(true);

        assertEquals(Proxy.ProxyType.SYSTEM, field.get(proxy));
    }

    @Test
    void builderCreationTest() throws NoSuchFieldException, IllegalAccessException {
        // The javadoc says:
        // Configure proxy from environment variables and system properties.
        // Params:
        // useIt – use system selector
        Proxy proxy = Proxy.builder()
                .useSystemSelector(true)
                .build();
        Field field = proxy.getClass().getDeclaredField("type");
        field.setAccessible(true);

        assertEquals(Proxy.ProxyType.SYSTEM, field.get(proxy));
    }
}
```
I've added the official javadoc in that code snippet to highlight that the behavior is different than expected.

The culprit seems to be this condition (which still exists in Helidon 3.0):
https://github.com/helidon-io/helidon/blob/84123e8caa7f6671efc396526404b2b39d0afbbf/webclient/webclient/src/main/java/io/helidon/webclient/Proxy.java#L429

Verdent:
Fixes #6446

### DIFF
--- a/config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
+++ b/config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
+++ b/config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.helidon.config.test.infra;
 
 import java.lang.annotation.ElementType;
--- a/config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
+++ b/config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
@@ -1,0 +1,17 @@
+package io.helidon.config.test.infra;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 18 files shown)


config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Annotation used to reset system properties to the state before test execution.
+ */
+@Retention(RetentionPolicy.RUNTIME)
```
tomas-langer, line 29: This is not the right location for this class. Please backport module `helidon-common-testing-junit5`, and mark existing classes that are now moved as deprecated

webclient/webclient/pom.xml
```diff
@@ -117,6 +117,12 @@
             <scope>provided</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
```
tomas-langer, line 120: This should not be here, as we are not testing config

tomas-langer: (see comment above - we need a new module)

tomas-langer (changes requested):

config/test-infrastructure/src/main/java/io/helidon/config/test/infra/ResetSystemProperties.java
Verdent: backported in the new commit

tomas-langer (approved):

PR #6549, opened 2023-04-04
Update MP OpenAPI to 3.1

### LINKED ISSUE #5820: 4.x: Upgrade MP OpenAPI to 3.1

tjquinno:
Resolves #5820 

Brief summary of the changes:

* Update Jandex version and references - The updated SmallRye OpenAPI code now uses 3.0.5 which now has the group ID `io.smallrye` instead of the former `org.jboss`. No change seems to be needed to the version of the Jandex maven plug-in.
* Several additional model objects in the MP OpenAPI API now implement `Extensible` which requires special handling when converting to and from YAML via SnakeYAML. 
* Include/exclude for scanning packages and classes changed from `Pattern` to strings. 
* Some adaptation to changes in SmallRye's `OpenApiConfig` interface and its implementation.
* Setting a Jersey server property based on a system property to (temporarily at least) work around a bug in the MP OpenAPI TCK. Some sub-resources are ambiguous in the TCK and Jersey fails by default. Our MP OpenAPI TCK driver `pom.xml` now sets a system property which `JaxRsService` and `JerseySupport` detect and set to suppress the failures.
* Update `hibernate-core` version to 6.2.0 (because it also uses Jandex and 6.2.0 works with Jandex 3)

### DIFF
--- a/archetypes/helidon/src/main/archetype/mp/common/common-mp.xml
+++ b/archetypes/helidon/src/main/archetype/mp/common/common-mp.xml
@@ -28,7 +28,7 @@
             <value key="parent-artifactId">helidon-mp</value>
             <list key="dependencies">
                 <map>
-                    <value key="groupId">org.jboss</value>
+                    <value key="groupId">io.smallrye</value>
                     <value key="artifactId">jandex</value>
                     <value key="scope">runtime</value>
                 </map>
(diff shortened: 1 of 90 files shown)


tjquinno:
Superseded by #6954

PR #6556, opened 2023-04-04
Fixed problem in AUTO_FLUSH backpressure strategy 

spericas:
Fixed problem in AUTO_FLUSH strategy that may result in pub-sub deadlock. Increment buffer sum before checking watermark and flushing.

### DIFF
--- a/microprofile/server/src/test/java/io/helidon/microprofile/server/AutoFlushTest.java
+++ b/microprofile/server/src/test/java/io/helidon/microprofile/server/AutoFlushTest.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


microprofile/server/src/test/java/io/helidon/microprofile/server/AutoFlushTest.java
```diff
@@ -0,0 +1,66 @@
...
+    @Inject
+    private WebTarget target;
+
+    @Path("auto-flush")
+    public static class AutoFlushResource {
+
+        @GET
+        public Response getDefaultMessage() {
+            return Response.ok((StreamingOutput) outputStream -> {
+                try (InputStream is = AutoFlushResource.class.getResourceAsStream("/zero.bin")) {
```
danielkec, line 48: Maybe generating the stream data on the fly would save 5Megs in our repo

spericas: Yes, I was debating about that too. I'll try it.

danielkec (approved):

PR #6571, opened 2023-04-06
Renamed receive() method to onMessage() in WsListener

spericas:
Renamed receive() method to onMessage() to make it consistent with the other methods in the interface, all of which use "on" as prefix. Part of doc issue #6467.


barchetta (approved):
LGTM

Issue #6567, opened 2023-04-05
Services <T> ServiceProvider<T> lookupFirst(ServiceInfoCriteria criteria) needs refactoring


trentjeff:
See comments on https://github.com/helidon-io/helidon/pull/6552/files/ecd9c3170d09dac84ebd6320bf50a54d698d6bb3#diff-af780da55376f950e23a7316b0e93035edf718a0dd0e69194e9a6e54a9b37c6d

`Services <T> ServiceProvider<T> lookupFirst(ServiceInfoCriteria criteria)` should be revisited to remove/replace the `<T>` usage on generics.

PR #6582, opened 2023-04-10
Fix pico lookup method returning <T> from generic criteria

trentjeff:
Fix for #6567

### DIFF
--- a/pico/runtime/src/main/java/io/helidon/pico/runtime/DefaultServices.java
+++ b/pico/runtime/src/main/java/io/helidon/pico/runtime/DefaultServices.java
@@ -216,15 +216,17 @@ class DefaultServices implements Services, ServiceBinder, Resettable {
     }
 
     @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public <T> Optional<ServiceProvider<T>> lookupFirst(Class<T> type,
                                                         boolean expected) {
         DefaultServiceInfoCriteria criteria = DefaultServiceInfoCriteria.builder()
                 .addContractImplemented(type.getName())
                 .build();
-        return lookupFirst(criteria, expected);
+        return (Optional) lookupFirst(criteria, expected);
     }
 
     @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public <T> Optional<ServiceProvider<T>> lookupFirst(Class<T> type,
                                                         String name,
                                                         boolean expected) {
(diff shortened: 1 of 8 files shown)


pico/runtime/src/main/java/io/helidon/pico/runtime/DefaultServices.java
```diff
@@ -216,46 +216,48 @@ public boolean reset(boolean deep) {
...
 
     @Override
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public <T> Optional<ServiceProvider<T>> lookupFirst(Class<T> type,
                                                         boolean expected) {
         DefaultServiceInfoCriteria criteria = DefaultServiceInfoCriteria.builder()
                 .addContractImplemented(type.getName())
                 .build();
-        return lookupFirst(criteria, expected);
+        return (Optional) lookupFirst(criteria, expected);
```
tomas-langer, line 225: This is exactly what I wanted to prevent. 
Is there a chance to add type to `ServiceInfoCriteria` instead? It could not use the generated builder, but it would be more user friendly

trentjeff: No, someone can just lookup RunLevel.STARTUP services for example.  Type is not required to be passed.

trentjeff: This change "fixes" the public contract at least, and the unfortunate casting is pushed down into the bowels of the impl.

tomas-langer: The default can be `ServiceInfoCriteria.Builder<Object>`, with a possibility to define a type (such as `<T> ServiceInfoCriteriaBuilder<T> contract(Class<T> type)`
This would allow us to make even internal implementation typed, while supporting use cases such as you describe

trentjeff: It would make it more ambiguous. First, you would need to then choose to put the `T` on the `serviceTypeName`, `contractImplemented`, or `externalContractImplemented`. Also note that these do not use `Class` today - they use String.  This seems like a real bad idea to me - it will lead to confusion.

trentjeff: `Contract` is not a singleton, it is a collection. Also, we would have to make analogous changes on `ServiceInfo` and possibly more builder types. It doesn't make sense.

tomas-langer: I think this would be more user friendly if we found a way, let's see once we use it ourself.

PR #6584, opened 2023-04-10
Adds more classes as part of overall JPA refactoring effort

ljnelson:


ljnelson:
This PR adds the "new generation" of JPA extension machinery to `integrations/cdi/jpa-cdi`.  In this PR it is not enabled. `PersistenceExtension` will be the new entry point and will replace the current `JpaExtension`. The other classes added here are replacements (of sorts) for some of the older classes. A subsequent PR will deprecate the older stuff and adjust tests.

arjav-desai (approved):
LGTM

PR #6597, opened 2023-04-12
Update documentation of composite provider flag.

### LINKED ISSUE #6173: Improve documentation for CompositeProviderFlag
The javadoc https://helidon.io/docs/v2/apidocs/io.helidon.security/io/helidon/security/CompositeProviderFlag offers very little guidance in helping users choose the appropriate configuration when composite providers are configured.

One recent question from slack has the user configuring two providers: (1) OID, with a fallback of (@) Basic-Auth. The user was configuring both providers as "MAY_FAIL". However, the appropriate configuration should be having both configured as "SUFFICIENT" with optional flag set to true.

This configuration, however, is entirely not obvious from reading the documentation. We should therefore improve the documentation - probably by providing examples for when/how it is appropriate for using compositions that include MAY_FAIL, SUFFICIENT, and optional=true - as well as the other flags available.

Also related to this is considering how the OID provider was not available due to a network outage (or perhaps a 404). We should also mention this case in particular to indicate something like "a failure to reach the security provider" will be treated as a failure, etc.

bad:

```
security:
  enabled: true
  provider-policy:
    type: "COMPOSITE"
    name: "composite-policy"
    default: true
    authentication:
      - name: "basic-authenticator"
        flag: "MAY_FAIL"
      - name: "oidc-authenticator"
        flag: "MAY_FAIL"
  providers:
  - name: "basic-authenticator"
    http-basic-auth:
      realm: "helidon"
      users:
      - login: "uimuser"
        password: "changeit"
  - name: "oidc-authenticator"
    oidc:
      identity-uri: "http://ohs1.oamdomain.org:30305"
      base-scopes: "FedRserver.Info"
      client-id: topologyClient
      client-secret: Welcome1
      token-endpoint-auth: CLIENT_SECRET_POST
      cookie-name: "OIDC_SESSION"
      cookie-same-site: "Lax"
      header-use: true
      audience: "FedRserver"
      redirect: true
      redirect-uri: "/sr/test/sol005Adapter"
```

Good (partially; needs optional: true for oidc):
```
security:
  enabled: true
  provider-policy:
    type: "COMPOSITE"
    name: "composite-policy"
    default: true
    authentication:
      - name: "basic-authenticator"
        flag: "SUFFICIENT"
      - name: "oidc-authenticator"
        flag: "SUFFICIENT"
```

tomas-langer:
Resolves #6173

### DIFF
--- a/security/security/src/main/java/io/helidon/security/CompositeProviderFlag.java
+++ b/security/security/src/main/java/io/helidon/security/CompositeProviderFlag.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,12 +27,23 @@ import java.util.Set;
 public enum CompositeProviderFlag {
     /**
      * Provider may fail, nothing happens if it does.
+     * <p>
+     * If succeeds, continues to the next provider.
+     * To create a provider that may fail, yet that is sufficient,
+     * please configure the flag to be {@link #SUFFICIENT}, and set the provider itself ot be optional
+     * (most Helidon providers support {@code optional} configuration option), so the provider abstains instead of fails
+     * when user cannot be authenticated using it.
      */
     MAY_FAIL(EnumSet.of(SecurityResponse.SecurityStatus.SUCCESS,
                         SecurityResponse.SecurityStatus.ABSTAIN,
                         SecurityResponse.SecurityStatus.FAILURE)),
     /**
      * If succeeds, no further providers are called, otherwise same as optional.
+     * <p>
+     * To create a provider that may fail, yet that is sufficient,
+     * use this flag, and set the provider itself ot be optional
+     * (most Helidon providers support {@code optional} configuration option, so the provider abstains instead of fails
+     * when user cannot be authenticated using it.
      */
     SUFFICIENT(EnumSet.of(SecurityResponse.SecurityStatus.SUCCESS, SecurityResponse.SecurityStatus.ABSTAIN)),
     /**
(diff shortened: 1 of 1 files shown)


security/security/src/main/java/io/helidon/security/CompositeProviderFlag.java
```diff
@@ -27,12 +27,23 @@
 public enum CompositeProviderFlag {
     /**
      * Provider may fail, nothing happens if it does.
+     * <p>
+     * If succeeds, continues to the next provider.
+     * To create a provider that may fail, yet that is sufficient,
+     * please configure the flag to be {@link #SUFFICIENT}, and set the provider itself ot be optional
```
ljnelson, line 33: typo

security/security/src/main/java/io/helidon/security/CompositeProviderFlag.java
```diff
@@ -27,12 +27,23 @@
...
+     * when user cannot be authenticated using it.
      */
     MAY_FAIL(EnumSet.of(SecurityResponse.SecurityStatus.SUCCESS,
                         SecurityResponse.SecurityStatus.ABSTAIN,
                         SecurityResponse.SecurityStatus.FAILURE)),
     /**
      * If succeeds, no further providers are called, otherwise same as optional.
+     * <p>
+     * To create a provider that may fail, yet that is sufficient,
+     * use this flag, and set the provider itself ot be optional
```
ljnelson, line 44: Typo

security/security/src/main/java/io/helidon/security/CompositeProviderFlag.java
```diff
@@ -27,12 +27,23 @@
...
+     * (most Helidon providers support {@code optional} configuration option, so the provider abstains instead of fails
```
ljnelson, line 45: Unclosed parenthesis

tomas-langer: fixed

Verdent (approved):

Issue #6600, opened 2023-04-12
[Archetype] - Create file validations


tvallin:
## Environment Details
* Helidon Version: 3.x - 4.x
* Helidon SE - MP - Nima
* JDK version: n/a
* OS: All
* Docker version (if applicable):

----------

## Problem Description

There are issues in archetype generated files (e.g. *.mustache).

## Steps to reproduce

```batch
cd archetypes
mvn install
```

Check generated files under `target/projects`

PR #6608, opened 2023-04-13
Create file validations

tvallin:
See https://github.com/helidon-io/helidon/issues/6600

### DIFF
--- a/archetypes/helidon/src/main/archetype/mp/custom/custom-mp.xml
+++ b/archetypes/helidon/src/main/archetype/mp/custom/custom-mp.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2021, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2021, 2023 Oracle and/or its affiliates.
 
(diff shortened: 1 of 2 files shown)


archetypes/helidon/src/main/archetype/mp/custom/custom-mp.xml
```diff
@@ -31,12 +31,6 @@
     <exec src="/mp/common/common-mp.xml"/>
     <source src="/mp/custom/observability.xml"/>
     <output>
-        <files if="${security}">
```
romain-grecourt, line 34: What's the issue with this block ? It's not a `.mustache` neither does the path contain `__pkg__`.

tvallin: If `security` is selected but does not contain `http-signature`, it results in copying the whole directory for some reason. Looks like a issue in build-tools. I did not investigate yet. I with create a unit test and check the behavior.

Other reason to remove this block is that in `common/security.xml, the file `keystore.p12` is already included so it does not make sense to have this logic in `custom-mp.xml`.

romain-grecourt (approved):

Issue #5713, opened 2022-12-16
Empty response shouln't be chunked


danielkec:
We should consider setting 0 content length to empty response by default to avoid chunked response with empty chunk. 

```java
res.status(301)
   .addHeader(Http.Header.LOCATION, "/")
   .send()
```

```
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 48 54 54 50 2f 31 2e 31 20 33 30 31 20 4d 6f 76 |HTTP/1.1 301 Mov|
|00000010| 65 64 20 50 65 72 6d 61 6e 65 6e 74 6c 79 0d 0a |ed Permanently..|
|00000020| 44 61 74 65 3a 20 46 72 69 2c 20 31 36 20 44 65 |Date: Fri, 16 De|
|00000030| 63 20 32 30 32 32 20 32 32 3a 33 32 3a 33 37 20 |c 2022 22:32:37 |
|00000040| 2b 30 31 30 30 0d 0a 4c 6f 63 61 74 69 6f 6e 3a |+0100..Location:|
|00000050| 20 2f 0d 0a 74 72 61 6e 73 66 65 72 2d 65 6e 63 | /..transfer-enc|
|00000060| 6f 64 69 6e 67 3a 20 63 68 75 6e 6b 65 64 0d 0a |oding: chunked..|
|00000070| 0d 0a                                           |..              |
+--------+-------------------------------------------------+----------------+


         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 30 0d 0a 0d 0a                                  |0....           |
+--------+-------------------------------------------------+----------------+
```

```java
res.status(301)
   .addHeader(Http.Header.LOCATION, "/")
   // force 301 to not use chunked encoding
   .addHeader(Http.Header.CONTENT_LENGTH, "0")
   .send()
```

```
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 48 54 54 50 2f 31 2e 31 20 33 30 31 20 4d 6f 76 |HTTP/1.1 301 Mov|
[...]

PR #6637, opened 2023-04-14
Response should not be chunked if there is no entity

klustria:
If response has no entity, make sure that `transfer-encoding:chunked` is not set in the header, but instead should have `content-length:0`.


klustria:
Goal of this PR is to resolve issue reported in https://github.com/helidon-io/helidon/issues/5713

klustria:
Hi @spericas, I made a change to this PR to simplify lengthOptimization and remove noEntity which I added previously to determine if there’s any data being sent. Our goal as we’ve discussed in you last review is to reduce multiple global flags to make the decision on when to do lengthOptimization. For this change, lengthOptimization will take place if there is 0 or 1 entity packet sent and response is not TransferEncodingChunked and not an SSE Event Stream. Previously, we only have lengthOptimization for 200 response, but I think we should do the same for other responses hence it was not part of the condition anymore.  I checked NIMA equivalent in nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1ServerResponse.java and I don’t see such restriction. I also added some logging in the code as we discussed. One question about logging though is it seems we always sorround it with the condition
```if (LOGGER.isLoggable(Level.FINER)) {```
I thought that below statement
```LOGGER.finer(() -> log(...```
would be enough to determine if it will be logged or not, hence my change did not have that condition under that assumption. Do you know why all the logs are enclosed with that condition?

File: webserver/webserver/src/main/java/io/helidon/webserver/BareResponseImpl.java [hunk 1]
```diff
@@ -202,7 +203,7 @@ public void writeStatusAndHeaders(Http.ResponseStatus status, Map<String, List<S
         boolean lengthSet = HttpUtil.isContentLengthSet(response);
         forcedChunked = HttpUtil.isTransferEncodingChunked(response);
         if (!lengthSet) {
-            lengthOptimization = status.code() == Http.Status.OK_200.code() && !forcedChunked && !isSseEventStream(headers);
+            lengthOptimization = isSseEventStream(headers) ? false : lengthOptimization;
```
spericas, line 206: lengthOptimization = !isSseEventStream(headers)

spericas: Or is it that we can do `lengthOptimization = !isSseEventStream(headers) && !forcedChunk` and make `forcedChunk` just a local variable instead of an instance variable. Seems like there's only one test for it here `lengthOptimization && !forcedChunked`, but I'm not 100% sure.

klustria: Elimintarted global variable `forcedChunked`. Previously, it exists to check if user added transfer-encoding-chunked into his response handler and only required when lengthOptimization is being determined. Because we set the response to transfer-encoding-chunked at the start of response processing if there is no content-length, hence why that state is needed via that global variable when we finally checked if lengthOptimization will take place. The new change removes the setting of the response to  transfer-encoding-chunked when length does not exist at the start (i.e. in `writeStatusAndHeaders`) and moves it to a part when header is set in onNextPipe where we know lengthOptimization is not required anymore. This way we can make sure that at the end in writeLastContent() that we can check if the response was forced chunked via the HttpUtil.isTransferEncodingChunked when we determine if lengthOptimization is needed.

spericas (approved):

Issue #6630, opened 2023-04-14
4.x: Inheritance in `DefaulQualifierAndValue.builder()` does not work


tomas-langer:
The following code does not compile:
```
QualifierAndValue qav = DefaultQualifierAndValue.builder()
                        .type(Named.class)
                        .value(cacheRecord.typeName().name() + "." + cacheRecord.methodName())
                        .build();
```
QualifierAndValue is actually required in some APIs (such as service infor criteria).

PR #6639, opened 2023-04-17
Inheritance in DefaulQualifierAndValue

trentjeff:


trentjeff:
Resolves https://github.com/helidon-io/helidon/issues/6630

common/types/src/main/java/io/helidon/common/types/DefaultAnnotationAndValue.java
```diff
@@ -227,7 +227,7 @@ public DefaultAnnotationAndValue build() {
         public Builder typeName(TypeName val) {
             Objects.requireNonNull(val);
             this.typeName = val;
-            return this;
+            return identity();
```
tomas-langer, line 230: This is not needed, as you are not returning a generic. If you use this "hackish" solution, no change required here.

pico/api/src/main/java/io/helidon/pico/api/DefaultQualifierAndValue.java
```diff
@@ -162,6 +162,26 @@ public static class Builder extends DefaultAnnotationAndValue.Builder {
         protected Builder() {
         }
 
+        @Override
+        public Builder typeName(TypeName val) {
+            return (Builder) super.typeName(val);
```
tomas-langer, line 167: This is the "hack" - as you do not have correctly generic parent, you cannot do this 100% right.
To avoid this cast, you can use
```
super.typeName(val);
return this;
```
to work around the explicit cast.

tomas-langer (changes requested):

common/types/src/main/java/io/helidon/common/types/DefaultAnnotationAndValue.java
trentjeff: pushed

pico/api/src/main/java/io/helidon/pico/api/DefaultQualifierAndValue.java
trentjeff: pushed

trentjeff: pushed

tomas-langer (approved):

Issue #6632, opened 2023-04-14
4.x: Pico interceptors are invoked even for methods that are not annotated


tomas-langer:
FallbackInterceptor is invoked even for constructor, even though it is not annotated with FtFallback (examples/nima/fault-tolerance)

Example:
https://github.com/tomas-langer/helidon/tree/ft-on-pico/examples/nima/fault-tolerance

Put a breakpoint in the FallbackInterceptor `proceed` method.

Issue #6629, opened 2023-04-14
4.x: Pico interception fails when proceed invoked more than once


tomas-langer:
Implementing retry mechanism - an interceptor.
When using retry.invoke(() -> chain.proceed(args))
I get the following exception on retry:
java.lang.IllegalStateException: Unknown call type: java.lang.String greetNamed
	at io.helidon.pico.runtime.Invocation.proceed(Invocation.java:127)
    at io.helidon.nima.faulttolerance.RetryInterceptor.lambda$proceed$0(RetryInterceptor.java:53)

See https://github.com/tomas-langer/helidon/tree/ft-on-pico/examples/nima/fault-tolerance
build and run the example, access endpoint at
`curl -i http://localhost:44193/greet/Github?throw=true`
Check the standard output of the process

PR #6642, opened 2023-04-18
Pico interceptor arguments are now correctly handled.

### LINKED ISSUE #6634: 4.x: Pico interceptor arguments are not correctly handled
the arguments to method `proceed(ctx, chain, args)` must be the arguments provided at the call site
chain.proceed(args) - the arguments sent to this method must be used to invoke the target method
right now, I always get an empty array, and the method is invoked with the same arguments the call was made (e.g. I cannot change arguments in interceptor, and I cannot use the arguments to call other methods)
See `FallbackInterceptor` - line 54, where the FallbackMethod must get the same arguments for the fallback to work. Now fails because I expect the array to contain two parameters.

Reproducer: https://github.com/tomas-langer/helidon/tree/ft-on-pico/examples/nima/fault-tolerance
Call `curl -i http://localhost:44193/greet/Github?throw=true` (port is random due to another problem)

tomas-langer:
Tests for all interceptor issues (disabled/commented out for other reported issues)

Resolves #6634 

Related to #6632 #6629

### DIFF
--- a/pico/tests/interception/pom.xml
+++ b/pico/tests/interception/pom.xml
@@ -1,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2023 Oracle and/or its affiliates.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>io.helidon.pico.tests</groupId>
+        <artifactId>helidon-pico-tests-project</artifactId>
+        <version>4.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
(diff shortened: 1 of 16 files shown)


pico/tests/interception/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+
+    <artifactId>helidon-pico-tests-interception</artifactId>
+    <name>Helidon Pico Test Interception</name>
+    <description>Test interception</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.pico</groupId>
+            <artifactId>helidon-pico-processor</artifactId>
+            <optional>true</optional>
```
trentjeff, line 39: add a note that this is for reactor ordering.

pico/tests/interception/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.pico</groupId>
```
trentjeff, line 42: you don't need pico maven module dep for this module.

pico/tests/interception/src/test/java/io/helidon/pico/tests/interception/InterceptionTest.java
```diff
@@ -0,0 +1,151 @@
...
+                () -> assertThat("Repeating last call", repeating, notNullValue()),
+                () -> assertThat("Repeating last call", repeating.methodName(), is("intercepted")),
+                () -> assertThat("Repeating last call", repeating.args(), is(new Object[] {"mod_hello", true, false, false}))
+        );
+    }
+
+    @Disabled("Known problem - issue #6629")
+    @Test
+    void testRepeat() {
+        String response = service.intercepted("hello", false, true, false);
```
trentjeff, line 131: instead of all of these boolean switches, can we just use a Builder arg - this will be more flexible for other testing scenarios...

tomas-langer: I would prefer to use an array of parameters, to make sure the parameters are correctly handled in the array

tomas-langer: Also adding builder is testing an additional component - this is as much "clean" test for interceptors as possible.

trentjeff (approved):

PR #6643, opened 2023-04-18
[4.x]: Use Hamcrest assertions instead of JUnit in tests/integration/jms (#1749)

Captain1653:
Part of #1749


Captain1653:
@barchetta [The same](https://github.com/helidon-io/helidon/pull/6645#issuecomment-1513756762) for 4.x ?

romain-grecourt (approved):

Issue #6516, opened 2023-03-30
Create Github Action for helidon-2.x branch


tvallin:
Move from Jenkins to GA

PR #6645, opened 2023-04-18
[2.x]: Use Hamcrest assertions instead of JUnit in tests/integration/jms (#1749)

Captain1653:
Part of #1749


Captain1653:
Hmm... Should GA actions be started after https://github.com/helidon-io/helidon/issues/6516 ? Did I get something wrong?

barchetta:
@Captain1653 you did nothing wrong! We had to update branch protection rules with the new GitHub Action based checks. Should work now 🤞

romain-grecourt (approved):

Issue #6647, opened 2023-04-18
4.x: Inject - interceptors should be called for all annotations (interface and type)


tomas-langer:
Currently annotation triggers on interface are ignored, and only annotation triggers on implementing type are used.
Also interceptor of constructor is not triggered correctly when creating an instance injected through contract.

See 
pico/tests/interception/src/main/java/io/helidon/pico/tests/interception/OtherContract.java

its implementation and  test
pico/tests/interception/src/test/java/io/helidon/pico/tests/interception/InterfaceInterceptionTest.java

both added by #6648 

The test also validates that interceptors are called for constructor if annotated with an interceptor trigger

tomas-langer:
In new service inject, interface interception is validated by `io.helidon.service.tests.inject.interception.InterfaceInterceptionTest`

tomas-langer:
And constructor interception as well.

PR #6648, opened 2023-04-18
Constructors are now intercepted with appropriate interceptors only.

### LINKED ISSUE #6632: 4.x: Pico interceptors are invoked even for methods that are not annotated

tomas-langer:
Resolves #6632 

Added tests for additional use cases (commented out the failing ones + added issue)

Created #6647

### DIFF
--- a/pico/runtime/src/main/java/io/helidon/pico/runtime/Invocation.java
+++ b/pico/runtime/src/main/java/io/helidon/pico/runtime/Invocation.java
@@ -118,7 +118,11 @@ public class Invocation<V> implements Interceptor.Chain<V> {
 
     @Override
     public V proceed(Object... args) {
-        if (!interceptorIterator.hasNext()) {
+        if (interceptorIterator.hasNext()) {
+            return interceptorIterator.next()
+                    .get()
+                    .proceed(ctx, this, args);
+        } else {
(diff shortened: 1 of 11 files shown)


pico/runtime/src/main/java/io/helidon/pico/runtime/Invocation.java
```diff
@@ -118,18 +118,18 @@ public static <T> List<Provider<T>> mergeAndCollapse(List<Provider<T>>... lists)
 
     @Override
     public V proceed(Object... args) {
-        if (!interceptorIterator.hasNext()) {
+        if (interceptorIterator.hasNext()) {
```
tomas-langer, line 121: it is easier to reason about an if statement if negation is not used (and I had to debug this method a lot).
This is a best practice we use in Helidon, fixed this instance as PR is related.

trentjeff (approved):

PR #6658, opened 2023-04-20
Interceptor Redo

trentjeff:
Resolves Issue #6629


pico/api/src/main/java/io/helidon/pico/api/Interceptor.java
```diff
@@ -50,6 +51,7 @@ public interface Interceptor {
          *
          * @param args the arguments passed
          * @return the result of the call
+         * @throws InvocationException if there is an errors during invocation chain processing
```
tomas-langer, line 54: "there is an error during" or "there are errors during"

pico/api/src/main/java/io/helidon/pico/api/Interceptor.java
```diff
@@ -32,6 +32,7 @@ public interface Interceptor {
      * @param args  the arguments to the call
      * @param <V>   the return value type (or {@link Void} for void method elements)
      * @return the return value to the caller
+     * @throws InvocationException if there is an errors during invocation chain processing
```
tomas-langer, line 35: either `is an error` or `are errors` - typo

tomas-langer (changes requested):
This PR should also uncomment some sections, and remove `@Disable` from a test in
`InterceptionTest` and `InterfaceInterceptionTest` in `helidon-pico-tests-interception`

pico/api/src/main/java/io/helidon/pico/api/Interceptor.java
trentjeff: pushed

pico/api/src/main/java/io/helidon/pico/api/Interceptor.java
trentjeff: pushed

trentjeff:
My assumptions are a little different from yours.

I expected that once the target of interception was successfully called once, then game over. The target can't be called again. However, If it threw an exception then it can be called again - but if it was completed then game over.

You assume you should be able to call the target an infinite number of times really.

I think both viewpoints have some merit, and perhaps we should make this into a qualifier type annotation on interceptors or something?

pico/tests/interception/src/main/java/io/helidon/pico/tests/interception/ReturningInterceptor.java
```diff
@@ -37,6 +37,7 @@ static Invocation lastCall() {
         return LAST_CALL.getAndSet(null);
     }
 
+    @SuppressWarnings("unchecked")
```
trentjeff, line 40: Curiously the SuppressWarnings("ALL") doesn't catch the unchecked vararg warnings in the generated interceptor - this did the trick though.

tomas-langer:
A qualifier such as `@Repeatable` that could be used on other annotations to mark them as "yes, we may want to invoke interceptor chain more than once" - I do not have problem with that

trentjeff:
Logged https://github.com/helidon-io/helidon/issues/6664 - Add @Repeatable meta annotation qualifier on InterceptorTrigger

tomas-langer (approved):

PR #6671, opened 2023-04-21
SE/Reactive archetype removal

### LINKED ISSUE #6655: 4.x: SE/Reactive archetype removal

tvallin:
fix #6655


archetypes/helidon/src/main/archetype/mp/custom/files/src/test/java/__pkg__/TestCORS.java.mustache
```diff
@@ -1,141 +1,108 @@
...
+import static org.hamcrest.Matchers.is;
 
+@HelidonTest
 class TestCORS {
 
-    private static WebClient client;
-    private static Server server;
+    @Inject
+    private WebTarget target;
+    private static final String ALLOW_RESTRICTED_HEADERS = "sun.net.http.allowRestrictedHeaders";
```
romain-grecourt, line 27: Where is that coming from ?

tvallin: You can see it from `BaseCrossOriginTest`. The system property is needed to be set to allow restricted header with Jersey client.

romain-grecourt: How did it work without that ?

tvallin: I think the reason is that reactive webClient is using `HelidonConnector` while now it is using the Jersey default `HttpUrlConnector` where this property manage restricted headers.

romain-grecourt (approved):

Issue #6674, opened 2023-04-21
4.x: Config driven and pico @Named fixes


tomas-langer:
The config driven should work as follows

configuration yaml:
```
server:
  port: 8080
ft:
  asyncs:
   first: 
    executor: "exec"
   second:
    executor: "service"
  bulkheads:
    - queue: 10
```

configuration beans:
```
@ConfigBean("ft.asyncs") // read it from ft.asyncs node, repeatable
AsyncConfig {..}

@ConfigBean("ft.bulheads") // read it from ft.bulkheads node, repeatable
BulkheadConfig {..}

@ConfigBean(value = "server", repeatable = false) // read it from server, not repeatable
ServerConfig {...}
```

When processing, the code to create the beans should be similar to this pseudo:
```
Config config; // located at the node of the config bean, such as `ft.asyncs`
if (configBeanAnnotation.repeatable()) {
  if (config.isList()) {
    - each node in the list is a config bean instance, named by the index (config.key().name()), we may want to introduce some custom config key to explicitly define a name, such as __name
  } else {
    - each child node is a config bean instance, named by the config node name (e.g. "first" and "second" for async)
    config.asNodeList().ifPresent(it -> it.forEach(configBeanCfg -> {
      String name = configBeanCfg.key().name(); // to be used with @Named
      String fullyQualifiedKey = configBeanCfg.key().toString(); // can also create a Named instance
  }
} else {
  // this is a non-repeatable config bean, just create a single instance from this place, not named (e.g. cannot be injected using @Named, as conceptually we cannot give it a name, we may support @default name + no name if we want to
}
```

To get instaces in Pico:
```
@Inject //config driven by ServerConfig
Server server;

@Inject //config driven by AsyncConfig
@Named("first")
Async firstAsync;

@Inject 
@Named("0") // index based naming - not much usable, but maybe somebody is just interested in the whole list and does not want to inject by name
Bulkhead bulkhead;
```

trentjeff:
After #6680 is merged, there will be this todo item:

```
    @Disabled("Will be addressed in #6674")
    @Test
    void namedConfiguredServices() {
        ConfigBeanRegistry cbr = (ConfigBeanRegistry) ConfigBeanRegistryHolder.configBeanRegistry().orElseThrow();
        Map<String, Collection<GeneratedConfigBean>> all = cbr.allConfigBeans();

//        System.out.println(all);
    }
```

In `NamedConfiguredByTest`.

PR #6680, opened 2023-04-24
GeneratedConfigBean should be easily available from Config-driven-services 

trentjeff:
This is a pre-req to issue #6674

### DIFF
--- a/builder/builder-config/src/main/java/io/helidon/builder/config/spi/GeneratedConfigBean.java
+++ b/builder/builder-config/src/main/java/io/helidon/builder/config/spi/GeneratedConfigBean.java
@@ -16,6 +16,10 @@
 
 package io.helidon.builder.config.spi;
 
+import java.util.Optional;
+
+import io.helidon.common.config.Config;
+
 /**
  * Every {@link io.helidon.builder.config.ConfigBean}-annotated type will also implement this contract.
  */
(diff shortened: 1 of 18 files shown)


builder/builder-config/src/main/java/io/helidon/builder/config/spi/GeneratedConfigBean.java
```diff
@@ -28,17 +32,25 @@ public interface GeneratedConfigBean extends GeneratedConfigBeanCommon {
...
 
     /**
-     * Set the instance id of this config bean.
+     * Returns the instance id assigned to this bean.
+     * Note that the instance id is typically assigned the {@link Config#key()} in most circumstances.
      *
-     * @param val the new instance identifier
+     * @return the instance id assigned to this bean
      */
-    void __instanceId(String val);
```
trentjeff, line 35: note: The setter was taken out of this interface and pushed down to the abstract impl.

builder/processor-tools/src/main/java/io/helidon/builder/processor/tools/GenerateMethod.java
```diff
@@ -67,6 +67,10 @@ static void internalMetaAttributes(StringBuilder builder) {
         builder.append("\tpublic static Map<String, Map<String, Object>> __metaAttributes() {\n"
                                + "\t\treturn ").append(BodyContext.TAG_META_PROPS).append(";\n"
```
trentjeff, line 68: Note: unfortunately we still need the static approach for other use cases.

tomas-langer (approved):

PR #6685, opened 2023-04-25
Update grpc-java version to 1.54.1 and clean up grpc tests

thegridman:
This PR is to update grpc-java to the latest version, which depends on a version of Netty closer to the one Helidon now depends on.


barchetta (approved):
LGTM. Thanks!

PR #6725, opened 2023-04-28
[3.x] LRA coordinator docker fix

### LINKED ISSUE #6470: [3.x] LRA Coordinator Docker build points to GitHub main branch
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version: 17
* OS: mac
* Docker version: Latest
 
----------

If we take `lra/coordinator/server/Dockerfile` in the `main` branch and in `helidon-3.x` they both have arguments pointing to the `main` branch:
```
ARG HELIDON_BRANCH=master
ARG HELIDON_REPOSITORY=oracle/helidon
```

This means, that even if I switch to `helidon-3.x` branch, it will still download 4.x dependencies during the docker build and fail will jdk-19 incompatibility issues.
So.. for `helidon-3.x` this ARG should change to point different branch in GitHub.

danielkec:
Fixes #6470

### DIFF
--- a/lra/coordinator/server/Dockerfile
+++ b/lra/coordinator/server/Dockerfile
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2021 Oracle and/or its affiliates.
+# Copyright (c) 2021, 2023 Oracle and/or its affiliates.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


lra/coordinator/server/Dockerfile
```diff
@@ -14,28 +14,34 @@
 # limitations under the License.
 #
 
-FROM maven:3.6.3-openjdk-17-slim as build
+FROM container-registry.oracle.com/java/openjdk:17 as build
```
barchetta, line 17: `container-registry.oracle.com/java/jdk-no-fee-term:17` will be a newer version and should still be freely available (no login, no license acceptance). Might want to give that a try.

Verdent (approved):

PR #6781, opened 2023-05-05
4.x: Truncate example token in README so it is not valid

barchetta:


ljnelson (approved):
Tough one to approve but I'll let it go this one time

PR #6790, opened 2023-05-09
Removed old bookstore-se app and renamed bookstore-nima to bookstore-se

spericas:
- Removed old bookstore-se app
- Renamed bookstore-nima to bookstore-se for consistency with other apps (e.g. quickstart)
- Updated `MainTest` to use blocking client
- Issue #6788


barchetta (approved):
LGTM

PR #6798, opened 2023-05-10
[3.x] Support for injection of ServerRequest and ServerResponse also via CDI

spericas:
Support for injection of ServerRequest and ServerResponse also via CDI. New functional test to verify both types of injection.

### DIFF
--- a/tests/functional/request-scope-injection/src/main/java/io/helidon/tests/functional/context/injection/CheckInjectionResource.java
+++ b/tests/functional/request-scope-injection/src/main/java/io/helidon/tests/functional/context/injection/CheckInjectionResource.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates.
+ *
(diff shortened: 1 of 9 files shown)


tests/functional/request-scope-injection/src/main/java/io/helidon/tests/functional/context/injection/CheckInjectionResource.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Inject
+    private ServerRequest serverRequestCdi;
+
+    @Inject
+    private ServerResponse serverResponseCdi;
+
+    @GET
+    public Response checkInjection() {
+        Objects.requireNonNull(serverRequest);
```
ljnelson, line 50: (You don't need any of these `Objects.requireNonNull()` statements since you would trigger "smart" NPEs at lines 54 and 55.)

ljnelson (approved):
One tiny comment but basically LGTM

Issue #6575, opened 2023-04-06
2.x Upgrade Weld


barchetta:
Upgrade Weld in 2.x to at least 3.1.9.Final

http://weld.cdi-spec.org/download/

PR #6805, opened 2023-05-11
2.x Upgrade Weld #6575

jbescos:
Relates to https://github.com/helidon-io/helidon/issues/6575


microprofile/weld/weld-core-impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -1074,4 +1068,24 @@ private static String getDefaultPackageReason(Class<?> clazz) {
         }
         return null;
     }
+
+    /**
+     * Delegates proxy creation via {@link ProxyServices} to the integrator or to our own implementation.
```
ljnelson, line 1073: Minor note/question: I see this delegating to `ProxyServices` in all cases and never to "our own implementation".

PR #6844, opened 2023-05-17
support lowercase for modifiers names

### LINKED ISSUE #6764: All modifier names should be kept in lower case
`BuilderTypeTools` and `TypeNames` and places within `Pico` treat them as Uppercase, but this is a mistake.  We should keep them all as they come from annotation processing, in lowercase form. 

Unfortunately, this will involve lots of places being changed.  It is therefore not as simple as it would appear.

arjav-desai:
Fixes #6764

### DIFF
--- a/builder/processor-tools/src/main/java/io/helidon/builder/processor/tools/BodyContext.java
+++ b/builder/processor-tools/src/main/java/io/helidon/builder/processor/tools/BodyContext.java
@@ -126,10 +126,29 @@ public class BodyContext {
                 searchForBuilderAnnotation("interceptorCreateMethod", builderTriggerAnnotation, typeInfo);
         this.interceptorCreateMethod = (interceptorCreateMethod == null || interceptorCreateMethod.isEmpty())
                 ? null : interceptorCreateMethod;
-        this.publicOrPackagePrivateDecl = (typeInfo.typeKind().equals(TypeInfo.KIND_INTERFACE)
-                                                   || typeInfo.modifierNames().isEmpty()
-                                                   || typeInfo.modifierNames().stream().anyMatch(TypeInfo.MODIFIER_PUBLIC::equalsIgnoreCase))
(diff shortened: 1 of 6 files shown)


builder/processor-tools/src/main/java/io/helidon/builder/processor/tools/BodyContext.java
```diff
@@ -126,10 +126,29 @@ public class BodyContext {
...
-                                                        ? "public " : "";
+        this.publicOrPackagePrivateDecl = (
+                typeInfo.typeKind().equals(TypeInfo.KIND_INTERFACE)
+                        || typeInfo.modifierNames().isEmpty()
+                        || typeInfo.modifierNames().stream().anyMatch(TypeInfo.MODIFIER_PUBLIC::equalsIgnoreCase))
+                ? "public " : "";
+    }
+
+    /**
+     * returns the bean attribute name of a particular method.
```
trentjeff: nit: Upper case first letter.

builder/processor-tools/src/main/java/io/helidon/builder/processor/tools/BodyContext.java
```diff
@@ -572,6 +574,69 @@ private static String searchForBuilderAnnotation(String key,
...
+    }
+
+    private static TypeName toCtorBuilderAcceptTypeName(TypeInfo typeInfo,
+                                                        boolean hasParent,
+                                                        TypeName parentAnnotationTypeName) {
+        if (hasParent) {
+            return typeInfo.typeName();
+        }
+
+        return (
```
trentjeff: very strange formatting.

arjav-desai: fixed with reformatting

trentjeff (approved):

Issue #6773, opened 2023-05-04
Refactor Custom Annotation Processor


trentjeff:
In `CustomAnnotationProcessor`:
```
                // TODO: the following are duplicates that should be removed - get them from the enclosingTypeInfo instead                
                .targetElementArgs(toArgs(typeToProcess))
                .targetElementAccess(toAccess(typeToProcess))
                .elementStatic(isStatic(typeToProcess));

```

PR #6883, opened 2023-05-22
Custom Annotation Processor Refinements

trentjeff:
Addresses issue #6773 . Removes a few duplicate APIs that are now part of `TypeInfo`.


dalexandrov (approved):
LGTM

PR #6907, opened 2023-05-25
6543 HTTP/2 header continuation

### LINKED ISSUE #6543: Nima - HTTP/2 Client - outbound continuation
Nima HTTP/2 client needs to be able to split too large header frame to 1 x frame followed by n x Continuation frame

[10.5.1.](https://httpwg.org/specs/rfc7540.html#rfc.section.10.5.1) [Limits on Header Block Size](https://httpwg.org/specs/rfc7540.html#MaxHeaderBlock)
A large header block ([Section 4.3](https://httpwg.org/specs/rfc7540.html#HeaderBlock)) can cause an implementation to commit a large amount of state. Header fields that are critical for routing can appear toward the end of a header block, which prevents streaming of header fields to their ultimate destination. This ordering and other reasons, such as ensuring cache correctness, mean that an endpoint might need to buffer the entire header block. Since there is no hard limit to the size of a header block, some endpoints could be forced to commit a large amount of available memory for header fields.

An endpoint can use the [SETTINGS_MAX_HEADER_LIST_SIZE](https://httpwg.org/specs/rfc7540.html#SETTINGS_MAX_HEADER_LIST_SIZE) to advise peers of limits that might apply on the size of header blocks. This setting is only advisory, so endpoints MAY choose to send header blocks that exceed this limit and risk having the request or response being treated as malformed. This setting is specific to a connection, so any request or response could encounter a hop with a lower, unknown limit. An intermediary can attempt to avoid this problem by passing on values presented by different peers, but they are not obligated to do so.

A server that receives a larger header block than it is willing to handle can send an HTTP 431 (Request Header Fields Too Large) status code [[RFC6585]](https://httpwg.org/specs/rfc7540.html#RFC6585). A client can discard responses that it cannot process. The header block MUST be processed to ensure a consistent connection state, unless the connection is closed.

danielkec:
Fixes #6543

### DIFF
--- a/nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2ConnectionWriter.java
+++ b/nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2ConnectionWriter.java
@@ -74,19 +74,62 @@ public class Http2ConnectionWriter implements Http2StreamWriter {
         // we must enforce parallelism of exactly 1, to make sure the dynamic table is updated
         // and then immediately written
 
+        int maxFrameSize = flowControl.maxFrameSize();
+
         return withStreamLock(() -> {
             int written = 0;
             headerBuffer.clear();
             headers.write(outboundDynamicTable, responseHuffman, headerBuffer);
-            Http2FrameHeader frameHeader = Http2FrameHeader.create(headerBuffer.available(),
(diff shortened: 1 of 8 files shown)


spericas (dismissed):
LGTM

spericas (approved):

PR #6915, opened 2023-05-25
2.x: Upgrade maven-dependency-plugin to 3.6.0

barchetta:
This upgrades the version of the maven-dependency-plugin used by the Helidon build and Helidon applications. At some point the behavior of `<excludeScope>test</excludeScope>` changed and is no longer what we want (see https://maven.apache.org/plugins/maven-dependency-plugin/copy-dependencies-mojo.html).


ljnelson (approved):
LGTM FWIW

PR #6916, opened 2023-05-26
Response should not be chunked if there is no entity

klustria:
NOTE: Difference with 3.x PR 6637 is that 2.x has WebSocket request and when it is used, writeStatusAndHeaders() needs to send the header right via initWriteResponse()


spericas (approved):

klustria:
This PR is for Backport issue https://github.com/helidon-io/helidon/issues/6901

Issue #6804, opened 2023-05-10
Nima WebClient Redirects support


romain-grecourt:

Verdent:
This issue needs more work, but for the most cases it will work just fine now. It should unblock others who were blocked by it. PR https://github.com/helidon-io/helidon/pull/6929 added this functionality.

barchetta:
Closing. Tracking further work with #6992

PR #6929, opened 2023-05-31
Níma WebClient redirect support

Verdent:
Níma WebClient redirect support - only HTTP 1 for now

See #6804

### DIFF
--- a/nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/ClientConfig.java
+++ b/nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/ClientConfig.java
@@ -66,4 +66,20 @@ public interface ClientConfig {
      */
     DnsAddressLookup dnsAddressLookup();
 
+    /**
+     * Whether to follow redirects.
+     *
+     * @return follow redirects
+     */
+    @ConfiguredOption("false")
+    boolean followRedirects();
+
+    /**
+     * Maximum number of redirects allowed.
+     *
+     * @return allowed number of redirects
+     */
+    @ConfiguredOption("5")
+    int maxRedirects();
+
 }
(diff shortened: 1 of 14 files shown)


romain-grecourt (dismissed):

romain-grecourt (approved):

nima/webclient/webclient/src/test/java/io/helidon/nima/webclient/HttpClientTest.java
```diff
@@ -113,6 +113,16 @@ public FakeHttpClientRequest fragment(String fragment) {
             return this;
         }
 
+        @Override
+        public FakeHttpClientRequest followRedirects(boolean followRedirects) {
+            return null;
```
tomas-langer, line 118: I think it should `return this` rather than `return null`

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/ClientConfig.java
```diff
@@ -66,4 +66,20 @@ public interface ClientConfig {
      */
     DnsAddressLookup dnsAddressLookup();
 
+    /**
+     * Whether to follow redirects.
+     *
+     * @return follow redirects
+     */
+    @ConfiguredOption("false")
```
tomas-langer, line 74: Is this the right default? I think it would be more convenient for users if redirects were followed by default.

tomas-langer (changes requested):

nima/webclient/webclient/src/test/java/io/helidon/nima/webclient/HttpClientTest.java
Verdent: ehm... yes... thank you.

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/ClientConfig.java
Verdent: I was using the same default as there is with reactive WebClient.

tomas-langer (approved):

Issue #6891, opened 2023-05-24
Prerequisite changes in support for Pico/OCI Extensibility


trentjeff:
In preparation for #6406 there needs to be a way to observe injection points so that the oci integration (i.e., generation of activators and the component module) can occur.

PR #6936, opened 2023-06-01
Preparation for Pico Extensibility Enhancements

trentjeff:
This resolves issue #6891

### DIFF
--- a/pico/api/src/main/java/io/helidon/pico/api/ServiceProviderBindable.java
+++ b/pico/api/src/main/java/io/helidon/pico/api/ServiceProviderBindable.java
@@ -111,12 +111,11 @@ public interface ServiceProviderBindable<T> extends ServiceProvider<T> {
     /**
      * Returns the previously assigned {@link PicoServices} instance.
      *
-     * @param expected true if the result is expected to be non-empty
      * @return the previously assigned pico services instance, or empty if never assigned
      *
      * @see #picoServices(Optional)
      */
-    Optional<PicoServices> picoServices(boolean expected);
+    Optional<PicoServices> picoServices();
 
     /**
      * Assigns the services instance this provider is bound to. A service provider can be associated with 0..1 services instance.
--- a/pico/api/src/main/java/io/helidon/pico/api/ServiceProviderBindable.java
+++ b/pico/api/src/main/java/io/helidon/pico/api/ServiceProviderBindable.java
@@ -109,10 +109,19 @@ public interface ServiceProviderBindable<T> extends ServiceProvider<T> {
     }
 
     /**
+     * Returns the previously assigned {@link PicoServices} instance.
+     *
+     * @return the previously assigned pico services instance, or empty if never assigned
+     * @see #picoServices(Optional)
+     */
+    Optional<PicoServices> safePicoServices();
+
+    /**
      * Assigns the services instance this provider is bound to. A service provider can be associated with 0..1 services instance.
      * If not set, the service provider should use {@link PicoServices#picoServices()} to ascertain the instance.
      *
      * @param picoServices the pico services instance, or empty to clear any active binding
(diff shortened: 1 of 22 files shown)


pico/api/src/main/java/io/helidon/pico/api/ServiceProviderBindable.java
```diff
@@ -108,11 +108,20 @@ default void rootProvider(ServiceProvider<T> rootProvider) {
         throw new UnsupportedOperationException();
     }
 
+    /**
+     * Returns the previously assigned {@link PicoServices} instance.
+     *
+     * @return the previously assigned pico services instance, or empty if never assigned
+     * @see #picoServices(Optional)
+     */
+    Optional<PicoServices> safePicoServices();
```
tomas-langer, line 117: Should be called `picoServices`, this name is strange.

tomas-langer: There is a conflict with `picoServices` on AbstractProvider - that method should be removed, and this one used instead (with `orElseThrow()` maybe)

tomas-langer (changes requested):

pico/api/src/main/java/io/helidon/pico/api/ServiceProviderBindable.java
trentjeff: Addressed and pushed.

pico/runtime/src/main/java/io/helidon/pico/runtime/AbstractServiceProvider.java
```diff
@@ -202,10 +202,16 @@ public Phase currentActivationPhase() {
...
      * Used to access the current pico services instance assigned to this service provider.
      *
+     * @param expected true to indicate that the return values is expected to be non-empty
      * @return the pico services assigned to this service provider
      */
-    public PicoServices picoServices() {
-        return Objects.requireNonNull(picoServices, description() + ": picoServices should have been previously set");
+    @Override
+    public Optional<PicoServices> picoServices(boolean expected) {
+        if (expected) {
```
tomas-langer, line 210: The whole idea of return an optional is to leave the decision whether it is expected or not on the caller of the method.
This method should just `return Optional.ofNullable(picoServices)`
The caller should throw an exception if they should be available

tomas-langer (changes requested):

pico/runtime/src/main/java/io/helidon/pico/runtime/AbstractServiceProvider.java
trentjeff: I wanted the provider to control the exception and not the caller. I'll address it in a slightly modified way.

trentjeff: pushed

pico/runtime/src/main/java/io/helidon/pico/runtime/AbstractServiceProvider.java
```diff
@@ -199,13 +199,17 @@ public Phase currentActivationPhase() {
...
-     */
-    public PicoServices picoServices() {
-        return Objects.requireNonNull(picoServices, description() + ": picoServices should have been previously set");
+    @Override
+    public Optional<PicoServices> picoServices() {
+        return Optional.ofNullable(picoServices);
+    }
+
+    PicoServices requiredPicoServices() {
+        PicoServices picoServices = picoServices().orElse(null);
```
tomas-langer, line 208: you can do
```
PicoServices picoServices = picoServices()
     .orElseThrow(() -> new PicoServiceProviderException(description() + ": picoServices should have been previously set", this));
```

tomas-langer (approved):

PR #6951, opened 2023-06-05
Adds Nima WebClient Shortcuts for media support

ljnelson:
This PR tries to do the equivalent of what #6564 does on the server side with as few changes as possible. This is not my area at all so please do let me know what I've done wrong.


klustria (commented):
Not really a MediaContext expert, but in comparison to the Server implementation, the changes seem to look fine to me. The only comment I may have, and maybe this is more of a  question... is it possible to add a test on this in nima/tests/integration/webclient to test out new shortcut like addMediaSupport like below:

            WebClient webClient = WebClient.builder()
                    ...
                    .addMediaSupport(JsonpSupport.create())
                    ...
                    .build();

klustria (dismissed):
LGTM, hence approving. Would be good to create a tracking issue to see if we can add a unit or an integration test for this.

romain-grecourt (dismissed):

klustria (approved):
LGTM....re-approving

PR #6959, opened 2023-06-07
4.x: Make check for audience claim in access token optional in OIDC provider

### LINKED ISSUE #5860: Make check for audience claim in access token optional in OIDC provider
It is not always possible to configure the Identity Provider to have compatible "aud" claim. It would be very nice to have the check for the "aud" claim optional (with default = true (do check claim) ).

## Environment Details
This is not version or SE/MP specific or platform specific.

Tomas-Kraus:
Added `optional-audience` config option to make audience claim optional.
Automatic audience claim generation is disabled when this option is set to true.

Resolves #5860

### DIFF
--- a/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
+++ b/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
@@ -959,8 +959,7 @@ public class Jwt {
      *                 issuer claim mandatory
      * @param audience validates that this JWT was issued for this audience. Setting this to non-null value will make
      *                 audience claim mandatory
-     * @param checkAudience whether audience claim check is turned on. Validation will fail when {@code true}
-     *                      and audience claim is null
+     * @param checkAudience whether audience claim validation should be executed
      * @return errors instance to check for validation result
      */
     public Errors validate(String issuer, String audience, boolean checkAudience) {
@@ -982,8 +981,7 @@ public class Jwt {
      *                 issuer claim mandatory
      * @param audience validates that this JWT was issued for this audience. Setting this to non-null value and with
      *                 any non-null value in the Set will make audience claim mandatory
-     * @param checkAudience whether audience claim check is configured as mandatory. Validation will fail when {@code true}
-     *                      and audience claim is null
+     * @param checkAudience whether audience claim validation should be executed
      * @return errors instance to check for validation result
      */
     public Errors validate(String issuer, Set<String> audience, boolean checkAudience) {
--- a/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
+++ b/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
@@ -953,6 +953,22 @@ public class Jwt {
 
     /**
      * Validates all default values.
+     * Values validated: {@link #validate(String, Set, boolean)}
+     *
+     * @param issuer   validates that this JWT was issued by this issuer. Setting this to non-null value will make
+     *                 issuer claim mandatory
+     * @param audience validates that this JWT was issued for this audience. Setting this to non-null value will make
+     *                 audience claim mandatory
+     * @param checkAudience whether audience claim check is turned on. Validation will fail when {@code true}
+     *                      and audience claim is null
+     * @return errors instance to check for validation result
+     */
+    public Errors validate(String issuer, String audience, boolean checkAudience) {
+        return validate(issuer, audience == null ? Set.of() : Set.of(audience), checkAudience);
+    }
+
+    /**
+     * Validates all default values.
      * Values validated:
      * <ul>
      * <li>{@link #expirationTime() Expiration time} if defined</li>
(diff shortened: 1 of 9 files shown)


Tomas-Kraus:
Fixed conflict after rebase.
@Verdent This PR is here more than month, please review it.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -501,4 +503,13 @@ String scopeAudience() {
     String name() {
         return TenantConfigFinder.DEFAULT_TENANT_ID;
     }
+
+    boolean optionalAudience() {
+        return optionalAudience;
+    }
+
+    void setOptionalAudience(boolean optional) {
```
tomas-langer, line 511: we do not use verbs in setters, it should be just `optionalAudience(boolean)`, unless there is a conflict (which should not be)

tomas-langer (changes requested):

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
Tomas-Kraus: optionalAudience(boolean) is config blueprint method so I could not use the same name as setter. The only way how to fix this is to define blueprint method in the BaseBuilder.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -362,6 +367,7 @@ public final class OidcConfig extends TenantConfigImpl {
     private final OidcCookieHandler tokenCookieHandler;
     private final OidcCookieHandler idTokenCookieHandler;
     private final OidcCookieHandler tenantCookieHandler;
+    private final boolean optionalAudience;
```
Verdent, line 370: never used

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -501,4 +515,9 @@ String scopeAudience() {
     String name() {
         return TenantConfigFinder.DEFAULT_TENANT_ID;
     }
+
+    boolean optionalAudience() {
+        return optionalAudience;
+    }
```
Verdent, line 521: Not needed

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -997,6 +1004,8 @@ public Builder config(Config config) {
             config.get("tenants").asList(Config.class)
                     .ifPresent(confList -> confList.forEach(tenantConfig -> tenantFromConfig(config, tenantConfig)));
 
+            config.get("optional-audience").asBoolean().ifPresent(this::optionalAudience);
```
Verdent, line 1007: This should be part of the BaseBuilder config method.

Verdent (changes requested):
Right now you are mixing two approaches here. We have tenant and overall OIDC config. I think placing this switch to the BaseBuilder is correct (BaseBuilder is used for both), since audience can be set different for each tenant. But in that case we need to be able also to get it from config in BaseBuilder and not in overall OIDC config only.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
Tomas-Kraus: removed

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
Tomas-Kraus: Moved all audience related things to BaseBuilder

security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
```diff
@@ -966,23 +982,43 @@ public Errors validate(String issuer, String audience) {
      *                 issuer claim mandatory
      * @param audience validates that this JWT was issued for this audience. Setting this to non-null value and with
      *                 any non-null value in the Set will make audience claim mandatory
+     * @param checkAudience whether audience claim check is configured as mandatory. Validation will fail when {@code true}
+     *                      and audience claim is null
```
Verdent, line 986: Please correct me if I am wrong, but this is not what this code does. If I am not mistaken, if set to `false`, it does not add the validation at all (which is fine by this statement), but how about the statement about failing validation in case of null audience? Isn't it actually null if set to optional? Also there is this null check for audience which will not add this check at all :-)

Tomas-Kraus: Forgot to fix this javadoc after I changed logic of the fix. Should be OK now.

Verdent (approved):

tomas-langer (approved):

Issue #6774, opened 2023-05-04
[4.x] WebClient should have a mode that is resilient to bad media/content types


trentjeff:
Here is a snippet of an HTTP response from a server:

```
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: text
< Content-Length: 280
< 
```
Some parts were removed for privacy issues

As you will notice the server is returning an invalid content type of `text` - see https://www.iana.org/assignments/media-types/media-types.xhtml#text for a list of all valid media/content types.

Our WebClient should support a mode (e.g., `relaxed` mode vs the current mode which I'll call `strict`) that will log a warning about these invalid usages, but should not throw an exception.  Even nicer would be to translate `text` -> `text/plain` for this specific case in `relaxed` mode.

I was able to reproduce this exception behavior in the following unit test of `MediaTypeTest` on main:

```
    @Test
    void parseEmptySubtypeParameterValue() {
        assertThat(HttpMediaType.create("text/html").parameters(), is(Map.of()));
        // vvv this one will fail
        assertThat(HttpMediaType.create("text").parameters(), is(Map.of()));
        assertThat(HttpMediaType.create("text; o1=; o2=v2").parameters(), is(Map.of("o2", "v2")));
    }
```

If we are in a relaxed mode, the WebClient should not throw an exception - we should just log a warning.  Doing so will make Helidon more resilient for users to consume since they currently have no other workaround other than to switch to HttpClient, etc.

Here is the full exception reported by the user from 3.1.0:
```
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Could not parse 'text'
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
        at io.helidon.common.reactive.Single.get(Single.java:730)
        ... a bunch more
Caused by: java.lang.IllegalArgumentException: Could not parse 'text'
        at io.helidon.common.http.MediaType.parse(MediaType.java:360)
        at java.base/java.util.Optional.map(Optional.java:260)
        at io.helidon.webclient.WebClientResponseHeadersImpl.contentType(WebClientResponseHeadersImpl.java:79)
        ... a bunch more
Caused by: java.lang.IllegalStateException: No more elements!
[...]

PR #6999, opened 2023-06-14
Issue #6774 - WebClient should have a mode that is resilient to bad media/content types

Tomas-Kraus:
Added relaxed mode to media type parser in media type and http modules of common.
This affects how HTTP1 headers are being parsed:

- default mode remains unchanged
- HTTP client may be configured to use relaxed mode which will cause relaxed rules to be applied on HTTP1 headers
- HTTP2 headers parsing remains unchanged
- currently there exist just a single rule: `text` -> `text/plain` to resolve issue #6774.

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
+++ b/common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
+++ b/common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,20 +17,15 @@
 package io.helidon.common.http;
 
 import java.nio.charset.StandardCharsets;
-import java.util.Optional;
 
 import io.helidon.common.buffers.Bytes;
 import io.helidon.common.buffers.DataReader;
 import io.helidon.common.buffers.LazyString;
-import io.helidon.common.media.type.ParserMode;
 
 /**
  * Used by both HTTP server and client to parse headers from {@link io.helidon.common.buffers.DataReader}.
  */
 public final class Http1HeadersParser {
-
-    private static final System.Logger LOGGER = System.getLogger(Http1HeadersParser.class.getName());
-
     // TODO expand set of fastpath headers
     private static final byte[] HD_HOST = (HeaderEnum.HOST.defaultCase() + ":").getBytes(StandardCharsets.UTF_8);
     private static final byte[] HD_ACCEPT = (HeaderEnum.ACCEPT.defaultCase() + ":").getBytes(StandardCharsets.UTF_8);
@@ -48,14 +43,10 @@ public final class Http1HeadersParser {
      *
      * @param reader         reader to pull data from
      * @param maxHeadersSize maximal size of all headers, in bytes
-     * @param parserMode     media type parsing mode
      * @param validate       whether to validate headers
      * @return a new mutable headers instance containing all headers parsed from reader
      */
-    public static WritableHeaders<?> readHeaders(DataReader reader,
-                                                 int maxHeadersSize,
-                                                 ParserMode parserMode,
-                                                 boolean validate) {
+    public static WritableHeaders<?> readHeaders(DataReader reader, int maxHeadersSize, boolean validate) {
         WritableHeaders<?> headers = WritableHeaders.create();
         int maxLength = maxHeadersSize;
 
@@ -67,10 +58,6 @@ public final class Http1HeadersParser {
 
             Http.HeaderName header = readHeaderName(reader, maxLength, validate);
             maxLength -= header.defaultCase().length() + 2;
-            // Skip spaces after header name
-            while (' ' == reader.lookup()) {
-                reader.skip(1);
-            }
             int eol = reader.findNewLine(maxLength);
             if (eol == maxLength) {
                 throw new IllegalStateException("Header size exceeded");
@@ -80,22 +67,7 @@ public final class Http1HeadersParser {
             reader.skip(2);
             maxLength -= eol + 1;
 
-            if (parserMode == ParserMode.RELAXED && header == HeaderEnum.CONTENT_TYPE) {
-                String valueString = value.toString();
-                Optional<String> maybeRelaxedMediaType = ParserMode.findRelaxedMediaType(valueString);
-                if (maybeRelaxedMediaType.isPresent()) {
-                    headers.add(Http.Header.create(header, maybeRelaxedMediaType.get()));
-                    LOGGER.log(System.Logger.Level.WARNING,
-                               () -> String.format("Invalid %s header value \"%s\" replaced with \"%s\"",
-                                                   HeaderEnum.CONTENT_TYPE.defaultCase(),
-                                                   valueString,
-                                                   maybeRelaxedMediaType.get()));
-                } else {
-                    headers.add(Http.Header.create(header, valueString));
-                }
-            } else {
-                headers.add(Http.Header.create(header, value));
-            }
+            headers.add(Http.Header.create(header, value));
             if (maxLength < 0) {
                 throw new IllegalStateException("Header size exceeded");
             }
--- a/common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
+++ b/common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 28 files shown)


common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
```diff
@@ -58,6 +67,10 @@ public static WritableHeaders<?> readHeaders(DataReader reader, int maxHeadersSi
 
             Http.HeaderName header = readHeaderName(reader, maxLength, validate);
             maxLength -= header.defaultCase().length() + 2;
+            // Skip spaces after header name
```
tomas-langer, line 70: This is not needed - it is already part of `LazyString` - there is a method `stripOws` that handles this problem.
This would only introduce a performance overhead

common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
```diff
@@ -67,7 +80,22 @@ public static WritableHeaders<?> readHeaders(DataReader reader, int maxHeadersSi
             reader.skip(2);
             maxLength -= eol + 1;
 
-            headers.add(Http.Header.create(header, value));
+            if (parserMode == ParserMode.RELAXED && header == HeaderEnum.CONTENT_TYPE) {
```
tomas-langer, line 83: This is not the right place to resolve this problem. The parsing should be as fast as possible, even if the value is wrong.
The value should be resolved and validated when needed - e.g. in the case of `ContentType`, when somebody calls the content type method on server request, or client response, where we map the value to actual `MediaType` (or `HttpMediaType`)

common/media-type/src/main/java/io/helidon/common/media/type/MediaTypeImpl.java
```diff
@@ -16,10 +16,30 @@
...
+
+    static MediaType parse(String fullType, ParserMode parserMode) {
         int slashIndex = fullType.indexOf('/');
         if (slashIndex < 1) {
+            if (parserMode == ParserMode.RELAXED) {
+                Optional<String> maybeRelaxedType = ParserMode.findRelaxedMediaType(fullType);
+                if (maybeRelaxedType.isPresent()) {
+                    String relaxedType = maybeRelaxedType.get();
+                    slashIndex = relaxedType.indexOf('/');
+                    LOGGER.log(System.Logger.Level.WARNING,
```
tomas-langer, line 33: Not a warning material - esp. if we explicitly set  `RELAXED` mode and provide a replacement, it may be a trace or debug statement.

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    STRICT,
+    /**
+     * Relaxed mode.
+     * Apply additional rules to identify unknown media types.
+     */
+    RELAXED;
+
+    // Relaxed media types mapping
+    private static final Map<String, String> RELAXED_TYPES = Map.of(
```
tomas-langer, line 40: This should be a map of `<String, MediaType>` if possible, so we do not need to parse it if replaced

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
```diff
@@ -0,0 +1,79 @@
...
+    );
+
+    /**
+     * Find relaxed media type mapping for provided value.
+     *
+     * @param value source media type value
+     * @return mapped media type value or {@code Optional.empty()}
+     *         when no mapping for given value exists
+     */
+    public static Optional<String> findRelaxedMediaType(String value) {
```
tomas-langer, line 57: It should be clear this method is only called when the value is not according to spec.
This actually feels like something we may want to provide full customization for using something like

```
interface MediaTypeParser {
  Optional<MediaType> map(String value);
}
```
That we would implement ourself using what we have in `MediaTypeImpl.parse()`,
and allow customizing it when setting up server or client (similar to the parser mode you have now). This would allow us to create our own parser with the mode (such as `HelidonMediaTypeParser.create()` and `HelidonMediaTypeParser.create(ParserMode)`, or even have it customizable using a builder to provide additional mappings, or to have a falblack type to use when parsing fails.
This seems to be a bit too hard-coded

tomas-langer (changes requested):

common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
Tomas-Kraus: Removed

common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
Tomas-Kraus: I'm reverting whole class (+ all required changes) to main version.

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
Tomas-Kraus: Changed to pkg visibility to remove it from API. Will keep hard-coded version now, but it may be rewritten in a better way.

common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
```diff
@@ -58,6 +67,10 @@ public static WritableHeaders<?> readHeaders(DataReader reader, int maxHeadersSi
 
             Http.HeaderName header = readHeaderName(reader, maxLength, validate);
             maxLength -= header.defaultCase().length() + 2;
+            // Skip spaces after header name
+            while (' ' == reader.lookup()) {
```
trentjeff, line 71: `java.lang.Character.isWhitespace()` would be a better choice here

common/media-type/src/main/java/io/helidon/common/media/type/MediaTypeImpl.java
trentjeff: +1; TRACE level maybe

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
trentjeff: Also, why is "text/plain" considered to be relaxed?  I would expect the map to only contain "text"

nima/http/media/multipart/src/main/java/io/helidon/nima/http/media/multipart/MultiPartSupport.java
```diff
@@ -45,8 +47,13 @@ public class MultiPartSupport implements MediaSupport {
             .withParameter("boundary", DEFAULT_BOUNDARY);
 
     private MediaContext context;
+    private final ParserMode parserMode;
 
-    private MultiPartSupport() {
+    private MultiPartSupport(Config config) {
+        ConfigValue<String> parserModeConfig = config.get("parser-mode").as(String.class);
```
trentjeff, line 53: where was the test for using config?

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
Tomas-Kraus: It contains only text. `text/plain` is target value.

common/http/src/main/java/io/helidon/common/http/Http1HeadersParser.java
Tomas-Kraus: This was reverted to main branch code.

common/http/src/test/java/io/helidon/common/http/MediaTypeTest.java
```diff
@@ -134,4 +136,20 @@ void testBuilt() {
         assertThat(mediaType.parameters(), is(Map.of("q", "0.1", "charset", "ISO-8859-2")));
         assertThat(mediaType.qualityFactor(), closeTo(0.1, 0.000001));
     }
+
+    // Calling create method with "text" argument shall throw IllegalArgumentException in strict mode.
+    @Test
+    void parseInvalidTextInStrictMode() {
+        assertThrows(IllegalArgumentException.class, () -> {
```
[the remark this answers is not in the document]
Tomas-Kraus: Message is being checked in io.helidon.nima.webclient.http1.HeadersTest#testInvalidTextContentTypeStrict(), but ok, let's have chech here too.

nima/http/media/multipart/src/main/java/io/helidon/nima/http/media/multipart/MultiPartSupport.java
Tomas-Kraus: This was reverted. No more tests are needed. :)

nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/http1/ClientResponseImpl.java
```diff
@@ -72,13 +73,16 @@ class ClientResponseImpl implements Http1ClientResponse {
     private long entityLength;
     private boolean entityFullyRead;
     private WritableHeaders<?> trailers;
+    // Media type parsing mode configured on client.
```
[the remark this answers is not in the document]
Tomas-Kraus: I don't see any Config in Http1ClientBuilder.
This value is coming from Http1ClientBuilder API and this is being tested in io.helidon.nima.webclient.http1.HeadersTest#testInvalidTextContentTypeStrict()

nima/http/media/multipart/src/main/java/io/helidon/nima/http/media/multipart/MultiPartImpl.java
```diff
@@ -38,8 +39,9 @@ class MultiPartImpl extends MultiPart {
     private ReadablePartAbstract inProgress;
     private boolean finished;
     private int index;
+    private final ParserMode parserMode;
```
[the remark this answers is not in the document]
Tomas-Kraus: Never heard about such a rule, but OK.

trentjeff (dismissed):

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
```diff
@@ -0,0 +1,57 @@
...
+    STRICT,
+    /**
+     * Relaxed mode.
+     * Apply additional rules to identify unknown media types.
+     */
+    RELAXED;
+
+    // Relaxed media types mapping
+    private static final Map<String, MediaType> RELAXED_TYPES = Map.of(
+            "text",  new MediaTypeImpl("text", "plain", "text/plain") // text -> text/plain
```
tomas-langer, line 41: Use `MediaTypes.TEXT_PLAIN` instead

tomas-langer (changes requested):
One small change please

common/media-type/src/main/java/io/helidon/common/media/type/ParserMode.java
Tomas-Kraus: Done.

tomas-langer (approved):

Issue #7016, opened 2023-06-16
[4.x] gRPC broken class loading when specifying protobuf package name


edvardfagerholmsc:
The following test program:

```java
package io.github.efagerho.loom;

import static io.helidon.nima.grpc.webserver.ResponseHelper.complete;

import io.github.efagerho.loom.grpc.Echo;
import io.github.efagerho.loom.grpc.HelloRequest;
import io.github.efagerho.loom.grpc.HelloResponse;
import io.grpc.stub.StreamObserver;
import io.helidon.logging.common.LogConfig;
import io.helidon.nima.grpc.webserver.GrpcRouting;
import io.helidon.nima.webserver.WebServer;

public class ServerMain {

    private static void hello(HelloRequest request, StreamObserver<HelloResponse> observer) {
        System.out.println("Handler called");
        String name = request.getName();
        complete(observer, HelloResponse.newBuilder().setResponse(String.format("Hello %s", name)).build());
    }

    public static void main(String[] args) {
        LogConfig.configureRuntime();

        WebServer ws = WebServer
            .builder()
            .port(8080)
            .addRouting(GrpcRouting.builder().unary(Echo.getDescriptor(), "EchoService", "Hello", ServerMain::hello)).build();

        System.out.format("gRPC endpoint configured with tls = %b\n", ws.hasTls());
        ws.start();
    }
}
```

Works with the .proto file below:

```
syntax = "proto3";

option java_multiple_files = true;
option java_package = "io.github.efagerho.loom.grpc";

service EchoService {
  rpc Hello (HelloRequest) returns (HelloResponse) {}
}

message HelloRequest {
  string name = 1;
}

message HelloResponse {
  string response = 1;
}
```

Some gRPC libraries like Rust's Tonic require a package name for the proto. However, if I modify the proto file by adding a `package echo;` declaration, then the Helidon Java server will fail to start with the following error:

```
Exception in thread "main" java.lang.IllegalArgumentException: Failed to load class "io.github.efagerho.loom.grpc.echo$HelloRequest" for grpc
        at io.helidon.nima.grpc.webserver.Grpc.load(Grpc.java:141)
        at io.helidon.nima.grpc.webserver.Grpc.grpc(Grpc.java:122)
        at io.helidon.nima.grpc.webserver.Grpc.unary(Grpc.java:53)
        at io.helidon.nima.grpc.webserver.GrpcRouting$Builder.unary(GrpcRouting.java:125)
[...]

PR #7019, opened 2023-06-16
gRPC invalid class loading with reflection

efagerho:
gRPC classes are loaded dynamically using protobuf reflection. getFullName() is used, which prepends the protobuf package name, which is not used by the Java package.

Fixes: https://github.com/helidon-io/helidon/issues/7016


tomas-langer:
Replaced with #7158

PR #7031, opened 2023-06-20
Removed calls to allowThreadLocals(true) to get ready for Java 21

kabutz:
The method allowThreadLocals() was removed in Java 21, so should be removed, especially since it's always set to "true" anyway.


tomas-langer:
Thanks for this PR!
 #7008 will conflict with this, due to changes to configuraiton objects (which removes part of the usage from configuration and server). We will need to revisit the code after we merge it and get rid of the rest.

romain-grecourt:
All the changes in this PR were implemented in https://github.com/helidon-io/helidon/pull/7222.

PR #7072, opened 2023-06-23
Document assumptions about types of current processing round

### LINKED ISSUE #7071: 4.x: Builders - document some assumptions better
CustomMethods l. 164 - <any> used, maybe switch to checking for error type?
FactoryMethods l. 100 - similar, explicitly check for unknown type, maybe add validation to last round?

tomas-langer:
Resolves #7071

### DIFF
--- a/common/processor/processor/src/main/java/io/helidon/common/processor/TypeFactory.java
+++ b/common/processor/processor/src/main/java/io/helidon/common/processor/TypeFactory.java
@@ -87,9 +87,10 @@ public final class TypeFactory {
         case TYPEVAR -> {
             return Optional.of(createFromGenericDeclaration(typeMirror.toString()));
         }
-        case WILDCARD -> {
+        case WILDCARD, ERROR -> {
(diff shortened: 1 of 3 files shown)


common/processor/processor/src/main/java/io/helidon/common/processor/TypeFactory.java
```diff
@@ -87,9 +87,10 @@ public static Optional<TypeName> createTypeName(TypeMirror typeMirror) {
         case TYPEVAR -> {
             return Optional.of(createFromGenericDeclaration(typeMirror.toString()));
         }
-        case WILDCARD -> {
+        case WILDCARD, ERROR -> {
```
trentjeff, line 90: shouldn't we have a unit test for these returns?

tomas-langer: Can we get a `TypeMirror` instance in a unit test? These methods did not have unit tests before, so I guess you investigated this?

trentjeff (approved):

PR #7080, opened 2023-06-23
Config with injection

### LINKED ISSUE #7079: Support Config in injection
Helidon Common Config should be injectable into services.
It would be also great to have a way to add custom config sources that are declared as services to the runtime config.

This will create a bigger distinction between the "bootstrap" config (by default uses `Config.create()`), and the "runtime" config, that is agumentted with discovered config sources.

There should still be a way to override configuration by a user, by explicitly setting a config instance (such as by introducing `GlobalConfig`.

tomas-langer:
Resolves #7079 

### Config documentation notes

Helidon config API (helidon-common-config)

There is a new class:
`io.helidon.common.config.GlobalConfig`

This class can be used to set global configuration to be used by Helidon components. The configuration must be set as the first thing, before Helidon components 
are invoked, or Helidon main class is called.
If no configuration is set, the usual `Config.create()` is used to obtain configuration either from default sources, or using meta configuration.

Helidon config
ConfigSource interface is now marked as a `@Contract` for Helidon Injection
ConfigProducer is a new `@Singleton` service that combines the discovered `ConfigSource` from Helidon Injection, with the discovered configuration (from GlobalConfig) and creates a new config (unless an explicit config instance is set using `GlobalConfig` - in such a case it is used AS IS). 
This means that when `io.helidon.common.Config` is injected into any service using Helidon Injection, this combined config instance will be provided.

### PR notes
helidon-config dependency had to be removed from all pico modules, so we can generate service code in config module. The tests were moved to appropriate location in `pico/tests`, currently using the same packages, to allow calling package local methods. 

Wow this caused quite a mayhem in our dependency management for `jakarta.inject-api`. Investigating

Follow up issue: https://github.com/helidon-io/helidon/issues/7081


pico/tests/resources-pico/src/test/resources/expected/module-info.java._pico_
```diff
@@ -43,4 +44,4 @@ module io.helidon.pico.tests.pico {
     exports io.helidon.pico.tests.pico.provider;
     // Pico application - @io.helidon.common.Generated(value = "io.helidon.pico.tools.ApplicationCreatorDefault", trigger = "io.helidon.pico.tools.ApplicationCreatorDefault")
```
ljamen, line 45: Are we still using Pico? I thought we were supposed to rename this?

ljamen (commented):
Reviewed java files. No docs currently associated with this draft PR. Documentation will be captured in a separate PR.

pico/tests/resources-pico/src/test/resources/expected/module-info.java._pico_
tomas-langer: It will be renamed (eventually before 4.x is released), not as part of this PR

tomas-langer:
Rebased on main and fixed APT dependencies.

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptorSupport.java
```diff
@@ -157,6 +157,8 @@ static ModuleInfoDescriptor create(String moduleInfo,
                         }
                     }
                     descriptor.addItem(provides.build());
+                } else if (line.startsWith("opens ")) {
+                    // follow up issue
```
danielkec, line 161: Can we leave the issue number in comment for later clarity?

danielkec (approved):

pico/tools/src/main/java/io/helidon/pico/tools/ModuleInfoDescriptorSupport.java
tomas-langer: Added to PR description

PR #7096, opened 2023-06-26
Blocking DB Client

### LINKED ISSUE #6991: Blocking DBClient Part 1
Here is the initial proposal for the update of DBClient to a blocking style API.

---

### DbExecute.

```java
public interface DbExecute extends AutoCloseable {
    Stream<DbRow> query(String statement, Object... parameters);
    Optional<DbRow> get(String statement, Object... parameters);
    long insert(String statement, Object... parameters);
    long update(String statement, Object... parameters);
    long delete(String statementName, Object... parameters);
    long dml(String statement, Object... parameters);
}
```

---

### AutoClosable

Note the use of `AutoClosable`.

In the blocking style, `onComplete` is achieved by the next statement.

```java
List<String> names = exec.query("SELECT * FROM names").toList();
System.out.println("done");
```

`onError` is achieved by a try-catch block:

```java
try {
    List<String> names = exec.query("SELECT * FROM names").toList();
    System.out.println("done");
} catch (DbClientException dce) {
    // ugh-oh
}
```

`DbExecute` should work like a session, where the object can be used multiple times to perform database operations.
If `DbExecute` is designed as single-use, we should introduce a `DbSession` class that implements `AutoClosable`.

The try-with-resources block represents code that uses a session:

```java
try (DbExecute exec = dbClient.execute()) {
  // use exec
}
```

---

### Transaction

`DbTransaction` extends `DbExecute` and defines a contract where `commit()` MUST be called.
When the auto `close()` is called, all non-committed and non rollbacked transactions are rollbacked.

```java
public interface DbTransaction extends DbExecute {
    void commit();
    void rollback();
}
```

```java
try (DbTransaction tx = dbClient.transaction()) {
    long count = tx.delete("DELETE FROM pokemons WHERE name = 'foo'");
    tx.commit();
    return count;
}
```

---

### Stream.close

Use `Stream` close handler as an API to close the related `ResultSet` (for JDBC). However since `Stream`  does not require try-with-resources, hook the closing of the stream as part of the enclosing `AutoClosable` (I.e. `DbExecute`).

This means, if the user uses a `Stream` produced by `DbStatementQuery` outside of a try-with-resources, then the stream is unusable.

The creation of the `Stream<DbRow>` can be done like this:

```java
Stream<DbRow> stream = StreamSupport.stream(new AbstractSpliterator<DbRow>(Long.MAX_VALUE, Spliterator.ORDERED) {
    @Override
    public boolean tryAdvance(Consumer<? super DbRow> action) {
        try {
            if (!rs.next()) {
                return false;
            }
            DbRow dbRow = createDbRow(rs, metadata, dbMapperManager, mapperManager);
            action.accept(dbRow);
            return true;
        } catch (SQLException ex) {
            throw new DbClientException(ex.getMessage(), ex);
        }
    }
}, false).onClose(() -> {
    try {
        rs.close();
    } catch (SQLException e) {
        throw new DbClientException("Failed to close result-set", e);
    }
});
```

---

### Boxed Long

Remove the result type parameter from `DbStatement` in order to avoid forcing a boxed `Long` as a return type.
This means that the `execute` method has to be defined in all sub interfaces of `DbStatement`:

```java
public interface DbStatementQuery extends DbStatement<DbStatementQuery> {
    Stream<DbRow> execute();
}
```

```java
public interface DbStatementGet extends DbStatement<DbStatementGet> {
    Optional<DbRow> execute();
}
```

```java
public interface DbStatementDml extends DbStatement<DbStatementDml> {
    long execute();
}
```

---

### Interceptor

Update the interceptor mechanism to be blocking friendly similar to `WebClientService`:

```java
public interface DbClientService {

    Object handle(Chain chain, DbClientServiceContext context);

    interface Chain {
        Object proceed(DbClientServiceContext context);
    }
}
```

---

### MongoDB

- Drop transaction support for `MongoDB`, throw an `UnsupportException` in `transaction()`
- Use the blocking API of Mongo directly 

---

### Misc

- Inline `dbclient/common` in the main module
- Remove all `ExecutorService` related code
- Remove all mentions of reactive, non-blocking

Tomas-Kraus:
This is still in progress. Currently working on common and JDBC basic functionality
Fixes #6991


tomas-langer (dismissed):
OK to merge this, as we need to speed up removal of reactive. Follow up issue exists...

tomas-langer (approved):

PR #7103, opened 2023-06-26
[4.x] Allow SE users work with OpenApiFeature instead of SeOpenApiFeature

### LINKED ISSUE #7041: SeOpenApiFeature should be renamed
The feature name should be `OpenApiFeature`

nima/openapi/src/main/java/io/helidon/nima/openapi/SeOpenApiFeature.java

tjquinno:
Resolves #7041 

Previously, SE developers would directly use `SeOpenApiFeature` in `nima/openapi`. We want them to _not_ use "SE" anything.

That class was the only one in `nima/openapi`. This PR:

1. Moves `SeOpenApiFeature` to `helidon/openapi`. 
2. Adds `OpenApiFeature#builder()` which returns an `OpenApiFeature.Builder` for a `SeOpenApiFeature` for SE developers to use.
3. Moves tests and supporting resource files from `nima/openapi` to `microprofile/openapi`. Those tests and resources refer to SmallRye OpenAPI elements. We continue to rely on SmallRye for our MP OpenAPI implementation but we will _not_ for SE. So those tests could not move to `helidon/openapi` or we'd have to add the SmallRye dependency to that component.
4. Updated the SE quickstart app to use the new `OpenApiFeature#builder` method.
5. Unplugs `nima/openapi` from the `nima` module.

This PR does not address doc changes; we have a separate issue to address the OpenAPI doc.


microprofile/openapi/src/test/java/io/helidon/microprofile/openapi/ServerModelReaderTest.java
```diff
@@ -0,0 +1,105 @@
...
+    public static void startup() {
+//        webServer = TestUtil.startServer(OPENAPI_SUPPORT_BUILDER);
+    }
+
+    @AfterAll
+    public static void shutdown() {
+//        TestUtil.shutdownServer(webServer);
+    }
+
+//    @Test
```
romain-grecourt, line 64: Why is the code below commented out ? Is it existing code that is moved ?

tjquinno: This test used to be in `helidon/openapi` in Helidon 3.x. I'm not sure where it went in the initial move to 4.x but we should have it. So I copied it from 3.x.

It needs to be in an MP-aware module, and `helidon/openapi` was MP-aware in 3.x but not in 4.x The APIs the test uses are different now and I did not want to take the time now to rewrite the test to use the MP-friendly test infrastructure.

I can remove this file from the PR and add it back in rewritten form later in a separate PR or leave it here and come back and edit it later.

tjquinno: I have removed the heavily-commented test and added #7106 to track the need to add it back.

openapi/src/main/java/io/helidon/openapi/OpenApiFeature.java
```diff
@@ -68,6 +68,15 @@ public abstract class OpenApiFeature extends HelidonFeatureSupport {
      */
     public static final String DEFAULT_CONTEXT = "/openapi";
 
+    /**
+     * Returns a new builder for preparing an SE variant of {@code OpenApiFeature}.
+     *
+     * @return new builder
+     */
+    public static Builder<?, ?> builder() {
```
tomas-langer, line 76: Returning `Builder<?, ?>` does not seem right. This is awkward when assigned to a variable. Is the abstraction really needed here? 
If we switch to new builder API, this should return the OpenApiConfig.Builder (and the MpOpenApiConfig would extend `OpenApiConfig.BuilderBase<?, ?>`).

tomas-langer: We can keep this until we refactor to new builder API

openapi/src/test/java/io/helidon/openapi/StaticFileOnlyOpenApiFeatureImpl.java
```diff
@@ -15,15 +15,13 @@
...
 import java.util.HashMap;
 import java.util.Map;
 
 import jakarta.json.Json;
 import jakarta.json.JsonBuilderFactory;
 import jakarta.json.JsonObject;
-import jakarta.json.stream.JsonParser;
 import org.yaml.snakeyaml.Yaml;
 
 public class StaticFileOnlyOpenApiFeatureImpl extends OpenApiFeature {
```
tomas-langer, line 27: This should not be public, nor should the builder method be public - it is an `Impl` class

tomas-langer (changes requested):

openapi/src/main/java/io/helidon/openapi/OpenApiFeature.java
tjquinno: For the sake of speed I will leave this as-is for now.

openapi/src/test/java/io/helidon/openapi/StaticFileOnlyOpenApiFeatureImpl.java
tjquinno: Removed `public` from the class.

romain-grecourt (approved):

romain-grecourt (approved):

tomas-langer (approved):

PR #7118, opened 2023-06-27
Lets unit tests validating JpaExtension and unit tests validating PersistenceExtension run side-by-side; continuation of overall fix for nested transaction problems

ljnelson:
This PR:
* deliberately duplicates the `chirp` series of unit tests as `chirp2` with changes appropriate to the more correct transactional semantics that `PersistenceExtension` enforces
* ensures that the `chirp` and `chirp2` suites can be run together on all JPA providers
* exercises and validates the deliberately obscure System property flags that switch `JpaExtension` and `PersistenceExtension` (its replacement) on and off, proving they can "live" together


integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
```diff
@@ -298,6 +299,23 @@ private <T> void makePersistencePropertyARepeatableQualifier(@Observes BeforeBea
...
+     *
+     * @exception NullPointerException if {@code event} is {@code null}
+     *
+     * @see JtaAdaptingDataSourceProvider
+     */
+    private void vetoDeprecatedJtaDataSourceProvider(@Observes ProcessBeanAttributes<JtaDataSourceProvider> event) {
+        if (!this.enabled) {
+            return;
+        }
+        event.veto();
```
tjquinno, line 316: If a developer includes a type with the deprecated class which we veto, should we log a warning or info message?

integrations/cdi/jpa-cdi/src/test/java/io/helidon/integrations/cdi/jpa/chirp2/TestCascadePersist2.java
```diff
@@ -0,0 +1,302 @@
...
+        this.cdiContainer = initializer.initialize();
+        /*
+        final DataSource ds = this.cdiContainer.select(DataSource.class).get();
+        assertThat(ds, notNullValue());
+        try (final Connection connection = ds.getConnection();
+             final Statement statement = connection.createStatement()) {
+            assertThat(statement, notNullValue());
+            statement.executeUpdate("RUNSCRIPT FROM 'classpath:chirp2.ddl'");
+        }
+        */
```
tjquinno, line 109: Maybe uncomment (if we want these lines) or remove (if we don't want them)?

tjquinno (commented):
A few mostly very minor notes.

integrations/cdi/jpa-cdi/src/test/java/io/helidon/integrations/cdi/jpa/chirp2/TestCascadePersist2.java
ljnelson: Derp; good catch; thanks

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
ljnelson: Ever so slightly tricky: this will only be the case so long as the old and the new stuff coexist, and there's nothing the user can do about it. I'll split the difference and log at the `FINE` level.

tjquinno (approved):

PR #7139, opened 2023-06-30
MP Metrics 5.0 support for 4.x

### LINKED ISSUE #5816: 4.x: Upgrade MP Metrics to 5.0

tjquinno:
Resolves #5816 

| 👀  | It's a long list of reviewers! Changes in the metrics API affected many parts of Helidon that use metrics, including several integrations. I've added people who IIRC have worked on such integrations so you can take a look. |
|----|:-----------------|

# Overview

Note: Look for 🚧 **_User action_** flags below for what developers need to do to adapt to the changes. The MP spec section on [migration hints](https://download.eclipse.org/microprofile/microprofile-metrics-5.0.0/microprofile-metrics-spec-5.0.0.html#_migration_hints) has information also. 

The latest MP Metrics release is 5.0.1. 

This PR contains all the code changes _that I know of_ to implement MP Metrics 5.0. As noted in separate meetings, the TCK is not yet running because the way we set up the Arquillian environment leads to many test failures. That remains a work in progress.

To detect other problems, I have commented out the invocation of the metrics TCK in this PR. 

What this PR does _not_ include:
* doc changes
* use of the new Helidon injection/builder approaches
* reactive removal
* removal of `nima` directories

# MP Metrics 5.0
There were breaking changes in the MP metrics libraries--to help simplify the API and bring MP metrics in closer alignment with Micrometer and OpenTelemetry metrics. Notable changes:

* metric types removed
  
  🚧 **_User action_** - use an alternative metric type and annotation

| Metric type/annotation removed |  Alternative |
|----------|-------------------|
| `ConcurrentGauge` <br> `@ConcurrentGauge` | `Gauge` <br> `@Gauge` |
| `Meter` <br>`@Metered` | `Counter` <br> `@Counted` - back-ends can track the changes in the counter value over time and create time series analysis from that (meter used to maintain historical data itself in the server) |
| `SimpleTimer` <br> `@SimplyTimed` | `Timer` <br> `@Timed` |

* metadata removals
  * type
  * display name
  
  🚧 **_User action_** - Remove code that sets the type or display name in the metadata programmatically and remove settings in the `@Metric` annotation.
* Other 
   * The notion of a registry "type" with fixed values of base, vendor, and application has given way to the idea of a "scope" instead. There are three built-in scopes--base, vendor, and application--but users can create their own scopes.
      * `MetricRegistry.Type` was an enum covering base, vendor, and application. It's gone because users can create their own so an enum doesn't really work.
      * `@RegistryType` is deprecated, intended to be replaced by `@RegistryScope`. Unfortunately, `@RegistryScope`  is not (yet) marked as a CDI qualifier (that'll change in an upcoming release of MP Metrics), so for now our CDI code still has producers for the three registry types using `@RegistryType` as well as a producer for injection sites that use `@RegistryScope` rather than `@RegistryType`.   
      
      🚧 **_User action_**  
      * Change uses of `MetricRegistry.Type` to a `String` with the corresponding name. The `MetricRegistry` interface declares `String` constants for the predefined scopes.
      *  Over time, change uses of `@RegistryType` (deprecated but not yet removed) to `@RegistryScope`.
  * All metrics that share the same name must:
      * Have the same metadata (this is not new)
      * Have the same tag _name_ set (this is new). Apparently the Prometheus exporter (not ours) needs this restriction for some reason in order to produce correct output; I am not at all familiar with that code but the restriction is written into the spec and checked by the TCK so our code enforces this.  
      
     🚧 **_User action_**  - Make sure all metrics that share the same name use the same tag names.
   * The metric registry API no longer permits you to create your own instance of one of the supported metric types (such as your own `Counter`) and register it. This is because Micrometer and OpenTel are totally in control of manufacturing the metric instances. 
     
     Some of our code (integrations, for example) used to do this in order to wrap other technology's measurements with custom implementations of the MP metrics (typically a counter). For the most part where I had to change our code to adjust, I changed such places to use gauges instead. In line with what Micrometer supports, The `Registry` interface usable from SE apps includes a way to register a read-only `Counter` by passing a `ToDoubleFunction` and an object to which to apply that function. But this is not visible through the MP `MetricRegistry` interface. In some places I used that approach.  
      
     🚧  **_User action_** - Use a `Gauge` metric type to expose a value that is updated outside of the metrics API. 

   * Change in `Gauge` to `Gauge<? extends Number>` - The `Gauge` metric was always intended to expose numeric data. The updated API now enforces this.
   
     🚧  **_User action_** - Declare each of your `Gauge` metrics so it reflects the specific numeric type the gauge exposes.

Metric types remaining from earlier releases are `Counter`, `Gauge<? extends Number>`, `Histogram`, and `Timer`. 

# Why do so many files have to change?
Our SE metrics implementation has always dependend--and, for the moment, still does depend--on the MP metrics types. We plan for that to change by introducing a neutral Helidon metrics API which will eventually have implementations which delegate to Micrometer and OpenTelemetry metrics. 
Then our MP metrics implementation will invoke the neutral Helidon API, as will customers who develop SE apps. 
People will be able to choose whether to use Micrometer or OTel underneat for either SE or MP apps.
But we have not yet defined or implemented the neutral API yet. (I have thought a lot about what that API should look like but the code is not written yet.)

So, the MP metrics changes rippled through our SE metrics API and implementation, and into other code that uses metrics, and into examples, and into tests.

# Other notes
Although this PR does not introduce the neutral Helidon metrics API, there are some beginning steps in that direction. 
For example, the new `MetricFactory` SPI interface encapsulates the details of creating new metrics. There are two implementations so far: one that delegates to Micrometer and one that provides no-op implementations of the various metrics.

# Other changes in behavior
## Prometheus output
In addition to the API changes, the Prometheus output now is slightly different. Each line showing a metric value (that is, _not_ the `# TYPE` or `# HELP` lines) has an additional tag to identify the scope where that metric was registered. In MP, the tag name is `mp_scope` per the spec; in SE it (for the moment) is `h_scope`. 

This is because our code now stores metrics in the Micrometer meter registry...a single meter registry for _all_ scopes we and our developers might create. That's the way Micrometer works. 

The Micrometer world does not use different meter registries for scoping but rather for dealing with different types of backends/consumers of metrics output. Typical applications use the Micrometer global registry for adding, removing, and querying meters. 
That global registry is a _composite_ registry, meaning you can add technology-specific meter registries to it. 
There is a Prometheus Micrometer meter registry Helidon creates and adds (if absent) to the global registry.
Applications create and add other types of meter registries to the global one to support other back-ends.

Further, the Prometheus Micrometer meter registry knows how to format its output so we will be able to remove our own code that does that. 

## JSON output
It's still there in Helidon, but the MP metrics spec no longer requires it. At this moment, I have not added back the JSON `OPTIONS` support for metadata. 
I might do that if I have time.

We might want to remove JSON support at some point--perhaps deprecate it in 4.0--but there was some Helidon code that used it for testing and there might be user code out there that relies on it. 
Users will have plenty to adapt to in 4.0 already, so it might be worth keeping JSON format output for now to ease that migration.

I already changed some of the JSON-dependent testing code to use Prometheus output instead; it is possible to do that everywhere if we need to.
Because there will be a cost to us in keeping it around--even if deprecated--until 5.0, we might want to remove it now. 
To my knowledge, MP metrics never marked the JSON output as deprecated during the MP metrics 4.x lifespan, for what that's worth.

# Current state

## Work done
1. AFAIK, all our MP and SE code and other files (excluding doc) are now consistent with the new MP Metrics spec and changes in the API. 
2. I've updated the MP fault tolerance dependency from 4.0 to 4.0.2 so it depends on the updated MP metrics library instead of the old one as 4.0 did.

## Work left to do for full MP metrics support in M1
1. Adapt to upcoming big merges as needed.
3. The MP spec contains a [section](https://download.eclipse.org/microprofile/microprofile-metrics-5.0.0/microprofile-metrics-spec-5.0.0.html#micrometer-implementations) with suggestions (not requirements) for vendors who choose to use Micrometer for the implementation, mostly how to plug in and configure Micrometer support for other back-ends. This is nice to have but is certainly not required for M1.


tomas-langer:
There seems to be a problem with the calls to registry. Calls to `/observe/metrics` fail with `404`. But if I call `/observer/metrics/base`, it succeeds, and then all calls to `/observer/metrics` succeed as well. 
Investigating

tjquinno:
Exactly. I also found this independently. Pushed a fix to the PR branch.

tjquinno:
We definitely need to do this for base (but only if it is enabled; remember metrics can be disabled by scope).

But you think also for vendor? Base is the only registry lazily instantiated upon first read.

tjquinno:
Still investigating, but the actual underlying cause might be different. 

With the change to the `scopes` method to always return `base` I am seeing failures in the bookstore functional test because the `request.count` metric from the key performance indicator metrics is not present in the output.

It seems that maybe the metrics feature post-setup code is not invoked when expected. That code should be fetching the base metrics registry to trigger its creation and also set up the KPI metrics...and neither seems to be happening and in the debugger I don't see the metrics feature `postSetup` method invoked.

tjquinno:
BTW, the latest pushes resolve the problem Tomas and I discovered with the base metrics.

klustria (approved):
As it relates to changes in `integrations/oci/metrics`, it LGTM, hence giving approval from that perspective.

metrics/metrics/src/main/java/io/helidon/metrics/HelidonFunctionalCounter.java
```diff
@@ -0,0 +1,96 @@
...
+     * @param scope scope of the registry
+     * @param metadata metadata describing the functional counter
+     * @param origin object which provides the counter value
+     * @param function function which, when applied to the origin, yields the counter value
+     * @param tags further identifies the functional counter
+     * @return new functional counter
+     * @param <T> type of the origin
+     */
+    static <T> HelidonFunctionalCounter<T> create(MeterRegistry meterRegistry,
+                                               String scope,
```
trentjeff, line 46: nit: alignment

metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactory.java
```diff
@@ -137,6 +150,27 @@ default void update(MetricsSettings metricsSettings) {
...
+     * selection and meter name selection.
+     *
+     * @param mediaType {@link io.helidon.common.media.type.MediaType} to control the output format
+     * @param scopeSelection {@link java.lang.Iterable} of individual scope names to include in the output
+     * @param meterNameSelection {@link java.lang.Iterable} of individual meter names to include in the output
+     * @return {@link String} meter exposition as governed by the parameters; {@code empty} if no metrics matched the selections
+     * @throws java.lang.IllegalArgumentException if the implementation cannot handle the requested media type
+     * @throws java.lang.UnsupportedOperationException if the implementation cannot expose its metrics
+     */
+    Optional<?> scrape(MediaType mediaType, Iterable<String> scopeSelection, Iterable<String> meterNameSelection);
```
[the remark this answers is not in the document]
tjquinno: We could return `null` instead if the provided selections result in no output, but it's against our guidelines to use `null` as part of our APIs.

metrics/api/src/test/java/io/helidon/metrics/api/TestDeletions.java
```diff
@@ -0,0 +1,58 @@
...
+    @BeforeEach
+    void clear() {
+        reg.removeMatching(MetricFilter.ALL);
+    }
+
+    @Test
+    void addCounterWithTag() {
+        Counter counter = reg.counter(COMMON_COUNTER_NAME, new Tag("myTag", "a"));
+        assertThat("New counter value", counter.getCount(), is(0L));
+        counter.inc();
```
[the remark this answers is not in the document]
tjquinno: The `@BeforeEach` exercises the metric deletion feature. The increment of the counters in each test verifies that the counter was, in fact, deleted correctly by that code. If the deletion does not work correctly, then the counter value in the _other_ test could be non-zero as a result of the increment at the end of the first test.

metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactory.java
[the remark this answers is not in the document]
trentjeff: its not the optional i am commenting on - it is the use of the wildcard generic

metrics/metrics/src/main/java/io/helidon/metrics/HelidonFunctionalCounter.java
tjquinno: Will change in this PR if there are also other, functional, changes needed.

metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactory.java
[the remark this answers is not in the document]
tjquinno: Thanks for clarifying. 

If you look at the two formatters that can be called, the Prometheus one returns an `Optional<String>` and the JSON one returns an `Optional<JsonObject>`.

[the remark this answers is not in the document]
trentjeff: in that case why not return a Builder with two Optional<> attributes?  I think that would be better than using generics.

examples/metrics/exemplar/src/test/java/io/helidon/examples/metrics/exemplar/MainTest.java
```diff
@@ -148,10 +148,14 @@ public void testMetrics() {
 
     }
 
+    // TODO Helidon will be delegating exemplar handling to the underlying implementation. This is a work in progress.
     private static String valueMatcher(String statName) {
         // application_timerForGets_mean_seconds 0.010275403147594316 # {trace_id="cfd13196e6a9fb0c"} 0.002189822 1617799841.963000
         return "application_" + GreetService.TIMER_FOR_GETS
-                + "_" + statName + "_seconds [\\d\\.]+ # \\{trace_id=\"[^\"]+\"\\} [\\d\\.]+ [\\d\\.]+";
+                // TODO Following is the original, exemplar-matching pattern. Suppressed temporarily while we migrate
```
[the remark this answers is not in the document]
tjquinno: Will update in this PR if there are other, functional, changes as well.

integrations/cdi/datasource-hikaricp/src/main/java/io/helidon/integrations/datasource/hikaricp/cdi/MicroProfileMetricsTrackerFactory.java
```diff
@@ -34,6 +34,7 @@ class MicroProfileMetricsTrackerFactory implements MetricsTrackerFactory {
         this.registry = null;
     }
 
+    // TODO change to RegistryScope once MP makes it a qualifier
```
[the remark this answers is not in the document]
tjquinno: Will update in this PR if there are other, functional, changes as well.

metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactory.java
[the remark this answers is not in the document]
tjquinno: As is, the value (if present inside the `Optional`) is sent directly as the entity in the response. The code in `MetricsFeature` has no need to know the specific type of the returned value. The media support handles whatever the actual type is in sending the response. It's very flexible if we ever need to add another formatter that produces another type. 

Adding another builder would (IMO) unnecessarily complicate things now,  and if we were to add other format types in the future then we'd have to update the builder as well. As is, a hypothetical new formatter would just return its type and we'd be done.

metrics/api/pom.xml
```diff
@@ -50,6 +50,14 @@
             <groupId>org.eclipse.microprofile.metrics</groupId>
             <artifactId>microprofile-metrics-api</artifactId>
         </dependency>
+        <dependency>
```
tomas-langer, line 53: Why is Micrometer an explicit dependency here? I thought the APIs should be MP and not Micrometere?

metrics/api/src/main/java/io/helidon/metrics/api/MetricsProgrammaticSettings.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Internal use class to hold a reference to the singleton.
+     */
+    class Instance {
+
+        private Instance() {
+        }
+
+        private static final LazyValue<MetricsProgrammaticSettings> INSTANCE = LazyValue.create(() ->
```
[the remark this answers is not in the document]
tomas-langer: This is the correct approach for now. We may want to use contract based lookup, but only when running within injection service registry. It still must work outside of it

metrics/api/src/main/java/io/helidon/metrics/api/MetricsProgrammaticSettings.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Programmatic (rather than configured) settings that govern certain metrics behavior.
+ * <p>
+ *     Implementations of this interface are typically provided by Helidon itself rather than
+ *     developers building applications and are not intended for per-deployment (or even per-application)
+ *     customization.
+ * </p>
+ */
+public interface MetricsProgrammaticSettings {
```
tomas-langer, line 32: Please add a follow up for naming here - we use `Config` as the suffix for configuring components in other modules, I think it should be used for this class as well

metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactory.java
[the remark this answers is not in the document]
tomas-langer: btw `Optional<Object>` would work the same and it is less unusual on the API

metrics/metrics/src/main/java/io/helidon/metrics/HelidonFunctionalCounter.java
```diff
@@ -0,0 +1,96 @@
...
+import org.eclipse.microprofile.metrics.Counter;
+import org.eclipse.microprofile.metrics.Metadata;
+import org.eclipse.microprofile.metrics.Tag;
+
+/**
+ * Implementation of a counter wrapper around an object capable of providing a double value.
+ *
+ * @param <T> type of the object providing the value
+ */
+class HelidonFunctionalCounter<T> extends MetricImpl implements Counter {
```
tomas-langer, line 31: Do I understand it correctly that this is basically a Gauge? Why do we still have it? Or is it represented differently by the format?

metrics/metrics/src/main/java/io/helidon/metrics/HelidonGauge.java
```diff
@@ -19,42 +19,198 @@
...
+     * To work around this, we keep track ourselves of the function and target or supplier which report the gauge value.
+     * Then our getValue() implementation simply invokes the function on the target or the supplier rather than delegating to
+     * the Micrometer gauge value() method (which would do exactly the same thing anyway).
+     *
+     * This way the typing works out (with the expected unchecked cast).
+     */
 
-    private HelidonGauge(String registryType, Metadata metadata, Gauge<T> metric) {
-        super(registryType, metadata);
+    private final io.micrometer.core.instrument.Gauge delegate;
```
tomas-langer, line 51: I thought we should have support for Micrometer and possibly other metrics implementation, but not hardcoded into our impl, but through  provider modules.
Why is there a hard dependency on Micrometer?

metrics/api/src/main/java/io/helidon/metrics/api/RegistryFactory.java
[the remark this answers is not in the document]
tjquinno: Added to follow-up tasks issue #7193

metrics/api/pom.xml
tjquinno: Yes, not ideal, but we plan for Helidon to introduce its own neutral metrics API and, once that is in place, we will change our MP metrics implementation to  depend on that rather than directly on Micrometer.

But due to scheduling complications earlier in the spring and given our goal of having MP metrics support in M1 we decided to go ahead with a first MP implementation without waiting for the neutral API, knowing we would change the MP implementation once the neutral API is in place. 

The neutral API will likely be very close to the Micrometer API, partly for the convenience of developers and partly because of Micrometer's maturity in this space. In that case, the changes we will need to make in our MP implementation, although not zero, will be relatively small. Much of the code already uses a `MetricFactory` SPI abstraction to help minimize the changes required later.

We would definitely have preferred to have the neutral API in place first, but I just couldn't make that happen in time for MP metrics in M1.

metrics/metrics/src/main/java/io/helidon/metrics/HelidonFunctionalCounter.java
tjquinno: Basically, this allows a `Counter`-style "wrapper" around an externally-updated value instead of a `Gauge`-style wrapper.

Earlier releases of MP metrics allowed developers to create their own implementations of metrics and register them. Helidon used this to register custom counters, for example. 

MP Metrics 5.0 no longer permits this, but the ability to have a `Counter`-like wrapper is useful (and present in Micrometer and OTel metrics) so will likely be in the Helidon neutral API as well.

metrics/metrics/src/main/java/io/helidon/metrics/HelidonGauge.java
tjquinno: See earlier note about the neutral API and timing.

metrics/api/pom.xml
spericas: We should try to fix this for M2. It doesn't look very nice. I should be able to help Tim.

tjquinno: Yes, we should implement our neutral API and adapt SE and MP to it soon.

tomas-langer (approved):
There are things to follow up on, let's merge so we have MP 6 complete

metrics/metrics/src/main/java/io/helidon/metrics/HelidonGauge.java
tomas-langer: I did not realize from the description that we must use Micrometer to be able to generate prometheus output (as we used to generate it ourself).
I have approved the PR, we need to fix this eventually when we switch to Helidon Metrics API

tjquinno: Given that our neutral API will layer on one or more actual implementations (Micrometer, OTel metrics), we should delegate to the underlying implementation to publish, whether Prometheus or other (push or pull) models, rather than duplicate that functionality in our own code.

Issue #6669, opened 2023-04-21
4.x: Níma webclient redesign


tomas-langer:
Currently we need to decide whether to use HTTP/2 or HTTP/1.1 client. I think we should support a use case, where I can tell the client "upgrade to HTTP/2 if available, use HTTP/1 otherwise" and use a single set of APIs for both.

We should also move all tests from reactive webclient to Níma, to support the same use cases

Also add possibilitty to get full response with entity - now we have to try with resources even if we read entity as string.
E.g.:
- `ClientRawResponse` - implements `Autocloseable` and must be closed
- `ClientResponse<Type>` - when you use `request(Type.class)`

Copied from duplicate:

Currently we can create version specific clients only. We should support a builder that uses common APIs (e.g. HTTP/1 as the common denominator), but allows intention to use a different protocol version.

Also `WebClient.builder()` should return `HttpClient.Builder` and not `Http1Client.Builder`
```
Http2Client http2 = Http2Client.builder()
                .build();
        Http1Client http1 = Http1Client.builder()
                .build();

        String s = http2.get("/test")
                .request()
                .as(String.class);

        HttpClient http = HttpClient.builder()
                .preferredVersions(Http2Client.PROTOCOL) // ALPN (http,h2,h3)
                .build();
        
        http.get("/test")
                .request() // uses the connection from SPI below
                .as(String.class);
        // SPI "Optional<ClientConnection> connect(Tls tls, Uri endpoint)"
```

Add configuration/builder option to use service loader (to try HTTP/2 for any request even if not expllcitly used). Default will be false, and use HTTP/1

PR #7149, opened 2023-07-05
4.x: Support for cookies in new WebClient

spericas:
Support for cookies in new WebClient. Only HTTP/1.1.

- Configuration seems like a work in progress: there's config in `WebClient`, in subclasses and there's also `Http1ClientConfig`. There is an overlap of config params at these locations. This PR _does not attempt to fix this_ but simply adds additional methods to configure cookies. All this needs to be cleaned up in a future PR.

- The cookies implementation is loosely based on the one 3.x, still using some JDK classes.

- A cookie store is associated with a WebClient instance. This required some changes in the implementation so that a client instance is provided instead of a config instance (`Http1ClientConfig`) in some places. Several of the changes are related to this. 

- There's a new test for cookies that is part of the PR.


nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/WebClientCookieManager.java
```diff
@@ -0,0 +1,130 @@
...
+     * @param cookiePolicy policy to accept cookies
+     * @param cookieStore storage for cookies
+     * @param defaultCookies default cookies to send
+     * @param acceptCookies flag to handler acceptance of cookies
+     * @return new cookie manager
+     */
+    public static WebClientCookieManager create(CookiePolicy cookiePolicy,
+                                         CookieStore cookieStore,
+                                         Map<String, String> defaultCookies,
+                                         boolean acceptCookies) {
```
Verdent, line 70: please check if formatting of this is correct (I am not sure if this is not just visual issue of github)

Verdent (changes requested):
Otherwise I think it is OK :-)

tomas-langer:
I have taken over this PR and I am making it a part of the webclient redesign effor (#6669)

PR #7151, opened 2023-07-05
2.x Update validate workflow to build using Java 17

barchetta:
This is an experiment to see how well 2.x branch builds when using Java 17


barchetta:
Ah yes, javadocs.

Issue #7143, opened 2023-07-03
Nima Proxy TODOs


jbescos:
This PR https://github.com/helidon-io/helidon/pull/6441 is missing some functionalities:

- [x] Support it in HTTP 2
- [x] Support Proxy authentication
- [ ] Support SOCKS ?? -> we are not going to support it
- [x] Some proxies respond with HTTP 1.0 protocol and Http1StatusParser will fail
- [x] HTTP CONNECT should specify the host:port of the remote (not the proxy)

PR #7168, opened 2023-07-11
CONNECT host:port is wrong

jbescos:
This address the point _HTTP CONNECT should specify the host:port of the remote (not the proxy)_
in https://github.com/helidon-io/helidon/issues/7143

In my opinion, it makes no sense because the header HOST is mandatory and it contains the remote host:port, but you can try it by yourself configuring a proxy in your browser. For example when you visit youtube you will see the next:
```
CONNECT www.youtube.com:443 HTTP/1.1
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:108.0) Gecko/20100101 Firefox/108.0
Proxy-Connection: keep-alive
Connection: keep-alive
Host: www.youtube.com:443
```
Some proxies could not work properly if the CONNECT is not set correctly.

I added also the keep alive headers.

### DIFF
--- a/common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
+++ b/common/http/src/main/java/io/helidon/common/http/HeaderEnum.java
@@ -77,6 +77,7 @@ enum HeaderEnum implements Http.HeaderName {
     LINK("Link"),
     LOCATION("Location"),
     PRAGMA("Pragma"),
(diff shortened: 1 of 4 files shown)


jbescos:
Moving PR to here: https://github.com/helidon-io/helidon/pull/7287

Issue #6732, opened 2023-05-01
3.x: database-mp native-image fails at runtime with `Unable to load class [org.hibernate.dialect.H2Dialect]`


barchetta:
# Environment Details
* Helidon Version: 3.2.0 (also 3.2.1)
* Helidon MP
* JDK version: java 17.0.3 2022-04-19 LTS
* OS: Linux
* GRAALVM native-image version: GraalVM 22.3.1 Java 17 CE

## Problem Description

helidon-database-mp archetype example compiles to native image, but fails at runtime:

```
2023.05.01 18:08:26 INFO io.helidon.common.LogConfig Thread[main,5,main]: Logging at runtime configured using classpath: /logging.properties
2023.05.01 18:08:26 INFO org.hibernate.jpa.internal.util.LogHelper Thread[main,5,main]: HHH000204: Processing PersistenceUnitInfo [name: pu1]
2023.05.01 18:08:26 WARN org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator Thread[main,5,main]: HHH000342: Could not obtain connection to query metadata
org.hibernate.boot.registry.selector.spi.StrategySelectionException: Unable to resolve name [org.hibernate.dialect.H2Dialect] as strategy [org.hibernate.dialect.Dialect]
	at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.selectStrategyImplementor(StrategySelectorImpl.java:155)
	at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:237)
	at org.hibernate.boot.registry.selector.internal.StrategySelectorImpl.resolveStrategy(StrategySelectorImpl.java:190)
	at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.constructDialect(DialectFactoryImpl.java:96)
	at org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl.buildDialect(DialectFactoryImpl.java:59)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:206)
	at org.hibernate.engine.jdbc.env.internal.JdbcEnvironmentInitiator.initiateService(JdbcEnvironmentInitiator.java:36)
	at org.hibernate.boot.registry.internal.StandardServiceRegistryImpl.initiateService(StandardServiceRegistryImpl.java:119)
[...]

Issue #3712, opened 2021-12-08
Add native-image related configuration files to Helidon modules


arjav-desai:
While testing native-image support for OCI ATP integration, we have found that we have to set following config 

1. In native-image.properties

```
Args=-H:+RemoveSaturatedTypeFlows \
  --initialize-at-run-time=oracle.ucp.common.Service
```

2. In (helidon specific) reflection-config.json 

```
[
  {
    "name": "oracle.ucp.jdbc.PoolDataSourceImpl",
    "allDeclaredConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allDeclaredFields": true,
    "allPublicFields": true
  }
]
```

else we run into build and runtime issue.

To ease the end user application development, we can set this on our UCP integration module instead.

PR #7170, opened 2023-07-11
native image config for data modules 3.x

arjav-desai:
#3712 #6732


integrations/db/ucp/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    <properties>
+        <spotbugs.skip>true</spotbugs.skip>
+        <maven.javadoc.skip>true</maven.javadoc.skip>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.oracle.database.jdbc</groupId>
+            <artifactId>ucp</artifactId>
+            <type>jar</type>
```
romain-grecourt, line 41: `<type>jar</type>` is the default, you can remove it.

PR #7172, opened 2023-07-12
Allow HTTP 1.0 responses

jbescos:
Relates to the point `Some proxies respond with HTTP 1.0 protocol and Http1StatusParser will fail` in https://github.com/helidon-io/helidon/issues/7143


jbescos:
Closing because I will try to address many proxy tasks in one PR

PR #7178, opened 2023-07-12
Fix path handling for metrics observe provider

### LINKED ISSUE #7044: 4.x: inconsistent metrics endpoint
Assuming the following dependencies:

```xml
<dependency>
    <groupId>io.helidon.nima.observe</groupId>
    <artifactId>helidon-nima-observe-metrics</artifactId>
</dependency>
<dependency>
    <groupId>io.helidon.metrics</groupId>
    <artifactId>helidon-metrics</artifactId>
</dependency>
```

The following code serves metrics at `/observe/metrics`:

```java
routing.addFeature(ObserveFeature.create());
```

However the following code serves metrics at `/metrics`:

```java
routing.addFeature(ObserveFeature.builder()
                                 .addProvider(MetricsObserveProvider.create())
                                 .build());
```

tjquinno:
Resolves #7044 

1. Looks like a line was omitted from the `MetricsObserveProvider#register` method when it was first created that deals with this scenario. (At least it seems that way comparing to health.)
2. A stray `/` in the `MetricsFeature` ctor needed to go away.

Also requires adding a protected method to `MetricsFeature` for visibility.

### DIFF
--- a/nima/observe/metrics/src/main/java/io/helidon/nima/observe/metrics/MetricsFeature.java
+++ b/nima/observe/metrics/src/main/java/io/helidon/nima/observe/metrics/MetricsFeature.java
@@ -182,6 +182,11 @@ public class MetricsFeature extends HelidonFeatureSupport {
     }
 
     @Override
+    protected void context(String context) {
+        super.context(context);
+    }
+
+    @Override
     protected void postSetup(HttpRouting.Builder defaultRouting, HttpRouting.Builder featureRouting) {
         configureVendorMetrics(defaultRouting);
     }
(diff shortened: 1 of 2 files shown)


nima/observe/metrics/src/main/java/io/helidon/nima/observe/metrics/MetricsFeature.java
```diff
@@ -181,6 +181,11 @@ public void afterStop() {
         }
     }
 
+    @Override
+    protected void context(String context) {
+        super.context(context);
```
romain-grecourt, line 186: Is this left-over? this override doesn't do anything.

tjquinno: Except make the method visible within package to `MetricObserveProvider` (same as with health).

romain-grecourt: roger that.

romain-grecourt (approved):

Issue #7051, opened 2023-06-22
4.x: Injection fails when starting server in parallel


tomas-langer:
Cannot start server in parallel when pico is used - fails with `service already initialized`, works fine when starting listeners in sequence

To reproduce:
 - `LoomServer.java`: uncomment lines 278+ and comment out 
 ```
        for (ServerListener listener : listeners.values()) {
            listener.start();
        }
```
- run `FtMain` in níma examples - you will get the exception

Issue #6498, opened 2023-03-28
Pico Services Concurrency


trentjeff:
`synchronized` was removed on Pico services `initializeServices()`, `reset()`, and `shutdown()`. A concurrent lock should be placed here instead to ensure (the somewhat unlikely) chance of someone calling reset() or shutdown() on a different thread while initialization is occurring.

Issue #6434, opened 2023-03-20
Pico shutdown should have a timeout configuration


trentjeff:
See DefaultPicoServices.doShutdown().

```
            return es.submit(shutdown)
                    // note to self: have an appropriate timeout config for this
//                    .get(cfg.activationDeadlockDetectionTimeoutMillis(), TimeUnit.MILLISECONDS);
                    .get();
```

trentjeff:
Most of this work apparently was already completed. Just putting some cleanup in https://github.com/helidon-io/helidon/issues/7051.

PR #7200, opened 2023-07-14
LoomServer parallel listener start

trentjeff:
Addresses issue #7051, #6434, and #6498.


trentjeff:
Note that #7051 was never really an issue for me locally or in the pipeline.  I need that this as an opportunity to address related issues (see description above).

tomas-langer (approved):

PR #7202, opened 2023-07-14
Remove example(s) that depend on node.js

### LINKED ISSUE #7144: Remove example(s) that depend on node.js
A lot of intermittent failures are seen because of these (due to failures to download node components).
Also this brings in a huge set of additional dependencies, and even though these are just for an example, I think it increases the complexity not in a useful way.

tvallin:
fix #7144


tvallin:
This example will be refactored instead of removed, therefor closing this PR.

PR #7228, opened 2023-07-21
Fix gRPC to calculate the correct class name for method request and response types

thegridman:
The current code in the `Grpc` class assumes that the request and response types for methods will be in the same package as the Java package option in the service photo file. This is not always true because you can import types from other port files. For example in Coherence we import some base Google proto types like `Empty`.

I'm also curios whether line 128 where `.setType` is called in the code below is correct.
```
        io.grpc.MethodDescriptor.Builder<ReqT, ResT> grpcDesc = io.grpc.MethodDescriptor.<ReqT, ResT>newBuilder()
                .setFullMethodName(io.grpc.MethodDescriptor.generateFullMethodName(serviceName, methodName))
                .setType(io.grpc.MethodDescriptor.MethodType.UNARY).setFullMethodName(path).setRequestMarshaller(reqMarshaller)
                .setResponseMarshaller(resMarshaller).setSampledToLocalTracing(true);
```

This code always calls `.setType(io.grpc.MethodDescriptor.MethodType.UNARY)` but the actual method may not be unary.


tomas-langer (approved):

tomas-langer:
The unary type is most likely wrong here.

tomas-langer (approved):

PR #7238, opened 2023-07-24
3.x: upgrade okio to 3.4.0

barchetta:
* upgrade okio to 3.4.0
* suppress graalvm sdk false positive

This version of okio has issues with Java modules (for example see https://github.com/square/okio/issues/1306).  This version of okio is written in Kotlin and splits the original `okio` artifact into `okio` (platform agnostic stuff) and `okio-jvm` (Java classes).  We don't need `okio` at runtime (only `okio-jvm`), and it was causing modularity problems. To work-around this:

1. Exclude `okio` from the runtime environment (`libs` folder). 
2. In `helidon-tracing-jaeger` add `requires kotlin.stdlib;` so that `kotlin-stdlib` is included in the module graph. We need this because `kotlin-stdlib` is a well-formed Java module, but `okio-jvm` (which depends on it) is not, and therefore we need some module to depend on `kotlin.stdlib` for it to be included in the module graph.

Okio is a transative dependency of okhttp which is a transitive dependency of a couple third parties (opentelemetry-exporter-jaeger for example).  okhttp is not a direct third party dependency of Helidon -- except for one or two test dependencies.

### DIFF
--- a/applications/parent/pom.xml
+++ b/applications/parent/pom.xml
@@ -105,6 +105,8 @@
                                 <overWriteIfNewer>true</overWriteIfNewer>
                                 <overWriteIfNewer>true</overWriteIfNewer>
                                 <includeScope>runtime</includeScope>
+                                <!-- Hack to work-around https://github.com/square/okio/issues/1306 -->
+                                <excludeArtifactIds>okio</excludeArtifactIds>
                             </configuration>
                         </execution>
                     </executions>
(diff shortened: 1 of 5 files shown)


barchetta:
Running into https://github.com/square/okio/issues/1306

romain-grecourt (approved):

PR #7263, opened 2023-07-28
Remove VirtualExecutorUtil.java

### LINKED ISSUE #4869: Remove VirtualExecutorUtil

tomas-langer:
Resolves #4869


barchetta (approved):
LGTM

PR #7281, opened 2023-07-31
Prototype builder update

### LINKED ISSUE #7212: 4.x MediaContext requires .build()
* Helidon Version: 4.0.0-ALPHA6
----------

The following does NOT include the default media support that is registered by `io.helidon.nima.http.media.MediaContextBuilderInterceptor`.

```java
 WebServer.builder().mediaContext(MediaContext.builder());
```

Instead you get a `MediaContext` that is empty and unable to read or write any entity types.

However if you invoke `.build()` everything works ok.

```java
WebServer.builder().mediaContext(MediaContext.builder().build());
```

The issue is that the builder implements the config type, and the config type is consumed by other parts.

```
// this works because the builder implements MediaContextConfig
// not because the method is overloaded to accept Supplier<MediaContextConfig>
// and the call the .get() is equivalent to .build()
WebServer.builder().mediaContext( MediaContext.builder());
```

### LINKED ISSUE #7283: 4.x: Builder interceptor to decorator
Builder interceptor - either change the generated code to use the return value, or change the interface and name to only accept builder instance (to allow mutation, not replacement)

tomas-langer:
Fixes #7212 
Fixes #7283 

Also changed interceptor to decorator, as interceptor not needed for builders


tomas-langer:
Rebased on main

romain-grecourt (commented):
The following does not compile anymore.

```java
WebServer.builder().mediaContext(MediaContext.builder());
```

The builder should have a variant of the `mediaContext` method that accepts `Supplier<MediaContext>`.

nima/webclient/api/src/main/java/io/helidon/nima/webclient/api/HttpClientConfigBlueprint.java
```diff
@@ -151,7 +151,7 @@ default ClientRequestHeaders defaultRequestHeaders() {
      *
      * @return media context
      */
-    @ConfiguredOption
+    @ConfiguredOption("create()")
```
romain-grecourt, line 154: Is "create()" a config key  ?

nima/webclient/http1/src/main/java/io/helidon/nima/webclient/http1/Http1ClientConfigBlueprint.java
```diff
@@ -17,17 +17,19 @@
...
  */
-@Prototype.Blueprint(builderInterceptor = Http1ClientConfigSupport.Http1BuilderInterceptor.class)
+@Prototype.Blueprint
 interface Http1ClientConfigBlueprint extends HttpClientConfig, Prototype.Factory<Http1Client> {
     /**
      * HTTP/1.1 specific configuration.
      *
      * @return protocol specific configuration
      */
+    @ConfiguredOption("create()")
```
romain-grecourt, line 33: Is "create()" a config key  ?

nima/webclient/api/src/main/java/io/helidon/nima/webclient/api/HttpClientConfigBlueprint.java
tomas-langer: create() is a static method on the type - this is the default value

tomas-langer: create() is a static method on the type - this is the default value

tomas-langer:
I would prefer to add the methods that accept supplier as a follow up to this issue. This is fixing the problem of sending builder where we should have sent a built value.

romain-grecourt (approved):

PR #7308, opened 2023-08-03
4.x: Change h2 dependency to provided

barchetta:
Changes the scope of our H2 dependency to `provided`.  

This means users will need to add an explicit dependency on H2 to use it.

### DIFF
--- a/integrations/db/h2/pom.xml
+++ b/integrations/db/h2/pom.xml
@@ -46,6 +46,7 @@
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
+            <scope>provided</scope>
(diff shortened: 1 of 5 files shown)


integrations/db/h2/pom.xml
```diff
@@ -46,6 +46,7 @@
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
+            <scope>provided</scope>
```
ljnelson, line 49: A question: my understanding is this Maven module exists so that users who want to use H2 in native image can just grab it and forget about it. My guess is that's why H2 is/was in `compile` scope here. Should it really be `provided` now?

PR #7313, opened 2023-08-04
Re-enabled test after integration with latest Jersey

spericas:
Re-enabled test after integration with latest Jersey. Some changes to the example given that Jersey RestClient does appear to provide request scope support with async (there is no propagation). This was used by the test to get the dynamic port in which the server runs. New code uses the Helidon context for that. Test no longer blocks when using `Asynchronous` in a RestClient interface.


barchetta (approved):
LGTM

PR #7335, opened 2023-08-08
Remove public constructors from production types.

### LINKED ISSUE #6414: SecurityPreMatchingFilter is overriding the SecurityContext when using webserver security
## Environment Details
* Helidon Version: 2.5.4
* Helidon MP
* JDK version: 17.0.3
* OS: Mac OS

----------

## Problem Description
An authenticated SecurityContext is being replaced by an unauthenticated one after all filters are executed
## Steps to reproduce
Setup Helidon Security using "web server security". Example:
```
security:
  providers:
    - spectra-auth:
        idcsTenantCertChain: '<idcsTenantCertChain>'
        authentication:
          authenticate: true
        audience: "<audience>"
        idcsUrl: "<idcsUrl>"
        scope: "<scope>"
  web-server:
    paths:
      # authenticate all paths
      - path: "/myapplication/*"
        authenticate: true
        spectra-auth:
          authentication:
            authenticate: true
``` 
Put a break point in the following lines in SecurityPreMatchingFilter:
```
    public void filter(ContainerRequestContext request) {
        SecurityTracing tracing = SecurityTracing.get(); // BREAK POINT 1
        SecurityContext securityContext = this.security().contextBuilder(Integer.toString(CONTEXT_COUNTER.incrementAndGet(), 36)).tracingSpan((SpanContext)tracing.findParent().orElse((Object)null)).build();
        Contexts.context().ifPresent((ctx) -> {
            ctx.register(securityContext);
        });
        ((Ref)this.injectionManager.getInstance((new GenericType<Ref<SecurityContext>>() {
        }).getType())).set(securityContext);  // BREAK POINT 2
        if (this.featureConfig().shouldUsePrematchingAuthentication()) {
            this.doFilter(request, securityContext);
        }

    }
```
In debug mode, execute a request using a valid Bearer Token.
This is the result when executing expressions in both break points:
- Break Point 1
  - Contexts.context().get().get(SecurityContext.class).get().isAuthenticated()
    - Result: true
  - Contexts.context().get().get(SecurityContext.class).get().userName()
    - Result: <authenticated_username>
- Break Point 2
  - Contexts.context().get().get(SecurityContext.class).get().isAuthenticated()
    - Result: false
  - Contexts.context().get().get(SecurityContext.class).get().userName()
    - Result: null

In other words this this the block replacing the existing SecurityContext:
```
        Contexts.context().ifPresent((ctx) -> {
            ctx.register(securityContext);
        });
```

tomas-langer:
Resolves #6414

### DIFF
--- a/nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2HuffmanEncoder.java
+++ b/nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2HuffmanEncoder.java
@@ -37,7 +37,7 @@ package io.helidon.nima.http2;
 import io.helidon.common.buffers.BufferData;
 
 /**
- * Implementation of HPack Huffman decoding and encoding.
+ * Implementation of HPack Huffman decoding.
  */
 public class Http2HuffmanEncoder {
     private static final int HUFFMAN_ENCODED = 1 << 7;
--- a/nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2HuffmanEncoder.java
+++ b/nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2HuffmanEncoder.java
@@ -37,7 +37,7 @@ package io.helidon.nima.http2;
 import io.helidon.common.buffers.BufferData;
 
 /**
- * Implementation of HPack Huffman decoding.
+ * Implementation of HPack Huffman encoding.
  */
 public class Http2HuffmanEncoder {
     private static final int HUFFMAN_ENCODED = 1 << 7;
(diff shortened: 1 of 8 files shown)


nima/http2/http2/src/main/java/io/helidon/nima/http2/Http2HuffmanEncoder.java
```diff
@@ -45,7 +45,16 @@ public class Http2HuffmanEncoder {
     /**
      * Huffman encoder.
      */
-    public Http2HuffmanEncoder() {
```
Verdent, line 48: Please check the javadoc of this class. It should get decoding part removed.

Verdent (changes requested):

Verdent (approved):

PR #7342, opened 2023-08-08
4.x: Add pull request template

barchetta:
### Description

Adds a template for pull requests.

### Documentation

None

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,10 @@
+
+### Description
+
+### Documentation
+
+If enhancement: provide description with example code/config snippet or pointer to issue with the description
+
+If feature: docs in PR or pointer to doc issue/PR
+
+If no doc impact: None
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,6 +5,6 @@
 
 If enhancement: provide description with example code/config snippet or pointer to issue with the description
 
-If feature: docs in PR or pointer to doc issue/PR
+If feature: summarize feature and provide pointer to doc issue
 
 If no doc impact: None


.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+
+### Description
+
+### Documentation
+
+If enhancement: provide description with example code/config snippet or pointer to issue with the description
```
romain-grecourt, line 6: Should these be in pseudo Markdown comments similar to what we do for the issue template ?
These can be cumbersome as sometime people edit the comments directly.

E.g.
```
[//]: # "enhancement: provide example code/config snippet or pointer to issue with the same"
[//]: # "feature: documentation changes are included in the PR or provide a pointer to a documentation issue/PR"
[//]: # "others: use "None" if the PR has no impact on documentation
```

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+
+### Description
+
```
romain-grecourt, line 3: - We should mention links to related issues, and the use of closing keywords.
- OK to repeat title in description

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
...
```
romain-grecourt, line 1: We should mention a few things related to PR titles:

Good:
- Prefix the title with the codeline E.g. `4.x: foo bar`
- Shorter titles are better (listing)

Bad:
- Fix issue 1234
- Backport issue 1234

.github/pull_request_template.md
barchetta: I intentionally went with a less-is-more approach on this first iteration. The main point in this is to get the Docs info populated and I don't want to bury that lede.

After this PR I can work on a more holistic template for "how to write a good PR" and then either put it in the template directly, or put it on a wiki page and link to it (to avoid clutter).

.github/pull_request_template.md
barchetta: Good ideas but see previous comment.

edbratt (approved):
This looks good to me. We should definitely try to improve our processes and this template but let's start simply first.

PR #7345, opened 2023-08-09
4.x: Support Http.Header.X_HELIDON_CN

### LINKED ISSUE #7220: 4.x Support `Http.Header.X_HELIDON_CN`
Add support for `X_HELIDON_CN` header.

Remove the following workaround and replace with usage of `X_HELIDON_CN` header:

- tests/integration/webclient/src/test/java/io/helidon/tests/integration/webclient/CertificateHelper.java
- examples/webserver/mutual-tls/src/main/java/io/helidon/webserver/examples/mtls/CertificateHelper.java

tvallin:
fix #7220


nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1Connection.java
```diff
@@ -57,6 +61,7 @@
  */
 public class Http1Connection implements ServerConnection, InterruptableTask<Void> {
     private static final System.Logger LOGGER = System.getLogger(Http1Connection.class.getName());
+    private static final Pattern CN_PATTERN = Pattern.compile("(.*)CN=(.*?)(,.*)?");
```
tomas-langer, line 64: This pattern is wrong, it would match also `OCN=Helidon`
Also there is no need to create the first capturing group (it can be a non-capturing group instead)

nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1Connection.java
```diff
@@ -136,6 +141,10 @@ public void handle(Semaphore requestSemaphore) throws InterruptedException {
                 currentEntitySizeRead = 0;
 
                 WritableHeaders<?> headers = http1headers.readHeaders(prologue);
+                ctx.remotePeer().tlsPrincipal()
+                        .flatMap(principal -> parseCN(principal.getName()))
+                        .or(() -> Optional.of("Unknown CN"))
```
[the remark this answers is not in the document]
tomas-langer: The same is required for HTTP/2 as well

nima/common/tls/src/main/java/io/helidon/nima/common/tls/TLSUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.security.Principal;
+import java.security.cert.Certificate;
+import java.security.cert.X509Certificate;
+import java.util.Optional;
+
+/**
+ * Utility class for TLS.
+ */
+public class TLSUtils {
```
tomas-langer, line 26: The class name should be `TlsUtils` instead

nima/common/tls/src/main/java/io/helidon/nima/common/tls/TLSUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    private TLSUtils() {
+    }
+
+    /**
+     * Parse the Common Name (CN) from certificates.
+     *
+     * @param certificates certificates
+     * @return  Common Name value
+     */
+    public static Optional<String> parseCN(Certificate[] certificates) {
```
tomas-langer, line 37: The method name should be `parseCn` instead

tomas-langer: Also as this accepts an array, the Javadoc is not valid. It should say it parses CN from the first certificate (if any present) or similar instead

nima/common/tls/src/main/java/io/helidon/nima/common/tls/TLSUtils.java
```diff
@@ -0,0 +1,56 @@
...
+        if (certificates.length >= 1) {
+            Certificate certificate = certificates[0];
+            X509Certificate cert = (X509Certificate) certificate;
+            Principal principal = cert.getSubjectX500Principal();
+
+            int start = principal.getName().indexOf("CN=");
```
tomas-langer, line 43: Similar to my previous comment, this would match `OCN=some-cn` as well, which would be invalid (consider DN `OCN=myName,CN=helidon.io`)

nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2Connection.java
```diff
@@ -624,6 +627,13 @@ private void doHeaders(Semaphore requestSemaphore) {
                                           requestHuffman,
                                           new Http2FrameData(frameHeader, inProgressFrame()));
         }
+        ctx.remotePeer().tlsCertificates()
```
tomas-langer, line 630: The X_HELIDON_CN header should not be created for each request on this connection. The remote peer stays the same for the duration of the connection, so we should create the header just once, and then reuse it. 
Also updating the headers later on is not a good choice (also because you are doing an instanceof), so I think this should be part of creation of the Http2Headers (maybe add `Headers` instance as a parameter and copy all headers from that instance to the new instance being created)

tomas-langer (changes requested):

romain-grecourt (dismissed):

http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
```diff
@@ -135,7 +137,9 @@ public static Http2Headers create(Http2Stream stream,
             stream.priority(priority);
         }
 
-        WritableHeaders<?> headers = WritableHeaders.create();
+        WritableHeaders<?> writableHeaders = headers == null
```
tomas-langer, line 140: a parameter to a public method in Helidon can never be `null`

http/http2/src/test/java/io/helidon/http/http2/Http2HeadersTest.java
```diff
@@ -263,6 +263,7 @@ private Http2Headers headers(String hexEncoded, DynamicTable dynamicTable) {
         return Http2Headers.create(stream,
                                    dynamicTable,
                                    Http2HuffmanDecoder.create(),
+                                   null,
```
tomas-langer, line 266: This should throw a NPE, use a header instance instead

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientStream.java
```diff
@@ -302,6 +302,7 @@ private Http2FrameData readOne() {
                     Http2Headers http2Headers = Http2Headers.create(this,
                                                                     connection.getInboundDynamicTable(),
                                                                     requestHuffman,
+                                                                    null,
```
tomas-langer, line 305: This should not be done, as the API is public.. So either create two methods (one with existing headers, one without them), or do it differently

tvallin: Noted, fixed

tomas-langer (approved):

PR #7346, opened 2023-08-09
4.x Http2Connection should ignore a max concurrent streams setting of zero from the client

thegridman:
The Http2Connection updates some of its configuration based on that sent by the client. One of the config items updated is the "max concurrent streams" setting. The problem is that if the client sends zero for this setting then things can break. I discovered this when a number of our Coherence gRPC tests broke when I switched the server to use Nima and the client is still using grpc-java and Netty.

Client libraries such as `grpc-java` hard code this value to be zero and this breaks gRPC because the server cannot send messages down multiple streams. I have added an example of this where I have an existing bi-di channel request open and then send another request from the client that causes the server to send a message down the bi-di channel. This will break without the fix in this PR.

The fix I have done here is to ignore changing the "max concurrent streams" setting on the server if the client sends zero. then the grpc-java clients work. It might well be that I have mis-understood something about "max concurrent streams" and how the server should handle a value of zero. Trying to find out anything on the internet was all a bit vague.

I added a test `nima/tests/integration/grpc/server/src/test/java/io/helidon/nima/tests/integration/grpc/webserver/EventsTest.java` which works with my fix and fails without it.

### DIFF
--- a/nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2Connection.java
+++ b/nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2Connection.java
@@ -242,26 +242,6 @@ public class Http2Connection implements ServerConnection, InterruptableTask<Void
 
             flowControl.resetMaxFrameSize(maxFrameSize.intValue());
         }
-
-        // Set server MAX_CONCURRENT_STREAMS limit when client sends number lower than hard limit
-        // from configuration. Refuse settings if client sends larger number than is configured.
-        // If the client sends zero the server MAX_CONCURRENT_STREAMS will not be changed.
-        this.clientSettings.presentValue(Http2Setting.MAX_CONCURRENT_STREAMS)
-                .ifPresent(it -> {
-                    if (it != 0) {
-                        if (http2Config.maxConcurrentStreams() >= it) {
-                            maxClientConcurrentStreams = it;
-                        } else {
-                            Http2GoAway frame =
-                                    new Http2GoAway(0,
-                                                    Http2ErrorCode.PROTOCOL,
-                                                    "Value of maximum concurrent streams limit " + it
-                                                            + " exceeded hard limit value "
-                                                            + http2Config.maxConcurrentStreams());
-                            connectionWriter.write(frame.toFrameData(clientSettings, 0, Http2Flag.NoFlags.create()));
-                        }
-                    }
-                });
     }
 
     /**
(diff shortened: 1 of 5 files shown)


danielkec (changes requested):
This looks more like a bug in the mentioned netty based client, citing the [spec](https://httpwg.org/specs/rfc9113.html#SETTINGS_MAX_CONCURRENT_STREAMS):
```
A value of 0 for SETTINGS_MAX_CONCURRENT_STREAMS SHOULD NOT be treated as special by endpoints. A zero value does prevent the creation of new streams; however, this can also happen for any limit that is exhausted with active streams. Servers SHOULD only set a zero value for short durations; if a server does not wish to accept requests, closing the connection is more appropriate.
```
Where `A zero value does prevent the creation of new streams`, we have to honor the 0 value like any other.

tomas-langer:
According to my understanding:
- we should ignore this setting from the client, as we do not support server push, and that is the only case where we create concurrent streams
- the value of `maxClientConcurrentStreams` can never be modified by the client, as this is the number of streams we allow the client to create, this is a big bug
- if we fix this bug, the problem will disappear

tomas-langer (dismissed):

nima/tests/integration/grpc/server/src/main/java/io/helidon/nima/tests/integration/grpc/webserver/EventService.java
```diff
@@ -0,0 +1,125 @@
...
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    private void send(Events.Message message, StreamObserver<Empty> observer) {
+        String text = message.getText();
+        for (Listener listener : listeners) {
+            listener.send(text);
+        }
```
danielkec, line 67: Can `ConcurrentModificationException` happen here? Would using iterator of defensive copy make sense here?

nima/http2/webserver/src/main/java/io/helidon/nima/http2/webserver/Http2Connection.java
```diff
@@ -242,23 +242,6 @@ public void clientSettings(Http2Settings http2Settings) {
...
-                    } else {
-                        Http2GoAway frame =
-                                new Http2GoAway(0,
-                                                Http2ErrorCode.PROTOCOL,
-                                                "Value of maximum concurrent streams limit " + it
-                                                        + " exceeded hard limit value "
-                                                        + http2Config.maxConcurrentStreams());
-                        connectionWriter.write(frame.toFrameData(clientSettings, 0, Http2Flag.NoFlags.create()));
-                    }
-                });
```
danielkec, line 261: :+1:

nima/tests/integration/grpc/server/src/main/java/io/helidon/nima/tests/integration/grpc/webserver/EventService.java
thegridman: I added an iterator here. Also I did the same in another method of that class, and generally tidied it up and added comments to describe what it is doing.

danielkec (approved):
Cool thx for the fix!

PR #7354, opened 2023-08-10
3.x Use text blocks in tests

Captain1653:
There are many lines in the code:
```

String value = "some text\n" +
"other text\n" +
"third text\n";
```

I think it can be replaced on [text blocks](https://openjdk.org/jeps/378).

I used IDE conversion.

### DIFF
--- a/config/config/src/test/java/io/helidon/config/MapConfigSourcePropertiesTest.java
+++ b/config/config/src/test/java/io/helidon/config/MapConfigSourcePropertiesTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2023 Oracle and/or its affiliates.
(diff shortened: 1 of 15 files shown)


tomas-langer (approved):
Hello @Captain1653, thanks for the contribution.
This one is under the "we know we can, the question is if we should". 
I am approving this PR, as you have invested your time into it, but kindly do not create such PRs unless there is an existing issue that was triaged by the team.

We do triage usually once a week to make sure the tasks that we work on are aligned with the Helidon project direction. Right now, as we are nearing 4.0.0 release, unexpected PRs are making our life a bit more complicated.

The best approach would be if you could join our public slack channel (see https://helidon.io/#community), so we could discuss such things in advance.

Thanks once again for the work you do for Helidon!

Captain1653:
@tomas-langer Thank you for the detailed explanation.

PR #7362, opened 2023-08-11
Create a different config setting for inbound and outbound HTTP1 header

klustria:
**Description**
Currently, Helidon 4 only support validate-headers config to validate headers for both inbound and outbound headers on both client and server. This change will bring separation on the configuration for both inbound and outbound headers on both client and server. The change will also remove performance impact for a singular config that would validate both outbound and inbound headers by default. By separating the configuration, only necessary headers will be validated by default. For example, the request headers on the client and the response headers on the server will not be validated by default because the user has control on their creation. On the other hand, the response headers on the client and the request headers on the server will be validated by default because the values they contain are unpredictable upon receipt and may contain malicious data. Following are new config options:
1. Client
   a. validate-response-header (inbound) - Validates the client response headers and will default to true because the content is unpredictable.
   b. validate-request-header (outbound) - Validates the client request headers and will default to false because the user has control on header creation.
2. Server 
    a. validate-request-header (inbound) - Validates the server request headers and will default to true because the content is unpredictable.
    b. validate-response-header (outbound) - Validates the client request headers and will default to false because the user has control on header creation.

**Details of the change:**
1. Webserver
   a. Http1ConfigBlueprint - replace validateHeaders() with validateRequestHeaders() and validateResponseHeaders() to allow separate validation control of request headers and response headers, respectively.
   b. Http1Connection and HttpServerResponse - will process validateRequestHeaders() and validateResponseHeaders() on their corresponding areas of validation.
3. Webclient
    a. Http1ClientProtocolConfigBlueprint = replace validateHeaders() with validateRequestHeaders() and validateResponseHeaders() to allow separate validation control of request headers and response headers, respectively.
    b. Http1CallChainBase, Http1CallOutputStreamChain and Http1CallEntityChain - will process validateRequestHeaders() and validateResponseHeaders() on their corresponding areas of validation.
    c. Http1ClientTest - Header validation related Unit test are adjusted to use validateRequestHeaders() and validateResponseHeaders() instead of their predecessor   validateHeaders()
 5. Integration tests for Webserver - all changed files will perform various scenarios to validate request or response headers on the client side.
 6. Integration tests for Webclient - all changed files will perform various scenarios to validate request or response headers on the client side.

**Documentation**
This is a configuration change so documentation will be collected from javadoc comments in Http1ConfigBlueprint.java for the Webserver and Http1ClientProtocolConfigBlueprint.java for Http1 Webclient.

Examples:
1. Http1 WebServer 
   ``` 
   ServerConnectionSelector http1 = Http1ConnectionSelector.builder()
             .config(Http1Config.builder()
                             .validateRequestHeaders(true) 
                             .validateResponseHeaders(false)
                             .build())
             .build();
   server.addConnectionSelector(http1);
   ```
2. Http1 WebClient 
    ``` 
    Http1Client client = Http1Client.create(clientConfig -> clientConfig.baseUri(baseURI)
              .protocolConfig(it -> {
                          it.validateRequestHeaders(true); 
                          it.validateResponseHeaders(false);
              })
    );
    ```


nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/http1/Http1ServerResponse.java
```diff
@@ -269,8 +274,11 @@ private void handleSinkData(Object data, MediaType mediaType) {
         }
     }
 
-    private static void writeHeaders(Headers headers, BufferData buffer) {
+    private static void writeHeaders(Headers headers, BufferData buffer, boolean validate) {
         for (HeaderValue header : headers) {
```
tomas-langer, line 278: I think this should be done as follows:
```
if (validate) {
   headres.forEach(HeaderValue::validate);
}

for (HeaderValue header : headers) {
  header.writeHttp1Header(buffer);
}
```

klustria: Updated based on comment. After rebase, it will look like this:

```
         if (validate) {
            headers.forEach(Header::validate);
        }
        for (Header header : headers) {
            header.writeHttp1Header(buffer);
        }
```

tomas-langer (approved):

Issue #7276, opened 2023-07-31
4.x: Fix all TLS tests and examples


tomas-langer:
We are currently using self-signed certificates without appropriate extension, so we usually have host name verification switched off.
Also we quite often have trust all configured.

All our test and examples should use a correct TLS setting, that does not require any lowering of security:
- trust store contains the same certificate as the opposite private key (so we trust the self-signed certificate)
- SAN should match `localhost`, so server certificate are correctly trusted by the clients

See tests/integration/webserver/upgrade/src/main/resources/certifiate_notest.md for reference.

Requires #7255 to be merged

PR #7364, opened 2023-08-11
 4.x: Fix all TLS tests and examples #7276 

jbescos:
Relates to https://github.com/helidon-io/helidon/issues/7276


nima/tests/integration/webclient/webclient/pom.xml
```diff
@@ -57,4 +57,20 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <build>
```
tomas-langer, line 60: Please remove this section.

tomas-langer (changes requested):

nima/tests/integration/webclient/webclient/pom.xml
jbescos: Done

tests/integration/webserver/upgrade/src/main/resources/certGen.sh
```diff
@@ -0,0 +1,57 @@
...
+function clean(){
+    FILE=$1
+    echo "=============================> Cleaning $FILE"
+    rm -f ${FILE}
+}
+
+function printKeystore(){
+    KEYSTORE=$1
+    echo "=============================> Print ${KEYSTORE}"
+    keytool -list -keystore ${FOLDER}/${KEYSTORE} -storepass ${PASSWORD} -v
```
romain-grecourt, line 33: Why the switch from `openssl` to `keytool` ? Instructions for openssl are ubiquitous.
Also, we should be consistent and have instructions or a script in each test where we have pkcs#12  files checked-in.

I would personally vote for a `README.md` in each test with dedicated instructions.

tomas-langer: I missed the keytool usage. Agree with @romain-grecourt  on this one. I did not want to have a script, as anything executable is a pain to maintain. Instructions should be enough.

jbescos: @romain-grecourt the reason to have it in keytool is because it exists in JDK, and JDK is needed in Helidon too. Openssl is not installed in Windows as far as I know.

@tomas-langer what is the difference in maintenance a script or a readme?. I guess you had some bad experiences in the past but I am curious to know what could it be.

romain-grecourt: `openssl` is more widely used, more versatile, and is applicable to non Java use-cases.
E.g. If we need to document end to end SSL encryption for users, we should use openssl. In this case having all command line instructions (public docs, and internal dev readmes) based on `openssl` is more consistent.

---

We don't need cross platform instructions (E.g. we use bash everywhere) ; `openssl` is not a build requirement, it is easy to install if needed.

jbescos: Okay, just note that we cannot create truststores in openssl (or that is my understanding), so we will need to do this:
```
        CertificateFactory cf = CertificateFactory.getInstance("X.509");
        X509Certificate cert = (X509Certificate) cf.generateCertificate(Resource.create("certificate.pem").stream());
        Tls clientTls = Tls.builder()
                .trust(trust -> trust.addCert(cert))
                .build();
```

Instead of this:
```
        Tls clientTls = Tls.builder()
                .trust(trust -> trust
                        .keystore(store -> store
                                .passphrase("password")
                                .trustStore(true)
                                .keystore(Resource.create("client.p12"))))
                .build();
```

jbescos: It is done

romain-grecourt: You don't need to use `.pem` ; see `certification_notes.md`.
Also, can you add a README.md for each test that has a checked-in keystores.

jbescos: I added the readmes.

But in the certificate_notes.md I don't see where it is written what the client should use. It cannot use the private key and the server keystore, because that is private information of the server.

jbescos:
I have big conflicts, I will close this PR and create it again.

jbescos:
Closing it because I create a new one:
https://github.com/helidon-io/helidon/pull/7393

PR #7365, opened 2023-08-11
Provider support in builders

### LINKED ISSUE #7295: Configured Builders need to allow for 0..1 providers to be declared
Ideally, we should be able to declare managers to be `Optional<TlsManager>`, or alternatively have a `ConfiguredOption` to say something like "atMostOne=true" - perhaps both so that we could still use `List` and just enforce the cardinality accordingly.  Currently, we need to enforce this through a builder interceptor/decorator which is not ideal.

Example:

```
    /**
     * List of managers (enforced to be at most 1) to manage the lifecycle of the certificate and {@link Tls} creation.
     * 
     * @return managers of the tls
     */
    // must be 0..1 
    @ConfiguredOption(provider = true, providerType = TlsManagerProvider.class, providerDiscoverServices = false)
    List<TlsManager> managers();
```

### LINKED ISSUE #7358: Add a way for the builder mechanism to provide access to the `Config` object passed to a builder
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
It would be very useful if the builder mechanism provided a way to retrieve, from the built `XXXConfig` object, the `Config` (if any) that was passed to the builder's `config(Config)` method.

For an example, suppose we create `MetricsConfigBlueprint` that starts out like this:
```java
@ConfigBean()
@Configured(root = true, prefix = MetricsConfigBlueprint.METRICS_CONFIG_KEY)
@Prototype.Blueprint(decorator = MetricsConfigBlueprint.BuilderDecorator.class)
@Prototype.CustomMethods(MetricsConfigSupport.class)
interface MetricsConfigBlueprint {
```

Tomas suggested these steps as a workaround which almost worked but exposed a small bug:

* Add a method to the blueprint:  
  ```java
  @ConfiguredOption(builderMethod = false, configured = false)
  Config metricsConfig();
  ```
  
* In the builder decorator method (run as `build()` executes):  
  ```java
      class BuilderDecorator implements Prototype.BuilderDecorator<MetricsConfig.BuilderBase<?, ?>> {

        @Override
        public void decorate(MetricsConfig.BuilderBase<?, ?> builder) {
            if (builder.config().isEmpty()) {
                builder.config(GlobalConfig.config().get(METRICS_CONFIG_KEY));
            }
            builder.metricsConfig(target.config().get());
        }
    }
  ```
* Write a custom method for the prototype that refers to the saved config instance:  
  ```java
  class MetricsConfigSupport {

    @Prototype.PrototypeMethod
    static Optional<String> lookupConfig(MetricsConfig metricsConfig, String key) {
        return metricsConfig.metricsConfig()
                .get(key)
                .asString()
                .asOptional();
    }
  ```

Tomas provided this patch that allows this approach to work:
```
Subject: [PATCH] Fix prototype builder for Config type
---
Index: builder/processor/src/main/java/io/helidon/builder/processor/GenerateAbstractBuilder.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/builder/processor/src/main/java/io/helidon/builder/processor/GenerateAbstractBuilder.java b/builder/processor/src/main/java/io/helidon/builder/processor/GenerateAbstractBuilder.java
--- a/builder/processor/src/main/java/io/helidon/builder/processor/GenerateAbstractBuilder.java	(revision af5d14b8a9565af4011713a94a631748b6b87322)
+++ b/builder/processor/src/main/java/io/helidon/builder/processor/GenerateAbstractBuilder.java	(date 1691673740771)
@@ -628,7 +628,7 @@
             pw.println("private Config config;");
         }
         for (PrototypeProperty child : typeContext.propertyData().properties()) {
-            if (!isBuilder || !child.typeHandler().actualType().equals(CONFIG_TYPE)) {
+            if (!isBuilder || !(child.typeHandler().actualType().equals(CONFIG_TYPE) && child.name().equals("config"))) {
                 pw.print(spacing);
                 pw.print(child.fieldDeclaration(isBuilder));
                 pw.println(";");
Index: builder/processor/src/main/java/io/helidon/builder/processor/PrototypeProperty.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/builder/processor/src/main/java/io/helidon/builder/processor/PrototypeProperty.java b/builder/processor/src/main/java/io/helidon/builder/processor/PrototypeProperty.java
--- a/builder/processor/src/main/java/io/helidon/builder/processor/PrototypeProperty.java	(revision af5d14b8a9565af4011713a94a631748b6b87322)
+++ b/builder/processor/src/main/java/io/helidon/builder/processor/PrototypeProperty.java	(date 1691673524635)
@@ -324,7 +324,7 @@
                             .map(typeHandler::toDefaultValue)
                             .orElse(null),
                     configuredAnnotation.getValue("builderMethod").map(Boolean::parseBoolean).orElse(true),
-                    configuredAnnotation.getValue("notConfigured").map(Boolean::parseBoolean).orElse(false),
+                    !configuredAnnotation.getValue("configured").map(Boolean::parseBoolean).orElse(true),
                     provider,
                     provider ? ProviderOption.create(configuredAnnotation) : null);
         }
```

tomas-langer:
Config method support in blueprints

Resolves #7295 
Resolves #7358

### DIFF
--- a/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
@@ -103,7 +103,7 @@ public final class Prototype {
          *                             node
          * @param <S>                  type of the expected service
          * @param <T>                  type of the configured service provider that creates instances of S
-         * @return list of discovered services
+         * @return list of discovered services, ordered by {@link io.helidon.common.Weight} (highest weight is first in the list)
          */
         default <S extends NamedService, T extends ConfiguredProvider<S>> List<S>
         discoverServices(Config config,
@@ -118,7 +118,8 @@ public final class Prototype {
          * Discover service from configuration.
          *
          * @param config               configuration located at the node of the service providers
-         *                             (either a list node, or object, where each child is one service)
+         *                             (either a list node, or object, where each child is one service - this method requires
+         *                             zero to one configured services)
          * @param serviceLoader        helidon service loader for the expected type
          * @param providerType         type of the service provider interface
          * @param configType           type of the configured service
@@ -126,7 +127,8 @@ public final class Prototype {
          *                             node
          * @param <S>                  type of the expected service
          * @param <T>                  type of the configured service provider that creates instances of S
-         * @return list of discovered services
+         * @return the first service (ordered by {@link io.helidon.common.Weight} that is discovered, or empty optional if none
+         *                             is found
          */
         default <S extends NamedService, T extends ConfiguredProvider<S>> Optional<S>
         discoverService(Config config,
--- a/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
@@ -21,6 +21,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.util.List;
+import java.util.Optional;
 
 import io.helidon.common.HelidonServiceLoader;
 import io.helidon.common.config.Config;
@@ -111,6 +112,29 @@ public final class Prototype {
                          Class<S> configType,
                          boolean allFromServiceLoader) {
             return ProvidedUtil.discoverServices(config, serviceLoader, providerType, configType, allFromServiceLoader);
+        }
+
+        /**
+         * Discover service from configuration.
+         *
+         * @param config               configuration located at the node of the service providers
+         *                             (either a list node, or object, where each child is one service)
+         * @param serviceLoader        helidon service loader for the expected type
+         * @param providerType         type of the service provider interface
+         * @param configType           type of the configured service
+         * @param allFromServiceLoader whether all services from service loader should be used, or only the ones with configured
+         *                             node
+         * @param <S>                  type of the expected service
+         * @param <T>                  type of the configured service provider that creates instances of S
+         * @return list of discovered services
+         */
+        default <S extends NamedService, T extends ConfiguredProvider<S>> Optional<S>
+        discoverService(Config config,
+                         HelidonServiceLoader<T> serviceLoader,
+                         Class<T> providerType,
+                         Class<S> configType,
+                         boolean allFromServiceLoader) {
+            return ProvidedUtil.discoverService(config, serviceLoader, providerType, configType, allFromServiceLoader);
         }
     }
 
--- a/builder/api/src/main/java/io/helidon/builder/api/ProvidedUtil.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/ProvidedUtil.java
@@ -63,15 +63,15 @@ final class ProvidedUtil {
      * Discover service from configuration.
      * This method looks for a single provider only.
      *
-     * @param config               configuration located at the node of the service providers
-     *                             (either a list node, or object, where each child is one service)
-     * @param serviceLoader        helidon service loader for the expected type
-     * @param providerType         service provider interface type
-     * @param configType           configured service type
+     * @param config           configuration located at the node of the service providers
+     *                         (either a list node, or object, where each child is one service)
+     * @param serviceLoader    helidon service loader for the expected type
+     * @param providerType     service provider interface type
+     * @param configType       configured service type
      * @param discoverServices whether all services from service loader should be used, or only the ones with configured
-     *                             node
-     * @param <T>                  type of the expected service
-     * @return list of discovered services
+     *                         node
+     * @param <T>              type of the expected service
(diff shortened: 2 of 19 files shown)


builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -112,6 +113,29 @@ public interface ConfiguredBuilder<BUILDER, PROTOTYPE> extends Builder<BUILDER,
                          boolean allFromServiceLoader) {
             return ProvidedUtil.discoverServices(config, serviceLoader, providerType, configType, allFromServiceLoader);
         }
+
+        /**
+         * Discover service from configuration.
+         *
+         * @param config               configuration located at the node of the service providers
+         *                             (either a list node, or object, where each child is one service)
+         * @param serviceLoader        helidon service loader for the expected type
```
trentjeff, line 122: does the service loader really need to be passed, or can it just be created directly?  When will we ever have a special loader?

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -112,6 +113,29 @@ public interface ConfiguredBuilder<BUILDER, PROTOTYPE> extends Builder<BUILDER,
...
+         * @param providerType         type of the service provider interface
+         * @param configType           type of the configured service
+         * @param allFromServiceLoader whether all services from service loader should be used, or only the ones with configured
```
trentjeff, line 125: Also mention somewhere how ranking/weight will be used.

builder/api/src/main/java/io/helidon/builder/api/ProvidedUtil.java
```diff
@@ -58,6 +59,60 @@ final class ProvidedUtil {
     private ProvidedUtil() {
     }
 
+    /**
+     * Discover service from configuration.
+     * This method looks for a single provider only.
```
trentjeff, line 64: mention weight

builder/processor/src/main/java/io/helidon/builder/processor/GenerateAbstractBuilder.java
```diff
@@ -314,8 +314,7 @@ public BUILDER config(Config config) {
             pw.println("@Override");
             pw.print(SOURCE_SPACING);
             pw.print(SOURCE_SPACING);
-            pw.print("public BUILDER");
-            pw.println(" config(Config config) {");
+            pw.println("public BUILDER config(Config config) {");
```
trentjeff, line 317: @Verdent fyi for your classmodel PR

builder/tests/builder/src/main/java/io/helidon/builder/test/testsubjects/WithProviderBlueprint.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.helidon.builder.api.Prototype;
+import io.helidon.config.metadata.Configured;
+import io.helidon.config.metadata.ConfiguredOption;
+
+@Prototype.Blueprint
+@Configured
+interface WithProviderBlueprint {
+    @ConfiguredOption(provider = true, providerType = SomeProvider.class)
+    SomeProvider.SomeService oneDiscover();
```
trentjeff, line 30: for completeness, it might be good to have `SomeNonExistantService` and repeat the same decls for to ensure they behave as expected,

config/metadata/src/main/java/io/helidon/config/metadata/ConfiguredOption.java
```diff
@@ -124,6 +124,11 @@
      * The provider must then be configured to {@link Configured#provides()} this type.
      * In addition, the {@link #providerType()} must be set to the correct service provider interface,
      * so code can be correctly generated.
+     * <p>
+     * If the method returns a list, the provider configuration must be under config key {@code providers} under
+     * the configured option. On the same level as {@code providers}, there can be {@code discover-services} boolean
+     * defining whether to look for services from service loader even if not configured in the configuration (this would
+     * override {@link #providerDiscoverServices()} defined on this annotation.
```
trentjeff, line 131: might be worth showing an example snipped of `config.yaml` as a <pre>

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
tomas-langer: You cannot create it in the other module, due to module system. The module that declares `uses` is the one that must create the `ServiceLoader` instance (or we would have to do some complex stuff using reflection, which we do not do by design).

trentjeff (approved):

PR #7384, opened 2023-08-15
4.x Support creating routes from standard gRPC  bindable services

thegridman:
The gRPC API needs to support being able to create routes from standard gRPC service implementations that have been written by extending protobuf generated services. This PR make it possible to create a set of routes by passing either a `BinadbleService` or `ServerServiceDefinition` to the `GrpcRouting` class.

These changes were required to get our Coherence gRPC client tests to pass when we build the Coherence server side on top of Nima. Our gRPC services have been written in the normal way by extending protobuf generated service classes. It would have been painful to have to register each service method individually with the `GrpcRouting` class.


tomas-langer (approved):
LGTM

Issue #7381, opened 2023-08-15
4.x: DOC for Helidon neutral metrics API


tjquinno:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------
## Problem Description
Related to #4868 

Documentation for the new Helidon metrics API needs to be written and it needs to replace the previous SE metrics doc. PR #7261 contains some information but the doc impact is much too large to capture in the PR alone.

(There will be some impact on the MP metrics doc as well but that will be handled as part of a separate issue when we convert our MP metrics implementation to use the new Helidon API.)

Tim will work with Lisa to create this doc.

Issue #4868, opened 2022-09-12
Metrics API for Helidon that is server agnostic and does not use MicroProfile metrics


tomas-langer:
We need to remove dependency on MP Metrics.
This requires us to introduce a metrics registry and metrics API (interfaces, annotations) to be used by Helidon.
We should also add support to add custom metrics (similar to health checks).

In addition we should add implementation for Níma that exposes Prometheus style endpoint

tomas-langer:
The current implementation in `main` uses MP Metrics, and has an integration for both reactive and Níma (and MP uses the Níma integration)

PR #7388, opened 2023-08-16
4.x: Neutral metrics API - revise `MetricsFeature` and rework Prometheus and JSON formatting

tjquinno:
### Description
Continuing work on #4868 
This PR works forward from #7261

These changes update some of the basic classes (not drastically) and start the changes to the `MetricsFeature` to use the new API.

## Internal changes in how we format `/metrics` output

To help keep dependencies cleaner, this PR rearranges how we format metrics output for the `/metrics` endpoint. The `MetricsFeature` class, primarily, needs to be able to format the output in arbitrary ways, selected by the media type. To do this without building hard dependencies from the feature module to Micrometer and the Micrometer Prometheus code (and potentially any other formatting approach), this PR introduces a service loader approach for formatting the output.

The feature code now service-loads all `MeterRegistryFormatterProvider` implementations (the PR includes Micrometer Prometheus and JSON) and then invokes the `formatter(MediaType)` method on each. If the returned `Optional<MeterRegistryFormatter>` is present then that provider handles that media type and the feature code uses that formatter. Otherwise the feature keeps looking for a provider that does handle the media type. If the feature finds none, then it returns a "406 - not acceptable" status.

### Documentation
See #7381


tjquinno:
Closing - created against the wrong base branch.

Issue #4238, opened 2022-05-18
Add Support to OCI Vault as a Config Source


klustria:
## Environment Details
* Helidon Version: 2.50
* Helidon MP

----------

## Problem Description
Find a suitable solution on how to use OCI Vault/Secrets as a config source

trentjeff:
Relates to https://github.com/helidon-io/helidon/issues/3649
and https://github.com/helidon-io/helidon/issues/4191

We also need to determine whether we need refresh value, so that it could be cached or refreshed on a time interval.

trentjeff:
Fwiw, I suggest we do this on 3.x and forward port it to 4.x

ljnelson:
Fully addressed in #7520 and #7555. Closing.

PR #7391, opened 2023-08-16
4.x: Adds OciSecretsConfigSourceProvider.java

ljnelson:
Addresses issue #4238.

Description: `OciSecretsConfigSourceProvider` is a Helidon SE `ConfigSourceProvider` that creates a `ConfigSource` that sources its information from an OCI Vault using the OCI [Secrets Retrieval](https://docs.oracle.com/en-us/iaas/tools/java/3.23.0/com/oracle/bmc/secrets/Secrets.html) and [Vaults](https://docs.oracle.com/en-us/iaas/tools/java/3.23.0/com/oracle/bmc/vault/Vaults.html) APIs.  The `ConfigSource` so created is, as I've been directed, a `NodeConfigSource` and a `PollableSource`.  Its "have I been modified?" logic is based on metadata also sourced from the vault.

This PR supersedes other PRs that I've made in this general area as it now follows the most recent instructions and directions in this area that I've received.  It should be easily back-portable to the `helidon-3.x` branch.  If I understand properly, it should be usable as-is by Helidon's MicroProfile Config implementation via already-present Helidon mechanisms for including `ConfigSourceProvider` implementations, though I'm not personally familiar yet with how to do this.

I will be closing and re-pointing other PRs here.

Once this PR is approved (if), I will backport it to `helidon-3.x`.

Documentation impact: There will need to be a couple of examples showing how to configure _this_ `ConfigSourceProvider` in a `meta-config.yaml`. (Examples already exist showing how to configure other `ConfigSourceProvider`s; this one is not very different.)  Javadoc already exists in this PR to show usage and a sample configuration.


integrations/oci/oci-secrets-config-source/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+    <dependencyManagement>
+        <dependencies>
+            <!--
+                OCI brings in two different versions of httpclient and httpcore (!); here we tell it to use the most
+                recent of the versions of each that it brings in.
+            -->
+            <dependency>
+                <groupId>org.apache.httpcomponents</groupId>
+                <artifactId>httpclient</artifactId>
+                <version>4.5.14</version>
```
trentjeff, line 48: versions should be externalized to above properties or in dependencies/pom.xml - deferring to @barchetta

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+ * <li>Ensure the {@code meta-config.yaml} classpath resource contains a {@code source} with a {@code type} of {@code
+ * oci-secrets} that looks similar to the following, substituting values as appropriate:<blockquote><pre>sources:
+ *  - type: 'oci-secrets'
+ *    properties:
+ *      compartment-ocid: 'your vault compartment OCID goes here'
+ *      vault-ocid: 'your vault OCID goes here'</pre></blockquote></li>
+ *
+ * </ol>
+ *
+ * <p>Refer to Helidon's documentation concerning meta-configuration for more details.</p>
```
trentjeff, line 94: This should also say something about `oci.yaml` and point to that runtime module.

Can we also references this class and that module in the README.md as well please.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+    /*
+     * Static fields.
+     */
+
+
+    private static final Logger LOGGER = System.getLogger(OciSecretsConfigSourceProvider.class.getName());
+
+    private static final Set<String> SUPPORTED_TYPES = Set.of("oci-secrets");
+
+    private static final double WEIGHT = 300D;
```
trentjeff, line 110: what is the significance for choosing 300D?  A short inline comment would be nice here.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+                // (It is not immediately clear why the OCI Java SDK requires a Compartment OCID, since a Vault OCID is
+                // sufficient to uniquely identify any Vault.)
+                this.loader = this::absentNodeContent;
+            } else {
+                ListSecretsRequest listSecretsRequest = ListSecretsRequest.builder()
+                    .compartmentId(compartmentOcid)
+                    .vaultId(vaultOcid)
+                    .build();
+                this.loader = () -> {
+                    Collection<? extends SecretSummary> secretSummaries = secretSummaries(vaultsSupplier, listSecretsRequest);
```
trentjeff, line 268: Can you please make this a private method - seems a bit long to inline like this.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+            return new Builder();
+        }
+
+        private static void completeAllSecretsTasks(Collection<? extends Callable<Void>> tasks, AutoCloseable secrets) {
+            RuntimeException re = null;
+            try (ExecutorService es = newVirtualThreadPerTaskExecutor()) {
+                for (Future<?> future : es.invokeAll(tasks)) {
+                    try {
+                        future.get();
+                    } catch (RuntimeException e) {
```
trentjeff, line 348: considering the duplication here at a few levels, wouldn't it be easier to just write a `handleException(AtomicReference<RuntimeException> re, Exception e)` method and just always delegate to it at everywhere to handle a plain old catch (Exception e)?

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+                }
+                throw new IllegalStateException(e.getMessage(), e);
+            }
+        }
+
+        @SuppressWarnings("checkstyle:linelength")
+        private static ValueNode valueNode(Secrets s, SecretSummary ss, Base64.Decoder d) {
+            return
+                ValueNode.create(new String(d.decode(((Base64SecretBundleContentDetails) (s.getSecretBundle(GetSecretBundleRequest.builder()
+                                                                                                                .secretId(ss.getId())
```
trentjeff, line 416: this is really hard to read due to spacing and compounding of actions - please consider breaking it up and realigning it.

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/UsageTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+final class UsageTest {
```
trentjeff, line 32: This is really an integration test.  We need other pure unit tests that will cover other code here as part of the normal pipeline build.

trentjeff (changes requested):

integrations/oci/oci-secrets-config-source/pom.xml
```diff
@@ -0,0 +1,143 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.helidon.integrations.oci</groupId>
+        <artifactId>helidon-integrations-oci-project</artifactId>
+        <version>4.0.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>helidon-integrations-oci-secrets-config-source</artifactId>
+    <name>Helidon Config OCI Secrets ConfigSourceProvider</name>
```
tomas-langer, line 27: The name should also follow naming convention:
"Helidon Integrations OCI Secrets Config Source"

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+     * metaConfig} is {@code null}
+     *
+     * @see #supported()
+     *
+     * @deprecated For use by the Helidon Config subsystem only.
+     */
+    @Deprecated // For use by the Helidon Config subsystem only.
+    @Override // ConfigSourceProvider
+    public ConfigSource create(String type, Config metaConfig) {
+        return this.supports(type) ? SecretBundleConfigSource.builder().config(metaConfig).build() : ConfigSources.empty();
```
tomas-langer, line 157: This method is only called for a supported type, you do not need to check it, and you do not need to return an empty config source.
The `type` support is built in for providers that have more than one supported type, so they know which one to build.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+     * @see #supported()
+     *
+     * @see #create(String, Config)
+     *
+     * @deprecated For use by the Helidon Config subsystem only.
+     */
+    @Deprecated // For use by the Helidon Config subsystem only.
+    @Override // ConfigSourceProvider
+    public boolean supports(String type) {
+        return type != null && this.supported().contains(type);
```
tomas-langer, line 196: type cannot `null`, as we follow our own guidelines
we do not want to be be null safe - the exact opposite, we expect any component to fail with a NPE if `null` is sent over public API.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+            /*
+             * Instance methods.
+             */
+
+
+            @Override // AbstractConfigSourceBuilder<Builder, Void>
+            protected Builder config(Config metaConfig) {
+                metaConfig.get("compartment-ocid")
+                    .asString()
+                    .flatMap(s -> s.isBlank() ? Optional.empty() : Optional.of(s))
```
tomas-langer, line 473: this can be replaced with `.filter(Predicate.not(String::isBlank))`

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+                for (Future<?> future : es.invokeAll(tasks)) {
+                    try {
+                        future.get();
+                    } catch (RuntimeException e) {
+                        if (re == null) {
+                            re = e;
+                        } else {
+                            re.addSuppressed(e);
+                        }
+                    } catch (Exception e) {
```
tomas-langer, line 354: please replace with `catch (InterrupteedExceptione)`
the block cannot throw any other checked exception, and you catch all runtime exceptions in the previous block

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+                if (e instanceof InterruptedException) {
+                    Thread.currentThread().interrupt();
+                }
+                re = new IllegalStateException(e.getMessage(), e);
+            } finally {
+                try {
+                    secrets.close();
+                } catch (RuntimeException e) {
+                    if (re == null) {
+                        throw e;
```
tomas-langer, line 377: you should never throw intentionally from a finally block. 
This whole construct is a bit strange, but it could be changed by assigning to `re` or suppressing, and the final `if (re != null) thorw re` could be done after the `finally` block is finished.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+                        re.addSuppressed(e);
+                    }
+                }
+                if (re != null) {
+                    throw re;
+                }
+            }
+        }
+
+        @SuppressWarnings("try")
```
tomas-langer, line 397: Why is this suppress warning present?

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+        private static Collection<? extends SecretSummary> secretSummaries(Supplier<? extends Vaults> vaultsSupplier,
+                                                                           ListSecretsRequest listSecretsRequest) {
+            try (Vaults v = vaultsSupplier.get()) {
+                return v.listSecrets(listSecretsRequest).getItems();
+            } catch (RuntimeException e) {
+                throw e;
+            } catch (Exception e) {
```
tomas-langer, line 404: Please replace all of these instances with catching the `InterruptedExceptinon` and handling it separately.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+                if (e instanceof InterruptedException) {
+                    Thread.currentThread().interrupt();
+                }
+                throw new IllegalStateException(e.getMessage(), e);
+            }
+        }
+
+        @SuppressWarnings("checkstyle:linelength")
```
tomas-langer, line 412: This should never be suppreessed.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
```diff
@@ -0,0 +1,526 @@
...
+ */
+public final class OciSecretsConfigSourceProvider implements ConfigSourceProvider, Weighted {
+
+
+    /*
+     * Static fields.
+     */
+
+
+    private static final Logger LOGGER = System.getLogger(OciSecretsConfigSourceProvider.class.getName());
```
tomas-langer, line 106: Field is unused

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/UsageTest.java
tomas-langer: Please do not declare as `final`

tomas-langer (changes requested):

integrations/oci/oci-secrets-config-source/pom.xml
ljnelson: Joe and I came up with this together. If you remove these, dependency convergence fails. What solution do the two of you propose? I am happy to follow directions.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Romain told me not to refer to Helidon documentation, only javadoc. I'll make sure to reference the `OciExtension` explicitly here.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: I followed examples of other `ConfigSourceProvider`s. To be frank I'm not sure if implementing `Weighted` is even needed. I am happy to follow consensus opinion. Should I remove it?

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: It closes over various variables. Do you want me to turn them into (less efficient, unneeded in some cases) instance variables?

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: I will look at exception handling again. In some cases I need to handle the fact that a client is `AutoCloseable`, but does not override `close()`, so it throws `Exception`, and in some of those cases there are multiple exceptions that can be thrown from within the try/finally block, and I want them all captured. In those cases, frequently I need to suppress other exceptions. It's complicated, and I agree that if there is a way to reduce the complexity I should take it.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: I will change this method to do no checking as requested.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: I will eliminate the `null` check.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Doh! I keep getting burned by the fact the `Config#asString()` method documentation says it returns an `Optional`. I will make this change.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Good catch; I missed that `ExecutorService` overrides `close()` to not throw `Exception`.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Unless I put it in the wrong place, it is because OCI clients are `AutoCloseable` and do not override `close()`. This means that `close()` can throw `Exception`. Warnings are emitted from the compiler in such a case that something in a try-with-resources block can throw an `InterruptedException`. They are annoying and I am handling the `InterruptedException` case so I felt they were appropriate. What is your suggested action?

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: OK. There will be duplicate code of course, since `Vaults#close()` throws `Exception` but I will make this change.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: OK; I'll remove this.

integrations/oci/oci-secrets-config-source/pom.xml
ljnelson: Double-checking to make sure: the only class in this Maven module is, in fact, a `ConfigSourceProvider`. I'll make the requested change unless I hear otherwise.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Oh sorry, now I see what you are saying. Will change.

ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/pom.xml
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/UsageTest.java
trentjeff: testing is good, so I would leave but rename it to IntegrationTest perhaps.  Mainly I want to see unit test coverage of something when oci is not in the environment.  Currently nothing will actually be unit tested.

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Done

integrations/oci/oci-secrets-config-source/src/main/java/io/helidon/integrations/oci/secrets/configsource/OciSecretsConfigSourceProvider.java
ljnelson: Moved to `@Weight(300D)`; documented Tomas' rationale in a comment

integrations/oci/oci-secrets-config-source/pom.xml
ljnelson: Replicated hack from `integrations/oci/cdi`. Done.

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/UsageTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+class UsageTest {
+
+    private UsageTest() {
+        super();
+    }
+
+    @Test
+    void testUsage() {
```
trentjeff, line 39: i'm still not seeing the name change or other pure unit test coverage here.

ljnelson: Could you give me an idea of what you would like to see tested? And what the name should be?

trentjeff: There is some 400+ lines in `OciSecretsConfigSourceProvider` that should be tested since as it is right now there is nothing currently test covered in the normal case where the pipeline or developer typically does not have an $HOME/.oci/config present in their environment.  I am suggesting we have some unit-test coverage on this code for the typical case.  Perhaps even consider making more methods package private to be able to test them in isolation, eg. ` static boolean isModified(Instant pollInstant, Instant closestSecretExpirationInstant)`, `static void completeAllSecretsTasks(Collection<? extends Callable<Void>> tasks, AutoCloseable secrets)`, ` Optional<NodeContent> load(Supplier<? extends Vaults> vaultsSupplier,
                                           Supplier<? extends Secrets> secretsSupplier,
                                           ListSecretsRequest listSecretsRequest)` with some fakes or mocks, etc.

trentjeff (approved):

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/IsModifiedTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static io.helidon.integrations.oci.secrets.configsource.OciSecretsConfigSourceProvider.SecretBundleConfigSource.isModified;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+class IsModifiedTest {
+
+    private IsModifiedTest() {
```
tomas-langer, line 28: please remove the constructors from test classes. There is no benefit in declaring them, and also declaring them private is not great here. Out of curiosity - why do you declare them at all?

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/UsageTest.java
```diff
@@ -0,0 +1,83 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assumptions.assumeFalse;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+class UsageTest {
+
+    private UsageTest() {
```
tomas-langer, line 34: As above - the constructor does not need to be explicit for test classes.

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/ValueNodeTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.Test;
+
+import static io.helidon.integrations.oci.secrets.configsource.OciSecretsConfigSourceProvider.SecretBundleConfigSource.valueNode;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+class ValueNodeTest {
+
+    private ValueNodeTest() {
```
tomas-langer, line 29: dtto.

tomas-langer (changes requested):

integrations/oci/oci-secrets-config-source/src/test/java/io/helidon/integrations/oci/secrets/configsource/IsModifiedTest.java
ljnelson: Habit, mostly. Obviously in a regular class if you declare no constructors, the implicit `public` zero-argument one is effectively declared for you. It hasn't been until recent JDKs that this often surprising state of affairs is warned about. Anyway, I'll remove them since these classes aren't `public` anyway.

tomas-langer (approved):

PR #7396, opened 2023-08-17
4.x: Media Context and streaming

### LINKED ISSUE #7006: Media support (such as Path) in Níma HTTP1 buffer into memory
Media support in Níma HTTP1 currently does not work properly. 
[HttpCallEntityChain](https://github.com/helidon-io/helidon/blob/main/nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/http1/HttpCallEntityChain.java#L107) buffers everything into the memory, instead of writing into the request OS. It should rather use something similar to [ClientConnectionOutputStream](https://github.com/helidon-io/helidon/blob/e3334d9c74d08f01a19b0ba92933131f6c54aa83/nima/webclient/webclient/src/main/java/io/helidon/nima/webclient/http1/HttpCallOutputStreamChain.java#L102) which performs sent entity analysis to determine proper `transfer-encoding` and `content-length`.

Once this is changed, it will break our current redirect support for entities introduced in https://github.com/helidon-io/helidon/issues/6804. Once the following issue is fixed https://github.com/helidon-io/helidon/issues/6992, it should introduce fix usable also by this issue.

Tomas Langer: 
For example when you call `ServerResponse.send(Paths.get("/my/file.txt"))` the file will be fully buffered into memory. 

Possible solutions (there may be more):
- have a `boolean streaming()` on media supports to tell the server and client that this really must be streamed and not cached in memory
- have some advanced logic to buffer a certain (max) amount in memory for any provider, and if more bytes would be written, start sending them over the network

In both cases we need to refactor the method `send(Object)` on server response (and appropriate one on client request)

tomas-langer:
### Description
Media writers now correctly handle streaming where needed, without buffering in memory.
Fixes of issues uncovered by this change.
This impacts custom Media supports only, the API remains the same.

- String support is always in-memory
- Path depends on file size (configurable) - smaller files are loaded into memory, bigger files use streaming

Fixed issues that were uncovered (as now we use streaming a bit more).

Resolves #7006 

Moved to M2 milestone, as this resolves some bigger issues (such as 100-Continue handling)

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -92,6 +92,38 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
         bufferData.write(Bytes.LF_BYTE);
     }
 
+    static WebClientServiceResponse createServiceResponse(HttpClientConfig clientConfig,
+                                                          WebClientServiceRequest serviceRequest,
+                                                          ClientConnection connection,
+                                                          DataReader reader,
+                                                          Http.Status responseStatus,
+                                                          ClientResponseHeaders responseHeaders,
+                                                          CompletableFuture<WebClientServiceResponse> whenComplete) {
+        WebClientServiceResponse.Builder builder = WebClientServiceResponse.builder();
+        AtomicReference<WebClientServiceResponse> response = new AtomicReference<>();
+
+        if (mayHaveEntity(responseStatus, responseHeaders)) {
+            // this may be an entity (if content length is set to zero, we know there is no entity)
+            builder.inputStream(inputStream(clientConfig,
+                                            connection.helidonSocket(),
+                                            response,
+                                            responseHeaders,
+                                            reader,
+                                            whenComplete));
+        }
+
+        WebClientServiceResponse serviceResponse = builder
+                .connection(connection)
+                .headers(responseHeaders)
+                .status(responseStatus)
+                .whenComplete(whenComplete)
+                .serviceRequest(serviceRequest)
+                .build();
+
+        response.set(serviceResponse);
+        return serviceResponse;
+    }
+
     @Override
     public WebClientServiceResponse proceed(WebClientServiceRequest serviceRequest) {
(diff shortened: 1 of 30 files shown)


danielkec (approved):

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -181,38 +213,21 @@ protected WebClientServiceResponse readResponse(WebClientServiceRequest serviceR
...
+                                     responseStatus,
+                                     responseHeaders,
+                                     whenComplete);
     }
 
-    private InputStream inputStream(HelidonSocket helidonSocket,
-                                    AtomicReference<WebClientServiceResponse> response,
-                                    ClientResponseHeaders responseHeaders,
-                                    DataReader reader) {
+    private static InputStream inputStream(HttpClientConfig clientConfig,
```
Verdent, line 225: Since you already moved `createServiceResponse` because of it became static, I think you should move this as well.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -240,7 +255,7 @@ private InputStream inputStream(HelidonSocket helidonSocket,
         }
     }
 
-    private boolean mayHaveEntity(Http.Status responseStatus, ClientResponseHeaders responseHeaders) {
+    private static boolean mayHaveEntity(Http.Status responseStatus, ClientResponseHeaders responseHeaders) {
```
Verdent, line 258: The same as above.

Verdent (changes requested):

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
Verdent: Not an issue after all.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
Verdent: Not an issue after all.

Verdent (approved):

PR #7400, opened 2023-08-17
4.x: Allowed values are now checked when validating a builder.

tomas-langer:
### Description
When using `@ConfiguredOption(allowedValues=...)` the generated builders now validate (during build) that the option is within the allowed values.

- for an option, if not `null`, a check is made of its `String.valueOf()` against the String allowed value set
- for a list of options, each value in the list is check (if any)

This change required modifying `Annotation` to support typed values and nested annotations, this has been done as well.

### Documentation
This enhancement enforces an existing API in `ConfiguredOption`.
Annotation now supports nested annotations and typed values. Methods to obtain a typed value are now added to the `Annotation` type (in `helidon-common-types`)

Resolves #6403 
Resolves #6231

PR #7402, opened 2023-08-18
4.x: Supplier in builder

### LINKED ISSUE #7323: Support Supplier in Config Builders
Passwords in the Vault may change over time.  And features that require lively updates for configuration-based artifacts (e.g., Tls w/ cert rotation in #7279), will need the ability to get the updated password value when the certificate is observed to have changed.

We therefore would like to support something like:
```java
    Supplier<char[]> keyPassword();
```

tomas-langer:
### Description
A blueprint can now define a `Supplier<T>`. If this type is `@Configured` and the method has `@ConfiguredOption`, the supplier will be used directly from `Config` (if used).
This means the supplier will provide the latest value if changes are enabled in config.

Added appropriate test with a mutable configuration.

Resolves #7323
May require rebase after #7400 is merged

### Documentation
This enhancement adds support for a typed `Supplier` in prototype blueprints (Helidon builder). 
The following blueprint:
```java
@Prototype.Blueprint
@Configured
interface SupplierBeanBlueprint {
    @ConfiguredOption
    Supplier<String> stringSupplier();
}
```

will work with `io.helidon.config.Config` that has support for changes (using polling, file watcher, or an event based config source) in the following way:
- if the `SupplierBean` is created with config (such as `SupplierBean.create(Config)` or `SupplierBean.Builder.config(Config)` is used), the returning supplier will always provide the latest known value from the configuration key `string-supplier`

Of course the supplier can also be set explicitly (`SupplierBean.Builder.stringSupplier(() -> "someValue)`) - in this case, the supplier that is provided by user will be used

### DIFF
--- a/builder/processor/src/main/java/io/helidon/builder/processor/TypeHandlerSupplier.java
+++ b/builder/processor/src/main/java/io/helidon/builder/processor/TypeHandlerSupplier.java
@@ -65,12 +65,30 @@ class TypeHandlerSupplier extends TypeHandler.OneTypeHandler {
             method.addLine(configGet(configured)
                                    + generateFromConfig(factoryMethods)
                                    + ".ifPresent(this::" + setterName() + ");");
+        } else if (actualType().isOptional()) {
+            method.add(setterName() + "(");
+            method.add(configGet(configured));
+            method.add(generateFromConfigOptional(factoryMethods));
+            method.addLine(".optionalSupplier());");
         } else {
             method.add(setterName() + "(");
             method.add(configGet(configured));
             method.add(generateFromConfig(factoryMethods));
             method.addLine(".supplier());");
         }
+    }
+
+    String generateFromConfigOptional(FactoryMethods factoryMethods) {
+        TypeName optionalType = actualType().typeArguments().get(0);
+        if (optionalType.fqName().equals("char[]")) {
+            return ".asString().as(String::toCharArray)";
+        }
+
+        TypeName boxed = optionalType.boxed();
+        return factoryMethods.createFromConfig()
+                .map(it -> ".map(" + it.typeWithFactoryMethod().genericTypeName().fqName() + "::" + it.createMethodName() + ")")
+                .orElseGet(() -> ".as(" + boxed.fqName() + ".class)");
+
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


builder/processor/src/main/java/io/helidon/builder/processor/TypeHandlerSupplier.java
```diff
@@ -0,0 +1,190 @@
...
+                .returnType(returnType, "updated builder instance")
+                .description(blueprintJavadoc.content())
+                .addJavadocTag("see", "#" + getterName() + "()")
+                .addParameter(param -> param.name(name())
+                        .type(argumentTypeName())
+                        .description(blueprintJavadoc.returnDescription()))
+                .accessModifier(setterAccessModifier(configured));
+        builder.typeName(Objects.class).addLine(".requireNonNull(" + name() + ");");
+        builder.addLine("this." + name() + " = " + name() + "::get;")
+                .addLine("return self();");
```
Verdent, line 187: Since these are changes on the same method builder, why not to just continue with the method chain? Also `.typeName(Objects.class).addLine(".requireNonNull(" + name() + ");")` is in above method body creations on the new line and here it is on the same. I would say it should be on the new line since it is chained method call.

Verdent (changes requested):

common/config/src/main/java/io/helidon/common/config/ConfigValue.java
```diff
@@ -284,4 +284,41 @@ default Stream<T> stream() {
...
+     * Returns a {@link Supplier} of an {@link Optional Optional&lt;T&gt;} of the configuration node.
+     * <p>
+     * Supplier returns a {@link Optional#empty() empty} if the node does not have a direct value.
+     *
+     * @return a supplier of the value as an {@link Optional} typed instance, {@link Optional#empty() empty} in case the node
+     * does not have a direct value
+     * @see #asOptional()
+     * @see #supplier()
+     */
+    Supplier<Optional<T>> optionalSupplier();
```
trentjeff, line 323: nice!

common/config/src/main/java/io/helidon/common/config/EmptyConfig.java
```diff
@@ -139,6 +140,21 @@ public T get() throws ConfigException {
...
+        }
+
+        @Override
+        public Supplier<T> supplier(T defaultValue) {
+            return () -> defaultValue;
+        }
+
+        @Override
+        public Supplier<Optional<T>> optionalSupplier() {
+            return this::asOptional;
```
trentjeff, line 156: i didn't see a test on this with nulls, etc.

trentjeff (changes requested):

common/config/src/main/java/io/helidon/common/config/ConfigValue.java
```diff
@@ -284,4 +284,41 @@ default Stream<T> stream() {
         return asOptional().stream();
     }
 
+    /**
+     * Returns a supplier of a typed value. The semantics depend on implementation, this may always return the
+     * same value, or it may provide latest value if changes are enabled.
```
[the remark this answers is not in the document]
tomas-langer: There is a clear statement "the semantics depend on implementation" - I cannot link to a type that is an implementation of this interface.

common/config/src/main/java/io/helidon/common/config/EmptyConfig.java
tomas-langer: Added a test and fixed the processor to correctly generate the config method

builder/processor/src/main/java/io/helidon/builder/processor/TypeHandlerSupplier.java
tomas-langer: Fixed

Verdent (approved):

trentjeff (approved):

builder/tests/builder/src/test/java/io/helidon/builder/test/SupplierTest.java
```diff
@@ -0,0 +1,146 @@
...
+        Config config = Config.just(testSource);
+        SupplierBean b = SupplierBean.create(config);
+
+        Supplier<Optional<String>> optionalSupplier = b.optionalSupplier();
+        Supplier<Optional<char[]>> optionalCharSupplier = b.optionalCharSupplier();
+
+        assertThat(optionalSupplier.get(), optionalValue(is(ORIGINAL_VALUE)));
+        assertThat(optionalCharSupplier.get(), optionalValue(is(ORIGINAL_VALUE_CHARS)));
+
+        testSource.update(null);
```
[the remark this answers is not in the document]
tomas-langer: method withing a test config source used only from the test itself

PR #7404, opened 2023-08-20
Change heap functions to get current memory usage

### LINKED ISSUE #7374: JVM Heap metrics never change
JVM Heap Metrics observed through the `/metrics` endpoint never vary during the lifetime of the JVM.

## Environment Details
* Helidon Version: 3.1.2
* Helidon MP
* JDK version: 17
* OS: Linux

----------

## Problem Description

The implementation of metrics in #3847 in [BaseRegistry](https://github.com/helidon-io/helidon/blob/helidon-3.x/metrics/metrics/src/main/java/io/helidon/metrics/BaseRegistry.java#L224) refactored the registration of `MEMORY_USED_HEAP` and other heap related metrics. The refactoring passes the result of calling `memoryBean.getHeapMemoryUsage()` as the object to be polled for metrics using the `MemoryUsage::getUsed` function reference.

The javadoc for the `getHeapMemoryUsage` method simply states, "Returns the current memory usage of the heap that is used for object allocation." I believe that `getHeapMemoryUsage()` is a point in time view of the metric based on the usage of the word "current". As a result, the metrics registry never sees a different value than that observed at the time of the metric registry construction. 

I have observed this in my own services and it is captured by Prometheus. The used and committed heap values never change for the life of a pod.

## Steps to reproduce

Start any metrics example and observe the values for the JVM heap memory. They will never fluctuate for the life of the JVM regardless of load or actual heap usage.

edthorne:
### Description

Resolves #7374 by providing inline function to get current heap memory metrics rather than a static snapshot of the heap memory metrics.

### Documentation

None

### DIFF
--- a/metrics/metrics/src/main/java/io/helidon/metrics/BaseRegistry.java
+++ b/metrics/metrics/src/main/java/io/helidon/metrics/BaseRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


tjquinno (approved):
@edthorne Thanks for the issue and the PR. (Although there is no test with the PR, I assume you have verified that your change allows the reported values to actually change.)

Maybe you already noticed, but this problem does not occur in 2.x. And I am actively working on the metrics code in 4.x and will incorporate this fix into those changes. So there's no need to create other issues for this problem in other releases.

PR #7426, opened 2023-08-23
Update Hibernate version to 6.2.7

### LINKED ISSUE #6561: Update to Jandex 3
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
We might want to adopt the latest Jandex release.

The Jandex 3 library (as of this writing, 3.0.5 is current) is now available. Further, there is a new release of the Jandex plug-in as well, also with a different groupID. 

The effect of the library change on Helidon would be in these general areas:

1. It has some incompatible API changes as compared to Jandex Some of our code and some code in 3rd-party libraries we currently depend on use APIs removed from Jandex 3.
2.  Also the group ID changed from `org.jboss` to `io.smallrye` (although the package name remains `org.jboss.jandex`).

See https://smallrye.io/blog/jandex-3-0-0/ for a summary of the changes.

Several areas in Helidon would be affected by the Jandex library change.
- [x] Update archetypes to generate `pom.xml` files depending on Jandex 3
- [x] Update example `pom.xml` files for Jandex 3 groupID change
- [x] Update integration `pom.xml` files for Jandex 3 groupID change
- [x] Update `helidon/microprofile` `pom.xml` files for Jandex 3 groupID change
- [x] Update `microprofile/lra/jax-rs` to avoid usage of APIs removed in Jandex 3
- [x] Update tests `pom.xml` for Jandex 3 groupID change
- [x] Update Jakarta persistence API to 3.1.0 (IIRC, because of the Hibernate update)

Tracked by a separate issue:
- #7448

tvallin:
### Description

Update Hibernate version to 6.2.7

Fix #6561

### Documentation

no impact


archetypes/helidon/src/main/archetype/mp/database/files/src/main/resources/hibernate.properties.mustache
```diff
@@ -1,3 +1,8 @@
 # Byte code for JPA must be generated at compile time.
 # This is a limitation of native image
-hibernate.bytecode.provider=none
+hibernate.bytecode.use_reflection_optimizer=false
+hibernate.connection.provider_class=org.hibernate.hikaricp.internal.HikariCPConnectionProvider
+hibernate.hikari.dataSourceClassName={{jdbcDataSource}}
+hibernate.hikari.dataSource.url={{databaseUrl}}
+hibernate.hikari.dataSource.user={{dbUser}}
+hibernate.hikari.dataSource.password={{password}}
```
ljnelson, line 8: Hmm; could you say why you needed to do this? Shouldn't be needed, but maybe you had a good reason to bypass our data source integration.

examples/integrations/cdi/pokemons/src/main/resources/META-INF/init_script.sql
```diff
@@ -17,9 +17,9 @@ INSERT INTO POKEMONTYPE VALUES (16, 'Dragon');
...
-INSERT INTO POKEMON VALUES (3, 'Squirtle', 11);
-INSERT INTO POKEMON VALUES (4, 'Caterpie', 7);
-INSERT INTO POKEMON VALUES (5, 'Weedle', 7);
-INSERT INTO POKEMON VALUES (6, 'Pidgey', 3);
+INSERT INTO POKEMON VALUES (1, 12, 'Bulbasaur');
+INSERT INTO POKEMON VALUES (2, 10, 'Charmander');
+INSERT INTO POKEMON VALUES (3, 11, 'Squirtle');
+INSERT INTO POKEMON VALUES (4, 7, 'Caterpie');
+INSERT INTO POKEMON VALUES (5, 7, 'Weedle');
+INSERT INTO POKEMON VALUES (6, 3, 'Pidgey');
```
ljnelson, line 25: This is the root of the problem, right? That for some reason Hibernate 6.2.x has changed the order of DDL/column generation? I would like to better understand why that is.

ljnelson: See #7433

PR #7442, opened 2023-08-24
4.x: A single set of APIs to create builders (all in builder-api).

### LINKED ISSUE #7443: 4.x: Design a single API to create Blueprints
Currently we have Config Metadata and Builder API that are combined to create builders, config metadata is additionally used to generate configuration documentation. Can this be merged into a single set of APIs?

### LINKED ISSUE #7444: 4.x: Design for mutable builders as part of other builders
Maybe support mutable builders as part of other builders (so we can update existing builder instead of creating a copy builder from the current result

tomas-langer:
API desing

Resolves #7443 
Resolves #7444

### Description

Provides a set of annotations to replace the ones from `helidon-config-metadata` for builders.
Required follow up
- change the config metadata processor to support the new APIs as well (and keep `config-metadata` only for non-builder use)
- modify all existing builders to use the new api
- fix default values for properties (now can be typed)

API in in `io.helidon.builder.api` - mostly type `Option` and modifications in `Prototype` - tried to move all single option related annotations to `Option`, kept type related annotations on `Prototype`, introduced `Description` as it is used in both.

Fully migrated the `helidon-builder-tests-builder` module to use the new API, to validate it works
Fully migrated `helidon-common-configurable` module to use the new API

Refactored `helidon-config-metadata-processor` to support both "old" config-metadata API and builder-api, added a test, validated on `common-configurable` module.

### Documentation

Enhances the builder API to be  a full blown approach that does not require any other module to create a fully configured builder. 
Types of interest (module `helidon-builder-api` in `builder/api` in the repository:
- `io.helidon.builder.api.Option` and all nested types
- `io.helidon.builder.api.Prototype` and all nested types
- `io.helidon.builder.api.Description`

Example code for `LruCacheBlueprint` (actual type updated in this PR, javadoc removed for clarity):
```java
@Prototype.Blueprint
@Prototype.Configured
interface LruCacheConfigBlueprint<K, V> extends Prototype.Factory<LruCache<K, V>> {
    @Option.DefaultInt(LruCache.DEFAULT_CAPACITY)
    @Option.Configured
    int capacity();
}
```


builder/api/src/main/java/io/helidon/builder/api/Description.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Custom description. The description is copied from javadoc by default.
+ * Use this annotation to customize the description, or when the javadoc is not available within
+ * the current module (e.g. when inherited from super type that is not part of the module).
+ */
+@Target({ElementType.METHOD, ElementType.TYPE})
+@Inherited
+@Retention(RetentionPolicy.CLASS)
```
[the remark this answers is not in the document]
tomas-langer: There is a problem when processing inhertited classes (for all annotations) - the sources are only available for the currently compiled module. If you inherit, or implement an interface from another module, we need class retention to be able to load those annotations.

builder/processor/src/main/java/io/helidon/builder/processor/Types.java
```diff
@@ -33,41 +33,72 @@ final class Types {
...
+    static final String PROTOTYPE_CONFIGURED = "io.helidon.builder.api.Prototype.Configured";
     static final String PROTOTYPE_CONFIGURED_BUILDER = "io.helidon.builder.api.Prototype.ConfiguredBuilder";
     static final String PROTOTYPE_BLUEPRINT = "io.helidon.builder.api.Prototype.Blueprint";
     static final String PROTOTYPE_FACTORY = "io.helidon.builder.api.Prototype.Factory";
     static final String PROTOTYPE_ANNOTATED = "io.helidon.builder.api.Prototype.Annotated";
     static final String CUSTOM_METHODS = "io.helidon.builder.api.Prototype.CustomMethods";
     static final String PROTOTYPE_CUSTOM_METHOD = "io.helidon.builder.api.Prototype.PrototypeMethod";
     static final String BUILDER_CUSTOM_METHOD = "io.helidon.builder.api.Prototype.BuilderMethod";
-    static final String SINGULAR = "io.helidon.builder.api.Prototype.Singular";
+    static final String DESCRIPTION = "io.helidon.builder.api.Description";
```
trentjeff, line 48: i wish there was a better way to declare these in java. this is fragile.

builder/tests/builder/src/main/java/io/helidon/builder/test/testsubjects/Level2Blueprint.java
```diff
@@ -29,36 +29,36 @@
...
     String getLevel0StringAttribute();
 
     /**
-     * Used for testing and demonstrating the use of {@link io.helidon.builder.api.Prototype.Singular}.
+     * Used for testing and demonstrating the use of {@link io.helidon.builder.api.Option.Singular}.
      *
      * @return ignored, here for testing purposes only
      */
-    @Prototype.Singular
+    @Option.Singular
```
trentjeff, line 45: Option seems appropriate for values, but I feel `Prototype` is/was more apropos for `Singular` as it aligns to the generated code.

builder/tests/builder/src/test/java/io/helidon/builder/test/DefaultValuesTest.java
```diff
@@ -0,0 +1,65 @@
...
+        assertThat(values.doubles(), hasItems(DEFAULT_DOUBLE, DEFAULT_DOUBLE));
+
+        assertThat(values.aLong(), is(DEFAULT_LONG));
+        assertThat(values.longs(), hasItems(DEFAULT_LONG, DEFAULT_LONG));
+
+        assertThat(values.anEnum(), is(DefaultValues.AnEnum.FIRST));
+        assertThat(values.enums(), hasItems(DefaultValues.AnEnum.FIRST, DefaultValues.AnEnum.SECOND, DefaultValues.AnEnum.THIRD));
+
+        assertThat(values.stringMap(), hasEntry("key1", "value1"));
+        assertThat(values.stringMap(), hasEntry("key2", "value2"));
```
trentjeff, line 63: might be nice to cover uri, date, and suppliers also.

common/key-util/src/main/java/io/helidon/common/pki/KeysBlueprint.java
```diff
@@ -84,15 +85,15 @@ interface KeysBlueprint {
...
+    @Option.Singular("certChain")
     List<X509Certificate> certChain();
 
     /**
      * The X.509 Certificates.
      *
      * @return the certificates configured or empty list if none configured
      */
-    @Prototype.Singular("cert")
+    @Option.Singular("cert")
```
trentjeff, line 96: "cert" was technically not needed here since the method name was plural.

common/processor/class-model/src/main/java/io/helidon/common/processor/classmodel/Field.java
```diff
@@ -163,6 +165,17 @@ && type().fqTypeName().equals(String.class.getName())
             return this;
         }
 
+        /**
+         * Set default value this field should be initialized with using the literal value provided.
```
trentjeff, line 169: review phrasing.

builder/processor/src/main/java/io/helidon/builder/processor/Types.java
tomas-langer: yes, but dependency on the module is worse. We want to keep the number of dependencies of annotation processors as low as possible

builder/tests/builder/src/main/java/io/helidon/builder/test/testsubjects/Level2Blueprint.java
tomas-langer: But the option is singular, not the prototype

builder/tests/builder/src/test/java/io/helidon/builder/test/DefaultValuesTest.java
tomas-langer: we do not support date and URI.

tomas-langer: Added a test for supplier with a default value.

trentjeff (approved):

builder/tests/common-types/src/main/java/io/helidon/common/types/AccessModifier.java
```diff
@@ -0,0 +1,55 @@
...
+    AccessModifier(String modifierName) {
+        this.modifierName = modifierName;
+    }
+
+    /**
+     * Return access modifier name which should be used in the generated component.
+     *
+     * @return modifier name
+     */
+    public String modifierName() {
```
trentjeff, line 51: how about just `name()` since it is already class named appropriately.

common/processor/processor/src/main/java/io/helidon/common/processor/ElementInfoPredicates.java
```diff
@@ -0,0 +1,151 @@
...
+        return element.modifiers().contains(TypeValues.MODIFIER_STATIC);
+    }
+
+    /**
+     * Predicate for private modifier.
+     *
+     * @param element typed element info to test
+     * @return whether the element has private modifier
+     */
+    public static boolean isPrivate(TypedElementInfo element) {
```
trentjeff, line 60: there are parts of the code where we check for package private - we should add that method as well.

tomas-langer:
After refactoring `common-configurable` to new buider API, I think this is ready for review.

builder/tests/builder/src/main/java/io/helidon/builder/test/testsubjects/Level2Blueprint.java
tomas-langer: Please check the usage - it looks logical, that all option (method) related annotations are prefixed with `Option`, and all prototype (class) related annotations are prefixed with `Prototype`.

builder/tests/common-types/src/main/java/io/helidon/common/types/AccessModifier.java
tomas-langer: This is existing code. We can change it, but it was not part of my PR

common/processor/processor/src/main/java/io/helidon/common/processor/ElementInfoPredicates.java
tomas-langer: Feel free to do that when you need it. I have only needed these methods in the code I was refactoring. I am sure there will be other useful predicates that we repeat...

common/configurable/src/main/java/io/helidon/common/configurable/ScheduledThreadPoolConfigBlueprint.java
```diff
@@ -0,0 +1,99 @@
...
+package io.helidon.common.configurable;
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+@Prototype.Blueprint
+@Prototype.Configured
+interface ScheduledThreadPoolConfigBlueprint extends Prototype.Factory<ScheduledThreadPoolSupplier> {
+    /**
+     * Default core pool size.
```
trentjeff, line 26: add {@value} in javadoc for these

common/configurable/src/main/java/io/helidon/common/configurable/ThreadPoolConfigBlueprint.java
```diff
@@ -0,0 +1,206 @@
...
+import java.util.Optional;
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+@Prototype.Blueprint
+@Prototype.Configured
+interface ThreadPoolConfigBlueprint extends Prototype.Factory<ThreadPoolSupplier> {
+    /**
+     * Default core pool size.
```
trentjeff, line 29: here too

http/http/src/main/java/io/helidon/http/RequestedUriDiscoveryContext.java
```diff
@@ -260,7 +260,7 @@ private void prepareAndCheckRequestedUriSettings() {
             }
             if (trustedProxies == null) {
                 trustedProxies = AllowList.builder()
-                        .addDenied(s -> true)
+                        .addDeniedPredicate(s -> true)
```
trentjeff, line 263: seems cleaner the old way. Typically `Predicate` is in the class name, not in method names.

builder/tests/common-types/src/main/java/io/helidon/common/types/AccessModifier.java
tomas-langer: So this cannot be done, as it would conflict with `Enum.name()`, which returns the name of the enum constant (such as `PUBLIC`

http/http/src/main/java/io/helidon/http/RequestedUriDiscoveryContext.java
tomas-langer: It cannot be done the old way, because there is `denied` already which is the exact string match, and we cannot use the same method for two different return types

tomas-langer: Trying to "hack" it by using `@Singular` with a duplicate method name

tomas-langer: yes, that seems to work

trentjeff (approved):

tomas-langer:
Had to do another rebase

trentjeff (approved):

romain-grecourt (approved):

PR #7455, opened 2023-08-25
3.x - Refactor TODO app example to remove the of use NodeJS/NPM

### LINKED ISSUE #7144: Remove example(s) that depend on node.js

romain-grecourt:
- Refactor UI to use JQuery + MustacheJS + Material icons (single .css file & single .js file)
- Refactor frontend service to use WebClient instead of JAXRS Client
- Simplify BackendServiceClient
- Update TodoServiceTest to write the backend mockup using WebServer instead of JAXRS
- Removed Docker and k8s to simplify the running steps
- Updated README.md

Fixes #7144


romain-grecourt:
This should be ported to 2.x and also to 4.x (with reactive -> blocking)

tomas-langer (approved):
Looks good, can you please create an issue to "forward-port" to 4.x?

Issue #7249, opened 2023-07-25
Use JSON-B instead of JSON-P in MP quickstarts


romain-grecourt:
MP quickstart and derivative use JSON-P to produce and consume JSON.
We need embrace a POJO based approach and have simple POJO for the entities.

Find all implementations of the MP greet service (E.g. search for `GreetResource`) and replace the usage of JSON-P with POJOs.

Do this on 3.x initially and port the changes to 2.x and 4.x

PR #7487, opened 2023-08-29
3.x: Use JSON-B instead of JSON-P in MP quickstarts

tvallin:
### Description

Update MP quickstarts and examples to use JSON-B + enable few example tests that are passing now.
See #7249. I will backport the changes if approved.

### Documentation

no impact


examples/microprofile/openapi-basic/pom.xml
```diff
@@ -44,6 +44,10 @@
             <groupId>io.helidon.microprofile.openapi</groupId>
             <artifactId>helidon-microprofile-openapi</artifactId>
         </dependency>
+        <dependency>
```
romain-grecourt, line 47: Is this example using both Jackson and Jsonb ? Is it possible to make it work on either but not both ?

tvallin: This example uses only JSON-B. I added` jackson` dependency because `helidon-microprofile-openapi` depends on `smallrye-open-api-core` where `jackson` is excluded so we use the latest version of `jackson`. 
I think `jackson` dependency was not there before because tests were disabled and the issue did not show up.

tvallin: As discussed, I removed the changes and I am going to open another issue to fix disabled tests.

examples/integrations/micrometer/mp/pom.xml
```diff
@@ -51,7 +51,12 @@
         </dependency>
         <dependency>
             <groupId>io.helidon.media</groupId>
-            <artifactId>helidon-media-jsonp</artifactId>
+            <artifactId>helidon-media-jsonb</artifactId>
```
barchetta, line 54: The `helidon-media-jsonb` dependency should not be needed. Can you remove it?

tvallin: Done

romain-grecourt (approved):

Issue #6518, opened 2023-03-30
Docs: fix `mutability-support.adoc`


tomas-langer:
section `Evolving API` describes method `Config.changes()` which no longer exists. This chapter should be revisited and fixed.

ljamen:
Hi @tomas-langer I will need additional information to address this doc issue. What is the fix for this section? Do you want it removed or updated for the new method? 

The note states:
Evolving API
This section describes the Config.changes() method. It is marked as **deprecated** because it returns an io.helidon.reactive.Flow.Publisher object. **In a future Helidon release** that requires Java 11 or later this method will be un-deprecated and **changed** — **or a similar method will be added** — so that the return type is java.util.concurrent.Flow.Publisher instead.

Any code you write using the existing Config.changes() method might need to change at that time.

PR #7500, opened 2023-08-30
3.x: Docs: Remove description of Config.changes() API

barchetta:
### Description

The `Config.changes()` API was removed after Helidon 1. This PR removes it from our documentation.

It also corrects the code snippet since the classpath config source does not support polling strategy. 

See #6518

### Documentation

Changes are in this PR

### DIFF
--- a/docs/se/config/mutability-support.adoc
+++ b/docs/se/config/mutability-support.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2018, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2018, 2023 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


docs/se/config/mutability-support.adoc
```diff
@@ -130,15 +118,15 @@ Config config = Config.create(
         ConfigSources.file("conf/config.properties")
                 .changeWatcher(FileSystemWatcher.create())                         // <2>
                 .optional(),
-        ConfigSources.classpath("application.properties")
+        ConfigSources.file("my.properties")
```
arjav-desai, line 121: Do we not support classpath any more? If so, there is a point above this piece of code that also needs to change i.e. 

> watch(Path watchedPath) - a filesystem-specific strategy to watch specified path. You can use this strategy with the file and classpath built-in config sources.

arjav-desai: My bad, you said polling is not supported and the above point is for watching so it should be fine.

arjav-desai (approved):

PR #7517, opened 2023-09-05
4.x: Value mapping API and implementation in modules.

### LINKED ISSUE #1897: mappable query params and path params

tomas-langer:
### Description
Resolves #1897 (mappable query params and path params)
Related to #1789 (align mappers)

### Documentation
This change introduces a few API changes that are needed to align method names and to implement the same interfaces.

#### New types
- `io.helidon.common.mapper.Value<T>` - a value, that supports mapping and a subset of `Optional` methods
- `io.helidon.common.mapper.OptionalValue<T>` - a value that may be empty, that supports mapping and full set of `Optional` methods (also extends `Value<T>`)

#### Modified types
- `io.helidon.common.config.ConfigValue` now extends `OptionalValue` (transitive impact on `io.helidon.config.ConfigValue`)
- `io.helidon.common.parameters.Parameters` now return `OptionalValue` instead of `Optional` in method `first`
- `io.helidon.common.parameters.Parameters` - renamed method `value` to `get`
- `Parameters` changes have transitive impact on `UriQuery`, `Uri` matrix parameters, `Uri` path parameters
- `io.helidon.dbclient.DbColumn` now implements `Value<String>` - as a result, methods `as(..)` are now renamed to `get(...)`, as they return an actual value, and method `value` is renamed to `get` (inherited from `Value`)

A very important change:
`io.helidon.http.Http.Header`:
- now extends `Value<String>`
- method `value()` will be removed, all usages should be replaced wil `.get()` (will create a follow up)
- method `value(Class<T>)` is removed

Example code:

HTTP headers (header once retrieved must have a value):
`req.headers().get(CONTENT_LENGTH).getLong()`

Query parameters (optional value):
`req.query().first("count").asInt().orElse(10)`

### DIFF
--- a/archetypes/helidon/src/main/archetype/se/custom/files/src/main/java/__pkg__/FileService.java.multipart.mustache
+++ b/archetypes/helidon/src/main/archetype/se/custom/files/src/main/java/__pkg__/FileService.java.multipart.mustache
@@ -92,7 +92,7 @@ final class FileService implements HttpService {
     }
 
     private void download(ServerRequest req, ServerResponse res) {
-        Path filePath = storage.resolve(req.path().pathParameters().value("fname"));
+        Path filePath = storage.resolve(req.path().pathParameters().get("fname"));
         if (!filePath.getParent().equals(storage)) {
             res.status(BAD_REQUEST_400).send("Invalid file name");
             return;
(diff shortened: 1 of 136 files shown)


tomas-langer:
Rebased

romain-grecourt (commented):
Good to go as-is. We can address follow-ups post M2.

romain-grecourt (approved):

PR #7524, opened 2023-09-06
4.x: introduce AfterStop annotation

### LINKED ISSUE #7518: 4.x `afterStop` method is not invoked during service shutdown
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
In `MetricsFeature` I added an implementation of `afterStop` (inherited ultimately from `ServerLifeCycle`) but it is never invoked, neither in a `@HelidonTest` or in MP quickstart.

## Steps to reproduce
Add `afterStop` to `MetricsFeature` and run it in MP.

@tomas-langer:
The problem is hidden in `io.helidon.webserver.servicecommon.HelidonFeatureSupport#setup(Builder, Builder)`
This method is either called from the feature's `setup(Builder)` or from our CDI extension, and as a result the feature is not registered directly with webserver routing (for the case it is called from CDI), as we only register the `#service()` 
When used in SE, we just register the feature on routing, so everything works find (unless you use the other method to support multiple sockets - one to server `/metrics` and one to watch for vendor metrics.

To correctly fix this:
1. we should always register the feature itself with the routing
2. we should support more than 2 possible endpoints (right now we can only have the "default" route and the "feature" route - but what if there are 4 sockets open, each one with its own routing setup?)

Workaround:
  change the `service()` method (such as in `MetricsFeature` to return an instance of an `HttpService` (not just a lambda) that has the `afterStop` method implemented with the desired tasks to do after the routing is stopped. This will work in both cases mentioned above.

tomas-langer:
### Description
Resolves #7518 
We cannot validate tasks done on server shutdown (both SE and MP), as `@AfterAll` is called first on the test class, and then on our extension.

### Documentation
This PR introduces `@AfterStop` annotation, that can be placed on a static method and validate things that should be done during shutdown. This is valid both for our MP testing (`@HelidonTest` annotation), and SE testing (`@ServerTest`, and `@RoutingTest` annotations).

For MP: `io.helidon.microprofile.tests.junit5.AfterStop`
For SE: `io.helidon.webserver.testing.junit5.AfterStop`

Example of usage:
```java
@AfterStop
static void afterStop() {
    assertThat("Before start should only have been called on server startup", FEATURE.beforeStart.get(), is(1));
    assertThat("After stop should have been called on server stop", FEATURE.afterStop.get(), is(1));
}
```

### DIFF
--- a/microprofile/tests/junit5/src/main/java/io/helidon/microprofile/tests/junit5/AfterStop.java
+++ b/microprofile/tests/junit5/src/main/java/io/helidon/microprofile/tests/junit5/AfterStop.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.microprofile.tests.junit5;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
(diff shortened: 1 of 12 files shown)


tjquinno (approved):
Approving. This is a useful addition to be able to clean up in the test environment. That said, I believe there is a bug in the server shutdown handling that fails to notify feature implementations of shutdown (detailed in the internal team slack channel). But that does not affect the suitability of this PR.

PR #7547, opened 2023-09-07
4.x Metrics followup

### LINKED ISSUE #7506: Metrics follow-up items
Items to follow-up on in metrics for 4.x

 - [x] Changes to get the build to pass for M2
     - [x] Main metrics
         - [x] SimpleMeterRegistryTests - 1 test disabled for pipeline
         - [x] MetricsConfigSupport#isMeterEnabled - needs to do the filtering; temporarily returns true always
         - [x] ScopeConfigSupport#isMeterEnabled - needs to do the meter-level filtering
     - [x] MP metrics
         - [x] BadGaugeTest
         - [x] TestBasicPerformanceIndicators.checkMetricsVendorUrl disabled
         - [x] HelloWorldAsyncResponseWithRestRequestTest
         - [x] whole HelloWorldAsyncResponseTest
         - [x] TestMetricsOnOwnSocket whole
         - [x] TestDisabledMetrics - leaving test disabled temporarily; no-op registry not provided if metrics are disabled
         - [x] MP metrics TCK disabled
         - [x] MpFeatureTest disabled
     - [x] Other tests
         - [x] FT MetricsTest - 3 tests
         - [x] Multiport functional test MainTest - commented code
         - [x] ServerMetricsCheckIT.testHttpMetrics integration/dbclient/app
         - [x] ProtectedMetricsTest test/integration/security/gh2297
         - [x] OCI metrics OciMetricsCdiExtensionTest - count is expected = 3 but is higher perhaps because gc based metrics are gauges not counters
         - [x] webclient/tests/webclient MetricsTest 1 test disabled
         - [x] tests/native-image/mp-1
     - [x] Examples
         - [x] filtering MP MainTest
         - [x] http status count mp

tjquinno:
### Description

Resolves #7506 (epic of TODO tasks from the M1 metrics work) and #7566 (issue for addressing all the epic task items).

This PR addresses the list of TODO items created as part of merging the neutral metrics API work into M2.

There are quite a few separate commits in this PR; I tried to separate the commits into related changes.

## NOTE
In this PR components that depend on metrics, in many cases, now have these dependencies:
* `helidon-webserver-observe-metrics` (which supports the `/metrics` endpoint and brings along the `helidon-metrics-api` and`helidon-metrics-providers-micrometer` artifacts as transitive dependencies)
* `helidon-metrics-system-meters` (the `base` meters that expose JVM and OS measurements) for those components that want them.

I made this adjustment after a discussion with several of the US folks but mostly Joe and Romain. This is consistent with the way components have dependencies on health in 4.x:
* `helidon-webserver-observe-health` supports `/health` and brings along `helidon-health` (the API and the implementation) 
* `helidon-health-checks` separately provides the Helidon-provided built-in health checks for those components that want them.

As a separate, broader topic we should think about whether there are useful bundles for health and metrics that would make sense, esp. for developers who are new to Helidon.

## General categories of changes

1. Re-enabling disabled tests, including the MP metrics TCK.
2. Make sure other components (esp. integrations, dbclient, etc.) with dep. on metrics have the correct dependencies: `helidon-metrics-providers-micrometer` and `helidon-metrics-system-meters` in many cases.
3. Reinstate the ability to disable metrics by scope and name:
   1. Config changes (regex expressions to include and exclude used to be `String`; now are `Pattern`)
   2. Logic changes (in a few places) to check whether a meter is enabled or not.
   3. Update the way config is used to disable metrics to the new config syntax.
4. Add support to the `NoOpMeterRegistry` to notify listeners when meters are added (or removed), primarily so the MP `Registry` can update the map between MP metrics and delegates. That's needed so we can return the same MP metric (rather than creating new no-op meters all the time) if the client retrieves by metric ID, for example.
5. Slight correction in the Prometheus output formatter to properly apply the scope selection when it is present.
6. Several of the system meters were declared as a functional counter when they should be gauges (or vice versa). The metrics TCK tells the difference.
7. Add in MP validate of tags to catch illegal use of reserved tag names (`mp_scope` and `mp_app`).
8. Some fixes in the JSON output formatter. Among other things, the JSON output no longer includes the scope tag and value; where suitable, meters of the same scope are grouped in the JSON object as before so the scope tag is superfluous and interfered with some tests.
9. Support for clean-up between tests in the same JVM.

### Documentation
No doc impact from this PR alone. Separate issue for metrics doc update.


examples/integrations/microstream/greetings-se/pom.xml
```diff
@@ -64,11 +64,15 @@
...
-            <artifactId>helidon-metrics</artifactId>
+            <groupId>io.helidon.webserver.observe</groupId>
+            <artifactId>helidon-webserver-observe-metrics</artifactId>
             <scope>runtime</scope>
         </dependency>
         <dependency>
+            <groupId>io.helidon.metrics</groupId>
+            <artifactId>helidon-metrics-system-meters</artifactId>
+            <scope>runtime</scope>
+        </dependency><dependency>
```
tomas-langer, line 75: missing end of line

examples/webserver/multiport/pom.xml
```diff
@@ -52,11 +52,15 @@
...
-            <artifactId>helidon-metrics</artifactId>
+            <groupId>io.helidon.webserver.observe</groupId>
+            <artifactId>helidon-webserver-observe-metrics</artifactId>
             <scope>runtime</scope>
         </dependency>
         <dependency>
+            <groupId>io.helidon.metrics</groupId>
+            <artifactId>helidon-metrics-system-meters</artifactId>
+            <scope>runtime</scope>
+        </dependency><dependency>
```
tomas-langer, line 63: missing end of line

metrics/api/src/main/java/io/helidon/metrics/api/NoOpMeterRegistry.java
```diff
@@ -108,21 +113,38 @@ public Iterable<String> scopes() {
...
+        M result = findOrRegister(NoOpMeter.Id.create(b.name(),
+                                                      b.tags()),
+                                  builder);
+        onAddListeners.forEach(listener -> listener.accept(result));
+        return result;
     }
 
     @Override
     public MeterRegistry onMeterAdded(Consumer<Meter> listener) {
+        onAddListeners.add(listener);
```
tomas-langer, line 125: Shouldn't this be guarded by a lock? I guess this can be multi-threaded. And if you call this and `onMeterRemoved` concurrently, you would get a concurrent modification exception.

metrics/api/src/test/java/io/helidon/metrics/api/TestScopeConfig.java
```diff
@@ -16,27 +16,26 @@
...
-// TODO enabled once patterns are back
-@Disabled
 class TestScopeConfig {
 
     @Test
     void checkEnableAllPattern() {
         ScopeConfig mts = ScopeConfig.builder()
                 .name("test")
-                .include(".*")
+                .include(Pattern.compile(".*"))
```
tomas-langer, line 38: You could add custom methods to builder also support `String` here if you wanted

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/KeyPerformanceIndicatorMetricsImpls.java
```diff
@@ -29,31 +31,26 @@
 
 class KeyPerformanceIndicatorMetricsImpls {
 
+    /**
+     * Name for metric recording number of requests currently deferred (received but not yet processing).
+     */
+    public static final String DEFERRED_NAME = "deferred";
```
tomas-langer, line 37: `public` is redundant here, as this is a package private class.

tomas-langer (changes requested):

examples/integrations/microstream/greetings-se/pom.xml
tjquinno: Fixed

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/KeyPerformanceIndicatorMetricsImpls.java
tjquinno: changed

metrics/api/src/main/java/io/helidon/metrics/api/NoOpMeterRegistry.java
tjquinno: Changed.

Both lists of listeners (onAdd and onRemove) are now `CopyOnWriteArrayList`, here and in `MMeterRegistry`. 

Update operations will be very rare compared to iterating over the lists.

metrics/api/src/test/java/io/helidon/metrics/api/TestScopeConfig.java
tjquinno: Added

tomas-langer (approved):

PR #7548, opened 2023-09-07
4.x: MP testing refactoring

tomas-langer:
### Description
1. Move user modules to `testing` to align with all other modules in Helidon (where we use `testing` for types to support users during testing of Helidon applications, and `tests` to contain our own tests
2. Refactor group ids (and some artifact ids) to align with what we should have in Helidon in MP `tests`
3. Rename  `restfull` to `restful`, as that is the correct name of the specification

### Documentation
There is no change in the types except for being moved to different modules/packages.

Two modules are now moved:
GroupId has changed from `io.helidon.microprofile.tests` to `io.helidon.microprofile.testing` for JUnit5 and TestNG support for MicroProfile

Junit5:
1. artifactId has changed from `helidon-microprofile-tests-junit5` to `helidon-microprofile-testing-junit5`
2. package has changed from `io.helidon.microprofile.tests.junit5` to `io.helidon.microprofile.testing.junit5`
3. module anme has changed from `io.helidon.microprofile.tests.junit5` to `io.helidon.microprofile.testing.junit5`

TestNG:
1. artifactId has changed from `helidon-microprofile-tests-testng` to `helidon-microprofile-testing-testng`
2. package has changed from `io.helidon.microprofile.tests.testng` to `io.helidon.microprofile.testing.testng`
3. module anme has changed from `io.helidon.microprofile.tests.testng` to `io.helidon.microprofile.testing.testng`


microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/AddConfigs.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.helidon.microprofile.tests.testng;
+package io.helidon.microprofile.testing.junit5;
```
dalexandrov, line 16: Why are you mixing `junit5` and `testng` ?

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/Configuration.java
```diff
@@ -33,7 +33,7 @@
      * If set to {@code true}, the existing (or default) MicroProfile configuration would be used. In this case it is
      * important to set property {@code mp.initializer.allow=true} in order CDI container to start, when used with
      * {@link HelidonTest}.
-     * By default uses a configuration constructed using all {@link io.helidon.microprofile.tests.junit5.AddConfig}
+     * By default uses a configuration constructed using all {@link AddConfig}
```
dalexandrov, line 36: Should we keep the full names in comments (I was addressed the same issue in my PRs)

tomas-langer: As long as it is from the same package, it is OK. The reason we do this is to have `import` statements on a class file only if used in code. As this does not introduce an import, I consider this OK

tomas-langer: Nevertheless this was the result of Idea refactoring, I did not want to do it...

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/AddConfigs.java
danielkec, line 16: Why move testng ones to junit and vice-versa?

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/AddConfigs.java
```diff
@@ -13,16 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.helidon.microprofile.tests.junit5;
+package io.helidon.microprofile.testing.testng;
```
danielkec, line 16: Vice-versa...

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/AddConfigs.java
tomas-langer: I did not. I refactored it using IDE by moving the correct ones around

danielkec (approved):

Verdent (dismissed):

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/Configuration.java
tomas-langer: Fixing

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/AddConfigs.java
tomas-langer: Fixing

tomas-langer: I have used `git mv` to move the files, yet still some of them appear to be incorrectly marked. I would prefer not to manually change the git patch, so I have no choice but to leave it as is

danielkec (approved):

PR #7556, opened 2023-09-08
4.x Refactor archetype metrics

### LINKED ISSUE #7549: 4.x Archetypes contain incorrect direct references to Micrometer
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Although the archetypes component builds without error messages after the recent metrics merge, the generated code is incorrect with quite a few direct uses of Micrometer. These are not new with the recent merge; I think they might have been this way since M1 although that is not very important.

### Background
As part of the metrics follow-up work I am reducing dependencies in other components to the minimum.

In the `pom.xml` for `webserver/observe/metrics`  change the build-time dependency on `helidon-metrics-providers-micrometer` to a test-scoped dependency on `helidon-metrics`. With that change built and installed locally the archetypes build failed in trying to build generated projects because the generated projects contain direct uses of Micrometer elements.

### Needed changes in the archetypes
For projects generated from the archetypes that use metrics, there should be a compile-time dependency on `helidon-metrics-api` and a run-time dependency on `helidon-metrics`. (The `helidon-metrics` component is a bundle of the built-in `base` metrics and the Micrometer implementation of the neutral Helidon metrics API.) 

The generated projects should contain no direct uses of anything from Micrometer, either in `pom.xml` dependencies or `module-info.java` or the code.

Specifically, in the generated `module-info.java` there should be no reference to any of the Micrometer modules (such as `micrometer.core`). In the generated Java code there should be no use of the Micrometer annotations or types.

For MP apps, the types and annotations should be from the packages`org.eclipse.microprofile.metrics` and `org.eclipse.microprofile.metrics.annotation`. MP apps should probably not even use the types from `io.helidon.metrics.api`;as MP apps they should use the MP metrics API.

For SE apps, the types should be from the `io.helidon.metrics.api` package and module.

Changing the dependency in `webserver/observe/metrics` as described above is a good starting point to see the problem, but making that change alone might not show all the places that need to be changed. 

## Steps to reproduce
In the `webserver/observe/metrics/pom.xml` remove the compile-time dependency on `helidon-metrics-providers-micrometer` and replace it with a test-scoped dependency on `helidon-metrics`. Build and install that component, then try to build the archetypes component. The attempts to build generated projects fails.

tvallin:
### Description

Fix #7549

### Documentation

No impact

### DIFF
--- a/archetypes/helidon/src/main/archetype/common/extra.xml
+++ b/archetypes/helidon/src/main/archetype/common/extra.xml
@@ -101,6 +101,11 @@ java -cp "target/classes:target/libs/*" {{package}}.WebClientMain PORT
                                     <value key="groupId">io.helidon.microprofile.metrics</value>
                                     <value key="artifactId">helidon-microprofile-metrics</value>
                                 </map>
+                                <map>
+                                    <value key="groupId">io.helidon.metrics</value>
+                                    <value key="artifactId">helidon-metrics</value>
+                                    <value key="scope">compile</value>
+                                </map>
                             </list>
                             <list key="modules">
                                 <value if="${flavor} == 'mp'">io.helidon.microprofile.faulttolerance</value>
--- a/archetypes/helidon/src/main/archetype/common/extra.xml
+++ b/archetypes/helidon/src/main/archetype/common/extra.xml
@@ -104,7 +104,7 @@ java -cp "target/classes:target/libs/*" {{package}}.WebClientMain PORT
                                 <map>
                                     <value key="groupId">io.helidon.metrics</value>
                                     <value key="artifactId">helidon-metrics</value>
-                                    <value key="scope">compile</value>
+                                    <value key="scope">runtime</value>
                                 </map>
                             </list>
                             <list key="modules">
--- a/archetypes/helidon/src/main/archetype/common/observability.xml
+++ b/archetypes/helidon/src/main/archetype/common/observability.xml
@@ -226,7 +226,8 @@ curl -H 'Accept: application/json' -X GET http://localhost:8080/metrics
                                         <map>
                                             <value key="groupId">io.helidon.metrics</value>
                                             <value key="artifactId">helidon-metrics-api</value>
-                                            <value key="scope">compile</value>
+                                            <value key="scope" if="${flavor} == 'mp'">runtime</value>
+                                            <value key="scope" if="${flavor} == 'se'">compile</value>
                                         </map>
                                         <map>
                                             <value key="groupId">io.helidon.metrics</value>
--- a/archetypes/helidon/src/main/archetype/common/observability.xml
+++ b/archetypes/helidon/src/main/archetype/common/observability.xml
@@ -55,14 +55,23 @@ curl -H 'Accept: application/json' -X GET http://localhost:8080/metrics
 
 ]]></value>
                                     </list>
-                                    <list key="dependencies" if="${flavor} == 'mp'">
-                                        <map order="800">
+                                    <list key="dependencies">
+                                        <map>
+                                            <value key="groupId">io.helidon.metrics</value>
+                                            <value key="artifactId">helidon-metrics</value>
+                                            <value key="scope">runtime</value>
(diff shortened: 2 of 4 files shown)


archetypes/helidon/src/main/archetype/common/extra.xml
```diff
@@ -101,6 +101,11 @@ java -cp "target/classes:target/libs/*" {{package}}.WebClientMain PORT
                                     <value key="groupId">io.helidon.microprofile.metrics</value>
                                     <value key="artifactId">helidon-microprofile-metrics</value>
                                 </map>
+                                <map>
+                                    <value key="groupId">io.helidon.metrics</value>
+                                    <value key="artifactId">helidon-metrics</value>
+                                    <value key="scope">compile</value>
```
tjquinno, line 107: IIUC this is used in the MP case. `helidon-metrics` includes `helidon-metrics-api` which we would normally not need at build-time for MP. It's certainly OK for an MP app to use the Helidon neutral metrics API but it's a bit unusual.

Can this be `runtime` scope instead or do we need it to be `compile`?

archetypes/helidon/src/main/archetype/common/observability.xml
```diff
@@ -201,18 +223,28 @@ curl -H 'Accept: application/json' -X GET http://localhost:8080/metrics
                             <output>
                                 <model>
                                     <list key="dependencies">
-                                        <map if="${flavor} == 'mp'">
-                                            <value key="groupId">io.helidon.integrations.micrometer</value>
-                                            <value key="artifactId">helidon-integrations-micrometer-cdi</value>
+                                        <map>
+                                            <value key="groupId">io.helidon.metrics</value>
+                                            <value key="artifactId">helidon-metrics-api</value>
+                                            <value key="scope">compile</value>
```
tjquinno, line 229: Similar question as above about the scoping for `helidon-metrics`. For SE apps that will use the neutral API it definitely makes sense to be compile-scoped. But for MP, runtime should (normally) be sufficient. That's assuming that the code uses the MP metrics API and not the Helidon neutral metrics API.

archetypes/helidon/src/main/archetype/common/observability.xml
```diff
@@ -249,58 +298,50 @@ curl -H 'Accept: application/json' -X GET http://localhost:8080/metrics
...
         assertThat(message, is("Hello Eric"));
-        double after = counter.count();
+        double after = counter.getCount();
         assertEquals(1d, after - before, "Difference in personalized greeting counter between successive calls");
     }]]></value>
                                     </list>
-                                    <list key="MainTest-static-fields" if="${flavor} == 'mp'">
-                                        <value><![CDATA[
-    @Inject
-    private MeterRegistry registry;]]></value>
```
tjquinno, line 266: I am not sure what code used the injected `MeterRegistry` before these changes. (Maybe none?) Certainly we should not inject a `MeterRegistry` but injecting the MP `MetricRegistry` and then using it is fine for an MP app. Certainly if the code would not use an injected MP `MetricRegistry` then there's no need to inject it.

tjquinno (commented):
A few questions for my understanding.

archetypes/helidon/src/main/archetype/common/observability.xml
tvallin: I changed `MeterRegistry` to `MetricRegistry` for MP application. You can see this code few lines above:
```xml
<list key="MainTest-static-fields" if="${flavor} == 'mp'">
                                        <value><![CDATA[
    @Inject
    private MetricRegistry registry;]]></value>
</list>
```
But as I moved it, it looks like I removed it but I just modified it.

archetypes/helidon/src/main/archetype/common/extra.xml
tvallin: Sure, done !

tjquinno (approved):

PR #7582, opened 2023-09-13
[4.x] Modifies Helidon Connector to use WebClient and also support HTTP/2

spericas:
## Problem Description
- Modifies Helidon connector to use WebClient and thus support HTTP/1.1 and HTTP/2
- Fixes problem in http2 module-info.java
- Updates tests to verify HTTP/2 using protocol negotiation (from HTTP/1.1 and using ALPN) and prior knowledge

## Documentation
See #7640


jersey/connector/src/main/java/io/helidon/jersey/connector/HelidonConnector.java
```diff
@@ -130,15 +153,23 @@ private Http1ClientRequest mapRequest(ClientRequest request) {
         });
 
         // SSL context
-        SSLContext sslContext = client.getSslContext();
-        httpRequest.tls(Tls.builder().sslContext(sslContext).build());
+        if (!hasTls) {
```
tomas-langer, line 156: This seems inversed? If does not have tls, configure it?

jersey/connector/src/main/java/io/helidon/jersey/connector/HelidonProperties.java
```diff
@@ -27,8 +32,44 @@ private HelidonProperties() {
...
+     * Property name to set a {@code Map<String, String>} instance with a list of
+     * default headers to be used by underlying {@link WebClient}.
+     * This property is settable on {@link jakarta.ws.rs.core.Configurable#property(String, Object)}.
+     *
+     * @see io.helidon.webclient.api.WebClientConfig.Builder#defaultHeadersMap(Map)
+     */
+    public static final String DEFAULT_HEADERS = "jersey.connector.helidon.defaultHeaders";
+
+    /**
+     * Property name to set a protocol ID for each request. You can use this property
```
tomas-langer, line 69: Maybe expand a bit on this property to mention that when TLS is used, we use ALPN, so it is not needed

jersey/tests/connector/pom.xml
```diff
@@ -53,6 +53,11 @@
             <artifactId>helidon-common-testing-http-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
tomas-langer, line 56: The connector requires http2 in its module-info.java, so the library must be a dependency - e.g. it should not be required here

tomas-langer (changes requested):

jersey/tests/connector/pom.xml
spericas: Roger

jersey/connector/src/main/java/io/helidon/jersey/connector/HelidonProperties.java
spericas: Done

tomas-langer (approved):

PR #7587, opened 2023-09-14
4.x: Allowed cipher suite test added

### LINKED ISSUE #5701: Intermittent failure in CipherSuiteTest
I get the following intermittent error in pipelines (at least 5 times already):
```
Error:  Failures: 
Error:    CipherSuiteTest.testUnsupportedAlgorithm:108 
Expected: an instance of javax.net.ssl.SSLHandshakeException
     but: <java.lang.IllegalStateException: (client reqID: 3) Connection reset by the host> is a java.lang.IllegalStateException
```

Findings after first analysis:
Test fails because the connection is sometimes closed before we receive the SSL Handshake Failure. 
The exception (as can be seen above) claims the connection was reset by host, so the server probably issues a close before the correct SSL error is sent back. I could not find the problem on the server side.
As this is a critical test failure that is blocking our pipelines, I have created a temporary fix that accepts connection close as a valid error. This must be fixed before release.
Could not reproduce on local Linux or MacOS, only fails on pipeline environment (Github Actions)

Verdent:
### Description
Allowed cipher suite test was removed during reactive removal and we have no replacement for it right now. This PR introduces this test back, but written for the new blocking server.

It also contains fix for incorrectly created URLs in our JUnit5 support, when TLS is enabled.

Fixes #5701

### Documentation

None


webserver/tests/webserver/src/test/resources/cipherSuiteConfig.yaml
```diff
@@ -0,0 +1,49 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+server:
+  port: 0
```
tomas-langer, line 18: This should nto be required - this is the default in Helidon SE

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/CipherSuiteTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+@ServerTest
+public class CipherSuiteTest {
```
tomas-langer, line 43: Please remove `public` modifier

tomas-langer (changes requested):

tomas-langer (approved):

PR #7590, opened 2023-09-14
4.x: Mark inject processor as preview (warning when executed).

### LINKED ISSUE #7413: 4.x: Mark Helidon inject modules as preview
We consider it a production ready feature, yet it may get some backward incompatible changes as we introduce Helidon declarative model.

Make sure that this is not reported for Helidon SE and MP applicaitons during startup.

tomas-langer:
### Description
Resolves #7413 

Helidon inject processor will print a warning when configured in a project.
This can be disabled through `-Ainject.acceptPreview=true` compiler argument

### Documentation
Documentation should mention that `helidon-inject` and all related modules are in `preview` - we may need to do backward incompatible changes to these modules, so if customer uses them directly or through annotation processor, they need to be aware of this.

### DIFF
--- a/inject/processor/src/main/java/io/helidon/inject/processor/BaseAnnotationProcessor.java
+++ b/inject/processor/src/main/java/io/helidon/inject/processor/BaseAnnotationProcessor.java
@@ -19,18 +19,22 @@ package io.helidon.inject.processor;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.annotation.processing.AbstractProcessor;
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.SourceVersion;
 import javax.lang.model.element.TypeElement;
+import javax.tools.Diagnostic;
 
 import io.helidon.common.types.TypeName;
+import io.helidon.inject.tools.Options;
 
 /**
  * Abstract base for all Helidon annotation processing.
  */
 abstract class BaseAnnotationProcessor extends AbstractProcessor {
+    private static final AtomicBoolean LOGGED_WARNING = new AtomicBoolean();
     private final System.Logger logger = System.getLogger(getClass().getName());
 
     private ActiveProcessorUtils utils;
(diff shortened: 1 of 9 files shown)


inject/processor/src/main/java/io/helidon/inject/processor/BaseAnnotationProcessor.java
```diff
@@ -53,6 +57,15 @@ public SourceVersion getSupportedSourceVersion() {
     public void init(ProcessingEnvironment processingEnv) {
         this.utils = new ActiveProcessorUtils(this, processingEnv);
         super.init(processingEnv);
+
+        if (!Options.isOptionEnabled(Options.TAG_ACCEPT_PREVIEW)
+                && LOGGED_WARNING.compareAndSet(false, true)) {
+            processingEnv.getMessager()
+                    .printMessage(Diagnostic.Kind.WARNING,
+                                  "Helidon Inject is a preview feature, and may introduce backward incompatible changes. "
+                                          + "It is a production quality code, but we may need to do fine tuning of APIs and SPIs."
```
trentjeff, line 66: Consider rephrasing (i.e., "drop 'a', etc.) "It is production quality code, but the API and SPIs may be modified in a future revision."

inject/tests/resources-inject/pom.xml
```diff
@@ -130,6 +125,11 @@
                     </annotationProcessorPaths>
                 </configuration>
                 <dependencies>
+                    <dependency>
+                        <groupId>io.helidon.inject</groupId>
+                        <artifactId>helidon-inject-processor</artifactId>
+                        <version>${helidon.version}</version>
+                    </dependency>
 				    <dependency>
```
trentjeff, line 133: nit: spacing.

tomas-langer: This is caused by the original PR not following Helidon code style and using tabs instead of spaces. Looks good in my IDE, as I have tab spacing set to the same value as the default spacing

tomas-langer: Fixed to spaces

inject/processor/src/main/java/io/helidon/inject/processor/BaseAnnotationProcessor.java
tomas-langer: Rephrased, hope it is OK now

trentjeff (approved):

Issue #7503, opened 2023-08-31
3.x: Enable tests that were disabled during renaming to `jakarta` packages work


tvallin:
## Environment Details
* Helidon Version: 3.x
* Helidon SE and MP
* JDK version: 17
* OS:
* Docker version (if applicable):

----------

## Problem Description

Tests have to be re enabled because things can break under the radar.

## Steps to reproduce

n/a

- [ ] Example Microprofile -> ImplicitHelloWorldTest
- [ ] Example Integrations CDI pokemons -> MainTest
- [ ] Example Integrations Micronaut -> MicronautExampleTest
- [ ] Example Integration Neo-4j-mp -> MainTest
- [ ] Example Microprofile Openapi-basic -> MainTest
- [ ] Example Openapi -> MainTest
- [ ] Example TODO-app -> BackendTests
- [ ] Microprofile FaultTolerance -> RetryTest
- [ ] Tests Functional BookStore -> MainTest#basicTestMetricsHealthSEModules
- [ ] Tests Functional BookStore -> MainTest#basicTestMetricsHealthMPModules
- [ ] Tests Functional MultipleApps -> MainTest
- [ ] Tests Integration Zipkin-mp-2.2 -> ZipkinMp22Test#invokeEndPoint

PR #7606, opened 2023-09-15
3.x: Enable tests that were disabled during renaming to jakarta packages work

tvallin:
### Description

fix https://github.com/helidon-io/helidon/issues/7503

### Documentation

no impact

### DIFF
--- a/examples/integrations/micronaut/data/src/main/java/io/helidon/examples/integrations/micronaut/data/BeanValidationExceptionMapper.java
+++ b/examples/integrations/micronaut/data/src/main/java/io/helidon/examples/integrations/micronaut/data/BeanValidationExceptionMapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2023 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,8 @@
 
 package io.helidon.examples.integrations.micronaut.data;
 
-import jakarta.validation.ConstraintViolationException;
+import javax.validation.ConstraintViolationException;
+
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.ExceptionMapper;
 import jakarta.ws.rs.ext.Provider;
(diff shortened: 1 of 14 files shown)


openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
```diff
@@ -561,8 +561,7 @@ private Optional<MediaType> chooseResponseMediaType(ServerRequest req) {
     private static class HelidonAnnotationScannerExtension implements AnnotationScannerExtension {
 
         @Override
-        public Object parseExtension(String key, String value) {
-
+        public Object parseValue(String value) {
```
tjquinno, line 564: I do not see why we need to create `parseValue` and use it from `parseExtension`. IIRC `parseValue` is a new method added to the SmallRye library in a later release (we use the later release in Helidon 4.x but not in 3.x). I notice that the new `parseValue` method here does not have `@Override` as it does in 4.x (which makes sense given that the method does not exist yet in SmallRye's library, at least from the Helidon 3.x perspective).

That said, the change seems to do no harm so I'm OK with it. If I have missed something about why it was added please let me know.

examples/integrations/micronaut/data/src/main/java/io/helidon/examples/integrations/micronaut/data/BeanValidationExceptionMapper.java
```diff
@@ -16,7 +16,8 @@
 
 package io.helidon.examples.integrations.micronaut.data;
 
-import jakarta.validation.ConstraintViolationException;
+import javax.validation.ConstraintViolationException;
```
tjquinno, line 19: Should this actually be changed from `jakarta` to `javax`? 

Same question for the similar change in `PetResource.java` below.

openapi/src/main/java/io/helidon/openapi/OpenAPISupport.java
tvallin: When enabling test on `openapi-basic` example. It throws a class def not found exception about `jackson`.
In 3.x, the smallrye library version used is `2.1.16` and the `AnnotationScannerExtension` uses the `parseValue` with default implementation and uses `jackson`.
The new `parseValue` from my PR does have the `@Override` annotation ... line 563.
You can easily reproduce the issue by running `openapi-basic` example and enabling the test.

examples/integrations/micronaut/data/src/main/java/io/helidon/examples/integrations/micronaut/data/BeanValidationExceptionMapper.java
tvallin: With the current version of `micronaut`, it does not handle `jakarta.validation`.

Here is the PR : 
Revert back to using javax.validation as jakarta.validation is not supported by micronaut [#4476 ](https://github.com/helidon-io/helidon/pull/4476)

tjquinno (approved):

PR #7612, opened 2023-09-18
4.x: Move archetype metrics under `GreetService`

### LINKED ISSUE #7611: 4.x QuickStart SE archetype creates unneeded and confusingly-named `MetricsService`
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP SE
* JDK version:
* OS:

----------

## Problem Description
The SE QuickStart archetype generates a `MetricsService` which is a variant of the `GreetingService` with a `Counter` meter registered and updated explicitly. 

1. The name of the service is misleading.
2. We should not need a separate application service to illustrate this. It could be part of the normal `GreetingService` instead which would be simpler for developers.

## Steps to reproduce
Run `helidon init --version 4.0.0-M2` and choose SE quickstart. The generated project contains the `MetricsService`.

tvallin:
### Description

fix #7611 

Removed the MetricsService and moved the `metrics` logic to `GreetService`

### Documentation

no impact


archetypes/helidon/src/main/archetype/common/observability.xml
```diff
@@ -328,36 +322,36 @@ with the number of time it was triggered.
 
 ```
 curl http://localhost:8080/observe/metrics
-# HELP allRequests_total
+# HELP allRequests_totalb
```
tjquinno, line 325: A typo with the extra "b" on the end?

tjquinno (commented):
I see that the generated test assertions now check only for the presence of the expected metrics and not the values. It would be nice if there was a way to verify the value as well, as we did before. I know it can be tedious to get the output, parse the output and save the current value, then access an endpoint, then retrieve the output and parse it again, and finally make sure the _change_ in value is correct (rather than the absolute value) but would that work?

archetypes/helidon/src/main/archetype/common/observability.xml
tvallin: Yes that's a typo ! And unexpected change, thanks for catching that !

tvallin:
Yes that would work. Other possibility is to use ordered test method and run metric test first so we can avoid parsing metrics endpoint output for SE. Would you like this approach ?

tjquinno (commented):
Mostly looks good, but I am reluctant to approve the PR because there will be more changes to how the archetype should generate the SE code once Tomas finishes some current work. 

Also, ideally we would avoid depending on test ordering among the test methods. If Romain is OK with it then I will be as well.

tvallin:
That's okay, I just removed the ordered test and parse the metrics instead to get the values.

tjquinno:
1. You have probably already seen the conflict now in `observerability.xml`.
2. I cloned your repo and switched to your `refactor-archetype` and rebuilt Helidon and the archetypes.
   
   When I run `helidon init` using the local `cli-data` the dialog for a custom SE project prompts for a metrics provider (Microprofile or Micrometer). If the user asks for metrics in the custom SE dialog, the pom _should not_ ask further metrics-related questions (similar to the recent change to the 3.x archetype).
   
   A generated SE project with metrics support should include these metrics-related  dependencies:   
   
   ```xml
        <dependency>
            <groupId>io.helidon.webserver.observe</groupId>
            <artifactId>helidon-webserver-observe-metrics</artifactId>
        </dependency>
        <dependency>
            <groupId>io.helidon.metrics</groupId>
            <artifactId>helidon-metrics-system-meters</artifactId>
            <scope>runtime</scope>
        </dependency>
   ```
   
   This might be a change from when we first talked about what metrics dependencies should be in 4.x SE projects. For now, at least, we have decided not to use the `helidon-metrics` bundle in our SE metrics examples.

tjquinno (approved):

PR #7625, opened 2023-09-19
4.x: Observability update

### LINKED ISSUE #7615: 4.x: Simplify observability customization
Currently it is quite hard to add custom health check using API 
- we need to create `HealthFeature`
- we need to create `HealthObserveProvider`
- we need to register it with `ObserveFeature` (and how does this work with service loaded observe providers?)

1. the `HealthObserveProvider` should not be created in this case (as it is only for service loader)
2. it should be clear that the observe feature will contain all other providers from classpath (e.g. instead of `addFeature` we should have `updateFeature` or similar)
3. try to make it more user friendly to customize things

tomas-langer:
### Description
Resolves #7615 

Update to observability features - design proposal (health checks updated)
Usage: https://github.com/helidon-io/helidon/blob/db0245d3b99d9bc0d58d42c4fe02f6a28bb92417/webserver/tests/observe/observe/src/test/java/io/helidon/webserver/tests/observe/ObserveTest.java


romain-grecourt (approved):

barchetta (approved):
LGTM! 🥇

tjquinno (approved):
LGTM. Thanks.

Issue #7508, opened 2023-09-03
Missing permutations in archetype integration tests


tvallin:
## Environment Details
* Helidon Version:
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description

The `filters.properties` contains this line: 

```
# force metrics.provider='microprofile' when tracing=true
tracing=!${tracing} || (${tracing} && ${metrics.provider} == 'microprofile')
```

The permutation contains only `microprofile` value as metric provider and escape others such as `micrometer`.

## Steps to reproduce

- Build archetype
- Check `permutations.txt`

Issue #7176, opened 2023-07-12
Cannot run archetype because `versions.xml` is missing


tvallin:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
Cannot run the archetype with the CLI. 

## Steps to reproduce
- Build archetype
- Run `helidon init --reset --url file:///path/to/helidon/archetypes/helidon/target/cli-data`
- Error: `error: /path/to/helidon/archetypes/helidon/target/cli-data/versions.xml (No such file or directory)`

PR #7638, opened 2023-09-20
[3.x] archetypes does not offer a metrics provider choice for custom/se

### LINKED ISSUE #7557: [3.x] archetypes should not offer a metrics provider choice for custom/se
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
In 3.x with `helidon init` for a custom SE project, we offer a choice of metrics providers: MicroProfile and Micrometer. These are not really equivalent providers.

The "normal" SE metrics in 3.x does rely on the MP metrics types (but not annotations, etc.) but it's kind of confusing to offer "MicroProfile" as an SE metrics provider. Further, the Micrometer support in 3.x is a completely separate integration; the Micrometer integration does not get the built-in metrics automatically registered, etc. It just makes it a little easier for developers to use the Micrometer API to talk directly to Micrometer meter registries.

We don't really have multiple providers that implement the SE metrics API, so we can and should remove the provider choice for metrics.

If a user chooses a custom SE project and chooses "yes" for observability and then "yes" for metrics, we should offer no further choices about metrics and the metrics support should be provided by `helidon-metrics`.

## Steps to reproduce
`helidon init`, choose SE, custom, and for observability respond "yes" and response "yes" to metrics. You then get two choices, Microprofile and Micrometer.

tvallin:
### Description

fix #7557

I backported critical changes about testing and compatibility. 
#7508
#7176

### Documentation

no impact

### DIFF
--- a/archetypes/helidon/filters.properties
+++ b/archetypes/helidon/filters.properties
@@ -36,9 +36,6 @@ health=!${health} || (${health.builtin})
 # force metrics.builtin=true
 metrics=!${metrics} || (${metrics.builtin})
 
-# force metrics.provider='microprofile' when tracing=true
-tracing=!${tracing} || (${tracing} && ${metrics.provider} == 'microprofile')
-
 # group extra options
 extra=${extra} == [] || \
   (${flavor} == 'se' && ${extra} == ['cors', 'webclient', 'fault-tolerance']) || \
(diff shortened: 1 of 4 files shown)


tjquinno (approved):
LGTM. I built your branch locally and ran `helidon init` and the metrics dialog is much better.

Issue #7628, opened 2023-09-19
4.x: demonstrate proper use of GlobalConfig in examples and archetypes


barchetta:
This issue depends on issue #7627 to be resolved first. Once it is we need to make sure that documentation, examples and archetypes illustrate the preferred pattern for setting up config. For example:

```
// This loads default config sources including `application.yaml`.  If you want to create custom config sources you can 
// do that here instead of just using the default config.
Config config = Config.create();
GlobalConfig.config(config);
```

With GlobalConfig we no longer need to pass config to the `routing` method found in many examples, nor to the `HttpService` found in many examples. Instead any customer code that needs config can do something like:

```
Config config = GlobalConfig.config().get("myapplication");
```

tvallin:
Close thanks to #7655 and #7664, thanks @barchetta

PR #7655, opened 2023-09-22
4.x: Update examples to use Config.global(config)

barchetta:
### Description

* Remove use of `io.common.config` from examples
* Use pattern of explicitly initializing global config using `Config.global(config)`
* Use pattern of Http Services having a no-arg constructor  (use `Config.global()` instead)

This PR is not necessarily comprehensive. I did not inspect every example, just hit the common and obvious ones.

See Also: #7628

### Documentation

No impact

### DIFF
--- a/examples/cors/src/main/java/io/helidon/examples/cors/Main.java
+++ b/examples/cors/src/main/java/io/helidon/examples/cors/Main.java
@@ -68,14 +68,14 @@ public final class Main {
      * @param routing routing builder
      */
     static void routing(HttpRouting.Builder routing) {
-        Config config = Config.global();
 
         // Note: Add the CORS routing *before* registering the GreetService routing.
-        routing.register("/greet", corsSupportForGreeting(config), new GreetService())
+        routing.register("/greet", corsSupportForGreeting(), new GreetService())
                 .addFeature(ObserveFeature.create());
     }
 
-    private static CorsSupport corsSupportForGreeting(Config config) {
+    private static CorsSupport corsSupportForGreeting() {
+        Config config = Config.global();
 
         // The default CorsSupport object (obtained using CorsSupport.create()) allows sharing for any HTTP method and with any
         // origin. Using CorsSupport.create(Config) with a missing config node yields a default CorsSupport, which might not be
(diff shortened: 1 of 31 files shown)


romain-grecourt (dismissed):

examples/cors/src/main/java/io/helidon/examples/cors/Main.java
```diff
@@ -65,14 +66,12 @@ public static void main(final String[] args) {
...
-    static void routing(HttpRouting.Builder routing, Config config) {
-
-        GreetService greetService = new GreetService(config);
+    static void routing(HttpRouting.Builder routing) {
+        Config config = Config.global();
 
         // Note: Add the CORS routing *before* registering the GreetService routing.
-        routing.register("/greet", corsSupportForGreeting(config), greetService)
+        routing.register("/greet", corsSupportForGreeting(config), new GreetService())
                 .addFeature(ObserveFeature.create());
```
tjquinno, line 75: As written, this code fetches the global config and passes it to `corsSupportForGreeting` whereas the greet service  invokes `Config.global()` itself to get config. Just to emphasize this new division of labor, how about if `corsSupportForGreeting` also fetched `Config.global()` for itself, rather than having this code fetch it and pass the config as a parameter?

examples/metrics/filtering/se/src/main/java/io/helidon/examples/metrics/filtering/se/Main.java
```diff
@@ -100,7 +100,7 @@ static void routing(HttpRouting.Builder routing, Config config, MetricsConfig.Bu
         MetricsObserver metrics = MetricsObserver.builder()
                 .metricsConfig(metricsConfigBuilder)
                 .build();
-        GreetService greetService = new GreetService(config, meterRegistry);
+        GreetService greetService = new GreetService(meterRegistry);
```
tjquinno, line 103: Re: line 97 above - For consistency and emphasis (of not passing config as parameters, for the most part) do we want to remove the `config` parameter from `routing`?

tjquinno (commented):
A couple questions.

examples/cors/src/main/java/io/helidon/examples/cors/Main.java
barchetta: Done

examples/metrics/filtering/se/src/main/java/io/helidon/examples/metrics/filtering/se/Main.java
barchetta: Done

tjquinno (approved):

tvallin (approved):

PR #7657, opened 2023-09-24
4.x: double-slash URI issue #7474

### LINKED ISSUE #7474: 4.x: HttpClient first URI segment is omitted if path starts with //
## Environment Details
* Helidon Version: 4.0.0-M1
* Helidon SE 
----------

When the path of a URI starts with `//` the first segment is omitted.
E.g. `http://localhost:8080//foo/bar` will make a request to `/bar` !

This is awkward as it is common to have a trailing slash in a URI (E.g. `http://localhost:8080/`), resolving that URI without normalizing can lead to a path that starts with a double slash.

E.g.

```java
String uri = "http://localhost:8080/"
String baseUri = uri + "/foo/bar";
```

```java
URI uri = URI.create("http://localhost:8080/")
String baseUri1 = uri + "/foo/bar"; // bad
URI baseUri2 = uri.resolve("/foo/bar"); // good
String baseUri3 = uri.normalize() + "/foo/bar"; // good
```

---

Reproducer:

```java
        .routing(r -> r.any("/{+}", (req, res) -> res.send(req.requestedUri().path().toString())))
        .build()
        .start();
try {
    Http1Client client = Http1Client.create();
    String baseUri = "http://localhost:" + server.port();
    System.out.println(client.get(baseUri + "//foo/bar").request(String.class).entity());
    // prints /bar
    System.out.println(client.get(baseUri + "/foo/bar").request(String.class).entity());
    // prints /foo/bar
} finally {
    server.stop();
}
```

---

Note that I ran into this while writing a unit test that retrieved the URI for given socket:
```java
@BeforeAll
static void init(@Socket("backend") URI uri) {
    // uri has a trailing /
}
```

danielkec:
Fixes #7474

### DIFF
--- a/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
@@ -126,8 +126,10 @@ class UriPathNoParam implements UriPath {
         }
 
         if (doubleSlash == 0) {
-            // RFC2396 net_path
-            // URI.create("//foo/bar").getPath() --> "/bar"
+            /*
+            RFC2396 - net_path starts with //, that would lead to loosing first path segment.
+            example: URI.create("//foo/bar").getPath() --> "/bar"
+            */
             rawPath = rawPath.substring(1);
         }
 
--- a/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
@@ -73,7 +73,7 @@ class UriPathNoParam implements UriPath {
     @Override
     public String path() {
         if (decodedPath == null) {
-            decodedPath = decode(rawPath);
+            decodedPath = decode(rawPath, false);
         }
         return decodedPath;
     }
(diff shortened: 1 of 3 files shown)


common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
```diff
@@ -121,10 +121,16 @@ private static String decode(String rawPath) {
...
         int doubleSlash = rawPath.indexOf("//");
 
-        if (percent == -1 && doubleSlash == -1 && dot == -1) {
+        if (!validate && percent == -1 && doubleSlash == -1 && dot == -1) {
             return rawPath;
         }
 
+        if (doubleSlash == 0) {
+            // RFC2396 net_path
+            // URI.create("//foo/bar").getPath() --> "/bar"
```
tomas-langer, line 130: wrong comment, good code I think

tomas-langer (changes requested):

common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
danielkec: Made better comment

tomas-langer (approved):

romain-grecourt (approved):

PR #7664, opened 2023-09-25
4.x: demonstrate proper use of GlobalConfig in archetypes

### LINKED ISSUE #7628: 4.x: demonstrate proper use of GlobalConfig in examples and archetypes

tvallin:
### Description

fix #7628

### Documentation

no impact


barchetta (approved):
There are some code indentation issues with the generated projects. I have created an issue to track that: #7665

PR #7666, opened 2023-09-25
Make SE metrics default scope application instead of nothing

### LINKED ISSUE #7658: 4.x SE meters should have default scope of `application` instead of no scope
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP SE
* JDK version:
* OS:

----------

## Problem Description
In the new Helidon metrics API, each meter can have a different scope (formerly called "registry type" in 3.x) or no scope. Currently, if the developer creates a meter without setting the scope explicitly or without setting the default scope in the `MetricsConfig` then the meter has no scope.

This is different from 3.x (in which the default scope was always `application`.) 

For ease of migration, the new metrics API in SE should use `application` as the default scope (unless the developer changes it using `MetricsConfig`).

tjquinno:
### Description
Resolves #7658 

Originally, the neutral Helidon metrics API (SE) assigned no default scope to new meters unless the developer set the default scope to a value. This PR changes that so the "default default" scope is `application` instead of nothing. 

Developers can still assign any specific scope to an individual meter when registering it and can also assign a different default scope (or clear the "default default" scope) using the `ScopingConfig` within `MetricsConfig`. 

### Documentation

Doc  impact: will be included in the metrics doc rewrite #7381

### DIFF
--- a/archetypes/helidon/src/main/archetype/common/observability.xml
+++ b/archetypes/helidon/src/main/archetype/common/observability.xml
@@ -385,13 +385,13 @@ allRequests_total 0.0
         assertThat(response.status().code(), is(200));
 
         String output = response.as(String.class);
-        String expected = MetricsService.ALL_GETS_TIMER_NAME + "_seconds_count " + 2;
+        String expected = MetricsService.ALL_GETS_TIMER_NAME + "_seconds_count{scope=\"application\",} " + 2;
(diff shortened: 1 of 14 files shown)


barchetta (commented):
Can you also enable the disabled tests in the examples in `examples/metrics` ?

barchetta (approved):

PR #7677, opened 2023-09-27
4.x: Upgrade ASM version used by  plugins to 9.5

barchetta:
### Description

Upgrades the version of ASM we use for compiler, javadoc, surefire and shade plugins to 9.5

### Documentation

No impact


ljnelson (approved):
LGTM as far as it goes; may end up needing it on other plugins as well until they catch up with JDK 21.

PR #7690, opened 2023-09-29
4.x: Move webserver/benchmark/jmh to tests/benchmark/jmh

barchetta:
### Description

Move `webserver/benchmark/jmh` to `tests/benchmark/jmh`

### Documentation

No impact


tomas-langer (approved):
I was expecting this to be moved to `webserver/tests`, but I think this is acceptable as well.

Issue #7698, opened 2023-10-02
[4.x] TlsManager enhancement for client-side


trentjeff:
https://github.com/helidon-io/helidon/issues/7279 introduced the TlsManager.

This issue will track enhancement to include TlsManager for WebClient and RestClient, and to ensure that there is a unique TlsManager per unique configuration so that we prevent the proliferation of TlsManager instances within the same JVM on the client side.

trentjeff:
See also https://gist.github.com/trentjeff/85a66a0ca61349c26af1e0426573e1b3

PR #7699, opened 2023-10-02
Partial fix for issue #7698 - TlsManager support for client-side

trentjeff:
This provides an enhancement whereas on the same JVM, using the same config bean, you will get the same `TlsManager` instance. This is to prevent a proliferation of manager instances in client-side usages - which will officially be coming in a separator PR later.

Base issue is #7698

### DIFF
--- a/common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerCache.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerCache.java
@@ -16,24 +16,40 @@
 
 package io.helidon.common.tls.spi;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Function;
 
 import io.helidon.common.tls.TlsManager;
 
 class TlsManagerCache {
-    private static final ConcurrentHashMap<Object, TlsManager> CACHE = new ConcurrentHashMap<>();
+    private static final ReentrantLock LOCK = new ReentrantLock();
+    private static final Map<Object, TlsManager> CACHE = new HashMap<>();
 
     private TlsManagerCache() {
     }
 
-    @SuppressWarnings("unchecked")
     static <T> TlsManager getOrCreate(T configBean,
                                       Function<T, TlsManager> creator) {
         Objects.requireNonNull(configBean);
         Objects.requireNonNull(creator);
-        return CACHE.computeIfAbsent(configBean, (Function<Object, TlsManager>) creator);
+        LOCK.lock();
+        try {
+            TlsManager manager = CACHE.get(configBean);
+            if (manager != null) {
+                return manager;
+            }
+
+            manager = creator.apply(configBean);
+            Object existing = CACHE.put(configBean, manager);
+            assert (existing == null);
+
+            return manager;
+        } finally {
+            LOCK.unlock();
+        }
     }
 
 }
--- a/common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerCache.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerCache.java
@@ -28,8 +28,8 @@ class TlsManagerCache {
     private TlsManagerCache() {
     }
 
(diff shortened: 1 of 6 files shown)


common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerCache.java
```diff
@@ -0,0 +1,38 @@
...
+
+import io.helidon.common.tls.TlsManager;
+
+class TlsManagerCache {
+    private static final ConcurrentHashMap<Object, TlsManager> CACHE = new ConcurrentHashMap<>();
+
+    private TlsManagerCache() {
+    }
+
+    static TlsManager getOrCreate(Object configBean,
```
tomas-langer, line 31: The signature should be:
```
static <T> TlsManager getOrCreate(T configBean, Function<T, TlsManager> creator)
```
if this would be the way to go, as that allows you to use a method referenc for the creator

common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerProvider.java
```diff
@@ -24,4 +28,22 @@
...
+    /**
+     * Provides the ability to have a unique {@link TlsManager} per unique {@link Config} instance provided.
+     *
+     * @param config     the raw config instance
+     * @param configBean the config bean instance
+     * @param name       the config bean instance name
+     * @param creator    the creator to apply if not already in cache, which takes the config bean instance
+     * @return the tls manager instance from cache, defaulting to creation from the {@code creator} if not in cache
+     */
+    static TlsManager getOrCreate(Config config,
```
tomas-langer, line 40: why is there a `name` and  `config` parameter, when theser are not used in the code? Please remove them
Also update the signature to what I wrote above

tomas-langer (changes requested):

common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerProvider.java
trentjeff: I was thinking we might need them for future use.  I'll remove them but at least consider that.

trentjeff: pushed

common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerCache.java
trentjeff: pushed

common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerProvider.java
```diff
@@ -24,4 +27,17 @@
  */
 public interface TlsManagerProvider extends ConfiguredProvider<TlsManager> {
 
+    /**
+     * Provides the ability to have a unique {@link TlsManager} per unique {@link Config} instance provided.
+     *
+     * @param configBean the config bean instance
+     * @param creator    the creator to apply if not already in cache, which takes the config bean instance
```
danielkec, line 34: We should either be very specific with this function being invoked inside synchronized block so no IO operations nor prolonged obstruction should happen in it, or synchronize the cache access ourselves with ReentrantLock

common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerProvider.java
```diff
@@ -24,4 +27,17 @@
...
+     * @param <T>        the type of the config bean
+     * @return the tls manager instance from cache, defaulting to creation from the {@code creator} if not in cache
+     */
+    static <T> TlsManager getOrCreate(T configBean,
+                                      Function<Object, TlsManager> creator) {
```
danielkec, line 39: Function parameter is not documented here, either document it is the config bean or use supplier

danielkec (changes requested):

common/tls/src/main/java/io/helidon/common/tls/spi/TlsManagerProvider.java
trentjeff: pushed

danielkec (approved):

PR #7705, opened 2023-10-03
4.x - named routing

### LINKED ISSUE #7702: 4.x - Add ability to configure routing for an existing socket
In 3.x users could define "named" sockets in configuration and set-up routing for those sockets programmatically.

In the current 4.x state (at least in 4.0.0-RC1), the only way to set-up routing for a socket is when creating the socket via the `putSocket` method on the server builder.

The workaround to emulate the 3.x use-case is to re-create the socket:

```
.putSocket("admin", socket -> socket.from(server.sockets().get("admin"))
                        .routing(Main::adminSocket))
```

----------

### Side by side comparison

Given the following `application.yaml`:

```yaml
server:
  port: 8080
  host: "localhost"
  sockets:
    - name: "admin"
      port: 8081
```

In 3.x:

```java
WebServer.builder(createPublicRouting())
    .config(config.get("server"))
    .addNamedRouting("admin", Routing.builder()
    .get("/admin", (req, res) -> res.send("Hello Admin!!"))
    .build())
```

In 4.x:

```java
WebServer.builder()
        .config(config.get("server"))
        .update(builder -> builder
                .putSocket("admin", socket -> socket.from(builder.sockets().get("admin"))
                        .routing(routing -> routing.get("/admin", (req, res) -> res.send("Hello Admin!!")))))
        .build()
```

romain-grecourt:
### Description

- Add new `.routing` variants in `WebServerConfig.Builder`
- Update `LoomServer` to use the new named routings
- Update multiport example to use the new API
- Fix multiport example README.md and dependencies (`/observe` and dependencies)

Fixes #7702

---

### Documentation

One can now register routing for a named socket created via `.config(config.get("server"))`. 

Before:
```java
server.routing(Main::publicRouting)
        // Add a set of routes on the named socket "admin"
        .putSocket("admin", socket -> socket.from(server.sockets().get("admin"))
                .routing(Main::adminSocket))
        // Add a set of routes on the named socket "private"
        .putSocket("private", socket -> socket.from(server.sockets().get("private"))
                .routing(Main::privateSocket));
```

After:
```java
server.routing(Main::publicRouting)
        .routing("admin", Main::adminSocket)
        .routing("private", Main::privateSocket);
```

### DIFF
--- a/examples/webserver/multiport/README.md
+++ b/examples/webserver/multiport/README.md
@@ -2,7 +2,7 @@
 
 It is common when deploying a microservice to run your service on
 multiple ports so that you can control the visibility of your
-service's endpoints. For example you might want to use three ports:
+service's endpoints. For example, you might want to use three ports:
 
 - 8080: public REST endpoints of application
 - 8081: private REST endpoints of application
@@ -16,7 +16,7 @@ as described above.
 
 The ports are configured in `application.yaml` by using named sockets.
 
-Seperate routing is defined for each named socket in `Main.java`
+Separate routing is defined for each named socket in `Main.java`
 
 ## Build and run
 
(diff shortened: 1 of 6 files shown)


tomas-langer (approved):
The API looks much better. The only problem I can see is if somebody combines both methods (we may want to warn/throw an exception for such cases).

Issue #7448, opened 2023-08-24
4.x: Update Hibernate version from 6.1.7 to 6.2.X 


barchetta:
Update Hibernate version from 6.1.7 to 6.2.X

As part of this remember to:
1. Remove exclusion of jandex from `hibernate-core` dependency (this was a work-around during upgrade to Jandex 3)
2. Remove force upgrade of `byte-buddy` (this was a work-around for Java 21 support)

Note that there have been a couple attempts at the upgrade:
- #7433
- #7426

lscoughlin:
can we get this added to the 3.x line as well?

PR #7719, opened 2023-10-04
[4.x] Switches default JPA CDI portable extension to PersistenceExtension from JpaExtension

ljnelson:
This PR "turns on" the "new" JPA machinery that has been lying somewhat dormant in Helidon since 3.x.  The "old" machinery remains, since the new machinery is a very different approach, and there may be production cases where the old machinery needs to be switched back on in case old behavior was being relied upon.

This PR also adds deprecation notices to various classes (most of which are internal anyway but it's good practice).

This PR also ensures that any usage of the `CDISEJtaPlatform` in combination with the ("new machinery") `PersistenceExtension` will cause the appropriate Hibernate-specific connection handling policy to be in effect, even if it isn't present in a `persistence.xml`.

This PR also fixes a setting-the-`autoCommit`-property-on-a-`Connection` bug that may be responsible for blocking #7448. We'll see.

Documentation impact: none.

### DIFF
--- a/integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/CDISEJtaPlatform.java
+++ b/integrations/cdi/hibernate-cdi/src/main/java/io/helidon/integrations/cdi/hibernate/CDISEJtaPlatform.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2021 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2023 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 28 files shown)


ljnelson:
@tjquinno added you as reviewer since you were kind enough to review the "new machinery" back in the day. Obviously check the Github actions etc. and if they're borked don't bother wasting your time; I'll fix whatever breaks first. Any early eyes-on from you would be great.

arjav-desai (approved):
LGTM

ljnelson:
Related to #4928, and may (finally) fix it.

PR #7720, opened 2023-10-04
Upgrade Weld and ClassFileWriter

### LINKED ISSUE #7692: 4.x: upgrade weld and jboss-classfilewriter
* Upgrade jboss-classfilewriter to 1.3.0.Final
* Upgrade Weld to 5.1.1.SP2

It would be good to upgrade jboss-classfilewriter because 1.2.5.Final is getting pretty old. Doing so requires at least Weld 5.1.1 to pick up https://github.com/weld/core/commit/19b0ffbd012f21f53e4aeb5488806fd1740c247d

Note that even after doing these upgrades I'm still hitting an `IllegalAccessException` that was supposed to be fixed by the above Weld commit, so I'm wondering if something needs to be changed in our code as well.

tomas-langer:
### Description
Resolves #7692 

### Documentation

no impact

### DIFF
--- a/microprofile/weld/weld-core-impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
+++ b/microprofile/weld/weld-core-impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
@@ -233,7 +233,7 @@ public class ProxyFactory<T> implements PrivilegedAction<T> {
     static String getProxyName(String contextId, Class<?> proxiedBeanType, Set<? extends Type> typeClosure, Bean<?> bean) {
         TypeInfo typeInfo = TypeInfo.of(typeClosure);
         final String className;
-        org.jboss.weld.bean.proxy.ProxyFactory.ProxyNameHolder holder;
+        ProxyNameHolder holder;
         if (typeInfo.getSuperClass() == Object.class) {
             // for classes that do not have an enclosing class, we want the super interface to be first
             if (proxiedBeanType.getEnclosingClass() == null) {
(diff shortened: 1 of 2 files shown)


microprofile/weld/weld-core-impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
```diff
@@ -632,7 +632,10 @@ private Class<T> createProxyClass(Class<?> originalClass, String proxyClassName)
 
     private ClassFile newClassFile(String name, int accessFlags, String superclass, String... interfaces) {
         try {
-            return new ClassFile(name, accessFlags, superclass, interfaces);
+            // We need to use a (non-deprecated) method that avoids instantiating DefaultClassFactory.INSTANCE
+            // If that happens, we will have module accessibility issues and the need to use --add-opens clausules
```
ljnelson, line 636: Typo: `clausules`

ljnelson (changes requested):

microprofile/weld/weld-core-impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java
tomas-langer: This is an exact copy from the source code of Weld, I want to keep our changes located in a single section in the file. So I will not fix this.

tomas-langer: see https://github.com/weld/core/blob/d6729a0bd53b85d7dc3177405a6da8724ec80bb5/impl/src/main/java/org/jboss/weld/bean/proxy/ProxyFactory.java#L517

ljnelson (approved):
Approving per request

PR #7731, opened 2023-10-05
4.x: Fix several issue in archetype and add `native-image.properties` to generated projects

### LINKED ISSUE #7714: Up level `Main.java.mustache` as it is generated for SE and MP flavor
## Environment Details
* Helidon Version: 4.x
* Helidon SE and Helidon MP
* JDK version: 21
* OS:

----------

## Problem Description

Up level `Main.java.mustache` to `archetype/common`

## Steps to reproduce
n/a

### LINKED ISSUE #7717: 4.x: add native-image.properties to archetypes
With the new version of graalVM native-image, we need to register all bean types to be initialized at build time (as they end up in the image heap). 
To support this, we should generated the following file:
`src/main/resources/META-INF/groupId/artifactId/native-image.properties`
with the following content:
```
Args=--initialize-at-build-time=packageName
```

`groupId`: maven groupId
`artifactId`: maven artifactId
`packageName`: package where we generate beans (such as JAX-RS resources), if multiple packages, then the top most package

### LINKED ISSUE #7722: 4.x: Archetypes - update MP Main class javadoc to document the module main class requirement
#7226 introduced module-info in the archetypes for all flavors.

Since [maven-jar-plugin:3.1.2](https://issues.apache.org/jira/browse/MJAR-238) `jar` is invoked with `-main-class` (see [usage](https://docs.oracle.com/en/java/javase/21/docs/specs/man/jar.html#operation-modifiers-valid-only-in-create-and-update-modes)), which sets the [ModuleMainClass](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.7.27)  module attribute.

However, when the module main class is set, the main class must be in a package that is either exported or opened by the module.

This means that when using the module system we can't reference a built-in main class, instead we have to create one in the module with a simple implementation that calls the built-in one.

#7711 added such Main class to MP archetypes. This issue is about documenting the new requirement in the javadoc of the generated MP main class:
- Tell users that the class is needed when using the module system
- Give a decent explanation of the requirement

See explanations below for the requirements of a module main class.

---

When using `jar --main-class`, the `jar` tool creates a binary form `ModuleDescriptor` that sets the following module attributes:
- [ModuleMainClass](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.7.27) with the value of `--main-class`
- [ModulePackages](https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.7.26) with the list of packages exported or opened by the module

The module descriptor is then validated with [ModuleDescriptor.read](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/module/ModuleDescriptor.html#read(java.io.InputStream)):
> This method works exactly as specified by the 2-arg [read](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/module/ModuleDescriptor.html#read(java.io.InputStream,java.util.function.Supplier)) method with the exception that a package finder is not used to find additional packages when the module descriptor read from the stream does not indicate the set of packages.

The 2-arg [ModuleDescriptor.read](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/module/ModuleDescriptor.html#read(java.io.InputStream,java.util.function.Supplier)) states:
> The set of packages that the packageFinder returns must include all the packages that the module exports, opens, as well as the packages of the service implementations that the module provides, and the package of the main class (if the module has a main class

This effectively means that the module main class must be in a package that is either exported or opened by the module.

If the module is not valid an [InvalidModuleDescriptorException](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/module/InvalidModuleDescriptorException.html) is thrown, and the exception message is displayed by `jar`. See the corresponding `jar` [source code](https://github.com/openjdk/jdk/blob/jdk-21%2B35/src/jdk.jartool/share/classes/sun/tools/jar/Main.java#L2111).

E.g.

```
jar: Package io.helidon missing from ModulePackages class file attribute
```

---

Some other interesting bits of information:

[ModuleDescriptor.packages()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/module/ModuleDescriptor.html#packages()):

> The set of packages includes all exported and open packages, as well as the packages of any service providers, and the package for the main class.

[ModulePackages]( https://docs.oracle.com/javase/specs/jvms/se21/html/jvms-4.html#jvms-4.7.26):

> The ModulePackages attribute indicates all the packages of a module that are exported or opened by the Module attribute, as well as all the packages of the service implementations recorded in the Module attribute. The ModulePackages attribute may also indicate packages in the module that are neither exported nor opened nor contain service implementations.

tvallin:
### Description

fix #7714
fix #7717
fix #7722

### Documentation

no impact


archetypes/helidon/src/main/archetype/mp/common/common-mp.xml
```diff
@@ -81,6 +81,34 @@
...
             </list>
+            <value key="main-class-javadoc"><![CDATA[
+/**
+ * Main entry point of the application.
+ * This class is required when using modules with Helidon MP.
+ * Since version 3.1.2, the maven-jar-plugin set module attribute such as `ModuleMainClass`. The main class must be
+ * in a package that is either exported or opened by the module. This means we cannot reference a built-in class and
+ * therefore this Main class is created with a simple implementation that calls the built-in one.
+ * See the Documentation for `--main-class` option used by maven-jar-plugin:
+ * <a href="https://docs.oracle.com/en/java/javase/21/docs/specs/man/jar.html#operation-modifiers-valid-only-in-create-and-update-modes">documentation</a>
```
romain-grecourt, line 92:
```suggestion
Note that this class is required when using modules as the module main class must be in a package that is either exported or opened by the module. This class provides a proper module module main class and simply calls {@link io.helidon.Main#main(String[])}.
```

romain-grecourt (approved):

PR #7741, opened 2023-10-06
[3.x] Fixes an issue where autoCommit is not restored properly in certain edge cases; ensures related proper Hibernate JTA settings are set by default

ljnelson:
A partial backport of #7719. 

Documentation impact: none


arjav-desai (approved):
LGTM

PR #7742, opened 2023-10-06
[4.x] Upgrades Hibernate to version 6.3.1.Final

ljnelson:
This PR upgrades Hibernate to verison 6.3.1.Final and adjusts certain `hibernate.properties` files found in integration tests to comply with Hibernate's ongoing series of settings removals.

Documentation impact: Hibernate 6.3.x has (somewhat oddly) removed support for the `none` bytecode provider, so any instances that end users have of `hibernate.properties` must remove any lines that resemble the following:
> hibernate.bytecode.provider=none

This may have an impact on native image.

More specifics: the `none` bytecode provider still exists, but throws an exception when its "reflection optimizer" is accessed. There used to be a guard in the Hibernate source code that would not attempt to access the reflection optimizer if the `none` bytecode provider was selected, but [that guard was removed](https://github.com/hibernate/hibernate-orm/pull/7197/files#diff-2907b1254eed5e31600371471560e28b875699a5e6266345062ba6f21aead4f9L290-L292), so any use of the `none` bytecode provider will result in an exception.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -63,7 +63,7 @@
         <version.lib.h2>2.2.220</version.lib.h2>
         <version.lib.hamcrest>1.3</version.lib.hamcrest>
         <version.lib.handlebars>4.3.1</version.lib.handlebars>
-        <version.lib.hibernate>6.1.7.Final</version.lib.hibernate>
+        <version.lib.hibernate>6.3.1.Final</version.lib.hibernate>
         <version.lib.hibernate-validator>7.0.2.Final</version.lib.hibernate-validator>
         <version.lib.hikaricp>5.0.1</version.lib.hikaricp>
         <version.lib.hystrix>1.5.18</version.lib.hystrix>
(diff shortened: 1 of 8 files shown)


ljnelson:
Hibernate 6.2.7 [changes the way column ordering is performed](https://stackoverflow.com/questions/76963362/did-hibernate-6-2-x-change-the-way-it-generates-ddl) so I will have to change all the various tests etc. to specify columns in their DDL.

tvallin:
See #7448

dependencies/pom.xml
```diff
@@ -63,7 +63,7 @@
         <version.lib.h2>2.2.220</version.lib.h2>
         <version.lib.hamcrest>1.3</version.lib.hamcrest>
         <version.lib.handlebars>4.3.1</version.lib.handlebars>
-        <version.lib.hibernate>6.1.7.Final</version.lib.hibernate>
+        <version.lib.hibernate>6.3.1.Final</version.lib.hibernate>
```
barchetta, line 66: After this upgrade we should no longer need to force upgrade `byte-buddy`. Could you verify that and then remove version management of byte-buddy from this file as well as top level `pom.xml` and `applications/mp/pom.xml`

ljnelson: Done

barchetta (approved):
LGTM

PR #7769, opened 2023-10-11
4.x: archetype - module-info opt-in

### LINKED ISSUE #7735: 4.x: archetype - module-info opt-in
## Environment Details
* Helidon Version: 4.0.0-RC1

## Problem Description

Create a new option that conditionally adds a module-info and a module main class(mp only).

This option should go under the `Package and Deployment` step:

```properties
id=jpms
name=Module support (JPMS)
description=Add a module-info
```

When flavor=mp, the description should be:

```properties
description=Add a module-info and a module main class
```

Add `jpms=false` as a preset to all app-types (other than custom).

This needs to be back-ported to 3.x as well.

tvallin:
### Description

fix #7735

By default, the `module-info.java` is not generated.

### Documentation

no impact

### DIFF
--- a/archetypes/helidon/src/main/archetype/common/packaging.xml
+++ b/archetypes/helidon/src/main/archetype/common/packaging.xml
@@ -92,6 +92,36 @@ kubectl delete -f app.yaml
                     </templates>
                 </output>
             </boolean>
+            <boolean id="jpms-se"
+                     name="Module support (JPMS)"
+                     description="Add a module-info"
+                     default="false"
+                     optional="true"
+                     if="${flavor} == 'se'">
+                <output>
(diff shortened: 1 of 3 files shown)


archetypes/helidon/src/main/archetype/common/packaging.xml
```diff
@@ -92,6 +92,36 @@ kubectl delete -f app.yaml
                     </templates>
                 </output>
             </boolean>
+            <boolean id="jpms-se"
```
romain-grecourt, line 95: Move these inputs under each flavor so that the ids are flavor less.

tvallin: Done

tvallin:
This is done by #7775

PR #7572, opened 2023-09-12
4.x: metrics and openapi endpoints should be authorized by default

tvallin:
### Description

fix #7439

### Documentation

no impact

PR #7777, opened 2023-10-11
4.x: Server Features

### LINKED ISSUE #7629: TracingFeature as an Observer
As of 4.0.0-M2, tracing is not part of the "ObserveFeature" providers, it is a separate ad-hoc feature.

Tracing logically belongs in the category of observability, it is odd to set up routing with both ObserveFeature and TracingFeature:

```java
routing.addFeature(TracingFeature.create(tracer))
            .addFeature(ObserveFeature.create())
```

We should implement TracingFeature as an "Observer":
- consistent with other observability related features (health, metrics)
- discoverable off class-path

### LINKED ISSUE #7630: HttpFeature discovery mechanism
The goal of this issue is to discuss and maybe implement a generic inversion of control for HttpFeature.

As of 4.0.0-M2, only `ObserveFeature` provides a way to discover automatically register HttpFeature in the routing. However there are still some Helidon features that need to be registered manually:

- OpenApiFeature
- SecurityFeature
- AccessLogFeature

This would re-enforce the meaning of features, they are pluggable and can be activated simply by having them on the classpath. 

```java
    // user only
    routing.register("/greet", GreetService::new);
}
```

This discovery mechanism could be built-in `HttpRouting.Builder` and enabled by default, with an option to disable the discovery:

```java
    routing.useSystemServices(false)  // 3.x style, what you see is what you get
            .register("/greet", GreetService::new)
}
```

---

A filtering mechanism would be needed, e.g. If a user wants to only load the Helidon features dynamically and avoid all others.

An implicit override mechanism, to programmatically customize features that would otherwise be registered dynamically.

E.g.
```java
    healthCheck = new TestHealthCheck("message");
    HealthObserver health = HealthObserver.create(healthCheck);
    routing.register("/greet", GreetService::new)
            .addFeature(ObserveFeature.builder()
                           .addObserver(health));
}
```

### LINKED ISSUE #7782: [4.x] NPE in HelidonFeatures at startup on uberjar'ed & AOT http-server (clj)
## Problem Description

This is from a "simple" hello world example using https://github.com/mpenet/mina (that depends on RC1), built as a uberjar, AOT'ed.

That build of mina has the following dependencies

(lots of clj/EDN coming, sorry)

```edn
 :deps {org.clojure/clojure {:mvn/version "1.12.0-alpha1"}
        com.github.strojure/zmap {:mvn/version "1.3.26"}
        io.helidon.http/helidon-http {:mvn/version "4.0.0-RC1"}
        io.helidon.webserver/helidon-webserver {:mvn/version "4.0.0-RC1"}
        io.helidon.webserver/helidon-webserver-websocket {:mvn/version "4.0.0-RC1"}
        io.helidon.webserver/helidon-webserver-http2 {:mvn/version "4.0.0-RC1"}}
```

We see the following npe in the logs on startup. 

```
Oct 12, 2023 7:37:07 AM io.helidon.common.features.HelidonFeatures features
INFO: Helidon SE 4.0.0-RC1 has no registered features
Oct 12, 2023 7:37:07 AM io.helidon.webserver.ServerListener start
INFO: [0x4bf77642] http://0.0.0.0:8080/ bound for socket '@default'
Exception in thread "features-thread" java.lang.NullPointerException: Cannot invoke "java.util.Set.forEach(java.util.function.Consumer)" because the return value of "java.util.Map.get(Object)" is null
	at io.helidon.common.features.HelidonFeatures.features(HelidonFeatures.java:217)
	at io.helidon.common.features.HelidonFeatures.lambda$print$0(HelidonFeatures.java:92)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Oct 12, 2023 7:37:07 AM io.helidon.webserver.LoomServer startIt
INFO: Started all channels in 82 milliseconds. 1245 milliseconds since JVM startup. Java 21+35-2513
```

Looking at the stacktrace it seems there's a nil guard missing here (well, likely before, but it pops there) https://github.com/helidon-io/helidon/blob/release-4.0.0-RC1/common/features/features/src/main/java/io/helidon/common/features/HelidonFeatures.java#L217

Steps to repro (copied from https://github.com/mpenet/mina/issues/11) :

- deps.edn:
  ```edn
  {:paths ["src"]
   :deps {com.s-exp/mina {:mvn/version "0.1.23"}}
   :aliases {:build {:deps {io.github.clojure/tools.build {:git/tag "v0.9.6" :git/sha "8e78bcc"}}
                     :ns-default build}}}

  ```
- build.clj:
  ```clojure
  (ns build
    (:require [clojure.tools.build.api :as b]))
  
  (def uber-file "foo.jar")
  
  (defn clean
    [_]
    (b/delete {:path "target"})
    (b/delete {:path uber-file}))
  
  (defn uber
    [_]
    (let [basis (b/create-basis {:project "deps.edn"})
          class-dir "target/classes"
          src-dirs ["src"]]
      (clean nil)
      (b/write-pom {:class-dir class-dir
                    :lib 'bar/foo
                    :version "0.1.0"
                    :basis basis
                    :src-dirs src-dirs})
      (b/copy-dir {:src-dirs (conj src-dirs "resources")
                   :target-dir class-dir})
      (b/compile-clj {:basis basis
                      :src-dirs src-dirs
                      :class-dir class-dir
                      :ns-compile '[foo.main]
                      :java-cmd (or (System/getenv "JAVA_CMD") "java")
                      :compile-opts {:direct-linking true}})
      (b/uber {:class-dir class-dir
               :uber-file uber-file
               :basis basis
               :main 'foo.main})))

  ```
- src/foo/main.clj:
  ```clojure
  (ns foo.main
    (:require
     [s-exp.mina :as mina])
    (:gen-class))
  
  (defn -main
    [& _args]
    (mina/start! (fn [_req]
                   {:status 200
                    :body "Hello"})
                 {:port 8080
                  :host "0.0.0.0"}))

  ```
- build the uberjar: `clojure -T:build uber`
- Run the jar: `java -jar foo.jar` to see the NPE occur but things work as expected. Running without AOT with `clojure -M -m foo.main` doesn't seem to have this issue.

## Environment Details
* Helidon Version: 4.0.0-RC1
* JDK version: openjdk 21 2023-09-19
* OS: Darwin wintermute2.lan 23.0.0 Darwin Kernel Version 23.0.0: Fri Sep 15 14:41:43 PDT 2023; root:xnu-10002.1.13~1/RELEASE_ARM64_T6000 arm64 

----------

tomas-langer:
Resolves #7630 
Resolves #7629
Resolves #7782 

### Description
Introduces server features
Changes:

* WebServer uses builders of built routing (to be able to update it from features)
* WebServer no longer falls back to the default routing for additional sockets (in Helidon SE)
* Introduced `ServerFeature` concept, server feature can access routing builders for all sockets on WebServer
* SecurityFeature is now a WebServer feature
* ContextFeature is now a WebServer feature
* ObserveFeature is now a WebServer feature
* OpenApiFeature is now a WebServer feature
* CorsFeature is a new WebServer feature
* TracingFeature is now an observability feature
* Features use common config dependency - can still pass `io.helidon.Config` instance to them, only changes in SPI
* Metrics in SE now require user in `observe` role, or `metrics.permit-all` set to `true`, otherwise 403 is returned
* OpeanAPI in SE now requires user in `openapi` role, or `openapi.permit-all` set to `true`, otherwise 403 is returned

### Documentation

This introduces a big change to way we can set up server - allows us to register features that need access to multiple sockets and their routing

### Todos

* archetypes are being improved, current code generated is not OK
* update docs


examples/metrics/exemplar/src/main/resources/application.yaml
```diff
@@ -22,3 +22,5 @@ server:
   host: 0.0.0.0
 tracing:
   service: "hello-world"
+metrics:
```
romain-grecourt, line 25: Is `permit-all` required ? If there is no security, users shouldn't be required to specify `permit-all`.

openapi/tests/gh-5792/src/test/java/io/helidon/openapi/tests/yamlparsing/SnakeYAMLV1Test.java
```diff
@@ -39,10 +41,16 @@ class SnakeYAMLV1Test {
...
+        server.addFeature(OpenApiFeature.builder()
+                                  .staticFile("target/test-classes/petstore.yaml")
+                                  .permitAll(true)
+                                  .build());
+    }
     @SetUpRoute
     static void routing(HttpRouting.Builder routing) {
-        routing.addFeature(OpenApiFeature.builder()
-                                   .staticFile("target/test-classes/petstore.yaml"));
+        routing.get("/greet", (req, res) -> res.send("Hi!"));
```
romain-grecourt, line 53: The `routing` method is not needed any more right ? that `/greet` endpoint seems to be unused.

romain-grecourt (commented):
Great work.
---
However, I have some concerns about the new `permit-all` requirement.

This is not good for end-user experience. If security is not enabled, server features should work out of the box, without any mandatory configuration.

openapi/tests/gh-5792/src/test/java/io/helidon/openapi/tests/yamlparsing/SnakeYAMLV1Test.java
tomas-langer: this is intentionally done so you cannot access the endpoint even if you do not have security configured. To protect people who forget they are exposing an endpoint with data that could be missused.

So you have two options:
1.  configure real security and protect the endpoint
2. explicitly mark the endpoint to `permit-all`, so you make an action of affirmation of this problem
3.

examples/metrics/exemplar/src/main/resources/application.yaml
tomas-langer: this is intentionally done so you cannot access the endpoint even if you do not have security configured. To protect people who forget they are exposing an endpoint with data that could be missused.

So you have two options:
1.  configure real security and protect the endpoint
2. explicitly mark the endpoint to `permit-all`, so you make an action of affirmation of this problem

openapi/tests/gh-5792/src/test/java/io/helidon/openapi/tests/yamlparsing/SnakeYAMLV1Test.java
tomas-langer: Current implementation of JUnit extensions requires the set up route method to create an instance of the client with correct routing (including server features).
To change this behavior, we would need to refactor that approach. I do not think I can do it quick enough.
Would it be better to have the method empty in the test class?

examples/metrics/exemplar/src/main/resources/application.yaml
barchetta: I am confused. I thought the decision was: if security is not configured for the application then the observability endpoints do not require authorization (and do not require permit-all). If you configure security for the application then the endpoints require authorization -- and you must add permit-all if you do not want the endpoints secured.

That's the behavior we agreed to and that was implemented  in PR #7572.

tomas-langer: The reason we have common-security is exactly to enforce security even when security is not present, for cases where you could shoot yourself in the leg.
So config, log can NEVER be exposed without authorization, unless you explicitly configure Helidon to allow it

openapi/tests/gh-5792/src/test/java/io/helidon/openapi/tests/yamlparsing/SnakeYAMLV1Test.java
romain-grecourt: I'd rather have it empty.

tomas-langer: Fixed

romain-grecourt (approved):
We will address the requirement for permit-all in a separate PR. Nice work.

PR #7778, opened 2023-10-11
4.x: Remove exclusion of jandex dependency now that hibernate has been upgraded

barchetta:
### Description

Removes dependency exclusion of `org.jboss:jandex` because it is no longer needed after hibernate upgrade (#7742).

Completes fix for #7448

### Documentation

No impact


ljnelson (approved):
LGTM

PR #7786, opened 2023-10-13
Mysql tests with testcontainers.

Tomas-Kraus:
Think we are not far from being done. I'd like to see all tests output and finish code cleanup 1st.


romain-grecourt:
Work was completed in #9104

PR #7806, opened 2023-10-16
DOCS: small updates for upgrade term for migration

ljamen:
### Documentation

If no doc impact: doc impact


ljamen:
fixed in #7715

PR #7829, opened 2023-10-18
4.x: Initial support for proxy protocol V1 and V2

spericas:
## Description

This PR includes an implementation of the proxy protocol V1 and V2 as described here https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt. This feature can be enabled on a listener configuration:

```
    /**
     * Enable support for proxy protocol for this socket.
     * Default is {@code false}.
     *
     * @return proxy support status
     */
    @Option.Default("false")
    boolean enableProxyProtocol();
```

Once enabled, it is disabled by default, the proxy protocol header MUST be present in the connection (only for the first request if multiple ones are sent over a long-lived connection). The feature is supported for HTTP/1.1 and HTTP2 and can also be used in combination with TLS (proxy protocol header is processed before the SSL handshake).

The proxy protocol version, either V1 or V2, is automatically discovered by inspecting the first few bytes of a connection. The data collected from the protocol header can be obtained from a `ServerRequest` as shown below:

```
ServerRequest request = ...
ProxyProtocolData data = request.proxyProtocolData().ifPresent(ppd -> ...);
```

It addition, a subset of the proxy protocol data is made available as the HTTP headers `X-Forwarded-For` and `X-Forwarded-Port`. These headers can also be used in combination with `RequestedUriDiscoveryContext` when configured to inspect the aforementioned headers.

Other than unit and integration tests, both the V1 and V2 protocol headers have been manually tested to be compatible with Wireshark's decoder.

## Documentation

I shall create a new follow-up issue to document this new features, if and when it is merged.


tjquinno:
The revsied `Http1Connection` and `Http2Connection` classes set the non-standard (but oft-used) `X-Forwarded-For` and `X-Forwarded-Port` headers but not the standard `Forwarded` header.

Should the code allow users to configure which family of header to use (maybe both??) and then set the corresponding header(s)? 

I see that the proxy spec (thanks for the link) uses the `X-Forwarded-*` headers in the text but also mentions in passing `Forwarded`.  It's not clear from my quick read whether the proxy spec _requires_ `X-Forwarded-*` or happened to use that as an example.

Given that our requested URI support allows users to configure one or the other or both families of headers for _that_ feature, should the proxy support do the same?

spericas:
> The revsied `Http1Connection` and `Http2Connection` classes set the non-standard (but oft-used) `X-Forwarded-For` and `X-Forwarded-Port` headers but not the standard `Forwarded` header.
> 
> Should the code allow users to configure which family of header to use (maybe both??) and then set the corresponding header(s)?
> 
> I see that the proxy spec (thanks for the link) uses the `X-Forwarded-*` headers in the text but also mentions in passing `Forwarded`. It's not clear from my quick read whether the proxy spec _requires_ `X-Forwarded-*` or happened to use that as an example.
> 
> Given that our requested URI support allows users to configure one or the other or both families of headers for _that_ feature, should the proxy support do the same?
> 
> Or, perhaps, just _always_ set both `Forwarded` and the `X-Forwarded-*` family?

It's obviously possible to support both. I wanted to the minimum without consulting config and without creating formatted strings, to avoid addition of more code to the request processing loop. Using the X headers seems more efficient and it is one of the options supported by the existing Uri mechanism.

common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/HexStringDecoder.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Utility method to decode hex strings. For example, "\0x0D\0x0A\0x0D\0x0A" is decoded
+     * as a 4-byte array with hex values 0D 0A 0D 0A.
+     *
+     * @param s string to decode
+     * @return decoded string as byte array
+     */
+    public static byte[] decodeHexString(String s) {
```
tomas-langer, line 33: Couldn't you use `java.util.HexFormat`?

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
```diff
@@ -327,6 +327,15 @@ interface ListenerConfigBlueprint {
      */
     Optional<Context> listenerContext();
 
+    /**
+     * Enable support for proxy protocol for this socket.
```
tomas-langer, line 331: I think it would be nice to have a bit of an explanation of what is meant by proxy prototocol, if there is a spec, maybe a link to it?

webserver/webserver/src/main/java/io/helidon/webserver/ProxyProtocolData.java
```diff
@@ -0,0 +1,130 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.webserver;
+
+/**
+ * Proxy protocol data parsed by {@link ProxyProtocolHandler}.
+ */
+public interface ProxyProtocolData {
```
tomas-langer, line 21: Can we give access to the actual bytes?

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -346,7 +346,8 @@ private void listen() {
                                                     socket,
                                                     serverChannelId,
                                                     router,
-                                                    tls);
+                                                    tls,
```
tomas-langer, line 349: ListenerConfig is available through `listenerContext`, not sue we need to send it separately, probably not really an optimization

tomas-langer (changes requested):

common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/HexStringDecoder.java
spericas: Nice, switch to use HexFormat.

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
spericas: Improved

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
spericas: Fixed

tomas-langer (approved):

PR #7861, opened 2023-10-19
2.x dep upgrades

barchetta:
### Description

Upgrades 
* kafka-clients to 3.6.0
* OCI SDK to 2.66.0
* okhttp3 to 4.12.0
* suppress a jgit false positive

The oktthp3 upgrade required various dependency hacks to resolve dependency convergence issues.

### Documentation

No impact

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -93,10 +93,7 @@
         <version.lib.jsonp-api>1.1.6</version.lib.jsonp-api>
         <version.lib.jsonp-impl>1.1.6</version.lib.jsonp-impl>
         <version.lib.junit>5.7.0</version.lib.junit>
-        <version.lib.kafka>3.4.0</version.lib.kafka>
-        <!-- Force upgrade of snappy. This should be removed once kafka-clients is upgraded -->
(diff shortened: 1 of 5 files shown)


dependencies/pom.xml
```diff
@@ -127,13 +124,14 @@
         <version.lib.narayana>5.12.0.Final</version.lib.narayana>
         <version.lib.netty>4.1.100.Final</version.lib.netty>
         <version.lib.netty-io_uring>0.0.19.Final</version.lib.netty-io_uring>
-        <version.lib.oci>2.60.1</version.lib.oci>
+        <version.lib.oci>2.66.0</version.lib.oci>
```
ljnelson, line 127: If you wanted, you could also remove the by-value import of the contents of their BOM, since it seems to now be valid in version 2.66.0.

barchetta: Hmmm...does not look like it is fixed in 2.66.0? https://github.com/oracle/oci-java-sdk/blob/8d42b2916aeeb3b27384ec8d6eac348f9eaaf170/bmc-bom/pom.xml#L10

ljnelson: Ah, you are right; missed the parent block

danielkec (approved):

PR #7869, opened 2023-10-20
4:x UriQueryEmpty + NoSuchElementException

### LINKED ISSUE #7867: Access to request().query().get() raises UnsupportedOperationException
## Environment Details
* Helidon Version: 4.0.0-RC1
* Helidon SE
* JDK version:
```
openjdk version "21" 2023-09-19
OpenJDK Runtime Environment (build 21+35)
OpenJDK 64-Bit Server VM (build 21+35, mixed mode, sharing)
```
* OS: Linux

----------

## Problem Description

The Javadoc for the `io.helidon.common.uri.UriQuery` interface's `get` method says:

```
    /**
     * Get the first value.
     *
     * @param name name of the parameter
     * @return first value
     * @throws NoSuchElementException in case the name is not present
     */
    String get(String name) throws NoSuchElementException;
```

Unfortunately, the implementation of `UriQueryEmpty` has:

```
  @Override
  public String get(String name) {
      throw new UnsupportedOperationException("Empty query");
  }
```

This means you'll get an `UnsupportedOperationException` instead of the expected `NoSuchElementException` if you're trying to get a parameter from a query that didn't contain any parameters.

## Steps to reproduce

Just try calling `ServerRequest.query().get("x")` for any incoming query that has no parameters.

romain-grecourt:
### Description

Update UriQueryEmpty to throw NoSuchElementException instead of UnsupportedOperationException as documented.

Fixes #7867

### Documentation

No doc impact.

### DIFF
--- a/common/uri/src/main/java/io/helidon/common/uri/UriQueryEmpty.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriQueryEmpty.java
@@ -39,22 +39,22 @@ final class UriQueryEmpty implements UriQuery {
 
     @Override
     public String getRaw(String name) throws NoSuchElementException {
-        throw new UnsupportedOperationException("Empty query");
+        throw new NoSuchElementException("Empty query");
     }
 
     @Override
     public List<String> getAllRaw(String name) {
-        throw new UnsupportedOperationException("Empty query");
+        throw new NoSuchElementException("Empty query");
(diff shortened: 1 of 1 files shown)


tomas-langer (approved):
Yes

barchetta (approved):

PR #7875, opened 2023-10-21
Simplify SE metrics config examples

tjquinno:
### Description

Replace complicated SE metrics examples with simple ones (don't use `server/observe/observers/metrics...` but just `metrics...` which still works.


tjquinno:
Obsolete draft PR.

PR #7889, opened 2023-10-25
3.x Examples validation flow for external examples repo

senivam:
### Description
The validation flow which will be used after examples are moved to the external repository 
### Documentation

GitHub actions change which will build examples from the external repository on any PR that relates to helidon-3.x
The validation is not falling so if the build does not pass (examples vs helidon PR), PR still can be merged but related checks will become red to warn the developer that something went wrong.


senivam:
image

This how it shall look when examples fail but PR can be still merged

romain-grecourt (approved):

PR #7994, opened 2023-11-13
4.x Add a new `addCheck` variant allowing the caller to set the health check name; add tests; revise doc

### LINKED ISSUE #7827: 4.x SE `HealthCheckResponse` no longer supports a name
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP SE
* JDK version:
* OS:

----------

## Problem Description
The `HealthCheckResponse` interface (and its `record` implementation `HealthCheckResponseImpl`) and the associated builder do not support a name.

The result is that developers cannot create a convenient lambda which returns a health check response that also has a name. Lambda-based responses appear, with details enabled in the output, like this:
```
{
  "status": "UP",
  "checks": [
    {
      "name": "Main$$Lambda/0x00000001320ac800",
      "status": "UP",
      "data": {
        "exampleHealthCheck": "looking-good"
      }
    },
...
}
```

Developers can, of course, write a class which implements `HealthCheck` and apply the name to the check (not the response) but that approach does not allow the convenience of using in-line lambdas.

## Steps to reproduce
Add a health check response in-line something like this:
```
ObserveFeature observe = ObserveFeature.builder()
    .addObserver(HealthObserver.builder()
    .addCheck(() -> HealthCheckResponse.builder()
             .status(readyTime.get() != 0)
             .detail("time", readyTime.get())
             .build(), HealthCheckType.READINESS)
```
You get a name like the example above.

tjquinno:
### Description
Resolves #7827

As Tomas suggested in the issue comments, this PR adds a variant to the `HealthObserver` builder that accepts a `Supplier<HealthCheckResult>`, a health check type, and (this is the new part) a name for the added health check.

(The original code provided no way to assign a name to a health check that was added by using a lambda or method reference for the health check response.)

### Documentation

Doc update included in the PR.

I revised the doc slightly so the example code uses a variable of type `Supplier<HealthCheckResponse>` instead of `HealthCheck` so the doc could show the use of the new method. IMO this is a better pattern to illustrate than creating a check _without_ an explicit name which yields the output I showed in the issue in which the generated name is shown as something like `"name": "Main$$Lambda/0x00000001320ac800"`.

The revised doc shows the code explicitly setting the name when invoking `addCheck` and the corresponding clearer output.

### DIFF
--- a/docs/se/health.adoc
+++ b/docs/se/health.adoc
@@ -93,9 +93,9 @@ A health check is a Java functional interface that returns a
 [source,java]
 .Health check with a lambda expression:
 ----
-HealthCheck hc = () -> HealthCheckResponse
+Supplier<HealthCheckResponse> hcr = () -> HealthCheckResponse
                 .builder()
-                .detail("exampleCheck", "looking-good")
+                .detail("exampleValue", "looking-good")
                 .status(HealthCheckResponse.Status.UP) // always "up"
                 .build();
 ----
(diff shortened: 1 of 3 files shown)


ljamen (approved):
LGTM

PR #7995, opened 2023-11-14
Fix repeated pseudo-header field `:status`

### LINKED ISSUE #7991: 4.x: gRPC handler send `:status` headers twice

## Environment Details
* Helidon Version:
  * Helidon SE
* JDK version: 21
* OS: Mac

----------

## Problem Description
Calls to a simple echo gRPC server from Postman or other gRPC clients result in an internal error. The error message is like `Received RST_STREAM with code 1`.

image

## Steps to reproduce

This issue can be reproduced every time a gRPC method is called from a strict gRPC client like postman. I've capture the data packets, and found the root cause of it: 2 headers frames both have set  the`:status` header.

Below are quoted from the [http/2 standard](https://datatracker.ietf.org/doc/html/rfc9113#PseudoHeaderFields)

>Pseudo-header fields MUST NOT appear in a trailer section.

>The same pseudo-header field name MUST NOT appear more than once in a field block. A field block for an HTTP request or response that contains a repeated pseudo-header field name MUST be treated as [malformed](https://datatracker.ietf.org/doc/html/rfc9113#malformed) ([Section 8.1.1](https://datatracker.ietf.org/doc/html/rfc9113#malformed)).

Apparently the current gRPC implementation does not conform to that.

### First headers frame
image

### Second headers frame
image

lilac:
### Description

Fix #7991

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -201,15 +201,13 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
                 // todo ignoring trailers
                 WritableHeaders<?> writable = WritableHeaders.create();
 
-                // write the expected gRPC headers for content type and status
-                writable.set(GRPC_CONTENT_TYPE);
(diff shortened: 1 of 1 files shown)


lilac:
@m0mus The issue had been moved to high priority backlog. Can someone be assigned to review it?

tomas-langer (approved):

PR #8010, opened 2023-11-15
4.x Fix nested config prefix for observer settings

### LINKED ISSUE #8009: 4.x MP metrics and health do not honor `server.features.observe.observers` config, but instead `observe.providers.*`
## Environment Details
* Helidon Version: 4.0.0
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
In MP apps, metrics and health follow config at the top level (`metrics` and `health`) but not config inside `server.features.observe.observers.metrics` or `...health` as SE apps do.

But they do respond to config at `observe.providers.xxx`.

This seems to be because the constructors for `MetricsCdiExtension` and `HealthCdiExtension` incorrectly pass hardcoded strings `"observe.providers.metrics"` and `"observe.providers.health"` to the superclass constructor as one of the config paths to consider.

I assume the intended behavior is for the CDI extensions to use either the top-level config for `metrics` and `health` or the config nested under `server...` but the paths do not have the correct prefix and use `providers` instead of `observers`.

## Steps to reproduce
1. Build and run the MP QuickStart app _without change_, then run this command: `curl -v http://localhost:8080/health`. You get the expected output with status 200
   ```json
   {"status":"UP","checks":[]}
   ```
   because the MP health CDI extension sets the health setting `details` to true programmatically.
2. Stop the server, then add the following config:
   ```properties
   server.features.observe.observers.health.details = false
   ```
3. Rebuild and rerun the server. The output is the same because the health CDI observer ignores the config under `server.features...`.

tjquinno:
### Description
Resolves #8009 

The metrics and health CDI extension constructors passed the incorrect config key for nested (non-top-level) config settings.

The original code passed `observe.providers.metrics` and `observe.providers.health`. I suspect this was correct before the restructuring of features and observers, and their config now appears under `server.features.observe.observers.metrics` and `...health` instead and use `observers` instead of `providers` in the prefix. I further suspect these strings just fell through the cracks during the refactoring.

This PR adds a superclass method which computes the correct prefix (so the hard-coded part of the prefix is in _one_ place) and adjusts the constructors to invoke it.

Note that the superconstructor accepts a list of config prefixes, and the leaf constructors pass both the above correct prefixes and the old (3.x) top-level prefixes (`metrics` and `health`).

### Documentation
Bug fix - no doc impact.

### DIFF
--- a/microprofile/service-common/src/main/java/io/helidon/microprofile/servicecommon/HelidonRestCdiExtension.java
+++ b/microprofile/service-common/src/main/java/io/helidon/microprofile/servicecommon/HelidonRestCdiExtension.java
@@ -255,6 +255,16 @@ public abstract class HelidonRestCdiExtension implements Extension {
     }
 
     /**
+     * Returns the config key for settings for the specified suffix nested within the server config tree.
+     *
+     * @param suffix the config key suffix (typically the name of the component: e.g., health)
+     * @return full nested config key for the specified suffix
+     */
+    protected static String nestedConfigKey(String suffix) {
+        return "server.features.observe.observers." + suffix;
+    }
+
+    /**
      * Configure with runtime config.
      *
      * @param config config to use
--- a/microprofile/health/src/main/java/io/helidon/microprofile/health/HealthCdiExtension.java
+++ b/microprofile/health/src/main/java/io/helidon/microprofile/health/HealthCdiExtension.java
@@ -58,7 +58,7 @@ public class HealthCdiExtension extends HelidonRestCdiExtension {
      * Creates a new instance of the health CDI extension.
      */
     public HealthCdiExtension() {
-        super(LOGGER, "observe.providers.health", "health");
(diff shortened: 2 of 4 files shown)


microprofile/service-common/src/main/java/io/helidon/microprofile/servicecommon/HelidonRestCdiExtension.java
```diff
@@ -254,6 +254,16 @@ protected HttpRouting.Builder routingBuilder(ServerCdiExtension server) {
                 : server.serverNamedRoutingBuilder(routingName);
     }
 
+    /**
+     * Returns the config key for settings for the specified suffix nested within the server config tree.
```
ljnelson, line 258: Minor comment: might be nice to append something like: "which is `server.features.observe.observers.`."

ljnelson (approved):

Issue #7986, opened 2023-11-10
websocket server send text error 1002


baichangda:
## Environment Details
 implementation group: 'io.helidon.webserver', name: 'helidon-webserver', version: '4.0.0'
 implementation group: 'io.helidon.webserver', name: 'helidon-webserver-static-content', version: '4.0.0'
 implementation group: 'io.helidon.webserver', name: 'helidon-webserver-websocket', version: '4.0.0'

----------

## Problem Description
helidon se websocket server send text to client in onOpen function produce client error 1002

## Steps to reproduce
websocket server code like this 
![image](https://github.com/helidon-io/helidon/assets/24662536/fab359b8-29d8-4aa9-b783-852eaf270e6f)
webcoekt client code like this
![image](https://github.com/helidon-io/helidon/assets/24662536/b555d1a0-8294-41ea-a6e8-cf69d34721b3)

1、js websocket connect to helidon websocket server succeed
2、in onOpen(...) callback send msg to client、send function is 'WsSession send(String text, boolean last)'
3、then js websocket client error
![image](https://github.com/helidon-io/helidon/assets/24662536/1da2b648-a085-48ec-a5a2-f3dc8b8cb176)

spericas:
@baichangda Seems like you're not showing the call to `send(String, boolean)` in your `onOpen`, right? Do you have sample code that matches your problem description? Is that in your `WsSession` class?

baichangda:
@spericas there is call stack image
![image](https://github.com/helidon-io/helidon/assets/24662536/fb1457ae-5365-4259-b2fc-ffd56f097345)
![image](https://github.com/helidon-io/helidon/assets/24662536/9c48e512-ec59-4393-8c52-4d1f1be4aacd)
![image](https://github.com/helidon-io/helidon/assets/24662536/013b8160-5a59-4419-9f82-52da9b19f73a)
![image](https://github.com/helidon-io/helidon/assets/24662536/dc827160-f160-4080-9654-403b99bb792c)

spericas:
@baichangda I'm not able to reproduce your problem using Helidon 4.0.0 and Chrome 119.0.6045.123. Please find a simple WS app attached that you can build and run as a server. For the client part, I'm just pasting some JavaScript code into the browser's console:

```
ws = new WebSocket("ws://localhost:8080/ws")
ws.onopen = function (evt) {
    console.log("onopen " + evt.type);
}
ws.onmessage = function (evt) {
[...]

PR #8025, opened 2023-11-17
Support for longer WebSocket frames

spericas:
### Description

1. Support for writing WS frames of more than 125 bytes, i.e. with lengths that fit in two bytes and eight bytes.
2. Configuration of max-frame-length to limit the size of a frame read by the server. The JDK WS client splits a long payload into 16K frames.
3. Some new tests for (1) and (2).

Related to issue #7986.

### Documentation

JavaDoc has been updated.

### DIFF
--- a/webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
+++ b/webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
@@ -45,11 +45,14 @@ import io.helidon.websocket.WsSession;
 public class WsConnection implements ServerConnection, WsSession {
     private static final System.Logger LOGGER = System.getLogger(WsConnection.class.getName());
 
+    static final String MAX_FRAME_LENGTH = "1048576";
+
     private final ConnectionContext ctx;
     private final HttpPrologue prologue;
     private final Headers upgradeHeaders;
     private final String wsKey;
     private final WsListener listener;
+    private final WsConfig wsConfig;
 
     private final BufferData sendBuffer = BufferData.growing(1024);
     private final DataReader dataReader;
@@ -75,6 +78,13 @@ public class WsConnection implements ServerConnection, WsSession {
         this.listener = wsRoute.listener();
         this.dataReader = ctx.dataReader();
         this.lastRequestTimestamp = DateTime.timestamp();
+        this.wsConfig = (WsConfig) ctx.listenerContext()
+                                      .config()
+                                      .protocols()
+                                      .stream()
+                                      .filter(p -> p instanceof WsConfig)
+                                      .findFirst()
+                                      .orElse(null);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
```diff
@@ -243,8 +248,8 @@ private boolean processFrame(ClientWsFrame frame) {
 
     private ClientWsFrame readFrame() {
         try {
-            // TODO check may payload size, danger of oom
-            return ClientWsFrame.read(ctx, dataReader, Integer.MAX_VALUE);
+            int maxFrameLength = wsConfig != null ? wsConfig.maxFrameLength() : Integer.MAX_VALUE;
```
danielkec, line 251: Default is 1Mb and default default 2Gb?

danielkec: Can the setting change between the frame reads? Could we sort it out once in the constructor?

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
```diff
@@ -85,14 +88,16 @@ private WsConnection(ConnectionContext ctx,
...
      */
     public static WsConnection create(ConnectionContext ctx,
                                       HttpPrologue prologue,
                                       Headers upgradeHeaders,
                                       String wsKey,
-                                      WsRoute wsRoute) {
-        return new WsConnection(ctx, prologue, upgradeHeaders, wsKey, wsRoute);
+                                      WsRoute wsRoute,
+                                      WsConfig wsConfig) {
+        return new WsConnection(ctx, prologue, upgradeHeaders, wsKey, wsRoute, wsConfig);
```
danielkec, line 100: Public API change in dot release, can we keep the old version and deprecate it?

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
spericas: No, it is just that `DirectWsConnection` is not connected to `WsConfig` at this time, so it may be null. I suppose I can make it 1 MB in both cases.

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
spericas: Good catch, let me see how we can do this differently.

danielkec (dismissed):

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
```diff
@@ -75,6 +78,13 @@ private WsConnection(ConnectionContext ctx,
         this.listener = wsRoute.listener();
         this.dataReader = ctx.dataReader();
         this.lastRequestTimestamp = DateTime.timestamp();
+        this.wsConfig = (WsConfig) ctx.listenerContext()
+                                      .config()
+                                      .protocols()
+                                      .stream()
+                                      .filter(p -> p instanceof WsConfig)
```
tomas-langer, line 85: You could work around the ugly cast by doing (after filter)
`.map(WsConfig.class::cast)`
which will change the stream to correctly typed one.

PR #8032, opened 2023-11-17
3.x: 8029 Bad media type logging level

### LINKED ISSUE #8029: [3.x] SEVERE log statement for a handled problem
Backport of #7110 for Helidon 3.x

danielkec:
Fixes #8029


tomas-langer (changes requested):
I do not like this fix. 
This is a pick-up point that covers errors that were unexpected.

A problem with parsing media type is an expected issue, and should be handled correctly (and return bad media type). Such a problem should never reach this place.

Also there should be a test verifying that a specific issue does not trigger this anymore.

tomas-langer (dismissed):
Looks good, but copyright

romain-grecourt (approved):

Issue #6545, opened 2023-04-03
Nima - HTTP/2 Client - graceful connection closing


danielkec:
- [x] Figure how gracefully and properly close connection pool(autocloseable client? sending goaway?) #7398

- [x] Figure how to detect stale connections in the pool(using PING frame?) #7398

Tomas:
- [x] we could send pings from our idle connection handler (that times out idle connections) - shorter period, either send ping or timeout, maybe with some logic to check that we got ACK for the ping before giving the connection to anybody #7398
- [ ] add shutdown hook(?) to handle static connection caches to gracefully shut them down

PR #8051, opened 2023-11-21
4.x: Graceful client connection close

danielkec:
Fixes part of #6545

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/spi/ClientConnectionCache.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/spi/ClientConnectionCache.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2023 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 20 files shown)


webclient/api/src/main/java/io/helidon/webclient/spi/ClientConnectionCache.java
```diff
@@ -0,0 +1,28 @@
...
+
+package io.helidon.webclient.spi;
+
+import io.helidon.webclient.api.ReleasableResource;
+
+/**
+ * Client connection cache with release shutdown hook to provide graceful shutdown.
+ */
+public abstract class ClientConnectionCache implements ReleasableResource {
+    protected ClientConnectionCache() {
```
tomas-langer, line 25: This will be created also for per-instance caches, so we may leak a lot of shutdown hooks. Please add a hook only for static JVM-wide caches (should be just two - one for HTTP/1 and one for HTTP/2), and do something different for the per-client caches.

tomas-langer (changes requested):

webclient/api/src/main/java/io/helidon/webclient/spi/ClientConnectionCache.java
danielkec: fixed

tomas-langer (approved):

Issue #7957, opened 2023-11-06
4.x: Helidon Inject - only API dependency to compile


tomas-langer:
The current implementation requires runtime to be present at compilation time (through dependencies of the generated code).

This should be changed, so only API is required.

tomas-langer:
An attempt will be done to have this backward compatible (by keeping the `bind` methods available and deprecated).

The new approach is being designes as follows:

1. `ServiceDescriptor` will be code generated - this type will depend only on API, and will use class model to be generated
2. The runtime will pick up the generated service descriptors using a module (similar to what is done right now) by binding the descriptors to the binder
3. Then an appropriate service provider will be discovered (through service loader, using an ID that is code generated together with the service descriptor (default is a `default` method on the interface)
4. These implementation take care of all the lifecycle tasks (activation, deactivation etc.)

Most of the code is either re-used or copied from existing implementation, with focus on:

- `null` is never used in APIs
- using `final` fields as much as possible (with the exception of state related fields, such as `status`, `serviceInstance` etc.)
- keep service sources as they are right now (e.g. user's code is not modified when creating a service, the generated code and injection runtime code is going to change)
- the concepts remain - source code generated, analysis of the world to make sure all injection points are satisfied, the capability to generate an application class that explictly binds injection points to service providers to prevent runtime discovery

tomas-langer:
Some conceptual changes are required to align with Helidon style and coding guidelines.
The PR will contain appropriate readme, high level info:

- there will be a new root module `codegen` to contain all tools related to code processing and generation, with implementations for annotation processing (APT) and classpath scanning (scan). Scan is expected to be used from maven plugin(s), and possibly from command line tools. We could also have a reflection based implementation of these if ever needed
- this results in using `codegen` instead of `processor` in most cases (change in modules)
- class model is moved to `codegen`
- Helidon copyright module is moved to `codegen`
- `tools` module is removed

trentjeff:
You need to add a `dependencies-test` making use of the enforcer plugin to ensure that there is no inject runtime in the cp. This is to avoid regressions in this area.

PR #8099, opened 2023-12-04
4.x: Requried changes for code generation

tomas-langer:
Note for release:
this should not be release as part of a dot release of Helidon. There may be backward incompatible changes required as I introduce Helidon Inject codegen for new approach. We should have the option to finalize the new modules before releasing them

### Description
Related to #7957 

This PR introduces `codegen` root module, and refactoring of Builder codegeneration to this new module set (to prove the usability). This required me to fix all `Blueprint` instances to only use builder API (and not config metadata).
New source that requires switch(object) for instanceof checks was not supported by checkstyle - upgraded, needed some additional fixes in our sources.
The module is a prerequisite to refactoring of Helidon Inject (so we can have dependency only on API and not on runtime in modules that provide services).

Everything should be backward compatible (introduced new modules, deprecated old ones).

I have deprecated `common/processor`, as that was too narrow focused (on annotation processing), we need to have wider support, and it did not feel right. The root `codegen` module is a better place for these classes.

Modules introduced (README in code is available):
- apt - Annotation processing support - main Helidon annotation processor that uses codegen
- class-model - Copy of the class model (deprecated the old one, introduced changes needed)
- codegen - utilities and main implementation of the analysis of extensions and their invocation (PR includes builder codegen extension), SPI to extend code generation and processing (some are copies of existing code)
- compiler - support for invoking compiler from java (used further down the road by the Maven plugin), partial copy of existing code
- helidon-copyright - copy of existing copyright
- scan - support for classpath scanning (partial copy, will be used from Maven plugin)

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -501,6 +501,36 @@
                 <version>${helidon.version}</version>
             </dependency>
             <dependency>
+                <groupId>io.helidon.codegen</groupId>
+                <artifactId>helidon-codegen</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.codegen</groupId>
+                <artifactId>helidon-codegen-apt</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.codegen</groupId>
+                <artifactId>helidon-codegen-scan</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.codegen</groupId>
+                <artifactId>helidon-codegen-compiler</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.codegen</groupId>
+                <artifactId>helidon-codegen-class-model</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.helidon.codegen</groupId>
+                <artifactId>helidon-codegen-helidon-copyright</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>io.helidon.common.processor</groupId>
                 <artifactId>helidon-common-processor</artifactId>
                 <version>${helidon.version}</version>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1325,6 +1325,11 @@
                 <artifactId>helidon-builder-processor</artifactId>
                 <version>${helidon.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.helidon.builder</groupId>
+                <artifactId>helidon-builder-codegen</artifactId>
+                <version>${helidon.version}</version>
+            </dependency>
 
             <!-- Injection Core -->
             <dependency>
--- a/codegen/README.md
+++ b/codegen/README.md
@@ -1,0 +1,57 @@
+Codegen
+----
+
+Code generation and code processing tools of Helidon.
+
+We see the following three environments that are used for code processing:
+
+1. Annotation processors
+2. Classpath scanning
+3. Reflection based in a running VM
+
+In Helidon, we do as much as possible using annotation processing and source code generation. 
+For the cases where we need to analyze external libraries, or arbitrary code, we use classpath scanning (such as from a Maven plugin or command line tool).
+Reflection can only be used in Helidon Microprofile, or in selected modules that are intentionally reflection based (this should be currently limited to Config object mapping module).
+
+
+# Modules
+
+This top level module contains the following modules:
+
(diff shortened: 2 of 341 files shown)


common/tls/src/main/java/io/helidon/common/tls/TlsConfigBlueprint.java
```diff
@@ -250,31 +254,33 @@ static List<X509Certificate> createTrust(Keys config) {
...
 
     /**
      * SSL session timeout.
      *
      * @return session timeout, defaults to 30 minutes
      */
-    @ConfiguredOption("PT30M")
+    @Option.Configured
+    @Option.Default("PT30M")
     Duration sessionTimeout();
```
danielkec, line 268: Did we change the default value from `0` between Helidon 3 and 4?

codegen/README.md
```diff
@@ -0,0 +1,57 @@
+Codegen
+----
+
+Code generation and code processing tools of Helidon.
+
+We see the following three environments that are used for code processing:
+
+1. Annotation processors
+2. Classpath scanning
```
trentjeff, line 9: I would argue we have these three:

1. Annotation processing.
2. Classpath scanning + Reflection based processing.
3. Build (e.g., maven-plugin) processing.

codegen/class-model/src/main/java/io/helidon/codegen/classmodel/Type.java
```diff
@@ -0,0 +1,111 @@
...
+            int i = 0;
+            for (TypeName param : instance.typeArguments()) {
+                if (i > 0) {
+                    nameBuilder.append(", ");
+                }
+                nameBuilder.append(param.resolvedName());
+                i++;
+            }
+            nameBuilder.append(">");
+        }
```
danielkec, line 75: Could be a little nicer maybe

```suggestion
        if (!instance.typeArguments().isEmpty()) {
            nameBuilder.append("<");
            nameBuilder.append(instance.typeArguments().stream()
                    .map(typeName -> typeName.resolvedName())
                    .collect(Collectors.joining(", ")));
            nameBuilder.append(">");
        }
```

codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeInfoFactory.java
```diff
@@ -0,0 +1,572 @@
...
+                        if (isCheckedException(ctx, it)) {
+                            TypeName typeName = AptTypeFactory.createTypeName(it).orElse(null);
+                            if (typeName == null) {
+                                return Stream.of();
+                            }
+                            return Stream.of(typeName);
+                        }
+                        return Stream.of();
+                    })
+                    .collect(Collectors.toSet());
```
danielkec, line 207: Could be like
```suggestion
            thrownChecked = ee.getThrownTypes()
                    .stream()
                    .filter(it -> isCheckedException(ctx, it))
                    .flatMap(it -> AptTypeFactory.createTypeName(it).stream())
                    .collect(Collectors.toSet());
```

danielkec (approved):
LGTM just small suggestions

codegen/README.md
tomas-langer: Maven plugin is expected to use classpath scanning, as it cannot use annotation processing. Also we currently do not use reflection based processing AFAIK.

tomas-langer:
> LGTM just small suggestions

I will apply them in next PR.

bom/pom.xml
```diff
@@ -500,6 +500,36 @@
                 <artifactId>helidon-common-tls</artifactId>
                 <version>${helidon.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.helidon.codegen</groupId>
```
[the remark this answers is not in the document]
tomas-langer: I do prefer the co-location by feature here (e.g. inject is a feature, and it has its own codegen). 
e.g. `helidon-inject-codegen` rather than `helidon-codegen-inject` (as I probably use the `inject` feature, and this is my `codegen` for that feature.
Regardless of the choice here, I expect us to have some kind of "bundle" of Helidon codegens that could be used by customers for their project, without the need to understand details (e.g. `helidon-bundle-apt`  or something like that). 

I do not think there is a single "right" answer for this choice.

common/tls/src/main/java/io/helidon/common/tls/TlsConfigBlueprint.java
tomas-langer: Seems like it. The default in Java is 24 hours

tomas-langer: Created #8127

tomas-langer:
I have pushed the changes (squashed) from this PR to branch [inject-update](https://github.com/helidon-io/helidon/tree/inject-updates). I will be creating all follow up PRs against that branch.

PR #8101, opened 2023-12-04
#8077 Client keep alive fix

### LINKED ISSUE #8077: 4.0.1: WebClient HTTP keep-alive support appears to be broken
## Environment Details
* Helidon Version: 4.0.0 and 4.0.1
* Helidon SE 
* JDK version: java 21.0.1 2023-10-17 LTS
* OS: MacOS

----------

## Problem Description

The Helidon 4 WebClient HTTP keep-alive support appears to be broken. Even with `keepAlive(true)` the client does not appear to reuse connections.

## Steps to reproduce

See the attached test case: [client.zip](https://github.com/helidon-io/helidon/files/13480592/client.zip)

This is based on the stand-alone webclient example that has been modified to set `keepAlive(true)` and to perform 10 GET requests sequentially.  After the test is run (see README) netstat shows 10 connections in TIME_WAIT instead of one:

```
tcp4       0      0  127.0.0.1.56044        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56045        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56046        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56047        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56048        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56049        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56050        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56051        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56052        127.0.0.1.8080         TIME_WAIT
tcp4       0      0  127.0.0.1.56053        127.0.0.1.8080         TIME_WAIT
```

danielkec:
Fixes #8077 and #8110

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
@@ -17,17 +17,15 @@
 package io.helidon.webclient.http1;
 
 import java.io.InputStream;
-import java.time.Duration;
 import java.util.List;
+import java.util.OptionalLong;
 import java.util.ServiceLoader;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.helidon.common.GenericType;
 import io.helidon.common.HelidonServiceLoader;
+import io.helidon.common.LazyValue;
 import io.helidon.common.buffers.BufferData;
 import io.helidon.common.media.type.ParserMode;
 import io.helidon.http.ClientRequestHeaders;
@@ -35,6 +33,7 @@ import io.helidon.http.ClientResponseHeaders;
 import io.helidon.http.ClientResponseTrailers;
 import io.helidon.http.HeaderNames;
 import io.helidon.http.HeaderValues;
+import io.helidon.http.Headers;
 import io.helidon.http.Http1HeadersParser;
 import io.helidon.http.Status;
 import io.helidon.http.media.MediaContext;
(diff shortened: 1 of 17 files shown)


webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -96,7 +97,7 @@ class Http1ClientResponseImpl implements Http1ClientResponse {
         this.whenComplete = whenComplete;
 
         if (responseHeaders.contains(HeaderNames.CONTENT_LENGTH)) {
```
tomas-langer, line 99: there is a method `contentLength()` on server response headers that can be used

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -166,10 +167,10 @@ public void close() {
                     } else if (entityLength == ENTITY_LENGTH_CHUNKED) {
                         if (hasTrailers) {
                             readTrailers();
```
tomas-langer, line 169: As far as I understand this, we should always check for `entityFullyRead` - and if not read, we should either consume the entity (e.g. throw it away), or close the connection (as we did originally).
Only when fully read should we release the connection (regardless of chunked encoding or content lenght). The only difference is that with chunked we may also need to read trailers...

tomas-langer (changes requested):

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
danielkec: thx!

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
danielkec: Fixed and trailered

tomas-langer (approved):

PR #8111, opened 2023-12-07
Update server's internal state if a listener fails to start

### LINKED ISSUE #8109: LoomServer does not update internal state when listener fails to start
If a listener fails to start (for example because of an unsupported TLS setting), the error is logged but the server's `isRunning` method still returns true after shutting down all other listeners. Internal server state needs to reflect failure to start. See `LoomServer::startIt`.

spericas:
### Description
Update server's internal state if a listener fails to start for some reason. Calls to `isRunning` must return `false` if listeners failed to start and the server isn't listening for connections.

Fixes #8109

### Documentation

None

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
@@ -210,6 +210,7 @@ class LoomServer implements WebServer {
             if (startFutures != null) {
                 startFutures.forEach(future -> future.future().cancel(true));
             }
+            running.set(false);
             return;
         }
         if (registerShutdownHook) {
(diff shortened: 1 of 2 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
```diff
@@ -210,6 +210,7 @@ private void startIt() {
             if (startFutures != null) {
                 startFutures.forEach(future -> future.future().cancel(true));
             }
+            running.set(false);
```
barchetta, line 213: Maybe it doesn't matter, but should this be moved up to line 208?

barchetta (approved):

Issue #8106, opened 2023-12-06
Upgrade grpc-java


barchetta:
## Problem Description

Upgrade grpc-java to 1.60.0 or newer: https://github.com/grpc/grpc-java/releases

Note that module names have bee added to most of the artifacts:
```
grpc-api
Automatic-Module-Name: io.grpc

grpc-core
Automatic-Module-Name: io.grpc.internal

grpc-inprocess
Automatic-Module-Name: io.grpc.inprocess

grpc-netty
Automatic-Module-Name: io.grpc.netty

grpc-protobuf
Automatic-Module-Name: io.grpc.protobuf

grpc-protobuf-lite
Automatic-Module-Name: io.grpc.protobuf.lite

grpc-services
Automatic-Module-Name: io.grpc.services

grpc-stub
Automatic-Module-Name: io.grpc.stub

grpc-util
Automatic-Module-Name: io.grpc.util
```

PR #8113, opened 2023-12-08
Update gRPC version to 1.60.0

thegridman:
### Documentation

This PR is to fix issue #8106

The gRPC Java libraries now have automatic module names in the jar files. These names differ slightly from those previously used by Helidon as they are all prefixed with `io.grpc` instead of just `grpc`. An issue resulting from this is that the gRPC API module has a module name of `io.grpc` which is the same as the repackaged gRPC module in Helidon. The Helidon repackaged io.grpc module no longer has any code, it is just a pom file with some dependencies. Its `io.grpc` module name is generated from the artifactId. We cannot change the artifactId but we can add a module-info to the jar to set its module name to be something other than `io.grpc`.

Usages of the repackaged `io.grpc` module have been removed from the rest of Helidon, other modules in Helidon now depend directly on whatever gRPC libraries they require. The  repackaged `io.grpc` module has been left in place so as not to break any customer's that depend on it. Having said that, customer code that uses gRPC libraries is almost certainly going to break anyway due to the change of module names, which is beyond our control.

### DIFF
--- a/grpc/io.grpc/pom.xml
+++ b/grpc/io.grpc/pom.xml
@@ -33,6 +33,7 @@
     <description>Bundle of grpc dependencies provided for backwards compatibility</description>
 
     <properties>
+        <maven.javadoc.skip>true</maven.javadoc.skip>
         <spotbugs.skip>true</spotbugs.skip>
     </properties>
 
--- a/grpc/io.grpc/pom.xml
+++ b/grpc/io.grpc/pom.xml
@@ -32,6 +32,10 @@
     <name>Helidon grpc-java Bundle</name>
     <description>Bundle of grpc dependencies provided for backwards compatibility</description>
 
+    <properties>
+        <spotbugs.skip>true</spotbugs.skip>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>io.grpc</groupId>
--- a/grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerConfiguration.java
+++ b/grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerConfiguration.java
@@ -108,6 +108,18 @@ public interface GrpcServerConfiguration {
     GrpcTlsDescriptor tlsConfig();
 
     /**
+     * Returns the limit for the rate of incoming RST_STREAM frames per connection to maxRstStream per second.
+     * When exceeded on a connection, the connection is closed. This can reduce the impact of
+     * an attacker continually resetting RPCs before they complete.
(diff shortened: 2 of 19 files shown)


grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerConfiguration.java
```diff
@@ -188,6 +202,7 @@ public Builder config(Config config) {
 
             name = config.get("name").asString().orElse(DEFAULT_NAME);
             port = config.get("port").asInt().orElse(DEFAULT_PORT);
+            maxRstFramesPerSecond = config.get("maxRstFramesPerSecond").asInt().orElse(Integer.MAX_VALUE);
```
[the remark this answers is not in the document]
danielkec: Would it be possible to stick to 200 resets per 30-second window as default? Also we already have `rapid-reset-check-period` and `max-rapid-resets` config keys for server in Helidon 4 would it be possible to use same config keys?

grpc/server/src/main/java/io/helidon/grpc/server/GrpcServerImpl.java
```diff
@@ -178,11 +178,17 @@ public CompletionStage<GrpcServer> start() {
...
+            if (maxRstFramesPerSecond <= 0) {
+                maxRstFramesPerSecond = Integer.MAX_VALUE;
+            }
+
             server = configureNetty(builder)
                     .directExecutor()
                     .addService(healthService)
                     .addService(ProtoReflectionService.newInstance())
                     .fallbackHandlerRegistry(handlerRegistry)
+                    .maxRstFramesPerWindow(maxRstFramesPerSecond, 1)
```
danielkec, line 191:
```suggestion
                    .maxRstFramesPerWindow(config.maxRapidResets(), rapidResetCheckPeriod().toSeconds())
```

danielkec (changes requested):

grpc/io.grpc/pom.xml
```diff
@@ -32,6 +32,10 @@
     <name>Helidon grpc-java Bundle</name>
     <description>Bundle of grpc dependencies provided for backwards compatibility</description>
 
+    <properties>
+        <spotbugs.skip>true</spotbugs.skip>
+    </properties>
```
barchetta, line 37: Looks like you also need to add `<maven.javadoc.skip>true</maven.javadoc.skip>`

danielkec (approved):

PR #8118, opened 2023-12-08
Fix premature access to RegistryFactory

### LINKED ISSUE #8117: 4.x: Pokemons example start-up logs that `RegistryFactory` is retrieved before being initialized
## Environment Details
* Helidon Version: 4.x
* Helidon MP
* JDK version: 21
* OS:

----------

## Problem Description

Pokemon example starts with following log:
```bash
Dec 08, 2023 10:05:55 AM io.helidon.microprofile.metrics.RegistryFactory getInstance
WARNING: Attempt to retrieve current io.helidon.microprofile.metrics.RegistryFactory before it has been initialized; 
using default Helidon meter registry and continuing
```

## Steps to reproduce

```bash 
cd examples/integrations/cdi/pokemons
mvn package
java -jar target/helidon-integrations-examples-pokemons.jar
```

@tjquinno made a first diagnosis that the `JpaExtension` has an earlier priority than `MetricsCdiExtension` and seek metrics before MP metrics initialize it.
Note that increasing the `JpaExtension` priority solves this issue for this specific use case.

tjquinno:
### Description
Resolves #8117 

The Helidon MP metrics `RegistryFactory` logged the warning because the JPA CDI extension was running some start-up complete code before the metrics CDI extension had a chance to prepare Helidon MP metrics.

The fix sets the priority on the JPA extension observer to be later than the corresponding priority on the metrics extension observer.

Not as part of the PR I temporarily changed the `RegistryFactory` code to thrown an exception instead of logging a warning, and the pipelines passed so it seems that no other CDI extensions need a similar adjustment.

### DIFF
--- a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
+++ b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
@@ -804,7 +804,7 @@ public class JpaExtension implements Extension {
      * @exception IOException if an input or output error occurs, typically because a {@code META-INF/persistence.xml}
      * resource was found but could not be loaded for some reason
      *
-     * @exception JAXBException if there was a problem {@linkplain Unmarshaller#unmarshal(Reader) unmarshalling} a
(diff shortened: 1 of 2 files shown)


ljnelson (approved):
Even in my current personal situation I can make sense of _this_ PR. Approved! Look! I Did Work™!

PR #8133, opened 2023-12-11
Inject updates 1

tomas-langer:
### Description
Review fixes from previous PR
Support for custom constants in builder
Support for List and Set codegen options
Type info factory fixes for both APT and classpath scanning (was using wrong type names, one requires $ for nested, one .)
Fixed `toConstant` utility method

Fixes in class model:

- javadoc is now correctly cleared in common component (aligned with javadoc)
- Type test
- Field can now accept Javadoc instance

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
@@ -27,23 +27,28 @@ import io.helidon.codegen.CodegenContext;
 import io.helidon.codegen.CodegenException;
 import io.helidon.codegen.ElementInfoPredicates;
 import io.helidon.codegen.classmodel.ContentBuilder;
+import io.helidon.codegen.classmodel.Javadoc;
 import io.helidon.common.Errors;
+import io.helidon.common.types.AccessModifier;
 import io.helidon.common.types.Annotation;
+import io.helidon.common.types.Modifier;
 import io.helidon.common.types.TypeInfo;
 import io.helidon.common.types.TypeName;
 import io.helidon.common.types.TypeNames;
 
 import static io.helidon.builder.codegen.Types.PROTOTYPE_BUILDER_METHOD;
+import static io.helidon.builder.codegen.Types.PROTOTYPE_CONSTANT;
 import static io.helidon.builder.codegen.Types.PROTOTYPE_CUSTOM_METHODS;
 import static io.helidon.builder.codegen.Types.PROTOTYPE_FACTORY_METHOD;
 import static io.helidon.builder.codegen.Types.PROTOTYPE_PROTOTYPE_METHOD;
 
 record CustomMethods(List<CustomMethod> factoryMethods,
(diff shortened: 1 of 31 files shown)


builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
```diff
@@ -225,6 +232,46 @@ private static GeneratedMethod factoryMethod(Errors.Collector errors,
...
+    private static List<CustomConstant> findConstants(TypeInfo customMethodsType,
+                                                      Errors.Collector errors) {
+        return customMethodsType.elementInfo()
+                .stream()
+                .filter(ElementInfoPredicates::isField)
+                .filter(ElementInfoPredicates.hasAnnotation(PROTOTYPE_CONSTANT))
+                .map(it -> {
+                    if (!it.elementModifiers().contains(Modifier.STATIC)) {
+                        errors.fatal(it,
+                                     "A field annotated with @Prototype.Constant must be static, final, "
```
trentjeff, line 244: Your message seems contradictory - it needs to be static ... "Not static." -> should be removed.

builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
```diff
@@ -225,6 +232,46 @@ private static GeneratedMethod factoryMethod(Errors.Collector errors,
...
+                                             + "and at least package local. Not static.");
+                    }
+                    if (!it.elementModifiers().contains(Modifier.FINAL)) {
+                        errors.fatal(it,
+                                     "A field annotated with @Prototype.Constant must be static, final, "
+                                             + "and at least package local. Not final.");
```
trentjeff, line 250: "Not final." should be removed as it is confusing.

builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
```diff
@@ -225,6 +232,46 @@ private static GeneratedMethod factoryMethod(Errors.Collector errors,
...
+                    }
+                    if (it.accessModifier() == AccessModifier.PRIVATE) {
+                        errors.fatal(it,
+                                     "A field annotated with @Prototype.Constant must be static, final, "
```
trentjeff, line 254: remove "Private."

builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
tomas-langer: It actually is correct - the field needs to be static. The "Not static." is the error message - your field is not static.

builder/codegen/src/main/java/io/helidon/builder/codegen/CustomMethods.java
tomas-langer: I will try to reword.

trentjeff (approved):

Issue #8102, opened 2023-12-05
4.x: Update performance tuning doc still mentioning Netty


danielkec:
https://helidon.io/docs/v4/#/mp/guides/performance-tuning

- [ ] Remove existing document
- [ ] Create a new performance tuning guide

barchetta:
Most of the current document is now longer applicable -- because we no longer use Netty and because we no longer need to manage thread pool sizes. Yay!

Here are some rough dumps of properties we can consider covering in a tuning guide (thanks to @danielkec) :

```
  # Depends on the workload and kernel version
  backlog: 8192
  receive-buffer-size: 64000
  write-queue-length: 32 # 0 means direct write

  # Depends on the workload and kernel version
  connection-options:
    # 0 mean indefinite(and less clutter on socket impl)
    read-timeout: PT0S     
    connect-timeout: PT0S
    
    # Can make quite difference between Mac, Linux kernels < 5.15 and Linux kernels > 5.15
    tcp-no-delay: true

    # when set, TCP autotuning is off(auto-tuning is better for variable or large payloads)
    socket-send-buffer-size: 64000     
    socket-receive-buffer-size: 64000

  # Turn off safety features
  # Carefull with this! Can be dangerous in production!
  protocols:
    "http_1_1":
      validate-request-headers: false
      validate-response-headers: false
      validate-path: false
      recv-log: false
      send-log: false
```

```
Some possible options to cover:

WebServer:
backlog
idle-connection-period
idle-connection-timeout
max-concurrent-requests
max-in-memory-entity
max-payload-size
max-tcp-connections
receive-buffer-size
write-buffer-size
write-queue-length

SocketOptions:
socket-receive-buffer-size
socket-send-buffer-size
tcp-no-delay
```

PR #8140, opened 2023-12-12
4.x tuning guide update

barchetta:
### Description

See #8102 

This PR removes the old tuning guides (which had incorrect information) and adds new guides than are more placeholders than full fledged documents. We need to decide what, if anything, we need to cover in a new tuning guide.

### Documentation

Provides updates to documentation.


spericas (approved):
LGTM

danielkec (approved):

PR #8142, opened 2023-12-12
4.x: Introduction of Helidon Inject Service module.

tomas-langer:
### Description
See readme in the module.

Introduces the minimal required set of types needed to define a service and to compile its code generated sources (code generation of the types is NOT part of this PR)


inject/service/src/main/java/io/helidon/inject/service/ServiceDescriptor.java
```diff
@@ -0,0 +1,111 @@
...
+        // and only add methods that are not already injected on existing type
+        Set<String> injectedMethods = myType.stream()
+                .filter(it -> it.elementKind() == ElementKind.METHOD)
+                .map(Ip::method)
+                .flatMap(Optional::stream)
+                .collect(Collectors.toSet());
+
+        result.addAll(superType.stream()
+                              .filter(it -> it.elementKind() == ElementKind.METHOD)
+                              .filter(it -> injectedMethods.add(it.method().orElse("")))
```
danielkec, line 106: Suspicious, IP with `elementType==METHOD` can have empty `.method()`?

inject/service/src/main/java/io/helidon/inject/service/Injection.java
```diff
@@ -0,0 +1,265 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.helidon.common.types.TypeName;
+
+import static java.lang.annotation.ElementType.TYPE;
```
spericas, line 27: (minor) ElementType.TYPE is used qualified and unqualified below

inject/service/src/main/java/io/helidon/inject/service/Injection.java
```diff
@@ -0,0 +1,265 @@
...
+    @Documented
+    @Retention(RetentionPolicy.CLASS)
+    @Service
+    @Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+    public @interface Singleton {
+        /*
+        Implementation note: we currently do not support custom scopes, so there is no Scope meta annotation.
+        If we decide to support scopes, we may want to introduce such an annotation.
+        */
+    }
```
spericas, line 138: What's the default scope if `@Singleton` isn't specified? Is that explained elsewhere?

spericas (approved):
Minor comments, LGTM.

Verdent (approved):

inject/service/src/main/java/io/helidon/inject/service/ServiceDescriptor.java
tomas-langer: because I do not like calling `.get()` on `Optional` when not checked

inject/service/src/main/java/io/helidon/inject/service/Injection.java
tomas-langer: Default is the same as annotated with `@Service`, I will update the docs (new instance injected per call to provider)

inject/service/src/main/java/io/helidon/inject/service/ServiceDescriptor.java
Verdent: Technically, you could use `.orElseThrow()`

inject/service/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>io.helidon.inject</groupId>
+        <artifactId>helidon-inject-project</artifactId>
+        <version>4.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>helidon-inject-service</artifactId>
```
trentjeff, line 29: maybe name this `inject-service-api`? Just feels weird to have inject-api depend on inject-service. Ideally it would have been reversed, right.

inject/service/src/main/java/io/helidon/inject/service/ServiceDescriptor.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * Inject fields and methods.
+     *
+     * @param ctx                  injection context
+     * @param interceptionMetadata interception metadata to support interception of field injection
+     * @param injected             mutable set of already injected methods from subtypes
+     * @param instance             instance to update
+     */
+
```
trentjeff, line 57: nit: remove space

trentjeff (approved):

inject/service/pom.xml
tomas-langer: I still think `inject-api` should go away, so we would have: `inject-service` and `inject`.
I do not want to name this module `inject-api`, as it would seem to be API of the Helidon Injection, which is not correct (as that is `Service` and related types). If we come up with better name than service, we can use it

trentjeff (approved):

inject/service/pom.xml
tomas-langer: Closing this conversation. We still have time to re-open it, as long as the branch remains open

PR #8147, opened 2023-12-13
2.x: Update gRPC version to 1.60.0

thegridman:
### Documentation

The gRPC Java libraries now have automatic module names in the jar files. These names differ slightly from those previously used by Helidon as they are all prefixed with io.grpc instead of just grpc. An issue resulting from this is that the gRPC API module has a module name of io.grpc which is the same as the repackaged gRPC module in Helidon. The Helidon repackaged io.grpc module no longer has any code, it is just a pom file with some dependencies. Its io.grpc module name is generated from the artifactId. We cannot change the artifactId but we can add a module-info to the jar to set its module name to be something other than io.grpc.

Usages of the repackaged io.grpc module have been removed from the rest of Helidon, other modules in Helidon now depend directly on whatever gRPC libraries they require. The repackaged io.grpc module has been left in place so as not to break any customer's that depend on it. Having said that, customer code that uses gRPC libraries is almost certainly going to break anyway due to the change of module names, which is beyond our control.

### DIFF
--- a/config/etcd/pom.xml
+++ b/config/etcd/pom.xml
@@ -57,8 +57,12 @@
         </dependency>
         <!-- etcd v3 -->
         <dependency>
-            <groupId>io.helidon.grpc</groupId>
-            <artifactId>io.grpc</artifactId>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-api</artifactId>
+        </dependency>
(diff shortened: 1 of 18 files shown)


danielkec (approved):

barchetta:
See #8106

PR #8155, opened 2023-12-15
4.x Several fixes to tracing config

### LINKED ISSUE #8148: Helidon Tracing configuration Paths
When adding Tracing and HealthChecks a Helidon 4.X appplication, traces are logged for health checks by default and cannot be disabled in configuration.

## Environment Details
* Helidon Version: 4.0.1
* Helidon SE
* JDK version: Zulu21.30+15-CA (build 21.0.1+12-LTS, mixed mode, sharing)
* OS: macOS Sonoma 14.2

----------

In `io.helidon.webserver.observe.tracing.TracingObserverConfig.BuilderBase#pathConfigs` the paths for `/metrics/*` and `/health/*` are by default disables - but in 4.X the path for the `ObserveFeature` was updated to `/observe/*` - so health becomes `/observe/health/*`.

This means that when deploying an Helidon 4.X application to Kubernetes we get a lot of traces from the health checks running.

Further, trying to disable this path by providing a `path` config in `application.yaml` does not work. [The documentation](https://helidon.io/docs/v4/#/se/tracing#_configuration) suggests the configuration key `tracing.paths` - but this doesn't seem to do anything. [The code](https://github.com/helidon-io/helidon/blob/50dd6c042f54bb5ee271c5b2ec360ff1ff969a44/webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserverConfigBlueprint.java#L80) seems to suggest the key `path-configs` 

```yaml
tracing:
  path-configs:
    - path: /observe/*
      traced-config: DISABLE
```

But that results in

> Exception in thread "main" io.helidon.config.ConfigMappingException: Mapping of value stored under key 'tracing.path-configs.0' to type 'interface io.helidon.webserver.observe.tracing.PathTracingConfig' has failed: No mapper configured

## Steps to reproduce

Create a Helidon 4.X application with both Tracing and HealthChecks enabled. Call `/observe/health` and see that a trace is logged. Try and disable in the config according to the documentation and also by looking at `TracingObserverConfig` (generated from blueprint).

tjquinno:
### Description
Resolves #8148 

Fixes included: 
1. Change `path-configs` to `paths` for config key within tracing to conform to the doc and to be more user-friendly.
2. Add observe-style paths for built-in services to disabled-by-default tracing paths.
3. Change the factory method and builder config method for `PathTracingConfig` to accept common `Config` as parameter instead of the impl `Config` type. This change is backward compatible with recompiling.

### Documentation
Bug fix. No doc change.

### DIFF
--- a/webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserverConfigBlueprint.java
+++ b/webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserverConfigBlueprint.java
@@ -58,14 +58,14 @@ interface TracingObserverConfigBlueprint extends ObserverConfigBase, Prototype.F
      * health could be enabled by setting {@code tracing.paths.0.path=/observe/health} and
      * {@code tracing.paths.0.enabled=true}.
      *
-     * We disable both the SE-style paths ({@code /observe/health}) and the MP-style paths ({@code /health}).
+     * By default we disable both the SE-style paths ({@code /observe/health}) and the MP-style paths ({@code /health}).
      */
     /**
      * Path specific configuration of tracing.
      *
      * @return configuration of tracing for specific paths
      */
-    @Option.Configured
+    @Option.Configured("paths")
     @Option.Singular
     @Option.DefaultCode("new @java.util.ArrayList@(@java.util.List@.of(PathTracingConfig.builder()\n"
             + "                                  .path(\"/metrics/*\")\n"
@@ -91,7 +91,7 @@ interface TracingObserverConfigBlueprint extends ObserverConfigBase, Prototype.F
             + "                                  .path(\"/observe/openapi/*\")\n"
             + "                                  .tracingConfig(TracingConfig.DISABLED)\n"
             + "                                  .build()))")
-    List<PathTracingConfig> paths();
+    List<PathTracingConfig> pathConfigs();
 
     /**
      * Tracer to use to extract inbound span context.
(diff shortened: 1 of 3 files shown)


webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserverConfigBlueprint.java
```diff
@@ -68,16 +70,28 @@ interface TracingObserverConfigBlueprint extends ObserverConfigBase, Prototype.F
...
             + "                                  PathTracingConfig.builder()\n"
             + "                                  .path(\"/openapi/*\")\n"
             + "                                  .tracingConfig(TracingConfig.DISABLED)\n"
+            + "                                  .build(), \n"
+            + "                                  PathTracingConfig.builder()\n"
+            + "                                  .path(\"/observe/openapi/*\")\n"
+            + "                                  .tracingConfig(TracingConfig.DISABLED)\n"
             + "                                  .build()))")
-    List<PathTracingConfig> pathConfigs();
+    List<PathTracingConfig> paths();
```
tomas-langer, line 94: This is a backward incompatible change in public API.
We can either consider it a bugfix (and alignment to docs), and change it, or we could have both methods (deprecate the `pathConfigs` and add a builder observer to correctly choose the right configured option)

tjquinno: Yes, I considered it a bug fix but I guess in my mind the real bug is the config key being `path-configs` instead of `paths`.

Another option: we could keep the old method name `pathConfigs` but set the config key explicitly to `paths` in the annotation.

tomas-langer: I think we need to keep the API backward compatible, so please either add a new method and deprecate existing (preferred option), or change the configuration key for this method.

tjquinno: The most recent push sets the config key explicitly to "paths" and changes the method names back to what they were to avoid the backward-incompatible change to the APIs.

tomas-langer (approved):

PR #8166, opened 2023-12-19
4.x Include scheme and port of origin and host in deciding whether to classify a request as CORS or not

### LINKED ISSUE #8088: 4.x CORS incorrectly treats a request containing origin and host headers that have the same node but different ports as a non-CORS request
## Environment Details
* Helidon Version: 4.0.1
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
In the CORS protocol, if the `Origin` and `Host` headers represent different locations then the request should be considered cross-origin. In this context, "location" includes the node name (or numeric address) and the port.

In 4.x, the Helidon CORS implementation incorrectly treats locations with the same node name but different ports as the same location. 

## Steps to reproduce
Slightly modify `examples/cors` to reproduce.
1. Add this to `logging.properties`:
   ```properties
   io.helidon.webserver.cors.level=FINER
   io.helidon.cors.level=FINER
   ```
2. `mvn clean package`
3. `java -jar target/helidon-examples-cors.jar`
4. Run this command:
   ```bash
   curl -i -X OPTIONS \
    -H "Access-Control-Request-Method: PUT" \
    -H "Origin: http://foo.com" \
    -H "Host: here.com" \
    http://localhost:8080/greet/greeting
   ```
   The HTTP response contains CORS headers, and the logging output from the server contains, among other things:
   ```
   [dump of the request] is cross-host: [header ORIGIN is present (http://foo.com), 
   header HOST is present (here.com), 
   header ORIGIN 'http://foo.com' does not match header HOST 'here.com']
   ```
   which is correct: the origin and host locations have different names (`foo.com` vs `here.com`) so it is a CORS request, and `http://foo.com` is a valid origin according to the CORS settings in the configuration.
5. Run this command in which the origin and host differ only in the port:
   ```bash
    curl -i -X OPTIONS \
    -H "Access-Control-Request-Method: PUT" \
    -H "Origin: http://foo.com" \
    -H "Host: foo.com:8010" \
    http://localhost:8080/greet/greeting
   ```
   The response is 404 because:
   * CORS has incorrectly determined that this is not a CORS request,
   * CORS has therefore passed the request through to the normal routing processing, and
   * there is no application endpoint for `OPTIONS` at `/greet/greeting`.
   
   The server log contains this (with some redactions for brevity):
   ```
   [dump of the request] is not cross-host: 
   [header ORIGIN 'http://foo.com' matches header HOST 'foo.com']
   passing normal request through unchanged
   ```

With the Helidon 3 version of the same example, these exact same steps yield correct results: CORS correctly identifies the last request as cross-origin, as shown by the response and the server logging output.

### LINKED ISSUE #8093: [4.x] CORS incorrectly disregards whether the incoming request is secure or not in deciding whether it's a CORS request
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
As explained here https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS (near the top):
> Cross-Origin Resource Sharing ([CORS](https://developer.mozilla.org/en-US/docs/Glossary/CORS)) is an [HTTP](https://developer.mozilla.org/en-US/docs/Glossary/HTTP)-header based mechanism that allows a server to indicate any [origins](https://developer.mozilla.org/en-US/docs/Glossary/Origin) (domain, _**scheme**_, or port) other than its own from which a browser should permit loading resources.

(emphasis added)

The incoming request determines the server's origin, including the scheme. 

The Helidon CORS processing does not consider whether the request is secure or not (that is, whether its scheme is `http` or `https`) in deciding whether the origin conveyed in the request's headers is the same as the server's origin.

## Steps to reproduce
Use the `examples/cors` example.

1. Edit `src/main/resources/logging.properties` to contain these lines (uncommented):
   ```properties
   io.helidon.webserver.cors.level=FINER
   io.helidon.cors.level=FINER
   ```
2. `mvn clean package`
3. `java -jar target/helidon-examples-cors.jar`
4. ```bash
   curl -i -X OPTIONS \
       -H "Access-Control-Request-Method: PUT" \
       -H "Origin: https://here.com" \
       -H "Host: here.com" \
       http://localhost:8080/greet/greeting`
   ```
The response is 404, because CORS incorrectly concluded the origin (`https` in the `Origin` header) and host (`http` as derived from the request in the code) shared the same location because it does not consider the scheme of the server's origin.

The logging output from the server contains this, also reflecting the error:
```list
is not cross-host: [header ORIGIN 'https://here.com' matches header HOST 'here.com']
```
which is true as far as it goes, but the CORS decision-making should also consider the host and origin schemes which it does not.

tjquinno:
### Description
Resolves #8088 
Resolves #8093 

Proper implementation of the CORS algorithm requires considering the scheme, node, and port of the origin (conveyed by the `Origin` header) and the host that was requested by the client (conveyed by the `Host` header or the `Forwarded` header or the `X-Forwarded-*` family of headers). A normal request in the CORS context is one that is _not_ cross-origin--that is, for which the origin and the host share the same scheme, node, and port.

The 4.x logic regressed from 3.x in not considering the scheme and port properly.

This PR does several things:
1. Adds an `isSecure` method to an adapter interface for requests. 
   
   The adapter insulates the CORS logic from differences in and dependence on the SE and MP request implementations. The PR also includes the enhancements to the SE and MP adapters to implement this new method.
3. Corrects the CORS logic which checks to see if the incoming request is a normal request (that is, non-CORS). 
   
   The CORS code first checks for whether the request is normal and, if so, skips the CORS processing. This normal check  failed to account properly for the scheme and port of the origin and host and therefore _incorrectly_ identified requests that needed CORS processing as normal (non-CORS) requests.
4. Adds numerous tests.

The CORS code does considerable logging (when turned on) to help users understand the decisions the code makes. This PR also introduces a new private record which represents the result of the "is the request normal" check, containing not only the true/false result but also the effective origin and host locations (full scheme, node, and port) that the is-normal check computed to decide the true/false result. The logging code now uses these intermediate results rather than having to recompute them for logging.

### Documentation
This is a bug fix. No doc impact.

### DIFF
--- a/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
+++ b/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
@@ -371,7 +371,7 @@ public class CorsSupportHelper<Q, R> {
     }
 
     // For testing
-    static RequestTypeInfo isRequestTypeNormal(String originHeader, UriInfo requestedHostUri) {
+    static RequestTypeInfo requestType(String originHeader, UriInfo requestedHostUri) {
         return RequestTypeInfo.create(originHeader, requestedHostUri);
     }
 
@@ -406,7 +406,7 @@ public class CorsSupportHelper<Q, R> {
             return true;
         }
 
-        RequestTypeInfo result = isRequestTypeNormal(originOpt.get(), requestAdapter.requestedUri());
+        RequestTypeInfo result = requestType(originOpt.get(), requestAdapter.requestedUri());
         LogHelper.logIsRequestTypeNormal(result.isNormal,
                                          silent,
                                          requestAdapter,
--- a/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
+++ b/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -371,8 +371,8 @@ public class CorsSupportHelper<Q, R> {
     }
 
     // For testing
-    static RequestTypeInfo isRequestTypeNormal(String originHeader, UriInfo requestedHostUri, boolean isRequestSecure) {
-        return RequestTypeInfo.create(originHeader, requestedHostUri, isRequestSecure);
+    static RequestTypeInfo isRequestTypeNormal(String originHeader, UriInfo requestedHostUri) {
+        return RequestTypeInfo.create(originHeader, requestedHostUri);
     }
 
     /**
@@ -380,18 +380,18 @@ public class CorsSupportHelper<Q, R> {
      * <p>
      *     We want to use the intermediate results for clearer logging if that's turned on without having to recompute it.
      * </p>
-     * @param originLocation full origin (including port)
+     * @param originLocation full origin (including scheme and port)
      * @param hostLocation   full host (including scheme and port)
      * @param isNormal       whether the origin and host information represent a normal (non-CORS) request
      */
     record RequestTypeInfo(String originLocation, String hostLocation, boolean isNormal) {
 
-        static RequestTypeInfo create(String originHeader, UriInfo requestedHostUri, boolean isRequestSecure) {
+        static RequestTypeInfo create(String originHeader, UriInfo requestedHostUri) {
             String originLocation = CorsSupportHelper.originLocation(originHeader);
             String hostLocation = CorsSupportHelper.hostLocation(requestedHostUri);
 
-            return new RequestTypeInfo(CorsSupportHelper.originLocation(originHeader),
-                                       CorsSupportHelper.hostLocation(requestedHostUri),
+            return new RequestTypeInfo(originLocation,
+                                       hostLocation,
                                        originLocation.equals(hostLocation));
         }
     }
@@ -406,7 +406,7 @@ public class CorsSupportHelper<Q, R> {
             return true;
         }
 
-        RequestTypeInfo result = isRequestTypeNormal(originOpt.get(), requestAdapter.requestedUri(), requestAdapter.isSecure());
+        RequestTypeInfo result = isRequestTypeNormal(originOpt.get(), requestAdapter.requestedUri());
         LogHelper.logIsRequestTypeNormal(result.isNormal,
                                          silent,
                                          requestAdapter,
--- a/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
+++ b/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
@@ -29,6 +29,7 @@ import java.util.function.Supplier;
 import java.util.logging.Logger;
 
 import io.helidon.common.config.Config;
+import io.helidon.common.uri.UriInfo;
 import io.helidon.cors.LogHelper.Headers;
 import io.helidon.http.HeaderName;
 import io.helidon.http.HeaderNames;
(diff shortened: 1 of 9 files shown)


tjquinno:
Converted to draft temporarily - there is another CORS-related issue I might also fix in this PR.

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
```diff
@@ -369,14 +370,68 @@ public Aggregator aggregator() {
...
+    static RequestTypeInfo isRequestTypeNormal(String originHeader, UriInfo requestedHostUri, boolean isRequestSecure) {
+        return RequestTypeInfo.create(originHeader, requestedHostUri, isRequestSecure);
+    }
+
+    /**
+     * Captures intermediate data and the final result in deciding whether a request is a normal (non-CORS) request or not.
+     * <p>
+     *     We want to use the intermediate results for clearer logging if that's turned on without having to recompute it.
+     * </p>
+     * @param originLocation full origin (including port)
```
ljnelson, line 383: (Layman question: but not scheme?)

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
```diff
@@ -369,14 +370,68 @@ public Aggregator aggregator() {
...
+     * @param hostLocation   full host (including scheme and port)
+     * @param isNormal       whether the origin and host information represent a normal (non-CORS) request
+     */
+    record RequestTypeInfo(String originLocation, String hostLocation, boolean isNormal) {
+
+        static RequestTypeInfo create(String originHeader, UriInfo requestedHostUri, boolean isRequestSecure) {
```
ljnelson, line 389: `isRequestSecure` is ignored? Or my eyes are bad?

ljnelson (dismissed):
Couple of questions but basically LGTM

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
tjquinno: The scheme is part of the origin location as well. Fixed comment.

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
tjquinno: Good point. It turns out that the `isRequestSecure` method is redundant with how the requested host URI is prepared. That URI always has the correct scheme, so we don't need `isRequestSecure`.

I've removed that method (added in this PR) from the interface and implementations, but I've kept those classes in the PR because I touched them and so I need to make the copyright year 2024 now even though they contain no other changes.

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
```diff
@@ -369,14 +370,68 @@ public Aggregator aggregator() {
         return aggregator;
     }
 
-    private boolean isRequestTypeNormal(CorsRequestAdapter<Q> requestAdapter, boolean silent) {
+    // For testing
+    static RequestTypeInfo isRequestTypeNormal(String originHeader, UriInfo requestedHostUri) {
```
ljnelson, line 374: Sorry; missed this before. Kind of an odd method name given that it does not return a `boolean`.

cors/src/main/java/io/helidon/cors/LogHelper.java
```diff
@@ -68,8 +68,21 @@ void setAndLog(BiConsumer<HeaderName, Object> consumer, String note) {
         }
     }
 
+
+    static <T> void logIsRequestTypeNormalNoOrigin(boolean silent, CorsRequestAdapter<T> requestAdapter) {
+        if (silent || !CorsSupportHelper.LOGGER.isLoggable(DECISION_LEVEL)) {
+            return;
+        }
+        if (CorsSupportHelper.LOGGER.isLoggable(DECISION_LEVEL)) {
```
ljnelson, line 76: You don't need this check, right? You already made it?

ljnelson (dismissed):
A couple of comments

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
tjquinno: Renamed.

cors/src/main/java/io/helidon/cors/LogHelper.java
tjquinno: Correct. Fixed.

ljnelson (approved):

PR #8175, opened 2023-12-20
4.x Make `RegistryFactory` and its `getInstance` and `getRegistry` methods public

### LINKED ISSUE #8161: 4.x MP metrics doc refers to `RegistryFactory` which is package private in 4.x
## Environment Details
* Helidon Version: 4.0.1
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
The MP metrics doc refers to the `RegistryFactory` which used to be public in SE and MP in 3.x but is now package private in MP and gone from SE.

We might need to make the class public again and make at least the static `getInstance()` method (no args) and the instance method `getRegistry(String)` method public. It's reasonable for an application to determine a scope name for a metric registry at runtime and need to be able to retrieve it. 

(If the developer knows the scope at build-time, s/he can use injection to obtain the corresponding `MetricRegistry`.)

tjquinno:
### Description
Resolves #8161 

In 3.x, the `RegistryFactory` class was public in the metrics API module and usable from both SE and MP. It had public methods `getInstance()` and `getRegistry()` and was injectable.

In 4.x, we moved it to MP because it no longer applied in SE with the new neutral metrics API. At that time, I also made the class package-private because `RegistryFactory` only had a role in MP metrics now, and I anticipated that users would `@Inject MetricRegistry` rather than work with the `RegistryFactory`.

The 4.x doc still referred to `RegistryFactory` and, in fact, it is still a valid use case to either `@Inject` the `RegistryFactory` or retrieve it programmatically using a static method so as to invoke its `getRegistry` method to locate a `MetricRegistry` programmatically. As we found out in our internal Slack channel, users had used these approaches in 3.x and we broke their code by removing this capability.

Restoring these features, essentially to how they worked in 3.x, enlarges our surface area only very slightly as compared to 4.0.1 and allows existing user code to continue to work with only a change in `import` statements (because `RegistryFactory` is in a different package in 4.x).

This PR does the following:
1. Make `RegistryFactory` and its `getInstance()` and `getRegistry(String)` methods public.
   1. Change the return type of `getRegistry` from `Registry` (a package-private class) to `MetricRegistry` (an MP metrics  interface which `Registry` implements).
   2. Add a package-private method `registry(String)` to `RegistryFactory` which returns a `Registry` for use by our own methods in the package.
2. Update the MP metrics doc to clarify and show how to get and use a `RegistryFactory` instance.

### Documentation
Changes included in the PR.

### DIFF
--- a/docs/src/main/asciidoc/includes/metrics/metrics-shared.adoc
+++ b/docs/src/main/asciidoc/includes/metrics/metrics-shared.adoc
@@ -153,24 +153,70 @@ In addition to your application {metrics}, the reports contain other
 To register or look up {metrics} programmatically, your service code uses the link:{microprofile-metrics-javadoc-url}/org/eclipse/microprofile/metrics/MetricRegistry.html[`MetricRegistry`] instance for the scope of interest: `base`, `vendor`, `application`, or a custom scope.
 
 ifdef::mp-flavor[]
-To get a `MetricRegistry` reference
+Either of the following techniques gets a `MetricRegistry` reference.
+Remember that injection works only if the class is a bean so CDI can inject into it.
 
-* `@Inject` the metric registry you want, perhaps also using the link:{microprofile-metrics-javadoc-annotation-url}/RegistryScope.html[`@RegistryScope`] annotation to select the registry type, or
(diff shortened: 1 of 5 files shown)


ljamen (approved):
Small doc updates - LGTM.

Issue #8191, opened 2024-01-04
OCI integration using Instance Principal fails


klustria:
## Environment Details
* Helidon Version: 4x
* Helidon MP

----------

## Problem Description
OCI Instance Principal Authentication fails with this error:

```
023.12.19 04:29:17 SEVERE io.helidon.inject.runtime.AbstractServiceProvider Thread[#1,main,5,main]: unable to activate: io.helidon.integrations.oci.sdk.runtime.OciAuthenticationDetailsProvider$$Injection$$Activator
java.util.NoSuchElementException: No instances of com.oracle.bmc.auth.AbstractAuthenticationDetailsProvider available for use. Verify your configuration named: oci
        at io.helidon.integrations.oci.sdk.runtime.OciAuthenticationDetailsProvider.select(OciAuthenticationDetailsProvider.java:123)
        at io.helidon.integrations.oci.sdk.runtime.OciAuthenticationDetailsProvider.first(OciAuthenticationDetailsProvider.java:94)
        at io.helidon.inject.runtime.AbstractServiceProvider.first(AbstractServiceProvider.java:338)
        at io.helidon.inject.runtime.AbstractServiceProvider.get(AbstractServiceProvider.java:324)
        at io.helidon.integrations.oci.sdk.runtime.OciExtension.lambda$ociAuthenticationProvider$4(OciExtension.java:203)
        at io.helidon.integrations.oci.sdk.cdi.OciExtension.lambda$installAdps$2(OciExtension.java:861)
        at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$CreateCallback.create(BeanConfiguratorImpl.java:372)
        at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$ImmutableBean.create(BeanConfiguratorImpl.java:511)
        at org.jboss.weld.contexts.AbstractContext.get(AbstractContext.java:96)
        ...
        ...
        ...
        at io.helidon.microprofile.cdi.HelidonContainerImpl.start(HelidonContainerImpl.java:256)
        at io.helidon.microprofile.cdi.Main.main(Main.java:80)
```

## Steps to reproduce
1.  Generate the latest Helidon MP 4 Quickstart with OCI using `helidon` tool.
2.  Build and deploy the app to an OCI instance. 
3.  Make sure that the home directory where the app is deployed does not have an oci config, i.e. no `~/.oci` directory.
4.  Run the application and the error will be reproduced

[...]

PR #8197, opened 2024-01-05
Fix issue for checking if app is running in an OCI instance which causes Instance Principal auth to fail

klustria:
### Description

The call to `Region.getRegionFromImds()` in `OciAvailabilityDefault.runningOnOci(OciConfig ociConfig)` is returning null causing Instance Principal authentication to not work because the code thinks it is not running in an OCI instance. To resolve this issue, passing in the Metadata Service Url in `Region.getRegionFromImds()` helps return a valid value.

Tested and validated this change to be working in a helidon app that is running in an OCI instance.

### Documentation

If no doc impact: None

### DIFF
--- a/integrations/oci/sdk/runtime/src/main/java/io/helidon/integrations/oci/sdk/runtime/OciAvailabilityDefault.java
+++ b/integrations/oci/sdk/runtime/src/main/java/io/helidon/integrations/oci/sdk/runtime/OciAvailabilityDefault.java
@@ -37,7 +37,7 @@ import static com.oracle.bmc.auth.AbstractFederationClientAuthenticationDetailsP
 @Singleton
 @Weight(ServiceInfoBasics.DEFAULT_INJECT_WEIGHT)
 class OciAvailabilityDefault implements OciAvailability {
-    static String OPC_PATH = getOpcPath();
+    private static final String OPC_PATH = getOpcPath();
 
     @Override
     public boolean isRunningOnOci(OciConfig ociConfig) {
@@ -74,7 +74,7 @@ class OciAvailabilityDefault implements OciAvailability {
         }
     }
 
-    static String getOpcPath() {
+    private static String getOpcPath() {
         String input = METADATA_SERVICE_BASE_URL;
         int index = -1;
         for (int nth = 3; nth > 0; nth--) {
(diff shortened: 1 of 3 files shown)


integrations/oci/sdk/runtime/src/main/java/io/helidon/integrations/oci/sdk/runtime/OciAvailabilityDefault.java
```diff
@@ -46,7 +46,7 @@ static boolean runningOnOci(OciConfig ociConfig) {
             return false;
         }
 
-        return (Region.getRegionFromImds() != null);
+        return (Region.getRegionFromImds("http://" + ociConfig.imdsHostName() + "/opc/v2/") != null);
```
trentjeff, line 49: What is the issue with calling the no arg method?  It seems like we should not need to pass the path.

If we really must take ownership of the path, then "/opc/v2/" should be configurable.  We should therefore follow-up to make that configurable - perhaps using `oci.yaml`.

klustria: without the arg, it is returning null. It can be fixed by using Region.registerFromInstanceMetadataService() but this uses http://169.254.169.254/opc/v2/. Our code however allows IMDS host to be configurable via ociConfig.imdsHostName(), hence why we cannot use  Region.registerFromInstanceMetadataService(). The IMDS URL is in the format, http://<IMDS HOST>/opc/v2/ hence why I derived it and hence why "/opc/v2/" does not need to be configurable?

klustria:
This PR is potential fix for https://github.com/helidon-io/helidon/issues/8191

integrations/oci/sdk/runtime/src/main/java/io/helidon/integrations/oci/sdk/runtime/OciAvailabilityDefault.java
```diff
@@ -71,4 +74,15 @@ static boolean canReach(String address,
         }
     }
 
+    private static String getOpcPath() {
+        String input = METADATA_SERVICE_BASE_URL;
+        int index = -1;
+        for (int nth = 3; nth > 0; nth--) {
+            index = input.indexOf("/", index + 1);
+            if (index == -1) {
```
trentjeff, line 82: Is this really what we want to do here, return null? Seems like we should be failing fast with an exception.  And since we want to catch this when we bump OCI versions, we should write a test for this method in the PR too.  That way we catch it in our build environment and not in the users.

klustria: Updated based on comments, i.e. getOpcPath() now throwing exception during failure and also added unit test coverage

trentjeff (approved):

Issue #8189, opened 2024-01-03
WebSocket upgrade request missing the Connection header


4e6:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 4.0.2
* Helidon SE 
* JDK version: openjdk 21.0.1
* OS: Archlinux
* Docker version (if applicable): -

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

When trying to create a WebSocket connection with `WsClient`, it fails to upgrade the connection

```
Exception in thread "main" io.helidon.webclient.websocket.WsClientException: Failed to upgrade to WebSocket. Response: 200 OK: Via: 1.1 fly.io
Content-Length: 630
Content-Type: text/plain
Date: Wed, 03 Jan 2024 19:57:01 GMT
Server: Fly/ec8196c09 (2024-01-02)
fly-request-id: 01HK8ED5H2PDAM0NC27TD3ZGQD-lhr

	at io.helidon.webclient.websocket@4.0.2/io.helicon.webclient.websocket.WsClientImpl.connect(WsClientImpl.java:123)
```

According to the [Mozilla doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#upgrading_to_a_websocket_connection), a WebSocket upgrade request should contain two headers:
```
Connection: Upgrade
Upgrade: websocket
```
Debugging showed that the upgrade request sets the `Upgrade` header but is missing the `Connection` one https://github.com/helidon-io/helidon/blob/68de0bb538483c5f76fdf78c686e273f15b4c684/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java#L130-L133

After setting the `Connection` header (see the example below), the connection was successfully established.

## Steps to reproduce
[//]: # "Step by step instructions to reproduce the problem"
[//]: # "Provide sample code/application if relevant"

This example results in an `io.helidon.webclient.websocket.WsClientException: Failed to upgrade to WebSocket.`

```
io.helidon.webclient.websocket.WsClient.builder()
    .build()
    .connect(
        new java.net.URI("wss://echo.websocket.org/"), 
        new io.helidon.websocket.WsListener() {}
    );
 ```

[...]

PR #8198, opened 2024-01-05
The Helidon WS client must include a Connection header

spericas:
### Description

The Helidon WS client must include a 'Connection' header with a value of 'Upgrade' as part of the initial request. Issue #8189.

### Documentation

None

### DIFF
--- a/webserver/tests/websocket/src/main/java/io/helidon/webserver/tests/websocket/EchoService.java
+++ b/webserver/tests/websocket/src/main/java/io/helidon/webserver/tests/websocket/EchoService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


webserver/tests/websocket/src/main/java/io/helidon/webserver/tests/websocket/EchoService.java
```diff
@@ -54,6 +55,10 @@ public void onClose(WsSession session, int status, String reason) {
     @Override
     public Optional<Headers> onHttpUpgrade(HttpPrologue prologue, Headers headers) throws WsUpgradeException {
         WritableHeaders<?> upgradeHeaders = WritableHeaders.create();
+
+        if (!headers.get(HeaderNames.CONNECTION).get().equals("Upgrade")) {
```
ljnelson, line 59: Naïve question: is case significant?

spericas: From what I could gather, yes

ljnelson: I ask because see also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism

spericas: I see, I will switch to insensitive for the test then

ljnelson (approved):

PR #8206, opened 2024-01-08
4.x: TLS default config values

### LINKED ISSUE #8127: 4.x: Check TLS session timeout
in `TlsConfigBlueprint`, we set `sessionTimeout` to 30 minutes (default value).
This is very different from 3.x, where the default was 0 - meaning that we did not update Java default (which is 24 hours).

This may cause regressions for long-running connections over TLS.

danielkec:
Fixes #8127

### Helidon 2, 3
While in Helidon 2 and 3 following logic was used for `sessionCacheSize` and `sessionTimeoutSeconds`:
```java
private long sessionCacheSize; // def 0
private long sessionTimeoutSeconds; //def 0
...
int sessionCacheSize = tlsConfig.sessionCacheSize();
if (sessionCacheSize > 0) {
  serverSessionContext.setSessionCacheSize(sessionCacheSize);
}
int sessionTimeoutSecs = tlsConfig.sessionTimeoutSeconds();
if (sessionTimeoutSecs > 0) {
  serverSessionContext.setSessionTimeout(sessionTimeoutSecs);
}
```

Resulting in an actual defaults being the unset defaults of [SSLSessionContext](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/javax/net/ssl/SSLSessionContext.html#getSessionCacheSize()):

- void setSessionTimeout(int seconds);  ----> `86400`
 seconds – the new session timeout limit in seconds; zero means there is no limit.
The JDK implementation returns the session timeout as set by the setSessionTimeout method, or if not set, a default value of `86400` seconds (24 hours).

 - void setSessionCacheSize(int size); ----> `20480`
 size – the new session cache size limit; zero means there is no limit.
 The JDK implementation returns the cache size as set by the setSessionCacheSize method, or if not set, the value of the javax.net.ssl.sessionCacheSize system property. If neither is set, it returns a default value of `20480`.  

### Helidon 4
In Helidon 4 we are currently using:
```java
@ConfiguredOption("1024")
int sessionCacheSize();
@ConfiguredOption("PT30M")
Duration sessionTimeout();
...
SSLSessionContext serverSessionContext = sslContext.getServerSessionContext();
if (serverSessionContext != null) {
  serverSessionContext.setSessionCacheSize(tlsConfig.sessionCacheSize());
  serverSessionContext.setSessionTimeout((int) tlsConfig.sessionTimeout().toSeconds());
}
```

This PR alignes the defaults with previous versions of Helidon.

### DIFF
--- a/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -50,7 +50,6 @@ public class ConfiguredTlsManager implements TlsManager {
     // secure random cannot be stored in native image, it must
     // be initialized at runtime
     private static final LazyValue<SecureRandom> RANDOM = LazyValue.create(SecureRandom::new);
-
     private final String name;
     private final String type;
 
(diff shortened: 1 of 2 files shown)


common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2024 Oracle and/or its affiliates.
```
tomas-langer, line 2: This is wrong copyright update, should be 2023, 2024

common/tls/src/main/java/io/helidon/common/tls/TlsConfigBlueprint.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2024 Oracle and/or its affiliates.
```
tomas-langer, line 2: dtto

tomas-langer (changes requested):

tomas-langer (approved):

PR #8210, opened 2024-01-08
4.x Allow programmatic look-up of MetricRegistry via CDI without NPE

### LINKED ISSUE #8209: 4.x `CDI.current().select(MetricRegistry.class).get()` throws NPE
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
Attempting a programmatic look-up of `MetricRegistry` triggers an NPE.

## Steps to reproduce
1. In the MP QuickStart app, add the following line at the beginning of `GreetResource#getDefaultMessage`:
   ```
   MetricRegistry appRegistry = CDI.current().select(MetricRegistry.class).get();
   ```
2. `mvn clean package`
The unit tests fail because of an NPE in `RegistryProducer` because the producer method assumes the `InjectionPoint` passed is non-null as is its `getAnnotated()` value but in this case `getAnnotated()` is, reasonably, null.

tjquinno:
### Description
Resolves #8209 

The producer method failed to check that the injection point and its `getAnnotated()` return value were non-null before using them.  Thus `CDI.current().select(MetricRegistry.class).get()` triggered an NPE.

(It turned out that the `InjectionPoint` passed is non-null but its `getAnnotated()` return value _was_ null in this case.)

### Documentation
Bug fix - no doc impact

### DIFF
--- a/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/RegistryProducer.java
+++ b/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/RegistryProducer.java
@@ -41,8 +41,8 @@ final class RegistryProducer {
     @Produces
     @Default
     public static org.eclipse.microprofile.metrics.MetricRegistry getScopedRegistry(InjectionPoint injectionPoint) {
-        Annotated annotated = injectionPoint == null ? null : injectionPoint.getAnnotated();
-        RegistryScope scope = annotated == null ? null : annotated.getAnnotation(RegistryScope.class);
+        Annotated annotated = (injectionPoint == null) ? null : injectionPoint.getAnnotated();
+        RegistryScope scope = (annotated == null) ? null : annotated.getAnnotation(RegistryScope.class);
         return scope == null
                 ? getApplicationRegistry()
                 : RegistryFactory.getInstance().getRegistry(scope.scope());
--- a/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/RegistryProducer.java
+++ b/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/RegistryProducer.java
@@ -1,5 +1,5 @@
 /**
- * Copyright (c) 2018, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/ProducerTest.java
```diff
@@ -89,4 +90,12 @@ void testRegistryFactoryProducer() {
         MetricRegistry customRegistry = registryFactory.getRegistry("myCustomScope");
         assertThat("Custom scoped MetricRegistry", customRegistry, notNullValue());
     }
+
+    @Test
+    void testDirectMetricRegistryLookup() {
+        MetricRegistry appRegistry = CDI.current().select(MetricRegistry.class).get();
+        assertThat("Directly-looked-up app registry", appRegistry, notNullValue());
+
```
spericas, line 98: Empty lines?

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/RegistryProducer.java
```diff
@@ -40,7 +41,8 @@ private RegistryProducer() {
     @Produces
     @Default
     public static org.eclipse.microprofile.metrics.MetricRegistry getScopedRegistry(InjectionPoint injectionPoint) {
-        RegistryScope scope = injectionPoint.getAnnotated().getAnnotation(RegistryScope.class);
+        Annotated annotated = injectionPoint == null ? null : injectionPoint.getAnnotated();
+        RegistryScope scope = annotated == null ? null : annotated.getAnnotation(RegistryScope.class);
```
spericas, line 45: Minor, parentheses around condition help readability in this case IMO.

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/ProducerTest.java
tjquinno: Fixed

microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/RegistryProducer.java
tjquinno: Changed

spericas (approved):

PR #8212, opened 2024-01-08
Add a little logic to CORS config processing and significantly update the CORS doc

### LINKED ISSUE #8026: 4.x CORS configuration ignored for automatically-discovered Helidon-provided services (health, metrics)
## Environment Details
* Helidon Version: 4.0.0
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
The Helidon-provided services (OpenAPI, metrics, health...maybe new observers as well??) should--and are documented as--supporting CORS via configuration.

But seemingly the service-specific configuration is ignored.

This could be related to #8024 as opposed to something specific to the built-in services, but double-checking the built-in service support for CORS merits its own issue.

This could be a side-effect of converting metrics and health to observers that are discovered implicitly. In 3.x the doc showed how to construct the `HealthSupport` or `MetricsSupport` instance and also provide the CORS settings as the services were added to the routing.

## Steps to reproduce
1. Start with the SE CORS example (because it already has the CORS-related dependencies).
2. Add the following to `application.yaml`:
   ```
   health:
     cors:
       allow-origins: ["http://here.com"]
       allow-methods: ["GET", "HEAD"]
   ```
3. Build and start the server.
4. Run `curl -i -H "Origin: http://bad.com" -H "Host: http://localhost:8080/observe/health` which should be rejected (bad.com is not one of the allowed origins) but the server returns the normal health response of 204.

tjquinno:
### Description
Resolves #8026 

The change in 4.x in how built-in features such as metrics and health are automatically discovered and activated did not revise the CORS doc so the descriptions particularly for how to control the CORS behavior for those features had fallen out of step with how Helidon behaves.

Also, because these services no longer require explicit developer code to add them to an SE web server, we needed to allow CORS to rely on configuration to establish the settings for those services. 

This PR contains two important changes:
1. It adds logic to the CORS `Aggregator` to allow the user to set up CORS behavior for _any_ endpoint--including the built-in services--using only configuration.
   
   (The role of the `Aggregator` in CORS has always been to pull together the CORS set-up from multiple places: explicitly coded `CorsSupport` instances, config, and MP annotations.)
2. It significantly updates the CORS-related doc:
   1. The doc now emphasizes the config-only approach and recommend it. This is much simpler, and it is functionally equivalent to the previous way in which developers created one or more `CorsSupport` instances and included it them in the routing for their services or endpoints.
   2. The doc now (correctly) describes how to control CORS for the built-in services--using config.

### Documentation
The PR includes doc updates.

### DIFF
--- a/cors/src/main/java/io/helidon/cors/Aggregator.java
+++ b/cors/src/main/java/io/helidon/cors/Aggregator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


tomas-langer (approved):
I cannot really review the docs language, for the rest I am OK

PR #8213, opened 2024-01-09
Remove references to any JDK from examples

senivam:
### Description

Removes references to JDKs from examples


senivam:
there is one reference to JDK (21) in archetypes:
```
archetypes/helidon/src/main/archetype/common/files/README.md.mustache
```
if it shall also be removed please let me know

barchetta (approved):
LGTM. I think it is OK to leave the JDK pre-req in the archetype since that is just one place to change.

PR #8216, opened 2024-01-09
4.x Add support for `@SpanAttribute` annotation, use entire path for REST resource span name

### LINKED ISSUE #8157: 4.x Add support for Otel `@SpanAttribute` annotation
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
As described in the OTel instrumentation doc [here](https://opentelemetry.io/docs/instrumentation/java/automatic/annotations/#adding-attributes-to-the-span-with-spanattribute) the `@SpanAttribute` annotation allows the developer to specify that each such annotated method parameter is to be added as an attribute to the telemetry span for that method. 

The MP telemetry spec specifically mentions this annotation in [its `@WithSpan` section](https://download.eclipse.org/microprofile/microprofile-telemetry-1.0/tracing/microprofile-telemetry-tracing-spec-1.0.html#_withspan).

Currently Helidon's MP telemetry implementation does not process `@SpanAttribute`.

### LINKED ISSUE #8162: 4.x OpenTelemetry span name includes only the last path segment, not the entire path
## Environment Details
* Helidon Version: 4.0.1
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
OpenTelemetry span names include only the last segment of the path, not the entire path.

This is a change from Helidon's prior OpenTracing behavior and seems to conflict with the OpenTelemetry semantic conventions for HTTP server-side [span naming](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name):
> HTTP server span names SHOULD be {method} {http.route}

`method` is the HTTP method (e.g., `GET`) and `http.route` is defined [here](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-semantic-conventions):
> The matched route, that is, the path template in the format used by the respective server framework.

## Steps to reproduce
1. Create a resource class with a `@Path`.
2. Add a subresource method with its own `@Path` specifying a subpath and with `@WithSpan`. 
The resulting span is named using only the last path segment, not the entire path.

tjquinno:
### Description
Resolves #8157 
Resolves #8162 

Highlights:
* Enhance `WithSpanInterceptor` so it also processes `@SpanAttribute` annotations.
* Add tests to check span name and attributes.
* Fix derivation of OTel span name for REST endpoints to reflect OpenTelemetry span naming semantic conventions.
* Add tests for REST endpoint span naming.

### Documentation
Bug fix; no doc impact.

### DIFF
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -15,6 +15,8 @@
  */
 package io.helidon.microprofile.telemetry;
 
+import java.util.Deque;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 
@@ -166,20 +168,20 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
         ExtendedUriInfo extendedUriInfo = (ExtendedUriInfo) requestContext.getUriInfo();
 
         // Derive the original path (including path parameters) of the matched resource from the bottom up.
-        StringBuilder derivedPath = new StringBuilder();
+        Deque<String> derivedPath = new LinkedList<>();
 
         Resource resource = extendedUriInfo.getMatchedModelResource();
         while (resource != null) {
-            derivedPath.insert(0, resource.getPath());
+            derivedPath.push(resource.getPath());
             if (!resource.getPath().startsWith("/")) {
-                derivedPath.insert(0, "/");
+                derivedPath.push("/");
             }
             resource = resource.getParent();
         }
 
-        derivedPath.insert(0, applicationPath())
-                .insert(0, requestContext.getMethod() + " ");
-        return derivedPath.toString();
+        derivedPath.push(applicationPath());
+        derivedPath.push(requestContext.getMethod() + " ");
+        return String.join("", derivedPath);
     }
 
     private String applicationPath() {
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -156,14 +156,16 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
     }
 
     private String spanName(ContainerRequestContext requestContext) {
-        // According to OpenTelemetry semantic conventions for spans, the span name for a REST endpoint should be
+        // According to recent OpenTelemetry semantic conventions for spans, the span name for a REST endpoint should be
         //
         // http-method-name low-cardinality-path
         //
         // where a low-cardinality path would be, for example /greet/{name} rather than /greet/Joe, /greet/Dmitry, etc.
+        // But the version of semantic conventions in force when the MP Telemetry spec was published did not include the
+        // http-method-name. So our code omits that for now to pass the MP Telemetry TCK.
 
         if (spanNameFullUrl) {
-            return requestContext.getMethod() + " " + requestContext.getUriInfo().getAbsolutePath().toString();
+            return requestContext.getUriInfo().getAbsolutePath().toString();
         }
         ExtendedUriInfo extendedUriInfo = (ExtendedUriInfo) requestContext.getUriInfo();
 
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -172,15 +172,17 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
 
         Resource resource = extendedUriInfo.getMatchedModelResource();
         while (resource != null) {
-            derivedPath.push(resource.getPath());
-            if (!resource.getPath().startsWith("/")) {
-                derivedPath.push("/");
+            String resourcePath = resource.getPath();
+            if (!resourcePath.equals("/") && !resourcePath.isBlank()) {
+                derivedPath.push(resourcePath);
+                if (!resourcePath.startsWith("/")) {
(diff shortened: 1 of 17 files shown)


microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -163,6 +153,47 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
...
+        //
+        // where a low-cardinality path would be, for example /greet/{name} rather than /greet/Joe, /greet/Dmitry, etc.
+
+        if (spanNameFullUrl) {
+            return requestContext.getMethod() + " " + requestContext.getUriInfo().getAbsolutePath().toString();
+        }
+        ExtendedUriInfo extendedUriInfo = (ExtendedUriInfo) requestContext.getUriInfo();
+
+        // Derive the original path (including path parameters) of the matched resource from the bottom up.
+        StringBuilder derivedPath = new StringBuilder();
```
ljnelson, line 169: Given all the `insert(0...)` work you're doing, maybe add things to a `Deque` instead and then flatten the `Deque` into a `StringBuilder` later. I'm not sure about `StringBuilder`'s innards but this might involve a heck of a lot of array copying and allocations. It also may not ultimately matter in the scheme of things but I'm thinking about cut-and-paste people down the line.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -163,6 +153,47 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
...
+
+    private String applicationPath() {
+        Application app = Contexts.context()
+                .flatMap(it -> it.get(Application.class))
+                .orElseThrow(() -> new IllegalStateException("Application missing from context"));
+
+        if (app == null) {
+            return "";
+        }
+        ApplicationPath applicationPath = getRealClass(app.getClass()).getAnnotation(ApplicationPath.class);
```
ljnelson, line 193: Since you've already used things like `ExtendedUriInfo` maybe there's a Jersey class that will give you whatever the `@ApplicationPath` annotation ends up setting indirectly?

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/WithSpanInterceptor.java
```diff
@@ -75,19 +78,42 @@ public Object interceptSpan(InvocationContext context) throws Exception {
...
+                    spanBuilder.setAttribute(attrName, (String) pValue);
+                } else if (paramType.isAssignableFrom(long.class) || paramType.isAssignableFrom(Long.class)) {
+                    spanBuilder.setAttribute(attrName, (long) pValue);
+                } else if (paramType.isAssignableFrom(double.class) || paramType.isAssignableFrom(Double.class)) {
+                    spanBuilder.setAttribute(attrName, (double) pValue);
+                } else if (paramType.isAssignableFrom(boolean.class) || paramType.isAssignableFrom(Boolean.class)) {
+                    spanBuilder.setAttribute(attrName, (boolean) pValue);
+                } else {
+                    spanBuilder.setAttribute(attrName, pValue.toString());
+                }
```
ljnelson, line 103: I'm pretty sure these are backwards. You want `String.class.isAssignableFrom(paramType)`. Or you want to do something like:
```
switch (paramType) {
case Class<?> x when Long.class.isAssignableFrom(paramType): // this may? work for primitives too? can't remember if autoboxing kicks in?
  spanBuilder.setAttribute(attrName, x.cast(pValue));
  break;
case Class<?> x when Double.class.isAssignableFrom(paramType):
  spanBuilder.setAttribute(attrName, x.cast(pValue));
  break;
case Class<?> x when Boolean.class.isAssignableFrom(paramType):
  spanBuilder.setAttribute(attrName, x.cast(pValue));
  break;
default:
  spanBuilder.setAttribute(attrName, pValue.toString());
  break;
}
```

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/WithSpanInterceptor.java
```diff
@@ -75,19 +78,42 @@ public Object interceptSpan(InvocationContext context) throws Exception {
...
+            }
+        }
+
+        // Start new Span
+        Span span = spanBuilder.startSpan();
 
         try (Scope scope = span.makeCurrent()) {
-            return context.proceed();
+            Object result = context.proceed();
+            span.end();
```
ljnelson, line 112: Double-checking you intended to hoist this out of the `finally` block.

ljnelson (changes requested):
Got some assignability stuff backwards, I think; have a peek

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Yes, `insert(0,...)` does some array copying. Changed.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: I didn't see a way to do that, after quite some looking.

If you have a specific suggestion I'd welcome it but I didn't find one. The security code Tomas pointed to uses the same technique to get the `Application`.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/WithSpanInterceptor.java
tjquinno: Very much so. There are two separate ways to end the span; one with an exception (see the `catch` code) and one without. Ending it in the `finally` block with no exception is _not_ correct.

ljnelson (dismissed):
LGTM (not an expert in this area)

tomas-langer (approved):
LGTM

PR #8225, opened 2024-01-12
Manage scopes correctly with baggage; allow baggage to be mutable to honor the Helidon `Span#baggage` semantics

### LINKED ISSUE #8187: 4.x Telemetry scopes are not closed properly when using baggage
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
In an app that uses OpenTelemetry tracing, adding baggage disrupts the handling of scopes.

Specifically, creating a span, activating it (which creates an autocloseable `Scope`), and then adding baggage to the span within the `try` block causes the `try-with-resources` close handling for the scope to act as a no-op.

## Steps to reproduce
You do not need to start Jaeger to receive the tracing; just ignore the runtime errors about being unable to send the tracing information.

1. Create a new SE app using `helidon init` and specify custom with tracing; take the other defaults.
2. Replace the generated `TracedService.TracedHandler#handle` method with this:
   ```
        @Override
        public void handle(ServerRequest req, ServerResponse res) {
            Tracer tracer = req.context().get(Tracer.class)
                    .orElseGet(Tracer::global);
            Span span = tracer.spanBuilder("custom-span")
                    .start();
            Span.current().ifPresentOrElse(s -> System.err.println("Current span before try-with-resources: " + s.context().spanId()),
                                           () -> System.err.println("No current span before try-with-resources"));

            try (Scope scope = span.activate()) {
                Span.current().ifPresentOrElse(s -> System.err.println("Current span in try-with-resources: " + s.context().spanId()),
                                               () -> System.err.println("No current span in try-with-resources"));
                span.addEvent("my nice log");

                res.send(message);
            } finally {
                span.end();
            }
            Span.current().ifPresentOrElse(s -> System.err.println("Current span after try-with-resources: " + s.context().spanId()),
                                           () -> System.err.println("No current span after try-with-resources"));
        }
   ```
   The changes do two things:
   1. Explicitly activate the custom span in the `try`, creating an auto-closable `Scope`.
   2. Display the current span just before, just inside, and just after the `try-with-resources` block.
3. Build and run the app and run `curl http://localhost:8080/traced`. You see the expected two different current spans at the expected times:
   ```
   Current span before try-with-resources: 268165d9440e2b8f
   Current span in try-with-resources: 1cfbe8a1c629213b
   Current span after try-with-resources: 268165d9440e2b8f
   ```
5. Add the following lines after the `span.addEvent` line in `TracedService.TracedHandler#handle`:
   ```
   span.baggage("myInfo", "myValue");
   Span.current().ifPresentOrElse(s -> System.err.println("Current span after setting baggage: " + s.context().spanId()),
                               () -> System.err.println("No current span after setting baggage"));
   ```
6. Stop, rebuild, and rerun the app and run `curl http://localhost:8080/traced` again.

The output reports the same span all three times:
```
Current span before try-with-resources: f3aa5b9b91198f31
Current span in try-with-resources: d93eaba9643f0c51
Current span after setting baggage: d93eaba9643f0c51
Current span after try-with-resources: d93eaba9643f0c51
```
The scope explicitly declared in the `try` that should be auto-closed is not.

This seems to occur because the Helidon `OpenTelemetrySpan#baggage(String, String)` method creates a new `Baggage` instance (as it should) and stores it in a new context _and makes the new context current_. This creates a new  active`Scope` but the scope is not exposed to the caller nor closed. 

Later, during the auto-close of the _explicit_ scope in the modified example code, the Helidon scope delegates the close operation to the OpenTelemetry scope which checks to see if the scope being closed is the current scope. _It is not_ in this scenario, so the OTel code ignores the attempt to close the scope. 

That is why the span ID after the try-with-resources block is unchanged once the baggage is added.

tjquinno:
### Description
Resolves #8187 

See the end of the issue description for an explanation of what was going wrong. Briefly, the Helidon classes wrapping baggage and spans were not managing the scopes for each correctly.

This PR addresses two basic problems:
* the scope management of baggage and spans
* the impedance mismatch between the Helidon neutral tracing API (which permits baggage key/value pairs to be set on a `Span`) and OpenTelemetry baggage which (1) is not tied to a span and (2) is (by default) immutable once placed in the Otel context (from where other code can access it). The default Otel baggage is fully formed (immutable) before being made "current" (added to the current context) and that model does not match the neutral Helidon tracing API.

The PR solves these problems as follows:
* Adds a mutable implementation of Otel's `Baggage` interface. Each span now instantiates this class and then the `baggage` method updates that field. Updates to the baggage are allowed before and after the baggage is added to the context.
* When a span is activated, now the code first makes the baggage "current" (pushes a new context that contains the baggage), then makes the span "current" (pushes a new context that contains the span). The now-current context knows about both the span and the baggage (as well as any other settings that were present in the original context before we made the baggage current). This is just the way Otel's context works: when you make a span or a baggage instance current, that creates a new context instance that inherits all the settings from the preceding current context plus whatever new key/value pair is being set.
* The `Span.activate()` method always has returned an instance of our implementation of `Scope`. That scope now stores internally both the Otel baggage scope and the Otel span scope. The `Scope#close()` method now closes the span scope, then the baggage scope. This resolves the original problem with scope management that resulted in filing the issue.

### Documentation
Bug fix; no doc update

### DIFF
--- a/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
+++ b/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.tracing.providers.opentelemetry;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.BiConsumer;
+
+import io.opentelemetry.api.baggage.Baggage;
+import io.opentelemetry.api.baggage.BaggageBuilder;
+import io.opentelemetry.api.baggage.BaggageEntry;
+import io.opentelemetry.api.baggage.BaggageEntryMetadata;
+
+class MutableOpenTelemetryBaggage implements Baggage {
+
+    private final Map<String, BaggageEntry> values = new LinkedHashMap<>();
+
+    MutableOpenTelemetryBaggage() {
+    }
+
+    private MutableOpenTelemetryBaggage(Builder builder) {
+        values.putAll(builder.values);
+    }
+
+    @Override
+    public int size() {
+        return values.size();
+    }
+
+    @Override
+    public void forEach(BiConsumer<? super String, ? super BaggageEntry> consumer) {
+        values.forEach(consumer);
+    }
+
+    @Override
+    public Map<String, BaggageEntry> asMap() {
+        return Collections.unmodifiableMap(values);
+    }
+
+    @Override
+    public String getEntryValue(String entryKey) {
+        BaggageEntry entry = values.get(entryKey);
+        return entry != null ? entry.getValue() : null;
+    }
+
+    @Override
+    public BaggageBuilder toBuilder() {
+        return new Builder(values);
+    }
+
+    void baggage(String key, String value) {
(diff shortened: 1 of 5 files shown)


tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    public BaggageBuilder toBuilder() {
+        return new Builder(values);
+    }
+
+    void baggage(String key, String value) {
+        values.put(key, new HBaggageEntry(value, new HBaggageEntryMetadata("")));
+    }
+
+    static class HBaggageEntry implements BaggageEntry {
```
tomas-langer, line 71: looks like a great candidate for a `record`

tomas-langer (approved):
Just  a small (optional) comment.

Issue #7206, opened 2023-07-16
Consider supporting Micronaut 4.0 in Helidon 4.x?


tjquinno:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
Micronaut has just released 4.0 (11 July 2023) which, among other things, has breaking changes regarding `javax` vs. `jakarta` in some areas including validation, as described in this excerpt from Slack:
> I'm having issue of imcompatibilty of the jakarta.validation and javax.validation.
The `jakarta.validation` (I'm refering to the constraints `import jakarta.validation.constraints.*` in to the model) doesn't work with micronaut `@Validated` in `CrudRepository` as it does not trigger the validation due the `helidon.integrations.micronaut` it only support` javax.validation` instead (tested it work).  
>  
> ```  
> version.lib.micronaut>3.4.3</version.lib.micronaut>, 
> version.lib.micronaut.data>3.3.0</version.lib.micronaut.data>  
> ``` 
> However with the use of `javax.validation.*` it create issue for manual validation, the `Validation.buildDefaultValidatorFactory() `doest work with` javax`, which trigger the error.  
>  
> ```  
> 1:57:56.883 [helidon-server-1] ERROR c.h.o.o.s.c.e.GlobalExceptionHandler - Unable to create a Configuration, because no Jakarta Bean Validation provider could be found. Add a provider like Hibernate Validator (RI) to your classpath.  
> javax.validation.NoProviderFoundException: Unable to create a Configuration, because no Jakarta Bean Validation provider could be found. Add a provider like Hibernate Validator (RI) to your classpath.
> ```

As Micronaut 4.0 contains breaking changes from Micronaut 3.x, we probably would not want to change the Micronaut dependency to 4.0 in a Helidon 3.x dot release.

But do we want to consider updating the Micronaut dependency to 4.0 in Helidon 4.x?

ASoftwareProgramer:
These are JAVA JDK **java.compiler** library, would this impact the integration ? I believe these namespace will not be change for the JAVA platform.

romain-grecourt:
> These are JAVA JDK **java.compiler** library, would this impact the integration ? I believe these namespace will not be change for the JAVA platform.

Correct, `jakarta.annotation:jakarta.annotation-api:jar:2.1.1` is not the replacement of `javax.annotation`.

ASoftwareProgramer:
[...]

PR #8227, opened 2024-01-12
3.x: Fix Micronaut integration in Helidon 3.x? #7206

jbescos:
### Description
Using Micronaut 4.x in Helidon 3.x: https://github.com/helidon-io/helidon/issues/7206

### Documentation

N/A

A couple of comments:

1. I had to remove `helidon/integrations/micronaut/cdi-processor/src/main/java/module-info.java` because there was one issue with Micronaut modules having same packages. I opened an issue: https://github.com/micronaut-projects/micronaut-core/issues/10299 . Anyway, as we use this module for annotation processing there is no real need to have a module-info.
2. helidon/integrations/micronaut/cdi/src/test/java/io/helidon/integrations/micronaut/cdi/TestBothBean.java is implementing one intrface with default methods and inherited annotations. I had to override the methods and add again the annotations. It seems a bug in Micronaut.


jbescos:
There is one issue with validation:
https://github.com/micronaut-projects/micronaut-core/issues/10370

jbescos:
Now Helidon examples exists in git@github.com:helidon-io/helidon-examples.git, and this makes a problem.

If you make a change in Helidon that requires modifications in examples, you are in a dead lock. You cannot submit in Helidon because examples fails, and you cannot submit in examples because the Helidon changes are not in place.

jbescos:
Examples fails because it also require changes:
https://github.com/helidon-io/helidon-examples/pull/32

barchetta:
As you see the examples checks are not required. Therefore, no deadlock.  When you are in this situation the procedure is:

1. Proceed with this PR. You will not be blocked by the examples check. You should comment as to why the examples fail to build (so we understand if you are introducing an incompatibility).
2. Create a PR against https://github.com/helidon-io/helidon-examples/tree/dev-3.x to adjust the example (maybe [you already have](https://github.com/helidon-io/helidon-examples/pull/34)). It will fail build check until the first PR is merged.
3. Once the first PR is approved you can merge it.
4. Then the second PR will pass the build check and you can have it reviewed and merge it.

Yes, this is inconvenient, but it's the cost of moving the examples out of the primary repository.

Issue #8217, opened 2024-01-10
Builder support for Map<String, String> as an option


Tomas-Kraus:
It would be nice to have support for `Config` options like
```
  properties:
    eclipselink.target-database: MySQL
    eclipselink.target-server: None
    eclipselink.logging.level: FINEST
    eclipselink.logging.level.weaver: OFF
    eclipselink.logging.parameters: true
    eclipselink.weaving: false

```
to be provided using
```
    @Option.Configured
    Map<String, String> properties();
```
in the ConfigBlueprint.

Will require to modify TypeHandlerMap to process and return whole nodes subtree (including child nodes).

tomas-langer:
There is a method `asMap()` on `Config`
This should be generated as `config.get("key").detach().asMap().orElseGet(Map::of)`

PR #8231, opened 2024-01-12
Support for Map<String, String> in configured builders.

tomas-langer:
### Description
Resovles #8217 

Adds explicit support for type `Map<String, String>`, and instead of doing the "complicated" get from config, it uses `config.detach().asMap()` to get the child values


Verdent (approved):

Tomas-Kraus (approved):
Looks good

PR #8233, opened 2024-01-12
4.x: Inject updates 8

### LINKED ISSUE #7957: 4.x: Helidon Inject - only API dependency to compile

tomas-langer:
### Description
Resolves #7957

 (this should be the last PR for Helidon Inject refactoring, follow up PRs will go against the `main` branch once again I hope)

- Removed types that are not related (will be re-introduced as part of Helidon Declarative)
- Documentation fixes and updates
- Fix and test for `@DrivenByName`
- Support for `QualifiedProvider` and tests
- Other smaller fixes (follow commits...)


inject/service/src/main/java/io/helidon/inject/service/QualifiedProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A provider to resolve qualified injection points of any type.
+ * <p>
+ * As compared to {@link io.helidon.inject.service.InjectionPointProvider}, this type is capable of resolving ANY injection
+ * point as long as it is annotated by the qualifier. The contract of the injection point depends on how the implementation
+ * service declares the type parameters of this interface. If you use any type other than {@link java.lang.Object}, that will
+ * be the only supported contract, otherwise any type is expected to be supported.
+ * <p>
+ * A good practise is to create an accompanying codegen extension that validates injection points at build time.
```
danielkec, line 34:
```suggestion
 * A good practice is to create an accompanying codegen extension that validates injection points at build time.
```

inject/service/src/main/java/io/helidon/inject/service/Injection.java
```diff
@@ -43,8 +43,8 @@ private Injection() {
     }
 
     /**
-     * Method, constructor, or method marked with this annotation is considered an injection point, and will be satisfied with
-     * services from the service registry.
+     * Method, constructor, or field marked with this annotation is considered as injectable, and its injection points
+     * will be satisfied with services from the service registry. An injection point is a filed, or a single parameter.
```
danielkec, line 47:
```suggestion
     * will be satisfied with services from the service registry. An injection point is a field, or a single parameter.
```

inject/maven-plugin/README.md
```diff
@@ -54,9 +56,9 @@ example usage:
...
...ng with the scope/cardinality for those instances. These instances are then delivered (as potentially injectable services) into other dependent services.  In this way Helidon Injection is simply acting as a broker and delivery mechanism between your Provider<> implementation(s) and the consumer that are using those service instances as injection points, etc. This is not meant to scare ore even dissuade you from using Provider<>, but merely to inform you that some of the deterministic behavior goes away under these circumstances using Provider instead of another Scope type like @Singleton. In many/most cases this is completely normal and acceptable. As a precaution, however, Helidon Injection chose to fail-fast at application creation time if your application is found to use jakarta.inject.Provider<T>. You will then need to provide a strategy/configuration in your pom.xml file to permit these types of usages. There are options to allow ALL providers (as shown in the above example), or the strategy can be dictated on a case-by-case basis. See the javadoc for [AbstractApplicationCreatorMojo](src/main/java/io/helidon/inject/maven/plugin/AbstractApplicationCreatorMojo.java) for details.
```
danielkec, line 61:
```suggestion
Also note that Helidon Injection strives to ensure your application stays as deterministic as possible (as shown by the <i>Injection__Application/i> generated class). But when the <i>jakarta.inject.Provider</i> type is used within your application then some of that deterministic behavior goes away. This is due to how Provider<>'s work since the implementation for the Provider (i.e., your application logic) "owns" the behavior for what actual concrete type that are created by it, along with the scope/cardinality for those instances. These instances are then delivered (as potentially injectable services) into other dependent services.  In this way Helidon Injection is simply acting as a broker and delivery mechanism between your Provider<> implementation(s) and the consumer that are using those service instances as injection points, etc. This is not meant to scare or even dissuade you from using Provider<>, but merely to inform you that some of the deterministic behavior goes away under these circumstances using Provider instead of another Scope type like @Singleton. In many/most cases this is completely normal and acceptable. As a precaution, however, Helidon Injection chose to fail-fast at application creation time if your application is found to use jakarta.inject.Provider<T>. You will then need to provide a strategy/configuration in your pom.xml file to permit these types of usages. There are options to allow ALL providers (as shown in the above example), or the strategy can be dictated on a case-by-case basis. See the javadoc for [AbstractApplicationCreatorMojo](src/main/java/io/helidon/inject/maven/plugin/AbstractApplicationCreatorMojo.java) for details.
```

danielkec (approved):
Just typos

Verdent (approved):

PR #8244, opened 2024-01-16
3.x Backport of scope/baggage fix

### LINKED ISSUE #8211: 3.x Backport of #8187 - Telemetry scopes are not closed properly when using baggage
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Backport of #8187.

tjquinno:
### Description
Resolves #8211 

Backport of PR #8225 - see the description there for full information.

### Documentation
Bug fix - no doc impact.

### DIFF
--- a/tracing/opentelemetry/pom.xml
+++ b/tracing/opentelemetry/pom.xml
@@ -113,7 +113,7 @@
                         </configuration>
                     </execution>
                     <execution>
-                        <!-- Run in a separate execution so other tests interfere. -->
+                        <!-- Run in a separate execution so other tests do not interfere. -->
                         <id>test-with-explicit-app</id>
                         <goals>
                             <goal>test</goal>
--- a/tracing/opentelemetry/pom.xml
+++ b/tracing/opentelemetry/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+    Copyright (c) 2022, 2024 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -66,6 +66,27 @@
         <dependency>
             <groupId>io.helidon.microprofile.server</groupId>
             <artifactId>helidon-microprofile-server</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-sdk-extension-autoconfigure</artifactId>
+            <version>${version.lib.opentelemetry-sdk-extension-autoconfigure}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-exporter-otlp</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common.testing</groupId>
+            <artifactId>helidon-common-testing-junit5</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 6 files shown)


tracing/opentelemetry/pom.xml
```diff
@@ -27,6 +27,10 @@
     <artifactId>helidon-tracing-opentelemetry</artifactId>
     <name>Helidon Tracing Open Telemetry</name>
 
+    <properties>
+        <version.lib.opentelemetry-sdk-extension-autoconfigure>1.22.0-alpha</version.lib.opentelemetry-sdk-extension-autoconfigure>
```
ljnelson, line 31: (I assume we're OK with this property at this level, and not in `dependencies/pom.xml` as would normally be the case.)

tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/MutableOpenTelemetryBaggage.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public void forEach(BiConsumer<? super String, ? super BaggageEntry> consumer) {
+        values.forEach(consumer);
+    }
+
+    @Override
+    public Map<String, BaggageEntry> asMap() {
+        return Collections.unmodifiableMap(values);
```
ljnelson, line 52: Depending on how often this method is called, you may want to store the unmodifiable view as a `final` instance field.

ljnelson (approved):
Couple of minor comments

tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/MutableOpenTelemetryBaggage.java
tjquinno: In the world of the neutral Helidon tracing API, the contents of the baggage data structure can change even once the span is started. So `values` can change at any time.

tracing/opentelemetry/pom.xml
tjquinno: My understanding is that for "tangential" dependencies we don't want to load up `dependencies/pom.xml` with them.

Now, maybe OTel is becoming central enough that we do want them there but ideally there should be direct dependencies on OTel only from one (maybe two?) artifacts.

Maybe that's just for integrations and this usage is less tangential than those. 

@tomas-langer @romain-grecourt Thoughts, folks?

tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/MutableOpenTelemetryBaggage.java
ljnelson: What I mean is:
```
// in your instance field declarations
private Map<String, BaggageEntry> values; // as before; don't have it in front of me
private Map<String, BaggageEntry> valuesView;

// In your class' constructor:
this.values = new HashMap<>(); // or whatever; all mutation methods target this
this.valuesView = Collections.unmodifiableMap(this.values); // return this from getters
```

tjquinno: Ah, yes. Got it. Will change. Was there a reason you didn't advocate for this:

```java
private final Map<String, BaggageEntry> values = new HashMap<>();
private final Map<String, BaggageEntry> valuesView = Collections.unmodifiableMap(values);
```

ljnelson: Not particularly; I tend to do all my instance field initializations in the constructor. Style and habit I suppose.

tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/MutableOpenTelemetryBaggage.java
```diff
@@ -0,0 +1,116 @@
...
+        return entry != null ? entry.getValue() : null;
+    }
+
+    @Override
+    public BaggageBuilder toBuilder() {
+        return new Builder(values);
+    }
+
+    void baggage(String key, String value) {
+        values.put(key, new HBaggageEntry(value, new HBaggageEntryMetadata("")));
```
ljnelson, line 68: Not thread safe. OK?

ljnelson (approved):

tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/MutableOpenTelemetryBaggage.java
tjquinno: Changed to save the view.

PR #8257, opened 2024-01-17
[4.x] Guard against NPE during early invocation of Span.current()

### LINKED ISSUE #8254: `TracerProviderHelper#currentSpan()` throws NPE in some situations
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
In some limited cases, the `TracerProviderHelper#currentSpan` method throws a NPE because `TRACER_PROVIDER` is null when `currentSpan` is invoked.

In at least two situations, customer code uses a custom `TracerProvider`. The static initializer in Helidon's `TracerProviderHelper` class uses service loading to locate and load such implementations, select one, and store a reference to it in the `TRACER_PROVIDER` static field. 

During its own initialization, that custom `TracerProvider` uses logging, and there is a custom logger in place which attempts to get the current span (if any) to add to log messages. And the attempt to get the current span ultimately invokes `TracerProviderHelper.currentSpan()` which depends on `TRACER_PROVIDER` being non-null.

In this scenario, though, `TRACER_PROVIDER` is still null because the static initializer has not yet finished. That triggers the NPE.

Although one could argue that the custom `TracerProvider` should not, even indirectly, invoke code that relies on `TRACER_PROVIDER` being non-null, this use case (and ones like it) would be difficult to untangle. Helidon could be of help to users in these situations by anticipating and dealing with this edge case.

tjquinno:
### Description
Resolves #8254 

See description for details and related 3.x PR #8256 for explanation.

### DIFF
--- a/tracing/tracing/src/main/java/io/helidon/tracing/TracerProviderHelper.java
+++ b/tracing/tracing/src/main/java/io/helidon/tracing/TracerProviderHelper.java
@@ -67,14 +67,23 @@ final class TracerProviderHelper {
     }
 
     static Tracer global() {
+        if (TRACER_PROVIDER == null) {
+            throw new IllegalStateException("Use before initialization has completed");
+        }
         return TRACER_PROVIDER.global();
     }
 
     static void global(Tracer tracer) {
+        if (TRACER_PROVIDER == null) {
+            throw new IllegalStateException("Use before initialization has completed");
+        }
         TRACER_PROVIDER.global(tracer);
     }
 
     static TracerBuilder<?> findTracerBuilder() {
+        if (TRACER_PROVIDER == null) {
+            throw new IllegalStateException("Use before initialization has completed");
+        }
         return TRACER_PROVIDER.createBuilder();
     }
 }
--- a/tracing/tracing/src/main/java/io/helidon/tracing/TracerProviderHelper.java
+++ b/tracing/tracing/src/main/java/io/helidon/tracing/TracerProviderHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2024 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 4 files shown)


tracing/tracing/src/main/java/io/helidon/tracing/TracerProviderHelper.java
```diff
@@ -58,7 +58,12 @@ private TracerProviderHelper() {
     }
 
     public static Optional<Span> currentSpan() {
-        return TRACER_PROVIDER.currentSpan();
+        /*
+        If a custom TracerProvider implementation indirectly tries to access the current span (for example, by triggering custom
+        logging that adds the current span to each message), then this method can be invoked before the static initializer has
+        completed and, therefore, before TRACER_PROVIDER is assigned.
+         */
+        return (TRACER_PROVIDER == null) ? Optional.empty() : TRACER_PROVIDER.currentSpan();
```
tomas-langer, line 66: I think that this is creating a bug. We should never reach this method within initalization.
If we do, then we should change the way we initialize, rather than checking if a constant field is null

tomas-langer (changes requested):

tracing/tracing/src/main/java/io/helidon/tracing/TracerProviderHelper.java
tjquinno: As I mentioned in our DMs, we could add a static boolean that indicates if the static initializer is in progress or not. Then `currentSpan` could check that boolean and, if true, return `Optional.empty()` right away rather than trying to use `TRACER_PROVIDER`. That might look slightly less odd than checking a supposedly `final` field for null, but it amounts to the same thing: checking whether the static initializer is in-progress or not. 

As we also discussed, I will also add code to throw exceptions from other methods that rely on `TRACER_PROVIDER` but are `void` or have non-`Optional` return types.

PR #8269, opened 2024-01-19
[DRAFT] Eclipse store integration

hrstoyanov:
### Description

Add [Eclipse Store](https://https://eclipsestore.io/) support to Helidon.

[Eclipse Store](https://eclipsestore.io/) is the successor of [Microstream](https://microstream.one), both of which are integrated into Helidon as extensions now, enabling smooth transition.

### Documentation

The official [Eclipse Store documentation](https://docs.eclipsestore.io) can be found here.

If no doc impact: None


hrstoyanov:
@fh-ms @zdenek-jonas  - this is the DRFAT merge request.  ^^^^

If Eclipse Store is released in about 2 weeks, in which version of Helidon 4.x will you be able to merge this extension in?

fh-ms:
[EclipseStore 1.1.0](https://github.com/eclipse-store/store/releases/tag/1.1.0) is available.

@hg-ms and I did a code review. Looks good for us.

hrstoyanov:
Thank you @fh-ms and @hg-ms for reviewing the code! The only additional commit I did was  to change the versions to 1.1.0 
 from 1.1.0-SNAPSHOT, as per the Eclipse Store/Serializer official releases.
 
 @tjquinno @ljnelson @barchetta 
 This pull request is now ready for your consideration - I removed [DRAFT] from the title and changed the description a bit.
 
 Any clue in which Helidon version it could be included, assuming we missed the 4.0.4 train?

hrstoyanov:
@fh-ms 
Please review the latest commit:
- deprecate MicroStream, invite users to switch to Eclipse Store.
- attempt to fix as many compiler warnings as possible.

fh-ms (approved):

hrstoyanov:
@fh-ms @zdenek-jonas. Please review the latest Eclipse Store 1.2.0 upgrade to this pull request. 

A couple of things:

- I added ```@Aot``` to the metadata in the module definitions (same as ```@Aot(true)```) to signify that in a GraalVM compile of an Helidon app, it is safe to include Exclipe Store 1.2.0. If EclipseStore 1.2.0 is not aot-compile safe,  we need to set  it back to ```@Aot(false)```  in the ES integrations module definitions.

- I wonder what is the value of Eclipse Store Cache integration - javax.cache API has not even been moved to jakarta namespace, neither has anyone bothered to modularize it, doubt anyone uses it. There are compiler warning because of automatic modules, etc.

hg-ms:
> @fh-ms @zdenek-jonas. Please review the latest Eclipse Store 1.2.0 upgrade to this pull request.

I did a review of the EclipseStore specific parts, everything is ok from our side.

Regarding Aot support: it should work but requires additional user actions because all user defined classes that are going to be persisted must be accessible by reflections. I’d suggest to add a description to @Aot documenting this.

As we don’t have a replacement for the javax.cache API we need to live with that.

romain-grecourt:
Did you run `etc/scripts/checkstyle.sh` and `etc/scripts/copyright.sh` ?

romain-grecourt:
static imports go last.

romain-grecourt:
> Do you want me to squash commits?

Sure. You can squash and rebase against the latest main.
That might clear up the copyright errors you have with the `.helidon-oidc-secret` files.

romain-grecourt:
You have 214 files changed, it does not look right. Let me fix it.

integrations/eclipsestore/metrics/src/test/java/io/helidon/integrations/eclipsestore/metrics/EclipseStoreHealthMetricsTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.mockito.Mockito;
+
+import java.util.Date;
+import java.util.Set;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+// TODO disabled
+@Disabled
```
romain-grecourt, line 36: Why is this test disabled ?

romain-grecourt (commented):
Can you add an example ?

security/providers/idcs-mapper/.helidon-oidc-secret
```diff
@@ -0,0 +1 @@
+108d6fd4-de62-4343-aa24-ad59db6f9846
```
romain-grecourt, line 1: I don't know what you are doing but you keep adding these files. You should update the top level .gitignore to ignore them.

examples/integrations/eclipsestore/greetings-mp/src/main/java/io/helidon/examples/integrations/eclipsestore/greetings/mp/GreetingProvider.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import io.helidon.integrations.eclipsestore.cdi.EclipseStoreStorage;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.inject.Inject;
+
```
romain-grecourt, line 28: `jakarta` and `org.eclipse` should not be separated.

There are 3 import groups: `java.*`, `io.helidon.*` and everything else

romain-grecourt: Groups are separated by a new line.

security/providers/idcs-mapper/.helidon-oidc-secret
romain-grecourt: That's the last time I do it. You need to learn how to do that yourself.

tomas-langer:
Thanks for this contribution. We are now making some decisions regarding the timing. I want to spend some time on this to use our new Builder APIs and do a bit of Helidon alignment, will let you know asap

fh-ms:
An update to 1.3.1 makes sense due to an [adjustment for Java 22](https://github.com/eclipse-serializer/serializer/issues/69).

We still have to test the CDI stuff with Helidon properly, and maybe we find a way to add the module-infos there, so we should wait until we remove the modules.

tomas-langer:
Regarding CDI integration - I think the Helidon MP integration is not "just" CDI, but it should also support MP Config as a source of configuration data. If that is the case, it may still be useful. If not, then I agree it can be removed and we can use the built-in CDI integration.

PR #6956, opened 2023-06-06
Remove metrics app registry clear-out code from metrics extension; add doc

tjquinno:
Resolves #4656 

Prior to this PR, the metrics CDI extension cleared the application metrics registry late in the CDI life cycle. Metrics registered earlier during start-up by developer code (such as from a developer-written extension) would disappear.

The clear-out code is needed to allow the TCK tests to work correctly. 

This PR does two main things:

1. Move the clear-out code from the normal metrics extension into a TCK-only extension.
2. Add documentation explaining that user code _should not_ register metrics before the `RuntimeStart` CDI event.

Helidon will not _prevent_ users from registering metrics prematurely. But the metrics system might not behave as configured if they do. The doc additions explain this.

PR #8292, opened 2024-01-24
Remove clear-out of registries from extension; do it in TCK-only shutdown observer; update doc

### LINKED ISSUE #8287: [4.x] MP Metrics clears app registry late in CDI start-up life cycle
Backport of #4656 for Helidon 4.x

tjquinno:
### Description
Resolves #8287 

(mostly a forward-port of PR #6956. See that PR for more info)

Early in Helidon 3 the metrics CDI extension cleared out the application `MetricRegistry` late in its initialization. This was primarily to allow TCK tests, which start up the container repeatedly, to work. (Some of the TCK tests reuse the same metric name for different metric types; if the registry was not cleared out between TCK tests the second attempt to register the same-named metric would fail.)

Later in Helidon 3 (but after the code base for 4 had been started), we changed this so a TCK-only CDI extension did the clean-out and we removed that code from the 3.x metrics CDI extension.

This PR does basically the same thing but in the Helidon 4 codebase. The PR also adds some doc comments (copied from the 3.x PR) warning users against registering metrics from their own CDI extensions until after the metrics extension has run and initialized metrics using the correct configuration. 

### Documentation
PR includes doc changes.

### DIFF
--- a/docs/src/main/asciidoc/mp/metrics/metrics.adoc
+++ b/docs/src/main/asciidoc/mp/metrics/metrics.adoc
@@ -109,6 +109,12 @@ Helidon automatically looks up the metric referenced from any injection site and
 
 include::{rootdir}/includes/metrics/metrics-shared.adoc[tag=metric-registry-api]
 
+=== Working with Metrics in CDI Extensions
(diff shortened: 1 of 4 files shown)


spericas (approved):
LGTM

PR #8320, opened 2024-01-31
4.x Add baggage API and allow access to baggage via `SpanContext` (as well as `Span`)

### LINKED ISSUE #8319: 4.x Provide a way to retrieve telemetry baggage from a `SpanContext` without having to start a `Span`
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
The Helidon neutral tracing API provides access to baggage associated with a `Span`. 

Sometimes, though, code has a `SpanContext` available (such as the one Helidon prepares for an incoming HTTP request) and for that code's purposes needs only the baggage, not an actual span.

The workaround is this:
```
SpanContext spanContext = // from somewhere
Span.Builder<?> b = tracer.spanBuilder("example");
Span span = b.parent(spanContext).start();
Optional<String> baggageValue = span.baggage("keyName");
```

This issue captures the request to add a way to retrieve baggage from a `SpanContext` without having to create a `Span`.

tjquinno:
### Description
Resolves #8319 

Users have requested a way to retrieve baggage from a `SpanContext` directly rather than having to use the `SpanContext` to create (and start and end) a `Span`. Currently only the `Span` type allows access to baggage.

New types and implementations:

1. The new `Baggage` interface provides read-only access to baggage with `get`, `keys`, and`containsKey` operations.
2. The new `WritableBaggage` interface extends `Baggage` and adds a `set` method.
3. The OpenTracing and OpenTelemetry providers (and the implicit no-op implementation in `tracing/tracing`) add implementations of these interfaces and add tests.

Changes to existing code:

1. The `SpanContext` interface adds the `Baggage baggage()` method.
2. The `Span` interface adds the `WritableBaggage baggage()` method.
3. The existing `Span` interface methods for directly manipulating baggage are marked as deprecated to encourage users to use the more powerful `Baggage` interfaces.

`Baggage` is read-only, typically based on the baggage data from an underlying `SpanContext` because we don't necessarily know how a `SpanContext` is derived--for example, it can be from incoming request headers--and so we should not allow callers to update baggage derived from span contexts.

In the Helidon neutral tracing API, though, the `Span` allows updates to the baggage we associate with the span, hence the `WritableBaggage` type.

## Very brief guided tour of the changes 
For other reasons we already had in the OpenTelemetry provider the `MutableOpenTelemetryBaggage` class which implements an Otel baggage-related interface. This PR expands that class to also implement the new Helidon `WritableBaggage` interface. 

The new `OpenTracingBaggage` class implements the read-only `Baggage` interface and it has a small inner subclass which implements `WritableBaggage`. We use an inner class here because in OpenTracing when span-related baggage is updated we need to propagate the update to the OpenTracing span itself as well as keeping track of it ourselves.

Both providers add tests to exercise their implementations of the new baggage API.

The OpenTracing provider component `pom.xml` now includes several Otel dependencies in test scope. This allows the tests to rely on the actual Otel extractor and propagator, for example, rather than no-ops that would cause tests to fail.

### Documentation
The PR includes a brief addition to the tracing doc page discussing baggage with links to the JavaDoc.


tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
```diff
@@ -65,9 +89,33 @@ public BaggageBuilder toBuilder() {
...
     }
 
+    @Override
+    public Optional<String> get(String key) {
+        return Optional.ofNullable(values.get(key).getValue());
+    }
+
+    @Override
+    public Set<String> keys() {
+        return Collections.unmodifiableSet(values.keySet());
```
ljnelson, line 104: Depending on the frequency of the call, consider storing an unmodifiable `Set` as an instance variable and returning it directly.

tracing/providers/opentracing/src/main/java/io/helidon/tracing/providers/opentracing/OpenTracingBaggage.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public Optional<String> get(String key) {
+        Objects.requireNonNull(key, "baggage key cannot be null");
+        return Optional.ofNullable(baggage.get(key));
+    }
+
+    @Override
+    public Set<String> keys() {
+        return Collections.unmodifiableSet(baggage.keySet());
```
ljnelson, line 62: Same as earlier comment

tracing/providers/opentracing/src/main/java/io/helidon/tracing/providers/opentracing/OpenTracingTracer.java
```diff
@@ -106,6 +106,9 @@ public <T> T unwrap(Class<T> tracerClass) {
         if (tracerClass.isAssignableFrom(delegate.getClass())) {
             return tracerClass.cast(delegate);
         }
+        if (tracerClass.isInstance(this)) {
```
ljnelson, line 109: Minor nit: depending on how "deep" you want/need the unwrapping to go you could do this whole general thing in a `while` loop

tracing/tracing/src/main/java/io/helidon/tracing/NoOpTracer.java
```diff
@@ -27,6 +28,28 @@ class NoOpTracer implements Tracer {
...
+            return Set.of();
+        }
+
+        @Override
+        public boolean containsKey(String key) {
+            return false;
+        }
+
+        @Override
+        public WritableBaggage set(String key, String value) {
```
ljnelson, line 48: (Naïve question: must it be a `WriteableBaggage` implementation if it's just going to ignore this call anyway?)

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
tjquinno: The `values` map can grow at any time, so the key set will not be stable. A `Set` instance field would need to be updated whenever baggage was added or a dirty bit set and the `Set` reconstituted lazily if dirty upon access. My gut feel is that `keys` will be infrequently invoked, at least seldom enough for optimizing this to be premature given the map's changeability.

tracing/providers/opentracing/src/main/java/io/helidon/tracing/providers/opentracing/OpenTracingBaggage.java
tjquinno: Ditto.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
ljnelson: Fair. (Do note that `Collections#unmodifiable...` methods account for changing values "underneath"; that's why they're called `unmodifiable` and not, say, `immutable`.)

tracing/tracing/src/main/java/io/helidon/tracing/NoOpTracer.java
tjquinno: In this `NoOpTracer` the inner classes `Span` and `SpanContext` have a `baggage()` method declared by the interfaces they implement to return a `WriteableBaggage`. Each of the `baggage()` methods returns the`EMPTY_BAGGAGE` constant so it needs to implement `WriteableBaggage` even though, as you point out, it doesn't really write anything.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/MutableOpenTelemetryBaggage.java
tjquinno: Exactly. We want the caller to see the accurate key set without being able to mess with its contents.

PR #8342, opened 2024-02-07
Fix  for #8344 - Incorrect SQL Type for null Parameters in JDBC Parameter Binding

bjcoombs:
### Summary:
This PR addresses an issue in the JDBC parameter binding logic where `null` values were incorrectly being set with `Types.VARCHAR` SQL type. This behavior lead to SQL exceptions when inserting `null` into non-`VARCHAR` columns in a database. The proposed change ensures that `null` values are set with `Types.NULL`, allowing the JDBC driver to correctly interpret the intended type based on the column definition.

### Problem Description:
When attempting to insert `null` values into a database using the Helidon JDBC client, the fixed SQL type of `Types.VARCHAR` for `null` parameters leads to type mismatch errors for columns of types other than `VARCHAR`. For example, inserting `null` into a `NUMERIC` column results in a `PSQLException` with a message indicating a type mismatch.

### Proposed Solution:
The fix involves changing the SQL type for `null` parameters from `Types.VARCHAR` to `Types.NULL` in the `setParameter` method of the JDBC parameter handling logic. This change allows the JDBC driver to determine the correct type for the `null` value based on the context of the column it is being inserted into, thus preventing type mismatch errors.

### DIFF
--- a/dbclient/jdbc/src/main/java/io/helidon/dbclient/jdbc/JdbcStatement.java
+++ b/dbclient/jdbc/src/main/java/io/helidon/dbclient/jdbc/JdbcStatement.java
@@ -15,14 +15,6 @@
  */
 package io.helidon.dbclient.jdbc;
 
-import io.helidon.dbclient.DbClientException;
-import io.helidon.dbclient.DbClientServiceContext;
-import io.helidon.dbclient.DbIndexedStatementParameters;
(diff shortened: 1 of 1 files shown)


bjcoombs:
fix for #8344

Tomas-Kraus (approved):
Looks good

PR #8357, opened 2024-02-09
4.x: Return 500 when 204 with entity is sent from routing.

tomas-langer:
### Description
Fixed a problem when user returned `204` with an entity in JAX-RS. 
With the default connector, the next (!) request would fail with Status code `-1` (caused by the client ignoring the entity according to the specification, but then the connection was broken, and the next response started reading the previous entity).

The client could handle this more gracefully (i.e. close the connection, report the problem), but it is the default one in Java...

Updated solution:
- if an attempt is done to write entity with `204` status, the status is changed to `500`, and everything else carries out as usual (as entity is allowed with 500). The server logs an error explaining the problem.

### DIFF
--- a/examples/metrics/http-status-count-se/src/test/java/io/helidon/examples/se/httpstatuscount/StatusService.java
+++ b/examples/metrics/http-status-count-se/src/test/java/io/helidon/examples/se/httpstatuscount/StatusService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package io.helidon.examples.se.httpstatuscount;
 
 import io.helidon.http.Status;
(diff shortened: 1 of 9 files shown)


romain-grecourt (approved):

danielkec (approved):

examples/metrics/http-status-count-se/src/test/java/io/helidon/examples/se/httpstatuscount/StatusService.java
```diff
@@ -43,6 +44,12 @@ private void respondWithRequestedStatus(ServerRequest request, ServerResponse re
             status = Status.INTERNAL_SERVER_ERROR_500.code();
             msg = "Unsuccessful conversion";
         }
-        response.status(status).send(msg);
+        Status httpStatus = Status.create(status);
+        response.status(status);
+        if (httpStatus != Status.NO_CONTENT_204) {
```
jbescos, line 49: Should it be here 304 and 205 too?

tomas-langer: Yes, fixed.

tomas-langer:
I have validated the implementation in `HttpUrlConnection` - it expects that there is NOT an entity (it does some side check with `Content-Length` header, which is not used in our case, we just use chunked encoding). 
So we must resolve this in Helidon.

danielkec (approved):

romain-grecourt (approved):

jbescos (approved):

jbescos:
> I have validated the implementation in `HttpUrlConnection` - it expects that there is NOT an entity (it does some side check with `Content-Length` header, which is not used in our case, we just use chunked encoding). So we must resolve this in Helidon.

Find here my findings about this: https://github.com/jakartaee/rest/issues/1199#issuecomment-1849416365

jbescos (commented):
I am reviewing this further and I am having questions. Lets try to clarify that.

Is Http1ServerResponse#nonEntityBytes always invoked?. I think it is. When there is 204, 205 or 304 with an entity we should log a warning and don't write the entity rather than failing with 500 error in my opinion.

Also the helidon client will not be able to handle chunked 204, 205 or 304 with entity. This is the comment wrote here: https://github.com/helidon-io/helidon/pull/8357#issuecomment-1940515055

PR #8360, opened 2024-02-09
Suppress Helidon handling of OpenTelemetry annotations if OTel Java agent is present

### LINKED ISSUE #8341: Helidon produces additional, unwanted spans on @WithSpan while OTEL agent is active
## Environment Details
* Helidon Version: 4.0.5
* Helidon MP
* JDK version: Java HotSpot(TM) 64-Bit Server VM (build 21.0.1+12-LTS-29, mixed mode, sharing)
* OS: Ubuntu 23.10

----------

## Problem Description
We use OTEL Java agent 1.32, and we are noticing that with Helidon OpenTelemetry support, we are getting an additional span for code annotated with `@WithSpan` . As the span information added by OTEL agent holds more meaningful information, we don't want to have the span introduced by Helidon. Note that besides these declarative spans, we also use programmatic spans that rely on injection of the `Tracer`. So, we need both Helidon OpenTelemetry support as well as OTEL agent.

The below outtake shows an example of what this looks like in Jaeger.

image

tjquinno:
### Description
Resolves #8341 

OTel provides a Java agent which automatically traces JAX-RS endpoints and methods with the OTel `@WithSpan` annotation. 

Helidon's MP Telemetry support also automatically traces JAX-RS endpoints (via a filter) and `@WithSpan` methods (via an interceptor). 

If both are present the service creates double spans. Yet users still want to keep `helidon-microprofile-telemetry` so they can `@Inject` tracers.

This PR does several things:
1. If the OTel agent is present, the telemetry CDI portable extension bypasses binding our interceptor to `@WithSpan` methods.
2. If the OTel agent is present, the telemetry JAX-RS filter bypasses Helidon's automatic JAX-RS endpoint handling.

(Testing this automatically would be quite involved. Because of the unique behavior and requirements of a Java agent in general and the OTel agent specifically, automating a test would require building a temporary JAR containing our in-memory span exporter plus other Otel-specific code to tell the agent to use that in-memory exporter and then adding system property assignments to the surefire test execution to control the agent. It could be done but is not part of this PR. I did manual tests to confirm the original behavior and to verify the fix.)

### Documentation
Bug fix - no doc change.

### DIFF
--- a/microprofile/telemetry/pom.xml
+++ b/microprofile/telemetry/pom.xml
@@ -140,6 +140,8 @@
                 <configuration>
                     <forkCount>1</forkCount>
                     <reuseForks>false</reuseForks>
+                    <!-- The default for the following is 5000 ms. Reducing it significantly speeds up tests. -->
+                    <argLine>-Dotel.bsp.schedule.delay=100</argLine>
                 </configuration>
             </plugin>
         </plugins>
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -99,6 +99,10 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
     @Override
     public void filter(ContainerRequestContext requestContext) {
 
+        if (Boolean.getBoolean(TelemetryCdiExtension.OTEL_AGENT_PRESENT)) {
+            return;
+        }
+
         if (LOGGER.isLoggable(System.Logger.Level.TRACE)) {
             LOGGER.log(System.Logger.Level.TRACE, "Starting Span in a Container Request");
         }
@@ -131,6 +135,10 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
 
     @Override
     public void filter(final ContainerRequestContext request, final ContainerResponseContext response) {
+
+        if (Boolean.getBoolean(TelemetryCdiExtension.OTEL_AGENT_PRESENT)) {
+            return;
+        }
 
         if (LOGGER.isLoggable(System.Logger.Level.TRACE)) {
             LOGGER.log(System.Logger.Level.TRACE, "Closing Span in a Container Request");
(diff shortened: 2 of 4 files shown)


microprofile/telemetry/pom.xml
```diff
@@ -140,6 +140,8 @@
                 <configuration>
                     <forkCount>1</forkCount>
                     <reuseForks>false</reuseForks>
+                    <!-- The default for the following is 5000 ms. Reducing it significantly speeds up tests. -->
+                    <argLine>-Dotel.bsp.schedule.delay=100</argLine>
```
spericas, line 144: Does this look sufficient for those days where the pipelines are dog slow?

tjquinno: I think this is OK.

This lower setting causes OTel to flush information to our test code much more often than the default. 

The `TestSpanExporter` class internally uses our `MatcherWithRetry` with default timeout of 60 seconds (which we could configure longer if in a pipeline if that became a problem).

The main reason I changed this setting is that the default means flushes occur every 5 seconds, per test, which really prolongs the running time of the tests. This speeds up the running time of those tests quite a bit. I don't think timing out will be a problem, but you're right--with our pipelines, one never knows for sure.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -92,13 +93,18 @@ public Iterable<String> keys(ContainerRequestContext containerRequestContext) {
     HelidonTelemetryContainerFilter(Tracer tracer, OpenTelemetry openTelemetry, org.eclipse.microprofile.config.Config mpConfig) {
         this.tracer = tracer;
         this.openTelemetry = openTelemetry;
+        isAgentPresent = Boolean.getBoolean(TelemetryCdiExtension.OTEL_AGENT_PRESENT);
```
tomas-langer, line 96: Why didn't you use `io.helidon.tracing.providers.opentelemetry.HelidonOpenTelemetry.AgentDetector`?

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryCdiExtension.java
```diff
@@ -28,8 +28,15 @@
  */
 public class TelemetryCdiExtension implements Extension {
 
+    /**
+     * Property name indicating the presence of the OpenTelemetry Java agent.
+     */
+    static final String OTEL_AGENT_PRESENT = "otel.agent.present";
```
tomas-langer, line 34: See above - I think this is already built in Helidon (`io.helidon.tracing.providers.opentelemetry.HelidonOpenTelemetry.AgentDetector`)

PR #8403, opened 2024-02-21
4.x: Fixes for problems discovered in service registry work

tomas-langer:
- handling of Optional<char[]> in builders
- handling of generic types built by builders in another builder
- missing primitive type mappers in common mappers
- support for mappable values in Parameters
- small build.sh fix to remove deleted scripts (replaced by another job)

### DIFF
--- a/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
@@ -21,7 +21,6 @@ import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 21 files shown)


builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -450,5 +451,39 @@ public interface OptionDecorator<B, T> {
          */
         String[] value();
     }
+
+    /**
+     * Utility class used from generated code for equals and hash code of specific cases of optional.
+     * This exists because of Optional of char arrays, where equals does not work as we need.
+     */
+    public static class OptionalUtil {
```
romain-grecourt, line 459: This nested utility class is polluting the Prototype API.
Instead shouldn't this be a general purpose utility class to support generated code for builders ? I.e. what if we need more utility in the future that aren't optional related.

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -450,5 +451,39 @@ public interface OptionDecorator<B, T> {
...
+        /**
+         * Equals that uses {@link java.util.Arrays#equals(char[], char[])} in case both optionals have a value.
+         *
+         * @param first  first optional
+         * @param second second optional
+         * @return whether the optionals are equals
+         */
+        public static boolean charArrayEquals(Optional<char[]> first, Optional<char[]> second) {
```
romain-grecourt, line 467: Isn't this violating the rule of not using optionals as fields or method parameters ?

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
tomas-langer: Yes, I will create a dedicated `GeneratedBuilder` or similar type.

romain-grecourt (approved):

PR #8410, opened 2024-02-22
Generalize the OCI metrics integration, using qualified injection of `Config`

tjquinno:
### Description
Configuration for the OCI metrics integration might come from different sources. This PR slightly generalizes how the `Config` is injected into the bean which starts up the integration.

The "start-up" bean now injects a `@Qualified` `Config` instance and provides a default producer for that qualified `Config` which is the normally-injected `Config` itself. There is no change in the behavior of this library when used as in the past.

The change is if another library includes a higher-priority bean which also produces the qualified `Config` bean, in which case this library uses _that_ config instance instead.

### Documentation
No functional change, and we do not intend for users to take advantage of this generalization themselves. 

No doc impact.

### DIFF
--- a/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetrics.java
+++ b/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetrics.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.integrations.oci.metrics.cdi;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.inject.Qualifier;
+
+/**
+ * Marks a {@link io.helidon.config.Config} object as containing settings for OCI metrics integration.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.PARAMETER})
+@Qualifier
+public @interface OciMetrics {
+}
--- a/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
+++ b/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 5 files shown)


integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetrics.java
```diff
@@ -0,0 +1,32 @@
...
+
+import javax.inject.Qualifier;
+
+/**
+ * Marks a {@link io.helidon.config.Config} object as containing settings for OCI metrics integration.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD, ElementType.PARAMETER})
+@Qualifier
+public @interface OciMetrics {
```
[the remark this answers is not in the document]
tjquinno: Yes, because other libraries which might provide a higher-priority producer will also need to specify this qualifier.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.helidon.config.Config;
+
+@Priority(0)
+class OciMetricsConfigProducer {
+
+    @Produces
+    @OciMetrics
+    public Config config(Config config) {
+        return config;
```
ljnelson, line 29: Double-checking that you want to return `@Default Config` as (also) `@OciMetrics Config`. I guess this is the "backstop" implementation, so that users supplying their own `@OciMetrics Config` will have their stuff picked up first?

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package io.helidon.integrations.oci.metrics.cdi;
+
+import javax.annotation.Priority;
+import javax.enterprise.inject.Produces;
+
+import io.helidon.config.Config;
+
+@Priority(0)
```
ljnelson, line 23: In _general_ if you can pick a constant from (bizarrely, counterintuitively, but hey, it's convention) [the `Interceptor.Priority` (!) constants](https://jakarta.ee/specifications/interceptors/2.1/apidocs/jakarta/interceptor/interceptor.priority#field.summary) it's best.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
```diff
@@ -0,0 +1,31 @@
...
+class OciMetricsConfigProducer {
+
+    @Produces
+    @OciMetrics
+    public Config config(Config config) {
```
ljnelson, line 28: This could be `private static` if you wanted; then no instance of `OciMetricsConfigProducer` will have to be created.

ljnelson (dismissed):
LGTM with some comments

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
tjquinno: Yes. I will add a comment explaining that this is intended.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
tjquinno: Will do.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetrics.java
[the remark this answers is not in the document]
ljnelson: Ah yes. Carry on.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsConfigProducer.java
tjquinno: I do not seem to be able to keep in my head the rules/conventions for visibility to CDI vs. Java declarations. I'll make this change.

ljnelson: To be fair: it's just another method. If you wanted to surface it, leave it as `public` for sure. But often producer methods are to be used solely by the container. If that's the case as I strongly guess it is, then I personally prefer to limit visibility wherever I can.

tjquinno:
Closing - this PR is obsolete in favor of superseding PR #8419

PR #8419, opened 2024-02-25
Generalize OCI metrics support; add tests for the generalization

tjquinno:
### Description
This PR slightly generalizes how the OCI metrics integration is set up.

Most of the work continues to be done in `OciMetricsBean` but this class can now be extended in two ways:
* customize what config prefix is used to fetch OCI metrics integration settings from the root node
* customize how the `OciMetricsSupport.Builder` is prepared
* allow other work around starting the `OciMetricsSupport` instance

The PR includes an example customization of the bean as well as tests using the customization to make sure the customized bean (designated as a CDI `@Alternate` and with a more urgent `@Priority` than the normal bean implementation) is used.

As noted in the original bean class comments, it's essential that the observer method in the bean run _after_ an observer method in the Helidon MP metrics extension. The PR adds a test using reflection to verify this requirement.

### Documentation
No impact.

This generalization is intended for use by library authors familiar with Helidon's workings as opposed to application developers. As a result, beyond the JavaDoc in the PR there is no need to update the user documentation.

### DIFF
--- a/integrations/oci/metrics/cdi/pom.xml
+++ b/integrations/oci/metrics/cdi/pom.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
-    Copyright (c) 2021, 2023 Oracle and/or its affiliates.
+    Copyright (c) 2021, 2024 Oracle and/or its affiliates.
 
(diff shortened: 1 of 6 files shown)


klustria (approved):
After going through with Tim, the changes LGTM

Issue #7060, opened 2023-06-23
Nima gRPC Client


romain-grecourt:
Implement a Nima based gRPC client.

PR #8423, opened 2024-02-26
gRPC Client Implementation

spericas:
### Description

New gRPC Client API and implementation for SE. Issue #7060.


all/pom.xml
```diff
@@ -974,6 +974,10 @@
             <groupId>io.helidon.webclient</groupId>
             <artifactId>helidon-webclient-websocket</artifactId>
         </dependency>
+        <dependency>
```
tomas-langer, line 977: I see different modules here and in bom pom. These should be aligned.

grpc/core/src/main/java/module-info.java
```diff
@@ -0,0 +1,36 @@
...
+    requires transitive io.helidon.common.context;
+    requires transitive io.helidon.common.config;
+    requires transitive io.grpc;
+    requires transitive io.grpc.stub;
+    requires transitive com.google.protobuf;
+    requires transitive io.grpc.protobuf;
+    requires transitive io.grpc.protobuf.lite;
+
+    requires java.logging;
+    requires jakarta.inject;
```
tomas-langer, line 31: This should not be used

grpc/core/src/main/java/module-info.java
```diff
@@ -0,0 +1,36 @@
...
+
+    requires transitive io.helidon.common.context;
+    requires transitive io.helidon.common.config;
+    requires transitive io.grpc;
+    requires transitive io.grpc.stub;
+    requires transitive com.google.protobuf;
+    requires transitive io.grpc.protobuf;
+    requires transitive io.grpc.protobuf.lite;
+
+    requires java.logging;
```
tomas-langer, line 30: This should not be used (we only use `System.Logger` in Helidon 4)

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcBaseClientCall.java
```diff
@@ -0,0 +1,209 @@
...
+
+    GrpcBaseClientCall(GrpcClientImpl grpcClient, MethodDescriptor<ReqT, ResT> methodDescriptor, CallOptions callOptions) {
+        this.grpcClient = grpcClient;
+        this.methodDescriptor = methodDescriptor;
+        this.callOptions = callOptions;
+        this.requestMarshaller = methodDescriptor.getRequestMarshaller();
+        this.responseMarshaller = methodDescriptor.getResponseMarshaller();
+    }
+
+    public Http2ClientConnection connection() {
```
tomas-langer, line 86: public methods on package private type

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcBaseClientCall.java
```diff
@@ -0,0 +1,209 @@
...
+
+    abstract void startStreamingThreads();
+
+    /**
+     * Unary blocking calls that use stubs provide their own executor which needs
+     * to be used at least once to unblock the calling thread and complete the
+     * gRPC invocation. This method submits an empty task for that purpose. There
+     * may be a better way to achieve this.
+     */
+    protected void unblockUnaryExecutor() {
```
tomas-langer, line 160: That is a strange requirement. Is there a chance we could work around this in a different way, or is it built into the `grpc` libraries?

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcChannel.java
```diff
@@ -0,0 +1,56 @@
...
+            MethodDescriptor<ReqT, ResT> methodDescriptor, CallOptions callOptions) {
+        MethodDescriptor.MethodType methodType = methodDescriptor.getType();
+        return methodType == MethodDescriptor.MethodType.UNARY
+                ? new GrpcUnaryClientCall<>(grpcClient, methodDescriptor, callOptions)
+                : new GrpcClientCall<>(grpcClient, methodDescriptor, callOptions);
+    }
+
+    @Override
+    public String authority() {
+        ClientUri clientUri = grpcClient.prototype().baseUri().orElseThrow();
```
tomas-langer, line 53: This is a strange code.
The `baseUri` is optional (by intention), and can be overridden for each client call (and may not be present at all).
Throwing and exception when not there seems to be wrong.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
```diff
@@ -0,0 +1,191 @@
...
+    protected void startStreamingThreads() {
+        // write streaming thread
+        writeStreamFuture = executor.submit(() -> {
+            try {
+                startWriteBarrier.await();
+                LOGGER.fine("[Writing thread] started");
+
+                boolean endOfStream = false;
+                while (isRemoteOpen()) {
+                    LOGGER.finest("[Writing thread] polling sending queue");
```
tomas-langer, line 106: For sockets, I would like to always include the socket id in the output. There is a method on socket context to log a message that will do that for you in a way consistent with other log statements (so we can correlate log output from multiple areas)

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
```diff
@@ -0,0 +1,191 @@
...
+                    BufferData bufferData = sendingQueue.poll(WAIT_TIME_MILLIS, TimeUnit.MILLISECONDS);
+                    if (bufferData != null) {
+                        if (bufferData == EMPTY_BUFFER_DATA) {     // end marker
+                            LOGGER.finest("[Writing thread] sending queue end marker found");
```
tomas-langer, line 110: dtto - these actually seem more like debugging statements than final logging statements

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
```diff
@@ -0,0 +1,191 @@
...
+                            if (!endOfStream) {
+                                LOGGER.finest("[Writing thread] sending empty buffer to end stream");
+                                clientStream().writeData(EMPTY_BUFFER_DATA, true);
+                            }
+                            break;
+                        }
+                        endOfStream = (sendingQueue.peek() == EMPTY_BUFFER_DATA);
+                        boolean lastEndOfStream = endOfStream;
+                        LOGGER.finest(() -> "[Writing thread] writing bufferData " + lastEndOfStream);
```
tomas-langer, line 119: If there is concatenation, it should be guarded by an if, or use template (supported by socket context)

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
```diff
@@ -0,0 +1,191 @@
...
+                        break;
+                    }
+
+                    // attempt to read and queue
+                    Http2FrameData frameData;
+                    try {
+                        frameData = clientStream().readOne(WAIT_TIME_MILLIS_DURATION);
+                    } catch (StreamTimeoutException e) {
+                        LOGGER.fine("[Reading thread] read timeout");
+                        continue;
```
tomas-langer, line 154: shouldn't this close this stream?

webclient/grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,46 @@
...
+         path = {"WebClient", "gRPC"}
+)
+module io.helidon.webclient.grpc {
+
+    requires static io.helidon.common.features.api;
+
+    requires transitive io.grpc;
+    requires transitive io.grpc.stub;
+    requires transitive io.helidon.builder.api;
+    requires transitive io.helidon.common.pki;
```
tomas-langer, line 35: Why is `pki` required here?

webclient/grpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,46 @@
...
+    requires transitive io.helidon.webclient.http2;
+    requires transitive io.helidon.webclient;
+
+    requires io.helidon.grpc.core;
+    requires java.logging;
```
tomas-langer, line 40: This should not be used

tomas-langer (commented):
Great work!

There are some error handling and logging issues (this being a draft), but the core of the work seems the right thing to do, thanks!

all/pom.xml
spericas: Sorry, didn't get this one. You mean for `helidon-webclient-grpc`?

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcBaseClientCall.java
spericas: I couldn't find a workaround for it so far. Yes, part of the grpc libraries.

webclient/grpc/src/main/java/module-info.java
spericas: It shouldn't

grpc/core/src/main/java/module-info.java
spericas: Roger, I thought either one was OK. Will switch code to `System.Logger` then.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
spericas: Even inside a lambda?

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
spericas: It's currently re-trying, although we could be more draconian I suppose

webclient/grpc/src/main/java/module-info.java
spericas: Roger

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcChannel.java
spericas: Good point, need to think about this one a bit more

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
spericas: Will check that

grpc/core/src/main/java/module-info.java
spericas: Roger

grpc/core/src/main/java/io/helidon/grpc/core/PriorityBag.java
```diff
@@ -0,0 +1,214 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.stream.Stream;
+
+// import io.helidon.common.Prioritized;
```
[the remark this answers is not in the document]
spericas: Will switch

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcChannel.java
spericas: After reviewing this, this base URI needs to be mandatory for gRPC, based on how clients/stubs are created. I replaced the code with a more informative error message.

spericas:
> After plugging the client into Coherence I am seeing a lot of severe level log messages due to timeouts. We use a bidirectional channel for Coherence events. The client makes the bi-di request which creates the bi-di channel, then the client can send messages up that channel to the server and the server can periodically send event messages back down the channel. It appears there is a 100ms timeout in the code that constantly logs the error. But, even though the error is logged a lot, everything appears to work, as the tests for events all pass.
> 
> The code is in GrpcClientCall line 152
> 
> ```
>                     try {
>                         frameData = clientStream().readOne(WAIT_TIME_MILLIS_DURATION);
>                     } catch (StreamTimeoutException e) {
>                         socket().log(LOGGER, ERROR, "[Reading thread] read timeout");
>                         continue;
>                     }
> ```

That timeout was too aggressive and has now been set to 2 seconds. We shall make this configurable eventually.

thegridman:
After running the latest build through our gRPC client tests we have the following issues.

- Still seeing constant timeout messages (which you mentioned above).
- Large Protobuf messages do not work, it appears they cannot be read on the server. 
- The fix for streams that do not send a response and just close the observer seems to work now. But what is not working is where the server throws an exception, this seems to just get lost and the call never ends on the client. To test this I added a method that specifically throws an exception and this fails. 

For the latest issues I updated the `GrpcStubTest` to add some reproducers. I have attached a patch file I exported from IntelliJ that you should be able to apply and reproduce the issues. All three additional tests will just hang as the client never completes the request.
[gRPC_Tests.patch](https://github.com/helidon-io/helidon/files/15091646/gRPC_Tests.patch)

thegridman:
I have been running our gRPC client tests using a build from this PR. The last time I ran them a few weeks ago they were ok. Now I have a failure in our fail-over tests. 

In this scenario the client has a bidirectional stream open to the server when the server is killed. When running the tests with Netty the client's StreamObserver onError method is called with a StatusRuntimeException with a Status of Unavailable. When using Helidon this does not happen, it appears that Helidon tries to reconnect to the server the next time a messages is sent up the stream by the client. This does not work, in our case because the server is gone and we need to fail over to a new server on another port. Helidon does not allow us to configure multiple server addresses, so this constant reconnect attempt fails until the client eventually gets a timeout exception.

We need the clients StreamObserver onError to be called, because basically that stream is dead. In our case we rely on state on the server that was initialised by previous messages, we cannot have a bidirectional stream transparently fail over behind our backs. 

In our client after the stream receives an error then the our client code can deal with reconnection. In our case we look up the new set of server endpoints and have to create a new Helidon client as the previous client only knows the single address we gave it previously. Obviously if our server was behind a load balancer that would be OK, but in a lot of cases there is no LB. With ManagedChannel in grpc-java there are various ways to pass multiple addresses or to do address lookup whenever the channel is about to connect, but Helidon does not have this. We have been able to work around this in our Helidon client code but it relies on the bidirectional (and presumably any other "stream" type calls, to be properly killed if the server disconnects.

thegridman:
Further to the above, I copied one of the bidirectional tests in `GrpcTest` and modified it to stop the `WebServer` between two calls to send a message. At no time does my StreamObserver receive an error, even when sending more messages, they are presumably and I see a sack trace in the log. I could fix this by changing `GrpcClientCall` line 128 that currently looks like this
```
            } catch (Throwable e) {
                socket().log(LOGGER, ERROR, e.getMessage(), e);
            }
            socket().log(LOGGER, DEBUG, "[Writing thread] exiting");
```
so that it calls the response listener like this:
```
            } catch (Throwable e) {
                socket().log(LOGGER, ERROR, e.getMessage(), e);
                responseListener().onClose(Status.UNKNOWN.withDescription(e.getMessage()).withCause(e), EMPTY_METADATA);
            }
            socket().log(LOGGER, DEBUG, "[Writing thread] exiting");
```
At least now my client StreamObserver onError is called when the send fails. But I still have not worked out how to call it when the socket closes.

thegridman:
After the latest changes I have run our 5,500 gRPC client tests using the Helidon client from this PR and they all pass (including the fail-over tests).

grpc/core/src/main/java/module-info.java
```diff
@@ -0,0 +1,33 @@
...
+
+    requires transitive io.helidon.common.context;
+    requires transitive io.helidon.common.config;
+    requires transitive io.grpc;
+    requires transitive io.grpc.stub;
+    requires transitive com.google.protobuf;
+    requires transitive io.grpc.protobuf;
+    requires transitive io.grpc.protobuf.lite;
+
+    requires jakarta.annotation;
```
tomas-langer, line 30: this dependency should not be here I think (and in pom file)

grpc/pom.xml
```diff
@@ -0,0 +1,38 @@
...
+        <version>4.0.0-SNAPSHOT</version>
+    </parent>
+    <packaging>pom</packaging>
+
+    <groupId>io.helidon.grpc</groupId>
+    <artifactId>helidon-grpc-project</artifactId>
+    <name>Helidon gRPC Project</name>
+
+    <description>gRPC support for Helidon</description>
+
```
tomas-langer, line 34: Please add this section to all new modules (unless in parent):
```
    <properties>
    </properties>
```

This will also catch some of problems with default public constructors

webclient/grpc/src/main/java/io/helidon/webclient/grpc/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Helidon WebClient HTTP/1.1 Support.
```
tomas-langer, line 18: Please fix comment.

grpc/core/src/main/java/io/helidon/grpc/core/MarshallerSupplier.java
```diff
@@ -0,0 +1,90 @@
...
+     * @return a {@link MethodDescriptor.Marshaller} for a type
+     */
+    <T> MethodDescriptor.Marshaller<T> get(Class<T> clazz);
+
+    /**
+     * Obtain the default marshaller.
+     *
+     * @return the default marshaller
+     */
+    static MarshallerSupplier defaultInstance() {
```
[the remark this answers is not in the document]
spericas: Renamed to `create`. Original name was from 3.x. See above.

grpc/core/src/main/java/io/helidon/grpc/core/MarshallerSupplier.java
```diff
@@ -0,0 +1,90 @@
...
+                    clazz.getName());
+            throw new IllegalArgumentException(msg);
+        }
+    }
+
+    /**
+     * A {@link MarshallerSupplier} implementation that
+     * supplies Protocol Buffer marshaller instances.
+     */
+    class ProtoMarshallerSupplier implements MarshallerSupplier {
```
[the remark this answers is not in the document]
spericas: Moved to upper level.

grpc/core/src/main/java/io/helidon/grpc/core/WeightedBag.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Add an element to the bag with a specific weight.
+     *
+     * @param value  the element to add
+     * @param weight the weight of the element
+     */
+    public void add(T value, double weight) {
+        if (value != null) {
```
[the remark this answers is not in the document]
spericas: Updated implementation to use weights greater than 0 and forbid `null` values

grpc/core/src/main/java/io/helidon/grpc/core/WeightedBag.java
```diff
@@ -0,0 +1,218 @@
...
+        return Stream.concat(stream, noWeightedList.stream());
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return stream().iterator();
+    }
+
+    private List<T> combine(List<T> list, T value) {
+        if (list == null) {
```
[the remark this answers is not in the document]
spericas: Fixed

grpc/core/src/main/java/io/helidon/grpc/core/WeightedBag.java
```diff
@@ -0,0 +1,218 @@
...
+     * Create a new {@link WeightedBag} where elements added with no weight will be last
+     * in the order.
+     *
+     * @param <T> the type of elements in the bag
+     * @return a new {@link WeightedBag} where elements
+     *         dded with no weight will be last in the
+     *         order
+     */
+    public static <T> WeightedBag<T> create() {
+        return withDefaultWeight(-1.0);
```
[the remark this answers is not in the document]
spericas: Fixed

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientImpl.java
```diff
@@ -0,0 +1,64 @@
...
+    private final Http2Client http2Client;
+    private final GrpcClientConfig clientConfig;
+
+    GrpcClientImpl(WebClient webClient, GrpcClientConfig clientConfig) {
+        this.webClient = webClient;
+        this.http2Client = webClient.client(Http2Client.PROTOCOL);
+        this.clientConfig = clientConfig;
+    }
+
+    public WebClient webClient() {
```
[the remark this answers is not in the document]
spericas: Fixed

webclient/grpc/src/main/java/io/helidon/webclient/grpc/package-info.java
spericas: Fixed

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
```diff
@@ -0,0 +1,249 @@
...
+    @Override
+    public void halfClose() {
+        socket().log(LOGGER, DEBUG, "halfClose called");
+        sendingQueue.add(EMPTY_BUFFER_DATA);       // end marker
+        startWriteBarrier.countDown();
+    }
+
+    @Override
+    public void sendMessage(ReqT message) {
+        socket().log(LOGGER, DEBUG, "sendMessage called");
```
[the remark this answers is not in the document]
spericas: I'm not sure exactly what you are suggesting here. Is this related to the large number of calls to `log()`?

PR #8427, opened 2024-02-27
4.x: Update `context-root` config documentation

### LINKED ISSUE #8365: 4.x: Document application context-root configuration
## Environment Details
* Helidon Version: 4.0.5
* Helidon MP

----------

## Problem Description

A user wants to configure the context root for the implicit (aka "synthetic")  JAXRS application.
The current [MP Server docs](https://helidon.io/docs/4.0.5/mp/server#_configuration_override_of_routing_path) mention the `routing-path` sub-key however it's mostly from the perspective of the `@RoutingPath` annotation.

One of the snippet uses an application like class name which indicates that it works also for `jakarta.ws.rs.Application`:
```yaml
io.helidon.examples.AdminApplication:
  routing-name:
    name: "management"
    required: true
  routing-path:
    path: "/management"
```

However this is written in YAML instead of `microprofile-config.properties` which is what is generated by the archetypes.

Also, when not using an application class explicitly it's unclear what the class name should be.
I.e. How to configure the context root for the most common case: using an implicit Application.

Here is the solution:
```properties
jakarta.ws.rs.core.Application.routing-path.path=/foo
```

---

So currently users have to figure out the following:

- the `routing-path` configuration sub-key works for JAXRS application
- know what the class name for the implicit application
- how to convert the YAML snippet to properties

We need to address all the items above.

tvallin:
### Description

fix #8365

Add details about application `context-root` configuration.

### Documentation

updated in this PR.

### DIFF
--- a/docs/src/main/asciidoc/mp/jaxrs/jaxrs-applications.adoc
+++ b/docs/src/main/asciidoc/mp/jaxrs/jaxrs-applications.adoc
@@ -69,7 +69,7 @@ NOTE: Helidon treats `@Path` and `@Provided` as bean-defining annotations but, a
 == Setting Application Path
 
 The application path, also known as context root, is the base URI used to serve all resource URIs provided
-by `@Path` annotation. This section describe how to set it with annotation or configuration files.
+by `@Path` annotation. This section describes how to set it with an annotation or configuration file.
 
 When an `Application` subclass is provided, use the `@ApplicationPath`:
 
@@ -97,7 +97,7 @@ The same configuration works for `.properties` file:
 io.helidon.examples.MyApplication.routing-path.path=/my-application
 ----
 
-If no `Application` is not provided, a _synthetic_ subclass is created and can be configured using
+If an `Application` is not provided, a _synthetic_ subclass is created and can be configured using
 this property:
 
 [source,properties]
--- a/docs/src/main/asciidoc/mp/jaxrs/jaxrs-applications.adoc
+++ b/docs/src/main/asciidoc/mp/jaxrs/jaxrs-applications.adoc
@@ -64,7 +64,46 @@ in each subclass using the collections in steps (2) and (3) only as defaults, i.
 both return empty sets.
 
 NOTE: Helidon treats `@Path` and `@Provided` as bean-defining annotations but, as stated above,
-`Application` subclasses may require additional annotations depending on the discovery mode.
+`Application` subclasses may require additional annotations depending on the discovery mode
+
+== Setting Application Path
+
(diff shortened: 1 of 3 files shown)


romain-grecourt (dismissed):

docs/src/main/asciidoc/mp/server.adoc
```diff
@@ -312,6 +312,20 @@ io.helidon.examples.AdminService:
     path: "/management"
 ----
 
+The same example configuration for `.properties` file:
```
barchetta, line 315: I think information about setting the JAX-RS application's context path should go in `mp/jaxrs/jaxrs-applications.adoc`.  Maybe add a new section after "Discovery of REST Beans" since it mentions `@ApplicationPath`.  Maybe:

1. Add a new section "Setting Application Path" (mention in description this is also know as context path)
2. Show an example of how to do it using `@ApplicationPath` (don't use AdminService, use something like MyApplication).
3. Show how to override that using configuration (when you have an Application class)
4. Show how to configure it if there is no Application class

Keep in mind the question we are trying to answer is "How do I set the context path of my JAX-RS application".  So it shouldn't be buried in the Helidon server routing feature.

docs/src/main/asciidoc/mp/jaxrs/jaxrs-applications.adoc
```diff
@@ -64,7 +64,46 @@ in each subclass using the collections in steps (2) and (3) only as defaults, i.
 both return empty sets.
 
 NOTE: Helidon treats `@Path` and `@Provided` as bean-defining annotations but, as stated above,
-`Application` subclasses may require additional annotations depending on the discovery mode.
+`Application` subclasses may require additional annotations depending on the discovery mode
+
+== Setting Application Path
+
+The application path, also known as context root, is the base URI used to serve all resource URIs provided
+by `@Path` annotation. This section describe how to set it with annotation or configuration files.
```
barchetta, line 72: "describe" -> "describes"

"with annotation" -> "with an annotation"

"files" -> "file"

docs/src/main/asciidoc/mp/jaxrs/jaxrs-applications.adoc
```diff
@@ -64,7 +64,46 @@ in each subclass using the collections in steps (2) and (3) only as defaults, i.
...
+----
+
+The same configuration works for `.properties` file:
+
+[source,properties]
+----
+io.helidon.examples.MyApplication.routing-path.path=/my-application
+----
+
+If no `Application` is not provided, a _synthetic_ subclass is created and can be configured using
```
barchetta, line 100: "If no Application" -> "If an Application class"

PR #8433, opened 2024-02-28
[4.x] Examples security validation

senivam:
### Description
Security validation for examples (to be extracted to the separate repository)


romain-grecourt (commented):
examples/webserver/mutual-tls/README.md needs to be updated as it hard-codes the old password

PR #8449, opened 2024-03-02
Fix OTel agent detector test that was polluting the JVM for later tests

### LINKED ISSUE #8448: MP telemetry test failures - intermittent, sometimes reliable in the pipeline or for certain developers locally; sometimes not
## Environment Details
* Helidon Version: 4.x
* JDK version:
* OS:

----------

## Problem Description
The tests in `helidon-microprofile-telemetry` would work locally but fail in pipelines or for other developers.

Santiago, in an unrelated PR #8327, added a workaround to run each test in its own JVM. 

This should not be needed. This issue is to investigate and find a less heavyweight solution.

tjquinno:
### Description
Resolves #8448 

Some pipeline runs and some developers' local builds experienced reliable failures, other pipeline runs and developers saw no problems.

There had been some conjecture that some static storage was contributing to this problem.

Instead, the problem turned out to be a bug in the `AgentDetectorTest` class.

#### Background
Our `AgentDetector` examines config and system properties (and the OTel context) to try to determine if the OTel agent is present. Our `OpenTelemetryProducer` uses the `AgentDetector` to decide whether to use the `OpenTelemetry` instance that the agent will already have set (if the agent is present) or to initialize and set an `OpenTelemetry` instance according to current config settings (if the agent is absent).

Some of our tests use an in-memory `TestSpanExporter` to see if spans were managed as expected. The use of this test exporter is conveyed through configuration that is set at test-time and so _will not_ be used if the agent is detected.

#### The problem

The `AgentDetectorTest` set a system property to make sure that the detector correctly responded, but the test did not reset or clear the system property after the test.

If the agent test ran before other tests that depended on our custom in-memory collector, the later tests would fail because tests would skip using the configuration to prepare OTel and, therefore, the in-memory collector. Because span data would not  be sent to the in-memory collector in those cases, tests that expected span data in order to check it would fail. The non-deterministic ordering of test execution explains why some builds saw failures while others did not.

#### The fix
The PR changes the `AgentDetectorTest` to restore or clear the system property at the end of the test to avoid disrupting later tests.

The PR also removes the changes to the surefire plug-in config that forces tests into their own JVMs since that workaround is no longer needed.

I verified the problem and the fix by temporarily forcing the order of the tests so the agent detector test preceded at least one of the tests that failed.

### Documentation
No impact.

### DIFF
--- a/microprofile/telemetry/pom.xml
+++ b/microprofile/telemetry/pom.xml
@@ -32,6 +32,11 @@
         Support for MicroProfile Telemetry
     </description>
 
+    <properties>
+        <!-- The default for the following is 5000 ms. Reducing it speeds up tests significantly. -->
+        <otel.bsp.schedule.delay>100</otel.bsp.schedule.delay>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>io.helidon.tracing.providers</groupId>
@@ -141,8 +146,9 @@
                     <execution>
                         <id>default-test</id>
                         <configuration>
-                            <!-- The default for the following is 5000 ms. Reducing it significantly speeds up tests. -->
-                            <argLine>-Dotel.bsp.schedule.delay=100</argLine>
+                            <systemPropertyVariables>
+                                <otel.bsp.schedule.delay>${otel.bsp.schedule.delay}</otel.bsp.schedule.delay>
+                            </systemPropertyVariables>
                         </configuration>
                     </execution>
                 </executions>
--- a/microprofile/telemetry/pom.xml
+++ b/microprofile/telemetry/pom.xml
@@ -143,7 +143,6 @@
                         <configuration>
                             <!-- The default for the following is 5000 ms. Reducing it significantly speeds up tests. -->
                             <argLine>-Dotel.bsp.schedule.delay=100</argLine>
-
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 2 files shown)


microprofile/telemetry/pom.xml
```diff
@@ -137,12 +137,15 @@
...
-                    <reuseForks>false</reuseForks>
-                    <!-- The default for the following is 5000 ms. Reducing it significantly speeds up tests. -->
-                    <argLine>-Dotel.bsp.schedule.delay=100</argLine>
-                </configuration>
+                <executions>
+                    <execution>
+                        <id>default-test</id>
+                        <configuration>
+                            <!-- The default for the following is 5000 ms. Reducing it significantly speeds up tests. -->
+                            <argLine>-Dotel.bsp.schedule.delay=100</argLine>
```
ljnelson, line 145: Consider, rather:
```
<systemPropertyVariables>
  <otel.bsp.schedule.delay>${otel.bsp.schedule.delay}</otel.bsp.schedule.delay>
</systemPropertyVariables>
```
…and, elsewhere in the `pom.xml`:
```
<properties>
  <otel.bsp.schedule.delay>100</otel.bsp.schedule.delay>
</properties>
```

tjquinno: Fixed

ljnelson (approved):

Issue #8164, opened 2023-12-19
4.x: Duplicate graphql.java module


romain-grecourt:
## Environment Details
* Helidon Version: 4.0.1
----------

## Problem Description

`com.graphql-java:graphql-java:18.6` depends on `com.graphql-java:java-dataloader:3.1.2` and both define `Automatic-Module-Name` to `com.graphqljava`.

## Steps to reproduce

```bash
jar --describe-module --file ~/.m2/repository/com/graphql-java/graphql-java/18.6/graphql-java-18.6.jar
No module descriptor found. Derived automatic module.

com.graphqljava@18.6 automatic
requires java.base mandated
contains graphql
contains graphql.analysis
contains graphql.cachecontrol
contains graphql.collect
contains graphql.com.google.common.base
contains graphql.com.google.common.collect
contains graphql.com.google.common.math
contains graphql.com.google.common.primitives
contains graphql.execution
contains graphql.execution.directives
contains graphql.execution.instrumentation
contains graphql.execution.instrumentation.dataloader
contains graphql.execution.instrumentation.fieldvalidation
contains graphql.execution.instrumentation.nextgen
contains graphql.execution.instrumentation.parameters
contains graphql.execution.instrumentation.threadpools
contains graphql.execution.instrumentation.tracing
contains graphql.execution.nextgen
contains graphql.execution.nextgen.result
contains graphql.execution.preparsed
contains graphql.execution.preparsed.persisted
contains graphql.execution.reactive
contains graphql.introspection
contains graphql.language
contains graphql.nextgen
contains graphql.normalized
contains graphql.parser
contains graphql.parser.antlr
contains graphql.relay
contains graphql.scalar
contains graphql.schema
contains graphql.schema.diff
contains graphql.schema.diff.reporting
contains graphql.schema.idl
contains graphql.schema.idl.errors
contains graphql.schema.impl
contains graphql.schema.transform
contains graphql.schema.usage
contains graphql.schema.validation
contains graphql.schema.visibility
contains graphql.util
contains graphql.validation
contains graphql.validation.rules
```

```bash
jar --describe-module --file ~/.m2/repository/com/graphql-java/java-dataloader/3.1.2/java-dataloader-3.1.2.jar
No module descriptor found. Derived automatic module.

com.graphqljava@3.1.2 automatic
requires java.base mandated
contains org.dataloader
contains org.dataloader.annotations
contains org.dataloader.impl
contains org.dataloader.registries
contains org.dataloader.stats
[...]

PR #8450, opened 2024-03-03
Bump graphql-java version to 21.x

agavrilov76:
### Description

The current version of GraphQL doesn't work well applications running as Java Modules. The following error occurs when running with `java -p`

`> helidon-quickstart-se $JAVA_HOME/bin/java  --list-modules -p target/libs/ 
Error occurred during initialization of boot layer java.lang.module.FindException: Two versions of module com.graphqljava found in target/libs (graphql-java-18.6.jar and java-dataloader-3.1.2.jar)`

Updating the library version solves the problem

### Documentation

None


agavrilov76:
Already addressed in
https://github.com/helidon-io/helidon/issues/8164

PR #8452, opened 2024-03-04
4.x span callbacks using restrictive types

### LINKED ISSUE #8352: [4.x] Invoke developer-provided tracing callbacks--e.g, as part of creating a new span builder and starting and ending a span
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Helidon's tracing support currently adds tags declared in configuration to all traces and/or spans it creates, regardless of which tracing implementation is present. 

It could be useful for developers to provide code to add tags to a new tracer or tags, baggage, or events to new span builders or spans dynamically before those objects are returned to the caller requesting them.

One way to accomplish that--and very useful other things as well--would be to support extensibility around creating a span builder and around starting and ending a span.

There has been some discussion among users of exposing tracing spans as Java Flight Recorder events. A well-designed extensibility approach could include that as a special case. It might make sense for Helidon to provide that particular optional tracing extension (if we adopt extensibility).

From user discussion about JFR integration:
> This helps co-relating. JVM level  JFR events/data such as GC ,  socket reads/writes, sample method profiles (Hot methods)   carrier thread Pinning, Thread cpu usage etc recorded by the JVM ,   with Span execution and. provide next level diagnostics on the events that  occurred during the  Span execution and provide in-depth insights into why a specific Span execution may be slow.
Specifically for cases where we have leaf spans with no child Spans , and also we see gaps between the  start of the  parent span and  start of the  child Span

tjquinno:
### Description
Resolves #8352 using restrictive adapters around existing types (`Span`, `Span.Builder`, `Scope`) passed to the developer's callback methods so user code cannot change the state of the span or scope.

Lifecycle events for callbacks:

* `beforeStart(Span.Builder<?>)`
* `afterStart(Span)`
* `afterActivate(Span, Scope)`
* `afterClose(Span, Scope)`
* `afterEnd(Span)` (successful)
* `afterEnd(Span, Throwable)` (unsuccessful)

### Documentation
Included in the PR.


tjquinno:
Superseded by #8619

Issue #8377, opened 2024-02-13
Some mocked OCI interfaces need to be updated every time OCI makes a backwards-incompatible release


ljnelson:
Here is an arbitrary example:

https://github.com/helidon-io/helidon/blob/16ca27b5da8151a1e7241f76993ad75874ac71e3/integrations/oci/metrics/cdi/src/test/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsCdiExtensionTest.java#L186

The `Monitoring` interface gained a few methods, which is a backwards-incompatible change. Currently, we solve this problem by manually adding "no-op" methods as appropriate.

Instead, for mock implementations of business interfaces such as these, we should use something like `java.lang.reflect.Proxy` or Mockito or something to implement only the parts we truly need to test.

klustria:
I think I have an idea on what to do with this, hence moving to Sprint Scope.

klustria:
Resolved via PR https://github.com/helidon-io/helidon/pull/8474.

PR #8474, opened 2024-03-12
Avoid implementing the OCI Monitoring interface and instead use Mockito to mock it.

klustria:
### Description
The test in **io.helidon.integrations.oci.metrics.cdi** uses an implemented OCI Monitoring class. When an OCI SDK upgrade is made and the Monitoring class has changed (for example, a new method is added or a method is removed or a method's signature has changed), this will lead to a need to update the implementation which is quite cumbersome. The goal of this change is to avoid that issue by using Proxy to mock only the Monitoring class methods that are required by the test. The issue was filed via https://github.com/helidon-io/helidon/issues/8377.

This change also includes a resolution to https://github.com/helidon-io/helidon/issues/7739 that is caused by the Monitoring.postMetricData() being called before the test metrics are registered.

### Solutions
1. For https://github.com/helidon-io/helidon/issues/8377:
   * Remove the implementation of Monitoring interface.
   * Create an extension that simulates OciExtension but will only process Monitoring annotation and if the annotation is found, that it will inject a mocked Monitoring bean .
2. For https://github.com/helidon-io/helidon/issues/7739:
    * If the mocked Monitoring.postMetricData is invoked, check if the # of expected metrics is greater than 0. If it is, then a signal will be sent to the test to proceed with validation. If not, then that means that the metrics has not been registered yet, so just wait for next invocation and try again.
     
### Documentation
No doc impact as this only affects the Unit Test.

### DIFF
--- a/integrations/oci/metrics/cdi/pom.xml
+++ b/integrations/oci/metrics/cdi/pom.xml
@@ -43,6 +43,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>io.helidon.integrations.oci.sdk</groupId>
+            <artifactId>helidon-integrations-oci-sdk-cdi</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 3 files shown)


integrations/oci/metrics/cdi/pom.xml
```diff
@@ -57,6 +62,11 @@
             <artifactId>hamcrest-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
```
ljnelson, line 67: I guess this is possible but another approach might be to use `java.lang.reflect.Proxy` etc. since `Monitoring` is an interface. Might be simpler.

klustria: Good point, and I've made the change to use `java.lang.reflect.Proxy` to simplify mocking and avoid adding the Mockito dependency..

barchetta (approved):

PR #8487, opened 2024-03-19
2.x: Reset fault tolerance method cache on CDI shutdown

tomas-langer:
so it can be started with a different configuration.

Resolves #8478 

### Description
When running tests the MP Fault tolerance method cache stayed in (it is static), not refreshing it based on new configuration.
This change makes sure the cache is cleared when CDI is shut down.

### Documentation
No impact

Issue #8478, opened 2024-03-14
3.x: Helidon MP FaultTolerance fails In HelidonTest


tejo93:
## Environment Details
* Helidon Version: 3.2.3
*  Helidon MP
* JDK version: 21
* OS: OL- 8 Linux
* Docker version (if applicable): N/A

----------

### I have this following configuration in src/main/resources/application.yaml

```java
# Application properties. This is the default greeting
app:
  greeting: Hello
counter: 0

# Microprofile server properties
server:
  port: 8080
  host: 0.0.0.0

io.helidon.examples.quickstart.mp.GreetResource/Retry/enabled: true
io.helidon.examples.quickstart.mp.GreetResource/Timeout/enabled: true
io.helidon.examples.quickstart.mp.GreetResource/Retry/maxRetries: 3
io.helidon.examples.quickstart.mp.GreetResource/Timeout/value: 5000
```

### Class GreetResource is annotated like this :

```java
@Path("/greet")
@Retry
@Timeout
public class GreetResource 
```

### Method getDefaultMessage :

```java
@GET
    @Produces(MediaType.APPLICATION_JSON)
    public JsonObject getDefaultMessage() throws InterruptedException {
        this.greetingProvider.setCounter(this.greetingProvider.counter()+1);
        System.out.println("call made to getDefaultMessage : " + this.greetingProvider.counter());
        Thread.sleep(100);
        if(this.greetingProvider.counter()==3) {
            return createResponse("World");
        }

        throw new RuntimeException("Default exception thrown");
    }
```

### Class GreetingProvider has following changes

```java
private final AtomicReference<Integer> counter = new AtomicReference<>();

@Inject
    public GreetingProvider(@ConfigProperty(name = "app.greeting") String message, @ConfigProperty(name = "counter") String counter) {
        this.message.set(message);
        this.counter.set(Integer.parseInt(counter));
    }

int counter() { return counter.get();}
void setCounter(int counter) {
        this.counter.set(counter);
    }
```

### I have test/resources/application.yaml

```java
io.helidon.examples.quickstart.mp.GreetResource/Retry/enabled: false
io.helidon.examples.quickstart.mp.GreetResource/Timeout/enabled: false
```

### Class MainTest

```java

@HelidonTest
@AddConfig(key ="io.helidon.examples.quickstart.mp.GreetResource/Retry/enabled", value = "true")
[...]

PR #8491, opened 2024-03-19
4.x: Introducing a test to validate that combination of config annotations…

tomas-langer:
… and test yaml works in HelidonTest together with FaultTolerance.
And actual fix of the problem - the same way this was fixed in 2.x.

Related to https://github.com/helidon-io/helidon/issues/8478
"forward port" of https://github.com/helidon-io/helidon/pull/8487

###Description
This was a bug in 2.x that is fixed as a side effect of other changes in 3.x. Adding a test to make sure we continue supporting this.


spericas (approved):

tomas-langer:
I just had a talk with a customer that claims this is not working in 4.x. I need to review their reproducer and see the status. I will either merge this PR, or update it with new test and a fix

spericas (approved):

PR #8523, opened 2024-03-22
4.x: Make HttpRouting an interface

### LINKED ISSUE #8572: 4.x: Make HttpRouting an interface
## Environment Details
* Helidon Version: 4.0.6
* Helidon SE

----------

## Problem Description
This is more of an improvement proposal but does address an actual issue as well (though probably not one that people will encounter in practice). Currently `HttpRoute` is an interface, whereas `HttpRouting` isn't (it's a final class). It would be good to be consistent and make both interfaces, as it allows for more configurability, and almost everything else that can be specified in a `WebServer` builder is an interface.

### Use Case
Being able to implement your own `HttpRouting` type can be useful when moving an existing service to Helidon; for example if there already is custom routing logic for it.

### Issue to Solve Along the Way
As for the actual issue (one might call it a bug) I encountered while working on this: other routing classes (such as `GrpcRouting`) aren't final, which one might understand as Helidon supporting custom implementations. However, if one does so, requests aren't properly routed anymore. The reason for this is that the `RouterImpl` indexes `Routing` instances by their class and its `routing` method is always called with a fixed class. Therefore, when trying to use a custom implementation, it wouldn't be found.

This issue needs to be addressed if one wants to make `HttpRouting` an interface. As a side-effect it would fix the "bug" just described, allowing the `RouterImpl` to handle custom implementations of other routing types.

### Compatibility
This change should be (mostly) backwards compatible. Currently, the `HttpRouting` class is final (so no existing code can break because it subclasses `HttpRouting`) and its constructor is private (so no existing code can break because the constructor won't exist anymore).

### Other Considerations
Even with `HttpRouting` becoming an interface, it wouldn't be straightforward to use a custom implementation. The `WebServer.Builder.routing` method takes a `HttpRouting.Builder`; providing a `HttpRouting` instance directly isn't possible. So, to use a custom routing implementation, one would also have to also create a custom `HttpRouting.Builder`, whose `build` method returns an instance of the custom routing implementation. This isn't great but also not a big deal as most users won't be doing this. Anyway, improving this is out of scope of this issue.

VerKWer:
### Description
Currently `HttpRoute` is an interface whereas `HttpRouting` isn't. Being able to implement your own `HttpRouting` type can be useful when moving an existing service to Helidon; for example if there already is custom routing logic for it. In addition, making this change seems ideologically correct because it allows for more configurability, and almost everything else that can be specified in a `WebServer` builder is an interface.

Another issue that's addressed as a side-effect is that while other routing classes (e.g. `GrpcRouting`) aren't final, using a custom implementation isn't easily possible. The reason for this is that the `RouterImpl` indexes `Routing` instances by their class and its `routing` method is always called with a fixed class. Therefore, when trying to use a custom implementation, it wouldn't be found.

### Documentation
The `HttpRouting` class is replaced by an interface and all implementation/builder code previously in it is moved to the new package-private `HttpRoutingImpl` class. This should not break existing applications because `HttpRouting` was `final` (so no subclasses can be defined) and its constructor was `private` (so it cannot be called directly).

The `routingType()` method is added to the `Routing` interface with a `default` implementation consistent with the previous behaviour. Its result is used by `RouterImpl` to determine the correct `Routing` instance to use. I'm not sure returning a `Class` here is the best solution, but it is what was previously used.

Resolves #8572

### DIFF
--- a/microprofile/websocket/src/main/java/io/helidon/microprofile/tyrus/TyrusRouting.java
+++ b/microprofile/websocket/src/main/java/io/helidon/microprofile/tyrus/TyrusRouting.java
@@ -42,6 +42,9 @@ public class TyrusRouting implements Routing {
         this.extensions = builder.extensions;
     }
 
+    @Override
+    public Class<? extends Routing> routingType() { return TyrusRouting.class; }
+
     /**
      * Builder for WebSocket routing.
      *
--- a/microprofile/websocket/src/main/java/io/helidon/microprofile/tyrus/TyrusRouting.java
+++ b/microprofile/websocket/src/main/java/io/helidon/microprofile/tyrus/TyrusRouting.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -43,7 +43,9 @@ public class TyrusRouting implements Routing {
     }
 
     @Override
-    public Class<? extends Routing> routingType() { return TyrusRouting.class; }
+    public Class<? extends Routing> routingType() {
+        return TyrusRouting.class;
+    }
 
     /**
      * Builder for WebSocket routing.
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcRouting.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcRouting.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
  *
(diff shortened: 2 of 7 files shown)


tomas-langer:
Hello @VerKWer - I have approved the pipeline run. 
In general we start from an issue in our issue tracker. If there is none yet created, kindly create it and link it to your PR. If there is one, just link it. This is the kind of work that requires approval of design. Although I agree with the approach you have taken, it would be better to discuss this first on an issue, or through our Slack channel, to make sure we are OK with such a change.

Also this is going into the internals of WebServer, so I will need to spend a bit of time making sure this does not introduce any kind of backward incompatibility.

Thanks for your contribution!

VerKWer:
Hey @tomas-langer! Thank you for having a look at the PR and pointing out the proper procedures; this is my first time contributing to Helidon, and I'll gladly create an issue to link to.

I did have similar thoughts about maybe first reaching out to check whether my contribution was welcome. To that end, I asked in the Nima Slack channel at the time but realise only now that that was more than four months ago on November 14 (time really does fly). I can't see the thread anymore, but I think it was you, who answered back then that you'd be open for the change, which was very encouraging. Still, I'll make sure to communicate better for future contributions, and to always create an issue first.

tomas-langer (approved):

PR #8554, opened 2024-03-25
[4.x] Observers now inherit weight of the ObserveFeature.

### LINKED ISSUE #8550: [4.x] Feature weight is not honored in server routing
When features are discovered automatically, and a weight is customized, it is ignored by routing in webserver.

To reproduce:
Quickstart SE
Update applicaiton.yaml with this:
```yaml
server:
  port: 8080
  host: 0.0.0.0
  # default behavior to discover Server features (observability, openapi, metrics)
  # features-discovers-services: true
  # all of the below is discovered automatically
  features:
    observe:
      weight: 120
```

In Main.java:
```java
routing.register("/greet", new GreetService())
                .get("/observe/health", (req, res) -> res.send("From routing"));
```

We expect that the endpoint `http://localhost:8080/observe/health` should return appropriate response from Health observer, and not the "From routing" response.

Right now it returns "From routing", so the change in configuration was not honored.

tomas-langer:
HttpFeatures now inherit weight of ServerFeatures. As ObserveFeature weight can be configured, we can control whether it is executed before or after user routing (default is after).

Resolves #8550 

This now works as designed. You can configure `ObserveFeature` weight through configuration to be higher than default (100) to be invoked first.

### DIFF
--- a/examples/security/outbound-override/src/main/java/io/helidon/security/examples/outbound/OutboundOverrideExample.java
+++ b/examples/security/outbound-override/src/main/java/io/helidon/security/examples/outbound/OutboundOverrideExample.java
@@ -24,6 +24,7 @@ import io.helidon.security.SecurityContext;
 import io.helidon.security.Subject;
 import io.helidon.webserver.WebServer;
 import io.helidon.webserver.WebServerConfig;
+import io.helidon.webserver.context.ContextFeature;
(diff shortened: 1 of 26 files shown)


tjquinno (dismissed):
LGTM

tjquinno (approved):
LGTM

PR #8567, opened 2024-03-26
Fix handling of baggage when retrieving current span

### LINKED ISSUE #8548: [3.x] - Not able to propagate baggage using helidon-tracing-jaeger
Backport of #8504 for Helidon 3.x

tjquinno:
### Description
Resolves #8548 

OpenTelemetry separates baggage from spans and stores them separately in the OTel context. In contrast, the Helidon tracing API treats baggage as a characteristic of a span.

Earlier PRs fixed a problem in which Helidon failed to store the baggage associated with a span correctly in the OTel context. This PR fixes a problem in reconstituting the correct current span from what is in the OTel context, making sure the new reconstituted span draws from _both_ the OTel span in the context _and_ the OTel baggage in the context.

The central fix is in `OpenTelemetryTracerProvider` and its `currentSpan` method (which now delegates to `activeSpan` so the Jaeger code can also use it). The `Span` constructed for `currentSpan`/`activeSpan` now holds a reference to the OTel `Baggage` instance from the Otel `Context`. 

Normally, that baggage will be our writeable baggage implementation of OTel's `Baggage` interface. That implementation allows the Helidon `Span` wrapper around the OTel span to permit updates to its baggage. By storing a reference to that baggage in the reconstituted span, changes to the baggage via the baggage methods on the reconstituted `Span` will write through to the baggage that is stored in the context. 

But, it's possible that _immutable_ Otel baggage might be in the OTel context when the current span is reconstructed. The Helidon `OpenTelemetrySpan` method for updating baggage therefore must now check to see if its baggage is writeable or not and act accordingly.

The PR also contains new tests for OTel and Jaeger (which relies on the OTel implementation internally).

### Documentation
Bug fix; no doc changes.

### DIFF
--- a/tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/OpenTelemetrySpan.java
+++ b/tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/OpenTelemetrySpan.java
@@ -23,6 +23,7 @@ import io.helidon.tracing.Scope;
 import io.helidon.tracing.Span;
 import io.helidon.tracing.SpanContext;
 
+import io.opentelemetry.api.baggage.Baggage;
 import io.opentelemetry.api.common.Attributes;
 import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.api.trace.StatusCode;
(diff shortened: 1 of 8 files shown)


tracing/opentelemetry/src/main/java/io/helidon/tracing/opentelemetry/OpenTelemetrySpan.java
```diff
@@ -100,7 +107,12 @@ public Scope activate() {
     public Span baggage(String key, String value) {
         Objects.requireNonNull(key, "baggage key cannot be null");
         Objects.requireNonNull(value, "baggage value cannot be null");
-        baggage.baggage(key, value);
+        if (baggage instanceof MutableOpenTelemetryBaggage mutableBaggage) {
```
tomas-langer, line 110: Is there maybe a way to switch to mutable baggage here and make it work, or is this a limitation of the open telemetry API?
If we have this check + exception, then an application that works fine with other tracing provider may fail with oTel

tomas-langer (approved):

PR #8472, opened 2024-03-12
Add `close()` method to `DbClient` to allow proper resource cleanup

bjcoombs:
### Description:
Fixes #8402

Currently, there is no way to explicitly close a `DbClient` instance, which leads to connection leaks and resource exhaustion, particularly when using JDBC with the Hikari connection pool. This issue becomes evident in testing scenarios where multiple tests create `DbClient` instances, causing the database to refuse new connections after a certain number of tests.

This pull request addresses this issue by:

1. Adding a `close()` method to the `DbClient` interface, allowing users to explicitly close the `DbClient` and release associated resources.

2. Implementing the `close()` method in the `DbClientBase` class with a default no-op implementation.

3. Overriding the `close()` method in the `JdbcClient` class to close the underlying connection pool when the `DbClient` is closed.

4. Introducing a new `CloseableJdbcConnectionPool` interface that extends `JdbcConnectionPool` and `AutoCloseable`, ensuring that the connection pool can be properly closed.

5. Updating the `HikariConnectionPool` class to implement the `CloseableJdbcConnectionPool` interface and close the `HikariDataSource` in the `close()` method.

With these changes, users can now explicitly close the `DbClient` when it is no longer needed, ensuring that resources are properly released. This is particularly useful in testing scenarios where multiple `DbClient` instances are created and need to be cleaned up to prevent resource leaks.

Furthermore, the `DbClient` interface now extends `AutoCloseable`, allowing it to be used in try-with-resources statements for automatic resource management.

Example usage:
```java
try (DbClient dbClient = DbClient.create(config)) {
    // Use the dbClient
    // ...
} // dbClient.close() will be called automatically
```

PR #8571, opened 2024-03-27
[4.x] Closeable jdbc pool

### LINKED ISSUE #8402: Close DBClient
I cannot see a way to `close()` a `DBClient`. 

I have a `JDBCClient` backed by a Hikari pool and I'm running tests. After about 15 tests I get

```
FATAL: sorry, too many clients already
```

## Environment Details
* Helidon Version: 4.0.5
* Helidon SE
* JDK version:  Zulu21.30+15-CA (build 21.0.1+12-LTS)
* OS: Sonoma 14.3.1

----------

## Problem Description

Whenever a `DBClient` is created, when using JDBC with Hikari it creates a connection pool. There is no way to shut down the `DBClient` afterwards so these connections stay open. After a number of tests - 15 in my case - Postgres refuses new connections.

To workaround I can force the database to restart (I use `TestContainers`) but that does make the tests slower.

tomas-langer:
### Description
Resolves #8402 

Replacement for https://github.com/helidon-io/helidon/pull/8472 - with my changes (as I cannot push to original branch)

### DIFF
--- a/dbclient/jdbc/src/main/java/io/helidon/dbclient/jdbc/JdbcClient.java
+++ b/dbclient/jdbc/src/main/java/io/helidon/dbclient/jdbc/JdbcClient.java
@@ -68,9 +68,9 @@ class JdbcClient extends DbClientBase implements DbClient {
 
     @Override
     public void close() {
-        if (connectionPool instanceof CloseableJdbcConnectionPool) {
+        if (connectionPool instanceof CloseableJdbcConnectionPool cjcp) {
             try {
-                ((AutoCloseable) connectionPool).close();
+                cjcp.close();
             } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


dbclient/jdbc/src/main/java/io/helidon/dbclient/jdbc/JdbcClient.java
```diff
@@ -68,9 +68,9 @@ public JdbcClientContext context() {
 
     @Override
     public void close() {
-        if (connectionPool instanceof CloseableJdbcConnectionPool) {
+        if (connectionPool instanceof CloseableJdbcConnectionPool cjcp) {
```
bjcoombs, line 71: 👍

PR #8576, opened 2024-03-27
4.x:  threading example

barchetta:
### Description

This PR adds an example that demonstrates two uses cases for application specific executor services:

1. Using a virtual thread executor to execute multiple tasks in parallel.
2. Using a platform thread executor to execute long-running CPU intensive operations.

### Documentation

Adds an example with a README

### DIFF
--- a/examples/webserver/threads/src/main/java/io/helidon/examples/webserver/threads/ThreadService.java
+++ b/examples/webserver/threads/src/main/java/io/helidon/examples/webserver/threads/ThreadService.java
@@ -1,0 +1,199 @@
+package io.helidon.examples.webserver.threads;
+
+import java.util.ArrayList;
+import java.util.Random;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.lang.System.Logger.Level;
+import java.util.concurrent.RejectedExecutionException;
+
+import io.helidon.common.configurable.ThreadPoolSupplier;
+import io.helidon.config.Config;
+import io.helidon.http.Status;
+import io.helidon.webclient.api.HttpClientResponse;
+import io.helidon.webclient.api.WebClient;
+import io.helidon.webserver.http.HttpRules;
+import io.helidon.webserver.http.HttpService;
+import io.helidon.webserver.http.ServerRequest;
+import io.helidon.webserver.http.ServerResponse;
+
+class ThreadService implements HttpService {
+
+    private static final System.Logger LOGGER = System.getLogger(ThreadService.class.getName());
+    private static final Random rand = new Random(System.currentTimeMillis());
+
+    // ThreadPool of platform threads.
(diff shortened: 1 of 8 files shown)


examples/webserver/threads/src/main/java/io/helidon/examples/webserver/threads/ThreadService.java
```diff
@@ -0,0 +1,208 @@
...
+         */
+        ThreadPoolSupplier platformThreadPoolSupplier = ThreadPoolSupplier.builder()
+                .config(appConfig.get("application-platform-executor"))
+                .build();
+        platformExecutorService = platformThreadPoolSupplier.get();
+
+        ThreadPoolSupplier virtualThreadPoolSupplier = ThreadPoolSupplier.builder()
+                .config(appConfig.get("application-virtual-executor"))
+                .build();
+        virtualExecutorService = virtualThreadPoolSupplier.get();
```
danielkec, line 69: `virtualThreadPoolSupplier` sounds really bad

examples/webserver/threads/src/main/java/io/helidon/examples/webserver/threads/ThreadService.java
```diff
@@ -0,0 +1,208 @@
...
+                }
+            }
+
+            // All parallel calls are complete!
+            response.send(String.join(":", responses));
+        } catch (ExecutionException e) {
+            LOGGER.log(Level.ERROR, e);
+            response.status(Status.INTERNAL_SERVER_ERROR_500).send();
+        }
+    }
```
danielkec, line 156: Maybe mimicking structured concurrency constructs with existing tooling would better?

```java
        var futures = new ArrayList<CompletableFuture<String>>();
        try (var exec = Executors.newVirtualThreadPerTaskExecutor()) {
            for (int i = 0; i < count; i++) {
                futures.add(CompletableFuture.supplyAsync(() -> callRemote(rand.nextInt(5)), exec));
            }

            String result = futures.stream()
                    .map(CompletableFuture::join)
                    .collect(Collectors.joining(":"));
        }
```

examples/webserver/threads/src/main/java/io/helidon/examples/webserver/threads/ThreadService.java
barchetta: Renamed to `virtualThreadSupplier`

examples/webserver/threads/src/main/java/io/helidon/examples/webserver/threads/ThreadService.java
barchetta: I intentionally chose to limit this to `Future` and direct use of `ExecutorServer.submit()` instead of introducing a large new API (`CompletableFuture` with its 70 methods) and indirect use of the `ExecutorService`. The existing code also parallels what is shown in `computeHandler` --  and I think that symmetry is important.

romain-grecourt (approved):

PR #8578, opened 2024-03-28
Fix OTel extract when there is no current context

### LINKED ISSUE #8573: `Span.current()` does not return `Optional.empty()` if there is no current span, contrary to the documented contract (OTel)
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
The Helidon tracing API documents that `Span.current()` will return `Optional.empty()` if there is no current span.

But, for OpenTelemetry (and, therefore, Jaeger) `Span.current()` always returns a value. If there was no current span it returns an `Optional` of a no-op span (trace and span IDs = 0).

## Steps to reproduce
With the Helidion OTel dependency in the project, invoke `Span.current()` before creating and activating a span. You get an `Optional` of a valid (but no-op) span.

tjquinno:
### Description
Resolves #8573 

The Helidon `OpenTelemetryTracer` class `extract` method (which creates a `SpanContext` from headers) incorrectly returned a no-op `SpanContext` if there was no current span, rather than the documented `Optional.empty()`. 

This PR fixes that by trying to construct an OTel `Span` first from the context, and only if that yields a span go ahead and return the span context.

The PR also adds a test for this and fixes an earlier test that did not supply enough headers for OTel to fully constitute a valid  span.

### Documentation
Bug fix; no doc impact

### DIFF
--- a/tracing/opentelemetry/src/test/java/io/helidon/tracing/opentelemetry/TestSpanAndBaggage.java
+++ b/tracing/opentelemetry/src/test/java/io/helidon/tracing/opentelemetry/TestSpanAndBaggage.java
@@ -191,8 +191,6 @@ class TestSpanAndBaggage {
 
     @Test
     void testExtractWithNoCurrentSpan() {
-        final String BAGGAGE_KEY = "mykey";
-        final String BAGGAGE_VALUE = "myvalue";
         final var tracer = io.helidon.tracing.Tracer.global();
 
         HeaderProvider headers = HeaderProvider.create(Map.of("not-a-trace", List.of("1234567890123456"),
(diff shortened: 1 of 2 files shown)


tracing/opentelemetry/src/test/java/io/helidon/tracing/opentelemetry/TestSpanAndBaggage.java
```diff
@@ -186,6 +189,19 @@ void testBaggageAddedAfterActivation() {
         }
     }
 
+    @Test
+    void testExtractWithNoCurrentSpan() {
+        final String BAGGAGE_KEY = "mykey";
+        final String BAGGAGE_VALUE = "myvalue";
```
ljnelson, line 195: Not used?

tjquinno: Fixed

ljnelson (approved):

Issue #8540, opened 2024-03-22
4.x [docs] MP does not handle built-in services (e.g. health) if user has resource with path `/`


tjquinno:
## Environment Details
* Helidon Version: 4.x
*  Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
In 3.x, a user could have an MP resource with path `/` and Helidon-provided services such as `/metrics` would still work.

In 4.x , if the user has a resource with `/` Helidon incorrectly routes requests for the Helidon-provided services to the user's `/` resource.

This is a major breaking change.

## Steps to reproduce
1. In the 3.x MP QuickStart example, change the path on `GreetResource` from `/greet` to `/`, rebuild, run, and access `curl http://localhost:8080/metrics`. It works. `curl http://localhost:8080/` and `/Joe` acts as they should: greetings to "World" and "Joe" respectively.
2. In the 4.x MP QuickStart example, change the path the same way, build, and run. The greeting requests work but accessing `/metrics` gives you "Hello metrics" instead of the actual metrics response.

tjquinno:
See also #8550

There should be a nice way to resolve this by setting the following config in `microprofile-config.properties`:
```properties
server.features.observe.weight = 120
```
or, equivalently, in SE config:
```yaml
server:
  features:
    observe:
      weight: 120
```

There is a bug (see above link) currently preventing this from working in 4.x but it is being actively worked.

tjquinno:
Keeping this issue open for adding doc describing the weighting technique.

Also, the MP doc does not have an observability page and should. That'll be part of this issue as well.

PR #8580, opened 2024-03-28
Add MP observability page; discuss weight setting to resolve some routing conflicts

tjquinno:
### Description
Resolves doc portion of #8540 

This PR adds a section to the SE observability page discussing how to use feature weights to resolve possible routing surprises if the application uses path `/` or `/{name}` (which would intercept the default build-in endpoints such as `/metrics` or `/health`).

It also adds a new MP observability page and abstracts common text into a file in `includes`.

### Documentation
This is a doc update PR.


barchetta (approved):
LGTM

PR #8583, opened 2024-03-28
Properly return Optional.empty() for current span if there is no current OTel span

### LINKED ISSUE #8575: [4.x] `Span.current()` does not return `Optional.empty()` if there is no current span, contrary to the documented contract (OTel)
Backport of #8573 for Helidon 4.x

tjquinno:
### Description
Resolves #8575 

Basically a forward-port of the 3.x fix for the same problem; see the comment there for details.  https://github.com/helidon-io/helidon/pull/8574#issue-2211654801

### Documentation
But fix; no doc impact

### DIFF
--- a/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerProvider.java
+++ b/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerProvider.java
@@ -65,6 +65,13 @@ public class OpenTelemetryTracerProvider implements TracerProvider {
     }
 
     /**
+     * Creates a new provider; reserved for service loading.
+     */
+    @Deprecated
+    public OpenTelemetryTracerProvider() {
+    }
+
+    /**
      * Register global tracer.
      *
      * @param tracer global tracer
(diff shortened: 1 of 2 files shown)


tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerProvider.java
```diff
@@ -90,10 +97,17 @@ public static Tracer globalTracer() {
      * @return optional of the current span
      */
     public static Optional<Span> activeSpan() {
-        // OTel returns a no-op span if there is no current one.
-        io.opentelemetry.api.trace.Span otelSpan = io.opentelemetry.api.trace.Span.current();
+        io.opentelemetry.context.Context otelContext = io.opentelemetry.context.Context.current();
```
ljnelson, line 100: I think this entire method can be just:
```
return Optional.ofNullable(io.opentelemetry.api.trace.Span.fromContextOrNull(io.opentelemetry.context.Context.current()))
  .map(otelSpan -> HelidonOpenTelemetry.create(otelSpan, io.opentelemetry.api.baggage.Baggage.current());
```
(Some `static` `import`s would make this a little easier, maybe.) What do you think?

ljnelson (approved):

PR #8609, opened 2024-04-03
Archetype - Add intellij configuration file for MP projects

### LINKED ISSUE #8577: [4.x] Add IntelliJ run configuration in archetypes
## Environment Details
* Helidon Version: 4.0.6

----------

## Problem Description

When using the built-in main class of Helidon, in a typical class-path setup with Helidon MP, it is not obvious how to run or debug from the IDE.

Users have to create a run configuration of type application and specify the built-in main class.
We can include this file directly in the archetype, so that projects generated with the archetype can be easily run from the IDE.

`.idea/runConfigurations/quickstart_mp.xml`:
```xml
<component name="ProjectRunConfigurationManager">
  <configuration default="false" name="quickstart-mp" type="Application" factoryName="Application">
    <option name="MAIN_CLASS_NAME" value="io.helidon.Main" />
    <module name="quickstart-mp" />
    <method v="2">
      <option name="Make" enabled="true" />
    </method>
  </configuration>
</component>
```

If we keep the default path like above we would need to update the generated `.gitignore` since `.idea` is excluded.

tvallin:
### Description

Add intellij configuration file to the archetype generated project and update the `.gitignore` file.

fix  #8577

### Documentation

no impact

### DIFF
--- a/archetypes/archetypes/src/main/archetype/common/files/.gitignore.mustache
+++ b/archetypes/archetypes/src/main/archetype/common/files/.gitignore.mustache
@@ -15,7 +15,9 @@ target/
 *.rar
 
 # IntelliJ Idea
-.idea/
+{{#gitignore-intellij}}
+{{.}}
+{{/gitignore-intellij}}
 *.iws
 *.ipr
 *.iml
--- a/archetypes/archetypes/src/main/archetype/mp/common/files/.idea/runConfigurations/configuration.xml.mustache
+++ b/archetypes/archetypes/src/main/archetype/mp/common/files/.idea/runConfigurations/configuration.xml.mustache
@@ -1,0 +1,9 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="{{artifactId}}" type="Application" factoryName="Application">
+    <option name="MAIN_CLASS_NAME" value="io.helidon.Main" />
+    <module name="{{artifactId}}" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
--- a/archetypes/archetypes/src/main/archetype/mp/common/common-mp.xml
+++ b/archetypes/archetypes/src/main/archetype/mp/common/common-mp.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2021, 2023 Oracle and/or its affiliates.
+    Copyright (c) 2021, 2024 Oracle and/or its affiliates.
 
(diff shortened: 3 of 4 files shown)


archetypes/archetypes/src/main/archetype/common/files/.gitignore.mustache
```diff
@@ -15,7 +15,9 @@ target/
 *.rar
 
 # IntelliJ Idea
-.idea/
+{{#gitignore-intellij}}
```
romain-grecourt, line 18: Use `!.idea/runConfigurations` instead. See https://git-scm.com/docs/gitignore
Also, use a generic model variable for the `.gitignore`, it does not have to be specific to IntelliJ.

archetypes/archetypes/src/main/archetype/mp/common/files/.idea/runConfigurations/configuration.xml.mustache
```diff
@@ -0,0 +1,9 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="{{artifactId}}" type="Application" factoryName="Application">
+    <option name="MAIN_CLASS_NAME" value="io.helidon.Main" />
```
romain-grecourt, line 3: The class name shouldn't be hard-coded in this file. E.g. what if jpms is enabled ?

archetypes/archetypes/src/main/archetype/mp/custom/custom-mp.xml
```diff
@@ -44,6 +44,8 @@
         <model>
             <value key="helidon-test">true</value>
             <value key="mainClass" if="${jpms}">${package}.Main</value>
+            <value key="idea-main-class" if="${jpms}">${package}.Main</value>
```
romain-grecourt, line 47: Can we have a mustache if-else condition for this instead ?

tvallin: Much better, thanks !

archetypes/archetypes/src/main/archetype/mp/custom/custom-mp.xml
```diff
@@ -44,6 +44,7 @@
...
```
romain-grecourt, line 47: You should be able to just use `mainClass` right ?

tvallin: Right ! I used another mustache value because of the condition but as it is part of the template now, it is not necessary anymore. Thanks !

romain-grecourt (approved):

PR #8614, opened 2024-04-04
4.x: Fix resolved URI query params (#8566)

### LINKED ISSUE #8566: WebClient - Uri resolved without query params when path and query params are mixed.
I would expect the resolved uri to be the same in all four examples below. Instead, the query params are missing when path and query params are mixed when building requests.

## Environment Details
* Helidon Version: 4.0.6
* Helidon SE
* JDK version: 21
* OS: Mac OX x

----------

## Problem Description
See tests below. The test fails for `test1` and `test2`. Would expect the same result as for the other two tests.

## Steps to reproduce
```
private final String expected = "https://eu.example.com:443/a/b?q=one";

    @Disabled
    @Test
    void test1() {
        var uri = WebClient.create()
                .get()
                .uri("https://eu.example.com/{path}/b")
                .pathParam("path", "a")
                .queryParam("q", "one")
                .resolvedUri();
         // actual: "https://eu.example.com/a/b"
        assertEquals(expected, uri.toString());
    }

    @Disabled
    @Test
    void test2() {
        var uri = WebClient.create()
                .get()
                .uri("https://{region}.example.com/a/b")
                .pathParam("region", "eu")
                .queryParam("q", "one")
                .resolvedUri();
        // actual: "https://eu.example.com/a/b"
        assertEquals(expected, uri.toString());
    }

    @Test
    void test3() {
        var uri = WebClient.create()
                .get()
                .uri("https://eu.example.com/a/b")
                .queryParam("q", "one")
                .resolvedUri();
        assertEquals(expected, uri.toString());
    }

    @Test
    void test4() {
        var uri = WebClient.create()
                .get()
                .uri("https://{region}.example.com/a/b?q={q}")
                .pathParam("region", "eu")
                .pathParam("q", "one")
                .skipUriEncoding(true)
                .resolvedUri();
        assertEquals(expected, uri.toString());
    }
```

VerKWer:
### Description
The `ClientRequestBase` class did not properly resolve URIs when path- and query parameters were mixed. More specifically, the query parameters were stripped if the URI template was absolute.

The reason for this is that `ClientRequestBase.resolveUri` first resolves path parameters (creating a `URI` instance) and then calls `ClientRequestBase.resolve` against that `URI`. If the constructed `URI` is absolute, resolving against it clears query parameters.

Note that there are two `ClientRequestBase.resolve` methods: one for `URI`s and one for `ClientUri`s. They behave differently regarding query parameters. The former mimics what `URI.resolve` does, which returns its argument if it is absolute; in particular, it removes any existing query parameters from the `ClientUri` if the argument is absolute. The latter method, however, leaves query parameters untouched. I'm not sure this difference is intentional but since these are public methods, their semantics shouldn't be changed. I added tests for this.

Resolves #8566

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
@@ -37,6 +37,7 @@ import io.helidon.common.context.Contexts;
 import io.helidon.common.tls.Tls;
 import io.helidon.common.uri.UriEncoding;
 import io.helidon.common.uri.UriFragment;
+import io.helidon.common.uri.UriQuery;
 import io.helidon.http.ClientRequestHeaders;
 import io.helidon.http.Header;
 import io.helidon.http.HeaderNames;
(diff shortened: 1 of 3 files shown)


tomas-langer:
I will create an updated PR for 4.x and for `main`, honoring this contribution.

I am closing this PR, as it is now obsolete. I will link the updated PR as soon as done.

Issue #8591, opened 2024-03-30
FT Asynchronous should run on platform thread by default


danielkec:
## Environment Details
* Helidon Version: 4.0.6
* Helidon SE or Helidon MP: Both

----------

## Problem Description
Fault tolerance asynchronous should always run on physical thread by default.
Currently it is not possible to configure MP for platform thread usage.

Cases for offloading with `@Asynchronous`:
* **Blocking code** - offloading VT->VT is not needed
* **Obstructing/long-running code** - offloading VT->VT is not practical as carrier thread from the same pool is used
* **Pinning code** - offloading VT->VT doesn't solve the issue

More complicated cases for VT's may exist, configurable option should exist(it doesn't in 4.0.6 MP).

## Steps to reproduce
```java
    @GET
    @Path("/platform")
    @Asynchronous
    public CompletionStage<ThreadInfo> getOnPlatformThread() {
        var name = Thread.currentThread().getName();
        var virtual = Thread.currentThread().isVirtual();
        ThreadInfo threadInfo = new ThreadInfo(name, virtual);
        return CompletableFuture.completedFuture(threadInfo);
    }
```

```java
➜  offloading-mp curl -s localhost:8080/thread-name/platform | jq
{
  "name": "helidon-ft-1: async",
  "virtual": true
}
```

spericas:
I agree there should be an option for this, but I'm not sure a platform thread should be default; I see people using asynchronous in 4.x without a good understanding of what they are doing, and suddenly creating a bunch of new platform threads may not be ideal. I will take a look.

spericas:
@danielkec One possibility would be to introduce a new annotation to control the executor service to use. Something along these lines:

```java
    @Asynchronous
    @InvokedUsing(SingleThreadExecutorSupplier.class)
    CompletableFuture<String> asyncWithExecutor() {
        called.set(!Thread.currentThread().isVirtual());
        FaultToleranceTest.printStatus("AsynchronousBean::asyncWithExecutor", "success");
        return CompletableFuture.completedFuture("success");
    }

    static class SingleThreadExecutorSupplier implements Supplier<ExecutorService> {

        @Override
        public ExecutorService get() {
            return Executors.newSingleThreadExecutor();
        }
    }
```

danielkec:
[...]

PR #8616, opened 2024-04-04
Support for a new annotation @WithExecutor to use with @Asynchronous in FT

spericas:
### Description

Support for a new annotation @WithExecutor to override the default executor service to use when making an asynchronous call. Uses CDI to find a provider based on this annotation. Issue #8591.

### Documentation

Documentation will be provided in a follow-up PR.

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
@@ -243,7 +243,7 @@ public class FaultToleranceExtension implements Extension {
         if (withExecutor != null) {
             Set<Bean<?>> beans = bm.getBeans(ExecutorService.class, withExecutor);
             if (beans.isEmpty()) {
-                throw new FaultToleranceDefinitionException("Unable to resolved named executor service '"
+                throw new FaultToleranceDefinitionException("Unable to resolve named executor service '"
                         + withExecutor.value() + "' at "
                         + method.getDeclaringClass().getName() + "::"
                         + method.getName());
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
@@ -17,12 +17,14 @@
 package io.helidon.microprofile.faulttolerance;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.ExecutorService;
 import java.util.stream.Collectors;
 
 import jakarta.annotation.Priority;
@@ -33,6 +35,7 @@ import jakarta.enterprise.inject.spi.AnnotatedConstructor;
 import jakarta.enterprise.inject.spi.AnnotatedField;
 import jakarta.enterprise.inject.spi.AnnotatedMethod;
 import jakarta.enterprise.inject.spi.AnnotatedType;
+import jakarta.enterprise.inject.spi.Bean;
 import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.enterprise.inject.spi.BeforeBeanDiscovery;
 import jakarta.enterprise.inject.spi.BeforeShutdown;
@@ -50,6 +53,7 @@ import org.eclipse.microprofile.faulttolerance.CircuitBreaker;
 import org.eclipse.microprofile.faulttolerance.Fallback;
 import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.Timeout;
+import org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException;
 import org.glassfish.jersey.process.internal.RequestScope;
 
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_BEFORE;
@@ -210,8 +214,6 @@ public class FaultToleranceExtension implements Extension {
                              @Initialized(ApplicationScoped.class) Object event) {
         if (FaultToleranceMetrics.enabled()) {
             getRegisteredMethods().forEach(annotatedMethod -> {
-                final AnnotatedType<?> annotatedType = annotatedMethod.getDeclaringType();
-
                 // Metrics depending on the annotationSet present
                 if (MethodAntn.isAnnotationPresent(annotatedMethod, Retry.class, bm)) {
                     new RetryAntn(annotatedMethod).validate();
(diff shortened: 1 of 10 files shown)


microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/AsynchronousBean.java
```diff
@@ -165,4 +169,34 @@ CompletableFuture<String> asyncCompletableFutureWithFallbackFailure() {
...
+        return Executors.newFixedThreadPool(10);
+    }
+
+    /**
+     * Bad executor name, should throw a {@link FaultToleranceException}.
+     *
+     * @return A future.
+     */
+    @Asynchronous
+    @WithExecutor("does-not-exist")
```
tomas-langer, line 198: Throwing an exception at runtime is not very user friendly.
Is there a chance to validate this during startup? Or maybe have an option to fallback to the default executor if the named one is not defined?

spericas: We could validate at startup I suppose. Silently falling back to the default one does not seem like a good idea to me when the task may depend on the right executor being used.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/WithExecutor.java
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Annotation used to specify an executor for an asynchronous call. Can be used
+ * e.g. to create platform instead of virtual threads. Targets only a method.
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+@Inherited
+@Qualifier
+@InterceptorBinding
```
ljnelson, line 37: Just double-checking that you want this to be both a qualifier and an interceptor binding. (That's normally weird, though permitted.)

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
```diff
@@ -334,7 +336,17 @@ private CompletableFuture<Object> callSupplierNewThread(FtSupplier<Object> suppl
 
         // Call supplier in new thread
         ClassLoader ccl = Thread.currentThread().getContextClassLoader();
-        CompletableFuture<Object> asyncFuture = Async.create().invoke(() -> {
+        AsyncConfig.Builder asyncBuilder = AsyncConfig.builder();
+
+        // Handle a user-provided executor via @WithExecutor
+        if (introspector.hasWithExecutor()) {
+            WithExecutor withExecutor = introspector.withExecutor();
+            ExecutorService executorService = CDI.current().select(ExecutorService.class, withExecutor).get();
```
ljnelson, line 344: (Per my earlier comment: I see here you are using it as a qualifier, so indeed this confirms that `@Qualifier` is certainly needed.)

danielkec:
I like it! Can we also add a build-in executor which would be default for `Asynchronous`?

Or something like that? 
```java
@WithExecutor(FtExecutors.PLATFORM)
```
Would be great if such executor would be configurable.

danielkec:
Also shutting down all executors with our qualifier on container shutdown would be nice

spericas:
> Can we also add a build-in executor which would be default for `Asynchronous`?

 The default is now defined in `io.helidon.faulttolerance.FaultTolerance` and shared between SE and MP. Seems like the right default to me.

> 
> Or something like that?
> 
> ```java
> @WithExecutor(FtExecutors.PLATFORM)
> ```
> 
> Would be great if such executor would be configurable.

We already have `ThreadPoolSupplier` as a means to configure and create an `ExecutorService`. We could define a qualifier and add a built-in CDI provider for this. However, it's only few lines of code for anyone to create such a provider, and that could be easily documented too. Is it really worth it?

spericas:
@danielkec I updated the main test to show how to use a configurable thread pool supplier to return an executor service for async. This could be easily documented.

spericas:
After discussing the use case further with @danielkec, we agreed to proceed with a two-pronged approach: (i) an extension for the FT module to support using a different executor service (this PR!) and (ii) a new, simpler annotation in another module to only handle running arbitrary methods in platform threads. There will be a new issue created for (ii).

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
```diff
@@ -230,11 +232,25 @@ void validateAnnotations(BeanManager bm,
...
             });
         }
     }
 
+    static void validateWithExecutor(BeanManager bm, Method method) {
+        WithExecutor withExecutor = method.getAnnotation(WithExecutor.class);
+        if (withExecutor != null) {
+            Set<Bean<?>> beans = bm.getBeans(ExecutorService.class, withExecutor);
+            if (beans.isEmpty()) {
+                throw new FaultToleranceDefinitionException("Unable to resolved named executor service '"
```
danielkec, line 246:
```suggestion
                throw new FaultToleranceDefinitionException("Unable to resolve named executor service '"
```

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/AsynchronousTest.java
```diff
@@ -123,4 +122,12 @@ void testAsyncCompletableFutureWithFallbackFailure() throws Exception {
         assertThat(completableFuture.get(), is("fallback"));
         assertThat(bean.wasCalled(), is(true));
     }
+
+    @Test
+    void testAsyncWithExecutor() throws Exception {
+        assertThat(bean.wasCalled(), is(false));
+        CompletableFuture<String> future = bean.asyncWithExecutor();
+        future.get();
```
danielkec, line 130: timeout

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/AsynchronousBean.java
```diff
@@ -165,4 +172,47 @@ CompletableFuture<String> asyncCompletableFutureWithFallbackFailure() {
...
+     * @return a future
+     * @see ExecutorServiceProducer#threadPoolSupplier()
+     */
+    @Asynchronous
+    @WithExecutor("platform-thread-executor")
+    CompletableFuture<String> asyncWithExecutor() {
+        Thread thread = Thread.currentThread();
+        called.set(thread.getName().contains("platform-threads") && !thread.isVirtual());
+        FaultToleranceTest.printStatus("AsynchronousBean::asyncWithExecutor", "success");
+        return CompletableFuture.completedFuture("success");
```
danielkec, line 189: Why not return the thread ref and doing the assertion in the test method? May be more clear what is wrong if the test fails

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/FaultToleranceExtension.java
spericas: Fixed!

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/AsynchronousTest.java
spericas: All these tests need to be "fixed" in that way. The implementation "should" never block that way.

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/AsynchronousBean.java
spericas: Again, all these tests can be improved, but that's for another PR.

PR #8621, opened 2024-04-05
3.x: Upgrade oci sdk to 3.39.0

barchetta:
### Description

 Upgrade oci sdk to 3.39.0

### Documentation

No impact


ljnelson (approved):
From what I remember about OCI 3.x and its pluggable client library, this should be fine, modulo dependency convergence issues if any.

PR #8624, opened 2024-04-06
4.x: Remove unused private methods from HelloWorldAsyncResponseWithRestRequestTest

Captain1653:
### Description
I've found two unused private methods in `HelloWorldAsyncResponseWithRestRequestTest`. Maybe they can be removed, I'm not sure.

They we added in [this PR](https://github.com/helidon-io/helidon/pull/7139/files#diff-46eebbbc9c442b375c4cbd9da88720b2922b103844cac47f2b832cc43ea94fc5) .


tjquinno (approved):
Looks OK to me. I think what happened is this: The newer MP metrics spec removed the requirement for JSON output from the `/metrics` endpoint. Our original metrics work for Helidon 4.0 excluded that support at first so various tests that used to rely on JSON were revised to depend on the Prometheus/OpenMetrics output.

Once we restored the JSON output, and because it's a bit easier to validate JSON output vs. Prometheus output, we converted tests back to use JSON. I suspect that's why these stale methods remained.

Issue #8021, opened 2023-11-16
Helidon ErrorHandlers class swallows exception object if response can't be reset


ritzow:
Helidon 4.0.0[ io.helidon.webserver.http.ErrorHandlers class line 212](https://github.com/helidon-io/helidon/blob/4.0.0/webserver/webserver/src/main/java/io/helidon/webserver/http/ErrorHandlers.java#L212).

A message is logged indicating the response could not be reset, but the error which caused it is ignored. The log message shpuld probably log the exception object to be handled by the underlying logging provider.

jbescos:
Merged

PR #8634, opened 2024-04-09
4.x: ErrorHandlers class swallows exception object if response can't be reset

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/8021

### Documentation
Probably we don't want to log the exception in the logger as a warning, but I understand it could be useful to include the root exception in the cause.


romain-grecourt (commented):
Can you add a test for this?

jbescos:
> Can you add a test for this?

Done

romain-grecourt (approved):

PR #8643, opened 2024-04-09
Proposal for a new annotation @ExecuteOn 

spericas:
### Description

Proposal for a new annotation @ExecuteOn to run any CDI bean method on a separate thread (platform or virtual) and always blocking the calling thread. See tests.


microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/AsyncPlatformInterceptor.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Intercepts a call to bean method annotated by {@code @AsyncPlatform}.
+     *
+     * @param context Invocation context.
+     * @return Whatever the intercepted method returns.
+     * @throws Throwable If a problem occurs.
+     */
+    @AroundInvoke
+    public Object runWithExecutor(InvocationContext context) throws Throwable {
+        AsyncPlatform asyncPlatform = context.getMethod().getAnnotation(AsyncPlatform.class);
```
ljnelson, line 60: (This isn't how you get annotations in CDI; it may be, for example, that a portable extension has removed the annotation or otherwise altered it; this won't pick up such alterations or changes. You want to somehow get an [`Annotated`](https://jakarta.ee/specifications/cdi/4.0/apidocs/jakarta.cdi/jakarta/enterprise/inject/spi/annotated) and look at _its_ annotations instead as the source of truth. This may require some refactoring.)

spericas: Right, code updated.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/AsyncPlatform.java
```diff
@@ -0,0 +1,55 @@
...
+    @Nonbinding
+    long value() default 5000L;
+
+    /**
+     * Waiting time unit.
+     *
+     * @return waiting time unit.
+     */
+    @Nonbinding
+    TimeUnit unit() default TimeUnit.MILLISECONDS;
```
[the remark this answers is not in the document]
spericas: Roger

microprofile/cdi/src/test/java/io/helidon/microprofile/cdi/AsyncPlatformTest.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    static AtomicBoolean success = new AtomicBoolean();
+
+    static class AsyncPlatformBean {
+
+        @AsyncPlatform
+        void cpuIntensive() {
+            Thread thread = Thread.currentThread();
+            success.set(!thread.isVirtual() && thread.getName().startsWith("my-platform-thread"));
```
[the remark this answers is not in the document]
spericas: Fixed

microprofile/cdi/src/test/java/io/helidon/microprofile/cdi/OnNewThreadTest.java
```diff
@@ -0,0 +1,121 @@
...
+        @Produces
+        @Named("my-executor")
+        ExecutorService myExecutor() {
+            return Executors.newFixedThreadPool(2);
+        }
+    }
+
+    @Test
+    void cpuIntensiveTest() {
+        OnNewThreadBean bean = CDI.current().select(OnNewThreadBean.class).get();
```
tomas-langer, line 94: you could select the bean just once after starting the container, to avoid the repeated code.

microprofile/cdi/src/test/java/io/helidon/microprofile/cdi/OnNewThreadTest.java
```diff
@@ -0,0 +1,121 @@
...
+        assertThat(bean.cpuIntensive(), is(true));
```
tomas-langer, line 95: A small issue for readability - maybe return a record(String name, boolean virtual), and assert against the expected value, so if a failure happens, you could see what thread was used to invoke the method, rather than getting just a `boolean`

microprofile/cdi/src/test/resources/application.yaml
```diff
@@ -0,0 +1,23 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+helidon:
+  async-platform:
+    executor-service:
```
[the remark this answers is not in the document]
spericas: Switched to "mp"

danielkec (approved):

Issue #7920, opened 2023-10-31
Add text block support to `@AddConfig`


romain-grecourt:
## Environment Details
* Helidon Version: 3.x 4.x
* Helidon MP
----------

## Problem Description

Add support for something similar to `@AddConfig` that uses text block.

E.g. using `@AddConfig`
```java
@HelidonTest
@AddConfig(key = "javax.sql.DataSource.example.dataSourceClassName", value = "org.h2.jdbcx.JdbcDataSource")
@AddConfig(key = "javax.sql.DataSource.example.dataSource.url", value = "jdbc:h2:mem:sample")
@AddConfig(key = "javax.sql.DataSource.example.dataSource.user", value = "sa")
@AddConfig(key = "javax.sql.DataSource.example.dataSource.password", value = "")
class MyTest {
    //...
}
```

E.g. using text blocks:
```java
@HelidonTest
@AddConfig(type = "properties", fragments = """
    javax.sql.DataSource.example.dataSourceClassName = org.h2.jdbcx.JdbcDataSource
    javax.sql.DataSource.example.dataSource.url = jdbc:h2:mem:sample
    javax.sql.DataSource.example.dataSource.user = sa
    javax.sql.DataSource.example.dataSource.password = 
    """)
class MyTest {
    //...
}
```

The key of this feature request is to support text blocks to express test configuration in a test classes. We may decide to update `@AddConfig` or create a new annotation.

dansiviter:
YAML support would be great too:

```java
@AddConfig(type = "yaml", fragments = """
    javax.sql.DataSource.example:
      dataSourceClassName: org.h2.jdbcx.JdbcDataSource
      dataSource:
        url: jdbc:h2:mem:sample
        user: sa
        password: ""
    """
```

jbescos:
Is this worth to do it?, there is already:

`@Configuration(configSources = "yourProperties.properties")`

dansiviter:
@jbescos As this is for tests adding indirection to a property file just makes maintenance and comprehension difficult.

danielkec:
And it will be cool!

jbescos:
My idea is to create a new annotation, because I think adding this into an existing one could be confusing.
`@ConfigBlob` or `@ConfigurationBlob`
```
String: type // Properties, YAML, etc
String: content // The text
```

jbescos:
Already in place

PR #8655, opened 2024-04-12
4.x: Add text block support

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/7920

### Documentation

None

### DIFF
--- a/microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/ConfigBlob.java
+++ b/microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/ConfigBlob.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.microprofile.testing.junit5;
+
+import java.io.IOException;
+import java.io.StringReader;
+import java.io.UncheckedIOException;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import java.util.Objects;
+import java.util.Properties;
+
+import io.helidon.config.mp.MpConfigSources;
+import io.helidon.config.yaml.mp.YamlMpConfigSource;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+/**
+ * Defines the configuration as a String in {@link #content()} for the
+ * given type.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Inherited
+public @interface ConfigBlob {
+
+    /**
+     * Specifies the format type of the {@link #content()}.
+     *
+     * It defaults to {@link Type#PROPERTIES}.
+     *
+     * @return the supported type
+     */
+    Type type() default Type.PROPERTIES;
+
+    /**
+     * Configuration content.
+     *
+     * @return String with content.
+     */
+    String content();
+
+    /**
+     * Different formats of the configuration content.
+     */
+    enum Type {
+
+        /**
+         * Properties format.
+         */
+        PROPERTIES {
+            @Override
+            protected ConfigSource configSource(String content) {
+                Objects.requireNonNull(content);
+                Properties p = new Properties();
+                try {
+                    p.load(new StringReader(content));
+                } catch (IOException e) {
+                    throw new UncheckedIOException(e);
+                }
+                return MpConfigSources.create("PropertiesConfigBlob", p);
+            }
+        },
+        /**
+         * YAML format.
+         */
+        YAML {
+            @Override
+            protected ConfigSource configSource(String content) {
+                Objects.requireNonNull(content);
+                return YamlMpConfigSource.create("YamlConfigBlob", new StringReader(content));
+            }
+        };
+
+        /**
+         * Create the ConfigSource given the content.
+         *
+         * @param content
+         * @return the configuration
+         */
+        protected abstract ConfigSource configSource(String content);
+    }
+}
--- a/microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
+++ b/microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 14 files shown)


microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
```diff
@@ -0,0 +1,100 @@
...
+     * @return the supported type
+     */
+    Type type() default Type.PROPERTIES;
+
+    /**
+     * Configuration content.
+     *
+     * @return String with content.
+     */
+    String content();
```
danielkec, line 58: Why not call the field `value`?

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/ConfigBlob.java
```diff
@@ -0,0 +1,100 @@
...
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+/**
+ * Defines the configuration as a String in {@link #content()} for the
+ * given type.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Inherited
+public @interface ConfigBlob {
```
romain-grecourt, line 42: I'm not convinced with that name as "blob" usually relates to binary stuff.

Can we come-up with something better ?
```java
@ConfigBlock
@ConfigChunk
@ConfigFragment
```

---

Also, should we use the prefix `Add` to be consistent with the current `AddConfig` annotation ?

```java
@AddConfig(key = "foo", value = "bar")
@ConfigBlock("""
    foo=bar
    bar=foo
""")
```

Compared to:

```java
@AddConfig(key = "foo", value = "bar")
@AddConfigBlock("""
    foo=bar
    bar=foo
""")
```

Compared to (re-using the `@AddConfig` annotation):

```java
@AddConfig(key = "foo", value = "bar")
@AddConfig(block = """
    foo=bar
    bar=foo
""")
```

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * Different formats of the configuration content.
+     */
+    enum Type {
```
romain-grecourt, line 63: This effectively hard-codes the config sources that are usable.
Shouldn't we instead support a model that can lead to any config source implementation on the classpath ?

E.g. `io.helidon.config.mp.spi.MpMetaConfigProvider` has a method `supportedTypes` that matches providers to "types".

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
jbescos: Value is more confusing to me because we use to see it together with the pair key/value. I'd expect that, if there is a value, there is also a key.

Content is more general in my opinion.

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/ConfigBlob.java
jbescos: `@AddConfigBlock` is probably better, because with the IDE you can autocomplete, and it will help users to know what do you have available just writting `@Add`

Regarding re-using `@AddConfig` I don't really support this. There are some combinations that makes no sense, for example when you specify `key`, `value` and `type`. It could confuse the user.

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
jbescos: I have taken a look into this. The problem is that what you have in classpath/modulepath will not necessarily match with the inputs of the user in the test.

You could have loaded YamlMetaConfigProvider and still the user sets in the test the next (in properties format):
```java
@AddConfigBlock( """
    foo=bar
    bar=foo
""")
```

I agree that we should have a common place for the `type`. For example in case in the future we add a new config support, like JSON or XML, we would need to update also the enum.

I can make the type as string. In this way, we can manage the string and hope that at some point we will have a common point to do this conversion.

romain-grecourt: The snippet you've shown above does not specify the type. It should specify the type with something that we can match against loaded providers. I mentioned `MpMetaConfigProvider` as an example of a similar mechanism.

We need a way to create a config source based on a type, `MpMetaConfigProvider` demonstrates something similar to what we need but does not seem directly usable.

Either we need a new SPI, or update the MpConfig API to add the capability to specify config sources as a pairs of type and reader.

JSON is already supported, see `hocon-mp`.

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/ConfigBlob.java
romain-grecourt: You are right that re-using the existing annotation could be confusing.

However with annotations, users have to be aware of them, they cannot be found naturally by navigating the code.
One could argue that having a new annotation isn't necessarily better for end users.

I don't care that much though :) As long as we pick a good name.

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
romain-grecourt: Using `value` over `content` would mean we could do this:
```java
// this would use the default type
@AddConfigBlock("""
    foo=bar
""")
```

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/ConfigBlob.java
jbescos: OK, so the task is going to be to create one SPI that will contain:
```
supportedTypes(): Collection<String>
create(Reader): ConfigSource

```
Then configure it in every module under /META-INF/services, iterate the ServiceLoaders to find which type is supported and invoke the create(Reader).

This can be done in a separate issue. In this task I can just set the `type` as string. Later we can use the type to do the process above.

tomas-langer:
This looks really useful!

jbescos:
Examples fails missing artifact:

```
 Error:  Failed to execute goal org.xolstice.maven.plugins:protobuf-maven-plugin:0.6.1:compile (default) on project helidon-examples-webserver-protocols: Unable to resolve artifact: Missing:
Error:  ----------
Error:  1) com.google.protobuf:protoc:exe:osx-aarch_64:3.5.1-1
Error:  
Error:    Try downloading the file manually from the project website.
Error:  
Error:    Then, install it using the command: 
Error:        mvn install:install-file -DgroupId=com.google.protobuf -DartifactId=protoc -Dversion=3.5.1-1 -Dclassifier=osx-aarch_64 -Dpackaging=exe -Dfile=/path/to/file
Error:  
Error:    Alternatively, if you host your own repository you can deploy the file there: 
Error:        mvn deploy:deploy-file -DgroupId=com.google.protobuf -DartifactId=protoc -Dversion=3.5.1-1 -Dclassifier=osx-aarch_64 -Dpackaging=exe -Dfile=/path/to/file -Durl=[url] -DrepositoryId=[id]
Error:  
Error:    Path to dependency: 
Error:    	1) io.helidon.examples.webserver:helidon-examples-webserver-protocols:jar:4.0.0-SNAPSHOT
Error:    	2) com.google.protobuf:protoc:exe:osx-aarch_64:3.5.1-1
```

romain-grecourt (approved):
LGTM, please file an issue to track the remaining work to support all config sources.

jbescos:
> LGTM, please file an issue to track the remaining work to support all config sources.

Thanks, here it is: https://github.com/helidon-io/helidon/pull/8655#pullrequestreview-2034162482

Issue #7737, opened 2023-10-05
Use LevelChangePropagator in logback examples


romain-grecourt:
## Environment Details
* Helidon Version: 4.0.0-RC1

----------

## Problem Description

When using [jul-to-slf4j](https://www.slf4j.org/legacy.html#jul-to-slf4jBridge) bridge with logback, the performance cost can be mitigated by using  [LevelChangePropagator](https://logback.qos.ch/manual/configuration.html#LevelChangePropagator).

```xml
<configuration debug="true">
  <contextListener class="ch.qos.logback.classic.jul.LevelChangePropagator">
    <resetJUL>true</resetJUL>
  </contextListener>
   <!-- rest of the configuration file .... -->
</configuration>
```

---

- We should update `examples/logging/slf4j` to use it so that we demonstrate a best practice.
- We should update the `logback.xml` snippet in the blog [article](https://medium.com/helidon/helidon-logging-and-mdc-5de272cf085d)
- We should be setup the bridge using `logging.properties` instead of code as that is a "drop-in" and can be used regardless of the Helidon flavor:

`logging.properties`:
```properties
handlers=org.slf4j.bridge.SLF4JBridgeHandler
```

PR #8656, opened 2024-04-12
4.x: Use LevelChangePropagator in examples/logging/slf4j (#7737)

Captain1653:
### Description
References #7737

Thoughts:
1. I'm not sure, but it can be something like that.  I've added `org.slf4j:jul-to-slf4j`, because it wasn't in the dependencies. I think, that it should be done.
2. It's little hard to understand, how to confirm, that the  performance cost was mitigated :). 

Questions:
1. I didn't add [debug="true"](https://logback.qos.ch/manual/configuration.html#configDebug), because it wasn't added before. Should I add it?
2. Should I do the same things for  `logback-aot` module? I think yes, but I can be wrong.


romain-grecourt (commented):
I'd want `debug="true"` as that a short-hand for a best practice.
`logback-aot` should also be updated however you will need to triple check that it works well with native image.

Captain1653:
I've found the "issue" with checkstyle.

Locally it doesn't check the rule `NewlineAtEndOfFile`, but CI does. Maybe I missed something.

CI

locally

I [checked the `checkstyle.xml`](https://github.com/helidon-io/helidon/blob/610f90b2d954a8c95775e861c2970f263c4357b4/etc/checkstyle.xml#L88)  and there is this rule.

romain-grecourt (approved):

PR #8662, opened 2024-04-16
Use delegation instead of inheritance from BufferedOutputStream

spericas:
### Description

Uses delegation instead of inheritance from BufferedOutputStream to ensure use of VT-friendly locks in the JDK code. Avoids thread pinning due to synchronized blocks in the JDK.

### Documentation

None


tjquinno (approved):

vasanth-bhat:
The original hung issue due to  this pinning is no longer reproduced in Helidon 4.0.8

Issue #7694, opened 2023-10-02
Ability to Inject MockBeans in Helidon


abhinavms:
## Problem
Helidon currently lacks an equivalent to Spring Boot's `@MockBean` annotation, which allows for the injection of mock beans into the application context, ensuring that mock dependencies are used when running the application. This feature is essential for improving integration tests in Helidon.

## Sample Usecase: Pact (Contract Testing)
One critical use case for this enhancement is contract testing, such as Pact testing, where microservices need to be tested in isolation to ensure their interactions conform to predefined contracts.

Suppose there are two microservices, X and Y, where X calls Y. Consumer tests are created by microservice X, specifying the endpoint, request, and expected response. For such tests to be effective, we need the ability to inject mock dependencies into the running server.

### Sample consumer test
```
{
  "consumer": {
    "name": "X"
  },
  "interactions": [
    {
      "description": "Check if the product exists",
      "providerStates": [
        {
          "name": "Product X010000021 exists"
        }
      ],
      "request": {
        "body": {
          "instance": "v2211",
          "pushToURL": "https://XXXXXX.com"
        },
        "headers": {
          "Content-Type": "application/json"
        },
        "method": "POST",
        "path": "/v1/product/check"
      },
      "response": {
        "status": 200
      }
    }
  ],
  "provider": {
    "name": "Y"
  }
}
```

This consumer test is run at the microservice Y, to valid if the request returns the expected response (In the above example POST to /v1/product/check with the provided request should return a status 200)

### Spring Boot Example
```
@ExtendWith(SpringExtension.class)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@Provider("Y")
public class PricingServiceProviderPactTest {

  @MockBean
  private ProductClient productClient;  // Replaces the bean with a mock in the application context

  @LocalServerPort
  private int port;

  @BeforeEach
  void before(PactVerificationContext context) {
    context.setTarget(new HttpTestTarget("localhost", port, "/"));
  }

  @TestTemplate
  @ExtendWith(PactVerificationInvocationContextProvider.class)
  void pactVerificationTestTemplate(PactVerificationContext context, HttpRequest request) {
    context.verifyInteraction();
  }

  @State("Product X010000021 exists")
  public void setupProductX010000021() {
    when(productClient.fetch((Set<String>) argThat(contains("X010000021")), any())).thenReturn(product);
  }
}
```
[...]

Issue #8339, opened 2024-02-06
4.x: Document CDI based mocking for unit tests


romain-grecourt:
## Environment Details
* Helidon Version: 4.0.4
* Helidon MP
----------

## Problem Description

Mockito's `@Mock` does not work with `@HelidonTest`.

---

Consider the following JAXRS resource and CDI bean:

```java
@ApplicationScoped
public class FooService {

    public String getFoo() {
        return "foo";
    }
}

@Path("/foo")
public class FooResource {

    @Inject
    private FooService fooService;

    @GET
    public String getFoo() {
        return fooService.getFoo();
    }
}
```

---

A test that combines both `@HelidonTest` and `@Mock` does not work:

```java
@HelidonTest
@ExtendWith(MockitoExtension.class)
class AcmeTest {
 
    @Inject
    private WebTarget target;

    @Mock
    private AcmeService acmeService;

    @Test
    void testFoo() {
        when(acmeService.getFoo()).thenReturn("bar");

        Response response = target.path("/foo").request().get();

        assertThat(response.getStatus(), is(200));
        assertThat(response.readEntity(String.class), is("bar"));
     }
}
```

----

### Current solution

Instead, we need to use CDI to provide an alternative. This solution works with Helidon 2.x, 3.x and 4.x.

```java
@HelidonTest
@Priority(1) // required for @Alternative
class FooTest {

    @Inject
    private WebTarget target;

    private FooService fooService;

    @BeforeEach
    void initMock() {
        // set fooService to a new mock before each test
        // use real methods for default answers
        fooService = Mockito.mock(FooService.class, InvocationOnMock::callRealMethod);
    }

    @Produces
    @Alternative
    FooService mockFooService() {
        // This makes FooResource inject our mock instead of the default singleton
        return fooService;
    }

    @Test
    void testMockedFoo() {
        // test the modified behavior
        when(fooService.getFoo()).thenReturn("bar");

        Response response = target.path("/foo").request().get();

        assertThat(response.getStatus(), is(200));
        assertThat(response.readEntity(String.class), is("bar"));
    }

    @Test
    void testFoo() {
        // test the non modified behavior
        Response response = target.path("/foo").request().get();

        assertThat(response.getStatus(), is(200));
        assertThat(response.readEntity(String.class), is("foo"));
    }
}
```

pom.xml:
```xml
<dependency>
    <groupId>org.mockito</groupId>
    <artifactId>mockito-core</artifactId>
    <version>5.8.0</version>
[...]

PR #8338, opened 2024-02-06
[4.x] Improve examples test coverage

tvallin:
### Description

fix #8337

Introduce integration test using test containers :
 - Database example is tested with a database ran in a docker container.
   - H2
   - MySql
   - Oracle-XE
 - Tracing example is tested with a tracing service ran in docker container.

### Documentation

no impact

PR #8674, opened 2024-04-18
4.x: Ability to Inject MockBeans in Helidon #7694

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/7694

### Documentation

This feature provides a simple mechanism for replacing CDI bean with mock instances created with Mockito.
It is similar to Spring's `@MockBean`, in fact it uses the same name.

The introduced API consist of a new Maven module `io.helidon.microprofile.testing:helidon-microprofile-testing-mocking` that provides one annotation `@MockBean`.

The annotation is designed to be used on fields of test classes annotated with `@HelidonTest`, the implementation relies only on CDI thus it works with either JUnit or TestNG.

The annotation has a parameter "answer" that can be used to set the default answer for the mocked beans.

See a snippet below:
```java
@HelidonTest
@AddBean(TestMockBeanAnswer.Resource.class)
@AddBean(TestMockBeanAnswer.Service.class)
class TestMockBeanAnswer {

    @MockBean(answer = Answers.CALLS_REAL_METHODS)
    private Service service;
    @Inject
    private WebTarget target;

    @Test
    void injectionTest() {
        String response = target.path("/test").request().get(String.class);
        assertThat(response, is("Not Mocked"));
        Mockito.when(service.test()).thenReturn("Mocked");
        response = target.path("/test").request().get(String.class);
        assertThat(response, is("Mocked"));
    }

    @Path("/test")
    public static class Resource {

        @Inject
        private Service service;

        @GET
        public String test() {
            return service.test();
        }
    }

    static class Service {

        String test() {
            return "Not Mocked";
        }

    }
}
```

The mocked beans of a test class can be customized with a CDI producer method.
E.g.

```java
@Produces
MockSettings mockSettings() {
    return Mockito.withSettings().defaultAnswer(Answers.RETURNS_DEFAULTS);
}
```


jbescos:
Setting it as a draft for now. I have a couple of CDI questions:

Question 1:

Currently `@MockTest` requires also` @Inject`:
```
    @Inject
    @MockBean
    private Service service;
```
Is there any way to not require `@Inject`?.

Question 2:

It does not work if the bean already exists in CDI. For example, this will not work because the service will be defined twice:
```
@HelidonTest
@AddBean(Service.class)
public class TestMockBean {

    @Inject
    @MockBean
    private Service service;
```
I didn't find how to replace the existing bean defined by the mocked one.

romain-grecourt:
> I didn't find how to replace the existing bean defined by the mocked one.
I'm not a CDI expert, but can you set the mocked bean as an alternative ? That's how the current solution (#8339) works.

jbescos:
> > I didn't find how to replace the existing bean defined by the mocked one.
> 
> I'm not a CDI expert, but can you set the mocked bean as an alternative ? That's how the current solution (#8339) works.

I think that requires a producer, and I cannot use a producer because you need to define the type. I will try it further.

jbescos:
For TestNG it seems we cannot use the @Inject in constructor, so the `@MockBean` applies only in fields.

For example, this simple test:
@HelidonTest
```java
class TestMockBeanField {

    private final WebTarget target;

    @Inject
    public TestMockBeanField(WebTarget target) {
        this.target = target;
    }

    @Test
    void test() {
    }
}
```

It fails with:

```
org.testng.TestNGException: 
An error occurred while instantiating class io.helidon.microprofile.tests.testing.testng.TestMockBeanField. Check to make sure it can be instantiated/accessed.
	at org.testng.internal.objects.SimpleObjectDispenser.createInstance(SimpleObjectDispenser.java:113)
	at org.testng.internal.objects.SimpleObjectDispenser.dispense(SimpleObjectDispenser.java:40)
	at org.testng.internal.objects.GuiceBasedObjectDispenser.dispense(GuiceBasedObjectDispenser.java:28)
	at org.testng.internal.ClassImpl.getDefaultInstance(ClassImpl.java:106)
	at org.testng.internal.ClassImpl.getInstances(ClassImpl.java:136)
	at org.testng.TestClass.getInstances(TestClass.java:129)
	at org.testng.TestClass.initTestClassesAndInstances(TestClass.java:109)
	at org.testng.TestClass.init(TestClass.java:101)
	at org.testng.TestClass.<init>(TestClass.java:66)
	at org.testng.TestRunner.initMethods(TestRunner.java:483)
	at org.testng.TestRunner.init(TestRunner.java:356)
	at org.testng.TestRunner.init(TestRunner.java:309)
	at org.testng.TestRunner.<init>(TestRunner.java:184)
	at org.testng.SuiteRunner$DefaultTestRunnerFactory.newTestRunner(SuiteRunner.java:652)
	at org.testng.SuiteRunner.init(SuiteRunner.java:224)
	at org.testng.SuiteRunner.<init>(SuiteRunner.java:116)
	at org.testng.TestNG.createSuiteRunner(TestNG.java:1375)
	at org.testng.TestNG.createSuiteRunners(TestNG.java:1349)
	at org.testng.TestNG.runSuitesLocally(TestNG.java:1191)
	at org.testng.TestNG.runSuites(TestNG.java:1114)
	at org.testng.TestNG.run(TestNG.java:1082)
	at org.apache.maven.surefire.testng.TestNGExecutor.run(TestNGExecutor.java:155)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.executeSingleClass(TestNGDirectoryTestSuite.java:102)
	at org.apache.maven.surefire.testng.TestNGDirectoryTestSuite.execute(TestNGDirectoryTestSuite.java:91)
	at org.apache.maven.surefire.testng.TestNGProvider.invoke(TestNGProvider.java:137)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:385)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:162)
	at org.apache.maven.surefire.booter.ForkedBooter.run(ForkedBooter.java:507)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:495)
```

microprofile/tests/testing/junit5/src/test/java/io/helidon/microprofile/tests/testing/junit5/TestMockBeanField.java
```diff
@@ -0,0 +1,97 @@
...
+        Mockito.when(service.test()).thenReturn("Mocked");
+        response = target.path("/test").request().get(String.class);
+        assertThat(response, is("Mocked"));
+        Mockito.when(otherService.test()).thenReturn("Mocked");
+        assertThat(otherService.test(), is("Mocked"));
+    }
+
+    @Produces
+    MockSettings mockSettings() {
+        return Mockito.withSettings().defaultAnswer(Answers.CALLS_REAL_METHODS);
```
romain-grecourt, line 67: How about we add an annotation attribute for the default answer ?

microprofile/testing/junit5/pom.xml
```diff
@@ -54,6 +54,11 @@
             <artifactId>junit-jupiter-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
romain-grecourt, line 57: Can you consult @tomas-langer about adding this dependency.

romain-grecourt (commented):
Looks nice so far.
We need to update the docs as well.

microprofile/tests/testing/junit5/src/test/java/io/helidon/microprofile/tests/testing/junit5/TestMockBeanField.java
jbescos: I didn't go for it because answer is only one of many Mockito settings. I thought if I am adding the answer, I would need to cover every other setting with annotation attributes, otherwise people will start asking why answer is supported in annotation, but other things are not.

microprofile/testing/junit5/pom.xml
jbescos: It is necessary to compile `helidon-microprofile-testing-junit5`. Users don't need to include mockito dependencies as long as they don't add `@MockBean`. In case they add `@MockBean` without having mockito dependency they will have a runtime error ClassNotFoundException.

I would personally remove every provided scope in the pom.

romain-grecourt: `HelidonJunitExtension` imports mockito so it's effectively required to have it on the test class-path.
I'm basically asking if we should modularize this. 

@tomas-langer please chime in.

microprofile/tests/testing/junit5/src/test/java/io/helidon/microprofile/tests/testing/junit5/TestMockBeanField.java
romain-grecourt: The default answer is one of the (if not THE) main setting, and `Answers` is an enum so it's compatible with an attribute.

The whole point of `@MockBean` is to avoid forcing users into CDI based mechanism, something as basic as the default answer should be supported without having to use `@Produces`.

microprofile/testing/junit5/pom.xml
tomas-langer: Import does not matter, code path that uses the imported type matters. I will check the code.

tomas-langer: If the dependency needs to be added by the user (i.e. we expect the user already has it on their classpath), it should be `provided`. If it is not required to run this code, it should be `<optional>true</optional>` instead.

microprofile/testing/junit5/src/main/java/io/helidon/microprofile/testing/junit5/HelidonJunitExtension.java
```diff
@@ -569,6 +613,25 @@ void registerOtherBeans(@Observes AfterBeanDiscovery event) {
...
                     .createWith(context -> getWebTarget(client, "@default"));
 
+            // Register all mocks
+            mocks.forEach(type -> {
+                event.addBean()
+                    .addType(type)
+                    .scope(ApplicationScoped.class)
+                    .alternative(true)
+                    .createWith(inst -> {
+                        Set<Bean<?>> beans = beanManager.getBeans(MockSettings.class);
```
tomas-langer, line 623: This is the code that would fail at runtime, if you use `@MockBean` and you do not have mockito on the classpath. We should never do this.
So either we need extensibility and have a mockito support for `MockBean`, or fallback to some other approach (such as creating a no-op proxy) if it is not available.

microprofile/testing/testng/pom.xml
```diff
@@ -44,6 +44,11 @@
             <artifactId>helidon-microprofile-cdi</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>provided</scope>
```
tomas-langer, line 50: Again, provided scope is not good here.
We expect TestNG (in this case), and Helidon as provided, as these must be present to even use this module, and users add them to their projects for testing.
Mockito is not an auto-include for testing, so we should not do this.

microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/HelidonTestNgListener.java
```diff
@@ -653,4 +704,15 @@ public Class<? extends Extension> value() {
         }
     }
 
+    /**
+     * Supports inline instantiation of the {@link Inject} annotation.
+     */
+    private static final class Literal extends AnnotationLiteral<Inject> implements Inject {
```
tomas-langer, line 710: This type already exists:
`jakarta.enterprise.inject.literal.InjectLiteral` there should be no need to create it here or in Junit extension, as we expect CDI to be on classpath (as otherwise this module has nothing to do)

microprofile/testing/junit5/pom.xml
romain-grecourt: I.e. we don't have to modularize, we only need to ensure that the normal code path doesn't fail because of the absence of Mockito.

tomas-langer (commented):
This looks like a great feature, but the hard dependency on `mockito` is something we probably do not desire.
It seems to me that the whole feature is implemented as a CDI extension. Can we maybe create a new module (depending on mockito), that would add the CDI extension and the annotation, and that could be used with both JUnit5 and testNG extension, because it is "just" a CDI portable extension?
i.e. `helidon-microprofile-testing-mock-beans`?

dependencies/pom.xml
```diff
@@ -501,6 +502,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>net.bytebuddy</groupId>
+                <artifactId>byte-buddy</artifactId>
+                <version>${version.lib.bytebuddy}</version>
+            </dependency>
```
jbescos, line 509: @barchetta to review this

microprofile/tests/testing/junit5/src/test/java/io/helidon/microprofile/tests/testing/junit5/TestMockBeanField.java
jbescos: Done

microprofile/testing/mock-beans/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+    <artifactId>helidon-microprofile-testing-mock-beans</artifactId>
+    <name>Helidon Microprofile Testing Mock Beans</name>
+
+    <description>
+        Integration with Mock Beans to support tests with CDI injection
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.microprofile.cdi</groupId>
```
tomas-langer, line 38: This should depend on CDI API only (not on Helidon MP, unless you actually use it).
Also this must be a `provided` dependency, as the application must already have it (otherwise there would be no beans to mock)

microprofile/testing/mock-beans/src/main/java/module-info.java
```diff
@@ -0,0 +1,32 @@
...
+module io.helidon.microprofile.testing.mockbeans {
+
+    requires jakarta.inject;
+    requires org.mockito;
+
+    requires transitive jakarta.cdi;
+
+    exports io.helidon.microprofile.testing.mockbeans;
+
+    provides jakarta.enterprise.inject.spi.Extension with MockBeansCdiExtension;
```
tomas-langer, line 31: please use fully qualified class name for `with`

tomas-langer (commented):
Looks good now, leaving to Romain to review.

tomas-langer (commented):
Looks good now, let's wait for library approval.

dependencies/pom.xml
barchetta: Do we use this dependency directly (I do not see where we do) or are we managing the version of a transitive dependency? If the later, is this for dependency convergence? Forced upgrade? I don't see an issue here, I just want to understand why. Thx.

microprofile/testing/mock-beans/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>jakarta.enterprise</groupId>
+            <artifactId>jakarta.enterprise.cdi-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+        </dependency>
```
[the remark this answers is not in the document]
romain-grecourt: We still expect direct mockito usage, so having it as provided can work. However we do still want to manage the version for convenience.

microprofile/testing/mock-beans/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.helidon.microprofile.testing</groupId>
+        <artifactId>helidon-microprofile-testing-project</artifactId>
+        <version>4.0.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>helidon-microprofile-testing-mock-beans</artifactId>
```
romain-grecourt, line 29: `mock-beans` feels too specific, how about `mocking` instead ?

I.e
- gav: `io.helidon.microprofile.testing:helidon-microprofile-testing-mocking`
- package: `io.helidon.microprofile.testing.mocking`

dependencies/pom.xml
tomas-langer: Dependency convergence - mockito and hibernate.

microprofile/testing/mock-beans/pom.xml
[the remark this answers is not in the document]
tomas-langer: Making this a `provided` dependency would not be the ideal solution I think. Customer can still version managed `mockito-core` to any compatible version they need.
My reasoning - Junit and Helidon are expected dependencies, as you are writing tests for a Helidon application.
Mockito is not "auto-include" for tests. 

So we can use `provided`, but it may confuse the users, as the feature would fail to compile unless they explicitly add `mockito` to their project as well.

microprofile/testing/mock-beans/pom.xml
tomas-langer: The `mock-beans` seems very specific, agreed. Maybe this deserves one more cycle to change the name.

microprofile/testing/mock-beans/pom.xml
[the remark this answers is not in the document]
romain-grecourt: I'm assuming that when using `@MockBean` users also always use Mockito explicitly.

E.g.
```java
import io.helidon.microprofile.testing.junit5.HelidonTest;
import io.helidon.microprofile.testing.mockbeans.MockBean;

import org.mockito.Mockito;

@HelidonTest
class MyTest {

    @MockBean
    private MyService service;

    @Inject
    private WebTarget target;

    @Test
    void testMyService() {
        Mockito.when(service.test()).thenReturn("Mocked");
        response = target.path("/test").request().get(String.class);
        assertThat(response, is("Mocked"));
    }
}
```

One may not actually not use Mockito directly (even though it doesn't make sense) in which case you'd have a runtime error. This is probably a good reason not to use `provided`.

Otherwise, users would have to resolve the `org.mockito.Mockito` import and would naturally add the mockito Maven dependency.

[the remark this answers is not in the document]
jbescos: To me it looks equally confusing that Helidon will not bring automatically junit or mockito as a transitive dependency. So it is consistent that if junit is not coming as a transitive dependency, then mockito is also not coming as a transitive dependency.

I am setting it as provided.

microprofile/testing/mock-beans/pom.xml
jbescos: I agree. Done.

barchetta (dismissed):
Dependencies look good to me. Thanks for all the work on this!

romain-grecourt (commented):
Looks good, please update the PR description to fill-in the documentation section:
- describe the feature (what it solves)
- describe the API (annotation + new GAV)
- add sample snippet

Let's also file an issue to create the actual docs for this feature.

jbescos:
> Looks good, please update the PR description to fill-in the documentation section:
> 
>     * describe the feature (what it solves)
> 
>     * describe the API (annotation + new GAV)
> 
>     * add sample snippet
> 
> 
> Let's also file an issue to create the actual docs for this feature.

Do we re-use this issue?: https://github.com/helidon-io/helidon/issues/8339

romain-grecourt:
> Do we re-use this issue?: #8339

Let's file a separate issue as #8339 has more to it than strictly document this new feature.
If we want to proceed with #8338 we have to figure out where it fits in the docs.

> I didn't work in any documentation issue before, is there any guide?

No, however  adding one more document is relatively easy as there is a lot of precedence.

---

The bullet items I provided was guidance not a template. It should read like a short documentation of the feature not like a form !

jbescos:
This is the new issue:
https://github.com/helidon-io/helidon/issues/8772

Are there more changes required in this PR or can it be approved/merged. I guess the documentation changes will be submitted in the new issue I created.

I still don't have any idea where/how should I document this. I will check it with @tvallin because he is in the same time zone than me.

romain-grecourt (approved):
LGTM, nice work.
I've updated the description.

PR #8707, opened 2024-04-29
[4.x] LazyConfigSource is now queried when an unknown node is requested

### LINKED ISSUE #7745: Custom Helidon SE LazyConfigSource can't resolve a non-existent key when config.get("non-existent-key") is called

## Environment Details
* Helidon Version: 3
* Helidon SE
* JDK version: 18
* OS: Mac

----------

## Problem Description
With a custom ConfigSource that implements the LazyConfigSource interface is used, config.get("non-existent-key") errors out with the exception "io.helidon.config.MissingValueException: Requested value for configuration key 'non.existent.key' is not present in the configuration" when we expect the method LazyConfigSource.node() be called to resolve the given key to an expected value.

Yes

io.helidon.config.MissingValueException: Requested value for configuration key 'non.existent.key' is not present in the configuration.
	at io.helidon.config.MissingValueException.create(MissingValueException.java:44)
	at io.helidon.config.ConfigValue.lambda$get$0(ConfigValue.java:109)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.helidon.config.ConfigValue.get(ConfigValue.java:109)
	at io.helidon.config.ConfigValues$GenericConfigValueImpl.lambda$supplier$0(ConfigValues.java:241)
	at io.helidon.examples.quickstart.se.ConfigTest.main(ConfigTest.java:58)

## Steps to reproduce
Please see the test code below

package io.helidon.examples.quickstart.se;

import io.helidon.config.Config;
import io.helidon.config.ConfigSources;
import io.helidon.config.spi.ConfigContext;
import io.helidon.config.spi.ConfigNode;
import io.helidon.config.spi.ConfigSource;
import io.helidon.config.spi.LazyConfigSource;

import java.util.Map;
import java.util.Optional;
import java.util.concurrent.ConcurrentHashMap;
import java.util.concurrent.atomic.AtomicInteger;
import java.util.function.Supplier;
import java.util.stream.Collectors;

public class ConfigTest {
    public static void main(String... args) {
        LazyImpl lazy = new LazyImpl();

        //Config config = Config.create();
        Config config = Config.builder()
                .sources(lazy, ConfigSources.file("src/test/resources/application.yaml"))
                //.disableCaching()
                //.failOnMissingKeyReference(false)
                //.failOnMissingValueReference(false)
                .build();

        try {
            String greeting = config.get("app.greeting").asString().get();
            System.out.println("greeting=" + greeting);
        }
        catch (Exception e) {e.printStackTrace();}

        try {
            String greeting = config.get("").asString().get();
            System.out.println("greeting=" + greeting);
        }
        catch (Exception e) {e.printStackTrace();}

        try {
            String greeting = config.get("nonexistentkey").asString().get();
            System.out.println("greeting=" + greeting);
        }
        catch (Exception e) {e.printStackTrace();}

        try {
            String nonExistentKey = config.get("non.existent.key").asString().get();
            System.out.println("non.existent.key=" + nonExistentKey);
        }
        catch (Exception e) {e.printStackTrace();}

        try {
            Supplier<String> nonExistentKeySupplier = config.get("non.existent.key").asString().supplier();
            System.out.println("non.existent.key(supplier)=" + nonExistentKeySupplier.get());
        }
        catch (Exception e) {e.printStackTrace();}

        System.out.println("printing all config options");
        for (Config configNode : config.asNodeList().get()) {
            if (configNode.key().name().equals("")) continue;
            configNode.traverse() // DEPTH-FIRST SEARCH
                    .filter(Config::isLeaf)
                    .forEach(node -> System.out.println(node.asString()));
        }

    }

}

final class LazyImpl implements ConfigSource, LazyConfigSource {
    private static final String DESCRIPTION = "LazyConfigSource-test";

    private final Map<String, String> values = new ConcurrentHashMap<>();

    @Override
    public void init(ConfigContext context) {
    }

    @Override
    public boolean exists() {
        return true;
    }

    @Override
    public String description() {
        return DESCRIPTION;
    }

    @Override
    public Optional<ConfigNode> node(String key) {
        System.out.println("lazy node creation for config option " + key);
        values.put(key, "dynamically-loaded-value");
        return Optional.ofNullable(values.get(key))
                .map(ConfigNode.ValueNode::create);
    }

    private void put(String key, String value) {
        values.put(key, value);
    }

    private void remove(String key) {
        values.remove(key);
    }
}

### LINKED ISSUE #8709: [4.x] Custom Helidon SE LazyConfigSource can't resolve a non-existent key when config.get('non-existent-key') is called
Backport of #7745 for Helidon 4.x

tomas-langer:
### Description
Previously `LazyConfigSource` was invoked only for nodes that existed in other config sources. If a node was not known by the config system, it would always return an empty value, even if one of the lazy config sources could provide a value for it.

This PR introduces a change where we first check if the key exists in any lazy source before caching it as non-existent.
As usual with Helidon Config, changes in time are not reflected, as the config tree is considered a snapshot of configuration (even though the lazy config source may introduce a value later in time, it is still considered immutable, and changes must be handled through change API).

Resolves #8709 

### Documentation
We are currently missing documentation of `LazyConfigSource` and I did not add it as part of this PR.

### DIFF
--- a/config/config/src/main/java/io/helidon/config/ProviderImpl.java
+++ b/config/config/src/main/java/io/helidon/config/ProviderImpl.java
@@ -34,6 +34,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import io.helidon.config.spi.ConfigFilter;
+import io.helidon.config.spi.ConfigNode;
 import io.helidon.config.spi.ConfigNode.ObjectNode;
 
 /**
@@ -114,6 +115,10 @@ class ProviderImpl implements Config.Context {
     @Override
     public synchronized Config last() {
         return lastConfig;
+    }
+
+    synchronized Optional<ConfigNode> lazyValue(String string) {
+        return configSource.lazyValue(string);
     }
 
     void onChange(Consumer<ConfigDiff> listener) {
--- a/config/config/src/main/java/io/helidon/config/ProviderImpl.java
+++ b/config/config/src/main/java/io/helidon/config/ProviderImpl.java
@@ -117,7 +117,7 @@ class ProviderImpl implements Config.Context {
         return lastConfig;
     }
 
-    synchronized Optional<ConfigNode> lazyValue(String string) {
+    Optional<ConfigNode> lazyValue(String string) {
         return configSource.lazyValue(string);
     }
 
--- a/config/config/src/main/java/io/helidon/config/ConfigFactory.java
+++ b/config/config/src/main/java/io/helidon/config/ConfigFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -70,8 +70,8 @@ final class ConfigFactory {
         this.provider = provider;
         this.aliasGenerator = aliasGenerator;
 
-        configCache = new ConcurrentHashMap<>();
-        timestamp = Instant.now();
+        this.configCache = new ConcurrentHashMap<>();
+        this.timestamp = Instant.now();
     }
 
     Instant timestamp() {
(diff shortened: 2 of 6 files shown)


config/config/src/main/java/io/helidon/config/ProviderImpl.java
```diff
@@ -116,6 +117,10 @@ public synchronized Config last() {
         return lastConfig;
     }
 
+    synchronized Optional<ConfigNode> lazyValue(String string) {
```
danielkec, line 120: Synchronized?

danielkec (dismissed):

config/config/src/main/java/io/helidon/config/ProviderImpl.java
tomas-langer: Fixed. Also fixed usage of `ConcurrentHashMap.computeIfAbsent` for possibly blocking operation.

tomas-langer:
Rebased on current `main` due to failure of examples build on MacOS, that had wrong protoc version.

danielkec (approved):

PR #8745, opened 2024-05-10
Refactor OCI metrics library a bit

tjquinno:
### Description
This PR refactors how the OCI metrics library code is organized just a bit so other implementations can extend common types without having to also include the OCI metrics library CDI module (which is helpful if the other implementation has its own CDI work to do).

### Documentation
No doc impact.


klustria:
LGTM

klustria (approved):
LGTM

PR #8757, opened 2024-05-13
4.x: Upgrade to MP Config 3.1 and fix an issue with profile specific properties

### LINKED ISSUE #8724: MicroProfile Config 3.1
Upgrade MicroProfile Config to version 3.1

* [3.1 Specification](https://github.com/eclipse/microprofile-config/releases/tag/3.1)
* [Overview Slide](https://docs.google.com/presentation/d/1A3Hbr-O7QFepUP5M0X2hKfqFiO1PZgnBWWhMyovu9JU/edit#slide=id.p9)

### LINKED ISSUE #8738: 4.x: MP Config does not correctly handling per property overriding with config profiles
## Environment Details
* Helidon Version: 4.0.8
* Helidon MP

----------

## Problem Description

Helidon MP Config is not correctly handling per property overriding precedence when config profiles ([spec](https://download.eclipse.org/microprofile/microprofile-config-3.1/microprofile-config-spec-3.1.html#_how_config_profile_works)) are used.  Note that this is not a change from the  MP Config 3.0 spec -- just a clarification.  

The proper behavior is that a profile-specific property only overrides the vanilla property per config source. Therefore if I specify the vanilla property in a config source with a high ordinal, it should trump any profile specific property from a config source at a lower ordinal. Our implementation appears to always place a higher precedence on the profile-specific property. Looks like this is the problematic code:

https://github.com/helidon-io/helidon/blob/a90531327f047aacd8fd16dc7785ba6bd44a19bf/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java#L114

https://github.com/helidon-io/helidon/blob/a90531327f047aacd8fd16dc7785ba6bd44a19bf/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java#L133

## Steps to reproduce

You can run the MP Config 3.1 TCK. See #8724.  Or

1. Create the MP quickstart using the Helidon CLI
2. Add this line to `microprofile-config.properties`: `%dev.app.greeting=Yo`
3. Build then Run the application with `java -Dmp.config.profile=dev -Dapp.greeting=Hola -jar target/quickstart-mp.jar`
4. Do `curl  -X GET http://localhost:8080/greet`

You'll see the application returns the property from the `dev` profile which is incorrect:
```
{"message":"Yo World!"}
```

According to the spec (and the 3.1 TCKs) it should return the value from the system property (`Hola`) since that ConfigSource has a higher ordinal.

barchetta:
### Description

1. Upgrade MicroProfile Config to version 3.1
2. Fix an issue (#8738) with profile specific property precedence that was found by a new TCK test 
3. Return a well-formed `ConfigValue` even when value is an unresolved expression (required by new TCK tests)
4. Add unit tests for the above issues

Fixes #8724
Fixes #8738

### Documentation

Updated with new MP Config version.

### Compatibility

MP Config 3.1 is fully backwards compatible with MP Config 3.0. But the bug fixes in this PR do alter the behavior of our implementation which potentially could impact a small set of users if they were depending on the buggy behavior. For this reason we should do this upgrade in 4.1 instead of 4.0.9.

### DIFF
--- a/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java
+++ b/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java
@@ -37,7 +37,6 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import io.helidon.config.ConfigSources;
 import io.helidon.config.mp.spi.MpConfigFilter;
 
 import org.eclipse.microprofile.config.Config;
@@ -343,7 +342,7 @@ class MpConfigImpl implements Config {
                         .map(it -> new ConfigValueImpl(propertyName, it, rawValue, source.getName(), source.getOrdinal()));
             } catch (NoSuchElementException e) {
                 // Property expression does not resolve
-                return Optional.empty();
+                return Optional.of(new ConfigValueImpl(propertyName, null, rawValue, source.getName(), source.getOrdinal()));
             }
         }
 
--- a/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java
+++ b/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java
@@ -92,8 +92,6 @@ class MpConfigImpl implements Config {
         this.converters.putIfAbsent(String.class, value -> value);
         this.configProfile = profile;
 
-        dumpConfigSources(this.sources);
-
         this.valueResolving = getOptionalValue("mp.config.property.expressions.enabled", Boolean.class)
                 .or(() -> getOptionalValue("helidon.config.value-resolving.enabled", Boolean.class))
                 .orElse(true);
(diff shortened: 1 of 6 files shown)


config/config-mp/src/main/java/io/helidon/config/mp/MpConfigImpl.java
```diff
@@ -107,13 +107,18 @@ class MpConfigImpl implements Config {
...
+
         if (configProfile == null) {
-            return findConfigValue(key)
-                    .orElseGet(() -> new ConfigValueImpl(key, null, null, null, 0));
+            return value;
         }
-        return findConfigValue("%" + configProfile + "." + key)
-                .or(() -> findConfigValue(key))
+
+        ConfigValue profileValue = findConfigValue("%" + configProfile + "." + key)
```
tomas-langer, line 118: here you could just use `orElse(value)`, and drop the `orElseGet()`

barchetta: Done

tomas-langer (approved):

PR #8766, opened 2024-05-15
Service Registry

### LINKED ISSUE #7052: Helidon Injection and ServiceLoader interaction
4.x or post 4.x

Currently we have a  lot of extension points done through service loader (Java `ServiceLoader`). We can expect that people using Helidon injection will attempt to create a config source and annotate it with `@Singleton` and expect it would be picked up as a conifg source (similar for media supports, security providers etc.).

We should prepare a strategy to handle this for 4.x (and post 4.x), as otherwise we would end up with two lifecycles that are not compatible or aligned.

Possible quick fix solution - if within injection, also add services discovered using service discovery, in addition to the services discovered using `ServiceLoader`. 

Some complications
- there are features that are specific depedening on the location - for example protocol configuration can be different for each socket - this woudl require named services (and a lot of documentation)
- "primordial services" - we need to bootstrap Helidon Injection, and the bootstrap cannot use discovered services - it must only use service loader based services (so "Primoridal config" could be done using `Config.create()`, and the runtime config could be a combination of discovered config source, parsers etc. with what is defined on classpath - such as meta configuration)

### LINKED ISSUE #8673: Minor issue with helidon 4.0.7?
I see these dependencies in my build:

```
io.helidon.common:helidon-common-tls:4.0.7
|    |    |    |    +--- io.helidon.common:helidon-common:4.0.7
|    |    |    |    +--- io.helidon.config:helidon-config:4.0.7
|    |    |    |    |    +--- io.helidon.common:helidon-common-config:4.0.7 (*)
|    |    |    |    |    +--- io.helidon.common:helidon-common:4.0.7
|    |    |    |    |    +--- io.helidon.common:helidon-common-media-type:4.0.7 (*)
|    |    |    |    |    \--- io.helidon.inject:helidon-inject-api:4.0.7
|    |    |    |    |         +--- io.helidon.common:helidon-common-types:4.0.7
|    |    |    |    |         |    +--- io.helidon.common:helidon-common:4.0.7
|    |    |    |    |         |    \--- io.helidon.builder:helidon-builder-api:4.0.7 (*)
|    |    |    |    |         +--- io.helidon.common:helidon-common:4.0.7
|    |    |    |    |         +--- io.helidon.common:helidon-common-config:4.0.7 (*)
|    |    |    |    |         +--- io.helidon.logging:helidon-logging-common:4.0.7
|    |    |    |    |         |    \--- io.helidon.common:helidon-common:4.0.7
|    |    |    |    |         +--- jakarta.inject:jakarta.inject-api:2.0.1
```

The issue is that **I only want to use Helidon SE 4.0.7,** so why am I getting **jakarta.inject:jakarta.inject-api:2.0.1**,  which I believe is part of Helidon MP???

### LINKED ISSUE #8806: 4.x: Deprecate OciConfig and related types in OCI SDK Runtime
This is a follow up issue for Service Registry PR (#8766)

The existing classes should be correctly deprecated (both blueprint, and the generated type).

tomas-langer:
### Introduction of Service Registry

Resolves #8806
Resolves #8673 
Resolves #7052 

Initial step for all tasks related with refactoring of Helidon Inject.

Update to existing modules to use Service Registry.

This changes modules that were backed by original Helidon Inject to use the new Service Registry. 
The API changes are only in the area of Helidon Inject, which was a preview feature.

There is a new module for OCI: `helidon-integrations-oci` that contains support for `Region` and `AuthenticationDetailsProvider` through Helidon Service Registry. OCI TLS support and config sources are now based on this new module.

Service Registry is currently documented through readme files in the repository structure. 

I have moved  the two config sources in OCI to a different folder names, as the current ones are not aligned with our guidelines (in addition to the changes described above).

Known issues (will be fixed with first set of review fixes):

1. All configuration options that are required before Config is set up will be prefixed with `helidon` - i.e. `helidon.oci.region`, `helidon.oci.atn-strategy` - these now use just `oci`, which may conflict with service configuration


danielkec (dismissed):

integrations/oci/oci/src/main/java/io/helidon/integrations/oci/AtnStrategyInstancePrincipal.java
```diff
@@ -63,18 +64,25 @@ public Optional<AbstractAuthenticationDetailsProvider> provider() {
...
         return LazyValue.create(() -> {
-            if (imdsVailable(config)) {
-                return Optional.of(InstancePrincipalsAuthenticationDetailsProvider.builder().build());
+            if (imdsAvailable(config)) {
+                var builder = InstancePrincipalsAuthenticationDetailsProvider.builder()
+                        .timeoutForEachRetry((int) config.atnTimeout().toMillis());
+
+                config.imdsBaseUri()
+                        .map(URI::toString)
+                        .ifPresent(builder::metadataBaseUrl);
```
danielkec, line 73: @klustria fiy

danielkec (dismissed):

danielkec (approved):

danielkec (approved):

Verdent (dismissed):

integrations/oci/oci/src/main/java/io/helidon/integrations/oci/AtnStrategyInstancePrincipal.java
klustria: Can we add `config.imdsBaseUri()` in builder if it exist? `InstancePrincipalsAuthenticationDetailsProvider.builder()` has a method `.metadataBaseUrl(metadataBaseUrl)` that allows setting of this. When this is not set, this API will use the default, but if set, then it can be used in combination with ssh tunnel to run Instance Principal authentication from a laptop.

tomas-langer: This is exactly as the code is written right now?

```
config.imdsBaseUri() // configured IMDS URI (optional, if not configured, use default)
                        .map(URI::toString) // as string
                        .ifPresent(builder::metadataBaseUrl); // set on the InstancePrincipalAtuhenticationDetailsProvider builder
```

klustria: Ignore my comment, you are right.

integrations/oci/oci/README.md
```diff
@@ -0,0 +1,43 @@
...
+The provider is looked up by using instances of `io.helidon.integrations.oci.spi.OciAtnStrategy`. The first service instance to provide an authentication provider is used, and no other service instance is called.
+
+The following out-of-the-box implementations exist:
+
+- Config based provider: uses `OciConfig` to create authentication provider; Weight: default - 10
+- Config file based provider: uses OCI config file (i.e. `~/.oci/config`) to create authentication provider; both file location and profile can be configured through `OciConfig`, Weight: default - 20
+- Resource principal provider: uses resource principal authentication provider; Weight: default - 30
+- Instance principal provider: uses instance principal authentication provider; Weight: default - 40
+
+To create a custom instance of authentication details provider, just create a new service for service registry
```
arjav-desai, line 25: We need to sync classes added in this module with what we already have in https://github.com/helidon-io/helidon/tree/main/integrations/oci/sdk/cdi/src/main/java/io/helidon/integrations/oci/sdk/cdi

integrations/oci/sdk/runtime/src/main/java/module-info.java
```diff
@@ -16,7 +16,10 @@
 
 /**
  * Helidon Injection Integrations to support OCI Runtime module.
+ *
+ * @deprecated replaced with {@code helidon-integrations-oci} module
  */
+@Deprecated(forRemoval = true, since = "4.1.0")
```
arjav-desai, line 22: Not all classes in this module are marked deprecated e.g. 
OciConfigBlueprint.java, OciRegionProvider.java. Do we need to?

tomas-langer: Neither of them are public.

tomas-langer: I will take a note to generate the `OciConfig` with deprecated annotation, and do it as a follow up

tomas-langer: Created #8806

danielkec (approved):

Issue #8767, opened 2024-05-16
Heap pressure and OutOfMemory(OOM) when looking up large number of config keys


logovind:
## Environment Details
* Helidon Version: 4.0.8
* Helidon SE or Helidon MP: Helidon MP 
* JDK version: JDK 21.0.3
* OS: Oracle Linux
* Docker version (if applicable): 

----------

## Problem Description
we tested Helidon Config with Helidon 4(Version 4.0.8) using sample app. During the test, we found in one of the scenario the sample app continuously look for some config entries. It results in heap growth leading to OOM. checking further we saw the growth is primarily coming from "io.helidon.config.mp.MpEnvironmentVariablesSource". The implementation of this has the cache with concurrent hashmap and all the looked up config items getting added to cache even when environment variable corresponding to the key does not exists.

logovind:
To simulate this issue we used following code in the sample app where it uses randomly gerated string as the key. 


test("Get_Config_with_default_value_test", responseMap, () -> StringUtils.equalsAnyIgnoreCase(config.getString(UUID.randomUUID().toString(), RANDOM), RANDOM));
test("Optional_config_test", responseMap, () -> config.getOptional(UUID.randomUUID().toString(), String.class).isEmpty());


Heap footprint:

Class Name                                              |   Objects | Shallow Heap |   Retained Heap
io.helidon.config.mp.MpConfigImpl                       |         1 |           32 | >= 63,09,50,784
io.helidon.config.mp.MpEnvironmentVariablesSource       |         2 |           48 | >= 37,46,57,888
io.helidon.config.mp.MpEnvironmentVariablesSource$Cached| 64,07,188 | 10,25,15,008 | >= 10,25,15,008

private final Map<String, Cached> cache = new ConcurrentHashMap<>();

we see that from the heap dump the cache which is implemented as concurrent hashmap in io.helidon.config.mp.MpEnvironmentVariablesSource ended up with ~3M Hash entries with those random keys based on the run from the sample app.

some of sample entries below: 

Key                                 |Value
9d3a23c9-0f74-43b2-8d9a-8a35d41e3062.   |    io.helidon.config.mp.MpEnvironmentVariablesSource$Cached @ 0xed496ba8
261469de-2937-4eeb-9875-2a0a282e19a3    |     io.helidon.config.mp.MpEnvironmentVariablesSource$Cached @ 0xeff02598
[...]

PR #8768, opened 2024-05-16
Switched implementation of MpEnvironmentVariablesSource to use an LRU cache

spericas:
### Description

Switched implementation of `MpEnvironmentVariablesSource` to use an LRU cache and avoid an OOM. See issue #8767.

### DIFF
--- a/config/config-mp/pom.xml
+++ b/config/config-mp/pom.xml
@@ -51,6 +51,10 @@
             <artifactId>microprofile-config-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-configurable</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.helidon.config</groupId>
             <artifactId>helidon-config-metadata</artifactId>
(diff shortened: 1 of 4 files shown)


tomas-langer (commented):
I am not sure I agree with this, as now for each property that is not part of the environment variables, we will have to go through all the steps again, whenever it is requested.

Can we maybe replace the cache with our `io.helidon.common.configurable.LruCache` that has a configurable size? 

This use case is wrong, and is actually expected to break us - and it would break SE Config as well, as we are intentionally caching the whole configuration tree.

spericas:
> I am not sure I agree with this, as now for each property that is not part of the environment variables, we will have to go through all the steps again, whenever it is requested.
> 
> Can we maybe replace the cache with our `io.helidon.common.configurable.LruCache` that has a configurable size?
> 
> This use case is wrong, and is actually expected to break us - and it would break SE Config as well, as we are intentionally caching the whole configuration tree.

Understood, let's consider that and also possibly using filters. I will close the PR for now.

tomas-langer (approved):

PR #8771, opened 2024-05-17
4.x - AbstractConfigurableExtension native-image fix

romain-grecourt:
### Description

Update AbstractConfigurableExtension to lookup config values lazily.
This resolves an issue with native-image where the config values are fixed at build-time.

### Documentation

None

### DIFF
--- a/integrations/cdi/common-cdi/configurable/src/main/java/io/helidon/integrations/cdi/configurable/ConfiguredProperties.java
+++ b/integrations/cdi/common-cdi/configurable/src/main/java/io/helidon/integrations/cdi/configurable/ConfiguredProperties.java
@@ -1,0 +1,261 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


integrations/cdi/common-cdi/configurable/src/main/java/io/helidon/integrations/cdi/configurable/ConfiguredProperties.java
```diff
@@ -0,0 +1,261 @@
...
+    @SuppressWarnings("unused")
+    private volatile Map<Object, Object> entries;
+    private final Map<Object, String> keys = new HashMap<>();
+    private final Config config;
+
+    ConfiguredProperties(Config config) {
+        this.config = config;
+    }
+
+    void configKey(String propertyName, String configKey) {
```
ljnelson, line 59: Minor: I find the name of this method to be strange. What is its purpose? How is it different from a straight `put`/`setProperty`?

integrations/cdi/common-cdi/configurable/src/main/java/io/helidon/integrations/cdi/configurable/ConfiguredProperties.java
```diff
@@ -0,0 +1,261 @@
...
+            throw new ExceptionInInitializerError(e);
+        }
+    }
+
+    @SuppressWarnings("unused")
+    private volatile Map<Object, Object> entries;
+    private final Map<Object, String> keys = new HashMap<>();
+    private final Config config;
+
+    ConfiguredProperties(Config config) {
```
ljnelson, line 55: You can remove the explicit dependency on MicroProfile Config and enable much simpler testing if, instead, you make this a `Function<? super String, ? extends Optional<String>>`. (Then later instead of `config.getOptionalValue(x, String.class)` you do `config.apply(x)`.)

ljnelson (dismissed):
Anyway minor comments LGTM in general

ljnelson (approved):

PR #8778, opened 2024-05-17
Use Helidon tracer, span builder, span types instead of OTel ones so we can trigger span listeners

### LINKED ISSUE #8370: [4.x] Should MP OpenTelemetry support add Helidon span, span context, and scope instances to request context?
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Currently, Helidon's MP Telemetry tracing support adds request context settings for the OTel span, span context, and scope so the developer's code can retrieve those if desired.

But Helidon does not also add an instance of  _Helidon_ span, span context, and scope.

This issue does not suggest adding a _new_ span; it simply suggests that our filter should add a Helidon `Span`, `SpanContext`, and `Scope` which are _wrappers_ around the OTel ones.

This allows user code to use the Helidon API to start a child span in their code, based on the existing span, span context, and scope already associated with the request. 

This could be particularly helpful for developers migrating 3.x MP Tracing applications (which could have used the Helidon types) to 4.x MP Telemetry.

### LINKED ISSUE #8774: Span listener not invoked when span is triggered by `@WithSpan` annotation or MP telemetry container or client filters
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
Registered `SpanListener` instances should be invoked for each span life cycle event, but they are not for spans created via the `@WithSpan` annotation.

The same is true for spans created by our MP telemetry filters.

tjquinno:
### Description
Resolves #8774
Resolves #8370

The Helidon MP telemetry handling of the `@WithSpan` annotation used the OTel types (tracer, span, and span builder). This made sense initially because `@WithSpan` is an OpenTelemetry annotation.

Similarly, our container and client filters for setting the correct span context used the OTel types.

But recently we added support for span listeners, and the listeners are invoked from the Helidon wrappers around the various tracing implementations. Because the code that handled `@WithSpan` and the two filters all used the OTel types directly instead of the Helidon wrappers the listeners were not invoked for spans created due to `@WithSpan` or the filters.

The main changes in this PR:
* `WithSpanInterceptor` now uses the Helidon tracing types. 
* `HelidonTelemetryContainerFilter` and `HelidonTelemetryClientFilter` now use the Helidon tracing types. As part of this change, the filters add the Helidon types to the request contexts so the backside filter method (invoked after the request has been processed) can obtain them and work with them. (#8370)

Testing changes:
* Addition of a new test to make sure that listeners are invoked for spans triggered by `@WithSpan`.
* Addition of a new test to make sure that the parent span is set automatically when there is a current one and that the OTel `SpanBuilder#setNoParent` method works as a way to create a root span.

### Documentation
This is a bug fix and has no doc impact.

### DIFF
--- a/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonOpenTelemetry.java
+++ b/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonOpenTelemetry.java
@@ -99,8 +99,8 @@ public final class HelidonOpenTelemetry {
      * @param helidonTracer Helidon {@link io.helidon.tracing.Tracer} to use in creating the wrapping span builder
      * @return Helidon {@link io.helidon.tracing.Span.Builder}
      */
-    public static  io.helidon.tracing.Span.Builder<?> create(SpanBuilder spanBuilder,
-                                                             io.helidon.tracing.Tracer helidonTracer) {
+    public static io.helidon.tracing.Span.Builder<?> create(SpanBuilder spanBuilder,
+                                                            io.helidon.tracing.Tracer helidonTracer) {
 
         return new OpenTelemetrySpanBuilder(spanBuilder, helidonTracer.unwrap(OpenTelemetryTracer.class).spanListeners());
     }
--- a/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonOpenTelemetry.java
+++ b/tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonOpenTelemetry.java
@@ -30,6 +30,7 @@ import io.helidon.tracing.SpanListener;
 import io.opentelemetry.api.OpenTelemetry;
 import io.opentelemetry.api.baggage.Baggage;
 import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
 import io.opentelemetry.api.trace.Tracer;
 
 import static io.opentelemetry.context.Context.current;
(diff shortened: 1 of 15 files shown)


tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonOpenTelemetry.java
```diff
@@ -91,6 +92,19 @@ public static io.helidon.tracing.Span create(Span span, Baggage baggage) {
...
     }
 
+    /**
+     * Wrap an open telemetry span builder.
+     *
+     * @param spanBuilder open telemetry span builder
+     * @param helidonTracer Helidon {@link io.helidon.tracing.Tracer} to use in creating the wrapping span builder
+     * @return Helidon {@link io.helidon.tracing.Span.Builder}
+     */
+    public static  io.helidon.tracing.Span.Builder<?> create(SpanBuilder spanBuilder,
```
tomas-langer, line 102: there are two spaces after `static` instead of one

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracer.java
```diff
@@ -107,6 +112,10 @@ public Tracer register(SpanListener listener) {
         return this;
     }
 
+    List<SpanListener> spanListeners() {
+        return Collections.unmodifiableList(spanListeners);
```
tomas-langer, line 116: `List.copyOf(spanListeners)` can be used instead of this.

tjquinno: I think the original is what we want.

It is possible for the developer to continue to add listeners after the tracer is created. (Unlikely, perhaps, but possible.) `List.of()` will exclude listeners added later because it makes a copy of the list _at that moment_; as written will include new ones.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonOpenTelemetry.java
tjquinno: Fixed.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracer.java
tomas-langer: It is not public API, so probably not important. If public, then we must make sure the docs are aligned with the method chosen.

tomas-langer (approved):

Issue #7792, opened 2023-10-14
4.x: Use `System.Logger` instead of JUL where applicable


tomas-langer:
Revisit all usages of Java Util Logging (JUL) and replace relevant cases with `System.Logger`, so we can remove the dependency on `java.logging` from most of our modules.

The dependency is expected in:
- `helidon-logging-jul` - the integration with JUL
- temporary as a workaround for shutdown hook (I think in webserver), unless already abstracted (or abstract it into our logging implementations)
- examples and test scopes (to correctly log in these cases)

jbescos:
There is one module that the API depends on Logger Level:
https://github.com/helidon-io/helidon/blob/e0a91cb10c1488516e8b14513a4e7a79402d6232/common/reactive/src/main/java/io/helidon/common/reactive/Single.java#L576

PR #8791, opened 2024-05-23
4.x: Use System.Logger instead of JUL where applicable #7792

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/7792

### Documentation
N/A

### DIFF
--- a/common/configurable/src/main/java/module-info.java
+++ b/common/configurable/src/main/java/module-info.java
@@ -24,7 +24,6 @@ module io.helidon.common.configurable {
     requires io.helidon.builder.api;
     requires io.helidon.common.context;
     requires java.management;
-    requires java.logging;
 
     requires transitive io.helidon.common.config;
     requires transitive io.helidon.common;
--- a/common/configurable/src/main/java/module-info.java
+++ b/common/configurable/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ module io.helidon.common.configurable {
     requires io.helidon.builder.api;
     requires io.helidon.common.context;
     requires java.management;
+    requires java.logging;
 
     requires transitive io.helidon.common.config;
     requires transitive io.helidon.common;
--- a/common/configurable/src/test/java/io/helidon/common/configurable/ThreadPoolTest.java
+++ b/common/configurable/src/test/java/io/helidon/common/configurable/ThreadPoolTest.java
@@ -27,6 +27,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
 import java.util.stream.IntStream;
 
 import org.awaitility.core.ConditionTimeoutException;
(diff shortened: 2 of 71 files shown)


jbescos:
There are some leftovers where I was not able to remove because of any of the next reasons:

1. There is a strong dependency to JUL in one public method.
2. Class requires dependencies on JUL (handlers, log records, etc)

```
$ grep -R 'requires\ java\.logging' --include 'module-info.java' ./
./webserver/observe/log/src/main/java/module-info.java:    requires java.logging;
./webserver/access-log/src/main/java/module-info.java:    requires java.logging;
./logging/jul/src/main/java/module-info.java:    requires java.logging;
./helidon/src/main/java/module-info.java:    requires java.logging;
./messaging/messaging/src/main/java/module-info.java:    requires java.logging;
./integrations/cdi/jta-weld/src/main/java/module-info.java:    requires java.logging;
./integrations/micrometer/micrometer/src/main/java/module-info.java:    requires java.logging;
./tracing/provider-tests/src/main/java/module-info.java:    requires java.logging;
./microprofile/openapi/src/main/java/module-info.java:    requires java.logging; // logging required for SnakeYAML logging workaround
./common/reactive/src/main/java/module-info.java:    requires java.logging;
./common/configurable/src/main/java/module-info.java:    requires java.logging;
./common/testing/junit5/src/main/java/module-info.java:    requires java.logging;

```

common/configurable/src/main/java/module-info.java
```diff
@@ -24,6 +24,7 @@
     requires io.helidon.builder.api;
     requires io.helidon.common.context;
     requires java.management;
+    requires java.logging;
```
tomas-langer, line 27: This should not be done. JUL is only used in tests.

webserver/observe/log/src/main/java/module-info.java
```diff
@@ -41,5 +41,5 @@
 
     provides io.helidon.webserver.observe.spi.ObserveProvider
             with io.helidon.webserver.observe.log.LogObserveProvider;
-	
+
```
tomas-langer, line 44: This change is wrong, please remove it (you may still need to change the copyright, unless you squash your commits locally).

tomas-langer (changes requested):
Looks great, there are still a couple of problems to be fixed.
Thanks

common/configurable/src/main/java/module-info.java
tomas-langer: This is a problem of your configuration of IDE. It should work just fine when you run it from command line

tomas-langer: In Idea, you need to "modify options", and "do not use --module-path" option in your runner configuration, otherwise it uses the module of the main source, and will not work as expected.

common/configurable/src/test/java/io/helidon/common/configurable/ThreadPoolTest.java
```diff
@@ -279,40 +276,6 @@ void testNullRejectionPolicy() {
         ), "rejectionPolicy is null");
     }
 
-    @Test
```
tomas-langer, line 282: Please return the test, it will work when you run it from Maven.

tomas-langer (changes requested):

common/configurable/src/main/java/module-info.java
jbescos: This error comes from command line, I always use it. Mockito is not able to mock final classes (I think the limitation comes from proxied classes).

Even if that test was enabled, still the test is not correct because I need to set `System.Logger log` in the class, and this is `static final`. I would need to modify the class to be able to test it somehow and I don't see a nice way to do it. Probably the nicest is to have a provider that returns the log instance, so I can inject this provider in the test.

Testing that a warning log is written is not really that important. I still think we can remove the test.

common/configurable/src/test/java/io/helidon/common/configurable/ThreadPoolTest.java
tomas-langer: So the problem comes from compiler plugin, as the tests seem to be compile with module path.
To fix the problem, add this section to `maven-compiler-plugin` in `pom.xml` of the module:
```
                <executions>
                    <execution>
                        <id>default-testCompile</id>
                        <configuration>
                            <!--
                            We use JUL from tests, need this for compilation
                             -->
                            <compilerArgs>
                                <compilerArg>--add-modules</compilerArg>
                                <compilerArg>java.logging</compilerArg>
                            </compilerArgs>
                        </configuration>
                    </execution>
                </executions>
```

and re-introduce the test as it was

tomas-langer (changes requested):

tomas-langer (approved):

Issue #7738, opened 2023-10-05
archetype: add a logging option


romain-grecourt:
## Environment Details
* Helidon Version: 4.0.0-RC1
----------

## Problem Description

Add a logging option to the Helidon archetype (Steps and input details are TBD). Support 3 values:
- JUL (default)
- Log4J
- SLF4J (with logback as provider)

The generated files for each value should perform an equivalent logging configuration.
I.e. configure the same loggers with equivalent levels. 
E.g. set the level for `org.jboss` to `WARNING` 

This should be back-ported to 3.x.

PR #8792, opened 2024-05-23
4.x: Archetype - Add SLF4J dependency

### LINKED ISSUE #8760: 4.x - Archetype: include an SLF4J provider when SLF4J is used
## Environment Details
* Helidon Version: 4.0.9

----------

## Problem Description

Some generate applications are missing an SLF4J provider. The following logs can be seen when starting the application:

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

In this particular case, `com.zaxxer:HikariCP` is the one dependency using SLF4J.

```shell
[INFO] +- io.helidon.integrations.cdi:helidon-integrations-cdi-datasource-hikaricp:jar:4.0.0-SNAPSHOT:runtime
[INFO] |  +- io.helidon.integrations.cdi:helidon-integrations-cdi-datasource:jar:4.0.0-SNAPSHOT:runtime
[INFO] |  |  \- io.helidon.integrations.cdi:helidon-integrations-cdi-configurable:jar:4.0.0-SNAPSHOT:runtime
[INFO] |  \- com.zaxxer:HikariCP:jar:5.0.1:runtime
[INFO] |     \- org.slf4j:slf4j-api:jar:2.0.9:runtime
```

We need to identify all the third-party dependencies supported by the archetype that may include `org.slf4j:slf4j-api` and add an SLF4J provider dependency in the generated pom:

```xml
<dependency>
    <groupId>org.slf4j</groupId>
    <artifactId>slf4j-jdk-platform-logging</artifactId>
</dependency>
```

We need to make sure that we add it only once.

## Steps to reproduce

```shell
helidon init --batch -Dflavor=mp -Dapp-type=database
cd database-mp
mvn package -DskipTests
java -jar target/database-mp.jar
```

tvallin:
### Description

fix #8760

### Documentation

No impact

### DIFF
--- a/archetypes/archetypes/src/main/archetype/mp/custom/database-outputs.xml
+++ b/archetypes/archetypes/src/main/archetype/mp/custom/database-outputs.xml
@@ -155,7 +155,7 @@
                         </map>
                         <map>
                             <value key="groupId">org.slf4j</value>
-                            <value key="artifactId">slf4j-simple</value>
+                            <value key="artifactId">slf4j-jdk14</value>
                         </map>
                     </list>
                 </model>
--- a/archetypes/archetypes/src/main/archetype/mp/custom/database-outputs.xml
+++ b/archetypes/archetypes/src/main/archetype/mp/custom/database-outputs.xml
@@ -153,6 +153,10 @@
                             <value key="artifactId">helidon-integrations-cdi-datasource-hikaricp</value>
                             <value key="scope">runtime</value>
                         </map>
+                        <map>
+                            <value key="groupId">org.slf4j</value>
+                            <value key="artifactId">slf4j-simple</value>
+                        </map>
                     </list>
                 </model>
             </output>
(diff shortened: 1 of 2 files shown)


archetypes/archetypes/src/main/archetype/mp/custom/database-outputs.xml
```diff
@@ -153,6 +153,10 @@
                             <value key="artifactId">helidon-integrations-cdi-datasource-hikaricp</value>
                             <value key="scope">runtime</value>
                         </map>
+                        <map>
+                            <value key="groupId">org.slf4j</value>
+                            <value key="artifactId">slf4j-simple</value>
```
romain-grecourt, line 158: We should probably use `slf4j-jdk14`.
When we work #7738 the dependency will be a choice.

tvallin: Done

romain-grecourt (approved):

Issue #7799, opened 2023-10-14
4.x: add `helidon-logging-jul` as a test dependency to some modules


tomas-langer:
Some of the modules now print a long stacktrace as the configured appender is not available (as `helidon-logging-jul` is no longer a transitive dependency coming from config as it used to be (wrongly)).

Check output of maven, and add the module as a test dependency to those modules that print the stacktrace (tests do not fail, just the logging configuraiton).

Search for `HelidonConsoleHandler` in the build output

jbescos:
Running the test I don't see any stack traces in the output console.

romain-grecourt:
> Running the test I don't see any stack traces in the output console.

You probably need to grep for the output files under target/surefire-reports

PR #8804, opened 2024-05-27
4.x: add helidon-logging-jul as a test dependency to some modules #7799

jbescos:
### Description

https://github.com/helidon-io/helidon/issues/7799

### Documentation

N/A

### DIFF
--- a/config/config/pom.xml
+++ b/config/config/pom.xml
@@ -99,6 +99,11 @@
             <artifactId>slf4j-jdk14</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.helidon.logging</groupId>
+            <artifactId>helidon-logging-jul</artifactId>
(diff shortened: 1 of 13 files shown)


common/testing/junit5/pom.xml
```diff
@@ -40,5 +40,10 @@
             <artifactId>junit-jupiter-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
tomas-langer, line 43: This does not make sense here - there is no `logging.properties`, so the logging is not required.

tomas-langer: As the issue describes, this should only be added to modules that use our appender. Please make sure that is the case for all added dependencies.

tomas-langer (changes requested):

jbescos:
I'm starting from scratch in other PR

PR #8823, opened 2024-05-30
Retrieve the correct requested URI info path value, indpt of the routing path used to locate the handler

### LINKED ISSUE #8818: Requested URI path omits some path segments depending on how the service/handler is registered
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
In 4.0.0, invoking `request.requestedUri().toUri().toString()`  returned the entire path.

Beginning in 4.0.1, it omits path elements that are mentioned when the service is registered.

## Steps to reproduce
1. Revise the SE QuickStart example to add this line to `GreetService#getDefaultMessageHandler`:
   
   ```java
   System.err.println("Requested URI: " + request.requestedUri().toUri().toString());
   ```
2. Make sure the pom specifies Helidon version 4.0.0.
3. Build and run, then run `curl http://localhost:8080/greet` and you see
   
   ```list
   Requested URI: http://localhost:8080/greet
   ```
4. Now, change the Helidon version in the pom's parent specification to 4.0.1, rebuild, rerun, and access the app again. You see this output (missing path):
   
   ```list
   Requested URI: http://localhost:8080/
   ```

tjquinno:
### Description
Resolves #8818 

The requested URI info logic used to fetch its path value from the incoming request `path()` return value. This is incorrect; that path is relative to the routing path which selected the handler, and that value has nothing to do with the path the client originally used in its outgoing request (which is what the requested URI feature is intended to reveal).

This PR changes the way the path is derived for the requested URI info to use the path from the prologue which reflects the entire path, not just the portion used in routing within the server.

It also adds a webserver test that makes sure that the path in the requested URI info is correct.

----

TL;DR - This bug escaped our attention until now because in 4.0.0 the server always invoked some CORS code, whether CORS was enabled or not. Part of that logic retrieved each request's requested URI info. That data is lazily created and cached. The routing that sends requests to CORS has no path prefix, so the `serverRequest.path()` value during the CORS-triggered handling just happened to include the entire path. 

Later, when a user handler retrieved the requested URI info from the request, even though at that moment the `request.path()` value was relative to the routing, the cached requested URI info with the original CORS-triggered value was returned.

Beginning in 4.0.1 the server stopped invoking CORS logic unless CORS was enabled. Beginning then, the requested URI information retrieved from a user handler that used path-based routing would be incorrect, omitting the part of the path used in the routing rule.

### Documentation
Bug fix; no doc update.

### DIFF
--- a/webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
+++ b/webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
@@ -16,16 +16,10 @@
 
 package io.helidon.webserver.tests.gh2631;
 
-import io.helidon.http.Status;
-import io.helidon.webclient.api.HttpClientResponse;
 import io.helidon.webclient.http1.Http1Client;
-import io.helidon.webclient.http1.Http1ClientResponse;
-import io.helidon.webserver.WebServer;
-import io.helidon.webserver.WebServerConfig;
 import io.helidon.webserver.http.HttpRouting;
 import io.helidon.webserver.testing.junit5.ServerTest;
 import io.helidon.webserver.testing.junit5.SetUpRoute;
-import io.helidon.webserver.testing.junit5.SetUpServer;
 
 import org.junit.jupiter.api.Test;
 
--- a/webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
+++ b/webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@ServerTest
+class Gh8818Test {
+    private final Http1Client client;
+
+    Gh8818Test(WebServer server) {
+        this.client = Http1Client.builder()
```
tomas-langer, line 38: You can use `Http1Client` as a parameter of your test. It will be created for the webserver that was started

webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
```diff
@@ -0,0 +1,59 @@
...
+                               .baseUri("http://localhost:" + server.port())
+                               .build();
+    }
+
+    @SetUpServer
```
tomas-langer, line 43: There is `SetUpRoute` annotation that you can use, the method parameter is then either `HttpRouting.Builder` or `HttpRules`.
This method is for adding server features and doing other server config related changes.

webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
tjquinno: Changed.

webserver/tests/requested-uri-path-gh8818/src/test/java/io/helidon/webserver/tests/gh2631/Gh8818Test.java
tjquinno: Changed.

tomas-langer (approved):

Issue #8807, opened 2024-05-28
No doc for `@AddConfigBlock`


tjquinno:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
Support for `@AddConfigBlock` was added but there was no doc update to the testing page(s).

PR #8825, opened 2024-05-31
Doc for @AddConfigBlock #8807

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/8807

### Documentation
Included in this PR

### DIFF
--- a/docs/src/main/asciidoc/mp/guides/testing-junit5.adoc
+++ b/docs/src/main/asciidoc/mp/guides/testing-junit5.adoc
@@ -104,6 +104,10 @@ The testing extension supports a few additional annotations that allow for finer
 
 This is useful when we want to modify configuration or beans between executions. In such a case, injection into fields is not possible, as we would need a different instance for each test.
 | `@AddConfig(key = "app.greeting", value = "Unite")` |  Defines a new configuration (either on class level, or method level) by adding a single configuration key/value.
+| `@AddConfigBlock(type = "properties", value = """ +
+    some.key1=some.value1 +
+    some.key2=some.value2 +
+""")` |  Defines a new configuration (either on class level, or method level) by adding a single configuration key/value.
 | `@Configuration(configSources = "test-config.properties")` | Adds a whole config source from classpath.
 |====================
 
--- a/docs/src/main/asciidoc/mp/guides/testing-junit5.adoc
+++ b/docs/src/main/asciidoc/mp/guides/testing-junit5.adoc
@@ -103,11 +103,11 @@ The testing extension supports a few additional annotations that allow for finer
 | `@HelidonTest(resetPerTest = true)` | Resets the container for each method.
 
 This is useful when we want to modify configuration or beans between executions. In such a case, injection into fields is not possible, as we would need a different instance for each test.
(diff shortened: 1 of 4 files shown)


tjquinno (changes requested):
A few changes, please.

docs/src/main/asciidoc/mp/guides/testing-junit5.adoc
```diff
@@ -104,6 +104,10 @@ The testing extension supports a few additional annotations that allow for finer
 
 This is useful when we want to modify configuration or beans between executions. In such a case, injection into fields is not possible, as we would need a different instance for each test.
 | `@AddConfig(key = "app.greeting", value = "Unite")` |  Defines a new configuration (either on class level, or method level) by adding a single configuration key/value.
+| `@AddConfigBlock(type = "properties", value = """ +
+    some.key1=some.value1 +
+    some.key2=some.value2 +
+""")` |  Defines a new configuration (either on class level, or method level) by adding a single configuration key/value.
```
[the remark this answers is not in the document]
jbescos: Thank you for the review. It is done now.

docs/src/main/java/io/helidon/docs/mp/guides/TestingJunit5Snippets.java
```diff
@@ -92,6 +93,12 @@ void testConfiguredGreeting(WebTarget webTarget) {
                 validate(webTarget, "/greet", "Unite World!");
             }
 
+            @Test
+            @AddConfigBlock("app.greeting=Unite")
```
tjquinno, line 97: The example should illustrate what we can do with this annotation that we cannot do with the other...declare _multiple_ settings as in the text you added to the doc.

tjquinno (changes requested):
Sorry, one more small change request.

docs/src/main/java/io/helidon/docs/mp/guides/TestingJunit5Snippets.java
jbescos: The problem is that the `helidon/examples/quickstarts/helidon-quickstart-mp` only uses one config property, so there is no point to add more.

I cannot create new config properties to inject in the test as a field, because of:
`java.lang.RuntimeException: When a class is annotated with @HelidonTest(resetPerTest=true), injection into fields or constructor is not supported, as each test method uses a different CDI container. Field private java.lang.String io.helidon.microprofile.tests.testing.junit5.TestAddConfigBlockProperties.value1 is annotated with @Inject`

Basically I need to create a new Snippet to test something like:
```
@HelidonTest
@AddConfigBlock("""
    some.key1=some.value1
    some.key2=some.value2
""")
class TestAddConfigBlockProperties {

    @Inject
    @ConfigProperty(name = "some.key1")
    private String value1;

    @Inject
    @ConfigProperty(name = "some.key2")
    private String value2;

    @Test
    void testValue() {
        assertThat(value1, is("some.value1"));
        assertThat(value2, is("some.value2"));
    }
}

```
Do you agree with it?

jbescos: I have pushed the change. Just let me know whether you are fine with it.

tjquinno (approved):

Issue #8816, opened 2024-05-30
4:x: Filter ordering should honor feature weight


tomas-langer:
This is the expected ordering of features:

| Feature                | Weight |
|------------------------|--------|
| Context                | 1100   |
| Access Log             | 1000   |
| Tracing                | 900    |
| CORS                   | 950    |
| Security               | 800    |
| Routing (all handlers) | 100    |
| OpenAPI                | 90     |
| Observe                | 80     |

Anything configured in routing should honor this ordering - i.e. if a filter is registered in AccessLog feature, it should be called before any filter registered in routing.

Right now it seems that all routing filters are called first, and then filters from features (in correct order based on their weight).

This must be fixed, and a test created to validate that this ordering is honored.

tomas-langer:
Current behavior:
As long as you use an HttpFeature to configure a filter, it will honor its weight
If you register filter using routing directly, it will be considered default weight, ordered by insertion
Behavior of filters added directly from ServerFeature.setup :
- it will be after filters in routing
- it will be before filters from HttpFeatures with less than default weight
- it will be in the order of the server features (depending on whether the feature is added by hand or discovered automatically)

tomas-langer:
Behavior before fix (where the filter comes from (weight)):
```
http-feature(999)
routing-first(N/A)
routing-second(N/A)
server-feature(99)
server-feature(1000)
http-feature(98)
```

Desired behavior after fix:
```
server-feature(1000)
http-feature(999)
routing-first(N/A)
routing-second(N/A)
server-feature(99)
http-feature(98)
```

RickyFrost:
Can we have a way to specify filter weight if using the following bootstrap?
`
final webServer = previouslyCreatedAndPolulatedServerBuilder.routing(
        HttpRouting.builder().addFilter(new MyFilter())
                // register other routings here...
                .build()
    ).build();
`

PR #8826, opened 2024-05-31
4.x: register routing in weighted order of Server and HTTP Features

tomas-langer:
Resolves: #8816 

### Description
Correctly handle order of Server Features and HTTP Features when registering non-HTTP Feature elements to the routing.

With the updated version, all elements (filters, routes, services) are ordered depending on the feature's weight. This is achieved by creating a new HttpFeature for each Server Feature with the same weight, that collects all registrations and applies them once the HTTP Features are ordered by weight. HttpFeature registered from a Server Feature is left intact and applied on the real builder (as this already works as it should)

### Documentation
This aligns with the internal documentation of feature weights. 

I have added update of WebServer documentation in the latest commit.

### DIFF
--- a/docs/src/main/asciidoc/se/webserver.adoc
+++ b/docs/src/main/asciidoc/se/webserver.adoc
@@ -285,6 +285,15 @@ include::{sourcedir}/se/WebServerSnippets.java[tag=snippet_7, indent=0]
 ----
 In this example, the `GET` handler matches requests to `/hello/subpath`.
 
+[[anchor-http-feature]]
+=== Using `HttpFeature`
+
+By implementing the `io.helidon.webserver.http.HttpFeature` interface, you can organize multiple routes and/or filters into
+a feature, that will be setup according to its defined `io.helidon.common.Weight` (or using `io.helidon.common.Weighted`).
+
+Each service has access to the routing builder. HTTP Features are configured for each routing builder. If there is a need
+to configure a feature for multiple sockets, you can use <<Server Features, Server Feature>> instead.
+
 == Request Handling
 
 Implement the logic to handle requests to WebServer in a `Handler`, which is a `FunctionalInterface`.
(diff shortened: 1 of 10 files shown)


tjquinno:
The coding changes look OK to me, although I do not know this area of the system.

The user whose questions raised our awareness of this issue has specifically asked for public documentation explaining, first, the ordering of filters vs. handlers (maybe we think it should be intuitive but we can at least state what that ordering is) and, second, how weighting affects that ordering.

Ideally such a doc update should be part of this PR...maybe a brief section on the webserver page in the request handling section?

RickyFrost:
Tim, I was also asking about the ordering of filters vs filters from features. It seems like an undue burden to create a ServerFeature, and an HttpFeature, just so I can correctly weight my filter, don't you think?

tomas-langer:
Updated documentation of WebServer

docs/src/main/asciidoc/se/webserver.adoc
```diff
@@ -389,6 +472,125 @@ include::{sourcedir}/se/WebServerSnippets.java[tag=snippet_21, indent=0]
...
+|xref:openapi/openapi.adoc[OpenAPI]
+|90
+
+|xref:observability.adoc[Observability]
+|80
+|===
+
+=== Context
+
+Context feature adds a filter that executes all request within the context of `io.helidon.common.context.Context`.
```
danielkec, line 525:
```suggestion
Context feature adds a filter that executes all requests within the context of `io.helidon.common.context.Context`.
```

danielkec (dismissed):
Just a typo

docs/src/main/asciidoc/se/webserver.adoc
```diff
@@ -389,6 +472,125 @@ include::{sourcedir}/se/WebServerSnippets.java[tag=snippet_21, indent=0]
...
+== Server Features
+Server features provide additional functionality to the WebServer, through modification of the server configuration,
+listener configuration, or routing.
+
+A server feature can be added by implementing `io.helidon.webserver.spi.ServerFeature`.
+Server features support automated discovery, as long as the implementation is available through Java `ServiceLoader`.
+Server features can also be added through configuration, as can be seen above in <<Configuration Options, Configuration Options>>,
+configuration key `features`.
+
+All features (both `ServerFeature` and `HttpFeature`) honor weight of the feature (defined either through `@Weight` annotation,
```
tjquinno, line 488: There's no discussion of `HttpFeature`s specifically which might be OK but people considering adding a feature, reading this, and seeing `HttpFeature` mentioned will probably wonder what the difference is and how to decide which type of feature they should implement for their use case. It would be good if the doc had a very brief explanation of the difference between then and how to choose. 

Maybe if this topic needs a more complete treatment (maybe throwing `HttpService` into the mix as well?) that should be done in a separate PR.

tomas-langer: Updated documentation of Routing - added a short description of `HttpFeature`

tjquinno (approved):
The doc updates do help. The coding changes LGTM but this is not an area of the system I know well.

PR #8838, opened 2024-06-04
4.x: Updates to support latest dev release of GraalVM native image

tomas-langer:
Updates to native image configuration to support latest release of GraalVM (tested with `OpenJDK Runtime Environment GraalVM CE 23-dev+23.1 (build 23+23-jvmci-b01)`).

There are a few changes that cause compilation failures, probably due to the change of "strict image heap" now being the default. In most cases the problems were caused by instances getting into the image heap that were not marked as `initialize-at-build-time`.

This PR fixes all the cases that were needed for our MP-1 integration test, and for a quickstart-mp (pipeline should find if there are other issues).

### DIFF
--- a/common/configurable/src/main/resources/META-INF/native-image/io.helidon.common/helidon-common-configurable/native-image.properties
+++ b/common/configurable/src/main/resources/META-INF/native-image/io.helidon.common/helidon-common-configurable/native-image.properties
@@ -1,0 +1,18 @@
+#
+# Copyright (c) 2024 Oracle and/or its affiliates.
+#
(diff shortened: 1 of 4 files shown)


tvallin (approved):

tomas-langer:
I have tried building the mp-2 example
- it fails with GraalVM dev build for Java 23 (a feature from the database fails with no such method error)
- it works fine with GraalVM for Java 21

lprimak:
Please see https://github.com/helidon-io/helidon/discussions/8867 and https://github.com/helidon-io/helidon/issues/8863
(haven't tried with .10 yet)
Thank you!

PR #8841, opened 2024-06-05
Fix tracer information propagation across threads using Helidon context

### LINKED ISSUE #8824: Helidon 4.0.9 gets NPE on context aware wrapped async virtual thread use in JaegerDataPropagationProvider.java:41
## Environment Details
* Helidon Version: 4.0.9 SE
* JDK version: Graal 21
* OS: OL 7u5

----------

## Problem Description
Working tests in 4.0.8 are getting NPE in 4.0.9

[ERROR] validate.helidon.tracing.TracerPropagationTest.concurrentVirtualThreadUseCanary  Time elapsed: 0.02 s  <<< ERROR!
java.lang.NullPointerException: Cannot invoke "io.helidon.tracing.providers.opentelemetry.OpenTelemetryDataPropagationProvider$OpenTelemetryContext.tracer()" because "delegate" is null
	at io.helidon.tracing.providers.jaeger@4.0.9/io.helidon.tracing.providers.jaeger.JaegerDataPropagationProvider$JaegerContext.<init>(JaegerDataPropagationProvider.java:41)

(complete stack will be uploaded)

## Steps to reproduce
See uploaded test class that passes in 4.0.8 and fails in 4.0.9

tjquinno:
### Description
Resolves #8824 

The earlier PR #8742 added propagation of tracer information for the OTel provider that was patterned after the long-standing Jaeger propagation. 

It turns out there was a flaw in that logic for choosing the span to propagate. Instead of using the current span as managed by the tracing implementation the code used whatever span, if any, that was already set in the current Helidon `Context` and actually ignored the tracing implementation's notion of the currently-active span. 

The key change here is that the propagation logic uses the span _from the tracing implementation's current span_ only. 

Any span that might already be set in the current context at the time propagation is being prepared is immaterial. For example, in the time since the span was set in that context some user code might have closed the scope which had made that span the current one, meaning no span was current. Or user code could have made some _other_ span the current span in the meantime.

The PR also contains a slight adaptation of a user test that had exposed this problem in the first place (see the attachment to the issue). The new test is in `provider-tests` so all provider implementations by default need to pass it. The test is written using the Helidon neutral tracing API. Because our Zipkin provider implementation does not implement the Helidon tracing API, this PR modifies the Zipkin build to exclude the new test. A separate issue will capture the possible change for Zipkin to implement the Helidon neutral tracing API at which point the new test from this PR should no longer be excluded from the Zipkin build.

### Documentation
Bug fix; no doc impact.

### DIFF
--- a/tracing/provider-tests/src/main/java/io/helidon/tracing/providers/tests/TestTracerAndSpanPropagation.java
+++ b/tracing/provider-tests/src/main/java/io/helidon/tracing/providers/tests/TestTracerAndSpanPropagation.java
@@ -36,123 +36,102 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.sameInstance;
 
 /**
- * Derived as closely as possible from a user-supplied reproducer. Comments are ours, added for explanation or questions.
+ * Tests that the current tracer and span are propagated through Helidon contexts.
+ * <p>
+ * Derived as closely as possible from a user-supplied reproducer. Basically, logging was changed to use system
+ * logging instead of SLF4J as in the original test and test assertions were converted to use Hamcrest. The key logic
+ * in the test is unchanged.
+ * </p>
  */
 class TestTracerAndSpanPropagation {
-        private static final System.Logger LOGGER = System
-                .getLogger(TestTracerAndSpanPropagation.class.getName());
+    private static final System.Logger LOGGER = System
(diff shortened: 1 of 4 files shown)


spericas (approved):
LGTM

tomas-langer (approved):

PR #8878, opened 2024-06-12
gRPC MP Implementation

spericas:
### Description

Integration of H3 gRPC MP Implementation into H4. gRPC in H3 MP started its own Netty-based server on a new port, the new implementation in H4 is fully intergrated with the web server, which is now capable of detecting the type of a connection and route its data to the correct _handler_. As a result of this new integration, there is new gRPC server configuration (there is no gRPC server anymore) and a few other integration points have changed.

### Module Changes

- New helidon-microprofile-grpc-core module
- New helidon-microprofile-grpc-server module 
- Additions and updates to helidon-grpc-core and helidon-webserver-grpc modules

### gRPC MP Support

Just like in H3, there is now support for gRPC annotated endpoints:

```java
    @Grpc.GrpcService
    @ApplicationScoped
    public static class EchoService {

        @Grpc.Unary("Echo")
        public void echo(Echo.EchoRequest request, StreamObserver<Echo.EchoResponse> observer) {
            try {
                String message = request.getMessage();
                Echo.EchoResponse response = Echo.EchoResponse.newBuilder().setMessage(message).build();
                complete(observer, response);
            } catch (IllegalStateException e) {
                observer.onError(e);
            }
        }
    }
```

In addition to annotated endpoints, the gRPC CDI extension will scan and automatically register any beans that implement `GrpcService` (from SE) and `BindableService` (io.grpc), just like it did in H3. 

Server interceptors from the io.grpc API are support _only on_ annotated types either by direct reference using the `@GrpcInterceptors` or via CDI name binding.

```java
    @Grpc.GrpcService
    @ApplicationScoped
    @EchoInterceptorBinding       // name binding
    @Grpc.GrpcInterceptors(EchoInterceptor1.class)   // direct reference
    public static class EchoService { ... }

    @Grpc.GrpcInterceptor
    @ApplicationScoped
    public static class EchoInterceptor1 implements ServerInterceptor {

        @Override
        public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call,
                                                                     Metadata headers,
                                                                     ServerCallHandler<ReqT, RespT> next) {
            ...
        }
    }

    @Grpc.GrpcInterceptor
    @EchoInterceptorBinding
    @ApplicationScoped
    public static class EchoInterceptor2 implements ServerInterceptor {

        @Override
        public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call,
                                                                     Metadata headers,
                                                                     ServerCallHandler<ReqT, RespT> next) {
            ...
        }
    }

```

Additional Features in PR

- The `GrpcMpContext` SPI has also been ported to H4. It enables users to write CDI extensions that are given access to the CDI `BeanManager` and the server's `GrpcRouting` builder.
- A Context interceptor is automatically registered for all annotated endpoints
- A Tracing interceptor is registered for all annotated endpoints if `grpc.tracing.enabled` is set to `true` (defaults to`false`)


bom/pom.xml
```diff
@@ -100,6 +100,16 @@
                 <artifactId>helidon-grpc-core</artifactId>
                 <version>${helidon.version}</version>
             </dependency>
+            <dependency>
```
tomas-langer, line 103: When updating `bom/pom.xml`, please update `all/pom.xml` as well.

grpc/core/pom.xml
```diff
@@ -26,13 +26,41 @@
...
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-configurable</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata</artifactId>
```
tomas-langer, line 50: This should never be a transitive dependency. It should be `<optional>true</optional>`

grpc/core/pom.xml
```diff
@@ -26,13 +26,41 @@
...
         </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-configurable</artifactId>
+        </dependency>
+        <dependency>
```
tomas-langer, line 44: This should never be a dependency (except for applications). You should depend on `helidon-common-config`

grpc/core/src/main/java/io/helidon/grpc/core/CollectingObserver.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.helidon.webserver.grpc;
+package io.helidon.grpc.core;
```
tomas-langer, line 17: This is a backward incompatible change, please keep a copy in original location, deprecated it and mark for removal.

grpc/core/src/main/java/io/helidon/grpc/core/CollectingObserver.java
```diff
@@ -25,69 +25,69 @@
...
      *
-     * @param collector        ToDo: Add JavaDoc
-     * @param responseObserver ToDo: Add JavaDoc
+     * @param collector the collector
+     * @param observer the observer
      */
-    public CollectingObserver(Collector<T, A, R> collector, StreamObserver<U> responseObserver) {
-        this(collector, responseObserver, null, null, null);
+    public CollectingObserver(Collector<T, A, R> collector, StreamObserver<U> observer) {
+        this(collector, observer, null, null, null);
```
tomas-langer, line 52: This needs to be refactored:
- we do not support `null` as a parameter of a public executable, either methods are needed for each combination, or use a builder
- we do not use public constructors except for type where it is a common practice (Exceptions, InputStream etc. - not sure if this applies), use static `create` factor methods.

grpc/core/src/main/java/io/helidon/grpc/core/GrpcTlsDescriptor.java
```diff
@@ -0,0 +1,202 @@
...
+
+import io.helidon.common.configurable.Resource;
+import io.helidon.config.Config;
+import io.helidon.config.metadata.Configured;
+import io.helidon.config.metadata.ConfiguredOption;
+
+/**
+ * GrpcTlsDescriptor contains details about configuring TLS of a {@link io.grpc.Channel}.
+ */
+public class GrpcTlsDescriptor {
```
tomas-langer, line 27: This class should not be really in `core`, as server connection TLS is configured on webserver, and can never be configured in a different way (I think). 
In the end the same should be valid for client, where we should use the uderlying configuration of the WebClient

grpc/core/src/main/java/io/helidon/grpc/core/GrpcTracingContext.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Optional;
+
+import io.helidon.tracing.Span;
+
+import io.grpc.Context;
+
+/**
+ * Contextual information related to Tracing.
+ */
+public final class GrpcTracingContext {
```
tomas-langer, line 27: Is tracing built-into grpc? 
We have removed it from Helidon WebServer and I would expect we do the same thing for grpc - i.e. tracing is a standalone module that can be added.

grpc/core/src/main/java/module-info.java
```diff
@@ -20,6 +20,14 @@
 module io.helidon.grpc.core {
 
     requires io.helidon.common;
+    requires io.helidon.http;
+    requires io.helidon.common.context;
+    requires io.helidon.config;
+    requires io.helidon.config.metadata;
+    requires io.helidon.tracing;
+
+    requires java.logging;
```
tomas-langer, line 29: `java.logging` should never be required, we use `System.Logger`

grpc/core/src/main/java/module-info.java
```diff
@@ -20,6 +20,14 @@
 module io.helidon.grpc.core {
 
     requires io.helidon.common;
+    requires io.helidon.http;
+    requires io.helidon.common.context;
+    requires io.helidon.config;
+    requires io.helidon.config.metadata;
```
tomas-langer, line 26: Config metadata can only be `requires static`, if used at all. 
New configurable types should be built using builder API, and that does not use config metadata

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/AbstractMethodHandlerSupplier.java
```diff
@@ -0,0 +1,291 @@
...
+        return annotation != null && methodType.equals(annotation.type());
+    }
+
+    /**
+     * A base class for method handlers.
+     *
+     * @param <ReqT>  the request type
+     * @param <RespT> the response type
+     */
+    public abstract static class AbstractHandler<ReqT, RespT> implements MethodHandler<ReqT, RespT> {
```
tomas-langer, line 79: Class should not be public (it is in a package private type)

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/AbstractServiceBuilder.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /**
+     * Obtain a list of all of the methods declared on the service class.
+     *
+     * @param clazz  the service class
+     * @return a list of all of the methods declared on the service class
+     */
+    protected List<Method> getAllDeclaredMethods(Class<?> clazz) {
+        List<Method> result = new LinkedList<>();
```
tomas-langer, line 153: So much reflection at runtime. Could this be handle by a CDI extension at least?

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/BidirectionalMethodHandlerSupplier.java
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * A supplier of {@link MethodHandler}s for bi-directional streaming gRPC methods.
+ */
+public class BidirectionalMethodHandlerSupplier extends AbstractMethodHandlerSupplier {
+
+    /**
+     * Create a supplier of handlers for bi-directional streaming methods.
+     */
+    // method is public because it is loaded via ServiceLoader
+    public BidirectionalMethodHandlerSupplier() {
```
tomas-langer, line 40: Please do not use public constructors.

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/GrpcInterceptors.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface GrpcInterceptors {
+    /**
+     * An ordered list of interceptors.
+     *
+     * @return the ordered list of interceptors
+     */
+    Class<?>[] value();
```
tomas-langer, line 50: This is a strange concepts.
Interceptors should be triggered based on annotations on the methods, not explicitly defined (also the developer does not need to be aware of interceptors that are available at runtime).
Do we need this?

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/Instance.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
+
+import io.grpc.Status;
+
+/**
+ * A supplier of instances of objects.
+ */
+public interface Instance {
```
tomas-langer, line 29: This is a very fishy thing. It uses reflection, name conflicts with CDI, and at least part of it is implemented in `LazyValue` in Helidon.
I would again question if this is needed, and if it is the right way to achieve what we want.

microprofile/grpc/core/src/main/java/module-info.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+module io.helidon.microprofile.grpc.core {
+
+    requires io.helidon.common;
+    requires io.helidon.microprofile.config;
+
+    requires transitive io.helidon.grpc.core;
+    requires transitive jakarta.cdi;
+
+    requires java.logging;
```
tomas-langer, line 28: java.logging should never be required, please use `System.Logger`

microprofile/grpc/server/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+            <groupId>io.helidon.webserver</groupId>
+            <artifactId>helidon-webserver-http2</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata</artifactId>
```
tomas-langer, line 50: This should not be required (already mentioned above)

microprofile/grpc/server/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+            <groupId>io.helidon.logging</groupId>
+            <artifactId>helidon-logging-jul</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.grpc</groupId>
+            <artifactId>grpc-inprocess</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
```
tomas-langer, line 91: This library should not be used at all in Helidon

microprofile/grpc/server/pom.xml
```diff
@@ -0,0 +1,164 @@
...
+        </dependency>
+        <dependency>
+            <!-- redirecting Weld logging to JUL logging -->
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
```
tomas-langer, line 96: This should never be a dependency of a Helidon module except for testing.

microprofile/grpc/server/src/main/java/module-info.java
```diff
@@ -0,0 +1,63 @@
...
+module io.helidon.microprofile.grpc.server {
+
+    requires io.helidon.common;
+    requires io.helidon.common.configurable;
+    requires io.helidon.common.context;
+    requires io.helidon.common.features.api;
+    requires io.helidon.config.mp;
+    requires io.helidon.config.objectmapping;
+    requires io.helidon.config;
+    requires io.helidon.config.metadata;
```
tomas-langer, line 40: max. `requires static` for config metadata.

microprofile/grpc/server/src/main/java/module-info.java
```diff
@@ -0,0 +1,63 @@
...
+    requires io.helidon.microprofile.grpc.core;
+    requires io.helidon.microprofile.server;
+    requires io.helidon.webserver.grpc;
+
+    requires io.grpc;
+    requires io.grpc.inprocess;
+    requires io.grpc.protobuf.lite;
+    requires com.google.protobuf;
+
+    requires java.logging;
```
tomas-langer, line 50: Please switch to system logger.

microprofile/grpc/server/src/main/java/module-info.java
```diff
@@ -0,0 +1,63 @@
...
+
+    requires microprofile.health.api;
+
+    exports io.helidon.microprofile.grpc.server;
+    exports io.helidon.microprofile.grpc.server.spi;
+
+    uses io.helidon.microprofile.grpc.server.spi.GrpcMpExtension;
+
+    provides jakarta.enterprise.inject.spi.Extension with GrpcMpCdiExtension;
```
tomas-langer, line 59: Please use fully qualified class name.

webserver/grpc/pom.xml
```diff
@@ -43,6 +43,14 @@
             <groupId>io.helidon.webserver</groupId>
             <artifactId>helidon-webserver-http2</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.helidon.grpc</groupId>
+            <artifactId>helidon-grpc-core</artifactId>
+        </dependency>
+        <dependency>
```
tomas-langer, line 50: Please reconsider the tracing approach. It should not be "hardcoded" dependency.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/ResponseHelper.java
tomas-langer, line 1: This was a public class, please deprecate and mark for removal.

tomas-langer (commented):
- I think the annotations should be in `helidon-microprofile-grpc-api` module
- please make sure all public constructors are replaced with factory methods (unless they are allowed by our guidelines)
- please make sure we never accept `null` as a parameter value
 
There is a lot of reflection used here. It may be better to use annotation processing to generate all the needed stuff to be reflection free (maybe even use ServiceRegistry for this?).

grpc/core/pom.xml
spericas: There was a class removed in the other PR and not this one, all these deps go away.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/ResponseHelper.java
spericas: Done

microprofile/grpc/server/src/main/java/module-info.java
spericas: Fixed

grpc/core/src/main/java/io/helidon/grpc/core/GrpcTlsDescriptor.java
spericas: Removed

microprofile/grpc/server/pom.xml
spericas: Fixed

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/BidirectionalMethodHandlerSupplier.java
spericas: These are public because of ServiceLoader, as explained in the comment

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/GrpcInterceptors.java
spericas: This is supported in addition to name binding. May be of some use for user-defined interceptors, but generally I agree with you. We had this in H3.

microprofile/grpc/core/src/main/java/io/helidon/microprofile/grpc/core/Instance.java
spericas: Renamed for now

grpc/core/src/main/java/io/helidon/grpc/core/GrpcTracingContext.java
spericas: Moved tracing to its own module

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcTracingConfigBlueprint.java
```diff
@@ -0,0 +1,63 @@
...
+     * @return the tracing name
+     */
+    Optional<GrpcTracingName> operationNameConstructor();
+
+    /**
+     * Set of attributes to trace.
+     *
+     * @return set of attributes to trace
+     */
+    Set<ServerRequestAttribute> tracedAttributes();
```
tomas-langer, line 62: It may be useful to add `@Singular` to have `addTracedAttribute(ServerRequestAttribute)` method generated on the builder (if this makes sense)

webserver/grpc/src/main/java/io/helidon/webserver/grpc/ResponseHelper.java
```diff
@@ -27,12 +27,17 @@
...
 import java.util.stream.Stream;
 
+import io.helidon.grpc.core.SafeStreamObserver;
+
 import io.grpc.stub.StreamObserver;
 
 /**
- * A number of helper methods to handle sending responses to a {@link io.grpc.stub.StreamObserver}.
+ * This class now lives here {@link io.helidon.grpc.core.ResponseHelper}.
+ * Will be removed from this location in a future version.
```
tomas-langer, line 36: Please add (to all newly deprecated things) a javadoc tag `@deprecated`, and if marked for removal, add a reference to the new one, such as:
```
@deprecated will be removed in a future version, please use {@link io.helidon.grpc.core.ResponseHelper} instead
```

tomas-langer (approved):

PR #8888, opened 2024-06-14
4.x: Add todo for duplicates in CompositeArrayBufferData and CompositeListBufferData

Captain1653:
### Description
I've found, that there are many duplicated methods in `CompositeArrayBufferData` and `CompositeListBufferData`

For example `writeTo(ByteBuffer writeBuffer, int limit)`
Screenshot 2024-06-14 at 09 50 17

Maybe I should create the issue, but I'm not sure. The todo can be useful.


tomas-langer:
TODO in code should not be used. We may have some, but it is a bug, not a feature.

Captain1653:
@tomas-langer Thank you for the explanation. Then I'll close this PR.

PR #8889, opened 2024-06-16
Fix multi-value query string parsing

### LINKED ISSUE #8886: UriQueryWriteable.fromQueryString does not support multi-value parameters
## Environment Details
* Helidon Version: 4.0.10
* Helidon SE

----------

## Problem Description

When using a client to make a request with multi-value query params, only the last value is retained.

## Steps to reproduce

```java
@Test
void testMultiValueQueryParam() {
    try (Http1ClientResponse response = client.get("/greet")
            .queryParam("foo", "bar1")
            .queryParam("foo", "bar2")
            .request()) {

        assertThat(response.lastEndpointUri().query().all("foo"), is(List.of("bar1", "bar2")));
        // fails because we get ["bar2"]
    }
}
```

It boils down to `UriQueryWriteable.fromQueryString`:

```java
@Test
void testFromQueryWithMultiValueParameter() {
    UriQueryWriteable uriQuery = UriQueryWriteable.create();
    uriQuery.fromQueryString("foo=bar1&foo=bar2");
    assertThat(uriQuery.all("foo"), is(List.of("bar1", "bar2"))); 
    // fails because we get ["bar2"]
}
```

tomas-langer:
Resolves #8886 

When parsing a query with multiple values, we incorrectly called `set` instead of `add`, so only the last value was used.

Also the query must be encoded when used for parsing - fixed javadoc, and usage.

### DIFF
--- a/common/uri/src/main/java/io/helidon/common/uri/UriQueryWriteable.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriQueryWriteable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


spericas (approved):
LGTM

PR #8890, opened 2024-06-16
4.x: Improved parsing of HTTP/1 prologue and headers.

tomas-langer:
### Description
Improved parsing of HTTP prologue and headers.

- comparison through long words instead of bytes (faster CPU handling)
- the "fast path" for requests that do not need headers is improved, `Connection` header has optimized processing, as it is always used by Helidon webserver, and original code always had to use `LazyString.stripOws` which was sub-optimal; new version does a fast check for `close` and `keep-alive` connection options and bypasses this
- lazy headers are not copied when creating Headers, so the `stripOws` method is only called when the header is requested by user (or when validation is turned on)


tomas-langer:
Relevant improvements (JMH tests run on my machine):
```
==================== HttpPrologueParsingJmhTest.techEmpower (+8.27%)
                              Baseline ████████████████████████████████████████████████████████████████████████████████████████████ 58250706.875 ops/s
                               Current ████████████████████████████████████████████████████████████████████████████████████████████████████ 63503835.499 ops/s

```

spericas (approved):
Difficult to check logic line by line, but general method seems sound. Nice work!

PR #8891, opened 2024-06-16
4.x: Fixed configuration metadata of blueprints that are configured and provide a service

### LINKED ISSUE #8776: `config-metadata.json` lacks configKey for ConfiguredProviders
## Environment Details
* Helidon Version: 4.0.8

----------

## Problem Description

[ConfiguredProvider](https://github.com/helidon-io/helidon/blob/4.0.8/common/config/src/main/java/io/helidon/common/config/ConfiguredProvider.java) defines a method `configKey()`. This key is not reflected anywhere in `config-metadata.json`

For example [Http1ProtocolConfigProvider](https://github.com/helidon-io/helidon/blob/4.0.8/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ProtocolConfigProvider.java) has a configKey of `http_1_1`. But the corresponding entry in config-metadata.json lacks that information:

```
  {
                "annotatedType": "io.helidon.webserver.http1.Http1Config",
                "type": "io.helidon.webserver.http1.Http1Config",
                "producers": [
                    "io.helidon.webserver.http1.Http1Config#create(io.helidon.common.config.Config)",
                    "io.helidon.webserver.http1.Http1Config#builder()"
                ],
                "provides": [
                    "io.helidon.webserver.spi.ProtocolConfig"
                ],
                "options": [
                 . . .
                ]
  }
```

## Background

My assumption is that `config-metadata.json` provides the schema for module configuration. It should be possible to derive a skeleton `application.yaml` using only the information provided in `config-metadata.json` files.  Currently that is not possible due to missing `configKey`

tomas-langer:
Updated generated documentation to latest state of code.

Resolves #8776 

There were two problems in the code:
1. The protocol config blueprints used wrong provides (must match the `@Option.Provider(ProtocolConfigProvider.class)` in listener config)
2. The protocol config blueprints lacked the configured key, and setting that it is not a root configuration

After fixing the above problems, the generated documentation correctly shows the configuration keys, and the listener config documentation now lists all available implementations.

In addition the same problem was with observe providers. Fixed all providers to add config key, re-generated the documentation.

Added validation that `@Provides` and `@Configured` must have a configuration key.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/Types.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/Types.java
@@ -43,6 +43,7 @@ final class Types {
     static final TypeName PROTOTYPE_ANNOTATED = TypeName.create("io.helidon.builder.api.Prototype.Annotated");
     static final TypeName PROTOTYPE_FACTORY = TypeName.create("io.helidon.builder.api.Prototype.Factory");
     static final TypeName PROTOTYPE_CONFIGURED = TypeName.create("io.helidon.builder.api.Prototype.Configured");
+    static final TypeName PROTOTYPE_PROVIDES = TypeName.create("io.helidon.builder.api.Prototype.Provides");
     static final TypeName PROTOTYPE_BUILDER = TypeName.create("io.helidon.builder.api.Prototype.Builder");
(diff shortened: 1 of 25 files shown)


barchetta (approved):
Yay!

romain-grecourt (approved):

PR #8898, opened 2024-06-20
Correct documentation for rest-request.enabled option

matthewdavidlloyd:
### Description

`rest-request-enabled` as described in the [web documentation](https://helidon.io/docs/v4/mp/metrics/metrics) does nothing, the correct option is `rest-request.enabled`.

### Documentation
None

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -151,7 +151,7 @@
         <version.lib.postgresql>42.7.2</version.lib.postgresql>
         <version.lib.prometheus>0.16.0</version.lib.prometheus>
         <version.lib.reactivestreams>1.0.4</version.lib.reactivestreams>
-        <version.lib.slf4j>2.0.9</version.lib.slf4j>
+        <version.lib.slf4j>2.0.16</version.lib.slf4j>
         <version.lib.smallrye-openapi>3.3.4</version.lib.smallrye-openapi>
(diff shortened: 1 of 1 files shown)


tjquinno (changes requested):
Hi, Matthew. We appreciate the contribution!

As you might know, we use a tool to generate the config `.adoc` files from the Java types that declare the configuration. (The tool is currently internal but will become public at some point.) 

It's perfectly OK to hand-edit those files, especially given that we do not make that tool available to you. But as part of this PR please also update the declaration of `MetricsConfigBlueprint#restRequestEnabled`  with this change:

`@Option.Configured` -> `Option.Configured("rest-request.enabled")`

This will make sure that future  _generated_ `.adoc` files will be correct and that the updates in your PR are not lost.

I was not aware you were working on this--we usually have an issue which is assigned to a developer (internal or external) who then links the resolving PR to the issue.  I have just merged changes to some of the same files. You'll need to rebase your PR to get those changes and then resolve the conflicts carefully. 

You'll notice from my changes that the default value for `rest-request.enabled` should be `false` rather than blank. My PR changed the `MetricsConfig` `.adoc` file but not the `MetricsObserver` one. 

As you revise your PR please make sure the default value for `rest-request.enabled` is `false` in both of the `.adoc` files.

Issue #8897, opened 2024-06-19
Provide adaptive concurrency limits


arouel:
# Why

There may be several reasons why adaptive concurrency limiting is preferred over using a fixed limit:

1. **Dynamic System Conditions**: In a distributed system, conditions such as load, resource availability, and topology can change frequently due to factors like auto-scaling, partial outages, code deployments, or fluctuations in traffic patterns. A fixed concurrency limit cannot adapt to these dynamic conditions, leading to either under-utilization of resources or overwhelmed services.

2. **Latency Sensitivity**: Different services or use cases may have varying sensitivity to latency. A fixed concurrency limit cannot account for these differences, potentially leading to either excessive queuing and high latency or under-utilization of resources. An adaptive approach can adjust the limit based on observed latencies, maintaining desired performance characteristics.

3. **Simplicity and Autonomy**: Manually determining and configuring fixed concurrency limits for every service or instance can be a complex and error-prone process, especially in large-scale distributed systems. An adaptive approach can autonomously and continuously adjust the limit without manual intervention, simplifying operations and reducing the risk of misconfiguration.

4. **Resilience and Self-Healing**: By automatically adjusting the concurrency limit based on observed conditions, an adaptive approach promotes resilience and self-healing capabilities. It allows services to shed excessive load during periods of high demand or resource constraints, preventing cascading failures and promoting graceful degradation.

While a fixed concurrency limit may be easier to reason about and configure initially, it lacks the flexibility and adaptability required in modern, dynamic distributed systems. An adaptive approach provides the ability to continuously optimize performance, resource utilization, and resilience in the face of changing conditions, ultimately leading to a more robust and efficient system.

# Suggestion

Ideally, a user would be able to describe the limiting algorithm in the [ListenerConfig](https://helidon.io/docs/v4/apidocs/io.helidon.webserver/io/helidon/webserver/ListenerConfig.html#maxConcurrentRequests()) that fit their needs instead of a fixed number for `maxConcurrentRequests`. The `Limit` and `Limiter` interfaces from [Netflix's concurrency limits library](https://github.com/Netflix/concurrency-limits/blob/main/concurrency-limits-core/src/main/java/com/netflix/concurrency/limits/Limit.java#L24) are a good starting point. In the first iteration we should provide the following implementations
- no limit,
- fixed limit,
- AMID limit.

[...]

PR #8913, opened 2024-06-26
Get annotation processing going with Eclipse Compiler for Java

arouel:
### Problem

I'm writing a HTTP feature for Helidon (related to #8897) and want to work with Eclipse IDE, but the Helidon annotation processors throw exceptions because the *Eclipse Compiler for Java* (ECJ) behaves slightly different compared to *javac*.

### Solution

Some changes were necessary to get the annotation processing going:
- Due to the implementation of [IdeFilerImpl:236](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7e0c89e8e422ceb8334eaaa84066fcad32ad6bf4/org.eclipse.jdt.apt.pluggable.core/src/org/eclipse/jdt/internal/apt/pluggable/core/filer/IdeFilerImpl.java#L236) there is no handling for `javax.tools.StandardLocation.SOURCE_PATH` implemented and therefore the search for `module-info.java` fails with a thrown `IllegalArgumentException`. Ignoring this exception when compiling with the ECJ is fine.
- There is also a case where the ECJ cannot resolve a type name and the `Element` parameter is null. Returning an empty `Optional` enables the caller to fall back.
- Due to the implementation of [IdeInputFileObject:50](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/e56c823649b351a74f287ec1479c9004ef6ed11f/org.eclipse.jdt.apt.pluggable.core/src/org/eclipse/jdt/internal/apt/pluggable/core/filer/IdeInputFileObject.java#L50) the deletion of a `FileObject` throws an `IllegalStateException`. Ignoring any exception when compiling with the ECJ should be fine.

I have successfully tested the changes suggested by myself in a recent Eclipse IDE (`Version: 2024-06 (4.32.0)`) on my local computer.

### DIFF
--- a/codegen/apt/src/main/java/io/helidon/codegen/apt/AptContextImpl.java
+++ b/codegen/apt/src/main/java/io/helidon/codegen/apt/AptContextImpl.java
@@ -145,7 +145,7 @@ class AptContextImpl extends CodegenContextBase implements AptContext {
             try (InputStream in = resource.openInputStream()) {
                 return Optional.of(ModuleInfoSourceParser.parse(in));
             }
-        } catch (IOException ignored) {
+        } catch (Exception ignored) {
             // it is not in sources, let's see if it got generated
         }
         // generated
--- a/codegen/apt/src/main/java/io/helidon/codegen/apt/FilerTextResourceImpl.java
+++ b/codegen/apt/src/main/java/io/helidon/codegen/apt/FilerTextResourceImpl.java
@@ -76,7 +76,11 @@ class FilerTextResourceImpl implements FilerTextResource {
     public void write() {
         if (modified) {
             if (originalResource != null) {
-                originalResource.delete();
+                try {
+                    originalResource.delete();
+                } catch (Exception ignored) {
+                    // The resource cannot be deleted, e.g. because ECJ has not implemented this method.
+                }
             }
             try {
                 FileObject newResource = filer.createResource(StandardLocation.CLASS_OUTPUT,
--- a/codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeInfoFactory.java
+++ b/codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeInfoFactory.java
@@ -255,7 +255,13 @@ public final class AptTypeInfoFactory extends TypeInfoFactoryBase {
                 .throwsChecked(thrownChecked)
                 .parameterArguments(params)
                 .originatingElement(v);
-        AptTypeFactory.createTypeName(v.getEnclosingElement()).ifPresent(builder::enclosingType);
+
+        // To be failure-tolerant, as the ECJ may not provide an enclosing element for a VariableElement.
+        Element enclosingElement = v.getEnclosingElement();
+        if (enclosingElement != null) {
+            AptTypeFactory.createTypeName(enclosingElement).ifPresent(builder::enclosingType);
+        }
+
         Optional.ofNullable(defaultValue).ifPresent(builder::defaultValue);
 
         return mapElement(ctx, builder.build());
--- a/config/metadata/processor/src/main/java/io/helidon/config/metadata/processor/ConfigMetadataHandler.java
+++ b/config/metadata/processor/src/main/java/io/helidon/config/metadata/processor/ConfigMetadataHandler.java
@@ -18,6 +18,9 @@ package io.helidon.config.metadata.processor;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 import java.util.ArrayList;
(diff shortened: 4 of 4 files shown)


codegen/apt/src/main/java/io/helidon/codegen/apt/AptContextImpl.java
```diff
@@ -101,7 +101,7 @@ private static Optional<ModuleInfo> findModule(Filer filer) {
             try (InputStream in = resource.openInputStream()) {
                 return Optional.of(ModuleInfoSourceParser.parse(in));
             }
-        } catch (IOException ignored) {
+        } catch (Exception ignored) {
```
arouel, line 104: Should we consider catching `IOException` and `IllegalArgumentException` instead?

tomas-langer: I think catching `Exception` is sufficient here, as it shows that there was "a" problem getting the `module-info.java`. As we have some fallback mechanism here, and we do not require it, it should be fine.
It may be interesting to find an alternative for Eclipse IDE (is there "any" way to get the module info?), because we use the module info to get the module name, which is used in some generated types (the worst case here is that the used name will be `unnamed/package-name`) for documentation purposes.

tomas-langer:
If you want to go forward with this PR, kindly sign the OCA, as otherwise we cannot accept your contribution.

codegen/apt/src/main/java/io/helidon/codegen/apt/AptContextImpl.java
arouel: I double-checked the ECJ implementation and it only handles `javax.tools.StandardLocation` values that point to generated source code. Which means, we can't do anything here. We would need to adapt the ECJ and implement the missing cases we need.

arouel:
@tomas-langer I added one more commit to handle the last case I had issues with when using Eclipse IDE. Can you have a look at it as well? I signed the OCA already, I guess it needs a bit to be approved.

codegen/apt/src/main/java/io/helidon/codegen/apt/FilerTextResourceImpl.java
```diff
@@ -76,7 +76,11 @@ public void lines(List<String> newLines) {
     public void write() {
         if (modified) {
             if (originalResource != null) {
-                originalResource.delete();
+                try {
```
tomas-langer, line 79: This means that it will only work with `clean compile` and will not be able to do incremental build.

codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
```diff
@@ -161,6 +161,10 @@ public static Optional<TypeName> createTypeName(TypeElement element, TypeMirror
      * @return the associated type name instance
      */
     public static Optional<TypeName> createTypeName(Element type) {
+        if (type == null) {
```
tomas-langer, line 164: This actually violates Helidon design (we only support non-null parameters - always). 
We can merge it, I will create a follow up to fix this.

tomas-langer (approved):

tomas-langer (changes requested):
So I cannot merge this:
1. The `ConfigMetadataHandler` needs copyright update (I could do this on my own)
2. The change you did for writing the stack trace is causing a spotbugs failure. So either this needs to be added as an exception (requires `etc/spotbugs/exclude.xml` file in the module + update of pom file (example can be found in `config/config` module. or it must be reverted to original

codegen/apt/src/main/java/io/helidon/codegen/apt/FilerTextResourceImpl.java
arouel: I assume deleting a resource only fails when using ECJ and `javac` wouldn't fail to delete the fail, right?

codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
arouel: Shall I add a comment here that this was added only to get ECJ running?

arouel: I have added a comment as to why this check was added.

arouel:
@tomas-langer I updated the PR, can you have a look again?

arouel:
@tomas-langer can you have a look at the PR again? I think I addressed all your points. It would be great to gave it to get Eclipse IDE running.

arouel:
I fixed the Checkstyle error just now. I don't know what we do with the other [failing job](https://github.com/helidon-io/helidon/actions/runs/11182713733/job/31539886256). Could it be flappy?

tomas-langer (approved):

PR #8921, opened 2024-06-28
4.x: Service registry OCI integration update

tomas-langer:
Additional configuration for metadata based authentication details providers:
- federation endpoint
- detect endpoint retries

Use `BasicAuthenticationDetailsProvider` instead of `Abstract`, as that is required by OCI APIs.


klustria (approved):
LGTM

PR #8927, opened 2024-06-30
4.x: Oci integration fixes

tomas-langer:
Fixed missed configured option in OciConfig.
Using a custom webclient request to check if IMDS endpoint is valid (as the default that used Region discovery has hardcoded 7 retries and increasing delays between them).
Added config-yaml as a runtime dependency, as without it the OCI configuration does not work.


klustria (approved):
LGTM

PR #8941, opened 2024-07-07
4.x: Remove unnecessary fields from Http1ClientResponseImpl

Captain1653:
### Description
I've found some unnecessary fields in `Http1ClientResponseImpl`. It's not a public class and these fields can be removed without breaking compatibility.
Screenshot 2024-07-07 at 17 11 36


romain-grecourt (approved):

Captain1653:
Resolved conflicts after [this commit](https://github.com/helidon-io/helidon/commit/00e114ba9b3fdb6368a6dc455465597d94ab131a).

romain-grecourt (dismissed):

tomas-langer:
@Verdent  could you please check this? The fact that we ignore the trailers if present may be a bug.

Verdent:
@Captain1653 Hi, please do not remove this field. It is absolutely correct it should be used in some way, but not removed. I will create an issue to investigate how it should be properly used and to add more per spec tests.

Captain1653:
@Verdent Sure, no problem. Then I'll close my PR.

Issue #8931, opened 2024-07-01
Enhancement : Helidon Jersey Connector :  Not reading the response fully  causes connection to be discarded 


vasanth-bhat:
## Environment Details
* Helidon Version: 4.0.8
* Helidon SE or Helidon MP  : Both
* JDK version: 21 
* OS:  Linux 

Helidon connector does not reuse connection , when the Response is not Full Read , even when the client program explicitly closes the response.  We again end up creating a new request  each connection. Even when we are reusing the same Jersey client instance in the client code across requests.     

Code Snippet is something like below where  just the status is read from response and  response is closed..
 
public  int  ping(String url) {  
         int status;      
         WebTarget webTarget = client.target(url); 
         try(Response response = webTarget.request().accept(MediaType.APPLICATION_JSON).get()){
             status = response.getStatus();
         }
         catch( .) {        
         }
         return status;
}
  
In the profile , we see that. the Http1ConenctionCache.keepAliveConenction()  does not find the connection in the cache always ends up creating a new connection.

We debugged the issue further ,  and found that the connection was never getting added to the  cache.   
The  adding to the cache happened only when  connection.releaseResource()   is  invoked.  This is invoked only when the response is fully read. ( entityFullyRead =true).  rr in case  where there is  no response body. (Content-Length ‘0’).  All other cases    connection.closeResource() Is invoked which closes the connection. 

    @Override
    public void close() {
        if (closed.compareAndSet(false, true)) {
            try {
                if (headers().contains(HeaderValues.CONNECTION_CLOSE)) {
                    connection.closeResource();
                } else {
                    **if (entityFullyRead || entityLength == 0) {**
                        connection.releaseResource();
                    } else {
                        connection.closeResource();
                    }
                }
            } finally {
                whenComplete.complete(null);
            }
        }
    }

vasanth-bhat:
To test further we  added additional line to read and discard the response. 
 
public  int  check(String url) {  
         int status;      
         WebTarget webTarget = client.target(url); 
         try(Response response = webTarget.request().accept(MediaType.APPLICATION_JSON).get()){
             status = response.getStatus();
             **String responseStr = response.readEntity(String.class);**
         }
         catch( .) {        
         }
         return status;
}
 
Now. We see that connections are re-used and. there is improvement in performance.  Below is  supporting data  from our tests. We see 29% improvement in throughput while using 22.6% less CPU.

vasanth-bhat:
[...]

PR #8943, opened 2024-07-08
Attempt to read an unconsumed response entity to allow connection caching

spericas:
### Description

Attempt to read an unconsumed response entity to allow connections to be re-used. Optimization only works for length-prefixed entities and when the entity has not been partially read --the goal is not to block the thread attempting to close the response while trying to improve connection caching. Issue #8931.

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
@@ -205,9 +205,7 @@ class Http1ClientResponseImpl implements Http1ClientResponse {
             return false;
         }
         try {
-            for (long i = 0; i < entityLength; i++) {
-                reader.read();
-            }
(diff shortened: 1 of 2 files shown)


webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -186,6 +188,34 @@ ClientConnection connection() {
...
+    private boolean consumeUnreadEntity() {
+        if (entityLength == ENTITY_LENGTH_CHUNKED) {
+            return false;
+        }
+        DataReader reader = connection.reader();
+        if (reader.available() != entityLength) {
+            return false;
+        }
+        try {
+            for (long i = 0; i < entityLength; i++) {
```
tomas-langer, line 208: reading by single bytes is very inefficient, you can use:
```
reader.skip(entityLength);
```

tomas-langer (approved):

PR #8947, opened 2024-07-09
Added EnvReader class for environment variable parsing

kinsleykajiva:
### Description

This PR introduces a new utility class, `EnvReader`, to the `io.helidon.config` package. The `EnvReader` class provides methods to read and parse environment variables from a file. This is particularly useful for applications that need to load configuration data from environment variables defined in a file.

Key features of the `EnvReader` class include:

- Reading environment variables from a file where each line represents an environment variable in the format `KEY=VALUE`.
- Filtering out any lines that start with `#` (commonly used for comments) and any lines that do not contain `=` (as they do not represent valid environment variables).
- Resolving self-references in values (e.g., `${OTHER_KEY}`) using the `resolveSelfReferences` method.
- Wrapping any `IOException` that occurs during file reading into a `RuntimeException` and rethrowing it.

This enhancement improves the flexibility of environment variable handling in the Helidon Config module and provides a robust solution for parsing environment variable files.

### Documentation

The `EnvReader` class is used to read and parse environment variables from a file. Here is an example of how to use it:

```java
Map<String, String> env = EnvReader.readEnvFile(".env");
for (Map.Entry<String, String> entry : env.entrySet()) {
    System.setProperty(entry.getKey(), entry.getValue());
}
```

In this example, the `readEnvFile` method reads environment variables from the `.env` file and returns a `Map` where each key-value pair represents an environment variable and its value. The `System.setProperty` method is then used to set each environment variable.

For more detailed information and examples, please refer to the `EnvReader` class documentation.

### DIFF
--- a/config/config/src/main/java/io/helidon/config/ConfigValues.java
+++ b/config/config/src/main/java/io/helidon/config/ConfigValues.java
@@ -16,11 +16,17 @@
 
 package io.helidon.config;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Supplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import io.helidon.common.GenericType;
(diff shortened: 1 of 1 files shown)


romain-grecourt:
Hello,

If you are interested in contributing to Helidon, make sure to sign the [OCA](https://oca.opensource.oracle.com/).

The support for environment variables and properties is built-in the core config module (I.e. io.helidon.config:helidon-config) because they are provided out of the box with the Java API (I.e. `System.getenv()`, `System.getProperties()`, `java.util.Properties`).

The class you are introducing (`EnvReader`) does not fit in `ConfigValues`. `ConfigValues` strictly provides factory methods for `ConfigValue` where `ConfigValue` is an `Optional` like API for config values.

See below a sketch for supporting `.env` files, this is more substantial that what you have submitted. Let us know if you are interested and we can create an issue to track the work.

---

### DotEnv

The `.env` file extension is associated with [dotenv](https://github.com/motdotla/dotenv), a popular NodeJS module.
There are implementations in various languages, E.g.
- https://github.com/theskumar/python-dotenv
- https://github.com/cdimascio/dotenv-java
- https://github.com/bkeepers/dotenv
- https://github.com/fireproofsocks/dotenvy

Here is the best description of the format that I found: https://hexdocs.pm/dotenvy/dotenv-file-format.html

### ConfigParser

In order to provide support for this format, you would need to create a new module that provides an implementation of `io.helidon.config.spi.ConfigParser`.

A config parser is essentially code that can parse the content of a file into a config node and that is associated with media types and file extensions. Here I think you would use `application/x-env` as the media type and `env` as the supported file extension.

See the following classes for inspiration:
- [YamlConfigParser](https://github.com/helidon-io/helidon/blob/4.0.10/config/yaml/src/main/java/io/helidon/config/yaml/YamlConfigParser.java)
- [PropertiesConfigParser](https://github.com/helidon-io/helidon/blob/main/config/config/src/main/java/io/helidon/config/PropertiesConfigParser.java)

There will be a few details to hash out:
- There is already a built-in mechanism for variable expansion, is it compatible with `.env` ?
- The variable names should follow the rules describes in [EnvironmentVariables](https://github.com/helidon-io/helidon/blob/4.0.10/config/config/src/main/java/io/helidon/config/EnvironmentVariables.java)

### Usage

```java
// parser is found with the file extension
Config config = Config.create(ConfigSources.file("/etc/foo.env"));
```

```java
// parser is found with the file extension
Config config = Config.create(ConfigSources.classpath("foo.env"));
```

```java
// parser is found with the media type
MediaType mt = MediaTypes.create("application/x-env");
Config config = Config.create(ConfigSources.create("""
FOO=BAR
""", mt));
```

```java
// Using a factory method provided by the public API for the dotenv config module
Config config = Config.create(DotEnvConfigSource.create("/etc/foo.env"));
```

PR #8992, opened 2024-07-16
4.0.11: Skip test if InstancePrincipal UT if Imds is available

barchetta:
### Description

Backport of #8985 to 4.0.11

TestInstancePrincipalsAuthenticationDetailsProvider is a negative test that validates a failure will take place if authentication is done on an environment with no Imds avaialable like an OCI instance, hence this will fail if the test is executed on an environment where Imds is available.

### Documentation

No impact


klustria (approved):
Looks great!

PR #8997, opened 2024-07-16
Add RW locking to better manage concurrency

### LINKED ISSUE #8980: 4.x `MeterRegistry#meters()` can throw `IllegalStateException`

## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP SE
* JDK version:
* OS:

----------

## Problem Description

Occasionally, a user is seeing code that has invoked `MeterRegistry#meters()` throw an exception like the one below.

Presumably, this is because a meter has been registered by another thread while the `meters()` method is running.

Helidon should protect against this.

```
java.lang.IllegalStateException: Accept exceeded fixed size of 104
	at java.base/java.util.stream.Nodes$FixedNodeBuilder.accept(Nodes.java:1233)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.HashMap$ValueSpliterator.forEachRemaining(HashMap.java:1787)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
	at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
	at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:622)
	at java.base/java.util.stream.ReferencePipeline.toList(ReferencePipeline.java:627)
	at io.helidon.metrics.providers.micrometer.MMeterRegistry.meters(MMeterRegistry.java:232)
```
## Steps to reproduce

tjquinno:
### Description
Resolves #8980 

The Helidon metrics implementation which uses Micrometer involves several interrelated data structures. Updates to these have always been protected by a lock, but the referenced issue reveals an oversight: read-only operations (which did not participate in locking) could fail in odd ways because the backing data structure for a returned list or set could change.

This PR enhances the locking scheme to use a `ReadWriteReentrantLock`. 

Notable points:
* Methods which read from the internal data structures now use the read lock.
* Methods which update the data structures now use the write lock.
* One method, `getOrCreate`, starts with a read lock and "promotes" to the write lock as needed. Of course `RWRL` does not itself support that but the changes to `getOrCreate` do. See the comments in the revision to the code for details.

### Documentation
Bug fix; no doc update.

### DIFF
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
@@ -560,11 +560,7 @@ class MMeterRegistry implements io.helidon.metrics.api.MeterRegistry {
         if (!isMeterEnabled(builder.name(), builder.tags(), builder.scope())) {
 
             io.helidon.metrics.api.Meter result = metricsFactory.noOpMeter(builder);
-            try {
-                onAddListeners.forEach(listener -> listener.accept(result));
-            } catch (Exception e) {
-                // Ignored.
-            }
+            onAddListeners.forEach(listener -> listener.accept(result));
             return result;
         }
         return null;
@@ -637,17 +633,14 @@ class MMeterRegistry implements io.helidon.metrics.api.MeterRegistry {
         }
 
         /*
-         "Promote" our lock from read to write.
+         Effectively, "promote" our lock from read (which we acquired a few lines above) to write. Because ReentrantReadWriteLock
+         does not actually support promoting, we have to release the read lock (which we did just above), acquire the write
+         lock, and recheck what we checked earlier while we had the read lock.
          */
 
         lock.writeLock().lock();
 
         try {
-            /*
-             Now that we have the write lock, check again to make sure the meter has not been registered since we checked
-             above. Things could have changed between when we released the read lock and acquired the write lock.
-             No need to check again if the meter is disabled. That is based on config which does not change at runtime.
-             */
             io.helidon.metrics.api.Meter previouslyRegisteredMeter = meterIfRegistered(mBuilder, id);
             if (previouslyRegisteredMeter != null) {
                 return previouslyRegisteredMeter;
@@ -786,7 +779,15 @@ class MMeterRegistry implements io.helidon.metrics.api.MeterRegistry {
             if (nativeMeter != null) {
                 MMeter<?> result = meters.get(nativeMeter);
                 delegate.remove(nativeMeter);
-                onRemoveListeners.forEach(listener -> listener.accept(result));
+                onRemoveListeners.forEach(listener -> {
+                    try {
+                        listener.accept(result);
+                    } catch (Exception ex) {
+                        LOGGER.log(Level.WARNING,
+                                   "Error invoking onRemoveListener " + listener.getClass().getName() + "; continuing",
+                                   ex);
+                    }
+                });
                 return Optional.of(result);
             }
             return Optional.empty();
@@ -820,7 +821,15 @@ class MMeterRegistry implements io.helidon.metrics.api.MeterRegistry {
                 scopeMembers.remove(removedHelidonMeter);
             }
         });
-        onRemoveListeners.forEach(listener -> listener.accept(removedHelidonMeter));
+        onRemoveListeners.forEach(listener -> {
+            try {
+                listener.accept(removedHelidonMeter);
+            } catch (Exception ex) {
+                LOGGER.log(Level.WARNING,
+                           "Error invoking onRemoveListener " + listener.getClass().getName() + "; continuing",
+                           ex);
+            }
+        });
         return removedHelidonMeter;
     }
 
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
@@ -25,7 +25,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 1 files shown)


metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -215,39 +221,65 @@ static MMeterRegistry applyMetersProvidersToRegistry(MetricsFactory factory,
...
     }
 
     @Override
     public Collection<io.helidon.metrics.api.Meter> meters(Predicate<io.helidon.metrics.api.Meter> filter) {
-        return meters.values()
-                .stream()
+        lock.readLock().lock();
+        List<io.helidon.metrics.api.Meter> temp;
+        try {
+            temp = new ArrayList<>(meters.values());
```
ljnelson, line 257: Hmm; this idiom with the streaming below it seems verbose. What about:
```
return List.copyOf(meters.values()).stream()
    .map(...)
    ....
```
…all here under lock instead?

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -215,39 +221,65 @@ static MMeterRegistry applyMetersProvidersToRegistry(MetricsFactory factory,
...
                 .toList();
+
     }
 
     @Override
     public Iterable<String> scopes() {
-        return scopeMembership.keySet();
+        lock.readLock().lock();
+        try {
+            return new HashSet<>(scopeMembership.keySet());
```
ljnelson, line 272: Is it OK if someone casts the returned `Iterable` to a `Set` and starts adding things to it? If not, consider:
```
return Set.copyOf(scopeMembership.keySet());
```

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -343,16 +379,21 @@ io.micrometer.core.instrument.MeterRegistry delegate() {
...
+        lock.readLock().lock();
+        try {
+            if (scopeSelection.iterator().hasNext()) {
+                Set<io.helidon.metrics.api.Meter> result = new HashSet<>();
+                for (String scope : scopeSelection) {
+                    if (scopeMembership.containsKey(scope)) {
+                        result.addAll(scopeMembership.get(scope));
+                    }
                 }
+                return result;
```
ljnelson, line 391: Watch mutability.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -512,10 +552,47 @@ void onMeterRemoved(Meter removedMeter) {
...
     }
 
+    private io.helidon.metrics.api.Meter noopMeterIfDisabled(io.helidon.metrics.api.Meter.Builder<?, ?> builder) {
+        if (!isMeterEnabled(builder.name(), builder.tags(), builder.scope())) {
+
+            io.helidon.metrics.api.Meter result = metricsFactory.noOpMeter(builder);
+            try {
+                onAddListeners.forEach(listener -> listener.accept(result));
+            } catch (Exception e) {
+                // Ignored.
```
ljnelson, line 566: Ding ding ding. My alarm bells go off; worthy of a log at least I would think

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -546,20 +625,34 @@ private <M extends Meter, HB extends MMeter.Builder<?, M, HB, HM>, HM extends MM
...
+                return (HM) foundMeter;
             }
-            return (HM) foundMeter;
+        } finally {
+            lock.readLock().unlock();
         }
 
-        lock.lock();
+        /*
+         "Promote" our lock from read to write.
```
ljnelson, line 640: …although you don't have any lock at this point?

ljnelson (commented):
Couple of gotchas

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tjquinno: Copying the array inside the lock but doing the rest of the work outside lets us release the lock sooner.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tjquinno: We've copied the elements into a new set that is not backed by the original, so the calling code can do what it wants with the returned set without disturbing our internal set.

Or did I miss your point?

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tjquinno: As above, we create a new result set that is not backed by our internal set, so I'm not sure why we'd restrict what the caller can do with the return value.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tjquinno: Well, not at that moment--hence the quotes around "promote" because in fact just a few lines above we _had_ acquired the read lock and the effect we want is to promote the lock from read to write as the comment a few lines below explains. 

I've rewritten the comments to remove any possible confusion.

spericas: LGTM. Acquiring WL and then checking condition again to account for the interim no lock was held.

spericas (approved):

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tjquinno: [I responded to this comment earlier but maybe I didn't save it.]

I've removed the try/catch block. Everywhere else where we invoke the listeners we (intentionally) let any exceptions bubble up. We should do the same here.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
ljnelson: I just meant: what are the semantics of the return value? I tend to prefer: "the return value is immutable". You're saying, I think: "the return value's mutability is undefined". Either can be true; just wanted clarification.

Issue #9022, opened 2024-07-22
Helidon Webclient (4.0.10) is not routing the requests through proxy configured using Proxy Builder. 


shashikumarmh77:
## Environment Details
* Helidon Version: 4.0.10
* Helidon SE or Helidon MP - Helidon SE
* JDK version: 21
* OS: Linux
* Docker version (if applicable):

----------

## Problem Description
I am trying to configure [MockServer](https://www.mock-server.com/) as a proxy to facilitate testing of REST requests.
In the following simplified test case,  http://echo.jsontest.com/1/one is invoked.  The actual endpoint returns {"1": "one"} but  MockServer configured as  proxy should return  mocked response of  {"1": "ONE"}
```
@Test
    public void helidonClientWithProxySetUsingSystem() throws ExecutionException, InterruptedException, IOException {
        final String URL = "http://echo.jsontest.com/1/one";
        var builder = WebClient.builder().followRedirects(true).relativeUris(true)
                // set proxy address.
                .proxy(Proxy.builder().type(Proxy.ProxyType.HTTP).host("localhost").port(1026))
                .baseUri(URL);
        var webClient = builder.build();
        var webClientRequestBuilder = webClient.method(Method.GET);
        System.out.println("Invoking " + URL);
        try (HttpClientResponse response = webClientRequestBuilder.request()) {
            System.out.println("Response code with WebClient:" + response.status().code());
        }
    }
```
However, the execution results in below exception.

```
o.helidon.common.buffers.DataReader$InsufficientDataAvailableException
	at io.helidon.common.buffers.DataReader.pullData(DataReader.java:83)
    ...
	at io.helidon.webclient.http1.Http1StatusParser.readStatus(Http1StatusParser.java:48)
	at io.helidon.webclient.http1.Http1CallChainBase.readResponse(Http1CallChainBase.java:218)
	at io.helidon.webclient.http1.Http1CallEntityChain.doProceed(Http1CallEntityChain.java:66)
	at io.helidon.webclient.http1.Http1CallChainBase.proceed(Http1CallChainBase.java:149)
	at io.helidon.webclient.api.ClientRequestBase.invokeServices(ClientRequestBase.java:395)
	at io.helidon.webclient.http1.Http1ClientRequestImpl.invokeWithServices(Http1ClientRequestImpl.java:205)
	at io.helidon.webclient.http1.Http1ClientRequestImpl.invokeRequestWithEntity(Http1ClientRequestImpl.java:195)
[...]

PR #9023, opened 2024-07-23
4.x: Helidon Webclient (4.0.10) is not routing the requests through proxy configured using Proxy Builder. #9022

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/9022

It will do HTTP CONNECT if any of these is true:
- You explicitly set in a flag.
- It is TLS
- It uses proxy authentication

### Documentation
None

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -90,6 +90,7 @@ public class Proxy {
     private final Optional<char[]> password;
     private final ProxySelector systemProxySelector;
     private final Optional<Header> proxyAuthHeader;
+    private final boolean forceHttpConnect;
 
     private Proxy(Proxy.Builder builder) {
         this.host = builder.host();
@@ -102,6 +103,7 @@ public class Proxy {
         this.port = builder.port();
         this.username = builder.username();
         this.password = builder.password();
+        this.forceHttpConnect = builder.forceHttpConnect();
 
         if (type == ProxyType.SYSTEM) {
             this.noProxy = inetSocketAddress -> true;
@@ -451,7 +453,8 @@ public class Proxy {
     private static Socket connectToProxy(WebClient webClient,
                                          InetSocketAddress proxyAddress,
                                          InetSocketAddress targetAddress,
-                                         Proxy proxy) {
+                                         Proxy proxy,
+                                         boolean tls) {
 
         WebClientConfig clientConfig = webClient.prototype();
         TcpClientConnection connection = TcpClientConnection.create(webClient,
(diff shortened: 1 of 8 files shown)


klustria (dismissed):
This LGTM already, so I'm approving. Also let's run this through @tomas-langer if he has any further comments

webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
```diff
@@ -667,6 +677,19 @@ public Builder type(ProxyType type) {
             return this;
         }
 
+        /**
+         * Forces HTTP CONNECT with the proxy server.
+         * Otherwise it will not execute HTTP CONNECT for HTTP protocol.
```
tomas-langer, line 682: This is not exact. Please reword it to make it clear that we talk about plain vs. TLS.
Both are HTTP protocol.

webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
```diff
@@ -469,26 +472,26 @@ private static Socket connectToProxy(WebClient webClient,
...
-            request.header(HeaderValues.CONNECTION_KEEP_ALIVE)
-                .header(PROXY_CONNECTION);
-        }
-        proxy.proxyAuthHeader.ifPresent(request::header);
-        // we cannot close the response, as that would close the connection
-        HttpClientResponse response = request.request();
-        if (response.status().family() != Status.Family.SUCCESSFUL) {
-            response.close();
-            throw new IllegalStateException("Proxy sent wrong HTTP response code: " + response.status());
+        if (proxy.forceHttpConnect || tls || proxy.username.isPresent()) {
```
tomas-langer, line 475: Proxy-Connection header should be set even when not using CONNECT (at least that is what curl does, and I would think that that is correct)

tomas-langer (changes requested):

webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
jbescos: From the code we have, it seems currently we only set `Proxy-Connection` in the first request to the Proxy. This request may not happen with this new changes. Should we send always `Proxy-Connection` to every request when using proxy?.

From the spec, `Proxy-Connection` is a redundant header that has the same effect than `Connection`. So there is no point on setting it.

webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
jbescos: Done

webclient/api/src/main/java/io/helidon/webclient/api/Proxy.java
klustria: Looks like curl removed `Proxy-Connection` based on specs as you mentioned, but put it back because of a slew of bugs filed against it when it was removed: https://stackoverflow.com/questions/62722430/why-does-curl-send-a-proxy-connection-header-even-though-the-rfc-seems-to-disco. Seems that there are still lots of proxy servers out there that it s expecting this header.

jbescos: Nice finding @klustria 

So, my understanding is that if proxy is set, we need to set `Proxy-Connection` in all the requests.

klustria: Yeah, that's what I understand too.

jbescos: Done. It is excluded in HTTP2 as it is said here: https://www.rfc-editor.org/rfc/rfc9113.html

[8.2.2. ](https://www.rfc-editor.org/rfc/rfc9113.html#section-8.2.2)[Connection-Specific Header Fields](https://www.rfc-editor.org/rfc/rfc9113.html#name-connection-specific-header-)

HTTP/2 does not use the Connection header field ([Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9113.html#RFC9110)]) to indicate connection-specific header fields; in this protocol, connection-specific metadata is conveyed by other means. An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields. This includes the Connection header field and those listed as having connection-specific semantics in [Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9113.html#RFC9110)] (that is, Proxy-Connection, Keep-Alive, Transfer-Encoding, and Upgrade). Any message containing connection-specific header fields MUST be treated as [malformed](https://www.rfc-editor.org/rfc/rfc9113.html#malformed) ([Section 8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#malformed)).
The only exception to this is the TE header field, which MAY be present in an HTTP/2 request; when it is, it MUST NOT contain any value other than "trailers".An intermediary transforming an HTTP/1.x message to HTTP/2 MUST remove connection-specific header fields as discussed in [Section 7.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.6.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9113.html#RFC9110)], or their messages will be treated by other HTTP/2 endpoints as [malformed](https://www.rfc-editor.org/rfc/rfc9113.html#malformed) ([Section 8.1.1](https://www.rfc-editor.org/rfc/rfc9113.html#malformed)).

klustria (approved):
LGTM

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
```diff
@@ -179,6 +180,14 @@ protected MediaContext mediaContext() {
         return super.mediaContext();
     }
 
+    @Override
+    protected void additionalHeaders() {
+        super.additionalHeaders();
+        if (proxy() != null && proxy().type() != ProxyType.NONE) {
```
tomas-langer, line 186: proxy() must never be null - this check should be useless.

tomas-langer (changes requested):

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
jbescos: Done

klustria (approved):
Looks good!

tomas-langer (approved):

PR #9026, opened 2024-07-23
Implements the gRPC MP Client API

spericas:
### Description

Implements the gRPC MP Client API. This API uses interfaces to create client proxies. It introduces two new annotations, `@GrpcProxy` for injection and `@GrpcChannel` to associate a channel with a proxy. Client proxies can be injected into CDI beans to access remote gRPC services.

A sample client interface is shown below:

```java
    @Grpc.GrpcService("EchoService")
    @Grpc.GrpcChannel("echo-channel")
    public interface EchoServiceClient {

        @Grpc.Unary("Echo")
        void echo(String request, StreamObserver<String> observer);
    }
```

A proxy instance can be injected as follows: 

```java
    @Inject
    @Grpc.GrpcProxy
    private EchoServiceClient proxyClient;
```

### Documentation

Will be provided in a follow-up PR.


examples/microprofile/grpc/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.webserver.testing.junit5</groupId>
+            <artifactId>helidon-webserver-testing-junit5-grpc</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.logging</groupId>
+            <artifactId>helidon-logging-jul</artifactId>
+            <scope>test</scope>
```
tomas-langer, line 67: This should be a `runtime` scope in examples.

examples/microprofile/grpc/src/test/java/io/helidon/examples/microprofile/grpc/StringServiceTest.java
```diff
@@ -0,0 +1,122 @@
...
+import jakarta.inject.Inject;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasSize;
+
+@HelidonTest
+@AddExtension(GrpcMpCdiExtension.class)
```
tomas-langer, line 40: If you have discovery enabled, this should not be needed - I suppose the CDI extensions are implemented as service loader services

examples/microprofile/grpc/src/main/resources/application.yaml
```diff
@@ -0,0 +1,47 @@
...
+    private-key:
+      keystore:
+        passphrase: "password"
+        resource:
+          resource-path: "server.p12"
+
+grpc:
+  channels:
+    string-channel:
+      port: 0
```
tomas-langer, line 35: How is the host & port discovered by the client? 
Also `grpc.channels` does not mention `client`, yet it seems to be a configuration for the client. Can you please explain this choice? I would expect `grpc.client.channels` or similar

microprofile/grpc/client/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+        </dependency>
+        <dependency>
+            <!--
+            need to use this version to work around the same groupId/artifactId and
+            different package names in Jakarta
+            This is needed for the Generated annotation in grpc sources
+            -->
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <scope>provided</scope>
```
tomas-langer, line 115: `optional` is sufficient, `provided` should be removed

microprofile/grpc/client/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+            </extension>
+        </extensions>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.helidon.builder</groupId>
```
tomas-langer, line 140: Please add this also as a plugin dependency, as otherwise the reactor may not do correct ordering (only for annotation processors/codegens)

microprofile/grpc/client/pom.xml
```diff
@@ -0,0 +1,161 @@
...
+                            <artifactId>helidon-builder-processor</artifactId>
```
tomas-langer, line 141: `helidon-builder-processor` is deprecated ans should not be used, please use a combination of `helidon-codegen-apt` and `helidon-builder-codegen`

tomas-langer: Do not forget to add the helidon copyright codegen extension

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/GrpcChannelsProvider.java
```diff
@@ -0,0 +1,216 @@
...
+
+        private Builder(Config config) {
+            // Add the default channel (which can be overridden in the config)
+            channel(DEFAULT_CHANNEL_NAME, GrpcChannelDescriptor.builder().build());
+
+            if (config == null) {
+                return;
+            }
+
+            Config channelsConfig = config.get(CFG_KEY_CHANNELS);
```
tomas-langer, line 185: This is an unusual usage of config, that is very hard to document right. 
Is there a way to do this with the standard usage of Helidon builders, or at least with appropriate config metadata documentation?

examples/microprofile/grpc/src/main/resources/application.yaml
spericas: 1. There's some ugly code that attempts to find the server port when running on the same VM and no port was specified in config. Only used for testing.
2. The use of `grpc.channels` instead of something like `grpc.client.channels` is carried over from 3.X. We can change that.

microprofile/grpc/client/pom.xml
spericas: Fixed, hopefully updated correctly.

microprofile/grpc/client/src/main/java/io/helidon/microprofile/grpc/client/GrpcChannelsProvider.java
spericas: I should be possible to convert this to a Helidon builder, I need to look into that.

spericas: Pushed commit to update the configuration of channels.

examples/microprofile/grpc/src/main/resources/application.yaml
spericas: Using `grpc.client.channels` in latest commit

spericas: Removed ugly code to find server port in favor for new interface implemented by client proxies to dynamically set a new port. Tests updated.

spericas:
Example has been moved here https://github.com/helidon-io/helidon-examples/pull/65

microprofile/grpc/client/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.webclient</groupId>
+            <artifactId>helidon-webclient-grpc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-tls</artifactId>
+        </dependency>
+        <dependency>
```
tomas-langer, line 60: This dependency should never be a runtime dependency (in any scope), unless you are writing a code generation module.

tomas-langer (changes requested):

microprofile/grpc/client/pom.xml
spericas: Removed

tomas-langer (approved):

PR #9045, opened 2024-07-28
Config reference documentation update

### LINKED ISSUE #9036: 4.x Generated config metadata for an enum on a `@Prototype.Configured` type does not list the valid values
## Environment Details
* Helidon Version: 4.x
* JDK version:
* OS:

----------

## Problem Description
The generated config metadata JSON does not list the allowed values for an enum property.

## Steps to reproduce
1. Create a new `public enum` type (such as `GcTimeType` in this PR #9032). 
2. On a `@Prototype.Configured` type (`MetricsConfigBlueprint` in the PR) add a property of that enum type:
   
   ```java
    @Option.Configured
    @Option.Default("COUNTER")
    GcTimeType gcTimeType();
   ```
3. Build the component.
4. Inspect the generated `META-INF/helidon/config-metadata.json` file. It describes the enum property but does not list allowed values.

tomas-langer:
Resolves #9036

This PR is a few commits on top of each other.

1. Inroduction of a JSON reader and writer for Metadata (so we can remove dependency on JSON-P/JSON-B from our tools, such as native image extension, config metadata write, and we can replace the text format in service registry with JSON which is better suited for the task)
2. A small fix for builder codegen, so it does not generate `service.loader` file when empty
3. `config-metadata` module is re-structured, as it now has multiple child modules; I did not change group id of existing modules for backward compatibility; introduced `codegen` and `docs` modules, deprecated `processor` module
4. Fix to all pom files to use `codegen`, add plugin dependencies to correctly order Maven reactor, remove `processor` dependencies from regular Maven dependencies etc.
5. A small fix to logging-jul, to correctly print out how logging is configured even when not running in native image
6. Update to `docs` pom file, to generate the config reference documentation with profile `configdocs`
7. Update to the actual `.adoc` files based on all of the above

This is also a must-have for 4.1.0, as service registry needs to use it


tomas-langer:
Closed as this is separated into multiple PRs

PR #9053, opened 2024-07-29
4.x: Config reference documentation

### LINKED ISSUE #7791: 4.x: Only use builder API for configuration metadata for blueprints and fix processor
Refactor all blueprints to only use API from `helidon-builder-api` instead of `helidon-config-metadata`
The features are aligned. If there is something that seems not possible, please check with the team.

This should allow us to remove the config metadata API from dependencies of such projects.

Simplify the config metadata processor, as it currently must support three parts:
1. "pure" config metadata usage on any type
2. config metadata usage on blueprints (this should be removed)
3. builder API on blueprints

### LINKED ISSUE #9036: 4.x Generated config metadata for an enum on a `@Prototype.Configured` type does not list the valid values

tomas-langer:
Resolves #9036
Resolves #7791

Please review by commits. The last one changes all of the generated files (mostly whitespace changes), so it looks big...

Config reference documentation updates:

1. config-metadata module is re-structured, as it now has multiple child modules; I did not change group id of existing modules for backward compatibility; introduced codegen and docs modules, deprecated processor module
2. Fix to all pom files to use codegen, add plugin dependencies to correctly order Maven reactor, remove processor dependencies from regular Maven dependencies etc.
3. Update to docs pom file, to generate the config reference documentation with profile configdocs
4. Update to the actual .adoc files based on all of the above

This is a preparation to remove all config reference docs from github repo if desired - we can generate it together with javadocs when building documentation (which would be a similar style to generate all of our docs directly from source material, rather than from intermediately generated files)


tomas-langer:
Rebased to resolve conflict with changes done by Metrics

Verdent (approved):

romain-grecourt (approved):

Issue #8833, opened 2024-06-03
4.x: OCI archetype server module throws `UnsatisfiedResolutionException`


barchetta:
## Environment Details
* Helidon Version: 4.0.8
* Helidon MP
* JDK version: 21.0.3
----------

## Problem Description

Running the OCI archetype project server results in an exception (see below). This works with 4.0.7.  Two problems:

1. The exception does not clearly indicate what class is not resolvable.
2. The sample fails to run

```
Exception in thread "main" jakarta.enterprise.inject.UnsatisfiedResolutionException
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.helidon.integrations.oci.sdk.cdi.OciExtension.lambda$installBasicAdp$25(OciExtension.java:1176)
	at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$CreateCallback.create(BeanConfiguratorImpl.java:372)
	at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$ImmutableBean.create(BeanConfiguratorImpl.java:511)
	at org.jboss.weld.contexts.AbstractContext.get(AbstractContext.java:96)
	at org.jboss.weld.bean.ContextualInstanceStrategy$DefaultContextualInstanceStrategy.get(ContextualInstanceStrategy.java:100)
	at org.jboss.weld.bean.ContextualInstance.get(ContextualInstance.java:50)
	at org.jboss.weld.manager.BeanManagerImpl.getReference(BeanManagerImpl.java:679)
	at org.jboss.weld.bean.builtin.InstanceImpl.getBeanInstance(InstanceImpl.java:262)
	at org.jboss.weld.bean.builtin.InstanceImpl.get(InstanceImpl.java:114)
	at org.jboss.weld.util.ForwardingWeldInstance.get(ForwardingWeldInstance.java:44)
	at io.helidon.integrations.oci.sdk.cdi.OciExtension.produceClient(OciExtension.java:1325)
	at io.helidon.integrations.oci.sdk.cdi.OciExtension.lambda$installServiceClient$27(OciExtension.java:1279)
	at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$CreateCallback.create(BeanConfiguratorImpl.java:372)
	at org.jboss.weld.bootstrap.events.configurator.BeanConfiguratorImpl$ImmutableBean.create(BeanConfiguratorImpl.java:511)
[...]

PR #9073, opened 2024-07-31
Update oci.auth-strategy values in generated OCI archetype to avoid UnsatisfiedResolutionException

klustria:
### Description
The allowed values for oci.auth-strategy has changed with the occurence of "_" replaced with "-", so the use of the following needs to be changed:
1. instance_principals to instance-principals
2. config_file to config-file
3. resource_principal to resource-principal

### Documentation

Looks like the values are not mentioned in the docs so no changes are needed there.

If no doc impact: None


klustria:
Tested by generating the oci archetype from a local build with the change and found it to have resolved the exception. The below command was used to generate the project:

```
helidon init --reset --url file:///${HOME}/helidon_project/helidon-main/archetypes/archetypes/target/cli-data --flavor mp --archetype oci --batch
```

barchetta (approved):

klustria:
This PR is for resolution of issue reported in https://github.com/helidon-io/helidon/issues/8833.

PR #9080, opened 2024-08-01
4.x: Use Hson.Struct instead of Hson.Object to prevent confusion with java.lang.Object

### LINKED ISSUE #9075: 4.x: Consider better class name than Object in metadata Hson
Current we use `Hson.Object` for object abstraction in output (similar to JsonObject).
This may create confusion if somebody does an import of the type, as it would get confused with `java.lang.Object`.

As this is designed for metadata handling within Helidon, we control our destiny, but still this may be useful (if we come up with a great alternative).

tomas-langer:
Resolves #9075 

The module is for Helidon internal metadata handling only, so no docs changes.


tomas-langer:
Thanks @romain-grecourt for the idea for naming!

PR #9085, opened 2024-08-01
Mark deprecations for Micrometer integration component

### LINKED ISSUE #8946: 4.x Deprecate Micrometer integration component and add "how to use Micrometer API" content to metrics doc
## Environment Details
* Helidon Version: 4.x
* JDK version:
* OS:

----------

## Problem Description
In 3.x Helidon supports integration with Micrometer as a parallel metrics implementation. That integration does not tie in to the main Helidon metrics implementation.

In 4.x the Helidon neutral metrics API is currently implemented using Micrometer. Developers who want to use the native Micrometer API can do so directly and their use of Micrometer fully interoperates with Helidon's.

We should deprecate the 3.x-era integration component in 4.x, remove it in 5, and in 4.x we should rewrite/supplement the metrics doc to describe how developers can use the Micrometer API directly (if they want to).

tjquinno:
### Description
Resolves #8946 

Releases before 4.x provided some convenient ways developers could use Micrometer in their applications, including some configuration help and CDI injection and interceptor support for the Micrometer annotations and meter types. Using those tools, the Micrometer meters were completely separate from the Helidon metrics managed using the Helidon metrics API.

Now that Helidon 4 uses Micrometer for its own metrics implementation, the former approach is obsolete, inefficient, and unneeded. 

As one step of providing even better integration with Micrometer, this PR marks as deprecated-for-removal the public types of the integration component; we plan to remove them in a future major release of Helidon.

We will update the documentation in a later PR. In the meantime see the possible release note text below.

### Documentation
There are existing files in our docs area for Micrometer integration, but there are no links to them on our 4.x doc pages. So there are no doc updates needed as part of this PR.

(possible short-term text for release note...)

Users who want to use Micrometer from their Helidon 4.x apps can take advantage of the fact that Helidon now uses Micrometer for its metrics implementation. The following are simple approaches for using Micrometer from a Helidon 4.x app:
* Use the Helidon neutral metrics API to register, locate, and update meters. Helidon registers meters  in the Micrometer global meter registry. Then, as needed, use the `unwrap` method that is provided on most of the Helidon metrics types to cast them to their underlying Micrometer implementation counterparts and operate on those Micrometer types.
* Use the Micrometer `Metrics.globalRegistry` to access the Micrometer global registry directly. Register and retrieve meters using the Micrometer API. 
* If it makes sense, use one API from some of your code and the other API from other parts of your application. 
* Make sure you access the Helidon global registry in your app before you access the Micrometer global registry. This allows Helidon to properly initialize its internal data structures.
* Include the tag `scope=application` when you register or look up a meter using the Micrometer API.
  
  The Helidon metrics API automatically adds a `scope` tag to each meter. Normally, meters Helidon registers from your application will have `scope=application`. If you look up such a meter using the Micrometer API you will need to add that tag for Micrometer to find it.
  
  Similarly, if you register a meter using the Micrometer API be sure to specify the `scope=application` tag so the Helidon API will find it successfully.

Further, for Helidon 4 MP apps:
* Add the following dependency to your project:
  ```xml
  <dependency>
      <groupId>io.helidon.integrations.micrometer</groupId>
      <artifactId>helidon-integrations-micrometer-cdi</artifactId>
  </dependency>
  ```
* Annotate CDI bean methods with either the MP metrics annotations or the Micrometer annotations `@Timed` or `@Counted`.  Helidon automatically registers those meters and updates them each time the annotated bean methods are invoked.
* Inject the Micrometer global registry, for example:
  ```java
  @Inject
  private io.micrometer.core.instrument.MeterRegistry meterRegistry;
  ```
  and then work with the injected meter registry.

In SE and MP, meters registered using either API are accessible to both because the meters are in the Micrometer global registry regardless.

(end of possible release note)

### DIFF
--- a/integrations/micrometer/cdi/pom.xml
+++ b/integrations/micrometer/cdi/pom.xml
@@ -79,6 +79,11 @@
             <optional>true</optional>
         </dependency>
         <dependency>
+            <groupId>io.helidon.microprofile.metrics</groupId>
+            <artifactId>helidon-microprofile-metrics</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
--- a/integrations/micrometer/cdi/src/test/java/io/helidon/integrations/micrometer/cdi/MixedAnnotationsResource.java
+++ b/integrations/micrometer/cdi/src/test/java/io/helidon/integrations/micrometer/cdi/MixedAnnotationsResource.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.integrations.micrometer.cdi;
+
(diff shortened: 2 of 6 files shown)


spericas (approved):
LGTM

PR #9096, opened 2024-08-03
4.x: Replace deprecated method Span.baggage(key, value) on Span.baggage().set(key, value)

Captain1653:
### Description

I've found, that the deprecated method `Span.baggage(String key, String value)` is used in the code. The method was marked as deprecated in [this commit](https://github.com/helidon-io/helidon/commit/0b16c29c22c0691a4aade898a7467692539c6263#diff-d0ca182466860d79537658e4b49b6bd56d6fd8c492719be03a70d86e371ee60bR132)
Screenshot 2024-08-03 at 16 25 04

It can be replaced on `Span.baggage().set(String key, String value)`.

### DIFF
--- a/tracing/providers/jaeger/src/test/java/io/helidon/tracing/providers/jaeger/JaegerBaggageTest.java
+++ b/tracing/providers/jaeger/src/test/java/io/helidon/tracing/providers/jaeger/JaegerBaggageTest.java
@@ -39,8 +39,8 @@ class JaegerBaggageTest {
     @Test
     void testBaggage(){
         Span span = tracer.spanBuilder("test-span").start();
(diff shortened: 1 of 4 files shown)


tjquinno (changes requested):
Just because methods are deprecated does not mean we immediately stop testing them.

As they stand today, the tests exercise _both_ ways of accessing the baggage which is a good thing.

If you want to add tests to work with other combinations of the deprecated API and the new one that's fine. But please do not remove more of the existing code.

Captain1653:
@tjquinno  Thank you for your explanation.

I was not going to add new tests for deprecated methods. I only wanted to replace them, but after your explanation, I've understood, that is better to close this PR (it is not a problem).

PR #9104, opened 2024-08-05
Refactor DbClient integration tests

romain-grecourt:
### Description

Refactor DbClient integration tests.

- Use TestContainers, with oraclelinux based images
- Add one sub-module per database, add modules for h2, mysql, pgsql, oracle, mongo
- Add missing bom and all entry for helidon-integrations-db-pgsql

Refactor tests/integration/harness
- Removed all the suite support (potentially add back later in a different shape)
- Reworked `HelidonProcessRunner` into a more general `ProcessRunner` API

### Documentation

None.


barchetta (approved):

Tomas-Kraus:
Do we really want to maintain such a huge amount of files for every single database vendor? I need to add few tests after fixing one issue in DbClient and it's terrible.

PR #9119, opened 2024-08-08
Add context propagation to the Zipkin tracing provider

### LINKED ISSUE #8848: Zipkin (and OpenTracing) tracing providers do not provide context propagation
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
The Jaeger and OpenTelemetry tracing providers both provide context propagation.

But neither the OpenTracing provider nor the Zipkin provider does so. 

Addressing #8846 would allow the Zipkin provider to do propagation very similarly to how our OTel provider does. That change plus this one would put Zipkin on an equal footing with OTel and Jaeger as providers.

Given that OpenTracing is essentially retired we might not choose to invest the time to add propagation for it. But Zipkin probably should do propagation.

tjquinno:
### Description
Resolves #8848 for Zipkin.

Changes:
1. Add a context propagator implementation for Zipkin. This permits Helidon to propagate Zipkin trace information across threads within a Helidon server. (That's distinct from trace propagation via HTTP headers which is handled in other code and already worked prior to this PR.)
2. Correct the handling of the global tracer in the Helidon OpenTracing provider component. 
   
   Our Zipkin tracing provider component relies to some extent on our OpenTracing provider types. If a user includes the Zipkin tracing provider then OpenTracing comes along and our Zipkin implementation uses OpenTracing.
   
   Previously, invoking `OpenTracingTracerProvider#global()` would return a new Helidon `Tracer` instance every time. Those instances would all wrap the single OpenTracing global tracer (whatever it is), but that's inefficient and caused failures in an existing propagation test in the `provider-tests` component. That test had been excluded for Zipkin because of the prior lack of propagation support. 
   
   This PR changes two aspects of the global tracer handling in OpenTracing:
   1. Use `LazyValue` for the Helidon global tracer maintained by our OpenTracing provider so we reuse it appropriately rather than creating a new one on each invocation of `global()`.
   2. Both in the `LazyValue` factory lambda and in the `OpenTracingTracerProvider#global(Tracer)` method not only set or update the Helidon global tracer lazy value but also manage OpenTracing's similar mechanism. OpenTracing lets us supply  code we want OpenTracing to run to create the OpenTracing global tracer. By specifying our code for this work we make sure our global Helidon tracer and the delegate global tracer in OpenTracing are in step.
3. Remove the exclusion in the Zipkin provider `pom.xml` of the common context propagation test. The pom does now  invoke it in a separate surefire execution to make sure the test has a functioning global tracer undisturbed by other tests.
4. (Otherwise untouched for this enhancement) Add public constructor to `ZipkinTracerProvider` to clear the Javadoc warning.
### Documentation
Bug fix. Our tracing documentation does not discuss context propagation across threads within a server, so this PR has no doc impact.

### DIFF
--- a/tracing/providers/zipkin/src/main/java/io/helidon/tracing/providers/zipkin/ZipkinDataPropagationProvider.java
+++ b/tracing/providers/zipkin/src/main/java/io/helidon/tracing/providers/zipkin/ZipkinDataPropagationProvider.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.tracing.providers.zipkin;
+
+import io.helidon.common.context.Contexts;
+import io.helidon.common.context.spi.DataPropagationProvider;
+
+import io.opentracing.Scope;
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.util.GlobalTracer;
+
+/**
+ * Data propagation provider for the Helidon Zipkin tracing provider.
+ */
+public class ZipkinDataPropagationProvider implements DataPropagationProvider<ZipkinDataPropagationProvider.ZipkinContext> {
+
+    private static final System.Logger LOGGER = System.getLogger(ZipkinDataPropagationProvider.class.getName());
+
+    @Override
+    public ZipkinContext data() {
+        Tracer tracer = Contexts.context()
+                .flatMap(ctx -> ctx.get(Tracer.class))
+                .orElseGet(GlobalTracer::get);
+        Span span = tracer.activeSpan();
+        return new ZipkinContext(tracer, span);
+    }
+
+    @Override
+    public void propagateData(ZipkinContext data) {
+        if (data != null && data.span != null) {
+            data.scope = data.tracer.activateSpan(data.span);
+        }
+    }
+
+    @Override
+    public void clearData(ZipkinContext data) {
+        if (data != null && data.scope != null) {
+            try {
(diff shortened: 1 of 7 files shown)


tjquinno:
I have asked Tomas about the test failure. Basically, now that Zipkin is properly doing context propagation the test which is failing is finding tracing information in the context where it does not expect it (because our Zipkin code did not do propagation previously).

tracing/providers/zipkin/src/main/java/io/helidon/tracing/providers/zipkin/ZipkinDataPropagationProvider.java
```diff
@@ -0,0 +1,74 @@
...
+import io.opentracing.Span;
+import io.opentracing.Tracer;
+import io.opentracing.util.GlobalTracer;
+
+/**
+ * Data propagation provider for the Helidon Zipkin tracing provider.
+ */
+public class ZipkinDataPropagationProvider implements DataPropagationProvider<ZipkinDataPropagationProvider.ZipkinContext> {
+
+    private static final System.Logger LOGGER = System.getLogger(ZipkinDataPropagationProvider.class.getName());
```
tomas-langer, line 31: There should be a public constructor (explicit and documented) if this is a service loader service.

tracing/providers/zipkin/src/main/java/io/helidon/tracing/providers/zipkin/ZipkinDataPropagationProvider.java
```diff
@@ -0,0 +1,74 @@
...
+            } catch (Exception e) {
+                LOGGER.log(System.Logger.Level.TRACE, "Cannot close tracing span", e);
+            }
+        }
+    }
+
+    /**
+     * Zipkin-specific propagation context.
+     */
+    public static class ZipkinContext {
```
tomas-langer, line 63: It seems that this class does not need to be public.

tracing/providers/zipkin/src/main/java/io/helidon/tracing/providers/zipkin/ZipkinDataPropagationProvider.java
tjquinno: Changed.

tracing/providers/zipkin/src/main/java/io/helidon/tracing/providers/zipkin/ZipkinDataPropagationProvider.java
tomas-langer: But in this case we can live with it, because it is only used in a class that must be public for service loader.

tomas-langer (approved):

Issue #7552, opened 2023-09-08
JPA - persistence.xml overridable by config 


danielkec:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon MP

----------

## Problem Description
It would make testing with JPA more user-friendly if persistence.xml was overridable by config, for example:
```
@HelidonTest
@AddConfig(key="oracle.ucp.jdbc.PoolDataSource.ds1.URL",
           value = "jdbc:h2:mem:test;MODE=Oracle;INIT=runscript from 'classpath:/META-INF/init_script.sql'")
@AddConfig(key="oracle.ucp.jdbc.PoolDataSource.ds1.connectionFactoryClassName",
           value = "org.h2.jdbcx.JdbcDataSource")
@AddConfig(key="oracle.ucp.jdbc.PoolDataSource.ds1.user", value = "sa")
@AddConfig(key="oracle.ucp.jdbc.PoolDataSource.ds1.password", value = "${EMPTY}")

@AddConfig(key="jpa.pu1.hibernate.dialect", value = "org.hibernate.dialect.H2Dialect")
@AddConfig(key="jpa.pu1.hibernate.show_sql", value = "true")
@AddConfig(key="jpa.pu1.hibernate.hbm2ddl.auto", value = "update")
class MainTest {
```

danielkec:
Something like this:
```diff
Index: integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/EntityManagerFactories.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/EntityManagerFactories.java b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/EntityManagerFactories.java
--- a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/EntityManagerFactories.java	(revision 2366fbc90dd0fe551ff69cf7ba217e1716e01729)
+++ b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/EntityManagerFactories.java	(date 1694178645722)
@@ -24,7 +24,7 @@
 import java.util.Set;
 import java.util.logging.Level;
[...]

PR #9131, opened 2024-08-12
4.x: JPA testing-related improvements

ljnelson:
This PR introduces MicroProfile Config as a dependency of the JPA CDI integration and uses it to blend MicroProfile Config properties with persistence unit properties. It also allows a user to specify an alternate location for a JPA descriptor, which can be useful in unit testing scenarios.

May close and/or address #7552.

### DIFF
--- a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
+++ b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
@@ -102,6 +102,8 @@ import jakarta.xml.bind.JAXBContext;
 import jakarta.xml.bind.JAXBException;
 import jakarta.xml.bind.Unmarshaller;
 
+import org.eclipse.microprofile.config.ConfigProvider;
+
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_AFTER;
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_BEFORE;
 
@@ -316,7 +318,10 @@ public class JpaExtension implements Extension {
         if (LOGGER.isLoggable(Level.FINER)) {
             LOGGER.entering(cn, mn);
         }
-        this.enabled = Boolean.parseBoolean(System.getProperty(this.getClass().getName() + ".enabled", "false"));
+        this.enabled =
+            Boolean.parseBoolean(ConfigProvider.getConfig()
+                                 .getOptionalValue(this.getClass().getName() + ".enabled", String.class)
+                                 .orElse("false"));
         if (LOGGER.isLoggable(Level.FINE) && !this.enabled) {
             LOGGER.logp(Level.FINE, cn, mn, "jpaExtensionDisabled", this.getClass().getName());
         }
--- a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
+++ b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
@@ -101,7 +101,6 @@ import jakarta.persistence.spi.PersistenceUnitInfo;
 import jakarta.xml.bind.JAXBContext;
 import jakarta.xml.bind.JAXBException;
 import jakarta.xml.bind.Unmarshaller;
-
 import org.eclipse.microprofile.config.ConfigProvider;
 
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_AFTER;
--- a/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
+++ b/integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
@@ -34,6 +34,7 @@ import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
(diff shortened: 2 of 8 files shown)


ljnelson:
Still needs documentation updates but before I document it I'll see what people have to say.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
```diff
@@ -1379,12 +1382,19 @@ private static JtaExtendedEntityManager produceJtaExtendedEntityManager(Instance
             }
         }
         SynchronizationType finalSyncType = syncType == null ? SYNCHRONIZED : syncType;
+        Config config = getOrDefault(instance.select(Config.class), selectionQualifiers);
+        Map<String, Object> stackedProperties =
+            new StackedMap<>(List.of(properties.keySet()::stream,
+                                     () -> stream(config.getPropertyNames()::spliterator,
```
tjquinno, line 1388: Not sure there's a good alternative, given that annoyingly (but understandably)  `Config#getPropertyNames` is [not guaranteed](https://github.com/eclipse/microprofile-config/blob/8e5f457bf1a16ff631f4ced0486e7a50e53247a9/api/src/main/java/org/eclipse/microprofile/config/Config.java#L235-L239) to be complete. Practically speaking the config sources present might indeed behave well.

If there were a good way to anticipate what the interesting property names would be--if present--so the code could ask for values explicitly that'd work around this corner case. I do not pretend to know all the details of what the code is accomplishing here to know if that's feasible.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
```diff
@@ -1455,23 +1463,30 @@ private static EntityManagerFactory produceEntityManagerFactory(Instance<Object>
         } catch (final ClassNotFoundException classNotFoundException) {
 
         }
+        Config config = getOrDefault(instance.select(Config.class), namedSelectionQualifiers);
+        properties =
+          new StackedMap<>(List.of(properties.keySet()::stream,
+                                   () -> stream(config.getPropertyNames()::spliterator,
```
tjquinno, line 1469: Similar thoughts about `getPropertyNames` as above.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
```diff
@@ -1481,12 +1496,20 @@ private static JtaEntityManager produceJtaEntityManager(Instance<Object> instanc
...
+            } else if (beanQualifier != Any.Literal.INSTANCE && !SyntheticJpaQualifiers.INSTANCE.contains(beanQualifier)) {
                 containerManagedSelectionQualifiers.add(beanQualifier);
                 selectionQualifiers.add(beanQualifier);
             }
         }
         SynchronizationType finalSyncType = syncType == null ? SYNCHRONIZED : syncType;
+        Config config = getOrDefault(instance.select(Config.class), selectionQualifiers);
+        Map<String, Object> stackedProperties =
+            new StackedMap<>(List.of(properties.keySet()::stream,
+                                     () -> stream(config.getPropertyNames()::spliterator,
```
tjquinno, line 1508: Similar re: `Config#getPropertyNames`.

tjquinno (commented):
Small changes (no functional difference) and a question about `Config#getPropertyNames`.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
```diff
@@ -316,7 +317,10 @@ public JpaExtension() {
         if (LOGGER.isLoggable(Level.FINER)) {
             LOGGER.entering(cn, mn);
         }
-        this.enabled = Boolean.parseBoolean(System.getProperty(this.getClass().getName() + ".enabled", "false"));
+        this.enabled =
+            Boolean.parseBoolean(ConfigProvider.getConfig()
+                                 .getOptionalValue(this.getClass().getName() + ".enabled", String.class)
+                                 .orElse("false"));
```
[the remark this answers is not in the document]
ljnelson: Sadly the contract of a MicroProfile `Converter` is all but undefined, so it's not guaranteed what will happen in the presence of `null` or empty `String` values. The contract of `Boolean#parseBoolean(boolean)`, on the other hand, is rigorous.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/PersistenceExtension.java
ljnelson: Right. The particular application in this case is: we are making a `Map` for supplying to a specific clunky JPA API (the persistence properties supplied to `createEntityManager`). The fact that these properties are represented in the JPA API as a `Map` is a pretty bad design choice, since for virtually all use cases other than logging or debugging the `Map` will be used by the persistence provider only as a device to "probe" for a value for a particular property known to it. Because of the contract of `Map`, however, we need to report its `keySet` and its `size` and so on, even though for most use cases this will never be called/used (the persistence provider will be looking for specific properties, so will, realistically, be calling `get(String)` and not much else). So not much else to do here, I think.

tjquinno (approved):

ljnelson:
I'll follow this up with additional PRs around documentation and testing; thanks for the approval.

PR #9155, opened 2024-08-16
Implements support for client gRPC channel injections

spericas:
### Description

Implements support for client gRPC channel injections (as in 3.x). Does not use CDI qualifiers for forward compatibility with our declarative API (annotations will be reused). New test that shows field and constructor injection.

### Documentation

Included as part of this PR.


tjquinno (changes requested):
A typo and possibly a change in the test code.

tjquinno (approved):

PR #9174, opened 2024-08-20
3.x: Upgrade jakarta inject, interceptor and transaction apis

barchetta:
### Description

Upgrades EE specs to the latest in the EE 9 platform:

* Inject: 2.0.0 to 2.0.1
* Interceptor: 2.0.0 to 2.0.1
* Transaction: 2.0.0 to 2.0.1

This results in the interceptor module name change from `jakarta.interceptor.api` to `jakarta.interceptor`

Also looks like JTA had module info changes requiring a dependency on CDI.


romain-grecourt (approved):

barchetta:
Due to the module name changes -- which can be considered disruptive -- I am closing this PR. If we ever do a 3.3.0 then we can consider these upgrades at that time.

PR #9193, opened 2024-08-22
4.x: Global instance handling change (and relevant changes to testing)

### LINKED ISSUE #9186: 4.x: Global values update
We have a few types that use static singletons to store "global" instances - configuration, service registry, metrics.

Introduce a new common type `io.helidon.common.GlobalInstances` to handle such singletons, that has a capability to be cleared.
This is a temporary solution - we want to move all instances to `ServiceRegisty`, and use the `GlobalServiceRegistry` to obtain them instead, as service registry can manage lifecycle and provides basic inversion of control, to obtain config etc.

Context approach was not working nicely, as you may access the global from a custom context (such as a server request), and then it would not be a global.
~~For testing purposes, it would be really great to have these instances bound to Helidon Context, instead of as singletons.~~
~~Proposal:~~
~~create `ContextValue` in `io.helidon.common.context` that would take care of this~~
~~register the values in either the global config (keeping current behavior of global singleton at runtime), or with a current ~~context (or its parent hierarchy) that has an ID starting with `test-`, to support test-scoped singletons~~
~~create an extension for testing that runs everything (before all, after all, each test method, constructor) in Helidon and create a new context for each test class being executed (i.e. all methods within a class share the same context)~~
~~modify existing "globals" to use the new approach~~

tomas-langer:
Resolves #9186

- using `ContextValue` that uses `Context` as the "backing" store
- modified Metrics, ServiceRegistry, Config, and MapperManager to use `ContextValue`
- `ContextValue` either uses the test context, or `global` context; this is achieved through a context registered under a known qualifier, to make sure it all works even when running in child context

Introduction of new test annotations
- configuration setup using annotations for Helidon SE (through service registry)
- static configuration update (if value was not yet read, and is not provided by existing config sources)
- introduction of new JUnit5 annotation + extension (resets global instances)

- Fix security's ClassToInstanceStore (discovered an issue while copying some ideas to GlobalInstances) 
- Add support for ExistingInstanceDescriptor to service registry (for testing purposes - bind an instance without a service descriptor) 
- Testing JUnit extensions now extend the new extension, so all tests inherit the "nice" behavior of global instances SE testing now supports ServiceRegistry (and is ready for future setup using Helidon Inject)


testing/testing/src/main/java/io/helidon/testing/TestConfig.java
```diff
@@ -0,0 +1,243 @@
...
+         */
+        Value[] value();
+    }
+
+    /**
+     * Custom configuration key/value pairs.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.TYPE)
+    public @interface Values {
```
romain-grecourt, line 146: How about `Block` to follow the current pattern used in MP testing?

common/common/src/main/java/io/helidon/common/GlobalInstances.java
```diff
@@ -0,0 +1,141 @@
...
+ * This is a temporary solution used before we migrate all our global singletons to service registry.
+ *
+ * @deprecated create a {@code io.helidon.service.registry.Service.Provider} to handle a registry service, and
+ *              then use {@code io.helidon.service.registry.GlobalServiceRegistry#registry()} to get an instance.
+ *              This will work nicely in testing (as global singletons are hard to re-set).
+ */
+@Deprecated(forRemoval = true, since = "4.2.0")
+@SuppressWarnings("unchecked")
+public final class GlobalInstances {
+    private static final Map<Class<?>, GlobalInstance> BACKING_MAP = new IdentityHashMap<>();
```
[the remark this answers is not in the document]
tomas-langer: It does not. Parallel testing will not work with any of our global singletons (even now).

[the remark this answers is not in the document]
tomas-langer: The latest commit introduces support for parallel testing, by creating a context for each test class, and storing it in the extension context.

testing/testing/src/main/java/io/helidon/testing/TestConfig.java
tomas-langer: Fixed. Also fixed the container annotations to be just plurals of the repeated annotations (so there is still `Values`, but now it is a container for `Value`, and there is a `Block` annotation for config source text block

common/context/src/main/java/io/helidon/common/context/ContextValue.java
```diff
@@ -0,0 +1,168 @@
...
+ * Otherwise, it is registered with {@link io.helidon.common.context.Contexts#globalContext()}.
+ * <p>
+ * The intention of this type is to have singletons for production runtime, while maintaining a separation during
+ * testing, where Helidon test extension run each part of the test class in a test context.
+ * <p>
+ * Note that the instance is not simply registered in the context, but it uses this class as a classifier.
+ *
+ * @param <T> type of the value stored in context
+ */
+public final class ContextValue<T> {
```
romain-grecourt, line 39: The name `ContextValue` is ambiguous.

I would expect `ContextValue` to follow the same pattern as `ConfigValue` / `OptionalValue` and replace `Optional` as the return type for the methods of `Context`.

Maybe that's not a bad idea, but that's out of scope for this PR.

---

This class is about singletons or "static values", the name and the context qualifier should reflect that.
The current qualifier is `global-instances`, but `Global` is already overloaded with `Context.globalContext()`.

I like the term `static` because it's short, however I'm not convinced with either `StaticValue` or `ContextStaticValue`. `ContextSingleton` does not seem too bad.

---

Another solution could be to separate the "mutable value" and the "static context". 

I.e. A generic `MutableValue`, similar to `LazyValue`, combined with `Contexts.staticContext()`.

E.g.

```java
static final MutableValue<Config> CONFIG_VALUE = MutableConfig.create(
    Contexts::staticContext,
    (c, v) -> c.set(Config.class, v),
    c -> c.get(Config.class).orElseThrow());
```

---

You stated earlier that one of the reasons we can't use service registry currently is because the global values are mutable.

Could we implement registry mutability with a special wrapper type like `MutableValue` ?

```java
static GenericType<MutableValue<Config>> MUTABLE_TYPE = new GenericType<MutableValue<Config>>() {};
static LazyValue<Config> DEFAULT_CONFIG = LazyValue.create(Config::empty);

static Config config() {
    return GlobalServiceRegistry.registry()
            .first(MUTABLE_TYPE)
            .map(MutableValue::get)
            .orElseGet(DEFAULT_CONFIG);
}
```

`GlobalServiceRegistry.registry()` would be based on a "static context".

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactoryProvider.java
```diff
@@ -51,7 +51,8 @@ public MetricsFactory create(Config rootConfig, MetricsConfig metricsConfig, Col
 
     @Override
     public void close() {
-        metricsFactories.forEach(MetricsFactory::close);
+        var toHandle = List.copyOf(metricsFactories);
```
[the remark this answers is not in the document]
tomas-langer: No, because `close` removes itself from the list, and there was a `ConcurrentModificationException`

common/context/src/main/java/io/helidon/common/context/Contexts.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2024 Oracle and/or its affiliates.
```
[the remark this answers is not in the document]
tomas-langer: Because of commit history - cannot resolve unless I squash all commits into one.

common/context/src/main/java/io/helidon/common/context/ContextValue.java
tomas-langer: Problem with Registry is also cyclic dependency - Registry must depend on config, and config already uses a "global" instance concep.

tomas-langer: Same with metrics - registry depends on metrics API

tomas-langer: Agreed on the naming. Will try a different one.

tomas-langer: Tried a different name, please re-check (also using a constant for the context registered)

common/context/src/main/java/io/helidon/common/context/ContextSingleton.java
```diff
@@ -0,0 +1,189 @@
...
+ * Note that the instance is not simply registered in the context, but it uses this class, and the "owner type" as a classifier
+ * to avoid conflicts.
+ *
+ * @param <T> type of the value stored in context
+ */
+public final class ContextSingleton<T> {
+    /**
+     * Classifier used to register a context that is to serve as the static context.
+     */
+    public static final String STATIC_CONTEXT_CLASSIFIER = "helidon-static-context";
```
romain-grecourt, line 42: How about doubling down on the term "singleton".

I.e. `ContextSingleton.CONTEXT_CLASSIFIER` 

```java
public static final String CONTEXT_CLASSIFIER="helidon-singleton-context"
```

common/context/src/main/java/io/helidon/common/context/ContextSingleton.java
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * This is similar to {@link java.lang.ThreadLocal}, except it uses {@code static} context to store a value.
+ * Static context is either a context available under special classifier {@value #STATIC_CONTEXT_CLASSIFIER}, or the
+ * {@link io.helidon.common.context.Contexts#globalContext()} (if the qualifier is not registered).
+ * <p>
+ * The intention of this type is to have singletons for production runtime, while maintaining a separation during
+ * testing, where Helidon test extensions run each part of the test class in a test context.
+ * <p>
+ * Note that the instance is not simply registered in the context, but it uses this class, and the "owner type" as a classifier
```
romain-grecourt, line 33: Here we describe a contract for the uniqueness which contradicts the semantic of a singleton.
I think we should remove that sentence and leave this as an implementation detail.

common/context/src/main/java/io/helidon/common/context/ContextSingleton.java
```diff
@@ -0,0 +1,189 @@
...
+package io.helidon.common.context;
+
+import java.util.NoSuchElementException;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Supplier;
+
+/**
+ * This is similar to {@link java.lang.ThreadLocal}, except it uses {@code static} context to store a value.
```
romain-grecourt, line 26: How about this ?

```java
/**
 * A context based "singleton" value holder that provides an indirection mechanism for static values.
 * "Singleton" values are stored and resolved from a context classified with {@value #STATIC_CONTEXT_CLASSIFIER}
 * in the {@link io.helidon.common.context.Contexts#currentContext() current context},
 * or the {@link io.helidon.common.context.Contexts#globalContext() global context if none is defined}.
 */
```

And then leave-out all the mentions to testing.

testing/testing/src/main/java/io/helidon/testing/TestConfig.java
```diff
@@ -0,0 +1,237 @@
...
+            AccumulatedOptions.class,
+            AccumulatedOptions::new);
+    private static final ReentrantLock LOCK = new ReentrantLock();
+
+    private TestConfig() {
+    }
+
+    /**
+     * Set a configuration key and value.
+     * This method CANNOT override an existing key, as such keys are already in the config snapshot.
```
danielkec, line 50: FAQ topic candidate

testing/testing/src/main/java/io/helidon/testing/TestRegistry.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public final class TestRegistry {
+    private TestRegistry() {
+    }
+
+    /**
+     * Mark a static method to be executed after the service registry has been shut-down.
+     * <p>
+     * This will only be executed after all pre-destroy methods have been called and the registry is no
+     * long available (i.e. you cannot obtain the registry instance that was used during all of your test
```
danielkec, line 35:
```suggestion
     * longer available (i.e. you cannot obtain the registry instance that was used during all of your test
```

danielkec (approved):

testing/testing/src/main/java/io/helidon/testing/TestConfigSource.java
```diff
@@ -0,0 +1,105 @@
...
+    TestConfigSource() {
+        TestConfig.register(this);
+    }
+
+    /**
+     * Get a new instance without any configured keys.
+     *
+     * @return instance of a test config source.
+     */
+    public static TestConfigSource create() {
```
romain-grecourt, line 64: We should provide another variant that takes a initial map.
E.g.
```java
TestConfigSource.create(Map.of("foo", "bar"));
```

testing/testing/src/main/java/io/helidon/testing/TestConfig.java
romain-grecourt: Do we event need this method ?
This looks like a hack that we aren't using, at least in this pull request.

One should be able to use this instead:
```java
@TestConfig.Source
static ConfigSource config() {
    var config = TestConfigSource.create();
    onPseudoEvent(() -> config.set("foo", "bar"));
    return config;
}
```

common/context/src/main/java/io/helidon/common/context/ContextSingleton.java
```diff
@@ -0,0 +1,184 @@
...
+ * A context based "singleton" value holder that provides an indirection mechanism for static values.
+ * "Singleton" values are stored and resolved from a context classified with {@value #STATIC_CONTEXT_CLASSIFIER}
+ * in the {@link io.helidon.common.context.Contexts#context()} current context},
+ * or the {@link io.helidon.common.context.Contexts#globalContext() global context if none is defined}.
+ *
+ * @param <T> type of the value stored in context
+ */
+public final class ContextSingleton<T> {
+    /**
+     * Classifier used to register a context that is to serve as the static context.
```
romain-grecourt, line 35: > Classifier used to register contexts as "singleton" capable.

common/context/src/main/java/io/helidon/common/context/ContextSingleton.java
```diff
@@ -0,0 +1,184 @@
...
+     */
+    public static final String STATIC_CONTEXT_CLASSIFIER = "helidon-singleton-context";
```
romain-grecourt, line 37:
```java
ContextSingleton.CLASSIFIER
```

is better than

```java
ContextSingleton.STATIC_CONTEXT_CLASSIFIER
```

At the very least we should not use "static", so either `SINGLETON_CONTEXT_CLASSIFIER`, `CONTEXT_CLASSIFIER` or just `CLASSIFIER`.

danielkec (approved):

tomas-langer:
Closing this PR after our extensive changes to Service Registry, where we decided to use ServiceRegistry as the main store for "global singletons".
The new PR will only contain changes needed to store service registry.

Replaced by #9656

PR #9197, opened 2024-08-25
4.x: Remove unused private static classes from io.helidon.config.BuilderImpl

Captain1653:
### Description
I've found two unused private static classes in `io.helidon.config.BuilderImpl`, but I'm not sure.  

Maybe it's a bug and these classes should be used? All tests are green.


romain-grecourt:
I think these static methods are vestiges of the old mp-config implementation (#1721) ; it seems safe to remove.

romain-grecourt (approved):

PR #9201, opened 2024-08-26
Add support for additional modifiers

### LINKED ISSUE #9202: Suport additional modifiers in codegen
The classmodel (`io.helidon.common.types.Modifier`) currently supports `abstract`, `final` and `static` modifiers from `java.lang.reflect.Modifier` - at least `transient` should be added, so clients can skip processing of non-persistent fields

tomas-langer:
Resolves #9202

Added modifiers:
 - transient
 - volatile
 - synchronized
 - native

Both to the enum, and to codegen (APT supports it already, as it uses the same strings, added support to scanning based codegen)


codegen/scan/src/main/java/io/helidon/codegen/scan/ScanTypeInfoFactory.java
```diff
@@ -458,6 +458,18 @@ private static Set<Modifier> toModifiers(ClassMemberInfo memberInfo) {
...
+            }
+            if (mi.isNative()) {
+                result.add(Modifier.NATIVE);
+            }
+        }
+
+        if (memberInfo instanceof FieldInfo fi) {
+            if (fi.isTransient()) {
+                result.add(Modifier.TRANSIENT);
+            }
```
lukasj, line 472: isn't `if fi.isVolatile()` missing here?

tomas-langer: It is not available. Added a workaround using `Modifier.isVolatile(fi.getModifiers())`

lukasj (approved):

PR #9206, opened 2024-08-26
4.x: UncheckedIOException no longer a special case

### LINKED ISSUE #9192: Default error handler does not return a 5xx on UncheckedIOException
## Environment Details
* Helidon Version: 4.1.0
* Helidon SE
* JDK version: 21.0.3
* OS: macOS

----------

## Problem Description
When a request handler throws an UncheckedIOException (such as when the handler uses a WebClient to call another service that isn't currently running resulting in a ConnectException wrapped in an UncheckedIOException) I expect the Helidon default error handler to return a 500 HTTP error. Instead the web server return nothing.

The cause is io.helidon.webserver.http.ErrorHandlers.runWithErrorHandling line 80 that catches and re-throws UncheckedIOException (and CloseConnectionException), with a comment saying 'these errors must "bubble up"'.

Maybe if the response is "broken" so returning an HTTP 500 can't work then an exception needs to bubble up. Perhaps the code assumes that these 2 exception categories (UncheckedIOException and CloseConnectionException) are due to the response. But that is not always correct, particularly since the Helidon WebClient throws those and lots of microservices call other microservices.

## Steps to reproduce

* `helidon init` -- the defaults
* Add a path to the routing (and `import java.io.*`): `.get("/throw", (req, res) -> { throw new UncheckedIOException(new IOException("ugh")); })`
*  `mvn package`
* `java -jar target/quickstart-se.jar`
* `curl -vi http://localhost:8080/throw

Expect `HTTP/1.1 500 Internal Server Error` but don't get anything.

tomas-langer:
Resolves #9192 

No longer using UncheckedIOException as a special case for the server.
Server I/O failures now throw a ServerConnectionException, and that extends the CloseConnectionException (which already has its special handling).

This change allows user to handle UncheckIOException as if it is any other throwable.

As part of this change:
- all reads on server (from server request reader) are now wrapped and no longer throw `UncheckedIOException`
- all writes on server (to server response writer) are surrounded by try/catch blocks to handle this exception.

This is a small behavior change. The original behavior was not documented, and the result was not ideal (as we would close connection even if the `UncheckedIOException` came from client call to another service).
The new behavior only closes connection with specific exception.

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
@@ -412,7 +412,7 @@ public class Http2Connection implements ServerConnection, InterruptableTask<Void
         if (frameHeader.length() == 0) {
             frameInProgress = BufferData.empty();
         } else {
-                frameInProgress = reader.readBuffer(frameHeader.length());
+            frameInProgress = reader.readBuffer(frameHeader.length());
         }
 
         receiveFrameListener.frame(ctx, streamId, frameInProgress);
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
@@ -412,7 +412,7 @@ public class Http2Connection implements ServerConnection, InterruptableTask<Void
         if (frameHeader.length() == 0) {
             frameInProgress = BufferData.empty();
         } else {
-            frameInProgress = reader.readBuffer(frameHeader.length());
+                frameInProgress = reader.readBuffer(frameHeader.length());
         }
 
         receiveFrameListener.frame(ctx, streamId, frameInProgress);
--- a/webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/UncheckedIoExceptionTest.java
+++ b/webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/UncheckedIoExceptionTest.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package io.helidon.webserver.tests;
 
 import java.io.IOException;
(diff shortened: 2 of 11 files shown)


webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
```diff
@@ -412,7 +412,7 @@ private void readFrame() {
         if (frameHeader.length() == 0) {
             frameInProgress = BufferData.empty();
         } else {
-            frameInProgress = reader.readBuffer(frameHeader.length());
+                frameInProgress = reader.readBuffer(frameHeader.length());
```
romain-grecourt, line 415: This whitespace change seems unintended.

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/UncheckedIoExceptionTest.java
```diff
@@ -0,0 +1,49 @@
...
+import io.helidon.webserver.testing.junit5.ServerTest;
+import io.helidon.webserver.testing.junit5.SetUpRoute;
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@ServerTest
+public class UncheckedIoExceptionTest {
```
romain-grecourt, line 34: should be package private

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/UncheckedIoExceptionTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @SetUpRoute
+    public static void routing(HttpRules rules) {
```
romain-grecourt, line 36: should be package private

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/UncheckedIoExceptionTest.java
```diff
@@ -0,0 +1,49 @@
...
+        rules.get("/fail", ((req, res) -> {
+            throw new UncheckedIOException("My Exception", new IOException("Outbound client failure"));
+        }));
+    }
+
+    @Test
+    public void testUncheckedIoTreatedAsAnyOther(Http1Client client) {
```
romain-grecourt, line 43: should be package private

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
```diff
@@ -363,7 +364,11 @@ private void route(HttpPrologue prologue, WritableHeaders<?> headers) {
         // Expect: 100-continue
         if (headers.contains(HeaderValues.EXPECT_100)) {
             if (this.http1Config.continueImmediately()) {
-                writer.writeNow(BufferData.create(CONTINUE_100));
+                try {
+                    writer.writeNow(BufferData.create(CONTINUE_100));
+                } catch (UncheckedIOException e) {
+                    throw new ServerConnectionException("Failed tow rite continue", e);
```
danielkec, line 370:
```suggestion
                    throw new ServerConnectionException("Failed to write continue", e);
```

danielkec (changes requested):

danielkec (approved):

romain-grecourt (approved):

Issue #9165, opened 2024-08-19
Server-sent event (SSE) objects cannot be created without data


tomwhoiscontrary:
## Environment Details
* Helidon Version: 4.1.0
* Helidon SE or Helidon MP: SE
* JDK version: Temurin 21.0.4+7
* OS: Ubuntu 22.04.4
* Docker version (if applicable): N/A

----------

## Problem Description

If you try this:

```
            SseEvent comment = SseEvent.builder().comment("this is a comment").build();
```

The call to build blows up with:

```
java.lang.IllegalArgumentException: Event must include some data
	at io.helidon.http.sse.SseEvent$Builder.build(SseEvent.java:233)
```

As far as i know, the SSE specification does not require data in an event. The WHATWG spec:

https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream

defines an event like this:

```
event         = *( comment / field ) end-of-line
```

where the asterisk means any number, including zero.

The SseSink code for serialising events explicitly handles data being absent:

https://github.com/helidon-io/helidon/blob/d4d538164bfbae0f16ca45f97647fb0ea520f968/webserver/sse/src/main/java/io/helidon/webserver/sse/SseSink.java#L96

This is a really weird thing to do, i know, but it would sometimes be useful to send an event with no data, and just a comment.

## Steps to reproduce

Happy to write a demonstration if that would be helpful, let me know.

PR #9207, opened 2024-08-27
Allows the creation of empty SSE events

spericas:
### Description

Allows an SSE event to be created using no data. Uses a static constant SseEvent.NO_DATA to avoid returning null in getters. Updates serialization code and tests. See issue #9165.

### DIFF
--- a/http/sse/src/main/java/io/helidon/http/sse/SseEvent.java
+++ b/http/sse/src/main/java/io/helidon/http/sse/SseEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


tjquinno (commented):
The changes look fine to me, although I do have one question.

I see the writer of the issue cited the SSE spec saying that "empty" events are legal.

I know very little about SSE, but is it in any way a potential problem that the `SseEvent` data for an empty event would contain this special constant value (as opposed to having no value at all)? For example, does this distort some representation of the event size or length or something like that that could otherwise be used to infer an "empty" event?

spericas:
> The changes look fine to me, although I do have one question.
> 
> I see the writer of the issue cited the SSE spec saying that "empty" events are legal.
> 
> I know very little about SSE, but is it in any way a potential problem that the `SseEvent` data for an empty event would contain this special constant value (as opposed to having no value at all)? For example, does this distort some representation of the event size or length or something like that that could otherwise be used to infer an "empty" event?

The code in `SseSink` currently checks for that constant and does not serialize anything into the stream. Would that answer your question?

tjquinno:
Yes. Thanks.

tjquinno (dismissed):

tjquinno (approved):

PR #9239, opened 2024-09-04
Logs action of closing an idle connection after timeout expires

spericas:
### Description

Logs action of closing an idle connection after timeout expires


spericas:
PR #9203

barchetta (approved):

Issue #9158, opened 2024-08-18
4.x: Helidon Service Inject


tomas-langer:
Core module that provides injection, interception, and events.

Follow up issues will be created for other features.

PR #9244, opened 2024-09-10
4.x: Helidon inject collateral 5

tomas-langer:
Related to #9158 

TypeInfo and TypedElementInfo now have originatingElementValue() that does not return an optional (and replaced usages)

Annotation now has metaAnnotation information (inherited annotations on the annotation or an annotation it is annotated with) All annotation values are validated, and classes modified to TypeName, and enums modified to EnumValue New modifiers added.
Added signature of TypedElementInfo to compare if methods or constructors have the same signature Added findInHierarchy to TypeInfo, so we can check if a type extends another type, or implements an interface A few small fixes (typos, almost valid regexp in tests etc.)


builder/tests/common-types/src/main/java/io/helidon/common/types/AnnotationSupport.java
```diff
@@ -619,22 +627,33 @@ private static Class<?> asClass(TypeName typeName, String property, Object value
...
-                return Class.forName(str);
-            } catch (ClassNotFoundException e) {
-                throw new IllegalArgumentException(typeName.fqName() + " property " + property
-                                                           + " of type String and value \"" + str + "\""
-                                                           + " cannot be converted to Class");
-            }
+
+        String className;
+
+        if (value instanceof TypeName tn) {
```
ljnelson, line 633: (This block could be a simple `switch` expression)

danielkec (dismissed):

romain-grecourt (approved):

Verdent (approved):

builder/tests/common-types/src/main/java/io/helidon/common/types/AnnotationSupport.java
tomas-langer: Fixed (and fixed another problem in this method).

danielkec (dismissed):
Switched to switch ...

danielkec (approved):

PR #9248, opened 2024-09-12
Policy validator configurable per endpoint in config

### LINKED ISSUE #8948: Support to add ABAC policy in configuration for explicit authorization
## Environment Details
* Helidon Version: 4.0.10
* Helidon SE or Helidon MP: Helidon MP
* JDK version: Oracle JDK 21.0.1
* OS: Windows (applicable on all platforms)

----------

## Problem Description
We need a support to define Authorization (AuthZ) rules at deployment time to protect a Jersey Resource.
For instance, 
* Deployment A will allow users with role/group "admin-users" access a resource while in deployment B, same resource will be allowed to users with role/group "super-users".
* The rules will not be limited to roles alone but will also consider data; for instance, a user with the role "regular-csr" may POST a refund up to a maximum of $1000.00.  

ABAC support currently available seems like a good fit; an expression like `inRole('admin-users") && object.amount <= 1000` would enforce the required restriction on resource.

But, at the moment, `object` requires the AuthZ as **explicit = true**.
If _application.yaml_ contains a rule specified in `security.web-server.paths[].abac`, then evaluation is done before the Resource code gets the control to check for authorization. Here, `object = null` and check fails.

At the same time, we can't fix the configuration in code as the rules can vary in different deployments.

## Expectation
Allow defining the rules in configuration file, (application.yaml), so it can be customized per deployment environment without changing the code.
Possibly, `security.web-server.paths[].abac` or `security.web-server.paths[].abac.policy-validator` can support a new key as **explicit** to decide if its evaluation should be deferred, so there is no impact to existing consumers.

Verdent:
### Description
Fixes #8948

### Documentation

Example for adding JAX-RS endpoint policy validator via config added
Fixed several legacy configurations in the same document

### DIFF
--- a/docs/src/main/asciidoc/includes/security/providers/abac.adoc
+++ b/docs/src/main/asciidoc/includes/security/providers/abac.adoc
@@ -119,9 +119,11 @@ Annotations: `@RolesAllowed`, `@RoleValidator.Roles`
 .Configuration example for `WebServer`
 ----
 security:
-  web-server.paths:
-    - path: "/user[/{*}]"
-      roles-allowed: ["user"]
+  features:
+    security:
+      paths:
+        - path: "/user[/{*}]"
+          roles-allowed: ["user"]
 ----
 
 [source,java]
@@ -160,10 +162,12 @@ Annotations: `@Scope`
 .Configuration example for `WebServer`
 ----
 security:
-  web-server.paths:
-    - path: "/user[/{*}]"
-      abac.scopes:
-        ["calendar_read", "calendar_edit"]
+  features:
+    security:
+      paths:
+        - path: "/user[/{*}]"
+          abac.scopes:
+            ["calendar_read", "calendar_edit"]
 ----
 
 [source,java]
@@ -185,14 +189,28 @@ Example of a policy statement: `${env.time.year >= 2017}`
 .Configuration example for `WebServer`
 ----
 security:
-  web-server.paths:
-    - path: "/user[/{*}]"
-      policy:
-        statement: "hasScopes('calendar_read','calendar_edit') AND timeOfDayBetween('8:15', '17:30')"
+  features:
+    security:
+      paths:
+        - path: "/user[/{*}]"
+          policy:
+            statement: "hasScopes('calendar_read','calendar_edit') AND timeOfDayBetween('8:15', '17:30')"
 ----
 
 [source,java]
 .JAX-RS example
 ----
 include::{sourcedir}/includes/security/providers/AbacSnippets.java[tag=snippet_4, indent=0]
+----
+
+[source,yaml]
+.Configuration example for `JAX-RS` over the configuration
+----
+security:
+  features:
+    security:
+      endpoints:
+        - path: "/somePath"
+          config:
+            abac.policy-validator.statement: "${env.time.year >= 2017}"
 ----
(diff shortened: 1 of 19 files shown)


docs/src/main/asciidoc/includes/security/providers/abac.adoc
```diff
@@ -119,9 +119,11 @@ Annotations: `@RolesAllowed`, `@RoleValidator.Roles`
 .Configuration example for `WebServer`
 ----
 security:
-  web-server.paths:
-    - path: "/user[/{*}]"
-      roles-allowed: ["user"]
+  features:
```
tomas-langer, line 122: This is wrong - either it can be in `security.web-server.paths`, or under `server.features...` - the root here is `security`

docs/src/main/asciidoc/includes/security/providers/abac.adoc
```diff
@@ -160,10 +162,12 @@ Annotations: `@Scope`
 .Configuration example for `WebServer`
 ----
 security:
-  web-server.paths:
-    - path: "/user[/{*}]"
-      abac.scopes:
-        ["calendar_read", "calendar_edit"]
+  features:
```
tomas-langer, line 165: Same as above

docs/src/main/asciidoc/includes/security/providers/abac.adoc
```diff
@@ -185,14 +189,28 @@ Example of a policy statement: `${env.time.year >= 2017}`
 .Configuration example for `WebServer`
 ----
 security:
-  web-server.paths:
-    - path: "/user[/{*}]"
-      policy:
-        statement: "hasScopes('calendar_read','calendar_edit') AND timeOfDayBetween('8:15', '17:30')"
+  features:
```
tomas-langer, line 192: Dtto

docs/src/main/asciidoc/includes/security/providers/abac.adoc
```diff
@@ -185,14 +189,28 @@ Example of a policy statement: `${env.time.year >= 2017}`
...
 [source,java]
 .JAX-RS example
 ----
 include::{sourcedir}/includes/security/providers/AbacSnippets.java[tag=snippet_4, indent=0]
 ----
+
+[source,yaml]
+.Configuration example for `JAX-RS` over the configuration
+----
+security:
```
tomas-langer, line 209: dtto

tests/integration/security/abac-policy/src/main/resources/logging.properties
```diff
@@ -0,0 +1,29 @@
...
+
+# Send messages to the console
+handlers=io.helidon.logging.jul.HelidonConsoleHandler
+
+# HelidonConsoleHandler uses a SimpleFormatter subclass that replaces "!thread!" with the current thread
+java.util.logging.SimpleFormatter.format=%1$tY.%1$tm.%1$td %1$tH:%1$tM:%1$tS %4$s %3$s !thread!: %5$s%6$s%n
+
+# Global logging level. Can be overridden by specific loggers
+.level=INFO
+AUDIT.level=FINEST
```
tomas-langer, line 28: Please remove FINEST log from tests.

tomas-langer (changes requested):

docs/src/main/asciidoc/includes/security/providers/abac.adoc
Verdent: aah.. yes... I missed it is under security and assumed it is the server.... My bad.

tomas-langer (approved):

PR #9260, opened 2024-09-18
4.x: JSON Formatter for JUL

### LINKED ISSUE #9259: There is no Logging Formatter that will handle JSON for JUL
## Environment Details
* Helidon Version: All

----------

## Problem Description
Currently, there is no appropriate Formatter for JSON with JUL. The alternative is to use SimpleFormatter like this: 

```
java.util.logging.SimpleFormatter.format={"ts":"%1tQ","timestamp":"%1$tY-%1$tm-%1$td %1$tH:%1$tM:%1$tS.%1$tL","level":"%4$s","logger":"%3$s","thread":"!thread!","msg":"%1$tQ - %5$s%6$s"}%n
```

but will break in instances where the message has multi-lines. The goal of this request is to  find a suitable approach that will handle this.

tomas-langer:
Resolves #9259 

### Description
JSON formatter for JUL and tests
A fix of escaping when writing in HSON.
Switched one packaging test to use JSON format.

### Documentation
There is currently no documentation of the logging modules.
I have added details about what this PR does and how to configure a JSON formatter to the linked issue

### DIFF
--- a/logging/jul/src/main/java/io/helidon/logging/jul/HelidonJsonFormatter.java
+++ b/logging/jul/src/main/java/io/helidon/logging/jul/HelidonJsonFormatter.java
@@ -1,0 +1,280 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.logging.jul;
+
+import java.io.PrintWriter;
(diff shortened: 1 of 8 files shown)


logging/jul/src/main/java/io/helidon/logging/jul/HelidonJsonFormatter.java
```diff
@@ -0,0 +1,280 @@
...
+ * <p>
+ * The configuration falls back to property {@value io.helidon.logging.jul.HelidonFormatter#JUL_FORMAT_PROP_KEY},
+ * and analyzes it to provide a "guessed" JSON structure.
+ * <p>
+ * Example (and also the default format):<br>
+ * {@value #DEFAULT_FORMAT}
+ */
+public class HelidonJsonFormatter extends Formatter {
+    static final String DEFAULT_FORMAT = "ts:%1$tQ,date:%1$tY.%1$tm.%1$td,time:%1$tH:%1$tM:%1$tS.%1$tL,level:%4$s,message:%5$s,"
+            + "exception:%6$s,thread:!thread!,logger:%3$s";
```
klustria, line 58: For default format, I think date and time is more than enough and ts that represents milliseconds since the beginning of the epoch will not be necessary in general.

klustria: Makes sense.

klustria (approved):
LGTM

PR #9271, opened 2024-09-23
Grouping Executors related methods into a single class

JaroslavTulach:
### Description

Enso has decided to use Helidon 4.0.x to emulate `WebSocket` on Graal.js. However we have problems to run our code on _native image_ with GraalVM for JDK21. [More info](https://github.com/enso-org/enso/pull/10783#discussion_r1768000821) available - in short: there is a bug in GraalVM for JDK21 that prevents usage of virtual threads and Graal.js. 

The bug was fixed in newer versions of GraalVM, but [backport is complicated](https://github.com/oracle/graal/pull/9516) and we don't have a workaround for the bug on the GraalVM side and we'd like to stick with JDK 21.

### Documentation

We have [experimentally verified](https://github.com/enso-org/enso/pull/10783/commits/c032d9db5c0e083a7c1e0f9cfd68e8312c8e96d3#r1770787594) that with small changes in Helidon code we can use `@Substitute` mechanism of GraalVM _native image_ to switch to _platform threads_. As we don't care about massive performance, but just want to get things work, it is a sufficient solution for us.

Can you accept this little refactoring into next version of Helidon, please? Thank you very much.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ExecutorsFactory.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ExecutorsFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


JaroslavTulach:
FYI: @ola-lis has created an official ticket in _My Oracle Support_ requesting fix for the [Helidon 4 + Graal.js on native-image for JDK21](https://github.com/enso-org/enso/pull/10783#issuecomment-2294854426) problem:

My Oracle Support request

E.g. by accepting this PR you'd help to resolve an official Oracle **Java SE JDK and JRE licensee** request.

tomas-langer (dismissed):

tomas-langer (approved):

Issue #9276, opened 2024-09-23
Allow Unbuffered OutputStream for Http1ServerResponse


VerKWer:
## Environment Details
* Helidon Version: 4.1.0
* Helidon SE
* JDK version: 21
* OS: Linux

----------

## Problem Description
Since #6509 was merged, the write buffer size of a web server must be strictly greater than zero. If a write buffer size of `0` is specified, the server will start and function normally, until some request handler tries to use the `OutputStream` for the request, at which point an exception is thrown.

This could be solved by checking the write buffer size when the server is instantiated, but I propose falling back to the previous, unbuffered behaviour in that case instead. For example, I'm working on moving an existing project to Helidon, which uses buffer pools to reduce GC pressure. Currently, the Helidon web server will always allocate a new buffer for every request that uses its `OutputStream`, and there is no way to avoid this.

## Steps to Reproduce
Create a new `WebServer` instance with `writeBufferSize(0)`. The server will start normally and responses can be sent using the `ServerResponse.send` methods. However, handling fails if `ServerResponse.outputStream()` is called because it tries to unconditionally allocate a `BufferedOutputStream`, for which `0` is an illegal buffer size.

## Other Considerations
I'm not sure about Http2. It seems that the `Http2ServerResponse` has no internal buffering mechanism? Any more information about this would be appreciated.

PR #9277, opened 2024-09-23
4.x: Allow Unbuffered OutputStream for Http1ServerResponse (#9276)

VerKWer:
### Description
Currently, when building a new `WebServer` instance, one can specify `writeBufferSize(0)`. The server builds and starts normally, but any http1 request handler that tries to use `ServerResponse.outputStream()` will throw an exception. This is obviously not very user friendly, and we should either fail when trying to build/start the server instance or support unbuffered response `OutputStream`s.

I propose the latter approach, as it was the default before #6509 was merged. An obvious use case for unbuffered response writer is if the user does their own buffer management. For example, I'm currently moving an existing project to Helidon, which uses its own buffer pool to reduce GC pressure. Not allowing a buffer size of `0` would mean that there is no way to avoid the allocation of a new buffer for every request.

Solves issue #9276

### Documentation
I added a new abstract class `ResponseOutputStream` in `Http1ServerResponse`, which serves as a base for the two existing classes `BlockingOutputStream` and `ClosingBufferedOutputStream`. That way, we can choose which one to use after checking the write buffer size.

I also added the `WriteBufferTest` class to test both the default and zero write buffer case.


webserver/webserver/src/test/java/io/helidon/webserver/http1/WriteBufferTest.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+class WriteBufferTest {
+
+    /**
+     * Test that a simple response can be sent using the {@link ServerResponse#outputStream()} using the default
+     * (non-zero) write buffer.
+     */
+    @Test
+    void defaultWriteBufferTest() throws Exception {
```
tomas-langer, line 44: Tests that require startup of a server should go to:
`webserver/tests/webserver`
where you can also use the `@ServerTest` annotation. See other tests in the module.
If you do not feel like changing the PR, let me know and I will take it over and do the rest.
Thanks for your contribution!

tomas-langer (changes requested):

webserver/webserver/src/test/java/io/helidon/webserver/http1/WriteBufferTest.java
VerKWer: Thanks for clarifying this. I'll gladly move the tests. How would I go about actually running the tests there? They don't seem to be part of the `webserver` module, and if I try using `webserver/tests` as a standalone project, compilation fails due to missing dependencies. I'm sure I'm missing something very obvious here.

VerKWer: Never mind, I forgot to activate the `tests` profile. I moved the relevant test to the correct module.

tomas-langer:
Hello, thanks a lot for your contribution, but this was already fixed.
The fix is directly in constructor of `io.helidon.webserver.http1.Http1ServerResponse.ClosingBufferedOutputStream#ClosingBufferedOutputStream`
There is a new test: `io.helidon.webserver.tests.OutputStreamBufferSizeTest`

romain-grecourt:
See #9313

Issue #9255, opened 2024-09-17
4.x Helidon gRPC client fail-over issues


thegridman:
## Environment Details
* Helidon Version: 4.1.1
* Helidon SE

## Problem Description

We are using the new Helidon gRPC client in Coherence so that customers using Helidon in their application do not have to use Netty for the gRPC client. Our main problem area is around fail-over, specifically when the server the client is connected to dies. This is a common occurrence in environments like Kubernetes, where losing a Pod any time without warning is just how it is.

We have a fail over test in our CI build that intermittently fails using the Helidon gRPC client, but is rock solid using gRPC Java and Netty. As I see it there are a few causes of this.

The Helidon gRPC client (which is basically the web client) can only be configured with a single URI. But in a typical Coherence application, we would run multiple servers and the client can connect to any. It might be fine if we put a load balancer over the top, but a lot of our customers do not do this. We have a discovery mechanism that allows the client to discover the list of server endpoints but we can only choose one to configure the client with. As the client can only have a single fixed URI, if that server fails the client (and gRPC Channel) is effectively dead, as there is no way to tell it to use an alternative endpoint. 

We have worked around this by having our own gRPC Channel implementation that wraps the Helidon gRPC client. Our Channel will do the endpoint lookup and pick one of the addresses to configure the Helidon client. If the Helidon Channel fails, we throw that Helidon client away, re-do endpoint discovery and create a new Helidon client for to connect to a different server. This mostly works, but is a bit of a hack in our opinion and not actually that reliable.

The problem then comes because it seems to take some time for the Helidon client to detect the server has gone. Presumably this is because the Helidon client is effectively in a polling loop with a timeout, so when the server dies, the error is not immediate. At least this is the case with our use of bi-directional streaming calls. This means it can take a while between the server dying and, the Helidon client detecting this, and our fail-over logic creating a new channel. This means that a request from our application after the server is dead may still be sent to the dead socket and immediately throw an exception. Presumably for other gRPC ,methods, unary for example, the failure will happen when the client tries to send the request, but we have not found a suitable place to be able to hook in any retry logic. 

[...]

PR #9285, opened 2024-09-25
4.x: Adds support to iterate over URIs when connecting to a gRPC service

spericas:
### Description

Adds support to iterate over URIs when connecting to a gRPC service. Defines a new interface `ClientUriSupplier` that can be used to dynamically provide a list of URIs to connect to. Each gRPC call will retrieve the next URI from the supplier. See issue #9255.

### DIFF
--- a/webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSupplier.java
+++ b/webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSupplier.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.webclient.grpc;
+
+import java.util.Iterator;
+
+import io.helidon.webclient.api.ClientUri;
+
+/**
+ * Interface implemented by all client URI suppliers.
+ */
+public interface ClientUriSupplier extends Iterator<ClientUri>, Iterable<ClientUri> {
+
+    @Override
+    default Iterator<ClientUri> iterator() {
+        return this;
+    }
+}
--- a/webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSuppliers.java
+++ b/webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSuppliers.java
@@ -16,6 +16,7 @@
 package io.helidon.webclient.grpc;
 
 import java.net.URI;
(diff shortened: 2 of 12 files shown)


webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSuppliers.java
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        /**
+         * Creates an ordered supplier.
+         *
+         * @param clientUris collection of client URIs
+         * @return new supplier
+         */
+        public static OrderedSupplier create(Collection<ClientUri> clientUris) {
+            return new OrderedSupplier(clientUris);
```
thegridman, line 58: The `RoundRobinSupplier` copies the collection passed into its `create` method (or rather it turns it into an array). In this method, the `OrderedSupplier` will create an iterator directly from the collection parameter. If this collection ever changes that will possibly cause a concurrent modification exception.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSuppliers.java
```diff
@@ -0,0 +1,210 @@
...
+            return new RandomSupplier(clientUris.toArray(new ClientUri[]{}));
+        }
+
+        protected RandomSupplier(ClientUri[] clientUris) {
+            this.clientUris = clientUris;
+        }
+
+        @Override
+        public boolean hasNext() {
+            return supplied;
```
thegridman, line 198: I see the code that uses `ClientUriSupplier` does not actually call `hasNext()` but this will return `false` if `next()` has not been called and only after `next()` is called will it return `true`, but then calling `next()` again will throw `NoSuchElementException`. I think this should be `return !supplied`

webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Iterator;
+
+import io.helidon.webclient.api.ClientUri;
+
+/**
+ * Interface implemented by all client URI suppliers.
+ */
+public interface ClientUriSupplier extends Iterator<ClientUri>, Iterable<ClientUri> {
+
+    @Override
```
thegridman, line 27: I'm not objecting to implementing both `Iterator` and `Iterable` (there is nothing additional we'd need to implement) but you never actually use this class as an `Iterable` anywhere.

thegridman (commented):
This all looks fine and will give us what we need to implement dynamic endpoints in Coherence.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSupplier.java
spericas: I understand, but I'd like to justify the creation of a new interface and it makes it a bit more friendly this way.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSuppliers.java
spericas: Yes, bug fixed

webclient/grpc/src/main/java/io/helidon/webclient/grpc/ClientUriSuppliers.java
spericas: Fixed

thegridman (approved):

thegridman (approved):

PR #9295, opened 2024-09-27
4.x: Concurrency limits module, and support in Helidon WebServer

### LINKED ISSUE #8897: Provide adaptive concurrency limits

### LINKED ISSUE #9229: Provide configurable option to queue requests when concurrency is limited with  "max-concurrent-requests"
## Environment Details
* Helidon Version:  4.x 
* Helidon SE or Helidon MP. : SE & MP 
* JDK version:  Java 21 
* OS:  Linux 

In. Helidon 4.x  with. Webserver that supports loom based virtual threads ,  uses the  new thread per request model .  So by design there is no longer  server thread pool  or  any associated  queues where requests get queued .   

By default. there is no limit on concurrency  and this can lead to issues  when resources such as DB connections,  external system integrations,  and. other such downstream resources  are limited.   This can lead to  performance degrade and also errors when requests timeout  waiting for such resources.  

To address this  Helidon  provides  the  "max-concurrent-requests" parameter on the Listener configuration.    While it helps to  limit the concurrency , the  services are running into issues when trying to use this  parameter to limit the concurrency 

 When the "max-concurrent-requests" parameter is set,  any  surge requests beyond the  limit get rejected and fail with 503.  There can be occasional surges that can cause the concurrency to go beyond the  configured limit, and such cases teh requests would error out. 
This behaviour   is  not consistent with  the behavior in earlier versions of Helidon where under this situation the requests would get queued in the queue associated with Helidon's server thread pool. 

It would be  good have an additional configurable options  in  Helidon 4 , where. one can  additionally enable queueing of requests ,  when   a limit is configured for max-concurrent-requests" 
Something like below

server :  
  max-concurrent-requests : 40 
  request-queue : 
     enable :  true 
     max :  100

tomas-langer:
### Description
Resolves #8897
Resolves #9229

### Documentation
This PR introduces 
- a new module `common/concurrency/limits` that provides API and SPI for concurrency limit implementations, and a couple of default implementations (AIMD, Fixed).
- a new module `webserver/concurrency-limits` that provides feature with a filter to impose limits within a filter in routing
- update of `webserver/webserver` to use a `Limit` instead of a `Semaphore` in connection handlers (backward compatible)

Configuration reference will be in the generated documentation.

#### Configure limits on WebServer
This will configure limit for a server listener (configurable per listener), enforced on the connection level (i.e. outside of routing and filters).

The default behavior is the same (unlimited).
If `server.max-concurrent-requests` is configured (value is not `-1`), it will be used and `concurrency-limit` configuration for the listener will be ignored.

Configuration:
```yaml
server:
  port: 8080
  concurrency-limit:
    aimd: # `limit type`
      # AIMD limit configuration
```
#### Configure limits for routing
This will configure limit for as a server feature, enforced in an HTTP filter.
Configuration:
```yaml
server:
  features:
    limits: # the feature is called `limits`
      enabled: true
      concurrency-limit: # `limit` configuration of the `limits` server feature
        fixed: # `limit type`
          permits: 1
          queue-length: 10
```


arouel (approved):
LGTM, great rework and integration

vasanth-bhat:
Have few clarifications

1.  Going through    Readme  , there is mention about 'Semaphore' concurrency limit type in  addition to aimd & BulkHead.  Is this same as. "basic" (BasicLimit)?   Also how is this different from specifying.  "server.max-concurrent-requests"? 

2.  For the  BulkHead,   it would be good to have metrics support ,   around  queueing ,  for example  %  queue full.  This is important in production to monitor  the queueing.  

3.  For AMID , it would be good to ahem a metrics around the dynamic concurrency limit in use. A Histogram may be useful here.

tomas-langer:
After discussing the needs in Helidon, we must do this using the `permit` approach, so this can be reused in webclient. 
I will refactor the PR.

arouel:
I wanted to bring up that metrics would be super important for production as @vasanth-bhat brought it up as well. Within my draft I figured quickly that the limiting works as expected but metrics would give me the insights I need to build convidence. Maybe this is something for a follow-up Pull Request, idk.

tomas-langer:
Metrics must be a follow up, as it would require much more work. I cannot just add a dependency on `helidon-metrics-api`, as it would introduce it everywhere (maybe that is the solution, but it requires a bit more thought).
I will finish this PR as the introduction of concurrency limits. We can look into the metrics as a follow up (this would make sense for fault tolerance as well, not just concurrency limits).

tomas-langer:
Pushed a new version, updated original description to reflect new implementation.
I have implemented the `fixed` limit as a more general feature - it supports queing now, so we no longer need fault tolerance for this.
How to add this to metrics needs to be a follow up.

tomas-langer:
Created a follow up issue: https://github.com/helidon-io/helidon/issues/9304

vasanth-bhat:
Went through the new changes.  

  I think it would help to clarify between  the below 2 , with example to understand  the difference between the two.   Do  both them provide same functionality but via different approaches ?

1. Fixed Limit with queueing support at the connection level  (or a server listener )
2. Fixed limit with queueing as server feature enforced via filter.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimit.java
```diff
@@ -0,0 +1,348 @@
...
+    /**
+     * Timeout of a request that is enqueued.
+     */
+    public static final String DEFAULT_QUEUE_TIMEOUT_DURATION = "PT1S";
+
+    static final String TYPE = "fixed";
+
+    private final FixedLimitConfig config;
+    private final LimiterHandler handler;
+    private int initialPermits;
```
danielkec, line 55: Can be final?

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -516,15 +516,35 @@ private void handle() {
...
                     ctx.log(LOGGER, TRACE, "Too many concurrent requests, rejecting request.");
                     response.status(Status.SERVICE_UNAVAILABLE_503)
                             .send("Too Many Concurrent Requests");
                     response.commit();
+                } else {
+                    LimitAlgorithm.Token permit = token.get();
+                    try {
+                        routing.route(ctx, request, response);
+                    } finally {
+                        if (response.status() == Status.NOT_FOUND_404) {
```
arouel, line 533: Should we allow a user to specify which 4xx codes can be ignored?

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
```diff
@@ -183,30 +187,35 @@ public void handle(Semaphore requestSemaphore) throws InterruptedException {
...
+                } else {
+                    LimitAlgorithm.Token permit = token.get();
 
+                    try {
+                        this.lastRequestTimestamp = DateTime.timestamp();
+                        route(prologue, headers);
+                        permit.success();
+                        this.lastRequestTimestamp = DateTime.timestamp();
+                    } catch (Throwable e) {
+                        permit.dropped();
```
arouel, line 215: Maybe something for a future PR: Should we allow a user to specify which exception can be recorded as ignored instead of being counted as dropped?

arouel (approved):

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+        if (maxLimit < minLimit) {
+            throw new ConfigException("Max limit must be higher than min limit");
+        }
+        if (initialLimit > maxLimit) {
+            throw new ConfigException("Initial limit must be lower than max limit");
+        }
+        if (initialLimit < minLimit) {
+            throw new ConfigException("Initial limit must be higher than minimum limit");
+        }
```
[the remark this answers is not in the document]
tomas-langer: ah, the messages are wrong, I see

danielkec (approved):

Verdent (approved):

PR #9298, opened 2024-09-27
4.1.2: Grouping Executors related methods into a single class

barchetta:
Backport #9271 to Helidon 4.1.2

### Description

Enso has decided to use Helidon 4.0.x to emulate `WebSocket` on Graal.js. However we have problems to run our code on _native image_ with GraalVM for JDK21. [More info](https://github.com/enso-org/enso/pull/10783#discussion_r1768000821) available - in short: there is a bug in GraalVM for JDK21 that prevents usage of virtual threads and Graal.js. 

The bug was fixed in newer versions of GraalVM, but [backport is complicated](https://github.com/oracle/graal/pull/9516) and we don't have a workaround for the bug on the GraalVM side and we'd like to stick with JDK 21.

### Documentation

We have [experimentally verified](https://github.com/enso-org/enso/pull/10783/commits/c032d9db5c0e083a7c1e0f9cfd68e8312c8e96d3#r1770787594) that with small changes in Helidon code we can use `@Substitute` mechanism of GraalVM _native image_ to switch to _platform threads_. As we don't care about massive performance, but just want to get things work, it is a sufficient solution for us.

Can you accept this little refactoring into next version of Helidon, please? Thank you very much.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ExecutorsFactory.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ExecutorsFactory.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


tomas-langer (approved):

JaroslavTulach:
Thank you very much guys! After [switching to Helidon 4.1.2](https://github.com/enso-org/enso/pull/10783/commits/ee82801c2e418ce37b18ccff276c1d8d5b3009fc) we can build our native image executable and it seems to work!

PR #9307, opened 2024-10-01
feat: Jwt -- don't force "upn" claim

### LINKED ISSUE #5151: unwanted "upd" claim in JWT - Oracle IDCS/IAM Domains having problems
https://github.com/helidon-io/helidon/blob/84123e8caa7f6671efc396526404b2b39d0afbbf/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java#L273-L275

These lines enforce that a JWT contains the claim `upd`. Unfortunately it is impossible to suppress this behavior. When you look at https://docs.oracle.com/en/cloud/paas/identity-cloud/rest-api/ClientAssertion.html, you see JWTs that Oracle IDCS/IAM Domains would expect. With the lines above the request will always fail as extra/unexpected claims lead to rejection

manger:
### Description

Fixes #5151

Adjusts the `Jwt` class so that `userPrincipal()` returns the value of a `upn`, `preferred_username`, or `sub` claim in that order of preference; but the class does not automatically insert a `upn` claim whenever a `preferred_username` or `sub` is present. This means `userPrincipal()` returns a value to satisfy the [Eclipse MicroProfile Interoperable JWT RBAC](https://download.eclipse.org/microprofile/microprofile-jwt-auth-2.1/microprofile-jwt-auth-spec-2.1.html). But the `Jwt` class is still usable for use-cases beyond that very specific profile where a `upn` claims in not required, not desired, and may even be forbidden.

A `upn` claim can still be set.

Unit tests are added to confirm that `Jwt.userPrincipal()` falls back to `sub` if `upn` is absent; and to confirm that `upn` is no longer automatically (and unexpectedly) added just because `sub` is set.

All the existing unit tests still pass.

### Documentation

The javadoc for `Jwt.userPrincipal()` and `Jwt.Builder.userPrincipal(String)` is adjusted.

### DIFF
--- a/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
+++ b/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
@@ -307,9 +307,7 @@ public class Jwt {
         this.cHash = JwtUtil.getByteArray(payloadJson, C_HASH, "c_hash value");
         this.nonce = JwtUtil.getString(payloadJson, NONCE);
         this.scopes = JwtUtil.toScopes(payloadJson);
-        this.userPrincipal = JwtUtil.getString(payloadJson, USER_PRINCIPAL)
-                .or(() -> preferredUsername)
-                .or(() -> subject);
+        this.userPrincipal = JwtUtil.getString(payloadJson, USER_PRINCIPAL);
     }
 
     private Jwt(Builder builder) {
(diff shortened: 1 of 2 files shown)


romain-grecourt:
@manger Do you intend on signing the OCA ?

manger:
Yes, I intend to sign the contributor agreement. Just need to navigate the org processes to do so.

Captain1653:
I can create PR with these changes from my account if it's still needed. Can I do it?

manger:
Yes, please do that.
I would still like to sign the contributor agreement but the process has been delayed with got complications. Please fix the issue without this specific PR.

Captain1653:
@manger FYI https://github.com/helidon-io/helidon/pull/9640 . 

Don't hesitate to write your comments/suggestions :)

PR #9313, opened 2024-10-02
Allows the webserver's write buffer size to be set to 0

spericas:
### Description

Allows the webserver's write buffer size to be set to 0 (or less). The underlying output stream shall not be wrapped into a buffered output stream one unless this value is greater than 0. See issue #9276.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
@@ -175,7 +175,8 @@ interface ListenerConfigBlueprint {
 
     /**
      * Initial buffer size in bytes of {@link java.io.BufferedOutputStream} created internally to
-     * write data to a socket connection. Default is {@code 4096}.
+     * write data to a socket connection. Default is {@code 4096}. Set buffer size to a value
+     * less than one to turn off buffering.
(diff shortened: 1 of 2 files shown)


tjquinno (approved):

VerKWer:
I guess the PR #9277 concerning the same issue can be closed as well then.

spericas:
> I guess the PR #9277 concerning the same issue can be closed as well then.

 Sorry, was not aware of that PR. I guess so then.

> Edit: Out of curiosity: I noticed you still wrap the `OutputStream` even in the unbuffered case. I know that this is not really significant but is the additional heap allocation of the wrapper really necessary?

Where is that exactly?

VerKWer:
Oh it's no problem at all. I'm happy that my issue gets fixed and glad for all the work people like you are putting in. I hope you don't mind my minor nitpick about the wrapping. The only reason I'm interested is that for the PR I proposed, no `ClosingBufferedOutputStream` is created if the buffer size is `0` (which was basically the situation before the buffer was introduced in the first place). But I'm not sure if the wrapper is needed further down the line, because I only have superficial familiarity with the code base.

spericas:
Yes, `ClosingBufferedOutputStream` is needed. Hopefully the PR addresses your use case. Should be in 4.1.2.

Issue #9243, opened 2024-09-09
HTTP trailers do not force chunked transfer encoding


rclewley:
## Environment Details
* Helidon Version: 4.10
* Helidon MP
* JDK version: JDK 21
* OS: macOS
* HTTP/1.1
* Docker version (if applicable):

----------

## Problem Description

If trailers are added to the a response, unless the transfer encoding has already been forced to chunked encoding, they will be output directly at the end of the request body of a non-chunked response. Trailers are only valid in chunked encoded responses in HTTP/1.1. 

As a result a couple of problems arise 
* Some clients, can't read the trailers, e.g Apache HTTP 5. Helidon WebClient does read them. 
* If the trailers aren't read, and they remain on the TCP stream, they will break the next request that re-uses the connection. 

[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"

If we augment the SimpleGreeter with some trailers

```
    public Message getDefaultMessage(@Context ResourceContext context) {
        String msg = String.format("%s %s!", message, "World");
        Message message = new Message();
        message.setMessage(msg);

        // Add a trailer.
        ServerResponse response = context.getResource(ServerResponse.class);
        // Add the Trailer Header, which lists that tailer names.
        response.header(HeaderNames.TRAILER, "Server-Timing");
        // Add the trailer.
        response.trailers().add(HeaderNames.create("Server-Timing"), "db;dur=53, app;dur=47.2");
        return message;
    }
```
and make a request, and monitor the TCP stream we get

```GET /simple-greet HTTP/1.1
Host: localhost:62804
User-Agent: Helidon 4.1.0
Connection: keep-alive
Content-Length: 0
Proxy-Connection: keep-alive

HTTP/1.1 200 OK
Date: Wed, 4 Sep 2024 15:26:55 +0100
Connection: keep-alive
Content-Length: 26
Content-Type: application/json
Trailer: Server-Timing

{"message":"Hello World!"}stream-result: 
Server-Timing: db;dur=53, app;dur=47.2```

What I would expect is a chunked response with the trailers in the final chunk. 

```GET /simple-greet HTTP/1.1
Host: localhost:63383
User-Agent: Helidon 4.1.0
Connection: keep-alive
Content-Length: 0
Proxy-Connection: keep-alive

HTTP/1.1 200 OK
Date: Wed, 4 Sep 2024 22:00:03 +0100
Connection: keep-alive
Content-Type: application/json
Trailer: Server-Timing
[...]

PR #9334, opened 2024-10-07
Ensures chunked encoding is enabled in all cases when there are trailers

spericas:
### Description

Ensures chunked encoding is enabled whenever there are trailers to be sent, even if a Content-Length header was specified. New test that adds a trailer and sets a Content-Length header. See issue #9243.

### Documentation

None


tjquinno (approved):
LGTM

PR #9335, opened 2024-10-08
Faster Jackson default settings

re-thc:
### Description

By default Jackson does not enable faster features to ensure 100% backwards compatibility. They should be safe and are already in use by many projects.

Note that `USE_FAST_DOUBLE_WRITER` has been ported to Java 20 and so has been left out.

### Documentation

None


spericas:
LGTM. Any concerns @tomas-langer?

tomas-langer (approved):

spericas (approved):

PR #9337, opened 2024-10-08
3.x: Support built-in health check config at `health.checks` (retaining now-deprecated `helidon.health` for compatibility)

### LINKED ISSUE #9322: 3.x Inconsistent config prefixes for health-related settings
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
The config keys for health-related settings mostly start with `health.` but for the built-in health checks which themselves are configurable those keys start with `helidon.health.` (such as `helidon.health.diskSpace.thresholdPercent` for example).

With the fix in place for #8617 the doc and the code now agree for all the settings, but the inconsistency among the config keys is awkward and confusing.

The settings which start with `helidon.health.` should be just `health.` instead. But we'll need to maintain backward compatibility
so we probably need to support both prefix styles for the settings that currently start with `helidon.health.`. There's an existing utility class that handles deprecated config keys including logging warnings.

tjquinno:
### Description
Resolves #9322 

The PR allows (and encourages, through changes in the doc and Javadoc) the use of `health.checks` (preferred over `helidon.health`) as the prefix for config keys to control the built-in health checks.

The now-deprecated `helidon.health` prefix still works although its use triggers a warning courtesy of the new utility class `DeprecatedMpConfig`. (This is patterned closely after the existing `DeprecatedConfig` class but that does not work with proxied MP `Config` objects.)

There are some public constants which retain their old values (even though the syntax for how they are computed has changed) so any user code relying on the previous values will continue to work unchanged.

The Javadoc and the `.adoc` doc now refer to `health.checks` as the config prefix.

A few tests added.

### Documentation
Doc is updated.

### DIFF
--- a/config/config-mp/src/main/java/io/helidon/config/mp/DeprecatedMpConfig.java
+++ b/config/config-mp/src/main/java/io/helidon/config/mp/DeprecatedMpConfig.java
@@ -13,5 +13,62 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.helidon.config.mp;public class DeprecatedMpConfig {
+package io.helidon.config.mp;
+
+import java.util.Optional;
+import java.util.logging.Logger;
+
+import org.eclipse.microprofile.config.Config;
+
+/**
+ * A utility class to handle MicroProfile configuration properties that should no longer be used.
+ * <p>
+ * For one major release, the property is retrieved through this class, to warn about the usage.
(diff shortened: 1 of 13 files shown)


ljnelson (approved):
LGTM

spericas (approved):

barchetta (approved):

PR #9352, opened 2024-10-09
4.x: Ensures there is no trailing whitespace in Java properties lines in persistence.adoc with Asciidoctor callouts at the expense of visual cleanliness

### LINKED ISSUE #9348: Persistence guide for UCP has incorrect JDBC URL property
https://github.com/helidon-io/helidon/blob/d47f53aea182799c1da5f1172b8acf289a24b3a8/docs/src/main/asciidoc/mp/persistence.adoc?plain=1#L438

There are two issues here:
1. `url` is not a supported Java Beans property; `URL` is. The documentation needs to be changed to reflect this.
2. Due to the way that Asciidoc does callouts, the trailing space at the end of the line before the pound sign (`#`) ends up being significant and the UCP integration passes it along as part of the URL. Because users often copy and paste things as-is, this needs to be fixed, either in the docs or (maybe? not sure?) as part of the UCP integration itself (i.e. by `trim()`ming user input).

### LINKED ISSUE #9350: Trailing whitespace needs to be removed in Persistence Guide properties snippets
https://github.com/helidon-io/helidon/blob/ed6cf6544616ba44c264a3d4ab7053b8025a981c/docs/src/main/asciidoc/mp/persistence.adoc?plain=1#L437

Note the trailing space at the end of the line before the pound sign (`#`), which exists only for the Asciidoc callout. I am not sure if this whitespace can be removed while preserving the callout but I need to look to see if it can.

Without a fix, a user who copies text from the persistence guide's rendered output will get the trailing space as part of the copy, and then the UCP won't be able to load the supplied class, because there is whitespace as part of the name.

It is possible as well that something like `String.trim()` should be applied selectively, but that means that our data source integration stuff will now "know" about certain "special" properties that need to be trimmed (at the moment it is entirely and deliberately agnostic). Of course one could also argue that the UCP or the data source or the driver should do this and I'm sure a strong case could be made for or against any or all of this.

To start with I will see if I can somehow eliminate the trailing space from the documentation.

ljnelson:
This PR updates the Persistence Guide so that a copy and paste of Java properties excerpts does not include trailing whitespace at the end of the line. (It also corrects a data source property, `url`, which should have been `URL`. See #9348.)

The tradeoff is that the Asciidoctor callout glyph is now smashed up against the end of the line.

Screenshot 2024-10-09 at 2 56 29 PM

Closes #9350.

Closes #9348.

### DIFF
--- a/docs/src/main/asciidoc/mp/persistence.adoc
+++ b/docs/src/main/asciidoc/mp/persistence.adoc
@@ -327,7 +327,7 @@ the `test` data source's associated connection pool or vendor-specific
 
 [source,properties]
 ----
-javax.sql.DataSource.test.foo.bar=baz# <1> <2> <3>
+javax.sql.DataSource.test.foo.bar=baz# <1><2><3>
 ----
 <1> The *_objecttype_* portion of the configuration property name is
     `javax.sql.DataSource`.
@@ -489,7 +489,7 @@ password:
 [source,properties]
 ----
 javax.sql.DataSource.test.dataSourceClassName = org.h2.jdbcx.JdbcDataSource# <1>
-javax.sql.DataSource.test.dataSource.url = jdbc:h2:mem:unit-testing;DB_CLOSE_DELAY=-1# <2> <3>
+javax.sql.DataSource.test.dataSource.url = jdbc:h2:mem:unit-testing;DB_CLOSE_DELAY=-1# <2><3>
 javax.sql.DataSource.test.dataSource.user = sa
 javax.sql.DataSource.test.dataSource.password =
 ----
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/mp/persistence.adoc
```diff
@@ -327,7 +327,7 @@ the `test` data source's associated connection pool or vendor-specific
 
 [source,properties]
 ----
-javax.sql.DataSource.test.foo.bar=baz # <1> <2> <3>
+javax.sql.DataSource.test.foo.bar=baz# <1> <2> <3>
```
barchetta, line 330: I think you need to remove spaces between callouts too
`javax.sql.DataSource.test.foo.bar=baz# <1><2><3>`

barchetta (approved):

PR #9353, opened 2024-10-10
4.x: Amends datasource-ucp to depend on ucp11 instead of ucp

### LINKED ISSUE #9349: UCP integration needs to depend on ucp11, not ucp
Because of https://github.com/helidon-io/helidon/commit/ef6649f236ecd3d35ec96a138c0735ad7a54856f, the Persistence Guide now makes reference to our `ojdbc` integration. That integration brings in the version of UCP that was compiled with JDK 11 (`ucp11`). The actual UCP CDI integration brings in the version of UCP that was compiled with JDK 8 (`ucp`), which is still supported. Nevertheless, the UCP CDI integration should include `ucp11`, not `ucp`.

Without a fix, users will end up with two copies of UCP in their classpath (`ucp` and `ucp11`) at runtime.

ljnelson:
This PR changes `integrations/cdi/datasource-ucp/pom.xml` to depend on `ucp11` instead of `ucp`, since the `pom.xml` was [previously amended](https://github.com/helidon-io/helidon/blob/4.1.2/integrations/cdi/datasource-ucp/pom.xml#L43-L47) (via https://github.com/helidon-io/helidon/commit/ef6649f236ecd3d35ec96a138c0735ad7a54856f) to include a dependency on our `ojdbc` integration rather than on the Oracle JDBC driver directly&mdash;a decision which had some unintended ripple effects. 

Part of this change seems to have been to [include the full production suite of artifacts related to the driver](https://github.com/helidon-io/helidon/blob/4.1.2/integrations/db/ojdbc/pom.xml#L42-L45). Notably, the suite of artifacts also includes `ucp11`, along with several other jars related to observability and so on.

Our `ojdbc` integration excludes `ucp` from its dependency tree. But that dependency is never included (despite the [Developers Guide For Oracle JDBC on Maven Central](https://www.oracle.com/database/technologies/maven-central-guide.html) claiming (incorrectly) that it is (see "Use Case 1: I Want the Production Jars", second bullet point, and contrast with what [the relevant `pom.xml` actually says](https://search.maven.org/artifact/com.oracle.database.jdbc/ojdbc11-production/21.9.0.0/pom))), so the exclusion is effectively a no-op. I'll fix this in a subsequent PR (though I'm not the original author, it seems simple enough).

Anyway, the net effect of all this, as #9349 indicates, is that a user of this project will currently end up with two copies of UCP on the classpath: one compiled with JDK 8 (`ucp`) dragged in by `integrations/cdi/datasource-ucp`, and another compiled with JDK 11 (`ucp11`) (dragged in by `integrations/db/ojdbc`).  The intent of this PR is to fix this problem and remove at least some of this confusion.

Closes #9349.

### DIFF
--- a/integrations/cdi/datasource-ucp/pom.xml
+++ b/integrations/cdi/datasource-ucp/pom.xml
@@ -47,7 +47,7 @@
         </dependency>
         <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
-            <artifactId>ucp</artifactId>
+            <artifactId>ucp11</artifactId>
             <scope>compile</scope>
         </dependency>
         <dependency>
--- a/integrations/cdi/datasource-ucp/pom.xml
+++ b/integrations/cdi/datasource-ucp/pom.xml
@@ -41,11 +41,6 @@
             <scope>compile</scope>
         </dependency>
         <dependency>
-            <groupId>io.helidon.integrations.db</groupId>
-            <artifactId>ojdbc</artifactId>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
             <artifactId>ucp11</artifactId>
             <scope>compile</scope>
@@ -75,10 +70,9 @@
 
         <!-- Runtime-scoped dependencies. -->
         <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>jandex</artifactId>
+            <groupId>io.helidon.integrations.db</groupId>
+            <artifactId>ojdbc</artifactId>
             <scope>runtime</scope>
-            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>io.helidon.microprofile.config</groupId>
@@ -86,6 +80,13 @@
             <scope>runtime</scope>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye</groupId>
+            <artifactId>jandex</artifactId>
+            <scope>runtime</scope>
+            <optional>true</optional>
+        </dependency>
+
         <!-- Test-scoped dependencies. -->
         <dependency>
             <groupId>org.junit.jupiter</groupId>
--- a/integrations/db/ojdbc/pom.xml
+++ b/integrations/db/ojdbc/pom.xml
@@ -45,10 +45,6 @@
             <type>pom</type>
             <exclusions>
                 <exclusion>
-                    <groupId>com.oracle.database.jdbc</groupId>
-                    <artifactId>ucp</artifactId>
-                </exclusion>
-                <exclusion>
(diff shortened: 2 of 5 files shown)


integrations/cdi/datasource-ucp/pom.xml
```diff
@@ -75,17 +70,23 @@
 
         <!-- Runtime-scoped dependencies. -->
         <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>jandex</artifactId>
+            <groupId>io.helidon.integrations.db</groupId>
+            <artifactId>ojdbc</artifactId>
```
ljnelson, line 74: The prior version had our `ojdbc` integration artifact brought in at `compile` time, which is not necessary (only `ucp11` is necessary at compile time). That means, too, that perhaps this dependency is not needed? I should say: it _definitely_ is not needed but I seem to recall others having strong opinions here, i.e. that pulling in the complete suite of Oracle production-level `ojdbc`-related artifacts at `runtime` is a convenient and proper thing to do. Personally I would vote to remove this dependency entirely and put the onus on the user to include it if she wants. What say others, particularly @arjav-desai  and @barchetta?

ljnelson: Sounds reasonable.

integrations/db/ojdbc/pom.xml
```diff
@@ -44,10 +44,6 @@
             <artifactId>ojdbc11-production</artifactId>
             <type>pom</type>
             <exclusions>
```
ljnelson, line 46: Note: I deliberately removed a non-functional `exclusion` that was excluding the `ucp` artifact, which does not appear in the `ojdbc11-production:pom` artifact's dependencies (thus having no effect). Since roughly February users have had `ucp11` on their classpath, since it was not being excluded. This PR preserves `ojdbc11-production:pom`'s set of dependencies, with the sole exception of `xmlparserv2` (we continue to exclude that and include `xmlparserv2_sans_jaxp_services` in its place as before).

ljnelson: Per instructions I will be re-introducing an `<exclusion>`, but this time one that will exclude `ucp*` (most particularly `ucp11`). This means indirect consumers of this `pom`-typed artifact who might have been accustomed to having `ucp11` on their classpath will now have to include `ucp11` explicitly. We need to update the examples to make this easy to understand.

ljnelson:
Partial mystery solved: the Oracle JDBC `ojdbc11` artifact, at version 21.15.0.0, has the following in its `native-image.properties` file it ships with:
```
  -H:EnableURLProtocols=http \
  -H:EnableURLProtocols=https
```
These lines are not present in the `native-image.properties` file shipped with `ojdbc11` version 23.6.0.24.10 (nor, I would think, should they be). So that explains that issue.

ljnelson:
Closing; incorporated in #9441

Issue #9354, opened 2024-10-10
h2c HTTP/1.1 upgrade to HTTP/2 fails if there is a request body


manger:
On upgrading a connection from HTTP/1.1 to HTTP/2 with `Upgrade: h2c` Helidon web server throws with `IllegalStateException: Invalid HTTP/2 connection preface` because it fails to read the request body before looking for the preface.

## Environment Details
* Helidon Version: 4.1.0
* Helidon SE
* JDK version: openjdk 21.0.2 2024-01-16
* OS: macOS Sonoma 14.7 (Darwin 23.6.0 arm64)
* Docker version (if applicable): n/a

----------

## Problem Description

A Helidon webserver supporting HTTP/2 (by depending on io.helidon.webserver:helidon-webserver-http2) tries to upgrade a POST request that has a `Upgrade: h2c` request header by returning `101 Switching Protocols` and `Upgrade: h2c`. The server then expects the HTTP/2 preface (`PRI * HTTP/2.0`), but fails as the request body comes before the preface. Perhaps an upgrade works for a GET, but not for requests with a body.

For a sample client request:
```
curl --http2 -d token=2977 http://127.0.0.1:8080/myapp
```
the server exception is:
```
WebServer socket] WARN io.helidon.webserver.http1.Http1Connection - Internal server error
io.helidon.http.RequestException: Internal error
        at io.helidon.http.RequestException$Builder.build(RequestException.java:139)
        at io.helidon.webserver.http1.Http1Connection.handle(Http1Connection.java:226)
        at io.helidon.webserver.ConnectionHandler.run(ConnectionHandler.java:165)
        at io.helidon.common.task.InterruptableTask.call(InterruptableTask.java:47)
        at io.helidon.webserver.ThreadPerTaskExecutor$ThreadBoundFuture.run(ThreadPerTaskExecutor.java:239)
        at java.base/java.lang.VirtualThread.run(VirtualThread.java:309)
Caused by: java.lang.IllegalStateException: Invalid HTTP/2 connection preface: 
+--------+-------------------------------------------------+----------------+
|   index|  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |            data|
+--------+-------------------------------------------------+----------------+
|00000000| 74 6f 6b 65 6e 3d 32 39 37 37                   |token=2977      |
+--------+-------------------------------------------------+----------------+
+--------+-------------------------------------------------+----------------+
[...]

PR #9359, opened 2024-10-10
4.x: Ignore protocol upgrades when initial request contains a payload

spericas:
### Description

Ignore protocol upgrades when initial request contains a payload. Payloads should not be sent until after the negotiation is complete to avoid interfering with the upgrade mechanism. Logs decision to ignore upgrade. New test. See issue #9354.

### DIFF
--- a/webserver/tests/upgrade/src/test/java/io/helidon/webserver/tests/upgrade/test/Http2UpgradeTest.java
+++ b/webserver/tests/upgrade/src/test/java/io/helidon/webserver/tests/upgrade/test/Http2UpgradeTest.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.webserver.tests.upgrade.test;
+
+import io.helidon.http.HeaderNames;
+import io.helidon.http.Status;
+import io.helidon.logging.common.LogConfig;
+import io.helidon.webclient.http1.Http1Client;
+import io.helidon.webclient.http2.Http2Client;
+import io.helidon.webclient.http2.Http2ClientProtocolConfig;
+import io.helidon.webserver.WebServer;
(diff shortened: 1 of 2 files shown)


webserver/tests/upgrade/src/test/java/io/helidon/webserver/tests/upgrade/test/Http2UpgradeTest.java
```diff
@@ -0,0 +1,84 @@
...
+                    .baseUri("http://localhost:" + port + "/echo")
+                    .build();
+            try (var res = http1Client.post()
+                    .header(HeaderNames.UPGRADE, "h2c")     // upgrade pretend
+                    .submit("payload")) {
+                assertThat(res.status(), is(Status.NOT_FOUND_404));
+            }
+
+            // verify success with smart client when payload is sent after upgrade
+            Http2Client http2Client = Http2Client.builder()
```
danielkec, line 68: I would also test second request on the same connection possibly with some other client, [vertx](https://github.com/helidon-io/helidon/blob/aa1e1d8e8209e3cfdd481715fd467d0a3dfca1d4/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/Continue100Test.java#L59) for example.

danielkec (changes requested):

webserver/tests/upgrade/src/test/java/io/helidon/webserver/tests/upgrade/test/Http2UpgradeTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.jupiter.api.Test;
+
+import static io.helidon.http.Method.POST;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+class Http2UpgradeTest {
+
+    @Test
+    void testHttp2UpgradeWithPayload() {
```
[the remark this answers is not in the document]
spericas: I didn't use it because the test creates multiple clients for h1 and h2 as part of the same test, seems cleaner this way I think

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
```diff
@@ -196,6 +196,8 @@ public void handle(Limit limit) throws InterruptedException {
                                 return;
                             }
                         }
+                    } else {
+                        ctx.log(LOGGER, WARNING, "Protocol upgrade for a request with a payload ignored");
```
danielkec, line 200: Wouldn't be a debug level better?

spericas: Maybe, but I would think this is something to be reported more as a warning to indicate that _upgrades are ignored due to an invalid upgrade request_. As DEBUG, it will likely always be unreported. After all, this is a condition that shouldn't really happen, and if it does, then it is nice to know about it.

danielkec (approved):

PR #9362, opened 2024-10-11
Fix DataReader.findNewLine with multiple lone EOL character

### LINKED ISSUE #9365: InsufficientDataAvailableException while reading Multipart request part data
## Environment Details
* Helidon Version: 4.1.2
* JDK version: java 22.0.2 2024-07-16 GraalVM 22.0.2+9.1 (build 22.0.2+9-jvmci-b01)
* OS: MacOS

----------

## Problem Description
Getting `io.helidon.common.buffers.DataReader$InsufficientDataAvailableException` while reading Multipart request part data in Helidon SE 4.1.2 application

The attached `helidon-multipart-example.zip` file contains a Helidon SE 4.1.2 application with a `/multi` endpoint that reproduces the issue.

I used two files: one is 8KB and the other is 90KB. For the 90KB file, I'm getting an InsufficientDataAvailableException, while the 8KB file works as expected with no errors.

### CURL
curl --location 'http://localhost:8080/multi' --form 'primaryFile=@"/path/to/your/success.pdf"'

curl --location 'http://localhost:8080/multi' --form 'primaryFile=@"/path/to/your/failure.pdf"'
[failure.pdf](https://github.com/user-attachments/files/17338261/failure.pdf)
[helidon-multipart-example.zip](https://github.com/user-attachments/files/17338262/helidon-multipart-example.zip)
[success.pdf](https://github.com/user-attachments/files/17338263/success.pdf)

## Error
`
Caused by: io.helidon.common.buffers.DataReader$InsufficientDataAvailableException
	at io.helidon.common.buffers.DataReader.pullData(DataReader.java:83)
	at io.helidon.common.buffers.DataReader.getBuffer(DataReader.java:225)
	at io.helidon.common.buffers.DataReader.readBuffer(DataReader.java:202)
	at io.helidon.http.media.multipart.ReadablePartNoLength$PartInputStream.readBuffer(ReadablePartNoLength.java:176)
	at io.helidon.http.media.multipart.ReadablePartNoLength$PartInputStream.ensureBuffer(ReadablePartNoLength.java:163)
	at io.helidon.http.media.multipart.ReadablePartNoLength$PartInputStream.read(ReadablePartNoLength.java:135)
	at java.base/java.io.InputStream.readNBytes(InputStream.java:412)
	at java.base/java.io.InputStream.readAllBytes(InputStream.java:349)
`

## Steps to reproduce

1. Create a Helidon SE 4.1.2 application.
2. Implement a multipart handler
3. Attempt to read the content of a multipart part

romain-grecourt:
### Description

`DataReader.findNewLine` returns max instead of new line index when there are multiple lone CR characters in the same buffer.

E.g. `\r\r\r\n`

The virtual index is incorrectly accumulated each time with the relative index, its value may exceed max which will skip any valid CRLN within max.

- Fix virtual index accumulation
- Re-use the branch that changes the buffer
- Add unit tests

Fixes #9365

### Documentation

None.

### DIFF
--- a/common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
@@ -395,56 +395,47 @@ public class DataReader {
      */
     public int findNewLine(int max) throws IncorrectNewLineException {
         ensureAvailable();
-        int idx = 0;
         Node n = head;
-        int indexWithinNode = n.position;
+        int idx = 0;
+        int fromIndexNode = n.position;
 
         while (true) {
             byte[] barr = n.bytes;
-            int maxLength = Math.min(max - idx, barr.length - indexWithinNode);
-            int crIndex = Bytes.firstIndexOf(barr, indexWithinNode, indexWithinNode + maxLength, Bytes.CR_BYTE);
+            int maxLength = Math.min(max - idx, barr.length - fromIndexNode);
+            int crIndexNode = Bytes.firstIndexOf(barr, fromIndexNode, fromIndexNode + maxLength, Bytes.CR_BYTE);
 
-            if (crIndex == -1) {
-                int lfIndex = Bytes.firstIndexOf(barr, indexWithinNode, indexWithinNode + maxLength, Bytes.LF_BYTE);
-                if (lfIndex != -1) {
+            if (crIndexNode == -1) {
+                int lfIndexNode = Bytes.firstIndexOf(barr, fromIndexNode, fromIndexNode + maxLength, Bytes.LF_BYTE);
+                if (lfIndexNode != -1) {
                     if (!ignoreLoneEol) {
-                        throw new IncorrectNewLineException("Found LF (" + (idx + lfIndex - n.position)
+                        throw new IncorrectNewLineException("Found LF (" + (idx + lfIndexNode - n.position)
                                                                     + ") without preceding CR. :\n" + this.debugDataHex());
                     }
                 }
-                // not found, continue with next buffer
-                idx += maxLength;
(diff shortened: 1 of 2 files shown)


tomas-langer:
What is this trying to fix? There is no issue attached.

The `\r` is not allowed as a character in HTTP unless it is the end of the line, and that must be `\r\n` according to the spec. 
So in general, we must not allow this. 

So please provide an explanation/issue why this is a good solution and a general one. Thanks!

romain-grecourt:
> What is this trying to fix? There is no issue attached.

The fix was created before the issue, that's unusual !

> The `\r` is not allowed as a character in HTTP unless it is the end of the line, and that must be `\r\n` according to the spec. So in general, we must not allow this.

`io.helidon.common.buffers.DataReader` is part of `common/buffers` and `DataReader.findNewLine()` has no HTTP semantics other than `newLine` means `\r\n`.

`common/buffers` shouldn't be responsible for enforcing HTTP semantics.

> So please provide an explanation/issue why this is a good solution and a general one. Thanks!

The documented contract is `return index of the new line, or max if not found`, if the buffer contains `\r\n` but we return `max` that's a bug.

With the current code If a buffer contains `\r\r\r\n` then max is returned instead of the actual new line index.
The calling code will then pull data to search for the new line and that will trigger a pull which may fail.

---

I will update the issue with more explanations.
There may be another latent bug in multipart as well, if I'm able to reproduce it I will shift this PR toward multipart.

spericas (approved):
LGTM. As I mentioned offline, it may be worth renaming some variables to distinguish between node and stream indexes. Should help understand the calculations better.

spericas (approved):

PR #9398, opened 2024-10-16
4.x: Fix MockBean to work with argument matchers

### LINKED ISSUE #9397: 4.x: MockBean fails with ArgumentMatcher
## Environment Details
* Helidon Version: 4.0.12
* Helidon MP
----------

## Problem Description

If a test uses an argument matcher, it fails with the following error:

```
org.mockito.exceptions.misusing.InvalidUseOfMatchersException: 
Misplaced or misused argument matcher detected here:
```

---

When mocked instances are created in ApplicationScoped the injected instance is a client proxy, and the mocked instance is lazily created on first usage. This breaks when using argument matchers.

A work-around is to call `toString()` on the mocked instance to force the creation of the mock before using matchers on the mocked instance.

## Steps to reproduce

```java
@MockBean
private Service service;

@Test
void testArgumentMatcher() {
    Mockito.when(service.test(anyString())).thenReturn("Mocked");
}
```

### LINKED ISSUE #9411: 4.x: HelidonTestNgListener creates new AnnotatedType
## Environment Details
* Helidon Version: 4.1.2
* Helidon MP

----------

## Problem Description

`HelidonTestNgListener` lets TestNG create the test instance and then uses CDI to perform [injection](https://github.com/helidon-io/helidon/blob/ed6cf6544616ba44c264a3d4ab7053b8025a981c/microprofile/testing/testng/src/main/java/io/helidon/microprofile/testing/testng/HelidonTestNgListener.java#L218).

It uses `beanManager.createAnnotatedType(clazz)` which discards anything done by extensions with `ProcessAnnotatedType`. 

E.g `MockBeansCdiExtension` adds an `@Inject` annotation [programmatically](https://github.com/helidon-io/helidon/blob/ed6cf6544616ba44c264a3d4ab7053b8025a981c/microprofile/testing/mocking/src/main/java/io/helidon/microprofile/testing/mocking/MockBeansCdiExtension.java#L51).

## Steps to reproduce

```java
@HelidonTest
@AddBean(TestMockBeanNoInject.Service.class)
class TestMockBeanNoInject {

    @MockBean
    private Service service;

    @Test
    void testArgumentMatcher() {
        assertThat(service, is(not(nullValue())));
    }

    @ApplicationScoped
    static class Service {
        String test(String str) {
            return "Not Mocked: " + str;
        }
    }
}
```

romain-grecourt:
### Description

- Call toString() to force initialization of the mocked instances
- Use produceWith instead of createWith
- Use addTransitiveTypeClosure instead of addType to support more than just one type
- Minor refactoring of the processMockBean method
- Re-work HelidonTestNgListener to initialize the testInstance with an extension
- Add tests

Fixes #9397
Fixes #9411

### Documentation

None.

### DIFF
--- a/microprofile/testing/mocking/etc/spotbugs/exclude.xml
+++ b/microprofile/testing/mocking/etc/spotbugs/exclude.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2024 Oracle and/or its affiliates.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+
+<FindBugsFilter
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="https://github.com/spotbugs/filter/3.0.0"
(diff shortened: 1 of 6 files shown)


jbescos:
It would be better to add the test also in testng module.

jbescos (approved):

abdelhak-zaaim (approved):

Issue #9396, opened 2024-10-15
Helidon-Se: 4.1.0 500 error when there is no-content instead of 204


ksivap147:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: Helidon-SE 4.1.0
* Helidon SE 
* JDK version: JDK21
* OS: Mac OS
* Docker version (if applicable): NA

----------

## Problem Description
I am seeing this behavior with Helidon 4.1.0 w.r.t 204 Status Code and gzip encoding. Client is sending a POST request for which there is no data, so the server sets status to 204 and does not send any content.
When there is no Accept-Encoding in request, client gets 204 and no body.
But when Accept-Encoding is gzip, even though I am sending a status of 204, client gets a 500. With this in the server log:
Attempt to send status 500 Internal Server Error with entity. Server responded with Internal Server Error. Please fix your routing, this is not allowed by HTTP specification, such responses MUST NOT contain an entity.
I believe this is because of the "Transfer-Encoding" header being added with value as chunked.

Request without gzip:
--------------
curl --request POST \
  --url http://100.111.154.123:8080/extract/ap \
  --header 'Accept: application/json' \
  --header 'Content-Type: application/json' \
  --data '....'

status: 
--------------
204 No Content

Response headers:
--------------
Content-Length: 0
Strict-Transport-Security: max-age=31536000; includeSubDomains
opc-request-id: /50cbc6a4b355b1a7bc19c5761f1bec38/50cbc6a4b355b1a7bc19c5761f1bec38

Request with gzip:
--------------
curl --request POST \
  --url http://100.111.154.123:8080/extract/ap \
  --header 'Accept: application/json' \
  --header 'Accept-Encoding: gzip' \
  --header 'Content-Type: application/json' \
  --data '...'

status: 
--------------
500 Internal Server Error

Response headers:
--------------
Content-Encoding: gzip
Strict-Transport-Security: max-age=31536000; includeSubDomains
Transfer-Encoding: chunked
## Steps to reproduce
[//]: # "Step by step instructions to reproduce the problem"
[//]: # "Provide sample code/application if relevant"

[...]

PR #9408, opened 2024-10-16
4.x: Explicitly sets the content length to zero when a 204 or related status code is returned

spericas:
### Description

Explicitly sets the content length to zero when a 204 or related status code is returned to avoid validation errors at a later time if chunked encoding is selected. New test. See issue #9396.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
@@ -165,7 +165,7 @@ class Http1ServerResponse extends ServerResponseBase<Http1ServerResponse> {
         if (isNoEntityStatus) {
             if (!headers.contains(HeaderNames.CONTENT_LENGTH)) {
                 headers.set(HeaderValues.CONTENT_LENGTH_ZERO);
-            } else if (headers.get(HeaderNames.CONTENT_LENGTH).asLong().getLong() > 0L) {
+            } else if (headers.get(HeaderNames.CONTENT_LENGTH).getLong() > 0L) {
                 throw new IllegalStateException("Cannot set status to " + status + " with header "
                                                         + HeaderNames.CONTENT_LENGTH + " greater than zero");
             }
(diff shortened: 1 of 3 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -693,9 +698,15 @@ private void sendFirstChunkOnly() {
         }
 
         private void sendHeadersAndPrepare() {
+            Status usedStatus = status.get();
+
             if (headers.contains(HeaderNames.CONTENT_LENGTH)) {
                 contentLength = headers.contentLength().orElse(-1);
                 isChunked = false;
+            } else if (isNoEntityStatus(usedStatus)) {
```
tomas-langer, line 706: I think this should be done in the `io.helidon.webserver.http.ServerResponse#status(io.helidon.http.Status)` only - when you set status, we should do the update of content length (unless already set). This seems too late.
There should also be a guarantee that you CANNOT send entity once you do this - i.e. if the user sets explicitly `chunked` transfer encoding, and you set `Content-Length=0`, and then the user sends an entity using an output stream, we have to provide a state that will not break the connection, and at the same time will provide a reasonable error message.

tomas-langer (changes requested):

abdelhak-zaaim (approved):

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
spericas: That approach will require a number of changes in other places. The related issue is undoubtedly an edge case, where the output stream is requested before the status is set; our internal logic is currently switching to chunked encoding when the output stream is requested, and setting the status at a later time will result in other exceptions in `BlockingOutputStream`. 

@ ksivap147 How much is your logic affected if you handle the 204 case separately, and not after the output stream was requested from the response?

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -159,6 +155,24 @@ static void nonEntityBytes(ServerResponseHeaders headers,
...
+    public Http1ServerResponse status(Status status) {
+        // set internal state
+        super.status(status);
+        isNoEntityStatus = isNoEntityStatus(status);
+
+        // check consistency if status code should not include entity
+        if (isNoEntityStatus) {
+            if (!headers.contains(HeaderNames.CONTENT_LENGTH)) {
+                headers.set(HeaderValues.CONTENT_LENGTH_ZERO);
+            } else if (headers.get(HeaderNames.CONTENT_LENGTH).asLong().getLong() > 0L) {
```
tomas-langer, line 168: `headers.contentLength().getAsLong()` can be used instead

tomas-langer (approved):

PR #9458, opened 2024-11-02
4.x: Replace deprecated method Scheduling.fixedRateBuilder() on Scheduling.fixedRate() in CoordinatorService

Captain1653:
### Description
I've found, that the deprecated method Scheduling.fixedRateBuilder() is used in the code. The method was marked as deprecated in [this commit](https://github.com/helidon-io/helidon/commit/e492423826d003cbf328cc9a0eb37df63aad5367#diff-00241b86b77b293b3a01442977eda4462b33de67f1d1d514fb5cb26bda7517a4R53).

image

It can be replaced on Scheduling.fixedRate()


danielkec:
Public API method removal is possible only in next major version, 5 in this case

Captain1653:
If it's ok to use deprecated methods until major version and refactor then no problem.

Should I close my PR, right?

danielkec (approved):
Oh my bad, this is usage not the method itself. Sorry for that.

romain-grecourt (approved):

PR #9497, opened 2024-11-13
2.x: netty 4.1.115, grpc-java 1.65.1

barchetta:
## Description

- Upgrade Netty to 4.1.115
- Upgrade grpc-java to 1.65.1 (matches what is currently in main. See #9105). Required by Netty upgrade.
- google-protobuf is already at 3.25.5 (matches what is in main)
- Drops dependency on protobuf lite in favor of the full version which we already depended on (like main).
- Upgrade dependency check plugin to 11.1.0


spericas (approved):
LGTM

Issue #9062, opened 2024-07-30
DB Client (PostgreSQL) Mapper failure on Boolean mapping


Isax03:
## Environment Details
* Helidon Version: 4.0.10
* Helidon SE
* JDK version: 21
* OS: Windows and Ubuntu

----------

## Problem Description

In the implementation of a `DbMapper` for an entity following the Entity Pattern, we encountered an issue when mapping boolean values from a PostgreSQL database. Specifically, the error arises when attempting to map PostgreSQL boolean types to `java.lang.Boolean` in Kotlin. The mapper is designed to extract various columns from a database row, including several boolean fields.

Here is a simplified version of the code causing the issue:

```kotlin
class EntityMapper : DbMapper<Entity> {
    override fun read(row: DbRow): Entity {
        val field1 = row.column("field1").get(UUID::class.java)
        val field2 = row.column("field2").get(String::class.java)
        val booleanField = row.column("boolean_field").get(Boolean::class.java)
        // Other fields...

        return Entity(
            field1 = field1,
            field2 = field2,
            booleanField = booleanField,
            // Other fields...
        )
    }

    // Other methods...
}
```

The relevant portion of the error message is:
```
io.helidon.http.RequestException: Failed to map class java.lang.Boolean to boolean: Failed to find mapper. Qualifiers: dbclient, source of class 'java.lang.Boolean'
...
Caused by: io.helidon.common.mapper.MapperException: Failed to map class java.lang.Boolean to boolean: Failed to find mapper.
...
```
This issue consistently occurs when mapping boolean fields from the database. The expectation was that the boolean fields in the database, represented as `boolean` in PostgreSQL, would be seamlessly mapped to `java.lang.Boolean` in the application layer. However, the error indicates that the mapping infrastructure does not have a predefined mapper for this conversion, resulting in a `MapperException`.

## Steps to Reproduce

1. Implement a `DbMapper` for an entity, including boolean fields, using Helidon DB Client.
2. Map the PostgreSQL boolean type fields to `java.lang.Boolean` in Kotlin.
3. Attempt to retrieve data from the database using the mapper.
4. Observe the error in the server logs indicating a failed mapping operation due to the lack of a suitable mapper for `java.lang.Boolean`.

This issue is reproducible consistently under the described conditions.

## Questions
1. Is there a specific configuration or additional setup required in Helidon to handle mapping between PostgreSQL boolean types and java.lang.Boolean?
[...]

PR #9514, opened 2024-11-20
Issue #9062 - DB Client (PostgreSQL) Mapper failure on Boolean mapping

Tomas-Kraus:
Added mappers for numbers and boolean values. Added boolean attribute to entity in integration tests.
Also, while writting the tests, I discovered limitations in `indexedParam` method of `DbStatement`. Added new method to allow specific parameters selection.
Fixes issue #9062.


Tomas-Kraus:
This won't be included in Helidon.

PR #9516, opened 2024-11-20
4.x: Update OCI SDK to 3.54.0

barchetta:
### Description

 Update OCI SDK to 3.54.0


ljnelson (approved):
LGTM assuming tests pass, and, as usual, recognizing that every upgrade is usually backwards incompatible in a minor way; see for example https://github.com/oracle/oci-java-sdk/releases/tag/v3.54.0 and look for Breaking Changes

Issue #9282, opened 2024-09-24
We should run h2spec on pipeline 


danielkec:
We should run h2spec test on the pipeline

Issue #9273, opened 2024-09-23
h2spec http2 5.1.2/1 fails


danielkec:
## Environment Details
* Helidon Version: 4.1.1

----------

## Problem Description
```
Hypertext Transfer Protocol Version 2 (HTTP/2)
  5. Streams and Multiplexing
    5.1. Stream States
      5.1.2. Stream Concurrency
        using source address 127.0.0.1:59142
        × 1: Sends HEADERS frames that causes their advertised concurrent stream limit to be exceeded
          -> The endpoint MUST treat this as a stream error of type PROTOCOL_ERROR or REFUSED_STREAM.
             Expected: GOAWAY Frame (Error Code: PROTOCOL_ERROR)
                       RST_STREAM Frame (Error Code: PROTOCOL_ERROR)
                       GOAWAY Frame (Error Code: REFUSED_STREAM)
                       RST_STREAM Frame (Error Code: REFUSED_STREAM)
                       Connection closed
               Actual: HEADERS Frame (length:52, flags:0x04, stream_id:1)
```

PR #9520, opened 2024-11-21
Http/2 revamp

### LINKED ISSUE #9272: h2spec http2 8.1.2.6 fails

## Environment Details
* Helidon Version: 4.1.1

----------

## Problem Description
```
  8. HTTP Message Exchanges
    8.1. HTTP Request/Response Exchange
      8.1.2. HTTP Header Fields
        8.1.2.6. Malformed Requests and Responses
          using source address 127.0.0.1:59750
          × 1: Sends a HEADERS frame with the "content-length" header field which does not equal the DATA frame payload length
            -> The endpoint MUST treat this as a stream error of type PROTOCOL_ERROR.
               Expected: GOAWAY Frame (Error Code: PROTOCOL_ERROR)
                         RST_STREAM Frame (Error Code: PROTOCOL_ERROR)
                         Connection closed
                 Actual: DATA Frame (length:2, flags:0x01, stream_id:1)
          using source address 127.0.0.1:59752
          × 2: Sends a HEADERS frame with the "content-length" header field which does not equal the sum of the multiple DATA frames payload length
            -> The endpoint MUST treat this as a stream error of type PROTOCOL_ERROR.
               Expected: GOAWAY Frame (Error Code: PROTOCOL_ERROR)
                         RST_STREAM Frame (Error Code: PROTOCOL_ERROR)
                         Connection closed
                 Actual: DATA Frame (length:2, flags:0x01, stream_id:1)
```

danielkec:
Fixes #9272, #9273 and #9282

- Add h2spec test
- Consuming request trailers
- Larger frame splitting fix
- Flow control update timeout
- Streamed payload larger than content length discovery

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,7 +30,6 @@ import io.helidon.common.socket.SocketContext;
 public class Http2ConnectionWriter implements Http2StreamWriter {
     private final DataWriter writer;
 
-    // todo replace with prioritized lock (stream priority + connection writes have highest prio)
     private final Lock streamLock = new ReentrantLock(true);
     private final SocketContext ctx;
     private final Http2FrameListener listener;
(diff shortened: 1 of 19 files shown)


http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
```diff
@@ -143,24 +142,14 @@ public int writeHeaders(Http2Headers headers,
...
-
-            writeData(dataFrame, flowControl);
-            bytesWritten += Http2FrameHeader.LENGTH;
-            bytesWritten += dataFrame.header().length();
-
-            return bytesWritten;
-        } finally {
-            streamLock.unlock();
-        }
+        // Executed on stream thread
```
spericas, line 145: Could you explain why the mutual exclusion is no longer needed here?

http/http2/src/main/java/io/helidon/http/http2/WindowSizeImpl.java
```diff
@@ -146,21 +143,51 @@ public long incrementWindowSize(int increment) {
...
             while (getRemainingWindowSize() < 1) {
                 try {
-                    updated.get().get(timeoutMillis, TimeUnit.MILLISECONDS);
-                } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                    if (LOGGER_OUTBOUND.isLoggable(DEBUG)) {
-                        LOGGER_OUTBOUND.log(DEBUG,
-                                            String.format("%s OFC STR %d: Window depleted, waiting for update.", type, streamId));
-                    }
+                    updatedSemaphore.drainPermits();
+                    var ignored = updatedSemaphore.tryAcquire(Math.min(backoff, 5000), TimeUnit.MILLISECONDS);
```
spericas, line 171: Maybe create some backoff constants for the upper and lower bounds?

http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
danielkec: Because method is executed only from a single stream thread only, lock was actually causing a deadlock when stream was expecting flow-control window update(wating for update blocks stream thread). Headers needs to be followed by data only from a single stream point of view, other frames from other streams needs and can be writable between those two.

spericas (approved):

tomas-langer (approved):

PR #9528, opened 2024-11-21
Changes dependencies on various Oracle JDBC artifacts from their compiled-with-Java-11 variants to their compiled-with-Java-17 variants

ljnelson:
This PR changes certain Oracle JDBC dependencies from their compiled-with-Java-11 variants to their compiled-with-Java-17 variants. For example, dependencies on `ucp11` are changed to be dependencies on `ucp17`.

The current version of the `ojdbc-bom` does not, because of a mistake, manage the `ojdbc17-production:pom` artifact, so this PR currently includes a stanza in `dependencies/pom.xml` to do this. This stanza should be removed if a new version of the `ojdbc-bom` (and its managed artifacts) is released, and then this PR should be modified to upgrade to that version.


barchetta:
This will change when we upgrade to 23. Closing.

PR #9534, opened 2024-11-26
Add MP metrics annotation support to MP REST clients

### LINKED ISSUE #6057: Helidon does not support MP metrics annotations on MP REST clients
## Environment Details
* Helidon Version: 3.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
Helidon does not support MicroProfile metrics annotations on MP REST clients.

The MP specs do not require this, but it would be a nice enhancement. 

This is different from #5953 which requests automatic metrics for all outgoing MP REST client calls.

tjquinno:
### Description
Resolves #6057 

This PR allows users to use the MicroProfile metrics `@Counted` and `@Timed` annotations on MicroProfile REST client interfaces and interface methods by adding a dependency in their project on a new Helidon component. 

Note that Helidon registers the metrics for a given REST client only once Helidon becomes aware of it which happens either when CDI produces that REST client or when an instance is created explicitly using the REST client builder approach. The metric registration can therefore occur after the app has started depending on how the app code is written to refer to REST clients.

## Details

The PR adds a new Helidon component `helidon-rest-client-metrics`.

The CDI extension does these basic tasks:
* Prepares the metrics registrations derived from the developer-provided annotations on REST client interfaces.
* Registers the metrics for a given interface when that interface becomes known to Helidon as a REST client.
* Encapsulates the metrics-related work required for each such metric before and after each outbound request through a REST client method.

The annotation scanning is pretty straightforward. It saves information about the metrics that will be needed for each REST client interface.

At runtime, when user code or CDI production creates a new REST client from an interface, the MP REST Client implementation calls back to a listener added in this PR. The listener:
* registers the metrics associated with that interface in the MP `MetricRegistry`, 
* prepares a data structure containing "before" invocation and "after" invocation work which updates the metrics, and
* adds a filter (also in this PR) to the REST client builder so the filter will be invoked whenever methods on that REST client are invoked.

When the filter is invoked, it basically delegates back to the CDI extension to perform the pre- and post-invocation metrics updating work.

### Documentation
The PR includes extensive doc updates. 

The existing REST client doc page is moved into a new `restclient` directory which is also where the new REST client metrics page now lives. Several pages with cross-references to the existing REST client page needed to be updated as well. Snippets moved as well.

The new REST client metrics page also has an extensive example discussion. There is no existing REST client example app and I have not at this point created one for REST client metrics. That would be a separate PR for the examples repo anyway.


spericas (approved):
LGTM

PR #9547, opened 2024-11-30
Initialize distribution customizations even before config is applied

### LINKED ISSUE #9546: 4.x Provide initial temp config for MP metrics `DistributionCustomizations`
## Environment Details
* Helidon Version: 4.1.4
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
MP Metrics 5.1 added the ability for users configure percentile and bucket settings in `microprofile-config.properties`. 

Helidon MP initializes these settings as part of the metrics CDI extension late-stage start-up work. As usual, in the absence of config settings for these values Helidon applies defaults.

A user reports that a test that uses MP metrics but which does not use `@HelidonTest` and therefore does not initialize the Helidon MP container began to fail starting with Helidon 4.1.4. That's because, without the metrics CDI extension running, no code initializes the new config settings. As a result, attempts by metrics to dereference the settings threw an NPE.

An easy workaround is to annotated such tests with `@HelidonTest`. It would also be simple enough for Helidon to provide initial default settings, subject to change during actual startup, to avoid such NPEs.

## Steps to reproduce
Create a test without `@HelidonTest`, run separately from tests with that annotation:
```java
class TestDistributionCustomizationsNoInit {

    private static MetricRegistry metricRegistry;

    @BeforeAll
    static void initRegistry() {
        metricRegistry = RegistryFactory.getInstance().getRegistry(MetricRegistry.APPLICATION_SCOPE);
    }

    @Test
    void checkDistributionCustomizations() {
        // The following triggers an NPE because this test does not use @HelidonTest
        // and therefore the normal metrics CDI extension initialization code --which sets up the distribution
        // customizations -- does not run.
        Timer timer = metricRegistry.timer("testTimer");
        assertThat("Timer", timer, notNullValue());
    }
} 
```

tjquinno:
### Description
Resolves #9546 

We added a new data structure, `DistributionCustomizations`, in Helidon 4.1.4 to support an enhancement in MP metrics 5.1 to allow configuration of percentiles, buckets, etc.

Under normal conditions, the Helidon MP metrics CDI extension invokes this type's `init` method with an MP config object to prepare the various customizations and all is well.

A user wrote a test that used MP metrics (a timer specifically) but did not mark the test with `@HelidonTest` so the initialization described above did not occur. That caused an NPE.

We have also since had a report of this happening in an app, in which a CDI extension attempts to register a timer _before_ the metrics CDI extension has prepared metrics properly. As with the test use case above, this means metrics tries to use the `instance` field before it has been set. 

This PR modifies the `DistributionCustomizations` class to initialize the static (non-final) `instance` variable at load-time without using config. This resolves the problem the user encountered with the test. 

Subsequently the metrics CDI extension overwrites that initial value with one derived from configuration.

### Documentation
No impact.

### DIFF
--- a/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/DistributionCustomizations.java
+++ b/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/DistributionCustomizations.java
@@ -62,7 +62,7 @@ class DistributionCustomizations {
     private static final Duration DEFAULT_TIMER_MIN = Duration.ofMillis(5);
     private static final Duration DEFAULT_TIMER_MAX = Duration.ofSeconds(10);
 
-    private static DistributionCustomizations instance;
+    private static DistributionCustomizations instance = new DistributionCustomizations();
     private final List<Percentiles> percentileCustomizations;
     private final List<SummaryBuckets> summaryBucketCustomizations;
     private final List<TimerBuckets> timerBucketCustomizations;
@@ -80,6 +80,13 @@ class DistributionCustomizations {
                                                      "mp.metrics.distribution.percentiles-histogram.enabled",
                                                      expression -> new SingleValuedCustomization<>(expression,
                                                                                                    Boolean::parseBoolean));
+    }
+
+    private DistributionCustomizations() {
+        percentileCustomizations = List.of();
+        summaryBucketCustomizations = List.of();
+        timerBucketCustomizations = List.of();
+        summaryBucketDefaultCustomizations = List.of();
     }
 
     static void init(Config mpConfig) {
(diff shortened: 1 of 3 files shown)


microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/DistributionCustomizations.java
```diff
@@ -62,7 +62,7 @@ class DistributionCustomizations {
     private static final Duration DEFAULT_TIMER_MIN = Duration.ofMillis(5);
     private static final Duration DEFAULT_TIMER_MAX = Duration.ofSeconds(10);
 
-    private static DistributionCustomizations instance;
+    private static DistributionCustomizations instance = new DistributionCustomizations();
```
ljnelson, line 65: (Could make `final`, I think.)

ljnelson (approved):

PR #9552, opened 2024-12-03
Add telemetry filter helper feature so developer code can influence automatic span creation

### LINKED ISSUE #9550: Allow for declarative exclusion of Telemetry spans being generated from HelidonTelemetryClientFilter
## Environment Details
* Helidon Version: 4.1.4
* Helidon MP
* OpenJDK 64-Bit Server VM (build 23+37-2369, mixed mode, sharing)
* OS: Ubuntu 24.04.1 LTS
----------

## Feature Request
We have a number of Microprofile Healthchecks, that of which leverage the REST client to check prerequisites for OCI ObjectStorage access. Because of the `HelidonTelemetryClientFilter` this now generates spans that are unwanted, and result in a considerable amount of noise. We are raising this feature request to ask for a mechanism that allows us to declaratively specify some exclusion(s) criteria that avoids creating these unwanted spans. 

## Workaround
The workaround suggested, in our case where we use OTEL, is to create a custom SpanProcessor that would then pick up information from a (Helidon) context to shortcircuit the span. From the calling code (i.e., a Health check component) the idea would be then to set up that piece of information and make it available through `runInContext`, for the custom SpanProcessor to  use.

tjquinno:
### Description
Resolves #9550 

Helidon provides container and client filters which always add spans to the incoming and outgoing calls (respectively).

As described in the related issue, sometimes developers want to be able to suppress the automatic creation of the spans by the filters.

This PR adds two new helper SPI interfaces, one for container filters and one for client filters. The new interfaces expose a single method which accepts the corresponding request context (container or client) and returns a `boolean` whether that helper votes to create the automatic span or not.

The container and client filters now check with all known helpers to find out whether to create the automatic span or not. As written, if at least one helper votes "no" then the span is _not_ created.

### Documentation
Doc page update will be in a later commit to this PR.

### DIFF
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryClientFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryClientFilter.java
@@ -17,8 +17,12 @@ package io.helidon.microprofile.telemetry;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.ServiceLoader;
 import java.util.Set;
 
+import io.helidon.common.HelidonServiceLoader;
+import io.helidon.common.LazyValue;
+import io.helidon.microprofile.telemetry.spi.HelidonTelemetryClientFilterHelper;
 import io.helidon.tracing.HeaderConsumer;
 import io.helidon.tracing.HeaderProvider;
 import io.helidon.tracing.Scope;
@@ -55,6 +59,11 @@ class HelidonTelemetryClientFilter implements ClientRequestFilter, ClientRespons
             Response.Status.Family.CLIENT_ERROR,
             Response.Status.Family.SERVER_ERROR);
 
+    private static final LazyValue<List<HelidonTelemetryClientFilterHelper>> HELPERS = LazyValue.create(
+            HelidonTelemetryClientFilter::helpers);
+
+    private static final String HELPER_START_SPAN_PROPERTY = HelidonTelemetryClientFilterHelper.class.getName() + ".startSpan";
+
     private final io.helidon.tracing.Tracer helidonTracer;
 
     @Inject
@@ -65,6 +74,16 @@ class HelidonTelemetryClientFilter implements ClientRequestFilter, ClientRespons
 
     @Override
     public void filter(ClientRequestContext clientRequestContext) {
+
+        boolean startSpan = HELPERS.get().stream().allMatch(h -> h.shouldStartSpan(clientRequestContext));
+        clientRequestContext.setProperty(HELPER_START_SPAN_PROPERTY, startSpan);
+        if (!startSpan) {
+            if (LOGGER.isLoggable(System.Logger.Level.TRACE)) {
+                LOGGER.log(System.Logger.Level.TRACE,
+                           "Client filter helper(s) voted to not start a span for " + clientRequestContext.getUri());
+            }
+            return;
+        }
 
         if (LOGGER.isLoggable(System.Logger.Level.TRACE)) {
             LOGGER.log(System.Logger.Level.TRACE, "Starting Span in a Client Request");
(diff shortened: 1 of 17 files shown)


microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryClientFilter.java
```diff
@@ -128,6 +152,10 @@ public void filter(ClientRequestContext clientRequestContext, ClientResponseCont
         clientRequestContext.removeProperty(SPAN_SCOPE);
     }
 
+    private static List<HelidonTelemetryClientFilterHelper> helpers() {
```
ljnelson, line 155: A quick comment without a lot of research behind it: I wonder if you shouldn't use the native Jersey features for this discovery operation. In CDI this would involve injecting an `Instance<...Helper>` at construction time. In "native" Jersey this would probably involve injecting a `Providers` to get hold of a `ContextResolver<...Helper>`. Rationale: Jersey already has a robust service discovery mechanism defined by spec. On the other hand maybe Helidon service loading has been deemed the architectural path forward here; I don't know.

Another related question I have here is that a helper that votes to not start a span is, notionally, just a filter that does nothing, right?

tjquinno: Good questions.

1. I did not think about that approach. 
   
   It would certainly be an alternative and the filters that are already in place are certainly Jakarta REST elements so that would be thematically consistent. 
   
   I guess my own thinking of late in general is to use the Helidon mechanism for ease of moving to a hypothetical future alternative. On the surface (I have not tried redoing the PR as you described) it seems as if there are fewer moving parts as written.
2. I suppose, because the existing filter passes the `ClientRequestContext` or `ContainerReqeustContext` to each helper I suppose a helper _could_ act as a filter by manipulating the context in the same way "real" filters can. That's certainly not the intent. 
   
    I also considered having the filter pass our own implementation of those Jakarta REST `XXXContext` interfaces that reject the mutator methods and delegate to the real context otherwise. But that seemed excessive (to me) to guard developers from themselves.
   
   I did think briefly about--and decided against--having the filter pass a number of parameters to the helper - such as all the items available from the context using the various `getXXX` methods. But that would have been quite cumbersome and there would be no built-in way of allowing access to the properties in the context.
   
   I also thought about having developers provide their own Jakarta REST _filter_ and communicate their desire to suppress the span creation by setting a property in the request context. That seems less transparent and would involve priorities for the developers' filters vs. ours so ours runs last; I didn't pursue this approach.

ljnelson: Right; the general pattern here is, loosely, you want some Filter-Like-Entities to be able to veto other FLEs (you want those helpers to say to downstream filters of a particular kind "please don't act"). I'm just wondering if there's a more idiomatic way to do it rather than introducing new members into the overall ontology. I'm probably just overthinking it.

tjquinno: I am leaning toward using the `Instance` approach and CDI.

From what I've found, the Jakarta `@Provider` approach would only support a single instance of the helper type being made available to the filters; users might have multiple helpers all of which need to participate in each vote.

It would be simpler for developers to annotate their helper implementation classes with a bean-defining annotation (such as `@ApplicationScoped`) (as in the CDI approach) than to create the Java service loader file under `META-INF/services` (as in the Java service loading approach). And as you pointed out earlier, using `Instance` allows the filter to retrieve all the implementations.

ljnelson: The only gotcha here is: for every contextual reference you get via `Instance#get()` (or `next()`) you should arrange to [destroy it if and only if it is in `@Dependent` scope](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#bm_obtain_instance).

tjquinno: (from an off-line conversation about the preceding comment) 

Conclusion: The filters which find the helpers are _not_ created per-request, so caching the helpers via the `Instance` in the filter constructor seems OK here. 

Reasoning: The filters last for the life of the application and so therefore any dependent-scoped helpers a developer might write--and which the filters might bring into existence via the `Instance`--should have the same lifetime as the filter. Because that's the lifetime of the app, the filter does not _really_ have to explicitly destroy any dependent-scoped helpers and there should not be any memory leak.

ljnelson: Right; the only real-world side effect of not arranging for dependent instances to be destroyed (one way or the other) is that any `@PreDestroy` methods they declare will not be executed. This is rarely a problem and I doubt it will crop up in this use case.

barchetta (approved):

PR #9502, opened 2024-11-14
4.x: Update to static content

tomas-langer:
Resolves #9471 

### Description

This PR introduces static content Server feature. This allows us to configure static content in config, with multiple handlers supported (both for `path` and `classpath` based static content).
The new configuration is using Helidon builders, and I have added the Config metadata plugin to have the reference available.
Old approach is deprecated and marked for removal, usages are changed to new approach, including MP.

### Documentation

Documentation updated both for SE and MP

PR #9558, opened 2024-12-04
4.1.5:  Update to static content

### LINKED ISSUE #9471: 4.x: Static content from classpath improvements
Currently we generate temporary files for extracted jar entries as `ws-****.je`

A few improvements could be done:
- use `helidon-` (or `helidon-ws`) prefix to clearly bind them to Helidon
- add configuration option to cache in-memory only (probably to a certain size), to avoid creating the files altogether

barchetta:
Backport #9502 to Helidon 4.1.5

Resolves #9471 

### Description

This PR introduces static content Server feature. This allows us to configure static content in config, with multiple handlers supported (both for `path` and `classpath` based static content).

The new configuration is using Helidon builders, and I have added the Config metadata plugin to have the reference available.

Old approach is deprecated and marked for removal, usages are changed to new approach, including MP.

### Documentation

Documentation updated both for SE and MP


barchetta:
After further review we will not be backporting  this enhancement to a patch release. Closing.

PR #9582, opened 2024-12-10
Makes ApplicationPath a bean defining annotation

spericas:
### Description

Makes ApplicationPath a bean defining annotation. New functional test that shows support for this new feature.

### Documentation

Needs to be updated.

### DIFF
--- a/tests/functional/jax-rs-bda/src/main/java/io/helidon/tests/functional/bda/HelloWorldApp1.java
+++ b/tests/functional/jax-rs-bda/src/main/java/io/helidon/tests/functional/bda/HelloWorldApp1.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2024 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


tests/functional/jax-rs-bda/src/main/java/io/helidon/tests/functional/bda/HelloWorldApp1.java
```diff
@@ -0,0 +1,34 @@
...
+package io.helidon.tests.functional.bda;
+
+import java.util.Set;
+
+import jakarta.ws.rs.ApplicationPath;
+import jakarta.ws.rs.core.Application;
+
+/**
+ * Use {@link ApplicationPath} as a bean-defining annotation (BDA). Note that
+ * there is no CDI scope annotation on the class.
```
ljnelson, line 25: (…which normally means it is therefore in `@Dependent` scope, since you've made it a bean. Is that OK?)

spericas: Yes, in the same way resources are in `@Dependent` scope by default. Jersey should still create an instance "per application".

romain-grecourt (approved):

PR #9619, opened 2024-12-23
Add metrics related to virtual threads

### LINKED ISSUE #9533: Evaluate adding metrics for virtual threads
## Environment Details
* Helidon Version: 4.1.4
* Helidon SE and MP
* JDK version: 21+
* OS: all

----------

This enhancement request is for tracking generation of additional metrics that will provide Helidon users information about virtual thread usage. As an example, current metrics provided in JVM include a count of active Platform Threads. Helidon users would like a metric that provides similar count of active of in-flight virtual threads. Users might also benefit from a runnig tally of current pinned threads.

tjquinno:
### Description
Resolves #9533 

### New meters related to virtual thread usage
* current number of active virtual threads (disabled by default)
* total number of virtual thread starts (disabled by default)
* total number of pinned threads
* timer (with histogram) of pinned threads' durations
* number of failed attempts to submit virtual threads to platform threads

The virtual thread count meters are disabled by default for performance reasons. Enable them by setting `metrics.virtual-threads.count.enabled=true` in configuration, but be aware doing so can degrade the server's performance.

### New meter related to platform threads
Helidon also now exposes the new system meter `thread.starts` which displays the total number of platform thread starts performed in the JVM since server start-up.

----
The PR adds a new `MetersProvider` implementation to the `helidon-metrics-system-meters` component. The new implementation registers for selected Java Flight Recorder events to track data related to virtual threads. 

The three virtual thread meters that are enabled by default should be rare so monitoring the JFR events for them adds very little overhead. In contrast, to maintain the current number of active virtual threads and the total number of virtual thread starts the added code must register for and respond to virtual thread start and end events which can be costly. That's why those two meters--and the registration of listeners for those events from JFR--are disabled by default.

### Documentation
Small additions to the SE and MP metrics guide and doc pages.

### DIFF
--- a/metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/VThreadSystemMetersProvider.java
+++ b/metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/VThreadSystemMetersProvider.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.metrics.systemmeters;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+import io.helidon.common.LazyValue;
+import io.helidon.metrics.api.Gauge;
+import io.helidon.metrics.api.Meter;
+import io.helidon.metrics.api.Metrics;
+import io.helidon.metrics.api.MetricsFactory;
+import io.helidon.metrics.api.Timer;
+import io.helidon.metrics.spi.MetersProvider;
+
+import jdk.jfr.consumer.RecordedEvent;
+import jdk.jfr.consumer.RecordingStream;
+
+/**
+ * Provides system meters for virtual threads using Java Flight Recorder events.
+ * <p>
+ * The virtual thread meters are all net changes since this object was initialized during metrics start-up. This
+ * currently happens before the server listeners start so the virtual thread data should be fairly accurate.
+ * <p>
+ * Further, we track both the number of pinned virtual threads and a distribution summary of their length. That's because
+ * distribution summaries reset after a while to manage storage and reduce the contribution of stale samples.
+ * Keeping a separate count gives some indication of the level of thread pinning over the entire lifetime of the server.
+ * <p>
+ * JFR delivers events in batches. For performance the values we track are stored as longs without
+ * concern for concurrent updates which should not happen anyway.
+ */
+public class VThreadSystemMetersProvider implements MetersProvider {
+
+    // Parts of the meter names.
+    static final String METER_NAME_PREFIX = "vthreads.";
+    static final String COUNT = "count";
+    static final String SUBMIT_FAILURES = "submitFailures";
+    static final String PINNED = "pinned";
+    static final String RECENT_PINNED = "recentPinned";
+    static final String STARTS = "starts";
+
+    private static final String METER_SCOPE = Meter.Scope.BASE;
+
+    private static final System.Logger LOGGER = System.getLogger(VThreadSystemMetersProvider.class.getName());
+    private final LazyValue<Timer> recentPinnedVirtualThreads = LazyValue.create(this::findPinned);
+    private long virtualThreadSubmitFails;
+    private long pinnedVirtualThreads;
+    private long virtualThreads;
+    private long virtualThreadStarts;
+
+    /**
+     * For service loading.
+     */
+    public VThreadSystemMetersProvider() {
+    }
+
+    @Override
+    public Collection<Meter.Builder<?, ?>> meterBuilders(MetricsFactory metricsFactory) {
+
+        var rs = new RecordingStream();
+
+        List<Meter.Builder<?, ?>> meterBuilders = new ArrayList<>(List.of(
+                Gauge.builder(METER_NAME_PREFIX + SUBMIT_FAILURES, () -> virtualThreadSubmitFails)
+                        .description("Virtual thread submit failures")
+                        .scope(METER_SCOPE),
+                Gauge.builder(METER_NAME_PREFIX + PINNED, () -> pinnedVirtualThreads)
+                        .description("Number of pinned virtual threads")
+                        .scope(METER_SCOPE),
+                Timer.builder(METER_NAME_PREFIX + RECENT_PINNED)
+                        .description("Pinned virtual thread durations")
+                        .scope(METER_SCOPE)));
+
+        listenFor(rs, Map.of("jdk.VirtualThreadSubmitFailed", this::recordSubmitFail,
+                             "jdk.VirtualThreadPinned", this::recordThreadPin));
+
+        if (metricsFactory.metricsConfig().virtualThreadCountEnabled()) {
+            meterBuilders.add(Gauge.builder(METER_NAME_PREFIX + COUNT, () -> virtualThreads)
+                                      .description("Active virtual threads")
+                                      .scope(METER_SCOPE));
+            meterBuilders.add(Gauge.builder(METER_NAME_PREFIX + STARTS, () -> virtualThreadStarts)
+                                      .description("Number of virtual thread starts")
+                                      .scope(METER_SCOPE));
+
+            listenFor(rs, Map.of("jdk.VirtualThreadStart", this::recordThreadStart,
+                                 "jdk.VirtualThreadEnd", this::recordThreadEnd));
+        }
+
+        rs.startAsync();
+        return meterBuilders;
+    }
+
+    private static void listenFor(RecordingStream rs, Map<String, Consumer<RecordedEvent>> events) {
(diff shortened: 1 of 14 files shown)


spericas:
Some general questions @tjquinno ...

> * current number of active virtual threads (disabled by default)
> * total number of virtual thread starts (disabled by default)
> * total number of pinned threads

 Is there also a counter for platform threads?

> * distribution summary (history) of pinned threads

 Do you have an example?

> * number of failed attempts to submit virtual threads to platform threads

 Under what conditions would this counter be incremented?

metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/VThreadSystemMetersProvider.java
```diff
@@ -0,0 +1,148 @@
...
+            meterBuilders.add(Gauge.builder(METER_NAME_PREFIX + COUNT, () -> virtualThreads)
+                                      .description("Active virtual threads")
+                                      .scope(METER_SCOPE));
+            meterBuilders.add(Gauge.builder(METER_NAME_PREFIX + STARTS, () -> virtualThreadStarts)
+                                      .description("Number of virtual thread starts")
+                                      .scope(METER_SCOPE));
+
+            listenFor(rs, Map.of("jdk.VirtualThreadStart", this::recordThreadStart,
+                                 "jdk.VirtualThreadEnd", this::recordThreadEnd));
+        }
```
spericas, line 102: Should we also track non-virtual thread start/end? Or is that already available elsewhere?

metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/VThreadSystemMetersProvider.java
```diff
@@ -0,0 +1,148 @@
...
+
+    private DistributionSummary findPinned() {
+        var result = Metrics.globalRegistry().summary(METER_NAME_PREFIX + PINNED, List.of());
+        if (result.isEmpty()) {
+            throw new IllegalStateException(METER_NAME_PREFIX + "pinned meter expected but not registered");
+        }
+        return result.get();
+    }
+
+    private void recordThreadStart(RecordedEvent event) {
```
spericas, line 126: Are these callbacks guarantied to be called sequentially?

tjquinno:
> Some general questions @tjquinno ...
> 
> > * current number of active virtual threads (disabled by default)
> > * total number of virtual thread starts (disabled by default)
> > * total number of pinned threads
> 
> Is there also a counter for platform threads?

The existing Helidon built-in meters created in [`SystemMetersProvider.java](https://github.com/helidon-io/helidon/blob/main/metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/SystemMetersProvider.java) include some platform-thread related ones:
* thread count
* daemon thread count
* maximum thread count

This JMX MBean makes a number of other thread-related values available: 
https://docs.oracle.com/en/java/javase/21/docs/api/java.management/java/lang/management/ThreadMXBean.html

The one available value we don't currently expose is probably the total started thread count. We could add that.

> 
> > * distribution summary (history) of pinned threads
(I edited the description to change "history" to "histogram" which I originally intended!)

> 
> Do you have an example?
> 

I don't have an _interesting_ example of the value. 

Here is an example of the output, taken from the updated doc that's part of this PR (obviously with no interesting output):
```json
  "vthreads.recentPinned": {
      "count": 0,
      "max": 0.0,
      "mean": 0.0,
      "total": 0.0,
      "p0.5": 0.0,
      "p0.75": 0.0,
      "p0.95": 0.0,
      "p0.98": 0.0,
      "p0.99": 0.0,
      "p0.999": 0.0
    },
 ```

> > * number of failed attempts to submit virtual threads to platform threads
> 
> Under what conditions would this counter be incremented?

[Here's](https://docs.oracle.com/en/java/javase/21/core/virtual-threads.html##JSCOR-GUID-F24A6627-4599-4FEF-B17F-DF35EBD34F7B) what the Java Flight Recorded doc says about the value it reports about pinned virtual threads.

> 
> > * distribution summary (history) of pinned threads
> > * number of failed attempts to submit virtual threads to platform threads
> 
> Under what conditions would this counter be incremented?
See the previous link.

File: metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigSupport.java [hunk 3]
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
  *
```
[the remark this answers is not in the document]
tjquinno: This is a remnant from an earlier iteration of the change I was contemplating. Not needed.

metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/VThreadSystemMetersProvider.java
tjquinno: I looked hard for a definitive statement one way or the other about this.

Although I did not find one in the official Javadoc the [JEP for `RecordingStream`s](https://openjdk.org/jeps/349) states this in explaining how JFR invokes the registered callbacks:
> Events stored in thread-local buffers are flushed periodically to the disk repository by the Java Virtual Machine (JVM) once every second. A separate thread parses the most recent file, up to the point in which data has been written, and pushes the events to subscribers. To keep overhead low, only actively subscribed events are read from the file. 

That the JEP describes _a_ (which implies a _single_) thread parsing the file suggests that JFR will not use multiple concurrent threads to notify subscribers. This (to me) strongly indicates that JFR's invocation of the callbacks is thread-safe.

Further, the code in this PR creates a new `RecordingStream` which implements `EventStream`, and the [Javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/jdk.jfr/jdk/jfr/consumer/EventStream.html) for `EventStream` states, among other things:
   > Events are delivered in batches. To receive a notification when a batch is complete, register an action using the [onFlush(Runnable)](https://docs.oracle.com/en/java/javase/21/docs/api/jdk.jfr/jdk/jfr/consumer/EventStream.html#onFlush(java.lang.Runnable)) method. This is an opportunity to aggregate or push data to external systems while the Java Virtual Machine (JVM) is preparing the next batch.
 
Plus, JFR goes to some lengths to impose as light a load as possible on the running JVM. It stands to reason that JFR would not invoke callbacks in a way that would require potentially-costly locking within the callbacks to prevent concurrent updates to data that might occur within the callbacks.

metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/VThreadSystemMetersProvider.java
tjquinno: As noted in a response to another comment, we already report as a meter the current platform thread count. We do not report the total number of platform thread starts. That value is available from the JMX `ThreadMXBean` so we could add that.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigSupport.java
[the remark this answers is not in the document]
tjquinno: Fixed. File no longer in the changed list.

tjquinno:
The latest push accomplished:
* Removed an unchanged file from the changed file list.
* Added the new _non-virtual_ thread system meter `thread.starts` to the built-in system meters
* Changed the meter type for `vthreads.recentPinned` from distribution summary to timer (the two are basically equivalent except that the timer type displays an elapsed time instead of the more general data value).

spericas (approved):

vasanth-bhat:
Few comments,  

1.  The implementation is  creating the RecordingStream without passing  any settings.  Is this done intentionally? Both Recording()  and RecordingStream()  have alternate constructors that take settings map as input. 

 In JFR stream or JFR, the details as to which events are captured,  threshold used for each event , stack capture etc depends on  settings used to create the  Recording (RecordingStream internally creates Recording) .    JDK  ships with 2 settings "default" and "profile".       

Below are settings for ”jdk.VirtualThreadPinned”  event  in  JDK bundled settings "default" and "profile" .  This would means when these settings are used,  a Pinned  JFR event would be recorded only for cases where the carrier thread was pinned for 20ms or longer.

 <event name="jdk.VirtualThreadPinned">
      <setting name="enabled">true</setting>
      <setting name="stackTrace">true</setting>
      <setting name="threshold">20 ms</setting>
 </event>

However many cases the system also run with custom settings via custom JFC file, and can can customise  settings for  both built-in JDK provided as well as custom defined JFR events.

2   ”jdk.VirtualThreadPinned”  JFR events are not generated  for all pinning cases.  For example carrier thread also get pinned  in Java-21 when the vthread mounted on them execute Object.wait().  But these do not generate ”jdk.VirtualThreadPinned”.   Same is true for pinning due to blocking operation in class initialiser  ( for ex : static blocks), or pining due to certain blocking operation in native code.   So primarily, we get pinning events only for blocking operation from sync blocks.    I guess this  is more for information/documentation purposes, so that one doesn't assume all pinning events are reported by this metric.

tjquinno:
We looked into this earlier and to get initial support for meters related to virtual threads in place quickly with the simplest ease-of-use we implemented the feature as written. 

Some examples of the complexities:

* The `settings` value in the `-XX:StartFlightRecording:settings` command-line qualifier can take multiple values. That means a Helidon user would need a way to choose which of the settings (among the custom ones specified on the command line or the predefined ones) to use as the basis for the Helidon `RecordingStream`. 
* The JFR-provided `Configuration` class allows look up by name any of the _predefined_ profiles...those whose `.jfc. files are in `JAVA_HOME`/lib/jfr`. But `Configuration.getConfiguration(String)` and `Configuration.getConfigurations()` do not support  config files located elsewhere. The Helidon user would need to provide Helidon configuration to specify which built-in name or custom file to use. The Helidon code would need to use that setting to select either a predefined configuration by name or a custom configuration by path. 

All doable, but for the first implementation we chose to simplify things.

I've had this sort of enhancement in mind but finally recorded it as an issue: https://github.com/helidon-io/helidon/issues/9652 (generally more likely to get visibility than comments on a previously-merged PR). Add any further comments you might have there.

PR #9640, opened 2025-01-08
4.x: UPN claim should be optional (#5151)

### LINKED ISSUE #5151: unwanted "upd" claim in JWT - Oracle IDCS/IAM Domains having problems

Captain1653:
### Description
Fixes #5151 

Original PR is [here](https://github.com/helidon-io/helidon/pull/9307).

Code is the same, but I have already signed OCA (my little help :) ).

Maybe, some changes in the docs should be done, but I'm not sure. I've read [this doc](https://helidon.io/docs/v4/mp/jwt), but I didn't find any mentions.

### DIFF
--- a/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
+++ b/security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -307,9 +307,7 @@ public class Jwt {
         this.cHash = JwtUtil.getByteArray(payloadJson, C_HASH, "c_hash value");
         this.nonce = JwtUtil.getString(payloadJson, NONCE);
         this.scopes = JwtUtil.toScopes(payloadJson);
-        this.userPrincipal = JwtUtil.getString(payloadJson, USER_PRINCIPAL)
-                .or(() -> preferredUsername)
-                .or(() -> subject);
+        this.userPrincipal = JwtUtil.getString(payloadJson, USER_PRINCIPAL);
     }
 
     private Jwt(Builder builder) {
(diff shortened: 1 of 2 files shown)


security/jwt/src/main/java/io/helidon/security/jwt/Jwt.java
```diff
@@ -663,11 +659,14 @@ public Optional<String> subject() {
...
     /**
      * User principal claim ("upn" from microprofile specification).
+     * Falls back "preferred_username" then "sub" claim.
      *
-     * @return user principal or empty if claim is not defined
+     * @return user principal or empty if claim and fallbacks are not defined
      */
     public Optional<String> userPrincipal() {
-        return userPrincipal;
+        return userPrincipal
```
Verdent, line 667: I think this causes inconsistency between the runtime model and the JWT which it generates. From my point of view this method should return only the set userPrincipal value and on places where this is used (and functionality based on the MP spec needs to be present), there might be those falbacks added. This causes backwards incompatible functionality, but on the other hand, I see this rather as a bug fix, so it should not be a problem.

Verdent: Any other opinions about this from others?

manger: Having the upn-or-preferred_usernane-or-sub logic in Jwt.Builder.build() is a bug.

Moving the login to Jwt.userPrincipal() fixes the bug and keeps backwards compatibility.

Moving it to a higher layer would be good for API consistency, which is probably worth the minor backwards incompatibility. But that should not stop fixing the immediate bug. Not being able to create JWTs with "sub" but not "upn" is the flaw needing fixing.

PR #9654, opened 2025-01-13
Metrics support for FT SE

spericas:
### Description

Basic support for metrics in FT SE. The following metrics are supported:

- ft.bulkhead.calls.total
- ft.bulkhead.waitingDuration
- ft.bulkhead.executionsRunning
- ft.bulkhead.executionsWaiting
- ft.circuitbreaker.calls.total 
- ft.circuitbreaker.opened.total
- ft.retry.calls.total
- ft.retry.retries.total
- ft.timeout.calls.total
- ft.timeout.executionDuration 

Metrics are disabled by default but can be enabled using `ft.metrics.enabled=true` or individually using `enableMetrics(boolean)` on the corresponding metric builder. For more information about these new metrics see `fault-tolerance.adoc` updates included in this PR.

### Documentation

Included as part of this PR.

### DIFF
--- a/docs/src/main/asciidoc/se/fault-tolerance.adoc
+++ b/docs/src/main/asciidoc/se/fault-tolerance.adoc
@@ -262,19 +262,13 @@ in Helidon when a method is decorated with multiple annotations.
 == Metrics
 
 The Helidon Fault Tolerance module has support for some basic metrics to monitor
-certain conditions. Metrics are disabled by default, but can be enabled via config by
-setting the property `ft.metrics.enabled=true` and by including an actual metrics
-implementation in your classpath.
+certain application conditions. Metrics are disabled by default, but can be enabled
+via config by setting the property `ft.metrics.enabled=true` and by including an actual
+metrics implementation in your classpath. For more information about metrics implementations
+see xref:{rootdir}/se/metrics/metrics.adoc[Helidon Metrics].
 
-```
-<dependency>
-    <groupId>io.helidon.metrics</groupId>
-    <artifactId>helidon-metrics</artifactId>
-</dependency>
-```
-
-The following tables lists all the metrics created by the Fault Tolerance module.
-Note that these metrics are generated per instance, and that each instance _must_
+The following tables list all the metrics created by the Fault Tolerance module.
+Note that these metrics are generated per command instance, and that each instance _must_
 be identified by a unique name --assigned either programmatically by
 the application developer or automatically by the API.
 
--- a/docs/src/main/asciidoc/se/fault-tolerance.adoc
+++ b/docs/src/main/asciidoc/se/fault-tolerance.adoc
@@ -277,7 +277,7 @@ the application developer or automatically by the API.
 |===
 ^|Name ^|Tags ^|Description
 |ft.bulkhead.calls.total | name="<bulkhead-name>" | Counter for all calls entering a bulkhead
-|ft.bulkhead.waitingDuration | name="<bulkhead-name>" | Histogram of waiting times to enter a bulkhead
+|ft.bulkhead.waitingDuration | name="<bulkhead-name>" | Distribution summary of waiting times to enter a bulkhead
 |ft.bulkhead.executionsRunning | name="<bulkhead-name>" | Gauge whose value is the number of executions running in a bulkhead
 |ft.bulkhead.executionsWaiting | name="<bulkhead-name>" | Gauge whose value is the number of executions waiting in a bulkhead
 |===
@@ -304,14 +304,14 @@ closed to open state
 |===
 ^|Name ^|Tags ^|Description
 |ft.timeout.calls.total | name="<timeout-name>" | Counter for all calls entering a timeout
-|ft.timeout.executionDuration | name="<timeout-name>" | Histogram of all execution durations in a timeout
+|ft.timeout.executionDuration | name="<timeout-name>" | Distribution summary of all execution durations in a timeout
 |===
 
 === Enabling Metrics Programmatically
 
 Metrics can be enabled programmatically either globally or, if disabled globally, individually for
 each command instance. To enable metrics globally, call `FaultTolerance.config(Config)` passing
-a Config instance that sets `ft.metrics.enabled=true`. This must be done on application startup,
+a Config instance that sets `ft.metrics.default-enabled=true`. This must be done on application startup,
 before any command instances are created.
 
 If metrics are not enabled globally, they can be enabled programmatically on each command instance
@@ -326,7 +326,6 @@ include::{sourcedir}/se/FaultToleranceSnippets.java[tag=snippet_8, indent=0]
 
 NOTE: The global config setting always takes precedence: that is, if metrics are enabled
 globally, they *cannot* be disabled individually by calling `enableMetrics(false)`.
-
 
 == Examples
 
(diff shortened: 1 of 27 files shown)


docs/src/main/asciidoc/se/fault-tolerance.adoc
```diff
@@ -259,6 +259,81 @@ and bulkhead is the last one (the first to be executed once a value is returned)
...
+
+Metrics can be enabled programmatically either globally or, if disabled globally, individually for
+each command instance. To enable metrics globally, call `FaultTolerance.config(Config)` passing
+a Config instance that sets `ft.metrics.enabled=true`. This must be done on application startup,
+before any command instances are created.
+
+If metrics are not enabled globally, they can be enabled programmatically on each command instance
+using the `enableMetrics(boolean)` method on its corresponding builder. For example, the
+following snippet shows how to create a `Retry` instance of name `my-retry` with metrics
+support enabled.
```
tjquinno, line 320: For metrics as a whole, setting `metrics.enabled` to `false` disables all metrics in which case any individual setting to enable an individual metric has no effect. If `metrics.enabled` is `true` (the default) then individual metrics can be disabled.

The FT metrics global setting as described has different semantics and could create some confusion for users.

Either approach would make sense, but I don't think we want two different meanings--one for metrics as a whole and one for FT metrics--for the umbrella "enabled" setting. 

And since the overall metrics interpretation has been in place for a while can we change the way the FT global setting works to align with that?

If it's useful to retain these semantics for FT metrics then we could change the key for the setting to something like `ft.metrics.default-enabled` or something like that.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/Bulkhead.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2025 Oracle and/or its affiliates.
```
tjquinno, line 2: Why remove 2020? (same for some other copyright headers in other files)

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/Bulkhead.java
```diff
@@ -32,6 +32,32 @@
  */
 @RuntimeType.PrototypedBy(BulkheadConfig.class)
 public interface Bulkhead extends FtHandler, RuntimeType.Api<BulkheadConfig> {
+
+    /**
+     * Counter for all the calls in a bulkhead.
+     */
+    String FT_BULKHEAD_CALLS_TOTAL = "ft.bulkhead.calls.total";
```
tjquinno, line 39: If we remove the `.total` suffix from the value then probably remove it from the name of the constant as well. (I won't repeat the same note for the other counters but it would apply to them too.)

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
```diff
@@ -0,0 +1,94 @@
...
+    static boolean enableMetrics() {
+        if (enableMetrics == null) {
+            Config config = FaultTolerance.config();
+            enableMetrics = config.get(FT_METRICS_ENABLED).asBoolean().orElse(false);
+        }
+        return enableMetrics;
+    }
+
+    static <T extends Number> void gaugeBuilder(String name, Supplier<T> supplier, Tag... tags) {
+        Gauge.Builder<T> builder = METRICS_FACTORY.get().gaugeBuilder(name, supplier).scope(VENDOR);
```
tjquinno, line 60: Not essential to change this, but for future reference you can avoid the explicit reference to the metrics factory by using, for example, `Gauge.builder(...)` or `Counter.builder(...)`, etc.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/TimeoutImpl.java
```diff
@@ -52,11 +67,20 @@ public String name() {
...
+        long start = metricsEnabled ? System.nanoTime() : 0L;
         if (!currentThread) {
             try {
-                return CompletableFuture.supplyAsync(supplier, executor)
+                T res = CompletableFuture.supplyAsync(supplier, executor)
                         .orTimeout(timeoutMillis, TimeUnit.MILLISECONDS)
                         .get();
+                if (metricsEnabled) {
+                    executionDurationMetric.record(System.nanoTime() - start, TimeUnit.NANOSECONDS);
+                }
```
tjquinno, line 82: IIUC this updates the timer only if the work completes successfully? Is that what we want?

tjquinno (changes requested):

docs/src/main/asciidoc/se/fault-tolerance.adoc
```diff
@@ -259,6 +259,81 @@ and bulkhead is the last one (the first to be executed once a value is returned)
...
+The Helidon Fault Tolerance module has support for some basic metrics to monitor
+certain conditions. Metrics are disabled by default, but can be enabled via config by
+setting the property `ft.metrics.enabled=true` and by including an actual metrics
+implementation in your classpath.
+
+```
+<dependency>
+    <groupId>io.helidon.metrics</groupId>
+    <artifactId>helidon-metrics</artifactId>
+</dependency>
```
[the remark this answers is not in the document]
spericas: That was updated with a link to the Metrics doc

docs/src/main/asciidoc/se/fault-tolerance.adoc
```diff
@@ -259,6 +259,81 @@ and bulkhead is the last one (the first to be executed once a value is returned)
...
+The following tables lists all the metrics created by the Fault Tolerance module.
+Note that these metrics are generated per instance, and that each instance _must_
+be identified by a unique name --assigned either programmatically by
+the application developer or automatically by the API.
+
+[cols="1,2,3"]
+.Bulkheads
+|===
+^|Name ^|Tags ^|Description
+|ft.bulkhead.calls.total | name="<bulkhead-name>" | Counter for all calls entering a bulkhead
```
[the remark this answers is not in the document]
spericas: These are named after (or identical to) the ones in MP FT

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/Bulkhead.java
spericas: Oops, that was a mistake

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/Bulkhead.java
spericas: See above

docs/src/main/asciidoc/se/fault-tolerance.adoc
```diff
@@ -259,6 +259,81 @@ and bulkhead is the last one (the first to be executed once a value is returned)
...
+|ft.bulkhead.waitingDuration | name="<bulkhead-name>" | Histogram of waiting times to enter a bulkhead
```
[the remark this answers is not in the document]
spericas: Roger

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/TimeoutImpl.java
spericas: Good point, for consistency it should be on finally I think

tjquinno (approved):

PR #9655, opened 2025-01-16
Adds an afterStart pseudo-event to LoomServer observable by Router implementations

ljnelson:
This PR adds an `afterStart` "event" in the same spirit as the existing `beforeStart` and `afterStop` "events" that are found in `ServerLifecycle` implementations (but dispatchable only via `Router` implementations).

I am not sure that I have written it correctly and expect it to be discussed thoroughly. My goal with this first pass has been to simply preserve the spirit and style of the preexisting "event" notifications in `LoomServer`, together with preserving design decisions made internally by members of the Helidon team.

`LoomServer` (package private) has a collection of (package private) `ServerListener` instances, indexed by name, that have access to `Router` implementations internally. I found that to preserve the spirit of the existing code I had to provide access to these `Router`s, since they are currently the only recipients of `beforeStart` and `afterStop` "events", and so, it seemed to me, should be the recipients of (the new) `afterStart` "event".

Internal discussions said that the notification mechanism should be serial, i.e. not in parallel, so I did not use the `parallel` method to dispatch the `afterStart` "event".

Currently if an `afterStart` "observer" throws an exception undefined behavior occurs. I believe that there will be many good opinions on what should happen instead.

Internal discussions also said that the new "event" payload should be the `WebServer` itself so there is now a bidirectional dependency between, e.g., `Router` and `WebServer`. I'm not sure if this was intended to be the case.

I also noticed that `Router` is effectively implementing the `ServerLifecycle` contract, but does not officially implement it, so I didn't add that.

I expect that there will be many changes requested, possibly even a completely different architecture for this originally-intended-to-be-small change.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
@@ -53,7 +53,6 @@ class LoomServer implements WebServer {
     private static final AtomicInteger WEBSERVER_COUNTER = new AtomicInteger(1);
 
     private final Map<String, ServerListener> listeners;
-    private final Runnable afterStartNotifier;
     private final AtomicBoolean running = new AtomicBoolean();
     private final Lock lifecycleLock = new ReentrantLock();
     private final ExecutorService executorService;
@@ -102,7 +101,6 @@ class LoomServer implements WebServer {
         });
 
         listeners = Map.copyOf(listenerMap);
-        afterStartNotifier = () -> listeners.values().forEach(l -> l.router().afterStart(this));
     }
 
     @Override
@@ -227,7 +225,7 @@ class LoomServer implements WebServer {
                 + uptime + " milliseconds since JVM startup. "
                 + "Java " + Runtime.version());
 
-        afterStartNotifier.run();
+        fireAfterStart();
 
         if ("!".equals(System.getProperty(EXIT_ON_STARTED_KEY))) {
             LOGGER.log(System.Logger.Level.INFO, String.format("Exiting, -D%s set.", EXIT_ON_STARTED_KEY));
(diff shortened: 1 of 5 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
```diff
@@ -53,6 +53,7 @@ class LoomServer implements WebServer {
     private static final AtomicInteger WEBSERVER_COUNTER = new AtomicInteger(1);
 
     private final Map<String, ServerListener> listeners;
+    private final Runnable afterStartNotifier;
```
romain-grecourt, line 56: Can this be a private method instead of a `Runnable` ?

ljnelson: OK

romain-grecourt (approved):

tjquinno:
I share what I think is your concern that, currently, routing has to be involved for these events to be delivered. 

Here's a concrete example that argues for supporting a non-routing-based way to learn of these events. 

The code which manages metrics for virtual threads starts a Java Flight Recorder `RecordingStream` to receive virtual-thread related events from JFR. That code, which does not and should not rely on any routing information, has no way to learn when the server is about to/has shut down, so it cannot explicitly close the `RecordingStream` when the server is shutting down.

(Recall that meters can be retrieved programmatically from code in the server and not only by the metrics endpoint, so it's important to maintain these measurements apart from if they are ever retrieved via HTTP.)

Practically speaking that's fine because we need the `RecordingStream` as long as the server is up, and JVM shutdown is the ultimate clean-up. 

But some recent work by @danielkec is somewhat disrupted by this - see #9687. 

Arguably `ServerLifecycle` should be outside the webserver component and its implementations notified without routing...it's probably rare, but there _could_ be Helidon servers that do not include a webserver but there are certainly components (like metrics) that should be able to respond to server lifecycle events without relying on routing.

tomas-langer (approved):

ljnelson:
Given the lack of discussion and two approvals I will merge this as is, with some mild reservations.

Issue #9650, opened 2025-01-10
4.x: remove grpc-netty from config/etcd


barchetta:
The module `config/etcd` has a dependency on `grpc-netty`:

https://github.com/helidon-io/helidon/blob/95af89efb622b90fe578cd2443c835ef2aa7afad/config/etcd/pom.xml#L65

With the new implementation of grpc in Helidon 4 it should be possible to remove this dependency.

Related to #9645

PR #9674, opened 2025-01-22
Use Helidon gRPC client in Config Etcd module

spericas:
### Description

Use Helidon gRPC client in config etcd module. Updates old IT tests to run with etcd V3 only. IT tests manually executed and shown to pass with all the latest changes and V3 etcd server. Issue #9650.

### Documentation. 

None


config/etcd/src/main/java/io/helidon/config/etcd/internal/client/v2/EtcdV2Client.java
```diff
@@ -55,7 +55,7 @@ public class EtcdV2Client implements EtcdClient {
      *
      * @deprecated
      */
-    EtcdV2Client(URI... uris) {
+    public EtcdV2Client(URI... uris) {
         etcd = new mousio.etcd4j.EtcdClient(uris);
```
spericas, line 59: Needs to be public for IT tests

barchetta (approved):

Issue #9707, opened 2025-01-28
Docs: Performance guide: remove deprecated config, add section on tuning file upload for Linux


barchetta:
## Environment Details
* Helidon Version: 4.1.6
* Helidon SE and MP
----------

## Problem Description

Update the  performance guides (https://helidon.io/docs/4.1.6/mp/guides/performance-tuning) to:

1. Remove deprecated/unsupported config options (such as `server.receive-buffer-size`). See issue #9706.
2. Add a paragraph concerning tuning file upload performance on Linux. See issue #9678

Issue #9706, opened 2025-01-28
Deprecate socket options no longer used in config


spericas:
## Problem Description
Certain socket options such as receive-buffer-size appear to be available in multiple config places, including `ListenerConfig` (and thus `WebServerConfig`), `ConnectionConfig` and `SocketOptions`. Only the last one appears to be in use at this time, so all the others should be deprecated and marked for future removal. There are likely other config properties that are also duplicated.

## Documentation
We need to review any docs that describe the use of these socket options

PR #9709, opened 2025-01-28
Deprecation of old config options now replaced by SocketOptions

spericas:
### Description

Deprecation of old config options now replaced by SocketOptions. We shall remove these in 5.0.

### Documentation

Updates included in this PR

### DIFF
--- a/docs/src/main/asciidoc/includes/guides/performance-tuning.adoc
+++ b/docs/src/main/asciidoc/includes/guides/performance-tuning.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2020, 2023 Oracle and/or its affiliates.
+    Copyright (c) 2020, 2025 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


barchetta:
See #9706 #9707

docs/src/main/asciidoc/includes/guides/performance-tuning.adoc
```diff
@@ -28,7 +28,6 @@ Still, there might be cases where you wish to change configuration options from
 For details on the following options please see:
 
 * xref:../../se/webserver.adoc#_configuration_options[WebServer Configuration]
-* xref:../../config/io_helidon_webserver_ConnectionConfig.adoc[WebServer Connection Configuration]
 * xref:../../config/io_helidon_common_socket_SocketOptions.adoc[WebServer Socket Configuration]
```
barchetta, line 32: How about adding a paragraph concerning tuning for file upload performance on Linux (setting `socket-receive-buffer-size`)

spericas: I've asked PSR to help us run a few tests on other types of environments (OCI). I think it would be nice to have that info before writing such a paragraph. Should we do that in a follow-up PR?

barchetta: Sounds good to me.

barchetta (approved):

PR #9715, opened 2025-01-29
Gracefully reject HTTP 1.0 connections 

spericas:
### Description

- Gracefully reject HTTP 1.0 connections with a 505 error code (and log entry) that are routed directly to our 1.1 provider
- If multiple providers are available, we cannot easily return 505 but we now detect HTTP 1.0 and log an informative message

### Documentation

None


barchetta (approved):
LGTM

PR #9716, opened 2025-01-30
Add support for span lifecycle callbacks for selected OTel objects

### LINKED ISSUE #9079: `SpanListener`s  do not get invoked when  native OTel  tracer is used via Inject or unwrap

tjquinno:
### Description
Resolves #9079 

### Release note
____
For several releases, Helidon's neutral tracing API (used primarily in Helidon SE applications) has invoked callbacks in your own `SpanListener` when a Helidon neutral `Span` or `Scope` changes state. See https://helidon.io/docs/latest/se/tracing#Tracing-callbacks. 

Separately, Helidon's support for MicroProfile Telemetry already lets your code inject an OpenTelemetry `Tracer` and the current OpenTelemetry `Span`. 

Starting with this release of Helidon, `SpanListener`s you register now receive lifecycle events for _selected_ OpenTelemetry objects:
* `Span` objects: 
   * the injected current `Span`
   * a `Span` created by a `SpanBuilder` that is obtained from the injected `Tracer`
* `Scope` objects returned from invoking `makeCurrent()` on the specific `Span` objects described above.

See the updated documentation under MP -> Telemetry for more information.
____

### PR Overview
We know of user applications which obtain an OpenTelemetry `Span` and type-check and cast it to `ReadableSpan` (another OpenTelemetry interface). To add the callback functionality while supporting that type-checking and casting usage on injected objects--inadvisable though such checking and casting of _injected_ instances might be--the changes in the PR wrap the injected OTel instances in Java proxies. This preserves type-checking and casting compatibility while adding the callback behavior.

The CDI producer methods now return proxies rather than the native OTel objects.

Each proxy invocation handler holds references to the OTel delegate and a corresponding Helidon wrapper object.

Basically, the proxies do additional work for only a few methods:
* methods which cause a state change in a span or scope
* methods which return another OTel type that exposes methods that cause a state change

For other methods the proxies just invoke the method on the delegate.

For operations which induce state changes, the proxy code invokes the corresponding method on the _Helidon wrapper_. Those methods do the correct notification (as they always have) and invoke the corresponding method on the OTel delegate (as they always have). Using the Helidon wrappers this way avoids replicating the wrappers' notification logic in the proxies.

There is one change to the public API to allow all this to work. The existing Helidon `Span` interface now has a new sub-interface `Span.Extended` which adds one additional method. Only our OpenTelemetry-based implementation of `Span` really needs this new method (so proxy invocation handler code can invoke it), but our OTel-based `Span` implementation `OpenTelemetrySpan` is in the `tracing/providers/opentelemetry` module and is package-private, so the `microprofile/telemetry` module cannot know about it. Adding the new public sub-interface and having `OpenTelemetrySpan` implement it minimizes the public changes while letting the proxy invoke the new method.

### Documentation
Update included in the PR.

### DIFF
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
@@ -1,0 +1,263 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.microprofile.telemetry;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import io.helidon.common.Errors;
+import io.helidon.tracing.providers.opentelemetry.HelidonOpenTelemetry;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.Scope;
+
+/**
+ * Provides proxies for OpenTelemetry types.
+ * <p>
+ * The proxies allow users to inject {@link io.opentelemetry.api.trace.Tracer} and {@link io.opentelemetry.api.trace.Span} and
+ * invoke state-changing operations on them and on related types ({@link io.opentelemetry.context.Scope}) in a way that notifies
+ * developer-provided {@link io.helidon.tracing.SpanListener} objects that have been registered with the Helidon tracer.
+ */
+class OtelProxy {
+
+    static OpenTelemetry openTelemetry(OpenTelemetry openTelemetry) {
+        return (OpenTelemetry) Proxy.newProxyInstance(OtelProxy.class.getClassLoader(),
+                                                      openTelemetry.getClass().getInterfaces(),
+                                                      new OpenTelemetryHandler(openTelemetry));
+    }
+
+    static Tracer tracer(OpenTelemetry openTelemetry, Tracer otelTracer) {
+        Class<?>[] allInterfaces = Arrays.copyOf(otelTracer.getClass().getInterfaces(),
+                                                 otelTracer.getClass().getInterfaces().length + 1);
+        allInterfaces[allInterfaces.length - 1] = ProxiedTracer.class;
+        return (Tracer) Proxy.newProxyInstance(OtelProxy.class.getClassLoader(),
+                                               allInterfaces,
+                                               new TracerHandler(otelTracer, HelidonOpenTelemetry.create(openTelemetry,
+                                                                                                         otelTracer,
+                                                                                                         Map.of())));
+    }
+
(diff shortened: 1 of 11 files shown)


microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
```diff
@@ -0,0 +1,263 @@
...
+                    TRACER_WITH_NAME_AND_VERSION = methodHelper.method("getTracer", String.class, String.class);
+                }
+            }
+
+        @Override
+            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+                if (TRACER_WITH_NAME.equals(method) || TRACER_WITH_NAME_AND_VERSION.equals(method)) {
+                    return OtelProxy.tracer(openTelemetry, (Tracer) method.invoke(openTelemetry, args));
+                } else {
+                    return method.invoke(openTelemetry, args);
```
ljnelson, line 108: You'll also want to [treat `equals` and `hashCode` methods very specially](https://javaclippings.wordpress.com/2009/03/18/dynamic-proxies-equals-hashcode-tostring/).

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
```diff
@@ -0,0 +1,263 @@
...
+
+        @Override
+            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+                if (method.equals(SPAN_BUILDER)) {
+                    io.helidon.tracing.Span.Builder<?> helidonSpanBuilder = helidonTracer.spanBuilder((String) args[0]);
+                    return OtelProxy.spanBuilder(helidonSpanBuilder);
+                } else if (method.equals(HELIDON_TRACER)) {
+                    return helidonTracer;
+                } else {
+                    return method.invoke(otelTracer, args);
```
ljnelson, line 136: See above.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
```diff
@@ -0,0 +1,263 @@
...
+                startInstant = Instant.ofEpochMilli(((TimeUnit) args[1]).toMillis((Long) args[0]));
+            } else if (START_SPAN.equals(method)) {
+                if (startInstant == null) {
+                    startInstant = Instant.now();
+                }
+                // The Helidon span builder will notify listeners and also delegate to the OTel span builder.
+                io.helidon.tracing.Span helidonSpan = helidonSpanBuilder.start(startInstant);
+                return OtelProxy.span((io.helidon.tracing.Span.Extended<Scope>) helidonSpan);
+            }
+            return method.invoke(otelSpanBuilder, args);
```
ljnelson, line 178: See above.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
```diff
@@ -0,0 +1,263 @@
...
+                    // Activating the Helidon span using helidonSpan.activate() would make the native OTel span stored inside the
+                    // Helidon span the current span for OTel. Normally that would be reasonable, but instead we need the current
+                    // OTel span to be our proxy so our proxy will be injected as the current span. That way, if user code
+                    // operates on an injected current OTel span it will actually operate on our proxy which will notify the
+                    // registered listeners.
+                    Scope otelScope = Context.current().with((Span) proxy).makeCurrent();
+                    io.helidon.tracing.Scope helidonScope = helidonSpan.activate(otelScope);
+                    return OtelProxy.scope(helidonScope, otelScope);
+                }
+                return method.invoke(otelSpan, args);
```
ljnelson, line 214: See above.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/OtelProxy.java
```diff
@@ -0,0 +1,263 @@
...
+                }
+            }
+
+        @Override
+            public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+                if (CLOSE.equals(method)) {
+                    helidonScope.close();
+                    return null;
+                } else {
+                    return method.invoke(otelScope, args);
```
ljnelson, line 234: See above.

tjquinno:
Using proxies seems to work in almost all respects but cannot be completely transparent in all the ways I wanted, so I'm reconsidering the approach and investigating a different implementation. Stay tuned.

tjquinno:
This PR is superseded by #9718.

Issue #9721, opened 2025-02-03
TLS TrustStore shouldn't include private key


wojtask9:
## Environment Details
* Helidon Version: 4.x
* Helidon SE
* JDK version: 21.x
* OS: Linux

----------

## Problem Description
[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"

Currently in Helidon is trying to add privateKey to trustStore (see https://github.com/helidon-io/helidon/blob/main/common/key-util/src/main/java/io/helidon/common/pki/KeysBuilderDecorator.java#L96). This is not neccessary. TrustStore should only contains certificates. 

It works if trustStore and keyStore are in the same file *.p12.

This can be done with `keytool` but using pure OpenSSL (starting from 3.3.x) *.p12 generated cannot be both keyStore and trusStore (see https://github.com/openssl/openssl/issues/22215 and https://docs.openssl.org/master/man1/openssl-pkcs12/#pkcs12-output-export-options for option -jdkTrust)

Helidon MTls examples supports both single/multiple keystore/truststore *.p12 files (https://github.com/helidon-io/helidon-examples/blob/helidon-4.x/examples/webserver/mutual-tls/automatic-store-generator.sh#L73) 

But with with multiple truststore/keystore we can see exception in logs using different files for keyStore and trustStore

```
2025.02.03 14:59:11 FINE io.helidon.common.pki.Keys Thread[#1,main,5,main]: Failed to read private key from default alias
io.helidon.common.pki.PkiException: Key stored under alias 1 is not a private key, but: null
	at io.helidon.common.pki@4.1.6/io.helidon.common.pki.PkiUtil.loadPrivateKey(PkiUtil.java:62)
	at io.helidon.common.pki@4.1.6/io.helidon.common.pki.KeysBuilderDecorator.updateFromKeystore(KeysBuilderDecorator.java:99)
	at io.helidon.common.pki@4.1.6/io.helidon.common.pki.KeysBuilderDecorator.lambda$decorate$0(KeysBuilderDecorator.java:37)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at io.helidon.common.pki@4.1.6/io.helidon.common.pki.KeysBuilderDecorator.decorate(KeysBuilderDecorator.java:37)
	at io.helidon.common.pki@4.1.6/io.helidon.common.pki.Keys$BuilderBase.preBuildPrototype(Keys.java:517)
[...]

PR #9722, opened 2025-02-03
Do not add privateKey if creating trustStore

wojtask9:
see https://github.com/helidon-io/helidon/issues/9721

### DIFF
--- a/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/ThreadPerTaskExecutor.java
+++ b/nima/webserver/webserver/src/main/java/io/helidon/nima/webserver/ThreadPerTaskExecutor.java
@@ -44,6 +44,7 @@ class ThreadPerTaskExecutor implements HelidonTaskExecutor {
     private final ThreadFactory factory;
     private final Map<Thread, Object> threadTasks = new ConcurrentHashMap<>();
     private final CountDownLatch terminationSignal = new CountDownLatch(1);
+    private final ClassLoader contextClassLoader;
 
     // states: RUNNING -> SHUTDOWN -> TERMINATED
     private static final int RUNNING = 0;
@@ -53,6 +54,7 @@ class ThreadPerTaskExecutor implements HelidonTaskExecutor {
 
     private ThreadPerTaskExecutor(ThreadFactory factory) {
         this.factory = Objects.requireNonNull(factory);
+        this.contextClassLoader = Thread.currentThread().getContextClassLoader();
     }
 
     static HelidonTaskExecutor create(ThreadFactory factory) {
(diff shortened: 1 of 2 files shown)


common/key-util/src/main/java/io/helidon/common/pki/KeysBuilderDecorator.java
```diff
@@ -93,7 +93,7 @@ private void updateFromKeystore(Keys.BuilderBase<?, ?> builder, KeystoreKeys key
         Optional<String> keyAliasConfigured = keystoreConfig.keyAlias();
         String keyAlias = keyAliasConfigured.orElse(KeystoreKeysBlueprint.DEFAULT_PRIVATE_KEY_ALIAS);
 
-        if (builder.privateKey().isEmpty()) {
+        if (builder.privateKey().isEmpty() || !keystoreConfig.trustStore()) {
```
tomas-langer, line 96: I think this should be `if (builder.privateKey().isEmpty() && !keystoreConfig.trustStore()) {` - i.e. only guess a private key if this is not a trust store and the private key is empty.

common/key-util/src/main/java/io/helidon/common/pki/KeysBuilderDecorator.java
```diff
@@ -106,7 +106,7 @@ private void updateFromKeystore(Keys.BuilderBase<?, ?> builder, KeystoreKeys key
             }
         }
         List<X509Certificate> certChain = null;
-        if (builder.certChain().isEmpty()) {
+        if (builder.certChain().isEmpty() || !keystoreConfig.trustStore()) {
```
tomas-langer, line 109: Same here.

tomas-langer (changes requested):

common/key-util/src/main/java/io/helidon/common/pki/KeysBuilderDecorator.java
wojtask9: yes. fixed

common/key-util/src/main/java/io/helidon/common/pki/KeysBuilderDecorator.java
wojtask9: @tomas-langer  fixed

romain-grecourt:
I tried rebasing and force pushing to your branch but it somehow closed the pull request. This could be related to the use of the master branch or a bug in GitHub.

Can you try to re-open or re-create this PR against a different branch ?

romain-grecourt:
Re-opening...

romain-grecourt:
Here is a backup of your commits, in case you can't find them: https://github.com/romain-grecourt/helidon/commits/9722-backup/

Issue #9734, opened 2025-02-05
Config for OCI integration module will not work with helidon.oci prefix as documented in the readme


klustria:
## Environment Details
* Helidon Version: 4.1.x
* Helidon SE
* JDK version: 21

----------

## Problem Description
The service registry OCI Integration Module is documented as prefixed by `helidon.oci`:  https://github.com/helidon-io/helidon/blob/main/integrations/oci/oci/README.md. However it does not work when configured to use that prefix.
 
## Steps to reproduce
1. Use helidon.oci to prefix parameter configuration in the oci-config.yaml, environment variables or system properties.
2. Retrieve an instance of an OCI authentication method and it will fail as the configuration will not take effect. The failure may manifest with this type of error message: 
    ```
    ServiceRegistry Contract io.helidon.integrations.oci.ImdsInstanceInfo is not supported, there are no service descriptors in this registry that can satisfy it.
    ```

----------

## Additional Information
There is a problem with the implementation of OciConfigProvider, where config.get("helidon.oci") was not implemented , so it currently uses the root of the configuration file.

Issue #9681, opened 2025-01-23
OKE Workload Provider from Service Registry fails when Instance Metadata Service does not exist


klustria:
## Environment Details:
* Helidon Version: v4.1.x
* Helidon SE
* JDK version: 21

## Problem Description
---
There are scenarios where in a Kubernetes pod, the Instance Metadata Service (IMDS) does not exist. Because of this OKE Workload Provider will fail if FederationEndpoint is not set. The problem is, currently we don't have a way to pass in FederationEndpoint to the instantiation of OKE Workload Provider.

The issue is manifested with this log:

```
2025-01-22 17:26:30.184  [main] WARN  c.o.b.a.AbstractFederationClientAuthenticationDetailsProviderBuilder - Attempt 1 - Rest call to get region from metadata service failed 
jakarta.ws.rs.ProcessingException: org.apache.http.conn.ConnectTimeoutException: Connect to 169.254.169.254:80 [/169.254.169.254] failed: Connection timed out
	at org.glassfish.jersey.apache.connector.ApacheConnector.apply(ApacheConnector.java:533)
	at org.glassfish.jersey.apache.connector.ApacheConnector.apply(ApacheConnector.java:540)
	at org.glassfish.jersey.client.ClientRuntime.lambda$createRunnableForAsyncProcessing$6(ClientRuntime.java:185)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
...
...
2025-01-22 17:28:43.375  [main] INFO  c.o.b.a.AbstractFederationClientAuthenticationDetailsProviderBuilder - Exiting retry 2 with wait time: 2117 millis
2025-01-22 17:29:35.996  [SIGTERM handler] DEBUG java.lang.Runtime - Runtime.exit() called with status: 143
java.lang.Throwable: Runtime.exit(143)
	at java.base/java.lang.Shutdown.logRuntimeExit(Shutdown.java:179)
	at java.base/java.lang.Shutdown.exit(Shutdown.java:160)
	at java.base/java.lang.Terminator$1.handle(Terminator.java:51)
	at java.base/jdk.internal.misc.Signal$1.run(Signal.java:218)
	at java.base/java.lang.Thread.run(Thread.java:1583)
2025-01-22 17:29:35.997  [helidon-shutdown-thread] INFO  io.helidon.Main - Shutdown requested by JVM shutting down
[...]

PR #9740, opened 2025-02-06
Use helidon.oci prefix for OCI Config and allow some oci auth types to accept federation-endpoint and tenancy-id

klustria:
### Description
The PR fixes Issues [9681](https://github.com/helidon-io/helidon/issues/9681) and [9734](https://github.com/helidon-io/helidon/issues/9734) which includes the following:
1. Allow instance-principal, resource-principal and oke-workload-identity to accept federation-endpoint and tenancy-id as config parameters. This is originally targeted just for oke-workload-identity where Instance Metadata Service (IMDS) does not work on an OKE environment. Because of these, it is unable to assemble the target endpoint as it needs the IMDS to retrieve the region. To resolve the issue, the federation-endpoint configuration is now allowed to be explicitly specified to avoid generation of endpoint using the region from IMDS. In some examples of the use of oke-workload-identity, the tenancy id is required, so this configuration parameter is also added as an option. Furthermore, because instance-principal and resource-principal providers extends AbstractRequestingAuthenticationDetailsProvider similar to oke-workload-instance, hence they are included in the change to allow those optional parameters.
2. Fix a bug where the oci configuration does not work when prefixed with "helidon.oci".
3. Add comprehensive testing coverage for above changes.

### Documentation
Related readme files were already updated as part of this change

If no doc impact: None

### DIFF
--- a/integrations/oci/authentication/resource/src/test/java/io/helidon/integrations/oci/authentication/resource/MockedAuthenticationMethodResourcePrincipal.java
+++ b/integrations/oci/authentication/resource/src/test/java/io/helidon/integrations/oci/authentication/resource/MockedAuthenticationMethodResourcePrincipal.java
@@ -25,7 +25,6 @@ import io.helidon.service.registry.Service;
 
 import com.oracle.bmc.auth.ResourcePrincipalAuthenticationDetailsProvider.ResourcePrincipalAuthenticationDetailsProviderBuilder;
 
-@Weight(Weighted.DEFAULT_WEIGHT)
 @Service.Provider
 class MockedAuthenticationMethodResourcePrincipal extends AuthenticationMethodResourcePrincipal {
 
--- a/integrations/oci/authentication/resource/src/test/java/io/helidon/integrations/oci/authentication/resource/MockedAuthenticationMethodResourcePrincipal.java
+++ b/integrations/oci/authentication/resource/src/test/java/io/helidon/integrations/oci/authentication/resource/MockedAuthenticationMethodResourcePrincipal.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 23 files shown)


integrations/oci/authentication/resource/src/test/java/io/helidon/integrations/oci/authentication/resource/MockedAuthenticationMethodResourcePrincipal.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Optional;
+import java.util.function.Supplier;
+
+import io.helidon.common.Weight;
+import io.helidon.common.Weighted;
+import io.helidon.service.registry.Service;
+
+import com.oracle.bmc.auth.ResourcePrincipalAuthenticationDetailsProvider.ResourcePrincipalAuthenticationDetailsProviderBuilder;
+
+@Weight(Weighted.DEFAULT_WEIGHT)
```
tomas-langer, line 28: Default weight, as the name shows is the default, so no need to annotate at all. 
If you want to replace a service with default weight, you would need to use higher than default.

integrations/oci/oci/src/main/java/io/helidon/integrations/oci/OciConfigProvider.java
```diff
@@ -71,7 +73,14 @@ private static void create() {
                 ConfigSources.systemProperties(),
                 ConfigSources.file("oci-config.yaml").optional(true),
                 ConfigSources.classpath("oci-config.yaml").optional(true));
-
-        ociConfig = OciConfig.create(config);
+        if (config.get(CONFIG_PREFIX).exists()) {
+            ociConfig = OciConfig.create(config.get(CONFIG_PREFIX));
+        } else {
+            ociConfig = OciConfig.create(config);
```
tomas-langer, line 79: I realized we cannot do this - as the configuration is not required (there are reasonable defaults), you would always warn, unless somebody adds the `oci-config.yaml` and adds at least one `helidon.oci` property there.
So either we warn if we find one of the supported options (I think this is too complicated to implement maybe?), or just log it in `TRACE`.

klustria: Yeah, 1st option is complicated, so I adopted the 2nd option to just log it in `TRACE`

integrations/oci/authentication/resource/src/test/java/io/helidon/integrations/oci/authentication/resource/MockedAuthenticationMethodResourcePrincipal.java
klustria: Good point. Remove all `@Weight` with DEFAULT_WEIGHT value. My goal is really to use  DEFAULT_WEIGHT as it is higher enough than the providers that I am overriding, so no need to use other values or a value higher than the default.

tomas-langer (approved):

klustria:
This PR should resolved theses issues: https://github.com/helidon-io/helidon/issues/9681 and https://github.com/helidon-io/helidon/issues/9734

Issue #9741, opened 2025-02-06
Helidon Inject documentation


Verdent:
Create Helidon Inject user documentation

PR #9742, opened 2025-02-06
Helidon Inject documentation

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/9741

### Documentation

Adds user documentation for Helidon Inject


docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,659 @@
...
+- <<Qualifiers, Qualifiers>>
+- <<Factories, Factories>>
+- <<Interceptors, Interceptors>>
+- <<Events, Events>>
+- <<Programmatic Lookup, Programmatic Lookup>>
+- <<Startup, Startup>>
+
+== Overview
+
+Injection is the basic building stone for inversion of control. Dependency injection provides a mechanism to get
```
edbratt, line 50: I would recommend we start with the basics of Injection and IoC. SE developers, prior to 4.2 would be focused only on procedural coding thus a very brief introduction to what is Injection, what IoC is, why use Injection; why consider adopting IoC will help users. Include why we didn't just adopt CDI (light-weight, smaller, faster, easier to understand, ...) I also think we should describe the goals achieved with Helidon Inject (How/Why is Helidon Inject different from other Injection frameworks that a reader might have used/heard about. Some potential differentiators: -- compile-time generation (immutable images, more secure)-- source-code, not byte-code (easier debugging/stack traces) ... (add more beneficial facets as you think is appropriate). 
Give the reader a bit of a primer on Inject, IoC, and set up reason to read further.

docs/src/main/java/io/helidon/docs/se/inject/BasicExample.java
```diff
@@ -0,0 +1,59 @@
...
+
+        void printGreeting(String name) {
+            System.out.println(greeter.greet(name));
+        }
+    }
+    // end::snippet_2[]
+
+    // tag::snippet_3[]
+    public static void main(String[] args) {
+        var registry = ServiceRegistryManager.create().registry();
```
tomas-langer, line 53: If you use service registry like this, you should keep the manager instance and shut it down after use (to correctly handle all `@PreDestroy` methods and clean-up resources.

docs/src/main/java/io/helidon/docs/se/inject/EventsExample.java
```diff
@@ -0,0 +1,101 @@
...
+        @Event.AsyncObserver
+        void event(MyEvent event) {
+            //Do something with the event
+        }
+    }
+    // end::snippet_5[]
+
+    // tag::snippet_6[]
+    @Service.Singleton
+    record MyIdProducer(@Service.Named("id") Event.Emitter<String> emitter) {
```
tomas-langer, line 81: It would be better to use custom qualifiers instead of a named qualifier.

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
```diff
@@ -0,0 +1,101 @@
...
+        @Override
+        public MyService get() {
+            return new MyService();
+        }
+    }
+    // end::snippet_1[]
+
+    // tag::snippet_2[]
+    @Service.Singleton
+    class MyServiceFactory implements Service.ServicesFactory<MyService> {
```
tomas-langer, line 47: `ServicesFactory` that creates a single unqualified instead does not make any sense.

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
```diff
@@ -0,0 +1,101 @@
...
+        public List<Service.QualifiedInstance<MyService>> services() {
+            return List.of(Service.QualifiedInstance.create(new MyService()));
+        }
+    }
+    // end::snippet_2[]
+
+    // tag::snippet_3[]
+    @Service.Singleton
+    @Service.Named("name")
+    class MyServiceFactoryWithQualifier implements Service.ServicesFactory<MyService> {
```
tomas-langer, line 58: And this is even worse, as this should be implemented as a named `Supplier`, not a `ServicesFactory`. 
This shows a very complicated way to do something very simple.
`ServicesFactory` should only be used for its purpose - creating zero to n instances at runtime, each with different qualifiers (i.e. multiple named instances, or multiple instances with some other qualifier). 
In most cases, the factory itself should not be qualified (this is now fixed in a PR for services factory: #9768)

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
```diff
@@ -0,0 +1,101 @@
...
+        @Override
+        public List<Service.QualifiedInstance<MyService>> services() {
+            return List.of(Service.QualifiedInstance.create(new MyService(), Qualifier.createNamed("name")));
+        }
+    }
+    // end::snippet_3[]
+
+    // tag::snippet_4[]
+    @Service.Singleton
+    class QualifiedFactory implements Service.QualifiedFactory<MyService, Service.Named> {
```
tomas-langer, line 68: Please never use `Named` in examples that are not showing how named works.
Use a custom qualifier (i.e. `@ConfigProperty` or something like that) for examples showing qualified factories (and anything else related to qualifiers).
Named has a slightly different handling.

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+    // end::snippet_4[]
+
+    // tag::snippet_5[]
+    @Service.Singleton
+    class InjectionPointFactory implements Service.InjectionPointFactory<MyService> {
+
+        @Override
+        public Optional<Service.QualifiedInstance<MyService>> first(Lookup lookup) {
```
tomas-langer, line 84: This example is just showing that you can implement a method of an interface, not how to query the lookup, which is the reason this factory exists.

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
```diff
@@ -0,0 +1,101 @@
...
+            return Optional.of(Service.QualifiedInstance.create(new MyService()));
+        }
+    }
+    // end::snippet_5[]
+
+    // tag::snippet_6[]
+    @Service.Singleton
+    @Service.Named("name")
```
tomas-langer, line 92: Again, using a name qualifier does not work nicely in this example.

docs/src/main/java/io/helidon/docs/se/inject/InterceptorExample.java
```diff
@@ -0,0 +1,74 @@
...
+    @Service.Singleton
+    static class MyServiceProvider implements Supplier<MyService> {
+        @Override
+        public MyService get() {
+            return new MyService();
+        }
+    }
+
+    @Service.Contract
+    @Interception.Delegate
```
tomas-langer, line 63: This is not an interception delegate example - this example is probably a bad copy paste?
To show interception, you annotate a service method with the intercepted annotation, or a contract interface method.

docs/src/main/java/io/helidon/docs/se/inject/ServiceRegistryExample.java
```diff
@@ -0,0 +1,31 @@
...
+package io.helidon.docs.se.inject;
+
+import io.helidon.service.registry.ServiceRegistryManager;
+
+class ServiceRegistryExample {
+
+    public void programmatic() {
+        // tag::snippet_1[]
+        // create an instance of a registry manager - can be configured and shut down
+        var registryManager = ServiceRegistryManager.create();
```
tomas-langer, line 25: Please add the shutdown example as well.

tomas-langer (changes requested):
Review of examples (`injection.adoc` will come later).

docs/src/main/java/io/helidon/docs/se/inject/EventsExample.java
Verdent: If I want this to make sense, I would need to paste custom qualifier creation there as well. What I am trying to do, is to examplain the emitter with qualifier. Since we do provider named one out of the box, it does not need to be made and this is te simplest example from my point of view.

docs/src/main/java/io/helidon/docs/se/inject/InterceptorExample.java
Verdent: It should not be here. Removed

docs/src/main/java/io/helidon/docs/se/inject/ServiceRegistryExample.java
Verdent: added

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
Verdent: Changed

docs/src/main/java/io/helidon/docs/se/inject/FactoryExample.java
Verdent: removed

docs/src/main/java/io/helidon/docs/se/inject/BasicExample.java
Verdent: Added.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,677 @@
...
+== Programmatic Lookup
+
+There are two primary ways to access services from the service registry:
+
+- *Dependency Injection* – Services are injected automatically at defined injection points.
+- *Programmatic Lookup* – Services are retrieved manually by accessing the service registry directly.
+
+If you want to use programmatic lookup, there are two main ways to access a `ServiceRegistry` instance:
+
+- Create a new ServiceRegistry instance and search for the desired service.
```
tomas-langer, line 614: You can use `GlobalServiceRegistry` to obtain a shared registry instance that is used by Helidon components - this should be the first way of doing programmatic work with service registry.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,677 @@
...
+- Inject the existing ServiceRegistry instance currently used by Helidon and retrieve services from it.
```
tomas-langer, line 615: This should only be used for advanced use cases, as it bypasses a lot of optimizations of the service registry (i.e. build time binding of services to injection points when using the Maven plugin)

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,677 @@
...
+[source,java]
+----
+include::{sourcedir}/se/inject/ServiceRegistryExample.java[tag=snippet_1, indent=0]
+----
+
+Note that all instances are created lazily, meaning the service registry does nothing by default.
+If a service performs any actions during construction or post-construction,
+you must first retrieve an instance from the registry to trigger its initialization.
+
+Special registry operations:
```
tomas-langer, line 634: These are advanced methods, which we should not even mention in documentation. Just document the methods that operate on contracts.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,677 @@
...
+- `Supplier<Optional<T>> supplyFirst(...)` - get a supplier of an optional instance
+- `Supplier<List<T>> supplyAll(...)` - get a supplier of all instances
+
+Lookup parameter options:
+
+- `Class<?>` - the contract we are looking for
+- `TypeName` - the same, but using Helidon abstraction of type names (may have type arguments)
+- `Lookup` - a full search criteria for a registry lookup
+
+== Startup
```
tomas-langer, line 657: This seems to be a bit unfinished, as we need explanation of `RunLevel` and how it is used in generated main class, and in `io.helidon.Main`.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,677 @@
...
+
+Methods that accept the bindings are on `io.helidon.service.registry.ServiceRegistryManager`:
+
+- `start(Binding)` - starts the service registry with the generated binding, initializing all singleton and per-lookup services annotated with a `@RunLevel` annotation (i.e. `start(ApplicationBinding.create())`)
+- `start(Binding, ServiceRegistryConfig)` - same as above, allows for customization of configuration, if used, do not forget to set discovery to `false` to prevent automated discovery from the classpath
+
+All options to start a Helidon application that uses service registry:
+
+- A custom Main method using `ServiceRegistryManager.start(...)` methods, or `ServiceRegistryManager.create(...)` methods
+- A generated `ApplicationMain` - optional feature of the Maven plugin, requires property `generateMain` to be set to `true`
```
tomas-langer, line 671: also mention it uses `RunLevel` actively, but via code generated class.
This is the only approach that is fully reflection free and skips lookups for injection points.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,677 @@
...
+Lookup parameter options:
+
+- `Class<?>` - the contract we are looking for
+- `TypeName` - the same, but using Helidon abstraction of type names (may have type arguments)
+- `Lookup` - a full search criteria for a registry lookup
+
+== Startup
+
+Helidon provides a Maven plugin (`io.helidon.service:helidon-service-maven-plugin`, goal `create-application`) to generate
+build time bindings, that can be used to start the service registry without any classpath discovery and reflection.
```
tomas-langer, line 660: `ApplicationBinding` provides build time bindings of service providers to injection points - this is to avoid lookup at runtime.
It does NOT elimiminate classpath discovery and reflection. To avoid that you need to use the generated `ApplicationMain` class.

Also classpath discovery does not mean classpath scanning (it sounds a bit bad in here) - we lookup all `service.registry` and `service.loader` resources from the classpath - we are not scanning for them (i.e. using only Java features that do not depend on how the classpath is created, so it works in native image as well)

Not sure how much of this should be in the docs, but it seems  a bit misleading right now.

tomas-langer (changes requested):

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+
+[source,java]
+----
+include::{sourcedir}/se/inject/FactoryExample.java[tag=snippet_2, indent=0]
+----
+
+=== QualifiedFactory
+link:{service-registry-base-url}/io/helidon/service/registry/Service.QualifiedFactory.html[`QualifiedFactory`]
+is strictly bound to a specific qualifier type.
+It will get executed only for injection points
```
tomas-langer, line 420: The special case should be used here for `java.lang.Object` as a contract - in such a case, the factory will be used for any contract, as long as the qualifier matches.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+
+Method calls on an instance created this way can’t be intercepted. To enable interception in such cases, we use the
+link:{service-registry-base-url}/io/helidon/service/registry/Interception.Delegate.html[`@Interception.Delegate`] annotation.
+However, keep in mind that usage of this annotation doesn’t add the ability to intercept constructor calls.
+To enable interception, this annotation must be present on the class that the factory produces.
+While it is not required on interfaces, it will still work correctly if applied there.
+
+If you need to enable interception for classes using delegation, you should make sure about the following:
+
+- The class must have accessible no-arg constructor (at least package local) and mustn’t be final
```
tomas-langer, line 537: `mustn't` seems wrong to me here, maybe some native speaker can help

tomas-langer: And it is still missing the requirement that the class itself has to be extensible (i.e. not `final`), as we are generating a subclass to handle interception.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+- The constructor should have no side effects, as the instance will act only as a wrapper for the delegate
+- All invoked methods must be accessible (at least package local) and non-final
+
+Therefore, using link:{service-registry-base-url}/io/helidon/service/registry/Interception.Delegate.html[`@Interception.Delegate`]
+on the classes can be considered as acknowledgement of the rules above.
```
tomas-langer, line 542: The annotation is not an `acknowledgement`, it is used to generate a subclass. It also validates the rules are followed, but a delegate for a class will not be generated, and interception cannot work, unless annotated

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+----
+
+== Events
+Events enable in-application communication between services by providing a mechanism to emit events and register consumers to handle them.
+
+A single event can be delivered to zero or more consumers.
+
+Key Terminology:
+
+- *Event Object* – Any object that is sent as an event.
```
tomas-langer, line 593: These should be links to the chapters below (`Event Object`, `Event Emitter` etc.)

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+Event emitters are code generated by Helidon. The emitter is generated when an injection point is discovered that expects it.
+
+All the emitters are generated as implementations of the
+link:{service-registry-base-url}/io/helidon/service/registry/Event.Emitter.html[`Event.Emitter`] interface.
+
+=== Event Producer
+An event producer is a service that triggers the event by using the event emitter.
+
+To emit an event, inject the desired
+link:{service-registry-base-url}/io/helidon/service/registry/Event.Emitter.html[`Event.Emitter`]
```
tomas-langer, line 620: The example and text contains too many `emit` words. Maybe rename the method in the event producer to `publish` or `produce` to get rid of at least one of them.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+instance which is used to create the service instance.
+
+To create a registry instance:
+
+[source,java]
+----
+include::{sourcedir}/se/inject/ServiceRegistryExample.java[tag=snippet_1, indent=0]
+----
+
+Keep in mind, that
```
tomas-langer, line 746: I would also add a note here that in most cases, the single, global service registry is sufficient (and expected to be used) for a single application.
A custom service registry should only be created for specific use cases.
When I read this, it seems the three options are equal for use, but that is not the case. 
The default use case is to use `Services` to access the registry, or use the `GlobalServiceRegistry` for more fine grained use cases.
Creating a new registry manager is only intended for custom Main classes, or advanced use cases. It should be clear here, as otherwise people will end up creating multiple registries and it will not work as expected.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+link:{service-registry-base-url}/io/helidon/service/registry/Service.RunLevel.html[`@Service.RunLevel`] annotations
+present on the service.
+See <<RunLevel,RunLevel>> chapter.
+
+It is important to note, that once you don’t need your service registry,
+method `shutdown` on the manager must be called to ensure proper termination of the service registry.
+
+=== RunLevel
+link:{service-registry-base-url}/io/helidon/service/registry/Service.RunLevel.html[`@Service.RunLevel`]
+is the annotation used for specifying the order in which certain services should be created.
```
tomas-langer, line 823: I would not use "order .. be created". 
The run level provides steps in which an application is started, and within each step (i.e. `@Service.RunLevel(10)`), you may have zero to n services to be started. 
The order of creation is defined by `@Weight`; Run level creates "groups" of services.

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+To start, add the
+link:{service-registry-base-url}/io/helidon/service/registry/Service.RunLevel.html[`@Service.RunLevel`]
+on the service.
+
+[source,java]
+----
+include::{sourcedir}/se/inject/RunLevelExample.java[tag=snippet_1, indent=0]
+----
+
+For better understanding, we can also add helpful method, which get executed upon service construction.
```
tomas-langer, line 835: This is probably obsoleted by switching to `start` method.

tomas-langer (changes requested):

docs/src/main/asciidoc/se/introduction.adoc
```diff
@@ -78,6 +78,13 @@ Build gRPC servers and clients.
 --
 Expose health statuses of your applications.
 --
+//Injection
+[CARD]
+.Injection
+[icon=message,link=injection.adoc]
```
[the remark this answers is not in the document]
Verdent: Ah... yes, I did not do this yet :-)

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+.Create simple injection point
+----
+include::{sourcedir}/se/inject/BasicExample.java[tag=snippet_2, indent=0]
+----
+
+Now it just needs to be tested. The easiest way is to make a main method. The following piece of code
+initializes Service registry. After that we search for our `GreetingInjectionService` and execute it
+to print out `Hello David!`. To find out more about this manual approach, please take a look into the <<Programmatic Lookup, Programmatic Lookup>> chapter.
+[source,java]
+.Lookup our created service and execute it manually
```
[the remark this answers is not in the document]
Verdent: Definitely! I did not know this exists

docs/src/main/asciidoc/se/injection.adoc
```diff
@@ -0,0 +1,850 @@
...
+.Named by type usage example
+----
+include::{sourcedir}/se/inject/QualifierExample.java[tag=snippet_4, indent=0]
+----
+
+The way it is used on the injection point, it is the same as it was in case of the
+link:{service-registry-base-url}/io/helidon/service/registry/Service.Named.html[`@Service.Named`].
+
+[source,java]
+.Named by type injection point
```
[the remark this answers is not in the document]
Verdent: Example added

docs/src/main/asciidoc/se/injection.adoc
Verdent: Hm... that was supposed to be that `mustn’t be final`. But perhaps it looked like I am talking about the constructor?

tomas-langer (approved):

Issue #9592, opened 2024-12-13
Cookie is not deleted properly


hrstoyanov:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 4.1.4
* Helidon SE or Helidon MP: SE
* JDK version: 22
* OS: MacOS

----------

## Problem Description

This api: 
```java 
ServerResonseImpl.clearCookie(String name)
``` 
appears buggy. It hard-codes the **Path**  attribute to "/" and pay no respect to the **Domain** attribute. **Name**, **Domain**, and **Path** are the primary identifiers for a cookie when it comes to deletion - you need to know these exactly to delete the correct cookie. The **Secure** and **HttpOnly** flags do not identify the cookie, but must be replicated when deleting the cookie.

Perhaps, it is best to remove this api altogether, and use SetCookie directly as a workaround (but you have to make START_OF_YEAR_1970) public:
```java
var deleteCookie = SetCookie
                .builder(cookieName, "00000000-0000-0000-0000-000000000000")
                .httpOnly(true)
                .domainAndPath(cookieDomainAndPath)
                .secure(request.isSecure())
                .sameSite(SetCookie.SameSite.STRICT)
                .expires(START_OF_YEAR_1970)
                .build();
        response.headers().addCookie(deleteCookie);
```

spericas:
@hrstoyanov I'm assuming you mean `ServerResponseHeadersImpl.clearCookie(String)`?

spericas:
Work in progress PR #9751

PR #9751, opened 2025-02-07
Improves handling of cookies in server responses

spericas:
### Description

Improves handling of cookies in server responses:

- A cookie's identity is now defined by its name, domain and path, not just its name as before
- A new method in `ServerResponseHeaders` enables clearing a cookie by passing all three values (a complete cookie)
- The equals and hashCode methods in `SetCookie` have been overridden as well
- Backward compatiblity with old clear cookie behavior is preserved, but Javadoc suggests using new method
- New unit tests added
- Issue #9592

### Documentation

None

### DIFF
--- a/http/http/src/main/java/io/helidon/http/ServerResponseHeaders.java
+++ b/http/http/src/main/java/io/helidon/http/ServerResponseHeaders.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -106,11 +106,23 @@ public interface ServerResponseHeaders extends ClientResponseHeaders,
 
     /**
      * Clears a cookie by adding a {@code Set-Cookie} header with an expiration date in the past.
+     * It is recommended to use {@link #clearCookie(SetCookie)} instead for better handling
+     * of Path and Domain.
      *
      * @param name name of the cookie.
      * @return this instance
      */
     ServerResponseHeaders clearCookie(String name);
+
+    /**
+     * Clears a cookie by adding a {@code Set-Cookie} header with an expiration date in the past.
+     *
+     * @param setCookie the cookie.
+     * @return this instance
+     */
+    default ServerResponseHeaders clearCookie(SetCookie setCookie) {
+        return clearCookie(setCookie.name());
+    }
 
     /**
      * Sets the value of {@link HeaderNames#LAST_MODIFIED} header.
--- a/http/http/src/main/java/io/helidon/http/ServerResponseHeadersImpl.java
+++ b/http/http/src/main/java/io/helidon/http/ServerResponseHeadersImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


http/http/src/main/java/io/helidon/http/ServerResponseHeadersImpl.java
```diff
@@ -63,15 +77,13 @@ public ServerResponseHeaders clearCookie(String name) {
                 if (!found) {
                     String[] values = new String[newValues.length + 1];
                     System.arraycopy(newValues, 0, values, 0, newValues.length);
-                    values[values.length - 1] = expiredCookie.text();
+                    values[values.length - 1] = expiredCookie.text();   // replace with expired
```
tjquinno, line 80: IIUC the logic in the "then" part of this `if` handles the case where the cookie requested to be cleared _was not_ present in the original cookie list and so the expired version of the cookie is being _added_ to the end of a newly-created list that has an extra slot to hold the extra (expired) cookie. 

The comment using "replace" here confused me for a moment I guess.

http/http/src/main/java/io/helidon/http/SetCookie.java
```diff
@@ -78,7 +88,7 @@ public static SetCookie parse(String setCookie) {
         String nameAndValue = cookieParts[0];
         int equalsIndex = nameAndValue.indexOf('=');
         String name = nameAndValue.substring(0, equalsIndex);
-        String value = nameAndValue.length() == equalsIndex ? null : nameAndValue.substring(equalsIndex + 1);
+        String value = nameAndValue.substring(equalsIndex + 1);
```
tjquinno, line 91: Not all of this is changed by this PR, I realize, but:
* This is a public method and there is no checking of the input string prior to the `split` invocation. So the array might be empty so `cookieParts[0]` would trigger a potentially-confusing `IndexOutOfBoundsException`.
* Similarly, the `=` might not appear in the `nameAndValue` string so by removing the check of `equalsIndex` the PR might introduce an `IndexOutOfBoundsException`.

Maybe, practically speaking, these are not serious issues. I know we would not want to slow down cookie processing if we can avoid it and maybe those two possible exceptions from Java are acceptable given that the cookie values are, realistically, very likely coming from well-behaved clients sending properly-formatted cookie expressions.

Just a thought.

http/http/src/main/java/io/helidon/http/ServerResponseHeaders.java
```diff
@@ -106,12 +106,24 @@ default ServerResponseHeaders addCookie(String name, String value) {
 
     /**
      * Clears a cookie by adding a {@code Set-Cookie} header with an expiration date in the past.
+     * It is recommended to use {@link #clearCookie(SetCookie)} instead for better handling
+     * of Path and Domain.
      *
      * @param name name of the cookie.
      * @return this instance
      */
     ServerResponseHeaders clearCookie(String name);
```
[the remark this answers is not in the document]
spericas: Well, thought about that, but this method is simpler to use and I suspect for most people sufficient (i.e., for most apps cookie names should be unique?)

http/http/src/main/java/io/helidon/http/ServerResponseHeadersImpl.java
spericas: Will fix comment

http/http/src/main/java/io/helidon/http/SetCookie.java
spericas: The check `nameAndValue.length() == equalsIndex` was removed _because it was wrong and always false_. An index by definition is either -1 or in the range `[0...length-1]`, so it cannot be equal to `length`. If -1, it will result in the empty string as value. I'll see about adding other checks for the params.

spericas: Added checks and new test

http/http/src/main/java/io/helidon/http/ServerResponseHeaders.java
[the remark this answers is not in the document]
tjquinno: Got it. I misinterpreted the comment but I now see it refers explicitly to better handling for path and domain.

tjquinno (approved):

PR #9765, opened 2025-02-10
4.1.7: Use helidon.oci prefix for OCI Config and allow some oci auth types to accept federation-endpoint and tenancy-id

barchetta:
Backport #9740 to Helidon 4.1.7

### Description
The PR fixes Issues [9681](https://github.com/helidon-io/helidon/issues/9681) and [9734](https://github.com/helidon-io/helidon/issues/9734) which includes the following:
1. Allow instance-principal, resource-principal and oke-workload-identity to accept federation-endpoint and tenancy-id as config parameters. This is originally targeted just for oke-workload-identity where Instance Metadata Service (IMDS) does not work on an OKE environment. Because of these, it is unable to assemble the target endpoint as it needs the IMDS to retrieve the region. To resolve the issue, the federation-endpoint configuration is now allowed to be explicitly specified to avoid generation of endpoint using the region from IMDS. In some examples of the use of oke-workload-identity, the tenancy id is required, so this configuration parameter is also added as an option. Furthermore, because instance-principal and resource-principal providers extends AbstractRequestingAuthenticationDetailsProvider similar to oke-workload-instance, hence they are included in the change to allow those optional parameters.
2. Fix a bug where the oci configuration does not work when prefixed with "helidon.oci".
3. Add comprehensive testing coverage for above changes.

### Documentation
Related readme files were already updated as part of this change

If no doc impact: None


klustria (approved):
LGTM

Issue #9731, opened 2025-02-05
OIDC Token Validation Fails on Every Alternate Request


jay-jariwa:
**Issue Description:**

---

We are integrating OIDC security provider with Oracle Access Manager (OAM) as the identity provider. When we make the first request to helidon server, the token is validated correctly, and we receive a response immeditely (~ 1 sec) . However, on the second request with the same token, there is no response for 25-30 seconds, and it eventually fails due to a timeout error.

The issue follows a consistent pattern:

1st request → ✅ Passes
2nd request → ❌ Fails (timeout)
3rd request → ✅ Passes
4th request → ❌ Fails (timeout)
5th request → ✅ Passes
(Repeats in the same cycle)

We found same pattern when we used sample OIDC project with branch helidon-4.x with below configutaion.

link - https://github.com/helidon-io/helidon-examples/tree/helidon-4.x/examples/microprofile/oidc

    - oidc:
        # use a custom name, so it does not clash with other examples
        cookie-name: "OIDC_EXAMPLE_COOKIE"
        # support for "Authorization" header with bearer token
        header-use: true
        # the default redirect-uri, where the webserver listens on redirects from identity server
        redirect-uri: "/oidc/redirect"
        audience: "ResourceServer"
        client-id: OAUTH2CLIENT
        client-secret: ***************
        identity-uri: http://xxx:7777/
        frontend-uri: "${security.properties.frontend-uri}"
        introspect-endpoint-uri: http:/xxx:14100/oauth2/rest/token/introspect?identityDomain=IDStore
        # We want to redirect to login page (and token can be received either through cookie or header)
        redirect: false

**Environment Details:**

---

**Product's Environment Details** 

Client Version: v1.28.2
Server Version: v1.28.2
Helidon : v4.1.6
JDK : 21
System Version: "Oracle Linux Server 8.9"

**Sample OIDC environment details**

Helidon : 4.1.6
JDK : 21 
System Version: macOS 14.7.1 (23H222)

**Error Logs:**

---

[Logs.txt](https://github.com/user-attachments/files/18667473/Logs.txt)

 **Additional Information:**

---

1. If a request is made for the first time, it is successfully validated. However, if another request is sent with the same token within ~30 seconds, it fails with a 401 Unauthorized error. If we wait for approximately 30 seconds before making the next request, the token is validated successfully again. This pattern repeats consistently. 

[...]

PR #9795, opened 2025-02-13
Connection cache 100 continue

### LINKED ISSUE #9736: 4.x: WebClient.readContinueTimeout does not work with cached connections
## Environment Details
* Helidon Version: 4.x
* Helidon SE

----------

## Problem Description

The idle thread starts the read before `SO_TIMEOUT` is set to the [continue timeout](https://github.com/helidon-io/helidon/blob/dec95f3c6d40de9da339b192413eda069fa9c9b0/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java#L393).
Thus, the timeout for the `100 Continue` response is the read timeout (defaults to 30 seconds).

I.e. If the server does not handle `Expect: 100-continue`, the request takes a long time on cached connections.

---

The client logic that [handles](https://github.com/helidon-io/helidon/blob/dec95f3c6d40de9da339b192413eda069fa9c9b0/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java#L399) socket read timeout does not work with cached connections.
The idle task completion [wraps](https://github.com/helidon-io/helidon/blob/dec95f3c6d40de9da339b192413eda069fa9c9b0/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java#L126) all exceptions with a RuntimeException which bubbles up and closes the connection.

I.e. If the server does not handle `Expect: 100-continue`, every other request takes a very long time.
The 2nd request times-out on the read timeout, and then fail.

---

I did some experimentation to "cancel" the idle thread.

It does seem to fix the use-case of a server with bad support for `Expect: 100-continue`.

```java
private void handle() {
    while (!canceled) {
        try {
            socket.setSoTimeout(500);
            next = upstream.read();
            if (next <= 0) {
                closed = true;
            }
            // restore SO_TIMEOUT
            socket.setSoTimeout((int) readTimeout.toMillis());
            return;
        } catch(SocketTimeoutException ignored) {
        } catch (IOException e) {
            closed = true;
            throw new UncheckedIOException(e);
        }
    }
}
```

## Steps to reproduce

See #9731

Verdent:
### Description

Fixes #9736
Fixes https://github.com/helidon-io/helidon/issues/9731

This issue introduces a new way of idle connection handling. It splits the desired idle timeout and uses small 500 ms timeouts until the overall desired timeout is reached. This helps to interrupt the idle timeout mechanism.

FYI: I will add some more comments to the code

### Documentation

None

### DIFF
--- a/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
+++ b/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,8 @@ package io.helidon.common.socket;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
+import java.net.Socket;
+import java.net.SocketTimeoutException;
 import java.util.Objects;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -36,13 +38,24 @@ import io.helidon.common.LazyValue;
  */
 class IdleInputStream extends InputStream {
 
+    /**
+     * This needs to cooperate with 100-continue timeout.
+     * Some low number is needed here for a quick iterations.
+     * Using these small iterations allows us to cancel the idle check much faster,
+     * without the need to wait for a full socket read timeout.
+     */
+    private static final int ITERATION_TIME_MILLIS = 101;
+
+    private final Socket socket;
     private final InputStream upstream;
     private final LazyValue<ExecutorService> executor;
     private volatile int next = -1;
     private volatile boolean closed = false;
+    private volatile boolean cancelled = false;
     private Future<?> idlingThread;
 
-    IdleInputStream(InputStream upstream, String childSocketId, String socketId) {
+    IdleInputStream(Socket socket, InputStream upstream, String childSocketId, String socketId) {
+        this.socket = socket;
         this.upstream = upstream;
         executor = LazyValue.create(() -> Executors.newThreadPerTaskExecutor(
                 Thread.ofVirtual()
@@ -107,10 +120,34 @@ class IdleInputStream extends InputStream {
 
     private void handle() {
         try {
-            next = upstream.read();
-            if (next <= 0) {
-                closed = true;
+            //Currently configured socket read timeout. This is intended to be restored after this method finishes.
+            int toRestore = socket.getSoTimeout();
+            int idleTimeoutIterations = Math.ceilDiv(Math.max(1, toRestore), ITERATION_TIME_MILLIS);
+            for (int i = 0; !cancelled; i++) {
+                try {
+                    //We need to check the current socket timeout before each iteration.
+                    //This time out could have changed,
+                    //and now it would represent the new timeout we should restore after this method ends.
+                    int currentSoTimeout = socket.getSoTimeout();
+                    if (currentSoTimeout != ITERATION_TIME_MILLIS) {
+                        toRestore = currentSoTimeout;
+                    }
+                    //Set iteration read timeout
+                    socket.setSoTimeout(ITERATION_TIME_MILLIS);
+                    next = upstream.read();
+                    if (next <= 0) {
+                        closed = true;
+                        return;
+                    }
+                    break;
+                } catch (SocketTimeoutException e) {
+                    if (i + 1 >= idleTimeoutIterations) {
+                        throw e;
+                    }
+                }
             }
+            //Idle checking thread was canceled or detected as not idle. Restore socket timeout it should have.
+            socket.setSoTimeout(toRestore);
         } catch (IOException e) {
             closed = true;
             throw new UncheckedIOException(e);
@@ -119,8 +156,10 @@ class IdleInputStream extends InputStream {
 
     private void endIdle() {
         try {
+            cancelled = true;
             idlingThread.get();
             idlingThread = null;
+            cancelled = false;
         } catch (InterruptedException | ExecutionException e) {
             closed = true;
             throw new RuntimeException("Exception in socket monitor thread.", e);
--- a/webclient/api/src/main/java/io/helidon/webclient/api/ClientConnection.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/ClientConnection.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
```diff
@@ -36,14 +40,23 @@
...
     private volatile boolean closed = false;
+    private volatile boolean cancelled = false;
     private Future<?> idlingThread;
 
-    IdleInputStream(InputStream upstream, String childSocketId, String socketId) {
+    IdleInputStream(Socket socket, InputStream upstream, String childSocketId, String socketId, Duration idleTimout) {
+        this.socket = socket;
         this.upstream = upstream;
+        this.idleTimeoutIterations = (int) (idleTimout.toMillis() / TIMEOUT);
+        this.lastIterationRemainder = (int) (idleTimout.toMillis() % TIMEOUT);
```
spericas, line 59: After looking at the final version of this, I think there's no need for the "remainder" logic TBH. Timeouts are approximations (especially at the socket layer), so it should be enough to divide by TIMEOUT or maybe add 1 to that. No need for finer control than that. That should simplify your logic a bit, making it easier to read.

romain-grecourt: +1 the loop should just increment the max iterations

romain-grecourt: Something like this maybe ?

```java
try {
    int timeout = socket.getSoTimeout();
    for (int i = 0; i < maxIterations && !cancelled; i++) {
        try {
            int t = socket.getSoTimeout();
            if (t != 500) {
                timeout = t;
            }
            socket.setSoTimeout(500);
            next = upstream.read();
            if (next <= 0) {
                closed = true;
            }
            break;
        } catch (SocketTimeoutException ignored) {
        }
    }
    socket.setSoTimeout(timeout);
} catch (IOException e) {
    closed = true;
    throw new UncheckedIOException(e);
}
```

Verdent: Alright, I have simplified the logic.

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
```diff
@@ -107,20 +122,46 @@ boolean isClosed() {
...
     private void handle() {
         try {
-            next = upstream.read();
-            if (next <= 0) {
-                closed = true;
+            //Currently configured socket read timeout. This is intended to be restored after this method finishes
+            int toRestore = -1;
+            for (int i = 0; !cancelled; i++) {
+                try {
+                    toRestore = timeoutToRestore(toRestore);
```
romain-grecourt, line 129: No need for a method, having the if statement here makes it more readable.

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
```diff
@@ -107,20 +122,46 @@ boolean isClosed() {
...
+                    socket.setSoTimeout(TIMEOUT);
+                    next = upstream.read();
+                    if (next <= 0) {
+                        closed = true;
+                        return;
+                    }
+                    break;
+                } catch (SocketTimeoutException e) {
+                    if (i + 1 >= idleTimeoutIterations) {
```
romain-grecourt, line 138: Do we need to throw the last socket timeout ?
Ins't it better to ignore it them and let the consuming code do another read ?

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
```diff
@@ -36,14 +40,25 @@
  */
 class IdleInputStream extends InputStream {
 
+    /**
+     * This needs to cooperate with 100-continue timeout.
+     * Some low number is needed here for a quick iterations.
+     */
+    private static final int TIMEOUT = 100;
```
romain-grecourt, line 47: Should it be configurable ?

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
Verdent: I am trying to have the same functionality as we had before and we were throwing the exception there.

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
Verdent: Hm... I see this as a cleaner option :-)

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
Verdent: I think it does not necessarily need to be. Or do you?

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
romain-grecourt: How is this better, it's 3 lines of code, it's used only once and does not mutate any state.

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
romain-grecourt: The value was 500, it's now 100 and it's a value we picked somewhat arbitrarily.
So why not make it configureable from the start.

tomas-langer: The 100 was chosen on my request. We need to have a reasonable small time, yet something that will not burn too much CPU.
The default timeout for 100 continue is 1 second, so we expect to spend max. 1.1 second waiting for it. Each higher number means more "error" introduced.

Of course if somebody sets a really small timeout for 100 continue, they would get up to 100 millis added on the first request (further request will skip 100 continue). I think this is a reasonable behavior for now.

Also configuration is not available in sockets, and this would cause a ripple effect in all APIs that use them. If we need to add configurability, we  would also need to add an object to send it around.

tomas-langer (dismissed):

tomas-langer (dismissed):

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
spericas: I agree with Romain, that should be inlined. Also assigning to a param is generally frowned upon.

webclient/api/src/main/java/io/helidon/webclient/api/ClientConnection.java
```diff
@@ -61,4 +61,18 @@ public interface ClientConnection extends ReleasableResource {
      * @param readTimeout connection read timeout
      */
     void readTimeout(Duration readTimeout);
+
+    /**
+     * Check whether this connection is allowed to send 100-Continue.
+     *
+     * @return whether 100-Continue is allowed
+     */
+    boolean allowExpectContinue();
```
romain-grecourt, line 70: I very much dislike the http specific method on the connection interface, although I understand its purpose.
This could be abstracted away as state with a word (E.g a long) and some mask constant.

Also since this is going to be truly implemented in only one implementation, we should probably use default method. These are public interfaces after all so that would preserve backward compatibility.

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
Verdent: You mean to use ternary operator? What is so bothering you guys in having this method?

Yeah, that parameter assignment was a mistake. I did not realize it is a parameter. It is fixed in the latest PR already.

romain-grecourt: Because it's just one if statement with an assignment, that shouldn't be in a method.

```java
int t = socket.getSoTimeout();
if (t != ITERATION_TIME_MILLIS) {
    toRestore = t;
}
```

Verdent: If this method bothers you so much, I will remove it.

Verdent: Alright, I will remove it, if it helps to get this merged.

tomas-langer (approved):

tomas-langer (approved):

romain-grecourt (approved):

PR #3232, opened 2021-07-28
BEP refactor ver2

danielkec:
Fixes #3216 and #3129

This PR reverts revert of #3130 and addresses issue with racing subscribe drain against complete.

PR #9803, opened 2025-02-14
4.x: Remove unnecessary check in BiConsumerChain

Captain1653:
### Description
There is duplicate check in `BiConsumerChain`. It was added [in this PR](https://github.com/helidon-io/helidon/pull/3232/files#diff-6084cd75144d370760d33a084cbc9ad084b77a526b3018a0d28f9f905372566bR53).

This check is always `false`, because of the early return above. Or there should be  the different check?
Screenshot 2025-02-14 at 18 42 17


danielkec (approved):
Huh :smile:

PR #9807, opened 2025-02-16
4.x: Clarify docs about built-in healthchecks and Graal VM (#9021)

### LINKED ISSUE #9021: 4.x: Clarify docs about DeadlockHealthCheck and Graal VM
## Environment Details
* Helidon Version: 4.x
* Helidon SE and Helidon MP
* JDK version: not applicable
* OS: not applicable
* Docker version (if applicable): not applicable

----------

## Problem Description
Helidon has [built-in health checks](https://helidon.io/docs/v4/se/health) .

There is a problem with `DeadlockHealthCheck` and Graal VM. This health check doesn't work for Graal VM. It was partially fixed [in this PR](https://github.com/helidon-io/helidon/commit/6a4f7c1a3aa090c0d081d5b51c8732c1aabbccdc#diff-034bb1da76df530d990e8e0ab86dd44aa760964e6e74c8f999b5ce978eddd2c3R73) and now for Graal VM it always returns `UP` and writes warnings, but there are no corresponding changes in the docs. I tried to find it, but I couldn't.

Is it worth to mention it somewhere?  Maybe create an additional column for this table? Something like `Graal VM Support` and values "yes"/"no"?
Screenshot 2024-07-20 at 17 41 51

Captain1653:
### Description
Fixes #9021 

I've changed docs according to the code:
Screenshot 2025-02-16 at 22 29 08

It looks now (in IDE preview):
Screenshot 2025-02-16 at 22 30 59

If it's wrong don't hesitate to close this PR :) . It's only my guess, how it can be done.

### DIFF
--- a/docs/src/main/asciidoc/se/health.adoc
+++ b/docs/src/main/asciidoc/se/health.adoc
@@ -272,18 +272,17 @@ common health check statuses:
 // Had to move the anchor to the heading above because the rendered page did not define the ID
 // correctly on the table so the link did not work. The link itself looks OK; just no ID generated on the page.
 //[[built-in-health-checks-table]]
-[cols="1,1,3,15,3,3"]
+[cols="1,1,3,15,3"]
 |=======
-|Built-in health check |Health check name |JavaDoc |Config properties (within `server.features.observe.observers.health`) |Default config value |Graal VM Support
+|Built-in health check |Health check name |JavaDoc |Config properties (within `server.features.observe.observers.health`) |Default config value
 
-|deadlock detection
+|deadlock detection &dagger;
 |`deadlock`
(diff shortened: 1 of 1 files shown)


Captain1653:
Maybe add some notes: "* Deadlock health check always returns UP fro Graal VM"?

tjquinno:
Nice update. 

Because (AFAIK) most users do not use GraalVM native image, I think it's less disruptive to readers of the doc page to avoid dedicating a new entire column to the yes/no "is the health check supported in GraalVM native image" information and instead use a footnote-style symbol by the health checks that are not supported and have the footnote just after the table.

I pushed that change to the branch and enabled the workflow.

barchetta (approved):

PR #9809, opened 2025-02-18
4.x: improvement of generated config docs from blueprints

tomas-langer:
Removed `return` javadocs from blueprint config documentation, as it duplicates information.

Fixes for lc4j integration to remove the `get` verb from javadoc of getters.

Previously (see current docs, example: https://helidon.io/docs/v4/config/io_helidon_http_media_MediaContext):
```
Media supports to use. 
This instance has priority over provider(s) discovered by service loader. 
The providers are used in order of calling this method, where the first support added is the
first one to be queried for readers and writers.

Media supports
```

After change:
```
Media supports to use.
This instance has priority over provider(s) discovered by service loader.
The providers are used in order of calling this method, where the first support added is the
first one to be queried for readers and writers.
```


integrations/langchain4j/providers/ollama/src/main/java/io/helidon/integrations/langchain4j/providers/ollama/OllamaLanguageModelConfigBlueprint.java
```diff
@@ -68,7 +68,7 @@ interface OllamaLanguageModelConfigBlueprint extends OllamaCommonConfig {
     Optional<Double> topP();
 
     /**
-     * Gets the penalty applied to repeated tokens during text generation.
+     * T penalty applied to repeated tokens during text generation.
```
tomas-langer, line 71: Fixing incorrect search & replace

danielkec (approved):

Issue #9698, opened 2025-01-27
Helidon Rest API error response contains reflected user input


aravind643:
Helidon Version: 4.1.4
Helidon SE or Helidon MP : Helidon MP
JDK version: jdk 21

---

When an Rest endpoint is invoked with invalid chars in the query params, the user input is reflected back in the response. This is not adhering to the security standards we have. Here is the description from the testing team.

---

#### Description
Input being returned in application responses is a prerequisite for many client-side vulnerabilities, including cross-site scripting, open redirection, content spoofing, and response header injection. Additionally, some server-side vulnerabilities such as SQL injection are often easier to identify and exploit when input is returned in responses.

#### Remediation

Do not reflect user input in the error response while generation error message. The error message should be generic.

---

Example: (In the URL, the query param contains invalid chars like "[", "]", ":")

```
Request:
GET /api/data?Class.classLoader.URLs[0]=testClassloaderManipulation1737604676.https://uigc90j45y1dlojlfkl4cipk8be22sqh.oastify.com HTTP/1.1
Authorization: Bearer [...]
User-Agent: PostmanRuntime/7.37.3
Accept: */*
Cache-Control: no-cache
Postman-Token: 6769d834-5d41-4758-8713-ec85575cabf4
Host: localhost
Accept-Encoding: gzip, deflate, br
Connection: keep-alive
```

Response:

```
HTTP/1.1 400 Bad Request
Date: Thu, 23 Jan 2025 04:19:26 GMT
Content-Type: text/plain
Content-Length: 177
Connection: keep-alive
Strict-Transport-Security: max-age=31536000; includeSubDomains
```

```
Query contains invalid char: 
Class.classLoader.URLs[0]=testClassloaderManipulation1737604676.https://uigc90j45y1dlojlfkl4cipk8be22sqh.oastify.com/, index: 22, char: &#x27;[&#x27;
```
---

I'm unable to intercept this request either by implementing ExceptionMapper or ContainerResponseFilter or ContainerRequestFilter to modify the response to contain generic message (without user input)

tomas-langer:
Changing messages of bad requests cannot be done using the usual error handlers, as the request never reaches routing (it must not reach routing, as it is invalid).
We have a tool in Helidon WebServer that allows modification of pre-routing errors. See `io.helidon.webserver.http.DirectHandlers` that can be registered on web server or a listener (this could be done even in MP by creating an extension and modifying the server builder before the server gets started by `ServerCdiExtension`)

[...]

PR #9811, opened 2025-02-18
Config to prevent reflection of user input when reporting errors

spericas:
### Description

New config section in listeners for error handling. Default settings prevent any entity from being returned to avoid reflecting any data from a request. Default settings can be modified to return safe messages and to log all messages. Issue #9698.

### Documentation

None

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
@@ -405,5 +405,6 @@ interface ListenerConfigBlueprint {
      *
      * @return optional error handling
      */
+    @Option.Configured
     Optional<ErrorHandling> errorHandling();
 }
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
@@ -400,4 +400,10 @@ interface ListenerConfigBlueprint {
         }
     }
 
+    /**
+     * Configuration for this listener's error handling.
+     *
+     * @return optional error handling
+     */
+    Optional<ErrorHandling> errorHandling();
 }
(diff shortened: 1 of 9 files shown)


tomas-langer (commented):
We are reflecting back a cleaned user input (it should not contain the actual illegal character, unless it is an OK entity to return - i.e. if `[` is forbidden, we can still return it, as it is not an HTML entity, so it does not matter
Even if we had a bug that returns illegal characters, the fix is to remove the illegal character from the returned string (it should be replaced with some other character, to keep the indexes - this should be already implemented).

spericas:
> We are reflecting back a cleaned user input (it should not contain the actual illegal character, unless it is an OK entity to return - i.e. if `[` is forbidden, we can still return it, as it is not an HTML entity, so it does not matter Even if we had a bug that returns illegal characters, the fix is to remove the illegal character from the returned string (it should be replaced with some other character, to keep the indexes - this should be already implemented).

Yes, that's the current behavior, but I still think it is unnecessary to return back all those characters as we are doing now. Pointing out the invalid char in the URI should be sufficient.

tomas-langer (changes requested):
I do not think that this is a bug - I have commented in detail on the original issue.

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
```diff
@@ -400,4 +400,11 @@ default void configureSocket(ServerSocket socket) {
         }
     }
 
+    /**
+     * Configuration for this listener's error handling.
+     *
+     * @return optional error handling
+     */
+    @Option.Configured
+    Optional<ErrorHandling> errorHandling();
```
tomas-langer, line 409: This could probably return `ErrorHandling` instead of `Optional<ErrorHandling>` using `@io.helidon.builder.api.Option.DefaultMethod("create")`, as both options have sensible defaults.
This would make usage a bit nicer.

tomas-langer (changes requested):

Issue #9704, opened 2025-01-27
Add support for Eureka service registration


barchetta:
This issue attempts to capture a problem description and a few hard-and-fast requirements of work around Eureka _service registration_. The following is an initial draft of the shape of the work, and is subject to edit and change at any time:

### Who
A developer of a Helidon-based microservice

### Wants To…
Have her microservice's coordinates (host, port, etc.) appear in a preëxisting (probably Spring-based/flavored) Eureka server somewhere such that developers of other microservices, Helidon-based or otherwise (probably Spring Boot-based), can discover it from that server, using Eureka-flavored facilities they are familiar with (probably Spring Cloud's `DiscoveryClient` machinery)

### Goals of the Solution
* **Configuration only.** The Helidon developer should ideally have to add only a minimum of configuration to her microservice, and perform no other work, to accomplish her goal
* **Unobtrusive.** In general, if some feature or facility required by the solution is not present, the solution should log a warning and take no action (rather than causing startup to fail, for example).
* **Late-bound.** If for whatever reason the web server never starts, for example, the solution should ideally not have performed any work at all. If the web server starts up successfully and the solution is notified that it should attempt service registration, it should acquire the information it needs at this stage (as late as possible) and not before. There may be exceptions.
* **Spring-friendly.** The primary beneficiaries of this work will be Spring Boot/Spring Cloud users who will be able to "see" Helidon-based microservices in the Eureka service catalogs they are used to interacting with. It is not an immediate goal to support other kinds of consumers.
* **Familiar configuration.** The only users of Eureka in any capacity (other than Netflix's presumed internal usage of it) are Spring users, and Spring Cloud Netflix makes use of the Eureka-native configuration keys. It follows that any users of the implementation of this enhancement in Helidon will be, or will have been, Spring users. Therefore, the implementation of this feature should reuse configuration names where possible. (So, for example, `eureka.instance` is an expected and familiar namespace under which Eureka registration configuration (such as `eureka.instance.leaseRenewalIntervalInSeconds`) should be stored.)

### Non-Functional Requirements
* The version of Eureka involved will be 2.0.4 (the version used, and at least partly maintained, by [Spring Cloud Netflix machinery version 4.2.0](https://search.maven.org/artifact/org.springframework.cloud/spring-cloud-starter-netflix-eureka-client/4.2.0/jar))
* No third-party HTTP clients should be used

### Issues
[...]

PR #9820, opened 2025-02-21
Adds afterStart(WebServer) implementations to various delegating ServerLifecycle implementations that were missing it

ljnelson:
PR #9655 added an `afterStart(WebServer)` method to `ServerLifecycle`. It was approved on February 4. There were more `ServerLifecycle` implementations than showed up at first glance. This PR attempts to ensure that all of the known `ServerLifecycle` implementations in the code base do the right thing with their `afterStart(WebServer)` implementations, so that they don't simply (incorrectly) inherit the (necessary) `default` "no-op" implementation.

This PR gates #9704 and #9789.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerFeatureContextImpl.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerFeatureContextImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


tjquinno (commented):
IIRC an app _can_ start multiple web servers (probably very rare, but I think possible?). The `afterStart` method accepts a `WebServer` as a parameter, but the `afterStop` method does not. Presumably some `afterStart` logic could do webserver-dependent things and might need to distinguish among which of the previously-started web servers has been stopped during its `afterStop` processing. Such code has no way of doing so with the current `afterStop` API.

Certainly if at most one webserver can ever be started in a JVM then this is a non-issue.

ljnelson:
> IIRC an app _can_ start multiple web servers (probably very rare, but I think possible?). The `afterStart` method accepts a `WebServer` as a parameter, but the `afterStop` method does not. Presumably some `afterStart` logic could do webserver-dependent things and might need to distinguish among which of the previously-started web servers has been stopped during its `afterStop` processing. Such code has no way of doing so with the current `afterStop` API.
> 
> Certainly if at most one webserver can ever be started in a JVM then this is a non-issue.

All fair points in my opinion. I'm also not particularly crazy about the fact that the "webserver is up!" "event" is buried in, for practical purposes, a routing, when the thing reacting to the event reception may have nothing to do with routing.

The scope of this PR, however, is simply to add (required) implementations of the `afterStart(WebServer)` method where I missed them the first time; they should have been included in #9655.

tjquinno (approved):
LGTM

PR #9846, opened 2025-02-26
4.x: CORS support helper change log level.

tomas-langer:
### Description
Info log level is only for a very limited number of messages that must be agreed on in advance. 

Configuration of CORS is not such a type of message. It may be printed multiple times if used directly by the user, and in general this does not seem to be a relevant information that MUST be printed on each startup.

I have switched the level to `DEBUG`, which is the next level.

This can always be easily enabled by changing the log level of this logger to verify the configuration when/if needed.

### DIFF
--- a/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
+++ b/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


romain-grecourt (approved):

cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
```diff
@@ -239,8 +238,10 @@ public CorsSupportHelper<Q, R> build() {
 
             CorsSupportHelper<Q, R>  result = new CorsSupportHelper<>(this);
 
-            LOGGER.log(System.Logger.Level.INFO,
-                    () -> String.format("CorsSupportHelper configured as: %s", result.toString()));
+            if (LOGGER.isLoggable(System.Logger.Level.DEBUG)) {
```
jbescos, line 241: I though that using a lambda in LOGGER.log avoids to use the `if (LOGGER.isLoggable(System.Logger.Level.DEBUG)) {`. Why is this needed?.

tjquinno (approved):

PR #9861, opened 2025-03-04
4.x Support via Helidon config for controlling OpenTelemetry tracing

### LINKED ISSUE #9595: Add support for OpenTelemetry configuration in Helidon SE
## Environment Details
* Helidon Version: 4.1.5
* Helidon SE 
* JDK version: 21
* OS:

----------

## Problem Description
otel configuration is not currently loaded by Helidon:

```
otel:
  sdk.disabled: false
  service.name: "todo:front"
  traces.exporter: otlp
```

We can work around this for now using the following code:

```Java
    private static Map<String, String> telemetryConfig(Config config) {
        Config otelConfig = config.get("otel");
        Map<String, String> result = new HashMap<>();
        otelConfig.asMap().ifPresent(result::putAll);
        result.putIfAbsent("otel.metrics.exporter", "none");
        result.putIfAbsent("otel.logs.exporter", "none");
        return result;
    }

    private static void  prepareOpenTelemetry(Config config) {
        String serviceName = config.get("otel.service.name").asString().orElse("HELIDON_SE_SERVICE");
        OpenTelemetry openTelemetry = AutoConfiguredOpenTelemetrySdk.builder()
                .addPropertiesCustomizer(x -> telemetryConfig(config))
                .addResourceCustomizer((r, props) -> r.toBuilder()
                        .put(SERVICE_NAME, serviceName)
                        .build())
                .build()
                .getOpenTelemetrySdk();
        if (openTelemetry == null) {
            openTelemetry = OpenTelemetry.noop();
        }
        io.opentelemetry.api.trace.Tracer otelTracer = openTelemetry.getTracer(serviceName);
        Tracer helidonTracer = HelidonOpenTelemetry.create(openTelemetry, otelTracer, Map.of());
        Contexts.globalContext().register(helidonTracer);
    }
```

However, it would be preferable to be able to load these either by config file or by environment variables.

## Steps to reproduce

tjquinno:
### Description

Resolves #9595 

This PR adds support for applying the tracing abstractions in Helidon config to OpenTelemetry, similarly to how Helidon handles config for Jaeger tracing. The OTel-based config also includes some additional OTel-specific settings.

Because the existing tracer builders use the older builder approach (and the builders implement a common interface) the new `OpenTelemetryTracerBuilder` follows the same pattern. Eventually we should convert all the tracer builders to the newer builder/config pattern.

Most of the changes are straightforward, using config settings in preparing the relevant OTel objects. (See the class-level Javadoc on the new `OpenTelemetryTracerBuilder` for details.)

One interesting aspect is that, to preserve existing behavior, the new OTel tracer builder as well as the Helidon OTel code that manages the Helidon global tracer in`OpenTelemetryTracerProvider` check for the OTel system property or env. var. that enables OTel auto-configuration. (This OTel feature allows developers or deployers to control OTel's behavior using properties and env. vars rather than programmatically in code.) Existing Helidon users relying on OTel might be depending on this, so the new OTel tracer builder is _not_ used if the user has enabled OTel auto-config. The `pom.xml` file runs most of the tests twice, once with auto-config (to replicate past behavior) and once without auto-config.

The PR changes some tests slightly to avoid problems related to attempts to assign OTel global objects multiple times in the same test JVM. 

An upcoming push to this PR will update how the telemetry CDI producer works. It currently relies on OTel auto-configuration and it will be revised to use the builder instead. 

### Documentation
Doc update will be in a later push to this PR.

### DIFF
--- a/docs/src/main/asciidoc/includes/attributes.adoc
+++ b/docs/src/main/asciidoc/includes/attributes.adoc
@@ -235,7 +235,9 @@ endif::[]
 :service-registry-base-url: {javadoc-base-url}/io.helidon.service.registry
 :telemetry-javadoc-base-url: {javadoc-base-url}/io.helidon.microprofile.telemetry
 :tracing-javadoc-base-url: {javadoc-base-url}/io.helidon.tracing
+:tracing-jaeger-provider-javadoc-base-url: {javadoc-base-url}/io.helidon.tracing.providers.jaeger
 :tracing-otel-provider-javadoc-base-url: {javadoc-base-url}/io.helidon.tracing.providers.opentelemetry
+:tracing-zipkin-provider-javadoc-base-url: {javadoc-base-url}/io.helidon.tracing.providers.zipkin
 :types-javadoc-base-url: {javadoc-base-url}/io.helidon.common.types
 
 :webclient-javadoc-base-url: {javadoc-base-url}/io.helidon.webclient
--- a/docs/src/main/asciidoc/includes/tracing/common-callbacks.adoc
+++ b/docs/src/main/asciidoc/includes/tracing/common-callbacks.adoc
@@ -116,9 +116,11 @@ The following tables list specifically what operations the proxies permit.
 
 |====
 
-=== Adding a Listener
+=== Adding and Removing a Listener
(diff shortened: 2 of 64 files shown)


tjquinno:
This PR will be superseded by a more comprehensive one.

PR #9865, opened 2025-03-04
4.x: Update all module to be compilable and testable with Java 24.

### LINKED ISSUE #9860: 4.x: Helidon build fails with Java 24
There are two problems:

1. Incorrect processing of error types in `AptTypeFactory`, where an inner type without a package does not resolve into a proper class name
2. Micronaut Data integration fails at test compile, due to a problem with one of the Micronaut annotation processors (under investigation, will reference issue)

As a result the repo cannot be built using Java 24 (using build 36 to validate).

tomas-langer:
Resolves #9860 

Updates that allow us to build and test using Java 24

Some tests are disabled in Java 24 (where we need to handle external libraries), some are updated to have a different behavior on Java 21 and 24.

A few changes were done to allow work on both versions.

### DIFF
--- a/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/HeadersServerTest.java
+++ b/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/HeadersServerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/HeadersServerTest.java
```diff
@@ -203,20 +203,32 @@ void serverInbound(WebServer server) throws IOException, InterruptedException {
...
+            }
+
+            // There is no way how to access GO_AWAY status code and additional data with JDK Http client
+            try {
+                HttpResponse<String> response = client.send(req.uri(base.resolve("/cont-in")).build(),
+                                                            HttpResponse.BodyHandlers.ofString());
+
+                int status = response.statusCode();
+                // since Java 24, we get a 400 back and not an exception
+                assertThat("IOException or status 400 was expected, but go status " + response.statusCode()
```
danielkec, line 223:
```suggestion
                assertThat("IOException or status 400 was expected, but got status " + response.statusCode()
```

danielkec (changes requested):

danielkec (approved):

barchetta (approved):

Issue #9506, opened 2024-11-19
4.x: MpConfigSources.create does not handle empty values


romain-grecourt:
## Environment Details
* Helidon Version: 4.1.4
* Helidon MP

----------

## Problem Description

`io.helidon.config.mp.MpConfigSources.create` does not handle empty values (I.e. `""`), which then results in `java.util.NoSuchElementException` because the MicroProfile config specification does not support empty values.

---

The following code detaches SE config as a map https://github.com/helidon-io/helidon/blob/f57be59aa42ee5ba1f12e85ce8ed43f069205d17/microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/AdHocConfigBuilder.java#L48

And then builds a MicroProfile Config from the map: https://github.com/helidon-io/helidon/blob/f57be59aa42ee5ba1f12e85ce8ed43f069205d17/microprofile/messaging/core/src/main/java/io/helidon/microprofile/messaging/AdHocConfigBuilder.java#L55

This makes it impossible to supply the following configuration:

```yaml
mp.messaging:
  connector:
      helidon-kafka:
        ssl.endpoint.identification.algorithm: ${EMPTY}
```

```
Exception in thread "main" java.util.NoSuchElementException: Property "ssl.endpoint.identification.algorithm" is not available in configuration
	at io.helidon.config.mp.MpConfigImpl.lambda$getValue$2(MpConfigImpl.java:127)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at io.helidon.config.mp.MpConfigImpl.getValue(MpConfigImpl.java:127)
	at io.helidon.config.mp.SeConfig.asMap(SeConfig.java:262)
	at io.helidon.messaging.connectors.kafka.KafkaConfig.create(KafkaConfig.java:63)
	at io.helidon.messaging.connectors.kafka.KafkaPublisher$Builder.config(KafkaPublisher.java:373)
	at io.helidon.messaging.connectors.kafka.KafkaConnector.getPublisherBuilder(KafkaConnector.java:197)
	at io.helidon.messaging.connectors.kafka.KafkaConnector$Proxy$_$$_WeldClientProxy.getPublisherBuilder(Unknown Source)
	at io.helidon.microprofile.messaging.OutgoingConnector.lambda$getPublisher$0(OutgoingConnector.java:74)
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1228)
[...]

PR #9890, opened 2025-03-12
3.x: Fix empty config values propagation (#9888)

### LINKED ISSUE #9888: 3.x: MpConfigSources.create does not handle empty values
Backport of https://github.com/helidon-io/helidon/issues/9506 for 3.x

Captain1653:
### Description
Fixes #9888 

The same [was done for 4.x](https://github.com/helidon-io/helidon/issues/9506)

### DIFF
--- a/microprofile/messaging/core/src/test/java/io/helidon/microprofile/messaging/connector/ConnectorConfigTest.java
+++ b/microprofile/messaging/core/src/test/java/io/helidon/microprofile/messaging/connector/ConnectorConfigTest.java
@@ -34,7 +34,6 @@ import org.eclipse.microprofile.reactive.messaging.spi.IncomingConnectorFactory;
 import org.eclipse.microprofile.reactive.streams.operators.PublisherBuilder;
 import org.eclipse.microprofile.reactive.streams.operators.ReactiveStreams;
 import org.hamcrest.Matchers;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


microprofile/messaging/core/src/test/java/io/helidon/microprofile/messaging/connector/ConnectorConfigTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Disabled("because @Inject and @HelidonTest(resetPerTest=true) throws exception")
```
Captain1653, line 46: This is a little hard  for me :) . Let me explain.

The same tests were added for 4.x and it works there, but for 3.x I have this exception
Screenshot 2025-03-12 at 09 46 34

The error message is clear. I can't inject, because of this parameter `resetPerTest=true`. It works for 4.x after [this refactoring](https://github.com/helidon-io/helidon/commit/1baabd5042389661b75238c774c38aceb132bd6b#diff-5dea65dfdf039d5e5a05021730d568db7512b0734e5580b1cc2239432b766267L156) .

But how can I fix this test for 3.x? Or should I remove these tests at all, because `AdHocConfigBuilderTest` does all necessary checks?

romain-grecourt: The test class only has one method so you don't need to use `resetPerTest = true`.
You can move the `@AddConfig` and `@AddBean` annotations on the class.

Captain1653: Thank you

romain-grecourt (approved):

Issue #9859, opened 2025-03-04
Allow helidon-jersey-connector configuration through the Helidon server config


brian-maulding:
## Environment Details
* Helidon Version: 4.1.6

----------

## Enhancement Description
HelidonConnector only allows you to provide configuration through `ClientConfig.property(HelidonProperties.CONFIG, config)`.  `Config.empty()` is used when the property is not set.  The Helidon server config should be used by default (when available) to allow for configuration without requiring code changes.

A Helidon application can use Jersey clients in many places.  All Jersey client instances will use HelidonConnector if the application adds a dependency to helidon-jersey-connector because it is loaded through the Service Provider Interface.  Some usages can come from third party libraries (for example, SDKs) where the application does not have any control over the code.  Therefore, it is not possible to configure the HelidonConnector for these client instances.  Using the Helidon server config by default would allow tuning HelidonConnector for all client instances.

As an example, we encountered performance problems for calls made through an SDK that could be resolved by setting `socket-options.tcp-no-delay=true`.  However, we cannot change the SDK code so we either live with the performance problems or stop using HelidonConnector everywhere.

romain-grecourt:
see also https://github.com/helidon-io/helidon/issues/7349

PR #9891, opened 2025-03-12
Connector configuration improvements

spericas:
### Description

Connector configuration improvements. Adds support to configure connector from global config (registry). Fixes a few edge cases in property mapping. Some new tests. Issue #9859.

### Documentation

Included in this PR.

### DIFF
--- a/jersey/connector/src/main/java/io/helidon/jersey/connector/HelidonConnector.java
+++ b/jersey/connector/src/main/java/io/helidon/jersey/connector/HelidonConnector.java
@@ -64,7 +64,7 @@ import static org.glassfish.jersey.client.ClientProperties.getValue;
 
 class HelidonConnector implements Connector {
     static final System.Logger LOGGER = System.getLogger(HelidonConnector.class.getName());
-    private static final String CONNECTOR_CONFIG_ROOT = "connector";
+    private static final String CONNECTOR_CONFIG_ROOT = "jersey.connector.webclient";
 
     private static final String HELIDON_VERSION = "Helidon/" + Version.VERSION + " (java "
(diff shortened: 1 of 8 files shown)


jersey/connector/src/main/java/io/helidon/jersey/connector/HelidonConnector.java
```diff
@@ -78,42 +77,60 @@ class HelidonConnector implements Connector {
...
-            builder.connectTimeout(Duration.ofMillis(getValue(properties, CONNECT_TIMEOUT, DEFAULT_TIMEOUT)));
+            Integer connectTimeout = getValue(properties, CONNECT_TIMEOUT, Integer.class);
+            if (connectTimeout != null) {
+                builder.connectTimeout(Duration.ofMillis(connectTimeout));
+            }
         }
         if (properties.containsKey(READ_TIMEOUT)) {
-            builder.readTimeout(Duration.ofMillis(getValue(properties, READ_TIMEOUT, DEFAULT_TIMEOUT)));
+            Integer readTimeout = getValue(properties, READ_TIMEOUT, Integer.class);
+            if (readTimeout != null) {
```
ljnelson, line 111: (If you're going to always call `getValue` and check to see whether it worked by checking for `null`, then you can eliminate the `containsKey` call.)

spericas: If it's there it can still return `null` if `getValue` is unable to convert the value to the type requested. In fact, it should at least log that condition and it isn't yet.

ljnelson: OK!

Verdent (approved):

PR #9911, opened 2025-03-16
3.x: Fix deadlock health check if invoking the ThreadMXBean fails (#9910)

### LINKED ISSUE #9910: 3.x Fix Deadlock health check  when it cannot access the JMX MBean to find deadlocks
Backport of https://github.com/helidon-io/helidon/issues/9691 for 3.x

Captain1653:
### Description
Fixes #9910 

The similar logic [was done for 4.x](https://github.com/helidon-io/helidon/pull/9694). Is it better  to be consistent between 3.x and 4.x ?

Unfortunately, we can't return the status `ERROR`, because for 3.x exists only statuses `UP` and `DOWN`.  But we can throw `HealthCheckException`. I hope, that it's appropriate.  Return `UP` when app can't get `ThreadMXBean` is too optimistic.

P.S. `DiskSpaceHealthCheck` in 3.x does the similar thing:
```
try {
      this.fileStore = Files.getFileStore(builder.path);
 } catch (IOException e) {
      throw new HealthCheckException("Failed to obtain file store for path " + builder.path.toAbsolutePath(), e);
 }
```

### DIFF
--- a/health/health-checks/src/main/java/io/helidon/health/checks/DeadlockHealthCheck.java
+++ b/health/health-checks/src/main/java/io/helidon/health/checks/DeadlockHealthCheck.java
@@ -94,8 +94,6 @@ public class DeadlockHealthCheck implements HealthCheck {
                 LOGGER.log(Level.FINEST, "Health check observed deadlocked threads: " + Arrays.toString(deadlockedThreads));
             }
         } catch (Throwable e) {
-            // Throw HealthCheckException, not report DOWN, because we do not know that
-            // there are deadlocks which DOWN should imply; we simply cannot find out.
             throw new HealthCheckException("Error invoking ThreadMXBean to find deadlocks; cannot complete this healthcheck", e);
         }
         return builder.build();
--- a/health/health-checks/src/main/java/io/helidon/health/checks/DeadlockHealthCheck.java
+++ b/health/health-checks/src/main/java/io/helidon/health/checks/DeadlockHealthCheck.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


health/health-checks/src/main/java/io/helidon/health/checks/DeadlockHealthCheck.java
```diff
@@ -80,17 +83,20 @@ public HealthCheckResponse call() {
...
-        boolean noDeadLock = false;
+        HealthCheckResponseBuilder builder = HealthCheckResponse.builder()
+                .name(NAME);
+
         try {
             // Thanks to https://stackoverflow.com/questions/1102359/programmatic-deadlock-detection-in-java#1102410
-            noDeadLock = (threadBean.findDeadlockedThreads() == null);
+            long[] deadlockedThreads = threadBean.findDeadlockedThreads();
+            if (deadlockedThreads != null) {
+                builder.down();
```
tjquinno, line 93: Returning `UP` is too optimistic, but returning `DOWN` is too pessimistic. We do not know that there is a deadlock (which is what `DOWN` should indicate); we cannot tell one way or the other.

Given there is no `ERROR` state available in the release, the code in this check should continue to throw an exception. 

The Helidon code which invokes this `call` method already catches exceptions and sets the "internalServerError` flag in the `HcResponse`. When the code in `HealthSupport#callHealthChecks` computes the overall status it should _ignore_ any `HcResponse` for which `internalServerError` has been set.

tjquinno (changes requested):
Please see the requested change.

health/health-checks/src/main/java/io/helidon/health/checks/DeadlockHealthCheck.java
Captain1653: Thank you for your explanation.

It's done. I'm not sure :) 

One small question for my curiosity:
If there is one internal error in response, then httpStatus=5xx and overallStatus = UP? Right?

tjquinno: If there is _at least_ one internal error, then Helidon should return `500`. I guess the state in the response details might as well be `UP` (that's what happens in the updated 4.x code) but, ideally, K8s or whatever is monitoring the health endpoint will disregard the body of the response due to the HTTP status being 500.

health/health/src/main/java/io/helidon/health/HealthSupport.java
```diff
@@ -212,6 +212,7 @@ HealthResponse callHealthChecks(List<HealthCheck> healthChecks) {
                 .collect(Collectors.toList());
 
         Status status = responses.stream()
+                .filter(this::internalErrorExcluded)
```
tjquinno, line 215: IMO it would be cleaner (less code added) to use
```java
.filter(Predicate.not(HcResponse::internalError))
```
here and remove the new  `internalErrorExcluded` method.

tjquinno (changes requested):
One change requested.

tjquinno (approved):

Captain1653:
@tjquinno Can it be merged? Please :)

PR #9922, opened 2025-03-20
4.x: OIDC Certificate bound access token

### LINKED ISSUE #9745: cross realm communication - mTLS
## Environment Details
* Helidon Version: 4.1.2

----------

## Enhancement Description
Our services are developed using Helidon 4.1.2 and hosted in a tenancy in OC1 realm , customer's IDCS tenancy is in OC2 realm . 
we need to support cross realm communication using mTLS from our services (OC1 realm) to IDCS(OC2 realm) . please evaluate if this could be supported using helidon role mapper lib  . Thanks!

Verdent:
### Description
Fixes #9745 

This PR adds:
- Full webclient configuration to the OIDC
- Certificate bound access token support

### Documentation

None

### DIFF
--- a/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
+++ b/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
@@ -1060,11 +1060,10 @@ public final class OidcConfig extends TenantConfigImpl {
          * @param config configuration located on node with OIDC configuration keys (e.g. client-id)
          * @return updated builder instance
          */
-        @ConfiguredOption(type = WebClientConfig.class,
-                          key = "webclient",
+        @ConfiguredOption(key = "webclient",
+                          type = WebClient.class,
                           description = "WebClient configuration used for outbound requests to the identity server. "
-                                  + "This configurations sets the configured values to "
-                                  + "the OIDC WebClient default configuration.")
+                                  + "This configuration sets the values to the OIDC WebClient default configuration.")
         public Builder config(Config config) {
             super.config(config);
             // mandatory configuration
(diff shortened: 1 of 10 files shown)


security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -1087,6 +1123,10 @@ public Builder config(Config config) {
             config.get("tenants").asList(Config.class)
                     .ifPresent(confList -> confList.forEach(tenantConfig -> tenantFromConfig(config, tenantConfig)));
 
+            config.get("webclient-config")
```
tomas-langer, line 1126: `webclient-config` in config seems like a too much information.
Can we just use `webclient`?

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -1184,7 +1224,8 @@ public Builder forceHttpsRedirects(boolean forceHttpsRedirects) {
          * @param relativeUris relative URIs flag
          * @return updated builder instance
          */
-        @ConfiguredOption("false")
+        @ConfiguredOption(value = "false", deprecated = true)
+        @Deprecated(since = "4.2.1", forRemoval = true)
```
tomas-langer, line 1228: Whenever you deprecate something (esp. when `forRemoval=true`), you need to add a javadoc `@deprecated` that explains what to do instead

tomas-langer: Please fix all in thisi class.

tomas-langer (dismissed):

tomas-langer (approved):

PR #9924, opened 2025-03-20
Fix problems in tracing with headers that appear multiple times in an incoming request

### LINKED ISSUE #9923: 4.x Setting `Baggage` header multiple times causes only the last one to be in the "current" baggage for Helidon MP Telemetry and SE tracing with Jaeger or Otel
## Environment Details
* Helidon Version: 4.x
* JDK version:
* OS:

----------

## Problem Description
HTTP allows the same header (in this base, `Baggage`) to be set multiple times. 

But in Helidon's MP Telemetry support only one occurrence is used in setting the current `Baggage` context.

Similarly, using SE only one occurrence is used.

### MP
Unzip, build, and run the attached `MultiBaggage.zip` project. Then access the app using 
```bash
curl -v -H "baggage:k1=v1b" -H "baggage:k2=v2b" localhost:8080/greet
```
The response displays the baggage-related header values seen followed by the baggage values available as the current baggage. The current baggage has only one setting (seemingly the last one).
```list
* Uses proxy env variable no_proxy == '127.0.0.1,localhost,oraclecorp.com,us.oracle.com,sca00kau.us.oracle.com,sca00kau'
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET /greet HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> baggage:k1=v1b
> baggage:k2=v2b
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Date: Thu, 20 Mar 2025 11:17:24 -0500
< Connection: keep-alive
< Content-Length: 62
< Content-Type: application/json
< 
* Connection #0 to host localhost left intact
{"message":"header=baggage=[k1=v1b, k2=v2b]...baggage=k2=v2b"}
```

### SE
Unzip, build, and run the attached `MultiBaggageSE.zip` project. Then access the app using the following command (it's different from the one above):
```bash
curl  -v -H "baggage:k1=v1b" -H "baggage:k2=v2b" localhost:8080/greet/baggage
```

The response does not show the baggage _values_ but it does show that only a single baggage _key_ `k1` is represented in the current baggage in the endpoint code.
```list
* Uses proxy env variable no_proxy == '127.0.0.1,localhost,oraclecorp.com,us.oracle.com,sca00kau.us.oracle.com,sca00kau'
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET /greet/baggage HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
> baggage:k1=v1b
> baggage:k2=v2b
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Date: Thu, 20 Mar 2025 13:33:26 -0500
< Connection: keep-alive
< Content-Length: 127
< Content-Type: application/json
< 
* Connection #0 to host localhost left intact
{"message":"Hello Baggage {k1=io.helidon.tracing.providers.opentelemetry.MutableOpenTelemetryBaggage$HBaggageEntry@489e9823}!"}
```

tjquinno:
### Description
Resolves #9923 

#### Release Note
____
Helidon now correctly handles the case where incoming SE and MP requests contain the same tracing-related heading more than once. Previously the tracing context would be set up using only one of the occurrences of the header.
____

#### PR overview

Helidon uses a Jakarta REST filter to establish the current tracing context from inbound requests in MP and a Helidon filter to do the same for inbound SE requests.

Both of those failed to handle cases in which a tracing-related header itself (baggage in the original case) appears more than once in the request. Both filters would use only one of the occurrences.

The PR addresses this in the SE and the MP filters and adds tests for both.

There were other small fixes added that were revealed by the new tests.

### Documentation
Bug fix - no doc impact.

### DIFF
--- a/microprofile/telemetry/pom.xml
+++ b/microprofile/telemetry/pom.xml
@@ -142,6 +142,20 @@
                         </path>
                     </annotationProcessorPaths>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>default-testCompile</id>
+                        <configuration>
+                            <compilerArgs>
+                                <!--
+                                Used by a test to send requests with repeating header names.
+                                -->
+                                <arg>--add-modules</arg>
+                                <arg>java.net.http</arg>
+                            </compilerArgs>
+                        </configuration>
+                    </execution>
+                </executions>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -278,19 +278,20 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
     private void handleBaggage(ContainerRequestContext containerRequestContext, Context context) {
         List<String> baggageProperties = containerRequestContext.getHeaders().get("baggage");
         if (baggageProperties != null) {
+            var baggageBuilder = Baggage.builder();
             for (String b : baggageProperties) {
                 String[] split = b.split("=");
                 if (split.length == 2) {
                     String[] valueAndMetadata = split[1].split(";");
                     String value = valueAndMetadata.length > 0 ? valueAndMetadata[0] : "";
                     String metadata = valueAndMetadata.length > 1 ? valueAndMetadata[1] : "";
-                    Baggage.builder()
-                            .put(split[0], value, BaggageEntryMetadata.create(metadata))
-                            .build()
-                            .storeInContext(context)
-                            .makeCurrent();
+                    baggageBuilder
+                            .put(split[0], value, BaggageEntryMetadata.create(metadata));
                 }
             }
+            baggageBuilder.build()
+                    .storeInContext(context)
+                    .makeCurrent();
         }
     }
 
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/RequestContextHeaderProvider.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/RequestContextHeaderProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2024, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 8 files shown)


microprofile/telemetry/pom.xml
```diff
@@ -142,6 +142,20 @@
...
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>default-testCompile</id>
+                        <configuration>
+                            <compilerArgs>
+                                <!--
+                                Used by a test to send requests with repeating header names.
+                                -->
+                                <arg>--add-modules</arg>
```
ljnelson, line 153: Just a comment, not a request for a change; it's too minor.

This is odd. I understand why it's here.

Maybe there's a way to run `testCompile` "without modules", in much the same way you can run Surefire "without modules" (via `useModulePath=false`). Otherwise this is a little leaky: there are other `testCompile` executions that could bypass this (I haven't tried it but I'd be willing to bet that something like a direct invocation of `compiler:testCompile` will trigger the `default-cli` execution and bypass this).

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -278,19 +278,20 @@ private String resolveTarget(ContainerRequestContext requestContext) {
...
                     String[] valueAndMetadata = split[1].split(";");
                     String value = valueAndMetadata.length > 0 ? valueAndMetadata[0] : "";
                     String metadata = valueAndMetadata.length > 1 ? valueAndMetadata[1] : "";
-                    Baggage.builder()
-                            .put(split[0], value, BaggageEntryMetadata.create(metadata))
-                            .build()
-                            .storeInContext(context)
-                            .makeCurrent();
+                    baggageBuilder
+                            .put(split[0], value, BaggageEntryMetadata.create(metadata));
```
ljnelson, line 289: Minor nit and I know it's not in the scope of this PR: the `split` calls will recompile their regex patterns each time

ljnelson (approved):
LGTM with two tiny comments

microprofile/telemetry/pom.xml
ljnelson: Aha, yes there is: https://maven.apache.org/plugins/maven-compiler-plugin/testCompile-mojo.html#useModulePath

Issue #9905, opened 2025-03-14
4.x Unable to use response.trailers().add(...) from a Filter


RickyFrost:
## Environment Details
* Helidon Version: 4.1.6
* Helidon SE
* JDK version: 21.x
* OS: Linux

----------

## Problem Description

There are many handlers ~30 in our service, all flowing through a single webserver.http.Filter doing things at various points in the lifecycle.
Example: do request setup before calling proceed, record certain statistics in the ::whenSent, or final cleanup in the filter/finally, etc

We are implementing Server-Timing using the response.trailers().add(...) in the ::whenSent.
But it turns out that event occurs too late (by definition -- send is already underway).

In the past we were running on H3.x and had a ::beforeSend event which would be the place to add the trailer header.
But this event was removed in H4.

Do we have another alternative to ::beforeSend in the H4.x environment?
Is there some other way than making an attempt at inlining in the ~30 handlers?
This seems error-prone and could be very inconsistent as the project grows and evolves.

## Steps to reproduce

This issue relates to Helidon-8266 (or may even be a dup) with the ::trailers use case.
https://github.com/helidon-io/helidon/issues/8266

It also relates to the trailers test already in Helidon, see below link. In all the tests, the trailers are added "inline" of the handler. Not a filter event.
https://github.com/helidon-io/helidon/blob/main/webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/HeadersServerTest.java

RickyFrost:
If it helps here is a super small outline of the filter code:
`
public class MyFilter {

    public void filter(FilterChain chain, RoutingRequest request, RoutingResponse response) {

        final var processTrailers = request.headers().contains(HeaderValues.TE_TRAILERS);
        if (processTrailers)
            response.headers().add(HeaderNames.TRAILER, SERVER_TIMING.defaultCase());

        // ...more processing            

        response.whenSent(() -> {
            if (processTrailers)
                response.trailers().add(SERVER_TIMING, timingValue);

            // ...more processing
        });

        filterChain.proceed();
    }
}
`

RickyFrost:
p.s. Is there is a better way to accomplish this?
I'm totally open to doing that instead of above.

spericas:
Indeed `whenSent` would be too late to add a trailer in H4. We may need to add a `beforeSent` event to support your use case. Looking for any other options ...

RickyFrost:
[...]

PR #9931, opened 2025-03-25
Adds support for new request processing event called beforeTrailers

spericas:
### Description

Adds support for new event called `beforeTrailers`. The consumer registered for this event has a last-minute chance to add HTTP trailers to a response. This can be useful in filters for example. Issue #9905.

### Documentation

None

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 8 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -187,6 +190,16 @@ public boolean isNexted() {
         return nexted;
     }
 
+    @Override
+    public ServerResponse beforeTrailers(Consumer<ServerResponseTrailers> beforeTrailers) {
+        this.beforeTrailers = beforeTrailers;
+        return this;
+    }
+
+    protected Consumer<ServerResponseTrailers> beforeTrailers() {
```
tomas-langer, line 199: Missing javadoc on a protected method of a public class.

spericas: Fixed

tomas-langer (approved):

PR #9953, opened 2025-04-01
4.x: Fix usage of types with static builder method that does not return Builder class

tomas-langer:


tjquinno (approved):
LGTM although I am not at all familiar with this part of the system!

PR #9955, opened 2025-04-01
4.x: Added support for `@Service.Replacement` to allow replacing list of services

### LINKED ISSUE #9819: Service registry `@Replacement` (our approach to alternatives)

tomas-langer:
Resolves #9819 

Adds support for `@Service.Replacement`

To replace instance for single injection point, simply use a service with higher weight
To replace instances where a List is injected, replacement can be used. The list will contain services ordered by weight (descending) up to the first instance of a `@Replacement` service (inclusive).

### DIFF
--- a/service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
+++ b/service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
@@ -301,10 +301,17 @@ class CoreServiceRegistry implements ServiceRegistry, Scopes {
                 ResolvedType theOnlyContractRequested = lookup.contracts().iterator().next();
                 Set<ServiceInfo> subsetOfMatches = servicesByContract.get(theOnlyContractRequested);
                 if (subsetOfMatches != null) {
-                    // the subset is ordered, cannot use parallel, also no need to re-order
-                    subsetOfMatches.stream()
-                            .filter(lookup::matches)
-                            .forEach(result::add);
+
+                    // the subset is ordered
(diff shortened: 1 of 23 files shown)


service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
```diff
@@ -319,11 +326,14 @@ public List<ServiceInfo> lookupServices(Lookup lookup) {
             // table scan :-(
             metrics.fullScan();
             // we need to go through each service descriptor if it matches
-            servicesByDescriptor.keySet()
+            var tmpResult = servicesByDescriptor.keySet()
                     .stream()
                     .filter(lookup::matches)
                     .sorted(SERVICE_INFO_COMPARATOR)
-                    .forEach(result::add);
+                    .collect(Collectors.toUnmodifiableList());
```
Verdent, line 333: Not exactly a problem I think, but would not `toList()` be nicer here? Or is there a reason for this approach?

Verdent (approved):

service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
tomas-langer: `.toList()` instead of `.collect(Collectors.toUnmodifiableList())` would not be a problem, just did not know about that method...

tomas-langer:
Closing this PR, will create a new one if this feature is required, we agreed to postpone for now.

PR #9957, opened 2025-04-01
Prototype of headers wrapper with rollback method

spericas:
### Description

Prototype of headers wrapper with rollback method to avoid copying headers before calling Jersey.

### Documentation

None

### DIFF
--- a/http/http/src/main/java/io/helidon/http/ServerResponseHeadersWrapper.java
+++ b/http/http/src/main/java/io/helidon/http/ServerResponseHeadersWrapper.java
@@ -38,7 +38,7 @@ import io.helidon.common.media.type.MediaType;
 public class ServerResponseHeadersWrapper implements ServerResponseHeaders {
 
     private final ServerResponseHeaders delegate;
-    private Function<ServerResponseHeaders, ServerResponseHeaders> rollbackFunction = h -> h;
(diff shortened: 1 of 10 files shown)


http/http/src/main/java/io/helidon/http/ServerResponseHeadersWrapper.java
```diff
@@ -0,0 +1,319 @@
...
+
+import io.helidon.common.media.type.MediaType;
+
+/**
+ * A response headers wrapper class with rollback capabilities.
+ */
+public class ServerResponseHeadersWrapper implements ServerResponseHeaders {
+
+    private final ServerResponseHeaders delegate;
+    private Function<ServerResponseHeaders, ServerResponseHeaders> rollbackFunction = h -> h;
```
ljnelson, line 41: (Could just use `Function.identity()`.)

ljnelson: (And just a quick sanity check: does `rollbackFunction` need to be `volatile`, given that it is reassigned, or is it guaranteed that will always be on one thread (I assume that must be the case)?)

spericas: It think it should all be handled by the same thread, but it is worth checking.

PR #9964, opened 2025-04-02
4.x: Upgrade byte-buddy to 1.17.5 and asm to 9.8 for Java 25 support

barchetta:
### Description

Upgrade byte-buddy to 1.17.5 and asm to 9.8 for Java 25 support


ljnelson (approved):
LGTM. Note that Byte Buddy's micro version has been changing on a weekly (or so) basis lately as Raph changes to using the new classfile API under the covers.

PR #9980, opened 2025-04-08
4.x Use `Timer` base units to format `Timer#toString` and JSON output

### LINKED ISSUE #9717: 4.x `MTimer#toString` and the JSON output for timer values use seconds (change from 3.x)
## Environment Details
* Helidon Version: 4.x
* JDK version:
* OS:

----------

## Problem Description
In two places, output related to `Timer` values is expressed in seconds, regardless of the setting of `baseUnit`:
* `MTimer#toString()` - uses `Duration` output with seconds units which is often not enough precision to display meaningful data for small values
* JSON-formatted output - uses seconds as units. Here the output is decimal (not truncated to an integer) but is a change from 3.x behavior.

### Some notes
The Micrometer `Timer` to which Helidon's API delegates does not itself support `baseUnits` as a builder setting. Micrometer's timer stores values in nanoseconds internally. The various methods on Micrometer's `Timer` which return the timer's data _do_ accept a `TimeUnit` parameter which determines the units in which the returned value (a `double`) is expressed.

That said, the Helidon `Timer.Builder` by virtue of extending `Meter.Builder`, does expose a `baseUnit(String)` method. The current Helidon implementation ignores that setting entirely when it builds the Micrometer `Timer`; there is no way to communicate this value to the Micrometer `Timer.Builder`.  

The Micrometer `Timer` does expose a `baseTimeUnit` method with this Javadoc:

> The base time unit of the timer to which all published metrics will be scaled

but Micrometer decides internally what this value will be based on what lower-level timer implementation(s) are used by the higher-level timer. For example, typically the actual timer Micrometer uses is a `PrometheusTimer` and its `baseTimeUnit` is seconds. 

One possibility could be for the Helidon `MTimer` implementation to save the specified base unit when the Helidon `Timer.Builder#baseUnit(String)` method is invoked and then for the Helidon code to use that in the `toString()` output.

Helidon could also use such a saved base unit value in formatting the JSON output. If the developer never specified a base unit to use, the default could be nanoseconds which (IIRC) would reinstate the 3.x behavior. 

## Steps to reproduce
1. Add a `Timer` to the SE QuickStart app's `GreetService`:
   
   ```
   private Timer greetingTimer;
   ```
   
   In the constructor:
   ```
      greetTimer = Metrics.globalRegistry().getOrCreate(Timer.builder("greetTimer");
   ```
   
   As the new first lines of `getDefaultMessageHandler`:
   ```
      greetTimer.record(123, TimeUnit.MILLISECONDS);
      LOGGER.log(System.Logger.Level.INFO, "Timer is now " + greetTimer);
   ```
2. Build and run the app.
3. `curl http://localhost:8080/greet` # to update the timer
4. `curl http://localhost:8080/observe/metrics -H "Accept: application/json"`

The server console output contains
```
Timer is now MTimer[id=PlainId[name='greetTimer', tags=[MTag[scope=application]]], scope='application', count=1, totalTime=PT0S]
```
(note the 0 reported time) and the metrics output includes
```
"greetTimer": {
      "count": 1,
      "max": 0.123,
      "mean": 0.123,
      "elapsedTime": 0.123,
      "p0.5": 0.121634816,
      "p0.75": 0.121634816,
      "p0.95": 0.121634816,
      "p0.98": 0.121634816,
      "p0.99": 0.121634816,
      "p0.999": 0.121634816
    }
```
The output is expressed in seconds.

tjquinno:
### Description
Resolves #9717 

## Release note
____
This change addresses two ways timer values are output:
* `MTimer#toString` continues to use `Duration` format but now shows as much precision as possible, regardless of any `baseUnit` setting on the timer.
* So users can control JSON output for timers, Helidon adds the new optional config setting `metrics.timers.json-units-default` which can have any `TimeUnit` value name.
   - If the config key is _absent_, Helidon expresses all timer values in seconds.
   - If the config key is _present_, Helidon uses the timer's explicit units setting if any was set and the config value otherwise.
   
   The default JSON output for timers in Helidon 4.x remains unchanged: JSON output for all timers is in seconds.
   
   For JSON output as in Helidon 3.x set `metrics.timers.json-units-default` to `NANOSECONDS`.
____

## Background
Helidon 4.x implements metrics using Micrometer, and although most Micrometer meters accept a `baseUnit` for storing the meter's values the `Timer` type does not. Instead, when code _retrieves_ a timer's value it specifies what Java `TimeUnit` the value should be scaled to. 

Helidon did not consistently deal with timer units.

## PR Overview

1. The `MTImer` class is the Helidon implementation of our `Timer` interface which delegates to Micrometer's `Timer`.
   * `MTimer` now stores the `baseUnit` setting if user code sets it and returns that value when asked. (It used to just use our superclass `MMeter#baseUnit` implementation which would return the `baseUnit` from the Micrometer delegate. Micrometer timers return seconds.)
   * `MTimer#toString` now retrieves the total time from the timer with as much precision as possible and (as before) formats it using `Duration`. (It used to retrieve the timer's `baseUnit` and retrieve the timer's value using that precision which was always seconds, leading to sometimes confusing output.)
   * Testing for this PR revealed an unrelated bug in the `Builder#buckets` method that is fixed.
2. The `MetricsConfigBlueprint` now has a new optional setting `metrics.timers.json-units-default`.
3. The `JsonFormatter` code now checks config for the new setting and computes the units to use for each timer accordingly (as described above).

### Documentation
Doc updates are included in the PR.

### DIFF
--- a/metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.TreeMap;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 
 import io.helidon.builder.api.Option;
@@ -60,6 +61,8 @@ interface MetricsConfigBlueprint {
      * Config key for KPI metrics settings.
      */
     String KEY_PERFORMANCE_INDICATORS_CONFIG_KEY = "key-performance-indicators";
+
+    TimeUnit DEFAULT_JSON_UNITS_DEFAULT = TimeUnit.SECONDS;
 
     @Prototype.FactoryMethod
     static List<Tag> createTags(Config globalTagExpression) {
@@ -249,6 +252,18 @@ interface MetricsConfigBlueprint {
     @Option.Configured
     @Option.Default(BuiltInMeterNameFormat.DEFAULT)
     BuiltInMeterNameFormat builtInMeterNameFormat();
+
+    /**
+     * Default units for timer output in JSON if not specified on a given timer.
+     * <p>
+     * If the configuration key is absent, the Helidon JSON output uses {@link java.util.concurrent.TimeUnit#SECONDS}.
+     * If the configuration key is present, Helidon formats each timer using that timer's specific units (if set) and
+     * the config value otherwise.
+     *
+     * @return default {@link java.util.concurrent.TimeUnit} to use for JSON timer output
+     */
+    @Option.Configured("timers.json-units-default")
+    Optional<TimeUnit> jsonUnitsDefault();
 
     /**
      * Reports whether the specified scope is enabled, according to any scope configuration that
--- a/docs/src/main/asciidoc/includes/metrics/metrics-shared.adoc
+++ b/docs/src/main/asciidoc/includes/metrics/metrics-shared.adoc
@@ -412,7 +412,7 @@ endif::mp-flavor[]
 <1> For _units_ specify any valid name for a
 link:{jdk-javadoc-url}/java.base/java/util/concurrent/TimeUnit.html[`TimeUnit`] value (`SECONDS`, `MILLISECONDS`, etc.)
 
-Helidon formats each timer's data as follows:
+If you have configured `json-units-default`, Helidon formats each timer's data as follows:
 
 1. If code set `baseUnit` on the timer, Helidon uses those units for that timer.
 2. Otherwise, Helidon uses the default units you configured.
--- a/docs/src/main/asciidoc/includes/metrics/metrics-shared.adoc
+++ b/docs/src/main/asciidoc/includes/metrics/metrics-shared.adoc
@@ -172,6 +172,7 @@ Clients can request a particular output format.
 | JSON | Header `Accept: application/json`
 |====
 
+[[scope-specific-retrieval]]
 Clients can also limit the report by specifying the scope as a query parameter in the request URL:
 
 * `{metrics-endpoint}?scope=base`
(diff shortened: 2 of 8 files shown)


tjquinno:
Especially to reviewers @ljnelson @spericas @barchetta ...

This PR addresses an unintentional change in behavior from 3.x to 4.x. As posted now, it does so in a way that _preserves_ the 4.x behavior by default (for consistency among 4.x releases) while allowing users a way to switch back to the "correct" 3.x default behavior via config.

As I think more about this, I am inclined to change the solution so the default behavior would become the "correct" 3.x behavior, and if users want to preserve the "incorrect" 4.x behavior they would need to set a config value. But I'd like others' thoughts on that.

Again, this fix affects `MTimer#toString` output (which seems less problematic if it were to change between releases within 4) and also the JSON output from `/metrics` which seems as if it could be more of a potential problem--for users who might be depending on the existing "incorrect" JSON output.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
```diff
@@ -250,6 +251,16 @@ static List<Tag> createTags(String pairs) {
...
+    /**
+     * Default base units for timers if not specified on the timer builder.
+     * If neither the configuration nor the builder specifies the timer units the results can depend on
+     * the underlying provider implementation of timers and, therefore, might not be predictable.
+     *
+     * @return default {@link java.util.concurrent.TimeUnit} to use as the base unit for timers
+     */
+    @Option.Configured("timers.base-units-default")
+    Optional<TimeUnit> timerBaseUnitsDefault();
+
```
spericas, line 263: Why not provide a default (for this default) and dtop the optional here? Seems simpler and predictable.

spericas: Not backward compatible maybe?

tjquinno: Yes, that was pretty much my thinking. The current (before the PR) 4.x behavior is that the Helidon layer delegates to the Micrometer layer to decide what units to apply. Using an `Optional` here allows us to detect when the user config actually specified something so we know whether to use that or delegate to Micrometer.

spericas: So Micrometer is not always using the same defaults?

tjquinno: In typical (default) usage, Micrometer will use Prometheus metric implementations (including timer) but that's not guaranteed. Users could have Micrometer itself use some other implementation underneath, in which case the default behavior might be different. The Helidon code would not necessarily see that because it uses the Micrometer API, and our code does not look behind the scenes to see if the user has plugged in some other underlying implementation under Micrometer.

I'm not aware of any of our users actually doing this, but there's no reason they couldn't and no inherent reason we'd hear about it.

spericas (dismissed):

tjquinno:
Reviewers @spericas @ljnelson @barchetta 

After discussions with Joe I've revised the changes a bit. I've rewritten the PR description to reflect the slightly different approach and the changes made by the PR, and I've added a doc update.

barchetta (approved):
LGTM

rhomast (approved):

PR #9988, opened 2025-04-10
4.x: Restore existing configuration on test container close

### LINKED ISSUE #9987: 4.x: `@HelidonTest` does not restore the original resolved config
## Environment Details
* Helidon Version: 4.2.0
* Helidon MP
----------

## Problem Description

`@HelidonTest` mutates the resolved config and does not restore it after the test are executed.
Thus if a test that relies on the default config runs after a test annotated with `@HelidonTest`, the resolved config instance may be incorrect.

E.g.

If the test class-path contains a file `mp-meta-config.yaml`, the default config should contain the config-sources configured by the meta config.

When using `@HelidonTest` the default behavior is to create a new config that does not include meta-configuration. One must use `@Configuration(useExisting = true)` to use the existing/default config.

Because `@HelidonTest` does not restore the original config, any test that does not use `@HelidonTest` or that uses `@Configuration(useExisting = true)` will resolve the config of the previous test / container.

## Workaround

If you clear the resolved before running your test, it will force the creation of the default config.

```java
@BeforeAll
static void beforeAll() {
    ConfigProviderResolver resolver = ConfigProviderResolver.instance();
    resolver.releaseConfig(resolver.getConfig());
}
```

romain-grecourt:
### Description

Restore existing configuration on test container close

- Add HelidonTestConfig.restore() to restore the original config (pre CDI container start)
- Call HelidonTestConfig.restore() as part of the CDI container shutdown.
- Update testng integration to aggressively hide the instrumented class in test results.

Fixes #9987

### Documentation

None.

### DIFF
--- a/microprofile/testing/testing/src/main/java/io/helidon/microprofile/testing/HelidonTestConfig.java
+++ b/microprofile/testing/testing/src/main/java/io/helidon/microprofile/testing/HelidonTestConfig.java
@@ -88,4 +88,13 @@ class HelidonTestConfig extends HelidonTestConfigDelegate {
             delegate = syntheticConfig;
         }
     }
+
+    /**
+     * Restore the original config.
+     */
+    void restore() {
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        ConfigProviderResolver resolver = ConfigProviderResolver.instance();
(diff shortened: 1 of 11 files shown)


danielkec (approved):

klustria (approved):
LGTM

PR #9997, opened 2025-04-11
Ensures SecretBundleNodeConfigSource can actually be polled

ljnelson:
While `SecretBundleNodeConfigSource` is set up to be pollable, and its `Builder` defines a public `pollingStrategy` method, the builder does not actually implement `PollableSource.Builder`, which turns out to be a requirement for pollable sources. This PR fixes the issue.


klustria (approved):
LGTM!

PR #10002, opened 2025-04-13
Leverages SequencedMap semantic methods in common LruCache (#10000)

### LINKED ISSUE #10000: Leverage SequencedMap in LruCacheImpl
Use the intentional SequencedMap #[putLast](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/util/SequencedMap.html#putLast(K,V))  and [#pollFirstEntry](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/util/SequencedMap.html#pollFirstEntry()) methods in enforcing Lru policy
I will submit PR resolving this issue

saladmaker:
fixes #10000

### DIFF
--- a/common/common/src/main/java/io/helidon/common/LruCacheImpl.java
+++ b/common/common/src/main/java/io/helidon/common/LruCacheImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2024, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/common/src/main/java/io/helidon/common/LruCacheImpl.java
+++ b/common/common/src/main/java/io/helidon/common/LruCacheImpl.java
@@ -15,9 +15,9 @@
  */
 package io.helidon.common;
 
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Optional;
+import java.util.SequencedMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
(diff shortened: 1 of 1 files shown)


common/common/src/main/java/io/helidon/common/LruCacheImpl.java
```diff
@@ -47,32 +47,31 @@ final class LruCacheImpl<K, V> implements LruCache<K, V> {
     public Optional<V> get(K key) {
         readLock.lock();
 
-        V value;
         try {
-            value = backingMap.get(key);
+            V value = backingMap.get(key);
+            if (null == value) {
```
tomas-langer, line 52: Moving this null check into the try/finally block will just make us hold the `readLock` a bit longer. Can you please explain the reason for moving it here?

common/common/src/main/java/io/helidon/common/LruCacheImpl.java
```diff
@@ -47,32 +47,31 @@ final class LruCacheImpl<K, V> implements LruCache<K, V> {
...
-            value = backingMap.get(key);
-            if (null == value) {
+            V freshValue = backingMap.get(key);
+            if (null == freshValue) {
                 return Optional.empty();
             }
+            //LRU policy
             backingMap.remove(key);
-            backingMap.put(key, value);
+            backingMap.putLast(key, freshValue);
```
tomas-langer, line 72: This is now inconsistent with the `put` operation - if `putLast` is sufficient, we can remove the line `backingMap.remove(key);`, if it is not sufficient, the line must be added to the `put` operation as well (where you removed it). 
Kindly align the operations with the miminal required code.

tomas-langer (changes requested):
Thanks for your contribution!
I have a added a couple of comments - the first one is just a question, the second one requires some work to align.

common/common/src/main/java/io/helidon/common/LruCacheImpl.java
saladmaker: Indeed! `putLast` will do just right.

PR #10010, opened 2025-04-14
4.x Tolerate if `PostRequestMetricsSupport` is missing from request context

### LINKED ISSUE #9995: 4.x Setting `metrics.rest-request.enabled=true` in MP multiport example throws exception
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
In a multi-port MP app, enabling REST request metrics using the SE-style property `metrics.rest-request.enabled=true` triggers an exception during endpoint access.

Using the MP-style property `mp.metrics.rest-request.enabled=true` (note the `mp.` prefix) does not cause an exception.

```listing
Caused by: java.util.NoSuchElementException: No value present
        at java.base/java.util.Optional.orElseThrow(Optional.java:377)
        at io.helidon.webserver.observe.metrics.PostRequestMetricsSupport.recordPostProcessingWork(PostRequestMetricsSupport.java:47)
        at io.helidon.microprofile.metrics.InterceptorSyntheticRestRequest.preInvocation(InterceptorSyntheticRestRequest.java:89)
        at io.helidon.microprofile.metrics.InterceptorSyntheticRestRequest.preInvocation(InterceptorSyntheticRestRequest.java:51)
        at io.helidon.microprofile.servicecommon.InterceptionRunnerImpl.run(InterceptionRunnerImpl.java:81)
        at io.helidon.microprofile.servicecommon.HelidonInterceptor.aroundInvocation(HelidonInterceptor.java:77)
        at io.helidon.microprofile.servicecommon.HelidonInterceptor$Base.aroundInvoke(HelidonInterceptor.java:136)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at org.jboss.weld.interceptor.reader.SimpleInterceptorInvocation$SimpleMethodInvocation.invoke(SimpleInterceptorInvocation.java:73)
        at org.jboss.weld.interceptor.proxy.InterceptorMethodHandler.executeAroundInvoke(InterceptorMethodHandler.java:84)
```

## Steps to reproduce
(Thanks to the user who pointed this out.)

To the Helidon MP multiport example https://github.com/helidon-io/helidon-examples/tree/helidon-4.x/examples/microprofile/multiport add `META-INF/microprofile-config.properties` (or make the corresponding change in `application.yaml`):
```properties
metrics.rest-request.enabled=true
```
Building with `mvn clean package` fails with test errors. 

So instead build and run this way:
```shell
mvn clean package -DskipTests
java -jar target/helidon-examples-microprofile-multiport.jar
```

Then access the app using
```shell
curl http://localhost:8080/hello
```
which triggers the exception.

tjquinno:
### Description
Resolves #9995 

## Release Note
____
Helidon MP apps now log a one-time warning instead of failing during request processing if multiple ports are configured and the SE setting `rest-request.enabled=true` is present.
____

## Background for reviewers
To support KPI metrics  on REST endpoints, Helidon MP adds a synthetic annotation to each REST endpoint and binds an interceptor (`InterceptorSyntheticRestRequest`) to that annotation. The interceptor (using `PostRequestMetricsSupport`) updates any KPI metrics measured on requests by fetching a metrics data structure from the request context to update.

That metrics data structure is placed into the request context by an SE-style filter.

If an MP app sets the SE-style setting `metrics.rest-request.enabled=true`--as opposed to the proper MP-style `mp.metrics.rest-request.enabled=true`--the filter is added only to the routing that also handles the metrics endpoint. (A separate enhancement request calls for a better approach to that; this PR addresses the current runtime failure in this scenario.) As a result, only requests on that socket have the metrics data structure in the request context. 

But, because _every_ REST request has the MP interceptor, the processing for all requests--regardless of socket--attempt to fetch the data structure.

The original code threw an exception if the expected data structure was missing from the request context.

The PR changes that so the code proceeds after logging a one-time warning message if it cannot find the data structure. 

Ideally we can remove the checking and the logging code once we redo how we handle KPI metrics on multiple sockets if  that update makes sure this situation cannot arise.

The PR adds an integration test for the failure scenario.

### Documentation
No impact.

### DIFF
--- a/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/PostRequestMetricsSupportImpl.java
+++ b/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/PostRequestMetricsSupportImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,8 +24,25 @@ import io.helidon.webserver.http.ServerResponse;
 
 class PostRequestMetricsSupportImpl implements PostRequestMetricsSupport {
 
+    private static final System.Logger LOGGER = System.getLogger(PostRequestMetricsSupportImpl.class.getName());
+
+    private static boolean loggedMessageAboutMissingKpiDataStructure = false;
+
     static PostRequestMetricsSupportImpl create() {
         return new PostRequestMetricsSupportImpl();
+    }
+
+    // @Deprecated - We should be able to remote this and its call once we update how KPI metrics are handled on multiple sockets.
+    @Deprecated(forRemoval = true, since = "4.2.1")
+    static void logMessageAboutMissingKpiDataStructure() {
+        if (!loggedMessageAboutMissingKpiDataStructure) {
+            loggedMessageAboutMissingKpiDataStructure = true;
+            LOGGER.log(System.Logger.Level.WARNING,
+                       """
+                               An expected metrics data structure is missing from the request context;
+                               an MP application might have set metrics.rest-request.enabled=true instead of assigning the proper
+                               key mp.metrics.rest-request.enabled=true. This message will not be repeated.""");
+        }
     }
 
     private final List<BiConsumer<ServerResponse, Throwable>> tasks = new ArrayList<>();
--- a/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/PostRequestMetricsSupportImpl.java
+++ b/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/PostRequestMetricsSupportImpl.java
@@ -32,8 +32,6 @@ class PostRequestMetricsSupportImpl implements PostRequestMetricsSupport {
         return new PostRequestMetricsSupportImpl();
     }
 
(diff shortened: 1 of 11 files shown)


webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/PostRequestMetricsSupportImpl.java
```diff
@@ -24,10 +24,25 @@
 
 class PostRequestMetricsSupportImpl implements PostRequestMetricsSupport {
 
+    private static final System.Logger LOGGER = System.getLogger(PostRequestMetricsSupportImpl.class.getName());
+
+    private static boolean loggedMessageAboutMissingKpiDataStructure = false;
```
ljnelson, line 29: Naïve comment: should this be `volatile` (and possibly atomic via `VarHandle`)? Don't know if multiple threads can get here but if they can then the message might be output twice if there's a race

ljnelson (approved):

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/PostRequestMetricsSupportImpl.java
tjquinno: No, not naïve and probably a valid point. 

But I hope this code will vanish when we (I guess that means when I) revisit how we deal with KPI metrics on multiple sockets. If the server output contains a small number >1 copies of the warning due to a heavy initial rush of incoming requests then I'm very OK with that.

PR #10015, opened 2025-04-15
4.x: Helidon Data review updates

tomas-langer:
Review updates:
- fix versions in bom pom
- remove versions from all pom
- remove gapi module
- update entity provider codegen, remove Class.forName
- renamed connection string in SqlConfigBlueprint to connection uri
- added version management for commons-compress and related libraries
- register JNDI in component using it

Other updates:
- added connection config blueprint used from providers and data sources
- removed data source common module
- changed the configuration structure to use `connection` for providers (not when using data source)

### Documentation

If no doc impact: None


Tomas-Kraus (approved):
Looks good.

PR #10017, opened 2025-04-15
Ensures that Helidon does not violate the CDI specification by implementing jakarta.enterprise.inject.Instance or jakarta.inject.Provider

ljnelson:
This PR ensures that Helidon's `CdiConfigExtension` does not attempt, deliberately or otherwise, to provide implementations of `jakarta.enterprise.inject.Instance` or `jakarta.inject.Provider`, as those interfaces [must be implemented directly by the CDI implementation](https://jakarta.ee/specifications/cdi/4.1/jakarta-cdi-spec-4.1#builtin_instance).

This PR makes a minimum number of changes to accomplish this goal.

### DIFF
--- a/microprofile/config/pom.xml
+++ b/microprofile/config/pom.xml
@@ -64,20 +64,9 @@
             <optional>true</optional>
         </dependency>
         <dependency>
-            <groupId>io.helidon.microprofile.testing</groupId>
-            <artifactId>helidon-microprofile-testing-junit5</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>io.helidon.microprofile.cdi</groupId>
             <artifactId>helidon-microprofile-cdi</artifactId>
             <scope>test</scope>
-            <exclusions>
-              <exclusion>
-                <groupId>io.helidon.microprofile.server</groupId>
-                <artifactId>helidon-microprofile-server</artifactId>
-              </exclusion>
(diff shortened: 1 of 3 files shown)


ljnelson:
Build failed; looks like a cycle in the build. Not sure why this didn't trigger locally. Will revamp the test.

ljnelson:
Looks like `HelidonTest` depends on a MicroProfile Config CDI implementation, which is the very project I'm working in here, so that's a cycle. Reworked the test to use `SeContainer` directly.

klustria (approved):
LGTM and have have tested the change to be working.

ljnelson:
No action item here on this closed and merged PR but there is something slightly unsettling to me about how an `Instance#isResolvable()` call will succeed, but then a `get()` call will throw a `NoSuchElementException` if the configuration property in question does not exist. There are areas of the MicroProfile Config specification that are vague and problematic; this may be one of them, or we may have another bug centered around how beans that produce configuration values are registered, or everything may be fine and this is Just How It Works. Obviously we pass the relevant TCK, but the TCK is problematic in many areas and may simply not cover this case.

To be clear, _this_ PR deliberately stays away from such things and only (properly and correctly) ensures that Helidon does not accidentally or deliberately try to register its own `Provider` and/or `Instance` implementations (which previously it was doing), so it remains correct and valid.

Issue #10022, opened 2025-04-16
Support Retrieval of the Route Template


SentryMan:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP

----------

## Enhancement Description

Support returning the matched route, that is, the path template as it was registered.

Example: `/path/{action}/{id?}`

This is required for OTEL agent integration

tjquinno:
This feature will also be needed for adding full OpenTelemetry support to Helidon SE for the same underlying reason: the OTel semantic conventions require a low-cardinality (small number of distinct values) path to be added as a tag to each span, whether created by the agent or not.

PR #10032, opened 2025-04-21
Provides access to matching pattern in ServerRequest

spericas:
### Description

Provides access to matching pattern in ServerRequest. See issue #10022.

### Documentation

None

### DIFF
--- a/http/http/src/main/java/io/helidon/http/PathMatchers.java
+++ b/http/http/src/main/java/io/helidon/http/PathMatchers.java
@@ -19,6 +19,7 @@ package io.helidon.http;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
@@ -61,22 +62,10 @@ public final class PathMatchers {
      *
      * @param pathToMatch the path must be prefixed by the provided path
      * @return prefix match path matcher
-     * @deprecated use {@link #prefix(String, String)} instead to preserve original pattern
      */
-    @Deprecated(forRemoval = true, since = "4.3.0")
     public static PathMatcher prefix(String pathToMatch) {
-         return prefix(pathToMatch, null);
-    }
-
-    /**
-     * Prefix match path matcher that stores original path pattern.
-     *
-     * @param pathToMatch matching path such as "/foo"
-     * @param pathPattern complete path pattern such as "/foo/*"
-     * @return prefix match path matcher
-     */
-    public static PathMatcher prefix(String pathToMatch, String pathPattern) {
-        return new PrefixPathMatcher(fixPrefix(pathToMatch), pathPattern);
+        Objects.requireNonNull(pathToMatch);
+        return new PrefixPathMatcher(fixPrefix(pathToMatch));
     }
 
     /**
@@ -180,7 +169,7 @@ public final class PathMatchers {
         }
 
         if (prefix) {
-            return prefix(checkPattern + "/", pathPattern);
+            return prefix(checkPattern + "/");
         }
         return exact(pathPattern);
     }
@@ -355,11 +344,9 @@ public final class PathMatchers {
     static final class PrefixPathMatcher implements PathMatcher {
         private final String prefix;
         private final String exactMatch;
-        private final String pathPattern;
 
-        PrefixPathMatcher(String prefix, String pathPattern) {
+        PrefixPathMatcher(String prefix) {
             this.prefix = prefix;
-            this.pathPattern = pathPattern;
             if (prefix.endsWith("/")) {
                 exactMatch = prefix.substring(0, prefix.length() - 1);
             } else {
@@ -420,7 +407,7 @@ public final class PathMatchers {
 
         @Override
         public Optional<String> matchingElement() {
-            return Optional.ofNullable(pathPattern);
+            return Optional.of(exactMatch + "/*");
         }
 
         @Override
--- a/http/http/src/main/java/io/helidon/http/PathMatchers.java
+++ b/http/http/src/main/java/io/helidon/http/PathMatchers.java
@@ -19,6 +19,7 @@ package io.helidon.http;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -60,9 +61,22 @@ public final class PathMatchers {
      *
      * @param pathToMatch the path must be prefixed by the provided path
      * @return prefix match path matcher
+     * @deprecated use {@link #prefix(String, String)} instead to preserve original pattern
      */
+    @Deprecated(forRemoval = true, since = "4.3.0")
     public static PathMatcher prefix(String pathToMatch) {
-        return new PrefixPathMatcher(fixPrefix(pathToMatch));
+         return prefix(pathToMatch, null);
+    }
+
+    /**
+     * Prefix match path matcher that stores original path pattern.
+     *
+     * @param pathToMatch matching path such as "/foo"
+     * @param pathPattern complete path pattern such as "/foo/*"
+     * @return prefix match path matcher
+     */
+    public static PathMatcher prefix(String pathToMatch, String pathPattern) {
+        return new PrefixPathMatcher(fixPrefix(pathToMatch), pathPattern);
     }
 
     /**
@@ -134,7 +148,7 @@ public final class PathMatchers {
                                                     + ", index: " + (iter.index() - 1));
         }
 
-        return new PatternPathMatcher(regexp.toString(), paramToGroupName);
+        return new PatternPathMatcher(regexp.toString(), paramToGroupName, pattern);
     }
 
     /**
@@ -166,7 +180,7 @@ public final class PathMatchers {
         }
 
         if (prefix) {
-            return prefix(checkPattern + "/");
+            return prefix(checkPattern + "/", pathPattern);
         }
         return exact(pathPattern);
     }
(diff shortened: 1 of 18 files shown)


tjquinno:
The approach here looks good to me.

One related but separate question is about filters. One use case--the one in the public slack channel that triggered this enhancement request--concerned adding a filter that would use the matched path. 

In that use case at least part of the filter's job would be to start an OpenTelemetry span according to the Otel semantic conventions, in which the name of the span uses low-cardinality route information--the matched path string  (`/greet/{name}`) not the actual path that is available via the request currently which can be high cardinality (`/greet/Joe` and `/greet/Julia` and many more).

To start the span, the code needs to know the name it wants to use for the span, and so for that reason it would be very useful to able to count on the matched path string being available via the request from filter code before invoking `filterChain#proceed`.

Are filters invoked before matching occurs or after? I didn't see this documented anywhere.

romain-grecourt:
> Are filters invoked before matching occurs or after? I didn't see this documented anywhere.

All filters are invoked before matching the routes, see https://github.com/helidon-io/helidon/blob/8da2cfa15998656b8589747487b8117356585339/webserver/webserver/src/main/java/io/helidon/webserver/http/Filters.java#L124

Both `path` and `matchingPattern` are updated while crawling, so the filters will get an empty string from `request. matchingPattern()`.

Maybe a callback to react when the matching occurs could be the solution.
E.g.

```java
routing.addFilter((chain, req, res) -> {
    req.whenMatched().thenRun(() -> {
        String pattern = req.matchingPattern();
        // will be invoked on every match
    });
    chain.proceed();
})
```

Another solution is to implement this use case using a non terminating handler:

```java
routing.any((req, res) -> {
    String pattern = req.matchingPattern();
    res.next();
})
```

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/MatchingPatternTest.java
```diff
@@ -0,0 +1,146 @@
...
+import io.helidon.webserver.testing.junit5.SetUpRoute;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@ServerTest
```
tomas-langer, line 35: Please add the same as a `@RoutingTest` to make sure all of this works without actual port open.

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutingRequest.java
```diff
@@ -38,4 +38,14 @@ public interface RoutingRequest extends ServerRequest {
...
      */
     RoutingRequest prologue(HttpPrologue newPrologue);
+
+    /**
+     * Update the pattern used to match this request. Such as "/foo/{bar}".
+     *
+     * @param matchingPattern the matching patter
+     * @return this instance
+     */
+    default RoutingRequest matchingPattern(String matchingPattern) {
```
tomas-langer, line 48: I do not find it great that this method is visible to users when using `ServerRequest`. This method should never ever be called by users, so it should be hidden from them.

webserver/webserver/src/main/java/io/helidon/webserver/http/HttpRoute.java
```diff
@@ -54,6 +54,15 @@ static HttpRouteImpl.Builder builder() {
      */
     Handler handler();
 
+    /**
+     * Path matcher for this HTTP route.
+     *
+     * @return path matcher or {@code null} if not applicable
+     */
+    default PathMatcher pathMatcher() {
```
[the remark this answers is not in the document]
spericas: Switching to `Optional`

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerRequest.java
```diff
@@ -120,4 +120,13 @@ public interface ServerRequest extends HttpRequest {
      * @see io.helidon.webserver.ListenerConfig#enableProxyProtocol()
      */
     Optional<ProxyProtocolData> proxyProtocolData();
+
+    /**
+     * The pattern used to match this request. Such as "/foo/{bar}".
+     *
+     * @return matching pattern or {@code ""} if not available
+     */
+    default String matchingPattern() {
```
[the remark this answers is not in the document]
spericas: Switching to `Optional`

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/MatchingPatternTest.java
spericas: Can you please elaborate more?

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutingRequest.java
spericas: Isn't that the case about all the methods in `RoutingRequest`?

webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/MatchingPatternTest.java
tomas-langer: `@ServerTest` will start the webserver and client will connect over network. `@RoutingTest` will not start a server and client will connect in-memory. So just exactly the same class, just a different annotation (or create an abstract class with all the tests and annotations except for the class annotation, and then create two classes that extend it, one with `@ServerTest` and one with `@RoutingTest`

spericas: Done

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutingRequest.java
spericas: Added comment

http/http/src/main/java/io/helidon/http/PathMatchers.java
```diff
@@ -60,9 +61,22 @@ public static PathMatcher exact(String pathToMatch) {
      *
      * @param pathToMatch the path must be prefixed by the provided path
      * @return prefix match path matcher
+     * @deprecated use {@link #prefix(String, String)} instead to preserve original pattern
      */
+    @Deprecated(forRemoval = true, since = "4.3.0")
     public static PathMatcher prefix(String pathToMatch) {
-        return new PrefixPathMatcher(fixPrefix(pathToMatch));
+         return prefix(pathToMatch, null);
```
tomas-langer, line 68: `null` is not allowed in Helidon as a parameter.
Also this method should not be removed, it still makes sense (as you can create a custom prefix path matcher using).
The default behavior can be to add `/*` to it for the `pathPattern`.

http/http/src/main/java/io/helidon/http/PathMatchers.java
```diff
@@ -60,9 +61,22 @@ public static PathMatcher exact(String pathToMatch) {
...
+
+    /**
+     * Prefix match path matcher that stores original path pattern.
+     *
+     * @param pathToMatch matching path such as "/foo"
+     * @param pathPattern complete path pattern such as "/foo/*"
+     * @return prefix match path matcher
+     */
+    public static PathMatcher prefix(String pathToMatch, String pathPattern) {
+        return new PrefixPathMatcher(fixPrefix(pathToMatch), pathPattern);
```
tomas-langer, line 79: There should be a check for not-null for both parameters.
I am not sure I understand why the second parameter is added.

http/http/src/main/java/io/helidon/http/PathMatchers.java
spericas: I wasn't sure adding "/*" was correct every time

http/http/src/main/java/io/helidon/http/PathMatchers.java
spericas: Yes, this seems to work fine, and `/foo/*/*` is handled as a `PatternPathMatcher`. See updates and new test.

tjquinno:
Quick note. Although invocations of the new `matchingPattern` method on request objects will not always be on the critical path of request processing, once Helidon more fully supports OpenTelemetry in SE--including adding spans for incoming requests--that code _will_ invoke `matchingPattern` on each request to assign the correct name to the new span. 

Certainly the rest of the code path for creating such a span will be quite a bit longer than checking an `Optional`, so I don't think this argues against using an `Optional` (instead of a separate boolean method that indicates if the matching pattern is valid). I just wanted to be clear that in some use cases this code will be invoked for each request.

spericas:
> Quick note. Although invocations of the new `matchingPattern` method on request objects will not always be on the critical path of request processing, once Helidon more fully supports OpenTelemetry in SE--including adding spans for incoming requests--that code _will_ invoke `matchingPattern` on each request to assign the correct name to the new span.
> 
> Certainly the rest of the code path for creating such a span will be quite a bit longer than checking an `Optional`, so I don't think this argues against using an `Optional` (instead of a separate boolean method that indicates if the matching pattern is valid). I just wanted to be clear that in some use cases this code will be invoked for each request.

So the span name will be derived from the matching pattern? That's odd

SentryMan:
> Users can accomplish filtering with an `any` handler as well. 

Does this work when a filter cancels a request with an error?

I have some filters that perform actions that can fail like validation where we can return a 400/500 instead of proceeding to a handler. The OTEL spans should ideally work for these cases.

tjquinno:
> So the span name will be derived from the matching pattern? That's odd
This link is also in a comment on the issue: https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/#http-route

tomas-langer (approved):

PR #10035, opened 2025-04-21
Adjusts OCI vault config sources to read only secret versions that are extant and in the Current rotation state

ljnelson:
This PR ensures that reads of OCI vault secrets are constrained to only those versions with the `Current` rotation state, and a lifecycle state that indicates current and ongoing existence.


klustria (approved):
Have tested these changes from my environment and everything looks to be working as expected, i.e. polling-strategy with lazy=false does not anymore produce intermittent NoSuchElementException problem.

PR #10050, opened 2025-04-24
Initial support for gRPC server metrics

spericas:
### Description

Support for gRPC-specific server metrics. Only records metrics when invocations are uccessfull (status is OK). Metrics disabled by default, and enabled via config. See https://grpc.io/docs/guides/opentelemetry-metrics/. 
 
### Documentation

TODO


spericas:
Closing since it needs to be rebased after https://github.com/helidon-io/helidon/pull/10089

Issue #10051, opened 2025-04-24
Support Automatic Instrumentation for OpenTelemetry


SentryMan:
## Environment Details
* Helidon Version: 4.x
* Helidon SE 

----------

## Enhancement Description

After consuming the changes of #10032 locally, I'm trying to prepare a PR for the OpenTelemetry Java Agent to add automatic instrumentation. 

It seems that we cannot get the matched pattern inside a filter, so the filter logic I have below doesn't work:

```java
final class OpenTelemetryFilter implements Filter {

  private final Instrumenter<ServerRequest, ServerResponse> instrumenter;

  OpenTelemetryFilter(Instrumenter<ServerRequest, ServerResponse> instrumenter) {
    this.instrumenter = instrumenter;
  }

  @Override
  public void filter(FilterChain chain, RoutingRequest req, RoutingResponse res) {

    Context parentContext = Context.current();
    if (!instrumenter.shouldStart(parentContext, req)) {
      chain.proceed();
      return;
    }

    Context context = instrumenter.start(parentContext, req);

    Throwable error = null;
    try (Scope ignored = context.makeCurrent()) {
      chain.proceed();
    } catch (Throwable t) {
      error = t;
      throw t;
    } finally {
      instrumenter.end(context, req, res, error);
    }
  }
}
```

spericas:
@SentryMan As explained in that PR, filters in Helidon SE are _pre-matching_, so it is impossible for that information to be available in that type of filter, the request simply hasn't matched yet. We'd need to explore other options such as callbacks or some other type of interception for this to work, but it can't be on a filter.

spericas:
There was a design decision to support _terminating_ and _non-terminating_ handlers. The latter type can call `res.next()` and resume the routing process. When calling a handler, Helidon does not know its type ahead of time, so it would need to call _post-matching_  filters (if they existed naturally) potentially multiple times, which makes the semantics tricky.

How is your `Instrumenter` above using the matched pattern in your example exactly?

SentryMan:
> How is your Instrumenter above using the matched pattern in your example exactly?

[...]

PR #10054, opened 2025-04-24
Use ofNullable for matching pattern

SentryMan:
### Description

When the matching pattern is not present, it seems that calling `ServerRequest#matchingPattern` throws an NPE.

```java
      WebServer.builder()
          .routing(
              b ->
                  b.addFilter(
                      (chain, req, res) -> {
                        req.matchingPattern();// throws NPE here
                        chain.proceed();
                      }))
          .port(8080)
          .build()
          .start();
```

part of #10051

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerRequest.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerRequest.java
@@ -198,7 +198,7 @@ class Http2ServerRequest implements RoutingRequest {
 
     @Override
     public Optional<String> matchingPattern() {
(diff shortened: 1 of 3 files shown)


romain-grecourt:
Can you add a unit test ?

SentryMan:
Soonest I can get back to this is Saturday. (I'm traveling)

SentryMan:
Alright back:

 > Can you add a unit test ?

How should I go about it? From what I can tell, `HttpRule` in the existing test doesn't let you register a filter.

SentryMan:
@romain-grecourt I have added some assertions

PR #9915, opened 2025-03-17
4.x: Helidon declarative HTTP, Fault Tolerance, and Scheduling

tomas-langer:
Resolves #9872 

First look at Helidon Declarative approach.
This PR brings support for

- HTTP server endpoints
- HTTP client
- scheduling
- fault tolerance

by a combination of
- generated source code
- interceptors

In addition it brings a new concept of `EntryPoint` - i.e. initial entries to the system from outside, such as HTTP request,  WebSocket request, grpc request, message received etc.
There is a set of tools to create entry point interceptors (one of the main reasons is to support security).

See `declarative/tests` for the usage (just two modules there with a few classes) - this should show all the currently supported features.

PR #10069, opened 2025-04-28
4.x: Codegen updates

tomas-langer:
### Description
Codegen changes needed for #9966 and #9915 

- support for decorators of builder methods that are `Set` or `List` type
- support for handling annotations on `module-info.java`
- support of upper and lower bounds when creating `TypeName` from `Type`

Each is a separate commit.

### DIFF
--- a/common/types/src/main/java/io/helidon/common/types/TypeNameSupport.java
+++ b/common/types/src/main/java/io/helidon/common/types/TypeNameSupport.java
@@ -273,10 +273,18 @@ final class TypeNameSupport {
             Stream.of(genericSection.split(","))
                     .map(String::trim) // remove possible spaces
                     .map(it -> {
-                        if (it.contains(".")) {
-                            return TypeName.create(it);
+                        if (it.contains(" extends ")) {
+                            // T extends io.helidon.webserver.spi.ServerFeature
+                            return createExtends(it);
+                        } else if (it.contains(" super ")) {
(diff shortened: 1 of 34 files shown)


common/types/src/main/java/io/helidon/common/types/TypeNameSupport.java
```diff
@@ -330,6 +338,38 @@ static TypeName createFromGenericDeclaration(String genericAliasTypeName) {
...
+        if (theOtherPart.contains(".")) {
+            builder.addUpperBound(TypeName.create(theOtherPart));
+        } else {
+            builder.addUpperBound(TypeName.createFromGenericDeclaration(theOtherPart));
+        }
+        return builder.build();
+    }
+
+    private static TypeName createSuper(String typeNames) {
+        // T extends io.helidon.webserver.spi.ServerFeature
```
ljnelson, line 358: This says "extends", but then the code uses `" super "`. Copy-and-paste bug?

tomas-langer: Fixed, luckily only the comment was copied, not the code.

ljnelson (approved):
I'll give you a green check to unblock you for the stuff in here I was able to understand. Someone else who is closer to all this should also weigh in

tomas-langer:
As I am the one who should understand this area, I did one more review of the code.

PR #10087, opened 2025-05-05
4.x: Validate that named routing has an existing named listener.

### LINKED ISSUE #9041: 4.x - Named routing validation
## Environment Details
* Helidon Version: 4.0.10
* Helidon SE
----------

## Problem Description

When named routing references a socket that is not configured, a warning should be emitted or an error should be thrown.

## Steps to reproduce

Consider the following config:

```properties
server.port=8080
server.sockets.0.name="admin"
server.sockets.0.port=8082
```

The config above has a typo, `"admin"` should not be quoted.
The code below should does not trigger a warning or an exception.

```java
WebServer.builder()
     .routing("admin", r -> r.get((req, res) -> res.send("OK")));
```

tomas-langer:
Resolves #9041 

### Description
This PR adds a check that all named routes have an associated named listener. If this is not met, and exception is thrown when the web server is created.

There is an option for backward-compatibility behavior added to configuration of webserver (both config and builder).

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.Timer;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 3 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java
```diff
@@ -191,6 +194,35 @@ public Context context() {
...
+        List<String> invalidRoutingNames = new ArrayList<>();
+
+        for (String routingName : routingNames) {
+            if (!namedListeners.contains(routingName)) {
+                invalidRoutingNames.add(routingName);
+            }
+        }
+
+        if (!invalidRoutingNames.isEmpty()) {
+            String message = "WebServer has configured named routing, but its"
```
romain-grecourt, line 210: How about "Socket not found for routing names: [admin, private]".

spericas (dismissed):

spericas (approved):

PR #10088, opened 2025-05-05
4.x: http module updates (constants for annotations, dependency plugin, javadocs fail on warning)

tomas-langer:
Switched fail on javadoc warning to true for http module

Switched dependency check plugin on for http module Fixed all discovered problems

Introduced constants for Method, MediaType, Status and Header names that can be used in annotations.

Related to #9872 

Resolves part of changes in #9915 to make review easier.

### DIFF
--- a/http/http/src/main/java/io/helidon/http/HeaderNameEnum.java
+++ b/http/http/src/main/java/io/helidon/http/HeaderNameEnum.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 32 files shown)


http/http/src/main/java/io/helidon/http/HeaderNameEnum.java
```diff
@@ -151,4 +151,84 @@ public String defaultCase() {
     public int index() {
         return index;
     }
+
+    static class Strings {
```
romain-grecourt, line 155: What's the reasoning for this class ? Could HeaderNames be used instead ?

spericas: Same question here. I also see in `HeaderNames`:

```
public static final String ACCEPT_NAME = Strings.ACCEPT_NAME;
```

romain-grecourt (approved):

PR #10089, opened 2025-05-06
Refactors code that handles gRPC interceptors in webserver

### LINKED ISSUE #10041: GrpcProtocolHandler does not take into account any registered interceptors
## Environment Details
* Helidon Version: 4.2.0
* Helidon MP
* JDK version: 21
* OS: Ubuntu

----------

## Problem Description
I've tried registering custom interceptors by using my own type extending `GrpcMpExtension` interface. This provides possibility to register services (which works perfectly) and interceptors (which aren't handled by `GrpcProtocolHandler`).

It looks like interceptors were omitted in `GrpcProtocolHandler` implementation.

## Steps to reproduce

Create your own type implementing `io.grpc.ServerInterceptor` and then register it using:
```
    @Override
    public void configure(GrpcMpContext context) {
        context.routing()
                .intercept(...)
```

spericas:
### Description

Refactors code that handles gRPC interceptors to properly support interception of methods regardless of how routes are set up. Fixed #10041.

### Documentation

None

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/ContextSettingServerInterceptor.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/ContextSettingServerInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcInterceptorUtil.java
```diff
@@ -0,0 +1,95 @@
...
+        // iterate the interceptors in reverse order to set up handler chain
+        for (int i = interceptorsList.size() - 1; i >= 0; i--) {
+            ServerInterceptor interceptor = interceptorsList.get(i);
+            interceptorsSet.add(interceptor);
+        }
+
+        // create interceptor chain
+        for (ServerInterceptor interceptor : interceptorsSet) {
+            handler = new InterceptingCallHandler<>(serviceDescriptor, interceptor, handler);
+        }
```
tjquinno, line 56: It looks as if the `Set` is just to enforce uniqueness and the `List` is so we can traverse it in reverse; otherwise we don't otherwise need access to either data structure. Instead, how about

```java
for (ServerInterceptor interceptor : interceptors.stream()
            .distinct()
            .toList()
            .reversed()) {
    handler = new InterceptingCallHandler<>(serviceDescriptor, interceptor, handler);
}
```
It's arguably more expressive of the intent. 

If this code is in a code path where performance is critical (I'm guessing it's not super-critical given the use of `WeightedBag#stream`) we might need to see whether creating and updating the `LinkedHashSet` or using `Stream#distinct` is more expensive.

`List#reversed` does not copy the list (although it does create a new object).

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcInterceptorUtil.java
```diff
@@ -0,0 +1,95 @@
...
+            this.callHandler = callHandler;
+        }
+
+        @Override
+        public ServerCall.Listener<ReqT> startCall(
+                ServerCall<ReqT, RespT> call,
+                Metadata headers) {
+            if (serviceDefinition != null && interceptor instanceof GrpcServiceDescriptor.Aware) {
+                ((GrpcServiceDescriptor.Aware) interceptor).setServiceDescriptor(serviceDefinition);
+            }
```
tjquinno, line 91:
```java
            if (serviceDefinition != null && interceptor instanceof GrpcServiceDescriptor.Aware awareInterceptor) {
                awareIinterceptor.setServiceDescriptor(serviceDefinition);
            }
```

tjquinno (commented):
A few possible changes.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/ContextSettingServerInterceptor.java
```diff
@@ -69,22 +69,24 @@ public <ReqT, ResT> ServerCall.Listener<ReqT> interceptCall(ServerCall<ReqT, Res
...
-        Map<Context.Key<?>, Object> contextMap = new HashMap<>();
-        contextMap.putAll(serviceDescriptor.context());
-        contextMap.putAll(methodDescriptor.context());
-        contextMap.put(GrpcServiceDescriptor.SERVICE_DESCRIPTOR_KEY, serviceDescriptor);
-
-        // set up context from gRPC API
-        for (Map.Entry<Context.Key<?>, Object> entry : contextMap.entrySet()) {
-            Context.Key<Object> key = (Context.Key<Object>) entry.getKey();
-            context = context.withValue(key, entry.getValue());
+        if (serviceDescriptor != null) {
```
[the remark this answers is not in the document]
spericas: Yes, because the `contextMap` is created only when the `serviceDescriptor` is provided

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcInterceptorUtil.java
spericas: Indeed, that's a lot nicer

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcInterceptorUtil.java
spericas: Will fix

tjquinno (approved):

Issue #9806, opened 2025-02-16
Add support for gRPC metrics


hyder:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP

----------

## Enhancement Description

We have metrics for REST but we don't yet have metrics for gRPC. We need gRPC metrics for both clients and servers. A good starting point and the type of metrics we can look for are https://grpc.io/docs/guides/opentelemetry-metrics/  and https://grpc.io/docs/guides/custom-backend-metrics/

hyder:
Additional ideas for grpc metrics

PR #10098, opened 2025-05-07
Support for gRPC server metrics

spericas:
### Description

Initial support for gRPC server metrics. Implements _Server Instruments_ section from this document  https://grpc.io/docs/guides/opentelemetry-metrics/, except that it only records sizes and durations for successful invocations (i.e., status tag is always OK). Frankly, it doesn't seem that useful to record these metrics for failed invocations, and it has the potential of creating a large number of metrics based on different status values.

The following metrics are created for each gRPC method (if invoked):

- grpc.server.call.started (counter for all invocations)
- grpc.server.call.duration (timer)
- grpc.server.call.sent_total_compressed_message_size (distribution summary)
- grpc.server.call.rcvd_total_compressed_message_size (distribution summary)

Metrics are _disabled_ by default and can be enabled via config. Partly addresses issue #9806.

### Documentation

In future PR.

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -20,6 +20,8 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.time.Duration;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -42,6 +44,13 @@ import io.helidon.http.http2.Http2StreamState;
 import io.helidon.http.http2.Http2StreamWriter;
 import io.helidon.http.http2.Http2WindowUpdate;
 import io.helidon.http.http2.StreamFlowControl;
+import io.helidon.metrics.api.Counter;
+import io.helidon.metrics.api.DistributionSummary;
+import io.helidon.metrics.api.MeterRegistry;
+import io.helidon.metrics.api.Metrics;
+import io.helidon.metrics.api.MetricsFactory;
+import io.helidon.metrics.api.Tag;
+import io.helidon.metrics.api.Timer;
 import io.helidon.webserver.http2.spi.Http2SubProtocolSelector;
 
 import io.grpc.Codec;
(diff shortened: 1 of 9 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -347,4 +394,52 @@ public MethodDescriptor<REQ, RES> getMethodDescriptor() {
...
+            MetricsFactory metricsFactory = MetricsFactory.getInstance();
+            MeterRegistry meterRegistry = Metrics.globalRegistry();
+
+            Tag grpcMethod = Tag.create("grpc.method", route.method().getFullMethodName());
+
+            Counter.Builder callStartedBuilder = metricsFactory.counterBuilder(
+                            "grpc.server.call.started")
+                            .scope(VENDOR);
+            callStartedBuilder.tags(List.of(grpcMethod));
+            Counter callStarted = meterRegistry.getOrCreate(callStartedBuilder);
```
tjquinno, line 415: Pure bike shedding comment...

The code could also chain the `tags` invocation onto the end of the `scope` invocation:
```java
Counter.Builder callStartedBuilder = Counter.builder("graph.server.call.started")
    .scope(VENDOR)
    .tags(List.of(grpcMethod));
Counter callStarted = meterRegistry(callStartedBuilder);
```
and depending on stylistic preferences (!) and readability could even do without the builder local variable:
```java
Counter callStarted = meterRegistry.getOrCreate(Counter.builder("graph.server.call.started")
                .scope(VENDOR)
                .tags(List.of(grpcMethod)));
```
(Same note for all the registrations.)

tjquinno (changes requested):
Some unimportant stylistic comments and some performance considerations.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -347,4 +396,45 @@ public MethodDescriptor<REQ, RES> getMethodDescriptor() {
...
     }
+
+    /**
+     * Initializes gRPC server metrics for the method being invoked. Note that
+     * duration and size metrics are currently recorded only for successful calls,
+     * using the {@link #OK_TAG}. If a call fails, it will only increment the number
+     * of started calls, but not record any of the other metrics.
+     */
+    private void initMetrics() {
+        if (grpcConfig.enableMetrics()) {
```
tjquinno, line 407: Minor: `initMetrics` is not invoked unless `grpcConfig.enableMetrics()` has already evaluated to `true` in the caller so the `if` here is superfluous.

tjquinno (dismissed):
Minor comment. Otherwise looks good to me.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
spericas: Good point, will fix

tjquinno (approved):

PR #10108, opened 2025-05-12
Updates to gRPC documentation

spericas:
### Description

Updates gRPC docs with new sections for server interceptors, server metrics and client metrics.

### DIFF
--- a/docs/src/main/asciidoc/se/grpc/client.adoc
+++ b/docs/src/main/asciidoc/se/grpc/client.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2019, 2024 Oracle and/or its affiliates.
+    Copyright (c) 2019, 2025 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -32,7 +32,8 @@ include::{rootdir}/includes/se.adoc[]
 ** <<Generated Stubs, Generated Stubs>>
 ** <<Service Descriptors, Service Descriptors>>
 ** <<Client URI Suppliers, Client URI Suppliers>>
-** <<gRPC Interceptors, gRPC Interceptors>>
+** <<Client Interceptors, Client Interceptors>>
+** <<Metrics, Metrics>>
 - <<Configuration, Configuration>>
 
 == Overview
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/se/grpc/client.adoc
```diff
@@ -191,6 +192,59 @@ with a list of interceptors to execute on every invocation.
 include::{sourcedir}/se/grpc/ClientSnippets.java[tag=snippet_10, indent=0]
 ----
 
+== Metrics
+
```
tjquinno, line 196: The Helidon `webclient/grpc` component depends on `helidon-metrics-api` which has only the no-op implementation. I know the doc links to the SE metrics doc page, but do we need to at least mention here (and in the server section) that to actually get metrics the developer would need to add a runtime dependency on `helidon-metrics-providers-micrometer` (or `helidon-metrics` which also includes the built-in system meters)?

docs/src/main/asciidoc/se/grpc/server.adoc
```diff
@@ -143,25 +146,112 @@ The service implementation will be very similar to our original implementation:
...
+it follows that for `MathService`, `Interceptor2` will be called _before_
+`Interceptor1` given its _higher_ weight of `Interceptor.USER + 100`.
+
+NOTE: Even though the gRPC API supports interception of methods via
+alternative mechanisms, it is recommended to use the `intercept` method
+on a `GrpcRouting` builder, as shown above, to ensure correct ordering
+based on weights.
+
+=== Metrics
+
```
tjquinno, line 194: Same question as earlier about maybe mentioning adding a dep. on `helidon-metrics-providers-micrometer` (or `helidon-metrics` which also includes the built-in system meters)?

tjquinno (changes requested):
A few requested changes and a few questions.

docs/src/main/asciidoc/se/grpc/server.adoc
spericas: See above

docs/src/main/asciidoc/se/grpc/client.adoc
spericas: Yeap, in other places in our docs we also reference the Metrics document. I think it is better to leave the explanation about implementations to that document instead of repeating it in multiple places. Easier to update in the future too.

Issue #10104, opened 2025-05-10
io.helidon.webclient.http1.Http1ClientRequestImpl not tolerant of mixed case header "Upgrade: WebSocket" during switching protocol check 101 upgrade


xio-1:
## Environment Details
* Helidon Version: 4.2.1
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

When using the websocket client with other websocket servers it your client code does not tollerate a mixed case response from the host of 'Upgrade: WebSocket' where such headers should be case tolerant to prevent the client rejecting the server upgrade approval.

Tested with undertow websocket server

## Steps to reproduce

Run undertow websocket server. connect to server using your WSClient impl and the client will reject the upgrade.
See https://github.com/helidon-io/helidon/blob/5da62343d8ca99b2d7bb8f602b3010d86a1c205c/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java#L157

spericas:
Can you share how the header looks like as returned by undertow?

barchetta:
For reference:
1. [Hypertext Transfer Protocol (HTTP) Upgrade Token Registry](https://www.iana.org/assignments/http-upgrade-tokens/http-upgrade-tokens.xhtml)
    * Specifies both "WebSocket" and "websocket"
2. [RFC 9110 HTTP Semantics: Upgrade Header](https://www.rfc-editor.org/rfc/rfc9110.html#name-upgrade)
    * States: "Although protocol names are registered with a preferred case, recipients SHOULD use case-insensitive comparison when matching each protocol-name to supported protocols."

PR #10115, opened 2025-05-13
Compares values of the Upgrade header ignoring case in WebClient

spericas:
### Description

Compares values of the Upgrade header ignoring case. Unfortunately, s…ome servers may return 'WebSocket' and others 'websocket' for example. See issue #10104.

### Documentation

None


webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
```diff
@@ -153,8 +153,8 @@ public UpgradeResponse upgrade(String protocol) {
         }
 
         if (response.status() == Status.SWITCHING_PROTOCOLS_101) {
-            // yep, this is the response we want
-            if (response.headers().contains(requestedUpgrade)) {
+            // is the upgrade request successful?
+            if (isUpgradeSuccessful(requestedUpgrade, response.headers().get(HeaderNames.UPGRADE))) {
```
ljnelson, line 157: I believe our [dev guidelines](https://github.com/helidon-io/helidon/blob/main/DEV-GUIDELINES.md#getters-and-setters) would say that this method should be, simply, `upgradeSuccessful` since `is` is a prefix of last resort.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
```diff
@@ -203,6 +203,22 @@ Http1ClientImpl http1Client() {
...
+    /**
+     * Check upgrade protocols. Protocol names are case insensitive.
+     *
+     * @param requestUpgrade request upgrade header
+     * @param responseUpgrade response upgrade header
+     * @return check if protocol upgrade can proceed
+     */
+    static boolean isUpgradeSuccessful(Header requestUpgrade, Header responseUpgrade) {
+        for (String protocol : requestUpgrade.allValues()) {
+            if (protocol.equalsIgnoreCase(responseUpgrade.get())) {
```
ljnelson, line 215: Maybe store the value of `responseUpgrade.get()` in a local variable outside of the loop?

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
spericas: Roger

ljnelson (approved):
LGTM

PR #10120, opened 2025-05-14
Simplify ready loop in WebClient proxy tests

barchetta:
We have seen cases where WebClient proxy tests are hanging in the validation pipeline and timing out after 30 minutes. See #10123

This PR simplifies the "ready" check for the test proxy server. Previously the ready check was a loop trying to repeatedly make connections to the proxy server until a connection was accepted. It then blocked on a latch to verify the connection has handled by the server.

The simplified version in this PR does away with the connection loop and simply blocks on the latch until the server decrements it after binding the server socket, but before accepting any connections.


spericas (approved):
Removing the potentially infinite loop seems like a good idea. We should also consider refactoring these tests to avoid starting and stopping a proxy _per unit test_; perhaps starting and stopping a single one before and after all tests instead.

Issue #9496, opened 2024-11-13
RST_STREAM sent by a Server Streaming client does not signal the gRPC layer


mattp-swirldslabs:
[//]: # "Please remove these comments"
[//]: # "Provide a general summary of the issue in the Title above"

## Environment Details
* Helidon Version: 4.1.1, 4.1.6
* Helidon SE
* JDK version: 21.0.5
* OS: Mac OS 14.6.1

----------

## Problem Description

RST_STREAM sent by a Server Streaming client does not signal the gRPC layer

## Steps to reproduce:

1) Start a Server Streaming connection to Helidon using a client like Postman and confirm you're receiving streaming data from the server.
2) Place a breakpoint in the [Http2ServerStream.rstStream() method](https://github.com/helidon-io/helidon/blob/51b942a90aa9f2416ce6b32687748ee2b7701e75/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java#L188-L198)
3) Use the Postman cancel button to send a `rst_stream` to terminate the connection
4) The state changes to `CLOSED` but the `Http2StreamWriter` (passed here: https://github.com/helidon-io/helidon/blob/51b942a90aa9f2416ce6b32687748ee2b7701e75/webserver/http2/src/main/java/io/helidon/webserver/http2/spi/Http2SubProtocolSelector.java#L60) continues to accept data from the gRPC layer.
5) Our application above the gRPC layer needs an exception or some signal from Helidon to know that it needs to stop streaming data.

* The "TODO" comment in the [Http2ServerStream.rstStream() method](https://github.com/helidon-io/helidon/blob/51b942a90aa9f2416ce6b32687748ee2b7701e75/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java#L188-L198) seems to indicate this might be a known issue

mattp-swirldslabs:
@danielkec Do you know if this issue was addressed as part of the "revamp" in #9520 ? Thanks!

danielkec:
Hi, I am afraid it is not, we have to yet address that

mattp-swirldslabs:
@danielkec Ok, thanks for the reply. Do you have an idea on the timeline or which release this might land in?

spericas:
[...]

PR #10128, opened 2025-05-16
Fixes problems cancelling a client-streaming gRPC call

spericas:
### Description

Fixes problems cancelling a client-streaming gRPC call. Here is a summary of the changes in this PR:

1. If a client-streaming call is cancelled before any data is sent, make sure final trailers frame includes gRPC type and status for compatibility (Postman)
2. When a gRPC server receives an EOS (end of stream), its internal state should be set to HALF_CLOSED_REMOTE (it was incorrectly set to HALF_CLOSED_LOCAL, causing it to exit the processing loop in the underlying HTTP/2 layer)
3. A RST_STREAM received for an HTTP/2 stream must be propagate to the corresponding sub-protocol handler

In addition to the new test in this PR, I've done some manual testing with Postman and grpcurl using client-streaming and bidi calls to ensure that proper cancellations are correctly handled. See issue #9496.

### Documentation

None

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -43,7 +43,6 @@ import io.helidon.http.http2.Http2FrameHeader;
 import io.helidon.http.http2.Http2FrameTypes;
 import io.helidon.http.http2.Http2Headers;
 import io.helidon.http.http2.Http2RstStream;
-import io.helidon.http.http2.Http2Settings;
 import io.helidon.http.http2.Http2StreamState;
 import io.helidon.http.http2.Http2StreamWriter;
 import io.helidon.http.http2.Http2WindowUpdate;
@@ -119,8 +118,6 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
                         Http2Headers headers,
                         Http2StreamWriter streamWriter,
                         int streamId,
-                        Http2Settings serverSettings,
-                        Http2Settings clientSettings,
                         StreamFlowControl flowControl,
                         Http2StreamState currentStreamState,
                         GrpcRouteHandler<REQ, RES> route,
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -94,12 +94,9 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
                                  DistributionSummary recvMessageSize) { }
     private static final LazyValue<Map<String, MethodMetrics>> METHOD_METRICS = LazyValue.create(ConcurrentHashMap::new);
 
-    private final HttpPrologue prologue;
     private final Http2Headers headers;
     private final Http2StreamWriter streamWriter;
     private final int streamId;
-    private final Http2Settings serverSettings;
-    private final Http2Settings clientSettings;
     private final GrpcRouteHandler<REQ, RES> route;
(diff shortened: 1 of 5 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -126,12 +125,9 @@ private record MethodMetrics(Counter callStarted,
                         Http2StreamState currentStreamState,
                         GrpcRouteHandler<REQ, RES> route,
                         GrpcConfig grpcConfig) {
-        this.prologue = prologue;
         this.headers = headers;
         this.streamWriter = streamWriter;
         this.streamId = streamId;
-        this.serverSettings = serverSettings;
-        this.clientSettings = clientSettings;
```
tjquinno, line 134: It looks as if three ctor parameters--`prologue, serverSettings,` and `clientSettings`--are no longer used. Can they be removed from the parameter list?

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -172,7 +168,8 @@ public Http2StreamState streamState() {
 
     @Override
     public void rstStream(Http2RstStream rstStream) {
-        listener.onComplete();
+        callCancelled = (rstStream.errorCode() == Http2ErrorCode.CANCEL);
```
tjquinno, line 171: Question: I don't know anything about the protocol. It looks as if we create a new `GrpcProtocolHandler` for each incoming request (the parameter list has a prologue, headers, etc. that would vary from one request to the next). 

Is there ever a time when a gRPC client would cancel the call but somehow the `GrpcProtocolHandler` would still be used?

My underlying question is, once `callCancelled` has been used below to choose the result status, is there any reason to set `callCancelled` back to `false`? I'm guessing not because it's the `close` method that uses `clientCancelled` but I just wanted to verify that.

(Same question about `headersSent`; would that _ever_ need to be set to `false` again? Again, I assume not but I'm just confirming.)

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -317,6 +315,7 @@ public void sendHeaders(Metadata headers) {
                                           streamId,
                                           HeaderFlags.create(END_OF_HEADERS),
                                           flowControl.outbound());
+                headersSent = true;
```
tjquinno, line 318: Question: Do we know for sure that `sendHeaders` will be invoked at most once, so there's no need to check `headersSent` at the beginning of the method to avoid sending them twice?

tjquinno (commented):
Possibly remove some now-unused constructor parameters. Also a few questions.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
spericas: We certainly can, it always seemed like objects we should be using but never did.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
spericas: Instances wouldn't be re-used, the underlying HTTP/2 stream is removed after that close method.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
spericas: Yes, here is the Javadoc for that method:

```
Send response header metadata prior to sending a response message. This method may only be called once and cannot be called after calls to sendMessage or close.
```

tjquinno (approved):

danielkec (approved):

PR #10137, opened 2025-05-20
4.x: Docs: Add SE logging guide

barchetta:
### Description

Documentation: Add SE logging guide.

See #9744

PR #10143, opened 2025-05-21
Enhance MDC to allow mapping a key to a supplier of string; enhance tracing to add `trace_id` 

### LINKED ISSUE #5741: Logging and tracing correlation
Add support for logging a trace id through Helidon MDC. This can be a separate module, or just update to existing tracing implementation, where we would add the top level span traceId to Helidon Logging MDC. Then users could use one of our logging integration modules to log the trace id with every single log statement.

This feature is important for correlation of traces and log statements (such as when using Verrazzano)

tjquinno:
### Description
Resolves #5741 

## Release note
____
If you use tracing, Helidon now automatically provides a logging MDC (mapped diagnostic context) key `trace_id` which you can use in logging messages to display the trace ID of the current span if there is one.

Also, Helidon's MDC provider implementation now permits MDC keys to be added with a `Supplier<String>` value or a `String`. (Previously only `String` values could be added.)
____

## PR Overview
(simplified since the original PR)

Two parts:

### Changes in logging
Previously, `HelidonMdc` (the main entry point for Helidon and user code to manage MDC) would simply delegate each method to the corresponding method on all `MdcProvider` instances. The providers would use their own context storage or context storage provided by the underlying logging implementation to store key/value pairs.

Now, `HelidonMdc` keeps its own `Map<String, Supplier<String>>` and first tries to resolve a look-up using that. If that map does not contain a supplier for a key, `HelidonMdc#get` delegates to the available MDC providers as it used to. 

The PR also includes changes to tests for the new supplier feature.

### Changes in tracing
The Helidon `TracerProviderHelper` class has always been initialized with whatever specific tracing provider library the user adds to her project. This init code now registers the `trace_id` MDC key with a supplier that gets the current span (if there is one) and returns the trace ID from it. 

This code uses the Helidon neutral tracing API to get the current span and, from it, the span context (which contains the trace ID. But this works because the Helidon neutral code to get the current span delegates to the underlying implementation (OpenTelemetry, Zipkin, etc.) so the `Supplier<String>` will always retrieve the current span, even if user code has set the current span using the underlying library's API, not the Helidon one.

The PR also adds a test to the provider tests that all providers run. It tests the MDC `trace_id`  functionality.

### Documentation
Depending on timing, PR #10137 could include a note about this or the new logging page that PR creates could be separately updated to describe the new `tracer_id` supported added by this PR.

### DIFF
--- a/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
+++ b/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
@@ -69,6 +69,19 @@ public class HelidonMdc {
     }
 
     /**
+     * Sets a value supplier <em>without</em> immediately getting the value and propagating the value to
+     * underlying logging implementations.
+     * <p>
+     * Normally, user code should use {@link #set(String, java.util.function.Supplier)} instead.
+     *
+     * @param key entry key
+     * @param valueSupplier  supplier of the entry value
+     */
+    public static void setDeferred(String key, Supplier<String> valueSupplier) {
+        SUPPLIERS.get().put(key, valueSupplier);
+    }
+
+    /**
      * Remove value with the specific key from all of the instances of {@link MdcProvider}.
      *
      * @param key key
--- a/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
+++ b/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
@@ -54,7 +54,6 @@ public class HelidonMdc {
      * @param value entry value
      */
     public static void set(String key, String value) {
-        SUPPLIERS.get().put(key, value::toString);
         MDC_PROVIDERS.forEach(provider -> provider.put(key, value));
     }
 
--- a/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
+++ b/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
@@ -32,9 +32,9 @@ import io.helidon.logging.common.spi.MdcProvider;
  * Helidon permits adding MDC entries using {@code Supplier<String>} values as well as direct {@code String} values.
  * Although some logging implementations provide their own context maps (for example {@code ThreadContext} in Log4J and
  * {@code MDC} in SLF4J), they map MDC keys to {@code String} values, not to arbitrary objects that would accommodate
- * {@code Supplier<String>}. Therefore, Helidon manages its own map of key/supplier pairs and resolves all lookups using that map.
- * <p>
- * Helidon also propagates key/string pair assignments to the logging implementations' context maps.
+ * {@code Supplier<String>}. Therefore, Helidon not only propagates every {@code set} operation to the loaded MDC providers,
+ * but also manages its own map of key/supplier pairs. Helidon resolves each lookup using that map
+ * if possible, delegating a look-up to the loaded MDC providers only if there is no supplier for a key.
  */
 public class HelidonMdc {
 
--- a/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
+++ b/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
@@ -18,6 +18,7 @@ package io.helidon.logging.common;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.function.Supplier;
@@ -106,7 +107,16 @@ public class HelidonMdc {
      * @return found value bound to key
      */
     public static Optional<String> get(String key) {
-        return SUPPLIERS.get().containsKey(key) ? Optional.of(SUPPLIERS.get().get(key).get()) : Optional.empty();
+        /*
+        User or 3rd-party code might have added values directly to the logger's own context store. So look in other
+        providers if our data structure cannot resolve the key.
+         */
+        return SUPPLIERS.get().containsKey(key)
+                ? Optional.of(SUPPLIERS.get().get(key).get())
+                : MDC_PROVIDERS.stream()
+                        .map(provider -> provider.get(key))
+                        .filter(Objects::nonNull)
+                        .findFirst();
     }
 
     static Map<String, Supplier<String>> suppliers() {
--- a/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
+++ b/logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.Map;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import io.helidon.logging.common.spi.MdcProvider;
+
+/**
+ * Implements {@link io.helidon.logging.common.spi.MdcProvider} using an internal thread local data structure.
+ */
+public class BaseMdcProvider implements MdcProvider {
```
tomas-langer, line 29: This class should be abstract.

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
```diff
@@ -0,0 +1,114 @@
...
+    static. That's because each instance should represent a different provider type and there should be exactly one instance
+    for each provider type. This way each different provider type has its own thread local map in case, for some reason,
+    the maps for different providers need to be different.
+     */
+    private final ThreadLocal<Map<String, Supplier<String>>> mdcSettings = ThreadLocal.withInitial(HashMap::new);
+
+    /**
+     * Creates a new instance.
+     */
+    public BaseMdcProvider() {
```
tomas-langer, line 42: Constructor should be protected.

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
```diff
@@ -0,0 +1,114 @@
...
+        mdcSettings.get().remove(key);
+    }
+
+    @Override
+    public void clear() {
+        mdcSettings.get().clear();
+    }
+
+    @Override
+    public String get(String key) {
```
tomas-langer, line 66: This is unfortunate, but as you say, it is the existing behavior :(

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
```diff
@@ -0,0 +1,114 @@
...
+        // earlier implementations.
+        return mdcSettings.get().containsKey(key) ? mdcSettings.get().get(key).get() : null;
+    }
+
+    /**
+     * Returns the current key/supplier MDC settings.
+     *
+     * @return copy of the current key/supplier map containing the current MDC settings
+     */
+    public Map<String, Supplier<String>> settings() {
```
tomas-langer, line 77: Settings feels like we are configuring something.
Mabe just `suppliers`?

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class BaseMdcProvider implements MdcProvider {
+
+    /*
+    Each instance of this class gets its own thread local map rather than usual practice of declaring thread local fields as
+    static. That's because each instance should represent a different provider type and there should be exactly one instance
+    for each provider type. This way each different provider type has its own thread local map in case, for some reason,
+    the maps for different providers need to be different.
+     */
+    private final ThreadLocal<Map<String, Supplier<String>>> mdcSettings = ThreadLocal.withInitial(HashMap::new);
```
tomas-langer, line 37: Thread local should be (by definition) static.

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
tomas-langer, line 77: These methods should not be public - they should only be used within the same package (i.e. `protected` here).

logging/common/src/main/java/io/helidon/logging/common/HelidonMdc.java
```diff
@@ -74,4 +85,20 @@ public static Optional<String> get(String key) {
...
 
+    /**
+     * Returns the MDC provider of the specified type, if any was found.
+     *
+     * @param providerType type of provider to find
+     * @return {@link io.helidon.logging.common.spi.MdcProvider} of the requested type
+     * @param <T> type of the provider
+     * @throws java.util.NoSuchElementException if Helidon found no MDC provider of the requested type
+     */
+    static <T extends MdcProvider> T provider(Class<T> providerType) {
```
tomas-langer, line 96: dtto - this should not exist.

logging/jul/src/main/java/io/helidon/logging/jul/JulMdc.java
```diff
@@ -15,22 +15,24 @@
...
 
-    private static final ThreadLocal<Map<String, String>> MDC_PROPERTIES = ThreadLocal.withInitial(HashMap::new);
-
     private JulMdc() {
         throw new IllegalStateException("This class cannot be instantiated");
     }
 
     static void put(String key, String value) {
-        MDC_PROPERTIES.get().put(key, value);
+        JulMdcProvider.instance().put(key, value);
```
tomas-langer, line 31: Do not use a singleton in any of the implementations. All methods are static and should be used as such.

tomas-langer (changes requested):

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
tjquinno: I think I see a way of reorganizing the code a bit that avoids duplicating code while addressing your comments. I'll do that.

That said, you didn't describe what problems you see this introducing.

I'm not sure how there could be one `MdcProvider` system-wide; the `HelidonMdc` code specifically allows for and invokes however many providers are found by service loading.

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
tjquinno: TL;DR - This will no longer be an issue if my planned restructuring works as I think it will.

____
Thread local is not static by definition, but certainly in most cases is it static.

As the comments in the code explain, in this case I was looking to maintain a separate thread local map for each provider. With that in mind, it made sense to have the thread local map be an instance field. 

After thinking about this some more we really do not need separate thread local maps for each provider, so in the restructuring I have in mind `HelidonMdc` will maintain a single thread local map and it will be static there.

tjquinno: No longer an issue with the rewrite of the PR.

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
tjquinno: No longer using a common base provider.

logging/common/src/main/java/io/helidon/logging/common/BaseMdcProvider.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    /**
+     * Returns the singleton instance of the MDC provider of the specified type.
+     *
+     * @param providerType type of provider for which the instance is requested
+     * @return provider of the specified type
+     * @param <T> type of the provider
+     */
+    protected static <T extends MdcProvider> T provider(Class<T> providerType) {
```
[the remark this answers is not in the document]
tjquinno: No longer using a common base provider.

logging/jul/src/main/java/io/helidon/logging/jul/JulMdc.java
tjquinno: No longer an issue with the changes to the PR.

tracing/provider-tests/src/main/java/io/helidon/tracing/providers/tests/TestMdc.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+
+class TestMdc {
+
+    @BeforeAll
+    static void beforeAll() throws IOException {
+        String loggingConfig = "# HelidonConsoleHandler uses a SimpleFormatter subclass that replaces \"!thread!\" with the current thread\n"
```
tomas-langer, line 42: I think that we do not need the comment for a test setup

tomas-langer (approved):

Issue #10109, opened 2025-05-12
Using gRPC tool ghz can result in connection closed issues


spericas:
## Problem Description

Using `ghz` as a client to test a Helidon SE (maybe MP as well) endpoint with high concurrency can result in scenerarios in which Helidon is attempting to write to a connetion that has been closed.

## Steps to reproduce

Run `ghz` with for a longer than the default 200 iterations. Sample stack trace:

```
May 12, 2025 2:14:21 PM io.helidon.grpc.core.SafeStreamObserver onCompleted
SEVERE: Caught exception handling onComplete
java.io.UncheckedIOException: java.net.SocketException: Socket closed
        at io.helidon.common.buffers.FixedBufferData.writeTo(FixedBufferData.java:74)
        at io.helidon.common.buffers.CompositeArrayBufferData.writeTo(CompositeArrayBufferData.java:48)
        at io.helidon.common.socket.PlainSocket.write(PlainSocket.java:136)
        at io.helidon.common.socket.SocketWriter.writeNow(SocketWriter.java:81)
        at io.helidon.http.http2.Http2ConnectionWriter.noLockWrite(Http2ConnectionWriter.java:200)
        at io.helidon.http.http2.Http2ConnectionWriter.writeHeaders(Http2ConnectionWriter.java:93)
        at io.helidon.webserver.grpc.GrpcProtocolHandler$1.close(GrpcProtocolHandler.java:374)
        at io.grpc.stub.ServerCalls$ServerCallStreamObserverImpl.onCompleted(ServerCalls.java:395)
        at io.helidon.grpc.core.SafeStreamObserver.onCompleted(SafeStreamObserver.java:83)
        at io.helidon.grpc.core.ResponseHelper.complete(ResponseHelper.java:53)
        at io.helidon.examples.webserver.grpc.StringService.upper(StringService.java:51)
        at io.grpc.stub.ServerCalls$UnaryServerCallHandler$UnaryServerCallListener.onHalfClose(ServerCalls.java:182)
        at io.grpc.PartialForwardingServerCallListener.onHalfClose(PartialForwardingServerCallListener.java:35)
        at io.grpc.ForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:23)
        at io.grpc.ForwardingServerCallListener$SimpleForwardingServerCallListener.onHalfClose(ForwardingServerCallListener.java:40)
[...]

PR #10147, opened 2025-05-22
gRPC Improvements

spericas:
### Description

Some improvements:

1. Avoids some buffer copies when sending and receiving gRPC payloads
2. New config property to completely disable compression that can be useful for perf testing

```
server:
  protocols:
    grpc:
      enable-compression: false
```

3. Ensures underlying HTTP/2 streams are properly removed from data structures after reaching CLOSED state
4. Logs all socket exceptions (e.g. broken pipes) in log level TRACE (see issue #10109)
5.  Proper synchronization for when an RST frame is received at connection level

### Documentation

None

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -168,6 +168,7 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
     public void rstStream(Http2RstStream rstStream) {
         callCancelled = (rstStream.errorCode() == Http2ErrorCode.CANCEL);
         listener.onCancel();
+        currentStreamState = nextStreamState(Http2StreamState.HALF_CLOSED_REMOTE);
     }
 
     @Override
@@ -210,8 +211,7 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
             // if EOS then half close remote
             if (header.flags(Http2FrameTypes.DATA).endOfStream()) {
                 listener.onHalfClose();
-                currentStreamState = Http2StreamState.HALF_CLOSED_REMOTE;
-
+                currentStreamState = nextStreamState(Http2StreamState.HALF_CLOSED_REMOTE);
                 // update metrics
                 if (grpcConfig.enableMetrics()) {
                     methodMetrics.recvMessageSize.record(bytesReceived);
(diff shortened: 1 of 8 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -167,6 +168,7 @@ public Http2StreamState streamState() {
     public void rstStream(Http2RstStream rstStream) {
         callCancelled = (rstStream.errorCode() == Http2ErrorCode.CANCEL);
         listener.onCancel();
+        currentStreamState = nextStreamState(Http2StreamState.HALF_CLOSED_REMOTE);
     }
```
danielkec, line 172: I don't think we have rstStream actually connected to anything yet, but when we do, it would be invoked from connection thread right?

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -160,13 +164,20 @@ public void init() {
...
+     * thread, so proper synchronization is required.
+     *
+     * @param rstStream RST stream frame
+     */
     @Override
     public void rstStream(Http2RstStream rstStream) {
-        callCancelled = (rstStream.errorCode() == Http2ErrorCode.CANCEL);
+        callCancelled.set(rstStream.errorCode() == Http2ErrorCode.CANCEL);
         listener.onCancel();
+        currentStreamState.set(nextStreamState(Http2StreamState.HALF_CLOSED_REMOTE));
```
danielkec, line 180:
```suggestion
        currentStreamState.updateAndGet(prevState -> nextStreamState(prevState, Http2StreamState.HALF_CLOSED_REMOTE));
```

danielkec: CAS shortcut to avoid RC

spericas: That looks better

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -103,19 +107,19 @@ private record MethodMetrics(Counter callStarted,
...
-    private boolean isIdentityCompressor;
+    private boolean identityCompressor;
     private long bytesReceived;
     private MethodMetrics methodMetrics;
     private long startMillis;
-    private boolean callCancelled;
 
-    GrpcProtocolHandler(HttpPrologue prologue,
-                        Http2Headers headers,
+    private final AtomicBoolean callCancelled = new AtomicBoolean();
```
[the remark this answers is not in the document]
spericas: Yes, there should only be one

danielkec (approved):

PR #10149, opened 2025-05-22
Helidon data update

### LINKED ISSUE #10141: Service registry support for Jakarta Persistence
Persistence unit context (EntityManaferFactory) and persistence session (EntityManager) should be available in service registry.
It should be created from Config node with following structure:

```
persistence-units:
  jakarta:
    - name: "PU-name"
      data-source: "test"
      provider-class-name: "org.eclipse.persistence.jpa.PersistenceProvider"
      init-script: "init-database.sql"
      drop-script: "drop-database.sql"
      properties:
        jakarta.persistence.schema-generation.database.action: drop-and-create
```

tomas-langer:
### Description
Resolves #10141

The test configuration is modified, see data/README.md

Simplified configuration of tests to support update of ports (where we need to start test containers on random ports, but configuration contains the default port of the database).

Separated data sources configuration and persistence unit configuration.

Removed `DataRegistry` and the root `data` configuration node, as it is no longer needed.

Simplified codegen, each repository now generates a single class (the implementation of the repository interface), and of course its service descriptor (we used to generate two).


Tomas-Kraus (approved):
Let's merge it. Cosmetic things can be fixed in next commit.

PR #10157, opened 2025-05-27
Helidon Data: update to configuration

tomas-langer:
Using root config node `data` for all new configurations (config sources, and persistence units)


Tomas-Kraus (approved):
LGTM

PR #10158, opened 2025-05-27
Fix for #10154 Graceful shutdown

### LINKED ISSUE #10154: 4.x WebServer Regression - Graceful shutdown broken by CRaC support changes
## Environment Details
* Helidon Version: 4.2.2
* Helidon SE
* JDK version: 21.0.5
* OS: MacOS

----------

## Problem Description

On `WebServer.stop()` there should be "Graceful Shutdown" waiting for active inflight requests to complete with the given `gracePeriod` [which defaults to 0.5 seconds].

The `gracePeriod` is no longer honoured, due to a change introduced by the support of CRaC feature.

The commit https://github.com/helidon-io/helidon/commit/ffb25fdea7c04baef00753fb0c0928d678deda44#diff-154ddda49d07442fa78e7da6ba0405cb5a6d3540a1e404bfe84b7114e562e8c0R242

... added the below 2 lines into `ServerListener.suspend(boolean shutdownExecutors)` method: 
```java
// Close all active connections
activeConnections().forEach(connection -> connection.close(true));
```
... this executes **_BEFORE_** the graceful shutdown of the `readerExecutor` and `sharedExecutor` which results in inflight active requests being interrupted (and no gracePeriod is honoured to allow the inflight requests to complete).

## Steps to reproduce
- Create a WebServer with a relatively slow endpoint (Say takes 5 seconds)
- Start the WebServer (with gracePeriod increased to 10 seconds to allow human interaction to simulate the issue)
- Hit the slow endpoint (to create an inflight request that we want to allow to gracefully complete)
- Stop the WebServer before the slow inflight request has completed

### Desired behaviour:
On WebServer.stop() the WebServer waits for the gracePeriod for the inflight request to complete.

### Observed behaviour:
On WebServer.stop() the WebServer immediately interrupts any inflight requests not honouring the configured gracePeriod.

# Suggested Fix options

### Suggestion 1. Close activeConnections only when !shutdownExecutors

```java
if (!shutdownExecutors) {
    // Close all active connections
    activeConnections().forEach(connection -> connection.close(true));
}
```
... with the thinking that closing all active connections is really only needed for the CRaC suspend case. 

### Suggestion 2. Move Close activeConnections to execute after shutdownExecutors completes

This allows the graceful shutdown to complete FIRST, and then after that close active connections. 

Note that both of these options fix my local test and Graceful shutdown is honoured.

rbygrave:
Fix #10154 by making the Close all active connections conditional on !shutdownExecutors, as that only needs to execute with CRaC suspend.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
@@ -238,10 +238,10 @@ class ServerListener implements ListenerContext {
         try {
             // Stop listening for connections
             serverSocket.close();
-            // Close all active connections
-            activeConnections().forEach(connection -> connection.close(true));
-
-            if (shutdownExecutors) {
+            if (!shutdownExecutors) {
+                // Close all active connections
+                activeConnections().forEach(connection -> connection.close(true));
+            } else {
                 // Shutdown reader executor
                 readerExecutor.terminate(gracePeriod.toMillis(), TimeUnit.MILLISECONDS);
                 if (!readerExecutor.isTerminated()) {


rbygrave:
>  ... this.canRun property inside Http1Connection.close() set to false.

Yes, I agree.


> Issue https://github.com/helidon-io/helidon/issues/10156sees to this and ensures that it does not accept any new request while considering the configured value of graceful shutdown.

That might be ok. I think the `Thread.sleep(gracePeriod)` looks more like a "shutdown delay" and isn't as desirable as the existing mechanism in ServerListener.suspend() that will complete faster / complete as soon as the inflight requests complete.


I've adjusted this PR to first use `close(false)` on the active connections [setting the canRun to false] ... and then AFTER the graceful completion of inflight requests perform `close(true)` on any active connections that remain.

rbygrave:
Is this issue waiting for some other approval or review? I'm pretty keen to see if this could be included in a release soon as we are seeing errors in K8s rollouts without this fix in place. If there is anything else I need to do please let me know.
Thanks, Rob.

rbygrave:
Any thoughts on a review for this issue?

rbygrave:
Just noting this didn't make the 4.2.7 release. Hopefully makes the next one.

romain-grecourt:
This has been merged in the main line, so it will be included in 4.3.0.
If we want this in 4.2.x we need to backport it.

Issue #10159, opened 2025-05-27
3.x, 4.x: Path matcher with pattern to match subpaths not matching as expected


tomas-langer:
## Environment Details
* Helidon Version: 3.x, 4.x
* Helidon SE
* JDK version: any
* OS: Ubuntu Linux

----------

Tasks:

- [x] 3.x examples fix - nothing to do
- [x] https://github.com/helidon-io/helidon-examples/issues/159

Pattern `/all/subpaths[/{*}]` is expected to match `/all/subpaths`, `/all/subpaths/`, and `/all/subpaths/someOther`.

When used in Helidon 3 and 4, the trailing slash pattern is not matched.

Also any additional path nesting will not be matched (not sure if the documentation describes this as something that should be matched). `/all/subpaths/level/level2` would not match as well.

In Helidon 4.x, we have a soluton by using `/all/subpaths/*`, which will match all of the above correctly.

Test to validate (4.x):

```java
       String pattern = "/all/subpaths[/{*}]";

        PathMatcher matcher = PathMatchers.create(pattern);

        PathMatchers.MatchResult match = matcher.match(UriPath.create("/all/subpaths"));
        assertThat("exact match", match.accepted(), is(true));

        match = matcher.match(UriPath.create("/all/subpaths/subpath/other"));
        assertThat("subpath match", match.accepted(), is(true));

        match = matcher.match(UriPath.create("/all/subpaths/"));
        assertThat("trailing slash match", match.accepted(), is(true));
```

Test to validate (3.x):
```java
        String pattern = "/all/subpaths[/{*}]";

        PathMatcher matcher = PathMatcher.create(pattern);

        PathMatcher.Result match = matcher.match("/all/subpaths");
        assertThat("exact match", match.matches(), is(true));

        match = matcher.match("/all/subpaths/subpath/other");
        assertThat("subpath match", match.matches(), is(true));

        match = matcher.match("/all/subpaths/");
        assertThat("trailing slash match", match.matches(), is(true));
```

tomas-langer:
These tests should be added to both versions. Also in 4.x, we need to add the following test (unless it already exists):

(this tests passes as expected in 4.x)

```java
String pattern = "/all/subpaths/*";

PathMatcher matcher = PathMatchers.create(pattern);

PathMatchers.MatchResult match = matcher.match(UriPath.create("/all/subpaths"));
assertThat("exact match", match.accepted(), is(true));

[...]

PR #10162, opened 2025-05-27
4.x: Path pattern fix for WebServer

tomas-langer:
Fix path pattern `{*}` to have the meaning we assigned to it in documentation.

Updated documentation and javadocs to use the correct patterns for "any path" using the 4.0 pattern of `/*` Added tests for the patterns.

#10159 for 4.x `main` branch

Resolves https://github.com/helidon-io/helidon/issues/10163

### DIFF
--- a/docs/src/main/asciidoc/se/webserver/webserver.adoc
+++ b/docs/src/main/asciidoc/se/webserver/webserver.adoc
@@ -248,6 +248,7 @@ You can use *path pattern* instead of _path_ with the following syntax:
 * `/foo/{:\d+}` - Unnamed regular expression segment with a specified expression
 * `/foo/{+var}` - Convenience shortcut for `{var:.+}`
 * `/foo/{+}` - Convenience shortcut for unnamed segment with regular expression `{:.+}`
+* `/foo/{*}` - Convenience shortcut for unnamed segment with regular expression `{:.*}`
 * `/foo[/bar]` - An optional block, which translates to the `/foo(/bar)?` regular expression
 * `/*` or `/foo*` - `*` Wildcard character can be matched with any number of characters.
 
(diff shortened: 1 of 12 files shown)


docs/src/main/asciidoc/se/webserver/webserver.adoc
```diff
@@ -248,6 +248,7 @@ You can use *path pattern* instead of _path_ with the following syntax:
 * `/foo/{:\d+}` - Unnamed regular expression segment with a specified expression
 * `/foo/{+var}` - Convenience shortcut for `{var:.+}`
 * `/foo/{+}` - Convenience shortcut for unnamed segment with regular expression `{:.+}`
```
spericas, line 250: Not directly related to this PR, but the "{+}" seems to be rendered as "{}". Should it be set to "{\+}"?

http/http/src/test/java/io/helidon/http/PathMatchersTest.java
```diff
@@ -32,6 +32,83 @@
...
+        assertThat("trailing slash match", match.accepted(), is(true));
+
+        match = matcher.match(UriPath.create("/all/subpathsbad"));
+        assertThat("trailing slash match", match.accepted(), is(false));
+    }
+
+    @Test
+    void testSubpathPattern() {
+        String pattern = "/all/subpaths[/{*}]";
+
```
spericas, line 77: I'm a bit confused. Aren't we saying in the doc above that `{*}` is the same as `{:.*}`? If so, why is this test's outcome different from the previous one?

tomas-langer: Will check and fix

spericas (approved):

Issue #10012, opened 2025-04-15
OCI Secrets as a Config does not work dynamically


klustria:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP

----------

## Problem Description
There are 2 issues:
1. Setting up OCI Secrets as a Config with polling-strategy to achieve dynamic configuration  behavior does not work and fails with this error:
    ```
    io.helidon.config.ConfigException: You are attempting to configure a polling strategy on a source builder that does not support it:      io.helidon.integrations.oci.secrets.configsource.SecretBundleNodeConfigSource$Builder
        at io.helidon.config.AbstractSourceBuilder.pollingStrategy(AbstractSourceBuilder.java:144)
        at  io.helidon.integrations.oci.secrets.configsource.SecretBundleNodeConfigSource$Builder.pollingStrategy(SecretBundleNodeConfigSource.java:429)
        at   io.helidon.integrations.oci.secrets.configsource.SecretBundleNodeConfigSource$Builder.pollingStrategy(SecretBundleNodeConfigSource.java:354)
        at java.base/java.util.Optional.ifPresent(Optional.java:178)
        at io.helidon.common.mapper.OptionalValue.ifPresent(OptionalValue.java:188)
        at io.helidon.config.AbstractSourceBuilder.config(AbstractSourceBuilder.java:95)
        at io.helidon.config.AbstractConfigSourceBuilder.config(AbstractConfigSourceBuilder.java:79)
    ```
2. If the above issue is resolved, a new intermittent issue happens wherein accessing the config property sometimes causes `NoSuchElementException` when the OCI secret is transitioning from Active to Updating state.

## Steps to reproduce
1. For the 1st issue, create an mp-meta-config.yaml with the following oci-secrets type:
    ```
     - type: 'oci-secrets'
       optional: true
       accept-pattern: '^database.password$'
       compartment-ocid: "ocid1.compartment.oc1..dummycompartmentocid"
       lazy: false
       vault-ocid: "ocid1.vault.oc1.iad.dummyvaultocid"
       polling-strategy:
         type: "regular"
         properties:
           interval: "PT5S"
    ```
2. For the 2nd issue, resolve the problem in the 1st issue, then retrieve the config that references an OCI secret. Make sure that there is only 1 secret in the Vault. Then update the OCI secret to a new value and retrieve it while it is updating. Using this MP code 
    ```
    public class GreetingProvider {
[...]

PR #10172, opened 2025-05-28
3.x: Adjusts OCI vault config sources to read only secret versions that are extant and in the Current rotation state

Captain1653:
### Description
It's backport for 3.x from [4.x](https://github.com/helidon-io/helidon/pull/10035) . As I can see, it can be done.

It's part of [this issue](https://github.com/helidon-io/helidon/issues/10012).

Should I create the separate issue for 3.x? Or it shouldn't be backported?


ljnelson (approved):
LGTM

PR #10177, opened 2025-05-29
Add support for `metrics.rest-request.enabled` (with dot instead of hyphen) in SE metrics config

### LINKED ISSUE #10168: Bug in metrics documentation
Helidon Version:
Documentation URL: https://helidon.io/docs/v4/mp/metrics/metrics

## Problem Description

The docs says that REST.request metrics are enabled using ```metrics.rest-request-enabled=true```

The correct configuration is ```metrics.rest-request.enabled=true```

tjquinno:
### Description
Resolves #10168 along with PR #10173.

## Release note
____
Helidon 4.x SE did not properly recognize the config key `metrics.rest-request.enabled` to turn on automatic REST request metrics. Instead the config key Helidon responded to was `metrics.rest-request-enabled` (using a hyphen instead of a dot separator). Helidon SE now honors both forms, although the `rest-request-enabled` form is deprecated and will likely be removed in a future major release.
____

## PR Notes
Helidon 4.x MP _does_ correctly recognize `metrics.rest-request.enabled` in `META-INF/microprofile-config.properties`. (The MP code has not used the generated `MetricConfig` class; rather its code directly accessed the MP config setting.)

Helidon 3.x SE and MP properly recognized the "dotted" key.

The PR:
* Explicitly sets the key for the `restRequestEnabled` blueprint method to `rest-request.enabled` (rather than letting it default).
* Adds a deprecated config setting with key `rest-request-enabled` to remain compatible with 4.x SE apps that might already be configured using the hyphen form of the key.

Any assignment of the deprecated setting--either via config using `rest-request-enabled` or via the new deprecated method on the builder--also assign the same value to the non-deprecated setting.

### Documentation
See #10173 for the related doc update.

### DIFF
--- a/metrics/api/pom.xml
+++ b/metrics/api/pom.xml
@@ -72,6 +72,11 @@
         </dependency>
         <dependency>
             <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
             <artifactId>helidon-config-yaml</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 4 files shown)


tomas-langer (approved):
I assume the change of the docs is going to be merged as part of another PR.

PR #10189, opened 2025-06-02
Add tracing support to concurrenty limits

### LINKED ISSUE #9968: Tracing and Accesslog support for time spent by a request waiting in the queue trying to acquire the Limit
## Environment Details
* Helidon Version: 4.x 
* Helidon SE  and  Helidon MP

----------

## Enhancement Description
Helidon 4.2.0 adds a Concurrency limit feature with Fixed and AIMD Limit , to  limit the concurrency with support for queueing additional requests.    While there are metrics  to monitor the overall queued requests and concurrency , currently there is.  no easy way to  get the data  on the time spent by a request waiting for the permit.  

The ask is to add 
1.  Tracing Span for  time spent by a request waiting for a permit 
2. Provide a field  in AccessLog to ,  to record the time spent by request waiting on the queue , This can be separate field and not added to server time.

tjquinno:
### Description
Resolves #9968 

Adds a tracing span each time a concurrency-limited request is queued.

### Documentation
Will be added as part of this PR.


common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
```diff
@@ -103,13 +107,32 @@ Optional<LimitAlgorithm.Token> tryAcquire(boolean wait) {
         }
         if (wait && queueLength > 0) {
             long startWait = clock.get();
+            Span.Builder<?> spanBuilder = null;
```
tomas-langer: Could this block be done as follows?
I think there are too many ifs...

```java
Span.Builder<?> spanBuilder = null;
            Span span;
            Scope scope;

            if (tracer == null) {
                span == null;
                scope = null;
            } else {
                spanBuilder = tracer.spanBuilder("limits-wait");
                var currentSpan = Span.current();
                if (currentSpan.isPresent()) {
                    currentSpan.get().context().asParent(spanBuilder);
                }
                span = spanBuilder.start();
                scope = span.activate();
            }

            token = handler.tryAcquire(true);
            if (token.isPresent()) {
                if (queueWaitTimer != null) {
                    queueWaitTimer.record(clock.get() - startWait, TimeUnit.NANOSECONDS);
                }
                if (scope != null) {
                    scope.close();
                    span.end();
                }
                return token;
            }
            if (scope != null) {
                span.status(Span.Status.ERROR);
                // shouldn't we log something with the span?
                scope.close();
                span.end();
            }
```

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimit.java
```diff
@@ -170,13 +174,32 @@ public Optional<Token> tryAcquire(boolean wait) {
         }
         if (wait && queueLength > 0) {
             long startWait = clock.get();
+            Span.Builder<?> spanBuilder = null;
```
tomas-langer: Same as in AimdLimit

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
```diff
@@ -103,13 +107,32 @@ Optional<LimitAlgorithm.Token> tryAcquire(boolean wait) {
...
+            if (tracer != null) {
+                spanBuilder = tracer.spanBuilder("limits-wait");
```
tomas-langer: Maybe use the limit type? i.e. `aimd-limit-wait` and `fixed-limit-wait`

tomas-langer (commented):
Except for the structure of conditions, this seems like a sound solution.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
tjquinno: I refactored this using a package-visible `record` that encapsulates all this logic. Also radically reduces the duplicated code in the two limits implementation classes by using the `record` from each.

Also added a log message (an event) to the span in the error case.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimit.java
tjquinno: Done.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
tjquinno: done.

spericas (approved):
LGTM

vasanth-bhat:
Is this going to be available in a 4.2.x release?

tjquinno:
Most likely, no.

We follow the [semantic versioning conventions](https://semver.org/) in which dot-dot releases are bug fixes. This change is an enhancement and will be in 4.3.0.

PR #10198, opened 2025-06-03
Manage REST request metrics for resource methods on a superclass

### LINKED ISSUE #10196: 4.x MP REST request metrics do not include inherited REST methods
## Environment Details
* Helidon Version: 4.x
* Helidon MP 
* JDK version:
* OS:

----------

## Problem Description
If a REST resource class inherits from another class that has a REST method, Jakarta REST (correctly) treats the inherited method as if it is declared on the subclass. But there is no `REST.request` metric created for the inherited method.

## Steps to reproduce
See [this StackOverflow entry](https://stackoverflow.com/questions/79648637/jax-rs-method-defined-in-parent-class-not-exposed-under-base-rest-request) for the idea. (It's not fully-correct Java but it communicates the idea).

From the reproducing test we'll be adding to Helidon (these are inner classes in the test class so they are `static` but that does not affect the problem):

```java
    public static abstract class BaseService {

        @GET
        @Path("/base")
        @Produces(MediaType.TEXT_PLAIN)
        public String getFromBase() {
            return "base";
        }
    }

    @RequestScoped
    @Path("/testRestResource")
    public static class ConcreteService extends BaseService {

        @GET
        @Path("/concrete")
        @Produces(MediaType.TEXT_PLAIN)
        public String getFromConcrete() {
            return "concrete";
        }
    }
```

The base `MetricRegistry` contains a `REST.request` metric only for the `getFromConcrete` method. It should also have one for the inherited `getFromBase` method.

In the test code, using the debugger to evaluate this expression
```java
metricRegistry.getMetrics((metricID, metric) -> 
        metricID.getName().equals("REST.request") 
        && metricID.getTags().get("class").endsWith("ConcreteService"))
```
yields
```list
{MetricID@10556} "MetricID{name='REST.request', 
    tags=[class="io.helidon.microprofile.metrics.ConcreteService",method="concrete"]}" -> 
    HelidonTimer@10557} "HelidonTimer{registryType='base', 
        metadata=DefaultMetadata{name='REST.request',
            unit='NANOSECONDS', 
            description='The number of invocations and total response time of this RESTful resource method since the start of the server. The metric will not record the elapsed time nor count of a REST request if it resulted in an unmapped exception. Also tracks the highest recorded time duration and the 50th, 75th, 95th, 98th, 99th and 99.9th percentile.'}, 
    count='0', 
    elapsedTime='PT0S'}"
```
That's a single timer, not the two there should be; the timer for the inherited method is missing.

tjquinno:
### Description
Resolves #10196

## Release note
____
Helidon MP, if so configured, automatically measures REST request methods. Helidon MP did not--but now does--register and update these metrics for REST resource methods _inherited from a superclass_. 
____

## PR Overview
Formerly, the metrics to be updated for REST request measurements failed to handle inherited methods in two ways:
* REST request methods that were inherited from an abstract superclass were simply ignored during annotation handling for REST resource methods.
* The REST request metrics were registered with a metric ID tag based on the _declaring_ class of the method rather than the _executing_ class.

This PR fixes both of those problems (both in the metrics CDI extension).

A data structure holding REST request methods is now also qualified by the executing class (instead of just the `Method` which ties to the declaring class) so, when the data structure is used later to actually register metrics, the correct execution class is available (not just the declaring class of each method).

Several non-public methods gained a new `Class` parameter. Internal uses in the extension class and several tests changed to align with the new signatures.

The PR also includes a new test to check that the inheritance use case works.

### Documentation
bug fix - no doc impact

### DIFF
--- a/microprofile/metrics/pom.xml
+++ b/microprofile/metrics/pom.xml
@@ -131,6 +131,7 @@
                                 <exclude>**/TestSelectivelyDisabledMetrics.java</exclude>
                                 <exclude>**/TestConfigProcessing.java</exclude>
                                 <exclude>**/TestDistributionCustomizationsNoInit.java</exclude>
+                                <exclude>**/TestRestServiceInheritance.java</exclude>
                             </excludes>
                             <systemPropertyVariables>
                                 <junit.jupiter.extensions.autodetection.enabled>true</junit.jupiter.extensions.autodetection.enabled>
@@ -206,6 +207,17 @@
                         <configuration>
                             <includes>
                                 <include>**/TestDistributionCustomizationsNoInit.java</include>
+                            </includes>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>test-rest-service-inheritance</id>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <includes>
+                                <include>**/TestRestServiceInheritance.java</include>
                             </includes>
                         </configuration>
                     </execution>
--- a/microprofile/metrics/pom.xml
+++ b/microprofile/metrics/pom.xml
@@ -219,6 +219,7 @@
                             <includes>
                                 <include>**/TestRestServiceInheritance.java</include>
                             </includes>
+                            <reuseForks>false</reuseForks>
                         </configuration>
                     </execution>
                 </executions>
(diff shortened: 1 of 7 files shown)


spericas (approved):
LGTM

PR #10214, opened 2025-06-05
[helidon-data] API Cleanup and TypeName constants tests in codegen modules

Tomas-Kraus:


data/data/src/main/java/io/helidon/data/SortSupport.java
```diff
@@ -18,52 +18,75 @@
...
+                .build();
+    }
 
     /**
      * Whether any order definitions are set.
      *
      * @return value of {@code true} when at least one order definition is set or {@code false} otherwise
      */
-    boolean isSorted();
+    @Prototype.PrototypeMethod
```
tomas-langer: This can be defined as a `default` method on the blueprint I think.

data/data/src/main/java/io/helidon/data/SortBlueprint.java
```diff
@@ -15,21 +15,25 @@
...
-    public boolean isSorted() {
-        return orderBy.isEmpty();
-    }
+    /**
+     * Order definitions for the query result.
+     * Default value is an empty {@link List} to return unsorted result.
+     *
+     * @return {@link List} of order definitions
+     */
+    @Option.Default("UNSORTED")
```
tomas-langer, line 37: Add also `@Option.Singular`

data/data/src/main/java/io/helidon/data/SortSupport.java
Tomas-Kraus: Done.

data/data/src/main/java/io/helidon/data/SortBlueprint.java
Tomas-Kraus: Done

PR #10228, opened 2025-06-09
4.x: Support for varargs in common types

### LINKED ISSUE #10194: 4.x: Builder support for vararg factory methods
Currently when the following method is declared:

```java
@Prototype.FactoryMethod
static Sort create(Order... orders) {
}
```

it gets generated as 
```java
static Sort create(Order[] orders) {
}
```

in the prototype. This should honor the vararg argument

tomas-langer:
### Description
Resolves #10194 

### Documentation
Support for varargs in analyzed types and in builder codegen.

Java has `vararg` as a property of a method (i.e. the method is vararg, not the parameter).
We already had support for `vararg` in class model, where it was added as a property of the parameter. And this is much easier to carry over through our code (i.e. if I know a parameter was declared as vararg, I can easily ask for it to be generated as such).

So I have kept this concept, and added support also in `TypeName` - of course the user should make sure that if the parameter is no longer the last one, it should not be declared as vararg (whereas if we kept it as a method property, it would not matter).

One reason to not have it as a method property - we would have to update `TypedElementInfo`, which is used for methods, fields, parameters, and constructors (where it would be a bit strange to have a `vararg` property)

### DIFF
--- a/builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameSupport.java
+++ b/builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameSupport.java
@@ -273,10 +273,18 @@ final class TypeNameSupport {
             Stream.of(genericSection.split(","))
                     .map(String::trim) // remove possible spaces
                     .map(it -> {
-                        if (it.contains(".")) {
-                            return TypeName.create(it);
+                        if (it.contains(" extends ")) {
+                            // T extends io.helidon.webserver.spi.ServerFeature
+                            return createExtends(it);
+                        } else if (it.contains(" super ")) {
+                            // T super java.lang.String
+                            return createSuper(it);
                         } else {
-                            return TypeName.createFromGenericDeclaration(it);
+                            if (it.contains(".")) {
+                                return TypeName.create(it);
+                            } else {
+                                return TypeName.createFromGenericDeclaration(it);
+                            }
                         }
                     })
                     .forEach(builder::addTypeArgument);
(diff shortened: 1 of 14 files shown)


builder/codegen/src/main/java/io/helidon/builder/codegen/BuilderCodegen.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
```
Verdent: Possible leftover?

builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameSupport.java
```diff
@@ -330,6 +338,38 @@ static TypeName createFromGenericDeclaration(String genericAliasTypeName) {
...
+    }
+
+    private static TypeName createSuper(String typeNames) {
+        // T super io.helidon.webserver.spi.ServerFeature
+        int index = typeNames.indexOf(" super ");
+
+        var builder = TypeName.builder()
+                .generic(true)
+                .className(typeNames.substring(0, index));
+        String theOtherPart = typeNames.substring(index + 9);
```
Verdent: Is the 9 correct here? When it was about `" extends "`, that would make sense that you skip 9 values, but should not it be 7 this time? If that is the case, I would say to use constant or variable for the `" extends "` and `" super "` and use `length` instead of the exact number. That would prevent possible mistakes in the future.

builder/codegen/src/main/java/io/helidon/builder/codegen/BuilderCodegen.java
tomas-langer: Will reset the branch and remove this.

builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameSupport.java
tomas-langer: This is code that exists in the production code. I will add a test to make sure this is correct (and fix it)

Verdent (approved):

PR #10236, opened 2025-06-12
4.x: TypeName multidimensional array support added

### LINKED ISSUE #9712: 4.x: TypeName#className does not properly work for arrays
https://github.com/helidon-io/helidon/blob/cffb8e2dd63238c5a9af3bd3c3d540802b7a9619/common/types/src/main/java/io/helidon/common/types/TypeNameBlueprint.java#L59-L65

This method does not return the proper class name for arrays.

`int[].class.getSimpleName()` returns `int[]`

However this method returns only `int`.

### LINKED ISSUE #9713: 4.x: TypeName does not support multidimensional arrays
Array related TypeName objects are always considered single dimensional, even though they were created from the multidimensional type.

Verdent:
### Description
Fixes #9712 
Fixes #9713

### Documentation

None

### DIFF
--- a/builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameBlueprint.java
+++ b/builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameBlueprint.java
@@ -18,6 +18,7 @@ package io.helidon.common.types;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import io.helidon.builder.api.Option;
@@ -88,7 +89,7 @@ interface TypeNameBlueprint {
     List<String> enclosingNames();
 
     /**
-     * Functions the same as {@link Class#isPrimitive()}.
+     * Functions similar as {@link Class#isPrimitive()}.
      *
      * @return true if this type represents a primitive type
      */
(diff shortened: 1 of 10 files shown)


builder/tests/common-types/src/main/java/io/helidon/common/types/TypeNameBlueprint.java
```diff
@@ -181,6 +182,14 @@ default String classNameWithEnclosingNames() {
     @Option.Redundant
     List<TypeName> upperBounds();
 
+    /**
+     * Component type of array.
+     *
+     * @return component type of array
+     */
+    @Option.Redundant
```
tomas-langer: This is not redundant.
Types `String[][]` and `String[]` are not the same

common/types/src/test/java/io/helidon/common/types/TypeNameTest.java
```diff
@@ -238,14 +238,14 @@ void primitiveArrayTypes() {
...
 
-        assertThat(create(boolean[].class).className(), is("boolean"));
-        assertThat(create(byte[].class).className(), is("byte"));
-        assertThat(create(short[].class).className(), is("short"));
-        assertThat(create(int[].class).className(), is("int"));
-        assertThat(create(long[].class).className(), is("long"));
-        assertThat(create(char[].class).className(), is("char"));
-        assertThat(create(float[].class).className(), is("float"));
-        assertThat(create(double[].class).className(), is("double"));
+        assertThat(create(boolean[].class).className(), is("boolean[]"));
```
tomas-langer: I am not sure this is a correct change. As we actually had test that validated this, it was the desired behavior...

tomas-langer (approved):

Issue #10239, opened 2025-06-12
Shutdown order of Helidon services is wrong


arouel:
## Environment Details
* Helidon Version: 4.2.2
* Helidon SE
* JDK version: 24
* OS: macOS 15.5

----------

## Problem Description

Due to a wrong shutdown order of Helidon services, we run into errors, because dependent services are already shut down.

#### current behavior

Helidon services get shut down in the same order as they got started according to their runlevel and weight definition.

#### expected behavior

Helidon services should be shutdown in reverse chronological order as they were started.

## Steps to reproduce

1. Create two services with a `RunLevel` annotation and assign one the number `RunLevel.STARTUP` and the other `RunLevel.STARTUP + 1`.
2. Start the services via the `ServiceRegistryManager.start`
3. Stop the services via the `ServiceRegistryManager.shutdown`

The services get shut down in the same order as the got started.

#### example code
```java
import io.helidon.service.registry.Service;
import io.helidon.service.registry.Service.PostConstruct;
import io.helidon.service.registry.Service.PreDestroy;
import io.helidon.service.registry.ServiceRegistryConfig;
import io.helidon.service.registry.ServiceRegistryManager;

public class StartStopTest {

    public static void main(String[] args) {
        var config = ServiceRegistryConfig.create();
        var manager = ServiceRegistryManager.start(config);
        manager.shutdown();
    }

    @Service.Singleton
    @Service.RunLevel(Service.RunLevel.STARTUP)
    static class Service1 {
        @PostConstruct
        void startUp() {
            System.out.println(getClass().getSimpleName() + " started");
        }
        @PreDestroy
        void shutDown() {
            System.out.println(getClass().getSimpleName() + " stopped");
        }
    }

    @Service.Singleton
    @Service.RunLevel(Service.RunLevel.STARTUP + 1)
    static class Service2 {
        @PostConstruct
        void startUp() {
            System.out.println(getClass().getSimpleName() + " started");
        }
        @PreDestroy
        void shutDown() {
            System.out.println(getClass().getSimpleName() + " stopped");
        }
    }
}
```

#### console output
```
Service1 started
Service2 started
Service1 stopped
Service2 stopped
```

PR #10240, opened 2025-06-12
Shutdown services in reverse order (fixes issue #10239)

arouel:
### Problem Description

Due to a wrong shutdown order of Helidon services, we run into errors, because dependent services are already shut down (see a detailed explanation in https://github.com/helidon-io/helidon/issues/10239).

### Documentation

With this change, Helidon services will be shut down in reverse chronological order in which they were started. This fixes issue https://github.com/helidon-io/helidon/issues/10239.


tjquinno:
@arouel Thanks for the issue and the PR. I've taken the liberty of contributing a small doc update to your PR to describe the new (correct) shutdown order.

arouel:
@tjquinno thank you for the refinement.

PR #10242, opened 2025-06-12
4.x: JSON-P/B configuration methods added

### LINKED ISSUE #9680: HTTP media support JSON-B/P and Jackson configuration
It is currently not possible configure JSON-B/P instances used in HTTP media support in any way. Method accepting custom instance are missing completely. `JacksonSupport` do have `create` method, which accepts custom `ObjectMapper`. It should be aligned in all our media supports.
https://github.com/helidon-io/helidon/blob/30f1bf66ffb8252dedf042962e4b6b6ba4e27955/http/media/jackson/src/main/java/io/helidon/http/media/jackson/JacksonSupport.java#L96-L98

In addition to that, all of our previously mentioned media supports (including Jackson) are missing configuration from the config. Config is passed to the create method, but ignored and never processed. See https://github.com/helidon-io/helidon/blob/30f1bf66ffb8252dedf042962e4b6b6ba4e27955/http/media/jsonb/src/main/java/io/helidon/http/media/jsonb/JsonbSupport.java#L73-L78

Verdent:
### Description
Fixes #9680 

### Documentation

None

### DIFF
--- a/builder/api/src/main/java/io/helidon/builder/api/Option.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Option.java
@@ -495,9 +495,9 @@ public final class Option {
     }
 
     /**
-     * Marker of how {@link java.util.Map} keys and values should be constructed.
-     * If used, it will use {@code io.helidon.common.config.Config#traverse} method to depth-first traversal of the node
-     * and its subtree.
+     * Definition of how {@link java.util.Map} keys and values should be constructed.
+     * If used, it will use the {@code io.helidon.common.config.Config#traverse} method to perform a depth-first traversal
+     * of the node and its subtrees.
      * <p>
      * For example:
      * <pre>{@code
--- a/builder/api/src/main/java/io/helidon/builder/api/Option.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Option.java
@@ -493,4 +493,38 @@ public final class Option {
          */
         Class<? extends Prototype.OptionDecorator<?, ?>> value();
     }
+
+    /**
+     * Marker of how {@link java.util.Map} keys and values should be constructed.
+     * If used, it will use {@code io.helidon.common.config.Config#traverse} method to depth-first traversal of the node
+     * and its subtree.
+     * <p>
+     * For example:
+     * <pre>{@code
+     * test-map:
+     *    key: "test-value1"
+     *    test-key:
+     *       second-part: "test-value2"
+     *       third-part: "test-value3"
+     * }</pre>
+     * <p>
+     * Will be handled as:
+     * <pre>{@code
+     * key: "key" value: "test-value1"
+     * key: "test-key.second-part" value: "test-value2"
+     * key: "test-key.third-part" value: "test-value3"
+     * }</pre>
+     */
+    @Target(ElementType.METHOD)
+    @Retention(RetentionPolicy.CLASS)
+    public @interface TraverseConfig {
+
+        /**
+         * Whether to use traverse method to handle map key and value.
+         *
+         * @return true to enable traverse and false to disable
+         */
+        boolean value() default true;
+
+    }
 }
--- a/builder/api/src/main/java/io/helidon/builder/api/Option.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Option.java
@@ -496,8 +496,13 @@ public final class Option {
 
     /**
      * Definition of how {@link java.util.Map} keys and values should be constructed.
-     * If used, it will use the {@code io.helidon.common.config.Config#traverse} method to perform a depth-first traversal
-     * of the node and its subtrees.
+     * <p>
+     * If this annotation is not used, traversed is automatically applied on String and primitive/boxed types.
+     * In all other cases, non-traverse approach is applied.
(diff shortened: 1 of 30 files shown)


http/media/jsonp/src/main/java/io/helidon/http/media/jsonp/JsonpSupport.java
```diff
@@ -90,8 +99,31 @@ public static MediaSupport create(Config config) {
...
+     * Creates a new builder.
+     *
+     * @return a new builder instance
+     */
+    public static JsonpSupportConfig.Builder builder() {
+        return JsonpSupportConfig.builder();
+    }
 
-        return new JsonpSupport(name);
+    static JsonpSupport create(Consumer<JsonpSupportConfig.Builder> consumer) {
```
tomas-langer: why is this method not public?

Verdent: Oh. Forgot to add public. Will fix that.

builder/api/src/main/java/io/helidon/builder/api/Option.java
```diff
@@ -481,4 +481,38 @@ private Option() {
          */
         Class<? extends Prototype.OptionDecorator<?, ?>> value();
     }
+
+    /**
+     * Definition of how {@link java.util.Map} keys and values should be constructed.
```
tomas-langer, line 498: There should be a mention of the default behavior somewhere on Option - maybe on this annotation.

- default for strings and primitive/boxed types should be traverse
- default for all other types is no-traverse

tomas-langer: Also this is only relevant if `Configured` is specified as well on the option, which should be mentioned here.

builder/codegen/src/main/java/io/helidon/builder/codegen/AnnotationDataOption.java
```diff
@@ -165,6 +168,9 @@ static AnnotationDataOption create(TypeHandler handler, TypedElementInfo element
         javadoc = processDeprecation(deprecationData, annotations, javadoc);
 
         TypeName decorator = optionDecorator(element);
+        traverseConfig = element.findAnnotation(OPTION_TRAVERSE_CONFIG)
+                .flatMap(Annotation::booleanValue)
+                .orElse(false);
```
tomas-langer: default should be depending on Map generic types, not `false`

config/config-mp/src/main/java/io/helidon/config/mp/SeConfig.java
```diff
@@ -167,8 +167,7 @@ public Stream<Config> traverse(Predicate<Config> predicate) {
                         .filter(predicate)
                         .map(node -> traverseSubNodes(node, predicate))
                         .reduce(Stream.empty(), Stream::concat))
-                .orElseThrow(MissingValueException.createSupplier(key()));
-
+                .orElseGet(Stream::of);
```
tomas-langer: This deserves a comment - otherwise the change is strange (i.e. "As javadoc describes we return an empty stream for missing config nodes")

http/media/jsonp/src/main/java/io/helidon/http/media/jsonp/JsonpSupport.java
```diff
@@ -67,7 +76,7 @@ private JsonpSupport(String name) {
      * @return a new {@link JsonpSupport}
      */
     public static MediaSupport create() {
-        return new JsonpSupport("jsonp");
+        return builder().name("jsonp").build();
```
tomas-langer: There is no need to specify the name, as it is default in the builder (I guess the same comment is valid for JSON-B and Jackson)

tomas-langer (changes requested):
We will now have generated configuration metadata for the providers (btw. we have now added GSON, which may benefit from the same features).
The documentation should be updated now to include the configuration options for each provider, if we document the media supports...

builder/api/src/main/java/io/helidon/builder/api/Option.java
Verdent: Makes sense.

http/media/gson/src/main/java/io/helidon/http/media/gson/GsonSupportConfigBlueprint.java
```diff
@@ -0,0 +1,60 @@
...
+import io.helidon.builder.api.Prototype;
+
+import com.google.gson.Gson;
+
+/**
+ * Configuration of the {@link GsonSupport}.
+ */
+@Prototype.Blueprint(decorator = GsonSupport.Decorator.class)
+@Prototype.Configured(value = "gson", root = false)
+interface GsonSupportConfigBlueprint extends Prototype.Factory<GsonSupport> {
```
tomas-langer, line 33: All the media support blueprints should provide a service `io.helidon.http.media.spi.MediaSupportProvider`, to be correctly mapped in documentation to `io.helidon.http.media.MediaContextConfig` - the generated docs for `mediaSupports` should list all the configurable media supports you add in this PR.

PR #10244, opened 2025-06-12
Adds support for JSON-RPC client and server APIs

spericas:
### Description

Adds support for JSON-RPC client and server APIs


jsonrpc/core/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>helidon-jsonrpc-project</artifactId>
+        <groupId>io.helidon.jsonrpc</groupId>
+        <version>4.3.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>helidon-jsonrpc-core</artifactId>
```
tomas-langer, line 28: As this module is not split into `api` and `core`, I think it could just be called `helidon-jsonrpc` instead of `helidon-jsonrpc-core`

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcErrorImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonObjectBuilder;
+import jakarta.json.JsonValue;
+
+/**
+ * An implementation of a JSON-RPC error.
+ */
+public class JsonRpcErrorImpl implements JsonRpcError {
```
tomas-langer: This class should not be public. Also `Impl` should never be used for public classes in Helidon.

The static factory methods should be on the `JsonRpcError` interface.

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcErrorImpl.java
```diff
@@ -0,0 +1,104 @@
...
+     * Create an instance from a code, a message and data.
+     *
+     * @param code    the error code
+     * @param message the message
+     * @param data    the associated data or {@code null}
+     * @return a new instance of this class
+     */
+    public static JsonRpcErrorImpl create(int code, String message, JsonValue data) {
+        Objects.requireNonNull(message, "message is null");
+        JsonObjectBuilder builder = Json.createObjectBuilder();
```
tomas-langer: There should be a JsonObjectBuilderFactory static instance somewhere in the module, as that should be more efficient than creating a new builder each time.

tomas-langer: a public or protected method in Helidon must never accept `null` as a valid value. `data` must never be null...

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcParams.java
```diff
@@ -0,0 +1,121 @@
...
+    String getString(String name);
+
+    /**
+     * Get a single param by name as a JSON value, if present.
+     *
+     * @param name param name
+     * @return an optional param value
+     * @throws ClassCastException if not a JSON object
+     */
+    Optional<JsonValue> optionalGet(String name);
```
tomas-langer: There is already a naming pattern in Helidon for this - the method should be either called `find` or `first` depending on the semantics.

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcParamsImpl.java
```diff
@@ -0,0 +1,105 @@
...
+import jakarta.json.JsonArray;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonString;
+import jakarta.json.JsonStructure;
+import jakarta.json.JsonValue;
+
+/**
+ * An implementation of {@link io.helidon.jsonrpc.core.JsonRpcParams}.
+ */
+public class JsonRpcParamsImpl implements JsonRpcParams {
```
tomas-langer: Implementation classes should never be public, please move the static methods to interface and make this package local (same as with JsonRpcError)

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcResult.java
```diff
@@ -0,0 +1,121 @@
...
+    String getString(String name);
+
+    /**
+     * Get a JSON object property value as a string, if present.
+     *
+     * @param name property name
+     * @return an optional property value
+     * @throws ClassCastException if not a JSON object or value not a string
+     */
+    Optional<JsonValue> optionalGet(String name);
```
tomas-langer: As above, also maybe this could be defined on some super-interface, as the methods seem to be duplicated

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcResultImpl.java
```diff
@@ -0,0 +1,104 @@
...
+
+import jakarta.json.JsonArray;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonString;
+import jakarta.json.JsonValue;
+
+/**
+ * An implementation of a JSON-RPC response result.
+ */
+public class JsonRpcResultImpl implements JsonRpcResult {
```
tomas-langer: dtto regarding public type and static methods

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
```diff
@@ -0,0 +1,72 @@
...
+import jakarta.json.JsonReader;
+import jakarta.json.JsonValue;
+import jakarta.json.bind.Jsonb;
+import jakarta.json.bind.JsonbBuilder;
+
+/**
+ * Provides JSONP to/from JSONB conversions. Not efficient, but simple and portable
+ * for now. A more efficient implementation should avoid serialization.
+ */
+public class JsonUtil {
```
tomas-langer, line 41: Static utility classes in general are not great. One problem in this case - what if the user wants to customize `JSON-B` instance?

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @param object the JSON object
+     * @param type   the JSONB type
+     * @param <T>    the type of the instance
+     * @return the JSONB instance
+     */
+    public static <T> T jsonpToJsonb(JsonObject object, Class<T> type) {
+        Objects.requireNonNull(object, "json object is null");
+        Objects.requireNonNull(type, "type is null");
+        String serialized = object.toString();
```
tomas-langer: JSON-B and JSON-P should work together, I was hoping there is a way to use the `JsonObject` directly instead of serilaizing and deserializing from a string.
If not available, please use a writer, as `object.toString()` uses pretty print (I think), and as such is less performance. Also a writer can be created from a factory and made faster.

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
```diff
@@ -0,0 +1,72 @@
...
+     * Convert a JSONB object into a {@link JsonValue}.
+     *
+     * @param object the object
+     * @return the JSON value
+     */
+    public static JsonObject jsonbToJsonp(Object object) {
+        Objects.requireNonNull(object, "json object is null");
+        String serialized = JSONB.get().toJson(object);
+        InputStream stream = new ByteArrayInputStream(serialized.getBytes(StandardCharsets.UTF_8));
+        try (JsonReader reader = Json.createReader(stream)) {
```
tomas-langer: This should use JsonReaderFactory.
Also we may want to allow customization of JSON-P, which is not possible (maybe this could be a follow up)

jsonrpc/core/src/main/java/module-info.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Helidon JSON-RPC core package.
```
tomas-langer: module, not package.

jsonrpc/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+        <version>4.3.0-SNAPSHOT</version>
+    </parent>
+    <packaging>pom</packaging>
+
+    <groupId>io.helidon.jsonrpc</groupId>
+    <artifactId>helidon-jsonrpc-project</artifactId>
+    <name>Helidon JSON-RPC Project</name>
+
+    <description>JSON-RPC support for Helidon</description>
+
```
tomas-langer, line 34: Please use this for all new modules:
```xml
    <properties>
    </properties>
```

webclient/jsonrpc/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+                        <version>${helidon.version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>io.helidon.codegen</groupId>
+                        <artifactId>helidon-codegen-helidon-copyright</artifactId>
+                        <version>${helidon.version}</version>
+                    </dependency>
+                </dependencies>
+            </plugin>
+            <plugin>
```
tomas-langer: This plugin should be configured for all of Helidon, not sure why is it needed explicitly?

webclient/jsonrpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import io.helidon.common.features.api.Features;
+
+/**
+ * Helidon WebClient JSON-RPC Support.
+ */
+@Features.Incubating
```
tomas-langer: Does this work without adding other feature annotations?

webserver/jsonrpc/pom.xml
```diff
@@ -0,0 +1,137 @@
...
+        </dependency>
+        <dependency>
+            <groupId>jakarta.json.bind</groupId>
+            <artifactId>jakarta.json.bind-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.http.media</groupId>
+            <artifactId>helidon-http-media-jsonp</artifactId>
+        </dependency>
+        <dependency>
```
tomas-langer, line 51: This should never be a dependency of our modules (`helidon-logging-jul`), only runtime of examples, or tests.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcErrorHandler.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Handler for a JSON-RPC erroneous request.
+     *
+     * @param req the server request
+     * @param jsonObject an invalid JSON-RPC request as a JSON object
+     * @return success or failure handling error. If the value returned is {@code true},
+     *         the error will not be reported to a client.
+     * @throws Exception if an unexpected condition is found
+     */
+    boolean handle(ServerRequest req, JsonObject jsonObject) throws Exception;
```
tomas-langer: I would expect an error handler to have a possibility to return a specific response, or provide an error (number + text or whatever is needed by JSON-RPC).

Also our methods should not throw `Exception` - even though this is something the user would implement, checked exceptions should not be accepted.

Overall I am not sure what this would solve, as you only can provide boolean...

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcHandlers.java
```diff
@@ -0,0 +1,114 @@
...
+     * @return the error handler or {@code null}
+     */
+    JsonRpcErrorHandler errorHandler() {
+        return errorHandler;
+    }
+
+    /**
+     * A builder for {@link io.helidon.webserver.jsonrpc.JsonRpcHandlers}.
+     */
+    public static class Builder implements io.helidon.common.Builder<JsonRpcHandlers.Builder, JsonRpcHandlers> {
```
tomas-langer: Do not use custom made builders, only in extreme cases. This should be using a blueprint.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
```diff
@@ -0,0 +1,165 @@
...
+import io.helidon.webserver.http.ServerResponse;
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.JsonValue;
+
+/**
+ * A representation of a JSON-RPC response.
+ */
+public interface JsonRpcResponse extends ServerResponse {
```
tomas-langer, line 31: I am not sure about extending `ServerResponse` here (and maybe the same for `ServerRequest above.

These interfaces have methods for handling entity, which should not be used when using JSON-RPC - such as methods `send(...)`

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
```diff
@@ -0,0 +1,165 @@
...
+     * until the response is delivered.
+     */
+    void send();
+
+    /**
+     * Get a complete response as a JSON object.
+     *
+     * @return a JSON object that represents the response
+     */
+    JsonObject asJsonObject();
```
tomas-langer, line 164: Why is this public API? Do we expect the user to get the response as a JSON when sending it?

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcRulesImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * An implementation of JSON-RPC rules.
+ */
+class JsonRpcRulesImpl implements JsonRpcRules {
+
+    private final Map<String, JsonRpcHandlers> rules = new HashMap<>();
```
tomas-langer: Routing should be always ordered, this uses a hash map.
If you register `/rpc/api/greet/world` first, and `/rpc/api/greet/*` , this may reverse the order, and the first route would never work.

webserver/testing/junit5/jsonrpc/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.webserver.testing.junit5</groupId>
+            <artifactId>helidon-webserver-testing-junit5</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.webserver</groupId>
+            <artifactId>helidon-webserver-jsonrpc</artifactId>
+            <version>${project.version}</version>
```
tomas-langer: There should never be a version specified for modules we release - these must be added in `bom` pom and in `all` pom, then version is not needed.

webserver/testing/junit5/jsonrpc/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.webclient</groupId>
+            <artifactId>helidon-webclient-jsonrpc</artifactId>
+            <version>${project.version}</version>
```
tomas-langer: version should be deleted

tomas-langer (changes requested):

jsonrpc/core/pom.xml
spericas: I prefer to leave this open for other sub-packages in the future (if not a showstopper).

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcErrorImpl.java
spericas: Will refactor this

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcErrorImpl.java
spericas: Will be refactored

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcParams.java
spericas: Good point

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcParamsImpl.java
spericas: See above

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
spericas: I went back and forth. But there maybe some useful methods like query that would need to be repeated (maybe others as well).

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
spericas: Unfortunately we need this in MCP for now because responses are sent over an SSE channel. Maybe we can deprecate in the future.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcRulesImpl.java
spericas: This is a bug, right

webserver/testing/junit5/jsonrpc/pom.xml
spericas: Forgot to remove version here

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
spericas: There are some alternatives, but none portable across JSON implementations AFAICT

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
spericas: Perhaps we can consider customizations in the next phase

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcRouting.java
```diff
@@ -0,0 +1,354 @@
...
+            if (!"2.0".equals(version)) {
+                return INVALID_REQUEST_ERROR;
+            }
+            String method = object.getString("method");
+            JsonRpcHandler handler = handlersMap.get(method);
+            if (handler == null) {
+                return METHOD_NOT_FOUND_ERROR;
+            }
+            return null;
+        } catch (ClassCastException e) {
```
tvallin, line 247:
```suggestion
        } catch (ClassCastException | NullPointerException e) {
```

If the Json Rpc is a response, it does not have a `method`, and the `object.getString("method");` throws a NPE. If you plan to handle `Responses` in a different way, than this change is not required.

spericas: I think the special case of a response arriving via the SSE channel in the MCP server should never hit this verification code. Did you actually encounter the NPE?

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcRulesImpl.java
spericas: Fixed

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
spericas: Yes, customizations are important, perhaps we can consider them in phase 2. As I see it, the main user of this API is MCP, so unless it is needed there, then we can defer at this time.

jsonrpc/core/src/main/java/module-info.java
spericas: Fixed

webserver/jsonrpc/pom.xml
spericas: I moved it to the tests directory, but now every JSON-RPC app will need to list it a dependency.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcErrorHandler.java
spericas: This was added mostly for the MCP server to handle responses to requests sent by the server (an odd condition that will likely go away in future versions where SSE is dropped). I understand your point, but we really need @tvallin feedback on this one.

webclient/jsonrpc/pom.xml
spericas: Removed

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcHandlers.java
spericas: Switched to a blueprint. I wanted to use just `method(String, JsonRpcHandler)` in the builder but using a blueprint it will be `putMethod(String, JsonRpcHandler)` being a `Map`. It would be nice to have full control of the name.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcRouting.java
spericas: Code has been updated to check if method is there to avoid NPEs

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcErrorHandler.java
spericas: Updated signature to return `Optional<JsonRpcError>`.  If the optional is empty, error is ignored; otherwise, the error returned is sent back to the client.

webclient/jsonrpc/src/main/java/module-info.java
spericas: Removed

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
tomas-langer: Still - please do not use `object.toString()`, as that does pretty printing

webclient/jsonrpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Helidon WebClient JSON-RPC Support.
+ */
+module io.helidon.webclient.jsonrpc {
```
tomas-langer, line 30: You have dependency on `io.helidon.common.features.api` (also in pom file), yet there is no feature defined here.

webserver/jsonrpc/src/main/java/module-info.java
```diff
@@ -0,0 +1,32 @@
...
+module io.helidon.webserver.jsonrpc {
+
+    requires io.helidon.builder.api;
+    requires io.helidon.common.config;
+    requires io.helidon.webserver;
+    requires io.helidon.jsonrpc.core;
+    requires jakarta.json;
+    requires jakarta.json.bind;
+
+    requires static io.helidon.common.features.api;
```
tomas-langer: Features are not used

webserver/jsonrpc/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.helidon.common.features</groupId>
+                            <artifactId>helidon-common-features-processor</artifactId>
```
tomas-langer: This is a wrong dependency - it should be a `codegen` for new modules.
Also it is actually not used in `module-info.java`

webserver/jsonrpc/pom.xml
tomas-langer: That is not true - user can choose any implementation they want, `helidon-logging-jul`, `helidon-logging-slf4j` etc. We must never have one of the implementations as a compile/runtime dependency of our modules

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcHandlers.java
tomas-langer: Maybe 
This is  a good enhancement, can you please create an issue for it?

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
tomas-langer: I like the delegate approach better, thanks

webclient/jsonrpc/src/main/java/module-info.java
spericas: Removed

webserver/jsonrpc/src/main/java/module-info.java
spericas: Fixed

webserver/jsonrpc/pom.xml
spericas: Ok, please review all the AP deps

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonUtil.java
spericas: Fixed, missed this before

webserver/jsonrpc/pom.xml
spericas: Sorry, I thought the comment was about `helidon-http-media-jsonp`. Sure we don't need the jul there.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcRouting.java
tvallin: Yes I encountered NPE because the response does not arrive through SSE channel but by MCP client POST.

The MCP server sends a json-rpc request using SSE channel and receive a json-rpc response as a HTTP POST.

It happens during a `Sampling` scenario.

spericas: @tvallin Can you check the latest changes and verify they satisfy your reqs?

tomas-langer (approved):

Issue #10257, opened 2025-06-14
Incorrect error message is returned in JwtAuthProvider when token is expired


zakerf:
## Environment Details
* Helidon Version: 4.2.3
* Helidon MP
* JDK version: 21
* OS: Ubuntu

----------

## Problem Description

If you send a token with a valid signature but it's expired then there is no clear error message returned in the response and/or logs.

Example log line:
2025.06.14 22:41:43 FINER AUDIT VirtualThread[#102,[0x4b492de8 0x3246856c] WebServer socket]/runnable@ForkJoinPool-1-worker-1: FAILURE authn.authenticate d567273b-f2db-4d36-a004-ed843d6b2467:2  io.helidon.microprofile.security.SecurityFilterCommon processAuthentication SecurityFilterCommon.java 243 :: "Provider io.helidon.microprofile.jwt.auth.JwtAuthProvider. Message: "

## Steps to reproduce
Use any JWT that's expired

PR #10258, opened 2025-06-14
fix: return validators errors if token validation fails

zakerf:
### Description

Fix for #10257. You can see that on L272 errors is also returned

### Documentation

If no doc impact: None


Verdent:
Hi @zakerf ,
before we can continue with your PR, you need to sign the OCA and also address the copyright issue.

tomas-langer:
Fixed by #10458

PR #10273, opened 2025-06-17
4.x: Check supertypes during validation of runtime types. Added test

### LINKED ISSUE #10261: 4.x Config/Builder annotation processor ignores inheritance during some validation
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP SE
* JDK version:
* OS:

----------

### Summary
Builder/config annotation processing fails to recognize that a runtime type implements `RuntimeType.Api<TheConfigType>` by virtue of extending a parameterized abstract class.

**Workaround**: Add `implements RuntimeType.Api<TheConfigType>` to the subclass.

### Detailed example

This example is drawn from health but it's not specific to health.

Common behavior among all built-in health checks could be defined in a configured blueprint this way:
```java
@Prototype.Configured
@Prototype.Blueprint
interface BuiltInCheckConfigBlueprint  {...}
```
The corresponding runtime type class would be declared like this:
```java
abstract class BuiltInCheck<CONFIG extends BuiltInCheckConfig> implements HealthCheck, RuntimeType.Api<CONFIG> {...}
```

A specific health check's blueprint looks like this:
```java
@Prototype.Configured
@Prototype.Blueprint
interface HeapMemoryCheckConfigBlueprint extends BuiltInCheckConfigBlueprint,
        Prototype.Factory<HeapMemoryCheck> {...}
```
and its runtime type class is like this:
```java
@RuntimeType.PrototypedBy(HeapMemoryCheckConfig.class)
class HeapMemoryCheck extends BuiltInCheck<HeapMemoryCheckConfig>
    // The following line seems to be needed but should not be because the superclass implements RuntimeTYpe.Api<CONFIG>.
    implements RuntimeType.Api<HeapMemoryCheckConfig> {...}
```

As noted in the last // comment above, the annotation processor's validation does not recognize that `HeapMemoryCheck` implements `RuntimeType.Api<HeapMemoryCheckConfig>` by virtue of extending `BuiltInCheck<HeapMemoryCheckConfig>`.

```list
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project helidon-test-builderconfiginheritance: 
Compilation failure: 
Compilation failure: 
[ERROR] /BuilderConfigInheritance/helidon-test-builderconfiginheritance/src/main/java/io/helidon/test/inh/HeapMemoryCheck.java:[9,1]
Type annotated with @RuntimeType.PrototypedBy(HeapMemoryCheckConfig.class) 
must implement RuntimeType.Api<HeapMemoryCheckConfig>
[ERROR] /BuilderConfigInheritance/helidon-test-builderconfiginheritance/src/main/java/io/helidon/test/inh/HeapMemoryCheck.java:[9,1] 
As io.helidon.test.inh.HeapMemoryCheckConfigBlueprint implements 
Prototype.Factory<io.helidon.test.inh.HeapMemoryCheck>, the runtime type must 
implement(or extend) interface io.helidon.builder.api.RuntimeType.Api<HeapMemoryCheckConfig>
```

## Steps to reproduce
See the attached project and its `README.md`. 

As attached the project builds because `HeapMemoryCheck`  (the subclass) has an explicit `implements RuntimeType.Api<HeapMemoryCheckConfig>` clause. 

Comment out the explicit `implements` and the build fails with the annotation processor validation error, even though the superclass declares the required `implements` clause (using a type parameter).

tomas-langer:
### Description
Resolves #10261 

Added test to verify the issue.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/ValidationTask.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/ValidationTask.java
@@ -39,8 +39,8 @@ abstract class ValidationTask {
                                            TypeInfo validatedType,
                                            TypeName implementedInterface,
                                            String message) {
-        boolean doesImplement = doesImplement(validatedType, implementedInterface);
-        if (!doesImplement) {
+
+        if (!doesImplement(validatedType, implementedInterface)) {
             errors.fatal(validatedType.typeName(), message);
         }
     }
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/ValidationTask.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/ValidationTask.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


builder/codegen/src/main/java/io/helidon/builder/codegen/ValidationTask.java
```diff
@@ -39,11 +39,21 @@ private static void validateImplements(Errors.Collector errors,
                                            TypeInfo validatedType,
                                            TypeName implementedInterface,
                                            String message) {
+        boolean doesImplement = doesImplement(validatedType, implementedInterface);
```
ljnelson: Is there a reason to introduce this local variable?

tomas-langer: not really

tjquinno (approved):
LGTM

PR #10274, opened 2025-06-17
Fix test problems with REST client metrics; also fix lack of support for on-demand registration of REST client metrics

### LINKED ISSUE #10264: 4.x MP REST client metrics support can throw `NullPointerException`
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
Originally reported by an internal user.

Under some circumstances the Helidon REST client metrics support CDI extension throws NPEs.

## Steps to reproduce
```bash
cd helidon/microprofile/tests/tck/tck-rest-client
```

In the Helidon MP REST client TCK project `pom.xml` there add this dependency:
```xml
        <dependency>
            <groupId>io.helidon.microprofile.rest-client-metrics</groupId>
            <artifactId>helidon-microprofile-rest-client-metrics</artifactId>
            <scope>test</scope>
        </dependency>
```

```bash
mvn clean test
```

Several tests fail with NPEs.

tjquinno:
### Description
Resolves #10264 

## Release note
----
* Helidon's support for MicroProfile REST Client allows applications to use an interface that is _not_ discovered or injected by CDI for a REST client. (See [this section](https://download.eclipse.org/microprofile/microprofile-rest-client-3.0/microprofile-rest-client-spec-3.0.html#_sample_builder_usage) in the MP REST Client spec.) Helidon's metrics support for REST clients did not correctly support such usage but this PR fixes that.
* Helidon's REST client metrics support now works properly in test scenarios where the Helidon container starts more than once in a JVM.
----

## PR overview 

1. Tests that run multiple REST clients in a reused JVM could fail.
   
   The code uses a CDI extension to register meters to be updated when user code invokes REST client methods, and a REST client filter updates those meters at call-time. The filter relies on data structures built by and stored in the  extension.
   
   The code had a _static_ field holding a reference to the extension. This has worked correctly in production situations with one Helidon container per JVM, but in test suites that run the Helidon container multiple times the listener would always use the _first_ extension because of the static field. This PR changes that to an _instance_ field which is functionally the same in production runs but solves the testing problem.
2. The code now correctly supports when user code registers a REST client interface that is not known ahead of time to CDI. (REST clients can be activated programmatically before CDI is fully initialized.) 
   
   The Helidon REST client metrics support did not handle this case, throwing an NPE. Now this scenario works correctly.   
3. To test the REST client metrics code more, this PR updates the _REST client_ TCK, adding a test dependency on REST client metrics. Even though the code in the REST client TCK does not have any metrics annotations, adding the dependency revealed a timing issue which this PR also resolves, described next.
4. Recall that the extension registers the meters to update, the listener learns when new REST clients are activated, and it registers a REST client filter with each activated REST client. The filter uses the CDI extension's data structures to update the appropriate timers and counters when the user code invokes uses the REST client.
   
   As mentioned above, in some cases, the REST client mechanism can initialize and use listeners and filters _before_ CDI has completed its startup so the extension might not be ready. The listener and filter in this component have a `LazyValue` for the extension and they also tolerate if loading the `LazyValue` fails because CDI is not available yet. As with any `LazyValue` on each usage the filter and listener invoke `lazyValue.get()` which tries to load the `LazyValue` again.  Eventually--once CDI is up--they obtain a valid reference to the extension. Until then no REST client metrics are supported on those early-activated REST clients.

### Documentation
No change. These are bug fixes.

### DIFF
--- a/microprofile/rest-client-metrics/src/main/java/io/helidon/microprofile/restclientmetrics/RestClientMetricsClientListener.java
+++ b/microprofile/rest-client-metrics/src/main/java/io/helidon/microprofile/restclientmetrics/RestClientMetricsClientListener.java
@@ -66,10 +66,10 @@ public class RestClientMetricsClientListener implements RestClientListener {
 
         /*
         In some cases, the system will invoke REST client listeners before CDI has completed its startup. The REST client metrics
-         CDI extension registers the meters to be updated for each REST client method, so this listener has nothing to do if
-         it runs before CDI has initialized the extension. That said, it needs to keep retrying to locate the CDI extension
-         because the extension might (should) become available later. The code below uses a utility method to access this
-         lazy value to allow that retry behavior.
+        CDI extension registers the meters to be updated for each REST client method, so this listener has nothing to do if
+        it runs before CDI has initialized the extension. That said, it needs to keep retrying to locate the CDI extension
+        because the extension might (should) become available later. The code below uses a utility method to access this
+        lazy value to allow that retry behavior.
          */
         private final LazyValue<RestClientMetricsCdiExtension> ext =
                 LazyValue.create(() -> CDI.current().getBeanManager().getExtension(RestClientMetricsCdiExtension.class));


spericas (approved):
LGTM

Issue #10203, opened 2025-06-04
Verify gRPC reflection service works correctly with MP service


spericas:
Helidon Version:

## Task Description

Verify gRPC reflection service works correctly with MP service. MP uses a slightly different path which may cause problems with the proto file access.

PR #10276, opened 2025-06-18
Adds support for new gRPC MP annotation 

spericas:
### Description

Adds support for a new @Grpc.Proto annotation to mark a gRPC MP method as returning the service's proto file descriptor. This is necessary when using the gRPC reflection service with MP. Also fixes a bug in setting up interceptors in routes and makes `ContextSettingServerInterceptor` immutable. Some new tests and test changes too. Issue #10203.

### Documentation

In a future PR.

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/ContextSettingServerInterceptor.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/ContextSettingServerInterceptor.java
@@ -34,26 +34,18 @@ import io.grpc.ServerInterceptor;
 import static io.helidon.grpc.core.GrpcHelper.extractMethodName;
 
 /**
- * A {@link ServerInterceptor} that sets values into the gRPC call context.
+ * A {@link io.grpc.ServerInterceptor} that sets values into the gRPC call context.
  */
 @Weight(InterceptorWeights.CONTEXT)
-class ContextSettingServerInterceptor implements ServerInterceptor, GrpcServiceDescriptor.Aware {
+class ContextSettingServerInterceptor implements ServerInterceptor {
 
-    /**
-     * The {@link GrpcServiceDescriptor} for the service being intercepted.
-     */
-    private GrpcServiceDescriptor serviceDescriptor;
+    private static final ContextSettingServerInterceptor INSTANCE = new ContextSettingServerInterceptor();
 
     private ContextSettingServerInterceptor() {
     }
 
-    /**
-     * Create a {@link ContextSettingServerInterceptor}.
-     *
-     * @return a {@link ContextSettingServerInterceptor} instance.
-     */
-    static ContextSettingServerInterceptor create() {
-        return new ContextSettingServerInterceptor();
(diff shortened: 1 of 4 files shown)


microprofile/grpc/server/src/main/java/io/helidon/microprofile/grpc/server/GrpcServiceBuilder.java
```diff
@@ -127,6 +129,24 @@ public GrpcServiceDescriptor build() {
         GrpcServiceDescriptor.Builder builder = GrpcServiceDescriptor.builder(serviceClass(), name)
                 .marshallerSupplier(getMarshallerSupplier());
 
+        // find protobuf file descriptor if available, used for gRPC reflection
+        Optional<AnnotatedMethod> protoMethod = methodList
```
tjquinno, line 135: Are there any requirements as to the visibility of the annotated method that should be checked here?

If a method has `@Grpc.Proto` but fails any of these checks should this code let the user know, maybe by logging a warning or throwing an exception? As written we just ignore such a method silently.

microprofile/grpc/tests/src/test/java/io/helidon/microprofile/grpc/tests/ReflectionServiceTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @ParameterizedTest
+    @CsvSource({"EchoService", "EchoService.Echo"})
+    void testFindSymbol(String symbol) throws InterruptedException {
+        TestObserver<ServerReflectionResponse> res = new TestObserver<>(1);
+        StreamObserver<ServerReflectionRequest> req = stub.serverReflectionInfo(res);
+        req.onNext(ServerReflectionRequest.newBuilder()
+                           .setFileContainingSymbol(symbol)
+                           .build());
+        req.onCompleted();
+        res.await(5, TimeUnit.SECONDS);
```
tjquinno, line 55: I have no reason to doubt that 5 s is long enough normally, but given how things run slowly in the pipeline jobs sometimes is this a risk for intermittent pipeline failures? I'd think the latch would typically permit the test to proceed quickly regardless of the timeout value, so does that argue for a much larger one knowing that ordinarily the latch will clear much sooner? 30 s? (I don't know; I'm just throwing numbers around.)

microprofile/grpc/server/src/main/java/io/helidon/microprofile/grpc/server/GrpcServiceBuilder.java
spericas: Code update to verify method's signature

microprofile/grpc/tests/src/test/java/io/helidon/microprofile/grpc/tests/ReflectionServiceTest.java
spericas: Set to 10 secs, that should be plenty

tjquinno (approved):

PR #10283, opened 2025-06-21
4.x: Remove unused code from io.helidon.metrics.providers.micrometer.MMeterRegistry

Captain1653:
### Description
I've found unused code in `MMeterRegistry`:
* package private overloaded method `create`
* private method `ensurePrometheusRegistryIsPresent`
* private class `ClockWrapper`

It can be safely removed without breaking compatibility.


tjquinno:
Let's not proceed with this PR right away. 

I'm concerned particularly that `ensurePrometheusRegistryIsPresent` would not be used from anywhere, and I want a chance to look into that.

Captain1653:
Sure, no problem.

I or you can convert this PR to draft if it's necessary.

Issue #9921, opened 2025-03-20
Helidon SE 4.X- Helidon Scheduler support


MohanramGovindarasu:
## Environment Details
* Helidon Version: 4.2
* Helidon SE or Helidon MP: Helidon SE

----------

## Enhancement Description

Heldion Scheduling- https://helidon.io/docs/latest/se/scheduling doesnt support listing jobs or deleting/cancelling jobs. We have used Helidon Schedular to call the java methods in async mode, it works fine. But we dont have complete control of the Schedular from the application code for the following reasons:

1. Listing of the Jobs
2. Get the status of the running Jobs
3. Deleting the Jobs 
4. Cancelling the Jobs
5. Managing Main Job and Sub-Job 

It will be good to have these features as part of the Helidon framework. As an alternative we are forced to uptake Quartz or any other other party tech stack which has these capabilities.

tomas-langer:
I will try to list my undestanding of the requirements, and I would have some additional questions to clarify them:

1. Listing jobs - we can add an ID to each job, which would allow you to list all the jobs, we cannot in most cases (unfortunately)  see which method would be invoked, as there is no such information for lambdas
2. Get the status - probably OK as well, we can see if the job is running, we could probably even show if the last execution failed etc. 
3. Deleting a job - I do not see much difference between deleting a job and cancelling a job, could you please explain the difference you expect here?
4. Cancelling a job - we can cancel the future (this is already supported in latest version, each task has a `close()` method that cancels it
5. Main/sub job - I have no idea what we could do here, can you maybe explain on an example?

tomas-langer:
Feature is now merged into `main`.
The features are reasonably simple - doing what was requested in scope of what our scheduling supports (starting/closing a job, listing jobs).
As no more details were provided for other requriements, I am closing this issue.
If this there are other features that would be helpful, please create a new issue describing them. Thanks!

MohanramGovindarasu:
Managing Main and SubJob:

- Only when all the subjob is done with their work, Main Job should be marked as done.
- Sub Jobs should have mechanism to notify Main Jobs once the sub job work is done, so that Main Job can spin off new subjobs based on the need.
-  Incase of failures, subjobs should have identified based on Batch ID/subjob ID. So that Main Job knows which subjob failed and retrigger/create new subjobs based on that.

PR #10289, opened 2025-06-24
4.x: Support for tracking scheduled tasks.

tomas-langer:
### Description
Support for tracking scheduled tasks.

A new class `ScheduledTasks` is used to track all tasks. Introduced a task ID - configurable, defaults to UUID, in declarative is the FQN of the class + `.methodName(..)` where the parameter is either empty, or one of `CronInvocation` or `FixedRateInvocation`.

The lookup will be opened as public API if we agree on this approach. 

Related to #9921

### DIFF
--- a/scheduling/src/main/java/io/helidon/scheduling/Task.java
+++ b/scheduling/src/main/java/io/helidon/scheduling/Task.java
@@ -49,6 +49,7 @@ public interface Task {
      * @return task ID
      */
     default String id() {
-        return description();
+        // we need a unique id in this VM, this is implemented by Helidon tasks
+        return String.valueOf(System.identityHashCode(this));
     }
 }
--- a/scheduling/src/main/java/io/helidon/scheduling/TaskConfigBlueprint.java
+++ b/scheduling/src/main/java/io/helidon/scheduling/TaskConfigBlueprint.java
@@ -44,7 +44,8 @@ interface TaskConfigBlueprint {
     /**
      * Identification of the started task. This can be used to later look up the instance, for example to cancel it.
      *
-     * @return task id, if not provided, a unique id will be generated
+     * @return task id, if not provided via configuration or builder, the configured task will generate a unique ID
      */
+    @Option.Configured
     Optional<String> id();
 }
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/scheduling/SchedulingExtension.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/scheduling/SchedulingExtension.java
@@ -206,14 +206,15 @@ class SchedulingExtension implements RegistryCodegenExtension {
         postConstruct.increaseContentPadding()
                 .increaseContentPadding()
                 .addContentLine(".taskManager(taskManager)")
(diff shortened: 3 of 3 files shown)


scheduling/src/main/java/io/helidon/scheduling/Task.java
```diff
@@ -42,4 +41,14 @@ public interface Task {
...
     }
+
+    /**
+     * ID used to identify this task.
+     * It should be unique, as it is used to identify a single task, for example to cancel it.
+     *
+     * @return task ID
+     */
+    default String id() {
+        return description();
```
danielkec: Description?

tomas-langer: There needs to be some default here, as otherwise this would be a backward incompatible change.

spericas: Isn't it better to use a UUID for uniqueness then? Or is that what it would make it backward incompatible?

scheduling/src/main/java/io/helidon/scheduling/TaskConfigBlueprint.java
```diff
@@ -31,4 +33,18 @@ interface TaskConfigBlueprint {
...
+     */
+    @Option.RegistryService
+    TaskManager taskManager();
+
+    /**
+     * Identification of the started task. This can be used to later look up the instance, for example to cancel it.
+     *
+     * @return task id, if not provided, a unique id will be generated
+     */
+    Optional<String> id();
```
spericas, line 50: Not sure I understand this config. So if an ID is specified then the config applies to a single task? Is the comment "a unique id will be generated" relevant here? It's a bit odd because it is an optional here.

scheduling/src/main/java/io/helidon/scheduling/Task.java
tomas-langer: You cannot use UUID, as that would require state, and this is a default method on an interface, which must return the same value each time it is invoked.

scheduling/src/main/java/io/helidon/scheduling/TaskConfigBlueprint.java
tomas-langer: Either you provide an ID using the builder, or one will be generated by the implementation.

scheduling/src/main/java/io/helidon/scheduling/Task.java
tomas-langer: I will change this to System.identityHasCode as that makes more sense here and is unique

scheduling/src/main/java/io/helidon/scheduling/TaskConfigBlueprint.java
tomas-langer: Updating this to be configurable, with a better description.

scheduling/src/main/java/io/helidon/scheduling/Task.java
spericas: Nice

spericas (approved):

PR #10291, opened 2025-06-24
4.x: CDI Bridge update, OCI extension update

### LINKED ISSUE #10096: AmbiguousResolutionException in Helidon 4.2.0 on a combination of Helidon dependencies
## Environment Details
* Helidon Version: 4.2.0 or greater 
* Helidon MP

----------

## Problem Description
When a combination of dependencies is used in Helidon MP 4.2.0 like below:
```xml
        <!-- Begin:  This combination of dependencies will cause AmbiguousResolutionException -->
        <dependency>
            <groupId>io.helidon.integrations.oci.metrics</groupId>
            <artifactId>helidon-integrations-oci-metrics-cdi</artifactId>
        </dependency>
        <dependency>
            <groupId>io.helidon.integrations.oci.sdk</groupId>
            <artifactId>helidon-integrations-oci-sdk-cdi</artifactId>
        </dependency>
        <dependency>
            <groupId>com.oracle.oci.sdk</groupId>
            <artifactId>oci-java-sdk-common</artifactId>
        </dependency>
        <dependency>
            <groupId>com.oracle.oci.sdk</groupId>
            <artifactId>oci-java-sdk-monitoring</artifactId>
        </dependency>
        <dependency>
            <groupId>io.helidon.integrations.oci</groupId>
            <artifactId>helidon-integrations-oci</artifactId>
            <scope>runtime</scope>
        </dependency>
        <!-- End -->
```
results into a runtime error:
```bash
Exception in thread "main" org.jboss.weld.exceptions.AmbiguousResolutionException: WELD-001335: Ambiguous dependencies for type AbstractAuthenticationDetailsProvider with qualifiers @Default
 Possible dependencies:
  - Configurator Bean [class io.helidon.integrations.oci.sdk.cdi.OciExtension, types: AbstractAuthenticationDetailsProvider, BasicAuthenticationDetailsProvider, qualifiers: @Any @Default],
  - Configurator Bean [interface com.oracle.bmc.auth.BasicAuthenticationDetailsProvider, types: AbstractAuthenticationDetailsProvider, BasicAuthenticationDetailsProvider, qualifiers: @Any @Default]
```
This is a regression as this is not happening prior to 4.2.0. For example, using 4.1.7 does not have this issue.

## Steps to reproduce
1. Create a Helidon quick start mp using helidon cli and use version 4.2.0 or higher.
2. Add the following dependencies in pom.xml
    ```xml
        <!-- Begin:  This combination of dependencies will cause AmbiguousResolutionException -->
        <dependency>
            <groupId>io.helidon.integrations.oci.metrics</groupId>
            <artifactId>helidon-integrations-oci-metrics-cdi</artifactId>
        </dependency>
        <dependency>
            <groupId>io.helidon.integrations.oci.sdk</groupId>
            <artifactId>helidon-integrations-oci-sdk-cdi</artifactId>
        </dependency>
        <dependency>
            <groupId>com.oracle.oci.sdk</groupId>
            <artifactId>oci-java-sdk-common</artifactId>
        </dependency>
        <dependency>
            <groupId>com.oracle.oci.sdk</groupId>
            <artifactId>oci-java-sdk-monitoring</artifactId>
        </dependency>
        <dependency>
            <groupId>io.helidon.integrations.oci</groupId>
            <artifactId>helidon-integrations-oci</artifactId>
            <scope>runtime</scope>
        </dependency>
        <!-- End -->
    ```
3. Build
    ```bash
    mvn clean install -DskipTests
    ```
4. Run
    ```bash
    ```
5. Yields the this exception:
   ```bash
   Exception in thread "main" org.jboss.weld.exceptions.AmbiguousResolutionException: WELD-001335: Ambiguous dependencies for type AbstractAuthenticationDetailsProvider with qualifiers @Default
   Possible dependencies:
    - Configurator Bean [class io.helidon.integrations.oci.sdk.cdi.OciExtension, types: AbstractAuthenticationDetailsProvider, BasicAuthenticationDetailsProvider, qualifiers: @Any @Default],
    - Configurator Bean [interface com.oracle.bmc.auth.BasicAuthenticationDetailsProvider, types:     AbstractAuthenticationDetailsProvider, BasicAuthenticationDetailsProvider, qualifiers: @Any @Default]
 	at org.jboss.weld.bean.builtin.InstanceImpl.checkBeanResolved(InstanceImpl.java:254)
	at org.jboss.weld.bean.builtin.InstanceImpl.get(InstanceImpl.java:113)
        ...
   ```
6. Reproducer is attached:
    [helidon-quickstart-mp.zip](https://github.com/user-attachments/files/20076909/helidon-quickstart-mp.zip)

## Root Cause Analysis
This issue started happening on this PR: [4.x: Langchain4j integration (#9733) - Revision #: 634923ca950f2ae470de1365005146a7bd53a8ba](https://github.com/helidon-io/helidon/pull/9733)

tomas-langer:
### Description
Resolves #10096 

### Documentation
Instead of trying to rely on the CDI bridge being the last extension with full knowledge of all other synthetic and declared beans, I modified the integration modules of OCI to check if registry contains the values needed for the OCI CDI extension to work. If so, registration of beans is skipped.

### DIFF
--- a/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
+++ b/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
@@ -66,7 +66,8 @@ public class OciMetricsBean extends OciMetricsSupportFactory {
 
     // Make Priority higher than MetricsCdiExtension so this will only start after MetricsCdiExtension has completed.
     void registerOciMetrics(@Observes @Priority(LIBRARY_BEFORE + 20) @Initialized(ApplicationScoped.class) Object ignore,
-                            Config rootConfig, BeanManager beanManager) {
+                            Config rootConfig,
+                            BeanManager beanManager) {
 
         var instance = beanManager.createInstance().select(Monitoring.class);
         if (instance.isUnsatisfied()) {
--- a/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
+++ b/integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import jakarta.annotation.Priority;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.context.Initialized;
 import jakarta.enterprise.event.Observes;
+import jakarta.enterprise.inject.spi.BeanManager;
 import jakarta.inject.Singleton;
 
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_BEFORE;
@@ -65,8 +66,13 @@ public class OciMetricsBean extends OciMetricsSupportFactory {
 
     // Make Priority higher than MetricsCdiExtension so this will only start after MetricsCdiExtension has completed.
     void registerOciMetrics(@Observes @Priority(LIBRARY_BEFORE + 20) @Initialized(ApplicationScoped.class) Object ignore,
-                            Config rootConfig, Monitoring monitoringClient) {
-        registerOciMetrics(rootConfig, monitoringClient);
+                            Config rootConfig, BeanManager beanManager) {
+
+        var instance = beanManager.createInstance().select(Monitoring.class);
+        if (instance.isUnsatisfied()) {
+            return;
+        }
+        registerOciMetrics(rootConfig, instance.get());
     }
 
     // For testing
--- a/integrations/oci/metrics/cdi/src/test/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsCdiExtensionTest.java
+++ b/integrations/oci/metrics/cdi/src/test/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsCdiExtensionTest.java
@@ -15,9 +15,7 @@
  */
 package io.helidon.integrations.oci.metrics.cdi;
 
-import java.lang.annotation.Annotation;
 import java.lang.reflect.Proxy;
-import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
@@ -42,15 +40,11 @@ import com.oracle.bmc.monitoring.model.MetricDataDetails;
 import com.oracle.bmc.monitoring.model.PostMetricDataDetails;
 import com.oracle.bmc.monitoring.requests.PostMetricDataRequest;
 import com.oracle.bmc.monitoring.responses.PostMetricDataResponse;
-
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.event.Observes;
-
 import jakarta.enterprise.inject.Default;
 import jakarta.enterprise.inject.spi.AfterBeanDiscovery;
 import jakarta.enterprise.inject.spi.Extension;
-import jakarta.enterprise.inject.spi.InjectionPoint;
-import jakarta.enterprise.inject.spi.ProcessInjectionPoint;
 import jakarta.enterprise.inject.spi.configurator.BeanConfigurator;
 import org.glassfish.jersey.ext.cdi1x.internal.CdiComponentProvider;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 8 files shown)


ljnelson:
I think this still needs more thinking.

This approach seems to rely on an unproven fact, namely that whenever the OCI CDI extension is present at runtime there will also exist, at runtime, another portable extension, somewhere, that will cause services in the registry to eventually show up as CDI beans. You and I know this is expected to be the service registry extension. But that does not seem to be enforced here, nor do I think it should be.

(This PR is technically therefore a backwards incompatible change: the Maven module containing the `ServiceRegistryExtension` does not currently appear in the `runtime` dependencies of the Maven module containing the `OciExtension`. It is, in other words, possible to run an application today that uses the `OciExtension` without the `ServiceRegistryExtension`, but this PR sort of indirectly assumes the presence of the `ServiceRegistryExtension` at runtime in all cases. Now, in practice, how many people actually use the `OciExtension` in isolation? I have no idea. Maybe the number will happen to be `0`.)

(The other lesser thing is: the `OciExtension` can be used, at the moment, outside of Helidon entirely. That's obviously not some kind of enormous goal, but it is a nice feature. This PR would couple it more tightly to Helidon's internals.)

I think we can still get out of this problem by slowing down and thinking harder about when services are added as beans and when beans are added as services. Maybe certain kinds of work needs to be done "first" and other kinds need to be done "last". Certainly the OCI extension needs to detect synthetic beans anyway. I haven't had time to fully deal with all this yet. Just my thoughts right now.

tomas-langer:
The problem I have encountered is very simple - synthetic bean events are reported by Weld after all "after bean discovery" events - i.e. you cannot add (or not add) a bean as a result of some other extension registering a bean.

Right now our code does not work, as I was expecting to get some reasonable ordering. This is an attempted fix, that does indeed assume something that is part of Helidon CDI module (i.e. it is always present in Helidon MP).

I would like some cleaner solution (a general one) to resolve this for any kind of bean.

Nevertheless for the specific problem here - OCI, it would be actually better if the extensions (Helidon OCI integration and the CDI OCI integration) managed to cooperate - i.e. if there is an AuthenticationDetailsProvider in the registry, that is the one that should be used, as the OCI integration for registry is more "low level" than the CDI one (it is expected to work before even config is created). So the final solution should still provide this OCI specific behavior.

ljnelson:
> I would like some cleaner solution (a general one) to resolve this for any kind of bean.

I think [there is one](https://github.com/helidon-io/helidon/issues/10096#issuecomment-2864047796) but it requires going to CDI 4.1. I'll see if I can prototype this, though I'm not up to speed on our Weld repackaging.

If you feel you need to merge this PR I have not requested changes or blocked it in any way.

integrations/oci/metrics/cdi/src/main/java/io/helidon/integrations/oci/metrics/cdi/OciMetricsBean.java
```diff
@@ -65,8 +66,13 @@ protected void activateOciMetricsSupport(Config rootConfig, Config ociMetricsCon
 
     // Make Priority higher than MetricsCdiExtension so this will only start after MetricsCdiExtension has completed.
     void registerOciMetrics(@Observes @Priority(LIBRARY_BEFORE + 20) @Initialized(ApplicationScoped.class) Object ignore,
-                            Config rootConfig, Monitoring monitoringClient) {
-        registerOciMetrics(rootConfig, monitoringClient);
+                            Config rootConfig, BeanManager beanManager) {
```
ljnelson: Instead of injecting `BeanManager` I think you can inject `Instance<Monitoring>` directly.

ljnelson:
It may turn out that we can get to a generic, non-service-registry-dependent, "pure" CDI resolution here if, simply, the `ServiceRegistryExtension` adds its synthetic beans [as alternatives](https://jakarta.ee/specifications/cdi/4.0/apidocs/jakarta.cdi/jakarta/enterprise/inject/spi/configurator/beanconfigurator#alternative(boolean)), with ordinary [priorities](https://jakarta.ee/specifications/cdi/4.0/apidocs/jakarta.cdi/jakarta/enterprise/inject/spi/configurator/beanconfigurator#priority(int)), and if the other Helidon-authored portable extensions do so as well, with really low priorities. I believe that [typesafe resolution itself may resolve any ambiguities in such a case](https://jakarta.ee/specifications/cdi/4.0/jakarta-cdi-spec-4.0#unsatisfied_and_ambig_dependencies) while still allowing a user to bring her own "normal" CDI beans to the table. Actively investigating.

ljnelson:
I believe this relatively simple approach may work, but it is too early to say for sure. I will come back to it on Monday.

ljnelson:
It's even simpler than that. It looks like a one- or two-line change to `ServiceRegistryExtension` does the trick. I'll put a tiny PR together on Monday.

ljnelson:
See also #10313

klustria:
@tomas-langer, I checked Laird's simple change in https://github.com/helidon-io/helidon/pull/10313 and that also resolves the issue. Can you please check if you are ok with that fix?

tomas-langer:
Closing this PR, as the PR by @ljnelson does a better and more universal job at fixing this problem.

PR #10314, opened 2025-07-02
Adds support for tracing to the gRPC client

spericas:
### Description

Support for tracing in gRPC client using a client interceptor. PR also introduces a new SPI for gRPC client services. Other services, such as context support, will be provided later.

### Documentation

TODO


webclient/grpc-tracing/src/main/java/io/helidon/webclient/grpc/tracing/GrpcClientTracing.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public String type() {
+        return "tracing";
+    }
+
+    @Override
+    public WeightedBag<ClientInterceptor> interceptors() {
+        WeightedBag<ClientInterceptor> interceptors = WeightedBag.create();
+        interceptors.add(new GrpcClientTracingInterceptor(), Weighted.DEFAULT_WEIGHT + 1000.0);
```
tjquinno: Is the idea that this weight helps this interceptor be have quite a bit heavier than default weight but leaves room if users _really_ wanted an interceptor with higher weight?

webclient/grpc-tracing/src/main/java/io/helidon/webclient/grpc/tracing/GrpcClientTracingInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+
+/**
+ * A gRPC channel interceptor for tracing.
+ */
+public class GrpcClientTracingInterceptor implements ClientInterceptor {
```
tjquinno: Can this be less visible than `public` given that Helidon-provided code in `GrpcClientTracing` adds it to the interceptors?

webclient/grpc-tracing/src/main/java/io/helidon/webclient/grpc/tracing/GrpcClientTracing.java
spericas: Yes, unfortunately, I don't think we have any ranges defined for this. I see many times we use 100, but some other times as much as 900. Perhaps we need to make this more consistent across our modules. I'm going to use 100 as I see it used more often.

webclient/grpc-tracing/src/main/java/io/helidon/webclient/grpc/tracing/GrpcClientTracingInterceptor.java
spericas: Yes, changing that

tjquinno (approved):

PR #10315, opened 2025-07-03
4.x: Use Maven 3.9.9

romain-grecourt:
### Description

Use a Maven 3.9.9 explicitly.
The current version available in the runners (3.9.10) causes issues, see https://github.com/apache/maven/issues/2487

### Documentation

None


ljnelson (approved):
LGTM. (Maven wrapper is still easier than this!)

romain-grecourt:
I don't mind using mvnw, but that should be done with more thoughts than just fixing a blocking issue.

PR #10318, opened 2025-07-03
4.x: Updated parsing of annotations from String for class model and builder

### LINKED ISSUE #10064: 4.x: Parsing of annotations from string for code generation needs to be improved
Helidon Version:

## Task Description

Method `io.helidon.codegen.classmodel.Annotation#parse`

The method must honore the types supported by annotations, so we can define more complex ones.
We can achieve this by analyzing the annotation `TypeInfo` - obtaining types for each property, and then parsing it correctly from the strings.
Another option is to have some explicit rules for defining types (i.e. 42S would be short, 43B  would be byte etc. to allow any type correctly).

Also currently if  chars , or = are part of the String value we fail horribly.

This is a TODO follow up (removed TODO from code).

tomas-langer:
- allows class model to print out any annotation structure,
- allows builder to add any annotation through its approach to generated methods.

Added test for all types of annotation properties, also combined with array types and nested annotations. The result is correctly using imports.

Resolves #10064

### DIFF
--- a/codegen/class-model/src/main/java/io/helidon/codegen/classmodel/AnnotationParser.java
+++ b/codegen/class-model/src/main/java/io/helidon/codegen/classmodel/AnnotationParser.java
@@ -23,6 +23,7 @@ import io.helidon.common.types.EnumValue;
 import io.helidon.common.types.TypeName;
 
 final class AnnotationParser {
+    private static final System.Logger LOGGER = System.getLogger(AnnotationParser.class.getName());
     private static final String CLASS_PREFIX = "class::";
     private static final String ENUM_PREFIX = "enum::";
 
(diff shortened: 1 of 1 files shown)


codegen/class-model/src/main/java/io/helidon/codegen/classmodel/AnnotationParser.java
```diff
@@ -0,0 +1,339 @@
...
+                        case 'n' -> '\n';
+                        case 'r' -> '\r';
+                        case 'f' -> '\f';
+                        case '\'' -> '\'';
+                        default -> throw new IllegalArgumentException("Invalid annotation value specified, invalid escaped "
+                                                                              + "value (expected one of t, n, r, f, '): "
+                                                                              + value);
+                    };
+                } else {
+                    throw new IllegalArgumentException("Invalid annotation value specified, char must be a single character,"
```
Verdent, line 111: I think you are missing one case. This is also a valid char: `'\u0046'`

tomas-langer: We can support it in the future, for now this is not supported.

Verdent (approved):

PR #10331, opened 2025-07-07
Allow option for adding Iterable<String> header values

klustria:
### Description
Allow option for adding Iterable<String> header values

### Documentation

2 additional APIs were added in HeaderValues:
1. create(HeaderName name, Iterable<String> values)
2. create(String name, Iterable<String> values)

example:
   ```java
   final String[] ordinalNumbers = {"First", "Second", "Third"};
   Iterable<String> ordinals = Arrays.asList(ordinalNumbers);

   Header header  = HeaderValues.create("ordinals", o);
   ``` 

Currently no documentation for HeaderValues usage, but if one exist, add the above example.

If no doc impact: None


tomas-langer:
We already have methods for a `Collection`, why is `Iterable` needed? Please describe the use case.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
```diff
@@ -0,0 +1,72 @@
...
+    HeaderValueIterable(HeaderName name, boolean changing, boolean sensitive, Iterable<String> values) {
+        super(name, changing, sensitive, values.iterator().next());
+
+        this.originalValues = values;
+    }
+
+    @Override
+    public HeaderWriteable addValue(String value) {
+        if (values == null) {
+            values = new ArrayList<>(getIterableSize() + 1);
```
tomas-langer: this is not a needed optimization. Just create a new ArrayList

http/http/src/main/java/io/helidon/http/HeaderValues.java
```diff
@@ -349,6 +349,36 @@ public static Header create(String name, Collection<String> values) {
...
+     * Create a new header. This header is considered unchanging and not sensitive.
+     *
+     * @param name   name of the header
+     * @param values Iterable&lt;String&gt; type values of the header, must contain at least one value (which may be an empty String)
+     * @return a new header
+     * @see #create(io.helidon.http.HeaderName, boolean, boolean, String...)
+     * @throws java.lang.IllegalArgumentException in case the collection is empty
+     */
+    public static Header create(HeaderName name, Iterable<String> values) {
+        System.out.println("Header.create(HeaderName name, Iterable<String> values)");
```
tomas-langer: Please remove this line

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
```diff
@@ -0,0 +1,72 @@
...
+        return this;
+    }
+
+    private int getIterableSize() {
+        int iterableSize = 0;
+        if (originalValues instanceof Collection) {
+            iterableSize = ((Collection<?>) originalValues).size();
+        } else {
+            // non-Collection type does not have size() so we need to count the elements
+            for (Object i : originalValues) {
```
tomas-langer: once you iterate an iterable, it is gone and cannot be re-iterated, this is wrong.
Maybe just create a list from the iterable values, or create it lazily when `valueCount` is used.

tjquinno:
> We already have methods for a `Collection`, why is `Iterable` needed? Please describe the use case.

See the linked issue description which mentions gRPC's `Metadata` type  as a specific example that makes header values available this way. I have edited the description to add a link to the Javadoc.

Given that dealing with headers can be in performance-critical sections of code, it could be useful to be able to avoid creating a throw-away collection from an `Iterable`.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
tjquinno: > once you iterate an iterable, it is gone and cannot be re-iterated, this is wrong. Maybe just create a list from the iterable values, or create it lazily when `valueCount` is used.

Java's `for` loop creates and uses an `Iterator` from the `Iterable`, IIRC; an `Iterator` is definitely single-use, but the `Iterable` can return an iterator any number of times. 

And the point behind the original issue to support `Iterable` directly was to avoid creating a throw-away intermediate data structure.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+package io.helidon.http;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+class HeaderValueIterable extends HeaderWritableValueBase {
```
tjquinno: Because an `Iterable` itself is not writeable, would it be a better fit if `HeaderValueIterable` extended `HeaderValueBase` instead of `HeaderWritableValueBase`? 

The use case that inspired the original issue is that a third-party library type (gRPC `Metadata`) makes header values available as an `Iterable<String>` without providing (direct) access to the underlying data structure which need not be a Java standard one. As such, the intent of exposing an `Iterable` would seem to be to provide read-only access to the values.

Having `HeaderValueIterable` extend `HeaderValueBase` instead would remove the need to maintain a separate, mutable `List` of values derived from the original `Iterable`. Instead, the required methods would use `Iterator`s as needed, obtained from the `Iterable`, without having to know what the underlying data structure might be and without having to maintain a copy of the values.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
klustria: From searching at whether Iterables can be iterated multiple times, here is the result and is in line with Tim's input:
```
... while an Iterator itself can only be used once, the Iterable object from which it is obtained can be iterated over multiple times by requesting a fresh Iterator for each new traversal
```
Also, I did a test to call valueCount() multiple times and it is correctly returning the right value everytime.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
klustria: Makes sense. @tomas-langer do you agree as well? Because of this feedback, I now made `HeaderValueIterable` extend `HeaderValueBase` and removed addValue() override method. Also made allValues() return the list from the Iterable every time it is called rather than maintaining a List in memory. Furthermore, the unit test for addValue() was also removed because of this change.

http/http/src/main/java/io/helidon/http/HeaderValues.java
klustria: Removed

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
```diff
@@ -55,18 +45,14 @@ private int getIterableSize() {
...
-        }
-        return values;
+        return StreamSupport.stream(originalValues.spliterator(), false).toList();
     }
 
     @Override
     public int valueCount() {
-        if (values == null) {
-            return getIterableSize();
+        if (valueCount == 0) {
```
tjquinno: Although this new incarnation of `HeaderValueIterable` does not _itself_ permit values to be added to the header, it's possible that the underlying `Iterable` that we are wrapping _might_ change in size or contents over time. It's probably very unlikely, but it's possible.

With that in mind, `valueCount()` might need to compute the size each time it is invoked.

The potential changeability over time of the underlying `Iterable` is another good reason to _not_ maintain a copy of the values inside the class. I know the implementation no longer does that but I'm just reinforcing the idea.

tjquinno (changes requested):
Another change in the handling of the size.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
klustria: Updated valueCount() to perform the count on every invocation to take into account underlying Iterables that has possibility to mutate. Also added a test to cover this scenario, i.e. to check that valueCount() and allValues() still work as expected even if the underlying Iterable has changed.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
ljnelson: ?! `for (Frob f : frobIterable)` is most certainly legal. An `Iterable` can produce an infinite number of `Iterator`s from its `iterator()` method. If it could not, it could not be a legal target of the enhanced `for` loop.

See https://docs.oracle.com/javase/specs/jls/se24/html/jls-14.html#jls-14.14.2 and search for "The precise meaning…" and following. From that document, here is the translation:

```
for (I #i = Expression.iterator(); #i.hasNext(); ) {
    {VariableModifier} T VarDeclId = (TargetType) #i.next();
    Statement
}
```

tjquinno: Yes. I too checked the JLS prior to commenting above, and when I read that section I interpreted "[t]he precise meaning" to be "the outwardly-observed net result," not necessarily "the implementation." It's a spec, after all.

That said, the Javadoc for the `iterator()` method of `Iterable` does not restrict how many times callers can invoke it on a given `Iterable` instance. If `iterator()` could be invoked at most once, `Iterable` would be next to useless.

ljnelson: Right; I was mostly responding to "once you iterate an iterable, it is gone and cannot be re-iterated", which just is not the case. (Certainly once you "run through" an `Iterator` and its `hasNext()` returns `false`, you're done.)

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
tomas-langer: We should never wrap a mutable object. If the object is possibly mutable, we should create an immutable copy

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
tomas-langer: Please keep consistency with collection and list - reuse `HeaderValueList` to support both writability of the header value, and immutability of the original value

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
tomas-langer: Also, as I noted above - just reuse the list based header value, and this is not needed

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
klustria: @tomas-langer, so the suggestion here is just to convert an Iterable to a List and just pass it on to `HeaderValueList` and not do `HeaderValueIterable` anymore? This will however negate @tjquinno 's hope of just processing an Iterable directly with no conversions providing some performance benefits. But my understanding is that we are ok without the performance benefit, but consistency with collection/list types is more important, is this correct?

tomas-langer: The consistency is important. You can have a specific class for iterable, but it should work in a similar way to list (i.e. support writability, not support mutable underlying source etc.).

klustria: Ok, I'll try and make a change to reuse the `HeaderValueList` for this purpose. FYI, @tjquinno, this will change from your original expectation of how we will process Iterables.

tjquinno: I guess I do not understand the concern with wrapping a mutable `Iterable` and delegating all operations to that `Iterable`.

Code that is reading values from any of our existing header values implementations does not know that the values are stable. 

Some other code could be updating a `HeaderValueList` for example. So that implies we are OK with code retrieving different values from a header by invoking `allValues` at different times. It might be unlikely in practice, but it's possible.

I do not see how that is different in principle from having a wrapper around a possibly-mutable `Iterable`.

klustria: I have the new code suggested by Tomas, but I will wait till we have a consensus. In the meantime, @tomas-langer can you please take a look at Tim's concern?

tomas-langer: The reason I do not want to use iterable provided by a user is because we would be dependent on the mutability constraints of an unkown implementation - what if we get a concurrent modification exception? We do not expect this at all.
So I would rather be safe here and make sure we are in full control of the values.

Is there a performance concern? If the iterable is maybe lazily created, it may be faster to use it directly. 
But then you have a huge overhead each time somebody asks for the size, as you need to iterate through all the values...

tjquinno: Copying the values is fine with me, given that this is the behavior we seem to want the implementations to have.

My main concern is that we seem to have made a design decision and built this behavior into our _implementations_ without ever declaring so in our Javadoc as part of the formal contract with callers and with developers of other implementations.

That's what led me initially to think it might be reasonable to wrap an `Iterable` rather than copy its values.

For sure the size issue could be a performance problem _if_ the `Iterable` is not also a `Collection` with its own `size` method. I don't know how likely that is, and I don't know how often callers invoke `size` vs. iterating over the values.

tjquinno:
Given the outcome of the discussion of what behavior we should implement, would it make sense to just have 
```java
public static Header create(HeaderName name, Iterable<String> values) {
    return create(name, StreamSupport.stream(values.spliterator(), false).toList());
}
```
and not even have a separate implementation for `HeaderValueIterable`?

This provides the convenience of the added factory method and therefore saves developera from having to create a list or array explicitly themselves, but the resulting `Header` still exhibits the snapshot behavior we want.

Although this would not create the list lazily upon first need, and doing that might have saved a little performance, there would be real savings only if the header values were never accessed which seems unlikely.

klustria:
> Given the outcome of the discussion of what behavior we should implement, would it make sense to just have
> 
> ```java
> public static Header create(HeaderName name, Iterable<String> values) {
>     return create(name, StreamSupport.stream(values.spliterator(), false).toList());
> }
> ```
> 
> and not even have a separate implementation for `HeaderValueIterable`?
> 
> This provides the convenience of the added factory method and therefore saves developera from having to create a list or array explicitly themselves, but the resulting `Header` still exhibits the snapshot behavior we want.
> 
> Although this would not create the list lazily upon first need, and doing that might have saved a little performance, there would be real savings only if the header values were never accessed which seems unlikely.

Yes, the code would be something like that. I'm casting if it is a collection so conversion is not needed, but otherwise, I'll use that same statement to convert non-collection type Iterables.

http/http/src/main/java/io/helidon/http/HeaderValueIterable.java
klustria: @tjquinno @tomas-langer, Kindly check the latest change. I am now reusing `HeaderValueList` and has also updated unit test for this change.

tjquinno:
> Yes, the code would be something like that. I'm casting if it is a collection so conversion is not needed, but otherwise, I'll use that same statement to convert non-collection type Iterables.

Ah, yes, of course. Sounds good.

http/http/src/test/java/io/helidon/http/HeaderValueIterableTest.java
```diff
@@ -0,0 +1,153 @@
...
+    // This will allow testing of a Collection and non-Collection type Iterable
+    private static Stream<Arguments> IterableObjects() {
+        return Stream.of(
+                // Collection type iterable
+                arguments(ordinalNumbers),
+                arguments(new HashSet<>(ordinalNumbers)),
+                // non-Collection type iterable
+                arguments(new CustomStringIterable(ordinalNumbers))
+        );
+    }
```
abdelhak-zaaim: Might want to cache these test arguments to avoid recreating HashSet/CustomStringIterable on every test run

klustria: Thanks for the suggestion, I implemented it.

klustria:
@tomas-langer @tjquinno , I have finalized the changes. Can you please review?

tomas-langer (approved):

PR #10353, opened 2025-07-09
Read timeout deprecated in the ConnectionKey

Verdent:
### Documentation

None


webclient/api/src/main/java/io/helidon/webclient/api/ConnectionKey.java
```diff
@@ -17,28 +17,189 @@
...
+    /**
+     * @param scheme           uri address scheme
+     * @param host             uri address host
+     * @param port             uri address port
+     * @param tls              TLS to be used in connection
+     * @param dnsResolver      DNS resolver to be used
+     * @param dnsAddressLookup DNS address lookup strategy
+     * @param proxy            Proxy server to use for outgoing requests
+     */
+    public ConnectionKey(String scheme,
```
tomas-langer: Please replace with `create` method instead of a public constructor.

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionKey.java
```diff
@@ -17,28 +17,189 @@
...
+            return true;
+        }
+        if (obj == null || obj.getClass() != this.getClass()) {
+            return false;
+        }
+        var that = (ConnectionKey) obj;
+        return Objects.equals(this.scheme, that.scheme)
+                && Objects.equals(this.host, that.host)
+                && this.port == that.port
+                && Objects.equals(this.readTimeout, that.readTimeout)
```
tomas-langer, line 179: Do not include read timeout in equals and hashCode - that is the reason we want this change.

webclient/tests/http1/src/test/java/io/helidon/webclient/http1/Http1ClientTest.java
```diff
@@ -518,7 +518,7 @@ private static void afterRedirectPut(ServerRequest req, ServerResponse res) {
     }
 
     private static void delayedHandler(ServerRequest req, ServerResponse res) throws IOException, InterruptedException {
-        TimeUnit.MILLISECONDS.sleep(10);
+        TimeUnit.SECONDS.sleep(1);
```
tomas-langer, line 521: Why did you increase the sleep here?

tomas-langer (changes requested):
After removing timeout, we must set the timeout before each outbound call according to the configuration of the current request/webclient.
This part is missing from this PR

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionKey.java
Verdent: Alright, I wanted to make this done the same way as it used to be, but without the timeout. So I kept the constructor. Will make create method.

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionKey.java
Verdent: I left it there for backwards compatibility. If it is not set (is null), this check will not fail. but if it is set, it will get validated the same way as before.

webclient/tests/http1/src/test/java/io/helidon/webclient/http1/Http1ClientTest.java
Verdent: Because it is too short for purposes where we are reusing the connection. Previously, since we operated with a different timeout, new connection was created, but now the old one was reused. because of our idle IS and how the waiting is done there, server always managed to respond before we got out of the waiting loop when cancelling the idle IS.

PR #10363, opened 2025-07-09
4.2.4: Use BOMs to manage versions of junit and langchain4j

barchetta:
Use BOMs to manage versions of junit and langchain4j

This does not upgrade langchain4j

This upgrades junit to 5.11.3 to align with the version of junit-platform-testkit


ljnelson (approved):
LGTM FWIW

barchetta:
The issues with the spring eureka integration tests is caused by the broken langchain4j bom's that are managing versions of things like junit and logback.  Langchain4j is fixing their bom's but for 0.36.2 they are broken.

So I'm going to bail on this PR - and we'll address this in 4.3.0 when we pick up later versions of langchain4j with fixed boms.

PR #10378, opened 2025-07-14
4.x: Update Jakarta EL implementation to org.glassfish.expressly:expressly:5.0.0

barchetta:
### Description

Helidon 4 supports Jakarta EE 10 which has EL 5.  But to date the EL implementation used by Helidon has been `org.glassfish:jakarta.el:4.0.2` which is for EL 4.

This PR upgrades the EL implementation to  `org.glassfish.expressly:expressly:5.0.0` (Expressly is the Eclipse project producing the EL implementation).

Note that the old EL implementation included the EL API. The new implementation (Expressly) does not include the API, instead it has it as a dependency.

We continue to manage the version of `org.glassfish:jakarta.el` so that we do not break existing application poms (which might assume Helidon manages the version of this artifact).

### DIFF
--- a/microprofile/tests/tck/tck-cdi-lang-model/pom.xml
+++ b/microprofile/tests/tck/tck-cdi-lang-model/pom.xml
@@ -75,8 +75,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.glassfish</groupId>
-            <artifactId>jakarta.el</artifactId>
+            <groupId>org.glassfish.expressly</groupId>
+            <artifactId>expressly</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -47,6 +47,8 @@
         <version.lib.database.messaging>19.3.0.0</version.lib.database.messaging>
         <version.lib.dropwizard.metrics>4.1.36</version.lib.dropwizard.metrics>
         <version.lib.eclipselink>4.0.7</version.lib.eclipselink>
+        <!-- el-impl has been replaced by expressly. But we still manage version to not break existing application poms -->
+        <version.lib.el-impl>4.0.2</version.lib.el-impl>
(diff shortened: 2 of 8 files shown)


ljnelson:
~Interesting title 😄~

Oh, wow, that's the new name of the project I guess!

microprofile/tests/tck/tck-cdi-lang-model/pom.xml
```diff
@@ -75,8 +75,8 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.glassfish</groupId>
-            <artifactId>jakarta.el</artifactId>
+            <groupId>org.glassfish.expressly</groupId>
```
ljnelson, line 78: Is it OK to change a TCK `pom.xml`?

barchetta: Not sure what you mean by "TCK pom.xml".  Isn't this pom defining the Helidon implementation (via dependencies) that will be tested by the TCK? So we would want that implementation to use the EL implementation that Helidon uses.

ljnelson: I confess I'm not sure. It seems like maybe it's an import of part of the TCK itself? Maybe this is all OK? See, for example, https://github.com/jakartaee/cdi-tck/blob/master/dist-build/lang-model-tck-runner/src/test/java/org/jboss/weld/lang/model/tck/LangModelTckTest.java and https://github.com/helidon-io/helidon/blob/main/microprofile/tests/tck/tck-cdi-lang-model/src/test/java/io/helidon/microprofile/cdi/lang/model/tck/LangModelTckTest.java

barchetta: As far as I can tell the TCK depends only on the el-api (as it should), and it is up to the implementation being tested to provide the implementation of el.

ljnelson: OK!

ljnelson (approved):

PR #10394, opened 2025-07-17
4.x: Declarative features configuration alignment and documentation

tomas-langer:
Update of FT features, added common configuration for name and `enabled`
Update to declarative features, to honor configuration overrides where reasonable Update to declarative documentation to describe approach to configuration

### DIFF
--- a/declarative/README.md
+++ b/declarative/README.md
@@ -24,8 +24,7 @@ Rules for Helidon Declarative:
    handle invocation, or implements an `io.helidon.service.registry.Interception.ElementInterceptor` that intercepts one specific
    element)
 5. If there is a good reason the user may want to override configuration in annotations with config, generate the code
-   appropriately (the configuration key should be type-config.<fully qualified class name>.<(method | field) name>.<type>, i.e.
-   `type-config.io.helidon.examples.decalarative.HttpEndpoint.retriable.retry`), or an explicit key configured on the annotation
+   appropriately, see `io.helidon.declarative.codegen.TypeConfigOverrides`, and the configuration section below
 6. If there is a good reason the user may want to use a custom named service implementation, provide a way to inject it (see Retry
    generated code for named retries, such as `GreetServiceEndpoint_retriable__Retry.java`)
 7. All features must be configured through service registry.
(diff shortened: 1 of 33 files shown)


danielkec (dismissed):

declarative/README.md
```diff
@@ -159,6 +160,76 @@ Parameters defined by qualifiers (may be an `Optional`, supports `Mappers`):
 
 ### Configuration
 
+Some annotation values may be overridden by configuration. The developer of the feature may decide to support this approach,
+though if supported, it must follow the guidelines defined here.
+
+1. Configuration key may be explicitly defined in an annotation to override the rules below
+2. The root configuration node for overriding annotation values is `type-config`
+3. Second level configuration node is the fully qualified class name of the annotated type (i.e. `com.example.MyType`)
```
ljnelson: So in Java properties format would this look like `type-config.com.foo.Bar.blat.frob`? How do I know what the class name is?

declarative/README.md
```diff
@@ -159,6 +160,76 @@ Parameters defined by qualifiers (may be an `Optional`, supports `Mappers`):
...
+
+Other features must be added as they are being developed
+
+Documented example:
+
+```yaml
+type-config:
+  "com.example.MyType":
+    server:
+      listener: "admin" # override socket name
```
ljnelson, line 256: (So in Java properties format this is `type-config.com.example.MyType.server.listener = admin`. Is that OK?

tomas-langer: Yes that is the key that would be used

declarative/README.md
tomas-langer: you do not need to know - the key is created by the component. As this is a fully qualified class name, there cannot be duplicate cases.
You could theoretically factor a duplicate key, but you would have to use package name starting with a capital letter, and class name starting with a lower case one - we do not need to support this

danielkec (approved):

tomas-langer:
Documentation of how to override using configuration could be achieved through a taglet for Javadoc...

Add the following comment on the annotation property definition (from `RestServer.Listener`):
```java
/**
 * Name of a routing to bind this application/service to.
 *
 * @return name of a routing (or listener host/port) on WebServer
 * @helidon.annotation-configured
 */
 String value();
```

Example of generated javadoc:
[String](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html) value
Name of a routing to bind this application/service to.

*Returns:*
name of a routing (or listener host/port) on WebServer

*Configuration override:*
This option can be overridden using configuration key:
type-overrides.{fully qualified annotated class name}.rest-server.listener.value

tomas-langer:
Closing this PR, as we have reached the following decision:

- support only expressions that reference config in annotations for now (of course limited to string options)
- remove support for any other configurability of declarative features

see PR #10491

PR #10405, opened 2025-07-19
4.x: Removed unused classes from microprofile/metrics

Captain1653:
### Description
I've found 3 unused package-private classes in `microprofile/metrics`:
* `InterceptionRunner`
* `InterceptionRunnerImpl`
* `InterceptionTargetInfo`

As I can see, they can be safely removed without breaking compatibility. Tests in the module `microprofile/metrics` are green.


Captain1653:
@tjquinno Can it be reviewed? Please :)

Issue #10380, opened 2025-07-14
Cannot define multiple `data` entries with a `SseEvent`


arouel:
## Environment Details
* Helidon Version: 4.2.3
* Helidon SE
* JDK version: 24
* OS: Linux

----------

## Problem Description

According to the [specification](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#examples) it is allowed to have multiple `data:` lines that build one message. Helidon seems to be missing this part from an API perspective, because I cannot create an event in the form of:
```
event: myevent
data: part1
data: part2
data: part3
```

## Steps to reproduce

Some might think it could be written with multiple emits like so
```java
void test(ServerRequest request, ServerResponse response) {
    try (SseSink sink = response.sink(SseSink.TYPE)) {
        sink.emit(SseEvent.builder()
                          .name("myevent")
                          .data("part1")
                          .build());
        sink.emit(SseEvent.builder()
                          .data("part2")
                          .build());
        sink.emit(SseEvent.builder()
                          .data("part3")
                          .build());
    }
}
```
but the result are three different events, because the events are separate by an additional newline like so:
```
event: myevent
data: part1

data: part2

data: part3

```

Shouldn't the `SseEvent` API take `0 - n` number of data entries to follow the specification? There are clients out there, that rely on multiple `data:` lines, each with a dedicated purpose.

spericas:
Yes, this should be properly supported, likely on the client API as well.

PR #10406, opened 2025-07-21
Adds support for multi-line SSE event

spericas:
### Description
Adds support for multi-line SSE events. These are concatenated using '\n' on the client side and reported as a single event. See issue #10380.

### Documentation

None


http/sse/src/main/java/io/helidon/http/sse/SseEvent.java
```diff
@@ -294,7 +294,16 @@ public Builder mediaType(MediaType mediaType) {
...
-            this.data = data;
+            // not set or an override?
+            if (this.data == NO_DATA || !(this.data instanceof String)) {
+                this.data = data;
+            } else {
+                // handle multi-line data
+                if (!(data instanceof String)) {
+                    throw new IllegalArgumentException("Cannot concatenate non-string event data");
+                }
+                this.data += "\n" + data;    // concatenate strings
```
tvallin, line 308: `/n` is by the spec and is not a platform specific new line right ?

webserver/tests/sse/src/test/java/io/helidon/webserver/tests/sse/SseClientTest.java
```diff
@@ -71,6 +72,20 @@ static void sseString1(ServerRequest req, ServerResponse res) {
         }
     }
 
+    static void sseString2(ServerRequest req, ServerResponse res) {
+        try (SseSink sseSink = res.sink(SseSink.TYPE)) {
+            sseSink.emit(SseEvent.builder()
+                                 .data("first line")
```
tvallin, line 78: Do we want to add new method to create a multi line data event like this ?

```java
Builder data(List<String> data);
```
or
```java
Builder data(String... data);
```

tvallin (commented):
Minor questions otherwise looks good !

http/sse/src/main/java/io/helidon/http/sse/SseEvent.java
spericas: That's right

webserver/tests/sse/src/test/java/io/helidon/webserver/tests/sse/SseClientTest.java
spericas: Yes, we can add a method like that. I'm not sure how popular this case is, but it can help.

tvallin (approved):

PR #10414, opened 2025-07-23
4.x Build limit-related span so its parent is set to any propagated parent

### LINKED ISSUE #10323: Concurrency Limit Span being created as an Orphan Span

tjquinno:
### Description
Resolves #10323 

### Documentation

If no doc impact: None

### DIFF
--- a/common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/spi/LimitAlgorithmListener.java
+++ b/common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/spi/LimitAlgorithmListener.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common.concurrency.limits.spi;
+
+/**
+ * Listener to key events as the limit algorithm processes a single instance of incoming work (such as a request).
+ * <p>
+ * Conceptually, an instance of work that is subject to concurrency limiting goes through two stages:
+ * <ol>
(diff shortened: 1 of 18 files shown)


common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/spi/LimitAlgorithmListener.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+public interface LimitAlgorithmListener {
+
+    /**
+     * Invoked when the limit algorithm accepts the work instance for immediate execution (without queueing).
+     *
+     * @param originName name associated with the origin of the work (a socket name, for example)
+     * @param limitName name of the limit algorithm in use
+     */
+    void onAccept(String originName, String limitName);
```
tomas-langer, line 44: I am missing some kind of context for the listener to work with.
For example - we may create a span when accepting the task, and we need the same instance when invoking `onDrop`, `onSuccess`, or `onIgnore` (maybe others, depends on the full flow).

I would expect such a context to be returned from the method that starts the processing (probably `onAccept`), and then passed as a parameter to the end state methods.

Otherwise you need to use some kind of a thread local or other algorithm to keep track of the state, which would be a big burden on the users of this feature.

So I would add a type argument (`LimitAlgorithmListener<T>`), have `T on Accept(...)` etc.

tomas-langer (changes requested):
Please do not merge this PR, we need to discuss how to continue first.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/spi/LimitAlgorithmListener.java
tjquinno: I know we will be discussing this and other points offline, but I did want to reply to your comment.

The _per-request_ listener instance _is_ essentially the context you are talking about.

The code that invokes the limit code prepares the per-request listener so it contains data the listener code needs (HTTP headers in the tracing case). it is passed _from_ the caller _to_ the limit algorithm rather than being _returned_ from a singleton listener and then passed _from_ the limit algorithm _to_ other singleton listener methods.

Using per-instance listeners avoided the need for a thread local or some other kind of data structure. That said, I agree it's a bit unusual to have per-request listener instances rather than a singleton listener. 

The approach you've outlined does not have a way of communicating the listener contexts back to the caller of the limit algorithm, and that's a critical piece. I have ideas for that which we can discuss off-line.

tjquinno:
This PR is superseded by #10464. Closing.

PR #10427, opened 2025-07-25
4.x: Common config now has a config factory

### LINKED ISSUE #10420: gRPC Client has an implicit dependency on helidon-config
## Environment Details
* Helidon Version: 4.2.4, also 4.3.0-SNAPSHOT
* JDK version: 21
* OS: Mac

----------

## Problem Description
When using the gRPC client, non-unary calls result in an exception during class initialization.
```java
Caused by: java.lang.RuntimeException: io.helidon.service.registry.ServiceRegistryException: There is no service in registry that matches this lookup: Lookup{serviceType=Optional.empty,scopes=[],qualifiers=[],contracts=[io.helidon.common.config.Config],contractType=Optional.empty,runLevel=Optional.empty,weight=Optional.empty,includeAbstract=false,dependency=Optional.empty,factoryTypes=[]}
	at io.helidon.metrics.api.MetricsFactoryManager.access(MetricsFactoryManager.java:188)
Caused by: java.lang.RuntimeException: io.helidon.service.registry.ServiceRegistryException: There is no service in registry that matches this lookup: Lookup{serviceType=Optional.empty,scopes=[],qualifiers=[],contracts=[io.helidon.common.config.Config],contractType=Optional.empty,runLevel=Optional.empty,weight=Optional.empty,includeAbstract=false,dependency=Optional.empty,factoryTypes=[]}

	at io.helidon.metrics.api.MetricsFactoryManager.getMetricsFactory(MetricsFactoryManager.java:129)
	at io.helidon.metrics.api.MetricsFactory.getInstance(MetricsFactory.java:56)
	at io.helidon.metrics.api.Tag.create(Tag.java:35)
	at io.helidon.webclient.grpc.GrpcBaseClientCall.<clinit>(GrpcBaseClientCall.java:86)
	... 1 more
Caused by: io.helidon.service.registry.ServiceRegistryException: There is no service in registry that matches this lookup: Lookup{serviceType=Optional.empty,scopes=[],qualifiers=[],contracts=[io.helidon.common.config.Config],contractType=Optional.empty,runLevel=Optional.empty,weight=Optional.empty,includeAbstract=false,dependency=Optional.empty,factoryTypes=[]}
Caused by: io.helidon.service.registry.ServiceRegistryException: There is no service in registry that matches this lookup: Lookup{serviceType=Optional.empty,scopes=[],qualifiers=[],contracts=[io.helidon.common.config.Config],contractType=Optional.empty,runLevel=Optional.empty,weight=Optional.empty,includeAbstract=false,dependency=Optional.empty,factoryTypes=[]}

	at io.helidon.service.registry.CoreServiceRegistry.supply(CoreServiceRegistry.java:192)
	at io.helidon.service.registry.CoreServiceRegistry.get(CoreServiceRegistry.java:180)
	at io.helidon.service.registry.CoreServiceRegistry.get(CoreServiceRegistry.java:175)
	at io.helidon.service.registry.ServiceRegistry.get(ServiceRegistry.java:48)
	at io.helidon.service.registry.Services.get(Services.java:156)
	at io.helidon.metrics.api.MetricsFactoryManager.externalMetricsConfig(MetricsFactoryManager.java:159)
	at java.base/java.util.Objects.requireNonNullElseGet(Objects.java:334)
	at io.helidon.metrics.api.MetricsFactoryManager.lambda$getMetricsFactory$5(MetricsFactoryManager.java:130)
	at io.helidon.metrics.api.MetricsFactoryManager.access(MetricsFactoryManager.java:186)
	... 5 more
```
This happens because the `Config` is not availableIn the service registry. When the `io.helidon.config:helidon-config` artifact is added to the class path, the problem goes away as the `Config` does get added to the service registry. However, neither the gRPC Client, nor the Metrics API artifacts have a dependency on the `helidon-config`, though it is seemingly required.

I was unable to find a workaround by calling `Services.set(Config.class...)` or similar to populate a `Config` into the registry.

## Steps to reproduce
I'm attaching a small program that reproduces the problem by directly instantiating the class that fails to initialize. In a real project, this class gets invoked through the public APIs.

[helidon-repro.zip](https://github.com/user-attachments/files/21393016/helidon-repro.zip)

tomas-langer:
So we have an instance in ServiceRegistry even when no implementation is on classpath

Resolves #10420

### DIFF
--- a/common/config/src/main/java/io/helidon/common/config/ConfigFactory.java
+++ b/common/config/src/main/java/io/helidon/common/config/ConfigFactory.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.common.config;
+
+import java.util.function.Supplier;
+
+import io.helidon.common.Weight;
+import io.helidon.common.Weighted;
+import io.helidon.service.registry.Service;
+
+/**
+ * Default factory for configuration, that returns the current global config, or creates a new config instance.
+ * This is to ensure we have a service registry instance even when there is no implementation on classpath.
+ */
+@SuppressWarnings("removal")
+@Weight(Weighted.DEFAULT_WEIGHT - 20)
+@Service.Singleton
+class ConfigFactory implements Supplier<Config> {
+    @Override
+    public Config get() {
+        // once GlobalConfig gets removed, we just return Config.create()
+        return GlobalConfig.configured() ? GlobalConfig.config() : Config.create();
+    }
+}
--- a/common/config/pom.xml
+++ b/common/config/pom.xml
@@ -46,6 +46,10 @@
             <artifactId>helidon-service-registry</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-types</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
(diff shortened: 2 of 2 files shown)


common/config/src/main/java/io/helidon/common/config/ConfigFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.function.Supplier;
+
+import io.helidon.common.Weight;
+import io.helidon.common.Weighted;
+import io.helidon.service.registry.Service;
+
+/**
+ * Default factory for configuration, that returns the current global config, or creates a new config instance.
+ * This is to ensure we have a service registry instance even when there is no implementation on classpath.
```
ljnelson: "service registry instance": is this a typo?

tomas-langer: I do not think so.
We need to have a `Config` instance in service registry, that is why this factory was created. I do not see the typo - maybe bad english on my side?

barchetta (approved):

common/config/src/main/java/io/helidon/common/config/ConfigFactory.java
ljnelson: I was trying to understand how the presence of a config would ensure we had a service registry.

tomas-langer: yep, not great on my side, by "service registry instance" I mean "an instance in service registry".

spericas (approved):
LGTM

Issue #10435, opened 2025-07-28
Add custom `Gson` type adapter factories during runtime


arouel:
Helidon Version: 4.2.x

## Enhancement Description

It is not currently possible to add custom [Gson](https://github.com/google/gson) type adapters in a modular way at runtime. Adding them requires manually instantiating [GsonSupport](https://github.com/helidon-io/helidon/blob/main/http/media/gson/src/main/java/io/helidon/http/media/gson/GsonSupport.java), which undermines the modular approach of extending Helidon with media type implementations.

To resolve this limitation, I suggest to utilize the [ServiceLoader](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/ServiceLoader.html) for adding custom `TypeAdapterFactory` instances as service providers. These providers are loaded during the initialization of `GsonSupport` by the `GsonMediaSupportProvider`. This allows in a modular way to extend Gson's functionality within Helidon.

### 1. Define the Data Class

First, define the class you want to serialize or deserialize.

```java
// src/main/java/my/pkg/Book.java
package my.pkg;

public record Book(String title, int pages) {
}
```

### 2. Implement the TypeAdapterFactory

Next, create a `TypeAdapterFactory` that handles the `Book` class.

```java
// src/main/java/my/pkg/BookTypeAdapterFactory.java
package my.pkg;

import com.google.gson.Gson;
import com.google.gson.TypeAdapter;
import com.google.gson.TypeAdapterFactory;
import com.google.gson.reflect.TypeToken;
import com.google.gson.stream.JsonReader;
import com.google.gson.stream.JsonWriter;

import java.io.IOException;

public class BookTypeAdapterFactory implements TypeAdapterFactory {

    private static final TypeAdapter<Book> instance = new TypeAdapter<>() {
        @Override
        public void write(JsonWriter writer, Book book) throws IOException {
            writer.beginObject();
            writer.name("title");
            writer.value(book.title());
            writer.name("pages");
            writer.value(book.pages());
            writer.endObject();
        }

        @Override
        public Book read(JsonReader reader) throws IOException {
            reader.beginObject();
            String title = null;
            int pages = 0;
            while (reader.hasNext()) {
                switch (reader.nextName()) {
                    case "title" -> title = reader.nextString();
                    case "pages" -> pages = reader.nextInt();
                    default -> reader.skipValue();
                }
            }
            reader.endObject();
            return new Book(title, pages);
        }
    };

    @Override
[...]

PR #10437, opened 2025-07-28
Add custom `Gson` type adapter factories during runtime

arouel:
### Description

It is not currently possible to add custom `Gson` type adapters in a modular way at runtime. Adding them requires manually instantiating `GsonSupport`, which undermines the modular approach of extending Helidon with media type implementations.

To resolve this, I have added the ability to add custom `TypeAdapterFactory` instances as service providers. These providers are loaded during the initialization of `GsonSupport` by the `GsonMediaSupportProvider` using Java's `ServiceLoader` API. This allows in a modular way to extend Gson's functionality within Helidon.

More details can be found in the issue https://github.com/helidon-io/helidon/issues/10435.


arouel:
@eiselems maybe this is of interest to you.

eiselems:
@tomas-langer sounds like something that I missed, maybe you can also have a look

arouel:
@Verdent @tomas-langer do you have feedback about this change request? I would appreciate if this get merged soon.

http/media/gson/src/main/java/io/helidon/http/media/gson/GsonSupport.java
```diff
@@ -82,6 +84,12 @@ public static MediaSupport create(Config config, String name) {
         Objects.requireNonNull(config, "Config must not be null");
         Objects.requireNonNull(name, "Name must not be null");
 
+        GsonBuilder gsonBuilder = new GsonBuilder();
+        // Enable the registering of custom type adapters by using service providers for TypeAdapterFactory.
+        for (var factory : ServiceLoader.load(TypeAdapterFactory.class)) {
```
tomas-langer: - there must be a `uses com.google.gson.TypeAdapterFactory` statement in `module-info.java`, otherwise this would fail on module path
- we never use Java service loader directly, please use `io.helidon.common.HelidonServiceLoader#create(java.lang.Class<T>)` - this supports `@Weight` to order providers
- the `gson` instance you create is lost

How to do this to align with current implementation:

- add the following method to `GsonSupportConfigBlueprint` to allow users to register them programmatically:
```
/**
 * Additional type adapter factories.
 *
 * @returns type adapter factories
 */
@Option.Singular("typeAdapterFactory")
List<TypeAdapterFactory> typeAdapterFactories();
```
- add the discovered factories to the builder created on line 85 (in the original code)
- update method `io.helidon.http.media.gson.GsonSupport.Decorator#decorate` - add the factories from the `target` builder to the `GsonBuilder` instance.

arouel:
@tomas-langer I made changes according to your points. Is there still something missing?

Issue #10370, opened 2025-07-11
Implement Observability support around LangChain4J API calls


arjav-desai:
See https://docs.langchain4j.dev/tutorials/observability

PR #10446, opened 2025-07-29
Add Observability support around LangChain4J API calls

arjav-desai:
### Description

Addresses #10370 

### Documentation

Doc is added with this PR.


integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
```diff
@@ -0,0 +1,131 @@
...
+    @Override
+    public void onResponse(ChatModelResponseContext chatModelResponseContext) {
+        final long endTime = System.nanoTime();
+        final long startTime = (Long) chatModelResponseContext.attributes().get(GEN_AI_CLIENT_OPERATION_START_TIME);
+
+        final ChatRequest chatRequest = chatModelResponseContext.chatRequest();
+        final ChatResponse chatResponse = chatModelResponseContext.chatResponse();
+
+        DistributionSummary.Builder inputClientTokenUsageBuilder = this.clientTokenUsageBuilder
+                .addTag(Tag.create("gen_ai.operation.name", "chat"))
```
tjquinno: To improve efficiency, some of the tags are constant and could be added to the builders once, at the top.

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
```diff
@@ -0,0 +1,131 @@
...
+                .addTag(Tag.create("gen_ai.request.model", chatRequest.modelName()))
```
tjquinno: Might this singleton service be invoked concurrently from multiple threads? 

If so, then there could be a race condition in which different threads are trying to set the tag--especially one that depends on the model name--and interfere with each other.

The builder supports a `from` method which is essentially a clone operation. If this code needs to be thread-safe then `onResponse` could create a new builder, on the new instance invoke `from` passing the original, unchanging ones initialized in the constructor. Then modify the tag on the new builder and use that updated new builder in the `getOrCreate` invocation.

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
```diff
@@ -0,0 +1,131 @@
...
+                .addTag(Tag.create("gen_ai.token.type", "input"));
```
tjquinno: I assume there are  relatively few distinct model names that appear in the requests and responses. 

In that case, this code might be must faster if it maintains a thread-safe map from model name to meter (not to builder, to the constructed meter).

Then `onResponse` would use `map.computeIfAbsent` to get the meter if it exists and create it just-in-time otherwise. That would save constructing the builder and invoking `getOrCreate` on subsequent requests involving the same model name.

(There would be a map for each of the builders that the constructor creates.)

arjav-desai: The model names are more than a few i.e. the provider list is https://docs.langchain4j.dev/integrations/language-models/ and some, if not most, of them support multiple models e.g. OpenAI has gpt-30, gpt-40 and now gpt-50 too.

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
arjav-desai: I will move this tag in constructor with review comments fix. Thanks.

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
tjquinno: Thanks for the link. Very useful.

I think it's very important for performance to use a local thread-safe map to hold the distribution summaries instead of creating builders and doing the meter registry lookup on each request.

If even one tag value is different between two meter IDs, then that means they are different meters. And presumably we want to be able to segregate measurements based on the model name: one pair of summaries  for Amazon Bedrock, a separate pair for Anthropic, etc. That's why the code uses the model name for the tag value. 

Also presumably there will be far more requests than there are distinct model names. We will save a lot of execution time using a map.

Maybe something like:
```java
    private record Summaries(DistributionSummary clientTokenUsage, 
                DistributionSummary clientOperationDuration) {}
    
    private Map<String, Summaries> summariesByModelName = new ConcurrentHashMap<>(); // now that pinning is fixed
```
Then `onResponse` would use 
```java
        summariesByName.computeIfAbsent(modelName, name -> {
        // Code to:
        //   - use the two template builders to create two model-specific builders
        //   - use registry.getOrCreate to register the two new model-specific summaries, 
        //   - create a new Summaries instance holding the two new distribution summaries for this model name
        });
````

Comment (arjav-desai) [reply integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java]: Fixed it.

File: integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java (hunk 2 above)
Comment (arjav-desai) [reply integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java]: Fixed it.

File: integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java [hunk 4]
```diff
@@ -0,0 +1,181 @@
...
+
+    @Override
+    public void onResponse(ChatModelResponseContext chatModelResponseContext) {
+        final long endTime = System.nanoTime();
+        final long startTime = (Long) chatModelResponseContext.attributes().get(GEN_AI_CLIENT_OPERATION_START_TIME);
+
+        final ChatRequest chatRequest = chatModelResponseContext.chatRequest();
+        final ChatResponse chatResponse = chatModelResponseContext.chatResponse();
+
+        responseInputTokenUsageSByModelName.computeIfAbsent(
```
tjquinno: In each case, save the return value from the `computeIfAbsent` call in a variable because it returns either the previously-stored value in the map or the new, just-stored one. Then invoke `record` on that variable, avoiding the need to look up the metric again in the next statement.

Otherwise this looks very good. It should perform much better.

tjquinno (changes requested):
Looking very good. One further requested change (same change in four places in the key method).

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
arjav-desai: Fixed it.

tjquinno (approved):

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
```diff
@@ -0,0 +1,184 @@
...
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+
+/**
+ * Creates metrics that follow the
+ * <a href="https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/">Semantic Conventions for GenAI Metrics</a>.
+ */
+@Service.Singleton
+@Service.Named("MetricsChatModelListener")
+public class MetricsChatModelListener implements ChatModelListener {
```
tomas-langer: This class should not be public

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
```diff
@@ -0,0 +1,184 @@
...
+    private final DistributionStatisticsConfig.Builder clientOperationDurationStatisticsConfigBuilder;
+    private final Map<String, DistributionSummary> responseInputTokenUsageByModelName = new ConcurrentHashMap<>();
+    private final Map<String, DistributionSummary> responseOutputTokenUsageByModelName = new ConcurrentHashMap<>();
+    private final Map<String, DistributionSummary> responseOperationDurationByModelName = new ConcurrentHashMap<>();
+    private final Map<String, DistributionSummary> errorOperationDurationByModelName = new ConcurrentHashMap<>();
+
+    /**
+     * Constructs a {@code MetricsChatModelListener} instance.
+     */
+    public MetricsChatModelListener() {
```
tomas-langer: Never use public constructors

tomas-langer (changes requested):

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
arjav-desai: Fixed it.

integrations/langchain4j/langchain4j/src/main/java/io/helidon/integrations/langchain4j/MetricsChatModelListener.java
arjav-desai: Fixed it.

tjquinno (dismissed):

Issue #10433, opened 2025-07-28
gRPC Client doesn't work with V2 blocking stubs


MariusVolkhart:
## Environment Details
* Helidon Version: 4.3.0-SNAPSHOT
* Helidon SE or Helidon MP: SE
* JDK version: 21
* OS: Mac

----------

## Problem Description
Helidon 4.3.0 upgrades the gRPC dependency to 1.73.0. In turn, [that exposes](https://github.com/grpc/grpc-java/releases/tag/v1.70.0) the [experimental "V2 blocking"](https://github.com/grpc/grpc-java/issues/10918) APIs based on `BlockingClientCall`. However, the Helidon gRPC client does not work with these APIs.

Specifically, when server streaming is involved, the gRPC client displays two behavior, both incorrect:

1. The client does not expose messages through `BlockingClientCall#read()` until the server calls `onComplete` or `onError`. This effectively means messages are stuck in the queue and the client deadlocks.
2. If the server does call `onComplete` of `onError`, only the first message of the stream is exposed through calls to `BlockingClientCall#read()`.

Using identical code in a production service, the gRPC Netty client works fine under these conditions and with these APIs. Swapping to the Helidon gRPC client, the stubs stop working correctly.

The problem is reproducible 100% of the time, and can be reasonably demonstrated in unit tests. This was done in https://github.com/helidon-io/helidon/compare/main...MariusVolkhart:helidon:mv/v2

These tests demonstrate usages of all the new APIs, and expose their various failure modes. For example,
```java
io.grpc.StatusException: UNKNOWN: This buffer has 3 bytes, requested to read at 3

	at io.grpc.Status.asException(Status.java:540)
	at io.grpc.stub.BlockingClientCall.read(BlockingClientCall.java:133)
	at io.grpc.stub.BlockingClientCall.read(BlockingClientCall.java:90)
	at io.helidon.webclient.grpc.tests.GrpcStubTest.testServerStreamingInfiniteV2(GrpcStubTest.java:151)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
[...]

PR #10453, opened 2025-07-30
4.x: Improves support for new gRPC V2 blocking stub calls

spericas:
### Description

Improves support for V2 blocking stub calls. Handles use case of messages in receiving queue that need to be delivered after all data from the server stream is consumed. Some new tests. Issue #10433.

### Documentation

None


tjquinno (approved):
LGTM

PR #10464, opened 2025-07-31
4.x Record limit outcomes, using them to build limit-related spans with correct parent

### LINKED ISSUE #10323: Concurrency Limit Span being created as an Orphan Span

tjquinno:
### Description
Resolves #10323

I have adjusted the approach since the PR was first created. This description describes the most recent design.

## Release note
_____
Helidon creates a tracing span to represent a request's waiting time due to concurrency limits. If the incoming request's headers propagate a tracing span the waiting-time span now has the propagated span as its parent.
_____

### TL;DR about the most recent changes
Previously, the PR added a new `Consumer<Outcome>` parameter to the existing methods. Although this changed the existing signatures as little as possible that is not how we will want the API to look in 5.0.

So instead of adding the `Consumer` parameter, this latest set of changes alters the `LimitAlgorithm` interface so all methods return the `LimitAlgorithm.Outcome` object (instead of the former `Optional<Token>`. Existing methods which return `Optional<Token>` remain for backward compatibility but are deprecated.

### Solution strategy
To minimize the performance impact of this fix, the general approach is as follows, specifically for the tracing span:
1. Once the limit algorithm decides how to deal with a request, use the request context to save the information needed to create the span.
2. Later, during request processing--in the existing tracing observer filter to be precise--create a "wait time" span if in fact the request spent time in the virtual limit algorithm queue. That filter already computed any propagated inbound parent span from the incoming request's headers, so that parent span is now also used in setting the parent for the new "wait time" span if needed.

That said, to accomplish this in a general way involves an enhancement to the limit algorithm mechanism to report the algorithm outcome back to the caller. This must be separate from the token; tokens are returned only when the request is accepted, but in general the caller might need to know about all outcomes, accepted _and_ rejected.

### Background
The existing limit algorithms essentially deal with an incoming item of work (such as an HTTP request) in two steps:
1. Disposition
   
   The algorithm decides how to deal with the work item:
   * immediate acceptance - The algorithm allows processing to start on the work item immediately, without queueing.
   * immediate rejection - There is no capacity for _running_ more work immediately (so the new work cannot be started right away) _and_ the virtual queue of waiting work  is already full so this new item cannot be added to the queue.
   * deferred - The algorithm cannot allow the work item to proceed right away but it might be able to later so it adds the work item to the virtual queue. 
     
     One of two things happens to a queued work item:
     * deferred acceptance - Enough work in progress completes so the limit algorithm removes the work item from the virtual  queue and allows it to proceed.
     * deferred rejection - The work item spends too long in the virtual queue (according to the algorithm's configuration) and so the algorithm does not allow the work item to run.
2. Processing
   
   If the algorithm accepts the work item--either immediately or deferred--it returns a `Token` to the caller. 
   
   Once the caller has processed the work item, it invokes exactly one of the following methods on the `Token` to indicate that the work item is no longer managed by the limit algorithm and why:
   * `success` - processed fully. This does _not_ necessarily mean that the work item itself succeeded, only that the system processed it fully and so the time used contributues meaningfully to any measurements of the system's response.
   * `ignore`- failed before meaningful progress was made. Any time spent _does not_ contribute to an accurate picture of how the system processes this type of work item.
   * `dropped` - processed signficantly but not fully. The work item might have timed out, for example. The time spent should be included in aggregate measurements.

## Basic concepts of the solution
The `LimitAlgorithm` interface now requires--and the two limit algorithms, fixed and AIMD, now implement--methods analogous to the previous `tryAcquire` and `invoke` ones but which return a new `LimitAlgorithm.Outcome` object instead of the former `Optional<Token>`. This object  represents the decision the algorithm made about a particular request. The new methods always return the `Outcome` to the caller, regardless of whether the limit algorithm granted a `Token` or not.

The webserver saves the outcome in the `RoutingRequest` it creates (for either http 1 or 2). When some other code asks for the request context, and (as before) the request creates the context on-first-access, that code also registers the `LimitOutcome` in the request context just in time.

## Details
* Add the `LimitAlgorithm.Outcome` interface.
  
  This represents all limit-related information about the limit algorithm's decision regarding a single work item. 
   * accepted or rejected
     * If accepted, the `Token` created (similar to before)
   * queued or not 
      * if queued, start and end wait times in nano system times
   
   The `Outcome` interface also includes static factory methods for the various combinations of outcome--accepted/rejected and immediate/deferred--so all implementations of `LimitAlgorithm` have easy ways to create suitable `Outcome` instances.

* Add the `LimitAlgorithm.Result` interface.
  
  Holds both the `Outcome` and the result from executing a `Callable`. When the caller passes in a `Callable<T>` we need a way to return both the `Outcome` and the callable's return value; this is the carrier for both.
* On `LimitAlgorithm`:
  * Deprecate the existing `tryAcquire` and `invoke` methods which returned `Optional<Token>` on `LimitAlgorithm`.  
  * Add replacements for those methods:
    * `Outcome tryAcquireOutcome` (with and without the `boolean` parameter)
    * `Result<T> call(Callable<T>)`
    * `Outcome run(Runnable)`

* Change `LimitHandlers.LimitHandler` (used internally, not public)
  * No longer implement `LimitAlgorithm` (it pretty much did so only to specify the old `Optional<Token> tryAcquire` behavior)
  * Add `Optional<Token> tryAcquireToken` which functions as the previous `tryAcquire` method did.

* Change the existing `FixedLimit` and `AimdLimit` implementations:
  * Provide Implementations for the new methods on `LimitAlgorithm`.
  * Implement the new non-public `LimitAlgorithmDeprecatedImpls` interface.
    
    This `LimitAlgorithmDeprecatedImpls` interface basically declares default implementations for methods that are now deprecated on `LimitAlgorithm`, but it does so using methods on `LimitAlgorithm` that are _not_ deprecated. This way, the fixed and AIMD implementations focus on implementing the non-deprecated methods, and therefore are _mostly_ as they will be when we remove all the deprecated items. This keeps the two implementation classes cleaner now and reduces our work later when we remove deprecated items: basically just remove the `implements LimitAlgorithmDeprecatedImpls` clause.

* The http1 and http2 request processing code (currently the only clients of the concurrency limit component) now saves the limit outcome in each request it creates. The `context()` methods create the request context on-first-access as before but now they also register the `LimitOutcome` in the newly-created request context.

* The existing tracing filter now also conditionally creates a span for the request's waiting time. The filter retrieves the limit outcome from the request context, if present, and uses the information in it to create the wait time span. 
  
  The filter already derived a parent span context if one was conveyed using the request's headers. The filter uses that same parent as the parent for the wait time span.

* The `TracingObserverConfigBlueprint` has the new `waitTracingEnabled()` method which defaults to `false`.

## Documentation
The PR includes a doc update illustrating the new tracing observer config setting.


tjquinno:
Re-open PR; this is the active one

webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserver.java
```diff
@@ -266,6 +274,38 @@ Find configuration of the web server span (can customize name, disable etc.)
             }
         }
 
+        private void recordLimitWaitingSpan(SpanContext inboundSpanContext, RoutingRequest req) {
+            Optional<Limit> limit = req.context().get(Limit.class);
+            if (limit.isEmpty()) {
+                return;
+            }
+            boolean tracingEnabled = (limit.get() instanceof FixedLimit fixedLimit)
```
tomas-langer: This is very strange - the tracing of limits should not be a function of the limit itself, rather a function of tracing observer, if this is implemented in tracing observer.
I think the option should be on tracing observer configuration, as then
- you could remove Limit itself from context
- you would get rid of this terrible nested ternal expression
- the configuration of tracing would be on a  component that actually understands tracing, rather then on some limit that should be abstracted away from tracing

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimit.java
```diff
@@ -50,9 +50,11 @@ public class AimdLimit implements Limit, SemaphoreLimit, RuntimeType.Api<AimdLim
     private final AimdLimitConfig config;
     private final AimdLimitImpl aimdLimitImpl;
 
+    private String originName;
+
     private AimdLimit(AimdLimitConfig config) {
         this.config = config;
-        this.aimdLimitImpl = new AimdLimitImpl(config);
+        this.aimdLimitImpl = new AimdLimitImpl(config, originName);
```
spericas: Is `originName` initialized at this time?

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
```diff
@@ -100,22 +102,42 @@ int currentLimit() {
...
+    Optional<LimitAlgorithm.Token> tryAcquire(boolean wait, Consumer<LimitOutcome> outcomeConsumer) {
+        Objects.requireNonNull(outcomeConsumer, "limit algorithm outcome consumer cannot be null");
         Optional<LimitAlgorithm.Token> token = handler.tryAcquire(false);
         if (token.isPresent()) {
+            LimitOutcomeImpl.Accepted outcome = LimitOutcomeImpl.createAccepted(originName, AimdLimit.TYPE);
+            if (token.get() instanceof OutcomeAwareToken outcomeAwareToken) {
+                outcomeAwareToken.outcome(outcome);
+            }
+            outcomeConsumer.accept(outcome);
             return token;
```
spericas: All this work above is wasted if the consumer is empty, right? Could this be avoided?

spericas: I guess this comment applies to similar patterns below. We should aim for near zero cost when consumers are dummy.

webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserver.java
spericas: Could you only register the `LimitOutcome` in the context when tracing is enabled and drop this (ugly) check here?

tjquinno: Enabling tracing on the limit config follows the pattern previously set by the existing limits metrics approach. Metrics for concurrency limits is enabled/disabled on the fixed and AIMD limit configs (they share no common supertype), so `tracingEnabled()` was added to the same places.

The config setting for enabling metrics or tracing for limits should be in the most natural place for users, separate from where in the code we choose to actually create the span (or update the metrics).

The work to create the waiting span is in the tracing observer partly in response to comments on earlier PRs to make sure it is out of the performance-critical code path and partly to reuse the parent span calculation. That's an internal implementation choice; the fact that the tracing observer code manages the limit span doesn't matter to users.

A more general approach Santiago suggested earlier--which was in an earlier PR but not in this one and could use similar techniques to support tracing and metrics and anything else--would introduce limit algorithm listeners that could be added explicitly to the limit builders (and/or discovered automatically). Each specific listener--a metrics limit listener and a tracing limit listener in particular--could have its own enabled/disabled setting. This would allow the metrics and tracing enabled settings to be moved away from the concurrency limit config itself. That said, the `enableMetrics` setting on the limit config is in released 4.2.4 code so we'd have to be careful if we did that.

Maybe the solution does need to introduce the listener concept and revise how metrics are handled as well.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
tjquinno: We could refactor this code a bit it so a private method accepts a null for the consumer and not use it if it is null.

webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserver.java
tjquinno: The major revision to the PR avoids this. Individual `LimitAlgorithmListener`s can be enabled or disabled in their own configuration.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimit.java
tjquinno: Good catch. It's available only when `AimdLimit.init` is invoked which in turn invokes `AimdLimitImpl.initMetrics`. 

Fixed.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitImpl.java
tjquinno: The new approach goes to extensive lengths to avoid work if it does not apply. Among them, internally the code uses `null` to indicate the lack of a consumer, avoiding  whatever overhead from invoking an empty consumer might not be optimized away.

spericas (dismissed):
Definitely simpler than the last version!

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimitConfigBlueprint.java
```diff
@@ -32,9 +32,10 @@
  * @see #queueLength()
  * @see #queueTimeout()
  */
-@Prototype.Blueprint
+@Prototype.Blueprint()
```
tomas-langer: This is not a needed change

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimitConfigBlueprint.java
```diff
@@ -32,9 +32,10 @@
...
 @Prototype.Configured(value = FixedLimit.TYPE, root = false)
 @Prototype.Provides(LimitProvider.class)
+@Prototype.RegistrySupport
```
tomas-langer: And I think this one is not needed either

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitConfigBlueprint.java
```diff
@@ -147,4 +147,5 @@ interface AimdLimitConfigBlueprint extends Prototype.Factory<AimdLimit> {
     @Option.Configured
     @Option.DefaultBoolean(false)
     boolean enableTracing();
+
```
tomas-langer: `enableTracing` should not exist on this blueprint

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitOutcome.java
```diff
@@ -0,0 +1,139 @@
...
+ * <h2>Execution result</h2>
+ * If the algorithm accepts a work item, then later the code using the limit algorithm should
+ * invoke exactly one of the token's {@code dropped}, {@code ignore}, or {@code success} methods.
+ */
+public interface LimitOutcome {
+
+    /**
+     * Describes the disposition of a work item based on the concurrency limit configuration and current state.
+     */
+    enum Disposition {
```
tomas-langer: Is there too many combinations to create a single enum with possible outcomes?
I would prefer if there was a direct mapping between the enum value and the interface the outcome implements, so there is no need to do `instanceof` checks anywhere when using this.

common/concurrency/limits/src/main/resources/META-INF/helidon/native-image/reflection-config.json
```diff
@@ -0,0 +1,11 @@
+{
+    "annotated":[
```
tomas-langer: This class no longer exists, so this file can probably be deleted as well

common/concurrency/limits/pom.xml
```diff
@@ -105,6 +105,11 @@
                             <artifactId>helidon-codegen-helidon-copyright</artifactId>
                             <version>${helidon.version}</version>
                         </path>
+                        <path>
```
tomas-langer, line 107: This does not seem to be used.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitAlgorithm.java
```diff
@@ -61,10 +62,36 @@ default <T> T invoke(Callable<T> callable) throws LimitException, Exception {
...
+     * @param callable                 callable to execute within the limit
+     * @param limitOutcomeConsumer     consumer of contexts provided by limit algorithm listeners
+     * @param <T>                      the callable return type
+     * @return result of the callable
+     * @throws LimitException      in case the limiter did not have an available permit
+     * @throws java.lang.Exception in case the task failed with an exception
+     */
+    default <T> T invoke(Callable<T> callable, Consumer<LimitOutcome> limitOutcomeConsumer)
+            throws LimitException, Exception {
+        return invoke(callable);
```
tomas-langer: when this default method is invoked (i.e. for a custom limit that was not updated to our new code), the `limitOutcomeConsumer` would never get called, and there would be no limit outcome present -> this will cause NPE down the code.
We should have some default limit outcome always sent, as otherwise we can end up with a lot of problems.

webserver/observe/tracing/pom.xml
```diff
@@ -70,6 +70,10 @@
             <artifactId>hamcrest-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
tomas-langer: Can you please move the compile dependency before test dependencies?

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/AimdLimitConfigBlueprint.java
tjquinno: Fixed

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimitConfigBlueprint.java
tjquinno: Fixed

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimitConfigBlueprint.java
tjquinno: Fixed

common/concurrency/limits/src/main/resources/META-INF/helidon/native-image/reflection-config.json
tjquinno: Fixed

common/concurrency/limits/pom.xml
tjquinno: Fixed

webserver/observe/tracing/pom.xml
tjquinno: Done

webserver/observe/tracing/src/main/java/io/helidon/webserver/observe/tracing/TracingObserver.java
tjquinno: The latest version of the PR excludes listeners altogether so this is no longer an issue.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitOutcome.java
tjquinno: I see your point.

Because there are two independent characteristics (accepted vs. rejected and immediate vs. deferred) it seems  clearer if we keep the existing `Disposition` and add a new enum to describe when the decision was made:
```
enum TIming {
  IMMEDIATE,
  DEFERRED
}
```
This way code can check one `if` condition to see if the request was accepted or rejected. Same for if it was decided immediately or deferred.

Expanding `Disposition` to cover all possibilities would combine two independent characteristics of the outcome.  Finding out if a request was accepted requires the `if` to check for _two_ values. Same for finding out if a request was deferred. 

Plus, although it's unlikely, if one new choice along one axis emerged down the road then we'd have a double increase in the number of values of the revised `Disposition`.

I've pushed a commit adding `Timing`.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitAlgorithm.java
tjquinno: Nice. Interesting and good catch.

We could document in the Javadoc that the consumer _might_ be called for every decision. But returning a default outcome sounds better.

I think that means there are new `Disposition` and  `Timing` values: `UNKNOWN`!

In the case you have identified we do not know anything about how the decision was made, so we should not pretend we do. I'll add that.

tjquinno: Nice. Interesting and good catch.

We could document in the Javadoc that the consumer _might not_ be called, depending on the implementation of the limit algorithm. But returning a default outcome sounds better.

There are new `Disposition` and  `Timing` values: `UNKNOWN`!

In the case you have identified we do not know anything about how the decision was made, so we should not pretend we do by using one of the other implementations with zeros or dummy values.

common/concurrency/limits/pom.xml
```diff
@@ -123,8 +123,8 @@
                         <version>${helidon.version}</version>
                     </dependency>
                     <dependency>
-                        <groupId>io.helidon.codegen</groupId>
-                        <artifactId>helidon-codegen-helidon-copyright</artifactId>
+                        <groupId>io.helidon.service</groupId>
```
tomas-langer, line 136: You hve removed the copyright library, which is wrong, it should stay in as annotation processor path

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitAlgorithm.java
```diff
@@ -41,54 +41,126 @@ public interface LimitAlgorithm {
...
-            runnable.run();
-            permit.success();
-        } catch (IgnoreTaskException e) {
-            permit.ignore();
-            e.handle();
-        } catch (Exception e) {
-            permit.dropped();
-            throw e;
+    @Deprecated(since = "4.3.0", forRemoval = true)
+    void invoke(Runnable runnable) throws LimitException, Exception;
```
tomas-langer: This is a backward incompatible change - method was `default`, must stay `default`

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitAlgorithmOutcomeImpl.java
```diff
@@ -0,0 +1,242 @@
...
+    public Disposition disposition() {
+        return disposition;
+    }
+
+    @Override
+    public Timing timing() {
+        return timing;
+    }
+
+    private void junk(Outcome outcome) {
```
tomas-langer: Please remove this.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
```diff
@@ -207,27 +206,27 @@ public void handle(Limit limit) throws InterruptedException {
...
-                    throw RequestException.builder()
-                            .setKeepAlive(false)
-                            .status(Status.SERVICE_UNAVAILABLE_503)
-                            .type(EventType.OTHER)
-                            .message("Too Many Concurrent Requests")
-                            .build();
-                } else {
-                    LimitAlgorithm.Token permit = token.get();
+                LimitAlgorithm.Outcome outcome = limit.tryAcquireOutcome(true);
+                if (outcome instanceof LimitAlgorithm.Outcome.Accepted accepted) {
```
tomas-langer: `instanceof` here seems like a wrong abstraction.
There should be a `boolean` or an enum method on the `outcome` to identify this.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
```diff
@@ -644,7 +648,7 @@ private void consumeEntity(Http1ServerRequest request, Http1ServerResponse respo
     }
 
     private void handleRequestException(RequestException e) {
-        // gather error handling properties
+         // gather error handling properties
```
tomas-langer: a new space added, seems wrong

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerRequest.java
```diff
@@ -142,6 +149,9 @@ public Context context() {
                     .parent(ctx.listenerContext().context())
                     .id("[" + serverSocketId() + " " + socketId() + "] http/1.1: " + requestId)
                     .build());
+            if (limitOutcome != null) {
```
tomas-langer: This null check should not be done - we should always have an outcome

common/concurrency/limits/pom.xml
tjquinno: I removed that because of your earlier comment https://github.com/helidon-io/helidon/pull/10464#discussion_r2269710938 

Maybe I misunderstood?

Restored.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/LimitAlgorithmOutcomeImpl.java
tjquinno: Fixed.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
tjquinno: `LimitAlgorithm.Outcome` does have the enum `Disposition` with `ACCEPTED` and `REJECTED` values, but an outcome properly has a token only if it represents an "accepted" outcome, so that is why the `token()` method should be only on the `Outcome.Accepted` interface. 

Even if the code in `Http1Connection` (and elsewhere) checked the enum first it would still need to cast the `Outcome` to `Outcome.Accepted` to get the token to operate on it, at least the way `Outcome` and `Outcome.Accepted` are now.

An alternative would have been to have an `Optional<Token>` at the higher-level `Outcome` interface, but IMO that would be less clear and would represent the same information in two places: both the enum and the `Optional<Token>` being present or absent would convey the same information.

And we wanted to have the enums clearly represent what a particular `Outcome` represents: accepted/rejected on one axis and immediate/deferred on another axis.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
tjquinno: Fixed

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerRequest.java
tjquinno: Changed.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
tomas-langer: I still prefer to check enum/boolean and then casting to doing instanceof

common/concurrency/limits/pom.xml
tomas-langer: Comments are positioned on the line that is marked - i.e. the comment is for the code under it, not above it. So just the way github views it.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/FixedLimit.java
```diff
@@ -321,15 +281,91 @@ public void init(String socketName) {
...
+    private class FixedToken implements Token {
         private final long startTime;
 
         private FixedToken(Supplier<Long> clock, AtomicInteger concurrentRequests) {
             startTime = clock.get();
             concurrentRequests.incrementAndGet();
         }
 
-        @Override
         public void dropped() {
```
tomas-langer: Why was the `Override` removed?

tjquinno: Good question. I do not remember removing them manually. Maybe some auto-reformatting at an intermediate point did that but that seems odd too.

I've restored them there and in `AimdLimitImpl.AimdToken` as well.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
tjquinno: Changed.

Now two lines of code
```java
                if (outcome.disposition() == LimitAlgorithm.Outcome.Disposition.ACCEPTED) {
                    LimitAlgorithm.Outcome.Accepted accepted = (LimitAlgorithm.Outcome.Accepted) outcome;
```
instead of one
```java
                if (outcome instanceof LimitAlgorithm.Outcome.Accepted accepted) {
```
but done in `Http1Connection` and `Http2ServerStream`.

common/concurrency/limits/pom.xml
tjquinno: Ah, OK.

PR #10473, opened 2025-08-01
4.x: Services.set for config will set both config contracts

tomas-langer:
When Services.set() is called with `io.helidon.config.Config`, also set the instances for common config.

This is a temporary solution until we have this available as part of service registry functionality (as we need to avoid reflection to analyze the contract).


barchetta (approved):
Look good. Thanks.

Issue #9702, opened 2025-01-27
Add support for Eureka service discovery


barchetta:
Add support for Eureka service discovery

- [ ] #9655 
- [ ] Add support for Eureka service registration
- [ ] Add support for Eureka service discovery

See also #420

ljnelson:
#10476 introduces the Discovery API. An immediately following PR will introduce the Eureka implementation of the API.

ljnelson:
Reopening since full documentation is next up

ljnelson:
@barchetta I believe this epic is complete.

ljnelson:
Hearing nothing further, closing as complete. #10641 is tangentially related.

Issue #7011, opened 2023-06-15
io.helidon.messaging.connectors.jms.AckMpTest randomly fails


Tomas-Kraus:
I saw this test failure several times, restarting the job usually helps.
```
[INFO] Running io.helidon.messaging.connectors.jms.AckMpTest
2023.06.15 09:57:15 INFO io.helidon.microprofile.cdi.HelidonContainerImpl Thread[#22,helidon-cdi-shutdown-hook,5,main]: Shutdown requested by JVM shutting down
2023.06.15 09:57:15 INFO io.helidon.nima.webserver.ServerListener Thread[#27,server-@default-listener,5,main]: [0x75af79bc] @default socket closed.
2023.06.15 09:57:15 INFO io.helidon.nima.webserver.LoomServer Thread[#22,helidon-cdi-shutdown-hook,5,main]: Níma server stopped all channels.
2023.06.15 09:57:15 INFO io.helidon.microprofile.server.ServerCdiExtension Thread[#22,helidon-cdi-shutdown-hook,5,main]: Server stopped in 8 milliseconds.
2023.06.15 09:57:15 INFO io.helidon.microprofile.cdi.HelidonContainerImpl Thread[#22,helidon-cdi-shutdown-hook,5,main]: Shutdown finished
[INFO] Async process complete, exit value = 0 for: [java, --enable-preview, -jar, /tmp/maven-exec15510170939562921931.jar, -classpath]
Error:  Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 6.693 s <<< FAILURE! - in io.helidon.messaging.connectors.jms.AckMpTest
Error:  io.helidon.messaging.connectors.jms.AckMpTest.resendAckTestPart2(SeContainer)  Time elapsed: 5.185 s  <<< FAILURE!
java.lang.AssertionError: 
Not all 3 items delivered in time.
Expected: iterable containing ["NO_ACK-1", "NO_ACK-2", "NO_ACK-3"]
     but: item 0: was "NO_ACK-2"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at io.helidon.messaging.connectors.mock.MockOutgoing.awaitData(MockOutgoing.java:166)
	at io.helidon.messaging.connectors.mock.MockOutgoing.awaitPayloads(MockOutgoing.java:194)
	at io.helidon.messaging.connectors.jms.AckMpTest.resendAckTestPart2(AckMpTest.java:116)
[...]

PR #10476, opened 2025-08-01
4.x: Introduces Discovery

ljnelson:
# Discovery

## Summary

*Discovery* transforms a supplied *discovery name* into an immutable *discovered set* of immutable resources suitable for it at a given moment of runtime.

In the first iteration of this feature, the resources are metadata-bearing URIs (*discovered URIs*), and can be used for any purpose. They are most often used to help connect to microservices on the network that may come and go over time.

Discovery is fundamentally a specification, with a corresponding vendor-independent Java API, that can be implemented by any number of implementations.

## Pull Request

This pull request introduces the Discovery API and its specification. Subsequent pull requests will:
* Introduce a Discovery implementation backed by [Netflix's Eureka server](https://github.com/Netflix/eureka/tree/v2.0.5)
    * This pull request will be submitted immediately after the current pull request is merged, _i. e._ the work for this pull request is complete
* Introduce documentation.
* Introduce integration with [Helidon's WebClient facility](https://helidon.io/docs/latest/se/webclient).

Other related pull requests may follow.

## Sample Javadoc

<details><summary><h3><code>io.helidon.discovery.Discovery</code></h3></summary>
﻿<a href="https://github.com/user-attachments/assets/68b5a4c4-844e-4d9b-b29b-23df532897da">io.helidon.discovery.Discovery</a></details>

<details><summary><h3><code>io.helidon.discovery.Discovery#uris</code></h3></summary>
<a href="https://github.com/user-attachments/assets/f5fb1e53-cdd8-4220-802d-180fdb396fa7">io.helidon.discovery.Discovery#uris</a></details>

<details><summary><h3><code>io.helidon.discovery.DiscoveredUri</code></h3></summary>
<a href="https://github.com/user-attachments/assets/e9b8ff44-a1a9-4860-8b43-50f0d212920a">io.helidon.discovery.DiscoveredUri</a></details>

## Requirements and Terminology

These requirements informed the design of the Discovery specification and its corresponding API. They, the full specification, and the terminology used to describe them, are present, normatively, in the API documentation. They are reproduced here in the form originally agreed to.

1. *Discovery* is fundamentally a Java API that provides a set of metadata-bearing *discovered URIs* when given a *discovery name* and a *default URI*. This provided set is the *discovered set*.
2. A discovered URI represented by an element in the discovered set may be used for any purpose by the caller.
3. A discovered URI must be immutable. Metadata borne by a discovered URI must be immutable and must be `String`-typed. A discovered URI may bear no metadata.
4. Two discovered URIs are considered *equal* if and only if the URIs they represent are equal. Notably, metadata is not included in equality calculations.
5. A discovered set must be immutable.
6. A discovered set must include a discovered URI that represents the default URI supplied to the operation that produced it. (A discovered set therefore will never be empty.)
7. Any two discovery operations may result in disjoint discovered sets.
8. Any two discovery operations may result in identical discovered sets.
9. Discovery must permit multiple implementations (*e. g.* Eureka, Consul, DNS, Kubernetes, etc.).
10. The presence of two or more *enabled* Discovery implementations at runtime in an application is prohibited.
11. The order of a discovered set may or may not be significant, depending on the Discovery implementation. A discovered set's order is *stable*, however.
12. No guarantees exist, or can be inferred, about the health or availability of an application described by a discovered URI.
13. Discovery names should be composed only from legal DNS hostname characters, since DNS-based Discovery implementations should be possible. In practice, this means alphabetic characters, digits, "`.`", and "`-`".
14. There must be no restriction on the schemes or general format of URIs used or supplied by Discovery (`http`, `https`, `mailto`, `jdbc`, `amqp`, etc.).
15. Any malfunctioning of a Discovery implementation must not prevent an application from continuing.
16. Discovery must be *opt-in*, not *opt-out*. Specifically:
    1. Discovery may be used *explicitly*, *i.e.* by a developer issuing deliberate method calls to the Discovery API and then making use of the discovered set of discovered URIs she receives.
    2. Discovery may be used *implicitly*, *i.e.* by future facilities in Helidon that can apply Discovery automatically, but there must still be some developer-visible manifestation of Discovery in such a case so that its effects can be anticipated.
17. Discovery must not be used to find, or connect to, an implementation of Discovery.

## Use Cases

These use cases informed the design of Discovery. They are deliberately in a form that is technology agnostic.

### Basic, Explicit

1. A developer or subsystem asks Helidon for a Discovery implementation without necessarily knowing in advance what implementation will be supplied.
2. Helidon provides such a Discovery.
3. The developer or subsystem asks Discovery for a set (the *discovered set*) of metadata-bearing URIs (*discovered URIs*) corresponding to a *discovery name*.
    1. She also supplies a *default URI* to use in case there is an error, or in case there would be no suitable discovered set otherwise.
4. The Discovery implementation provides such a discovered set.
    1. At a minimum, the discovered set contains a representation of the default URI.
5. The developer or subsystem selects an element from the discovered set for use according to the application's requirements and heuristics.
6. The developer or subsystem uses the URI the selected discovered URI represents in subsequent development.

### Testing, Explicit

1. A developer wishes to test a part of her application that happens to use the Discovery API (a *Discovery-using component*).
    1. This Discovery API usage may not be, itself, under test.
2. She ensures that a Discovery implementation is present in her project's runtime (and hence test) dependencies.
3. She configures this Discovery implementation to be disabled at test time, possibly because the test environment does not supply required external resources.
    1. Note that this implies any Discovery implementation may be so configured.
4. She executes the test.
5. The Discovery-using component is, by definition, prepared for a Discovery implementation to be disabled.
    1. For example, depending on the final shape of the Discovery API, it might supply a default value at discovery time, or it might test the discovered set to see if it is empty before using a default value instead.
6. The Discovery-using component is prepared for this case and uses a default value instead (see 5.1 above).
7. The test runs successfully, with "real" Discovery never actually being performed.


discovery/README.html
romain-grecourt, line 1: Is this file supposed to be included ?

ljnelson: No; removed

ljnelson:
This PR is related to #9702 but will not close it when merged.

ljnelson:
Interesting; it appears that while Github will let you include images in the PR description, if you attempt to click on it, the URI that Github generates to the image results in a 404. Not sure how to work around that. Obviously the "cure" is to build the javadocs yourself.

ljnelson:
I've fixed the javadoc image problem by duplicating the link target manually.

ljnelson:
Rerunning failed jobs due to #7011

discovery/discovery/etc/spotbugs/exclude.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+
+<FindBugsFilter
```
tjquinno: This file seems to have an empty top-level element. Do we need it?

discovery/discovery/src/main/java/io/helidon/discovery/Discovery.java
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * <h2>Discovered Sets</h2>
+ *
+ * <p>Invocations of the {@link #uris(String, URI)} method return immutable {@link SequencedSet}s of {@link
+ * DiscoveredUri}s. These are <dfn>discovered sets</dfn>. Discovered sets have at least the following
+ * characteristics:</p>
+ *
+ * <ol>
+ *
+ * <li>They are <dfn>immutable</dfn>, following the <a
```
tjquinno, line 73: Hmm. If returned discovered sets are _unmodifiable_ then the _caller_ cannot change them but it's conceivable that the backing data structure _could_ change so they are not necessarily _immutable_. Maybe clarify what the contract requires or expects of implementations on this score?

tjquinno (commented):
A couple relatively minor notes. Quite thorough!

discovery/discovery/src/main/java/io/helidon/discovery/Discovery.java
ljnelson: My intent was to indicate via the word "immutable" that the Java Collection Framework "unmodifiable" semantics apply, and also, since elsewhere I document that no reference is retained to the collection, that JCF "immutable" semantics apply as well. As [the Java Collections Framework states](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Collection.html#unmodifiable), "However, if an unmodifiable collection contains all immutable elements, it can be considered effectively immutable" and "However, if the backing collection of an unmodifiable view is effectively immutable, or if the only reference to the backing collection is through an unmodifiable view, the view can be considered effectively immutable."

discovery/discovery/etc/spotbugs/exclude.xml
ljnelson: Hmm, probably not, although I'm not sure of the convention. Do we keep an "empty" `exclude.xml` "just in case", or do we introduce it only when necessary?

ljnelson: I went ahead and removed it

ljnelson: Hmm, that seems to result in https://github.com/helidon-io/helidon/actions/runs/16684210383/job/47230134932?pr=10476#step:3:4742. I'll put it back.

ljnelson: Er, no, it was just a reference in the `pom.xml` file. Carry on 🤦🏻

discovery/discovery/src/main/java/io/helidon/discovery/Discovery.java
tjquinno: I did see the note at line 95. I guess I reasoned that "They and their contents"  there could be read to mean that the returned unmodifiable `SequentialSet` itself is not retained, but that might allow the values to still be kept in some other backing structure. It's probably not a big deal.

ljnelson: Not trying to shut you down. If you have better wording in mind I'm willing to apply it!

tjquinno: No; what's there conveys what's needed, I think. I was just desperate to find _something_ to comment about!

ljnelson:
Merged with @barchetta's additional blessing

PR #10483, opened 2025-08-04
io.helidon.messaging.connectors.jms.AckMpTest randomly fails #7011

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/7011

I think with `ActiveMQJMSConstants.INDIVIDUAL_ACKNOWLEDGE` it should work too:
https://activemq.apache.org/components/artemis/documentation/latest/pre-acknowledge.html#individual-acknowledge

The issue seems to be in how complex it works internally, when `"NO_ACK-1"` is processed before `"2"`, it looks that `"NO_ACK-1"` is acked.

I tried to update ActiveMQ version and it didn't help.

With this change test will not fail.


romain-grecourt:
Fixed in https://github.com/helidon-io/helidon/pull/10524

PR #10489, opened 2025-08-05
4.x: Deprecate common config and replace its usages

### LINKED ISSUE #10416: 4.x: Deprecate common config
Helidon Version: 4.x

## Task Description

We have created `io.helidon.common.config:helidon-common-config` to work around cyclic dependencies in original `helidon-inject`, which is now removed from the project.

`helidon-service-registry` does not suffer from this cyclic dependency issue, builders have been updated to not suffer from this issue.

As a result, the module is no longer needed.

My recommendation is to deprecate the module, change all our modules to use `helidon-config` as we used to, and forget about this.

This change is backward incompatible, so we have to do the actual removal (and some of code generation changes) only in 5. This impacts generated builders that are configurable, and a lot of public APIs that were modified to accept common config...

The decision made by the team:

- deprecate common config
- update all Helidon code to switch from common config to Helidon config
- keep the module until 5, so code generated with previous 4.x releases is valid

tomas-langer:
### Description

Resolves #10416 

- deprecate common config
- update codegens and APIs
- update changelog


config/config/src/main/java/io/helidon/config/ConfigKeyImpl.java
romain-grecourt, line 1: Copyright only

CHANGELOG.md
```diff
@@ -13,10 +13,22 @@ For Helidon 1.x releases please see [Helidon 1.x CHANGELOG.md](https://github.co
...
 ## [4.3.0-SNAPSHOT]
 
-This minor release of Helidon contains important bugfixes and enhancements and is recommended for all users of Helidon 4. It is API compatible with Helidon 4.2.X.
+This minor release of Helidon contains important bugfixes and enhancements and is recommended for all users of Helidon 4.
 
 A minimum of Java 21 is required to use Helidon 4.
 
+### BREAKING CHANGES
+
+- As part of deprecation of module `helidon-common-config`, we have changed all public APIs to accept only
```
romain-grecourt: `io.helidon.common.config.Config` is deprecated and all usages within the Helidon components are replaced with `io.helidon.config.Config`.

The builder code generator breaks compatibility for builders that use `@io.helidon.builder.api.Option.Configured`. 

The signatures of the config related methods now use `io.helidon.config.Config` instead of `io.helidon.common.config.Config`:

```java
 // before
BUILDER config(io.helidon.common.config.Config config);
static FooBar create(io.helidon.common.config.Config config);

// after
BUILDER config(io.helidon.config.Config config);
static FooBar create(io.helidon.config.Config config);
```

This means that existing code explicitly using `io.helidon.common.config.Config` will need to upcast to `io.helidon.config.Config` to compile.

```java
io.helidon.common.config.Config config = io.helidon.common.config.Config.create();
FooBarConfig.create((io.helidon.config.Config) config);
```

romain-grecourt (commented):
See comments about the incompatibilities notes in CHANGELOG.md.

Is it possible to overload methods that use `io.helidon.common.config.Config` with a variant that use `io.helidon.config.Config` to preserve compatibility ?

E.g.

```java
static FooBar create(io.helidon.config.Config config) {
    return builder().config(config).build();
}

static FooBar create(io.helidon.common.config.Config config) {
    if (config instanceof Config cfg) {
        return create(cfg);
    } else {
        LOGGER.log(Level.WARNING,
            "io.helidon.common.config.Config is deprecated!");
        return create(Config.toConfig(config)); // wrap
    }
}

Builder config(io.helidon.config.Config config) {
    return this;
}

Builder config(io.helidon.common.config.Config config) {
    if (config instanceof Config cfg) {
        return config(cfg);
    } else {
        LOGGER.log(Level.WARNING,
            "io.helidon.common.config.Config is deprecated!");
        return config(Config.toConfig(config)); // wrap
    }
}
```

tomas-langer:
I have refactored the PR, it now should be fully backward compatible (i.e. you should not be required to change your code when upgrading to a this Helidon version).
See changelog for details.

pom.xml
```diff
@@ -571,6 +571,11 @@
                                 <failOnWarning>true</failOnWarning>
                                 <ignoredDependencies>
                                     <dependency>io.helidon.common.features:helidon-common-features-api:*</dependency>
+                                    <!-- used by generated code, and as such not correctly identified by the plugin -->
+                                    <dependency>io.helidon.config:helidon-config</dependency>
```
romain-grecourt: We didn't need to add these before, why now ?
Looks like we bind this to the `verify` phase, so generated code should be visible to the plugin.

CHANGELOG.md
```diff
@@ -13,10 +13,20 @@ For Helidon 1.x releases please see [Helidon 1.x CHANGELOG.md](https://github.co
...
 ## [4.3.0-SNAPSHOT]
 
-This minor release of Helidon contains important bugfixes and enhancements and is recommended for all users of Helidon 4. It is API compatible with Helidon 4.2.X.
+This minor release of Helidon contains important bugfixes and enhancements and is recommended for all users of Helidon 4.
 
 A minimum of Java 21 is required to use Helidon 4.
 
+### BREAKING CHANGES
+
+- As part of deprecation of module `helidon-common-config`, `helidon-config` MUST be on classpath for any component that uses
```
romain-grecourt: `io.helidon.common.config.Config` is deprecated in favor of `io.helidon.config.Config`.

`io.helidon.config:helidon-config` is now required by generated builders that use @io.helidon.builder.api.Option.Configured

`io.helidon.config:helidon-config` depends on `io.helidon.common.config:helidon-common-config`, thus the incompatibility occurs when depending only `io.helidon.common.config:helidon-common-config`.

CHANGELOG.md
```diff
@@ -13,10 +13,20 @@ For Helidon 1.x releases please see [Helidon 1.x CHANGELOG.md](https://github.co
...
+    configuration. Helidon modules are updated to include this dependency where possible
+- Configured builders generated with Helidon 4.3.0 will not work with older versions of Helidon, as the API they use from
```
romain-grecourt: What is the added API that will cause the breakage ? Can you add a code snippet to show how it will break.

CHANGELOG.md
```diff
@@ -13,10 +13,20 @@ For Helidon 1.x releases please see [Helidon 1.x CHANGELOG.md](https://github.co
...
 A minimum of Java 21 is required to use Helidon 4.
 
+### BREAKING CHANGES
+
+- As part of deprecation of module `helidon-common-config`, `helidon-config` MUST be on classpath for any component that uses
+    configuration. Helidon modules are updated to include this dependency where possible
+- Configured builders generated with Helidon 4.3.0 will not work with older versions of Helidon, as the API they use from 
+    `helidon-config` module does not exist yet
+- Configured builders generated with Helidon 4.2.4+ will work with Helidon 4.3.0
+- Configured builders generated with previous versions of Helidon will not work, and must be re-generated
```
romain-grecourt: Can you illustrate what will break ?

pom.xml
tomas-langer: before it only contained helidon-common-config, which was used in other places, so it never complained. 
I will try to investigate a bit

tomas-langer: The problem - for example in `helidon-http` module:
- the module depends on `helidon-common-configurable` that depends on `helidon-config`
- the module uses `helidon-config` in testing

If I do not declare `helidon-config` at all in this module, I get this error:
```
[ERROR] Used undeclared dependencies found:
[ERROR]    io.helidon.config:helidon-config:jar:4.3.0-SNAPSHOT:compile
[ERROR] Non-test scoped test only dependencies found:
[ERROR]    io.helidon.config:helidon-config:jar:4.3.0-SNAPSHOT:compile
```
If I declare it in `test` scope, the build fails (as there is a module dependency required on `io.helidon.config`, which is not on module path

tomas-langer: And if I declare it in `compile` scope, I get the following error:

```
[ERROR] Non-test scoped test only dependencies found:
[ERROR]    io.helidon.config:helidon-config:jar:4.3.0-SNAPSHOT:compile
```
(helidon-http does not generate code that uses config, so my comment in the POM was not correct)

tomas-langer: I will do that exclusion for the http module, checking if other modules are impacted as well.

tomas-langer:
I have addressed all the comments I hope.

common/config/src/main/java/io/helidon/common/config/ConfigBuilderSupport.java
```diff
@@ -26,7 +26,10 @@
 /**
  * Methods used from generated code in builders when
  * {@link io.helidon.builder.api.Prototype.Configured} is used.
+ *
+ * @deprecated use {@code io.helidon.config.ConfigBuilderSupport} instead
  */
+@Deprecated(forRemoval = true, since = "4.3.0")
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 public final class ConfigBuilderSupport {
```
romain-grecourt, line 34: Should we make `io.helidon.common.config.ConfigBuilderSupport` extend `io.helidon.config.ConfigBuilderSupport` and have the generated configured builders still use `io.helidon.common.config.ConfigBuilderSupport` to preserve "forward compatibility" ?

Would that only leave the new dependency as the only breaking change ?

CHANGELOG.md
```diff
@@ -13,10 +13,24 @@ For Helidon 1.x releases please see [Helidon 1.x CHANGELOG.md](https://github.co
...
 ## [4.3.0-SNAPSHOT]
 
-This minor release of Helidon contains important bugfixes and enhancements and is recommended for all users of Helidon 4. It is API compatible with Helidon 4.2.X.
+This minor release of Helidon contains important bugfixes and enhancements and is recommended for all users of Helidon 4.
 
 A minimum of Java 21 is required to use Helidon 4.
 
+### BREAKING CHANGES
+
+- `io.helidon.common.config.Config` is deprecated in favor of `io.helidon.config.Config`.
```
romain-grecourt: The deprecation of `io.helidon.common.config.Config` in favor of `io.helidon.config.Config` is not actually a breaking change, it could be documented in a different section (DEPRECATIONS?).

We need to make this easy to understand for end-users, right now it's confusing.
Maybe @barchetta can help.

common/config/src/main/java/io/helidon/common/config/ConfigBuilderSupport.java
tomas-langer: common config cannot depend on config

tomas-langer:
Examples fail because the `Configuration` class was moved from common config to config, and it is used in the Declarative example

tomas-langer:
Rebased on latest `main` and fixed Eureka provider dependencies (as it now needs full config)

romain-grecourt (approved):

config/config/src/main/java/io/helidon/config/ConfigKeyImpl.java
romain-grecourt: Still there, but oh well `¯\_(ツ)_/`

PR #10491, opened 2025-08-05
4.x: Declarative configuration expression

tomas-langer:
Allow expression that use configuration in annotation values for Helidon Declarative.

Changed the following:
- scheduling (both annotations)
- rest client URI

The expression can use `${config.key:default-value}` pattern, if default value is not provided, it must exist in configuration under the provided key.

Updated codegen, scheduling annotations, and rest client annotation (it now has a mandatory value to contain the URI).

There is no other way to use configuration for declarative now.

As a follow up, other codegens for declarative should be refactored to use the same expression resolution (i.e. server endpoint, fault tolerance).

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/DelcarativeConfigSupport.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/DelcarativeConfigSupport.java
@@ -23,8 +23,8 @@ import static io.helidon.declarative.codegen.DeclarativeTypes.CONFIG_BUILDER_SUP
 /**
  * Support to generate consistent configuration for declarative types.
  */
-public final class Configurability {
-    private Configurability() {
+public final class DelcarativeConfigSupport {
+    private DelcarativeConfigSupport() {
     }
 
     /**
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/restclient/RestClientExtension.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/restclient/RestClientExtension.java
@@ -43,7 +43,7 @@ import io.helidon.common.types.TypeInfo;
 import io.helidon.common.types.TypeName;
 import io.helidon.common.types.TypeNames;
 import io.helidon.common.types.TypedElementInfo;
-import io.helidon.declarative.codegen.Configurability;
+import io.helidon.declarative.codegen.DelcarativeConfigSupport;
 import io.helidon.declarative.codegen.http.HttpFields;
 import io.helidon.declarative.codegen.http.RestExtensionBase;
 import io.helidon.declarative.codegen.model.http.ClientEndpoint;
@@ -555,10 +555,10 @@ class RestClientExtension extends RestExtensionBase implements RegistryCodegenEx
     }
 
     private void constructorUriHandling(Constructor.Builder ctr, ClientEndpoint endpoint) {
-        Configurability.assignResolveExpression(ctr,
-                                                "config",
-                                                "uri",
-                                                endpoint.uri());
+        DelcarativeConfigSupport.assignResolveExpression(ctr,
+                                                         "config",
+                                                         "uri",
+                                                         endpoint.uri());
 
         String path = endpoint.path().orElse("/");
         if (path.startsWith("/")) {
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/scheduling/SchedulingExtension.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/scheduling/SchedulingExtension.java
@@ -37,8 +37,8 @@ import io.helidon.common.types.TypeInfo;
 import io.helidon.common.types.TypeName;
 import io.helidon.common.types.TypeNames;
 import io.helidon.common.types.TypedElementInfo;
-import io.helidon.declarative.codegen.Configurability;
 import io.helidon.declarative.codegen.DeclarativeTypes;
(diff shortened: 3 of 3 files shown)


tomas-langer:
Example fix linked

common/config/src/main/java/module-info.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
```
spericas: No need for this?

microprofile/scheduling/src/test/resources/test.properties
```diff
@@ -17,11 +17,9 @@
 test-cron-expr=0/3 * * * * ? *
 
 # Overriding annotation values
-io.helidon.microprofile.scheduling.SchedulingTest.overriddenValuesCron.schedule.cron=0/2 * * * * ? *
-io.helidon.microprofile.scheduling.SchedulingTest.overriddenValuesCron.schedule.concurrent=false
-
-io.helidon.microprofile.scheduling.SchedulingTest.overriddenValuesFixed.schedule.delay-by=PT1S
-io.helidon.microprofile.scheduling.SchedulingTest.overriddenValuesFixed.schedule.interval=PT2S
+override.cron=0/2 * * * * ? *
```
romain-grecourt: `overrides`, plural, seems better.

declarative/tests/scheduling/src/main/resources/application.yaml
```diff
@@ -15,9 +15,7 @@
 #
 
 override:
```
romain-grecourt: `overrides`, plural, seems better.

common/config/src/main/java/io/helidon/common/config/ConfigBuilderSupport.java
```diff
@@ -243,6 +252,44 @@ public static <T extends NamedService> List<T> discoverServices(Config config,
                                             existingValue);
     }
 
+    /**
```
romain-grecourt: Can this go in `io.helidon.config.ConfigHelper` since we are trying to move away from `io.helidon.common.config`.

microprofile/scheduling/src/test/resources/test.properties
tomas-langer: it is a test, and the key is arbitrary. Do you want me to change this?

common/config/src/main/java/io/helidon/common/config/ConfigBuilderSupport.java
tomas-langer: I will change that once we go fully for config. Right now `io.helidon.config` may not be a dependency of the generated code, so we cannot really do it.
This PR is to decide if this concept is OK, I will rebase it on the config change once it is done.

tomas-langer:
Updated this PR and the examples PR. This is now the intended target solution for 4.3.0

config/config/src/main/java/io/helidon/config/ConfigBuilderSupport.java
```diff
@@ -31,6 +33,13 @@
  */
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 public final class ConfigBuilderSupport {
+    // matches string between ${ } with a negative lookbehind if there is not backslash
```
romain-grecourt: `s/not/no/g`

declarative/codegen-model/src/main/java/io/helidon/declarative/codegen/model/http/ClientEndpointBlueprint.java
```diff
@@ -30,7 +30,7 @@ interface ClientEndpointBlueprint extends RestEndpointBlueprint {
      *
      * @return URI if configured
      */
-    Optional<String> uri();
+    String uri();
```
romain-grecourt, line 33: That interface has not been released yet right ?

declarative/codegen/src/main/java/io/helidon/declarative/codegen/Configurability.java
```diff
@@ -0,0 +1,70 @@
...
+package io.helidon.declarative.codegen;
+
+import io.helidon.codegen.classmodel.ContentBuilder;
+
+import static io.helidon.declarative.codegen.DeclarativeTypes.CONFIG_BUILDER_SUPPORT;
+
+/**
+ * Support to generate consistent configuration for declarative types.
+ */
+public final class Configurability {
```
romain-grecourt: This is `public` because of the io.helidon.declarative.codegen.* java packages right ?
Do we have a way to identify such classes as "project private" ?

declarative/codegen-model/src/main/java/io/helidon/declarative/codegen/model/http/ClientEndpointBlueprint.java
tomas-langer: Correct, declarative is introduced in 4.3.0

declarative/codegen/src/main/java/io/helidon/declarative/codegen/Configurability.java
tomas-langer: I consider any public class to be part of the public API of Helidon

romain-grecourt: How about `CodegenSupport`,  or `ConfigExpressions` ?

Issue #10486, opened 2025-08-05
4.1.4 "Illegal character in query" error during websocket connection when urls have encoded query parameters (with spaces) in them


amalbose:
## Environment Details
* Helidon Version: 4.1.4
* Helidon MP
* JDK version: 21
* OS: mac

----------

## Problem Description
When websocket urls have encoded parameters in them (with spaces), it fails to connect throwing `Illegal character in query` error
eg: ws://localhost:8081/websocket/?u=First%20Last

This issue is always reproducible with encoded parameters in the url

From debugging, I could see that UriQueryImpl#value() is used to generate the the query string in io.helidon.microprofile.tyrus.TyrusUpgrader#protocolHandshake
As seen in the screenshot this generates the query params by joining the decoded query params. So in our case “[First%20Last](ws://localhost:8081/websocket/?u=First%20Last)” will be decoded to “[First Last](ws://localhost:8081/websocket/?u=First%20Last)” and then joined to generate the invalid url “[u=First Last](ws://localhost:8081/websocket/?u=First%20Last)”
This url is used to generate the TyrusSession (second screenshot) (in org.glassfish.tyrus.core.TyrusEndpointWrapper#onConnect) where getURI will throw “Reason Phrase cannot exceed 123 UTF-8 encoded bytes“ Exception

[...]

PR #10492, opened 2025-08-05
Use rawQuery() instead of query() in Tyrus integration code

spericas:
### Description

Use rawQuery() instead of query() in the integration code to avoid decoding escaped chars such as %20 before passing the URI to Tyrus. See issue #10486.

### Documentation

None


tjquinno (approved):
LGTM

Issue #10471, opened 2025-08-01
4.x Intermittent pipeline test failure in `SseClientTest.testSseJson1`


tjquinno:
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Intermittent pipeline test failure:

https://github.com/helidon-io/helidon/actions/runs/16668098611/job/47178543193?pr=10470#step:3:6321

```list
Error:  Errors: 
Error:    SseClientTest.testSseJson1:146 » InsufficientDataAvailable
[INFO] 
Error:  Tests run: 24, Failures: 0, Errors: 1, Skipped: 0
```

jbescos:
I cannot reproduce it locally after 189 tries.

tjquinno:
That might be, but it is still happening in the pipelines. From a few hours ago: https://github.com/helidon-io/helidon/actions/runs/16739471713/job/47385035288

tjquinno:
Happening very frequently.

jbescos:
I guess the server socket wrote in the stream and closed it, but maybe the stream was buffered and it was not flushed. In this way the client could get that error. I am going to investigate it.

jbescos:
This is the stack when data is written by the SseSink:

```
System Thread [[0x6cff6fb4 0x0481208b] WebServer socket] (Suspended)	
	GrowingBufferData.writeTo(OutputStream) line: 66	
	PlainSocket.write(BufferData) line: 136	
	SocketWriterDirect(SocketWriter).writeNow(BufferData) line: 81	
	DataWriterSseSink.emit(SseEvent) line: 124	
	DataWriterSseSink.emit(SseEvent) line: 49	
	SseBaseTest.sseJson1(ServerRequest, ServerResponse) line: 69	
	0x00007f56941b8478.handle(ServerRequest, ServerResponse) line: not available	
	HttpRoutingImpl$RoutingExecutor.doRoute(ConnectionContext, RoutingRequest, RoutingResponse) line: 174	
	HttpRoutingImpl$RoutingExecutor.call() line: 132	
	HttpRoutingImpl$RoutingExecutor.call() line: 110	
	ErrorHandlers.runWithErrorHandling(ConnectionContext, RoutingRequest, RoutingResponse, Callable<Void>) line: 76	
	Filters.filter(ConnectionContext, RoutingRequest, RoutingResponse, Callable<Void>) line: 83	
[...]

PR #10493, opened 2025-08-06
4.x Intermittent pipeline test failure in SseClientTest.testSseJson1 #10471

jbescos:
### Description
https://github.com/helidon-io/helidon/issues/10471

See the discussion of the issue.


webserver/tests/sse/src/test/java/io/helidon/webserver/tests/sse/SseBaseTest.java
```diff
@@ -45,18 +45,28 @@ protected WebServer webServer() {
...
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+    }
+
     static void sseString1(ServerRequest req, ServerResponse res) {
         try (SseSink sseSink = res.sink(SseSink.TYPE)) {
             sseSink.emit(SseEvent.create("hello"))
                     .emit(SseEvent.create("world"));
+            // Fix for github actions: https://github.com/helidon-io/helidon/issues/10471
```
tomas-langer: Sleep is in general not a good solution to intermittent failures, as the amount of time required may change depending on the runner.
If we want to fix tests properly, we should have a mechanism to correctly chain the events (i.e. using some Java concurrent constructs, such as a CountDownLatch).

jbescos: Right, I will do that.

What about the flush in PlainSocket#close?. I honestly think we should have more issues if that part is missing. Find more context here: https://github.com/helidon-io/helidon/issues/10471#issuecomment-3157721221

jbescos:
Even with CountDownLatch set to 30 seconds, sometimes it fails. The only difference is that previously it was failing `testSseJson1` and now it is failing `testSseJson2`.

I don't know what to do, because this is happening in github actions only.

jbescos:
Closing, this is not a good fix.

PR #10494, opened 2025-08-06
4.x: Extract Config override functionality into a separate module

### LINKED ISSUE #10415: 4.x: Deprecate config "overrides" feature
Helidon Version: 4.x

## Task Description

Overrides in Helidon config are not used by our team, and we never received any feedback for this feature.
Either we have created it without a single bug, or it is not used. 

As it is really unexpected to have bugless features, I would assume this feature is not used. 
It is quite a complicated implementation, and it would be better if we could remove it from our code base.

This task has several subtasks

- [ ] deprecate all APIs for config overrides (i.e. `io.helidon.config.spi.OverrideSource`), mark it for removal, add link to this issue with a message "if you are using this feature, please let us now in the referenced issue, as otherwise it will be removed"
- [ ] add warning log statement (invoked exactly once per runtime) when the feature is used with the same message and link
- [ ] backport this change to Helidon 3 and 2, to make sure we prevent future upgrade failures

Depending on how long we manage to have this deprecation in place, we can remove it from Helidon either in 5 or in 6

tomas-langer:
and deprecate the existing APIs and SPIs for Config overrides.

Updated documentation

Resolves #10415 

After reviewing changes for examples, I realized that a change in the override source (when using the filter) does not trigger a change in the config itself (as there is no such support in configuration).

I will try to figure out if this is a problem for the change


tomas-langer:
Closing this PR for now, as I need to decide how to proceed.

PR #10495, opened 2025-08-06
4.x: Remove unused class InterceptorBase and his subclass

Captain1653:
### Description
I have found unused class `package-private` class `InterceptorBase` and his unused subclass. These classes can be safely removed without breaking compatibility.
Tests are green.


Captain1653:
@tjquinno Can it be reviewed? Please :)

Issue #10472, opened 2025-08-01
Infinite loop during webservice request in io.helidon.http.http2.Http2Headers.DynamicTable#add


stoerr:
## Environment Details
* Helidon Version: 4.3.2
* Helidon MP
* JDK version: OpenJDK Runtime Environment Temurin-21.0.3+9
* OS: MacOS 15.5

----------

## Problem Description

The DynamicTable.add() method in Http2Headers can enter an infinite loop because it can happen that currentTableSize > maxTableSize but the headers list is empty, causing the evict() method to do nothing and the while loop to continue indefinitely. That caused our server to run at 100% CPU after a load test, even with no outside traffic anymore.

It happened during a load test, but is difficult to reproduce. However, a possible cause of the problem can be seen in the code, though it's hard to test whether that was the actual cause of the problem in our case. So I suggest to fix this bug; we'll see whether the problem reoccurs.

The effect of the bug is that after a load test the server continued to run at 100% after a load test. With a debugger I verified that currentTableSize > maxTableSize while the headers table was empty.

Compare [Http2Headers.java line 1201](https://github.com/helidon-io/helidon/blob/release-4.2.4/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java#L1201) : the evict method was called again and again and did nothing because the headers table was empty, but the loop condition was met since currentTableSize > maxTableSize. 

Code inspection revealed one likely cause of that at [HttpsHeaders.java line 1180](https://github.com/helidon-io/helidon/blob/release-4.2.4/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java#L1180) : here the headers table is cleared, while the currentTableSize (which is supposed to track the size of the headers) is not reset:

```java
        void maxTableSize(long number) {
            if (number > protocolMaxTableSize) {
                throw new Http2Exception(Http2ErrorCode.COMPRESSION, "Attempt to set larger size than protocol max");
            }
            this.maxTableSize = number;
            if (maxTableSize == 0) {
                headers.clear();
            }
            while (maxTableSize < currentTableSize) {
                evict();
            }
        }
```

Here are the other relevant code fragments. The infinite loop was in the while method:

[method add](https://github.com/helidon-io/helidon/blob/release-4.2.4/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java#L1192C1-L1211C10)

```java
        int add(HeaderName headerName, String headerValue) {
            String name = headerName.lowerCase();
[...]

PR #10502, opened 2025-08-07
4.x Infinite loop during webservice request in io.helidon.http.http2.Http2Headers.DynamicTable#add #10472

jbescos:
### Description

This PR fixes an infinite loop, although I am not sure there are more ways to reproduce it. There could be a concurrency issue in webclient, but I cannot reproduce it.

Maybe fixes https://github.com/helidon-io/helidon/issues/10472

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -1175,6 +1175,7 @@ public class Http2Headers {
             }
             this.maxTableSize = number;
             if (maxTableSize == 0) {
+                currentTableSize = 0;
                 headers.clear();
             }
             while (maxTableSize < currentTableSize) {
--- a/http/http2/src/test/java/io/helidon/http/http2/DynamicTableTest.java
+++ b/http/http2/src/test/java/io/helidon/http/http2/DynamicTableTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
```diff
@@ -1129,7 +1131,8 @@ void write(BufferData buffer) {
      */
     public static class DynamicTable {
         private final List<DynamicHeader> headers = new ArrayList<>();
-        private volatile long protocolMaxTableSize;
+        private final Lock lock = new ReentrantLock();
```
tomas-langer: Why did you introduce a lock to this class?

jbescos: It is explained in the issue https://github.com/helidon-io/helidon/issues/10472 . Basically it needs to be thread safe.

There are some locks to avoid DynamicTable is accessed concurrently here:
https://github.com/helidon-io/helidon/issues/10472#issuecomment-3154127754

But there is other usage of this class coming from this stack trace that is not guarded by previous locks:
https://github.com/helidon-io/helidon/issues/10472#issuecomment-3168387480

tomas-langer: the dynamic table should not be thread safe
- server writer has its own instance, and correctly locks access
- server reader is a single thread (the connection thread) and should not need to lock access
- I am not sure about the client reading of header, that seems to be a bug right now, that the usage of the table is not locked, but that should be implemented by locking in the client stream, not by making the table thread safe

danielkec: I do agree with Tomas, this is server's hot path. We need to move the locking in the client's path only

jbescos: OK, I need to be able to reproduce this first. I will contact with the issue filer.

jbescos:
Setting it as a draft for now.

jbescos:
I am struggling to reproduce it. This PR fixes something, as we can see in the `#testIssue10472`, but I am not sure this will fix the original issue.

tomas-langer:
This PR will not fix the original issue.
Even though the fix is needed for other reasons, this does not resolve the client issue where we need to guard access to the header table (as it is not thread safe, and client accesses it from multiple threads).
This must be fixed in the client implementation.

PR #10505, opened 2025-08-07
Suppress False Positive for Angus Mail

barchetta:
### Description

Suppress False Positive for Angus Mail. Helidon has a dependency on Angus Activation, but not Angus Mail.


ljnelson:
False *positive*?

romain-grecourt (approved):

Issue #10422, opened 2025-07-24
Caching of the access token for oidc client credential flow


botsteve:
Helidon Version: Mp 4.2.4

## Enhancement Description

In the current implementation, a new access token is fetched for every outgoing request to the protected API. This enhancement introduces a caching mechanism to reuse the access token across multiple requests, provided it is still valid .If the cached token is expired or missing, a new token will be fetched and cached.

Benefits:

- Reduces the number of calls to the OAuth token endpoint.
- Improves performance by avoiding unnecessary latency.

PR #10509, opened 2025-08-08
Cache access token for client credentials flow

botsteve:
### Documentation

- Uses a single CachedToken instance to cache the client credentials access token per provider/config, not per outbound target or tenant.
- Access to the cache is protected by a ReentrantLock  for thread safety.
- If the OIDC server does not provide expires_in as it is a RECOMMENDED but not REQUIRED field in the response https://datatracker.ietf.org/doc/html/rfc6749, the caching is skipped in order to avoid scenarios where you get stuck with an expired cached access token.
- The CachedToken class uses a 30-second buffer to refresh the token before it actually expires.

Closes the issue https://github.com/helidon-io/helidon/issues/10422

### DIFF
--- a/security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcProvider.java
+++ b/security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcProvider.java
@@ -18,11 +18,14 @@ package io.helidon.security.providers.oidc;
 
 import java.lang.System.Logger.Level;
 import java.lang.annotation.Annotation;
+import java.time.Duration;
+import java.time.Instant;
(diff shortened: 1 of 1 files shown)


botsteve:
@romain-grecourt or @Verdent if you can review sometime. Thanks!

botsteve:
@Verdent, if you have some time to review this, please

botsteve:
Hei @Verdent, Happy New Year! Can you spare some time to merge this if it's ok?

botsteve:
@barchetta , @Verdent , can this MR be merged, would appreciate if somebody can take a look.

PR #10510, opened 2025-08-08
Revert "Add tracing support to concurrenty limits"

tjquinno:
Reverts helidon-io/helidon#10189

The span added is not correctly parented, and there is some concern about whether Helidon should add a span for waiting time due to concurrency limits.


ljnelson (approved):
This looks like a reasonably straightforward removal to me.

tjquinno:
I should have marked this as draft earlier. I think we're hoping to supersede the earlier implementation (which this would revert) with a better one #10464 . Stay tuned.

tjquinno:
Closing without merging. #10464 supersedes this.

PR #10511, opened 2025-08-08
4.x: Allow full customization of singular setter for collections and maps.

### LINKED ISSUE #10302: Allow full control over map names in blueprints
## Enhancement Description

 When using a `Map` with the singular option in a blueprint, a method name can be specified but this only controls the method's name suffix. For example, using `@Option.Singular("foo")` generates a method named `putFoo` in the builder. It would be nice to have full control over this method name and be able to drop the `put` prefix. IMO, fluent APIs read better without prefixes such as `put` in cases like this.

tomas-langer:
### Description
Resolves #10302 

Simply set `io.helidon.builder.api.Option.Singular#withPrefix` to false, and the name  in value of the same annotation will be used as the full name of the "add" or "put" method.

This is to enable nicer names when adding a value to a map, that is more aligned with its behavior (as `put` may not always be the best solution here).

### DIFF
--- a/builder/tests/builder/src/test/java/io/helidon/builder/test/ComplexCaseTest.java
+++ b/builder/tests/builder/src/test/java/io/helidon/builder/test/ComplexCaseTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


builder/tests/builder/src/test/java/io/helidon/builder/test/ComplexCaseTest.java
```diff
@@ -37,6 +37,20 @@
...
+    void testCustomAddName() {
+        var configBean = MyConfigBean.builder()
+                .setName("cfb")
+                .build();
+
+        var complexCase = ComplexCase.builder()
+                .setName("name")
+                .setClassType(Object.class)
+                .allowConfigBean(configBean)                .build();
+
```
spericas, line 51: Missing new line here?

spericas: Other than that LGTM

tomas-langer: fixed

spericas (approved):

PR #10522, opened 2025-08-12
4.x: Fix wrong method generated for factory methods in builder

### LINKED ISSUE #10516: 4.x: Factory method from Config that returns a list breaks generated code
- create a blueprint with a configured list
- create a factory method that returns a list

The generated code will have the following method:

```java
        public BUILDER processors(Consumer<SpanProcessorConfig.Builder> consumer) {
            Objects.requireNonNull(consumer);
            var builder = SpanProcessorConfig.builder();
            consumer.accept(builder);
            this.processors(builder.build());
            return self();
        }
```

where `this.processors` is a list, so it does not make sense at all.

tomas-langer:
Config factory methods for types that are also configured blueprints no longer generate an invalid method if the return type is a list.

Added tests

Resolves #10516


tjquinno (approved):
LGTM. I'll mark this approved although I'm not an expert in this area!

PR #10531, opened 2025-08-13
Add debug logging when instantiating a service (resolves #10530)

### LINKED ISSUE #10530: Make service instantiation comprehensible during runtime
Helidon Version: 4.2.x

## Enhancement Description

To better understand when a service gets instantiated by a `ServiceRegistryManager` and when `postConstruct` and `preDestroy` is called, we suggest to add debug log messages to each `ServiceDescriptor` to be generated.

This helps in debugging scenarios where one wants to understand the load order of each service.

arouel:
### Description

With this change, we provide debug log messages when a service gets instantiated or when the methods `postConstruct` or `preDestroy` are called.

This helps a developer to understand the lifecycle of an individual service but also the lifecycle of all services together.

This Pull Request resolves https://github.com/helidon-io/helidon/issues/10530.

### Documentation

The resulting log messages look like the following in case debug logging is enabled.

```
10:44:45.915 [main] DEBUG test.TestService - instantiate (weight = 900.0, run level = 10.0)
10:44:45.915 [main] DEBUG test.TestService - postConstruct
10:44:51.464 [main] DEBUG test.TestService - preDestroy
```

### DIFF
--- a/service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
+++ b/service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
@@ -200,6 +200,7 @@ public class ServiceDescriptorCodegen {
                 .sortStaticFields(false);
 
         singletonInstanceField(classModel, service);
+        singletonLoggerField(classModel, service);
 
         Map<String, GenericTypeDeclaration> genericTypes = genericTypes(classModel, params, methods);
 
@@ -1062,6 +1063,17 @@ public class ServiceDescriptorCodegen {
                 .defaultValueContent("new " + service.descriptorType().className() + "<>()"));
     }
 
+    private void singletonLoggerField(ClassModel.Builder classModel, DescribedService service) {
+        // singleton instance of the descriptor
+        classModel.addField(instance -> instance
+                .accessModifier(AccessModifier.PRIVATE)
+                .isStatic(true)
+                .isFinal(true)
+                .type(TypeName.create("java.lang.System.Logger"))
+                .name("LOGGER")
+                .defaultValueContent("System.getLogger(\"" + service.serviceDescriptor().typeName().genericTypeName() + "\")"));
+    }
+
     private void injectionPointFields(ClassModel.Builder classModel,
                                       TypeInfo service,
                                       Map<String, GenericTypeDeclaration> genericTypes,
@@ -1644,6 +1656,8 @@ public class ServiceDescriptorCodegen {
                 .map(ParamDefinition::ipParamName)
                 .collect(Collectors.joining(", "));
 
+        method.addContentLine("LOGGER.log(System.Logger.Level.DEBUG, \"instantiate (weight = \" + weight() + \", run level = \" + runLevel().orElse(null) + \")\");");
+
         if (interceptedMethods) {
             // return new MyImpl__Intercepted(interceptMeta, this, ANNOTATIONS, casted params
             method.addContent("return new ")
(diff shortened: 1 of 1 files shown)


service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -1062,6 +1063,17 @@ private void singletonInstanceField(ClassModel.Builder classModel, DescribedServ
                 .defaultValueContent("new " + service.descriptorType().className() + "<>()"));
     }
 
+    private void singletonLoggerField(ClassModel.Builder classModel, DescribedService service) {
+        // singleton instance of the descriptor
+        classModel.addField(instance -> instance
+                .accessModifier(AccessModifier.PRIVATE)
+                .isStatic(true)
+                .isFinal(true)
+                .type(TypeName.create("java.lang.System.Logger"))
```
tomas-langer: Please use `System.Logger.class`, as that is part of JDK and there is no need to abstract it to a TypeName. TypeName would require a new constant + test to verify it is valid, so this is an easier path.

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -1644,6 +1656,8 @@ private void createInstantiateBody(TypeName serviceType,
                 .map(ParamDefinition::ipParamName)
                 .collect(Collectors.joining(", "));
 
+        method.addContentLine("LOGGER.log(System.Logger.Level.DEBUG, \"instantiate (weight = \" + weight() + \", run level = \" + runLevel().orElse(null) + \")\");");
```
tomas-langer: This should be done slightly differently to avoid possible issues with name conflicts and imports.
Also when using any parameters, the log statement should be guarded by `LOGGER.isLoggable()`

```java
method.addContent("LOGGER.log(")
                .addContent(System.Logger.Level.class)
                .addContentLine(".DEBUG, \"Instantiate (weight = \" + weight() + \", run level....")
```

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -1922,6 +1936,7 @@ private void postConstructMethod(ClassModel.Builder classModel, DescribedService
                     .addAnnotation(Annotations.OVERRIDE)
                     .addParameter(instance -> instance.type(typeName)
                             .name("instance"))
+                    .addContentLine("LOGGER.log(System.Logger.Level.DEBUG, \"postConstruct\");")
```
tomas-langer: The `Level` should be added as a class, so we can sort out imports (example above)

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -1936,6 +1951,7 @@ private void preDestroyMethod(ClassModel.Builder classModel, DescribedService se
                     .addAnnotation(Annotations.OVERRIDE)
                     .addParameter(instance -> instance.type(typeName)
                             .name("instance"))
+                    .addContentLine("LOGGER.log(System.Logger.Level.DEBUG, \"preDestroy\");")
```
tomas-langer: dtto

tomas-langer (commented):
This seems like a nice addition for troubleshooting, just a few comments regarding style.

If you do not feel like fixing the style comments, let me know, I can pick the PR and continue on it.

Thanks for the contribution!

tomas-langer (dismissed):

PR #10532, opened 2025-08-13
4.x: Updated TLS implementation to handle endpoint identification algorithm for server

### LINKED ISSUE #10429: mTLS shouldn't require setting endpoint-identification-algorithm in the config
Helidon Version: 4

## Enhancement Description
In order for mTLS to work on the server, the TLS config the server uses must have `endpoint-identification-algorithm` explicitly set to `NONE`. This is rather unintuitive. There has been discussion around the underlying concept on the OpenJDK bug tracker in the past: https://bugs.openjdk.org/browse/JDK-8229972?focusedId=14659802&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14659802

#8012 discusses this configuration settings as well, identifying that it is counter-intuitive. The [response given](https://github.com/helidon-io/helidon/issues/8012#issuecomment-1814995452) by the OP explains the problem.

Essentially, **the request is for `endpoint-identification-algorithm` to automatically be set to `NONE` in server configs**. A strong case can be made that it should be set to `NONE` _even if_ a user sets it to `HTTPS` explicitly.

tomas-langer:
The algorithm is now ignored for server sockets, as is mentioned in:  https://bugs.openjdk.org/browse/JDK-8229972?focusedId=14659802&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14659802
 
Added test, and verified on our example.

### Description
Resolves #10429

### DIFF
--- a/common/tls/src/main/java/io/helidon/common/tls/Tls.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/Tls.java
@@ -180,7 +180,8 @@ public class Tls implements RuntimeType.Api<TlsConfig> {
 
             // create a copy of SSLParameters, as otherwise we may use wrong endpoint identification algorithm for server
             SSLParameters parameters = copySslParameters();
-            parameters.setApplicationProtocols(parameters.getApplicationProtocols());
+            parameters.setApplicationProtocols(this.sslParameters.getApplicationProtocols());
+            parameters.setEndpointIdentificationAlgorithm("");
 
             socket.setSSLParameters(parameters);
             return socket;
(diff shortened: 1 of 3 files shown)


common/tls/src/main/java/io/helidon/common/tls/Tls.java
```diff
@@ -273,6 +263,29 @@ Optional<X509TrustManager> trustManager() {
         return tlsManager.trustManager();
     }
 
+    private SSLParameters copySslParameters() {
+        // this copy does not set application protocols (each client request may have different
```
danielkec:
```suggestion
        // this copy does not set application protocols (each client request may have different ones)
```

danielkec (approved):

PR #10548, opened 2025-08-15
4.x Allow configuring OpenTelemetry overall and tracing using Helidon builders and config

### LINKED ISSUE #9595: Add support for OpenTelemetry configuration in Helidon SE

tjquinno:
### Description
Resolves #9595 

## Release Note
____
Helidon SE now has additional support for OpenTelemetry.

* Configuring OpenTelemetry and OpenTelemetry tracing using Helidon SE config
  
  By adding a dependency on the new component `io.helidon.telemetry:helidon-telemetry-opentelemetry-config` developers can configure the behavior of OpenTelemetry programmatically using Helidon SE builders declaratively using Helidon SE config. 
  
  Some of the available settings:
  ```yaml
  telemetry:
    service: "test-otel"
    global: false
    signals:
      tracing:
        attributes:
          attr1: 12
          attr2: true
        sampler:
          type: "always_off"
        exporters:
          - type: otlp
            compression: gzip
        processors:
          - type: batch
            max-queue-size: 20
  ```
* Using OpenTelemetry semantic conventions for auto-created spans
  
   For some time, Helidon SE has automatically created tracing spans for each incoming HTTP request if an app depends on `io.helidon.webserver.observe:helidon-webserver-observe-tracing`. This code prepared the span name and span tags to convey information about the request with each span.
  
  By adding a dependency on the new component `io.helidon.webserver.observe:helidon-webserver-observe-telemetry-tracing` to an app, developers can instead have those span names and tags conform to the OpenTelemetry tracing semantic conventions.

Note that these enhancements are _incubating_ changes. We intend to keep them but we might withdraw them or change the public APIs and behavior in dot releases. (See https://helidon.io/docs/v4/apidocs/io.helidon.common.features.api/io/helidon/common/features/api/Incubating.html)
____

### New component `helidon-telemetry-opentelemetry-config`

This new component contains two main categories of code:
* Configurable builders for top-level `telemetry` settings and for tracing level `tracing` settings. 
  
  See the example config below; note the use of the `signals` key to group the various Otel signals; tracing is the only one this PR supports but others signals such as logging or metrics can be added smoothly in future PRs and fit nicely in this structure.
  
  There are numerous supporting types (enums for types of items, etc.) and factory methods to convert config to Otel types.
* Service registry interfaces and classes to register the Otel `OpenTelemetry` object derived from configuration. 
  
  The factory class (`HelidonOpenTelemetryServiceFactory implements Supplier<OpenTelemetry>`) injects `Config` and uses the builders, etc. to construct the `OpenTelemetry` and, if the config has `global` as true, asks Otel to make the new instance its (Otel's) global instance.

### Modified `helidon-webserver-observe-tracing`
The code that prepares a span for incoming tracing requests now retrieves a service from the registry of type `TracingSemanticConventions` and invokes methods on that to augment the span builder and span it creates for each HTTP request.

This component now also implements that interface using the logic it has always used as a `@Service.Singleton` with a  lower-than-default weight so another implementation takes precedence if one is visible.

### New component `helidon-webserver-observe-telemetry-tracing`
This component implements the OpenTelemetry semantic conventions for automatically-created incoming spans for HTTP requests.

### Fuller example

The following config (this example works with the code) is fairly straightforward. One interest aspect: Application code using the Otel API does not set exporters programmatically; instead it sets exporters on a processor and then adds the processor to the tracer provider builder. A configured provider can specify a list of exporter names, and those names must be defined in the config. A provider that specifies no exporter names gets all declared exporters.

About attributes: The Otel API allows apps to assign string, `Number`, and boolean values. We should be able to expose only a single `attributes` setting. Our code would try to interpret a value as a boolean, then as a numeric value, and if all those fail then treat it as a string and then invoke the corresponding method to set the attribute.

Example config:
```yaml
  telemetry:
    service: "test-otel"
    global: false
    signals:
      tracing:
        attributes:
          attr1: "some text"
          attr2: true
          attr3: 24.5
        sampler:
          type: "always_off"
        exporters:
          - type: otlp
             protocol: http/proto
             name: my-oltp
           - type: zipkin
         processors:
           - max-queue-size: 21.     # Uses all defined exporters because no names are specified.
             type: batch
           - type: simple
             exporters: ["my-oltp"]. # Uses only the named exporter(s).
```

### Documentation
TBD soon in a separate PR. See issue #10556


telemetry/opentelemetry-config/pom.xml
```diff
@@ -0,0 +1,252 @@
...
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-configurable</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.service</groupId>
+            <artifactId>helidon-service-registry</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common.features</groupId>
```
tomas-langer: This should be an optional dependency (features api)

telemetry/opentelemetry-config/pom.xml
```diff
@@ -0,0 +1,252 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-semconv</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-extension-trace-propagators</artifactId>
+        </dependency>
+        <dependency>
```
tomas-langer, line 83: config-metadata should only be used when not using blueprints, probably not needed here

telemetry/opentelemetry-config/pom.xml
```diff
@@ -0,0 +1,252 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
```
tomas-langer, line 157: There is a features-api dependency, yet no codegen configured for it.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/ExporterType.java
```diff
@@ -0,0 +1,86 @@
...
+        this.defaultProtocol = defaultProtocol;
+        this.defaultPort = defaultPort;
+        this.defaultPath = defaultPath;
+    }
+
+    ExporterType() {
+        this(null, 0, null);
+    }
+
+    String defaultProtocol() {
```
tomas-langer: This can return null - it would be better to use an Optional, even though it is not a public API, as it is easy to fall into NPE

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetryImpl.java
```diff
@@ -0,0 +1,70 @@
...
+import io.helidon.builder.api.RuntimeType;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.OpenTelemetry;
+
+/**
+ * Helidon management of OpenTelemetry.
+ */
+@RuntimeType.PrototypedBy(OpenTelemetryConfig.class)
+public class HelidonOpenTelemetryImpl implements HelidonOpenTelemetry, RuntimeType.Api<OpenTelemetryConfig> {
```
tomas-langer: Impl classes must not be public.
You can add the `PrototypedBy` on the `HelidonOpenTelemetry` interface.

tomas-langer: Or just remove `public` from the class declaration.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetryServiceFactory.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.function.Supplier;
+
+import io.helidon.config.Config;
+import io.helidon.service.registry.Service;
+
+import io.opentelemetry.api.OpenTelemetry;
+
+@Service.Singleton
+@Service.RunLevel(Service.RunLevel.STARTUP)
```
tomas-langer, line 29: Do we really need a run level for this? 
Run level should be used if the component needs to open sockets, start processing repeating tasks etc. If the `OpenTelementry` instance may be needed by other services, it does not need run level, and it would be created when needed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * OpenTelemetry {@link io.opentelemetry.context.propagation.TextMapPropagator} instances added explicitly by the app.
+     * <p>
+     * Default: {@value ContextPropagationType#DEFAULT_NAMES}.
+     *
+     * @return explicitly-added
+     */
+    @Option.Configured
+    @Option.Singular
+    @Option.DefaultCode("ContextPropagationType.DEFAULT_PROPAGATORS")
```
tomas-langer: The class `ContextPropagationType` may not be imported, in this case it may be better to use the fully qualified class name.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigSupport.java
```diff
@@ -0,0 +1,113 @@
...
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.context.propagation.TextMapPropagator;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+
+class OpenTelemetryConfigSupport {
+
```
tomas-langer, line 37: the class should be final, missing private constructor

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigSupport.java
```diff
@@ -0,0 +1,113 @@
...
+                return;
+            }
+
+            target.openTelemetry(target.enabled()
+                                         ? openTelemetry(target)
+                                         : OpenTelemetry.noop());
+        }
+
+    }
+    static class CustomMethods {
```
tomas-langer, line 85: Same as above - final, private constructor, this should fail our checkstyle plugin

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryTracingConfigBlueprint.java
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Constructed span processors.
+     *
+     * @return span processors
+     */
+    @Option.Singular
+    List<SpanProcessor> processors();
+
+    /**
+     * Span exporters.
```
tomas-langer, line 74: Maybe a bit broader explanation of what the key is in the map?

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SamplerConfigSupport.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.telemetry.otelconfig;
+
+import io.helidon.builder.api.Prototype;
+import io.helidon.common.config.Config;
+
+class SamplerConfigSupport {
```
tomas-langer: Considering the number of blueprints, we could live with a single top level class - i.e. OtelConfigSupport, and then have all the inner classes in this one, just name them accordingly.
Not a huge problem, as this is all package local.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SamplerConfigBlueprint.java
```diff
@@ -0,0 +1,49 @@
...
+    @Option.Default(SamplerType.DEFAULT_NAME)
+    SamplerType type();
+
+    /**
+     * Sampler parameter.
+     *
+     * @return sampler parameter
+     */
+    @Option.Configured
+    Optional<Number> param();
```
tomas-langer: This will most likely not work - either change to `Optional<Double>`, or you should add a factory method.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SamplerType.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Parent-based trace ID ration-based sampler.
+     */
+    PARENT_BASED_TRACE_ID_RATIO("parentbased_traceidratio");
+
+    static final String DEFAULT_NAME = "PARENT_BASED_ALWAYS_ON";
+    static final SamplerType DEFAULT = PARENT_BASED_ALWAYS_ON;
+
+    private final String name;
```
tomas-langer: `name` is already used by `enum` type itself. 
I have used `text` in other enums that need a textual representation that is not the same as `#name()`. 
This should be aligned in all the enums in this PR, but at least do not use `name`

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanExporterConfigBlueprint.java
```diff
@@ -0,0 +1,36 @@
...
+interface SpanExporterConfigBlueprint {
+
+    /**
+     * Span exporter type.
+     *
+     * @return exporter type
+     */
+    @Option.Configured
+    @Option.Default("DEFAULT")
+    ExporterType type();
```
tomas-langer, line 33: The `ExporterType` enum is not declared as `public`, yet it will be part of public API.
I think this is valid for all the enums in this PR - please make sure they are `public` when used in  any `Blueprint`.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanProcessorType.java
```diff
@@ -0,0 +1,60 @@
...
+            if (spanProcessorType.processorType.equals(value) || spanProcessorType.name().equals(value)) {
+                return spanProcessorType;
+            }
+        }
+        throw new IllegalArgumentException("Unknown span processor type: " + value + "; expected one of "
+                                                   + Arrays.toString(SpanProcessorType.values()));
+
+    }
+
+    static SpanProcessorType from(Config config) {
```
tomas-langer: This should not be needed - Config provides an enum mapper that is quite clever, so as long as:
- the values use `_` same as enum values
- the values are case insensitively equal

The mapping will work out of the box, and no custom factory method is required

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/ZipkinExporterConfigSupport.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    static class CustomMethods {
+
+        private CustomMethods() {
+        }
+
+        @Prototype.FactoryMethod
+        static SpanBytesEncoder createSpanBytesEncoder(Config config) {
```
tomas-langer: This factory method is not needed (one of the examples where this will be handled by config, as I mentioned above)

telemetry/opentelemetry-config/src/main/java/module-info.java
```diff
@@ -0,0 +1,52 @@
...
+ * limitations under the License.
+ */
+
+import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.HelidonFlavor;
+
+/**
+ * OpenTelemetry implementation for telemetry.
+ */
+@Features.Name("Telemetry")
```
tomas-langer, line 23: This is `OpenTelemetry`, not `Telemetry`, right?

telemetry/opentelemetry-config/src/main/java/module-info.java
```diff
@@ -0,0 +1,52 @@
...
+    requires io.opentelemetry.extension.trace.propagation;
+    requires io.opentelemetry.sdk;
+    requires io.opentelemetry.sdk.common;
+    requires io.opentelemetry.sdk.trace;
+
+    requires zipkin2;
+    requires zipkin2.reporter;
+
+    requires static io.helidon.common.features.api;
+    requires io.opentelemetry.semconv;
```
tomas-langer: Missing an `exports` statement.

telemetry/opentelemetry-config/src/main/java/module-info.java
```diff
@@ -0,0 +1,52 @@
...
+@Features.Path({"Telemetry"})
+@Features.Incubating
+module io.helidon.telemetry.otelconfig {
+
+    requires io.helidon.builder.api;
+    requires io.helidon.common.configurable;
+    requires io.helidon.config;
+    requires io.helidon.service.registry;
+
+    requires io.opentelemetry.api;
```
tomas-langer, line 37: There are a few dependencies declared as `optional`, yet here they are as `requires` - all optional dependencies must be declared as `requires static`, as otherwise they may not be available at runtime, and this module would not load.

telemetry/opentelemetry-config/src/test/java/io/helidon/telemetry/otelconfig/TestServiceBootstrap.java
```diff
@@ -0,0 +1,74 @@
...
+                                  protocol: http/proto
+                                  name: my-oltp
+                                - type: zipkin
+                              processors:
+                                - max-queue-size: 21
+                                  type: batch
+                        """,
+                MediaTypes.APPLICATION_YAML));
+
+        Services.set(Config.class, config);
```
tomas-langer, line 60: If you want to use `Services.set` in testing, you should annotate your class as `@Testing.Test`, as otherwise you will not be able to call it more than once (the annotation ensures that you run within a different global service instance).

telemetry/opentelemetry-config/pom.xml
tjquinno: Fixed.

telemetry/opentelemetry-config/pom.xml
tjquinno: Fixed

telemetry/opentelemetry-config/pom.xml
tjquinno: I think my one problem was I did not invoke the annotation processor. I've added that.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/ExporterType.java
tjquinno: It turns out the code that uses this enum does not use these defaults, so I am removing them entirely.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetryImpl.java
tjquinno: Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetryServiceFactory.java
tjquinno: My thinking was this...

The factory instantiates `HelidonOpenTelemetry` which, if `global` is `true` will use the OTel API to assign the configured `OpenTelemetry` instance as the OTel global instance (as well as register it in our registry).

OTel permits setting its global instance only once, similar to our `Services.set` method. User code could attempt to retrieve the OTel global instance using the OTel API directly without or before some other code gets the `OpenTelemetry` instance from our registry (which would trigger the factory). 

In that case, OTel would use its default config (or config set using OTel environment variables and system properties) instead of using the Helidon config.

So having the run level is an attempt to have our factory prepare the global `OpenTelemetry` instance  using the Helidion configuration _before_ user code might use the OTel API directly, thereby ignoring the configuration.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
tjquinno: Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigSupport.java
tjquinno: Fixed

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigSupport.java
tjquinno: Fixed

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryTracingConfigBlueprint.java
tjquinno: Done

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SamplerConfigBlueprint.java
tjquinno: Changed to `Optional<Double>`.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanExporterConfigBlueprint.java
tjquinno: Unfortunately, I think I have to change all the enums to  `public`.

We _should_ be able to keep them package private. They are used only in blueprints that themselves should not be exposed to developers, and I had marked (most) of the related methods also with access "". 

But the generated getter method on the `BuilderBase` for the enum is public even though the generated setter uses the specified access.

See https://github.com/helidon-io/helidon/issues/10552

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanProcessorType.java
tjquinno: Ah, yes. I remember writing a lot of that mapper back in the day!

I've changed many of the enums so the declared enum names match (except for case) the lower-case values to take advantage of this. Doing so eliminated several of the config support classes as a bonus.

Fixed

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/ZipkinExporterConfigSupport.java
tjquinno: Fixed

telemetry/opentelemetry-config/src/main/java/module-info.java
tjquinno: Changed to `Telemetry/OpenTelemetry/Config` after off-line chat.

Fixed

telemetry/opentelemetry-config/src/main/java/module-info.java
tjquinno: Fixed.

telemetry/opentelemetry-config/src/main/java/module-info.java
tjquinno: Fixed.

telemetry/opentelemetry-config/src/test/java/io/helidon/telemetry/otelconfig/TestServiceBootstrap.java
tjquinno: Fixed

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SamplerConfigSupport.java
tjquinno: Removing the custom enum config-to-enum conversion methods and their invocations from support classes (relying on the automatic support for enums in config) also let me remove some of the support classes.

For the moment I'll leave the others as top-level classes--five remain.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SamplerType.java
tjquinno: Changing to use the bulit-in enum config mapping removed the need for most per-enum conversions.

Any remaining enums that need to match use clear field names.

Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanExporterConfigBlueprint.java
tjquinno: All changed to public.

tjquinno:
Recent additional enhancement to the PR... 

By adding a component (`helidon-webserver-observe-telemetry-tracing`) to an app a user can have Helidon follow the OpenTelemetry semantic conventions on spans (name and tags attached). Without adding that component users continue to see span names and tags as they have before.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetryServiceFactory.java
tomas-langer: RunLevel is used only in Declarative anyway, so i am afraid this will not achieve what you want to achieve.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryTracingConfigSupport.java
```diff
@@ -0,0 +1,134 @@
...
+
+    }
+
+    static class CustomMethods {
+
+        private CustomMethods() {
+        }
+
+        @Prototype.FactoryMethod
+        static Object createAttributes(Config config) {
```
tomas-langer: I prefer using typed approach. 
In other builders, we ended up with multiple configurable options - i.e. 
`stringAttributes`, `booleanAttributes`, `doubleAttributes` etc.
This approach is very error prone, as you may have a valid String value `true`, which is not supposed to be a boolean, and coercing it to a boolean may fail further down.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OtlpExporterProtocolType.java
```diff
@@ -0,0 +1,70 @@
...
+                                                   + Arrays.toString(OtlpExporterProtocolType.values()));
+    }
+
+    /**
+     * Maps a config node's string value to the corresponding type.
+     *
+     * @param config config node
+     * @return matching type
+     */
+    static OtlpExporterProtocolType from(Config config) {
```
tomas-langer, line 67: This method should not be needed or used, config will do it correctly.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanExporterConfigBlueprint.java
tomas-langer: If we do not expect the user to use the enumeration in code (and they are only configuration related), then maybe mark them as deprecated until we resolve the issue?

tomas-langer: On the other hand - the fact that the setters are package private only does not change the fact that the blueprint method return type will be public anyway, so what the builder base does is irrelevant - if the prototype is public, so must be the return types.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanProcessorConfigBlueprint.java
```diff
@@ -0,0 +1,52 @@
...
+     * @return span exporter name
+     */
+    @Option.Configured
+    List<String> exporters();
+
+    /**
+     * Span processor type.
+     * @return span processor type
+     */
+    @Option.Configured
```
tomas-langer, line 47: I am not sure I understand why the setter is package private only. Do we expect the user to never use this prototype directly?

telemetry/opentelemetry-config/src/main/java/module-info.java
```diff
@@ -0,0 +1,59 @@
...
+import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.HelidonFlavor;
+
+/**
+ * OpenTelemetry implementation for telemetry.
+ */
+@Features.Name("Telemetry")
+@Features.Description("Support for Telemetry")
+@Features.Flavor({HelidonFlavor.SE, HelidonFlavor.MP})
+@Features.Path({"Telemetry/OpenTelemetry/Config"})
```
tomas-langer: the path must be an array, each element is one path element, i.e "Telemetry", "OpenTelemetry", "Config"

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanExporterConfigBlueprint.java
tomas-langer: You can also use:
```
@Prototype.Blueprint(isPublic = false)
```

To generate a package private prototype - then you do not need to care about public return types, as the user has no way of obtaining an instance of it.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryTracingConfigSupport.java
tjquinno: **About attributes:**
That's precisely why the first step in `createAttributes` is to check for the escape `\"...\"` (for example `\"true\"` or any other type) to force its interpretation as a string. Tested in `TestAttributeValueConversions`.

It would be nice if users could employ the standard `"true"` syntax for JSON or `!!str true` for YAML. But AFAIK our config does not preserve the inferred data type from the JSON or YAML parser so that information is no longer available when our code consumes config. That would be a nice enhancement, at least as a hint that consuming code could take into account.

I strongly suspect that in by far most cases the natural inferred typing is what the user wants. And in the rarer cases where that is _not_ what the user wants, such as where `true` needs to be treated as a string or `12` needs to be a double, the `\"...\"` notation is non-standard, yes, but it works for strings and the standard Java notation of `12D` works. 

This algorithm allows the user to use an easy and natural approach in the (by far) usual cases but provides escapes when the user needs them.

Forcing _all_ users to deal with separate `attributes` (only for strings), `boolean-attributes`, `double-attributes`, etc. which is unneeded in most cases is not necessary and is user-unfriendly.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OtlpExporterProtocolType.java
tjquinno: This is required.

The OTel value for one of the choices is `http/proto` (slash, not underscore).

For consistency with the values that OTel config uses we want to support those values, including this oddball one.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/SpanProcessorConfigBlueprint.java
tjquinno: For uniformity and consistency I'll change that to public.

lMy thinking was that users should not need to invoke _that method_.

`SpanProcessorConfigBlueprint` is used for the simple processor (it has no other settings) and is a super-blueprint for `BatchSpanProcessorBlueprint`. 

The containing blueprint, `OpenTelemetryTracingConfigBlueprint`, exposes `List<SpanProcessor>` which refers to the OTel type. A user would typically hand-build a span processor using the OTel API and then add it to the tracing builder rather than programmatically using our config API. 

But I suppose we don't want to prevent users from using the config builder approach.

tjquinno: Fixed

telemetry/opentelemetry-config/src/main/java/module-info.java
tjquinno: Oof. Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetryServiceFactory.java
tjquinno: True, that does not work automatically. That said, the public Javadoc for `RunLevel.STARTUP` includes code that SE developers can use to bootstrap using run levels. Developers would need to do this themselves to achieve the goal I have in mind, but if we include the `@RunLevel` then at least developers _can_ do that.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryTracingConfigSupport.java
tjquinno: After off-line discussion, relenting and adopting separate typed sections for each data type accepted.

```yaml
telemetry:
    service: "test-otel"
    global: false
    signals:
        tracing:
            attributes:
                strings:
                    attr1: 12
                    attr5: "any old thing"
                    attr7: something
                longs:
                    attr2: 12
                doubles:
                    attr3: 24.5
                    attr6: 12
                booleans:
                    attr4: true
```

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryTracingConfigBlueprint.java
```diff
@@ -0,0 +1,132 @@
...
+    @Option.Configured("attributes.strings")
+    @Option.Singular
+    Map<String, String> stringAttributes();
+
+    /**
+     * Boolean attributes.
+     *
+     * @return boolean attributes
+     */
+    @Option.Configured("attributes.booleans")
```
tomas-langer, line 102: There is also an option to have these with package private access  (i.e. used for configuration) and then have a `Map<String, Object>` as a not-configured option that will eventually host all of them. And then if a user uses the builder directly, they could just call `putAttribute(String, Object)` in code. Not a blocking issue.

PR #10554, opened 2025-08-15
4.x: Introduces a Eureka-backed Discovery implementation

ljnelson:
# Eureka Discovery

## Summary

Introduces a [Eureka version 2](https://github.com/Netflix/eureka/tree/v2.0.5)-backed implementation of [Discovery](https://github.com/helidon-io/helidon/pull/10476) ("Eureka Discovery"), together with unit and integration tests.

## Pull Request

This pull request introduces a [Eureka version 2](https://github.com/Netflix/eureka/tree/v2.0.5)-backed implementation of [Discovery](https://github.com/helidon-io/helidon/pull/10476) ("Eureka Discovery"), along with unit tests and an integration test that exercises an actual Eureka server in conjunction with Eureka Discovery.

Subsequent pull requests will:

* Introduce comprehensive documentation.
* Introduce integration with [Helidon's WebClient facility](https://helidon.io/docs/latest/se/webclient).

Other related pull requests may follow.

## Sample Javadoc

<details><summary><h3><code>io.helidon.discovery.providers.eureka.EurekaDiscovery</code></h3></summary><a href="https://github.com/user-attachments/assets/73aeabab-7f79-4572-81e9-584bdc861152">io.helidon.discovery.providers.eureka.EurekaDiscovery</a></details>

<details><summary><h3><code>io.helidon.discovery.providers.eureka.EurekaDiscoveryConfig</code></h3></summary><a href="">io.helidon.discovery.providers.eureka.EurekaDiscoveryConfig</a></details>

<details><summary><h3><code>io.helidon.discovery.providers.eureka.CacheConfig</code></h3></summary><a href="">io.helidon.discovery.providers.eureka.CacheConfig</a></details>

Note as well that javadoc is available at the `private` level.

## Sample Usage

Via [Helidon Inject](https://helidon.io/docs/latest/se/injection) (configuration is applied automatically, if it is found):
```
import java.net.URI;
import java.util.Objects;
import java.util.SequencedSet;

import io.helidon.discovery.DiscoveredUri;
import io.helidon.discovery.Discovery;
import io.helidon.service.registry.Service.Inject;
import io.helidon.service.registry.Service.Singleton;

@Singleton // for example
public class MyClass {
    private final Discovery d;

    @Inject
    public MyClass(Discovery d) {
        this.d = Objects.requireNonNull(d, "d");
    }

    public void sample() {
        SequencedSet<DiscoveredUri> uris = this.d.uris("EXAMPLE", URI.create("http://example.com/"));
    }
}
```
Via the [Helidon Builder API](https://helidon.io/docs/latest/se/builder):
```
Discovery d = EurekaDiscovery.builder()
    .client(http1ClientBuiltElsewhere) // the client must have its baseUri property set to be useful
    .build();
```

## Sample Configuration

All configuration is optional, except for the option describing where the Eureka server is.

In YAML format, often in an `application.yaml` classpath resource:
```
discovery:
  eureka:
    cache:
      enabled: true # true by default
      compute-changes: true # true by default; compute cache state or replace in full
      defer-sync: false # false by default; start cache synchronization immediately
      fetch-thread-name: "Eureka registry fetch thread" # this is the default value
      sync-interval: "PT30S" # how often to synchronize cache; 30 seconds by default
    client: # Http1ClientConfig controlling how Eureka is contacted/called
      base-uri: "http://localhost:8761/eureka" # required; where the Eureka server lives
    enabled: true # true by default; set to false if you want to disable Eureka Discovery;
                  # see Discovery Use Cases for more details
    prefer-ip-address: false # false by default; if true, try to use IPs instead of hostnames for URI host components
```
## Requirements and Terminology

These requirements (together with those of any [Discovery](https://github.com/helidon-io/helidon/pull/10476) implementation, and others not fully specified) informed the design of Eureka Discovery specification and its corresponding API. They, the full specification, and the terminology used to describe them, are present, normatively, in the API documentation, a sample of which is included above. They are reproduced here in the form originally agreed to.

1. Eureka Discovery provides the ability to discover URI resources from a known [Eureka version 2](https://github.com/Netflix/eureka/tree/v2.0.5) server.
2. Eureka Discovery is not, and will not evolve to be, a full reimplementation of [the native Eureka client](https://javadoc.io/doc/com.netflix.eureka/eureka-client/latest/index.html).
3. Eureka Discovery does only enough to enable efficient discovery operations and [honor the requirements of all Discovery implementations](https://github.com/helidon-io/helidon/pull/10476).
    1. Most notably, Eureka Discovery does not have any peer-to-peer semantics.
4. Any caching used by Eureka Discovery must be optional.
5. Eureka Discovery must be safe for concurrent use by multiple threads.

## Use Cases and Scenarios

These use cases informed the design and implementation of Eureka Discovery. They are deliberately in a form that is technology agnostic. Please also see the [Discovery use cases](https://github.com/helidon-io/helidon/pull/10476).

### Disabling caching

1. A developer configures her usage of Discovery to be provided by Eureka Discovery.
2. She configures Eureka discovery to not use its own caching, but instead to contact the Eureka server for every request.
3. Her usage of the (definitionally implementation agnostic) Discovery instance she acquires does not change.
4. She observes from logs that the Eureka Discovery contacts the server with every discovery operation she performs.
    1. She notices that when the server is down, or another connectivity problem occurs, her application continues to function, using default values supplied to Discovery, in accordance with the [requirements](https://github.com/helidon-io/helidon/pull/10476).

### Enabling caching, adjusting intervals

1. A developer configures her usage of Discovery to be provided by Eureka Discovery.
2. She configures Eureka discovery to use a local cache of discovered information. (This is, in fact, its default.)
3. She adjusts the refresh interval of the cache to be only a few seconds. (Its default emulates what Eureka users are accustomed to, _e. g._ in the vicinity of 30 seconds or so.)
4. She observes from logs that the Eureka Discovery contacts the server regularly every few seconds to see if anything has changed on the server side.

### DIFF
--- a/discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
+++ b/discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
@@ -15,7 +15,6 @@
  */
 package io.helidon.discovery.providers.eureka;
 
-import java.time.Duration;
 import java.util.Optional;
 
 import io.helidon.builder.api.Option;
@@ -42,13 +41,15 @@ import io.helidon.webclient.http1.Http1Client;
 interface EurekaDiscoveryConfigBlueprint extends Prototype.Factory<EurekaDiscovery> {
 
     /**
-     * Whether to use the discovery cache; {@code true} by default.
+     * The {@link CacheConfig} to use controlling how a local cache of Eureka server information is used.
      *
-     * @return {@code true} if the discovery cache should be used
+     * @return a {@link CacheConfig}
+     *
+     * @see CacheConfig
      */
     @Option.Configured("cache")
-    @Option.DefaultBoolean(true)
-    boolean cache();
+    @Option.DefaultCode("CacheConfig.create()")
+    CacheConfig cache();
 
     /**
      * The {@link Http1Client} to use to communicate with the Eureka server.
@@ -86,48 +87,5 @@ interface EurekaDiscoveryConfigBlueprint extends Prototype.Factory<EurekaDiscove
      */
     @Option.Configured("prefer-ip-address")
     boolean preferIpAddress();
-
-    /**
-     * Whether only changes to service information in the Eureka server should be fetched; {@code true} by default.
-     *
-     * @return whether only changes to service information in the Eureka server should be fetched
-     */
-    @Option.Configured("registry-fetch-changes")
-    @Option.DefaultBoolean(true)
-    boolean registryFetchChanges();
-
-    /**
-     * The time between retrievals of service information from the Eureka server; 30 seconds by default.
-     *
-     * @return the time between retrievals of service information from the Eureka server
-     *
-     * @see Duration#parse(CharSequence)
-     */
-    @Option.Configured("registry-fetch-interval")
-    @Option.Default("PT30S")
-    Duration registryFetchInterval();
-
-    /**
-     * The name of the {@link Thread} used to retrieve service information from the Eureka server; "Eureka registry
-     * fetch thread" by default.
-     *
-     * @return the name of the {@link Thread} used to retrieve service information from the Eureka server
-     *
-     * @see Thread.Builder#name(String)
-     */
-    @Option.Configured("registry-fetch-thread-name")
-    @Option.Default("Eureka registry fetch thread")
-    String registryFetchThreadName();
-
-    /**
-     * Whether to eagerly start the {@link Thread} that retrieves service information from the Eureka server; {@code
-     * true} by default.
-     *
-     * @return {@code true} if the thread should be started as soon as possible; {@code false} if it should be started
-     * as late as possible
-     */
-    @Option.Configured("registry-fetch-thread-start-eagerly")
-    @Option.DefaultBoolean(true)
-    boolean registryFetchThreadStartEagerly();
 
 }
--- a/discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/CacheConfigBlueprint.java
+++ b/discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/CacheConfigBlueprint.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.helidon.discovery.providers.eureka;
+
+import java.time.Duration;
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+import io.helidon.builder.api.Prototype.Blueprint;
+
+/**
+ * Prototypical state for the portion of Eureka Discovery configuration related to a local cache of Eureka server
+ * information.
+ */
+@Blueprint
+@Prototype.Configured
+interface CacheConfigBlueprint {
+
+    /**
+     * Whether the state of the cache should be computed from changes reported by Eureka, or replaced in full; {@code true} by default.
+     *
+     * @return whether the state of the cache should be computed
+     */
+    @Option.Configured("compute-changes")
+    @Option.DefaultBoolean(true)
+    boolean computeChanges();
+
+    /**
+     * Whether to defer immediate cache synchronization; {@code false} by default.
+     *
+     * @return {@code true} if cache synchronization should be deferred until it is actually needed
+     */
+    @Option.Configured("defer-sync")
+    @Option.DefaultBoolean(false)
+    boolean deferSync();
+
+    /**
+     * Whether a local cache of Eureka information is used or not; {@code true} by default.
+     *
+     * @return {@code true} if the cache should be used
+     */
+    @Option.Configured("enabled")
+    @Option.DefaultBoolean(true)
+    boolean enabled();
+
+    /**
+     * The name of the {@link Thread} used to retrieve service information from the Eureka server; "Eureka registry
+     * fetch thread" by default.
+     *
+     * @return the name of the {@link Thread} used to retrieve service information from the Eureka server
+     *
+     * @see Thread.Builder#name(String)
+     */
+    @Option.Configured("fetch-thread-name")
+    @Option.Default("Eureka registry fetch thread")
+    String fetchThreadName();
+
+    /**
+     * The time between retrievals of service information from the Eureka server; 30 seconds by default.
+     *
+     * @return the time between retrievals of service information from the Eureka server
+     *
+     * @see Duration#parse(CharSequence)
+     */
+    @Option.Configured("sync-interval")
+    @Option.Default("PT30S")
+    Duration syncInterval();
+
+}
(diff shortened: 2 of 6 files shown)


discovery/providers/eureka/pom.xml
```diff
@@ -0,0 +1,180 @@
...
+                            <artifactId>helidon-config-metadata-codegen</artifactId>
+                            <version>${helidon.version}</version>
+                        </path>
+                        <path>
+                            <groupId>io.helidon.service</groupId>
+                            <artifactId>helidon-service-codegen</artifactId>
+                            <version>${helidon.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
```
tjquinno, line 168: Helidon components' `pom.xml` files tend to declare dependencies here on the same annotation processors declared as `<path>` to make sure the anno processors themselves are up-to-date before use.

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscovery.java
```diff
@@ -0,0 +1,278 @@
...
+import java.util.function.Consumer;
+
+import io.helidon.builder.api.RuntimeType.Api;
+import io.helidon.builder.api.RuntimeType.PrototypedBy;
+import io.helidon.discovery.DiscoveredUri;
+import io.helidon.discovery.Discovery;
+import io.helidon.discovery.providers.eureka.EurekaDiscoveryConfig.Builder;
+
+/**
+ * A {@link Discovery} implementation that communicates with <a
```
tjquinno, line 29: All critically-important info here.

Presumably the eventual doc page (I know the plan is for a separate PR for that) will include much of the same.

At that time would we trim this Javadoc significantly to avoid overlap and, therefore, the need  to maintain the information in two places? If so, what about  including at least a preliminary version of the `.adoc` file in this PR and moving this information there? 

I know there are more steps required than just writing an `.adoc` file...plugging it into the doc tree, choosing a relevant but distinct glyph, etc. so I get the appeal of deferring. It just would be great to have a human- and search-engine-visible doc page touting this feature when the feature arrives, even if the write-up will be expanded later.

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
```diff
@@ -0,0 +1,133 @@
...
+    boolean preferIpAddress();
+
+    /**
+     * Whether only changes to service information in the Eureka server should be fetched; {@code true} by default.
+     *
+     * @return whether only changes to service information in the Eureka server should be fetched
+     */
+    @Option.Configured("registry-fetch-changes")
+    @Option.DefaultBoolean(true)
+    boolean registryFetchChanges();
```
tjquinno: IIUC from the PR description, these `registry*` settings apply only if `cache` is `true`.

Would it be worth it to restructure the config a bit, something like:
```yaml
eureka:
  registry-cache:
    enabled: t/f
    changes-only:
    interval:
    thread:
      name:
      start-eagerly:
```
Some of my thinking...

The repeated `registry-fetch-` prefix suggests there's a natural grouping.

Maybe it's obvious that it's the registry that's being cached so `registry-cache` would be just `cache`? 

Are there cache things other than fetching going on? If not, do we need `fetch` in the key names?

Maybe the `thread` subheading is overkill, or maybe `interval` would belong under there also?

I'm honestly not sure of all this; just asking the questions.

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
```diff
@@ -0,0 +1,133 @@
...
+     * The name of the {@link Thread} used to retrieve service information from the Eureka server; "Eureka registry
+     * fetch thread" by default.
+     *
+     * @return the name of the {@link Thread} used to retrieve service information from the Eureka server
+     *
+     * @see Thread.Builder#name(String)
+     */
+    @Option.Configured("registry-fetch-thread-name")
+    @Option.Default("Eureka registry fetch thread")
+    String registryFetchThreadName();
```
tjquinno: This is exceedingly flexible. 

IIUC there would be at most one Eureka instance in a JVM, so maybe a constant thread name would be OK? (I know it's defaulted so the user does not have to configure it explicitly.)

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryImplFactory.java
```diff
@@ -0,0 +1,113 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A {@link Supplier} of {@link Optional} {@link EurekaDiscoveryImpl} instances that conforms to the requirements of a
+ * <a href="https://helidon.io/docs/latest/se/injection#_factories">Helidon Inject <dfn>factory</dfn></a>.
+ *
+ * @see EurekaDiscoveryImpl
+ */
+@Singleton // the scope of the *product*
+final class EurekaDiscoveryImplFactory implements Supplier<Optional<EurekaDiscoveryImpl>> {
```
tjquinno, line 38: Could this be declared as `EurekaDiscoveryFactory implements Supplier<EurekaDiscovery>`? I know this is not a public class so it's not externally visible, but just as a general practice?

Users or other code in other components would use something like `Services.get(EurekaDiscover.class)` (not the impl class) so maybe it's aesthetically more pleasing if not functionally different if this class' parameterized type aligns.

Plus, although to me it seems unlikely I suppose there could in some cases be other suppliers of `EurekaDiscovery` with a different implementation with different `@Weight` and having them all implement `Supplier<EurekaDiscovery>` would help highlight that they are "competing" providers.

discovery/providers/eureka/pom.xml
ljnelson: Interesting. I was following https://helidon.io/docs/latest/se/injection#_build_time which does not do this. Do you have a sense of which is better?

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
ljnelson: My thinking was people are likely to set only one option at a time and it seemed cruel for programmatic usage to force them to invoke builders two or three levels deep to set a single `boolean`. There may be a middle ground here though.

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryImplFactory.java
ljnelson: The name could change, but it seems important for intra-package classes to be able to inject `EurekaDiscoveryImpl` as well as `EurekaDiscovery` and `Discovery`. (The "normal" JSR-330/CDI approach is to have your producers/factories/whatever declare their product type to be the implementation. Barring some wild voodoo, the potential injectable types are then those of the transitive type closure of the implementation type.)

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscovery.java
ljnelson: I have many thoughts. The short version is: javadoc is not just documentation, but the authoritative specification and usage reference. Anything else is non-normative. I'll take the hit on possibly redundant information/documentation.

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
ljnelson: This was more like: every time I use a constant someone inevitably comes out of the woodwork and wants to customize it. I've just gotten in the mental habit of allowing everything to be set. Maybe this is one step too far; not sure.

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
ljnelson: Struck a compromise; vetted it

discovery/providers/eureka/src/main/java/io/helidon/discovery/providers/eureka/EurekaDiscoveryConfigBlueprint.java
ljnelson: On further thought I think I will leave the option to configure this value. I've relocated it in the last round of feedback. Programmatic usage can result in multiple instances of `EurekaDiscovery` so you want to be able to tweak their aspects.

ljnelson:
Native image failed (I've made no changes in that area, nor is this PR included in that test). I rebased nonetheless and we'll see what happens next.

PR #10562, opened 2025-08-18
4.x: Javadoc on blueprint can now contain simple references to other methods.

### LINKED ISSUE #10553: Javadoc references to prototype members become references to blueprint members, which are not visible
Given:
* a `FooConfigBlueprint` that is, by definition, not `public`
* with a method `boolean foo();` and 
* a `FooConfig` that is generated according to its rules

…then if a runtime type (`Foo`) wants to refer in its javadoc to `FooConfig#foo()`, via `{@link FooConfig#foo()}`, this reference will become unlinkable, because javadoc will turn it into a reference to `FooConfigBlueprint#foo()`, which is not visible.

tomas-langer:
Generated prototypes now copy all methods from blueprint, including javadoc, so javadoc references work as expected.

A blueprint may now add default methods as prototype methods, to support backward compatibility of the generated prototype interface.

Resolves #10553

### DIFF
--- a/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
+++ b/builder/api/src/main/java/io/helidon/builder/api/Prototype.java
@@ -417,12 +417,21 @@ public final class Prototype {
     }
 
     /**
-     * Include default methods from the blueprint and its super types.
+     * Include default methods from the blueprint and its super types (interface methods declared as {@code default}).
      * <p>
      * If the value is specified, only methods with names matching the values are included, as long as they qualify as
      * blueprint methods (i.e. they have a non-void return type and zero parameters).
      * <p>
      * This can be used to define backward compatible changes.
+     * <p>
+     * Behavior for handling {@code default} methods on interfaces in blueprints:
+     * <ul>
+     *     <li>If this annotation is NOT present on a blueprint, methods declared as {@code default} are ignored</li>
+     *     <li>If this annotation is present on a blueprint and has empty value, all methods declared as {@code default} are
+     *          included as long as they are property getters (non-void, no parameters, not private)</li>
+     *     <li>If this annotation is present on a blueprint and has non-empty value, only methods with names matching a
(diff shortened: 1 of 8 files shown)


builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -415,5 +415,26 @@ default void decorateAddSet(B builder, Set<T> optionValues) {
          */
         boolean value() default true;
     }
+
+    /**
+     * Include default methods from the blueprint and its super types.
```
ljnelson: "default" as in Java language `default` interface methods? Or some other meaning of default?

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -415,5 +415,26 @@ default void decorateAddSet(B builder, Set<T> optionValues) {
...
+     * <p>
+     * If the value is specified, only methods with names matching the values are included, as long as they qualify as
```
ljnelson, line 422: So it's a filter that allows everything to pass through unless it is configured otherwise?

tomas-langer: If no annotation is present - does not include `default` interface methods
If annotation is present without value - include all `default` interface methods that match property getter
If annotation is present and has a value - include only `default` interface methods that match property getter and one of the values

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
tomas-langer: I have updated the javadoc to be clear, I hope enough.

builder/api/src/main/java/io/helidon/builder/api/Prototype.java
tomas-langer: And I have updated the javadoc of the class to include the definition I provided above.

ljnelson (approved):

PR #10573, opened 2025-08-20
Fix error in checking for out-of-order values in configured customizations

### LINKED ISSUE #10572: 4.x MP metrics incorrectly flags correctly-arranged default settings as out-of-order
## Environment Details
* Helidon Version:
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
MP metrics, and our implementation, support the use of configuration to assign settings such as bucket values for histograms. 

Some such settings should be in increasing order but the code which enforces that is flawed and always issues a warning even when values are in order.

## Steps to reproduce
```bash
cd microprofile/tests/tck/tck-metrics
mvn test -D-Dmaven.surefire.debug
```
Set a breakpoint at `DistributionCustomizations:268`. 

The values passed in are in order, but the logic at 268 incorrectly judges that the first one is out of order so it judges the entire list to be out of order.

tjquinno:
### Description
Resolves #10572 

The check for out-of-order values had some bad logic. Fixed it and added a couple of tests. Had to refactor some code to make it accessible to the test.

### Documentation
Bug fix; no impact.

### DIFF
--- a/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/DistributionCustomizations.java
+++ b/microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/DistributionCustomizations.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2024, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


microprofile/metrics/src/main/java/io/helidon/microprofile/metrics/DistributionCustomizations.java
```diff
@@ -241,14 +243,12 @@ protected MultiValuedCustomization(String nameExpressionAndValues,
...
-        private T[] values(String valuesString,
-                           Class<T> type,
-                           Function<String, T> valueParser,
-                           Consumer<T> valueChecker) {
+        static <T extends Comparable<T>> ValuesResult<T> values(String valuesString,
+                                                                Class<T> type,
+                                                                Function<String, T> valueParser,
+                                                                Consumer<T> valueChecker,
+                                                                Supplier<String> namePrefix,
+                                                                Supplier<Boolean> hasTrailingWildcard) {
```
ljnelson, line 251: Nit: could use `BooleanSupplier`, yeah?

Issue #10575, opened 2025-08-20
4.x: io.helidon.microprofile.testing.AddConfig* do not require a reset


romain-grecourt:
## Environment Details
* Helidon Version:
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description

The [javadoc](https://github.com/helidon-io/helidon/blob/ededc0bb1306f4f478df200f80decbb77b21f0f4/microprofile/testing/testing/src/main/java/io/helidon/microprofile/testing/AddConfig.java#L30) of the `@io.helidon.microprofile.testing.AddConfig*` annotations states that they imply a dedicated CDI container. However that behavior is effectively not [implemented](https://github.com/helidon-io/helidon/blob/ededc0bb1306f4f478df200f80decbb77b21f0f4/microprofile/testing/testing/src/main/java/io/helidon/microprofile/testing/HelidonTestInfo.java#L333).

[//]: # "Describe the bug in detail highlighting current behavior vs expected behavior"
[//]: # "State if the problem is easily reproducible or happens intermittently"
[//]: # "Include stack traces or command outputs"

## Steps to reproduce

```java
@HelidonTest
class Aest {

    @Inject
    @ConfigProperty(name = "bar", defaultValue = "not-bar")
    String bar;

    @Test
    @AddConfig(key="bar", value="bar1")
    public void test1() {
        assertThat(bar, is("bar1"));
    }
}
```

## Workaround

The `@Configuration` annotation can be used to trigger a dedicated CDI container.

```java
@HelidonTest
class Test {

    @Inject
    @ConfigProperty(name = "bar", defaultValue = "not-bar")
    String bar;

    @Test
    @Configuration
    @AddConfig(key="bar", value="bar1")
    public void test1() {
        assertThat(bar, is("bar1"));
    }
}
```

PR #10583, opened 2025-08-21
4.x: io.helidon.microprofile.testing.AddConfig* do not require a reset

al3xandru:
### Description

4.x: io.helidon.microprofile.testing.AddConfig* do not require a reset

### Documentation

If no doc impact: None


al3xandru:
Where should the test be placed? I couldn't figure out the correct location.

Captain1653:
Hi @al3xandru 

I'm not a maintainer Helidon, but allow me give some advices:
1. It's better to add issue id in your commit message and it will be like that:
`marks test for reset when AddConfig is used on method (#10575)` 
It will link your PR to the original issue.

2. In your PR description better to add this `Fixes #10575` . It will give additional information for maintainers.

3. About your fix. I think you also should add `|| !descriptor.addConfigBlocks().isEmpty()` , because annotation `AddConfigBlock` has the same statement in javadoc `If used on a method, the container will be reset regardless of the test lifecycle.`

I think, that then core team will give you some additional info about tests.

Kind regards

al3xandru:
Not a PR to be proud of 😂

romain-grecourt (approved):

PR #10590, opened 2025-08-22
Add config support for retry policy on OTLP exporters

tjquinno:
### Description

1. Add missing config support for OpenTelemetry `RetryPolicy`. Add test.
2. Correct Javadoc on two `@hidden` items.

### Documentation
No impact.

### DIFF
--- a/telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/RetryPolicyConfigBlueprint.java
+++ b/telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/RetryPolicyConfigBlueprint.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/RetryPolicyConfigBlueprint.java
```diff
@@ -0,0 +1,65 @@
...
+@Prototype.Blueprint
+interface RetryPolicyConfigBlueprint {
+
+    /**
+     * Maximum number of retry attempts.
+     *
+     * @return max retry attempts
+     */
+    @Option.Configured
+    Optional<Integer> maxAttempts();
```
ljnelson, line 38: Nit: Can't this just be `int`?

tjquinno: When creating config for OpenTelemetry types, I've scrupulously used `Optional<X>`. OTel provides its own defaults if values are not specified, and I do not want to replicate in our config declarations any of that OTel default value information. If we were being careful we'd have to track all those default values from OTel across their releases to make sure our default remains current.

Also, the various factory methods and/or builder decorators have a pattern of

```java
var nativeBuilder = // Create an OTel builder for the thing.
...
config.maxAttempts().ifPresent(nativeBuilder::setMaxAttempts);
...
```
which works well.

PR #10620, opened 2025-08-29
4.x: Upgrades ojdbc* and ucp* artifacts to 23.9.0.25.07

ljnelson:
This pull request updates the `version.lib.ojdbc` property in `dependencies/pom.xml` to be, effectively, `23.9.0.25.07`.

This release of the `ojdbc*` and `ucp*` artifacts apparently removes a GraalVM native image `Feature` implementation that was erroneously calling non-existing native-image-related classes. It appears that this release has solved a family of problems that has prevented Helidon from uptaking recent versions of the `ojdbc*` and `ucp*` artifacts.

A subsequent pull request may follow to replace usages of `ojdbc11` with `ojdbc17` instead. This pull request deliberately does not make this change.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -147,7 +147,7 @@
             UCP versions 21.10.0.0 and up throw NPEs. There is a test to catch them.
             Until this bug is fixed do not upgrade past version 21.9.0.0.
         -->
-        <version.lib.ojdbc>${version.lib.ojdbc.family}.6.0.24.10</version.lib.ojdbc>
+        <version.lib.ojdbc>${version.lib.ojdbc.family}.9.0.25.07</version.lib.ojdbc>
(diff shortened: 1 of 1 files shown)


dependencies/pom.xml
```diff
@@ -147,7 +147,7 @@
             UCP versions 21.10.0.0 and up throw NPEs. There is a test to catch them.
```
barchetta, line 147: Should this comment  be updated or removed? Approving if you want to address in follow-up PR.

barchetta (approved):

dependencies/pom.xml
ljnelson: Good catch. I’ll merge and then follow up.

ljnelson:
Related: #10640

Issue #9351, opened 2024-10-09
Oracle database dependencies need to be refreshed


ljnelson:
We need to decide on what our policy is regarding what version of the Oracle Database to support out of the box, and then implement it properly. See https://support.oracle.com/knowledge/Oracle%20Database%20Products/742060_1.html for helpful guidance.

[As of this issue's writing](https://github.com/helidon-io/helidon/blob/4.1.2/dependencies/pom.xml#L132-L136), we support version 21.9.0.0, an innovation release. The referenced document indicates this is a non-long-term-stable release.

(Note that as noted in the `dependencies/pom.xml` file some versions of UCP in particular after 21.9.0.0 throw `NullPointerException`s so we need to proceed with caution.)

If we were to support an LTS release, and not an "innovation" release, then we should choose between 23ai and 19c. The obvious choice would seem to be 23ai, but I'm guessing that most of our users probably don't work with it yet.

ljnelson:
The JDBC drivers page has an interoperability matrix: https://www.oracle.com/database/technologies/appdev/jdbc-downloads.html#Q7

The JDBC drivers page also [says](https://www.oracle.com/database/technologies/appdev/jdbc-downloads.html#:~:text=19c%20is%20a%20Long%20Term%20Release%20with%20a%20premier%20support%20until%20April%202024%20and%20extended%20support%20until%20April%202027.%20This%20is%20the%20recommended%20version.), about 19c, "This is the recommended version" but it's unclear whether that's just an oversight given that the rest of the page pushes the 23ai drivers.

ljnelson:
The 23ai drivers have one ever-so-slightly [backwards-incompatible change regarding `setBoolean`](https://download.oracle.com/otn-pub/otn_software/jdbc/23c/JDBC-UCP-ReleaseNotes-23ai.txt?AuthParam=1728668129_07da2d53114e742f97151c1c459a62cd#:~:text=%2D%20setBoolean(int%20parameterIndex,instead%20of%20setBoolean.) so it is worth paying attention.

ljnelson:
[...]

PR #10640, opened 2025-09-03
4.x: uses the 17 variants of ojdbc and ucp artifacts

ljnelson:
This pull request uses the `17` variants of the `ojdbc*` and `ucp*` libraries.

Following https://github.com/helidon-io/helidon/issues/9351#issuecomment-2411990030, we want to make sure we're using the latest and greatest artifacts where possible.

### DIFF
--- a/data/sql/datasource/ucp/pom.xml
+++ b/data/sql/datasource/ucp/pom.xml
@@ -72,7 +72,7 @@
         </dependency>
         <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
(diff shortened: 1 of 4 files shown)


ljnelson:
@barchetta I'd like to get your opinion on this, particularly with respect to the backwards compatibility of dependencies. Most of our stuff depends on our `integrations/db/ojdbc` artifact, so things like the `ucp11` vs. `ucp17` artifact dependencies are transitive.

If/when this PR can be merged, #9351 can be closed as well.

PR #10648, opened 2025-09-05
4.x: Global config now works consistent with Testing.Test and global service registry

### LINKED ISSUE #10633: Helidon SE 4.2.5: Config.global() returns incorrect results when used with @Testing.Test
## Environment Details
* Helidon Version: 4.2.5
* Helidon SE
* JDK version: 21
* OS: OL9

----------

## Problem Description

Config.global() doesn't respond correctly when used with @Testing.Test annotation to get unique per-test Config instances. The current workaround is to rely solely on Services.get(Config.class) instead of Config.global(). The problem is, a non-trivial amount of our codebase currently calls Config.global().

## Steps to reproduce

First attempt was to use this test below. It fails because extension is per class, not test.
`
@Testing.Test
final class ConfigSeparationTest {

    @Test void one() {
        Services.set(Config.class, Config.just(MapConfigSource.create(Map.of("key","one"))));
        assertEquals("one", Config.global().get("key").asString().get());
    }

    @Test void two() {
        Services.set(Config.class, Config.just(MapConfigSource.create(Map.of("key","two"))));
        assertEquals("two", Config.global().get("key").asString().get());
    }

    @Test void three() {
        Services.set(Config.class, Config.just(MapConfigSource.create(Map.of("key","three"))));
        assertEquals("three", Config.global().get("key").asString().get());
    }
}
`

Which led to this attempt. Notice the WA already being applied to use Services.get(Config.class) instead of Config.global()

`
@ExtendWith(ConfigSeparationTest.MyExtension.class)
final class ConfigSeparationTest {
    @Test
    void one() {
        Services.set(Config.class, Config.just(MapConfigSource.create(Map.of("key", "one"))));
        assertEquals("one", Services.get(Config.class).get("key").asString().get());
    }

    @Test
    void two() {
        Services.set(Config.class, Config.just(MapConfigSource.create(Map.of("key", "two"))));
        assertEquals("two", Services.get(Config.class).get("key").asString().get());
    }

    @Test
    void three() {
        Services.set(Config.class, Config.just(MapConfigSource.create(Map.of("key", "three"))));
        assertEquals("three", Services.get(Config.class).get("key").asString().get());
    }

    static class MyExtension implements Extension,
                                        InvocationInterceptor {

        @Override
        public void interceptTestMethod(Invocation<Void> inv,
                                        ReflectiveInvocationContext<Method> ic,
                                        ExtensionContext ec) throws Throwable {
            try {
                var id = "test-" + ic.getTargetClass().getName() + "-" + ic.getExecutable().getName();
                var ctx = Context.builder().id(id).build();
                ctx.register("helidon-registry-static-context", ctx);
                ctx.supply("helidon-registry", ServiceRegistry.class, () -> ServiceRegistryManager.create().registry());
                Contexts.runInContext(ctx, () -> {
                    try {
                        return inv.proceed();
                    } catch (RuntimeException e) {
                        throw e;
                    } catch (Throwable e) {
                        throw new UncheckedException(e);
                    }
                });
            } catch (UncheckedException e) {
                throw e.getCause();
            }
        }
    }
}
`

tomas-langer:
### Description
Resolves #10633 

This change makes sure that if global config is set using service registry, the service registry will be used for further lookups, to avoid setting a static global config when user did not intend to.

This is now aligned with `@Testing.Test` extension (and webserver and MP testing).

### DIFF
--- a/common/config/src/main/java/io/helidon/common/config/GlobalConfig.java
+++ b/common/config/src/main/java/io/helidon/common/config/GlobalConfig.java
@@ -47,7 +47,7 @@ public final class GlobalConfig {
     private static final System.Logger LOGGER = System.getLogger(GlobalConfig.class.getName());
     private static final AtomicBoolean LOGGED_REGISTERED = new AtomicBoolean(false);
     private static final Config EMPTY = Config.empty();
-    private static final AtomicReference<Boolean> GLOBAL_FROM_REGISTRY = new AtomicReference<>();
+    private static final AtomicBoolean GLOBAL_FROM_REGISTRY = new AtomicBoolean();
     private static final LazyValue<Config> DEFAULT_CONFIG = LazyValue.create(() -> {
         List<io.helidon.common.config.spi.ConfigProvider> providers =
                 HelidonServiceLoader.create(ServiceLoader.load(io.helidon.common.config.spi.ConfigProvider.class))
@@ -84,7 +84,7 @@ public final class GlobalConfig {
      */
     @Deprecated(forRemoval = true, since = "4.2.0")
     public static Config config() {
-        if (Boolean.TRUE.equals(GLOBAL_FROM_REGISTRY.get())) {
+        if (GLOBAL_FROM_REGISTRY.get()) {
             return Services.get(Config.class);
         }
 
@@ -115,7 +115,7 @@ public final class GlobalConfig {
     public static Config config(Supplier<Config> config, boolean overwrite) {
         Objects.requireNonNull(config);
 
-        if (Boolean.TRUE.equals(GLOBAL_FROM_REGISTRY.get())) {
+        if (GLOBAL_FROM_REGISTRY.get()) {
             return Services.get(Config.class);
         }
 
--- a/common/config/src/main/java/io/helidon/common/config/GlobalConfig.java
+++ b/common/config/src/main/java/io/helidon/common/config/GlobalConfig.java
@@ -47,6 +47,7 @@ public final class GlobalConfig {
     private static final System.Logger LOGGER = System.getLogger(GlobalConfig.class.getName());
     private static final AtomicBoolean LOGGED_REGISTERED = new AtomicBoolean(false);
     private static final Config EMPTY = Config.empty();
+    private static final AtomicReference<Boolean> GLOBAL_FROM_REGISTRY = new AtomicReference<>();
     private static final LazyValue<Config> DEFAULT_CONFIG = LazyValue.create(() -> {
         List<io.helidon.common.config.spi.ConfigProvider> providers =
                 HelidonServiceLoader.create(ServiceLoader.load(io.helidon.common.config.spi.ConfigProvider.class))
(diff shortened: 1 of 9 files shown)


common/config/src/main/java/io/helidon/common/config/GlobalConfig.java
```diff
@@ -47,6 +47,7 @@ public final class GlobalConfig {
     private static final System.Logger LOGGER = System.getLogger(GlobalConfig.class.getName());
     private static final AtomicBoolean LOGGED_REGISTERED = new AtomicBoolean(false);
     private static final Config EMPTY = Config.empty();
+    private static final AtomicReference<Boolean> GLOBAL_FROM_REGISTRY = new AtomicReference<>();
```
danielkec: Could be AtomicBoolean or even just primitive volatile boolean

tomas-langer: I needed three states (null/true/false) originally and then refactored it to just need two. Will use AtomicBoolean

config/config/src/main/java/io/helidon/config/Config.java
```diff
@@ -422,7 +422,7 @@ static Config global() {
             return BuilderImpl.GlobalConfigHolder.get();
         }
         Config config = Services.get(Config.class);
-        io.helidon.common.config.GlobalConfig.config(() -> config, true);
+        io.helidon.common.config.GlobalConfig.config(config, true);
```
romain-grecourt, line 425: Can we remove this line ?

common/config/src/main/java/io/helidon/common/config/GlobalConfig.java
romain-grecourt, line 1: How about this:

```java
// package io.helidon.common.config
class ConfigFactory implements Supplier<Config> {
    @Override
    public Config get() {
        return Config.create();
    }
}
```

```java
// package io.helidon.common.config
class GlobalConfig {
    @Service.Singleton
    static final class Holder {
        private final AtomicReference<Config> ref = new AtomicReference<>();
    }

    public static Config config() {
        ServiceRegistry registry = GlobalServiceRegistry.registry();
        Config config = registry.get(Holder.class).ref.get();
        if (config == null) {
            config = registry.get(Config.class);
        }
        return config;
    }

    public static Config config(Supplier<Config> supplier, boolean overwrite) {
        Objects.requireNonNull(supplier);
        // TODO consider exposing CoreServiceRegistry.set
        //  to avoid the double static context lookup
        Holder holder = Services.get(Holder.class);
        Config config = holder.ref.get();
        if (overwrite || config == null) {
            config = supplier.get();
            try {
                Services.set(Config.class, config);
            } catch (Exception e) {
                holder.ref.set(config);
            }
        }
        return config;
    }
}
```

tomas-langer: I am not sure how this solves the problem. The second class that needs to explicitly set an instance, or use a service registry instance is the `io.helidon.config.Config` method.
We need to distinguish the cases, where a global instance was obtained from a registry (and continue obtaining it from registry), and when it was explicitly set by a user (and we must use that instance from that point in time). This is for backward compatibility within 4.x version.

romain-grecourt: If a user uses the Global config setter explicitly then the Global config getter should always return the explicitly set instance, isn't that the expected behavior ?

We should minimize the side effects of the global config getters. I.e It should at most trigger a registry access and trigger the default config supplier.

config/config/src/main/java/io/helidon/config/Config.java
tomas-langer: No we cannot, because common config needs to know that it should use service registry next time an instance is requested.

romain-grecourt (approved):

PR #10655, opened 2025-09-08
For MP CORS, suppress explanation of rejection in response; log it instead

### LINKED ISSUE #10630: 4.x CORS response contains unnecessary information
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
When CORS rejects a request the response can be less verbose.

## Steps to reproduce
Set a breakpoint at line 112 of the existing unit test `CorsTest` in `webserver/cors`. Then run the test with the debugger and examine the status (numeric and text).

tjquinno:
### Description
Resolves #10630 

Identical in spirit to PR #10631 but for MP.

When we reject a CORS request, log the underlying reason instead of sending the reason back in the response.

### Documentation
Bug fix; no impact.


spericas (approved):
LGTM

PR #10660, opened 2025-09-09
4.x: Metadata module and loading of files compatible with single jars

### LINKED ISSUE #10657: 4.x: Design metadata handling for Helidon
Currently we have a few files in our `META-INF/helidon` directory. 
These files are loaded by ways specific to each of the components.

We should design a way to handle these files that is shared, and that:

1. Can survive shading/assembly plugin
2. Still has a single location of a resource to avoid classpath scanning under "normal" deployments (i.e. a manifest)

Initial design

- identify what kind of deployment we are in (shaded jar, multi-jar classpath)
- choose either classpath scanning or manifest based discovery
- each metadata file (generated files, and files that require merging for runtime) should be in a unique location (i.e. a unique directory within each module), file names MUST be constant
- a `manifest` file exists in each module with metadata files that contains a list of all metadata files (each line is one metadata file location)

tomas-langer:
Resolves #10657

Helidon Metadata
-----

This module is responsible for discovering Helidon metadata both when running on "proper" classpath or module path,
and when running on "uber" classpath (i.e. fat jars, shaded jars, etc.).

Important note - Metadata discovery only works on unique file names. As we gather multiple files from different locations,
the file name is the only element we can use for grouping.

# Files

The following files are expected in a single location on classpath, and will not work with shaded jar, unless correctly merged (this will not cause runtime failures):
- `default-media-types.properties` - default media types (there should be exactly one instance of this file used by Helidon)
- `config-metadata.json` - configuration metadata, may be used by tooling to find all configurable options (i.e. from IDE), not used at runtime of Helidon
- `feature-metadata.properties` - deprecated feature metadata

The following files are expected to be under `META-INF/helidon` directory structure.
Each file then lives in a directory structure depending on module type (see JPMS modules and regular modules). 

The following files are expected by Helidon; this list may be extended in the future (`${file-name}` in the location patterns):

- `service-registry.json` - information for Helidon `ServiceRegistry`
- `feature-registry.json` - feature metadata, used at startup and when building native image
- `service.loader` - Java `ServiceLoader` services that should be discovered by Helidon `ServiceRegistry`
- `serial-config.properties` - configuration for Java serialization
- `media-types.properties` - customized media types

All the files mentioned above will also be discovered from `META-INF/helidon` directory (directly) for backward compatibility,
testing, and ease of use when not shading. 
A general rule is that each resource is used just once (i.e. if a duplicate record in manifest exists).

## JPMS modules

Each file will be in a directory structure `META-INF/helidon/${jpms-module-name}/${file-name}`, where:

- `jpms-module-name` is the name of the module (Java module system)
- `file-name` is one of the supported Helidon metadata files

## Regular modules

Each file will be in a directory structure `${package-name}-${unique-id}/${file-name}`, where:

- `package-name` is the package name considered "top" package of the module by Helidon codegen
- `unique-id` is a unique identifier generated at build time by Helidon codegen (i.e. a UUID, or a timestamp)
- `file-name` is one of the supported Helidon metadata files

# Discovery

If there is more than one `META-INF/MANIFEST.MF` files on the classpath, we follow the multiple resources approach;
if there is one or less manifest files, we follow the classpath scanning approach.

## Multiple resources approach

The discovery process:

1. Lookup all `META-INF/helidon/manifest` files
2. Read each file as lines, where lines prefixed with `#` are considered comments, and empty lines are ignored
3. Each line is a location of a metadata file on the classpath (expected to be unique)
4. The metadata module will group all files by file name, to allow other components to discover them

## Classpath scanning approach

The discovery process:

1. Classpath must be either file based or module-file based (i.e. jars and directories), cannot work with custom classloaders
2. We analyze all files under `META-INF/helidon` except for the manifest file
3. The metadata module will group all files by file name, to allow other components to discover them

# Options

The metadata module can have options configured to modify its behavior.
Options may also be configured when invoking `Metadata` programmatically.

The following options are currently supported:

- `mode` - which mode to use, see Metadata.Mode, defaults to `AUTO`, can be changed using a Java system property
            (i.e. `-Dio.helidon.metadata.mode=RESOURCES`)
- `location` - location from which we start classpath discovery, when used, defaults to `META-INF/helidon`
  (this is expected to contain the manifest file, and possible metadata files that are not part of the manifest for backward
  compatibility)
- `metadata-files` - list of file names of metadata files to find (comma separated list, defaults to all supported files)
- `manifest-file` - location of the main manifest file


metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscoveryImpl.java
```diff
@@ -0,0 +1,473 @@
...
+            case RESOURCES -> createFromResources(config);
+            case SCANNING -> createFromClasspathScanning(config);
+            case NONE -> new MetadataDiscoveryImpl(Map.of());
+        };
+    }
+
+    static MetadataDiscovery createFromClasspathScanning(MetadataDiscoveryConfig config) {
+
+        // all files and directories that are part of the current classpath/module path
+        Set<Path> configuredClasspath = new HashSet<>();
```
romain-grecourt: Should this call `modules()` instead ?

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
```diff
@@ -0,0 +1,216 @@
...
+     * @param fileName name of the file
+     * @return list of metadata instances, empty list if none found (never null)
+     * @throws java.lang.NullPointerException if {@code fileName} is null
+     */
+    List<MetadataFile> list(String fileName);
+
+    /**
+     * Modes of discovery.
+     */
+    enum Mode {
```
romain-grecourt, line 64: How about a sealed interface instead ?

E.g.

```java
public sealed interface MetadataDiscovery permits MetadataDiscovery.Resources,
                                                  MetadataDiscovery.Scanning,
                                                  MetadataDiscovery.Noop {

    List<MetadataFile> list(String fileName);

    static MetadataDiscovery create() {
        String mode = System.getProperty("io.helidon.metadata.mode", "auto");
        return switch (mode) {
            case "auto" -> auto(Thread.currentThread().getContextClassLoader());
            case "resources" -> resources(Thread.currentThread().getContextClassLoader());
            case "scanning" -> scanning();
            default -> noop();
        };
    }

    static MetadataDiscovery auto(ClassLoader classLoader) {
        return MetadataDiscoverySupport.auto(classLoader);
    }

    static MetadataDiscovery resources(ClassLoader classLoader) {
        return new Resources(classLoader);
    }

    static MetadataDiscovery scanning() {
        return new Scanning();
    }

    static MetadataDiscovery noop() {
        return new Noop();
    }

    final class Resources implements MetadataDiscovery {

        LazyValue<Map<String, List<MetadataFile>>> files;

        private Resources(ClassLoader classLoader) {
            files = LazyValue.create(() -> MetadataDiscoverySupport.manifests(classLoader));
        }

        @Override
        public List<MetadataFile> list(String fileName) {
            Objects.requireNonNull(fileName, "fileName is null");
            return files.get().getOrDefault(fileName, List.of());
        }
    }

    final class Scanning implements MetadataDiscovery {

        LazyValue<Map<String, List<MetadataFile>>> files;

        private Scanning() {
            files = LazyValue.create(MetadataDiscoverySupport::scan);
        }

        @Override
        public List<MetadataFile> list(String fileName) {
            Objects.requireNonNull(fileName, "fileName is null");
            return files.get().getOrDefault(fileName, List.of());
        }
    }

    final class Noop implements MetadataDiscovery {

        private Noop() {
        }

        @Override
        public List<MetadataFile> list(String fileName) {
            return List.of();
        }
    }
}
```

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Helidon metadata discovery.
+ */
+public interface MetadataDiscovery {
+    /**
+     * Default location of the metadata files, this is expected to be a directory that is the same in every
+     * jar file. Files in this location will be explicitly added only if they match the provided metadata file name.
+     */
+    String LOCATION = "META-INF/helidon";
```
romain-grecourt: Can these constants go in a separate class, E.g. `MetadataFiles` ?

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscoveryConfig.java
romain-grecourt, line 1: Do we need a builder ?

- `classloader` can be a factory method parameter
- `location` can be hard-coded to `META-INF/helidon`
- `metadataFiles` is only used to limit the discovered files, can be a factory method parameter, or removed

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
```diff
@@ -0,0 +1,216 @@
...
+         */
+        AUTO,
+        /**
+         * Uses resources discovered from classpath, expecting duplicate resources.
+         * <p>
+         * NOTE: This approach will not work with unmerged manifests.
+         */
+        RESOURCES,
+        /**
+         * Classpath scanning mode, ignores all manifest files on classpath, and discovers all metadata files.
```
romain-grecourt:
```
Discovers metadata files by scanning the file entries defined by the {@code java.class.path} and {@code jdk.module.path} system properties.
<p>
NOTE: This mode adds a startup performance overhead and requires file based class loaders.
```

tests/integration/packaging/se-1/src/main/resources/logging.properties
```diff
@@ -19,11 +19,12 @@
 
 # Send messages to the console
 handlers=io.helidon.logging.jul.HelidonConsoleHandler
-# Use JSON formatter with default format
-io.helidon.logging.jul.HelidonConsoleHandler.formatter=io.helidon.logging.jul.HelidonJsonFormatter
+java.util.logging.SimpleFormatter.format=%1$tY.%1$tm.%1$td %1$tH:%1$tM:%1$tS %4$s %3$s !thread!: %5$s%6$s%n
```
romain-grecourt: Why not use HelidonConsoleHandler here ?

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscoveryImpl.java
tomas-langer: oh, yes, I have started extracting that part and forgot about it.

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
tomas-langer: That is an implementation detail that does not help with anything and you duplicate the code (as both `Resources` and `Scanning` is exactly the same implementation).
Also removing the enum seems wrong, as it is used in configuration, and enum is strictly better than string constants

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
tomas-langer: Not sure I see the reason for that. I guess we could, but then the type would only contain the constants and is useless otherwise.

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscoveryConfig.java
tomas-langer: there are multiple parameters, most of them optional.
I wanted this usable by others as well - that is the reason I did not hardcode the values. If the values are hardcoded, this would only ever be usable by Helidon itself with no extensibility outside of our modules.

tests/integration/packaging/se-1/src/main/resources/logging.properties
tomas-langer: This was an accident, I needed to analyze the output and it was hard with the JSON formatter

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscoveryConfig.java
tomas-langer: It is also very useful for testing the module, as I can have more than one set of test metadata. If I remove this, I will need to create at least three new modules to test it.

romain-grecourt: We don't need to a design a generic API, this should be Helidon centric.

Why would you need 3 new modules ?

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
romain-grecourt: The main idea is to remove the configuration and to get away with a just an interface and factory method(s), no builder, no configuration.

There could be only implementation, that takes a `Supplier<Map<String, List<MetadataFile>>>` as argument, and all the logic would be in the factory method(s).

Having the enum and only one factory method seems fine.

E.g.

```java
public interface MetadataDiscovery {

    List<MetadataFile> list(String fileName);

    enum MODE {
        AUTO,
        RESOURCES,
        SCANNING,
        NONE
    }

    static MetadataDiscovery create(Mode mode) {
        // ...
    }

    static MetadataDiscovery getInstance() {
        // uses "io.helidon.metadata.mode" system property
    }
}
```

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
romain-grecourt: It makes the API cleaner.

We can also just move the individual resource file name constants in their respective modules, and only have two methods to resolve the resource paths.

E.g.

```java

static String path(String module, String filename) {
    // ...
}

static String path(String filename) {
    // ...
}
```

tomas-langer: No we cannot, because the names of files must be present at discovery.

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscoveryConfig.java
tomas-langer: 1 module to test "good metadata"
1 module to test non-existent metadata in `manifest` file (as this fails at discovery)
1 module to test classpath scanning

tomas-langer: I will try to refactor this to remove the configurability.

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
romain-grecourt: Why "must" ?

If the constants remain, I would prefer to have them be in a separate class.

tomas-langer: Because the discovery actually uses the file names - for example when doing classpath scanning, we have to look for some files, and this is the list.
We could change it to just collect all files under `META-INF/helidon` - then we could remove the constants and move them back to the modules.

tomas-langer: And we do not want to create a new instance of discovery for each component that looks for these files, as that may end up with multiple classpath scans, or even just multiple loads of `manifest` files.

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param fileName name of the file
+     * @return list of metadata instances, empty list if none found (never null)
+     * @throws java.lang.NullPointerException if {@code fileName} is null
+     */
+    List<MetadataFile> list(String fileName);
+
+    /**
+     * Modes of discovery.
+     * Can be controlled through system property {@value MetadataDiscoveryImpl#SYSTEM_PROPERTY_MODE}.
```
romain-grecourt, line 62: In the value going to be inlined in the javadoc ? Otherwise we might have to move the constant since MetadataDiscoveryImpl is not public.

metadata/metadata/pom.xml
romain-grecourt, line 1: Should we make this a common module instead ? I.e. common/metadata

tests/integration/packaging/se-1/pom.xml
```diff
@@ -230,5 +230,44 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <!-- Please do not use this style of packaging -->
```
romain-grecourt: This is a test, who's reading this comment ? :D

tests/integration/packaging/se-1/src/main/assembly/fat-assembly.xml
romain-grecourt, line 1: Should we add this to all packaging test projects ?

metadata/metadata/src/main/java/io/helidon/metadata/MetadataDiscovery.java
tomas-langer: This is rendered in javadoc (tested with local build of `mvn install -Pjavadoc`:
```
Modes of discovery. Can be controlled through system property "io.helidon.metadata.mode".
```

tests/integration/packaging/se-1/pom.xml
tomas-langer: I try to share this as much as I can, even in tests...

romain-grecourt (approved):

PR #10663, opened 2025-09-10
Add support for exception mappers to JSON-RPC

spericas:
### Description
Add support for exception mappers to JSON-RPC. A mapper can optionally return a JsonRpcError. If it does not, then the JSON-RPC response is returned to the client. Also fixes minor issue JSON-RPC client. New tests.

### Documentation
In Javadoc only.

### DIFF
--- a/webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcExceptionHandler.java
+++ b/webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcExceptionHandler.java
@@ -21,8 +21,8 @@ import io.helidon.jsonrpc.core.JsonRpcError;
 import io.helidon.webserver.ServerLifecycle;
 
 /**
- * An exception handler that can be registered to map exceptions thrown in JSON-RPC
- * method handlers.
+ * An exception handler that can be registered to map exceptions thrown in method
+ * handlers to {@code JsonRpcError}s.
  */
 @FunctionalInterface
 public interface JsonRpcExceptionHandler extends ServerLifecycle {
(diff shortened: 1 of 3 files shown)


webserver/pom.xml
```diff
@@ -29,6 +29,16 @@
     <groupId>io.helidon.webserver</groupId>
     <artifactId>helidon-webserver-project</artifactId>
     <name>Helidon WebServer Project</name>
+    <dependencies>
+        <dependency>
+            <groupId>io.helidon.builder</groupId>
+            <artifactId>helidon-builder-api</artifactId>
+        </dependency>
+        <dependency>
```
tvallin: This dependency is duplicate.

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcHandlersBlueprint.java
```diff
@@ -42,4 +42,12 @@ interface JsonRpcHandlersBlueprint {
...
      */
     Optional<JsonRpcErrorHandler> errorHandler();
+
+    /**
+     * Return a map of throwable to exception handlers.
+     *
+     * @return a map of method names to handlers
+     */
+    @Option.Singular(value = "exception", withPrefix = false)
+    Map<Class<? extends Throwable>, JsonRpcExceptionHandler> exceptionMap();
```
tvallin, line 52: This allow registering a mapper for a single JSON-RPC handler right ? Can we had a global handler that would be called if there is no specific handler exception mapper registered ?

webserver/pom.xml
spericas: Oh, this must be Intellij ...

PR #10668, opened 2025-09-11
4.x: Fixed missing required query and header parameters in declarative

### LINKED ISSUE #10666: 4.x: Declarative - if path param, header etc. not found, we end up with a 500
When using Helidon declarative HTTP, and a parameter is "injected" to a method (such as Query parameter, header parameter) and it is not available in the actual request, we throw an exception that ends with internal server error.

We need to figure out what kind of exception should be thrown (such as bad request, or some custom we could map in exception mapper) for non-optional parameters and handle it nicely.

We must never end with a log statement or 500, as this can be crafted by client (i.e. it would be an attack vector).

This feature is not yet released.

tomas-langer:
Missing header or query parameter that is required in declarative method will now end with a bad request, rather than internal server error.

Resolves #10666

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/webserver/ParamProviderHttpHeader.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/webserver/ParamProviderHttpHeader.java
@@ -107,7 +107,9 @@ class ParamProviderHttpHeader extends AbstractParametersProvider implements Http
                     .addContent(BAD_REQUEST_EXCEPTION)
                     .addContent("(\"Header \" + ")
                     .addContent(headerConstantName)
-                    .addContentLine(".defaultCase() + \" is not present in the request.\"));")
+                    .addContent(".defaultCase() + ")
+                    .addContentLiteral(" is not present in the request.")
+                    .addContentLine("));")
                     .decreaseContentPadding()
                     .decreaseContentPadding();
         }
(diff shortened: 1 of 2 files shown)


declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/webserver/ParamProviderHttpHeader.java
```diff
@@ -93,13 +101,22 @@ public boolean codegen(ParameterCodegenContext ctx) {
...
+            // add .orElseThrow() in case the header is missing
+            contentBuilder.addContentLine("")
+                    .addContent(".orElseThrow(() -> new ")
+                    .addContent(BAD_REQUEST_EXCEPTION)
+                    .addContent("(\"Header \" + ")
                     .addContent(headerConstantName)
-                    .addContent(").");
-            getMethod(contentBuilder, parameterType);
-            contentBuilder.addContentLine(";");
+                    .addContentLine(".defaultCase() + \" is not present in the request.\"));")
```
danielkec:
```suggestion
                    .addContent(".defaultCase() + ")
                    .addContentLiteral(" is not present in the request.")
                    .addContentLine("));")
```

declarative/tests/http/src/main/resources/logging.properties
```diff
@@ -20,3 +20,5 @@ java.util.logging.SimpleFormatter.format=%1$tY.%1$tm.%1$td %1$tH:%1$tM:%1$tS.%1$
 # Global logging level. Can be overridden by specific loggers
 .level=INFO
 io.helidon.level=INFO
+
+io.helidon.webserver.level=FINEST
```
danielkec: Intentional?

danielkec (approved):
Just nitpicks

danielkec (approved):

PR #10672, opened 2025-09-11
Move feature info from metrics API to Micrometer implementation component

### LINKED ISSUE #10665: 4.x: Feature is on metrics API instead of implementation
Currently we have a `Metrics` feature reported even if metrics are not used by application, because the `@Features` annotation is on the API module.

The annotations (and codegen) should be moved to implementation(s) instead, so it is not reported when not used.

tjquinno:
### Description
Resolves #10665 

### Release Note
____
Previously, Helidon would report that the "metrics" feature was included in an app if the metrics API in `helidon-metrics-api` was included in the project, regardless of whether a metrics implementation was also included.

Helidon now reports that the metrics feature is present only if a metrics implementation (Micrometer) is present in the app.
____

### Documentation
None

### DIFF
--- a/metrics/api/pom.xml
+++ b/metrics/api/pom.xml
@@ -92,13 +92,7 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <forceJavacCompilerUse>true</forceJavacCompilerUse>
                     <annotationProcessorPaths>
-                        <path>
-                            <groupId>io.helidon.common.features</groupId>
-                            <artifactId>helidon-common-features-codegen</artifactId>
-                            <version>${helidon.version}</version>
-                        </path>
                         <path>
                             <groupId>io.helidon.config.metadata</groupId>
                             <artifactId>helidon-config-metadata-codegen</artifactId>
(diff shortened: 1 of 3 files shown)


metrics/api/pom.xml
```diff
@@ -53,11 +53,6 @@
             <groupId>io.helidon.service</groupId>
             <artifactId>helidon-service-registry</artifactId>
         </dependency>
-        <dependency>
```
tomas-langer, line 56: There is also the annotation processor configuration - you should remove the `helidon-common-features-codegen` from there.

metrics/providers/micrometer/pom.xml
```diff
@@ -59,6 +59,11 @@
             <groupId>io.helidon.service</groupId>
             <artifactId>helidon-service-registry</artifactId>
         </dependency>
+        <dependency>
```
tomas-langer, line 62: And the codegen is probably missing from here

metrics/providers/micrometer/src/main/java/module-info.java
```diff
@@ -14,10 +14,20 @@
...
+import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.HelidonFlavor;
+
 /**
  * Micrometer adapter for Helidon metrics API.
  */
+@Features.Name("Metrics")
+@Features.Description("Metrics")
+@Features.Flavor(HelidonFlavor.SE)
+@Features.Path({"Metrics"})
```
tomas-langer: Please change this to `"Metrics", "Micrometer", as this is a "subfeature" of metrics.
If that does not work correctly (i.e. `Metrics` would not be visible as a feature when you run an application), then just change the `Description` to contain the information about Micrometer.

Also similar change should be done to any other metrics provider we have

PR #10674, opened 2025-09-12
Lc4j doc update

danielkec:
Small fixes and SE/MP guide


danielkec:
> The convention is one "*Snippets.java" file per .adoc file. The package structures follows the directory structure of the .adoc files.
> 
> You can use nested classes for namespaces instead of java packages.

The idea is that references to other snippet classes are not duplicated, in this case there are common classes, mp and se specific ones, eg. I can't have 3 inner classes of the same name.

romain-grecourt:
> I can't have 3 inner classes of the same name.

Something like this ?

https://github.com/helidon-io/helidon/blob/75392a92c691b67043f6bb8f5f621e3183166aff/docs/src/main/java/io/helidon/docs/mp/TelemetrySnippets.java#L70

https://github.com/helidon-io/helidon/blob/75392a92c691b67043f6bb8f5f621e3183166aff/docs/src/main/java/io/helidon/docs/mp/TelemetrySnippets.java#L136

danielkec:
> > I can't have 3 inner classes of the same name.
> 
> Something like this ?
> 
> https://github.com/helidon-io/helidon/blob/75392a92c691b67043f6bb8f5f621e3183166aff/docs/src/main/java/io/helidon/docs/mp/TelemetrySnippets.java#L70
> 
> https://github.com/helidon-io/helidon/blob/75392a92c691b67043f6bb8f5f621e3183166aff/docs/src/main/java/io/helidon/docs/mp/TelemetrySnippets.java#L136

As I said imports would be broken, it would have to look like this:
```
class LangChain4jSnippets {
    class SnippetBase {
        // tag::snippet_1[]
        @Path("/chat")
        public class PirateResource {

            @Inject
            LangChain4jSharedSnippets.SnippetBase.PirateService pirateService;

            @POST
            public String chat(String message) {
                return pirateService.chat(message);
            }
        }
        // end::snippet_1[]
    }
```

romain-grecourt:
```java
class LangChain4jSnippets {
    // stub
    static class PirateService {
        String chat(String msg) {
            return null;
        }
    }
        
    class Snippet1 {

        // tag::snippet_1[]
        @Path("/chat")
        public class PirateResource {

            @Inject
            PirateService pirateService;

            @POST
            public String chat(String message) {
                return pirateService.chat(message);
            }
        }
        // end::snippet_1[]
    }
}
```

danielkec:
I do yield to convention

romain-grecourt (approved):

PR #10676, opened 2025-09-12
In MP CORS do not add CORS headers if request is rejected

### LINKED ISSUE #10630: 4.x CORS response contains unnecessary information

tjquinno:
### Description
Resolves #10630 

### Release note ###
____
Helidon MP no longer adds CORS headers if the request is rejected.
____

The MP CORS code path was still adding CORS headers even if the CORS processing was rejecting the request. This PR suppresses the addition of headers in that case. The status in the response adapter will have been set by CORS code to 403 Forbidden by the time `prepareResponse` is running, so the change in the PR simply returns immediately in that case rather than adding CORS headers to the response.

PR also adds a test.

### Documentation
No impact.

### DIFF
--- a/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
+++ b/cors/src/main/java/io/helidon/cors/CorsSupportHelper.java
@@ -315,7 +315,8 @@ public class CorsSupportHelper<Q, R> {
     }
 
     /**
-     * Prepares a response with CORS headers, if the supplied request is in fact a CORS request.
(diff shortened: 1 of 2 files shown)


spericas (approved):
LGTM

PR #10681, opened 2025-09-15
Recommend Java 25

barchetta:
### Description

Recommend Java 25 in documentation.

Java 25 contains performance enhancements and fixes that benefit virtual threads. We recommend it!


ljnelson:
(Super minor nit: you use both "Java 25+" and "Java 25 or newer" idioms; maybe pick one and stick with it.)

barchetta:
> (Super minor nit: you use both "Java 25+" and "Java 25 or newer" idioms; maybe pick one and stick with it.)

Fixed -- left "+" usage in one spot.

PR #10684, opened 2025-09-17
Reusable telemetry tracing test utilities

tjquinno:
### Description
Adds telemetry testing utilities to simplify recording and checking for expected OpenTelemetry span data.

### Release Note
___
Developers can use utility classes in the new `helidon-telemetry-testing-tracing` component to capture and inspect OpenTelemetry span data in their unit tests.
___

The PR adds a new test-time component for telemetry. It uses a span exporter provided by OpenTelemetry that sends JSON-encoded span data to a `java.util.logging.Logger`, and it includes code to extract the JSON data into Java data structures suitable for use from developer (or Helidon) unit tests.

In the past, various Helidon components have used an in-memory collector for similar purposes but have replicated that test code in each component where it was needed. Eventually we should convert those tests to use this new component.

A soon-to-be-added telemetry example already uses this component.

The OpenTelemetry exporter determines the JSON structure. The Java types in this new component map very directly to that structure, so although they might appear a bit odd at first glance it's important that they track the underlying JSON structure.

### Documentation
`README.md` included in the PR.

### DIFF
--- a/telemetry/testing/tracing/pom.xml
+++ b/telemetry/testing/tracing/pom.xml
@@ -31,6 +31,7 @@
 
     <properties>
         <spotbugs.exclude>etc/spotbugs/exclude.xml</spotbugs.exclude>
+        <javadoc.fail-on-warnings>true</javadoc.fail-on-warnings>
     </properties>
 
     <dependencies>
@@ -51,5 +52,20 @@
             <artifactId>hamcrest-all</artifactId>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>check-dependencies</id>
+                        <phase>verify</phase>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
 
 </project>


telemetry/testing/tracing/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.helidon.telemetry.testing</groupId>
+        <artifactId>helidon-telemetry-testing-project</artifactId>
+        <version>4.3.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>helidon-telemetry-testing-tracing</artifactId>
```
tomas-langer, line 28: All new production modules should have:

Property:
```xml
```

And plugin:
```xml
            <plugin>
                <artifactId>maven-dependency-plugin</artifactId>
                <executions>
                    <execution>
                        <id>check-dependencies</id>
                        <phase>verify</phase>
                    </execution>
                </executions>
            </plugin>
```

tjquinno: Fixed. I guess the dev-guidelines document needs an update ;-) .

PR #10701, opened 2025-09-23
4.x: Declarative security

### LINKED ISSUE #10607: Security annotations and runtime (most likely requires backward incompatible changes)

tomas-langer:
### Description
Resolves #10607

TBD:
- docs update - this will be done in a separate PR after both this and Declarative Validation are merged, as validation is doing too many changes to make it reasonable in this PR

### DIFF
--- a/security/providers/common/src/main/java/io/helidon/security/providers/common/spi/AnnotationAnalyzer.java
+++ b/security/providers/common/src/main/java/io/helidon/security/providers/common/spi/AnnotationAnalyzer.java
@@ -168,7 +168,7 @@ public interface AnnotationAnalyzer {
         if (declaredMethod.getParameterCount() != expectedArguments.size()) {
             return false;
         }
-        // same number of parameters, same name, let's got through parameters
+        // same number of parameters, same name, let's go through parameters
         for (int i = 0; i < declaredMethod.getParameterTypes().length; i++) {
             Class<?> parameterType = declaredMethod.getParameterTypes()[i];
             if (!expectedArguments.get(i).typeName().equals(TypeName.create(parameterType))) {
--- a/webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityInterceptor.java
+++ b/webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityInterceptor.java
@@ -110,7 +110,8 @@ class HttpSecurityInterceptor implements HttpEntryPoint.Interceptor {
         UriInfo requestedUri = req.requestedUri();
         SecurityContext securityContext = req.context()
                 .get(SecurityContext.class)
-                .orElseThrow();
+                .orElseThrow(() -> new IllegalStateException("No security context in request context. "
+                                                                     + "Make sure the ContextFeature is added to WebServer"));
 
         String resourceType = ctx.serviceInfo().serviceType().fqName();
         var securityEnvironment = SecurityEnvironment.builder(security.serverTime())


microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ServiceRegistryExtension.java
```diff
@@ -174,6 +174,10 @@ void registryToCdi(@Observes @Priority(Interceptor.Priority.PLATFORM_AFTER + 200
                 // we do not want to re-insert CDI beans into CDI, obviously
                 continue;
             }
+            if (service.contracts().contains(ResolvedType.create("io.helidon.security.Security"))) {
```
romain-grecourt: Should this be a constant ?

security/security/src/main/java/io/helidon/security/SecurityService.java
```diff
@@ -14,18 +14,26 @@
...
+import io.helidon.common.config.Config;
+import io.helidon.service.registry.Service;
 
-@AiProvider.ConfigKey("open-ai")
-@AiProvider.ModelConfig(value = OpenAiModerationModel.class)
-@AiProvider.ModelConfig(value = OpenAiEmbeddingModel.class)
-@AiProvider.ModelConfig(value = OpenAiLanguageModel.class)
-interface OpenAiEmbLc4jProvider {
+@Service.Singleton
+class SecurityService implements Supplier<Security> {
```
romain-grecourt: Do we have a pattern for naming these ? Should it be called Factory or Provider instead ?

tomas-langer: It should be called factory, but not that important, as the class is not public, so it cannot be used and we are free to change it.
Also this was changing a few times as I was designing the Service registry API, so some of my implementations were left with original names.

security/providers/abac/src/test/java/io/helidon/security/providers/abac/AbacProviderDeprecatedApiTest.java
```diff
@@ -0,0 +1,154 @@
...
+        ProviderRequest request = Mockito.mock(ProviderRequest.class);
+        when(request.endpointConfig()).thenReturn(ec);
+
+        AuthorizationResponse response = provider.authorize(request);
+
+        assertThat(response.status(), is(SecurityResponse.SecurityStatus.FAILURE));
+        assertThat(response.description(), not(Optional.empty()));
+        response.description()
+                .ifPresent(desc -> {
+                    assertThat(desc, containsString("Attrib1 attribute annotation is not supported"));
```
danielkec, line 67: Didn't it change to fqName?

danielkec: Ah contains .. my bad

security/providers/common/src/main/java/io/helidon/security/providers/common/spi/AnnotationAnalyzer.java
```diff
@@ -43,33 +50,134 @@ default void init(Config config) {
...
+            throw new SecurityException("Cannot analyze endpoint class, as it cannot be obtained", e);
+        }
+    }
+
+    private boolean matchesMethod(Method declaredMethod, List<TypedElementInfo> expectedArguments) {
+        // now validate the parameter types
+        if (declaredMethod.getParameterCount() != expectedArguments.size()) {
+            return false;
+        }
+        // same number of parameters, same name, let's got through parameters
```
danielkec:
```suggestion
        // same number of parameters, same name, let's go through parameters
```

webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityInterceptor.java
```diff
@@ -0,0 +1,614 @@
...
+        - we may want to code generate the security handler for each method, and just look it up based on name
+            (such `as io.helidon.examples.declarative.GreetEndpoint.greet(java.lang.String)`)
+        - for now, we create the handler in memory and cache it, and then invoke it
+         */
+
+        SecurityTracing tracing = SecurityTracing.get(req.context());
+        UriInfo requestedUri = req.requestedUri();
+        SecurityContext securityContext = req.context()
+                .get(SecurityContext.class)
+                .orElseThrow();
```
danielkec: Error message would be nice

PR #10754, opened 2025-10-07
4.x snakeyaml 2.5

### LINKED ISSUE #9805: Helidon MP OpenAPI shows warning in logs
## Environment Details
* Helidon Version: 4.1.4
* Helidon MP
* JDK version: Java(TM) SE Runtime Environment Oracle GraalVM 21.0.2+13.1
* OS: MAC

----------

## Problem Description

We see logs like below when we use Helidon open-api
```java
Server started on http://localhost:7072 (and all other host addresses) in 2819 milliseconds (since JVM startup).
Failed to find field for org.eclipse.microprofile.openapi.models.media.Discriminator.mapping
Failed to find field for org.eclipse.microprofile.openapi.models.media.Discriminator.mapping
Failed to find field for org.eclipse.microprofile.openapi.models.Components.schemas
Failed to find field for org.eclipse.microprofile.openapi.models.Components.schemas
Failed to find field for org.eclipse.microprofile.openapi.models.Components.responses
Failed to find field for org.eclipse.microprofile.openapi.models.Components.responses
Failed to find field for org.eclipse.microprofile.openapi.models.Components.parameters
Failed to find field for org.eclipse.microprofile.openapi.models.Components.parameters
Failed to find field for org.eclipse.microprofile.openapi.models.Components.examples
Failed to find field for org.eclipse.microprofile.openapi.models.Components.examples
Failed to find field for org.eclipse.microprofile.openapi.models.Components.requestBodies
Failed to find field for org.eclipse.microprofile.openapi.models.Components.requestBodies
Failed to find field for org.eclipse.microprofile.openapi.models.Components.headers
Failed to find field for org.eclipse.microprofile.openapi.models.Components.headers
Failed to find field for org.eclipse.microprofile.openapi.models.Components.securitySchemes
Failed to find field for org.eclipse.microprofile.openapi.models.Components.securitySchemes
Failed to find field for org.eclipse.microprofile.openapi.models.Components.links
Failed to find field for org.eclipse.microprofile.openapi.models.Components.links
Failed to find field for org.eclipse.microprofile.openapi.models.Components.callbacks
Failed to find field for org.eclipse.microprofile.openapi.models.Components.callbacks
```

## Steps to reproduce
```java
1. extract the helidon-mp.zip
2. run `mvn clean install && java -jar sample-ms/target/sample-ms.jar`
3. Check the logs.

[helidon-mp.zip](https://github.com/user-attachments/files/18813206/helidon-mp.zip)

### PR DESCRIPTION
**Pull request by tjquinno**
### Description
Resolves #9805

### Release Note

____
Helidon MP no longer triggers warning log messages from SnakeYAML while processing certain OpenAPI documents.
____

###  PR Background

SnakeYAML does a pretty good job matching YAML tags to Java field or method names to map values back and forth. And, it provides ways for "substituting" one property for another. For example, the MP OpenAPI model types support default values but the YAML tag is `default` and the MP types use `defaultValue` and we use a SnakeYAML property substitute for that.

Two other similar but not identical cases in OpenAPI concern the `$ref` YAML property and extension properties which, in YAML, start with `x-`. The MP OpenAPI types represent these using 
```
String getRef()
void setRef(String)

Map<String, Object> getExtensions()
void setExtensions(Map<String, Object>)
```
and related `addExtension` and `removeExtension` methods.

Originally, we received advice from the SnakeYAML team that property substitution was the easiest way to deal with the later two cases as well, so we did. But these are not true property "substitutions" as the "original" property in these cases does not exist in the SnakeYAML-derived property lists. As a result, SnakeYAML would log warnings when we tried to "substitute" for non-existent properties. The "new" properties were added correctly so the mapping between YAML and Java types worked correctly. 

So Helidon would, by default, set the logging level for the SnakeYAML class which did this logging to `SEVERE` using the JUL API so the innocuous but alarming warnings would not appear. (A system property allowed users to choose to preserve the warnings.)

The issue related to this PR resulted from a user using a non-JUL logging implementation in which the warnings were not suppressed.

### Changes overview

Adopts SnakeYAML 2.5 (was 2.4 previously).

The changes in the PR abandon the use of property substitution for these two cases. Instead, Helidon now includes  custom SnakeYAML implementation of a `PropertyUtils` class which knows how to deal with extensions and refs. 

The serialization code also is changed to explicitly deal with the mapping between "$ref" in YAML and "ref" in the Java types.

Two methods on `ExpandedTypeDescription` that used to have to be invoked explicitly are no longer needed and have no bodies, but they are public methods on a public class so we need to keep them in the very unlikely case a user has extended or used `ExpandedTypeDescription` themselves.

The PR also revises some other non-public methods to remove some special processing for ref or extensions now that we are handling those a bit differently.

The code no longer by default suppresses warnings from the SnakeYAML logger although it still respects the user's setting for that property.

The PR includes a new test to make sure that there are no logging messages from SnakeYAML (with the previous suppression turned off).

### Documentation
No impact.

### DIFF
--- a/microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/ExtendedPropertyUtils.java
+++ b/microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/ExtendedPropertyUtils.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.microprofile.openapi;
+
+import java.beans.IntrospectionException;
+import java.beans.PropertyDescriptor;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.microprofile.openapi.models.Extensible;
+import org.eclipse.microprofile.openapi.models.Reference;
+import org.yaml.snakeyaml.introspector.BeanAccess;
+import org.yaml.snakeyaml.introspector.MethodProperty;
+import org.yaml.snakeyaml.introspector.Property;
+import org.yaml.snakeyaml.introspector.PropertyUtils;
+
+/**
+ * Specialized SnakeYAML implementation of {@link org.yaml.snakeyaml.introspector.PropertyUtils} that essentially adds ad hoc
+ * properties to the SnakeYAML type if the underlying type meets certain criteria.
+ */
+class ExtendedPropertyUtils extends PropertyUtils {
+
+    private static final Map<Class<?>, InferredProperty> TYPE_INFO = Map.of(
+            Extensible.class, InferredProperty.create("extensions"),
+            Reference.class, InferredProperty.create("ref"));
+
+    @Override
+    public Set<Property> getProperties(Class<?> type, BeanAccess bAccess) {
+        var result = super.getProperties(type, bAccess);
+        TYPE_INFO.forEach((t, info) -> {
+            if (t.isAssignableFrom(type)) {
+                result.add(info.property(type));
+            }
+        });
+        return result;
+    }
+
+    @Override
+    protected Map<String, Property> getPropertiesMap(Class<?> type, BeanAccess bAccess) {
+        var result = super.getPropertiesMap(type, bAccess);
+        TYPE_INFO.forEach((t, info) -> {
+            if (t.isAssignableFrom(type)) {
+                result.put(info.propertyName, info.property(type));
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by barchetta — APPROVED** [review]
Approve snakeyaml upgrade

File: microprofile/openapi/src/main/java/io/helidon/microprofile/openapi/ExtendedPropertyUtils.java [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+
+    private static final Map<Class<?>, InferredProperty> TYPE_INFO = Map.of(
+            Extensible.class, InferredProperty.create("extensions"),
+            Reference.class, InferredProperty.create("ref"));
+
+    @Override
+    public Set<Property> getProperties(Class<?> type, BeanAccess bAccess) {
+        var result = super.getProperties(type, bAccess);
+        TYPE_INFO.forEach((t, info) -> {
+            if (t.isAssignableFrom(type)) {
```
ljnelson, line 45: Just being careful: this tests to see if `t` is a supertype of `type` (equivalently: if `type` is a subtype of `t`). Oftentimes people get this backwards. Not sure if this is one of those times.

tjquinno: We're good here, but it's good to double-check.

The `t` type variable will be either the interface `Extensible` or the interface `Reference` given the contents of `TYPE_INFO`.  The code is doing what we want:  checking to see if an instance of the type `type` can be converted to `Extensible` (or `Reference` the second time through the loop).

PR #10755, opened 2025-10-08
Helidon Declarative Validation

### LINKED ISSUE #10735: Declarative Validation

tomas-langer:
Resolves #10735 

## Declarative Validation

Validation provides capabilities to validate types and method parameters/return values.
Both features require code-generation.

Type validation code-generation of type validators is triggered by the presence of the `@Validation.Validated` annotation on a type.

Method validation code-generation of interceptors is triggered by the presence of an annotation "meta-annotated"
with `@Validation.Constraint` on a method, its parameter, or type arguments of its parameters or return type.
In addition the `Check.Valid` annotation also triggers code-generation, and can be used to validate against a type validator
mentioned above.

### Declaration

Annotations:
- All annotations from `io.helidon.validation.Check` class to trigger validation using an interceptor
- `@Validation.Validated` - on a type to generate type validator for a type

### Configuration
There is currently no configuration for validation.

### Implementation

Each constraint has a dedicated validator provider service, with (default-weight - 30) weight. 
The providers are annotated with a named annotation that matches the constraint annotation.

This allows our users to override the implementation using their custom services.

The validation process works as follows:
1. an interceptor is generated for any method that has constraint annotations or valid annotation
2. the interceptor obtains all constraint validation instances and type validators needed to validate a call
3. the interceptor validates the call and throws a `ValidationException` if validation fails

The type validation works as follows:
1. a type validator is generated for any type that has a `@Validation.Validated` annotation
2. the type validator obtains all constraint validation instances and type validators needed to validate an instance
3. the type validator validates the type and returns response with possible constraint violations

Important types:
- `Check` - a container class for all constraint annotations
- `Validation` - a container class for validation annotations that are not constraints
- `ValidationException` - throws when validation fails in an interceptor
- `Validator` - programmatic API to validate instances and their properties (only for validated types), can be obtained from service registry
- `ConstraintValidatorProvider` - service registry service that validates a single constraint annotation type
- `ConstraintValidator` - created for each annotated element using the type of the element and the constraint annotation
- `validators` package contains built-in constraint validator providers

Supported concepts:
- any service method annotated with a constraint annotation will be intercepted and validated
- any service method that has parameters with at least one constraint annotation will be intercepted and validated
- for identification of "what to validate", `@Check.Valid` is a constraint annotation
- annotations on type arguments of method parameters and method return type are supported, as long as they are on 
    `Optional`, `Map`, `Collection` - i.e. `List<@Check.Valid String>` will trigger an interceptor and will be validated
- types annotated with `@Validation.Validated` will be validated when used in a method or constructor that uses `@Check.Valid` on the typed parameter
- same rules as for generation of interceptors apply for type validation: constraints are honored, including getters,
    fields, and type arguments
- a user may create a "compound annotation" - an annotation that has one or more constraint annotations, these will be honored as
    if the element was directly annotated with the constraints
- a user may create a custom constraint annotation (annotation meta-annotated with `@Validation.Constraint`), such annotations
    may also be meta-annotated with additional constraints; a custom constraint annotation requires a custom validator provider


declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
```diff
@@ -0,0 +1,480 @@
...
+                .stream()
+                .filter(ElementInfoPredicates::isMethod)
+                .filter(Predicate.not(ElementInfoPredicates::isPrivate))
+                .filter(Predicate.not(ElementInfoPredicates::isStatic))
+                .filter(ElementInfoPredicates::hasNoArgs)
+                .filter(Predicate.not(ElementInfoPredicates::isVoid))
+                .filter(it -> needsWork(constraintAnnotations, it))
+                .forEach(element -> {
+                    String propertyName = element.elementName();
+                    if (propertyName.startsWith("get") && propertyName.length() > 3) {
```
Verdent: Add check that the 4th character is an upper case one, because otherwise methods with non complient getter names would pass this.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
```diff
@@ -0,0 +1,480 @@
...
+                        propertyName = propertyName.substring(3);
+                        if (propertyName.length() == 1) {
+                            propertyName = propertyName.toLowerCase(Locale.ROOT);
+                        } else if (!Character.isUpperCase(propertyName.charAt(1))) {
```
Verdent: I think this is a mistake. You wanted the character at the index 0, right?

Verdent: Also, if you add the check for the 4tch character to be upper case as I wrote before, this check would be redundant :-)

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
```diff
@@ -0,0 +1,480 @@
...
+        type.elementInfo()
+                .stream()
+                .filter(ElementInfoPredicates::isMethod)
+                .filter(Predicate.not(ElementInfoPredicates::isPrivate))
+                .filter(ElementInfoPredicates::hasNoArgs)
+                .filter(Predicate.not(ElementInfoPredicates::isVoid))
+                .filter(it -> needsWork(constraintAnnotations, it))
+                .forEach(element -> {
+                    String propertyName = element.elementName();
+                    if (propertyName.startsWith("get") && propertyName.length() > 3) {
```
Verdent: The same as before.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
tomas-langer: This is not redundant, nor is it a mistake - this is for cases such as `getADSL` - we want to keep all capital...

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
tomas-langer: I have moved this to a separate method to avoid the duplication of code.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
tomas-langer: Updated the implemenation, extracted the method.

tomas-langer:
Examples of generated files:
- `ValidatedType__Validated` - for a validated record
- `ValidatedService__ValidationInterceptor_1` - for an intercepted method that has bot constraints and a validated type

[ValidatedType__Validated.java](https://github.com/user-attachments/files/22954425/ValidatedType__Validated.java)
[ValidatedService__ValidationInterceptor_1.java](https://github.com/user-attachments/files/22954447/ValidatedService__ValidationInterceptor_1.java)

Verdent (approved):

codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ImportOrganizer.java
```diff
@@ -241,6 +238,17 @@ private void resolveFinalImports() {
...
         }
 
+        private Type topLevelType(Type type) {
+            Type lastType = type;
+            Optional<Type> declaringClass = type.declaringClass();
+            while (declaringClass.isPresent()) {
+                lastType = declaringClass.get();
+                declaringClass = lastType.declaringClass();
+
+            }
```
danielkec, line 248:
```suggestion
            if (declaringClass.isPresent()) {
                lastType = declaringClass.get();
                declaringClass = lastType.declaringClass();

            }
```

common/types/src/main/java/io/helidon/common/types/TypeInfo.java
```diff
@@ -1416,21 +1628,21 @@ protected TypeInfoImpl(TypeInfo.BuilderBase<?, ?> builder) {
                 this.typeName = builder.typeName().get();
                 this.rawType = builder.rawType().get();
                 this.declaredType = builder.declaredType().get();
-                this.description = builder.description();
+                this.description = builder.description().map(Function.identity());
```
danielkec, line 1631: What does this do?

service/codegen/src/main/java/module-info.java
```diff
@@ -25,6 +25,7 @@
     requires io.helidon.metadata.hson;
     requires io.helidon.service.metadata;
     requires io.helidon.metadata;
+    requires jdk.jshell;
```
danielkec: What for?

codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ImportOrganizer.java
danielkec: my bad

tomas-langer: As the comment on the call site suggests, we need the top level, not just the parent:
```
//If processed type is inner class, we will be importing (top level) parent class
```

common/types/src/main/java/io/helidon/common/types/TypeInfo.java
tomas-langer: This is code generated.

service/codegen/src/main/java/module-info.java
tomas-langer: IDE glitch, will remove

common/types/src/main/java/io/helidon/common/types/TypeInfo.java
tomas-langer: It is needed if there are generics defined. I do not want to "clean-up" the generated code for `common-types`, as it is easier to check for changes when we need an update.

tomas-langer:
Using Oracle DB lite for test of Helidon data hoping that it shortens the time to run the pipeline, as it was timing out

Issue #10514, opened 2025-08-09
Tracing enhancement in Helidon SE for gRPC


hyder:
Helidon Version: 4.2.5

## Enhancement Description
When using Helidon SE, in order for tracing to work, we need to add an interceptor. This works automatically in Helidon MP.

```Java
        WebServer server = WebServer.builder()
                .config(config.get("server"))
                .routing(Main::routing)
                .addRouting(GrpcRouting.builder()
                        .intercept(GrpcTracingInterceptor.create(tracer, GrpcTracingConfig.create()))
                        .service(new GreetingService(tracer)))
                .addFeature(ObserveFeature.builder()
                        .addObserver(TracingObserver.create(tracer))
                        .build()
                )
                .build()
                .start();
```

We also need an example of how to add tracing in the methods e.g.

```Java
        Span.Builder<?> spanBuilder = tracer.spanBuilder("greet");
        Span.current().ifPresent(s->s.context().asParent(spanBuilder));
        Span span = spanBuilder.start();
        String greeting = greetMessage.getGreeting().isEmpty() ? "greeting": greetMessage.getGreeting() ;

        try(Scope scope = span.activate()) {
            complete(observer, Greet.GreetMessage.newBuilder().setGreeting(greeting).build());
            span.end();
        }
        catch (Exception e) {
            span.end(e);
        }
```

spericas:
Regarding the `GrpcTracingInterceptor` part of the issue, I don't think we currently have an SPI to add gRPC features in the Webserver. That is why that manual step is needed. The HTTP part uses an `HttpFeature` that can add filters to the HTTP routing. gRPC needs a mechanism like that for the WebServer.

Note that we have an SPI in the `GrpcClient` for that based on `GrpcClientServiceProvider`.

PR #10771, opened 2025-10-21
New gRPC server SPI and gRPC tracing refactoring

spericas:
### Description

- New SPI for gRPC server similar to that in gRPC client
- New module for gRPC tracing that uses the SPI
- Issue #10514

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
@@ -28,6 +28,7 @@ import io.helidon.webserver.spi.ProtocolConfigProvider;
 @Prototype.Blueprint
 @Prototype.Configured(root = false, value = GrpcProtocolProvider.CONFIG_NAME)
 @Prototype.Provides(ProtocolConfigProvider.class)
+@Prototype.IncludeDefaultMethods("grpcServices")
 interface GrpcConfigBlueprint extends ProtocolConfig {
 
     /**
@@ -74,5 +75,7 @@ interface GrpcConfigBlueprint extends ProtocolConfig {
     @Option.Singular
     @Option.Configured
     @Option.Provider(value = GrpcServerServiceProvider.class, discoverServices = false)
-    List<GrpcServerService> grpcServices();
+    default List<GrpcServerService> grpcServices() {
+        return List.of();
+    }
 }


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
```diff
@@ -53,12 +57,22 @@ interface GrpcConfigBlueprint extends ProtocolConfig {
...
-     * disabled, no compression will be ever be used by the server even if a
+     * disabled, no compression will ever be used by the server even if a
      * client-compatible compressor is found.
      *
      * @return true if compression is enabled
      */
     @Option.Configured
     @Option.DefaultBoolean(true)
     boolean enableCompression();
+
```
tomas-langer, line 69: This will be a backward incompatible change. This can now be worked around, as you can make it a default method, and then add it as a builder method through annotation.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
```diff
@@ -53,12 +57,22 @@ interface GrpcConfigBlueprint extends ProtocolConfig {
...
+    /**
+     * gRPC server services. These services will not be discovered automatically.
+     *
+     * @return services to use
+     */
+    @Option.Singular
+    @Option.Configured
+    @Option.Provider(value = GrpcServerServiceProvider.class, discoverServices = false)
```
tomas-langer, line 77: Why is `discoverServices` set to `false`? We only want explicitly configured services to be added?

spericas: That's how it was set up for the client SPI, so this would be for consistency.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
spericas: Let me see if I can figure out how to do the second part ...

tomas-langer: `io.helidon.builder.api.Prototype.IncludeDefaultMethods` - you can list names of methods to include.

tomas-langer (approved):

Issue #10752, opened 2025-10-06
ClientRequestBase incorrectly rejects non-lowercase schemes


ljnelson:
See:

https://github.com/helidon-io/helidon/blob/ac8fd4a2b827e81167552529af9b54b147b7b339/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java#L478-L485

And:

https://github.com/helidon-io/helidon/blob/4.3.1/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java#L67

[RFC 3986](https://www.rfc-editor.org/rfc/rfc3986#appendix-A) and [RFC 2234](https://www.rfc-editor.org/rfc/rfc2234#section-6.1) combine to say that `scheme`s may be any case.

PR #10786, opened 2025-10-27
4.x: Web Client Discovery integration

ljnelson:
# Web Client Discovery

## Summary

This PR integrates [Discovery](https://helidon.io/docs/latest/se/discovery) with Helidon [Web Client](https://helidon.io/docs/latest/se/webclient) by introducing a suitable [`WebClientService`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/spi/WebClientService.html) implementation (as required for such integrations).

## Pull Request

The heart of the pull request is the `WebClientDiscovery` interface (an extension of the [`WebClientService` interface](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/spi/WebClientService.html)) and the `DefaultWebClientDiscovery` class that implements it. All other `.java` files are boilerplate or scaffolding related to the required usage of [Helidon Builder](https://helidon.io/docs/latest/se/builder).

The full specification of this integration can be found in the documentation of the `WebClientDiscovery#handle(io.helidon.webclient.spi.WebClientService.Chain, io.helidon.webclient.api.WebClientServiceRequest)` method.

### Documentation

Documentation included in this pull request consists of:

* Full javadoc (class and method specifications)
* Style- and length-compatible additions to `webclient.adoc`
* A new section in `discovery.adoc`

Thanks in particular to @tjquinno, @romain-grecourt and @barchetta for their collective design and documentation help.

## Requirements

* [Discovery is opt-in, not out-out](https://github.com/helidon-io/helidon/pull/10476), so Web Client Discovery must be opt-in as well.

* Discovery itself can be disabled, and `WebClientServiceProvider` implementations can be disabled. When either is disabled, the implementation must continue by using any supplied URIs "as is".

* Discovery must not prevent requests from completing, even in the face of errors, so Web Client Discovery must not either.

* It must be easy to see (in configuration, etc.) which Web Client requests are subject to discovery, and which are not.

* Given a [`WebClientServiceRequest`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/WebClientServiceRequest.html), the implementation must be able to retrieve or infer a [discovery name and a default URI](https://helidon.io/docs/latest/apidocs/io.helidon.discovery/io/helidon/discovery/Discovery.html) from its contents.

* Given a [`UriInfo`](https://helidon.io/docs/latest/apidocs/io.helidon.common.uri/io/helidon/common/uri/UriInfo.html), the implementation must have a deterministic algorithm for combining path information that is part of the discovered URI returned by discovery, path information that is part of the default URI [submitted as part of discovery](https://helidon.io/docs/latest/apidocs/io.helidon.discovery/io/helidon/discovery/Discovery.html#uris(java.lang.String,java.net.URI)), and remaining path information that is part of the `UriInfo` that is not part of either.

## Invariants and Constraints

* **Invariant:** The implementation fundamentally must be a [`WebClientService`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/spi/WebClientService.html) created by a [`WebClientServiceProvider`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/spi/WebClientServiceProvider.html) implementation and therefore must adhere to its requirements as described in its [javadocs](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/spi/WebClientServiceProvider.html).

* **Invariant:** [Helidon Builder API](https://helidon.io/docs/latest/se/builder) must be used.

* **Invariant:** Textual representations of URIs [must not be used as configuration keys](https://github.com/helidon-io/helidon/pull/10786#discussion_r2486475100).

* **Invariant:** [Properties](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/HttpConfigBase.html#properties()) [must not be used](https://github.com/helidon-io/helidon/pull/10786#discussion_r2471171488) as the primary means of configuration.

* **Constraint:** [`ClientRequestBase`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/ClientRequestBase.html) [supports only URIs whose schemes are `http` or `https`](https://github.com/helidon-io/helidon/blob/4.3.1/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java#L478-L485). (See also [issue 10752](https://github.com/helidon-io/helidon/issues/10752).)

* **Constraint:** [`ClientUri`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/ClientUri.html) [does not support opaque URIs](https://github.com/helidon-io/helidon/blob/4.3.1/webclient/api/src/main/java/io/helidon/webclient/api/ClientUri.java#L112-L114).

* **Constraint:** [`ClientUri`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/ClientUri.html) [paths](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/ClientUri.html#path()) are never empty, even when explicitly [set](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/ClientUri.html#path(java.lang.String)) to be empty.

* **Constraint:** [`ClientUri#create()`](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/ClientUri.html#create()) returns a `ClientUri` with a [`scheme` component of `http`](https://github.com/helidon-io/helidon/blob/4.3.1/common/uri/src/main/java/io/helidon/common/uri/UriInfoBlueprint.java#L37), a [`host` component of `localhost`](https://github.com/helidon-io/helidon/blob/4.3.1/common/uri/src/main/java/io/helidon/common/uri/UriInfoBlueprint.java#L45), a [`port` component of `80`](https://github.com/helidon-io/helidon/blob/4.3.1/common/uri/src/main/java/io/helidon/common/uri/UriBuilderSupport.java#L37-L39) and a [`path` of `/`](https://github.com/helidon-io/helidon/blob/4.3.1/common/uri/src/main/java/io/helidon/common/uri/UriInfoBlueprint.java#L73).

## Sample Javadoc

<details><summary><h3><code>io.helidon.webclient.discovery.WebClientDiscovery</code></h3></summary>
<p>Javadoc HTML pages are not capturable by full-screen screenshot tools. Screenshots are accordingly paged below.</p>

<h4>Page 1</h4>
<a href="https://github.com/user-attachments/assets/a48997fe-b889-40ca-bcf7-e29f0d4ab280">io.helidon.webclient.discovery.WebClientDiscovery Page 1</a>

<h4>Page 2</h4>
<a href="https://github.com/user-attachments/assets/a5bde806-3229-40c6-ae80-c64359086dea">io.helidon.webclient.discovery.WebClientDiscovery Page 2</a>

<h4>Page 3</h4>
<a href="https://github.com/user-attachments/assets/23f82f02-3452-42d3-a468-4cd56ad433f2">io.helidon.webclient.discovery.WebClientDiscovery Page 3</a>

<h4>Page 4</h4>
<a href="https://github.com/user-attachments/assets/8775e492-b3af-43be-86f8-721866778cc3">io.helidon.webclient.discovery.WebClientDiscovery Page 4</a>

</details>


docs/src/main/asciidoc/se/webclient.adoc
```diff
@@ -407,29 +407,56 @@ include::{sourcedir}/se/WebClientSnippets.java[tag=snippet_8,indent=0]
...
 === Adding Service to WebClient
 
-WebClient currently supports 3 built-in services namely `metrics`, `tracing` and `security`.
+WebClient currently supports four built-in services, namely
+xref:discovery.adoc#_web_client_discovery_integration[`discovery`], `metrics`, `tracing` and `security`.
 
 ==== Enabling the service
 
-In order for a service to function, their dependency needs to be added in the application's pom.xml. Below are examples on how to enable the built-in services:
+In order for a service to function, their dependency needs to be added in the application's `pom.xml`. Below are examples on how to enable the built-in services:
```
tjquinno: As long as we're editing this line, maybe also change "their dependency" to "its dependency" or "its dependencies" to match the antecedent "a service".

webclient/discovery/src/main/java/io/helidon/webclient/discovery/DefaultWebClientDiscovery.java
```diff
@@ -0,0 +1,336 @@
...
+
+        static final Pattern WHITESPACE_AND_COMMAS_PATTERN = Pattern.compile("[\\s,]+");
+
+        DiscoveryRequest {
+            requireNonNull(discoveryName, "discoveryName");
+            requireNonNull(defaultUri, "defaultUri");
+            requireNonNull(extraPath, "extraPath");
+        }
+
+        static Optional<DiscoveryRequest> of(Map<String, String> properties, UriInfo uriInfo) {
```
tjquinno: With the necessarily-intricate logic here, users might be puzzled if they mistype a property setting in some way and the result is not what they expect. 

I'm not exactly sure what might be most useful but some debug-level logging messages that explain why the code is doing what it's doing could be helpful here.

ljnelson: Excellent point.

ljnelson: Addressed in https://github.com/helidon-io/helidon/pull/10786/commits/1e5d5661d3cff09e0a74238b9c186787caf25e8e

docs/src/main/asciidoc/se/webclient.adoc
ljnelson: Addressed in https://github.com/helidon-io/helidon/pull/10786/commits/1e5d5661d3cff09e0a74238b9c186787caf25e8e

webclient/discovery/src/main/java/io/helidon/webclient/discovery/DefaultWebClientDiscovery.java
```diff
@@ -0,0 +1,345 @@
...
+    @Override // WebClientDiscovery (RuntimeType.Api<WebClientDiscoveryConfig>)
+    public WebClientDiscoveryConfig prototype() {
+        return this.prototype;
+    }
+
+    @Override // WebClientDiscovery (WebClientService)
+    public WebClientServiceResponse handle(Chain chain, WebClientServiceRequest request) {
+        ClientUri clientUri = request.uri();
+        if (LOGGER.isLoggable(DEBUG)) {
+            LOGGER.log(DEBUG, "Initial ClientUri: " + clientUri);
```
tomas-langer: Two debug statements may end up disconnected (i.e. multiple lines may exist between the first and second debug statement). Unless there is any kind of correlation id that can be used to map them, the connection would be lost.
Either log it as a single statement, or add correlation information.

webclient/discovery/src/main/java/io/helidon/webclient/discovery/WebClientDiscoveryProvider.java
```diff
@@ -0,0 +1,74 @@
...
+     * @exception NullPointerException if {@code config} is {@code null}
+     * @see WebClientDiscovery#handle(io.helidon.webclient.spi.WebClientService.Chain,
+     * io.helidon.webclient.api.WebClientServiceRequest)
+     * @see WebClientDiscoveryConfig#builder()
+     */
+    @Override // WebClientServiceProvider
+    @SuppressWarnings("removal") // Config is deprecated for removal, but our contract (WebClientServiceProvider) mandates its use
+    public WebClientDiscovery create(Config config, String name) {
+        return WebClientDiscoveryConfig.builder()
+            .config(config)
```
tomas-langer, line 70: you should set the name on the builder as well, so the resulting service is correctly named.

webclient/discovery/src/main/java/module-info.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Provides {@linkplain io.helidon.discovery.Discovery discovery} support for Helidon {@linkplain
+ * io.helidon.webclient.api.WebClient WebClient}.
+ *
+ * @see io.helidon.webclient.discovery.WebClientDiscoveryProvider
+ */
+module io.helidon.webclient.discovery {
+
+    requires io.helidon.webclient.api;
```
tomas-langer: APIs exposed through public types should be `requires transitive`

webclient/discovery/src/test/java/io/helidon/webclient/discovery/TestURI.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.isEmptyString;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.sameInstance;
+
+class TestURI {
```
tomas-langer: Invalid test class name - all test classes should match pattern `.*Test`

tomas-langer: This comment is valid for all types in this test package.

webclient/discovery/src/test/java/io/helidon/webclient/discovery/URITest.java
```diff
@@ -0,0 +1,106 @@
...
+
+        // "if the path ["/"] of this URI [u1] is not a prefix of the path [""] of the given URI [u0], then the given
+        // URI [u0] is returned."
+        u2 = u1.relativize(u0);
+        assertThat(u2, sameInstance(u0));
+    }
+
+    @Test
+    void testRelativizeEquals() {
+        URI u0 = URI.create("http://example.com");
```
tomas-langer: Can you explain why you are testing `URI` type that is part of Java?

webclient/discovery/src/test/java/io/helidon/webclient/discovery/UriEncodingTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+class TestUriEncoding {
+
+    private TestUriEncoding() {
+        super();
+    }
+
+    @Test
+    void testDecode() {
+        ClientUri c = ClientUri.create();
```
tomas-langer: This test does not seem to belong here at all - this should be tested as part of `io.helidon.webclient.api` module.
If this test is missing, please add it. If it exists, please remove this one.

webclient/discovery/src/test/java/io/helidon/webclient/discovery/WebClientTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+class TestWebClient {
+
+    private TestWebClient() {
+        super();
+    }
+
+    @Test
+    void showThatWebClientRequiresLowercaseHttpAndHttpsSchemesOnly() {
+        WebClient c = WebClient.builder().build();
```
tomas-langer: Again not sure what you are testing here related to discovery - just the fact that webclient does not support `otherscheme`?

webclient/discovery/src/test/resources/application.yaml
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+TestWebClientConfig-testPropertiesFromConfig:
+  base-uri: http://prod.example.com/foo
+  properties:
```
tomas-langer: The reason we have webclient service is to be able to configure them with the webclient in a manner that makes it clear to what the configuration belongs. This approach you use is not doing that.

Please use the same approach as with other services (both in webserver and webclient) 
I would expect the configuration to be similar to:

```yaml
webclient-config-key:
  base-uri: http://prod.example.com/foo
  services:
    discovery:
      # some structure here specific to discovery, i.e. 
      "TEST": 
          prefix-uri: "http://test.example.com"
          name: "TEST"
```

This configuration should then be processed by the service, and used when processing a request.

webclient/discovery/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+        <groupId>io.helidon.webclient</groupId>
+        <artifactId>helidon-webclient-project</artifactId>
+        <version>4.3.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>helidon-webclient-discovery</artifactId>
+    <name>Helidon WebClient Discovery</name>
+
+    <properties>
+        <java.util.logging.config.file>src/test/resources/logging.properties</java.util.logging.config.file>
```
tomas-langer, line 31: All new modules should have `<javadoc.fail-on-warnings>true</javadoc.fail-on-warnings>` property configured.

webclient/discovery/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
```
tomas-langer, line 103: All new modules should have this plugin configured:

```xml
            <plugin>
                <artifactId>maven-dependency-plugin</artifactId>
                <executions>
                    <execution>
                        <id>check-dependencies</id>
                        <phase>verify</phase>
                    </execution>
                </executions>
            </plugin>
```

webclient/discovery/src/test/java/io/helidon/webclient/discovery/WebClientTest.java
ljnelson: In general, I left in tests I used to explore contracts that were not immediately apparent to me. This is, as you say, to "prove" that any strategy using, for example, a `discovery:` scheme or anything like that wouldn't work. Should I remove such exploratory tests?

webclient/discovery/src/test/resources/application.yaml
ljnelson: So don't use client properties, but ensure this information is part of the configuration of the service itself? Just want to be clear. (The client properties approach allows a user to supply or alter them in a one-off fashion without having to use the programmatic discovery API. That may or may not be important.)

webclient/discovery/src/test/java/io/helidon/webclient/discovery/WebClientTest.java
tomas-langer: It seems to me that the tests are misleading when you try to understand what the module does. 
Maybe this would be good to keep in some other repository, where we could keep such "reference" tests when somebody else needs to understand the behavior of URI and webclient. 
So I think these should be removed from here.

webclient/discovery/src/test/resources/application.yaml
tomas-langer: That seems the right way here - we are configuring the discovery service (or maybe I do not understand what you are doing here).

webclient/discovery/src/test/java/io/helidon/webclient/discovery/WebClientTest.java
ljnelson: Shall I move this test in particular to the `webclient` module? It seems important to test _somewhere_ for an (error?) condition that is [not explicitly documented](https://helidon.io/docs/latest/apidocs/io.helidon.webclient.api/io/helidon/webclient/api/HttpClient.html#get(java.lang.String))

webclient/discovery/src/test/resources/application.yaml
ljnelson: OK. I will redo the PR and documentation and specification to include this change, and will follow the suggested YAML above (or something similar to it).

Would a structure like this be OK?
```
webclient-config-key:
  services:
    discovery:
      names:
        "https://prod.example.com:443": "prod"
```
Or the inverse:
```
webclient-config-key:
  services:
    discovery:
      prefix-uris:
        prod: "https://prod.example.com:443/"
```

webclient/discovery/src/main/java/module-info.java
ljnelson: OK. Is there a reason [the compiler is not configured to expose these warnings](https://github.com/helidon-io/helidon/blob/4.3.1/pom.xml#L277-L281)? In the meantime I'll make sure `-Xlint:all` is on for this Maven module anyway

webclient/discovery/src/test/java/io/helidon/webclient/discovery/URITest.java
ljnelson: This was another exploratory test. It led me, among other useful things, to find [JDK-6523089](https://bugs.openjdk.org/browse/JDK-6523089). Shall I remove it?

webclient/discovery/src/test/java/io/helidon/webclient/discovery/UriEncodingTest.java
ljnelson: This was another exploratory test I wrote while exploring different strategies for making discovery "visible" in URIs that helped me (and, I hoped, other readers) to understand that URI decoding as performed by the `UriEncoding` class surprisingly turns pluses (`+`) into spaces (` `), which is a feature of [`application/x-www-form-urlencoded` parsing](https://url.spec.whatwg.org/#urlencoded-parsing), not [percent encoding/decoding](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1). Shall I (a) move it or (b) remove it?

webclient/discovery/src/main/java/io/helidon/webclient/discovery/WebClientDiscoveryProvider.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/3cb5be0e56a9957a6b7348c2bd6430fc552145d8

webclient/discovery/src/main/java/module-info.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/3cb5be0e56a9957a6b7348c2bd6430fc552145d8

webclient/discovery/src/test/java/io/helidon/webclient/discovery/TestURI.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/3cb5be0e56a9957a6b7348c2bd6430fc552145d8

webclient/discovery/pom.xml
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/3cb5be0e56a9957a6b7348c2bd6430fc552145d8.

webclient/discovery/pom.xml
ljnelson: Without further configuration, this fails; resolved successfully in https://github.com/helidon-io/helidon/pull/10786/commits/3cb5be0e56a9957a6b7348c2bd6430fc552145d8

webclient/discovery/src/test/resources/application.yaml
ljnelson: One last question: my original approach permits one-off discovery usage triggered by client properties. I'm not saying it's right, just that it's possible, which is why I went this route. For example:
```
var c = webclientbuilder.build(); // with the discovery service on/enabled/present
// For this particular request, set up discovery a particular way:
c.property("helidon-discovery-TEST-prefix-uri", "http://test.example.com/test")
  .get(someUriSomeonePassedInThatMayOrMayNotMatch);
```
(Typed without proofing but you get the idea.)

To be very clear: we want to disable this use case, correct?

webclient/discovery/src/main/java/io/helidon/webclient/discovery/DefaultWebClientDiscovery.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/commit/3cb5be0e56a9957a6b7348c2bd6430fc552145d8

webclient/discovery/src/test/resources/application.yaml
tomas-langer: I am not against augmenting the configuration via properties, but it should not be the main means of configuration.
I would start without it, with the option to add it in the future, once we know how people intend to use this feature.

tomas-langer: Regarding the first question - I do not understand the configuration use case enough to know the correct answer to this. One thing I would avoid is to have URI as a key anywhere, as that makes it really hard to override using system properties/environment variables.

I.e. `client.services.discovery.prod.uri=https://prod.example.com.433` is OK, but `client.services.discovery.names.https://prod.example.com:443=prod` is not OK

webclient/discovery/src/test/java/io/helidon/webclient/discovery/WebClientTest.java
tomas-langer: If there is something that is not tested, of course add it to the `webclient` module (either directly to it, or to the `tests` submodule depending on what it does)

webclient/discovery/src/test/java/io/helidon/webclient/discovery/UriEncodingTest.java
tomas-langer: Move it to webclient API tests please.

webclient/discovery/src/test/java/io/helidon/webclient/discovery/URITest.java
tomas-langer: yes, please remove tests related to JDK classes.

webclient/discovery/src/test/resources/application.yaml
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/94c1613005ba25167e31dd262561ef2eedaa63b3.

webclient/discovery/src/test/java/io/helidon/webclient/discovery/WebClientTest.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/94c1613005ba25167e31dd262561ef2eedaa63b3.

webclient/discovery/src/test/java/io/helidon/webclient/discovery/URITest.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/94c1613005ba25167e31dd262561ef2eedaa63b3.

webclient/discovery/src/test/java/io/helidon/webclient/discovery/UriEncodingTest.java
ljnelson: Resolved in https://github.com/helidon-io/helidon/pull/10786/commits/94c1613005ba25167e31dd262561ef2eedaa63b3.

PR #10791, opened 2025-10-30
Allow and document how users control SE tracing for OpenTelemetry using Helidon config under `tracing`

### LINKED ISSUE #10790: 4.x - Allow users to use `tracing` config with the Helidon OpenTelemetry tracing provider
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Helidon now supports programmatic or config settings to control OpenTelemetry, and we have added a telemetry page describing this.

But we need to update the _tracing_ code and doc pages and/or guides so users can use the `tracing` config settings to control the OpenTelemetry tracing behavior.

tjquinno:
### Description
Resolves #10790 

### Release Note
____
Users incorporating the Helidon tracing provider for OpenTelemetry--by depending on the `helidon-tracing-providers-opentelemetry` component--can now control many aspects of OpenTelemetry tracing behavior using the Helidon `tracing` config section and, programmatically, using the corresponding `OpenTelemetryTracer.Builder`. Note that users can more fully control OpenTelemetry using the `telemetry` config section and builder, but adding support for the `tracing` config section helps users move from other tracing providers (especially Jaeger) to OpenTelemetry by supporting very similar configuration files.

As noted in the accompanying doc update, if you specify both `telemetry` and `tracing` config Helidon uses the `telemetry` settings.
____

### PR Background
Our SE tracing page contains details about using and configuring Jaeger and Zipkin tracing but not OpenTelemetry. Our code did not formerly support using the `tracing` config block or a tracing builder if the project uses the Helidon OpenTelemetry tracing provider.

This PR allows users to use the `tracing` config block to control the OpenTelemetry tracing provider, and/or programmatically use a builder, and adds relevant documentation.

#### Docs
A new OTel tracing config page, and updates to the tracing doc page.

#### `microprofile/cdi`
Use the same workaround as with security because with this PR we have both a service registry supplier and a CDI producer for `OpenTelemetry`.

#### Jaeger tracing provider
Some miscellaneous clean-up and removal of explicitly setting the global OpenTelemetry instance in the Jaeger builder because that is now handled elsewhere.

#### OpenTelemetry tracing provider

* Config blueprint and blueprint support types. 
* To conform to the older tracing pattern the new `OpenTelemetryTracerBuilder` extends the builder generated from the blueprint and also implements `TracerBuilder`.
* Two enums for context propagation type and OTLP exporter protocol type referenced by the config.
* A `Supplier<OpenTelemetry>` for publicizing the `OpenTelemetry` instance created based on the tracing settings. Note that this service factory has lower weight than the one in the previously-added telemetry component so OpenTelemetry  settings under `telemetry` take precedence over those under `tracing`.

#### Tracing component
* Add `ExtendedTracerConfigBlueprint`. These are settings common to the Jaeger and OpenTelemetry tracing set-up. Eventually we should redo the Jaeger component to use the new builder/config support and its blueprint should also extend `ExtendedTracerConfig`. That is not part of this PR and purposefully excluded to try to control the scope of this PR.
* Adds a new `config` method to `TracerBuilder` accepting `Config`, joining the method that accepts common config.

Note also that this PR does not change how users configure OTel behavior using MP. That is still done according to the MP Telemetry spec and not using our `telemetry` settings.

### Documentation
Includes doc updates.

### DIFF
--- a/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ServiceRegistryExtension.java
+++ b/microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ServiceRegistryExtension.java
@@ -180,7 +180,7 @@ public class ServiceRegistryExtension implements Extension {
                 continue;
             }
             if (service.contracts().contains(SECURITY)) {
-                // workaround - we must use `SecurityCdiExtension` , as it has public methods with security builder...
+                // workaround - we must use `SecurityCdiExtension`, as it has public methods with security builder...
                 continue;
             }
             if (service.contracts().contains(TRACER)) {


docs/src/main/asciidoc/includes/tracing/tracer-otel.adoc
```diff
@@ -0,0 +1,107 @@
...
+// end::otel-dependency[]
+
+// tag::otel-configuration[]
+
+== Configuring OpenTelemetry Tracing
+Because OpenTelemetry comprises several signals, of which tracing is only one, you configure OpenTelemetry tracing differently from other implementations of tracing.
+
+For complete information on configuring OpenTelemetry using Helidon {flavor-uc}, see xref:{rootdir}/{flavor-lc}/open-telemetry.adoc[the Helidon {flavor-uc} OpenTelemetry documentation].
+
+The following example shows only some of the settings you can use to control OpenTelemetry tracing.
```
rhomast: Where do you find the rest of the settings you can use?

docs/src/main/asciidoc/includes/tracing/tracer-otel.adoc
```diff
@@ -0,0 +1,107 @@
...
+telemetry.signals.tracing.exporters.0.type=otlp <3>
+----
+endif::mp-flavor[]
+<1> Specifies the OpenTelemetry service name.
+<2> Selects the `simple` span processor.
+<3> Selects which span exporter to use.
+
+The `simple` processor type sends each span to the backend as soon as it ends, rather than grouping multiple spans into batches, and so is appropriate for testing and demos but not for production deployments.
+
+The `otlp` exporter type matches the exporter dependency example shown above. If you try to configure a particular exporter type without added the corresponding dependency to your project, the server fails during startup.
```
rhomast:
```suggestion
The `otlp` exporter type matches the exporter dependency example shown above. If you try to configure a particular exporter type without adding the corresponding dependency to your project, the server fails during startup.
```

docs/src/main/asciidoc/includes/tracing/tracer-otel.adoc
tjquinno: The link in the preceding sentence describes them all. I've rewritten the two sentences and brought them into one paragraph to clarify.

docs/src/main/asciidoc/includes/tracing/tracer-otel.adoc
tjquinno: Fixed.

tjquinno:
After a discussion with other team members I'm enlarging the scope of the PR a bit beyond just doc. So I've marked it as a draft.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerBlueprintSupport.java
```diff
@@ -0,0 +1,298 @@
...
+
+        /**
+         * Adds a string-valued tag.
+         *
+         * @param builder builder
+         * @param name    tag name
+         * @param value   tag value
+         */
+        @Prototype.BuilderMethod
+        static void addTracerTag(OpenTelemetryTracerConfig.BuilderBase<?, ?> builder, String name, String value) {
```
tomas-langer, line 217: This is not needed - you can change the name of the singular setter and ignore the prefix.
See `io.helidon.webserver.jsonrpc.JsonRpcHandlersBlueprint#handlersMap`

tomas-langer: There seem that other methods follow the same pattern.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerBuilder.java
```diff
@@ -0,0 +1,71 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.tracing.providers.opentelemetry;
+
+import io.helidon.tracing.TracerBuilder;
+
+class OpenTelemetryTracerBuilder
```
tomas-langer, line 21: This class is unexpected. This seems like a subset of something a prototype would contain.

tracing/providers/opentelemetry/src/main/java/module-info.java
```diff
@@ -25,11 +25,16 @@
...
+    requires io.opentelemetry.exporter.otlp;
     requires io.opentelemetry.extension.trace.propagation;
+    requires io.opentelemetry.sdk;
+    requires io.opentelemetry.sdk.common;
+    requires io.opentelemetry.sdk.trace;
     requires io.opentelemetry.semconv;
 
     requires transitive io.helidon.tracing;
     requires transitive io.opentelemetry.api;
+    requires java.logging;
```
tomas-langer: This is wrong (probably added by IDE?).
We use `System.getLogger`, not JUL

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
```diff
@@ -0,0 +1,215 @@
...
+     */
+    @Option.Configured("service")
+    String serviceName();
+
+    /**
+     * URI for the collector to which to send tracing data.
+     *
+     * @return tracing collector URI
+     */
+    Optional<URI> collectorUri();
```
tomas-langer, line 48: This can be `@Configured` as well, URI is supported

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
```diff
@@ -0,0 +1,215 @@
...
+    @Option.Configured("path")
+    Optional<String> collectorPath();
+
+    /**
+     * Tracer-level tags with {@code String} values added to all reported spans.
+     *
+     * @return tracer-level string-valued tags
+     */
+    @Option.Configured("tags")
+    @Option.Singular("tracerTag")
```
tomas-langer: No need to define singular name if it is just removing the `s` from its plural form

tracing/tracing/src/main/java/io/helidon/tracing/SamplerType.java
```diff
@@ -14,7 +14,21 @@
...
  */
 
+package io.helidon.tracing;
+
 /**
- * Provides integration with a Mock model.
+ * Sampler types (abbreviated list).
  */
-package io.helidon.integrations.langchain4j.providers.mock;
+public enum SamplerType {
```
tomas-langer, line 22: I am not sure about this - we are limiting this to an enum, and there may be a new sampler type in the future in an implementation, that would force us  to do a new release to support it. Similar for the enum below - is this part of a specification? If not, we should not use an enum IMHO.

tracing/providers/opentelemetry/src/main/java/module-info.java
tjquinno: Ugh. yes. fixed.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerBlueprintSupport.java
tjquinno: The interface `TracerBuilder` (which the builder needs to implement) declares these methods with _primitive_ types, but the generated methods use the corresponding _class_ types so the signatures do not match.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/OpenTelemetryTracerBuilder.java
tjquinno: In the tracing subsystem, the builders must implement the separate legacy `TracerBuilder` interface. 

From our earlier chat there is no way to indicate to the code generator that the generated builder should implement an additional interface.

Or did you have some other change in mind?

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
tjquinno: Changed.

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
tjquinno: Left-over from an experiment. Removed.

tracing/tracing/src/main/java/io/helidon/tracing/SamplerType.java
tjquinno: Understood, but here's why the PR uses these two.

A motivating factor for adding tracing-based config support for OpenTelemetry tracing is consistency with other tracing provider configuration (esp. Jaeger) and, as a nice side-benefit, so users have an easier transition from Jaeger (for example) to OpenTelemetry while preserving (or mostly preserving) their existing config files or builder code. 

These two sampler types--`CONSTANT` AND `RATIO`--are not actually themselves OpenTelemetry sampler types but rather the ones supported by the Jaeger tracer builder. The Jaeger tracer builder--like the new code in the PR--also converts those values to the corresponding OpenTelemetry ones in setting up OTel using the OTel API internally.

The doc update in this PR points out to users that they can use the `telemetry` settings for much more complete control over OpenTelemetry's behavior (as compared to what's available via the `tracing` settings), and there a user _can_ choose any sampler OpenTelemetry supports.

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ServiceRegistryExtension.java
```diff
@@ -179,13 +180,17 @@ void registryToCdi(@Observes @Priority(Interceptor.Priority.PLATFORM_AFTER + 200
                 continue;
             }
             if (service.contracts().contains(SECURITY)) {
-                // workaround - we must use `SecurityCdiExtension`, as it has public methods with security builder...
+                // workaround - we must use `SecurityCdiExtension` , as it has public methods with security builder...
```
tomas-langer: not sure why this line was changed, the space does not belong there

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonTracingBasedOpenTelemetryServiceFactory.java
```diff
@@ -0,0 +1,82 @@
...
+import io.helidon.config.Config;
+import io.helidon.service.registry.Service;
+import io.helidon.tracing.Tracer;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.OpenTelemetry;
+
+@Service.Singleton
+@Service.RunLevel(Service.RunLevel.STARTUP)
+@Weight(Weighted.DEFAULT_WEIGHT - 100)
```
tomas-langer: This is not a good weight - you set it to 0, it should be a positive number (and never 0)

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
```diff
@@ -0,0 +1,216 @@
...
+    SamplerType samplerType();
+
+    /**
+     * Parameter value used by the selected sampler; interpretation depends on the sampler type..
+     *
+     * @return sampler parameter value
+     */
+    @Option.Configured
+    @Option.DefaultDouble(1.0d)
+    Double samplerParam();
```
tomas-langer: you can use primitive type here, as we have we do not support `null` as a value

microprofile/cdi/src/main/java/io/helidon/microprofile/cdi/ServiceRegistryExtension.java
tjquinno: Accidental. Fixed.

tracing/providers/opentelemetry/src/main/java/io/helidon/tracing/providers/opentelemetry/HelidonTracingBasedOpenTelemetryServiceFactory.java
tjquinno: Fixed.

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
tjquinno: Changed to `double`

Issue #10777, opened 2025-10-23
Severe throughput degradation when Helidon gRPC server runs across network boundary (Docker/remote)


AlfredoG87:
# ⚠️ Helidon gRPC Server Severe Throughput Regression Across Docker/Network Boundary

## Environment

| Component | Details |
|------------|----------|
| **Helidon Version** | 4.3.1 |
| **Helidon Flavor** | SE (gRPC Server & Client) |
| **JDK** | Eclipse Temurin 21.0.2+13 (LTS) |
| **OS** | macOS 14.x (Apple Silicon) |
| **Container Runtime** | Docker Desktop 4.33.1 / Engine 27.3 |
| **Hardware** | MacBook Pro M1 |

---

## Description

When the **Helidon gRPC server** runs locally in the JVM, throughput matches Netty (≈600–1 900 MB/s).  
When the same Helidon server runs inside Docker or across a real network, throughput drops by **10–20×**, while the Netty server remains stable.

This regression is consistently reproducible and isolated to the Helidon HTTP/2 gRPC server.

---

## Reproduction Repository

**Repo:** [https://github.com/AlfredoG87/helidon-vs-netty-benchmark-capacity](https://github.com/AlfredoG87/helidon-vs-netty-benchmark-capacity)

Includes identical implementations of Helidon and Netty servers and clients, integration tests, and Docker setup.

---

## Steps to Reproduce

1. Clone and enter repo:
   ```
   git clone https://github.com/AlfredoG87/helidon-vs-netty-benchmark-capacity
   cd helidon-vs-netty-benchmark-capacity
   ```

2. Run local benchmark (no Docker):
   ```
   ./gradlew test --tests org.example.benchmark.ThroughputIntegrationTest
   ```

3. Build Docker images:
   ```
   ./gradlew distTar
   ./docker/build.sh 0.0.1-SNAPSHOT netty
   ./docker/build.sh 0.0.1-SNAPSHOT helidon
   ```

4. Start servers (Netty on 50051, Helidon on 50052):
   ```
   docker compose -f docker/docker-compose.yaml up --build
   ```

5. Run clients targeting Docker servers:
   ```
   ./gradlew run --args="client netty http://localhost:50052/ 1000 1024"
   ./gradlew run --args="client helidon http://localhost:50052/ 1000 1024"
   ```

---

### Local JVM Servers

| Server | Client | Payload KB | Messages | Duration (s) | MB/s |
[...]

PR #10797, opened 2025-11-03
gRPC improvements and a bug fix for large data uploads

spericas:
### Description

This PR introduces a bug fix and some optimizations for handling large method params in gRPC calls. 

1. Bug fix was discovered using certain clients that can upload very small data chunks, and can cause the gRPC server handler to read more bytes than those available. The fix now waits for the next chunk (or chunks) before processing.
2. It also includes some optimizations to avoid creating some reading and writing buffers, as well as a limit on how much a reading buffer can grow that is adjustable via config.
3. There's a new test that verifies uploads of large byte arrays.
4. Related to issue #10777. Although there's additional TCP tuning required for that issue.

### Documentation

None.

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -347,14 +347,12 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
     static Http2StreamState nextStreamState(Http2StreamState currentStreamState,
                                             Http2StreamState desiredStreamState) {
         return switch (desiredStreamState) {
-            case HALF_CLOSED_LOCAL ->
-                    currentStreamState == Http2StreamState.HALF_CLOSED_REMOTE
-                            ? Http2StreamState.CLOSED
-                            : Http2StreamState.HALF_CLOSED_LOCAL;
-            case HALF_CLOSED_REMOTE ->
-                    currentStreamState == Http2StreamState.HALF_CLOSED_LOCAL
-                            ? Http2StreamState.CLOSED
-                            : Http2StreamState.HALF_CLOSED_REMOTE;
+            case HALF_CLOSED_LOCAL -> currentStreamState == Http2StreamState.HALF_CLOSED_REMOTE
+                    ? Http2StreamState.CLOSED
+                    : Http2StreamState.HALF_CLOSED_LOCAL;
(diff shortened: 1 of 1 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -185,35 +191,61 @@ public void rstStream(Http2RstStream rstStream) {
...
     public void data(Http2FrameHeader header, BufferData data) {
         try {
             boolean isCompressed = false;
 
-            while (data.available() > 0) {
-                // start of new chunk?
+            // check for any unread data received before
+            BufferData newData;
+            if (unreadBufferData != null) {
+                newData = BufferData.create(unreadBufferData, data);
```
tjquinno, line 209: I don't know how `BufferData` works, in particular can the `data` parameter passed to this method get additional data over time after this method has been invoked? 

I ask because, even though `BufferData.create` seems pretty efficient I wondered if it's worth further optimizing the case where `data` has no data available and then just assigning `newdata = unreadBufferData` in that case. I'm not arguing for or against because I don't know! Just thinking out loud and wondering about how often that might actually happen.

tjquinno (dismissed):
LGTM. One _possible_ performance optimization noted but I don't know if it'd be worth it. Change if it makes sense, otherwise good to go.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
spericas: The condition `unreadBufferData != null` is very rare. In fact, it is the crux of the bug fix and was never uncovered until this perf testing. Also `data.available > 0` is almost always true when entering the method (i.e. before hitting the while loop). In fact, the issue was uncovered with a `data.available() == 1` entering the method, which seems to happen from time to time with certain clients.

AlfredoG87:
@spericas, am glad to see this merged 💯 
Great turn around time, I understand that there are still some optimizations to really be able to close https://github.com/helidon-io/helidon/issues/10777

however this PR would allow us some improvements right? can't wait to test it out. any `patch` release planning where this can be cherry-picked? what is the plan for the release? or are we waiting for a complete resolution before release?

spericas:
@AlfredoG87 Improvements are only for Helidon server (not client yet, so just use Netty). There's another PR #10800 that is relevant as well. Both "should" be part of the next release 4.3.2.

PR #10816, opened 2025-11-06
4.x: Set all contracts when calling Services.set

### LINKED ISSUE #10447: 4.x: Deprecated Config.global(Config) method does not have a full replacement
## Environment Details
* Helidon Version: 4.2.x
* Helidon SE

`io.helidon.config.Config.config(Config)` method has the following deprecation comment:
```
use {@link io.helidon.service.registry.Services#set(Class, Object[])} to register a static instance for the
     *      global service registry; when using a custom service registry instance, set is on the registry configuration builder
```

1. This does not work in 4.2.4 - `io.helidon.config.Config` is not a valid contract (this should be fixed in 4.2.5)
2. The recommended approach would only register a config instance for `io.helidon.config.Config`, but not for `io.helidon.common.config.Config`; the latter is used by internal components of Helidon to retrieve configuration from service registry, and as such it should use the same instance

We must fix this and validate that when `Services.set(io.helidon.config.Config.class, configInstance)` is called, it would be available both as common Config instance and as config Config instance...

Related to #10416

tomas-langer:
resolves #10447

Code generate a contrat map (what inherited contracts a contract implements)
Use the information when `Services.set` or `Services.add` is called.

Also removes the workaround we have for `Config` right now.

### DIFF
--- a/service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
+++ b/service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
@@ -486,18 +486,7 @@ class CoreServiceRegistry implements ServiceRegistry, Scopes {
             stateWriteLock.unlock();
         }
     }
-/*
-TODO add and add qualified also must use all contracts
-            ResolvedType contractType = ResolvedType.create(contract);
-            Set<ResolvedType> actualContracts = contractMap.get(contractType);
-            if (actualContracts == null) {
-                actualContracts = Set.of(contractType);
-            }
-            for (ResolvedType actualContract : actualContracts) {
-                // set for each type that is valid for this contract
-                doSet(actualContract, instances);
-            }
- */
+
     <T> void add(Class<T> contract, double weight, T instance) {
         if (!allowLateBinding) {
             throw new ServiceRegistryException("This service registry instance does not support late binding, as it was "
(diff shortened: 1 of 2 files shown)


service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -1463,41 +1553,34 @@ private void providedTypeMethod(ClassModel.Builder classModel, DescribedService
...
-                .addContentCreate(providedType));
+        String name = typeConstant(fieldHandler, providedType);
 
         // TypeName providedType()
         classModel.addMethod(method -> method.addAnnotation(Annotations.OVERRIDE)
                 .returnType(TypeNames.TYPE_NAME)
                 .name("providedType")
-                .addContentLine("return PROVIDED_TYPE;"));
+                .addContent("return ")
+                .addContent(name)
```
danielkec, line 1563: Nitpick: Variable is called `constant` on line 1575 but `name` here, could be more readable following some convention

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -2312,26 +2437,26 @@ private void addMethodElementField(TypeInfo typeInfo,
...
+                .from(method)
+                .annotations(elementAnnotations)
+                .build();
+        classModel.addField(constant -> constant
+                .description("Element info for method: {@code " + method.signature() + "}.")
+                .accessModifier(AccessModifier.PUBLIC)
+                .isStatic(true)
+                .isFinal(true)
+                .type(TypeNames.TYPED_ELEMENT_INFO)
+                .name(constantName)
```
danielkec, line 2457: Huh

tomas-langer: Not sure I undestand this comment...

danielkec: 😄 My bad, thought its the constant again

PR #10832, opened 2025-11-14
Fix hidden edge-case bug in compareNotNullUnsigned (builds on previous PR #9379)

abdelhak-zaaim:
### Summary
This PR fixes a hidden bug in `compareNotNullUnsigned(byte[] data1, byte[] data2)`.  

Although PR #9379 corrected similar issues in `compareNotNull(byte[] data1, byte[] data2) ...`,  
the unsigned byte array version still had subtle edge cases that returned incorrect results.

### Problem
- Arrays with different lengths but first differing element determining the order returned wrong result.

### DIFF
--- a/integrations/db/h2/src/main/java/io/helidon/integrations/db/h2/BitsSubstitution.java
+++ b/integrations/db/h2/src/main/java/io/helidon/integrations/db/h2/BitsSubstitution.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


abdelhak-zaaim:
> Helidon uses H2 version [2.2.240](https://github.com/helidon-io/helidon/blob/2f533b59c73a0f509a7db484c7ca04c6b3904af8/dependencies/pom.xml#L66) ; see the corresponding [Bits](https://github.com/h2database/h2database/blob/version-2.4.240/h2/src/main/org/h2/util/Bits.java) class.
> 
> It only contains the `uuidToBytes` methods. We should remove all other method substitutions, including the one you're fixing: `compareNotNullUnsigned`.

Thanks for the clarification
Just to confirm, should I update this PR to remove the unused substitution methods (including `compareNotNullUnsigned`), or would you prefer that I simply close this pull request?

romain-grecourt:
Sorry I missed your comment. Yes you can update this pull request.

PR #10859, opened 2025-11-26
Log warnings (if not configured to suppress) upon multiple instantiations of `MMeterRegistry`

### LINKED ISSUE #10858: 4.x Add warning message if more than one `MeterRegistry` is created
Helidon Version: 4.x

## Task Description
In Helidon MP, the last part of the `MetricsCdiExtension` start-up uses configuration to set up the behavior of the metrics subsystem and then creates the global `MeterRegistry`.

In two cases we know about, MP user apps and/or libraries have either their own CDI extension which registers meters or trigger async work that registers meters, causing a race condition registering meters _before_ the Helidon MP metrics extension has initialized metrics according to the configuration. This causes two Helidon meter registries to be created only one of which respects the user-specified configuration.

The outwardly-observable symptom in these two cases has been a deadlock partly in Micrometer code, caused by the race condition creating the two global Helidon `MMeterRegistry` instances which share the same single underlying Micrometer meter registry instance. Helidon does its own locking to prevent harmful concurrent access to _its_ meter registry, but in this scenario there are two instances so two locks. These two instances of the Helidon meter registry are unaware of each other's use of the same underlying Micrometer meter registry which does its own locking, hence the deadlock.

Using multiple meter registries is a legitimate use case but should be _very_ rare. (Some of our own unit tests use this feature precisely to isolate meter registrations in different tests from each other. Overwhelmingly, production apps should use a single meter registry.)

This issue proposes that Helidon log a warning message when any additional `MMeterRegistry` is instantiated.

The first warning would include  _two_  stack traces, one for the first instantiation and one for the second. Any subsequent instantiations would trigger a warning including a stack trace for only the new instantiation. This would allow users to see exactly what code is triggering the multiple instantiations. That is in contrast to the current deadlock situations we have seen; the deadlock is partly caused by async code and the thread dump's stack trace for the asynchronous work therefore does not directly show the code that caused the instantiation of the second meter registry.

A new configuration setting `metrics.warn-on-multiple-registries` will default to `true` but could be set to `false` to suppress the warnings.

This problem would at least partly--perhaps wholly--be resolved once Helidon metrics migrates to use the service registry to store the "global" meter registry rather than the current implementation which uses `static`s. But this migration would affect some delicate areas of the code beyond the `MMeterRegistry` constructor, hence this proposal for a less-intrusive approach that could be easily removed once the migration to use the service registry actually occurs.

tjquinno:
### Description
Resolves #10858 

## Release Note
____
Helidon now logs warnings if more than one meter registry is instantiated (unless suppressed by setting `metrics.warn-on-multiple-registries` to `false`).
____

## PR background
The Helidon implementation of the `MeterRegistry` interface for the Micrometer implementation (`MMeterRegistry`) now:
* Records in a static field the stack trace of its first instantiation.
* On instantiations 2 through n logs a warning message, showing the original stack trace and the stack trace of the subsequent instantiation so developers can identify exactly which code is triggering the later instantiations.

See the issue for a summary of the most frequent case where this has been a problem.

The PR includes tests which divert `System.err` to a local stream so it can be inspected to make sure the expected log messages appear and unexpected ones do not appear.

### Documentation
No doc update.


metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -556,6 +565,51 @@ void onMeterRemoved(Meter removedMeter) {
         }
     }
 
+    // For testing.
+    static void clearMultipleInstantiationInfo() {
+        hasLoggedFirstMultiInstantiationWarning = false;
```
tomas-langer: Field are not volatile and not guarded by a lock - this may not work as expected.

metrics/providers/micrometer/src/test/java/io/helidon/metrics/providers/micrometer/TestMultipleRegistryLogging.java
```diff
@@ -0,0 +1,121 @@
...
+
+class TestMultipleRegistryLogging {
+
+    private static ByteArrayOutputStream baos;
+    private static PrintStream originalErr;
+    private static PrintStream testPrintStream;
+
+    @BeforeAll
+    static void beforeAll() {
+        baos = new ByteArrayOutputStream();
```
tomas-langer: Stealing the system error output will also disable output of Junit framework itself.
I think we are using a custom in-memory logging appender in some other test, which is a better approach to check this.

tjquinno: Added a new `MultiStream` to `helidon-tests` which copies all output to the original `OutputStream` (e.g., `System.err`) and all provided  "copy" `OutputStream`s.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tjquinno: Fixed.

PR #10867, opened 2025-12-01
Deprecated public constructor of DataReader, replaced with static create method

tomas-langer:
Replaced usages.
Fixed an issue in readable part of multipart.

### DIFF
--- a/common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
@@ -38,7 +38,9 @@ public class DataReader {
      * Data reader from a supplier of bytes.
      *
      * @param bytesSupplier supplier that can be pulled for more data
+     * @deprecated use {@link #create(java.util.function.Supplier)} instead
      */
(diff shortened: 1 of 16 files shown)


http/media/multipart/src/main/java/io/helidon/http/media/multipart/ReadablePartLength.java
```diff
@@ -105,7 +105,7 @@ public int read() {
             }
 
             partRemaining--;
-            return dataReader.read();
+            return dataReader.read() & 0xFF;
```
romain-grecourt: Please create an issue and fix in a separate PR.

Is it feasible to write a unit test ?

tomas-langer: Removed, will create a separate PR

tomas-langer: PR: https://github.com/helidon-io/helidon/pull/10875
Added a unit test that failed before the fix.

PR #10869, opened 2025-12-02
Declarative metrics.

### LINKED ISSUE #10725: Declarative Metrics

- [ ] Update API (annotations, interceptors etc.)
- [ ] Implement Codegen
- [ ] Add support for new annotations to Helidon MP extension
- [ ] Add support to test new features (unless works out of the box)

tomas-langer:
Timed, Counted, Gauge annotations. Also adds support for injection of `io.helidon.metrics.api.MeterRegistry`.

Codegen to generate interceptors and gauge registrars.

Resolves #10725 

Example of usage:

Counted with custom tags:
```java
@Http.GET
@Http.Path("/counted")
@Metric.Counted(tags = @Metric.Tag(key = "location", value = "method"))
String counted() {
    return "counted";
}
```

Timed with custom name:
```java
@Http.GET
@Http.Path("/timed")
@Metric.Timed(value = "my-timed-metric", absoluteName = true)
String timed() {
    return "timed";
}
```

Gauge (unit is required):
```java
@Metric.Gauge(unit = Meter.BaseUnits.BYTES)
int gaugeValue() {
    return gaugeValue.get();
}
```

The new codegen is part of the existing `helidon-declarative-codegen` module.

### DIFF
--- a/docs/src/main/asciidoc/service-registry/service_reference.adoc
+++ b/docs/src/main/asciidoc/service-registry/service_reference.adoc
@@ -131,4 +131,16 @@ Note: this is a work in progress, not listing the full set of contracts yet!
 |N/A
 |Health check instances to be added to WebServer health observer
 |N/A
+|`io.helidon.metrics.api`
+`MetricsFactory`
+|90
+|`io.helidon.metrics.api`
+|Factory of meter registries
+|N/A
+|`io.helidon.metrics.api`
+`MeterRegistry`
+|90
+|`io.helidon.metrics.api`
+|The "global" meter registry, can be used to create/get custom metrics that cannot be achieved through interception
+|N/A
 |===
--- a/metrics/api/src/main/java/io/helidon/metrics/api/MeterRegistry.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/MeterRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/metrics/api/src/main/java/io/helidon/metrics/api/MeterRegistryFactory.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/MeterRegistryFactory.java
@@ -18,9 +18,12 @@ package io.helidon.metrics.api;
 
 import java.util.function.Supplier;
 
+import io.helidon.common.Weight;
+import io.helidon.common.Weighted;
 import io.helidon.service.registry.Service;
 
 @Service.Singleton
+@Weight(Weighted.DEFAULT_WEIGHT - 10)
 class MeterRegistryFactory implements Supplier<MeterRegistry> {
     private final MetricsFactory metricsFactory;
 
--- a/metrics/api/src/main/java/io/helidon/metrics/api/Metrics.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/Metrics.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/metrics/api/src/main/java/io/helidon/metrics/api/MetricsFactoryFactory.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/MetricsFactoryFactory.java
@@ -18,10 +18,13 @@ package io.helidon.metrics.api;
 
 import java.util.function.Supplier;
 
+import io.helidon.common.Weight;
+import io.helidon.common.Weighted;
 import io.helidon.config.Config;
 import io.helidon.service.registry.Service;
 
 @Service.Singleton
+@Weight(Weighted.DEFAULT_WEIGHT - 10)
 class MetricsFactoryFactory implements Supplier<MetricsFactory> {
     private final Config config;
 
--- a/metrics/api/src/main/java/module-info.java
+++ b/metrics/api/src/main/java/module-info.java
@@ -25,6 +25,7 @@ module io.helidon.metrics.api {
     requires io.helidon.builder.api;
     requires io.helidon.service.registry;
     requires static io.helidon.config.metadata;
+    requires io.helidon.common;
 
     exports io.helidon.metrics.api;
     exports io.helidon.metrics.spi;


docs/src/main/java/io/helidon/docs/se/inject/DeclarativeExample.java
```diff
@@ -190,6 +191,28 @@ public ValidatorResponse check(ValidatorContext context, Object value) {
...
+    // end::snippet_11[]
+
+    // tag::snippet_12[]
+    @Service.Singleton
+    static class ServiceWithAGauge {
+        private volatile int perentage = 0;
+
+        @Metric.Gauge(unit = "percent")
+        int gauge()  {
+            return this.perentage;
```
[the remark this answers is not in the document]
tomas-langer: Fixed.

declarative/tests/metrics/src/main/java/io/helidon/declarative/tests/metrics/TestEndpoint.java
```diff
@@ -0,0 +1,74 @@
...
+    @Http.Path("/time")
+    String time() {
+        return "time: " + meterRegistry.clock().monotonicTime();
+    }
+
+    @Http.POST
+    @Http.Path("/gauge")
+    String gauge(@Http.Entity String value) {
+        gaugeValue.set(Integer.parseInt(value));
+        return "gauge set";
```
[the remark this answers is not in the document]
tomas-langer: This is a test where we control the submitted values, not needed.

metrics/api/src/main/java/io/helidon/metrics/api/Metric.java
```diff
@@ -0,0 +1,233 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.helidon.service.registry.Interception;
+
+/**
+ * Annotations for declarative metrics.
+ */
+public final class Metric {
```
tjquinno: The same package already contains an interface `Metrics` which does general work with respect to the metrics subsystem. Adding this new `Metric` class to the same package in the same component seems unnecessarily confusing. 

Adding the annotations declared here to the existing `Metrics` interface instead would reduce the number of types a developer needs to deal with.

docs/src/main/asciidoc/se/injection/declarative.adoc
```diff
@@ -371,3 +372,29 @@ For each constraint annotation, there MUST be a service that validates it.
...
+- Counter
+- Timer
+- Gauge
+
+Method annotations:
+
+- link:{metrics-javadoc-base-url}/io/helidon/metrics/api/Metric.Counted.html[`io.helidon.metrics.api.Metric.Counted`] - adds a counter metric to the metric registry for method executions
+- link:{metrics-javadoc-base-url}/io/helidon/metrics/api/Metric.Timed.html[`io.helidon.metrics.api.Metric.Timed`] - adds a timer metric to the metric registry for method executions
+- link:{metrics-javadoc-base-url}/io/helidon/metrics/api/Metric.Gauge.html[`io.helidon.metrics.api.Metric.Gauge`] - marks a method that returns a number as a gauge
+
```
tjquinno, line 412: Mention the `@Tag` annotation and its usage in the doc.

docs/src/main/java/io/helidon/docs/se/inject/DeclarativeExample.java
```diff
@@ -190,6 +191,27 @@ public ValidatorResponse check(ValidatorContext context, Object value) {
     }
     // end::snippet_10[]
 
+    // tag::snippet_11[]
+    @Service.Singleton
+    static class MeteredService {
+        @Metric.Counted
```
tjquinno: For either `@Counted` or `@Timed` add at least one tag to show its usage.

metrics/api/src/main/java/io/helidon/metrics/api/Metric.java
```diff
@@ -0,0 +1,233 @@
...
+     * <ul>
+     *     <li>The type that contains a metered method, such tags will be applied to all meters within this type</li>
+     *     <li>The method that has a metric annotation, such tags will be applied to all meters on the method</li>
+     *     <li>In meters annotations {@code tags} property, such tags will be applied only to the single meter</li>
+     * </ul>
+     */
+    @Repeatable(Tags.class)
+    @Target({ElementType.TYPE, ElementType.METHOD})
+    @Retention(RetentionPolicy.CLASS)
+    public @interface Tag {
```
tjquinno: Defining `@Tag` and `@Tags` this way allows the compiler and IDEs to enforce the syntax, but it's very verbose.

What about using either something similar to the familiar Java `Map.of()` style (a sequence of strings, alternating keys and values) or `@Tags({"k1=v1", "k2=v2"})`?

True, neither the Java compiler nor IDEs would be able to enforce either of these, but the annotation processor certainly could do so at build time and the syntax would be much simpler for humans to read and write.

tjquinno (commented):
Some thoughts.

tjquinno:
No `@Inject` support of meters as part of this PR?

Do we intend to add that later? Not at all?

romain-grecourt (commented):
Please provide an overview of the API, and basic usage in the PR description.

metrics/api/src/main/java/io/helidon/metrics/api/Metric.java
tomas-langer: I think the clarity and validation is worth this price. You do not need to learn the syntax used, no need for validation...

tomas-langer: Also the `Tag` annotation can be defined on a type, or on the method and will be used by each metric on that type (or method). 

i.e.:
```java
@Metric.Tag(key = "endpoint", value = "TestEndpoint")
@Metric.Tag(key = "application", value = "MyNiceApp")
class TestEndpoint {
```

which still seems better to me than introducing a new annotation (not used in the metrics) that would have to honor the same syntax, i.e.:

```java
@Metric.Tag("endpoint=TestEndpoint")
@Metric.Tag("application=MyNiceApp")
```

I will always prefer the first option - we have a chance to use the language to make it clear....

tomas-langer:
> No `@Inject` support of meters as part of this PR?
> 
> Do we intend to add that later? Not at all?

I am not sure we need to right now.
Can you point me to an MP example that uses this? Is this a first class use case?

metrics/api/src/main/java/io/helidon/metrics/api/Metric.java
tomas-langer: We would need to deprecate all methods on `Metrics` and either remove them (or move them to another place). 
But I understand this concern and that may be the right solution.

tjquinno:
> > No `@Inject` support of meters as part of this PR?
> > Do we intend to add that later? Not at all?
> 
> I am not sure we need to right now. Can you point me to an MP example that uses this? Is this a first class use case?

See the MP spec https://download.eclipse.org/microprofile/microprofile-metrics-5.1.2/microprofile-metrics-spec-5.1.2.html; search for `@Inject` and the `Colours` example class.

This is probably not a necessary feature for declarative. It leads to some more complexity (the `@Metric` MP annotation, for example, so developers can specify all information needed to identify exactly which meter the system should inject).

Because the PR adds the `Supplier<MeterRegistry>`, IIUC that's all that's required so developers can `@Inject` the meter registry and explicitly look up any specific meter their code needs.

This discussion reminds me that the doc should mention that code can `@Inject` the `MeterRegistry`.

tomas-langer:
For injectable contracts - I have rebased on the PR with service registry reference and will update it.

PR #10880, opened 2025-12-03
Do not use timeout on offering return of a connection to the pool.

### LINKED ISSUE #10861: Race condition in IdleInputStream when reusing WebClient connections under high load
## Environment Details
* Helidon Version: 4.3.1
* Helidon SE
* JDK version: 25 (temurin)
* OS: MacOS 26.1, Rocky8

----------

## Problem Description

I am observing an issue where the response time of our HTTP server (which acts as an aggregator using Helidon WebServer & WebClient to fetch data from multiple external HTTP APIs) gradually degrades over several hours to a day. The system handles hundreds of HTTP requests per second. Restarting the application resolves the latency issue temporarily.

While investigating the root cause of this performance degradation, I discovered a potential race condition in `IdleInputStream` related to the WebClient connection pool. Although I am not yet 100% certain this is the sole cause of the latency accumulation, it is a definite issue where the idle socket monitoring thread may conflict with active connection usage.

#### Technical Details
1.  When a connection is returned to the pool, `idle()` is called in `Http1ConnectionCache`:
    https://github.com/helidon-io/helidon/blob/90f69d9eb7a9f42009652ea6f69b54575bd08f4b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ConnectionCache.java#L186-L187

2.  Inside `idle()`, the `idlingThread` is assigned:
    https://github.com/helidon-io/helidon/blob/90f69d9eb7a9f42009652ea6f69b54575bd08f4b/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java#L110-L115

3.  However, if the connection is immediately retrieved from the pool for a new request (which happens frequently under high load), the check for `idlingThread` inside `read()`:
    https://github.com/helidon-io/helidon/blob/90f69d9eb7a9f42009652ea6f69b54575bd08f4b/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java#L68-L71
    may race with the assignment or execution of the idle monitor.

Specifically, it appears possible for the idle monitoring thread to run or continue running even after the connection has been leased again and should be considered "active".

## Steps to reproduce

I have created a reproduction case that patches `IdleInputStream.java` to add debug logging (`active` flag) and runs a load test.
- diff io/helidon/common/socket/IdleInputStream.java
```java
diff --git a/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java b/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
index b62420e261..d77dbe1c2a 100644
--- a/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
+++ b/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
@@ -52,6 +52,7 @@ class IdleInputStream extends InputStream {
     private volatile int next = -1;
     private volatile boolean closed = false;
     private volatile boolean cancelled = false;
+    private volatile boolean active = true;
     private Future<?> idlingThread;

     IdleInputStream(Socket socket, InputStream upstream, String childSocketId, String socketId) {
@@ -69,6 +70,7 @@ class IdleInputStream extends InputStream {
         if (idlingThread != null) {
             endIdle();
         }
+        active = true;
         if (next < 0) {
             return upstream.read();
         } else {
@@ -83,6 +85,7 @@ class IdleInputStream extends InputStream {
         if (idlingThread != null) {
             endIdle();
         }
+        active = true;
         if (next < 0) {
             return upstream.read(b, off, len);
         } else {
@@ -108,6 +111,7 @@ class IdleInputStream extends InputStream {
      * in blocking manner to detect severed connection.
      */
     void idle() {
+        active = false;
         if (idlingThread != null) {
             return;
         }
@@ -124,6 +128,9 @@ class IdleInputStream extends InputStream {
             int toRestore = socket.getSoTimeout();
             int idleTimeoutIterations = Math.ceilDiv(Math.max(1, toRestore), ITERATION_TIME_MILLIS);
             for (int i = 0; !cancelled; i++) {
+                if (active) {
+                    System.out.println("active=true at IdleInputStream.handle() socket: " + socket);
+                }
                 try {
                     //We need to check the current socket timeout before each iteration.
                     //This time out could have changed,
@@ -168,6 +175,7 @@ class IdleInputStream extends InputStream {
             idlingThread.get();
             idlingThread = null;
             cancelled = false;
+            active = true;
         } catch (InterruptedException | ExecutionException e) {
             closed = true;
             throw new RuntimeException("Exception in socket monitor thread.", e);
```

- src/example/Main.java
```java
package example;

import io.helidon.http.HeaderNames;
import io.helidon.logging.common.LogConfig;
import io.helidon.webclient.api.HttpClientRequest;
import io.helidon.webclient.api.HttpClientResponse;
import io.helidon.webclient.api.WebClient;
import io.helidon.webserver.WebServer;
import io.helidon.webserver.http.HttpRouting;
import io.helidon.webserver.http.ServerRequest;
import io.helidon.webserver.http.ServerResponse;

import java.time.Duration;
import java.util.ArrayList;
import java.util.concurrent.StructuredTaskScope;

public class Main {

    static WebClient client = WebClient
            .builder()
            .followRedirects(false)
            .keepAlive(true)
            .shareConnectionCache(true)
            .connectionCacheSize(100)
            .connectTimeout(Duration.ofMillis(1000))
            .readTimeout(Duration.ofMillis(1000))
            .build();

    public static void main(String[] args) {
        LogConfig.configureRuntime();

                .host("0.0.0.0")
                .port(8080)
                .routing(Main::routing)
                .build()
                .start();

        System.out.println("WEB server is up!");

        for (int i = 0; i <= 100000; i++) {
            try (var scope = StructuredTaskScope.open(
                    StructuredTaskScope.Joiner.<String>awaitAll(),
                    cf -> cf.withTimeout(Duration.ofMillis(1000)))) {
                for (int j = 0; j <= 100; j++) {
                    final int i2 = i;
                    final int j2 = j;
                    StructuredTaskScope.Subtask<String> t = scope.fork(() -> {
                        multiShot(i2, j2);
                    });
                }
                try {
                    scope.join();
                } catch (StructuredTaskScope.TimeoutException e) {
                    scope.close();
                }
            }
        }
    }

    static void multiShot(int i, int j) {
        try (var scope = StructuredTaskScope.open(
                StructuredTaskScope.Joiner.<String>awaitAll(),
                cf -> cf.withTimeout(Duration.ofMillis(200)))) {
            ArrayList<StructuredTaskScope.Subtask<String>> arr = new ArrayList<>();
            long start = System.currentTimeMillis();
            for (int k = 0; k < 10; k++) {
                StructuredTaskScope.Subtask<String> t = scope.fork(() -> {
                    HttpClientRequest req = client.get();
                    req.uri("http://localhost:8080/hello");
                    try (HttpClientResponse res = req.request()) {
                        String s = res.as(String.class);
                        return s;
                    }
                });
                arr.add(t);
            }
            try {
                scope.join();
            } catch (StructuredTaskScope.TimeoutException e) {
                scope.close();

            }
            long numOfSuccess = arr.stream()
                    .filter((t) -> t.state() == StructuredTaskScope.Subtask.State.SUCCESS && t.get().equals("hello"))
                    .count();
            long end = System.currentTimeMillis();
            long duration = (end - start);
            if (i % 100 == 0 && j == 0) {
                System.out.println("i: " + i + ", success: " + numOfSuccess + "/" + 10 + " duration: " + duration + "ms");
            }
        }
    }

        routing.get("/hello", Main::helloResponse);
    }

    static void helloResponse(ServerRequest request, ServerResponse response) {
        try {
            Thread.sleep(Duration.ofMillis(10));
        }
        response.header(HeaderNames.CONTENT_TYPE, "text/plain")
                .send("hello");;
    }
}
```

- Output
```
WEB server is up!
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53075]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53067]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53079]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53046]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53069]
i: 0, success: 10/10 duration: 193ms
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53799]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=54812]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=55051]
...omitted...
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53710]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53828]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=53824]
i: 100, success: 0/10 duration: 207ms
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=57233]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=58915]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=58914]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=62788]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=63604]
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=63605]
i: 200, success: 10/10 duration: 30ms
active=true at IdleInputStream.handle() socket: Socket[addr=localhost/127.0.0.1,port=8080,localport=50145]
```
The log output confirms that `IdleInputStream.handle()` (the monitor thread) is running and sees `active=true` , meaning the connection is simultaneously being used by the application logic.

### LINKED ISSUE #10866: Question: Is the timed offer in Http1ConnectionCache redundant with pool size configuration?
Helidon Version: 4.3.1

## Enhancement Description

In `Http1ConnectionCache.finishRequest()`, a timed `offer()` with a 10ms timeout is used when returning
connections to the pool:

https://github.com/helidon-io/helidon/blob/2fdbbbdce95eb91c3feb283573d51b8caa2cfdc7/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ConnectionCache.java#L186

I'm wondering if this introduces unnecessary complexity.

#### My concern

The connection cache size is already configurable via connectionCacheSize. If the timed wait succeeds in
returning a connection to the pool (because another thread freed up a slot within those 10ms), this achieves
essentially the same effect as having configured a slightly larger pool size in the first place.

In other words, we now have two separate mechanisms that both control "how eagerly we retain connections":
1. The pool size configuration
2. The timed wait when returning connections

This feels like a form of dual management for the same concern. If users want to retain more connections, they
can simply increase the pool size—a single, clear knob to control this behavior.

#### Question

Would it be simpler to use a non-blocking offer() instead?

```java
if (connectionQueue.offer(conn)) {
```

This would consolidate the connection retention policy into a single configuration: pool size alone determines
how many connections are kept, without the additional timed-wait mechanism.

Is there a specific scenario where having both mechanisms provides value that pool size alone cannot achieve?
I'd appreciate any insight into the design decision here.

Thank you!

tomas-langer:
Remove timeout from parameters, as it is not used (and is configured after a connection is created) Fix race condition.

Resolves #10866 
Resolves #10861

### DIFF
--- a/webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/ResponseListenerTest.java
+++ b/webserver/tests/webserver/src/test/java/io/helidon/webserver/tests/ResponseListenerTest.java
@@ -16,6 +16,10 @@
 
 package io.helidon.webserver.tests;
 
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import io.helidon.http.Status;
@@ -33,8 +37,8 @@ import static org.hamcrest.MatcherAssert.assertThat;
 @ServerTest
 class ResponseListenerTest {
     private static final AtomicBoolean BEFORE_SEND_CALLED = new AtomicBoolean(false);
-    private static final AtomicBoolean WHEN_SENT_CALLED = new AtomicBoolean(false);
     private static final AtomicBoolean WHEN_BEFORE_SEND = new AtomicBoolean(false);
+    private static CompletableFuture<Boolean> whenSentCompleted = new CompletableFuture<>();
 
     private final Http1Client client;
 
@@ -48,11 +52,11 @@ class ResponseListenerTest {
             res.beforeSend(() -> {
                         res.status(Status.PARTIAL_CONTENT_206);
                         BEFORE_SEND_CALLED.set(true);
-                        if (WHEN_SENT_CALLED.get()) {
+                        if (whenSentCompleted.isDone()) {
                             WHEN_BEFORE_SEND.set(true);
                         }
                     })
-                    .whenSent(() -> WHEN_SENT_CALLED.set(true));
+                    .whenSent(() -> whenSentCompleted.complete(true));
 
             res.send("done");
         });
@@ -60,13 +64,13 @@ class ResponseListenerTest {
 
     @AfterEach
     void afterEach() {
-        WHEN_SENT_CALLED.set(false);
+        whenSentCompleted = new CompletableFuture<>();
         WHEN_BEFORE_SEND.set(false);
         BEFORE_SEND_CALLED.set(false);
     }
 
     @Test
-    void testCalls() {
+    void testCalls() throws ExecutionException, InterruptedException, TimeoutException {
         var response = client.get("/").request(String.class);
 
         assertThat(response.status(), is(Status.PARTIAL_CONTENT_206));
@@ -74,6 +78,9 @@ class ResponseListenerTest {
 
         assertThat("whenSent called before beforeSend", WHEN_BEFORE_SEND.get(), is(false));
         assertThat("beforeSend not called", BEFORE_SEND_CALLED.get(), is(true));
-        assertThat("whenSent not called", WHEN_SENT_CALLED.get(), is(true));
+        // there is a race condition - `whenSent` is called after the response is sent, so we may get a response here before
+        // it is completed by the server; ergo we must use a future. The other values are guaranteed to be set before
+        // the response is sent
+        assertThat("whenSent not called", whenSentCompleted.get(5, TimeUnit.SECONDS), is(true));
     }
 }


common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
```diff
@@ -100,6 +100,9 @@ public int read(byte[] b, int off, int len) throws IOException {
     public void close() throws IOException {
         upstream.close();
         closed = true;
+        if (idlingThread != null) {
+            idlingThread.cancel(true);
+        }
```
hozumi, line 105: Thanks for the fix.

I'm wondering if `idlingThread` is properly cancelled when a connection is discarded because the connection pool is full. Is `IdleInputStream.close()` actually called from anywhere? It doesn't seem to be invoked in TcpClientConnection.closeResource() or PlainSocket.close().
https://github.com/helidon-io/helidon/blob/13d0e60cbe9034db5d5625ae4518564ac192c25d/webclient/api/src/main/java/io/helidon/webclient/api/TcpClientConnection.java#L199
https://github.com/helidon-io/helidon/blob/13d0e60cbe9034db5d5625ae4518564ac192c25d/common/socket/src/main/java/io/helidon/common/socket/PlainSocket.java#L111-L117

I'm not fully familiar with the relationship between `HelidonSocket` and `java.net.Socket` , so I'm not exactly sure where and when they should be closed. It might be fine since j`ava.net.SocketException: Socket closed` is caught here and the future completes anyway.
https://github.com/helidon-io/helidon/blob/de462e411cc1adbacd3d1eaafb57ccc8f59ea5b6/common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java#L162

tomas-langer: The caller that wants to return the connection to the pool checks the result, and if false, the connection is closed (and the idle thread cancelled as part of it). 
Even if it was not cancelled, the underlying socket is closed, the the read would return and the idle thread would finish.

hozumi: I understand the mechanism you described – that makes sense.
However, it still looks to me like `IdleInputStream.close()` is never explicitly called anywhere. Is that observation correct?
It might be out of scope here, but I was just curious.

tomas-langer: The observation is correct, it is never closed explicitly.

tomas-langer:
Rebasing, I will try to figure out the `IdleInputStream.close()`...

common/socket/src/main/java/io/helidon/common/socket/IdleInputStream.java
tomas-langer: Now closing the stream explicitly in `PlainSocket`

Issue #9869, opened 2025-03-05
4.x: Micronaut Data cannot build on Java 24 with the current MN version


tomas-langer:
## Environment Details
* JDK version: 24 build 36

----------

## Problem Description
Failure to build tests for Micronaut Data integration on Java 24

## How to reproduce:
Switch to Java 24
Remove `not-java-21` profile in `integrations/micronaut/data/pom.xml`
Make `java_21` profile active by default, or move its content to `build` of the project
Run build that compiles test sources (with the annotation processors of Micronaut)

The build fails - there is no message at all. The message can be seen when the `javac` command gets executed manually from command line:

```
error: Unable to implement Repository method: DbPetRepository.delete(Object entity). Cannot query persistentEntity [Pet] on non-existent property: entity
```

Everything works fine on Java 21

jbescos:
In a first view, there are some micronaut classes that are supposed to be generated. With jdk24 they are not.

And when I check with -X the output, in jdk21 we see:
`[DEBUG]   (f) annotationProcessorPaths = [io.micronaut:micronaut-inject-java:4.2.2.jar, io.micronaut:micronaut-core-processor:4.2.2.jar, io.micronaut.validation:micronaut-validation-processor:4.2.0.jar, io.micronaut.data:micronaut-data-processor:4.4.1.jar, io.helidon.integrations.micronaut:helidon-integrations-micronaut-cdi-processor:4.4.0-SNAPSHOT.jar]`

Meanwhile in jdk24 we see:
`[DEBUG]   (f) annotationProcessorPaths = [io.helidon.common.features:helidon-common-features-codegen:4.4.0-SNAPSHOT.jar]`

PR #10883, opened 2025-12-04
4.x: Micronaut Data cannot build on Java 24 with the current MN versi…

jbescos:
Resolves https://github.com/helidon-io/helidon/issues/9869


jbescos:
@barchetta , this PR contains a newer version of micronaut, and they changed the module names, and that could make it backwards incompatible. Could you write that in the release notes, please?.

PR #10894, opened 2025-12-09
UNIX Domain Socket support

### LINKED ISSUE #3241: Bind to Unix Domain Sockets
## Environment Details
* Helidon Version: 2.3.2
* JDK version: 16+
* OS: Linux

----------

## Problem Description
Permit binding to Unix Domain Sockets (via `java.net.UnixDomainSocketAddress` in Java 16+) to aid performance and security. This is especially useful in K8s to allow local binding within a Pod rather than exposing ports. Unfortunately that may also mean replacing the existing port (`int`) and binding address (`InetAddress`) configuration with `java.net.SocketAddress`.

## Steps to reproduce
N/A

tomas-langer:
- add support for UNIX domain sockets to webserver and webclient
- use SocketChannel in server listener (required for the above)
- add support for socket channels when accepting connections to webserver

Resolves #3241 

### Description
UNIX domain sockets are only supported when `SocketChannel` is used.
This PR replaces server socket with a `SocketChannel`, and by default now uses `SocketChannel` when accepting connections.

A `bind-address` can be now configured both on the server and the client with a prefix `unix:` to mark that the listener (or the client) should bind to a UNIX Domain socket.

Example:

```yaml
server:
  bind-address: "unix:/tmp/benchmark.sock"
```

The socket is created and for the server also deleted on listener shutdown. 
Note that JVM does not remove the socket file, so if the server is shutdown in a non-standard way and the file is not deleted, the server will not start, unless the file is deleted manually.

The most complex part is the support for `SSLEngine` with the new `TlsNioSocket`. 

A few tests had to be changed, as the behavior changed (i.e. we get a different exception as we no longer use `Socket`).

At least according to my local benchmark runs, this is faster than the original code.
Results:

```
Pipelining
Original:       8_466_935.32 req/s
NIO:           10_543_301.99 req/s

No pipelining
Original:       2_193_370.62 req/s
NIO:            2_266_242.64 req/s
```

### DIFF
--- a/common/socket/src/main/java/io/helidon/common/socket/TlsSocket.java
+++ b/common/socket/src/main/java/io/helidon/common/socket/TlsSocket.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


tomas-langer:
Unfortunatelly `wrk` does not support UNIX domain sockets, so I could not run a benchmark against it. From my small tests with `http2`, the results were very similar to TCP sockets.

common/buffers/src/main/java/io/helidon/common/buffers/FixedBufferData.java
```diff
@@ -129,7 +138,7 @@ public FixedBufferData write(int value) {
 
     @Override
     public int writeTo(ByteBuffer writeBuffer, int length) {
-        int toWrite = Math.min(writeBuffer.limit() - writeBuffer.position(), this.length - readPosition);
+        int toWrite = Math.min(writeBuffer.limit() - writeBuffer.position(), writePosition - readPosition);
```
spericas, line 141: A bug we had before? Looks good now

common/socket/src/main/java/io/helidon/common/socket/SocketOptionsBlueprint.java
```diff
@@ -130,6 +131,26 @@ default void configureSocket(Socket socket) {
...
     }
 
+    /**
+     * Configure socket with defined socket options.
+     *
+     * @param socket socket to update
+     */
+    @SuppressWarnings("unchecked")
+    default void configureSocket(SocketChannel socket) {
+        var logger = System.getLogger(SocketOptions.class.getName());
```
spericas, line 141: Why not static? Or at least in catch block

common/buffers/src/main/java/io/helidon/common/buffers/FixedBufferData.java
tomas-langer: Yes, this was a bug. Luckily we never used `ByteBuffer` before

common/socket/src/main/java/io/helidon/common/socket/SocketOptionsBlueprint.java
tomas-langer: Because it is an interface and it would become publicly accessible.

spericas (dismissed):
LGTM

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -238,6 +246,14 @@ private void suspend(boolean shutdownExecutors) {
         try {
             // Stop listening for connections
             serverSocket.close();
+            if (configuredAddress instanceof UnixDomainSocketAddress udsa) {
```
romain-grecourt, line 249: Can this be abstracted so that this code resides in the unix-domain-socket module ?

webclient/http2/src/test/resources/logging-test.properties
```diff
@@ -24,3 +24,4 @@ java.util.logging.SimpleFormatter.format=%1$tH:%1$tM:%1$tS %4$s %3$s %5$s%6$s%n
 
 # Global logging level. Can be overridden by specific loggers
 .level=INFO
+io.helidon.level=FINEST
```
danielkec: Intentional?

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
danielkec: Do we need extra module? Why?

common/socket/src/main/java/io/helidon/common/socket/TlsNioSocket.java
```diff
@@ -0,0 +1,436 @@
...
+    Optional<Certificate[]> tlsPeerCertificates() {
+        try {
+            return Optional.of(engine.getSession().getPeerCertificates());
+        } catch (SSLPeerUnverifiedException e) {
+            return Optional.empty();
+        }
+    }
+
+    Optional<Principal> tlsPrincipal() {
+        return Optional.of(engine.getSession().getLocalPrincipal());
```
danielkec:
```suggestion
        return Optional.ofNullable(engine.getSession().getLocalPrincipal());
```

common/socket/src/main/java/io/helidon/common/socket/TlsNioSocket.java
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    Optional<Certificate[]> tlsCertificates() {
+        return Optional.of(engine.getSession().getLocalCertificates());
```
danielkec:
```suggestion
        return Optional.ofNullable(engine.getSession().getLocalCertificates());
```

common/socket/src/main/java/io/helidon/common/socket/TlsNioSocket.java
tomas-langer: This is copied from `TlsSocket`, where we have the same issue. Will fix both.

common/socket/src/main/java/io/helidon/common/socket/TlsNioSocket.java
```diff
@@ -0,0 +1,436 @@
...
+                    doHandshake(handshakeStatus);
+                }
+            }
+        }
+    }
+
+    private SSLEngineResult wrapAndSend(ByteBuffer appData, boolean ignoreClose) throws SSLException {
+        if (closed && !ignoreClose) {
+            throw new UncheckedIOException(new SSLException("Engine is closed"));
+        }
```
danielkec, line 407: Is unchecked ex. wrapping needed here when `throws SSLException` is declared on the method already?

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
tomas-langer: The difference between TCP socket and UNIX domain socket right now is only when creating the address and when shutting down the listener.

This is a very small change, and the abstraction to another module would require a lot of new integration points. 
I would cross that bridge if we need to integrate a different `SocketChannel` based listener.

danielkec (approved):

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
romain-grecourt: > Do we need extra module? Why?

Actually, my bad. I thought that we were adding a new module.
If the UNIX domain socket support comes out of the webserver module then there is no need to abstract anything.

PR #10897, opened 2025-12-10
Fix a problem where we kept bad candidates for builder methods.

### LINKED ISSUE #10896: Blueprint generation fails when a void non-default method is used.
We can (theoretically) define any method on the blueprint that is non-default, and then add an implementation using custom methods.
This is currently not possible, as even though we get a warning the method is not valid, it is still added as an option candidate.

These methods should be warned and ignored.

Desired outcome:
```
[WARNING] Builder definition methods cannot have void return type (must be getters): io.helidon.builder.tests.custom.voidmethod.SomeTypeBlueprint.doSomething
[WARNING] Builder definition methods cannot have parameters (must be getters): io.helidon.builder.tests.custom.voidmethod.SomeTypeBlueprint.doSomething
```

Current outcome:
compilate of `SomeType` fails, because it has a `void` field.

tomas-langer:
### Description
Resolves  #10896

non-default `void` method on blueprint is now accepted - warned.
It can be implemented using a custom method.

This is not a recommended pattern, though it can be useful when implementing third party interfaces that we consider prototypes.

Recommended solution for our own prototypes - only add the custom method to prototype, do not have it on blueprint.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryOption.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryOption.java
@@ -221,7 +221,7 @@ final class FactoryOption {
         // we must validate - any candidate that is void, or has parameters is a bad candidate
         var collector = Errors.collector();
         for (TypedElementInfo candidate : candidates) {
-            if (candidate.typeName().equals(TypeNames.PRIMITIVE_VOID)) {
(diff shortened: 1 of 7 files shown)


tomas-langer:
The generated `SomeType` in the test will generate the following method:

```java
    @Override
    default void doSomething(List<String> first, List<String> second) {
        SomeTypeSupport.doSomething(this, first, second);
    }
```

and compiles...

PR #10900, opened 2025-12-11
4.x: Upgrades Narayana to 7.1.0.Final

ljnelson:
This PR:
* Upgrades Narayana to ~7.3.3.Final~ 7.1.0.Final (see comments below)
* ~Upgrades `jboss-logging` to 3.6.1.Final (the [version required by Narayana 7.3.3.Final](https://github.com/jbosstm/narayana/blob/7.3.3.Final/boms/main/pom.xml#L63)). Note that this is higher than [the version Weld 5.1.6.Final (the version our repackaging uses) uses](https://github.com/weld/core/blob/5.1.6.Final/pom.xml#L81). It does not appear after cursory testing to cause problems, but we'll let the pipelines sort it out.~ Ensures `jboss-logging` is at version ~3.5.0.Final, used by both our repackaged Weld 5.1.6.Final and Narayana 7.1.0.Final~ 3.5.3.Final (see comments)
* Removes an unused? reference to `jboss-logging-annotations` which may have had a purpose once upon a time but doesn't seem to now.

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -94,7 +94,7 @@
         <version.lib.jaxb-core>4.0.5</version.lib.jaxb-core>
         <version.lib.jaxb-impl>4.0.5</version.lib.jaxb-impl>
         <version.lib.jboss.classfilewriter>1.3.1.Final</version.lib.jboss.classfilewriter>
-        <version.lib.jboss.logging>3.5.0.Final</version.lib.jboss.logging>
+        <version.lib.jboss.logging>3.5.3.Final</version.lib.jboss.logging>
         <version.lib.jaxb-runtime>4.0.5</version.lib.jaxb-runtime>
         <version.lib.jersey>3.1.11</version.lib.jersey>
         <!-- Jetbrains annotations: dependency convergence requirement, we never use this directly -->


ljnelson:
Packaging failures are due to: https://github.com/jboss-logging/jboss-logging/blob/3.6.1.Final/src/main/java/org/jboss/logging/Logger.java#L2572

This is called by, effectively: https://github.com/weld/core/blob/5.1.6.Final/impl/src/main/java/org/jboss/weld/logging/VersionLogger.java#L35 (our Weld repackaging [uses a slightly altered version of 5.1.6.Final](https://github.com/helidon-io/helidon/blob/4.3.2/dependencies/pom.xml#L169))

Sample output:
```
Caused by: java.lang.IllegalArgumentException: This library does not have private access to interface org.jboss.weld.logging.VersionLogger
	at org.jboss.logging@3.6.1.Final/org.jboss.logging.Logger.getMessageLogger(Logger.java:2572)
	at org.jboss.logging@3.6.1.Final/org.jboss.logging.Logger.getMessageLogger(Logger.java:2552)
	at weld.core.impl@4.4.0-SNAPSHOT/org.jboss.weld.logging.VersionLogger.<clinit>(VersionLogger.java:35)
```
(`weld.core.impl@4.4.0-SNAPSHOT` in this run is our repackaged Weld module which as previously noted is actually version 5.1.6.Final.)

ljnelson:
Upgrading Weld requires repackaging and fixing our rewrite of its proxy machinery. If we decide to do this, that's a whole different PR.

Holding our repackaged Weld version at 5.1.6.Final, our only option is to upgrade to a version of Narayana that still uses `jboss-logging` at 3.5.0.Final (the same version used by Weld 5.1.6.Final). That version is, broadly, 7.1.2.Final.

However, for whatever reason, [a key component of Narayana that we use was never released at 7.1.2.Final](https://central.sonatype.com/artifact/org.jboss.narayana.jta/narayana-jta). Instead, the most recent compatible release was 7.1.0.Final.

So I'll make this PR be an upgrade to Narayana 7.1.0.Final, where we will be forced to remain until we decide to upgrade and re-re-package Weld, whereupon we should upgrade `jboss-logging` and Narayana appropriately.

ljnelson:
Aligning jboss-logging at 3.5.0.Final, the version required by both Weld 5.1.6.Final and Narayana 7.1.0.Final, causes our modularity tests to fail. Bumping it back to 3.5.3.Final, a rather arbitrary choice, appears to fix them.

barchetta (approved):

ljnelson:
@tomas-langer gives his approval out-of-band. Merging.

PR #10902, opened 2025-12-11
Export package from OpenAPI UI integration module; with integration test

### LINKED ISSUE #10889: OpenApiUi class is inaccessible in Modular applications
## Environment Details
* Helidon Version: **4.3.2**
* Helidon SE or Helidon MP: **Helidon SE**
* JDK version: **25**
* OS: **Linux**

----------

## Problem Description
I am building a Helidon SE application using Java Modules (`module-info.java`). I need to configure the OpenAPI UI programmatically, exactly as described in the [official documentation](https://helidon.io/docs/v4/se/openapi/openapi-ui) (Customizing the UI Behavior).

The documentation instructs developers to use `OpenApiUi.builder()`. However, the module `io.helidon.integrations.openapi.ui` **does not export** the package `io.helidon.integrations.openapi.ui`.

Because the package is encapsulated, the Java compiler forbids access to `OpenApiUi` when `requires io.helidon.integrations.openapi.ui;` is used in a modular application.

**Compiler Error:**
```text
error: package io.helidon.integrations.openapi.ui is not visible
import io.helidon.integrations.openapi.ui.OpenApiUi;
                                      ^
  (package io.helidon.integrations.openapi.ui is declared in module io.helidon.integrations.openapi.ui, which does not export it)
```

**Current Workaround:**
I am forced to use compiler arguments to break encapsulation:
`--add-exports=io.helidon.integrations.openapi.ui/io.helidon.integrations.openapi.ui=my.module.name`

## Steps to reproduce
1. Create a Helidon SE project using Java 21+.
2. Add a `module-info.java` file to the project.
3. Add the dependency `io.helidon.integrations.openapi-ui:helidon-integrations-openapi-ui`.
4. In `module-info.java`, add: `requires io.helidon.integrations.openapi.ui;`.
5. Attempt to use the builder in Java code:

```java
package com.example.myapi;

import io.helidon.integrations.openapi.ui.OpenApiUi; // <--- Compiler Error Here

public class DocService {
    public void init() {
        // As per documentation
        var ui = OpenApiUi.builder()
                .build();
    }
}
```

tjquinno:
### Description
Resolves #10889 

### Release Note
____
The Helidon OpenAPI UI integration component now exports its package so applications can set up the UI behavior programmatically and not only using config files.
____

The PR adds an `exports` line to the OpenAPI UI `module-info.java` and adds an integration test.

### Documentation
No impact.


tests/integration/gh-10889-openapi-ui-module/helidon-tests-integration-openapi-ui-module/src/main/java/io/helidon/tests/integration/openapiui/module/GreetService.java
romain-grecourt, line 1: We should only include the required files in tests, let's not include all the files from the quickstart archetype.

---

You could run the test as a module with a module-info under `src/test/java`, however this will likely require you to "modularize" the test dependencies.
If you go this way you don't need a Main class you can simply use `@ServerTest` and setup `OpenApiFeature` with `@SetUpServer`.

Or, you can re-use the integration test harness (`tests/integration/harness`) used by the packaging integration tests (`tests/integration/packaging`) to exercise JPMS related packaging (jlink-image, module-path, jlink-image + module-path).

---

I suggest updating `tests/integration/packaging/se-1` to setup OpenAPI with UI and a update the existing test to verify that the UI is served.

tjquinno: I've removed some/all of the extraneous files.

Because the bug came from an end user writing an app I wanted to mimic that environment as much as possible, hence I opted to build a separate app for the test.

I suppose the test could've been added to se-1 but that would have perturbed that app's composition--adding OpenAPI and the UI. It seemed reasonable and expedient to just keep this integration test of the UI integration--one we actually recommend _against_ including in apps--isolated by itself.

romain-grecourt: Does the current test fail without the module-info change ?
I don't see anything that runs the test with the module path, am I missing something ?

tjquinno:
The test doesn't compile without the fix.

romain-grecourt (commented):
I disagree with the test.

It enforces that users can compile with the module-path and import from `io.helidon.integrations.openapi.ui`.

This should require 3 files: `pom.xml`, `module-info.java`, `Main.java`.

If all we do is compile for a test, then we should use the maven-invoker-plugin.

---

If the goal is to make a functional test for this issue, then the test needs to execute with the module-path, otherwise do a compile only test.

Instead of copying files from `helidon init`, I'd like to see minimum required code and dependencies to test the OpenAPI UI.

I.e.,  A single endpoint (E.g. `/test`) with the corresponding OpenAPI spec and minimum dependencies.

---

The easiest is to update `tests/integration/packaging/se-1`, it's what they are intended for.

tjquinno:
I've removed the integration test and used the invoker plug-in to build a small project contained in the component itself.

romain-grecourt (approved):

PR #10934, opened 2025-12-12
Declarative tracing

### LINKED ISSUE #10731: Declarative Tracing/Telemetry

- [x] Update API (annotations, interceptors etc.)
- [x] Implement Codegen
- [x] Add support for new annotations to Helidon MP extension
- [x] Add support to test new features (unless works out of the box)

tomas-langer:
### Description
Adds declarative tracing.
Resolves #10731 

### Documentation

Adds `Tracing.Traced` and `Tracing.TagParam` annotations.
Adds code generator for element interceptors for annotated methods.

Example of usage:

```java
@RestServer.Endpoint
@Http.Path("/endpoint")
@Service.Singleton
@Tracing.Traced(tags = @Tracing.Tag(key = "endpoint", value = "Test"), kind = Span.Kind.SERVER)
class TestEndpoint { 
    @Http.GET
    @Http.Path("/greet")
    @Tracing.Traced(value = "explicit-name", tags = @Tracing.Tag(key = "custom", value = "customValue"))
    String greet(@Http.HeaderParam("User-Agent") @Tracing.TagParam String userAgent) {
        return "Hello!";
    }
}
```

Test failure in examples is a verified race condition that is not caused by this PR.

### DIFF
--- a/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/ContentLengthTest.java
+++ b/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/ContentLengthTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -166,14 +166,10 @@ class ContentLengthTest {
                             "Request data length doesn't correspond to the content-length header.",
                             TIMEOUT);
 
-        // content length discrepancy is discovered when consuming request data
-        var e = consumeExceptionFuture.get(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
-        assertThat(e, Matchers.instanceOf(RequestException.class));
-        assertThat(e.getMessage(), is("Stream is closed."));
-
-        // stream is closed, sending is not possible
-        e = sendExceptionFuture.get(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
-        assertThat(e, Matchers.instanceOf(IllegalStateException.class));
-        assertThat(e.getMessage(), is("Stream is already closed."));
+        /*
+        As in the previous test, this may not reach routing at all (depends on environment, buffer sizes etc.).
+        The original block of code should have been removed when changes for unix domain sockets were done, as in
+        longerData() above.
+         */
     }
 }


tracing/tracing/src/main/java/io/helidon/tracing/Tracing.java
```diff
@@ -0,0 +1,112 @@
...
+     * Tags are added to spans (alternative term is Span Attribute).
+     * <p>
+     * Tag key is either explicitly defined, or inferred from the parameter name, tag value is the parameter value.
+     * <p>
+     * Note that when this annotation is used with runtime processing (i.e. in Helidon MP), the value MUST be defined,
+     * as names of parameters are not retained in compiled Java classes.
+     */
+    @Target(ElementType.PARAMETER)
+    @Retention(RetentionPolicy.RUNTIME)
+    public @interface TagParam {
```
tjquinno: Mentioned in a separate discussion that, to me, `@ParamTag` would convey the meaning more clearly.

tracing/providers/jaeger/src/main/java/io/helidon/tracing/providers/jaeger/JaegerTracerBuilder.java
```diff
@@ -622,6 +622,17 @@ JaegerTracerBuilder exporter(SpanExporter spanExporter) {
         return this;
     }
 
+    /**
+     * Add a span exporter.
+     *
+     * @param spanExporter span exporter to use
+     * @return updated builder instance
+     */
+    public JaegerTracerBuilder addSpanExporter(SpanExporter spanExporter) {
```
tjquinno, line 625: The package-private `exporter` method does the same thing and is used from one line in a test. Maybe remove `exporter` and change the test to use this new equivalent public method.

tracing/tracing/src/main/java/io/helidon/tracing/TracerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.function.Supplier;
+
+import io.helidon.common.Weight;
+import io.helidon.common.Weighted;
+import io.helidon.config.Config;
+import io.helidon.service.registry.Service;
+
+@Service.Singleton
+@Weight(Weighted.DEFAULT_WEIGHT - 20)
+class TracerFactory implements Supplier<Tracer> {
```
tjquinno, line 28: Also mentioned In a separate discussion...

Existing Helidon and user code might use the `TracerProvider.global(myTracer)` method to assign a global tracer. Granted, we are moving away from statics and so-called global instances toward the service registry instead, and we need a `Supplier<Tracer>` to support injection in declarative.

That said, users with code that invokes `TracerProvider.global(myTracer)` might expect the "global" tracer their code registers to be the one injected.

We could either make sure that this service factory and `TracerProvider` work together in some way or clearly document that the injected tracer will always be derived from configuration. 

A `Tracer` factory in this PR also needs to account for the fact that the relatively recently-added support for `telemetry` configuration includes `HelidonOpenTelemetryServiceFactory` (a `Supplier<OpenTelemetry>`) which invokes `io.helidon.telemetry.otelconfig.HelidonOpenTelemetry.create(otelConfig)` which instantiates `HelidonOpenTelemetryImpl` which itself creates a tracer and passes it to `Tracer.global(tracer)` (which ultimately delegates to `TracerProvider.global(tracer)`).

tjquinno: Also worth noting that a parallel PR #10791 also has a `Tracer` service factory. We need to make sure that these separate parts of the code work together so the correct `Tracer` is registered and injected.

tomas-langer: This is working through the following calls:
- `Tracer.global(Tracer)`
- `TracerProviderHelper.global(Tracer)`
- Services.set(Tracer.class, tracer);

So if the user invokes `Tracer.global(Tracer)` before it is used anywhere, the instance in the registry will be this custom instance, and whenever an injection point of this type is encountered, the custom instance will be injected. In such a case the `TracerFactory` will never be triggered.

tjquinno: Oh, right. I was thinking about the new service factory and hadn't internalized the changes to `TracerProviderHelper` which register the tracer directly.

tracing/providers/jaeger/src/main/java/io/helidon/tracing/providers/jaeger/JaegerTracerBuilder.java
tomas-langer: Using `addSpanExporter` instead of `exporter`, removed old package private method.

tracing/tracing/src/main/java/io/helidon/tracing/Tracing.java
tomas-langer: Changed according to your comment.

PR #11000, opened 2025-12-29
Helidon SE Scheduling: Add ZoneId support in cron builder

### LINKED ISSUE #10984: Helidon SE Scheduling: Add ZoneId support in cron builder
The `io.helidon.scheduling.Cron` [builder](https://helidon.io/docs/v4/se/scheduling#_cron) in Helidon SE 4.3.2 currently lacks the ability to define a specific `ZoneId` for task execution. It relies on the system default timezone.

This limitation prevents accurate scheduling for regions observing Daylight Saving Time (DST). For example, a task scheduled for `09:00 America/New_York` cannot be represented by a static UTC cron expression, as the offset shifts between UTC-4 and UTC-5 throughout the year.

**Proposed Enhancement**

Expose a method in the builder to accept a `ZoneId`:

```java
Cron.builder()
     .expression("0 0 9 * * ?")
     .zone(ZoneId.of("America/New_York")) // Proposed API
     .task(task)
     .build();
```

Since the underlying dependency (`cron-utils`) already supports `ZonedDateTime` for execution time calculations, this enhancement would likely involve propagating the configured zone to the internal next-execution check.

denyshorman:
**Fixes #10984**

**Summary**
This PR adds the ability to configure a specific `ZoneId` for Cron tasks in Helidon SE. Previously, Cron expressions were always evaluated against the system's default timezone (usually UTC), making it difficult to schedule tasks for specific "wall clock" times in regions observing Daylight Saving Time.

**Changes**
*   **API:** Added `zone(ZoneId zone)` to the `Cron` builder and configuration blueprint.
*   **Implementation:** Updated `CronTask` to calculate `ZonedDateTime.now(zone)` using the configured timezone instead of the system default.
*   **Default Behavior:** The default value is `ZoneId.systemDefault()`, ensuring full backward compatibility for existing applications.
*   **Documentation:** Added a "Timezone Configuration" section to `scheduling.adoc`.
*   **Tests:** Added `cronWithCustomZone` test to verify execution respects the configured zone when it differs from the system zone.

**Example Usage**
```java
Cron.builder()
    .expression("0 0 9 * * ?")
    .zone(ZoneId.of("America/New_York")) // Now supported
    .task(task)
    .build();
```

### DIFF
--- a/scheduling/src/test/java/io/helidon/scheduling/CronSchedulingTest.java
+++ b/scheduling/src/test/java/io/helidon/scheduling/CronSchedulingTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2021, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,8 +17,11 @@
 package io.helidon.scheduling;
 
 import java.time.Duration;
+import java.time.Instant;
+import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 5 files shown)


scheduling/src/test/java/io/helidon/scheduling/CronSchedulingTest.java
```diff
@@ -334,4 +337,89 @@ void cronDisabledProgrammatically() throws InterruptedException {
...
+            // switch to Tokyo to guarantee a difference.
+            if (testZoneNow.getHour() == systemNow.getHour()) {
+                testZone = ZoneId.of("Asia/Tokyo");
+                testZoneNow = ZonedDateTime.now(testZone);
+            }
+
+            // Create cron for the current hour/minute in test zone
+            // Format: "sec min hour day month ? year"
+            var cronExpression = String.format("* %d %d %d %d ? %d",
+                    testZoneNow.getMinute(),
```
danielkec: I think there is an RC each time timeZoneNow is on top of a minute

denyshorman: Good catch on the minute rollover. I scrapped the previous approach entirely. Moved to a forward-looking calculation.

Issue #10845, opened 2025-11-20
Helidon SE Config Bridge to Helidon MP


klustria:
Helidon Version: 4.x

## Enhancement Description
Request that a Helidon Declarative Config Source can be automatically propagated to MP Config. This is to provide convenience for creating just one module for a Custom Config Source that would work on both Helidon SE and MP.

Created a custom SE meta-config, but the config did not get propagated in a Helidon MP application.

klustria:
Resolved via PR https://github.com/helidon-io/helidon/pull/11029.

PR #11029, opened 2026-01-15
Allow service registry injected ConfigSources to work in Helidon MP

klustria:
### Description
Allow service registry injected ConfigSources to work in Helidon MP

### Documentation

If no doc impact: None

### DIFF
--- a/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigBuilder.java
+++ b/config/config-mp/src/main/java/io/helidon/config/mp/MpConfigBuilder.java
@@ -365,7 +365,7 @@ class MpConfigBuilder implements Builder<MpConfigBuilder, Config>, ConfigBuilder
                         .forEach(source -> targetConfigSources.add(new OrdinalSource(source))));
     }
 
-    private void addServiceRegistrySources(List<OrdinalSource> targetConfigSources) {
+    private static void addServiceRegistrySources(List<OrdinalSource> targetConfigSources) {
         Services.all(io.helidon.config.spi.ConfigSource.class)
                 .forEach(it ->
                     targetConfigSources.add(new OrdinalSource(MpConfigSources.create(it)))


config/config-mp/src/main/java/io/helidon/config/mp/MpConfigBuilder.java
```diff
@@ -363,6 +365,14 @@ private void addDiscoveredSources(List<OrdinalSource> targetConfigSources) {
                         .forEach(source -> targetConfigSources.add(new OrdinalSource(source))));
     }
 
+    private void addServiceRegistrySources(List<OrdinalSource> targetConfigSources) {
+        Services.all(io.helidon.config.spi.ConfigSource.class)
+                .forEach(it -> {
+                    System.out.println("addServiceRegistrySources configsource: " + it.description());
```
ljnelson: Missed a `System.out.println` I guess

klustria: Thanks for catching this @ljnelson. I thought I removed them all.  In any case, I've updated the code to remove it.

klustria:
Resolves https://github.com/helidon-io/helidon/issues/10845

config/config-mp/src/main/java/io/helidon/config/mp/MpConfigBuilder.java
```diff
@@ -288,6 +289,7 @@ public Config build() {
         if (useDiscoveredSources) {
             addDiscoveredSources(ordinalSources);
         }
+        addServiceRegistrySources(ordinalSources);
```
ljnelson: I wonder: are service registry sources "discovered"? Should this line move up to be protected by the `useDiscoveredSources` guard?

klustria: Discussed with @tomas-langer and he agreed with your assessment, so I made the change as you've suggested. Thanks @ljnelson for that feedback.

config/config-mp/src/main/java/io/helidon/config/mp/MpConfigBuilder.java
```diff
@@ -363,6 +365,13 @@ private void addDiscoveredSources(List<OrdinalSource> targetConfigSources) {
                         .forEach(source -> targetConfigSources.add(new OrdinalSource(source))));
     }
 
+    private void addServiceRegistrySources(List<OrdinalSource> targetConfigSources) {
```
ljnelson: Can this be `static`?

klustria: I think it can be, but then we will have to change other similar methods right beside it to make them all be also static so there is consistency, which I prefer not to do in this change as it makes backport. Is there any strong argument though for doing that?

ljnelson: If there is no reference to internal instance state, then it is the Right Thing To Do™. I'm generally and personally in favor of making such targeted "right thing to do" fixes in PRs, so long as it doesn't get out of hand. In terms of: "does this fix a bug", no, there's no immediate motivation.

klustria: Pushed a new change to make addServiceRegistrySources() as a static method.

ljnelson: Hmm, the commit seems to have made `addDiscoveredConverters` `static` which also needs to be done, but didn't change `addServiceRegistrySources`

klustria: Thanks for catching that. I extended the static change to `addServiceRegistrySources()` this time.

PR #11044, opened 2026-01-16
Add support for OpenTelemetry metrics configuration using Helidon config

### LINKED ISSUE #11041: Allow configuration of OpenTelemetry metrics using Helidon config
Helidon Version: 4.x

## Enhancement Description
Building on the earlier enhancement to allow config of OTel _tracing_ using Helidon config under `telemetry`, add the ability to configure OTel `metrics` under `telemetry.signals.metrics`.

tjquinno:
### Description
Resolves #11041 

#### Release Note
____
Helidon now allows users to set up Helidon configuration, both programmatically and under the `telemetry.signals.metrics` config key, to control the behavior of OpenTelemetry metrics. 

(This augments the previously-releasee enhancement which added support for OpenTelemetry tracing config under `telemetry.signals.tracing`.)
____

#### PR Overview
The PR adds quite a few new types to `helidon-telemetry-otel-config`.

There are three general categories of additions:
* Enums which correspond to discrete choices for OpenTelemetry singletons for various types.
* Config blueprints for the key new config types
* Config support classes with various config factory methods to create OTel types from Helidon config (typically using OTel builders of the corresponding type) and some other methods.

Note that issue this PR resolves describes only configuration of OpenTelemetry's metrics behavior. This PR _does not_ create an implementation of the Helidon metrics API using OpenTelemetry metrics. 
### Documentation
Includes updates to the telemetry doc page.

### DIFF
--- a/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
+++ b/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


docs/src/main/asciidoc/se/open-telemetry.adoc
```diff
@@ -431,6 +444,120 @@ The link:https://opentelemetry.io/docs/languages/java/sdk/#sampler[OpenTelemetry
...
+            name: counter-selector
+            type: counter
+            meter-name: my-counter
+
+----
+<1> Introduces the metrics configuration.
+<2> Introduces the first metric exporter (with name `exp-1`).
+<3> Indicates to accumulate measurement values since the previous transmission.
+<4> Prescribes to aggregate histograms for transmission using the OpenTelemetry `BASE2_EXPONENTIAL_BUCKET_HISTOGRAM` technique with the specified maximum number of buckets and maximum scale.
+<5> Introduces the second metric exporter (with nane `exp-2`).
```
rhomast: Typo: with name `exp-2`

tjquinno: Good catch. Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/AggregationConfigBlueprint.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.telemetry.otelconfig;
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+@Prototype.Configured
```
tomas-langer, line 25: Missing javadoc on both the class and the method. This will end up in a public class that is not correctly documented.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/TypedAttributes.java
```diff
@@ -0,0 +1,79 @@
...
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+import io.opentelemetry.api.common.AttributesBuilder;
+
+/**
+ * Abstraction of typed attributes settable on OpenTelemetry elements.
+ */
+@Prototype.Configured
```
tomas-langer: This annotation will be ignored.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/ViewRegistrationConfigBlueprint.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    @Option.Configured
+    Optional<Predicate<String>> attributeFilter();
+
+    /**
+     * Instrument selector, configurable using {@link io.helidon.telemetry.otelconfig.InstrumentSelectorConfig}.
+     *
+     * @return instrument selector
+     */
+    @Option.Configured
```
tomas-langer, line 73: It may be useful to explain how this works in config - there is a custom method, so the details are hidden and not documented anywhere

tjquinno: After an off-line discussion I am pausing this PR to await enhancements in the config doc tooling that might affect this PR.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/AggregationConfigBlueprint.java
tjquinno: Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/TypedAttributes.java
tjquinno: removed

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/ViewRegistrationConfigBlueprint.java
tjquinno: It turns out that the config for OTel metrics in this PR is not a good fit for the recent enhancement to config and blueprints for 3rd-party types, which is what I was pausing for. 

In most of the OTel config here, the code needs to check the `type` (as with exporters, readers, and instrument-selector) or the `protocol` (for exporters of type `otlp`) to know how to use the config settings to manufacture the correct OpenTelemetry object. The added 3rd-party type support works nicely if the config settings map more statically to the 3rd-party type; sadly that's not the case for most of these new config blueprints.

Re: adding to the blueprint Javadoc explaining the lower-level config settings... 

I had previously added Javadoc `{@link}`s to the Javadoc for these cases leading to the generated config class. Readers could use this to browses the config reference page for that type to find the config table. The links do not render as links in the generated config tables. For this reason, I had also already made sure the documentation page  includes the generated config doc for each nested section, not just the high level. The documentation page describes fully what config settings are available for each section.

I am reluctant to embed further hand-written descriptions of the config settings in the Javadoc because, eventually, the Javadoc `@{link}` directives should be made to work on the rendered pages, and further we do not want to replicate this information in hand-written form in case of future changes.

Issue #11058, opened 2026-01-23
4.x Filters invoked twice when Http1DirectJunitExtension


jbescos:
## Problem Description
Filters are invoked two times because this [line](https://github.com/helidon-io/helidon/blob/bc50411bebe8f40eb31b938ed40a6990272065b4/webserver/webserver/src/main/java/io/helidon/webserver/http/HttpRoutingImpl.java#L238) adds an instance on build.

And [here](https://github.com/helidon-io/helidon/blob/bc50411bebe8f40eb31b938ed40a6990272065b4/webserver/testing/junit5/junit5/src/main/java/io/helidon/webserver/testing/junit5/Http1DirectJunitExtension.java#L177) and [here](https://github.com/helidon-io/helidon/blob/bc50411bebe8f40eb31b938ed40a6990272065b4/webserver/testing/junit5/junit5/src/main/java/io/helidon/webserver/testing/junit5/Http1DirectJunitExtension.java#L188) it invokes #build.

Fix is to use a copy of the builder.

Update:
The copy of the builder does not help either, because the [copy](https://github.com/helidon-io/helidon/blob/f99ff6ffc868c9f103d70a5d90bca5c3efb21d67/webserver/webserver/src/main/java/io/helidon/webserver/http/HttpRoutingImpl.java#L308) just passes the same reference of features. This issue potentially affects production code, not only the test.

## Steps to reproduce

- Create a ServerFeature registered through Java ServiceLoader that adds a filter
- Create a test using @RoutingTestinvoke a webserver endpoint
- You can see that the filter was invoked twice

PR #11059, opened 2026-01-23
4.x Filters invoked twice when Http1DirectJunitExtension

jbescos:
### Description

Solves https://github.com/helidon-io/helidon/issues/11058

Description in it


romain-grecourt:
This issue applies to all usages of WebServer, there isn't anything specific to Http1DirectJunitExtension right ?

jbescos:
> This issue applies to all usages of WebServer, there isn't anything specific to Http1DirectJunitExtension right ?

Potentially yes, but it is only seen so far only in Http1DirectJunitExtension, because it depends on the usage you make with the builder. In the original issue you have the links to these places.

PR #11068, opened 2026-01-26
Correctly flush data over network from Jersey output stream.

### LINKED ISSUE #9183: 4.0.6: Chunked transfer no longer sends data over network on Flush (MP)
## Environment Details
* Helidon Version: 4.0.6 onwards
* Helidon MP
* JDK version:  openjdk-22.0.2
* OS: Windows

----------

## Problem Description
Creating a Rest endpoint in java that attempts to chunk back results to a client, doesn't come back chunked but comes back with one single result.
Using Helidion 4.0.6 this problem is easily reproducible.

## Steps to reproduce
# Step:
Compile Programs like the following and run in Helidon Mp contained 4.0.6:-

```java
import jakarta.enterprise.context.ApplicationScoped;
import jakarta.ws.rs.GET;
import jakarta.ws.rs.Path;
import jakarta.ws.rs.Produces;
import jakarta.ws.rs.core.MediaType;
import jakarta.ws.rs.core.Response;
import jakarta.ws.rs.core.StreamingOutput;

import java.io.BufferedWriter;
import java.io.OutputStreamWriter;
import java.io.Writer;

@Path("/test/")
@ApplicationScoped
public class StreamResource {

    @GET
    @Path(value = "/stream")
    @Produces(MediaType.TEXT_PLAIN)
    public Response stream() {

            StreamingOutput stream = output -> {
                Writer writer = new BufferedWriter(new OutputStreamWriter(output));
                for (int i = 0; i < 10; i++) {
                    writer.write("This is the " + i + "th input for the result stream." + "\n");
                    writer.flush();
                    try {
                        Thread.sleep(2000);
                    }
                }

            };
            return Response.ok(stream).build();
        }

}
```
#####
call  endpoint from client, like the following: curl -N -v "http://localhost:8080/api/v1/test/stream"

pre 4.0.6, results will comeback in chunks, one line at a time, 4.0.6 and above results come back in one single chunk

tomas-langer:
Resolves #9183 

### Description

Jersey now fixed a bug where the flushing stream was ignored, so now this works as expected - we can call `flush()` on a JAX-RS output stream and data will be written over the network (and it will trigger chunked encoding, unless content-length is set).

Added tests.

### DIFF
--- a/jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
+++ b/jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
@@ -599,10 +599,9 @@ public class JaxRsService implements HttpService {
             }
             ensureDelegate();
             if (position == 0 && !written) {
-                // no op if nothing was written
+                delegate.close();
                 return;
             }
-            ensureDelegate();
             delegate.write(buffer, 0, position);
             written = true;
             position = 0;
--- a/jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
+++ b/jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@ import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 
 import io.helidon.common.config.Config;
@@ -53,6 +54,7 @@ import io.helidon.webserver.http.ServerRequest;
 import io.helidon.webserver.http.ServerResponse;
 
 import jakarta.ws.rs.NotFoundException;
+import jakarta.ws.rs.RuntimeType;
 import jakarta.ws.rs.core.Application;
 import jakarta.ws.rs.core.GenericType;
 import jakarta.ws.rs.core.Response;
@@ -61,6 +63,7 @@ import org.glassfish.jersey.CommonProperties;
 import org.glassfish.jersey.internal.MapPropertiesDelegate;
 import org.glassfish.jersey.internal.inject.InjectionManager;
 import org.glassfish.jersey.internal.util.collection.Ref;
+import org.glassfish.jersey.io.spi.FlushedCloseable;
 import org.glassfish.jersey.server.ApplicationHandler;
 import org.glassfish.jersey.server.ContainerException;
 import org.glassfish.jersey.server.ContainerRequest;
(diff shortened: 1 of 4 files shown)


jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -394,11 +404,21 @@ public OutputStream writeResponseStatusAndHeaders(long contentLengthParam,
             Response.StatusType statusInfo = containerResponse.getStatusInfo();
             res.status(Status.create(statusInfo.getStatusCode(), statusInfo.getReasonPhrase()));
 
-            if (contentLength > 0) {
-                res.header(HeaderValues.create(HeaderNames.CONTENT_LENGTH, String.valueOf(contentLength)));
-            }
             // in case there is an exception during close operation, we would lose the information and wait indefinitely
-            this.outputStream = new ReleaseLatchStream(cdl, res.outputStream());
+            // sp we ise ReleaseLatchStream
```
spericas, line 408: Couldn't decode this comment

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -497,6 +522,99 @@ public void close() throws IOException {
...
+        }
+
+        @Override
+        public void flush() throws IOException {
+            if (position == 0 && !written) {
+                // no op if nothing was written
+                return;
+            }
+            ensureDelegate();
+            delegate.write(buffer, 0, position);
```
spericas, line 589: This write() may be called with position == 0. Should it be guarded?

jersey/webserver/src/test/java/io/helidon/jersey/webserver/BufferingOutputStreamTest.java
```diff
@@ -0,0 +1,92 @@
...
+        assertThat(baos.size(), is(4));
+        o.write(Bytes.SEMICOLON_BYTE);
+        assertThat(baos.size(), is(4));
+        o.flush();
+        o.close();
+
+        assertThat(baos.toString(), is("::  ;"));
+        assertThat(contentLength.get(), nullValue());
+    }
+    @Test
```
spericas, line 64: No new line between tests

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -497,6 +522,99 @@ public void close() throws IOException {
...
+        public void close() throws IOException {
+            if (!written) {
+                contentLengthConsumer.accept(position);
+            }
+            ensureDelegate();
+            if (position == 0 && !written) {
+                delegate.close();
+                return;
+            }
+            delegate.write(buffer, 0, position);
```
spericas, line 605: Same here

microprofile/tests/server/src/test/java/io/helidon/microprofile/tests/server/FlushStreamTest.java
```diff
@@ -0,0 +1,224 @@
...
+    @Test
+    void testUnFlushedStream(WebTarget target) {
+        Response response = target.path("/test/stream/unflushed")
+                .request()
+                .get();
+
+        assertThat(response.getStatus(), is(Status.OK_200_CODE));
+
+        assertThat(response.getHeaderString(HeaderNames.TRANSFER_ENCODING_NAME),
+                   nullValue());
```
spericas, line 87: Will there a content-length here then?

tomas-langer: There should be, but the test was badly written, and after fixing the test, I found out it still does not work as expected...

tomas-langer:
Closing the pull request as there is no way to fix this in Helidon right now.

PR #11072, opened 2026-01-27
[4.x] Merged config

danielkec:
### Description
Creates MegedConfig which can reach to fallback config when property is not present in the primary config.

### DIFF
--- a/config/config/src/main/java/io/helidon/config/MergedConfig.java
+++ b/config/config/src/main/java/io/helidon/config/MergedConfig.java
@@ -29,7 +29,7 @@ import io.helidon.config.spi.ConfigMapper;
 
 /**
  * Configuration that merges two {@code Config} instances, primary and fallback.
- * When property is not present in the primary configuration, fallback is used.
+ * When a property is not present in the primary configuration, fallback is used.
  * <p>
  * When a key is present in both the {@code primary} and {@code fallback} configurations,
  * the value from the {@code primary} configuration takes precedence.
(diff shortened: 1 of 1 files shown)


config/config/src/main/java/io/helidon/config/MergedConfig.java
```diff
@@ -0,0 +1,242 @@
...
+                                         }
+                                         return c;
+                                     }),
+                             fallbackDelegate.traverse(predicate)
+                                     .filter(c -> !existingKeys.contains(c.key())));
+    }
+
+    @Override
+    public <T> T convert(Class<T> type, String value) throws ConfigMappingException {
+        try {
```
tomas-langer: Same as mapper - just use primary

config/config/src/main/java/io/helidon/config/MergedConfig.java
```diff
@@ -0,0 +1,242 @@
...
+     *
+     * @param primary  the primary configuration, which has precedence in the case of duplicate keys
+     * @param fallback the fallback configuration, queried for keys not found in the primary one
+     * @return a merged {@code Config} instance
+     */
+    public static Config create(Config primary, Config fallback) {
+        return new MergedConfig(primary, fallback);
+    }
+
+    private MergedConfig(Config primary, Config fallback) {
```
tomas-langer: Constructor should be before methods.

config/config/src/main/java/io/helidon/config/MergedConfig.java
danielkec: thx, fixed

config/config/src/main/java/io/helidon/config/MergedConfig.java
danielkec: thx, fixed

PR #11074, opened 2026-01-27
JSON generator fixes

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11079
Fixes https://github.com/helidon-io/helidon/issues/11080

### Documentation

None

### DIFF
--- a/http/media/json/src/main/java/io/helidon/http/media/json/HelidonJsonSupportConfigBlueprint.java
+++ b/http/media/json/src/main/java/io/helidon/http/media/json/HelidonJsonSupportConfigBlueprint.java
@@ -20,7 +20,6 @@ import io.helidon.builder.api.Option;
 import io.helidon.builder.api.Prototype;
 import io.helidon.http.media.spi.MediaSupportProvider;
 import io.helidon.json.binding.JsonBinding;
-import io.helidon.json.binding.JsonBindingConfig;
 
 /**
  * Configuration blueprint for Helidon JSON media support.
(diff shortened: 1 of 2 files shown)


http/media/json/src/main/java/io/helidon/http/media/json/HelidonJsonSupportConfigBlueprint.java
```diff
@@ -43,15 +42,6 @@ interface HelidonJsonSupportConfigBlueprint extends Prototype.Factory<HelidonJso
...
     String name();
 
-    /**
-     * Configuration of the Helidon JSON Binding component.
-     *
-     * @return json binding configuration
-     */
-    @Option.DefaultMethod("create")
-    @Option.Configured(merge = true)
-    JsonBindingConfig jsonBindingConfig();
```
tomas-langer, line 53: I do not understand why this was removed? It seems we may want to configure the instance...

I would rather change the `HelidonJsonSupport.Decorator` to:
```java
    static class Decorator implements Prototype.BuilderDecorator<HelidonJsonSupportConfig.BuilderBase<?, ?>> {

        @Override
        public void decorate(HelidonJsonSupportConfig.BuilderBase<?, ?> target) {
            if (target.jsonBinding().isEmpty()) {
                target.jsonBinding(JsonBinding.create(target.jsonBindingConfig()));
            }
        }
    }
```

tomas-langer:
Also the name of the PR is misleading, as you are doing much more.
This PR is missing an issue it fixes...

http/media/json/src/main/java/io/helidon/http/media/json/HelidonJsonSupportConfigBlueprint.java
Verdent: Because there is nothing to configure from the config in this configuration.

Verdent:
Issues created and title changed

http/media/json/src/main/java/io/helidon/http/media/json/HelidonJsonSupportConfigBlueprint.java
Verdent: One can easily create the whole instance of `JsonBinding` if they need to register some special converters, since this is only usable at runtime.

PR #11075, opened 2026-01-27
Update gRPC reflection service to support imports

spericas:
### Description

Update gRPC reflection service (v1 and v1alpha) to correctly support file queries for imported proto files. New tests. See issue #11076.

### Documentation

None


webserver/tests/grpc/src/test/java/io/helidon/webserver/tests/grpc/ReflectionServiceTest.java
```diff
@@ -151,7 +164,7 @@ void testExtension() throws InterruptedException {
                                            .build())
                            .build());
         req.onCompleted();
-        res.await(500, TimeUnit.SECONDS);
+        res.await(10, TimeUnit.SECONDS);
```
tjquinno, line 167: For future reference we have a `MatcherWithRetry` which is tailor-made to help with situations like this.

PR #11088, opened 2026-01-29
 4.3.4: Fix issues with trailers in HTTP/2

### LINKED ISSUE #10996: HTTP/2 stream not correctly closed when using trailers and stream
## Environment Details
* Helidon Version: 4.x
* Helidon SE & MP

----------

## Problem Description
When a request contains `TE: trailers` and an output stream is used on the server side (i.e. when using JAX-RS in Helidon MP), we send the data without "end-of-stream" flag, but fail to follow up with the expected trailer with "end-of-stream" flag set.
This results in a stream that is not closed, even though all the data was sent.
This can be verified using `curl` with:

```shell
curl -i -vvv --http2 --insecure -H "TE: trailers" https://localhost:8443/no-end-of-stream
```

If the `TE: trailers` header is removed, this does not fail. The `-vvv` is to enable full verbosity, where the `eos` (end of stream) flag can clearly be seen.

barchetta:
Backport #10997 to Helidon 4.3.4

- Header now uses the same separator for creating and parsing a single value (comma and a space)
- Http2Headers now send all values, not just he first one 
- Trailers are considered available only if they are set in the "trailer" response header, otherwise stream ends with data 
- Updated tests to ensure we get the trailers as expected, and all trailer names as well
- Sending stream result OK when using streaming, as we announced the trailer in trailer header

Resolves #10996 

The original issue was caused by our implementation not sending trailers when we had a single pass-through in the output stream.
Additional issues discovered when troubleshooting the issue and adding tests.

### DIFF
--- a/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/CutConnectionTest.java
+++ b/webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/CutConnectionTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.


barchetta:
Test failure:
```
Error:  io.helidon.webserver.tests.http2.CutConnectionTest.testStringRoute  Time elapsed: 0.061 s  <<< FAILURE!
java.lang.AssertionError: 

Expected: is <FINE>
     but: was <FINER>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
	at io.helidon.webserver.tests.http2.CutConnectionTest.testStringRoute(CutConnectionTest.java:118)
```

tomas-langer (approved):

PR #11097, opened 2026-01-29
Move MP config encryption from SE module.

### LINKED ISSUE #11096: 4.3.3 SE service is getting errors re: io.helidon.config.mp.spi from io.helidon.config.encryption
## Environment Details
* Helidon Version: 4.3.3
* Helidon SE 
* JDK version: 21.x
* OS: OL9

----------

## Problem Description

Error occurred during initialization of boot layer
java.lang.module.ResolutionException: Module io.helidon.config.encryption does not read a module that exports io.helidon.config.mp.spi

See here:
[INFO]    \- io.helidon.config:helidon-config-encryption:jar:4.3.3:compile
[INFO]       +- io.helidon.common:helidon-common-key-util:jar:4.3.3:compile
[INFO]       \- io.helidon.common:helidon-common-crypto:jar:4.3.3:compile

and yet config.mp is required.

See also:
https://github.com/helidon-io/helidon/blob/helidon-4.3.x/config/encryption/src/main/java/module-info.java
where line 30 has "requires static io.helidon.config.mp"
And line 39 where it provides a mp config filter.

## Steps to reproduce
Try to use helidon-config-encryption in a JPMS enabled module for Helidon SE 4.3.3.

tomas-langer:
### Description
Resolves #11096 

- remove dependency on MP APIs from `helidon-config-encryption`, as service implementation implies module dependency (i.e. `requires static` does not work)
- move MP config encryption to `helidon-config-mp` module, to keep backward compatibility (i.e. we do not need an additional module, and it is anyways included by `helidon-microprofile-config`)

### DIFF
--- a/config/config-mp/etc/spotbugs/exclude.xml
+++ b/config/config-mp/etc/spotbugs/exclude.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2021, 2022 Oracle and/or its affiliates.
+    Copyright (c) 2021, 2026 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -32,6 +32,11 @@
         <Class name="io.helidon.config.mp.MpConfigBuilder" />
         <Method name="toPath" />
         <Bug pattern="PATH_TRAVERSAL_IN" />
+    </Match>
+    <Match>
+        <!-- Path from config or code -->
+        <Class name="io.helidon.config.mp.MpEncryptionFilter" />
+        <Bug pattern="PATH_TRAVERSAL_IN"/>
     </Match>
     <Match>
         <!-- We need to re-throw NPEs and IllegalArgumentExceptions, rest is handled -->
--- a/config/config-mp/pom.xml
+++ b/config/config-mp/pom.xml
@@ -55,6 +55,10 @@
             <artifactId>helidon-common-configurable</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-key-util</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.helidon.config</groupId>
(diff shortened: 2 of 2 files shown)


config/config-mp/src/main/java/io/helidon/config/mp/MpEncryptionFilter.java
```diff
@@ -14,15 +14,22 @@
  * limitations under the License.
  */
 
-package io.helidon.config.encryption;
+package io.helidon.config.mp;
```
romain-grecourt, line 17: `io.helidon.config.encryption.MpEncryptionFilter` is a public class and is now moved to `io.helidon.config.mp.MpEncryptionFilter`.

This is not backward compatible, we should make sure this is documented.

tomas-langer: is release notes sufficient? Or should I change something in the sources?

tomas-langer: Updated the PR - keeping the old class in with no body, and explanation and pointer.

romain-grecourt: If the class does not implement `MpConfigFilter`, it's still incompatible.

The first iteration of the PR was fine, if we are okay we introducing a backward incompatibility in a dot-dot release.

Otherwise, we have to have config-encryption depend on config-mp. (I.e `require io.helidon.config.mp` and non optional dependency).

tomas-langer: That brings in the MP Config API dependency, which is unfortunate in a Helidon SE application. 

The only possible issue is if somebody explicitly uses this class. 
Filters are ONLY loaded using ServiceLoader - i.e. there is no API where you can register such a filter. 
As the class is final, users should never have a reference to it in their code.

I think this makes our decision a bit easier and we should introduce this (slightly) backward incompatibility to fix a dependency tree issue.

romain-grecourt (approved):

PR #11110, opened 2026-02-01
SE Testing with config block

danielkec:
Mirrors MP `@ConfigBlock` [feature](https://helidon.io/docs/v4/mp/testing/testing#_configuration) to SE config.

## Description
This PR adds configuration block annotations for Helidon SE testing, mirroring the existing MicroProfile @ConfigBlock feature. This enables developers to inject test configurations inline using YAML or properties format, both at class and method levels.

## Changes:

Added `@AddConfigBlock`, `@AddConfigBlocks`, and `@Configuration` annotations for SE testing configuration
Implemented TestConfigFactory to process configuration blocks and merge them with existing configurations
Enhanced TestJunitExtension to support `@Service.Named` parameter resolution and per-method configuration triggers.

```java
Testing.Test
@Testing.AddConfigBlock(value = """
        test-a:
            test-c: value-c-in-parent-block
            test-d: value-d-in-parent-block
            test-e: value-e-in-parent-block
        """, type = "yaml")
class TestConfigBlockPerMethod {

    @Testing.Configuration(configSources = "custom-config-a.yaml")
    @Testing.AddConfigBlock(value = """
            test-a.test-d=value-d-in-method-block
            """)
    @Test
    void overrideExistingAndParentBlock(Config c) {
...
```

### DIFF
--- a/testing/junit5/src/main/java/io/helidon/testing/junit5/Testing.java
+++ b/testing/junit5/src/main/java/io/helidon/testing/junit5/Testing.java
@@ -76,7 +76,7 @@ public final class Testing {
     @Repeatable(AddConfigBlocks.class)
     public @interface AddConfigBlock {
         /**
-         * Specifies the configuration format. Possible values are: 'yaml' and `properties`.
+         * Specifies the configuration format. Possible values are: 'yaml' and 'properties'.
          * <p>
          * The default format is 'properties'
          *
@@ -111,7 +111,7 @@ public final class Testing {
      * }
      * </pre>
      *
-     * @see AddConfigBlocks
+     * @see AddConfigBlock
      */
     @Inherited
     @Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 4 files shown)


romain-grecourt:
The MP annotations filled a gap where there the JUnit lifecycle hooks weren't usable and the API to produce MP config programmatically is arguably not graceful.

---

In SE, we should make sure that JUnit lifecycle hooks can be used to specify config using the service registry:
```java
@BeforeAll
static void setUp() {
    Services.set(Config.class, Config.just(ConfigSources.create("""
            foo: bar
            """, MediaTypes.APPLICATION_YAML)));
}
```

The current SE JUnit extensions are flawed and don't support the code above correctly, they should.

---

If we want to introduce annotations for declaring test configuration in SE, it should be for style only, not functional.

The `@Configuration` annotation differs from the other ones for legacy reasons, it shouldn't be mirrored.

Instead, we should probably de-compose it with `@AddXX` annotations:

E.g.
```java
@AddConfigSource("foo.yaml")
@AddMetaConfig
```

PR #11113, opened 2026-02-03
Update builders to support third party types as runtime types.

### LINKED ISSUE #11104: Support third party types as "RuntimeType" from builders
Helidon Version: 4

## Enhancement Description

We support the concept of "builder -> prototype -> runtime type", where method `build` of both the builder and prototype creates the `runtime type`, and a method `buildPrototype()` on builder creates the prototype.

It would be nice to extend this to "builder -> prototype -> third party runtime type" which is currently not supported.

Example: 
Our `Keys` prototype supports building `PrivateKey`. We could design a `PrivateKey` prototype that would buiuld `PrivateKey` (a Java type).
This would allow us to generate config reference which would show the configuration options of a private key (now it just shows the type `PrivateKey`, as the mapping is hiddent behind a custom method).

Proposed design:
1. Blueprint of prototype that can build the third party runtime type:
```java
@Prototype.Blueprint
@Prototype.Configured
@Prototype.FactoryType(PrivateKey.class)
interface PrivateKeyConfigBlueprint extends Prototype.Factory<PrivateKey> {
}
```

- the  `FactoryType` annotation links this to a class with static method `PrivateKey create(PrivateKeyConfig)`

2. Blueprint that uses `PrivateKey` as a configured option:
```java
@Prototype.Blueprint
@Prototype.Configured
interface TlsConfigBlueprint {
    @Option.Configured
    PrivateKey privateKey();
}
```
- we will look into the current package (of `TlsConfigBlueprint`) if there is a `PrivateKeyConfig` type that matches the requirements (is a `Prototype.Factory<PrivateKey>`)

3.  Blueprint that uses `PrivateKey` as a configured option from a different module:
```java
@Prototype.Blueprint
@Prototype.Configured
interface TlsConfigBlueprint {
    @Option.Configured
    @Option.Prototype(PrivateKeyConfig.class)
    PrivateKey privateKey();
}
```
- as we cannot discover the `PrivateKeyConfig` from the `PrivateKey` class easily, we need to tell the blueprint which prototype to use

4. The resulting prototype that uses the third party type
(regardless whether this is the same module, or referenced)

The following methods will be added to the builder:
- `privateKey(PrivateKey)` - existing behavior
- `privateKey(PrivateKeyConfig)` - new method, invoking `privateKey(pkConfig.build())`
- `privateKey(Consumer<PrivateKeyConfig.Builder>) - new method

And the `config(Config)` method of the builder will use the following (or similar):
```java
config.get("private-key").as(PrivateKeyConfig::create).ifPresent(this::privateKey);
```

5. The generated config metadata JSON

The type mentioned in the JSON will be the `PrivateKeyConfig` type, so we can generate a nice reference to it from our config reference pages, and autocomplete can work in IDEs.

tomas-langer:
### Description
Resolves #11104 

Updated documentation.
Updated config metadata documentation to correctly refernece such types (a few other improvements added as the result did not look nice when the page contained `java_lang_System_Logger` instead of `java.lang.System.Logger` for third party types, or when a type in config reference document lacked the containing class and showed `Logger`  instead of `System.Logger`).

This PR adds capability to create prototypes that build a third party type. 
Such a prototype can then be used from other prototypes to add nice setters, and to support reading from configuration.

Example for `System.Logger`:

Prototype that is a factory:
```java
@Prototype.Blueprint
@Prototype.Configured
@Prototype.CustomMethods(LoggerConfigSupport.class)
interface LoggerConfigBlueprint extends Prototype.Factory<System.Logger> {
    /**
     * Name of the logger.
     *
     * @return logger name
     */
    @Option.Configured
    String name();
}
```

The required custom methods to enable this (a factory method from prototype to runtime type):
```java

class LoggerConfigSupport {
    @Prototype.RuntimeTypeFactoryMethod
    static System.Logger createLogger(LoggerConfig config) {
        return System.getLogger(config.name());
    }
}
```

This already allows us:

```java
System.Logger logger = LoggerConfig.create(config.get("my-logger")).build();
logger = LoggerConfig.builder()
      .name("my.logger.Logger")
      .build();
```

The second advantage of this approach is when we want to reference this type from another prototype:
```java
@Prototype.Blueprint
@Prototype.Configured
interface UsingConfigBlueprint {
    @Option.Configured
    String stringOption();

    @Option.Configured
    @Option.PrototypedBy("LoggerConfig")
    System.Logger logger();
}
```

As we can see, we use the runtime type on the builder, and we tell it that the associated prototype is `LoggerConfig` in the same package (if the package is different, fully qualified class name must be provided - this is because types generated in the same annotation processing round resolve to an error in annotation values).

The generated prototype builder will have the following methods:
```java
public BUILDER logger(System.Logger logger);
public BUILDER logger(Consumer<LoggerConfig.Builder> consumer);
public BUILDER logger(LoggerConfig logger)
```

Allowing setting an explicit instance or the configuration prototype.
The config method will use the following:

```java
config.get("logger").as(LoggerConfig::create).map(LoggerConfig::build).ifPresent(this::logger);
```

allowing us to seamlessly integrate third party runtime type into our configuration system using a "config" prototype

### DIFF
--- a/docs/src/main/asciidoc/se/builder.adoc
+++ b/docs/src/main/asciidoc/se/builder.adoc
@@ -310,7 +310,7 @@ Creates a runtime type from a consumer to configure the prototype builder.
 The blueprint must include the `@Prototype.Blueprint` annotation and extend `Prototype.Factory<RuntimeType>`.
 
 
-Note that requirements 3 - 5 can be omitted when using a third party runtime type (i.e. `PrivatgeKey`), in such a case the setup should be:
+Note that requirements 3 - 5 can be omitted when using a third party runtime type (i.e. `PrivateKey`), in such a case the setup should be:
 
 1. Create a blueprint that extends `Prototype.Factory<RuntimeType>` (same as above)
 2. Add a custom method annotated with `@Prototype.RuntimeTypeFactoryMethod` (custom methods can be configured on blueprint using `@Prototype.CustomMethods`, the custom method is a static method with signature `RuntimeType create(Prototype)`), method name is arbitrary


builder/api/src/main/java/io/helidon/builder/api/Option.java
```diff
@@ -532,4 +532,24 @@ private Option() {
...
+    @Retention(RetentionPolicy.CLASS)
+    public @interface PrototypedBy {
+        /**
+         * Type of the prototype that is a factory of the type returned by the annotated method.
+         * If the type is in the same package as the type of the annotated method, class name is sufficient, otherwise
+         * a fully qualified name must be used.
+         *
+         * @return prototype class that builds the type of the annotated option
+         */
+        String value();
```
tjquinno, line 553: Can the type here be `Class` instead of `String` so the compiler can check it? IIRC that was the approach described in the issue write-up.

docs/src/main/asciidoc/se/builder.adoc
```diff
@@ -309,6 +309,14 @@ Creates a runtime type from a consumer to configure the prototype builder.
 5. *Prototype Integration:* +
 The blueprint must include the `@Prototype.Blueprint` annotation and extend `Prototype.Factory<RuntimeType>`.
 
+
+Note that requirements 3 - 5 can be omitted when using a third party runtime type (i.e. `PrivatgeKey`), in such a case the setup should be:
```
tjquinno: Typo: `PrivatgeKey` -> `PrivateKey`

builder/api/src/main/java/io/helidon/builder/api/Option.java
tomas-langer: Unfortunatelly not - the problem is that if you reference a prototype, that is in the same module, this results in an error element in the annotation value (with no way to extract the actual class name).

I think we could allow blueprint class (if in the same module) to be used here, but that may be confusing as well (and also it may be in a different package).

tjquinno: I am not an expert, but I thought this is what `AnnotationMirror` is for.

tomas-langer: You will get an annotation value that is of type `Error`. In debug, you can see it is an unresolved class described as `<any?>.Prototype`, but this value cannot be obtained through public APIs. 
So the result we get is `<error>` (or something like that). 

I guess the reason is that this class is not yet generated when compiler analyzes the source code, so it is not aware of it. And as we generate both types (the prototype and the using type) in the same annotation round, triggered by the same annotation, we get this error instead of the class name.

I am of course using the features of the annotation processors, such as `AnnotationMirror`. The exact code is:

```java
private static Map<String, Object> extractAnnotationValues(AnnotationMirror am,
                                                               Elements elements) {
        return extractAnnotationValues(elements, elements.getElementValuesWithDefaults(am));
    }

    private static Map<String, Object>
    extractAnnotationValues(Elements elements, Map<? extends ExecutableElement, ? extends AnnotationValue> values) {
        Map<String, Object> result = new LinkedHashMap<>();
        values.forEach((el, val) -> {
            String name = el.getSimpleName().toString();
            // the value we get is an error, not a class
            Object value = val.accept(new ToAnnotationValueVisitor(elements), null);
            if (value != null) {
                result.put(name, value);
            }
        });
        return result;
    }
```

tjquinno: No, I wasn't suggesting that our annotation processor wasn't using `AnnotationMirror` or wasn't using it effectively.

I was saying that I thought the intent from the JDK of `AnnotationMirror` was to allow an annotation processor to deal with such a case and I was lamenting that it does not.

ljnelson: (Just a flyby but I believe that this is exactly the case that [`MirroredTypeException#getTypeMirror()`](https://docs.oracle.com/en/java/javase/25/docs/api/java.compiler/javax/lang/model/type/MirroredTypeException.html#getTypeMirror()) was made for, as [documented in the `AnnotatedConstruct#getAnnotation(Class)` contract](https://docs.oracle.com/en/java/javase/25/docs/api/java.compiler/javax/lang/model/AnnotatedConstruct.html#getAnnotation(java.lang.Class)). So maybe in a `try` block you can catch the `MirroredTypeException` and then get the `TypeMirror`; if it's a `DeclaredType` you can get its element via `declaredType.asElement()` which in the case in question would be a `TypeElement`, which would then allow you to get its full name via `getQualifiedName()`. Maybe for some reason this won't work in this situation.)

tomas-langer: The problem is this line: 
`Object value = val.accept(new ToAnnotationValueVisitor(elements), null);`

The `AnnotationValue` is an error type (there is no exception thrown, as we do not use `Class` to obtain the annotation itself). I should get a `TypeMirror`, but the compiler is not capable of providing it.

ljnelson: I see. I know that `ErrorType` "is a" `DeclaredType` whose `asElement()` returns a `TypeElement` from which you can get a qualified name. I'm assuming you've tried that and it simply isn't filled in for some reason. I see that you [do something similar](https://github.com/helidon-io/helidon/blob/main/codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java#L193-L195) but maybe the `asElement` approach would be worth checking out (I'm not sure offhand what the specification for `TypeMirror#toString()` is in the case of `ErrorType`).

tomas-langer: it is an error that has no public methods, and `toString()` returns `<error>`

tomas-langer: Screenshot From 2026-02-05 18-25-18

tomas-langer: The type is `com.sun.tools.javac.code.Attribute.UnresolvedClass` - this should not be public API

tomas-langer: This would get me to the actual name of the class:
`((Attribute.UnresolvedClass) val).classType.asElement().getSimpleName()`

But as I said, the `com.sun` type is not part of the public API

tomas-langer:
Investigating if we can use the `.class` in annotation, as that would be much cleaner for users.

builder/api/src/main/java/io/helidon/builder/api/Option.java
ljnelson: A shame. See https://bugs.openjdk.org/browse/JDK-8229535. So the question is: do we design an API properly that relies on an internal hack, hoping the bug will be fixed, or do we recognize the bug as a fact of the world and compromise on the API design? Drawbacks either way.

tomas-langer: I cannot get to the value even using reflection, as I get this exception:

```
java.lang.IllegalAccessException: class io.helidon.codegen.apt.AptAnnotationFactory cannot 
access class com.sun.tools.javac.code.Attribute$UnresolvedClass (in module jdk.compiler) because 
module jdk.compiler does not export com.sun.tools.javac.code to unnamed module @4768b95c
```

ljnelson: Right; integrity by default

tomas-langer: I will merge this PR, so I unblock Tim. If we find a workaround, we can create a new annotation with Class value and deprecate the current one.

PR #11132, opened 2026-02-05
Use a new profile `long-tests` for tests that use testcontainers

### LINKED ISSUE #11130: Move all long-running test-container test to a separate profile.
Helidon Version: 4.x

When running a local build, the database tests using test containers take a very long time, and always impact network on my local machine.

I would like to move these tests to a separate profile `testcontainers` that will be enabled by default, to allow disabling it when needed.

tomas-langer:
or are generally long running.

This is to allow quicker local builds.
The profile is enabled by default, so pipelines are not impacted.

Profile name is `long-tests`
To run build without it, use `mvn clean install -Ptests,-long-tests`

### Description
Resolves #11130

### DIFF
--- a/tests/integration/dbclient/pom.xml
+++ b/tests/integration/dbclient/pom.xml
@@ -39,7 +39,7 @@
             <id>long-tests</id>
             <activation>
                 <property>
-                    <name>!helidon.long-tests</name>
+                    <name>!long-tests</name>
                 </property>
             </activation>
             <modules>


data/tests/pom.xml
```diff
@@ -50,12 +50,25 @@
...
-        <module>eclipselink</module>
-        <module>hibernate</module>
     </modules>
 
+    <profiles>
+        <profile>
+            <id>long-tests</id>
+            <activation>
+                <property>
+                    <name>!helidon.long-tests</name>
```
romain-grecourt:
```suggestion
                    <name>!long-tests</name>
```

tests/integration/dbclient/pom.xml
```diff
@@ -32,10 +32,23 @@
...
-        <module>oracle</module>
-        <module>mongodb</module>
     </modules>
+
+    <profiles>
+        <profile>
+            <id>long-tests</id>
+            <activation>
+                <property>
+                    <name>!helidon.long-tests</name>
```
romain-grecourt:
```suggestion
                    <name>!long-tests</name>
```

tests/integration/jpa/pom.xml
```diff
@@ -37,9 +37,22 @@
...
-        <module>pgsql</module>
-        <module>h2</module>
     </modules>
+
+    <profiles>
+        <profile>
+            <id>long-tests</id>
+            <activation>
+                <property>
+                    <name>!helidon.long-tests</name>
```
romain-grecourt:
```suggestion
                    <name>!long-tests</name>
```

tests/integration/pom.xml
```diff
@@ -94,4 +89,23 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
+
+    <profiles>
+        <profile>
+            <id>long-tests</id>
+            <activation>
+                <property>
+                    <name>!helidon.long-tests</name>
```
romain-grecourt:
```suggestion
                    <name>!long-tests</name>
```

data/tests/pom.xml
tomas-langer: The property is intentionally called `helidon.long-tests` - as this would be specified using a system property, and I wanted to avoid naming conflicts.
The profile name is different, as profiles are managed by us.
Also I expect users to switch this profile off using the profile features of Maven, not through the system property.
The property name is a workaround to have an "always on profile"

romain-grecourt: My issue is the use of "helidon" as a prefix.

I'm not worried about conflicts. These properties are inherited from the effective pom which we control or the properties passed to the Maven command line.

How about `profiles.long-tests` instead ?

PR #11138, opened 2026-02-06
Create config from text and media type.

### LINKED ISSUE #11133: MediaType api enhancement
- Add properties `text/x-java-properties` to MediaTypes
- Add helper method to config `Config.create(String text, MediaType type)`

tomas-langer:
### Description
Resolves #11133 

### Documentation
Adds a new factory method to config that accepts string and media type.
Adds a new media type for Java properties to `MediaTypes`, so we do not need to re-create the media type over and over again

### DIFF
--- a/config/config/src/main/java/io/helidon/config/Config.java
+++ b/config/config/src/main/java/io/helidon/config/Config.java
@@ -403,7 +403,7 @@ public interface Config extends io.helidon.common.config.Config {
     }
 
     /**
-     * Create configuration from a string. The string format must match the provided media type, and a parse must be
+     * Create configuration from a string. The string format must match the provided media type, and a parser must be
      * present for that media type. {@link io.helidon.common.media.type.MediaTypes#TEXT_PROPERTIES} is always supported
      * (Java properties media type).
      * <p>


config/config/src/main/java/io/helidon/config/Config.java
```diff
@@ -348,6 +350,26 @@ static Config create(Supplier<? extends ConfigSource>... configSources) {
...
+     * @param configContent config content, such as YAML, properties, HOCON
+     * @param mediaType     media type of the provided {@code configContent}, such as
+     *                      io.helidon.common.media.type.{@link io.helidon.common.media.type.MediaTypes#APPLICATION_YAML};
+     *                      appropriate parser must be available on classpath
+     * @return a new config instance from the provided content
+     * @throws java.lang.RuntimeException in case there is no config parser for the provided media type, or if the content
+     *                                    cannot be parsed into a valid configuration
+     */
+    static Config create(String configContent, MediaType mediaType) {
+        return Config.just(ConfigSources.create(configContent, mediaType));
```
romain-grecourt: Is there supposed to be a difference between `create` and `just` ?
Isn't `create` supposed to add the default config sources as well ?

tomas-langer: I will change it. I wanted to slowly move away from `just`, as it is not aligned with our factory methods (as these should use `create`).
But in context of config, it makes a lot of sense.

config/config/src/main/java/io/helidon/config/Config.java
```diff
@@ -400,6 +402,26 @@ static Config just(Supplier<? extends ConfigSource>... configSources) {
                 .build();
     }
 
+    /**
+     * Create configuration from a string. The string format must match the provided media type, and a parse must be
```
danielkec:
```suggestion
     * Create configuration from a string. The string format must match the provided media type, and a parser must be
```

danielkec (approved):

PR #11155, opened 2026-02-13
CORS updated and declarative CORS

### LINKED ISSUE #7631: Decoupled CORS supports
Currently (as of 4.0.0-M2), the CORS support is an adhoc solution implemented in various component that need it (E.g. OpenApi).
The configuration is nested under each component.

How about having a centralized solution implemented as a filter, similar to SecurityFeature.

E.g. 
```yaml
cors:
    paths:
        - path: "/greeting[/{*}]"
          methods: ["GET"]
          allow-origins: ["http://foo.com", "http://there.com", "http://other.com"]
          allow-methods: ["PUT", "DELETE"]
```

```java
    routing.addFeature(CorsFeature.create())
           .register("/greet", GreetService::new)
}
```

### LINKED ISSUE #10726: Declarative CORS

tomas-langer:
### Description
Resolves #7631 
Resolves #10726

### Documentation

This PR deprecates existing CORS support (and the `cors` module).

The new approach:

- a single enforcement of CORS from a single server feature
- configurable (backward compatible with current version)
- additional CORS Path configurations can be added to service registry, or manually configured when setting up the `CorsFeature`

Existing APIs are mostly deprecated, the CorsPathConfig replaces `CrossOriginConfig` (as it is based on builder API).

~There is one change that should not have impact on function - we now return `204` as the "normal" response from pre-flight checks (204 is "No-Content"), as opposed to original 200 ("OK"), as 204 is a better fit.
This breaks test in examples, that I will fix together with Declarative CORS example.~

Documentation updated (may require a bit more work).
~Archetypes are broken, investigating (I want to remove deprecated stuff)~

An example snippet of the new top-level configuration:
(this is not changed, except `Max-Age` is a duration - we still support seconds, but Duration is preferred):
```yaml
cors:
  add-defaults: false # this is a new option, not to add defaults that allow all origins and methods
  paths:
    - "path-pattern": "/observe/health/*"  # protect anything under /observe/health (inclusive)
      "allow-origins": ["*"]
      "allow-methods": ["HEAD", "GET"]
      "max-age": "PT1H" # used to be just second, now can be duration string, seconds, or 11111 ms
```

The example above is our recommended configuration (and documented) even before this PR.
There is an option to define CORS per endpoint in its configuration, but we do not have that as a recommended approach already.

### DIFF
--- a/microprofile/cors/src/test/java/io/helidon/microprofile/cors/ArchetypeTest.java
+++ b/microprofile/cors/src/test/java/io/helidon/microprofile/cors/ArchetypeTest.java
@@ -91,7 +91,8 @@ public class ArchetypeTest {
                 .header("Access-Control-Request-Method", "PUT")
                 .options();
 
-        assertThat("pre-flight status", r.getStatus(), is(200));
+        // status is 204, as that is the status provided by Jersey on void methods
+        assertThat("pre-flight status", r.getStatus(), is(204));
         MultivaluedMap<String, Object> responseHeaders = r.getHeaders();
         assertThat("Header " + HeaderNames.ACCESS_CONTROL_ALLOW_METHODS_NAME,
                    r.getHeaders().get(HeaderNames.ACCESS_CONTROL_ALLOW_METHODS_NAME),


declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
```diff
@@ -0,0 +1,477 @@
...
+                              Set<String> allowedHeaders,
+                              Set<String> exposedHeaders,
+                              Boolean allowCredentials,
+                              Long maxAgeSeconds) {
+        CorsConfig() {
+            this(null, null, null, null, null, null, null);
+        }
+
+        CorsConfig(boolean defaults) {
+            this(true, null, null, null, null, null, null);
```
tjquinno: Why is the  `defaults` parameter declared and passed if it is ignored?

declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
```diff
@@ -0,0 +1,477 @@
...
+
+        // now check if we need a class level config, or if one of the option methods covers it
+        boolean wantTypeConfig = true;
+        for (CorsElement optionMethod : optionMethods) {
+            var path = optionMethod.path();
+            if (path.isEmpty()) {
+                wantTypeConfig = false;
+            } else {
+                if (path.get().equals("/")) {
+                    wantTypeConfig = false;
```
tjquinno, line 99: Once set to `false` `wantTypeConfig` can never become `true` again so these could both `break`.

docs/src/main/asciidoc/se/cors.adoc
```diff
@@ -71,75 +71,36 @@ include::{rootdir}/includes/dependencies.adoc[]
...
 
 [[se-api-routing-example]]
 === Sample Routing Setup Using the `CrossOriginConfig` API
 
 The link:{helidon-github-examples-url}/quickstarts/helidon-quickstart-se[Helidon SE Quickstart application]
  lets you change the greeting by sending a `PUT` request to the `/greet/greeting` resource.
 
-This example, based on the QuickStart greeting app, uses the low-level `CrossOriginConfig` API and
-the `CorsSupport` API to influence the xref:{rootdir}/se/webserver/webserver.adoc#routing[routing],
+This example, based on the QuickStart greeting app, uses the CORS APIto influence the xref:{rootdir}/se/webserver/webserver.adoc#routing[routing],
```
tjquinno: Typo: "APIto" -> "API to"

microprofile/cors/src/main/java/io/helidon/microprofile/cors/CorsCdiExtension.java
```diff
@@ -16,35 +16,37 @@
...
 
 import static jakarta.interceptor.Interceptor.Priority.LIBRARY_BEFORE;
 
 /**
  * CDI extension for processing CORS-annotated types.
  * <p>
- *     Pre-computes the {@link CrossOriginConfig} for each method which should have one and makes sure that the
- *     {@link CrossOrigin} annotation appears only on methods which also have {@code OPTIONS}.
+ * Pre-computes the {@link io.helidon.webserver.cors.CorsPathConfig} for each method which should have one and makes
+ * sure that CORS annotations appears only on methods which also have {@code OPTIONS}.
```
tjquinno: Typo: "annotations appears -> "annotations appear" (remove "s" from "appears")

webserver/cors/src/main/java/io/helidon/webserver/cors/CorsValidator.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import io.helidon.http.HeaderNames;
+import io.helidon.http.Method;
+import io.helidon.webserver.http.ServerRequest;
+import io.helidon.webserver.http.ServerResponse;
+
+class CorsValidator {
```
tjquinno, line 29: The preceding CORS implementation used logging to explain in quite some detail the decisions it was making and why. 

When users have questions about why a request is or is not treated as a CORS request or what type of CORS request it is viewed as, it has been very effective to have users enable that logging. In many cases they can resolve their questions themselves. 

And, if they need more help from us, we have that information to guide us.

Not carrying that logging forward will be a loss for users and us.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
tomas-langer: To separate it from the constructor with no parameters, that configures all `null` and not defaults

webserver/cors/src/main/java/io/helidon/webserver/cors/CorsValidator.java
tomas-langer: This is the output for failure:
```
2026.02.18 21:59:09.630 FINER [0x094a412f 0x0ed18e7c] CORS allowing method; actual method: PUT, allowedMethods: [DELETE, LIST, PUT]
2026.02.18 21:59:09.630 FINE [0x094a412f 0x0ed18e7c] CORS denying request; actual origin: http://not.allowed, allowedOrigins: [http://bar.foo, http://foo.bar], allowedOriginPatterns: []
```

And for success:
```
2026.02.18 22:01:19.964 FINER [0x5ead4b1d 0x25d58b6f] CORS allowing method; actual method: PUT, allowed all
2026.02.18 22:01:19.965 FINER [0x5ead4b1d 0x25d58b6f] CORS allowing all origins; actual origin: http://foo.bar
2026.02.18 22:01:19.965 FINER [0x5ead4b1d 0x25d58b6f] CORS allowing headers; actual headers: [], allowedHeaders: []
2026.02.18 22:01:19.965 FINE [0x5ead4b1d 0x25d58b6f] CORS allowing request
```

tomas-langer: Is there anything else we need to log? 
We have to be very careful as this is happening for each request to Helidon.
I think we can log quite a lot when in pre-flight, but loggin initial analysis for normal requests would lead to performance impact

tomas-langer: If you are interested in headers being sent around, that can be enabled by tracing the HTTP request. I would prefer not to duplicate this logging in CORS

tomas-langer: With full webserver logging:

failure:
```
2026.02.18 22:16:13.010 FINE [0x78c3a129 0x272b3326] recv headers: 
Access-Control-Request-Headers: X-foo
Access-Control-Request-Headers: X-bar
Access-Control-Request-Headers: X-oops
Access-Control-Request-Method: PUT
Host: localhost:38961
Origin: http://not.allowed
User-Agent: Helidon 4.4.0-SNAPSHOT
Connection: keep-alive
Content-Length: 0
Proxy-Connection: keep-alive

2026.02.18 22:16:13.010 FINER [0x78c3a129 0x272b3326] CORS allowing method; actual method: PUT, allowedMethods: [LIST, DELETE, PUT]
2026.02.18 22:16:13.010 FINE [0x78c3a129 0x272b3326] CORS denying request; actual origin: http://not.allowed, allowedOrigins: [http://foo.bar, http://bar.foo], allowedOriginPatterns: []
2026.02.18 22:16:13.010 FINE [0x78c3a129 0x272b3326] send status: 403 Forbidden
2026.02.18 22:16:13.010 FINE [0x78c3a129 0x272b3326] send headers: 
Connection: keep-alive
Content-Length: 0 (changing)
```

Success:
```
2026.02.18 22:16:13.017 FINE [0x78c3a129 0x115c6dab] recv headers: 
Access-Control-Request-Method: PUT
Host: localhost:38961
Origin: http://foo.bar
User-Agent: Helidon 4.4.0-SNAPSHOT
Connection: keep-alive
Content-Length: 0
Proxy-Connection: keep-alive

2026.02.18 22:16:13.018 FINER [0x78c3a129 0x115c6dab] CORS allowing method; actual method: PUT, allowed all
2026.02.18 22:16:13.018 FINER [0x78c3a129 0x115c6dab] CORS allowing all origins; actual origin: http://foo.bar
2026.02.18 22:16:13.018 FINER [0x78c3a129 0x115c6dab] CORS allowing headers; actual headers: [], allowedHeaders: []
2026.02.18 22:16:13.018 FINE [0x78c3a129 0x115c6dab] CORS allowing request
2026.02.18 22:16:13.018 FINE [0x78c3a129 0x115c6dab] send status: 204 No Content
2026.02.18 22:16:13.018 FINE [0x78c3a129 0x115c6dab] send headers: 
Access-Control-Allow-Methods: * (changing)
Access-Control-Allow-Origin: * (changing)
Access-Control-Max-Age: 3600 (changing)
Connection: keep-alive
Content-Length: 0
```

declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
tjquinno: So even if the caller passes `defaults` as `false` the code still sets it to `true`?

webserver/cors/src/main/java/io/helidon/webserver/cors/CorsValidator.java
tjquinno: There used to be logging explaining which type of request CORS decided a request was and why (at trace level IIRC). 

Users who choose to enable very detailed logging in order to diagnose a problem would expect some performance cost. 

Perhaps seeing what headers are or are not returned is enough to allow people to infer what type of request Helidon decided it was.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
tomas-langer: This is a private record in a package private class.

spericas:
Is there anything in the CORS spec about 200 vs. 204? I fear using 204 could break things. A Google search suggests that it may not be supported by some old browsers.

tomas-langer:
> Is there anything in the CORS spec about 200 vs. 204? I fear using 204 could break things. A Google search suggests that it may not be supported by some old browsers.

A successful HTTP response to a [CORS-preflight request](https://fetch.spec.whatwg.org/#cors-preflight-request) is similar, except it is restricted to an [ok status](https://fetch.spec.whatwg.org/#ok-status), e.g., 200 or 204.

tomas-langer:
I will push another update reverting to 200 code

tomas-langer:
I have found a problem I am trying to resolve - currently we have tests in archetypes that are run both as `@ServerTest` and `@RoutingTest`.
Unfortunatelly `RoutingTest` is exactly what the name says - it can only test routing. And `ServerFeature` is not part of routing, because the routing test bypasses webserver.
As a result, the "global" CORS is not configured in `RoutingTest` and the test fails.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
tjquinno: I get that this cannot be invoked from outside this type (today, at least). It just seems very odd to pass a parameter that is ignored. 

If we don't want the caller of the constructor to have to pass all those nulls when invoking `new CorsConfig(true, null, null, null, null, null, null)` directly then we could add to `CorsConfig`
```java
    static final CorsConfig DEFAULTS = new CorsConfig(true, null, null, null, null, null, null);
````
and the caller could use that.

Since there are several uses of `new CorsConfig()` as well, we could add a constant for that as well. The JVM might optimize the explicit constructor calls anyway but it would make the intent very clear.

tjquinno (commented):
One more try on the ignored parameter.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/cors/CorsPathConfigGenerator.java
tomas-langer: changed to use the parameter

PR #11168, opened 2026-02-17
Support OpenTelemetry server HTTP semantic conventions metrics

### LINKED ISSUE #11042: Honor the OTel metrics semantic conventions for HTTP traffic (which meters are recorded, names, tags, etc.)
Helidon Version: 4.x

## Enhancement Description
Allow users to easily have Helidon emit metrics which conform to the OpenTelemetry metrics semantic conventions for HTTP traffic.

### LINKED ISSUE #11123: allow fine tuning on which endpoints/methods to include in the REST_requestxxx metrics
Helidon Version: 4.3

For helidon mp, when rest request metrics is enabled such as below, metrics (REST_requestxxx) are generated for all REST endpoints.
metrics:
  rest-request:
    enabled: true

It would be nice to provide finer control on which endpoints/methods to target.

tjquinno:
### Description
Resolves #11042
Resolves #11123

### Release Note
_____
#### Selecting which requests are automatically measured
Helidon can automatically collect metrics for REST endpoint traffic:
* the automatic KPI metrics (simple counter of all requests or extended KPI metrics fi configured)
* the Helidon MP `REST.request` timers 
* metrics prescribed by OpenTelemetry semantic conventions for HTTP server traffic (new feature - see below)
 
You can now control which endpoints are measured by adding configuration under `server.features.observe.observer.metrics.auto` to control which sockets are included in the measurements and which paths are enabled or disabled. This is very similar to the approach taken to configure selective tracing of REST endpoints. You can do the same programmatically.

The updated metrics doc pages describe this new feature in detail.

#### Collecting metrics according to the OpenTelemetry metrics semantic conventions 
Users can now enable OpenTelemetry semantic conventions for metrics which measure incoming HTTP traffic by adding a dependency on the new `io.helidon.webserver.observe:helidon-webserver-observe-telemetry-metrics` component.

You can control which endpoints Helidon includes using the selection feature described just above.

This feature relies on Helidon's OpenTelemetry support. Be sure to set up--either programmatically or using configuration--the telemetry configuration including a metrics exporter, and be sure to add a dependency on the OpenTelemetry exporter you choose.

The updated Helidon SE and MP metrics documentation pages describe both of these new features.
_____

#### PR Overview
The existing metrics observer component now has a new SPI contract for providers of an HTTP filter that implements automatic HTTP metrics.  

The metrics observer code finds all instances of the provider contract and invokes their `filter` method, passing the metrics observer configuration. By default, the provided filters are added for each socket; config can select which sockets should be so filtered. 

The PR also contains a new component has an implementation of this provider contract. 

This implementation registers and updates metrics in compliance with the OTel metrics HTTP semantic conventions. The code uses the OpenTelemetry API directly to register and update the meters. That's why users need to set up `telemetry` with a metrics exporter. (Using the Helidon neutral metrics API in a MP app failed because the semantic conventions use tag names with dots, and MP Metrics prohibits that.)

Existing code already implements some "legacy" HTTP traffic metrics (KPI metrics, either basic or extended). This PR also applies the new path selections added to the metrics observer config to those legacy metrics. If the user omits the `auto` 

### Documentation
Doc update included.


webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+
+import io.helidon.builder.api.Prototype;
+import io.helidon.common.uri.UriPath;
+import io.helidon.http.Method;
+
+class AutoHttpMetricsConfigSupport {
+
+    private static final List<AutoHttpMetricsPathConfig> MEASUREMENT_DISABLED_HELIDON_ENDPOINTS = List.of(
+            AutoHttpMetricsPathConfigBlueprint.disabled("/metrics"),
```
tomas-langer: Maybe this should be `/metrics/*` to cover all possible sub-paths?
Similar we could just use `/observe` to cover all of observability endpoints

tomas-langer: Blueprint definitely should not be used here.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
```diff
@@ -0,0 +1,92 @@
...
+@Prototype.CustomMethods(AutoHttpMetricsPathConfigSupport.CustomMethods.class)
+interface AutoHttpMetricsPathConfigBlueprint {
+
+    /**
+     * Creates a metrics path config that disables automatic metrics collection for the specified exact path.
+     *
+     * @param path exact path to exclude
+     * @return metrics path config disabling collection for the path
+     */
+    static AutoHttpMetricsPathConfig disabled(String path) {
```
tomas-langer: Please do not add static methods to blueprints (ever).
If you want this to be part of public API, you should be able to add a `@Prototype.FactoryMethod` to custom methods.
If this is internal only, put the methods somewhere else.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsObserverConfigBlueprint.java
```diff
@@ -40,6 +40,14 @@ interface MetricsObserverConfigBlueprint extends ObserverConfigBase, Prototype.F
     @Option.Default("metrics")
     String name();
 
+    /**
+     * Automatic metrics collection settings.
+     *
+     * @return auto metrics collection settings
+     */
+    @Option.Configured("auto")
+    Optional<AutoHttpMetricsConfig> autoHttpMetrics();
```
tomas-langer: This is a backward iincompatible change - this method should declared as `default`, and add `@Prototype.IncludeDefaultMethods` annotation to the blueprint.

I do not understand the (commonly used) appraoch of having a different method name from the configuration key.

Maybe the config key should be `auto-http-metrics`, and not just `auto`? 
I think it is easier to understand config/builder when the method name matches the config key (matches as in camel case vs. snake case)

webserver/observe/telemetry/metrics/src/main/java/module-info.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.HelidonFlavor;
+
+@Features.Preview
+@Features.Name("OpenTelemetry Server HTTP Semantic Conventions")
+@Features.Description("Automatic metrics for compliance with OpenTelemetry server HTTP semantic conventions")
+@Features.Path({"Metrics", "Automatic Metrics", "Server", "HTTP", "OpenTelemetry"})
+@Features.Flavor(HelidonFlavor.SE)
+
```
tomas-langer: An excess line break

webserver/observe/telemetry/metrics/src/main/java/module-info.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+/**
+ * Support for OpenTelemetry metrics semantic conventions.
+ */
+
+import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.HelidonFlavor;
+
+@Features.Preview
+@Features.Name("OpenTelemetry Server HTTP Semantic Conventions")
```
tomas-langer: Can we find a shorter name? Names are used in `Path`, and this would not look great when you print the features out

webserver/observe/telemetry/metrics/src/main/java/module-info.java
```diff
@@ -0,0 +1,43 @@
...
+    requires io.helidon.service.registry;
+    requires io.helidon.webserver;
+
+    requires io.opentelemetry.api;
+    requires io.opentelemetry.semconv;
+    requires io.helidon.common.features.api;
+    requires micrometer.core;
+    requires io.helidon.metrics.api;
+    requires io.opentelemetry.sdk.metrics;
+    requires java.logging;
```
tomas-langer: We should never require `java.logging` 

Is it correct that this module does not `export` anything, nor does it provide a service?

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelope.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Models either:
+ *  - a single ResourceMetrics JSON object (your log line), OR
+ *  - items inside the wrapped ExportMetricsServiceRequest.resourceMetrics[].
+ *  <p>
+ *  Mostly generated by Oracle generative chat.
```
tomas-langer: What? How is this relevant to this interface?

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelope.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ */
+public interface MetricsEnvelope {
+    Resource resource();
```
tomas-langer: Missing javadoc on public interface methods

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelope.java
```diff
@@ -0,0 +1,82 @@
...
+    List<ScopeMetrics> scopeMetrics();
+
+    interface Resource {
```
tomas-langer: Missing javadocs everywhere

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelopeImpl.java
```diff
@@ -0,0 +1,186 @@
...
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Implementation of MetricsEnvelop. Mostly generated by Oracle generative chat.
+ */
+public final class MetricsEnvelopeImpl implements MetricsEnvelope {
```
tomas-langer: an impl class should never be public

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelopeImpl.java
```diff
@@ -0,0 +1,186 @@
...
+    private final Resource resource;
+    private final List<ScopeMetrics> scopeMetrics;
+
+    public MetricsEnvelopeImpl(Resource resource, List<ScopeMetrics> scopeMetrics) {
```
tomas-langer: we never use public constructors (except for exceptions)

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelopeImpl.java
```diff
@@ -0,0 +1,186 @@
...
+        this.resource = resource;
+        this.scopeMetrics = scopeMetrics == null ? List.of() : List.copyOf(scopeMetrics);
+    }
+
+    @Override public Resource resource() { return resource; }
```
tomas-langer: This formatting is not aligned with Helidon

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsExport.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.webserver.observe.telemetry.metrics;
+
+import java.util.List;
+
+/**
+ * Mostly generated by Oracle generative chat.
```
tomas-langer: What kind of Javadoc is this?

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
tjquinno: Moved `disabled` from the blueprint interface to the support class.

tjquinno: Also changed the patterns.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
tjquinno: removed

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsObserverConfigBlueprint.java
tjquinno: Added a default implementation of the method. Added the annotation to the type. 

I used `auto` for its brevity; removed it so the name will be defaulted.

webserver/observe/telemetry/metrics/src/main/java/module-info.java
tjquinno: fixed

webserver/observe/telemetry/metrics/src/main/java/module-info.java
tjquinno: Shortened.

webserver/observe/telemetry/metrics/src/main/java/module-info.java
tjquinno: `java.logging` was not used; it must have slipped in; removed.

The module contains exactly one type, a `@Service.Singleton`. Are you saying this must be declared as a `provides` in `module-info.java`?

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelope.java
tjquinno: removed

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelope.java
tjquinno: type removed

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelope.java
tjquinno: types removed

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelopeImpl.java
tjquinno: types removed

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelopeImpl.java
tjquinno: types removed

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsExport.java
tjquinno: types removed

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/MetricsEnvelopeImpl.java
tjquinno: type removed

webserver/observe/telemetry/metrics/src/main/java/module-info.java
tomas-langer: It is missing an `export`, so types from this module will not be visible - including the generated service descriptor.
A module without `exports` or `provides` seems to make no sense to me.

For singleton, you still need `exports` (not `provides`)

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/HelloWorldPathMethodSelectionTest.java
```diff
@@ -0,0 +1,213 @@
...
+    @Inject
+    MetricRegistry registry;
+
+    @Inject
+    @RegistryType(type = MetricRegistry.Type.BASE)
+    MetricRegistry restRequestMetricsRegistry;
+
+    @BeforeAll
+    public static void initialize() {
+        System.setProperty("jersey.config.server.logging.verbosity", "FINE");
```
tomas-langer: Do not set logging to lower levels than INFO (unless we are investigating something on pipelines).

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/HelloWorldPathMethodSelectionTest.java
```diff
@@ -0,0 +1,213 @@
...
+    // checks those metrics.
+    static <T> T runAndPause(Callable<T> c) throws Exception {
+        T result = c.call();
+        pause();
+        return result;
+    }
+
+    static void pause() {
+        try {
+            TimeUnit.MILLISECONDS.sleep(500L);
```
tomas-langer: Half a second is quite a long time. Is there a chance to implement this with no sleeps?
Or at least implement the sleeping approach on the side that tests (i.e. check if the result is OK, if not, wait for a bit, then test again) - this way we can use a cycle with shorter sleep times, so on a fast machine we are done faster, and on a slow machine we can give it even seconds..

tomas-langer: This is IMHO an intermittent failure waiting to happen

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/HelloWorldPathMethodSelectionTest.java
```diff
@@ -0,0 +1,213 @@
...
+        var headResponse = webTarget.path("/helloworld").request(MediaType.TEXT_PLAIN_TYPE).head();
+        assertThat("HEAD response", headResponse.getStatus(), is(200));
+        assertThat("HEAD header", headResponse.getHeaders(), hasEntry("X-result", List.of("good")));
+
+        // This should NOT create a timer.
+        var withNameResponse = webTarget.path("/helloworld/withArg/Joe").request(MediaType.TEXT_PLAIN_TYPE).get();
+        assertThat("With name response", withNameResponse.getStatus(), is(200));
+
+        // Wait for the metrics to update.
+        pause();
```
tomas-langer: the same as above - do not pause and then expect things will be OK. If the process of metrics is asynchronous, half a second may not be enough, but it may be ready in 100 ms.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigBlueprint.java
```diff
@@ -0,0 +1,83 @@
...
+    List<AutoHttpMetricsPathConfig> autoHttpMetricsPathConfigs();
+
+    /**
+     * Default suppressions for Helidon-provided paths plus explicitly-configured settings.
+     *
+     * @hidden internal use only
+     * @return effective auto metrics settings
+     */
+    @Option.Access("")
+    List<AutoHttpMetricsPathConfig> effectivePathConfigs();
```
tomas-langer: Do we really need two lists here? Can we update the list in a decorator to achieve what is needed?

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
```diff
@@ -0,0 +1,113 @@
...
+         * endpoints with measurement disabled. Users can enable automatic metrics for such an endpoint
+         * by adding explicit configuration for it with {@code enabled} set to {@code true}.
+         *
+         * @param config automatic metrics configuration
+         * @param method HTTP method from the request
+         * @param uriPath URI path from the request
+         * @return whether the request should be measured, based on the configuration
+         */
+        @Prototype.PrototypeMethod
+        static boolean isMeasured(AutoHttpMetricsConfig config, Method method, UriPath uriPath) {
```
tomas-langer: Are these methods really needed on the prototype? This seems like an implementation detail, that could be hidden in the component using the prototype.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
```diff
@@ -0,0 +1,113 @@
...
+         * meters for the specified meter name. Telemetry implementations can permit users to configure whether certain
+         * attributes--typically ones that might be expensive to obtain--are to set on automatic meters.
+         *
+         * @param config {@link io.helidon.webserver.observe.metrics.AutoHttpMetricsConfig} to check for opt-in settings
+         * @param meterName name of the meter to check for
+         * @param attributeName name of the attribute to opt in
+         * @return true if the attribute should be added to the meter; false otherwise
+         */
+        @Prototype.PrototypeMethod
+        static boolean isOptedIn(AutoHttpMetricsConfig config, String meterName, String attributeName) {
```
tomas-langer: Same comment as above

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    static class BuilderDecorator implements Prototype.BuilderDecorator<AutoHttpMetricsConfig.BuilderBase<?, ?>> {
+
+        @Override
+        public void decorate(AutoHttpMetricsConfig.BuilderBase<?, ?> target) {
+
+            var fullList = new ArrayList<>(target.autoHttpMetricsPathConfigs());
+            fullList.addAll(MEASUREMENT_DISABLED_HELIDON_ENDPOINTS);
+            target.effectivePathConfigs(fullList);
```
tomas-langer: As I mentioned on the blueprint, the `effectivePathConfigs` is not really needed here, you could just set it to `target.autoHttpMetricsPathConfigs(fullList)`

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
```diff
@@ -0,0 +1,78 @@
...
+    @Option.DefaultBoolean(true)
+    boolean enabled();
+
+    /**
+     * Path matcher for this path config for which this path config applies.
+     *
+     * @return path matcher
+     */
+    @Option.Configured
+    PathMatcher path();
```
tomas-langer: This is inconsistent with CORS path config.
Also the user (if using this imperatively) probably has a path string, rather than PathMatcher.
I prefer to use `String` here, and if you want a method on builder that uses `PathMatcher`, you can add it as a custom builder method (again similar thing done in CORS path config for headers and method).
Same comment applies to `Method` below.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
```diff
@@ -0,0 +1,78 @@
...
+    List<Method> methods();
+
+    /**
+     * Method predicate for checking a request's HTTP method against this config's methods.
+     *
+     * @return method predicate for the configured HTTP methods
+     * @hidden internal use only
+     */
+    @Option.Access("")
+    MethodPredicate methodPredicate();
```
tomas-langer, line 75: This seems like an implementation detail - the method predicate can be created in the component that uses this prototype, it is not needed here.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigSupport.java
```diff
@@ -0,0 +1,75 @@
...
+
+        /**
+         * Checks whether the {@link UriPath} matches the path in the path config.
+         *
+         * @param config path config settings
+         * @param uriPath {@code UriPath} to test
+         * @return true if the {@code UriPath} matches the config's path matcher; false otherwise
+         */
+        @Prototype.PrototypeMethod
+        static boolean matchesPath(AutoHttpMetricsPathConfig config, UriPath uriPath) {
```
tomas-langer: There custom meethods are an implementation detail, and can be on the component using the prototype.

webserver/observe/metrics/src/test/java/io/helidon/webserver/observe/metrics/TestAutoMetricsConfig.java
```diff
@@ -0,0 +1,121 @@
...
+                                methods: ["HEAD","OPTIONS"]
+                                enabled: false
+                              - path: "/hi"
+                                enabled: false
+                              - path: "/another/{name}"
+                                methods: ["PUT","POST","OPTIONS"]
+                                enabled: false
+                              - path: "/another/{name}"
+                """;
+        var configFromText = Config.just(ConfigSources.create(configText, MediaTypes.APPLICATION_X_YAML))
```
tomas-langer: you can now use `Config.just(configText, MediaTypes.APPLICATION_YAML)` (the `_X_` in media type is not needed)

webserver/observe/metrics/src/test/java/io/helidon/webserver/observe/metrics/TestAutoMetricsConfig.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    void testWithNoAutoConfig() {
+        String configText = """
+                server:
+                  features:
+                    observe:
+                      observers:
+                        metrics:
+                """;
+        var config = AutoHttpMetricsConfig.create(Config.just(ConfigSources.create(configText, MediaTypes.APPLICATION_X_YAML))
```
tomas-langer: can be simplified, slightly wrong media type

webserver/observe/telemetry/metrics/src/main/java/module-info.java
```diff
@@ -0,0 +1,41 @@
...
+    requires io.helidon.service.registry;
+    requires io.helidon.webserver;
+
+    requires io.opentelemetry.api;
+    requires io.opentelemetry.semconv;
+    requires io.helidon.common.features.api;
+    requires micrometer.core;
+    requires io.helidon.metrics.api;
+    requires io.opentelemetry.sdk.metrics;
+
```
tomas-langer, line 40: Please add `exports io.helidon.webserver.observe.telemetry.metrics;`

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
```diff
@@ -0,0 +1,306 @@
...
+                          auto-http-metrics:
+                            paths:
+                              - path: /greet
+                                methods: ["OPTIONS"]
+                                enabled: false
+                            sockets: ["@default","private"]
+                            opt-in: ["http.server.request.duration:server.address"]
+                """;
+        checkPort = false; // because we did not opt in for the server.port.
+        Config config = Config.just(ConfigSources.create(configText, MediaTypes.APPLICATION_YAML));
```
tomas-langer: Can be shortened, ConfigSources no longer needed here

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
```diff
@@ -0,0 +1,306 @@
...
+import io.helidon.webclient.http1.Http1Client;
+import io.helidon.webclient.http1.Http1ClientResponse;
+import io.helidon.webserver.WebServer;
+import io.helidon.webserver.WebServerConfig;
+import io.helidon.webserver.testing.junit5.ServerTest;
+import io.helidon.webserver.testing.junit5.SetUpServer;
+import io.helidon.webserver.testing.junit5.Socket;
+
+import io.opentelemetry.exporter.logging.otlp.OtlpJsonLoggingMetricExporter;
+import jakarta.json.Json;
```
tomas-langer: You can now use our own JSON library `helidon-json` that contains `JsonObject` abstraction. 
We  should never use `jakarta` in Helidon SE except for integrations with Jakarta specific libraries (such as media types for JSON-P and JSON-B)

webserver/observe/telemetry/metrics/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+            <artifactId>helidon-telemetry-opentelemetry-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common.features</groupId>
+            <artifactId>helidon-common-features-api</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.config</groupId>
+            <artifactId>helidon-config-metadata</artifactId>
```
tomas-langer: This dependency should not be needed.

webserver/observe/telemetry/metrics/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.helidon.common.features</groupId>
+                            <artifactId>helidon-common-features-codegen</artifactId>
+                            <version>${helidon.version}</version>
+                        </path>
+                        <path>
+                            <groupId>io.helidon.config.metadata</groupId>
```
tomas-langer: There is only a singleton, config metadata is not required here

webserver/observe/telemetry/metrics/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+                            <version>${helidon.version}</version>
+                        </path>
+                        <path>
+                            <groupId>io.helidon.codegen</groupId>
+                            <artifactId>helidon-codegen-apt</artifactId>
+                            <version>${helidon.version}</version>
+                        </path>
+                        <path>
+                            <groupId>io.helidon.builder</groupId>
+                            <artifactId>helidon-builder-codegen</artifactId>
```
tomas-langer: There is no blueprint, this is not required

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/HelloWorldPathMethodSelectionTest.java
tjquinno: Accidentally kept from another test's code I started with. Removed.

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/HelloWorldPathMethodSelectionTest.java
tjquinno: Replaced with our `assertWithRetry`.

microprofile/metrics/src/test/java/io/helidon/microprofile/metrics/HelloWorldPathMethodSelectionTest.java
tjquinno: fixed

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigBlueprint.java
tjquinno: The existing decorator assigned the effective list, but you're right we can just update the original one.

Changed.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
tjquinno: Two other components use `isMeasured`:
* The metrics CDI extension calls it so its Jakarta REST container filter can tell if a path/method combination should be measured as a `REST.request` MP timer.
* The OpenTelemetry semantic conventions implementation in `helidon-webserver-observe-telemetry-metrics` calls it to decide whether to measure a given request.

We wouldn't replicate the logic in those two places.

If your point is that the decision if a given path and method are to be measured is not inherently the responsibility of the config itself, then we could move this logic to a public static utility method: something like `public static boolean isMeasured(autoHttpMetricsConfig, path, method)`.

There is no such utility class already in `helidon-webserver-observe-metrics` now but we could add one for this method.

Would that address your concern?

For `isOptedIn`, only the `helidon-webserver-observe-telemetry-metrics` component uses `isOptedIn` to properly prepare the attributes (tags) to attach to each timer, so that logic could move into that component. And that makes sense because the opt-in behavior is (currently, at least) OTel-specific.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
tjquinno: See response above. Moved into the one component that used this method.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigSupport.java
tjquinno: changed

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
tjquinno: See above--multiple components, not just one, need this information. 

Even if it is an implementation detail, associating the non-public `MethodPredicate` directly with the config instance allows us to construct the `MethodPredicate` once when the path config is built and then use the predicate many times as each request is evaluated, thus avoiding creating a new predicate for each request.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigSupport.java
tjquinno: ~~Moved to public utility class.~~

After off-line discussion with Tomas, we decided to stay with the prototype method approach, for now at least.

webserver/observe/metrics/src/test/java/io/helidon/webserver/observe/metrics/TestAutoMetricsConfig.java
tjquinno: Changed to use the newer method. Removed the X - I'd missed that `application/yaml` finally got an official designation!

webserver/observe/metrics/src/test/java/io/helidon/webserver/observe/metrics/TestAutoMetricsConfig.java
tjquinno: changed

webserver/observe/telemetry/metrics/src/main/java/module-info.java
tjquinno: Added.

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
tjquinno: Changed.

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
tjquinno: Changed

webserver/observe/telemetry/metrics/pom.xml
tjquinno: removed

webserver/observe/telemetry/metrics/pom.xml
tjquinno: removed

webserver/observe/telemetry/metrics/pom.xml
tjquinno: removed

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
tjquinno: Changed. The public API now uses strings. The builder decorator now sets the `PathMatcher` and the `MethodPredicate` once (those settings are package private, not public). This way the code in other components can use them rather them than having to recreate them from the text values as each request is checked or create them in the other components.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsPathConfigBlueprint.java
tomas-langer: You do not need to create a new predicate per request, you can store it as a field in the component using the config.
There is a helper method: `io.helidon.http.Method#predicate(java.util.Collection<io.helidon.http.Method>)`

PR #11169, opened 2026-02-17
Update documentation with support for configuration expressions in scheduling

### LINKED ISSUE #11137: Scheduling an execution via annotation Cron ignores expression override via config
## Environment Details
* Helidon Version:4.3.3
* Helidon SE or Helidon MP : MP
* JDK version: OpenJDK 64-Bit Server VM (build 21.0.2+13-58, mixed mode, sharing)
* OS: Windows 10

----------

## Problem Description
I have a method which I need to execute every day. I use the Scheduling feature of Helidon for that.

When I use the annotation io.helidon.microprofile.scheduling.Scheduled everything works fine. I use MicroProfile Config to adjust the time of execution for the scheduled task.

```
transfer.MyImporter.executeImport.schedule.expression = 0 * 10 ? * *
```

That works fine.

But the annotation Scheduled is deprecated and io.helidon.scheduling.Scheduling.Cron should be used. But when using Cron the override with MicroProfile Config doesn't work. Even when I use the config property "cron" instead of "expression" it does not work.

The Cron annotation is working with the value I specify in the class source code (in the annotation) but not with MicroProfile Config neither with "cron" nor with "expression".

tomas-langer:
Resolves #11137 

Updated documentation to describe configurability of annotations for scheduling (both in declarative and in MP)

### DIFF
--- a/docs/src/main/asciidoc/mp/scheduling.adoc
+++ b/docs/src/main/asciidoc/mp/scheduling.adoc
@@ -139,7 +139,7 @@ The default value is an expression of `0 15 8 ? * *`.
 app:
   schedule:
     cache:
-      cron: "* * * * * ? "
+      cron: "* * * * * ?"
 ----
 
 The configured values would be executing every 1 second.


docs/src/main/asciidoc/mp/scheduling.adoc
```diff
@@ -103,24 +86,63 @@ include::{sourcedir}/mp/SchedulingSnippets.java[tag=snippet_4, indent=0]
...
+=== Fixed Rate
+The Fixed rate annotation can have the delay by and value overridden using config.
+
+.Annotation that allows config overrides
+[source,java]
+----
+include::{sourcedir}/mp/SchedulingSnippets.java[tag=snippet_2, indent=0]
+----
+
+The default values are 5 minutes for delay-by, and 10 minutes for interface, and could be overridden:
```
danielkec:
```suggestion
The default values are 5 minutes for delay-by, and 10 minutes for interval, and could be overridden:
```

docs/src/main/asciidoc/mp/scheduling.adoc
```diff
@@ -103,24 +86,63 @@ include::{sourcedir}/mp/SchedulingSnippets.java[tag=snippet_4, indent=0]
...
 [source,yaml]
 .Overriding annotated values from config
 ----
-fully.qualified.ClassName.methodName:
+app:
   schedule:
-    cron: "* * * * * ?"
-    concurrent: false
+    cache:
+      cron: "* * * * * ? "
```
danielkec:
```suggestion
      cron: "* * * * * ?"
```

danielkec (changes requested):

Issue #11173, opened 2026-02-18
Build fails on Java 26 due to javadoc warning


tomas-langer:
Run 

`mvn clean install -Pjavadoc -DskipTests -rf :helidon-webclient-grpc`

on Java 26

The following failure can be seen:

```
[INFO] --- javadoc:3.6.2:jar (attach-javadocs) @ helidon-webclient-grpc ---
[INFO] No previous run data found, generating javadoc.
[WARNING] Javadoc Warnings
[WARNING] warning: unknown enum constant Kind.LIST
[WARNING] reason: class file for io.helidon.config.metadata.ConfiguredOption$Kind not found
[WARNING] 1 warning
```

PR #11174, opened 2026-02-18
Fixes for Javadoc when building on Java 26 EA

tomas-langer:
The build now passes when building javadoc on Java 26.

Tried with:
```
OpenJDK Runtime Environment (build 26+35-2893)
```

Add support for links to grpc in Javadoc.

Resolve issue where javadoc failed with warning that Kind.LIST enum constant cannot be found.

Fixed an invalid link to grpc.

Resolves https://github.com/helidon-io/helidon/issues/11173


romain-grecourt (approved):

tomas-langer:
We are checking if there is a chance to fix this in a different manner - i.e. so we do not add transitive dep. on the metadata

romain-grecourt (dismissed):

Issue #100, opened 2018-10-09
Add a way to provide supported content types to JsonSupport 


romain-grecourt:
## Support any content type for JSON support

The content-types supported by the current JSON support utility are hard-coded.
We need to provide a way for users to specify the content type that are associated with JSON payload.

## Steps to reproduce

Register JSON support, send JSON payload with a post request with content-type application/x-www-form-urlencoded

romain-grecourt:
Actually this needs to be a general mechanism for any support media type.
@tomas-langer what do you think ?

tomas-langer:
Not sure this should be a general mechanism (as we are using a `Service` to implement the JSON support). 
Maybe we could have an abstract media type support that has a builder that can be configured, though the logic is not simple (as we must take into account the `Accept` and `Content-Type` headers).

romain-grecourt:
A base service class for media type support services would do.

spericas:
After the discussion earlier in the week, can we close this stating that we need to support `application/x-www-form-urlencoded` instead?

Captain1653:
Is it issue still relevant? 

Now helidon [supports specific media-type for json](https://github.com/helidon-io/helidon/pull/1091/files#diff-aee943b82a47e2a0334b4755bd37e31d455192bdd4c83821af33a1a964681226R81).

Also helidon [supports JSON-B and JSON-P media-types](https://helidon.io/docs/v4/se/webserver/webserver).

Should this issue be closed or I get something wrong?

romain-grecourt:
See https://github.com/helidon-io/helidon/blob/cbe1edf84eaac7c18400949b8d47438bc33cd2ef/http/media/jsonp/src/main/java/io/helidon/http/media/jsonp/JsonpSupport.java#L224

PR #11176, opened 2026-02-18
Helidon JSON media support for JsonValue subclasses.

### LINKED ISSUE #11027: Improve compile-time json HTTP media support
Create http media module structure - one module for JsonObject, one for binding.

tomas-langer:
A bit of refactoring of JSON Binding media support. Relevant fixes in other classes

Resolves #11027 
Related to #100 

Updated Health observer to use the new JSON library instead of JSON-P

Example of configuration of supported types and content-type:

```
server:
  media-context:
    media-supports:
      json:
        content-type: "application/x-json; charset=UTF-8"
        accepted-media-types: ["application/json", "application/x-json", "text/json"]
```

This configuration would:
- send `Content-Type: application/x-json; charset=UTF-8" header for client requests with entity, and server responses with entity (i.e. you call `ServerResponse.send(JsonValue)`)
- when server response is generated, this support would be matched if the `Accept` header from server request matches one of the listed media types

### DIFF
--- a/http/media/json/src/main/java/module-info.java
+++ b/http/media/json/src/main/java/module-info.java
@@ -16,7 +16,6 @@
 
 import io.helidon.common.features.api.Features;
 import io.helidon.common.features.api.HelidonFlavor;
-import io.helidon.http.media.json.JsonMediaSupportProvider;
 
 /**
  * Helidon JSON Media Support.
@@ -45,5 +44,5 @@ module io.helidon.http.media.json {
     exports io.helidon.http.media.json;
 
     provides io.helidon.http.media.spi.MediaSupportProvider
-            with JsonMediaSupportProvider;
+            with io.helidon.http.media.json.JsonMediaSupportProvider;
 }
--- a/http/media/json/src/test/java/io/helidon/http/media/json/JsonMediaTest.java
+++ b/http/media/json/src/test/java/io/helidon/http/media/json/JsonMediaTest.java
@@ -48,6 +48,7 @@ When adding/updating tests in this class, consider if it should be done
     - JacksonMediaTest
     - JsonbMediaTest
     - GsonMediaTest
+    (and all other JSON based tests)
  */
 class JsonMediaTest {
     private static final Charset ISO_8859_2 = Charset.forName("ISO-8859-2");


http/media/json/src/main/java/module-info.java
```diff
@@ -0,0 +1,49 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import io.helidon.common.features.api.Features;
+import io.helidon.common.features.api.HelidonFlavor;
+import io.helidon.http.media.json.JsonMediaSupportProvider;
```
Verdent: Not exactly a mistake, but I know you prefer to have the package specified as part of the "with" section of the "provides"

http/media/json/src/test/java/io/helidon/http/media/json/JsonMediaTest.java
```diff
@@ -0,0 +1,229 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+/*
+When adding/updating tests in this class, consider if it should be done
+ in the following tests a well:
+    - JacksonMediaTest
```
Verdent, line 48: Maybe this should contain also JsonBindingMediaTest. This change should be probably updated in all other media support test classes.

json/json/src/main/java/io/helidon/json/JsonObject.java
```diff
@@ -370,6 +370,30 @@ public void toJson(JsonGenerator generator) {
...
+    public boolean equals(Object o) {
+        if (!(o instanceof JsonObject that)) {
+            return false;
+        }
+        if (that == this) {
+            return true;
+        }
+
+        this.ensureResolvedKeys();
+        that.ensureResolvedKeys();
```
Verdent, line 389: One call is enough :-)

http/media/json/src/main/java/module-info.java
tomas-langer: Result of IDE changes after renaming, fixing.

http/media/json/src/test/java/io/helidon/http/media/json/JsonMediaTest.java
tomas-langer: Updated the comment in this class.

json/json/src/main/java/io/helidon/json/JsonObject.java
tomas-langer: it is one call - once per `this`, once per `that`

Verdent: Oh it is "that", I missed that!

Verdent (approved):

Issue #11178, opened 2026-02-19
Support sending HAProxy proxy protocol information using WebClient.


dtetreau-Oracle:
Helidon Version: 4

## Enhancement Description

In https://github.com/helidon-io/helidon/issues/11119, I requested the ability to read more data from the HAProxy Proxy Protocol handler when using Helidon's WebServer. In addition to that feature request, I would also like the ability to send proxy protocol data when using Helidon's WebClient. Note that Helidon's WebClient current "proxy" support is for HTTP proxies, not the HAProxy Proxy Protocol. To support the the [proxy protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt), the WebClient would need to be configured with the proxy data (for example, src/destination addresses and ports) when constructed, and then the WebClient would need to encode the proxy data and write it to the Socket/SocketChannel when the TCP/UDS connection is first established. This data would need to be sent before the TLS handshake occurs on the connection.

I took a look into the current code to see whether there was a way of potentially doing this already. The closest I found was the Helidon SocketOptions interface. Unfortunately, that would work for UnixDomainSocketClientConnection, but not for TcpClientConnection, because in the TCP initialization code, `socketOptions.configureSocket()` is called immediately before `socket.connect()`, but this use case needs to be able to access the socket immediately after it is connected.

PR #11179, opened 2026-02-19
Add ConnectionListener interface to WebClient config to enable writing Proxy Protocol information

dtetreau-Oracle:
### Description

This PR adds a new `ConnectionInitializer` interface to the Helidon WebClient which gives WebClient users the ability to access the underlying Socket after it has been connected, but before any TLS/HTTP traffic is sent over the socket. As currently written, this is a generic feature. The specific reason I would like this is to allow me to implement https://github.com/helidon-io/helidon/issues/11178 in my own code by providing a `ConnectionInitializer` implementation which writes the HAProxy Proxy Protocol data to the socket.

### Documentation

I don't see this being a widely used feature, so I haven't updated any documentation in this PR yet.

### DIFF
--- a/webclient/grpc-tracing/src/main/java/io/helidon/webclient/grpc/tracing/GrpcClientTracingInterceptor.java
+++ b/webclient/grpc-tracing/src/main/java/io/helidon/webclient/grpc/tracing/GrpcClientTracingInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2025, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 
+import io.helidon.service.registry.Services;
 import io.helidon.tracing.HeaderConsumer;
 import io.helidon.tracing.Scope;
 import io.helidon.tracing.Span;
@@ -53,15 +54,16 @@ class GrpcClientTracingInterceptor implements ClientInterceptor {
             @Override
             public void start(Listener<ResT> responseListener, Metadata headers) {
                 LOGGER.log(Level.DEBUG, "Call start; metadata: {0}", headers);
+                var tracer = Services.get(Tracer.class);
                 // Start a new span for the outgoing gRPC client call.
-                var outgoingClientSpanBuilder = Tracer.global()
+                var outgoingClientSpanBuilder = tracer
                         .spanBuilder(method.getServiceName() + "-" + method.getFullMethodName())
                         .kind(Span.Kind.CLIENT);
                 Span.current().ifPresent(parent -> outgoingClientSpanBuilder.parent(parent.context()));
                 outgoingClientSpan = new AtomicReference<>(outgoingClientSpanBuilder.start());
 
                 try {
-                    Tracer.global().inject(outgoingClientSpan.get().context(), null, new GrpcHeaderConsumer(headers));
+                    tracer.inject(outgoingClientSpan.get().context(), null, new GrpcHeaderConsumer(headers));
                     LOGGER.log(Level.DEBUG, "After injecting span context; metadata: {0}", headers);
                     outgoingClientScope = outgoingClientSpan.get().activate();
 
--- a/webclient/tests/grpc/src/test/java/io/helidon/webclient/grpc/tests/GrpcProxyProtocolTest.java
+++ b/webclient/tests/grpc/src/test/java/io/helidon/webclient/grpc/tests/GrpcProxyProtocolTest.java
@@ -23,7 +23,6 @@ import com.google.protobuf.ByteString;
 import com.google.protobuf.Empty;
 import io.grpc.stub.StreamObserver;
 import io.helidon.config.Config;
-import io.helidon.testing.junit5.Testing;
 import io.helidon.webclient.api.ClientUri;
 import io.helidon.webclient.api.ConnectionListener;
 import io.helidon.webclient.grpc.ClientUriSuppliers;
@@ -45,7 +44,6 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
-@Testing.Test
 public class GrpcProxyProtocolTest {
     @TempDir
     public Path tempDir;


webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java
```diff
@@ -118,7 +118,15 @@ public final void run() {
         // proxy protocol before SSL handshake
         if (listenerConfig.enableProxyProtocol()) {
             ProxyProtocolHandler handler = new ProxyProtocolHandler(socket.socket(), channelId);
-            proxyProtocolData = handler.get();
+            try {
+                proxyProtocolData = handler.get();
+            } catch (IOException | RuntimeException e) {
```
dtetreau-Oracle: I added this catch block because during testing, I noticed that when the ProxyProtocolHandler threw an exception, the client side eventually failed with a read timeout, not with an immediate error. Looking at how SSL and other socket establishment errors are handled below, I suspect this was overlooked during the initial proxy protocol feature addition.

common/socket/src/main/java/io/helidon/common/socket/SocketOptionsBlueprint.java
```diff
@@ -139,10 +139,15 @@ default void configureSocket(Socket socket) {
     @SuppressWarnings("unchecked")
     default void configureSocket(SocketChannel socket) {
         var logger = System.getLogger(SocketOptions.class.getName());
+        final var supportedOptions = socket.supportedOptions();
```
tomas-langer: we do not use `final` declaration on variables

webclient/api/src/main/java/io/helidon/webclient/api/BytesWritingConnectionListener.java
```diff
@@ -0,0 +1,50 @@
...
+
+    BytesWritingConnectionListener(final byte[] bytes) {
+        this.bytes = bytes;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void socketConnected(final ConnectedSocket socket) throws IOException {
```
tomas-langer: And neither do we use `final` on parameters.

webclient/api/src/main/java/io/helidon/webclient/api/BytesWritingConnectionListener.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+/**
+ * A {@link ConnectionListener} writes a series of bytes to the socket.
+ */
+final class BytesWritingConnectionListener implements ConnectionListener {
+    private final byte[] bytes;
+
+    BytesWritingConnectionListener(final byte[] bytes) {
```
tomas-langer: Please remove `final` from parameter

webclient/api/src/main/java/io/helidon/webclient/api/BytesWritingConnectionListener.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public void socketConnected(final ConnectedSocket socket) throws IOException {
+        socket.socket().getOutputStream().write(bytes);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void socketConnected(final ConnectedSocketChannel socket) throws IOException {
```
tomas-langer: Please remove `final` from parameter

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionListener.java
```diff
@@ -0,0 +1,95 @@
...
+    static ConnectionListener noop() {
+        return new DefaultConnectionListener();
+    }
+
+    /**
+     * Returns a {@link ConnectionListener} which writes bytes to the socket once the socket is connected.
+     * @param bytes The bytes to write.
+     * @return The listener which will write the given bytes to the socket.
+     */
+    static ConnectionListener writeBytes(byte[] bytes) {
```
tomas-langer: We use `create` as the verb for factory methods. So this should be either called `create(byte[])`, or `createWriteOnConnect` (or something similar - this is a  bit of a mouthful..)

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionListener.java
```diff
@@ -0,0 +1,95 @@
...
+     * @param socket The newly connected socket channel.
+     */
+    void socketConnected(ConnectedSocketChannel socket) throws IOException;
+
+    /**
+     * Context information about a newly connected {@link Socket}.
+     * @param socket The socket itself.
+     * @param channelId The channel id.
+     */
+    record ConnectedSocket(Socket socket, String channelId) {
```
tomas-langer: Records should not be used as public API, as they are harder to handle in a backward compatible manner.
It would be better to have an interface  (that can be implementated by a record.

The main reason is if you expose a constructor, you can never change such API to an interface. So in generate you cannot use public constructors in Helidon at all.

tomas-langer: Do we need the distinction of Socket vs. Channel on this level? Maybe a single socket interface could be used with a method to write `BufferData`

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConnectionContextImpl.java
```diff
@@ -0,0 +1,83 @@
...
+import io.helidon.webserver.ProxyProtocolData;
+
+
+/**
+ * {@link GrpcConnectionContext} which exposes specific information from a {@link ConnectionContext}.
+ */
+final class GrpcConnectionContextImpl implements GrpcConnectionContext {
+    private final ConnectionContext connectionContext;
+
+    GrpcConnectionContextImpl(final ConnectionContext connectionContext) {
```
tomas-langer: Please do not use `final` for parameters

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -157,9 +162,15 @@ public void init() {
                 methodMetrics.callStarted.increment();
             }
 
+            // Include the GrpcConnectionContext in the Helidon Context so that the gRPC customer
+            // handler can access the peer info and proxy protocol data.
+            final Context context = Contexts.context().orElse(Context.create());
```
tomas-langer: Please do not use `final` here.

tomas-langer: This is a very dangerous call that is most likely not going to work as you expect.
Vy default we no longer run all requests in a dedicated context.

I think the best code here would be:

```java
Context context = Context.create(connectionContext.listenerContext().context());
```

`connectionContext` is the `ConnectionContext` from `io.helidon.webserver` - you use it to create the `GrpcConnectionContext` - so either expose the `listenerContext` method as well, or send the connection context around.
This will create an instance of Helidon context with the listener context as parent (which is the right thing to do here), whereas you code would use the global context as parent.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -157,9 +162,15 @@ public void init() {
...
+            context.register(GrpcConnectionContext.class, connectionContext);
+
             // initiate server call
             ServerCallHandler<REQ, RES> callHandler = route.callHandler();
-            listener = callHandler.startCall(serverCall, GrpcHeadersUtil.toMetadata(headers));
+            listener = Contexts.runInContext(context, () ->
```
tomas-langer: The context method has a preformance overhead.

Is there a chance to provide the `GrpcConnectionContext` through some grpc specific APIs that we have control over, rather than using context here? If not, let's keep it in...

webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java
tomas-langer: `IOException` cannot be thrown here, `RuntimeException` is sufficient

webserver/webserver/src/main/java/io/helidon/webserver/ProxyProtocolHandler.java
```diff
@@ -98,23 +97,17 @@ private static RequestException badProtocolException(String messageSuffix) {
...
-    ProxyProtocolHandler(Socket socket, String channelId) {
-        this.socket = socket;
+    ProxyProtocolHandler(SocketChannel socketChannel, String channelId) {
+        this.socketChannel = socketChannel;
         this.channelId = channelId;
     }
 
-    @Override
-    public ProxyProtocolData get() {
+    public ProxyProtocolData get() throws IOException {
```
tomas-langer: Never throw a checked exception in Helidon

webserver/webserver/src/main/java/io/helidon/webserver/ProxyProtocolHandler.java
```diff
@@ -98,23 +97,17 @@ private static RequestException badProtocolException(String messageSuffix) {
...
         LOGGER.log(Level.DEBUG, "Reading proxy protocol data for channel %s", channelId);
-
-        try {
```
tomas-langer, line 113: This code was correct, please revert it.

tomas-langer (commented):
Thanks for the contribution!

I have reported a few violation of Helidon code style, mostly cosmetic fixes.
One bigger thing to solve is the public API where you use `record`, and I think this could be hidden behind a single interface - I tried to provide detailed comments.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
dtetreau-Oracle: Looking for opportunities to expose this through the `io.grpc.Context` instead of the Helidon Context.

common/socket/src/main/java/io/helidon/common/socket/SocketOptionsBlueprint.java
dtetreau-Oracle: Done.

webclient/api/src/main/java/io/helidon/webclient/api/BytesWritingConnectionListener.java
dtetreau-Oracle: Done.

webclient/api/src/main/java/io/helidon/webclient/api/BytesWritingConnectionListener.java
dtetreau-Oracle: Done.

webclient/api/src/main/java/io/helidon/webclient/api/BytesWritingConnectionListener.java
dtetreau-Oracle: Done.

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionListener.java
dtetreau-Oracle: Done, switched `noop()` --> `createNoop()` and `writeBytes()` --> `createWriteOnConnect`.

webclient/api/src/main/java/io/helidon/webclient/api/ConnectionListener.java
dtetreau-Oracle: Removed records in favor of interfaces.

Regarding distinguishing Socket vs Channel, although my own very specific use cases could be served by simply exposing a way to write data, I felt that the cases were distinct enough to keep them separate. This is also similar to existing patterns in Helidon such as the SocketOptions which has separate `configureSocket(Socket)` and `configureSocket(SocketChannel)` methods. If you would like me to combine them into one, however, then I will do so.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConnectionContextImpl.java
dtetreau-Oracle: Done

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
dtetreau-Oracle: Removed final, and I was able to set the necessary information directly on the `io.grpc.Context` instead of the Helidon `Context`, so I think this addresses your concern.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
dtetreau-Oracle: I've changed this code to now set a key on the `io.grpc.Context`. Note that the `GrpcConnectionContext` interface is not (and cannot be) located in the `helidon-grpc-core` module, so I could not define the grpc `Context.Key` in the existing `ContextKeys` class and instead had to create a new `ServerContextKeys` class in `helidon-webserver-grpc` module.

webserver/webserver/src/main/java/io/helidon/webserver/ProxyProtocolHandler.java
dtetreau-Oracle: Done

webserver/webserver/src/main/java/io/helidon/webserver/ProxyProtocolHandler.java
dtetreau-Oracle: Done

tomas-langer (approved):
LGTM

PR #11183, opened 2026-02-20
Generate service descriptors without deprecated code for annotations.

### LINKED ISSUE #11182: Service descriptors are generated with deprecated APIs
This can be seen in examples for declarative, where build reports a lot of warnings, such as:

```
[WARNING] /home/tomas/Projects/oracle/helidon/helidon-examples/examples/declarative/webserver/hello-world/target/generated-sources/annotations/io/helidon/examples/declarative/webserver/helloworld/HelloWorldEndpoint__ServiceDescriptor.java:[54,37] putValue(java.lang.String,java.lang.Object) in io.helidon.common.types.Annotation.BuilderBase has been deprecated and marked for removal
```

tomas-langer:
Align common types in builder tests with common/types.

Resolves #11182 

We currently use `Annotation.Builder.putValue` which is deprecated.
Added `property(String, Object)` method to the builder to simplify usage, and changed codegen to generate service descriptors using the new method.

### DIFF
--- a/logging/log4j/pom.xml
+++ b/logging/log4j/pom.xml
@@ -81,12 +81,29 @@
                 <configuration>
                     <annotationProcessorPaths>
                         <path>
+                            <groupId>io.helidon.codegen</groupId>
+                            <artifactId>helidon-codegen-apt</artifactId>
+                            <version>${helidon.version}</version>
+                        </path>
+                        <path>
                             <groupId>io.helidon.common.features</groupId>
                             <artifactId>helidon-common-features-codegen</artifactId>
                             <version>${helidon.version}</version>
                         </path>
                     </annotationProcessorPaths>
                 </configuration>
+                <dependencies>
+                    <dependency>
+                        <groupId>io.helidon.codegen</groupId>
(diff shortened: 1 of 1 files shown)


builder/tests/common-types/src/main/java/io/helidon/common/types/AnnotationPropertyImpl.java
```diff
@@ -40,4 +41,28 @@ public Object value() {
...
+    public String toString() {
+        if (constantValue == null) {
+            return String.valueOf(value);
+        }
+        return constantValue.type().fqName() + "." + constantValue.name();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(value);
```
Verdent, line 55: Shouldnt be the constantValue addressed here as well?

tomas-langer: No, because the same annotation type with the same value is the same instance - even if one would use a constant and another one would use the value.

Verdent (approved):

PR #11184, opened 2026-02-20
API stability processor and API

### LINKED ISSUE #11124: Mechanism to specify and enforce API stability
## Background

Today Helidon has the ability to specify the stability of a module using the `io.helidon.common.features.api` annotations in `module-info` files (see [example module-info.java](https://github.com/helidon-io/helidon/blob/023951b080920a4057e993cfda5f1d4ea7fb79c8/integrations/langchain4j/langchain4j/src/main/java/module-info.java#L25)).

This approach has some limitations:

1. Granularity: `@Features` can only be applied at the module level -- not at the class/type level.
2. Enforcement: use of preview/incubating features is only indicated at runtime.  There is no mechanism to enforce or warn at build time.

## Enhancement Description

Note: examples and API details given below should be considered illustrative.  Options need to be discussed. 

### Overview

Create one or more annotations that can be applied at the type level to indicate API stability.  Check API stability at user application compile time and issue warnings when Preview, Incubating or Private APIs are used, with options to suppress the warnings or fail the build.

Example:

```java
@Stability(PREVIEW)
public class MyApi {
}
```

Possible stability levels roughly align with JDK terminology:

* PRIVATE:  Helidon project private. Do not use. Can change at any time.
* INCUBATING:  Highly unstable API. Very likely to change including the possibility of complete removal. Feel free to experiment with but do so at your own risk.
* PREVIEW: API is on path to being stable in roughly the shape it now appears. API changes are possible, but feature is documented and supported.
* STABLE: API is stable. Any changes will be backwards compatible or go through deprecation process.

Use of a single annotation with an enum reduces the ambiguity of what happens if you have multiple stability levels on a class.

### Suppression

It should be possible to suppress warnings in code. One option is to support `java.lang.SupressWarnings`. For example

```java
@SuppressWarnings("helidon:api")
```
### Control compile time behavior

By default this mechanism issues warnings at compile time.  It should be possible to control this behavior. One option is to use the `javac` `-A` flag:

```
-Ahelidon.api=fail
-Ahelidon.api=suppress
-Ahelidon.api=warn
```

### Deprecation

Deprecation would continue to be done with the `java.lang.Deprecated` annotation.  

### Helidon Requirements

It would be a project requirement that every public class/type must be annotated with a stability level. Ideally we have tooling to check for this.

## References

While we want to align with JDK terminology it is also useful to see how JUnit handles this. Note that in the JDK the term "Experimental" applies to JVM features and not JDK APIs.

* [JEP 11: Incubator Modules](https://openjdk.org/jeps/11)
* [JEP 12: Preview Features](https://openjdk.org/jeps/12)
* [JUnit API Evolution](https://docs.junit.org/6.0.2/api-evolution.html)

tomas-langer:
### Description
Resolves #11124

Adds API stability support for Helidon and carries it through Helidon enforcement, user validation, and generated code paths.

This PR:
- introduces `io.helidon.common.Api` with `@Api.Stable`, `@Api.Preview`, `@Api.Incubating`, `@Api.Deprecated`, `@Api.Internal`, `@Api.Since`, and suppression constants
- adds `helidon-codegen-api-stability`, an annotation processor for user applications that fails by default for `@Api.Internal` and `@Api.Incubating`, warns by default for `@Api.Preview` and `@Api.Deprecated`, and supports both `@SuppressWarnings` on containing elements and `-Ahelidon.api` / `-Ahelidon.api.*` overrides
- adds `helidon-codegen-api-stability-enforcer`, an annotation processor for Helidon modules that requires public top-level APIs to declare exactly one stability annotation while still allowing members, constructors, and nested types to carry more specific annotations
- publishes both processors from the codegen reactor and exposes them via the BOM, `helidon-all`, and `helidon-bundles-apt`
- enables the enforcer in `common/common` and `common/mapper`
- annotates Helidon public APIs in `common/common` and `common/mapper`, and updates selected APIs in config, http, service-registry, webclient, and webserver to use the new stability model
- updates builder, declarative, service, and JSON code generation paths so generated code propagates stability annotations or suppresses diagnostics where needed
- adds processor and enforcer regression coverage, including a Maven-backed consumer test that verifies fail, warn, and suppression behavior through the real reactor wiring, plus supporting build and diagnostic fixes

Detailed stability rules, suppression keys, and processor options are documented in `codegen/api-stability/README.md`.


codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityScanner.java
```diff
@@ -0,0 +1,281 @@
...
+
+    Collection<Ref> deprecatedApis() {
+        return deprecatedApis;
+    }
+
+    private Void visit(TreePath path, Supplier<Void> superCall) {
+
+        if (path != null) {
+            var elt = trees.getElement(path);
+            if (elt instanceof TypeElement || elt instanceof ExecutableElement) {
```
danielkec, line 108: This returns before visiting the qualifier, so Outer.Nested and IncubatingApi.stableMethod() miss the outer @Api.Incubating type and produce no diagnostic.

codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityScanner.java
```diff
@@ -0,0 +1,281 @@
...
+        }
+        return Optional.empty();
+    }
+
+    private Element classElement(CompilationUnitTree compilationUnit) {
+        // imports belong to the compilation unit, and not the top level class
+        // we want to be able to suppress warnings on the first class of the compilation unit
+        for (Tree typeDecl : compilationUnit.getTypeDecls()) {
+
+            if (typeDecl.getKind() == Tree.Kind.CLASS) {
```
danielkec: Imports only attach to the first top-level CLASS here. A top-level interface with @SuppressWarnings("helidon:api:internal") still gets an unsuppressible import error because classElement() returns null.

tomas-langer: It is for all top level elements, I will fix this.

codegen/api-stability/README.md
```diff
@@ -0,0 +1,94 @@
...
+- methods, constructors, and nested types may also carry their own stability annotations
+- they may also be annotated with `@Api.Since("version")` to announce the version that the stability state changed
+- in Helidon, this is verified by an API Stability Enforcer annotation processor, that must be configured for `default-compile`
+
+Users:
+- API Stability annotation processor should be present in user's projects in compiler configuration
+- user's build will fail if they use `Internal` or `Incubating` APIs, unless they suppress warnings, or configure the annotation processor to ignore/warn
+- user's build will print warnings if they use `Preview` or `Deprecated` APIs (also can be suppressed)
+
+# Stability states and transition
```
romain-grecourt: The wording of this section and the two nested sections imply a state machine, but it's not.. right ?

common/common/src/main/java/io/helidon/common/Api.java
```diff
@@ -0,0 +1,192 @@
...
+ * </ul>
+ *
+ * In case you decide to use on of the APIs that are not stable (i.e. to provide early feedback to us, or with
+ * the knowledge you may need to change sources when using the next minor/dot/patch release of Helidon), you can
+ * add {@link java.lang.SuppressWarnings} with the constants provided in this class, such as {@link #SUPPRESS_PREVIEW},
+ * or you can suppress all warnings/errors with {@link #SUPPRESS_ALL}.
+ */
+@Api.Stable
+// we expect people to use the constants defined here
+public final class Api {
```
romain-grecourt, line 50: I don't agree with having the annotations usable on methods.

Type hierarchies may be allowed to mix stability levels, so this implies methods or fields with diverging levels.

If we allow internal API in type hierarchies, we should require method overrides to reset the level correctly.

common/common/src/main/java/io/helidon/common/Api.java
```diff
@@ -0,0 +1,192 @@
...
+     * A deprecated API.
+     * <p>
+     * Deprecated APIs will be removed in the next major version of Helidon (or later).
+     * <p>
+     * Mutually exclusive with other API stability annotations.
+     */
+    @Api.Stable
+    @Target({ElementType.TYPE, ElementType.CONSTRUCTOR, ElementType.METHOD})
+    @Retention(RetentionPolicy.CLASS)
+    public @interface Deprecated {
```
romain-grecourt: I don't understand the need for a custom Deprecated annotation. Why is `java.lang.Deprecated` not usable ?

common/common/src/main/java/io/helidon/common/Api.java
tomas-langer: The intention is to have a `Stable` class/interface API, with an `Incubating` or `Preview` method, so we can evolve the APIs.
The intention is not to have `Internal` methods on stable APIs, or `Stable` methods on `Internal`, `Preview`,  `Incubating`, or `Deprecated` APIs.

I think the first use case is a must have, so I would keep the ability to annotate methods. Maybe just change the rules to make sure it is only going from "higher" to "lower" stability, and never the other way around

common/common/src/main/java/io/helidon/common/Api.java
tomas-langer: It is usable, but it follows a different path in validation and suppression.
This is consistent and covers the full lifecycle of Helidon APIs, rather then covering most of the lifecycle and then stepping aside to `@Deprecated` for deprecation.

codegen/api-stability/README.md
tomas-langer: It is almost a state machine, and it makes sense to document the expected transitions. 
The stability annotation is a state - it is part of the lifecycle of the service. You cannot have more than one, there is a clearly defined transition, and there is initial and termination "state" (even though the termination state means it is deleted).

tomas-langer:
> Should we create a separate issue to track the APIs to be annotated?
> 
> #11124 mentions all APIs to be annotated, however this PR is only the fundamental building block.
> 
> #11124 targets 4.5.0 ; but this PR targets main. We need to either re-target this PR or the issue ; and create a mirror.
> 
> We also need to clarify the difference in defaults for warn/fail. 4.x will warn by default ; 27+ will fail by default.

I have created a backport: #11498 
I kept the update of existing APIs out of scope except for the ones included as a "proof"

As a follow up, we will need to update the backport with the list of APIs we must change for 4.x.

I have also created a follow up issue: #11500

codegen/api-stability/README.md
romain-grecourt: Thinking about this sub-system in state-machine terms makes it complicated, we should keep it simple.

Each annotation should describe its contract without mentioning or codifying against other stability levels.
E.g In practice a PREVIEW API will never be demoted to INCUBATING, but does it need to be codified ?

We could have a notion of order `INTERNAL < INCUBATING < PREVIEW < STABLE` although this can be kept as an implementation detail of the processors.

common/common/src/main/java/io/helidon/common/Api.java
romain-grecourt: See my other comment about state machine and complexity.
There shouldn't be a lifecycle, at best ordering which is possibly only an implementation detail.

Deprecation should be adjacent to the stability, anything can be deprecated at any time, it's an indication that the symbol will go away.

common/common/src/main/java/io/helidon/common/Api.java
romain-grecourt: Do we have an existing use-case that justifies having the per-method granularity ?

---

Assuming the stability levels are ordered `STABLE > PREVIEW > INCUBATING > INTERNAL`, we define the main rule as follows:

The visible methods of a type hierarchy must be defined by a type with greater or equal stability than the declared type. Method overrides can be used to override stability.

This would effectively mean:
- Stable APIs can only have stable members throughout their type hierarchy
- Preview APIs should never expose non-preview direct members
- Incubating APIs should never expose non-incubating direct members
- Internal APIs should never expose non-internal direct members

With that rule in place, having the per-method granularity does not make sense as it would only allow declaring direct member of higher stability.

tomas-langer: Stable class can have a preview/incubating method. That would not be possible if we cannot annotate a method.

romain-grecourt: Can you justify that requirement ?

IMO Stable APIs should be 100% Stable, updates are done with extreme caution. Mistakes cannot be reverted without deprecation cycles.

tomas-langer: For example adding a new property to the webserver to handle some performance optimization. Or maybe a feature flag for an incubating feature that is built-into the webserver.

Something like:
```
server:
  enable-nonblocking-listener: true # incubating feature
  non-blocking: # configuration of incubating feature
    workers: 14  
```

The methods in builder are not stable - if this would be better, we will just remove them in the future.

codegen/api-stability/README.md
tomas-langer: I have updated the documention and implementation.
The transition rules are removed, it is no longer a state machine.

common/common/src/main/java/io/helidon/common/Api.java
tomas-langer: I have updated the description and implementation.
Nested is still allowed - though there is a hard rule that non-Stable types cannot contain Stable elements

common/common/src/main/java/io/helidon/common/Api.java
tomas-langer: Removed, updated code and docs

common/common/src/main/java/io/helidon/common/Api.java
romain-grecourt: If we have method stability, the rules I laid out above don't apply.

How about the following rule:

> Method stability must be lower or equal to the enclosing type stability.
> Overridden method stability must be lower or equal to the enclosing type stability.

codegen/api-stability-enforcer/src/test/java/io/helidon/codegen/api/stability/enforcer/ApiStabilityEnforcerProcessorTest.java
```diff
@@ -0,0 +1,251 @@
...
+    void testMissingTopLevelAnnotationFails() throws IOException {
+        var messages = compileWithApi(new ApiStabilityEnforcerProcessor(),
+                                      new JavaSourceFromString("TestApi.java", """
+                                              package io.helidon.common;
+
+                                              public class TestApi {
+                                              }
+                                              """));
+
+        assertThat(messages,
```
romain-grecourt, line 69: Should these tests assert the full diagnostic line to ensure that we have rich diagnostics with line / column and a caret that points at the symbol ?

codegen/api-stability/src/test/java/io/helidon/codegen/api/stability/ApiStabilityProcessorTest.java
```diff
@@ -0,0 +1,928 @@
...
+        var stderr = new StringWriter();
+        javac.run(new PrintWriter(stdout), new PrintWriter(stderr), args.toArray(String[]::new));
+
+        return (stderr.toString() + System.lineSeparator() + stdout)
+                .lines()
+                .filter(line -> !line.isBlank())
+                .toList();
+    }
+
+    private static List<String> compile(Processor processor,
```
romain-grecourt: Should this be using `io.helidon.codegen.testing.TestCompiler` ?

codegen/api-stability-enforcer/src/test/java/io/helidon/codegen/api/stability/enforcer/ApiStabilityEnforcerProcessorTest.java
```diff
@@ -0,0 +1,251 @@
...
+                    }
+                }
+                """));
+        sources.addAll(List.of(compilationUnits));
+        return compile(processor, sources.toArray(JavaFileObject[]::new));
+    }
+
+    private static List<String> compile(Processor processor,
+                                        JavaFileObject... compilationUnits) throws IOException {
+        var compiler = javax.tools.ToolProvider.getSystemJavaCompiler();
```
romain-grecourt, line 233: Should this be using `io.helidon.codegen.testing.TestCompiler` ?

tomas-langer: These tests cannot use TestCompiler, as that requires types from modules that use this processor - so it would create a cyclic dependency.

codegen/api-stability/src/test/java/io/helidon/codegen/api/stability/ApiStabilityProcessorTest.java
tomas-langer: Working on this...

codegen/api-stability-enforcer/src/test/java/io/helidon/codegen/api/stability/enforcer/ApiStabilityEnforcerProcessorTest.java
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        var messages = new ArrayList<String>();
+        for (var diagnostic : diagnostics.getDiagnostics()) {
+            messages.add(diagnostic.toString());
+        }
+        return messages;
+    }
+
+    private static String fullError(String fileName,
```
romain-grecourt: Why not assert with a text block ?

tomas-langer: Because of system specific behavior - this is different on Linux and on Windows. The `fullError` method takes care of that.

romain-grecourt: Is it possible to normalize the diagnostic message at `diagnostic.toString()` instead ?
This would make the diagnostics assertions more readable and more maintainable, where one can just copy paste the error message (at least when not on Windows).

tomas-langer: Refactored to use a text block

tomas-langer:
I have discovered two issues:

1. `DeclarativeClient` generated type does not suppress warnings, and compilation of example fails
2. We cannot use `-Ahelidon.api.incubating` option (or other options of the stability annotation processor), as these conflict with the check in Helidon codegen, as it does not recognize this as a valid `helidon` option

tomas-langer:
> I have discovered two issues:
> 
> 1. `DeclarativeClient` generated type does not suppress warnings, and compilation of example fails
> 2. We cannot use `-Ahelidon.api.incubating` option (or other options of the stability annotation processor), as these conflict with the check in Helidon codegen, as it does not recognize this as a valid `helidon` option

Fixed now

PR #11187, opened 2026-02-23
Helidon JSON polymorphism support

### LINKED ISSUE #11121: JSON Binding polymorphism support
JSON Binding support should be able to deserialize polymorphic types properly.

Usecase:
```java
interface Animal {}
class Dog implements Animal {}
class Cat implements Animal {}
```
When deserializing to an Animal type, there should be a mechanism, which would help deserializer to determine which implementations should be used to deserialize into.

### LINKED ISSUE #11185: Helidon JSON field name hash conflicts
Generated converter should generate special handling for the affected field name such a collision is detected.

Verdent:
### Description

Fixes #11121 
Fixes #11185 

### Documentation

None

### DIFF
--- a/json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
+++ b/json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
@@ -197,7 +197,7 @@ record ConvertedTypeInfo(TypeName converterType,
         if (polymorphic.isPresent()) {
             Annotation annotation = polymorphic.get();
             key = annotation.stringValue("key").orElse(key);
-            defaultSubtype = annotation.typeValue("defaultSubtype")
+            defaultSubtype = annotation.typeValue()
                     .filter(it -> !it.equals(OBJECT));
         }
 


json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +230,64 @@ private Json() {
 
     }
 
+    /**
+     * Polymorphism supporting annotation.
+     * This annotation defined which are the known subtypes of the annotated type
```
tomas-langer: "defined which are the known" -> "defines the known"

json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +230,64 @@ private Json() {
...
+     * and how to successfully deserialize into them.
+     */
+    @Retention(RetentionPolicy.RUNTIME)
```
tomas-langer: Retention should be `CLASS`

json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +230,64 @@ private Json() {
...
+    @Target(ElementType.TYPE)
+    public @interface TypeInfo {
```
tomas-langer: The name conflicts with our comon types `TypeInfo`, and `Type` conflicts with Java - we need a different name for this annotaiton.

json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +230,64 @@ private Json() {
...
+
+        /**
+         * Allowed aliases of the handled type.
+         *
+         * @return allowed aliases
+         */
+        Subtype[] value();
```
tomas-langer: The name should probably be Alias, if it is allowed aliases, and not `Subtype`

json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +230,64 @@ private Json() {
...
+
+        /**
+         * Key used for keeping the alias information.
+         * Default value is {@code @type}.
+         *
+         * @return key name
+         */
+        String key() default "@type";
```
tomas-langer, line 249: Maybe note that we are talking about a JSON key?

json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +230,64 @@ private Json() {
...
+
+        /**
+         * The default implementation to be used when alias is not provided.
+         * The default value {@link java.lang.Object} is intended as a "not specified"
+         * marker. In that case, no default implementation will be used and an exception
+         * will be thrown instead.
+         *
+         * @return the default implementation of the annotated type.
+         */
+        Class<?> defaultImplementation() default Object.class;
```
tomas-langer: This is a bit too long for me. Maybe `DefaultType`? 

It may make sense to keep `Subtype` annotaiton, and call this `defaultSubtype`, to make it clear that we are talking about one of them.
`defaultAlias` is unfortunately not great, as that cannot be a class.

json/json/src/main/java/io/helidon/json/JsonParser.java
```diff
@@ -133,6 +133,20 @@ static JsonParser create(JsonValue value) {
...
+     * Create a new JSON parser that pretends to be at the beginning of an object.
+     * <p>
+     * This method wraps an existing parser to ensure that parsing begins
+     * at the start of a JSON object ('{'). It is expected the provided parser is at the key start or object end.
+     * </p>
+     *
+     * @param parser the base parser to wrap
+     * @return a new JsonParser instance that starts at object beginning
+     */
+    static JsonParser objectStart(JsonParser parser) {
```
tomas-langer: If this is a factory method, it should start with `create`, this could conflict with an actual method on the parser instance.

json/json/src/main/java/io/helidon/json/JsonParser.java
```diff
@@ -403,4 +417,30 @@ default JsonException createException(String message, byte c) {
...
+     */
+    void mark();
+
+    /**
+     * Discards the current mark.
+     * <p>
+     * This method clears any previously set mark. Should be always called when backtracking is no longer needed.
+     * </p>
+     */
+    void dumpMark();
```
tomas-langer: Please use `clearMark`

json/binding/src/main/java/io/helidon/json/binding/Json.java
Verdent: I can change the name, but it should not be SubType/Alias annotation, because it would require to set the alias value as well and it would not work well here.

json/binding/src/main/java/io/helidon/json/binding/Json.java
```diff
@@ -230,4 +231,75 @@ private Json() {
 
     }
 
+    /**
+     * Polymorphism supporting annotation.
+     * This annotation defines the known subtypes of the annotated type
```
tomas-langer: This docs should be now changed.

json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
```diff
@@ -77,10 +80,10 @@ record ConvertedTypeInfo(TypeName converterType,
             "REVERSE_ALPHABETICAL", Comparator.reverseOrder());
 
     public static ConvertedTypeInfo create(TypeInfo typeInfo, CodegenContext ctx) {
```
tomas-langer: This method should not be public (the type is not public).
And methods should be ordered by visibility (i.e. package private static method must always be before private static method(s))

json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
```diff
@@ -113,22 +116,101 @@ public static ConvertedTypeInfo create(TypeInfo typeInfo, CodegenContext ctx) {
                                                            properties,
                                                            resolvedGenerics))
                 .or(() -> processBuilderInfo(typeInfo, properties, ctx, resolvedGenerics));
+        Optional<PolymorphicInfo> polymorphismInfo = typeInfo.findAnnotation(JsonTypes.JSON_POLYMORPHIC)
```
tomas-langer, line 117: There is a method `hasAnnotation`, so this could be done as a simple if. But if you prefer this approach, I am OK with it.

json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
```diff
@@ -113,22 +116,101 @@ public static ConvertedTypeInfo create(TypeInfo typeInfo, CodegenContext ctx) {
...
+                                     extraProperties,
                                      jsonProperties,
                                      orderComparator,
                                      creatorInfo,
                                      builderInfo,
+                                     polymorphismInfo,
                                      genericParamsWithIndexes);
     }
 
+    private static void discoverExtraProperties(LinkedHashMap<String, String> extraProperties,
```
tomas-langer: Using collection implementation as a parameter is unusual. Maybe `SequencedMap` would be sufficient here?

json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
```diff
@@ -113,22 +116,101 @@ public static ConvertedTypeInfo create(TypeInfo typeInfo, CodegenContext ctx) {
...
+            discoverExtraProperties(extraProperties, interfaceInfo, newTargetType);
+        }
+        if (typeInfo.superTypeInfo().isPresent()) {
+            discoverExtraProperties(extraProperties, typeInfo.superTypeInfo().get(), newTargetType);
+        }
+    }
+
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
+    private static String resolveAliasValue(Annotation subtypeAnnotation) {
+        String alias = subtypeAnnotation.stringValue("alias").get();
```
tomas-langer: Using `.get()` on optionals without checking. I know there is a default value, but if you used 
`subtypeAnnotation.stringValue("alias").orElse("")`
it would be clear what we expect as the default, and if we ever decide not to load defaults, this will just fail.

json/codegen/src/main/java/io/helidon/json/codegen/JsonConverterGenerator.java
```diff
@@ -84,40 +85,361 @@ static void generateConverter(ClassBase.Builder<?, ?> classBuilder, ConvertedTyp
...
     }
 
-    private static void addConfigurationMethod(Method.Builder method, Map<String, TypeToConfigure> toConfigure) {
+    private static void generatePolySerializeMethod(ClassBase.Builder<?, ?> classBuilder,
+                                                    Method.Builder method,
+                                                    ConvertedTypeInfo converterInfo,
+                                                    PolymorphicInfo polymorphicInfo,
+                                                    Map<String, TypeToConfigure> toConfigure) {
+        method.name("serialize")
+                .addParameter(param -> param.name("generator").type(JsonTypes.JSON_GENERATOR))
```
tomas-langer: There is now a shortcut method `.addParameter(TypeName, String)` , so no need for this complex code (maybe it is `(String, TypeName)`, but it is there).

json/json/src/main/java/io/helidon/json/ObjectStartParser.java
```diff
@@ -0,0 +1,328 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.json;
+
+/**
+ * An implementation of the {@link io.helidon.json.JsonParser} which enforces object start as the current value.
```
tomas-langer, line 20: And maybe add "and delegates all other calls to the provided JsonParser"?

json/codegen/src/main/java/io/helidon/json/codegen/ConvertedTypeInfo.java
Verdent: public removed. One method moved.

json/codegen/src/main/java/io/helidon/json/codegen/JsonConverterGenerator.java
Verdent: Oh, I did not know that. Thanks.

json/binding/src/main/java/io/helidon/json/binding/Json.java
Verdent: Ah, missed that. Thanks

tomas-langer (approved):

PR #11213, opened 2026-02-26
Fix SSL performance regression in Nio

re-thc:
### Description

In the new Nio path exceptions are thrown instead of `int` for statuses and the stack trace dominates performance.

This PR restores "previous" performance with the old engine making it parity. Pre-PR there's about ~5-10% difference.

### Documentation

No change.


tomas-langer (changes requested):
`io.helidon.common.socket.TlsNioSocket:212` and `io.helidon.common.socket.TlsNioSocket:226` now cache peer principal and certificate lookups,  but that cache is only cleared when renegotiated() runs at   `io.helidon.common.socket.TlsNioSocket:483`.

 Because `io.helidon.common.socket.PeerInfoImpl:100` and `io.helidon.common.socket.PeerInfoImpl:123` keep resolving through suppliers on the socket, callers that retain an existing PeerInfo instance across renegotiation used to see  the new session identity automatically.

 After this patch, that same PeerInfo can keep  returning the old principal or certificate chain unless the caller explicitly asks the  socket for a fresh remotePeer() first.

 The blocking TLS implementation still re-reads the current session on every call at `io.helidon.common.socket.TlsSocket:111` and  `io.helidon.common.socket.TlsSocket:119`, so this is a behavior regression in the public PeerInfo API.

re-thc:
Updated this so it is session-aware and added regression test on it.

PR #11231, opened 2026-02-27
4.x Add support for exporting Helidon metrics via OTLP

### LINKED ISSUE #11043: Support export of Helidon metrics via OpenTelemetry metrics exporter(s)
Helidon Version: 4.x

## Enhancement Description
Provide a simple way for users to have Helidon transmit data for metrics managed using the Helidon metrics API using the OpenTelemetry OTLP protocol.

tjquinno:
### Description
Resolves #11043 

### Release note
____
As a preview feature, Helidon services can now periodically export Helidon metrics data to backends via OTLP, using configuration or programmatic set-up of the server.

In the `metrics` or `server.features.observe.observes.metrics` config section, add a new `publishers` section. For example:
```yaml
metrics:
  publishers:
    - type: prometheus
    - type: otlp
      url: "http://localhost:9090/api/v1/otlp/v1/metrics"
      interval: PT15S
```
This configuration:
* Preserves the traditional Prometheus (or JSON) format output at the Helidon metrics endpoint.
* Transmits Helidon Metrics data to the specified backend URL every 15 seconds. 

If you specify any entries at all under `publishers`, Helidon does not keep the traditional Prometheus behavior, so providing just the `otlp` entry above would implicitly disable the traditional Prometheus behavior.

Note that configuring multiple publishers can affect performance.
____

### PR Overview
Our (currently) only metrics implementation uses Micrometer. 

A few notes about Micrometer:
* Micrometer's global registry is composite and can contain zero or more "sub-registries," and a composite registry delegates each registry operation to all sub-registries. 
* Different types of registries can publish metrics data differently.
* Each registry creates its own implementation of each meter that is registered and applies each update operation to all implementations of that meter.

A few notes about the Helidon metrics implementation:
* When code registers (or unregisters) a meter using the Helidon metrics API, Helidon delegates to Micrometer's global registry which in turn delegates to all its sub-registries.
* Up to now, Helidon has always added a Micrometer Prometheus meter registry to the composite Micrometer global registry. This allows us to both have actual meter implementations that maintain their data and also publish using the Prometheus format at our endpoint.
* Helidon's metrics observer (which supports the metrics endpoint) uses the Prometheus meter registry directly to prepare its output.

To add the ability to publish using OTLP, the design direction we agreed on was for Helidon internally to use the Micrometer OTLP meter registry. It can be configured: URL to send to, how often to send, etc. We wanted to allow users to keep the old Prometheus behavior as well if they want to, although having multiple meter registries will affect performance; we have not studied by how much.

The external UX, then, is to add configuration (or do the equivalent programmatically) for zero or more metrics "publishers" as a neutral abstraction. To support OTLP, users configure an `otlp` publisher. 

Internally, that translates into preparing a Micrometer OTLP meter registry and adding it to the Micrometer global registry.

Taking this approach not only enables OTLP export but allows us or users to add support for other publishing types--that is, Micrometer meter registry types. (There's one that exports to Datadog, for example.)

#### Config, publishers, providers
`MetricsConfig` now supports the `publishers` section using these new types:

* `MetricsPublisherConfig` - a POJO interface used as a super-interface for concrete publisher config blueprints
* `MetricsPublisher` - POJO interface (which extends `NamedService` for configured provider support)
* `spi.MetricsPublisherProvider` - provider so configured provider support can find implementations

### Changes in `helidon-metrics-providers-micrometer`
* Add two sets of subtypes for the ones listed above
   * Prometheus
   * OTLP

In Micrometer, each meter registry type typically has its own config type (e.g., `OtlpConfig` for `OtlpMeterRegistry`) which gathers the configurable settings for that type of meter registry. Our new `OtlpPublisherConfigBlueprint` and `PrometheusPublisherConfigBlueprint` interfaces follow those Micrometer config types very closely.

#### Our Micrometer-specifid `MetricsFactory` implementation 
The pre-existing `MicrometerMetricsFactory`, during init, used to make sure that there was a Prometheus meter registry in the Micrometer global registry, adding one if not. (In normal situations there was never one already there so we would always add the Prom. one.)

Now, that code uses the publishers to provide the Micrometer meter registries to add to the global composite registry, maintaining backward compatibility if no publishers are explicitly set up.

### Changes in `helidon-webserver-observer-metrics`
To decide whether to set up the metrics endpoint, the `MetricsFeature` now also checks to see if there are any _enabled_ Prometheus meter registries derived from the `publishers` config. If not, it does not prepare the metrics endpoint.

(As a future enhancement, if we wanted to we _could_ relax this restriction. As long as the underlying implementation had meters that retained their data, we could write our own code to retrieve all meters from the Helidon meter registry and prepare the Prometheus-format output ourselves.)

### Documentation
Update included in this PR.

### DIFF
--- a/metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisherConfig.java
+++ b/metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisherConfig.java
@@ -16,6 +16,8 @@
 
 package io.helidon.metrics.api;
 
+import java.util.Optional;
+
 import io.helidon.builder.api.Option;
 
 /**
@@ -24,11 +26,10 @@ import io.helidon.builder.api.Option;
 public interface MetricsPublisherConfig {
 
     /**
-     * Whether the configured publisher is enabled.
+     * Name for this instance of the publisher.
      *
-     * @return true if enabled, false otherwise
+     * @return name of the publisher instance
      */
     @Option.Configured
-    @Option.DefaultBoolean(true)
-    boolean enabled();
+    Optional<String> name();
 }
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisher.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisher.java
@@ -87,7 +87,7 @@ public class OtlpPublisher implements MicrometerMetricsPublisher,
 
     @Override
     public String name() {
-        return OtlpPublisherProvider.TYPE;
+        return config.name().orElse(OtlpPublisherProvider.TYPE);
     }
 
     @Override
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisherConfigBlueprint.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisherConfigBlueprint.java
@@ -24,6 +24,7 @@ import java.util.concurrent.TimeUnit;
 import io.helidon.builder.api.Option;
 import io.helidon.builder.api.Prototype;
 import io.helidon.config.Config;
+import io.helidon.metrics.api.MetricsPublisherConfig;
 import io.helidon.metrics.spi.MetricsPublisherProvider;
 
 import io.micrometer.registry.otlp.AggregationTemporality;
@@ -34,7 +35,7 @@ import io.micrometer.registry.otlp.AggregationTemporality;
 @Prototype.Configured(value = OtlpPublisherProvider.TYPE, root = false)
 @Prototype.Blueprint
 @Prototype.Provides(MetricsPublisherProvider.class)
-interface OtlpPublisherConfigBlueprint extends Prototype.Factory<OtlpPublisher> {
+interface OtlpPublisherConfigBlueprint extends MetricsPublisherConfig, Prototype.Factory<OtlpPublisher> {
 
     /**
      * Whether the configured publisher is enabled.
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
@@ -95,7 +95,7 @@ public class PrometheusPublisher implements MicrometerMetricsPublisher,
 
     @Override
     public String name() {
-        return PrometheusPublisherProvider.TYPE;
+        return config.name().orElse(PrometheusPublisherProvider.TYPE);
     }
 
     @Override
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisherConfigBlueprint.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisherConfigBlueprint.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 
 import io.helidon.builder.api.Option;
 import io.helidon.builder.api.Prototype;
+import io.helidon.metrics.api.MetricsPublisherConfig;
 import io.helidon.metrics.spi.MetricsPublisherProvider;
 
 /**
@@ -29,7 +30,7 @@ import io.helidon.metrics.spi.MetricsPublisherProvider;
 @Prototype.Configured(value = PrometheusPublisherProvider.CONFIG_KEY, root = false)
 @Prototype.Blueprint
 @Prototype.Provides(MetricsPublisherProvider.class)
-interface PrometheusPublisherConfigBlueprint extends Prototype.Factory<PrometheusPublisher> {
+interface PrometheusPublisherConfigBlueprint extends MetricsPublisherConfig, Prototype.Factory<PrometheusPublisher> {
 
     /**
      * Whether the configured publisher is enabled.


metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
```diff
@@ -252,6 +253,26 @@ static List<Tag> createTags(String pairs) {
...
+    @Option.Provider(value = MetricsPublisherProvider.class, discoverServices = false)
+    @Option.Singular
+    List<MetricsPublisher> publishers();
+
+    /**
+     * Indicates if the publishers node has children.
+     *
+     * @return true if the publishers config node exists, false otherwise
+     */
+    boolean publishersConfigured();
```
tomas-langer: Please do not use this method.
If this is created using API and not from config, you would set this to false even if the user configures publishers by hand, and then the state would be wrong.

I think we should have a clear rule, that if no publishers are configured, we use the default one (i.e. we will not allow the case where you explicitly configure an empty list).

If you really need to have this option, then create a method `disablePublishers` or something similar, that would default to `false`, and if set to true, we would ignore `publishers` (this can still be done in the decorator).

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigSupport.java
```diff
@@ -126,6 +126,9 @@ public void decorate(MetricsConfig.BuilderBase<?, ?> builder) {
             if (builder.scoping().isEmpty()) {
                 builder.scoping(ScopingConfig.create());
             }
+            builder.publishersConfigured(builder.config()
```
tomas-langer: This is not a safe way to resolve this - see above.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsFactory.java
```diff
@@ -45,6 +45,11 @@
  */
 public interface MetricsFactory {
 
+    /**
+     * Qualifier for context setting indicating if any pull publishers are present.
+     */
+    String PULL_PUBLISHERS_PRESENT = "pull-publishers";
```
tomas-langer: Using this as a context qualifier for a `boolean` seems a bit too vague. I think that `io.helidon.metrics` prefix would be better here.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisher.java
```diff
@@ -0,0 +1,34 @@
...
+package io.helidon.metrics.api;
+
+import io.helidon.common.config.NamedService;
+import io.helidon.service.registry.Service;
+
+/**
+ * Behavior common to all metrics publishers.
+ */
+@Service.Contract
+public interface MetricsPublisher extends NamedService {
```
tomas-langer, line 26: An interface without methods seems useless. Do we expect the component using the configuration to do instanceof checks?

metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisherConfig.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public interface MetricsPublisherConfig {
+
+    /**
+     * Whether the configured publisher is enabled.
+     *
+     * @return true if enabled, false otherwise
+     */
+    @Option.Configured
+    @Option.DefaultBoolean(true)
```
tomas-langer: Please copy this method to each blueprint. 
We need to eventually solve the `enabled` problem for the whole of Helidon, and abstracting it away will make our life more complicated in the future.

metrics/api/src/main/java/io/helidon/metrics/spi/MetricsPublisherProvider.java
```diff
@@ -0,0 +1,29 @@
...
+
+import io.helidon.common.config.ConfiguredProvider;
+import io.helidon.metrics.api.MetricsPublisher;
+import io.helidon.service.registry.Service;
+
+/**
+ * Marks each configured provider of a {@link io.helidon.metrics.api.MetricsPublisher}.
+ */
+@Service.Contract
+public
```
tomas-langer: Please remove the line break after public, this is a strange formatting.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisher.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public boolean enabled() {
+        return config.enabled();
+    }
+
+    @Override
+    public String name() {
+        return OtlpPublisherProvider.TYPE;
```
tomas-langer: usually this should be `config.name()`

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisherConfigBlueprint.java
```diff
@@ -0,0 +1,143 @@
...
+    Optional<String> prefix();
+
+    /**
+     * URL to which to send metrics telemetry.
+     *
+     * @return backend URL
+     */
+    @Option.Default("http://localhost:4318/v1/metrics")
+    @Option.Configured
+    Optional<String> url();
```
tomas-langer, line 65: Methods named `url` that return `String` are suspisious. If this is to align with MM API, then it may be used, if we design Helidon APIs it should be a different name.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisherConfigBlueprint.java
```diff
@@ -0,0 +1,143 @@
...
+
+import io.micrometer.registry.otlp.AggregationTemporality;
+
+/**
+ * Settings for an OTLP publisher.
+ */
+@Prototype.Configured(value = OtlpPublisherProvider.TYPE, root = false)
+@Prototype.Blueprint
+@Prototype.Provides(MetricsPublisherProvider.class)
+interface OtlpPublisherConfigBlueprint extends MetricsPublisherConfig, Prototype.Factory<OtlpPublisher> {
```
tomas-langer: There is a lot of options here, but according to my search only `enabled`, `prefix`, `url`, and `interval`  are used.
Please either remove unused options, or implement all of them

metrics/providers/micrometer/pom.xml
```diff
@@ -92,6 +96,11 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
                     <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.helidon.config.metadata</groupId>
+                            <artifactId>helidon-config-metadata-codegen</artifactId>
+                            <version>${helidon.version}</version>
+                        </path>
```
tomas-langer, line 103: This section is missing (and similar in `dependencies`):
```xml
                        <path>
                            <groupId>io.helidon.codegen</groupId>
                            <artifactId>helidon-codegen-helidon-copyright</artifactId>
                        </path>
```

This is needed to generate copyrights in prototypes (which are part of our public API and we deliver their source code as well)

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
tjquinno: Here is the problem.

What if a user specifies this:
```
publishers:
    prometheus:
        enabled: false
```
The user has explicitly told us to NOT use the prometheus publisher which is the default for the Micrometer metrics provider.

But, as written `publishers` are handled by the configured provider mechanism which sees that the publisher is disabled and so _does not add it to the list_. 

If, as you suggest, the Micrometer provider code were to look at the `List<MetricsPublisher>` it would find it empty, and would add the default...directly contradicting what the user said.

How do we resolve this?

tjquinno: I've pushed a commit that I think addresses this. Instead of `publishersConfigured` being a configurable setting it is now just a custom prototype method that looks both at the contents of the `List<MetricsPublisher>` and the config node to handle the cases I described above;

metrics/api/src/main/java/io/helidon/metrics/api/MetricsFactory.java
tjquinno: Changed.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisher.java
tjquinno: I think we need this to define the metrics configuration in a provider-independent way...the expanded `MetricsConfigBlueprint` now has a `List<MetricsPublisher> publishers()` method. 

Concrete publishers are specific to a metrics provider but the config in `helidon-metrics-api` needs to be  provider-agnostic. 

Is there an alternative?

Regarding `instanceof` checks, the `MicrometerMetricsFactory` code does just that for the following reason...

Before this change, `MicrometerMetricsFactory` needed to check whether a Micrometer Prometheus meter registry was already set up in the Micrometer global registry and, if not, add one. Two goals:
* Provide a Micrometer meter registry to create meters that actually worked (that is, not no-op meters).
* Support the metrics endpoint Prometheus-format output.

In this PR, users can add or configure the publishers, possibly including a Prometheus one. `MicrometerMetricsFactory` filters the publishers by type using `instanceof` to see if one is a Prometheus registry and, if not, adds the default Prometheus one.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisherConfig.java
tjquinno: changed

metrics/api/src/main/java/io/helidon/metrics/spi/MetricsPublisherProvider.java
tjquinno: Ugh. Fixed.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisherConfigBlueprint.java
tjquinno: I intentionally used the names and types from the Micrometer `OtlpConfig` type. 

I thought about declaring our config name and type to be `URI uri()` which our code could obviously convert to a `String` when preparing the MM `OtlpConfig` instance. But there is always the chance--slim though it might be--that MM could support special keywords for the `url`. Plus there is clarity for users keeping the same names and types.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisherConfigBlueprint.java
tjquinno: Good note. I must have been interrupted as I was completing that part of the code. Fixed.

metrics/providers/micrometer/pom.xml
tjquinno: Added

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
tomas-langer: The main reason I do not like this is that you try to use the config node. 
Please do not use the config at all.
In this case, if the user defines a disabled provider, it will be in the list - i.e. the list will not be empty, and no default will be added.
The default should be added in the decorator in any case

metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisher.java
tomas-langer: The alternative is that there will be a MM (or OTel?) speicfic blueprint defining the `publishers`, as they may not make sense for other providers

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
tjquinno: Doesn't `ProvidedUtil.servicesFromObject` explicitly ignore configured services it finds that are disabled?

https://github.com/helidon-io/helidon/blob/41b6e3702ceead6427017e6bf21f61f0209c19f8/config/config/src/main/java/io/helidon/config/ProvidedUtil.java#L302

I observed this, and because code elsewhere in the PR depended on this behavior that's why I added a test case to the new `TestPublishers#checkPublishersPresent` test in the PR which configures only a disabled `test-publisher` and checks that the size of the `metricsConfig.publishers()` list is 0. The test passes, so the list is empty if the configured metrics publisher is disabled.

As for adding the default in the decorator, a default would have to be an actual implementation that depends on the specific metrics provider on the classpath. 

We could expand the `MetricsPublisherProvider` interface to include a `MetricsPublisher defaultPublisher()` method. Then the decorator could use the service loader to find the concrete implementation of the `MetricsPublisherProvider` that's available and invoke it's `defaultPublisher()` method if it (the decorator) saw that the `publishers` list is empty.

Or is there a way to get the "correct" provider from the configured provider mechanism which already knows about it. I don't think there is--the `ProvidedUtil#servicesFromList` code I looked at uses the retrieved providers internally without saving them anywhere accessible.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsPublisher.java
tjquinno: It seemed logical to me that any conceivable metrics provider (Micrometer, OpenTelemetry, ...) would likely have _some_ way of making the metrics data available to external services..._publishing_ the metrics, in other words. 

Certainly the specific settings that control a publisher's behavior would vary among metrics providers (which the configured providers mechanism handles), but (IMO) a functional metrics implementation would have a way to publish the data that would need to be configured.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/OtlpPublisher.java
tjquinno: changed

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
tomas-langer: I see there is a problem in unfinished design of the providers - we use `enabled` from config, but do not have it enforced in imperative API. 
I will create an issue for the enabled concept in blueprints/providers, so we do not lose track of it.

PR #11233, opened 2026-02-27
File ConfigSource fails generating MD5 digest.

klustria:
### Description
When Sun JCE provider is removed, and a FIPS provider is added, retrieving content of a File ConfigSource will fail on generating an MD5 digest.

### Documentation
File ConfigSource will generate a digest using MD5 by default, and will fall back to SHA-256 if it fails. Alternatively,  a digest algorithm can be specified using **io.helidon.config.file.digest.algorithm** system property.

If no doc impact: JavaDoc is updated to reflect the change, but can also be noted on Advance Configuration Topics of Config from the Helidon Documentation.

### DIFF
--- a/config/config/src/main/java/io/helidon/config/FileSourceHelper.java
+++ b/config/config/src/main/java/io/helidon/config/FileSourceHelper.java
@@ -228,39 +228,22 @@ public final class FileSourceHelper {
 
     static MessageDigest digest() {
         String algorithm = System.getProperty(PROPERTY_DIGEST_ALGORITHM);
-        String[] algorithms;
-        if (algorithm != null) {
-            algorithms = new String[] {algorithm};
-        } else {
-            // MD5 will be the default, and will fall back to SHA-256 if it fails
-            algorithms = new String[] {ALGORITHM_MD5, ALGORITHM_SHA256};
+        if (algorithm == null) {
+            return digest(ALGORITHM_MD5, ALGORITHM_SHA256);
         }
-        return getDigest(algorithms);
+        return digest(algorithm);
     }
 
-    // If multiple algorithms are specified, this will fall back to the next algorithm if the current one fails
-    private static MessageDigest getDigest(String[] algorithms) {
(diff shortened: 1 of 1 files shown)


config/config/src/main/java/io/helidon/config/FileSourceHelper.java
```diff
@@ -218,12 +226,41 @@ private static FileLock lockFile(Path filePath, FileInputStream fis) throws IOEx
         }
     }
 
-    private static MessageDigest digest() {
-        try {
-            return MessageDigest.getInstance("MD5");
-        } catch (NoSuchAlgorithmException e) {
-            throw new ConfigException("Cannot get MD5 digest algorithm.", e);
+    static MessageDigest digest() {
```
tomas-langer, line 229: Can you please simplify this to this code:

```java
    static MessageDigest digest() {
        String algorithm = System.getProperty(PROPERTY_DIGEST_ALGORITHM);
        if (algorithm == null) {
            return digest(ALGORITHM_MD5, ALGORITHM_SHA256);
        }
        return digest(algorithm);
    }

    private static MessageDigest digest(String... algorithms) {
        for (String algorithm : algorithms) {
            try {
                return MessageDigest.getInstance(algorithm);
            } catch (NoSuchAlgorithmException e) {
                LOGGER.log(Level.TRACE,
                           () -> "Cannot get " + algorithm + " digest algorithm", e);
            }
        }
        throw new ConfigException("Cannot get digest algorithm(s): " +  String.join(", ", algorithms));
    }
```

1. thre is no reason to do a `!= null`, when you can do `==` - readability
2. early return - rather than if/else and return after, we can do one if with early return
3. creation of array does not make sense here, when we can use varargs
4. the null check + length check seems overprotective for a private method that is called only from one other method
5. the cycle to find a valid algorithm does not need to be so detailed, we can throw after we cycle through all options

klustria: Updated based on feedback, thanks for simplifying the code!

PR #11234, opened 2026-03-01
Update of JSON media supports

### LINKED ISSUE #100: Add a way to provide supported content types to JsonSupport

### LINKED ISSUE #11192: Update JSON-Binding Media Support to customize content type/accepted types
## Task Description
Similar to what is done in Helidon JSON media support.

Add tests for this customizable configuration (both for JSON and JSON Binding)

### LINKED ISSUE #11216: `JacksonSupport` to support `+json` suffixes
Helidon Version: 4.3.4

## Enhancement Description

Related to #11192 except for Jackson. Support various JSON types such as `application/geo+json` and `application/json-patch+json` with Jackson. It may just a change to use `HttpMediaTypes.JSON_PREDICATE` rather than the existing checking of just testing `MediaTypes.APPLICATION_JSON`.

tomas-langer:
- consistent approach to analysis of media types
- consistent approach in handling charsets
- consistent approach in configuring supports
- added methods with `io.helidon.config.Conifg` and deprecated old ones
- now using service registry instead of service loader in GSON (backward compatible)
- media types are now validated against the `Accept` header of request (for server responses)
- content type and accepted types are now configurable for all JSON supports
- failure to parse charset now returns 415

As part of testing, I found out we have a lot of warnings for generated builders, as they use deprecated Config, added suppress warnings annotation.

Resolves #100 
Probably resolves #11216 
Resolves #11192 

Change in behavior:

- json binding frameworks now always return "COMPATIBLE" and never "SUPPORTS" - this is to make sure we choose a media support that matches the type exactly over media supports that can handle anything, even when the order is not good (i.e. we have JSON Binding first, and JSON second, and you want to write `JsonObject` - binding will return compatible, JSON will return supports, JSON will be used).
- all JSON media supports now can be configured for explicit supported content type (what to set in writer if content type is not set on server response and client request)
- all JSON media supports now can be configured for accepted media types (which types to accept for reader - i.e. we support JSON and JSON-patch for most cases, with added `text/plain` for `JSON-P`, where we have some tests that expect that (discussion pending if this should be supported or not)
- unified check whether a reader/writer should be applied, to be as consistent as feasible
- unified getter for `Charset` to use for reading/writing that throws the correct exception, and looks in the right places
- unified place to configure the `Content-Type`

~There is current a fail in examples, that I will fix either in parallel or once this is merged - the declarative metrics example does not depend on `helidon-http-media-json` and it uses the Helidon version of `JsonObject` to read response entity, so it should explicitly depend on it - fix is here: https://github.com/helidon-io/helidon-examples/pull/260~ (Example PR is now merged)

### DIFF
--- a/http/http/src/main/java/io/helidon/http/HeaderNames.java
+++ b/http/http/src/main/java/io/helidon/http/HeaderNames.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/http/media/media/pom.xml
+++ b/http/media/media/pom.xml
@@ -42,10 +42,6 @@
         </dependency>
         <dependency>
             <groupId>io.helidon.common</groupId>
-            <artifactId>helidon-common-mapper</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.helidon.common</groupId>
             <artifactId>helidon-common-config</artifactId>
         </dependency>
         <dependency>


http/media/media/src/main/java/io/helidon/http/media/EntityWriterBase.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        var acceptCharset = serverRequestHeaders.find(HeaderNames.ACCEPT_CHARSET)
+                .map(Header::getString);
+
+        if (acceptCharset.isPresent()) {
+            var responseContentType = config.contentType()
+                    .withCharset(acceptCharset.get());
+            serverResponseHeaders.contentType(responseContentType);
+            return acceptCharset.map(EntityIoBase::charset);
```
danielkec: 🤖 Support lists and q-values, pick best supported charset, or ignore invalid values instead of returning 415

tomas-langer: `Accept-Charset` should not be used - it is not sent by browsers, and should be ignored by servers. Removed support for it.

romain-grecourt (approved):

PR #11236, opened 2026-03-02
Object deserialization fix

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11237

### Documentation

None


json/binding/src/main/java/module-info.java
```diff
@@ -37,6 +37,7 @@
 
     requires io.helidon.config;
     requires io.helidon.common;
+    requires io.helidon.common.buffers;
```
danielkec, line 40: Intentional?

Verdent: Yes, it is needed because of the Bytes constants used in the ObjectConverter

PR #11249, opened 2026-03-03
Fix addContentLiteral in content builder in class model

### LINKED ISSUE #11245: Class model method `addContentLiteral` does not escape special characters
A simple newline will break it.

tomas-langer:
To ensure we escape special characters (quotes, backslash, new line, tab).

Added a test, updated javadoc.

Resolves #11245

### DIFF
--- a/codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ContentBuilder.java
+++ b/codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ContentBuilder.java
@@ -84,7 +84,7 @@ public interface ContentBuilder<T extends ContentBuilder<T>> {
     default T addContentLiteral(String literal) {
         if (literal.contains("\n")) {
             // the simplest solution is to use multiline string
-            return addContent("\"\"\"\n" + literal + "\n\"\"\"");
+            return addContent("\"\"\"\n" + literal + "\"\"\"");
         }
         // escape tabs, double quote, and backslashes
         String toWrite = literal;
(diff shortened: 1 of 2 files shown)


codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ContentBuilder.java
```diff
@@ -74,12 +74,24 @@ default T addContentLine() {
...
+     * otherwise it will be a quotes string, with escaped tabs, quotes, and backslashes.
      *
      * @param literal the literal string to be added as content
      * @return updated builder instance
      */
     default T addContentLiteral(String literal) {
-        return addContent("\"" + literal + "\"");
+        if (literal.contains("\n")) {
+            // the simplest solution is to use multiline string
+            return addContent("\"\"\"\n" + literal + "\n\"\"\"");
```
tvallin: This adds as a new line at the end of the literal. I do not think this is the expected behavior.

Here is a test to highlight my thought:
```java
@Test
void testWithoutNewLineAtTheEnd() {
    String foo = "foo\nbar";
    var c = new TestContentBuilder();
    c.addContentLiteral(foo);
    assertThat(c.generatedString(), is("    \"\"\"\nfoo\\nbar\"\"\""));
}
```

This would fail : 
```
Expected: is "    \"\"\"\nfoo\nbar\"\"\""
     but: was "    \"\"\"\nfoo\nbar\n\"\"\""
```

tomas-langer: yes, my mistake

PR #11252, opened 2026-03-03
Fix builder regression - generate singular setter with consumer of prototype builder

### LINKED ISSUE #11248: [Blueprint] @Option.Singular does not generate Consumer<Builder> method to the list
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description

@Option.Singular does not generate Consumer\<Builder\> method to the list. This used to work in 4.3.4 release.

## Steps to reproduce

From the `tests/builder` directory, create the following classes:

```java
interface D extends RuntimeType.Api<DConfig> {
    static DConfig.Builder builder() {
        return DConfig.builder();
    }

    static D create(DConfig config) {
        return new D() {
            @Override
            public D build() {
                return this;
            }

            @Override
            public DConfig prototype() {
                return config;
            }
        };
    }

    static D create(Consumer<DConfig.Builder> consumer) {
        return builder().update(consumer).build();
    }
}
```

```java
@Prototype.Blueprint
interface DConfigBlueprint extends Prototype.Factory<D> {

    String name();
}
```

```java
@Prototype.Blueprint
interface UsingDBlueprint {

    @Option.Singular
    List<D> dList();
}
```

Add the following test:

```java
@Test
void test() {
    UsingD.builder().addDList(builder -> builder.name("name")).build();
}
```

The method `addDList` is not generated and create a compile error.

tomas-langer:
For cases where the prototype is a factory of runtime type.
Updated existing test.

Resolves #11248

### DIFF
--- a/builder/tests/factory/src/main/java/io/helidon/builder/tests/factory/UsingBlueprint.java
+++ b/builder/tests/factory/src/main/java/io/helidon/builder/tests/factory/UsingBlueprint.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


tvallin (approved):

tomas-langer:
Added a commit to re-trigger pipelines, as they did not start for last push

PR #11235, opened 2026-03-02
Doc for using the OTel API directly from a Helidon app

tjquinno:
### Description
Resolves #10911 
Resolves #10354 

* Create a new directory `se/telemetry`.
* Move the existing `se/telemetry.adoc` file under the new directory.
* Add a new `se/telemetry/using-opentelemetry-api.adoc` under the new directory.
* Update the `sitegen.yaml` and telemetry snippets files.

There is a companion PR for the examples repo to add a small example app.

### Documentation
This is a doc PR.

PR #11260, opened 2026-03-04
4.x Support OpenTelemetry logging set-up using Helidon configuration

### LINKED ISSUE #11259: 4.x Allow OpenTelemetry logging configuration using Helidon config
Helidon Version: 4.x

## Enhancement Description
Helidon already supports preparing the OpenTelemetry runtime environment using Helidon config for OTel as a whole, tracing, and metrics.

This enhancement adds configuration of OTel logging.

tjquinno:
### Description
Resolves #11259 

#### Release note
____
Users can now control the runtime set-up of OpenTelemetry logging using Helidon config in the `telemetry.signals.logging` config section.
____

#### PR overview
This PR follows the pattern established earlier when we added tracing and metrics config support for OTel.

To add support for logging, there are a few new blueprint interfaces and blueprint support classes.

Working on this enhancement revealed a handful of settings omitted from metrics and tracing, mostly to do with the exporters which send data to other servers.

Also, although OTel uses different exporter types for the three signals (tracing, metrics, and logging) and different processor types, the configuration for exporters does not depend on the signal. Same for processors. So this PR also refactors those blueprints and related code somewhat to take advantage of the consistency across the signals. The code still creates the correct particular type depending on which signal is involved

Now that we have all three signals covered I see ways to simplify some of the code, a task for a future date. But the structure of the `telemetry` section and its subsections will very likely remain as-is. This entire feature is marked as preview, so if a future refactoring could modify the blueprints and therefore the associated generated types although we will try to minimize any backward-incompatible changes.

### Documentation
Update forthcoming in a separate PR. (#11235 already touches and moves the affected file.)

### DIFF
--- a/docs/src/main/asciidoc/se/telemetry/open-telemetry.adoc
+++ b/docs/src/main/asciidoc/se/telemetry/open-telemetry.adoc
@@ -51,7 +51,7 @@ Helidon SE supports OpenTelemetry in several important ways:
 ** Programmatically, using the OpenTelemetry SDK API and the Helidon OpenTelemetry API
 
 * Conforms to the link:{semconv-link}[OpenTelemetry semantic conventions] for automatically-created spans and metrics for HTTP requests
-* Allows xref:{docdir}/metrics/metrics.adoc#usage-publishing[publishing Helidon metrics] to backend systems using OTLP.
+* Allows xref:{rootdir}/se/metrics/metrics.adoc#usage-publishing[publishing Helidon metrics] to backend systems using OTLP.
 
 OpenTelemetry models observability as a set of link:https://opentelemetry.io/docs/concepts/signals/[_signals_]. Each signal--for example metrics, tracing, and logging--is an origin of monitoring data, and each has configurable settings which control its behavior.
 


telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/LogExporterType.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Console {@link io.opentelemetry.exporter.logging.LoggingSpanExporter}.
+     */
+    CONSOLE,
+
+    /**
+     * JSON logging to console {@link io.opentelemetry.exporter.logging.otlp.OtlpJsonLoggingSpanExporter}.
+     */
+    LOGGING_OTLP;
```
tomas-langer, line 45: This name seems wrong - the docs says `JSON to Console`, yet the name seems a bit more abstract.
I would expect the name to be `CONSOLE_JSON`, or the description to be different explaining why the name is not matching the behavior.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -103,12 +113,20 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
...
      */
     Optional<MeterProvider> meterProvider();
 
+    /**
+     * Sets the logger provider that OpenTelemetry should use. Applications can invoke this method to set the OpenTelemetry
+     * logger provider rather than having Helidon prepare one from the logger-related config settings.
+     *
+     * @return OpenTelemetry logger provider
+     */
+    Optional<LoggerProvider> loggerProvider();
```
tomas-langer: I cannot find any usage in the code for this method (and for `meterProvider` and `tracerProvider` - these methods seem redundant.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -93,6 +94,15 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
...
     Optional<OpenTelemetryMetricsConfig> metricsConfig();
 
+    /**
+     * OpenTelemetry logging settings.
+     *
+     * @return logging settings
+     */
+    @Option.Access("")
+    @Option.Configured("signals.logging")
+    Optional<OpenTelemetryLoggingConfig> loggingConfig();
```
tomas-langer: Please add comment (probably just code comment, not javadoc) that this is used only in the decorator, as otherwise this method is never used.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryLoggingConfigBlueprint.java
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Minimum severity level of log records to process. Defaults to {@code Severity.UNDEFINED_SEVERITY_NUMBER}.
+     *
+     * @return minimum severity level
+     */
+    @Option.Configured
+    Optional<Severity> minimumSeverity();
+
+    /**
+     * Whether to include <em>only</em> log records from traces which are sampled. Defaults to {@code false}.
```
tomas-langer: The statement is not true - this method returns `Optional`, so it cannot default to `false`.
Either change the javadoc to "uses library default", or change the method to return `boolean`.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryLoggingConfigSupport.java
```diff
@@ -0,0 +1,104 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.telemetry.otelconfig;
+
+import io.helidon.builder.api.Prototype;
+import io.helidon.common.Errors;
+import io.helidon.common.config.Config;
```
tomas-langer: This class is deprecated and marked for removal, you should not be using it (io.helidon.common.config.Config).

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
tjquinno: I had included these methods so  developers, if they wanted, could construct their own instance programmatically and have Helidon use those. I had neglected to use those values in the decorator.

That said, it turns out that the type of the setting would need to be not the `XxxProvider` (from the OTel API) but the OTel type `SdkXxxProvider`. That would force our API to expose the OTel _SDK_ API (not the pure runtime  API only) and I don't think we should do that. 

So I will remove those three methods from the blueprint.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
tjquinno: Changed. Same for the methods for the other two signals' config.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryLoggingConfigBlueprint.java
tjquinno: Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryLoggingConfigSupport.java
tjquinno: Ugh. Fixed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/LogExporterType.java
tjquinno: (thought I had posted this earlier) 

The enum names correspond as closely as possible to the OpenTelemetry values. See https://opentelemetry.io/docs/languages/java/configuration/#properties-logs and the `otel.logs.exporter` choices, one of which is `logging-otlp`.

Issue #11253, opened 2026-03-03
Upgrade Hibernate to 6.6


barchetta:
## Task Description

Upgrade [Hibernate to 6.6 ](https://hibernate.org/orm/releases/)in Helidon 4.

Related to #10441

PR #11267, opened 2026-03-04
4.x: Upgrades Hibernate to 6.6.44.Final

ljnelson:
Addresses #11253 

Subject to pipelines passing; see #11188


ljnelson:
Looks like native image tests are gating this, will pause for now

PR #11281, opened 2026-03-09
4.x: Improve annotation processing diagnostics

romain-grecourt:
### Description

Improve annotation processing diagnostics

- Add originatingElement to io.helidon.common.types.Annotation
- Update CodegenExceptiont to take multiple originating elements using a varag (backward compatible)
- Update AptProcessor to delegate the logging of CodegenException to AptLogger
- Update AptLogger to unwrap all originating elements correctly

### Documentation

None

### DIFF
--- a/builder/tests/common-types/src/main/java/io/helidon/common/types/AnnotationBlueprint.java
+++ b/builder/tests/common-types/src/main/java/io/helidon/common/types/AnnotationBlueprint.java
@@ -100,6 +100,15 @@ interface AnnotationBlueprint {
     List<Annotation> metaAnnotations();
 
     /**
+     * The element used to create this instance.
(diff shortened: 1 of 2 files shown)


common/types/src/main/java/io/helidon/common/types/AnnotationBlueprint.java
```diff
@@ -99,6 +99,15 @@ interface AnnotationBlueprint {
     @Option.Singular
     List<Annotation> metaAnnotations();
 
+    /**
```
tomas-langer, line 102: This change should be done in `builder/tests/common-types` and the module should be built, the result can then be compared with `common/types`. This is to ensure that common types follow builder approach.

PR #11282, opened 2026-03-09
4.x: Prettify class-model annotation rendering and support nested annotations

romain-grecourt:
## Description

Render annotation over multiple lines if needed to make it prettier. 
This is useful for annotation with deep nested data.

Also make various updates to support deep nested annotations:
- Update class model to wrap `io.helidon.common.types.Annotation` recursively to `io.helidon.codegen.classmodel.Annotation`
- Fix `AnnotationProperty` to unwrap nested values correctly
- Deprecate `AnnotationProperty.constantValue()` for removal in favor of `AnnotationProperty.value()`
- Add an Implementation of `ConstantValue` to carry the constant type and names as well as the value
- Added tests

### Documentation

None

### DIFF
--- a/codegen/class-model/etc/spotbugs/exclude.xml
+++ b/codegen/class-model/etc/spotbugs/exclude.xml
@@ -22,7 +22,6 @@
         xmlns="https://github.com/spotbugs/filter/3.0.0"
         xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0/spotbugs/etc/findbugsfilter.xsd">
 
-    <!-- These paths/URLs come from code or config, not open user input -->
     <Match>
         <Class name="io.helidon.codegen.classmodel.AnnotationParameter" />
         <Method name="isValueType"/>
--- a/codegen/class-model/src/main/java/io/helidon/codegen/classmodel/Annotation.java
+++ b/codegen/class-model/src/main/java/io/helidon/codegen/classmodel/Annotation.java
@@ -122,7 +122,7 @@ public final class Annotation extends CommonComponent {
         builder.typeName(type().genericTypeName());
         for (var parameter : parameters) {
             var value = toTypesValue(parameter.value());
-            builder.putProperty(parameter.name(), AnnotationProperty.create(value));
+            builder.property(parameter.name(), value);
         }
         return builder.build();
     }
--- a/common/types/src/main/java/io/helidon/common/types/AnnotationProperty.java
+++ b/common/types/src/main/java/io/helidon/common/types/AnnotationProperty.java
@@ -68,7 +68,7 @@ public interface AnnotationProperty {
                                                            + " Constant: " + constantType.fqName() + "." + constantName);
             }
         }
-        return new AnnotationPropertyImpl(new EnumValueImpl(constantType, constantName));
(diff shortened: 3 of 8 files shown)


codegen/class-model/etc/spotbugs/exclude.xml
```diff
@@ -0,0 +1,31 @@
...
+    limitations under the License.
+
+-->
+
+<FindBugsFilter
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xmlns="https://github.com/spotbugs/filter/3.0.0"
+        xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.0/spotbugs/etc/findbugsfilter.xsd">
+
+    <!-- These paths/URLs come from code or config, not open user input -->
```
tomas-langer: Wrong comment for this type of match. Also why is this happening? Maybe there is a real problem?

codegen/class-model/src/main/java/io/helidon/codegen/classmodel/Annotation.java
```diff
@@ -115,52 +118,126 @@ public io.helidon.common.types.Annotation toTypesAnnotation() {
...
-            if (value instanceof Annotation annot) {
-                builder.putValue(parameter.name(), annot.toTypesAnnotation());
-            } else {
-                builder.putValue(parameter.name(), parameter.value());
-            }
+        var builder = io.helidon.common.types.Annotation.builder();
+        builder.typeName(type().genericTypeName());
+        for (var parameter : parameters) {
+            var value = toTypesValue(parameter.value());
+            builder.putProperty(parameter.name(), AnnotationProperty.create(value));
```
tomas-langer: There is now a shortcut method `builder.property(String, Object)` to bypass the need to create `AnnotationProperty`

common/types/src/main/java/io/helidon/common/types/AnnotationProperty.java
```diff
@@ -106,8 +99,12 @@ static <T extends Enum<T>> AnnotationProperty create(Class<T> type, T value) {
      * {@link EnumValue}.
      *
      * @return constant value if defined
+     * @deprecated use {{@link #value()}} and downcast to {@link ConstantValue}
      */
-    Optional<ConstantValue> constantValue();
+    @Deprecated(since = "4.4.0", forRemoval = true)
+    default Optional<ConstantValue> constantValue() {
```
tomas-langer, line 105: This is wrong. Both constant value and value must be present, as otherwise you may not be able to obtain the actual value of the annotation if needed.

Example:
```
// in class Weighted
public static final double DEFAULT_WEIGHT = 100.0;

// this is what we want to generate
@Weight(Weighted.DEFAULT_WEIGHT)
```

The Annotation property must contain both `100.0` as the value, and `io.helidon.common.Weighted` as the constant class and `DEFAULT_WEIGHT` as the constant name)

romain-grecourt: For posterity, the fix implemented added a new method `value()` to `ConstantValue`.

`EnumValue` (sub-type of `ConstantValue`) has been updated to override the new method with a default implementation that returns itself (same semantic as `enum`).

A new implementation of `ConstantValue` is added (`ConstantValueImpl`) to retain both the value and the constant ref pair (type name and ref name).

PR #11283, opened 2026-03-09
4.x: Codegen Unit Testing

romain-grecourt:
### Description

Codegen Unit Testing

Add a new module helidon-codegen-testing to facilitate writing unit tests for code generators.
The module contains a unit test that demonstrates how to use it.

### Documentation

None

### DIFF
--- a/codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
+++ b/codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
@@ -22,6 +22,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Supplier;
 
 import javax.annotation.processing.Processor;
@@ -199,6 +200,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addProcessor(Processor processor) {
+            Objects.requireNonNull(processor, "processor is null");
+
             this.processors.add(() -> processor);
             return this;
         }
@@ -210,6 +213,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addProcessor(Supplier<Processor> processor) {
+            Objects.requireNonNull(processor, "processor is null");
+
             this.processors.add(processor);
             return this;
         }
@@ -246,6 +251,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addClasspath(Class<?> clazz) {
+            Objects.requireNonNull(clazz, "clazz is null");
+
             classpath.add(TestPaths.paths(clazz));
             return this;
         }
@@ -282,6 +289,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addClasspathEntry(Path path) {
+            Objects.requireNonNull(path, "path is null");
+
             this.classpath.add(path);
             return this;
         }
@@ -293,6 +302,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addClasspathEntries(List<Path> paths) {
+            Objects.requireNonNull(paths, "paths is null");
+
             classpath.addAll(paths);
             return this;
         }
@@ -304,6 +315,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addModulePath(Class<?> clazz) {
+            Objects.requireNonNull(clazz, "clazz is null");
+
             modulepath.add(TestPaths.paths(clazz));
             return this;
         }
@@ -340,6 +353,8 @@ public final class TestCompiler {
          * @return this builder
          */
         public Builder addModulePathEntry(Path path) {
+            Objects.requireNonNull(path, "path is null");
+
             this.modulepath.add(path);
             return this;
         }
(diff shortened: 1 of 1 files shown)


codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
```diff
@@ -0,0 +1,407 @@
...
+
+    /**
+     * Compile result.
+     *
+     * @param success      success
+     * @param classOutput  class output location
+     * @param sourceOutput source output location
+     * @param diagnostics  rendered diagnostics
+     */
+    public record Result(boolean success,
```
tomas-langer: We should not use records as parts of public API, as they have a public constructor, so they make backward compatibility hard to maintain (i.e. if we ever decide to switch to interface).

Also inner types should be last in the source file.

codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
```diff
@@ -0,0 +1,407 @@
...
+        opts.add("--release");
+        opts.add("21");
+    }
+
+    /**
+     * Copy an existing instance.
+     *
+     * @param testCompiler instance to copy
+     */
+    public TestCompiler(TestCompiler testCompiler) {
```
tomas-langer: We never use public constructors in Helidon.

codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    /**
+     * Add processors.
+     *
+     * @param processors processors
+     * @return this instance
+     */
+    @SafeVarargs
+    public final TestCompiler processors(Supplier<Processor>... processors) {
```
tomas-langer: This seems to be a builder of something, it should follow builder pattern.

codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
```diff
@@ -0,0 +1,407 @@
...
+/**
+ * Fluent API to invoke the Java compiler programmatically for testing.
+ * <p>
+ * <b>This is NOT part of any supported API.
+ * If you write code that depends on this, you do so at your own risk.
+ * This code and its internal interfaces are subject to change or deletion without notice.</b>
+ * </p>
+ */
+@SuppressWarnings("ALL")
+public class TestCompiler {
```
tomas-langer: This class is not consistent with methods we use on builders to set/add things in collections.

codegen/testing/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    </parent>
+    <artifactId>helidon-codegen-testing</artifactId>
+    <name>Helidon Codegen Testing</name>
+    <description>
+        Tools for testing codegen extensions
+    </description>
+
+    <properties>
+        <helidon.services.skip>true</helidon.services.skip>
+        <maven.deploy.skip>true</maven.deploy.skip>
```
tomas-langer, line 35: This means we will not be able to use this approach in any other repository (such as MCP, which contains its own code generators)

codegen/testing/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+            <groupId>io.helidon.codegen</groupId>
+            <artifactId>helidon-codegen</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-all</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
```
tomas-langer, line 55: does this need to be a transitive dependency?

codegen/testing/pom.xml
romain-grecourt: Do you think this is ready to be consumed outside of the Helidon repository in its current shape ?

codegen/testing/src/main/java/io/helidon/codegen/testing/TestCompiler.java
romain-grecourt: The blueprint generate builder have methods named after the property and `addXXX` variants.
Can you describe the method names you want ? 

```java
TestCompiler.builder()
	.currentRelease()
	.addProcessors(new ProcessorImpl())
	.addClasspath(AcmeAnnotation.class)
	.printDiagnostics(false)
```

Doesn't look as good as:
```java
TestCompiler.builder()
	.currentRelease()
	.processors(new ProcessorImpl())
	.classpath(AcmeAnnotation.class)
	.printDiagnostics(false)
```

tomas-langer: I think that this is fine and aligned with Helidon,  and that is how I have refactored it:

```java
TestCompiler.builder()
	.currentRelease()
	.addProcessor(new ProcessorImpl())
	.addClasspath(AcmeAnnotation.class)
	.printDiagnostics(false)
```

codegen/testing/pom.xml
tomas-langer: Let' wait with that

Issue #11298, opened 2026-03-10
[4.3.4] SSE bug: Http1ServerResponse.sink() does not mark response as having an entity, causing false "A route MUST call either send, reroute, or next" error


hrstoyanov:
## Environment Details
* Helidon Version: 4.3.4
* Helidon SE or Helidon MP: SE
* JDK version: 25.0.2
* OS: MacOs Tahoe

----------

Another show-stopper for me:

### Bug description

Http1ServerResponse.sink() creates a sink that writes directly to the low-level dataWriter, but never marks the response as having an entity. This causes Helidon's routing to think no response was produced.

The check (HttpRoutingImpl.java, lines 196-211)

After every handler returns, the router checks in order:

1. response.shouldReroute()  → false
2. response.isNexted()       → false
3. response.hasEntity()      → false  ← BUG: should be true
4. → throws "A route MUST call either send, reroute, or next"

Why hasEntity() returns false?

Http1ServerResponse.hasEntity() (line 274):

```java
public boolean hasEntity() {
    return isSent || streamingEntity;
}
```

- isSent — only set to true in the closeRunnable (line 326), which runs when the sink is closed, not when it's created. An SSE sink is long-lived — it may not close until much later.
- streamingEntity — only set to true by outputStream() (line 451). The DataWriterSseSink bypasses outputStream() entirely — it writes directly to ctx.dataWriter().writeNow() (DataWriterSseSink line 124).

So, after the handler calls response.sink(SseSink.TYPE), writes status+headers, and emits SSE events, neither flag is set.

## Suggested fix

In Http1ServerResponse.sink() (line 308), after the sink is successfully created, mark the response:

```java
public <X extends Sink<?>> X sink(GenericType<X> sinkType) {
    for (SinkProvider<?> p : SINK_PROVIDERS) {
        if (p.supports(sinkType, request)) {
            X sink = (X) p.create(...);
            this.isSent = true;   // ← the fix: sink took over the response
            return sink;
        }
    }
    ...
}
```

Setting isSent = true immediately is correct because the sink's constructor (DataWriterSseSink line 76) already calls writeStatusAndHeaders() — the HTTP status and headers have been flushed to the wire. The response is sent at that point. 

Setting isSent = true in sink() is the right approach because:
     1. Headers are already written - DataWriterSseSink constructor calls writeStatusAndHeaders() immediately
     2. Sink owns the response - Once created, only the sink should write to the response
[...]

PR #11305, opened 2026-03-11
Fix: Http1ServerResponse.sink() does not mark response as having an entity

hrstoyanov:
### Problem
When `Http1ServerResponse.sink()` creates an SSE sink, it writes status+headers to the wire immediately, but never marks the response as having an entity. This causes Helidon's routing layer to throw:

```
IllegalStateException: A route MUST call either send, reroute, or next on ServerResponse
```

even though the response was already sent via the sink.

### Root Cause
- `isSent` is only set in the sink's `closeRunnable()`, which runs when the sink closes
- `streamingEntity` is only set by `outputStream()`, which SSE sinks bypass entirely
- After the handler returns, `hasEntity()` returns `false`, triggering the false error

### Solution
Set `isSent = true` immediately after `p.create()` succeeds in the `sink()` method, because the sink's constructor already writes status+headers to the wire and has taken control of the response.

### Changes
- `helidon/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java`
  - Line ~313: Primary sink creation path - capture sink, set `isSent`, return
  - Line ~343: Deprecated fallback path - capture sink, set `isSent`, return

### Testing
Verified with Server-Sent Events (SSE) handlers that emit events without calling `res.send()`. The routing layer now correctly recognizes the response as handled without requiring workarounds.

### Impact
This fix allows SSE-based handlers to work correctly without requiring developers to call `res.send()` after emitting SSE events (which was a workaround for this bug).

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
@@ -310,7 +310,7 @@ class Http1ServerResponse extends ServerResponseBase<Http1ServerResponse> {
         for (SinkProvider<?> p : SINK_PROVIDERS) {
             if (p.supports(sinkType, request)) {
                 try {
-                    return (X) p.create(new SinkProviderContext() {
+                    X sink = (X) p.create(new SinkProviderContext() {
                         @Override
                         public ServerResponse serverResponse() {
                             return Http1ServerResponse.this;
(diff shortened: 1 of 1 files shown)


hrstoyanov:
I have signed OCA in the past, please accept the  fix for #11298 - it is very critical.

romain-grecourt:
Can you find yourself in the list of contributors ? https://oca.opensource.oracle.com/?ojr=contrib-list

Did you use the correct email on your commits ?

hrstoyanov:
Superseded by PR #11309

PR #11306, opened 2026-03-11
Fix #11299: Prevent write-after-done error in metrics OPTIONS request

### LINKED ISSUE #11299: 4.x Metrics endpoint incorrectly handles unrecognized accepted media type on OPTIONS request
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Sending a `GET` to `/metrics` with `Accept` specifying a bad media type causes a strange write-after-done error.

It turns out that the `MetricsFeature#optionsMatching` method fails to `return` after preparing and sending the error response, causing the rest of the method to try to perform normal processing which tries to send more information to the client even thought the response has already gone.

## Steps to reproduce
Run any app that supports the metrics endpoint.

Run `curl -H "Accept: x/x" -X OPTIONS http://localhost:8080/metrics`

The server reports an exception like this:
```list
2026.03.10 11:15:06 WARNING io.helidon.webserver.http.ErrorHandlers VirtualThread[#85,[0x63d2b0de 0x0224fc34] WebServer socket]/runnable@ForkJoinPool-1-worker-7: [0x63d2b0de 0x0224fc34] Request failed: HttpPrologue[protocol=HTTP, protocolVersion=1.1, method=OPTIONS, uriPath=/metrics, query=, fragment=], cannot send error response, as response already sent
```

wadhwaroh-lang:
Prevents write-after-done error when OPTIONS request has unrecognized media type. Method now properly returns after sending error response instead of continuing execution.

### Description

Fixes #11299: Prevent write-after-done error in metrics OPTIONS request

The `MetricsFeature#optionsMatching` method fails to return after sending an error response for unrecognized media types on OPTIONS requests to the `/metrics` endpoint. This causes the method to continue execution and attempt to send additional data to an already-closed response, resulting in a "write-after-done" error.

**Problem Reproduction:**
```bash
curl -H "Accept: x/x" -X OPTIONS http://localhost:8080/metrics

### Documentation
Root Cause: The optionsMatching method sends a 405 METHOD_NOT_ALLOWED response when bestAcceptedForMetadata(req) returns null (indicating an unrecognized media type), but then continues execution instead of returning. This causes the subsequent call to getOrOptionsMatching() to attempt further writes to the closed response.

Solution: Added a missing return statement immediately after sending the error response in the optionsMatching method to prevent continued execution.

None - This is a bug fix for incorrect error handling. No documentation changes required.

### DIFF
--- a/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
+++ b/webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
@@ -314,6 +314,7 @@ class MetricsFeature {
             res.header(ALLOW, "GET");
             res.status(METHOD_NOT_ALLOWED_405);
             res.send();
+            return; 
         }
 
         getOrOptionsMatching(mediaType, req, res, () -> outputMetadata(mediaType,


wadhwaroh-lang:
“Same fix already merged via #11311”

PR #11311, opened 2026-03-11
Address error when an unrecognized media type is specified in an OPTIONS request

### LINKED ISSUE #11299: 4.x Metrics endpoint incorrectly handles unrecognized accepted media type on OPTIONS request

tjquinno:
### Description
Resolves #11299 

There was a missing `return` after sending the "not allowed" response that incorrectly allowed the code to try to add more to the response even after it had been sent.

### Documentation
Bug fix. No doc update.


webserver/observe/metrics/src/test/java/io/helidon/webserver/observe/metrics/TestOptionsNoLogging.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    void testOptionsDoesNotLog() {
+        try (InMemoryLoggingHandler loggingHandler = InMemoryLoggingHandler.create(Logger.getLogger(""));
+             HttpClientResponse response = client.method(Method.OPTIONS)
+                     .path("/observe/metrics")
+                     .accept(MediaTypes.create("x/x"))
+                     .request()) {
+            assertThat("Response", response.status().code(), is(Status.METHOD_NOT_ALLOWED_405.code()));
+            // Need to wait for the server to try to add more to the already-sent response in order to trigger the warning message.
+            pause(250);
```
spericas: This looks problematic in a slow pipeline run, doesn't it? Could we use an "assert eventually" strategy here instead?

tjquinno: Tomas had a similar comment in the companion PR which deals with the backport of this to `helidon-4.x`.

As I said there, the challenge is that the desired outcome is no log message...which is also the result if the test checks too soon; by its nature the bug occurs--and the message is logged--because the server tries to add more to the response after it has sent the response, so the client can easily receive the response and check the log handler and find no messages because the server has not gotten that far yet.

Therefore, a test would have to wait the "correct" amount of time after receiving the response before checking the log handler, but of course it's hard to know exactly what that is.

This is a simple enough fix so we could probably remove the test entirely. I've reproduced the problem without the fix and verified that the fix resolves it.

spericas: Better not add a problematic test like this

tjquinno: I have removed the test.

PR #11314, opened 2026-03-11
4.x backport options metrics feature fix

### LINKED ISSUE #11313: [4.x backport] Metrics endpoint incorrectly handles unrecognized accepted media type on OPTIONS request
Backport of #11299  to `helidon-4.x`

tjquinno:
### Description
Resolves #11313 

Backport of #11311 to helidon-4.x


webserver/observe/metrics/src/test/java/io/helidon/webserver/observe/metrics/TestOptionsNoLogging.java
```diff
@@ -0,0 +1,75 @@
...
+    @Test
+    void testOptionsDoesNotLog() {
+        try (InMemoryLoggingHandler loggingHandler = InMemoryLoggingHandler.create(Logger.getLogger(""));
+             HttpClientResponse response = client.method(Method.OPTIONS)
+                     .path("/observe/metrics")
+                     .accept(MediaTypes.create("x/x"))
+                     .request()) {
+            assertThat("Response", response.status().code(), is(Status.METHOD_NOT_ALLOWED_405.code()));
+            // Need to wait for the server to try to add more to the already-sent response in order to trigger the warning message.
+            pause(250);
```
tomas-langer: This is not good for our pipelines. An intermittent failure waiting to happen.
It will slow down build on fast machines, and break it on slow machines. 
Try to find a better way to handle this.
- a loop with shorter delays, but more repetitions (i.e. wait up to 5 seconds, but in 100 ms cycles, or increment the sleep time for each cycle)
- or have some way to handle this with a timeout (i.e. a countdownlatch counted down by the logging handler when a matching record is received)

tjquinno: I wish this were possible, but the retry approach cannot work here.

The log message will appear only if the bug occurs; if the bug is fixed then there is no log message. 

Therefore, there is no condition to retry until it become true (within a time limit) because the "success" condition (no log message) is also what we find if the log messages are checked too soon, before the logging has had a chance to occur. The nature of the bug is that the response is sent and then the server tries to add more to the response, triggering the log message.

This is a simple enough fix so we could probably remove the test entirely. I've reproduced the problem without the fix and verified that the fix resolves it.

tjquinno: I have removed the test.

PR #11320, opened 2026-03-12
Refactor OCI metrics for current builder/config and for extension

### LINKED ISSUE #42: Refactor OCI metrics integration to use newer blueprint/builder/service provider style
Helidon Version: 4.x

## Task Description

The existing OCI metrics integration uses the older builder approach and config keys in camel case rather than hyphenated.

In this task I'll refactor the code there to preserve backward compatibility with the older implementation but primarily support the newer approach and conventions.

The changes will also add a Helidon service provider for the OCI `Monitoring` instance.

### LINKED ISSUE #11319: Refactor OCI metrics integration to use newer blueprint/builder/service provider style
Helidon Version: 4.x

## Task Description

The existing OCI metrics integration uses the older builder approach and config keys in camel case rather than hyphenated.

In this task I'll refactor the code there to preserve backward compatibility with the older implementation but primarily support the newer approach and conventions.

The changes will also add a Helidon service provider for the OCI `Monitoring` instance.

tjquinno:
### Description
Resolves helidon-io/helidon-extensions#42 

~~Depends on helidon-io/helidon#11334~~ (now merged)

### Release Note
____
Users incorporating Helidon's OCI metrics integration into their services should migrate to a new configuration pattern (using the following dependency):
```xml
<dependency>
    <groupId>io.helidon.integration.oci.metrics</groupId>
    <artifactId>helidon-integrations-oci-metrics</artifactId>
</dependency>
```

```yaml
metrics:
  publishers:
    - type: oci
      compartment-id: my-compartment
      resource-group: my-rg
      initial-delay: PT5S
      delay: PT2.5S
```
See the documentation under `integrations -> OCI -> metrics` for all the available settings.

The functionality is the same as with the earlier, now-deprecated configuration style: determining how Helidon periodically publishes Helidon metrics data to the OCI monitoring back-end system.

Developers can also obtain the relevant `Monitoring` instance using `Services.get(Monitoring.class)`.
____

### PR overview

The existing OCI Metrics integration uses the older-style builder and config approach, and also uses camel-cased instead of hyphenated config keys.

This PR re-implements the integration as follows:
* Adds a `@Service.Singleton Supplier<Monitoring>` which uses the `OciConfig` from the service registry to prepare the instance.
* Adds the `OciMetricsConfigBlueprint` which constructs the new `OciMetricsService` (intended to be functionally equivalent to the now-deprecated `OciMetricsSupport`). The config part now uses the current builder/config style and hyphenated config keys for the properties.
* The OCI metrics "push" feature now primarily appears as a metrics publisher in the same way we now do with the recently-added `otlp` publisher and the `prometheus` publisher. Settings for the `oci` publisher appear like this:
   
   ```yaml
   metrics:
     publishers:
       - type: oci
         compartment-id: my-compartment
         resource-group: my-rg
         initial-delay: PT5S
         delay: PT2.5S
   ```
   
   As before, the settings described above control how the OCI metrics service prepares and schedules the transmission of the metrics data...in other words, these values control the _use of_ the `Monitoring` instance which actually connects to the backend; these settings do not set up the `Monitoring` instance itself. That is handled in one of two ways:
   * The developer can programmatically provide the `Monitoring` instance to the builder. 
   * If the developer does not provide it to the builder, the code gets the `Monitoring` instance from the service registry which uses conventional OCI configuration to prepare the `Monitoring` instance.
* The pre-existing `OciMetricsSupport` class now delegates to the new `OciMetricsService` class so we have the logic in one place, but `OciMetricsSupport` continues to present the older public API to the users (API and config keys).
* The new `OciMetricsService` class _is not_ an `HttpService`. The old one was so it could be informed of webserver start-up and shutdown. Instead, the new class sets up the scheduled executor (to send the data periodically) upon construction of the runtime object from config and registers as a Helidon shutdown hook so it can stop the executor when the server is shutting down.
* The new class also participates in CRaC so the publisher resumes pushing when the app is resumed.
* The test directory includes a blueprint and a configured provider for config which extends the new `OciMetricsConfig` which make sure we can extend the public OCI metrics types in a way that provides synonyms for certain configured properties.

The original API remains intact for backward compatibility although is marked as deprecated.

### Documentation
Update to SE OCI Integration page included in the PR.


integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsService.java
```diff
@@ -0,0 +1,255 @@
...
+    void prepareAutoStartup() {
+        startup();
+    }
+
+    /**
+     * Prepares for automatic shutdown of the periodic "push" processing when the server shuts down. Might be overridden
+     * by a subclass that provides for explicit start-up and shutdown.
+     */
+    void prepareAutoShutdown() {
+        Main.addShutdownHandler(this);
```
danielkec, line 148: This lifecycle is now tied to `Main.addShutdownHandler(this)`, which runs only during JVM shutdown. Because the scheduler is started during construction, an `oci` publisher created from `metrics.publishers` can keep running after `WebServer.stop()` in the same JVM. That is a behavior change from the previous `HttpService`-based lifecycle.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsConfigSupport.java
```diff
@@ -0,0 +1,37 @@
...
+class OciMetricsConfigSupport {
+
+    private OciMetricsConfigSupport() {
+    }
+
+    static class BuilderDecorator implements Prototype.BuilderDecorator<OciMetricsConfig.BuilderBase<?, ?>> {
+        @Override
+        public void decorate(OciMetricsConfig.BuilderBase<?, ?> target) {
+            if (target.monitoringClient().isEmpty()) {
+                target.monitoringClient(Services.get(Monitoring.class));
```
danielkec, line 33: This eagerly resolves `Services.get(Monitoring.class)` during config decoration. That means `type: oci` with `enabled: false` can still fail startup if OCI auth or `Monitoring` is not configured, because the lookup happens before the later `enabled()` guard in `OciMetricsService`.

docs/src/main/asciidoc/se/integrations/oci.adoc
```diff
@@ -84,9 +85,19 @@ include::{rootdir}/includes/dependencies.adoc[]
 </dependency>
 ----
 
+[source,xml]
+.Adding the Helidon OCI Metrics Integration dependency
+----
+<dependency>
+    <groupId>io.helidon.integration.oci.metrics</groupId>
```
tomas-langer, line 92: wrong group id - missing `s` in `integrations`

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsService.java
```diff
@@ -0,0 +1,255 @@
...
+        scheduledExecutorService = Executors.newScheduledThreadPool(1, Thread.ofVirtual().factory());
+        scheduledExecutorService.scheduleAtFixedRate(() -> Contexts.runInContext(ctx, this::pushMetrics),
+                                                     prototype.initialDelay().toNanos(),
+                                                     prototype.delay().toNanos(),
+                                                     TimeUnit.NANOSECONDS);
+    }
+
+    private void pushMetrics() {
+        List<MetricDataDetails> allMetricDataDetails = ociMetricsData.getMetricDataDetails();
+        LOGGER.log(System.Logger.Level.TRACE, String.format("Processing %d metrics", allMetricDataDetails.size()));
```
tomas-langer, line 192: This should be guarded by `if (LOGGER.isLoggable(....)`, as it invokes a `String.format` every time, and calls `.size()` every time.

integrations/oci/metrics/metrics/src/main/java/io/helidon/integrations/oci/metrics/OciMetricsService.java
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Prepares for automatic shutdown of the periodic "push" processing when the server shuts down. Might be overridden
+     * by a subclass that provides for explicit start-up and shutdown.
+     */
+    void prepareAutoShutdown() {
+        Main.addShutdownHandler(this);
+    }
+
+    private void prepareTransmission() {
+        LOGGER.log(System.Logger.Level.TRACE, "Starting OCI Metrics publisher {0}", name());
```
tomas-langer, line 152: Please update all log statements with level below `INFO` to be guarded by `isLoggable`

tomas-langer (changes requested):

Issue #9164, opened 2024-08-19
Server-sent event (SSE) streams do not send a status-line until an event is emitted


tomwhoiscontrary:
## Environment Details
* Helidon Version: 4.1.0
* Helidon SE or Helidon MP: SE
* JDK version: Temurin 21.0.4+7
* OS: Ubuntu
* Docker version (if applicable): N/A

----------

## Problem Description

If you send an SSE response using an SseSink as described here:

https://helidon.io/docs/v4/se/sse

But time elapses between creating the sink and emitting the first event, then the HTTP status-line is not sent until the event is emitted. I think this is because the status-line and headers are not pushed out until the first byte is written to the output stream, which commits them.

If you are emitting events slowly (say, for real world events which do not happen often), this leads to a slightly odd HTTP transaction where the client sends a request, then nothing happens for minutes, hours, etc, then the server sends a status-line, headers, and an event. This might not be illegal, but it is definitely unhelpful. For example, browsers typically have a 300 second response timeout; if they have not seen a status-line in five minutes, they will abandon the request. In non-browser clients, I typically use short timeouts, on the order of seconds, to detect wedged servers.

I think it would be preferable if the status-line and headers were sent immediately on the creation of the sink. There could then be a long wait until the first event is sent, but that's fine - clients expect that.

## Steps to reproduce

I'm happy to write a self-contained demonstration of this if it's helpful. But i'm in a rush right now - please let me know if you would like such a thing.

tomwhoiscontrary:
Great, i look forward to trying this out!

PR #11325, opened 2026-03-13
Fix HTTP/2 SSE sink support and isSent flag handling (#11053, #11298)

### LINKED ISSUE #11053: Support SSE on HTTP/2

### LINKED ISSUE #11298: [4.3.4] SSE bug: Http1ServerResponse.sink() does not mark response as having an entity, causing false "A route MUST call either send, reroute, or next" error

hrstoyanov:
Implements sink() method in Http2ServerResponse to enable Server-Sent Events (SSE) and other streaming use cases for HTTP/2 connections, and fixes incorrect isSent flag handling in both HTTP/1 and HTTP/2 server responses.

## Summary

This PR consolidates fixes for two related issues:
- **Issue #11053**: HTTP/2 SSE sink support - implements the missing sink() method in Http2ServerResponse
- **Issue #11298**: isSent flag not set immediately when sink() is called, causing IllegalStateException

### Http2ServerResponse.java
- Added sink(GenericType<X>) method implementation to support Sink providers (e.g., SseSink) for HTTP/2 server responses
- Implemented handleSinkData() helper method to properly write sink data through the media context
- Added SINK_PROVIDERS static field using HelidonServiceLoader to discover SinkProvider implementations

### Http1ServerResponse.java
- Fixed isSent flag handling: now set to true immediately when sink() is called, not just in the sink's closeRunnable
- Prevents IllegalStateException from routing layer claiming no send/reroute/next was called

### module-info.java
- Added 'uses io.helidon.webserver.http.spi.SinkProvider;' declaration for HTTP/2 module

### New Test: Http1ServerResponseSinkIsSentTest.java
- Verifies the fix for issue #11298
- Tests SSE endpoint using sink() without explicit send() call
- Validates isSent flag is properly set when sink is created

## Technical Details

The sink() method:
1. Iterates through discovered SinkProviders to find one supporting the requested sink type
2. Creates SinkProviderContext with server response, request, connection context, and close runnable
3. Sets isSent = true immediately upon successful sink creation
4. Falls back to deprecated create() method for backward compatibility
5. Throws HttpException (406 NOT_ACCEPTABLE) if no provider found

## Testing
- New test: Http1ServerResponseSinkIsSentTest verifies isSent behavior
- Existing SSE tests validate end-to-end functionality

## Related Issues
Fixes #11053
Fixes #11298

### DIFF
--- a/webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
+++ b/webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
@@ -24,19 +24,14 @@ import java.nio.charset.StandardCharsets;
 import java.util.Optional;
 
 import io.helidon.common.GenericType;
-import io.helidon.common.buffers.BufferData;
 import io.helidon.common.media.type.MediaType;
 import io.helidon.common.media.type.MediaTypes;
-import io.helidon.http.DateTime;
 import io.helidon.http.Header;
 import io.helidon.http.HeaderNames;
 import io.helidon.http.HttpMediaType;
-import io.helidon.http.ServerRequestHeaders;
 import io.helidon.http.ServerResponseHeaders;
 import io.helidon.http.Status;
 import io.helidon.http.WritableHeaders;
-import io.helidon.http.encoding.ContentEncoder;
-import io.helidon.http.encoding.ContentEncodingContext;
 import io.helidon.http.media.EntityWriter;
 import io.helidon.http.media.MediaContext;
 import io.helidon.http.sse.SseEvent;
@@ -64,8 +59,6 @@ class DataWriterSseSink implements SseSink {
     private static final byte[] SSE_DATA = "data:".getBytes(StandardCharsets.UTF_8);
     private static final byte[] SSE_EVENT = "event:".getBytes(StandardCharsets.UTF_8);
     private static final byte[] SSE_COMMENT = ":".getBytes(StandardCharsets.UTF_8);
-    private static final byte[] OK_200 = "HTTP/1.1 200 OK\r\n".getBytes(StandardCharsets.UTF_8);
-    private static final byte[] DATE = "Date: ".getBytes(StandardCharsets.UTF_8);
     private static final WritableHeaders<?> EMPTY_HEADERS = WritableHeaders.create();
 
     private final ServerResponse response;
@@ -80,23 +73,22 @@ class DataWriterSseSink implements SseSink {
         this.mediaContext = ctx.listenerContext().mediaContext();
         this.closeRunnable = context.closeRunnable();
 
-        // output stream to write headers
-        OutputStream headersOutputStream = new DataWriterOutputStream(ctx.dataWriter());
-
-        // check for content encoding
-        ContentEncoder encoder = null;
-        ServerRequestHeaders requestHeaders = context.serverRequest().headers();
-        ContentEncodingContext encodingContext = ctx.listenerContext().contentEncodingContext();
-        if (encodingContext.contentEncodingEnabled() && requestHeaders.contains(HeaderNames.ACCEPT_ENCODING)) {
-            encoder = encodingContext.encoder(requestHeaders);
-            encoder.headers(response.headers());        // adds Content-Encoding
+        // validate and configure SSE headers on the response
+        ServerResponseHeaders headers = response.headers();
+        HttpMediaType ct = headers.contentType().orElse(null);
+        if (response.status().code() != Status.OK_200.code()
+                || ct != null && !CONTENT_TYPE_EVENT_STREAM.values().equals(ct.mediaType().text())) {
+            throw new IllegalStateException("ServerResponse instance cannot be used to create SseSink");
         }
+        if (ct == null) {
+            headers.add(CONTENT_TYPE_EVENT_STREAM);
+        }
+        headers.set(CACHE_NO_CACHE_ONLY);
 
-        // write status and headers before encoding stream
-        writeStatusAndHeaders(headersOutputStream);
-
-        // set the final output stream
-        this.outputStream = (encoder != null) ? encoder.apply(headersOutputStream) : headersOutputStream;
+        // use the response's output stream — it handles protocol-specific
+        // header framing (HTTP/1.1 text vs HTTP/2 HPACK HEADERS frames)
+        // and content encoding
+        this.outputStream = response.outputStream();
     }
 
     @Override
@@ -157,49 +149,6 @@ class DataWriterSseSink implements SseSink {
         try {
             outputStream.close();
             ctx.serverSocket().close();
-        } catch (IOException e) {
-            throw new UncheckedIOException(e);
-        }
-    }
-
-    void writeStatusAndHeaders(OutputStream headersOutputStream) {
-        try {
-            ServerResponseHeaders headers = response.headers();
-
-            // verify response has no status or content type
-            HttpMediaType ct = headers.contentType().orElse(null);
-            if (response.status().code() != Status.OK_200.code()
-                    || ct != null && !CONTENT_TYPE_EVENT_STREAM.values().equals(ct.mediaType().text())) {
-                throw new IllegalStateException("ServerResponse instance cannot be used to create SseSink");
-            }
-
-            // start writing status line
-            headersOutputStream.write(OK_200);
-
-            // serialize a date header if not included
-            if (!headers.contains(HeaderNames.DATE)) {
-                headersOutputStream.write(DATE);
-                byte[] dateBytes = DateTime.http1Bytes();
-                headersOutputStream.write(dateBytes);
-            }
-
-            // set up and write headers
-            if (ct == null) {
-                headers.add(CONTENT_TYPE_EVENT_STREAM);
-            }
-            headers.set(CACHE_NO_CACHE_ONLY);
-            BufferData buffer = BufferData.growing(512);
-            for (Header header : headers) {
-                header.writeHttp1Header(buffer);
-            }
-            headersOutputStream.write(buffer.readBytes());
-
-            // complete heading
-            headersOutputStream.write('\r');        // "\r\n" - empty line after headers
-            headersOutputStream.write('\n');
-
-            // write response heading to the output
-            headersOutputStream.flush();
         } catch (IOException e) {
             throw new UncheckedIOException(e);
         }


hrstoyanov:
@spericas i killed the previous PR #11309 and replaced it with this one (singled summary commit, checkstyle fixes.)

romain-grecourt:
Next time, just squash, amend the commit to update/remove the author and force push.

hrstoyanov:
@spericas let me know of any issues, and if none - when this can be shipped

spericas:
> @spericas let me know of any issues, and if none - when this can be shipped

Did you see my comment about using the HTTP/1.1 `DataWriterSseSink` in the closed issue? I don't think that works correctly for HTTP/2. Have you tried your implementation with an HTTP/2 client?

hrstoyanov:
Where did you make this comment?

I added some more code changes over the weekend (DataWriterSseSink) and I am running Helidon http2 SSE in my apps fine now.

hrstoyanov:
@spericas I see you made the comment in the old pr that I closed. Sorry.

Check the latest commit in this new pr - it adds more code changes (see DataWriterSseSink) .  I am running a web app using sse over http2 comfortably

spericas:
> @spericas I see you made the comment in the old pr that I closed. Sorry.
> 
> Check the latest commit in this new pr - it adds more code changes (see DataWriterSseSink) . I am running a web app using sse over http2 comfortably

I see it, but we had an issue in the past with transfer encoding and using the output stream instead of a writer. We need to verify this again. We recently added support for content encoding as well, but that should be easier using the output stream. I enabled the pipeline for now.

spericas:
There are still a couple of problems with this PR. The first relates to the closing of streams vs. connections (need to investigate this further), and the second to HTTP/2 testing and chunked encoding with HTTP/1.1. I'm working on some improvements. Will create a PR against this PR's source branch.

hrstoyanov:
Thanks @spericas - whatever works best, feel free to kill/replace this PR, **but remember - this PR addresses two separate SSE issues** - we want both fixed!

And anything I can help/test - let me know. I am desperate  at this point, SSE is critical for some UI frameworks (data-star.dev) and LLM streaming responses apis

spericas:
> Thanks @spericas - whatever works best, feel free to kill/replace this PR, **but remember - this PR addresses two separate SSE issues** - we want both fixed!
> 
> And anything I can help/test - let me know. I am desperate at this point, SSE is critical for some UI frameworks (data-star.dev) and LLM streaming responses apis

@hrstoyanov Yes, the `isSent` fix is correct and will be part of it. Will try to get this resolved quickly.

spericas:
@hrstoyanov The problem that I alluded earlier related to the use of output streams is described here https://github.com/helidon-io/helidon/issues/9164 It relates to buffering the existing logic for non-SSE case.

spericas:
@hrstoyanov It was easier and more expedient to create a new PR https://github.com/helidon-io/helidon/pull/11506 I think it addresses your needs but also prevents the introduction of a regression due to unwanted buffering when using output streams (which was the reason we had stopped using them before). There's additional testing as well in the new PR. Could you try it out in your use case?

hrstoyanov:
Thank you @spericas . Will test today or tomorrow!

hrstoyanov:
@spericas I will put your new PR through real-life tests this week, but an AI automated review  **says you PR is much better** and I agree completely! Adding it here for your review -  **the most valuable feedback is Section 8 which lists potential issues with you new PR:**

----
# Helidon SSE Fix — Comparative Review Report

**Date:** 2026-04-01  
**Author:** hrstoyanov  
**Subject:** Review of PR [#11506](https://github.com/helidon-io/helidon/pull/11506) by `spericas` vs our PR [#11325](https://github.com/helidon-io/helidon/pull/11325)

---

## 1. Background & Issues Addressed

| Issue | Title | Status |
|-------|-------|--------|
| [#11053](https://github.com/helidon-io/helidon/issues/11053) | Support SSE on HTTP/2 | Open — `Http2ServerResponse` does not implement `sink()`, so `res.sink(SseSink.TYPE)` throws `UnsupportedOperationException` over HTTP/2 |
| [#11298](https://github.com/helidon-io/helidon/issues/11298) | `Http1ServerResponse.sink()` does not mark `isSent`, causing false routing error | Open — routing layer throws "A route MUST call either send, reroute, or next" because `hasEntity()` returns `false` |
| [#9164](https://github.com/helidon-io/helidon/issues/9164) | SSE streams do not send status-line until first event is emitted | Closed (PR #9224 in 4.1.2) — original fix **bypassed** `OutputStream` entirely and wrote directly to the low-level `DataWriter` to avoid buffering/chunked-encoding problems |

### The Critical History (Issue #9164)

In Helidon 4.1.0, SSE used the normal `response.outputStream()`. This caused two problems:

1. **Buffering:** The `BlockingOutputStream` buffers the first chunk before sending headers. This meant status-line + headers were not sent until the first SSE event was emitted — violating client timeout expectations (e.g., browser 300s timeout).
2. **Chunked encoding:** Using `OutputStream` forced chunked `Transfer-Encoding`, which is redundant for SSE (the protocol's `\n\n` framing is self-delimiting).

**Fix in PR #9224 (Sep 2024):** `spericas` rewrote `DataWriterSseSink` to bypass `OutputStream` completely. It wrote HTTP/1.1 status line (`HTTP/1.1 200 OK\r\n`) and headers manually via `DataWriterOutputStream` wrapping `ctx.dataWriter()`. This eliminated buffering and chunked encoding but **hardcoded HTTP/1.1 text framing**, making it impossible to work with HTTP/2 (which uses HPACK binary header frames).

---

## 2. Our PR #11325 — What We Did

Our PR addressed both #11053 and #11298 with two key changes:

### 2.1 `Http2ServerResponse.sink()` Implementation
- Added the missing `sink(GenericType<X>)` method mirroring `Http1ServerResponse`
- Used `HelidonServiceLoader` to discover `SinkProvider` implementations
- Provided `SinkProviderContext` with `serverResponse()`, `serverRequest()`, `connectionContext()`, and `closeRunnable()`
- Set `isSent = true` immediately upon sink creation

### 2.2 `DataWriterSseSink` Refactoring (2nd commit)
- **Removed** direct `DataWriter` access and manual HTTP/1.1 header writing
- **Delegated** to `response.outputStream()` — which is protocol-aware (HTTP/1.1 uses text headers + chunked; HTTP/2 uses HPACK HEADERS frame + DATA frames)
- Removed `writeStatusAndHeaders()` method entirely
- Removed manual content-encoding setup (the OutputStream handles it)

### 2.3 `isSent` Fix
- Set `isSent = true` immediately in `sink()` for both HTTP/1.1 and HTTP/2

---

## 3. Spericas' Concerns

From the PR #11325 conversation:

> *"I see it, but we had an issue in the past with transfer encoding and using the output stream instead of a writer. We need to verify this again."*

> *"There are still a couple of problems with this PR. The first relates to the closing of streams vs. connections (need to investigate this further), and the second to HTTP/2 testing and chunked encoding with HTTP/1.1."*

> *"The problem that I alluded earlier related to the use of output streams is described here #9164. It relates to buffering the existing logic for non-SSE case."*

### Legitimate Concerns

1. **Buffering regression:** Our PR re-introduced `response.outputStream()` in `DataWriterSseSink`, which uses `BlockingOutputStream`. This stream **buffers the first write** — it stores the first chunk in `firstBuffer` and doesn't send headers until the second write triggers `sendHeadersAndPrepare()`. For SSE, this means if only one event is emitted before a long delay, headers might not reach the client immediately.

2. **Chunked encoding over HTTP/1.1:** Using `OutputStream` re-introduces chunked `Transfer-Encoding` for HTTP/1.1 SSE streams. While functionally correct, this was intentionally removed in PR #9224 because SSE doesn't need it and some clients may handle it poorly.

3. **Close semantics — stream vs. connection:** Our `close()` in `DataWriterSseSink` called `ctx.serverSocket().close()`, which **closes the entire TCP connection**, not just the HTTP/2 stream. On HTTP/2, multiple streams share one connection — closing the socket kills all concurrent streams. This was a genuine bug.

---

## 4. PR #11506 Analysis — What Spericas Changed

### 4.1 Files Modified (10 files)

| File | Change Summary |
|------|---------------|
| `SinkProviderContext.java` | Added `flushHeaders()` default method to the SPI |
| `Http1ServerResponse.java` | Added `flushHeaders()`, changed `closeRunnable` to `this::commit`, set `isSent = true` immediately |
| `Http2ServerResponse.java` | Full `sink()` implementation, `handleSinkData()`, `flushHeaders()`, `BlockingOutputStream.flushHeaders()` |
| `DataWriterSseSink.java` | Major refactor — uses `response.outputStream()` + `context.flushHeaders()` |
| `SseHttp2Test.java` | New HTTP/2 SSE test |
| `SseServerTest.java` | New test: `testSseHeadersAreSentBeforeFirstEvent` |
| `SseClientTest.java` | Re-enabled (removed `@Disabled`) |
| `SimpleSseClient.java` | Added chunked transfer-encoding support + `awaitHeaders()` |
| `SseBaseTest.java` | Added `sseDelayed()` test handler |
| `webserver/tests/http2/pom.xml` | Added `helidon-webserver-sse` dependency |

### 4.2 Key Design Decisions

#### A) `flushHeaders()` — The Buffering Solution

This is the **core innovation** over our PR. Instead of accepting the buffering delay or bypassing `OutputStream`, spericas adds a new `flushHeaders()` method at three levels:

1. **SPI:** `SinkProviderContext.flushHeaders()` — default no-op method
2. **Http1ServerResponse.BlockingOutputStream:**
   ```java
   void flushHeaders() {
       if (closed || !firstByte) return;        // already sent
       if (firstBuffer != null) {               // data buffered, trigger normal path
           write(BufferData.empty());
           return;
       }
       sendHeadersAndPrepare();                  // no data yet, send headers only
       firstByte = false;
   }
   ```
3. **Http2ServerResponse.BlockingOutputStream:** Same pattern using HTTP/2's `sendHeadersAndPrepare()` (HPACK frames)

**Usage in DataWriterSseSink constructor:**
```java
this.outputStream = response.outputStream();
context.flushHeaders();   // <-- immediately flush status+headers
```

This elegantly solves the buffering problem: the SSE sink uses the protocol-aware `OutputStream` (so it works on both HTTP/1.1 and HTTP/2), but **immediately forces headers out** before any events are written. Clients see `200 OK` + `text/event-stream` right away.

#### B) Close Semantics — Stream vs. Connection

Our PR's `close()`:
```java
public void close() {
    closeRunnable.run();
    outputStream.close();
    ctx.serverSocket().close();  // ← KILLS entire connection!
}
```

Spericas' PR #11506 `close()`:
```java
public void close() {
    outputStream.close();    // close OutputStream first (content encoders can finish trailers)
    closeRunnable.run();     // then commit response
}
```

Key improvements:
- **No `ctx.serverSocket().close()`** — on HTTP/2, this would destroy all multiplexed streams on the connection. The OutputStream close + commit is sufficient.
- **Reversed order:** `outputStream.close()` before `closeRunnable.run()`. This is important because content encoders (gzip, deflate) need to write trailer bytes when the stream closes. If we commit first, those trailers have nowhere to go.
- **`closeRunnable` now points to `this::commit`** instead of the old lambda that set `isSent` + called `afterSend()` + `request.reset()`. This is correct because `isSent` is already set in `sink()`, and `commit()` handles the proper end-of-stream signaling.

#### C) `isSent` Fix — Identical to Ours

Both PRs set `isSent = true` immediately after `p.create(...)` returns. ✅

#### D) `closeRunnable` Change in Http1ServerResponse

Our branch (and current upstream):
```java
public Runnable closeRunnable() {
    return () -> {
        Http1ServerResponse.this.isSent = true;   // redundant now
        afterSend();
        request.reset();
    };
}
```

Spericas' PR #11506:
```java
public Runnable closeRunnable() {
    return Http1ServerResponse.this::commit;
}
```

This is cleaner — `commit()` calls `outputStream.commit()` which handles:
- Flushing remaining data
- Sending end-of-stream markers (HTTP/2 empty DATA frame, HTTP/1.1 chunk terminator)
- Running the responseCloseRunnable (which calls `afterSend()`)

#### E) Chunked Encoding Support in Test Client

Since using `OutputStream` re-introduces chunked encoding on HTTP/1.1, spericas updated `SimpleSseClient` with a proper `ChunkedInputStream` decoder. This is pragmatic — chunked encoding is a natural consequence of streaming via OutputStream and is correct per HTTP/1.1 spec.

### 4.3 Test Coverage

| Test | What It Validates |
|------|------------------|
| `SseHttp2Test.testSseClose` | HTTP/2 SSE works; closing sink only closes the stream (not the connection); subsequent requests on the same connection succeed |
| `SseServerTest.testSseHeadersAreSentBeforeFirstEvent` | Headers arrive within 500ms even though first event is delayed by 1000ms — proves `flushHeaders()` works |
| `SseClientTest` (re-enabled) | Full SSE client test suite — was disabled, now works with chunked encoding support |

---

## 5. Comparative Assessment

| Aspect | Our PR #11325 | Spericas PR #11506 | Verdict |
|--------|--------------|-------------------|---------|
| HTTP/2 `sink()` implementation | ✅ Implemented | ✅ Implemented (identical approach) | Tie |
| `isSent` immediate flag | ✅ Fixed | ✅ Fixed (same approach) | Tie |
| Protocol-aware OutputStream | ✅ Uses `response.outputStream()` | ✅ Uses `response.outputStream()` | Tie |
| Eager header flushing | ❌ Not addressed — headers buffered until first write | ✅ `flushHeaders()` forces immediate headers | **#11506 wins** |
| Close semantics | ⚠️ `ctx.serverSocket().close()` kills entire connection | ✅ Only closes stream + commits response | **#11506 wins** |
| Close ordering | ⚠️ `closeRunnable` before `outputStream.close()` | ✅ `outputStream.close()` before `closeRunnable` (encoders can finish) | **#11506 wins** |
| Content encoding compat | ⚠️ Untested with content-encoding | ✅ Close-then-commit order allows encoder trailers | **#11506 wins** |
| HTTP/2 test | ❌ None | ✅ `SseHttp2Test` validates multiplexing survival | **#11506 wins** |
| Delayed header test | ❌ None | ✅ `testSseHeadersAreSentBeforeFirstEvent` | **#11506 wins** |
| Chunked encoding handling | ❌ Test client can't decode chunked | ✅ `ChunkedInputStream` in test client | **#11506 wins** |
| Re-enabled test suites | ❌ `SseClientTest` still `@Disabled` | ✅ Re-enabled | **#11506 wins** |

---

## 6. Assessment of Spericas' Buffering Concern (Issue #9164)

**Was the concern valid?** Yes, absolutely.

The original PR #9224 bypassed `OutputStream` specifically to avoid buffering delays. Our PR #11325 reintroduced `OutputStream` without solving the buffering problem. The `BlockingOutputStream` always holds back the first chunk:

```java
// BlockingOutputStream.write(BufferData)
if (firstByte && firstBuffer == null) {
    firstBuffer = buffer.copy();    // <-- buffers first write, does NOT send headers
    return;
}
```

This means for an SSE sink that calls `response.outputStream()` but doesn't write any data yet, the headers sit in the buffer indefinitely. Browsers with 5-minute timeouts would give up.

**Spericas' `flushHeaders()` solution** is the correct fix. It gives sink implementations a way to say "I need headers sent NOW, even though I haven't written data yet" — without bypassing the protocol-aware OutputStream. This preserves:
- HTTP/1.1: Proper text header framing + chunked encoding
- HTTP/2: HPACK HEADERS frame encoding
- Content encoding: Handled transparently by the OutputStream stack

---

## 7. Conclusion & Recommendation

**PR #11506 is a superior solution.** It addresses all the issues our PR #11325 addressed, plus:

1. Solves the header-buffering problem that was the root cause of #9164
2. Fixes the dangerous `serverSocket().close()` that would break HTTP/2 multiplexing
3. Fixes close ordering for content-encoding compatibility
4. Includes comprehensive test coverage for HTTP/2, delayed headers, and chunked encoding
5. Re-enables the previously disabled `SseClientTest` suite


---

## 8. Issues & Recommendations for PR #11506

Despite being the superior solution, PR #11506 has several issues that should be raised with `spericas` before merging.

### 8.1 🔴 HTTP/2 `closeRunnable` Does NOT Call `commit()` — Stream May Not Terminate

**The most critical issue.** There is an asymmetry between HTTP/1 and HTTP/2:

**Http1ServerResponse (PR #11506):**
```java
public Runnable closeRunnable() {
    return Http1ServerResponse.this::commit;   // ← calls commit()
}
```

**Http2ServerResponse (PR #11506):**
```java
public Runnable closeRunnable() {
    return () -> {
        Http2ServerResponse.this.isSent = true;   // ← redundant (already set at sink() line 129)
        request.reset();
        // ← NO commit() call!
        // ← NO afterSend() call!
    };
}
```

**Consequences:**

1. **End-of-stream frame may not be sent.** On HTTP/2, `commit()` calls `BlockingOutputStream.commit()` which sends the end-of-stream DATA frame (`stream.writeData(BufferData.empty(), true)`). Without it, the HTTP/2 stream hangs open indefinitely — the client never knows the SSE stream ended.

2. **`whenSent()` listeners never fire.** The `afterSend()` method (which runs `whenSent` callbacks) is only called inside `BlockingOutputStream`'s `responseCloseRunnable`. For HTTP/1, `commit()` triggers this chain. For HTTP/2, the lambda bypasses `commit()` entirely, so `afterSend()` is never called.

3. **`isSent = true` is redundant.** It's already set at line 129 in `sink()`, immediately after the sink is created. Setting it again in the closeRunnable is harmless but misleading.

4. **`request.reset()` is called in HTTP/2 but not in HTTP/1.** In HTTP/1, `commit()` → `BlockingOutputStream.commit()` → `responseCloseRunnable.run()` which calls `afterSend()` + `request.reset()`. But in HTTP/2, the closeRunnable directly calls `request.reset()` without going through the commit chain.

**Recommended fix:**
```java
// Http2ServerResponse — match Http1's approach
public Runnable closeRunnable() {
    return Http2ServerResponse.this::commit;
}
```

### 8.2 🔴 `DataWriterSseSink.close()` — `outputStream.close()` Is a No-Op Without Content Encoding

Both `Http1ServerResponse.BlockingOutputStream.close()` and `Http2ServerResponse.BlockingOutputStream.close()` are **explicit no-ops**:

```java
// Http1ServerResponse.BlockingOutputStream
@Override
public void close() {
    // no-op
}

// Http2ServerResponse.BlockingOutputStream
@Override
public void close() {
    // does nothing, we expect commit()
}
```

In the new `DataWriterSseSink.close()`:
```java
public void close() {
    outputStream.close();      // ← no-op when no content encoding!
    closeRunnable.run();
}
```

The PR's commit message says *"close the SSE output stream before committing so content encoders can finish writing trailers"*. This design only achieves its goal when content encoding is active — the encoder wraps the stream, and closing the encoder flushes final compressed bytes. Without encoding, `close()` does nothing.

This isn't strictly a bug (the `closeRunnable` then calls `commit()` which properly terminates the stream), but the design intent is fragile. For HTTP/1, the actual `OutputStream` returned by `response.outputStream()` is `ClosingBufferedOutputStream`, whose `close()` calls `closingDelegate.closing()` + `delegate.close()` — but `delegate.close()` on a `BufferedOutputStream` wrapping the no-op `BlockingOutputStream.close()` would flush the `BufferedOutputStream` buffer and then call the no-op close. So `ClosingBufferedOutputStream.close()` does flush, which is important.

**Recommendation:** Add a comment in `DataWriterSseSink.close()` explaining the design:
```java
public void close() {
    // Close the output stream first — when content encoding is active,
    // this flushes final encoder trailer bytes (e.g., gzip footer).
    // For Http1 ClosingBufferedOutputStream, this also flushes the write buffer.
    // For raw BlockingOutputStream (no encoding), this is a no-op.
    outputStream.close();
    closeRunnable.run();
}
```

### 8.3 🟡 Content-Encoded SSE: Close-Then-Commit Ordering Needs Test Coverage

When content encoding is active, the output stream chain looks like:

```
DataWriterSseSink
  → EncoderOutputStream (gzip/deflate)
    → ClosingBufferedOutputStream (Http1) or BlockingOutputStream (Http2)
```

The close ordering in `DataWriterSseSink.close()`:
1. `outputStream.close()` → closes the encoder → writes compression trailers (e.g., gzip CRC32 + ISIZE)
2. `closeRunnable.run()` → calls `commit()` → sends terminating chunk (Http1) or end-of-stream frame (Http2)

This ordering is correct in theory — the encoder must finish writing before the transport layer terminates. However:
- There is **no test** for SSE + content encoding + proper stream termination
- The gzip/deflate footer bytes need to pass through `BlockingOutputStream.write()` during the encoder's `close()` — after `flushHeaders()` has already been called, so `firstByte` is `false` and data writes should work. But if `commit()` was somehow called first (race condition or future refactoring), the `closed` flag would block writes.

**Recommendation:** Add an integration test:
```java
@Test
void testSseWithGzipEncoding() throws Exception {
    // SSE endpoint with Accept-Encoding: gzip
    // Verify events are properly decompressed
    // Verify stream terminates cleanly
}
```

### 8.4 🟡 `flushHeaders()` Silently No-Ops if `outputStream` Is Null

In both `Http1ServerResponse` and `Http2ServerResponse`:
```java
void flushHeaders() {
    if (outputStream != null) {
        outputStream.flushHeaders();
    }
    // ← silently does nothing if outputStream is null
}
```

This works for SSE because the sink constructor calls `response.outputStream()` before `context.flushHeaders()`. But a future `SinkProvider` implementor who calls `context.flushHeaders()` before obtaining the output stream would get silent failure — headers never flush, and there's no error or log message.

**Recommendation:** Either throw an `IllegalStateException` if `outputStream == null`, or at minimum log a warning:
```java
void flushHeaders() {
    if (outputStream != null) {
        outputStream.flushHeaders();
    } else {
        LOGGER.log(System.Logger.Level.WARNING,
                   "flushHeaders() called before outputStream was created; headers not flushed");
    }
}
```

### 8.5 🟢 `SINK_PROVIDERS` Static Field Placed Among Instance Fields

In `Http2ServerResponse`:
```java
private boolean isSent;
private boolean streamingEntity;
private long bytesWritten;
private static final List<SinkProvider> SINK_PROVIDERS = ...;  // ← static among instance fields
private BlockingOutputStream outputStream;
```

Java convention groups `static final` constants at the top of the class, near `LOGGER`. The field should be moved up next to:
```java
private static final System.Logger LOGGER = System.getLogger(Http2ServerResponse.class.getName());
private static final List<SinkProvider> SINK_PROVIDERS = ...;
```

### 8.6 🟢 `sink()` Method Is Copy-Pasted Between Http1 and Http2

The `sink()` method in `Http1ServerResponse` and `Http2ServerResponse` is now ~40 lines of nearly identical code. The only differences are:
- The `closeRunnable` implementation (which, per issue 8.1, should be the same anyway)
- The `flushHeaders()` delegation target

This could be extracted to a `protected` method in `ServerResponseBase<T>` parameterized by a few lambdas, eliminating the duplication. This would also prevent future drift between the two implementations.

### Summary Table

| # | Severity | Issue | Fix |
|---|----------|-------|-----|
| 8.1 | 🔴 High | HTTP/2 `closeRunnable` doesn't call `commit()` — end-of-stream not sent, `whenSent` listeners skipped | Change to `Http2ServerResponse.this::commit` |
| 8.2 | 🔴 High | `outputStream.close()` is no-op without encoding — design intent unclear | Add clarifying comment, verify with/without encoding |
| 8.3 | 🟡 Medium | No test for SSE + content encoding close ordering | Add gzip SSE integration test |
| 8.4 | 🟡 Medium | `flushHeaders()` silently no-ops if called before `outputStream()` | Throw or log warning |
| 8.5 | 🟢 Low | `SINK_PROVIDERS` static field misplaced among instance fields | Move to class top |
| 8.6 | 🟢 Low | `sink()` duplicated across Http1/Http2 responses | Extract to `ServerResponseBase` |

spericas:
@hrstoyanov Hmmm maybe you picked up an older version. I just addressed the duplication based on your analysis, but the important ones (8.1 and 8.2) were already there. Namely,

  1. Section 8.1 is stale and no longer applies. In the current branch, HTTP/2 closeRunnable() already returns
     this::commit, exactly as the review recommends, so end-of-stream and afterSend() go through the normal
     commit path. See Projects/helidon-4/webserver/http2/src/main/java/io/helidon/webserver/http2/
     Http2ServerResponse.java:122 and Projects/helidon-4/webserver/http2/src/main/java/io/helidon/webserver/
     http2/Http2ServerResponse.java:334.
  2. Section 8.3 is also already addressed by the current tests. We now have compression coverage in Projects/
     helidon-4/webserver/tests/sse/src/test/java/io/helidon/webserver/tests/sse/SseCompressionTest.java, and
     the full reactor SSE run passes. So “needs test coverage” was valid earlier, but not now.
  3. Section 8.2 is only a documentation concern, not a behavior bug. In Projects/helidon-4/webserver/sse/src/
     main/java/io/helidon/webserver/sse/DataWriterSseSink.java:139, outputStream.close() is intentionally
     there to finalize encoder state before commit(). For HTTP/1 the returned stream is the buffered wrapper,
     not the raw no-op BlockingOutputStream, so close() still flushes buffered bytes before commit. For HTTP/2
     without content encoding it is mostly a no-op, but closeRunnable.run() immediately commits the stream, so
     termination is still correct. A clarifying comment would help, but this is not a merge blocker.

Let me know about the use case testing when you have a chance.

hrstoyanov:
@spericas thank you for your effort!
Closing this PR now, I saw the other (new) one got merged. Will make a build and report back if I see any issues.

PR #11349, opened 2026-03-17
Support Optional declarative client 404 responses

### LINKED ISSUE #11302: Add support for Optional response types to typed rest client
WebServer supports sending of `Optional<?>` - if empty, a `404` is returned, otherwise the value is returned as entity.

We should support the same on declarative rest client

```
@Http.GET
@Http.Path("/customer/{id}")
Optional<Customer> findCustomer(@Http.PathParam("id") customerId);
```

This method will current fail to generate valid code.

Desired state:

1. invoke the request with the provided customer id (i.e. `http://localhost:8080/customer/41`)
2. request `Customer` type
3. check if response status is `404 Not-Found` - if so, return empty optional, otherwise return the type

oleksii-tumanov:
## Summary
- support Optional<T> declarative client methods
- return Optional.empty() for 404 responses (when the client return type is Optional)
- add tests

Fixes #11302

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/restclient/RestClientExtension.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/http/restclient/RestClientExtension.java
@@ -467,7 +467,7 @@ class RestClientExtension extends RestExtensionBase implements RegistryCodegenEx
         var returnType = method.returnType();
         boolean useGenericType;
         String genericTypeField;
-        if (hasResponse && !returnType.typeArguments().isEmpty() && !returnType.isOptional()) {
+        if (hasResponse && !returnType.typeArguments().isEmpty()) {
             // we have a return type, and it has type arguments
             useGenericType = true;
             TypeName genType = TypeName.builder(TypeNames.GENERIC_TYPE)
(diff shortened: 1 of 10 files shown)


tomas-langer:
- Blocking: ClientResponseTypedImpl.java:105 now returns Optional.empty() for
    every 404 before the entity is consumed, and declarative clients just return that cached
    value from /tmp/pr11349-RestClientExtension.java:539 without closing the typed response. Any
    404 that still carries a payload now drops that body and can leave the underlying response
    unread, which is a connection-reuse regression for optional typed calls.
  - Blocking: DefaultErrorHandling.java:53 skips the entire RestClient.ErrorHandler
    chain for Optional<?> 404 responses. That suppresses user-provided handlers as well as the
    built-in fallback, so existing custom 404 mapping logic no longer runs for optional
    declarative methods.

oleksii-tumanov:
Thank you @tomas-langer

I rebased on "main".

  Addressed both points:
  - **DefaultErrorHandling** no longer bypasses the RestClient.ErrorHandler chain (for Optional<?>  404), so custom handlers still run.
  - **ClientResponseTypedImpl** now returns Optional.empty() for optional with 404 responses. If 404 includes a payload, it consumes/closes the response so the body is not left unread.

  I also added regression coverage for:
  - 404 with entity in the webclient tests
  - custom optional 404 error handling

tomas-langer (dismissed):

tomas-langer:
I will do a few follow up changes.

tomas-langer (approved):

oleksii-tumanov:
thank you @tomas-langer 
Seems my account does not appear to have write/merge permission on the upstream repo. 
Can you please merge the PR?

tomas-langer:

PR #11351, opened 2026-03-17
4.x: Replace usage of io.helidon.build.common.test.utils.JUnitLauncher with EngineKit

romain-grecourt:
### Description

Decouple JUnit dependency from build-tools.

- Replace usage of `io.helidon.build.common.test.utils.JUnitLauncher` with `EngineKit`
- Remove parameter resolver, use class protection domain to resolve test-classes
- Remove dependency on `helidon-build-common-test-utils`

### Documentation

None.


tomas-langer:
As there is no associated issue, can you please create a PR against `main` once merged?

PR #11359, opened 2026-03-18
Use EngineKit instead of io.helidon.build.common.test.utils.JUnitLaun…

romain-grecourt:
### Description

Decouple JUnit dependency from build-tools.

- Replace usage of `io.helidon.build.common.test.utils.JUnitLauncher` with `EngineKit`
- Remove parameter resolver, use class protection domain to resolve test-classes
- Remove dependency on `helidon-build-common-test-utils`

### Documentation

None.

### DIFF
--- a/service/tests/maven-plugin/pom.xml
+++ b/service/tests/maven-plugin/pom.xml
@@ -45,8 +45,8 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-params</artifactId>
+            <groupId>io.helidon.common.testing</groupId>
+            <artifactId>helidon-common-testing-junit5</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


service/tests/maven-plugin/pom.xml
```diff
@@ -136,5 +140,28 @@
...
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-invoker-plugin</artifactId>
+                        <configuration>
+                            <environmentVariables>
+                                <!--suppress LongLine -->
+                                <MAVEN_DEBUG_OPTS>-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:8001</MAVEN_DEBUG_OPTS>
```
danielkec, line 159: The new `debug-invoker` profile activates when `env.MAVEN_DEBUG_OPTS` is present, but then it overwrites that variable with a hard-coded JDWP config on `*:8001`. That means any existing `MAVEN_DEBUG_OPTS` from the outer build gets hijacked instead of forwarded. Would it make sense to use a separate activation flag, or to pass through `${env.MAVEN_DEBUG_OPTS}` so callers can choose the debug settings?

romain-grecourt: Activating on `env.MAVEN_DEBUG_OPTS` enables using `mvnDebug` as-is, this is very practical.

The hard-coded port on `8001` is debatable, it could use a property with the default value, although I'm not sure we will ever need to customize that.

The same configuration has been used in build-tools for many years.

PR #11362, opened 2026-03-19
JSON-RPC reworked to use Helidon JSON (27.x)

### LINKED ISSUE #11356: Replace JSON-P with Helidon JSON in JSON-RPC

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11356

### Documentation

None

### DIFF
--- a/webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
+++ b/webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2025, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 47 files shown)


webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
```diff
@@ -69,11 +69,11 @@ default JsonRpcResponse rpcId(String rpcId) {
 
     /**
      * Set a result as an arbitrary object that can be mapped to JSON. This
```
tvallin, line 71: My understanding is that the class defining the object must be annotated with `@Json.Entity`, right ? Maybe we could update the javadoc to reflect that instead of the `arbitrary object`, or would it work even without the annotation ?

tvallin (approved):
Minor comment, otherwise looks good to me !

webserver/jsonrpc/src/main/java/io/helidon/webserver/jsonrpc/JsonRpcResponse.java
```diff
@@ -45,7 +45,7 @@ public interface JsonRpcResponse extends ServerResponse {
      * @return this response
      */
     default JsonRpcResponse rpcId(int rpcId) {
-        return rpcId(Json.createValue(rpcId));
+        return rpcId(JsonNumber.create(rpcId));
```
tomas-langer, line 48: There is no `int` override for `JsonNumber.create(double)`, so this will modify the expected value from `1` to `1.0`

tomas-langer (changes requested):

PR #11366, opened 2026-03-19
JSON-P removal in smaller modules

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11377
Fixes https://github.com/helidon-io/helidon/issues/11376
Fixes https://github.com/helidon-io/helidon/issues/11378
Fixes https://github.com/helidon-io/helidon/issues/11379
Fixes https://github.com/helidon-io/helidon/issues/11380
Fixes https://github.com/helidon-io/helidon/issues/11381
Fixes https://github.com/helidon-io/helidon/issues/11382

### Documentation

None

### DIFF
--- a/http/media/json/src/test/java/io/helidon/http/media/json/JsonMediaTest.java
+++ b/http/media/json/src/test/java/io/helidon/http/media/json/JsonMediaTest.java
@@ -77,7 +77,7 @@ class JsonMediaTest {
         res.supplier().get()
                 .write(JSON_OBJECT_TYPE, createObject("test-title"), os, headers);
 
-        assertThat(headers, HttpHeaderMatcher.hasHeader(HeaderNames.CONTENT_TYPE, HttpMediaTypes.JSON.text()));
+        assertThat(headers, HttpHeaderMatcher.hasHeader(HeaderNames.CONTENT_TYPE, HttpMediaTypes.JSON_UTF_8.text()));
         String result = os.toString(StandardCharsets.UTF_8);
         assertThat(result, containsString("\"title\""));
         assertThat(result, containsString("\"test-title\""));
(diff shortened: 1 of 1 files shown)


tomas-langer:
This is even more complicated for `JwtHeaders.headerJson()` and `headerClaim(String), and `headerClaims()`, as these return Jakarta `JsonObject`. We will need to discuss how to handle these cases when backporting!

tomas-langer:
I think this should be split into at least two PRs - one with changes that do not impact the public APIs (such as observability changes), and one or more for modules that do change public APIs

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcFeature.java
```diff
@@ -491,34 +487,44 @@ private String processJsonResponse(ServerRequest req,
...
                                        JsonObject stateCookie,
                                        Tenant tenant) {
-        String accessToken = json.getString("access_token");
-        String idToken = json.getString("id_token", null);
-        String refreshToken = json.getString("refresh_token", null);
-        JwtHeaders jwtHeaders = JwtHeaders.parseToken(idToken);
-        TenantConfig tenantConfig = tenant.tenantConfig();
+        String accessToken = json.stringValue("access_token")
+                .orElseThrow(() -> new IllegalStateException("Token endpoint response is missing access_token"));
+        Optional<String> idToken = json.stringValue("id_token");
```
danielkec, line 492: This makes `id_token` optional, but the auth request still always sends and stores a nonce. If the token endpoint returns only `access_token`, we now skip nonce validation entirely and still complete the login. That seems like a security behavior change unrelated to the JSON migration; can we keep requiring `id_token` here, or gate the relaxed behavior explicitly and add a callback test for the missing-`id_token` case?

Verdent:
This work will go through 4.x first, since we need to be sure to create a proper deprecations and dual JSON using where needed. It will also be split to a small PRs.

PR #11369, opened 2026-03-19
Add test to make sure validation fails when constraint is declared on private element

### LINKED ISSUE #11360: Generated validator type is invalid for constraints on private fields.
## Environment Details
* Helidon Version: 4.4.0
* Helidon SE
* JDK version: 25
* OS: Windows

----------

## Problem Description

Adding `@Validation.Validated` with field-level constraints (`@Validation.Integer.Min`, `@Validation.String.NotEmpty`) to a JPA `@Entity` class causes a compilation failure. The `ValidatedTypeGenerator` annotation processor produces a `*__Validated.java` file with two issues:

1. Unreachable statement (compile error)

The generated `check()` and `checkProperty()` methods contain a switch with only a default branch that throws. The trailing }; is unreachable:

```
  public void check(ValidationContext ctx, Pokemon instance, String propertyName) {
      switch (propertyName) {
          default -> throw new ValidationException("Invalid property name: " + propertyName);
          };  // unreachable statement
  }
```

2. No validation logic generated for field constraints

Despite `@Validation.Integer.Min(1)` on the id field and `@Validation.String.NotEmpty` on name, the generated switch bodies have no cases for these properties. The `check(ctx, instance)` method body is empty.

## Reproducer

[data.zip](https://github.com/user-attachments/files/26100008/data.zip)

 Based on `helidon-examples/examples/declarative/data`. Changes:
  - Added `helidon-validation` dependency to `pom.xml`
  - Added `@Validation.Validated` to `Pokemon` class
  - Added `@Validation.Integer.Min(1)` to id field
  - Added `@Validation.String.NotEmpty` to name field

tomas-langer:
Update implementation.

### Description
Resolves #11360 

Original code only gathered elements that were annotated and supported. New implementation first gathers all annotated element, and then fail if some of them are not supported.

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java
@@ -219,12 +219,24 @@ class ValidatedTypeGenerator {
                         return true;
                     }
                     var kind = it.kind();
-                    if (kind == ElementKind.FIELD
-                            || kind == ElementKind.METHOD
-                            || kind == ElementKind.CLASS
-                            || kind == ElementKind.CONSTRUCTOR) {
-                        // these are all supported
+                    if (kind == ElementKind.METHOD) {
+                        if (ElementInfoPredicates.isStatic(it)) {
+                            // static methods not supported
+                            return true;
+                        }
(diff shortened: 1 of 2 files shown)


tomas-langer:
Added https://github.com/helidon-io/helidon/issues/11372 as a follow up

PR #11400, opened 2026-03-23
Scope SE fault-tolerance handlers to FtService routes

edbratt:
## Summary
- remove route-wide SE fault-tolerance error handlers from generated `Main`
- keep `Main` limited to registering `FtService`
- handle fault-tolerance-specific exceptions inside generated `FtService`

## Problem
When the SE `custom` starter included fault tolerance together with unrelated routes, the generated `Main` installed global error handlers, including a catch-all `Throwable` handler. That changed exceptions from unrelated routes as well.

In the verified database case, `/db` `NotFoundException` responses that should remain `404` were converted to `500`, causing the generated project tests to fail.

## Fix
Scope the fault-tolerance error handling to `FtService` routes only:
- `TimeoutException` -> `408 timeout`
- `CircuitBreakerOpenException` -> `503 circuit breaker`
- retry exhaustion -> `500` with the generated runtime exception message

## Result
Generated SE starters keep the expected fault-tolerance behavior on FT routes without changing the error handling of unrelated routes such as `/db`.

### DIFF
--- a/archetypes/archetypes/src/main/archetype/se/custom/extra-output.xml
+++ b/archetypes/archetypes/src/main/archetype/se/custom/extra-output.xml
@@ -40,23 +40,8 @@
                     <value key="artifactId">helidon-fault-tolerance</value>
                 </map>
             </list>
-            <list key="Main-helidon-imports" if="${extra} contains 'fault-tolerance'">
-                <value>io.helidon.http.Status</value>
-                <value>io.helidon.faulttolerance.BulkheadException</value>
-                <value>io.helidon.faulttolerance.CircuitBreakerOpenException</value>
-                <value>io.helidon.faulttolerance.TimeoutException</value>
-            </list>
             <list key="Main-routing-builder" if="${extra} contains 'fault-tolerance'">
-                <value><![CDATA[.register("/", new FtService())
-                .error(BulkheadException.class,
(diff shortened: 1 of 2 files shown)


archetypes/archetypes/src/main/archetype/se/custom/extra-output.xml
```diff
@@ -40,23 +40,8 @@
                     <value key="artifactId">helidon-fault-tolerance</value>
                 </map>
             </list>
-            <list key="Main-helidon-imports" if="${extra} contains 'fault-tolerance'">
-                <value>io.helidon.http.Status</value>
-                <value>io.helidon.faulttolerance.BulkheadException</value>
```
romain-grecourt, line 45: BulkheadException is not handled anymore ?

archetypes/archetypes/src/main/archetype/se/custom/files/src/main/java/__pkg__/FtService.java.mustache
wadhwaroh-lang, line 115: Looks like timeout and circuit breaker got local catch blocks here, but the old Main also had a BulkheadException -> 503 "bulkhead" handler that got removed. Should we catch BulkheadException in bulkheadHandler too, so that behavior doesn’t disappear?

tomas-langer:
I am closing this PR as there is no issue for it, and it has been abandoned. If this seems like a fix that you need, kindly submit an issue, and we will pick it up

PR #11405, opened 2026-03-24
4.x: Apply DbMapper to MongoDB DML statements

danielkec:
### Description

Refs #10819

Apply custom `DbMapper` conversion to MongoDB DML statements so mapped values are expanded consistently across named, indexed, and collection-based parameter binding.

This change:
- normalizes mapped values recursively before MongoDB JSON substitution
- preserves structured map, list, and array values instead of stringifying them
- flattens mapped named parameters so sample-style `$id`/`$name` placeholders work when binding a mapped object under one key
- adds regression coverage for `namedParam`, `indexedParam`, `addParam`, `params(List<?>)`, `params(Object...)`, `params(Map<String, ?>)`, and the attached reproducer

Forward port tracked in #11407.

### Documentation

None


danielkec:
Forward port PR: #11407 (`main`).

Keep #10819 open until both #11405 and #11407 are merged.

PR #11409, opened 2026-03-24
Remove deprecated types from common mapper.

### LINKED ISSUE #11406: Remove deprecated types from common-mapper
Related to #9568

tomas-langer:
Add methods to Value and OptionalValue without a Mappers instance, to get Mappers only when needed.

Resolves #11406 

Also removes the need to provide `Mappers` when creating `Value` and `OptionalValue` , and it is requested only when actually needed.
This is to remove `Services.get()` on hot-path of applications that do not need mapping.

### DIFF
--- a/http/http/src/main/java/io/helidon/http/ContentDisposition.java
+++ b/http/http/src/main/java/io/helidon/http/ContentDisposition.java
@@ -31,9 +31,7 @@ import java.util.regex.Pattern;
 
 import io.helidon.common.GenericType;
 import io.helidon.common.mapper.MapperException;
-import io.helidon.common.mapper.Mappers;
 import io.helidon.common.mapper.Value;
-import io.helidon.service.registry.Services;
 
 /**
  * A generic representation of the {@code Content-Disposition} header.
@@ -187,7 +185,7 @@ public class ContentDisposition implements Header {
 
     @Override
     public Value<String> asString() {
-        return Value.create(Services.get(Mappers.class), name(), get(), GenericType.STRING, "http", "header");
+        return Value.create(name(), get(), GenericType.STRING, "http", "header");
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


danielkec:
One remaining JPMS/lazy-lookup issue: `http/http` still calls `Services.get(Mappers.class)` directly in `HeaderValueBase` and `ContentDisposition`. That makes mapper resolution eager again and forces `helidon-http` to require `service.registry` directly. The fix should be to use the new lazy `Value.create(name, ...)` overloads there too, then drop the direct `Services` usage/module dependency.

http/http/src/main/java/io/helidon/http/HeaderValueBase.java
```diff
@@ -64,12 +65,12 @@ public String getString() {
...
     @Override
     public <T> T get(Class<T> type) {
-        return MapperManager.global().map(get(), String.class, type, QUALIFIER);
+        return Services.get(Mappers.class).map(get(), String.class, type, QUALIFIER);
     }
 
     @Override
     public <N> Value<N> as(Function<? super String, ? extends N> mapper) {
-        return Value.create(MapperManager.global(), name(), mapper.apply(get()), QUALIFIER);
+        return Value.create(Services.get(Mappers.class), name(), mapper.apply(get()), QUALIFIER);
```
danielkec: See PR comment above: lazy `Value.create(...)` should be used here too. https://github.com/helidon-io/helidon/pull/11409#issuecomment-4125764167

http/http/src/main/java/io/helidon/http/ContentDisposition.java
```diff
@@ -186,7 +187,7 @@ public Value<Boolean> asBoolean() {
 
     @Override
     public Value<String> asString() {
-        return Value.create(MapperManager.global(), name(), get(), GenericType.STRING, "http", "header");
+        return Value.create(Services.get(Mappers.class), name(), get(), GenericType.STRING, "http", "header");
```
danielkec: Same here; please use the lazy `Value.create(...)` overload. https://github.com/helidon-io/helidon/pull/11409#issuecomment-4125764167

PR #11415, opened 2026-03-27
Fix HTTP/2 client ping handling

### LINKED ISSUE #11414: Http2ClientStream.sendPing() issues
## Environment Details
* Helidon Version: 4.4.0
* Helidon SE
* JDK version: 21
* OS: Mac/OS-agnostic

----------

## Problem Description
The `Http2ClientStream.sendPing()` has two issues:
1. It uses a `static final Http2FrameData HTTP2_PING` instance to hold the frame data, but it never calls `rewind()` on the data buffer with the ping payload. The first call to the `sendPing()` method will produce a healthy ping. The second call to the method will result in sending the ping header to the wire and will not "send" an empty payload (e.g. won't send a single byte of the payload to the wire) because the BufferData.available() that holds the ping payload would return 0. An HTTP2 server on the other side of the connection may then receive bytes from another frame that the client sends after the second ping and interpret them as the ping payload. This breaks the HTTP2 connection eventually because the server starts receiving garbage frames. See a simulator test below.
2. It bypasses the ping()/pong() protocol of the underlying `Http2ClientConnection`. Since the server will send PING ACK frames back to the pings initiated by the `Http2ClientStream.sendPing()`, the `Http2ClientConnection` will process them which will result in accumulating permits in the `Http2ClientConnection.pingPongSemaphore`. This will interfere with subsequent ping()/pong() calls in the connection class because they will report success due to the accumulated permits while the connection may already be broken very badly.

In our own GRPC client implementation on top of the Helidon HTTP2 Client, we had to work-around this issue by writing our own `sendPing()` like this:

https://github.com/hashgraph/pbj/pull/763

However, we're a bit concerned about our implementation because it bypasses the synchronization implemented in the `Http2ConnectionWriter` class around the Http2ClientConnection's DataWriter. We write to the DataWriter directly instead, which may result in thread races in rare circumstances. We have to do that because there's not an public API to access the `Http2ConnectionWriter` object.

So ideally, we'd like this issue to be fixed on the Helidon side. There may be various ways to do that. For example, you could make `Http2ClientStream.sendPing()` delegate to the `Http2ClientConnection.ping()` method. Alternatively, you could nuke `Http2ClientStream.sendPing()` and make the `Http2ClientConnection.ping()` public instead.

Also, it may be a good idea to make the `Http2ClientConnection.writer()` public to support more advanced client protocols and/or simplify workarounds that applications may need to implement, such as our PR linked above.

## Steps to reproduce
A simulator that simulates the `Http2ClientStream.sendPing()` behavior:
```java
package com.hedera.pbj.grpc.client.helidon;

import io.helidon.common.buffers.BufferData;
import io.helidon.common.buffers.DataWriter;
import io.helidon.common.socket.SocketContext;
import io.helidon.http.http2.Http2ConnectionWriter;
import io.helidon.http.http2.Http2FrameData;
import io.helidon.http.http2.Http2Ping;
import org.junit.jupiter.api.extension.ExtendWith;
import org.mockito.ArgumentCaptor;
import org.mockito.Mock;
import org.mockito.junit.jupiter.MockitoExtension;
import java.util.List;
import static org.mockito.Mockito.doNothing;

@ExtendWith(MockitoExtension.class)
public class PingTest {
    @Mock DataWriter dataWriter;
    @Mock SocketContext ctx;

    private static final Http2FrameData HTTP2_PING = Http2Ping.create().toFrameData();

    @Test
    public void testPing() throws Exception {
        final ArgumentCaptor<BufferData> bufferDataArgumentCaptor = ArgumentCaptor.forClass(BufferData.class);
        doNothing().when(dataWriter).writeNow(bufferDataArgumentCaptor.capture());

        final Http2ConnectionWriter http2ConnectionWriter = new Http2ConnectionWriter(ctx, dataWriter, List.of());

        // Simulate two sendPing() calls, just like the Http2ClientStream performs them effectively:
        http2ConnectionWriter.write(HTTP2_PING);
        http2ConnectionWriter.write(HTTP2_PING);

        bufferDataArgumentCaptor.getAllValues().forEach(bd -> System.err.println(bd.debugDataHex()));
    }
}

```

This test produces the following output:
```
+--------+-------------------------------------------------+----------------+
|   index|  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |            data|
+--------+-------------------------------------------------+----------------+
|00000000| 00 00 08 06 00 00 00 00 00                      |.........       |
+--------+-------------------------------------------------+----------------+
+--------+-------------------------------------------------+----------------+
|   index|  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |            data|
+--------+-------------------------------------------------+----------------+
|00000000| 00 00 00 00 00 00 00 00                         |........        |
+--------+-------------------------------------------------+----------------+

+--------+-------------------------------------------------+----------------+
|   index|  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |            data|
+--------+-------------------------------------------------+----------------+
|00000000| 00 00 08 06 00 00 00 00 00                      |.........       |
+--------+-------------------------------------------------+----------------+
+--------+-------------------------------------------------+----------------+
|   index|  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |            data|
+--------+-------------------------------------------------+----------------+
+--------+-------------------------------------------------+----------------+
```

As you can see, the payload of the second ping is empty. Only the header frame gets written.

danielkec:
### Description

Fixes #11414

`Http2ClientStream.sendPing()` reused drained frame data and bypassed the connection ping accounting. Route stream pings through the connection writer, validate health-check ACK payloads, and add regression coverage for repeated stream pings and unmatched ACKs.

This change:
- routes `Http2ClientStream.sendPing()` through `Http2ClientConnection.sendPing()` so each send writes a fresh PING frame
- tracks health-check ping ACKs by payload and documents the single in-flight health-check ping assumption
- adds `BufferData.writeInt64(long)` for the 8-byte ping payload encoding used by the HTTP/2 client
- adds Mockito-based regression coverage for repeated stream pings, matching ACKs, and timeout on unmatched ACKs

### Documentation

None

### DIFF
--- a/webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionPingTest.java
+++ b/webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionPingTest.java
@@ -22,7 +22,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 import io.helidon.common.buffers.BufferData;
 import io.helidon.common.buffers.DataReader;
@@ -42,6 +41,7 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 class Http2ClientConnectionPingTest {
+    private static final Duration TEST_WAIT_TIMEOUT = Duration.ofSeconds(10);
     private static final Http2StreamConfig STREAM_CONFIG = new Http2StreamConfig() {
         @Override
         public boolean priorKnowledge() {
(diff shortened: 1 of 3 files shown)


webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionPingTest.java
```diff
@@ -0,0 +1,213 @@
...
+        private Http2ClientConnection connection() {
+            return connection;
+        }
+
+        private List<BufferData> writes() {
+            return dataWriter.writes;
+        }
+
+        private BufferData awaitWrite() throws InterruptedException {
+            BufferData write = dataWriter.asyncWrites.poll(5, TimeUnit.SECONDS);
```
spericas: Could 5 secs not be enough on a slow pipeline run?

webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionPingTest.java
```diff
@@ -0,0 +1,213 @@
...
+    void pingWaitsForMatchingAck() throws Exception {
+        MockedConnectionTestContext test = new MockedConnectionTestContext(Duration.ofSeconds(1));
+        CompletableFuture<Long> payloadFuture = new CompletableFuture<>();
+
+        Thread.ofPlatform().start(() -> {
+            try {
+                long pingPayload = readPingPayloadId(test.awaitWrite());
+                payloadFuture.complete(pingPayload);
+                test.connection().pong(pingPayload + 1);
+                Thread.sleep(50);
```
spericas: Why is this sleep necessary here? And why 50 ms?

danielkec: Huh thx!

webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionPingTest.java
danielkec: There is no real socket latency on that path, mocked CaptureWriter writes imediatly

PR #11421, opened 2026-03-27
[4.x] Correctly handle default origin logic in Websockets

spericas:
### Description

Correctly handle default origin logic in Websockets

### Documentation

Updated as part fo PR

### DIFF
--- a/webserver/tests/websocket/src/test/java/io/helidon/webserver/tests/websocket/WebSocketOriginTest.java
+++ b/webserver/tests/websocket/src/test/java/io/helidon/webserver/tests/websocket/WebSocketOriginTest.java
@@ -82,6 +82,9 @@ class WebSocketOriginTest {
                                   .endpoint("/single", WebSocketOriginTest::single));
     }
 
+    /**
+     * Verify that an allowlisted origin is accepted.
+     */
     @Test
     void testSingle() throws ExecutionException, InterruptedException, TimeoutException {
         int port = webServer.port();
@@ -109,6 +112,9 @@ class WebSocketOriginTest {
         assertThat(received, hasItem("LOWER"));
     }
 
+    /**
+     * Verify that a non-allowlisted origin is rejected.
+     */
     @Test
     void testRejectedOrigin() {
         int port = webServer.port();
@@ -125,6 +131,9 @@ class WebSocketOriginTest {
         assertThat(handshakeException.getResponse().statusCode(), is(403));
     }
 
+    /**
+     * Verify that same-host origins are still rejected when an explicit allowlist is configured.
+     */
     @Test
     void testSameHostNotAllowlistedRejected() {
         int port = webServer.port();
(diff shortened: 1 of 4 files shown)


webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsUpgrader.java
```diff
@@ -205,6 +205,67 @@ protected Set<String> origins() {
...
+            originUri = new URI(origin);
+        } catch (URISyntaxException e) {
+            return false;
+        }
+
+        if (originUri.getScheme() == null || originUri.getHost() == null) {
+            return false;
+        }
+
+        UriInfo requestedUri = ctx.listenerContext()
```
danielkec: This makes the default origin check depend on requested-URI discovery. Behind a TLS-terminating proxy, `requestedUri` will usually resolve to `http://...` unless `Forwarded` / `X-Forwarded-*` discovery is configured, while the browser still sends `Origin: https://...`. That means a legitimate same-origin WebSocket handshake starts failing with 403 in a common deployment shape. Can we either avoid deriving this check from backend transport state, or document and test the proxy/TLS requirement explicitly?

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsUpgrader.java
```diff
@@ -205,6 +205,67 @@ protected Set<String> origins() {
...
+                         ctx.localPeer().address(),
+                         prologue.uriPath().rawPath(),
+                         ServerRequestHeaders.create(headers),
+                         prologue.query(),
+                         ctx.isSecure());
+
+        return normalizeOrigin(originUri).equals(normalizedRequestedUri(requestedUri));
+    }
+
+    private static String normalizedRequestedUri(UriInfo requestedUri) {
```
danielkec: We normalize the `Origin` side to lowercase, but `normalizedRequestedUri` compares `requestedUri.scheme()` and `requestedUri.host()` as-is. Those components are case-insensitive, so a valid same-origin request can be rejected if `Host` or trusted forwarded metadata uses different casing. Can we normalize both sides before comparing?

webserver/tests/websocket/src/test/java/io/helidon/webserver/tests/websocket/WebSocketOriginTest.java
```diff
@@ -102,6 +109,22 @@ public CompletionStage<?> onText(java.net.http.WebSocket webSocket,
         assertThat(received, hasItem("LOWER"));
     }
 
+    @Test
+    void testRejectedOrigin() {
```
danielkec, line 119: The new contract says that a non-empty `origins` set is an explicit allowlist, but this test only covers one allowlisted cross-origin and one blocked cross-origin. It would be worth adding an unlisted same-origin case (for example `http://localhost:<port>`) and a no-`Origin` case here; otherwise `(allowlisted || same-origin)` would still pass these tests while violating the documented behavior.

PR #11432, opened 2026-03-30
11430: Fix bulkhead waiting metric accounting

### LINKED ISSUE #11430: Bulkhead metrics race condition
Bulkhead metrics fails tests intermittently due to a race condition.

tomas-langer:
Resolves #11430

Fix bulkhead queue metric accounting so `ft.bulkhead.executionsWaiting` stays incremented while a queued call is actually blocked, and add regression coverage for the previous behavior.

### DIFF
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/BulkheadImpl.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/BulkheadImpl.java
@@ -17,18 +17,16 @@
 package io.helidon.faulttolerance;
 
 import java.lang.System.Logger.Level;
-import java.util.IdentityHashMap;
+import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Queue;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/BulkheadImpl.java
```diff
@@ -132,28 +132,36 @@ public <T> T invoke(Supplier<? extends T> supplier) {
...
-            } finally {
-                try {
-                    if (metricsEnabled) {
-                        waitingDurationMetric.record(System.nanoTime() - start, TimeUnit.NANOSECONDS);
-                        callsWaiting.decrementAndGet();
-                    }
-                } finally {
-                    inProgressLock.unlock(); // we have enqueued, now we can wait
+                if (waitingCounted) {
+                    waitingDurationMetric.record(System.nanoTime() - start, TimeUnit.NANOSECONDS);
```
danielkec: `waitingDuration` is still recorded before `barrier.waitOn()`, so it measures enqueue time, not the actual time spent waiting to enter the bulkhead.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/BulkheadImpl.java
```diff
@@ -131,29 +131,37 @@ public <T> T invoke(Supplier<? extends T> supplier) {
...
+                callsRejected.incrementAndGet();
                 throw new BulkheadException("Bulkhead queue \"" + name + "\" is full");
             }
-            barrier.waitOn();
+            long waitStartedAt = 0L;
+            if (waitingCounted) {
+                waitStartedAt = System.nanoTime();
+            }
+            try {
+                barrier.waitOn();
```
danielkec: Since `cancelSupplier()` now retracts the same barrier, waking up here no longer guarantees the caller actually acquired the permit. The code below still calls `dequeued()` before checking `cancelledSuppliers`, so cancelled queued calls report a dequeue event even though they never run. I think the fix is to move the listener notification after the cancellation check, for example:

```java
if (cancelledSuppliers.remove(supplier)) {
    return null;
}
listeners.forEach(l -> l.dequeued(supplier));
```

spericas: Indeed

PR #11443, opened 2026-03-30
2.x: 2.6.15 release prep

barchetta:
### Description

- Update CHANGELOG 
- Update owasp dependency check plugin

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,7 +119,7 @@
         <version.plugin.source>3.0.1</version.plugin.source>
         <version.plugin.spotbugs>4.4.2.2</version.plugin.spotbugs>
         <version.plugin.findsecbugs>1.11.0</version.plugin.findsecbugs>
-        <version.plugin.dependency-check>12.1.9</version.plugin.dependency-check>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -119,7 +119,7 @@
         <version.plugin.source>3.0.1</version.plugin.source>
         <version.plugin.spotbugs>4.4.2.2</version.plugin.spotbugs>
         <version.plugin.findsecbugs>1.11.0</version.plugin.findsecbugs>
-        <version.plugin.dependency-check>12.1.9</version.plugin.dependency-check>
+        <version.plugin.dependency-check>12.2.0</version.plugin.dependency-check>
```
danielkec, line 122: Does this still end up using whatever dependency-check version happens to be installed locally, rather than the `12.2.0` bumped here? `etc/scripts/owasp-dependency-check.sh` still calls `org.owasp:dependency-check-maven:aggregate` without an explicit version, so I’m wondering whether this change actually affects that script path.

barchetta: Not sure what you mean by "installed locally". The project pom manages the version of `dependency-check-maven` which is invoked by the script.

Issue #11450, opened 2026-03-31
Helidon JSON byte[] serialization methods


Verdent:
Our JSON support is missing serialization methods to the byte[]. It has only String/OutpuStream/Writer variations.

Issue #11449, opened 2026-03-31
JsonNumber is serialized as a String


Verdent:
Helidon JsonNumber is currently serialized as a String. It should be a String only in NaN and Infinite cases.

PR #11451, opened 2026-03-31
Json improvements and fixes

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11450
Fixes https://github.com/helidon-io/helidon/issues/11449

This PR includes several bug fixes and enhancements.

### Documentation

None

### DIFF
--- a/json/json/src/main/java/io/helidon/json/JsonParser.java
+++ b/json/json/src/main/java/io/helidon/json/JsonParser.java
@@ -445,6 +445,15 @@ public interface JsonParser {
     JsonException createException(String message);
 
     /**
+     * Create a JsonException with the given message and cause.
+     *
+     * @param message the exception message
+     * @param e the cause
+     * @return a JsonException
+     */
+    JsonException createException(String message, Exception e);
+
+    /**
      * Create a JsonException with the given message and found byte information.
      *
      * @param message the exception message
--- a/json/json/src/main/java/io/helidon/json/JsonParserStream.java
+++ b/json/json/src/main/java/io/helidon/json/JsonParserStream.java
@@ -283,16 +283,25 @@ final class JsonParserStream extends JsonParserBase {
 
     @Override
     public JsonException createException(String message) {
+        return new JsonException(exceptionMessage(message));
+    }
+
+    @Override
+    public JsonException createException(String message, Exception e) {
+        return new JsonException(exceptionMessage(message), e);
+    }
+
+    private String exceptionMessage(String message) {
         int start = Math.max(currentIndex - 10, 0);
         int length = Math.min(currentIndex + 10, bufferLength - start);
         int dataIndex = currentIndex - start;
         BufferData bufferData = BufferData.create(buffer, start, length);
 
-        return new JsonException(message + "\n"
-                                         + "Error at JSON index: " + currentIndex + "\n"
-                                         + "Data index: " + dataIndex + "\n"
-                                         + "Data: \n"
-                                         + bufferData.debugDataHex(false));
+        return message + "\n"
+                + "Error at JSON index: " + currentIndex + "\n"
+                + "Data index: " + dataIndex + "\n"
+                + "Data: \n"
+                + bufferData.debugDataHex(false);
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


json/json/src/main/java/io/helidon/json/JsonParser.java
```diff
@@ -84,7 +90,11 @@ static JsonParser create(byte[] json, int start, int length) {
             throw new JsonException("Invalid start/length: start="
                                             + start + ", length=" + length + ", array length=" + json.length);
         }
-        return new JsonParserArray(json, start, length);
+        JsonParserArray parser = new JsonParserArray(json, start, length);
```
tomas-langer, line 95: Regresses JsonParser.create(byte[], int, int) for zero-length slices.
 The new whitespace-skip fast path unconditionally reads json[start], so a call like JsonParser.create(new byte[] {'{'}, 1, 0) now throws ArrayIndexOutOfBoundsException instead of a parser-level JsonException.

Verdent: oh... thanks

json/json/src/main/java/io/helidon/json/JsonParserStream.java
```diff
@@ -391,84 +446,410 @@ char processEscapedSequence() {
     int decodeUtf8(int position, byte currentByte) {
         if ((currentByte & 0xE0) == 0xC0) {
             // 2-byte UTF-8 sequence: 110xxxxx 10yyyyyy -> U+0080 to U+07FF
-            int c2 = readNextByte() & 0x3F; // Second byte must be 10yyyyyy
+            ensure(1);
+            int c2 = requireUtf8Continuation(readNextByte()); // Second byte must be 10yyyyyy
```
danielkec: This adds UTF-8 validation for `readString()`, but `readChar()` still goes through `decodeUtf8ToChar()` below, which still accepts overlong or invalid UTF-8 on the stream path. A quoted `0xC0 0xAF` sequence still comes back as `'/'` here instead of failing.

json/json/src/main/java/io/helidon/json/Parsers.java
```diff
@@ -88,4 +89,149 @@ private static String hex(char c) {
...
+        for (int i = start; i < end; i++) {
+            byte b = buffer[i];
+            if (b == '\\') {
+                return decodeEscapedJsonString(buffer, start, end);
+            }
+            if (isControlCharacter(b)) {
+                throw new JsonException("Unescaped control character in JSON string");
+            }
+        }
+        return new String(buffer, start, length, StandardCharsets.UTF_8);
```
danielkec: This fallback still uses Java's replacement decoder for raw bytes, so malformed UTF-8 now slips through `readJsonString().value()` as U+FFFD instead of being rejected. That leaves the `JsonValue` path less strict than `readString()` and binding.

json/json/src/main/java/io/helidon/json/JsonParser.java
```diff
@@ -80,11 +86,17 @@ static JsonParser create(byte[] json) {
...
-        if (start < 0 || length < 0 || start > json.length || start + length > json.length) {
+        if (json.length == 0) {
+            throw new JsonException("Empty byte array provided");
+        } else if (start < 0 || length < 0 || start > json.length || start + length > json.length) {
             throw new JsonException("Invalid start/length: start="
                                             + start + ", length=" + length + ", array length=" + json.length);
         }
-        return new JsonParserArray(json, start, length);
+        JsonParserArray parser = new JsonParserArray(json, start, length);
+        if (WHITESPACE_CHARS[json[start] & 0xff]) {
```
danielkec, line 96: This dereferences `json[start]` even when `length == 0`. `JsonParser.create(new byte[]{1}, 1, 0)` now throws `ArrayIndexOutOfBoundsException` here instead of a `JsonException`, so the zero-length slice needs to be rejected before the whitespace skip.

json/json/src/main/java/io/helidon/json/JsonParserStream.java
Verdent: Fixed

json/json/src/main/java/io/helidon/json/Parsers.java
Verdent: Fixed

json/json/src/main/java/io/helidon/json/JsonParser.java
Verdent: Fixed

tomas-langer (approved):

PR #11509, opened 2026-04-01
Resolve deprecations in module `helidon-common-reactive` (27.x)

### LINKED ISSUE #11466: Resolve deprecations in module `helidon-common-reactive` (27.x)
- [ ] io.helidon.common.reactive.IoMulti#createOutputStream:60
- [ ] io.helidon.common.reactive.IoMulti#builderOutputStream:87
- [ ] io.helidon.common.reactive.IoMulti#createInputStream:116
- [ ] io.helidon.common.reactive.IoMulti#builderInputStream:145
- [ ] io.helidon.common.reactive.MultiFromOutputStream.MultiFromOutputStream:36
- [ ] io.helidon.common.reactive.SingleSubscriberHolder#<init>:44

tomas-langer:
Resolves #11466

Remove the deprecated `IoMulti` alias methods, delete `MultiFromOutputStream`, and keep `SingleSubscriberHolder` construction behind `create()` so `helidon-common-reactive` only exposes the supported reactive APIs.

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/SingleSubscriberHolder.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/SingleSubscriberHolder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/reactive/src/test/java/io/helidon/common/reactive/MultiFromInputStreamTest.java
+++ b/common/reactive/src/test/java/io/helidon/common/reactive/MultiFromInputStreamTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2020, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/common/reactive/src/test/java/io/helidon/common/reactive/MultiFromOutputStreamTest.java
+++ b/common/reactive/src/test/java/io/helidon/common/reactive/MultiFromOutputStreamTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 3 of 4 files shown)


common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamMulti.java
```diff
@@ -16,19 +16,172 @@
...
+    }
+
+    private void publishBufferedMaybe() throws IOException {
+        if (byteBuffer.position() > 0) {
+            publish();
+        }
+    }
+
+    private void publish(byte[] b, int off, int len) throws IOException {
+        ByteBuffer emitBuffer = ByteBuffer.allocate(len - off);
```
danielkec, line 132: `len` here is already the number of bytes to write, not the end index. When `off > 0`, allocating `len - off` makes the buffer too small and `emitBuffer.put(b, off, len)` overflows. I reproduced this with `write(new byte[]{10, 20, 30, 40}, 1, 2)`, which throws `BufferOverflowException`. This should allocate `len`, and it would be good to add a regression test for the non-zero-offset path.

tomas-langer: This is an exact copy of the code from `helidon-4.x` after removing the deprecated type. If this is real, please submit an issue against 4.x and `main`...

danielkec: Filed #11521 for `main` and #11522 for `4.x`. This bug predates #11509, so it should be tracked separately from this PR.

common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamMulti.java
```diff
@@ -16,19 +16,172 @@
...
+     * <li><b>param</b> {@code n} the requested count.</li>
+     * <li><b>param</b> {@code demand} the current total cumulative requested count,
+     * ranges between [0, {@link Long#MAX_VALUE}] where the max indicates that this
+     * publisher is unbounded.</li>
+     * </ul>
+     *
+     * @param requestCallback to be executed
+     * @return this output stream multi
+     */
+    public OutputStreamMulti onRequest(BiConsumer<Long, Long> requestCallback) {
```
danielkec, line 73: This changes the bytecode signature of the supported `OutputStreamMulti.onRequest(...)` API. On `main`, callers compiled against `OutputStreamMulti` link to the inherited method returning `MultiFromOutputStream`; after this change they look for that old descriptor at runtime and hit `NoSuchMethodError`. I reproduced it by compiling `IoMulti.outputStreamMulti().onRequest(...)` against current `main` and running it against this PR. If we need binary compatibility here, this PR has to preserve the old descriptor somehow rather than only redeclaring the method with return type `OutputStreamMulti`.

tomas-langer: Expected - the super class was removed as it was deprecated.

tomas-langer: Expected - the super class was removed as it was deprecated.

Issue #11521, opened 2026-04-02
OutputStreamMulti.write overflows for non-zero offsets


danielkec:
Discovered while reviewing #11509. This bug is not introduced by that PR; it is already present on the `main` branch.

`io.helidon.common.reactive.OutputStreamMulti#write(byte[], int, int)` delegates to a helper that currently allocates the buffer as `ByteBuffer.allocate(len - off)`. In the `OutputStream` contract, `len` is already the number of bytes to write, not an end index, so any call with `off > 0` can overflow before the item is emitted.

Actual behavior:
- `write(new byte[] {10, 20, 30, 40}, 1, 2)` throws `BufferOverflowException`

Expected behavior:
- the publisher emits the slice `{20, 30}`

## Reproducer
```java
OutputStreamMulti publisher = IoMulti.outputStreamMulti();
publisher.subscribe(new Flow.Subscriber<ByteBuffer>() {
    @Override
    public void onSubscribe(Flow.Subscription subscription) {
        subscription.request(Long.MAX_VALUE);
    }

    @Override
    public void onNext(ByteBuffer item) {
    }

    @Override
    public void onError(Throwable throwable) {
        throwable.printStackTrace();
    }

    @Override
    public void onComplete() {
    }
});

publisher.write(new byte[] {10, 20, 30, 40}, 1, 2);
```

This currently fails with a `java.nio.BufferOverflowException` from `OutputStreamMulti.publish(byte[], int, int)`.

## Suggested fix
Allocate `len`, not `len - off`, in `publish(byte[], int, int)`.

## Example Regression Test
```java
@Test
void testByteArrayWriteWithOffset() throws IOException {
    OutputStreamMulti publisher = IoMulti.outputStreamMulti();
    TestSubscriber<ByteBuffer> subscriber = new TestSubscriber<>();

    publisher.subscribe(subscriber);
    subscriber.requestMax();

    publisher.write(new byte[] {10, 20, 30, 40}, 1, 2);
    publisher.close();

    subscriber.assertItemCount(1).assertComplete();
    assertArrayEquals(new byte[] {20, 30}, subscriber.getItems().get(0).array());
}
```

This can be added alongside the existing tests in `MultiFromOutputStreamTest`.

wadhwaroh-lang:
@danielkec  I have created a fix for this in PR https://github.com/helidon-io/helidon/pull/11536
Can you please review

danielkec:
4.x backport is tracked in #11522 and proposed in #11583.

PR #11536, opened 2026-04-05
Fix OutputStreamMulti buffer size for write(byte[], off, len)

wadhwaroh-lang:
This change fixes a bug in io.helidon.common.reactive.OutputStreamMulti where write(byte[], int, int) delegates to publish(byte[], int, int), which incorrectly allocated the emit buffer as ByteBuffer.allocate(len - off).

Per the OutputStream contract, the third argument is the number of bytes to write, not an end index. Using len - off can allocate too few bytes when off > 0, causing a BufferOverflowException before the data is published (e.g. write(new byte[] {10, 20, 30, 40}, 1, 2) should emit {20, 30}).

Change: allocate ByteBuffer.allocate(len) and keep put(b, off, len) unchanged.

Regression test: testByteArrayWriteWithOffset in MultiFromOutputStreamTest, matching the scenario in the issue.

This bug was called out while reviewing #11509 but is not introduced by that PR; it is already present on main (see [#11521](https://github.com/helidon-io/helidon/issues/11521)).

### DIFF
--- a/common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamMulti.java
+++ b/common/reactive/src/main/java/io/helidon/common/reactive/OutputStreamMulti.java
@@ -129,7 +129,7 @@ public class OutputStreamMulti extends OutputStream implements Multi<ByteBuffer>
     }
 
     private void publish(byte[] b, int off, int len) throws IOException {
-        ByteBuffer emitBuffer = ByteBuffer.allocate(len - off);
+        ByteBuffer emitBuffer = ByteBuffer.allocate(len);
         emitBuffer.put(b, off, len);
         emitBuffer.flip();
         doPublish(emitBuffer);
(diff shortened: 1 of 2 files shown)


wadhwaroh-lang:
@danielkec @tomas-langer  Can you please review? I am unable to add labels for the same( don't have permissions). Thanks!

etc/scripts/build-examples.sh
```diff
@@ -48,6 +48,24 @@ version() {
 HELIDON_VERSION=$(version "${WS_DIR}/bom/pom.xml")
 readonly HELIDON_VERSION
 
+# Drop stale helidon-examples (e.g. from a restored CI workspace cache) when its
```
romain-grecourt: Why are you updating this script ?

wadhwaroh-lang: @romain-grecourt We hit a fork Actions failure: stale helidon-examples from cache (4.4.1) vs main repo 27.x, so build-examples.sh failed the version check. I added a guard to remove a mismatched checkout; that’s separate from #11521, so I removed it from this PR and can propose it separately if Helidon wants it upstream.

romain-grecourt: If you want this to be fixed, file a different issue / PR.

wadhwaroh-lang: Understood — I’ve removed the script change from this PR so #11536 is only #11521. I’ll track the fork CI / stale helidon-examples fix in a separate issue and PR if you want it upstream. @romain-grecourt

danielkec:
Thanks @wadhwaroh-lang for the fix, the regression test, and the open source contribution to Helidon.

PR #11555, opened 2026-04-07
27: Fixes type use annotations issue

### LINKED ISSUE #11532: APT/codegen drops top-level TYPE_USE annotations on primitive method return types
**Title**  
APT/codegen drops top-level `TYPE_USE` annotations on primitive method return types

**Summary**  
Helidon’s APT/codegen metadata pipeline appears to preserve top-level `TYPE_USE` annotations on **declared** method return types, but drops them on **primitive** method return types.

As a result, a method like:

```java
public @ExampleAnnotation long primitiveValue();
```

does not expose `@ExampleAnnotation` through the resulting `TypedElementInfo.typeName().annotations()` metadata, while an otherwise equivalent declared return type does.

This seems to be a Helidon codegen/APT issue, not a Java annotation-model issue.

**Expected behavior**  
For both of the following methods:

```java
public @ExampleAnnotation long primitiveValue();
public @ExampleAnnotation ExampleType declaredValue();
```

the top-level return-type-use annotation `@ExampleAnnotation` should be visible in the Helidon metadata model, specifically on the return `TypeName` exposed by `TypedElementInfo.typeName()`.

**Actual behavior**  
- `@ExampleAnnotation ExampleType declaredValue()` → annotation is visible on `TypedElementInfo.typeName().annotations()`
- `@ExampleAnnotation long primitiveValue()` → annotation is **not** visible on `TypedElementInfo.typeName().annotations()`

**Minimal example**

Example annotation:

```java
package example;

import java.lang.annotation.Retention;
import java.lang.annotation.Target;

import static java.lang.annotation.ElementType.TYPE_USE;
import static java.lang.annotation.RetentionPolicy.CLASS;

@Target(TYPE_USE)
@Retention(CLASS)
public @interface ExampleAnnotation {
}
```

Example source:

```java
package example;

record ExampleType(long value) {
}

public interface ExampleFrob {
    @ExampleAnnotation long primitiveValue();
    @ExampleAnnotation ExampleType declaredValue();
}
```

**Observed through Helidon metadata APIs**  
Using Helidon’s APT/codegen path to build `TypeInfo` / `TypedElementInfo` from `ExampleFrob`:

- `TypedElementInfo` for `declaredValue` has `typeName().annotations()` containing `example.ExampleAnnotation`
- `TypedElementInfo` for `primitiveValue` does **not** have `typeName().annotations()` containing `example.ExampleAnnotation`

**Likely cause**  
`AptTypeFactory.createTypeName(TypeMirror)` appears to return immediately for primitive `TypeKind`s via `TypeName.create(...)`, without preserving `TypeMirror.getAnnotationMirrors()`.

By contrast, the `DeclaredType` path explicitly copies annotation mirrors onto the resulting `TypeName`.

**Probable affected area**  
- `codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java`

ljnelson:
Resolves #11532

Preserve type-use annotations when converting compiler type mirrors into codegen metadata, including primitive returns, array dimensions and components, type variables, and wildcards. Keep generic variable names separate from their annotations and preserve recursive generic bounds.

Add focused APT regression coverage for annotation placement and generic type structure.

### DIFF
--- a/codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
+++ b/codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
@@ -150,12 +150,12 @@ final class AptTypeFactory {
                 default -> throw new IllegalStateException("Unknown primitive type: " + kind);
             };
 
-            return Optional.of(TypeName.create(type));
+            return Optional.of(withTypeUseAnnotations(TypeName.create(type), typeMirror, elements));
         }
 
         switch (kind) {
         case VOID -> {
-            return Optional.of(TypeName.create(void.class));
+            return Optional.of(withTypeUseAnnotations(TypeName.create(void.class), typeMirror, elements));
         }
         case TYPEVAR -> {
             if (!inProgress.add(typeMirror)) {
@@ -170,7 +170,7 @@ final class AptTypeFactory {
                 handleBounds(elements, inProgress, typeVar.getUpperBound(), builder::addUpperBound);
                 handleBounds(elements, inProgress, typeVar.getLowerBound(), builder::addLowerBound);
 
-                return Optional.of(builder.build());
+                return Optional.of(withTypeUseAnnotations(builder.build(), typeMirror, elements));
             } finally {
                 inProgress.remove(typeMirror);
             }
@@ -185,10 +185,10 @@ final class AptTypeFactory {
             handleBounds(elements, inProgress, vt.getExtendsBound(), builder::addUpperBound);
             handleBounds(elements, inProgress, vt.getSuperBound(), builder::addLowerBound);
 
-            return Optional.of(builder.build());
+            return Optional.of(withTypeUseAnnotations(builder.build(), typeMirror, elements));
         }
         case ERROR -> {
-            return Optional.of(TypeName.create(typeMirror.toString()));
+            return Optional.of(withTypeUseAnnotations(TypeName.create(typeMirror.toString()), typeMirror, elements));
         }
         // this is most likely a type that is code generated as part of this round, best effort
         case NONE -> {
@@ -201,10 +201,12 @@ final class AptTypeFactory {
 
         if (typeMirror instanceof ArrayType arrayType) {
             TypeName typeName = createTypeName(elements, inProgress, arrayType.getComponentType()).orElseThrow();
-            return Optional.of(TypeName.builder(typeName)
-                                       .componentType(typeName)
-                                       .array(true)
-                                       .build());
+            return Optional.of(withTypeUseAnnotations(TypeName.builder(typeName)
+                                                       .componentType(typeName)
+                                                       .array(true)
+                                                       .build(),
+                                           typeMirror,
+                                           elements));
         }
 
         if (typeMirror instanceof DeclaredType declaredType) {
@@ -220,15 +222,7 @@ final class AptTypeFactory {
                 return Optional.empty();
             }
 
-            var annotationMirrors = declaredType.getAnnotationMirrors();
-            if (!annotationMirrors.isEmpty() && elements != null) {
-                // we cannot do this if elements is null
-                var newResultBuilder = TypeName.builder(result);
-                for (AnnotationMirror annotationMirror : annotationMirrors) {
-                    newResultBuilder.addAnnotation(AptAnnotationFactory.createAnnotation(annotationMirror, elements));
-                }
-                result = newResultBuilder.build();
-            }
+            result = withTypeUseAnnotations(result, typeMirror, elements);
 
             if (typeParams.isEmpty()) {
                 return Optional.ofNullable(result);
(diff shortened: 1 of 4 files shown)


codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
```diff
@@ -243,6 +237,36 @@ private static Optional<TypeName> createTypeName(Elements elements, Set<TypeMirr
...
+     *
+     * <p>This method helps fix <a href="https://github.com/helidon-io/helidon/issues/11532">Github issue 11532</a>.</p>
+     *
+     * @param typeName the non-{@code null} typeName that may need augmenting
+     * @param typeMirror the non-{@code null} {@link TypeMirror} {@code typeName} represents
+     * @param elements an {@link Elements}; if {@code null} no action will be taken
+     * @return the possibly augmented {@link TypeName}, or the supplied {@link TypeName} if no augmentation was needed
+     * @throws NullPointerException if {@code typeName} or {@code typeMirror} is {@code null}
+     */
+    private static TypeName withTypeUseAnnotations(TypeName typeName, TypeMirror typeMirror, Elements elements) {
```
romain-grecourt: This method is named after `TYPE_USE` but there is nothing specific to `TYPE_USE` in its body.

It wraps every call to `TypeName.create` or `TypeName.builder().build()` ; so it should probably have a generic name.

Not sure about Javadoc on private methods, that's never rendered.

ljnelson: Would renaming this to `withTypeMirrorAnnotations` suffice?
Should I remove method documentation? I wasn't sure if people would understand what this method does.

ljnelson: As I continue to think about this: certainly in the Java AST any `AnnotationMirror` present on a `TypeMirror` is a type use annotation (declaration annotations are present on `Element`s). I think that's probably why the method was named this way. Do you still want the rename?

romain-grecourt: That method wraps all TypeName instances created in AptTypeFactory, the implementation is indirectly about TYPE_USE, but if we need to make more fixes in the future then the name will have to be changed.

ljnelson: If you have a suggestion I am ready to apply it. `maybeDecorateTypeName`? `maybeTransformTypeName`? `maybeAnnotateTypeName`? `augmentTypeNameIfNecessary`?

tomas-langer:
Can you re-check on current version of `helidon-4.x`? This is probably fixed as part of my fixes for declarative validation - issue https://github.com/helidon-io/helidon/issues/11868 (not yet forward ported to `main`)
PR for `main` is #11874

codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
```diff
@@ -201,10 +201,12 @@ private static Optional<TypeName> createTypeName(Elements elements, Set<TypeMirr
 
         if (typeMirror instanceof ArrayType arrayType) {
             TypeName typeName = createTypeName(elements, inProgress, arrayType.getComponentType()).orElseThrow();
-            return Optional.of(TypeName.builder(typeName)
-                                       .componentType(typeName)
-                                       .array(true)
-                                       .build());
+            return Optional.of(withTypeUseAnnotations(TypeName.builder(typeName)
```
tomas-langer, line 204: Building the array from `TypeName.builder(typeName)` copies the component type's annotations into the enclosing array. With the new primitive handling, `@TypeUseMarker int[]` is represented with the annotation on both the array and its `int` component, so it becomes indistinguishable from `int @TypeUseMarker []`; multidimensional arrays promote annotations further.

Build the array wrapper without copying component annotations, then add only the annotations from the current `ArrayType`. Add positional tests for both syntaxes, and preferably a multidimensional case, that assert `typeName().annotations()` separately from `componentType().annotations()`.

codegen/apt/src/main/java/io/helidon/codegen/apt/AptTypeFactory.java
```diff
@@ -243,6 +237,36 @@ private static Optional<TypeName> createTypeName(Elements elements, Set<TypeMirr
         throw new IllegalStateException("Unknown type mirror: " + typeMirror);
     }
 
+    /**
+     * Augments the supplied {@link TypeName}, if necessary, to carry any type use annotations present on the supplied
+     * {@link TypeMirror} from which it was originally built, by way of the [@link
```
tomas-langer, line 242: `[@link AptAnnotationFactory#createAnnotation(AnnotationMirror, Elements)}` is not a valid Javadoc inline tag and renders as malformed literal text. Use `{@link AptAnnotationFactory#createAnnotation(AnnotationMirror, Elements)}` or remove the private-method Javadoc if it is not intended to be rendered.

danielkec (approved):
Reviewed annotation placement and recursive generic bounds. Focused tests and generated-metadata roundtrips pass.

PR #11556, opened 2026-04-07
27: Adds blueprint support "inside" dbclient without changing existing public API

ljnelson:
### Description

Replaces several hand-built DbClient builder internals with blueprint-generated state while preserving existing public APIs and usage patterns. There is no immediate forcing function for this refactor but it might gradually bring DbClient more in line with existing Helidon conventions and open up the possibility for better integration down the line. First and foremost: this PR is intended to do no harm.

This change keeps the public builder surfaces intact, but moves builder storage/boilerplate behind generated internal state for:

- `DbClient.Builder`
- `DbClientBuilderBase`
- `JdbcClientBuilder`
- `MongoDbClientBuilder`
- `DbStatements.Builder`
- `DbMapperManager.Builder`

The result is intended to be backward-compatible: existing builder entry points, method names, fluent chaining, and behavior remain unchanged.

### Notes

- `DbClient.Builder` remains as-is
- JDBC and Mongo builder-specific state now follow the same pattern as the shared dbclient builder base.
- `DbStatements.Builder` and `DbMapperManager.Builder` were migrated using the same compatibility-first approach.

### Documentation

No impact.

### DIFF
--- a/dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateBlueprint.java
+++ b/dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateBlueprint.java
@@ -24,6 +24,8 @@ import io.helidon.builder.api.Prototype;
 @Prototype.CustomMethods(DbStatementsBuilderStateSupport.class)
 interface DbStatementsBuilderStateBlueprint {
 
-    @Option.Singular("configuredStatement")
-    Map<String, String> configuredStatements();
+    @Option.Singular(value = "addStatement", withPrefix = false)
+    @Option.Configured
+    Map<String, String> statements();
+
 }
--- a/dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateSupport.java
+++ b/dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateSupport.java
@@ -27,6 +27,6 @@ final class DbStatementsBuilderStateSupport {
     static void addStatement(DbStatementsBuilderState.BuilderBase<?, ?> builder, String name, String statement) {
         Objects.requireNonNull(name, "Statement name must be provided");
         Objects.requireNonNull(statement, "Statement body must be provided");
-        builder.putConfiguredStatement(name, statement);
+        builder.addStatement(name, statement);
     }
 }
--- a/dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatements.java
+++ b/dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatements.java
@@ -88,14 +88,14 @@ public interface DbStatements {
          */
         public Builder config(Config config) {
             config.detach().asMap()
-                    .ifPresent(this.delegate::addConfiguredStatements);
+                    .ifPresent(this.delegate::addStatements);
             return this;
         }
 
         @Override
         public DbStatements build() {
             return new DbStatements() {
-                private final Map<String, String> statements = new HashMap<>(delegate.configuredStatements());
+                private final Map<String, String> statements = new HashMap<>(delegate.statements());
 
                 @Override
                 public String statement(String name) {


dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateBlueprint.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Map;
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+@Prototype.Blueprint(isPublic = false, createEmptyPublic = false)
+@Prototype.CustomMethods(DbStatementsBuilderStateSupport.class)
+interface DbStatementsBuilderStateBlueprint {
+
+    @Option.Singular("configuredStatement")
```
tomas-langer: No need to define the `value` if it is just the name of the method "sans" the `s` at the end, you can simply use:
```
@Option.Singular
Map<String, String> configuredStatements();
```

This would create `putConfiguredStatement` method.

tomas-langer: To avoid the method created belo (addStatement), do this:

```java
@Option.Singular(value = "addStatement", withPrefix = false)
@Option.Configured
Map<String, String> statements();
```

tomas-langer: The `Configured` part is not needed, but it seems this should be supporting configuration, as the original builders support it

ljnelson: I defer completely to you on this. I'll change `configuredStatements()` to `statements()` such that it is, exactly:
```
@Option.Singular(value = "addStatement", withPrefix = false)
@Option.Configured
Map<String, String> statements();
```

ljnelson: Addressed in https://github.com/helidon-io/helidon/pull/11556/commits/6d8c0c243b132687201013590cd30e00fe6e2d65

ljnelson: TBH I'm not sure I interpreted your instructions correctly so feel free to correct if necessary

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbClientBuilderStateBlueprint.java
```diff
@@ -0,0 +1,51 @@
...
+    @Option.DefaultBoolean(false)
+    boolean missingMapParametersAsNull();
+
+    Optional<DbStatements> statements();
+
+    Optional<Mappers> mapperManager();
+
+    Optional<DbMapperManager> dbMapperManager();
+
+    @Option.Singular("mapperProvider")
```
tomas-langer, line 45: Do not specify singular names where they are inferred correctly from the option, this is redundant code. This can be just `@Option.Singular`

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbClientBuilderStateBlueprint.java
```diff
@@ -0,0 +1,51 @@
...
+    List<DbMapperProvider> mapperProviders();
+
+    @Option.Singular("service")
```
tomas-langer, line 48: Either call the method `services`, or use the default `clientService` singular. I do not see a reason why this should differ.

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateBlueprint.java
```diff
@@ -0,0 +1,31 @@
...
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+@Prototype.Blueprint(isPublic = false, createEmptyPublic = false)
+@Prototype.CustomMethods(DbStatementsBuilderStateSupport.class)
+interface DbStatementsBuilderStateBlueprint {
+
+    @Option.Singular(value = "addStatement", withPrefix = false)
+    @Option.Configured
```
tomas-langer, line 28: The configured only makes sense if the whole blueprint is configured. I am not sure how you read configuration with this new approach. If you are  not using the `config` method of this blueprint, then teh `@Configured` does not make sense.

dbclient/dbclient/src/main/java/io/helidon/dbclient/DbStatementsBuilderStateSupport.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Objects;
+
+import io.helidon.builder.api.Prototype;
+
+final class DbStatementsBuilderStateSupport {
+    private DbStatementsBuilderStateSupport() {
+    }
+
+    @Prototype.BuilderMethod
+    static void addStatement(DbStatementsBuilderState.BuilderBase<?, ?> builder, String name, String statement) {
```
tomas-langer, line 27: This should conflict with the `addStatement` method automatically generated by the builder. This whole class should be removed.

tomas-langer:
I am not sure I see the benefit of this work. It adds a few internal-only prototypes, so the number of classes has increased, and nothing is really removed.

ljnelson:
PR withdrawn by reuqest

PR #11642, opened 2026-04-13
4.x: tree-ify config reference documentation

romain-grecourt:
### Description

Treeify config metadata docs around resolver nodes.

- Inline dotted prefixes and dotted option keys into the resolved config tree.
- Proof canonical YAML generation with a unit test
- Page models instead of adapted maps
- Synthetic names based on tree paths
- Usages tracked only from resolved tree nodes.
- Inline the MP config sections to make the config reference SE only

### Documentation

This PR updates the config reference documentation generator and thus impacts the documentation.


tomas-langer (approved):
Validated the docs build locally and compared the generated config-doc output against helidon-4.x. No blocking issues found.

PR #11658, opened 2026-04-15
Re-encoding already encoded content

### LINKED ISSUE #11648: Re-encoding already encoded content
Reproducer:

Requests comes with `Accept-Encoding: br;q=1, gzip;q=0.8`
We have `helidon-http-encoding-gzip` on the classpath
The route returns `Content-Encoding: br`

Expected behavior:
The client receives a response with `Content-Encoding: br`

Actual behavior:
The client receives a resposne with `Content-Encoding: br` and `Content-Encoding: gzip` - i.e. dual encoding

tomas-langer:
Resolves #11648

Avoid gzip re-encoding when `Content-Encoding` is already set, let HTTP/1.1 and HTTP/2 webclients pass through unsupported response encodings, and add buffered plus streamed regression coverage for pre-encoded `br` responses.


spericas (approved):
LGTM

PR #11660, opened 2026-04-15
Graceful handling of exceptions when writing to response

### LINKED ISSUE #11653: Graceful handling of exceptions when writing to response
Helidon 4.4.1

When writing to a peer that is gone, we get an `UncheckedIOException` with something similar to `Broken pipe`, `Connection reset by peer`, `Engine is closed`. These all mean the peer is gone, and the connection is dead.

Our writers shoud throw a "known" exception (such as `ConnectionCloseException` that we throw when we cannot read), so we do not need to add custom error handling by the user to all the places this can happen.

tomas-langer:
Resolves #11653

Wrap dead-peer write failures across the remaining affected server protocols so they surface as graceful connection-close exceptions instead of raw transport or cancellation stack traces. This covers HTTP/1 response writes, HTTP/2 connection ping acknowledgments, WebSocket frame writes, and gRPC header/data/trailer writes plus peer-cancelled callback paths.

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
@@ -133,7 +133,7 @@ public class Http2Connection implements ServerConnection, InterruptableTask<Void
         this.connectionWriter = new Http2ConnectionWriter(ctx,
                                                           ctx.dataWriter(),
                                                           List.of(new Http2LoggingFrameListener("send")));
-        this.connectionChecks = new Http2ConnectionChecks(http2Config, connectionWriter, this);
+        this.connectionChecks = new Http2ConnectionChecks(http2Config, this);
         this.subProviders = subProviders;
         this.requestDynamicTable = Http2Headers.DynamicTable.create(
                 serverSettings.value(Http2Setting.HEADER_TABLE_SIZE));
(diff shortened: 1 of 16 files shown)


danielkec (commented):
Inline review comments.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
```diff
@@ -748,17 +749,29 @@ private void doPriority() {
...
                                                           Http2FrameTypes.PING,
                                                           Http2Flag.PingFlags.create(Http2Flag.ACK),
                                                           0);
         ping = null;
-        connectionWriter.write(new Http2FrameData(header, frame));
+        writeConnectionFrame(new Http2FrameData(header, frame));
         state = State.READ_FRAME;
     }
 
+    private void writeConnectionFrame(Http2FrameData frame) {
```
danielkec: This helper is not used by `Http2ConnectionChecks.closeConnection(...)` or `Http2Upgrader.upgrade(...)`, so HTTP/2 still has direct dead-peer write paths. Can we route those writes through this wrapper too?

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
```diff
@@ -336,10 +340,18 @@ private WsSession send(ServerWsFrame frame) {
             }
         }
         sendBuffer.write(frame.payloadData());
-        ctx.dataWriter().writeNow(sendBuffer);
+        writeFrame(sendBuffer);
         return this;
     }
 
+    private void writeFrame(BufferData frameData) {
```
danielkec, line 347: This covers frame writes after upgrade, but `WsUpgrader.upgrade(...)` still writes the `101 Switching Protocols` response directly. A dead peer during handshake will still bypass the graceful close path.

danielkec (approved):

Issue #11672, opened 2026-04-16
4.x - Replace JSON-P with Helidon JSON in helidon-integrations-oci


Verdent:
Relates to https://github.com/helidon-io/helidon/issues/11379

PR #11675, opened 2026-04-16
integrations-oci JSON-P deprecation

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11672

### Documentation

None

### DIFF
--- a/integrations/oci/oci/src/main/java/io/helidon/integrations/oci/ImdsInstanceInfoBlueprint.java
+++ b/integrations/oci/oci/src/main/java/io/helidon/integrations/oci/ImdsInstanceInfoBlueprint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2024, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


integrations/oci/oci/src/main/java/io/helidon/integrations/oci/ImdsInstanceInfoBlueprint.java
```diff
@@ -94,8 +94,20 @@ interface ImdsInstanceInfoBlueprint {
...
-     * @return Full information about the Instance as a {@link jakarta.json.JsonObject}
+     * @deprecated for removal, use {@link #helidonJsonObject()}.
+     * @return Full information about the instance as a {@link jakarta.json.JsonObject}
      */
+    @Deprecated(forRemoval = true, since = "4.5.0")
     @Option.Configured
-    JsonObject jsonObject();
+    jakarta.json.JsonObject jsonObject();
+
+    /**
```
tomas-langer, line 104: The blueprint should have a method
```java
io.helidon.json.JsonObject json();
```

And then a decorator that ensures alignment (i.e. the user should be free to only configure Helidon JSON, as otherwise they cannot migrate away from the current method).
The decorator should check that either of the objects is available and update the other one. In case that both are filled, use the Helidon JSON.

integrations/oci/oci/src/main/java/io/helidon/integrations/oci/ImdsInstanceInfoBlueprint.java
```diff
@@ -94,8 +94,20 @@ interface ImdsInstanceInfoBlueprint {
...
+     * Instance Data.
+     *
+     * @return Full information about the instance as a {@link io.helidon.json.JsonObject}
+     */
+    default JsonObject helidonJsonObject() {
```
tomas-langer, line 109: Remove this method, it will be covered by the one I mentioned above.

tomas-langer, line 109: Also - never use the product name in method name - just use `json` instead, and if `json` is used, use `jsonObject`.

tomas-langer (changes requested):

tomas-langer (approved):

Issue #11565, opened 2026-04-08
API Stability annotations on Helidon APIs (4.x)


tomas-langer:
Helidon Version: 4.x

Follow up for #11564

## Task Description

- [ ] Change declarative and JSON modules to be preview instead of incubating (we are not going to remove them)
- [ ] Make sure all APIs that are currently marked as `@Deprecated` instead of "preview", "incubating", or "internal" are correctly annotated, and the deprecated annotation is removed
- [ ] Create a follow up issue in examples to correctly suppress warnings for such APIs wherever used
- [ ] Validate that examples do not print any compiler warnings

tomas-langer:
The main steps have been done. There may be minor changes added to further 4.x releases as needed

PR #11688, opened 2026-04-17
Mark ServiceLoader constructors internal (4.x)

tomas-langer:
Related to #11565

Marks ServiceLoader-only public constructors as @Api.Internal and moves the generated binding suppression to configure(...), where the generated constructor references are emitted.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConnectionProvider.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConnectionProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ServiceLoader;
 
+import io.helidon.common.Api;
 import io.helidon.common.HelidonServiceLoader;
 import io.helidon.webserver.ProtocolConfigs;
 import io.helidon.webserver.http1.spi.Http1UpgradeProvider;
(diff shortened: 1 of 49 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConnectionProvider.java
```diff
@@ -39,13 +40,9 @@ public class Http1ConnectionProvider implements ServerConnectionSelectorProvider
...
     /**
-     * Create a new instance with default configuration.
-     * To customize instance programmatically, use {@link Http1ConnectionSelector}
-     * instead.
-     *
-     * @deprecated to be used solely by {@link java.util.ServiceLoader}
+     * Required public constructor for {@link java.util.ServiceLoader}.
      */
-    @Deprecated
+    @Api.Internal
```
danielkec, line 45: `webserver/testing/junit5/junit5/src/main/java/io/helidon/webserver/testing/junit5/DirectClientConnection.java` still does `new Http1ConnectionProvider()` under `@SuppressWarnings("deprecation")`; with this ctor now internal, that call/suppression should be updated too.

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsUpgradeProvider.java
```diff
@@ -30,12 +31,9 @@ public class WsUpgradeProvider implements Http1UpgradeProvider<WsConfig> {
...
 
     /**
-     * Create a new instance with default configuration.
-     *
-     * @deprecated This constructor is only to be used by {@link java.util.ServiceLoader},
-     *      use {@link WsUpgrader#create(WsConfig)} for manual setup
+     * Required public constructor for {@link java.util.ServiceLoader}.
      */
-    @Deprecated()
+    @Api.Internal
```
danielkec, line 36: `webserver/websocket/src/test/java/io/helidon/webserver/websocket/WsUpgradeProviderConfigTest.java` still does `new WsUpgradeProvider()`, so this change turns that test into an internal-API use as well.

tomas-langer: There is no need to suppress tests, we do not enforce any restrictions on this in Helidon itself.

danielkec (approved):

PR #11699, opened 2026-04-20
Implement gRPC per-call deadline support (fixes #11698)

### LINKED ISSUE #11698: gRPC grpc-timeout header not sent or parsed; CallOptions.getDeadline() not propagated to server
## Problem

Helidon's gRPC implementation does not propagate per-call deadlines across the wire:

- The client never writes the `grpc-timeout` HTTP/2 header, even when `CallOptions.getDeadline()` is set via `stub.withDeadlineAfter(...)`
- The server never parses the `grpc-timeout` header, so `Context.current().getDeadline()` is always null in server handlers
- Deadline propagation across service-to-service calls does not work — a server handler making an outbound gRPC call does not inherit the incoming deadline

## Expected Behavior

- `stub.withDeadlineAfter(n, unit)` causes the `grpc-timeout` header to be written on the outbound HTTP/2 HEADERS frame
- The server parses `grpc-timeout` and makes the remaining time available via `Context.current().getDeadline()`
- A server handler making an outbound gRPC call automatically inherits the remaining context deadline
- `DEADLINE_EXCEEDED` (code 4) is returned when a deadline expires, not `CANCELLED` (code 1)

## Affected Modules

- `grpc/core` — `grpc-timeout` header encoding/decoding
- `webclient/grpc` — write header and enforce deadline client-side
- `webserver/grpc` — parse header and propagate into `Context`

MariusVolkhart:
Helidon's gRPC client stored CallOptions but never called getDeadline(). The grpc-timeout HTTP/2 header was never written on outbound calls and never parsed on inbound ones, so stub.withDeadlineAfter() silently did nothing and
Context.current().getDeadline() was always null in server handlers.

## Header encoding/decoding (grpc/core)

The gRPC over HTTP/2 spec defines a compact timeout format: a positive integer up to 8 ASCII digits followed by a unit suffix (H/M/S/m/u/n). GrpcHeadersUtil.encodeTimeout() picks the largest unit where the value fits in 8 digits, keeping wire overhead minimal. decodeTimeout() is the inverse. Both are used by client and server; unit tests cover all six units, the 8-digit boundary, round-trips, and malformed-input rejection.

## Client-side enforcement (webclient/grpc)

GrpcBaseClientCall now computes an effective deadline as min(CallOptions.getDeadline(), Context.current().getDeadline()). Taking the minimum of both sources gives automatic deadline propagation for free: a server handler that makes an outbound gRPC call inherits the remaining context deadline without any manual wiring.

The fail-fast check in start() runs before clientConnection() so an already-expired deadline returns DEADLINE_EXCEEDED without opening a TCP connection. When a live deadline is set, grpc-timeout is written into the outbound HEADERS frame after metadata conversion to prevent metadata from accidentally overwriting it.

A virtual thread (Thread.sleep) enforces the deadline at runtime. Virtual threads park without holding a carrier thread, which fits Helidon 4.x's threading model and avoids a ScheduledExecutorService.

Both GrpcClientCall (streaming) and GrpcUnaryClientCall (unary) lacked a race-safe close guard: the streaming call had none at all, and the unary call used volatile boolean, which has a TOCTOU window when the deadline thread and the read thread both see false before either writes true. Both now use AtomicBoolean.compareAndSet so exactly one thread wins and calls responseListener.onClose().

## Server-side enforcement (webserver/grpc)

ContextSettingServerInterceptor is already the assembly point for the gRPC Context before Contexts.interceptCall(). Adding deadline parsing here means every handler type — GrpcService, BindableService, and interceptor-wrapped handlers — gets deadline enforcement automatically.

Context.withDeadline() returns a CancellableContext that schedules a ScheduledFuture. If the call completes before the deadline the timer must be explicitly canceled; otherwise the ScheduledFuture holds a live reference until the original deadline fires. A ForwardingServerCallListener wraps onComplete and onCancel to call CancellableContext.cancel() in a finally block.

GrpcProtocolHandler previously returned CANCELLED (code 1) whenever the call was canceled by the client. A deadline expiry is semantically DEADLINE_EXCEEDED (code 4). The requestDeadline field is set at init() time from the grpc-timeout header; at close() time, if callCancelled is true and the deadline has expired, DEADLINE_EXCEEDED is returned instead.

### Documentation

 I don't think there's anything to document here since this isn't an API change.

### DIFF
--- a/grpc/core/pom.xml
+++ b/grpc/core/pom.xml
@@ -86,5 +86,10 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.helidon.common.testing</groupId>
+            <artifactId>helidon-common-testing-junit5</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/grpc/core/src/main/java/io/helidon/grpc/core/GrpcHeadersUtil.java
+++ b/grpc/core/src/main/java/io/helidon/grpc/core/GrpcHeadersUtil.java
@@ -17,6 +17,7 @@
 package io.helidon.grpc.core;
 
 import java.util.Base64;
+import java.util.Optional;
 
 import io.helidon.http.HeaderName;
 import io.helidon.http.HeaderNames;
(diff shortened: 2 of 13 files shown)


MariusVolkhart:
Based on feedback on another PR, will review this more carefully before re-submitting.

Issue #11700, opened 2026-04-21
webclient/grpc: Missing `te: trailers` header in outbound gRPC requests


MariusVolkhart:
### Description

The Helidon gRPC webclient does not include the `te: trailers` header in outbound request HEADERS frames.

The [gRPC-over-HTTP/2 protocol spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) lists `TE` in the `Call-Definition` as a non-optional field:

```
Call-Definition → Method Scheme Path [Authority] TE [Timeout] Content-Type ...
TE → "te" "trailers"  ; Used to detect incompatible proxies
```

HTTP/2-aware intermediaries use this header to decide whether to forward HTTP/2 trailers end-to-end (RFC 7540 §8.1.2.2). Since gRPC status codes and trailing metadata arrive exclusively in trailers, intermediaries that strip them cause every RPC to fail silently with no `grpc-status`.

For reference, grpc-java includes this header unconditionally in every client request — it is hardcoded into [`GrpcHttp2OutboundHeaders.clientRequestHeaders()`](https://github.com/grpc/grpc-java/blob/61d5492/netty/src/main/java/io/grpc/netty/GrpcHttp2OutboundHeaders.java#L43) alongside `content-type` and `user-agent`.

### Impact

No impact for direct client-to-server connections (development/testing). In production deployments with infrastructure in the path — Envoy-based service meshes (Istio), AWS ALB, GCP HTTP(S) LB, nginx as an HTTP/2 proxy — trailers may be silently dropped, making every gRPC call fail or appear to hang with no error.

PR #11701, opened 2026-04-21
Send te: trailers on every gRPC client request (#11700)

MariusVolkhart:
The gRPC-over-HTTP/2 spec lists TE as a non-optional field in Call-Definition. Without it, HTTP/2-aware intermediaries (Envoy, AWS ALB, GCP LB, nginx) may silently strip trailers, discarding grpc-status and trailing metadata from every RPC. Direct connections are unaffected, but any production deployment with a proxy in the path is broken in a way that is hard to diagnose: calls appear to hang or fail with no status code.

grpc-java hardcodes te: trailers alongside content-type in every outbound HEADERS frame (GrpcHttp2OutboundHeaders.clientRequestHeaders).

Adds GRPC_TE_TRAILERS constant to GrpcBaseClientCall and sets it in setupHeaders(). Verified by a server-side ServerInterceptor in an integration test that captures the raw Metadata received over the wire, proving the header survives the full HTTP/2 encoding pipeline and not just that it was added to a WritableHeaders object in-process.

### Documentation

No API changes or behavior changes needed from consumers. Closes an implementation gap relative to reference grpc-java.


spericas (approved):
LGTM

Issue #11720, opened 2026-04-21
gRPC server sends CANCELLED/UNKNOWN instead of RESOURCE_EXHAUSTED for oversized inbound messages


MariusVolkhart:
## Summary

When an inbound message exceeds `GrpcConfig.maxReadBufferSize()`, Helidon currently returns `CANCELLED` or `UNKNOWN` to the client. Other gRPC implementations return `RESOURCE_EXHAUSTED` with a descriptive message:

- grpc-java: [`MessageDeframer.processHeader()`](https://github.com/grpc/grpc-java/blob/v1.73.0/core/src/main/java/io/grpc/internal/MessageDeframer.java#L388-L393) throws `Status.RESOURCE_EXHAUSTED.withDescription("gRPC message exceeds maximum size ...").asRuntimeException()`

There is no specification mandate for `RESOURCE_EXHAUSTED` in this case, but it is the convention established by grpc-java and aligns with what clients are likely to handle.

## Impact

- Clients that check for `RESOURCE_EXHAUSTED` to implement retry-or-reduce-payload logic will not trigger correctly
- The error message received by the client does not indicate why the call failed or what the size limit is
- The server logs an ERROR for a condition that is part of normal operation (client sent an oversized message)

## Proposal

When message size exceeds the configured limit, close the call with `Status.RESOURCE_EXHAUSTED` and a message indicating the limit and actual size, matching grpc-java's behavior.

Issue #11719, opened 2026-04-21
gRPC server default maxReadBufferSize rejects messages that other implementations accept


MariusVolkhart:
## Summary

`GrpcConfig.maxReadBufferSize()` defaults to 2 MB. Because gRPC uses length-prefixed framing — each message declares its full byte size upfront and must be buffered entirely before deserialization — this value also serves as the maximum inbound message size.

Other widely-used gRPC implementations default to 4 MB:

- grpc-java: [`GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE = 4 * 1024 * 1024`](https://github.com/grpc/grpc-java/blob/v1.73.0/core/src/main/java/io/grpc/internal/GrpcUtil.java#L212)
- grpc-go: `defaultMaxRecvMsgSize = 4 * 1024 * 1024` in `rpc_util.go`

There is no specification requirement for this value, but 4 MB is the de facto ecosystem standard.

## Impact

A message between 2 MB and 4 MB succeeds against a grpc-java or grpc-go server with default configuration, but is rejected by a Helidon server. This affects common payloads like image data in proto fields, large list responses, and batch operations. APIs ported from grpc-java or grpc-go will fail without configuration changes.

## Proposal

Change the `maxReadBufferSize()` default from 2 MB to 4 MB to match the ecosystem standard.

MariusVolkhart:
This issue fails to adequately explain the landscape. I'll reopen with a better explanation.

spericas:
Fixed in 27

PR #11723, opened 2026-04-21
Align gRPC server inbound message size limit with ecosystem defaults and improve oversized message error handling

MariusVolkhart:
This PR addresses two related issues with gRPC inbound message size handling.

**#11719 — Default max message size (2 MB → 4 MB)**

`GrpcConfig.maxReadBufferSize()` previously defaulted to 2 MB. Other gRPC implementations default to 4 MB:
- grpc-java: [`GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE`](https://github.com/grpc/grpc-java/blob/v1.73.0/core/src/main/java/io/grpc/internal/GrpcUtil.java#L212)
- grpc-go: `defaultMaxRecvMsgSize` in `rpc_util.go`

Messages between 2 MB and 4 MB would succeed against any of those servers but be rejected by Helidon with default configuration. The default is now 4 MB to match. The limit remains configurable via `GrpcConfig.Builder.maxReadBufferSize()`.

**#11720 — RESOURCE_EXHAUSTED status for oversized messages**

When a message exceeded the size limit, the server threw a plain exception, which was caught and converted to `listener.onCancel()` and an ERROR log. The client received `CANCELLED` or `UNKNOWN` with no indication of the cause.

`allocateReadBuffer()` now throws `StatusRuntimeException(RESOURCE_EXHAUSTED)` with a description of the limit and actual size, following the convention established by grpc-java's `MessageDeframer.processHeader()`. A dedicated `catch (StatusRuntimeException)` block in `data()` closes the call with the correct status trailers and logs at DEBUG rather than ERROR.

**Testing**

- `GrpcProtocolHandlerTest.defaultMaxReadBufferSizeMatchesGrpcEcosystemStandard` — unit test asserting the 4 MB default
- `GrpcProtocolHandlerTest.oversizedMessageThrowsResourceExhausted` — unit test asserting `allocateReadBuffer()` throws `RESOURCE_EXHAUSTED`
- `GrpcMaxMessageSizeErrorTest.oversizedMessageReturnsResourceExhausted` — integration test verifying the client receives `Status.RESOURCE_EXHAUSTED`

**Documentation**

The maximum inbound message size the server tolerates by default increased from 2MB to 4MB. This is not a new feature or an API change, but an implementation detail change that could be of interest to some users.

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
@@ -80,11 +80,13 @@ interface GrpcConfigBlueprint extends ProtocolConfig {
     }
 
     /**
-     * Max size of gRPC reading buffer. If receiving an entity larger than this,
-     * processing will be aborted. This can help prevent DoS attacks. Default
-     * set to 2 MB.
+     * Max inbound message size. If receiving a message larger than this,
+     * processing will be aborted with a {@code RESOURCE_EXHAUSTED} status.
+     * This can help prevent DoS attacks. Matches the gRPC ecosystem standard of
+     * 4 MB ({@code GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE} in grpc-java,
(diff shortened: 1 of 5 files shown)


MariusVolkhart:
This PR lacks context. Will reopen with additional changes that explain the situation better.

PR #11725, opened 2026-04-21
Fix gRPC server inbound message size default and error status

### LINKED ISSUE #11719: gRPC server default maxReadBufferSize rejects messages that other implementations accept

### LINKED ISSUE #11720: gRPC server sends CANCELLED/UNKNOWN instead of RESOURCE_EXHAUSTED for oversized inbound messages

MariusVolkhart:
## Summary

- Align `maxReadBufferSize()` default with the 4 MB ecosystem standard (grpc-java, grpc-go). The previous 2 MB default caused Helidon to reject messages that succeed against every other standard gRPC server. Fixes #11719
- Return `RESOURCE_EXHAUSTED` status when an inbound message exceeds the limit. Previously, the server threw a plain exception that was caught as a cancellation — the client received `CANCELLED` or `UNKNOWN` with no indication of the cause. Fixes #11720

## Documentation

- The default `maxReadBufferSize` changed from 2 MB to 4 MB. Users who rely on the 2 MB limit to constrain inbound message size need to set it explicitly after upgrading.
- Oversized inbound messages now return gRPC status `RESOURCE_EXHAUSTED` with a descriptive message instead of `CANCELLED`/`UNKNOWN`. Client code that handles size violations by checking for `CANCELLED` should be updated to check for `RESOURCE_EXHAUSTED`.

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
@@ -89,7 +89,8 @@ interface GrpcConfigBlueprint extends ProtocolConfig {
      * and must be buffered entirely before deserialization — this effectively
      * limits the maximum inbound message size.
      *
-     * <p>Messages larger than this are rejected. Defaults to 4 MB, matching
+     * <p>Messages larger than this are rejected with a
+     * {@code RESOURCE_EXHAUSTED} status. Defaults to 4 MB, matching
      * the gRPC ecosystem standard ({@code GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE}
      * in grpc-java,
      * <a href="https://github.com/grpc/grpc-java/blob/v1.73.0/core/src/main/java/io/grpc/internal/GrpcUtil.java#L212">source</a>).
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcConfigBlueprint.java
@@ -83,13 +83,20 @@ interface GrpcConfigBlueprint extends ProtocolConfig {
     }
 
     /**
-     * Max size of gRPC reading buffer. If receiving an entity larger than this,
-     * processing will be aborted. This can help prevent DoS attacks. Default
-     * set to 2 MB.
+     * Maximum read buffer size for inbound gRPC messages.
+     * <p>
+     * Because gRPC uses length-prefixed framing — each message declares its full size upfront
+     * and must be buffered entirely before deserialization — this effectively
+     * limits the maximum inbound message size.
+     *
+     * <p>Messages larger than this are rejected. Defaults to 4 MB, matching
+     * the gRPC ecosystem standard ({@code GrpcUtil.DEFAULT_MAX_MESSAGE_SIZE}
+     * in grpc-java,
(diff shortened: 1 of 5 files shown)


spericas:
Some likely problems were found with this PR:

- P1 Early close stops draining inbound DATA and can stall the HTTP/2 connection
This new StatusRuntimeException path closes the call from GrpcProtocolHandler.data(), which moves the sub-protocol state to HALF_CLOSED_LOCAL. Http2ServerStream stops its read loop as soon as the handler reaches that state, but the connection thread still accepts later DATA frames and keeps enqueueing them into the per-stream ArrayBlockingQueue of size 32. If a client keeps sending after the oversized first message, that queue will fill and inboundData.put(...) will block the connection thread, creating a connection-level DoS/regression that did not exist before because the old path kept draining frames.

- P2 Status trailers are dropped for all StatusRuntimeExceptions
The new catch block closes the call with new Metadata() instead of propagating the trailers carried by the StatusRuntimeException. Because this handler now catches every StatusRuntimeException, not just the oversize-buffer one, any exception that includes response metadata (for example rich error details or custom debugging trailers) will lose that metadata on the wire. The existing gRPC helper code explicitly preserves trailers when translating status exceptions, so this looks like an unintended regression.

spericas:
@MariusVolkhart Not sure if you reviewed the P1 and P2 comments. Also, there are some pipeline failures. Let me know if you want me to take over it or if you plan to update the PR. Thanks!

MariusVolkhart:
I will update, apologies for the slow reply.

PR #11737, opened 2026-04-21
BufferDataInputStream must return -1 at EOF, not 0 or throw

### LINKED ISSUE #11735: BufferDataInputStream.read() returns 0 instead of -1 at EOF, hanging gRPC calls with read-until-EOF marshallers
`GrpcProtocolHandler.BufferDataInputStream` does not conform to the `InputStream` contract. Its `read(byte[], int, int)` overload returns `0` when the buffer is exhausted, and its no-argument `read()` throws `ArrayIndexOutOfBoundsException`. The `InputStream` specification requires all overloads to return `-1` at end of stream.

**Impact**

Any gRPC service whose `MethodDescriptor.Marshaller.parse()` implementation reads until EOF — for example by calling `stream.readAllBytes()` — will hang indefinitely. The server thread spins inside `readAllBytes()`'s internal loop because the loop condition `n >= 0` stays true as long as `read()` keeps returning `0`. `listener.onHalfClose()` is never called and no response is sent.

Marshallers generated from protobuf definitions are not affected because they read a known, fixed number of bytes and never request more.

**Root cause**

`InputStream.readAllBytes()` delegates to `readNBytes(Integer.MAX_VALUE)`, which loops while the return value of `read(byte[], int, int)` is `>= 0`. `FixedBufferData.read(byte[], int, int)` returns `0` when no bytes remain (it computes `toRead = Math.min(len, available)`, which is `0` when `available == 0`). Without an `available()` guard in `BufferDataInputStream`, this `0` propagates directly to the caller.

The same pattern was handled correctly in `BufferData.toInputStream()`, which has existed since the class was written. `BufferDataInputStream` was introduced later as a separate class to implement the `KnownLength` marker interface and did not replicate the guard.

**Affected versions**

4.x (confirmed on 4.4.1)

**Fix**

Add `available()` guards to all three `read` overrides in `BufferDataInputStream` before delegating to `BufferData`:

```java
@Override
public int read() {
  return bufferData.available() > 0 ? bufferData.read() : -1;
}

@Override
public int read(byte[] b) {
  if (bufferData.available() == 0) return -1;
  return bufferData.read(b);
}

@Override
public int read(byte[] b, int off, int len) {
  if (bufferData.available() == 0) return -1;
  return bufferData.read(b, off, len);
}
```

MariusVolkhart:
BufferDataInputStream delegated directly to BufferData without translating to InputStream semantics. BufferData.read() throws ArrayIndexOutOfBoundsException when exhausted; the bulk read overloads return 0. The InputStream contract requires -1 at EOF. Without the guard, any marshaller that reads until EOF (e.g. readAllBytes()) loops or throws instead of completing, and the gRPC call hangs indefinitely.

Add available() checks before each delegation. The same pattern already existed in BufferData.toInputStream() but was missed when BufferDataInputStream was written as a separate class to implement the KnownLength marker interface.

Fixes #11735

### Documentation

None

This is the same as #11736, but for `main` whereas the other is for `helidon-4.x`

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -600,16 +600,21 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
 
         @Override
         public int read() {
-            return bufferData.read();
+            // BufferData.read() throws when exhausted; InputStream requires -1 at EOF
+            return bufferData.available() > 0 ? bufferData.read() : -1;
         }
 
         @Override
         public int read(byte[] b) {
+            // BufferData.read(byte[]) returns 0 when exhausted; InputStream requires -1 at EOF
+            if (bufferData.available() == 0) return -1;
             return bufferData.read(b);
         }
 
         @Override
         public int read(byte[] b, int off, int len) {
+            // BufferData.read(byte[],int,int) returns 0 when exhausted; InputStream requires -1 at EOF
+            if (bufferData.available() == 0) return -1;
             return bufferData.read(b, off, len);
         }
 
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -607,15 +607,13 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
         @Override
         public int read(byte[] b) {
             // BufferData.read(byte[]) returns 0 when exhausted; InputStream requires -1 at EOF
-            if (bufferData.available() == 0) return -1;
-            return bufferData.read(b);
+            return bufferData.available() == 0 ? -1 : bufferData.read(b);
         }
 
         @Override
         public int read(byte[] b, int off, int len) {
             // BufferData.read(byte[],int,int) returns 0 when exhausted; InputStream requires -1 at EOF
-            if (bufferData.available() == 0) return -1;
-            return bufferData.read(b, off, len);
+            return bufferData.available() == 0 ? -1 : bufferData.read(b, off, len);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -600,16 +600,21 @@ static class BufferDataInputStream extends InputStream implements KnownLength {
...
         public int read() {
-            return bufferData.read();
+            // BufferData.read() throws when exhausted; InputStream requires -1 at EOF
+            return bufferData.available() > 0 ? bufferData.read() : -1;
         }
 
         @Override
         public int read(byte[] b) {
+            // BufferData.read(byte[]) returns 0 when exhausted; InputStream requires -1 at EOF
+            if (bufferData.available() == 0) return -1;
```
tomas-langer: either use ternary expression (if simple to read), or use a full if with `{`, i.e. this should be:


```
if (bufferData.available() == 0) {
    return -1;
}
return bufferData.read(b);
```

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -600,16 +600,21 @@ static class BufferDataInputStream extends InputStream implements KnownLength {
...
             return bufferData.read(b);
         }
 
         @Override
         public int read(byte[] b, int off, int len) {
+            // BufferData.read(byte[],int,int) returns 0 when exhausted; InputStream requires -1 at EOF
+            if (bufferData.available() == 0) return -1;
```
tomas-langer: Same as above

tomas-langer (changes requested):

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
MariusVolkhart: Done, and I applied the same change to https://github.com/helidon-io/helidon/pull/11736

I'll check my other PRs for similar patterns. Thanks for the review

tomas-langer (approved):

PR #11740, opened 2026-04-22
Fix writeTo(OutputStream) writing excess bytes after partial read

### LINKED ISSUE #11738: writeTo(OutputStream) writes excess bytes after partial read in FixedBufferData and ReadOnlyArrayData
## Summary

Two `BufferData` implementations write more bytes than intended from
`writeTo(OutputStream)` when some bytes have already been consumed from the
buffer before the call.

## Root Cause

`OutputStream.write(byte[], int off, int len)` expects `len` as the third
argument — a byte count, not an end position. Both affected implementations
pass a position or total length instead of the number of remaining bytes.

### FixedBufferData

    // FixedBufferData.java
    out.write(bytes, readPosition, writePosition);
    //                             ^^^^^^^^^^^^^
    //                             BUG: end position, not length

The correct form, already used by `GrowingBufferData` in the same module:

    out.write(bytes, readPosition, writePosition - readPosition);

When `readPosition == 0` the values are equal and the bug is invisible. When
`readPosition > 0`, `write` reads past the end of valid data into uninitialized
array memory.

Example:
- 10 bytes written → writePosition = 10
- 3 bytes consumed → readPosition = 3
- `out.write(bytes, 3, 10)` copies indices 3–12, but only 3–9 hold valid data
- Indices 10–12 are uninitialized; those 3 extra bytes are written to the stream

### ReadOnlyArrayData

    // ReadOnlyArrayData.java
    out.write(bytes, offset + position, length);
    //                                  ^^^^^^
    //                                  BUG: total length, not remaining length

The correct form, consistent with how `available()` is already computed in the
same class (`return length - position`):

    out.write(bytes, offset + position, length - position);

With a tightly-sized array this throws `IndexOutOfBoundsException`. With an
oversized or reused array it silently writes excess bytes — the same failure
mode as `FixedBufferData`.

## Impact

The output stream receives more bytes than the buffer's available data. No
exception is thrown in the silent case; the extra bytes contain uninitialized
array memory. If the buffer was previously reset via `reset()` and reused, the
bytes past the valid range may contain data from a prior use of the same array.

`CompositeArrayBufferData` and `CompositeListBufferData` both delegate to their
children's `writeTo` and are not independently affected.

## Affected versions

Confirmed present on both active branches:
- `main` (Helidon 27) as of commit `43ebfa7`
- `helidon-4.x` as of commit `23a22ca`

MariusVolkhart:
Fixes #11738

This is functionally identical to #11739, cherry-picked onto `main`.

## Changes

`OutputStream.write(byte[], off, len)` takes a byte count as the third
argument, not an end position. Two implementations passed the wrong value:

- `FixedBufferData`: passed `writePosition` (an end position) instead of
  `writePosition - readPosition`
- `ReadOnlyArrayData`: passed `length` (total slice length) instead of
  `length - position`

Both bugs are invisible when the read position is zero — the position and
the count happen to be equal. They manifest once any bytes have been consumed
from the buffer before `writeTo` is called: the stream receives more bytes than
`available()` reports, sourced from uninitialized array memory beyond the valid
data range. With a tightly-sized array, `ReadOnlyArrayData` throws
`IndexOutOfBoundsException` instead.

`GrowingBufferData` already used the correct form and served as the reference.
`CompositeArrayBufferData` and `CompositeListBufferData` delegate to their
children and are not independently affected.

Regression tests added to `BufferDataTest` (covers both `FixedBufferData`
variants via the existing parameterized setup) and `ReadOnlyArrayDataTest`.

## Documentation

No changes required.

### DIFF
--- a/common/buffers/src/main/java/io/helidon/common/buffers/FixedBufferData.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/FixedBufferData.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -68,7 +68,7 @@ class FixedBufferData implements BufferData {
     @Override
     public void writeTo(OutputStream out) {
         try {
-            out.write(bytes, readPosition, writePosition);
+            out.write(bytes, readPosition, writePosition - readPosition);
             readPosition = writePosition;
         } catch (IOException e) {
             throw new UncheckedIOException(e);
--- a/common/buffers/src/main/java/io/helidon/common/buffers/ReadOnlyArrayData.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/ReadOnlyArrayData.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


tomas-langer:
Can you please rebase on latest `main` and fix the copyright issues? Otherwise examples will not build correctly.

Let me know if you would prefer that I take the PR over.

Thanks for your contribution!

tomas-langer (approved):

Issue #11768, opened 2026-04-23
4.x - Replace JSON-P with Helidon JSON in JSON-RPC


Verdent:
Relates to https://github.com/helidon-io/helidon/issues/11356

PR #11769, opened 2026-04-23
JSON-P removed from the JSON-RPC (4.x)

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/11768

### Documentation

None

### DIFF
--- a/docs/src/main/asciidoc/se/jsonrpc/client.adoc
+++ b/docs/src/main/asciidoc/se/jsonrpc/client.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2025 Oracle and/or its affiliates.
+    Copyright (c) 2025, 2026 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -93,9 +93,12 @@ a JSON-RPC error response.
 
 Every JSON-RPC response contains either a result or an error, and that is
 the reason why `res.result()` returns an optional value. The last step
-shows how the result is mapped to a `StartStopResult` instance using JSON-B.
-See xref:{rootdir}/se/jsonrpc/server.adoc[JSON-RPC Server] for more
-information on these types.
+shows how the result is mapped to a `StartStopResult` instance using Helidon
+JSON binding. See xref:{rootdir}/se/jsonrpc/server.adoc[JSON-RPC Server] for
+more information on these types.
+
+NOTE: Custom types used with Helidon JSON binding must be annotated with
+`@Json.Entity` and compiled with the Helidon JSON annotation processor.
 
 === Batch Requests
 
@@ -117,7 +120,7 @@ The response of type `JsonClientBatchResponse` shall include an entry
 for each of the invocations in the request. In this example, we can test
 that the response returned HTTP status 200 and has a size of 2, and
 then verify the results by binding them to `StartStopResult` instances
-using JSON-B.
+using Helidon JSON binding.
 
 [source,java]
 ----
--- a/docs/src/main/asciidoc/se/jsonrpc/client.adoc
+++ b/docs/src/main/asciidoc/se/jsonrpc/client.adoc
@@ -98,7 +98,9 @@ JSON binding. See xref:{rootdir}/se/jsonrpc/server.adoc[JSON-RPC Server] for
 more information on these types.
 
 NOTE: Custom types used with Helidon JSON binding must be annotated with
-`@Json.Entity` and compiled with the Helidon JSON annotation processor.
+`@Json.Entity` and compiled with the Helidon JSON annotation processor. See
+xref:{rootdir}/se/json/json.adoc#_enabling_code_generation[Enabling Code
+Generation] for the Maven annotation processor setup.
 
 === Batch Requests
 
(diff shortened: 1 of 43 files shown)


jsonrpc/core/pom.xml
```diff
@@ -60,10 +75,40 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.helidon.bundles</groupId>
```
tomas-langer: Never use the bundle in Helidon itself.

webserver/tests/jsonrpc/pom.xml
```diff
@@ -71,4 +76,39 @@
...
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.helidon.bundles</groupId>
```
tomas-langer: Do not use the bundle in this project.

tomas-langer (changes requested):

jsonrpc/core/pom.xml
```diff
@@ -38,18 +38,33 @@
...
         </dependency>
         <dependency>
             <groupId>io.helidon.common.features</groupId>
             <artifactId>helidon-common-features-api</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
+            <artifactId>helidon-common-types</artifactId>
+            <scope>test</scope>
```
tomas-langer: This is most likely a wrong dependency - common types should be a transitive dependency of one of the existing dependencies, so it should not be added in test scope, as that removes it from runtime scope.

jsonrpc/core/pom.xml
```diff
@@ -38,18 +38,33 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.service</groupId>
```
tomas-langer: Same here

jsonrpc/core/pom.xml
```diff
@@ -38,18 +38,33 @@
...
+            <artifactId>helidon-service-registry</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.common</groupId>
```
tomas-langer: And here. This is probably a result of the dependency check plugin. Just disable the plugin.

webclient/jsonrpc/src/main/java/io/helidon/webclient/jsonrpc/JsonRpcClientBatchResponseImpl.java
```diff
@@ -53,16 +51,16 @@ class JsonRpcClientBatchResponseImpl implements JsonRpcClientBatchResponse {
 
     @Override
     public int size() {
-        return asJsonArray().size();
+        return asJsonArray().values().size();
```
tomas-langer: This migration introduces avoidable allocation on the batch hot path. `JsonArray.values()` returns a copied list, so `size()` copies the whole batch here, `get(...)` copies again when building response wrappers, and `iterator()` creates another copy. The server batch path has the same pattern in `JsonRpcRouting`. Please cache the values once for a parsed array, or use/add a non-copying size/iteration API.

jsonrpc/core/src/main/java/io/helidon/jsonrpc/core/JsonRpcParams.java
```diff
@@ -17,55 +17,54 @@
...
     /**
-     * Create an instance from a JSON structure.
+     * Create an instance from a JSON value.
      *
-     * @param params the structure
+     * @param params the value
      * @return a new instance of this class
      */
-    static JsonRpcParams create(JsonStructure params) {
+    static JsonRpcParams create(JsonValue params) {
```
tomas-langer, line 36: The new signature accepts `JsonValue`, but the implementation still only accepts JSON object or array values and rejects scalar/null values. Please keep the public contract aligned with the JSON-RPC constraint here, including documenting the `IllegalArgumentException` for non-object/non-array values.

docs/src/main/asciidoc/se/jsonrpc/client.adoc
```diff
@@ -93,9 +93,12 @@ a JSON-RPC error response.
...
 the reason why `res.result()` returns an optional value. The last step
-shows how the result is mapped to a `StartStopResult` instance using JSON-B.
-See xref:{rootdir}/se/jsonrpc/server.adoc[JSON-RPC Server] for more
-information on these types.
+shows how the result is mapped to a `StartStopResult` instance using Helidon
+JSON binding. See xref:{rootdir}/se/jsonrpc/server.adoc[JSON-RPC Server] for
+more information on these types.
+
+NOTE: Custom types used with Helidon JSON binding must be annotated with
+`@Json.Entity` and compiled with the Helidon JSON annotation processor.
```
tomas-langer: This adds the new Helidon JSON binding requirement, but the guide still only shows the JSON-RPC runtime dependency. Since the examples now rely on `@Json.Entity`, please include or link the required Helidon JSON annotation processor setup so users copying the guide can compile the custom types.

tomas-langer (changes requested):
Requesting changes for the change-caused issues called out inline.

tomas-langer (dismissed):

tomas-langer (approved):

PR #11772, opened 2026-04-23
BufferDataInputStream: implement Drainable, skip, and readAllBytes

### LINKED ISSUE #11742: BufferDataInputStream: implement gRPC performance interfaces
`GrpcProtocolHandler.BufferDataInputStream` currently implements `KnownLength`, which lets gRPC pre-allocate a buffer of the right size. Three additional opportunities exist for more efficient byte processing on the inbound message path:

**`io.grpc.Drainable`**

Callers that detect `Drainable` can drain bytes directly to a destination `OutputStream` without an intermediate copy buffer. Implementing `drainTo(OutputStream)` delegates to `BufferData.writeTo()`, which writes the underlying byte array directly.

**`InputStream.readAllBytes()`**

The default `InputStream.readAllBytes()` implementation accumulates bytes into a growing `byte[]`, doubling it on overflow. Since `available()` returns the exact byte count, `readAllBytes()` can allocate a single exact-size array and fill it in one call.

**`InputStream.skip(long)`**

The default `InputStream.skip(long)` reads and discards bytes via a temporary buffer. Since `BufferData` tracks its read position explicitly, `skip()` can advance the read position directly in O(1) without touching any bytes.

MariusVolkhart:
`BufferDataInputStream` already implements `KnownLength`, letting gRPC pre-allocate a buffer of the right size. This PR adds three more optimizations on the inbound message path:

**`Drainable.drainTo(OutputStream)`**

Callers that detect `Drainable` can drain bytes directly to a destination `OutputStream` without an intermediate copy buffer. The implementation delegates to `BufferData.writeTo()`, which writes the underlying byte array directly.

**`InputStream.readAllBytes()`**

The default implementation accumulates bytes into a growing `byte[]`, doubling it on overflow. Since `available()` returns the exact byte count, the override allocates a single exact-size array and fills it in one call.

**`InputStream.skip(long)`**

The default implementation reads and discards bytes via a temporary buffer. Since `BufferData` tracks its read position explicitly, the override advances the read position directly in O(1) without touching any bytes.

Closes #11742

## Documentation

No changes required.

### DIFF
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -640,9 +640,7 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
         @Override
         public byte[] readAllBytes() {
             // single exact-size allocation; avoids the default growing-array loop
-            byte[] result = new byte[bufferData.available()];
-            bufferData.read(result);
-            return result;
+            return bufferData.readBytes();
         }
     }
 }
--- a/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
+++ b/webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
@@ -66,6 +66,7 @@ import io.grpc.Compressor;
 import io.grpc.CompressorRegistry;
 import io.grpc.Decompressor;
 import io.grpc.DecompressorRegistry;
+import io.grpc.Drainable;
 import io.grpc.KnownLength;
 import io.grpc.Metadata;
 import io.grpc.MethodDescriptor;
@@ -591,7 +592,7 @@ class GrpcProtocolHandler<REQ, RES> implements Http2SubProtocolSelector.SubProto
      * knowledge for optimizations. It can also copy a byte array directly on a
      * single read.
      */
-    static class BufferDataInputStream extends InputStream implements KnownLength {
+    static class BufferDataInputStream extends InputStream implements KnownLength, Drainable {
         private final BufferData bufferData;
 
         BufferDataInputStream(BufferData bufferData) {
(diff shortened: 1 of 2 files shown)


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -620,5 +621,28 @@ public int read(byte[] b, int off, int len) {
...
+            // advance readPosition directly; avoids the default read()-in-a-loop
+            int toSkip = (int) Math.min(bufferData.available(), n);
+            bufferData.skip(toSkip);
+            return toSkip;
+        }
+
+        @Override
+        public byte[] readAllBytes() {
+            // single exact-size allocation; avoids the default growing-array loop
+            byte[] result = new byte[bufferData.available()];
```
tomas-langer: There is a method `BufferData.readBytes()` doing exactly what you do here.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -620,5 +621,28 @@ public int read(byte[] b, int off, int len) {
         public int available() {
             return bufferData.available();
         }
+
+        @Override
+        public int drainTo(OutputStream target) {
+            int count = bufferData.available();
```
tomas-langer, line 627: the method `writeTo` returns the number of bytes actually written, so this can be simplified to `return bufferData.writeTo(target)`

MariusVolkhart: `BufferData.writeTo(OutputStream)` returns `void`. There's an overload for `ByteBuffer` that returns `int`.

tomas-langer: Sorry, my bad

tomas-langer (approved):

PR #11783, opened 2026-04-24
API stability annotations follow up (4.x)

### LINKED ISSUE #11565: API Stability annotations on Helidon APIs (4.x)

tomas-langer:
Resolves #11565

Finishes the remaining 4.x API-stability cleanup after PR #11688 by replacing the remaining Javadoc-only stability signals with explicit annotations where appropriate, keeping the metrics snapshot contract public, and updating the related module wiring.

There may be additional changes in the future, if we decide to change public APIs to internal (i.e. some codegen APIs)

### DIFF
--- a/common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/TemporaryFolderExt.java
+++ b/common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/TemporaryFolderExt.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2018, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -45,8 +45,10 @@ import org.junit.jupiter.api.extension.ExtensionContext;
  * The extension automatically deletes the temporary files after all tests in
  * the test class have finished.
  *
+ * @deprecated Use JUnit 5 {@link org.junit.jupiter.api.io.TempDir @TempDir} support for temporary files and
+ *         directories instead.
  */
-@Deprecated
+@Deprecated(since = "4.5.0", forRemoval = true)
 public class TemporaryFolderExt implements BeforeEachCallback, AfterEachCallback {
 
     private Path root;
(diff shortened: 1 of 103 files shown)


common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/TemporaryFolderExt.java
```diff
@@ -45,8 +45,10 @@
  * The extension automatically deletes the temporary files after all tests in
  * the test class have finished.
  *
+ * @deprecated Use JUnit 5 {@link org.junit.jupiter.api.io.TempDir @TempDir} support for temporary files and
+ *         directories instead.
  */
-@Deprecated
+@Deprecated(since = "4.5.0", forRemoval = true)
```
romain-grecourt, line 51: Should this be both deprecated and internal ?

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
```diff
@@ -120,7 +120,7 @@
  *
  * @deprecated Please use {@link PersistenceExtension} instead.
  */
-@Deprecated(since = "4.0")
+@Deprecated(since = "4.0.0", forRemoval = true)
 @SuppressWarnings("checkstyle:IllegalToken") // deprecated, to be removed
 public class JpaExtension implements Extension {
```
romain-grecourt, line 126: Should the constructor be marked internal ?

common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/TemporaryFolderExt.java
tomas-langer: I don't think that makes sense, as it is removed in 27. The internal information would not really help I think.

integrations/cdi/jpa-cdi/src/main/java/io/helidon/integrations/cdi/jpa/JpaExtension.java
tomas-langer: yes

tomas-langer: Also this whole module is deprecated and will be removed...

common/testing/junit5/src/main/java/io/helidon/common/testing/junit5/TemporaryFolderExt.java
tomas-langer: I kept it only deprecated for removal, as it feels much stronger than internal already.

romain-grecourt (approved):

PR #11786, opened 2026-04-24
Add validation codegen tests#11372

### LINKED ISSUE #11372: Add validation codegen tests
As part of solving https://github.com/helidon-io/helidon/issues/11360 I have added new codegen tests (the module did not exist before) to validate that codegen correctly fails on the reported issue (i.e. private field with a constraint).

We should add tests to validate that we correctly fail for all unsupported elements, and pass for all supported elements.

shankaritchapurapu:
### Description

Resolves https://github.com/helidon-io/helidon/issues/11372

https://github.com/helidon-io/helidon/issues/11360 resolves an issues where unsupported type (private field with constraint) is causing Codegen failure. 

This is a followup change to add tests to Codegen to cover for all supported and unsupported type of kind/elements.

Supported types
==============
class: non-private non-static field + supported methods
record: record components
interface: supported methods

Unsupported Types
===============
private field
static field
private method
static method
method with args
void method
constructor
Negative test for unsupported @Validation.Validated type kind (enum), asserting the illegal type error

### Documentation
None

Testing
-------
Build project and run the tests successfully with the following command (locally)
mvn -f declarative/tests/codegen/pom.xml -Dtest=ValidationCodegenTest test

### DIFF
--- a/declarative/tests/codegen/src/test/java/io/helidon/declarative/codegen/validation/ValidationCodegenTest.java
+++ b/declarative/tests/codegen/src/test/java/io/helidon/declarative/codegen/validation/ValidationCodegenTest.java
@@ -31,13 +31,12 @@ import io.helidon.validation.Validation;
 
 import org.junit.jupiter.api.Test;
 
-import static io.helidon.codegen.testing.CodegenMatchers.matches;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.MatcherAssert.assertThat;
 
-public class ValidationCodegenTest {
+class ValidationCodegenTest {
     private static final List<Class<?>> CLASSPATH = List.of(
             Generated.class,
             Validation.Constraint.class,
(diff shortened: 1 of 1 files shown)


declarative/tests/codegen/src/test/java/io/helidon/declarative/codegen/validation/ValidationCodegenTest.java
```diff
@@ -47,69 +46,245 @@ public class ValidationCodegenTest {
...
+        assertThat("Build should fail", result.success(), is(false));
         assertThat(diagnostics, containsString("error:"));
-        assertThat(diagnostics, containsString("private Integer id"));
-        assertThat(diagnostics, containsString("Only non-private fields and getter methods are supported"));
+        for (String diagnosticPart : diagnosticParts) {
+            assertThat(diagnostics, containsString(diagnosticPart));
+        }
+    }
+
+    private static String capitalize(String name) {
```
tomas-langer: If this depends on codegen, there is already io.helidon.codegen.CodegenUtil#capitalize - so no need to create this method

tomas-langer (changes requested):

tomas-langer (approved):

PR #11790, opened 2026-04-24
Fix timeout current-thread completion logging (27.x)

### LINKED ISSUE #11789: [27.x] Fault tolerance Timeout logging error message when invocation completes before timeout
Carry-forward of #11560 for `main` / `27.x`.

Issue #11560 reports that `TimeoutConfig.builder().currentThread(true)` logs `Delayed runnable was unexpectedly interrupted` even when the invocation completes before the configured timeout.

The same cancellation path is still present on current `main` (`96ac8fb6a0`): `TimeoutImpl` cancels the sleeping monitor future with `cancel(true)`, and `FaultTolerance.delayedRunnable` logs the resulting `InterruptedException` as an error even though this interruption is expected during normal completion.

This issue tracks the 27.x fix and regression coverage.

tomas-langer:
Resolves #11789

Avoid logging an error when a `currentThread(true)` timeout monitor is interrupted by normal cancellation after the supplier completes. Add regression coverage that forces the monitor thread into the sleeping state and verifies normal completion does not emit the spurious log message.

### DIFF
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
@@ -87,8 +87,9 @@ public final class FaultTolerance {
             try {
                 Thread.sleep(millis);
             } catch (InterruptedException e) {
-                // should never be interrupted
-                LOGGER.log(ERROR, "Delayed runnable was unexpectedly interrupted");
+                // Future.cancel(true) interrupts the sleeping task to cancel the delayed action.
+                Thread.currentThread().interrupt();
+                return;
             }
             runnable.run();
         };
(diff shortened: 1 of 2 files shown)


fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FaultTolerance.java
```diff
@@ -87,8 +87,9 @@ static Runnable delayedRunnable(Runnable runnable, long millis) {
             try {
                 Thread.sleep(millis);
             } catch (InterruptedException e) {
-                // should never be interrupted
-                LOGGER.log(ERROR, "Delayed runnable was unexpectedly interrupted");
+                // Future.cancel(true) interrupts the sleeping task to cancel the delayed action.
```
tomas-langer, line 90: This thread is under our full control, so we are the ones interrupting it.

tvallin (approved):

PR #11793, opened 2026-04-24
Fix builder generic type variable handling (27.x)

### LINKED ISSUE #11533: Builder codegen generates invalid source for generic blueprints with properties typed by blueprint type variables
Helidon Builder appears to generate invalid Java source for a generic blueprint when one of the blueprint's property types refers to the blueprint’s own type variables.

### Minimal example

```
@Prototype.Blueprint
interface ExampleBlueprint<A, T> {
  Function<A, T> fn();
}
```

### Expected

Generated code should preserve the type variables correctly, e.g. something equivalent to:

```
Function<A, T>
```

### Actual

Generated code qualifies the type variables as if they were real concrete types (!), producing invalid source like:

```
Function<some.package.A, some.package.T>
```

or similar variants using renamed type parameters.

That does not compile, because `A` and `T` are type variables, not package-level classes.

### Impact

This prevents builderizing generic types whose properties include generic JDK functional interfaces such as:

- `Function<A, T>`
- `Supplier<T>`
- etc.

### Workaround

Keep the generic type and its builder handwritten for now.

tomas-langer:
Resolves #11533

Preserve generic declarations when builder codegen repairs blank-package type names so type variables such as `T` and `X` are not rewritten as same-package classes. Add a focused builder regression case that uses `Function<T, X>` on a generic blueprint.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryOption.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryOption.java
@@ -884,15 +884,69 @@ final class FactoryOption {
                 .build();
     }
 
-    // if no package, assume same package as blueprint (generated as part of this annotation round)
+    // Blank-package names are either same-round generated types or type variables.
+    // Keep the latter unqualified and recursively repair the former.
     static TypeName fixEmptyPackage(TypeName typeName, String packageName) {
-        if (typeName.packageName().isBlank()) {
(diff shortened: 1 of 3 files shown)


danielkec (approved):
LGTM. I reviewed the builder codegen change and the focused generics regression. The added `Function<T, X>` path generates the expected unqualified type variables, and the focused `GenericTest` path passes.

PR #11812, opened 2026-04-26
Avoid unnecessary DNS resolution without proxy

### LINKED ISSUE #11800: Http1CallChainBase.prologue() does unnecessary DNS resolution per request when proxy is not used
Helidon Version: 4.4.0

## Enhancement Description

In `Http1CallChainBase.prologue()` , `new InetSocketAddress(uri.host(), uri.port())` is constructed on every request. The constructor performs DNS resolution under the hood, but the resulting `uriAddress` is only used by `proxy.isNoHosts(uriAddress)`, which is short-circuited when `proxy == Proxy.noProxy()`:

https://github.com/helidon-io/helidon/blob/4e4bb38c5d1be40d9c43fc342c51091c7f790fef/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java#L188-L194

For users not using a proxy (a common case), the DNS resolution is wasted work on the hot path of every request — including requests served by reused keep-alive connections that would otherwise touch no DNS at all.

### Why this matters

In most setups, the JVM's positive DNS cache (`networkaddress.cache.ttl`, default 30s) absorbs the cost so it's typically unnoticed. In my case though — through my own oversight — I was running with the k8s default `ndots:5` combined with multiple search domains in `/etc/resolv.conf`, which made each cache-miss fire many backend DNS queries. Any slowness on the resolver side then directly added to request latency. While this was a misconfiguration on my side, it illustrates how the unnecessary per-request DNS resolution can amplify into a noticeable problem in environments that aren't ideally tuned. Even in well-tuned environments, removing per-request work that yields no functional benefit seems desirable.

### Suggested fix

Inline the `new InetSocketAddress(...)` so it is only evaluated when actually needed (short-circuited by
`||`):

```java
String requestUri = proxy == Proxy.noProxy()
        || (proxy.type() == Proxy.ProxyType.HTTP && proxy.isNoHosts(new InetSocketAddress(uri.host(),
uri.port())))
        || (proxy.type() == Proxy.ProxyType.SYSTEM && !proxy.isUsingSystemProxy(absoluteUri))
        || clientConfig.relativeUris()
        ? "" // don't set host details, so it becomes relative URI
        : absoluteUri;
```

wadhwaroh-lang:
Delay construction of the proxy no-host address until the HTTP proxy branch needs it, keeping the common no-proxy request path free of per-request DNS resolution.

### Description

Fixes #11800.

`Http1CallChainBase.prologue()` previously created an `InetSocketAddress` for every request before checking whether proxy handling was needed. `InetSocketAddress(String, int)` can perform DNS resolution, so this added unnecessary per-request work when `proxy == Proxy.noProxy()`.

This change moves the `InetSocketAddress` construction into the HTTP proxy `isNoHosts(...)` branch, allowing the common no-proxy path to short-circuit before any address resolution.

### Documentation

None. This is an internal performance optimization with no user-facing API or configuration changes.

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -185,9 +185,9 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
             // https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html which states: "The absoluteURI form is REQUIRED when the
             // request is being made to a proxy."
             String absoluteUri = uri.scheme() + "://" + uri.host() + ":" + uri.port();
-            InetSocketAddress uriAddress = new InetSocketAddress(uri.host(), uri.port());
             String requestUri = proxy == Proxy.noProxy()
-                    || (proxy.type() == Proxy.ProxyType.HTTP && proxy.isNoHosts(uriAddress))
+                    || (proxy.type() == Proxy.ProxyType.HTTP
+                            && proxy.isNoHosts(new InetSocketAddress(uri.host(), uri.port())))
                     || (proxy.type() == Proxy.ProxyType.SYSTEM && !proxy.isUsingSystemProxy(absoluteUri))
                     || clientConfig.relativeUris()
                     ? "" // don't set host details, so it becomes relative URI


wadhwaroh-lang:
@hozumi @romain-grecourt  @danielkec  Can either of you please review? Thanks!

hozumi:
LGTM, Thank you for fixing this!

tomas-langer (approved):

Issue #2990, opened 2021-04-30
Enable custom values to be added to the ExecutionContext in graphql calls


schoenobates:
## Environment Details
* Helidon Version: 2.2.2
* Helidon SE or Helidon MP: SE
* JDK version: 11
* OS: MacOS 11.2.3

----------

## Problem Description

When handling a graphql invocation, it would be useful to attach things to the execution context - e.g. the current Authorisation header, a request ID for logging etc etc.

Currently there is no way to do this, but perhaps the following would be good:

+ an update to the `ExecutionContext` to add a context value: `void setContextValue(String, Object)`
+ corresponding getter (similar perhaps to the config style getters which give access to type safety)
+ builder method on the InvocationHandler builder to add handlers which can add things to the context (e.g. a token from a header for example, or so kind of dynamically calculated value).

tmiddlet2666:
@schoenobates 
You have mentioned the ability to setContextValue() on the ExecutionContext.
Are you wanting this to work in Helidon SE or Helidon MP?

schoenobates:
Hi @tmiddlet2666 - I was playing around with this on SE. I had a prototype going in a repo.

mrguamos:
any update on this? I need to get the Authorization header in the `ServerRequest` but this is not being passed down from `GraphQlService` to `InvocationHandler` or is there any other way to secure Graphq

```java
    private void graphQlPost(ServerRequest req, ServerResponse res) {
        LinkedHashMap entity = JSONB.fromJson(req.content().inputStream(), LINKED_HASH_MAP_GENERIC_TYPE.type());
        processRequest(res,
                       (String) entity.get("query"),
                       (String) entity.get("operationName"),
                       toVariableMap(entity.get("variables")));
    }
```

I'm using Helidon 4 SE

shankaritchapurapu:
PR -> https://github.com/helidon-io/helidon/pull/11855

shankaritchapurapu:
Merged PR https://github.com/helidon-io/helidon/pull/11855

PR #11855, opened 2026-05-04
Enable custom values to be added to the ExecutionContext in graphql calls

shankaritchapurapu:
resolves ... #2990: Enable custom values to be added to the ExecutionContext in graphql calls

### Description
Added support for per-invocation GraphQL context values.

ExecutionContext now lets callers set and retrieve named values, including typed retrieval via contextValue(String, Class<T>). ExecutionContextImpl stores those values in a concurrent map.

InvocationHandler.Builder now accepts addContextHandler(...) callbacks. Each callback runs before GraphQL execution, so it can attach request-specific data such as an authorization token, request ID, or logging metadata to the ExecutionContext.

InvocationHandlerImpl applies those handlers before creating the ExecutionInput, making the populated ExecutionContext available to GraphQL Java data fetchers.

A new test verifies both typed context access and that context values set by a builder handler are visible during data fetching.

### Documentation

InvocationHandlerContextTest.java shows example of setting the values to the ExectionContext

### DIFF
--- a/graphql/server/src/main/java/io/helidon/graphql/server/ExecutionContext.java
+++ b/graphql/server/src/main/java/io/helidon/graphql/server/ExecutionContext.java
@@ -20,6 +20,8 @@ import java.util.Optional;
 
 /**
  * GraphQL execution context to support partial results and context values.
+ * Context values are backed by Helidon common context and are also available from
+ * {@link io.helidon.common.context.Contexts#context()} during GraphQL execution.
  */
 public interface ExecutionContext {
     /**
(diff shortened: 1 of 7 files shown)


graphql/server/src/main/java/io/helidon/graphql/server/ExecutionContext.java
```diff
@@ -16,10 +16,39 @@
...
+ * GraphQL execution context to support partial results and context values.
  */
 public interface ExecutionContext {
+    /**
+     * Set a context value.
+     *
+     * @param name name of the context value
+     * @param value context value
+     */
+    void setContextValue(String name, Object value);
```
tomas-langer: These methods should be `default`, as this would break compatibility of existing implementations (even in tests)

tomas-langer:
We have `helidon-common-context` feature, that can be added to webserver through `helidon-webserver-context` module. Maybe this could be reused for this purpose?

tomas-langer (approved):

PR #11867, opened 2026-05-05
 Decouple concurrency limit metrics from socket naming

### LINKED ISSUE #11826: Decouple concurrency Limit metrics from socket naming
## Problem

The common concurrency `Limit` API is currently initialized with `Limit.init(String socketName)`, and the built-in limit metrics use that value as the `socketName` metric tag.

That was adequate when WebServer request limits were listener/socket scoped, but the abstraction is broader than WebServer sockets:

- `Limit` lives in `common/concurrency/limits` and can be used outside WebServer.
- WebServer now has listener-owned limits and binding-specific limits.
- Binding-specific limit metric values such as `@default-unix` or `@default-quic` can be useful under the existing `socketName` tag, but they are no longer strictly socket names.

This leaves generic `Limit` API and documentation tied to WebServer socket terminology.

## Desired Outcome

Decouple `Limit` initialization and metrics tagging from socket-specific naming while preserving compatibility for existing users and current WebServer metrics.

Possible direction:

- Introduce a generic initialization context for `Limit`, such as an origin name plus metric tags.
- Let built-in limit implementations use caller-provided metric tags instead of always creating `socketName=<value>`.
- Keep `init(String)` as a compatibility path, or deprecate it only after a replacement exists.
- Update WebServer to pass listener or binding origin information through the generic API.

## Acceptance Criteria

- The common `Limit` API and Javadocs no longer require socket terminology for non-WebServer usage.
- Built-in limit metrics can be initialized with a generic origin and/or caller-provided tags.
- WebServer can continue to expose listener-level and binding-specific request-limit metrics in a compatible or clearly documented way.
- Migration impact for existing metrics using the `socketName` tag is explicitly documented.

## Context

This came up during the WebServer transport binding work for #11811. The current branch documents binding-specific values such as `@default-unix` under the existing `socketName` tag, but the broader API cleanup should be handled as a separate issue.

shankaritchapurapu:
resolves https://github.com/helidon-io/helidon/issues/11826

### Description
The ticket is about decoupling common concurrency Limit metrics from WebServer socket naming. Previously, the common API had:

`Limit.init(String socketName)`
and built-in metrics always turned that into socketName=<value>. That made a common module depend conceptually on WebServer socket terminology.
**API Change**

In 
`Limit.java`
, we added a nested runtime initialization type:

`Limit.InitializationContext`
It carries:

originName
metricTags
New usage:

`limit.init(Limit.InitializationContext.create(originName, tags));`
The old API remains:

`limit.init(String originName)`
So existing users are not broken.

Metrics Change

In SemaphoreMetrics.java, metrics registration now accepts caller-provided tags from Limit.InitializationContext.

Legacy behavior is preserved through init(String):

> default origin -> no socketName tag
> non-default origin -> socketName=<originName>
> So existing WebServer dashboards continue to work.

Limit Implementations Updated

Updated these to use Limit.InitializationContext:

> SemaphoreLimitBase.java
> AimdLimit.java
> AimdLimitImpl.java

They now store context.originName() for outcomes and pass context.metricTags() to metrics.

WebServer Updated

In 
ServerListener.java
, WebServer now initializes limits with:

`Limit.InitializationContext`
But it intentionally preserves the existing WebServer tag contract:

> default socket -> no socketName
> named socket -> socketName=<socketName>

**Docs Updated**

concurrency-limits.adoc
 now explains:

common Limit can use caller-provided metric tags
WebServer keeps socketName compatibility
existing dashboards using socketName continue to work

**Tests Added/Updated**

Added LimitMetricsTest.java to verify:

> custom context tags pass through unchanged
> map tags convert correctly
> legacy named origin creates socketName
> default legacy origin omits socketName

Updated FixedLimitMetricsTest.java to verify named socket metrics expose socketName="admin" via /observe/metrics.

**Net result:** 
common Limit no longer requires socket terminology, while WebServer metrics remain backward-compatible.

### Documentation
=== Limit Initialization and Metric Tags

Concurrency limits can be initialized with a generic origin and optional metric
tags using `Limit.InitializationContext`. This avoids coupling the common
`Limit` API to WebServer socket terminology.

```

Limit limit = FixedLimit.builder()
        .permits(100)
        .enableMetrics(true)
        .build();

limit.init(Limit.InitializationContext.create(
        "batch-import",
        Map.of("component", "inventory",
               "origin", "batch-import")));
```

Built-in limit metrics use the tags provided in the initialization context.

For compatibility, `Limit.init(String)` is still supported. When WebServer
initializes listener or binding limits, it preserves the existing metrics
behavior: non-default origins use the `socketName=<origin>` tag, while the
default origin omits the `socketName` tag. Existing dashboards that rely on
`socketName` continue to work.

### DIFF
--- a/common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
+++ b/common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
@@ -16,7 +16,11 @@
 
 package io.helidon.common.concurrency.limits;
 
+import java.util.List;
+import java.util.Objects;
+
 import io.helidon.config.NamedService;
+import io.helidon.metrics.api.Tag;
 import io.helidon.service.registry.Service;
 
 /**
@@ -35,8 +39,75 @@ public interface Limit extends LimitAlgorithm, NamedService {
      * Initialization method for this limit. This method can be used for any
      * task, including metrics initialization.
      *
-     * @param socketName socket name for this limit such as {@code "@default"}
+     * @param context the context for limit initialization
      */
-    default void init(String socketName) {
+    default void init(InitializationContext context) {
+        init(context.originName());
+    }
+
+    /**
+     * Initialization method for this limit. This method can be used for any
+     * task, including metrics initialization.
+     *
+     * @param originName origin name for this limit, such as {@code "@default"}
+     */
+    default void init(String originName) {
+    }
+
+    /**
+     * Runtime context used when initializing a {@link Limit}.
+     */
+    interface InitializationContext {
+        /**
+         * Create a limit context with no metric tags.
+         *
+         * @param originName origin name of the work protected by the limit
+         * @return a new limit context
+         */
+        static InitializationContext create(String originName) {
+            return createContext(originName, List.of());
+        }
+
+        /**
+         * Create a limit context with metric tags.
+         *
+         * @param originName origin name of the work protected by the limit
+         * @param metricTags metric tags to use when registering limit metrics
+         * @return a new limit context
+         */
+        static InitializationContext create(String originName, List<Tag> metricTags) {
+            return createContext(originName, metricTags);
+        }
+
+        /**
+         * Origin name of the work protected by the limit.
+         *
+         * @return origin name
+         */
+        String originName();
+
+        /**
+         * Metric tags to use when registering limit metrics.
+         *
+         * @return metric tags
+         */
+        List<Tag> metricTags();
+
+        private static InitializationContext createContext(String originName, List<Tag> metricTags) {
+            Objects.requireNonNull(originName);
+            List<Tag> copiedTags = List.copyOf(metricTags);
+
+            return new InitializationContext() {
+                @Override
+                public String originName() {
+                    return originName;
+                }
+
+                @Override
+                public List<Tag> metricTags() {
+                    return copiedTags;
+                }
+            };
+        }
     }
 }
--- a/common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
+++ b/common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
@@ -50,7 +50,9 @@ public interface Limit extends LimitAlgorithm, NamedService {
      * task, including metrics initialization.
      *
      * @param originName origin name for this limit, such as {@code "@default"}
+     * @deprecated use {@link #init(InitializationContext)} to provide initialization details explicitly.
      */
+    @Deprecated
     default void init(String originName) {
     }
 
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
@@ -190,13 +190,6 @@ class ServerListener implements ListenerContext {
         ith.start();
     }
 
-    private static InitializationContext limitContext(String socketName) {
-        if (WebServer.DEFAULT_SOCKET_NAME.equals(socketName)) {
-            return InitializationContext.create(socketName);
-        }
-        return InitializationContext.create(socketName, List.of(Tag.create("socketName", socketName)));
-    }
-
     @Override
     public MediaContext mediaContext() {
         return mediaContext;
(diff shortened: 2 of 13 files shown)


shankaritchapurapu:
Do we need the backward compatible API limit.init(String originName)?

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
```diff
@@ -35,8 +40,101 @@ public interface Limit extends LimitAlgorithm, NamedService {
...
      */
-    default void init(String socketName) {
+    @Deprecated
+    default void init(String originName) {
+    }
+
+    /**
+     * Runtime context used when initializing a {@link Limit}.
+     */
+    final class InitializationContext {
```
tomas-langer: Change this to interface with the static factory methods. The rule is that if we would end up with more than two static factory methods, we should introduce a builder. I think we can live without it for now (see later comments).
The implementation should be a package private or private inner class.

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
```diff
@@ -35,8 +40,101 @@ public interface Limit extends LimitAlgorithm, NamedService {
...
+        }
+
+        /**
+         * Create a limit context with metric tags.
+         *
+         * @param originName origin name of the work protected by the limit
+         * @param metricTags metric tag names and values to use when registering limit metrics
+         * @return a new limit context
+         */
+        public static InitializationContext create(String originName, Map<String, String> metricTags) {
```
tomas-langer: Do not add this method, methods with maps as parameters are not great for the future, and this is not needed, as it can be easily converted to `List<Tag>`

common/concurrency/limits/src/main/java/io/helidon/common/concurrency/limits/Limit.java
```diff
@@ -35,8 +40,101 @@ public interface Limit extends LimitAlgorithm, NamedService {
...
+        }
+
+        /**
+         * Create a limit context preserving the legacy WebServer socket-name metric tag behavior.
+         * The {@code socketName} tag is omitted for the default socket.
+         *
+         * @param socketName WebServer socket name
+         * @return a new limit context
+         */
+        public static InitializationContext createForLegacySocketName(String socketName) {
```
tomas-langer: Remove this method, change the call sites instead to use the new factory method. There is no need to have a "legacy" method when this is a new interface.

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -196,6 +197,10 @@ private void initServerThread() {
                 .unstarted(this::listen);
     }
 
+    private static Limit.InitializationContext limitContext(String socketName) {
+        return Limit.InitializationContext.createForLegacySocketName(socketName);
```
tomas-langer: if this stay a one line helper method, remove it (as it is only called from one place). If it will become more complicated (i.e. checking if default, and adding tags otherwise, keep it)

shankaritchapurapu (commented):
Addressed comments

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -196,6 +198,13 @@ private void initServerThread() {
                 .unstarted(this::listen);
     }
 
+    private static Limit.InitializationContext limitContext(String socketName) {
```
tomas-langer: Please move the private method after `public` and package private methods.

tomas-langer (approved):

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -187,6 +190,13 @@ class ServerListener implements ListenerContext {
         ith.start();
     }
 
+    private static InitializationContext limitContext(String socketName) {
```
tomas-langer: This should be after public and package private methods.

tomas-langer (changes requested):

shankaritchapurapu (commented):
Refactored code as per review comments.

tomas-langer (approved):

Issue #11853, opened 2026-05-03
Unix Domain Socket


tomas-langer:
This is an epic to resolve Unix Domain Socket (UDS) issues in Helidon

All of these fixes must be ported to `main`

- [x] UDS client TLS handshake can block indefinitely (fixed on `main` by #11930)
- [x] UDS client TLS failure can leave the channel open (fixed on `main` by #11930)
- [x] UDS client docs do not describe transport semantics (fixed on `main` by #11930)
- [x] UDS test module has bad artifact id (typo) (fixed on `main` by #11930)

tomas-langer:
All the issues are now resolved.

PR #11871, opened 2026-05-06
Fix WebServer TLS over Unix domain sockets (4.x)

### LINKED ISSUE #11845: WebServer TLS over Unix domain sockets fix (4.x)
## Summary

Backport the WebServer TLS-over-Unix-domain-socket fix to `helidon-4.x`.

Source issue: #11844

The same server-side TLS setup bug is present on `origin/helidon-4.x`: the NIO TLS path calls `SocketChannel.socket().getRemoteSocketAddress()` before creating the server `SSLEngine`. That works for Internet socket channels, but `SocketChannel.socket()` is unsupported for Unix-domain socket channels.

## Evidence

On `origin/helidon-4.x`:

- `webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java:283-291`

  ```java
  private HelidonSocket createNioSocket(Tls tls, SocketChannel channel, String channelId) {
      if (tls.enabled()) {
          var address = channel.socket().getRemoteSocketAddress();

          SSLEngine engine;
          if (address instanceof InetSocketAddress isa) {
              engine = tls.sslContext().createSSLEngine(isa.getHostName(), isa.getPort());
          } else {
              engine = tls.sslContext().createSSLEngine();
          }
  ```

- `webserver/tests/unix-domain-socket` currently has only the plain Unix-domain socket test and no TLS UDS regression coverage or TLS test resources.

## Expected behavior

TLS-enabled Unix-domain socket listeners should accept and serve HTTPS requests over UDS. The server-side TLS path should only use host/port peer information when the channel remote address is an `InetSocketAddress`; Unix-domain channels should use an address-less server `SSLEngine`.

## Suggested fix

Backport the 27.x fix: use `SocketChannel.getRemoteAddress()` in the NIO TLS path and create the host/port `SSLEngine` only for `InetSocketAddress`. For other socket address types, create the engine without peer host/port.

Add focused coverage under `webserver/tests/unix-domain-socket` with TLS cert/key resources. At minimum, cover a TLS-enabled Unix-domain WebServer serving a request over UDS.

tomas-langer:
Resolves #11845

Fixes TLS handling for WebServer Unix-domain-socket listeners by avoiding TCP-only peer lookup in the NIO TLS path. Adds the minimal WebClient UDS TLS support needed for the regression coverage, including logical TLS peer handling, ALPN, handshake timeout cleanup, and UDS-aware connection keys for HTTP/1, HTTP/2, generic WebClient, and gRPC.

Also corrects the Unix-domain-socket test module artifact id tracked under #11853.

There will be a follow up set of PRs to resolve all the client UDS issues, so please do not report them - I already know.


webclient/api/src/main/java/io/helidon/webclient/api/ClientRequest.java
```diff
@@ -95,6 +95,10 @@ default T path(String uri) {
      * Socket address to use. This can be used to connect to a UNIX domain socket.
      * Note that if UNIX domain socket address is configured, all other address configuration is ignored.
      * This includes if it is configured as a base address on the client.
+     * <p>
+     * The socket address overrides the physical transport only. The request URI, resolved against the client base URI,
```
danielkec, line 99: This contract is not true yet: `ClientRequestBase.address(UnixDomainSocketAddress)` still rewrites the logical URI to `localhost:0`, so Host/SNI/cache keys use the wrong authority. Preserve `clientUri` for UDS and store only the socket address as the transport override.

webserver/tests/unix-domain-socket/src/test/java/io/helidon/webserver/tests/unix/domain/socket/UnixDomainSocketTest.java
```diff
@@ -83,4 +95,240 @@ public void test() {
...
+        WebServer server = startTlsServer(address, "Hello Secure World!", "Hello Secure HTTP/2!");
+        WebClient webClient = WebClient.builder()
+                .shareConnectionCache(false)
+                .protocolPreference(List.of(Http2Client.PROTOCOL_ID, Http1Client.PROTOCOL_ID))
+                .baseUri("https://localhost")
+                .tls(clientTls())
+                .build();
+        Http1Client http1Client = Http1Client.builder()
+                .shareConnectionCache(false)
+                .baseUri("https://localhost")
```
danielkec, line 113: This only covers `localhost`, which hides the UDS authority rewrite. Add a non-localhost logical authority/certificate case, such as `https://service.example` over UDS, to verify Host/SNI/cache keys stay logical while transport uses the socket path.

common/task/src/main/java/module-info.java
```diff
@@ -19,6 +19,8 @@
  */
 module io.helidon.common.task {
 
+    requires io.helidon.common;
```
danielkec, line 22: `io.helidon.common` appears to be used only for the class-retention `@Api.Internal` annotation on `DeadlineGuard`; can this be `requires static io.helidon.common` with an optional Maven dependency to avoid adding a runtime dependency to `common-task`?

PR #11874, opened 2026-05-06
Fix codegen TypeName annotation handling (27.x)

### LINKED ISSUE #11869: Forward-port #11868: Codegen Core And Class Model Issues for Validation
Forward-port the codegen core and class-model fixes required by declarative validation support for interface-declared constraints.

The source 4.x issue is #11868.
Blocks #11866.

Fixed issues to carry forward:

- `TypeHierarchy` only reported some annotations nested inside generic type arguments and missed annotations declared directly on a `TypeName`, wildcard bounds, and array component types. Add `typeNameAnnotations(...)` and coverage for bounded and array component cases.
- Validation needs to merge type-use annotations from contract signatures into implementation signatures. Add `TypeHierarchy.mergeTypeNameAnnotations(...)` so structurally matching nested type arguments, wildcard bounds, and component types retain annotations.
- `ContentSupport.addCreateTypeName(...)` uses a string `TypeName.create(...)` form that cannot faithfully recreate annotated nested type names. Emit builder-based `TypeName` construction when nested annotations are present.
- The helper for generated string literals in `ContentSupport` must use `ContentBuilder#addContentLiteral(...)` instead of hand-built string quoting.
- `ContentSupport.addCreate(...)` for typed elements does not include enclosing type information. Add enclosing-type emission.

tomas-langer:
Resolves #11869

Forward-port of #11870 from the 4.x line.

Summary:
- preserve nested type-use annotations when discovering and merging TypeName metadata across type arguments, wildcard bounds, and array component types
- emit builder-based TypeName construction when nested annotations must be preserved
- use faithful string literal escaping for generated annotation values and include typed-element enclosing type metadata
- add focused TypeHierarchy and compile-through TypeName codegen coverage

### DIFF
--- a/codegen/codegen/src/main/java/io/helidon/codegen/TypeHierarchy.java
+++ b/codegen/codegen/src/main/java/io/helidon/codegen/TypeHierarchy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2024, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -32,6 +33,7 @@ import io.helidon.common.types.TypeInfo;
 import io.helidon.common.types.TypeName;
 import io.helidon.common.types.TypedElementInfo;
 
+import static io.helidon.common.types.TypeNames.GENERATED;
 import static io.helidon.common.types.TypeNames.INHERITED;
 import static java.util.function.Predicate.not;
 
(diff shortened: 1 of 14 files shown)


codegen/codegen/src/main/java/io/helidon/codegen/TypeHierarchy.java
```diff
@@ -230,75 +244,263 @@ public static Set<TypeName> nestedAnnotations(CodegenContext ctx, TypeInfo typeI
...
+        Set<TypeName> processedTypes = new HashSet<>();
+        String packageName = type.typeName().packageName();
+
+        type.superTypeInfo().ifPresent(it -> collectInheritedMethods(
+                processedTypes,
+                prototypes,
+                it,
+                element,
+                packageName));
+        type.interfaceTypeInfo().forEach(it -> collectInheritedMethods(
```
danielkec, line 325: This still misses generic contract overrides. For example, `Contract<T>.validate(List<@NotBlank T>)` implemented as `Contract<String>.validate(List<String>)` will not populate `prototypes`, because inherited method lookup later requires exact parameter `TypeName` equality (`List<String>` vs `List<T>`), so the inherited type-use annotation is not merged.

This appears to be present in the original 4.x PR #11870 as well. Should we create a follow-up issue to track that?

tomas-langer: This should now be fixed.

Verdent (approved):

PR #11875, opened 2026-05-06
Fix service registry codegen for interface-contract validation (4.x)

### LINKED ISSUE #11872: 4.x: Fix service registry codegen for interface-contract validation
Service registry and service codegen need follow-up work for the interface-method validation path split out of #11865 after #11870.

When constraints are declared only on service contracts, generated validation must still be invoked consistently for registry services and factory-provided services. The current service registry and service codegen paths still need to preserve interception wrappers, factory behavior, and generated descriptor metadata across direct and broad lookup shapes.

Expected behavior:
- factory-provided service lookup routes provided-contract results through interception wrappers while preserving concrete provider lookup compatibility;
- broad empty-contract lookup does not expose raw factory-provider descriptors that can bypass generated wrappers;
- generated factory wrappers intercept the provider factory SPI methods such as `get()`, `services()`, `first(...)`, and `list(...)`;
- supplier, services, injection-point, qualified, and optional-supplier factory wrappers preserve provider-specific `first(...)` and `list(...)` behavior without eager instance creation or null delegates;
- factory adapter raw-method selection matches the expected factory SPI signatures, not only method name and arity;
- direct provider-contract lookup preserves the requested provider contract in `ServiceInstance.contracts()`;
- generated wrapper descriptors retain provider metadata such as `@Weight` and `@Service.RunLevel`;
- registry codegen extension processing can opt into service-contract annotation discovery, includes contract-triggered annotations in the round type map, and avoids repeated uncached contract scans.

Add focused service-registry and service-codegen regression coverage for the lookup and factory shapes above.

tomas-langer:
Resolves #11872

Fixes service registry/codegen handling for services whose validation constraints are discovered through implemented service contracts.

The change updates service codegen and registry metadata so generated interception wrappers preserve service/factory contract information, route factory-provided services through the correct wrappers, and keep element interceptor qualifiers exact for each generated service type and method signature.

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
@@ -17,7 +17,10 @@
 package io.helidon.declarative.codegen.validation;
 
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -74,8 +77,16 @@ class InterceptorGenerator {
             // ignore annotations, these are good
             return;
         }
+        if (factoryInterceptionWrapper(type)) {
+            // The original factory provider owns the element interceptors used by the generated wrapper.
+            return;
+        }
 
         if (!isService(type)) {
+            if (hasOnlyNonPrivateInterfaceMethodValidation(type)) {
+                // Interface method constraints are validated by interceptors generated for service implementations.
+                return;
+            }
             if (!type.hasAnnotation(VALIDATION_VALIDATED)) {
                 throw new CodegenException(VALIDATION_VALIDATED.fqName()
                                                    + " annotation is required on non-service type that has constraints "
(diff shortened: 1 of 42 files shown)


service/codegen/src/main/java/io/helidon/service/codegen/ServiceRegistryCodegenExtension.java
```diff
@@ -216,46 +220,266 @@ private RegistryRoundContext createRoundContext(RoundContext roundContext,
...
+            }
+        }
+    }
+
+    private void addEffectiveMethodElements(List<TypedElementInfo> elements,
+                                            Set<ElementSignature> signatures,
+                                            List<TypedElements.ElementMeta> elementMetas) {
+        for (TypedElements.ElementMeta elementMeta : elementMetas) {
+            TypedElementInfo element = elementMeta.effectiveElement();
+            if (element.kind() == ElementKind.METHOD && signatures.add(element.signature())) {
```
danielkec: This can skip constrained provided-contract methods when their signature collides with a factory SPI method, for example `Supplier.get()` vs `ProvidedApi.get()`, so validation is not generated. Consider keeping provider/provided signatures separate, or merging by declaring contract/return type instead of `ElementSignature` alone.

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -2391,12 +2403,424 @@ private void generateDelegationService(RegistryRoundContext roundContext,
...
+                .decreaseContentPadding()
+                .decreaseContentPadding()
+                .addContent("} : delegate");
+    }
+
+    private void injectionPointFactoryFirstBody(ContentBuilder<?> content,
+                                                DescribedService service,
+                                                Optional<InterceptedTypeGenerator.MethodDefinition> firstMethod,
+                                                Optional<InterceptedTypeGenerator.MethodDefinition> listMethod) {
+        if (listMethod.isPresent()) {
```
danielkec: If a provider customizes both `first(...)` and `list(...)`, this generated `first(...)` no longer calls the provider-specific `first(...)`, so selection semantics can change. Prefer invoking the intercepted `first(...)` when it exists, and only fall back to `list(...).getFirst()` when `first(...)` is inherited/default.

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -2391,12 +2403,424 @@ private void generateDelegationService(RegistryRoundContext roundContext,
...
+                .decreaseContentPadding()
+                .decreaseContentPadding()
+                .addContent("} : delegate");
+    }
+
+    private void qualifiedFactoryFirstBody(ContentBuilder<?> content,
+                                           DescribedService service,
+                                           Optional<InterceptedTypeGenerator.MethodDefinition> firstMethod,
+                                           Optional<InterceptedTypeGenerator.MethodDefinition> listMethod) {
+        if (listMethod.isPresent()) {
```
danielkec: If a provider customizes both qualified `first(...)` and `list(...)`, this generated `first(...)` no longer calls the provider-specific `first(...)`, so qualifier selection semantics can change. Prefer invoking the intercepted `first(...)` when it exists, and only fall back to `list(...).getFirst()` when `first(...)` is inherited/default.

service/codegen/src/main/java/io/helidon/service/codegen/ServiceTypes.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /**
+     * Whether the type is a Helidon service type.
+     *
+     * @param type type to check
+     * @return whether the type is a service
+     */
+    static boolean isService(TypeInfo type) {
+        if (type.hasAnnotation(ServiceCodegenTypes.SERVICE_ANNOTATION_PROVIDER)) {
```
danielkec, line 49: `@Service.Describe` is also consumed by `ServiceExtension`, but this predicate does not treat it as a service trigger. Described services with constraints only on contracts can be skipped; include `SERVICE_ANNOTATION_DESCRIBE` here before scanning contract annotations.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
```diff
@@ -74,8 +77,16 @@ void process(TypeInfo type) {
             // ignore annotations, these are good
             return;
         }
+        if (factoryInterceptionWrapper(type)) {
+            // The original factory provider owns the element interceptors used by the generated wrapper.
+            return;
+        }
 
         if (!isService(type)) {
+            if (hasOnlyNonPrivateInterfaceMethodValidation(type)) {
```
danielkec, line 86: This now silently accepts any constrained non-service interface, even if it is not a `@Service.Contract` and no service implementation will be processed. Consider restricting this bypass to known service contracts, or keep the previous `@Validated` error for ordinary interfaces.

service/codegen/src/main/java/io/helidon/service/codegen/ServiceDescriptorCodegen.java
```diff
@@ -214,17 +219,17 @@ public void service(TypeName generator, RegistryRoundContext roundCtx, Collectio
...
                 contracts = service.providedDescriptor().contracts();
-                factoryContracts = service.serviceDescriptor().contracts();
+                factoryContracts = service.providerFactoryContracts();
             } else {
                 // check if contracts are intercepted
                 var providedElements = service.providedDescriptor().elements();
 
                 if (providedElements.methodsIntercepted()) {
                     // remove contracts from the original service, service descriptor will only be used to instantiate provider
                     contracts = Set.of();
```
danielkec, line 231: This hides the raw factory provider from provided-contract and broad lookups, but concrete provider lookup still returns the raw descriptor by service type, so it can bypass the generated wrapper. With this PR's supplier validation shape, `registry.get(SupplierProvidedServiceProvider.class).get().get()` returns the invalid provided service without `ValidationException`, while `registry.get(SupplierProvidedService.class).get()` is validated. That leaves a provider-class lookup path outside the factory interception guarantee from #11872.

danielkec: out of scope

danielkec (approved):

PR #11885, opened 2026-05-06
Fix TLS reload without initial managers (4.x)

### LINKED ISSUE #11884: 4.x: ConfiguredTlsManager can fail TLS reload after starting without managers
## Problem

`ConfiguredTlsManager.reload(...)` delegates reload requests to reloadable key and trust managers, but those reloadable managers are not initialized when the initial TLS configuration starts without X509 key or trust managers. A later reload that supplies a key or trust manager can fail with `NullPointerException` instead of the existing controlled not-reloadable exception.

## Expected behavior

Reload after startup without the corresponding manager should fail with a controlled not-reloadable exception and leave the current TLS state unchanged.

## Source

Found during the common source audit.

## Related issue

Main-line tracking issue: #11883

## Validation

Focused regression tests on `helidon-4.x` first reproduced `NullPointerException` on reload after managerless startup. The same tests pass after installing not-reloadable manager sentinels, preflighting reloads before mutation, and preserving TLS manager state on failed reinitialization.

tomas-langer:
Resolves #11884

Fixes reload handling for configured TLS managers that started without key or trust managers. The reload path now uses controlled not-reloadable sentinels, validates both sides before mutating either manager, and keeps TLS manager state publication consistent across reload and failed reinitialization.

### DIFF
--- a/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2024 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


danielkec (approved):
Reviewed locally. No blocking findings.

PR #11907, opened 2026-05-07
Fix ListContext unregister identity handling (4.x)

### LINKED ISSUE #11904: 4.x: ListContext unregister removes replacement instance
## Problem

`Context.unregister(instance)` is implemented by `ListContext` as removal by runtime class. If one instance is registered and then another instance of the same class replaces it, unregistering the old instance removes the currently registered replacement.

## Reproducer

A focused regression on `helidon-4.x` follows this shape:

1. Create a context.
2. Register `first`.
3. Register `second` of the same class.
4. Call `unregister(first)`.
5. Read `get(TestValue.class)`.

Expected result: `Optional.of(second)`.

Actual result: `Optional.empty`.

Affected source:

- `common/context/context/src/main/java/io/helidon/common/context/ListContext.java`

## Proposed Fix

Keep registration replacement by type, but make `unregister(instance)` remove only a stored instance registration whose contained object is the same instance by identity. The fix should cover classified and unclassified registries without changing public API shape.

tomas-langer:
Resolves #11904

Updates `Context.unregister(...)` handling so `ListContext` removes registrations by stored instance identity instead of runtime type. Loaded supplied values can be removed by identity without forcing unloaded suppliers to resolve, and the public unregister defaults now reject `null` inputs consistently.

### DIFF
--- a/common/context/context/src/main/java/io/helidon/common/context/Context.java
+++ b/common/context/context/src/main/java/io/helidon/common/context/Context.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2019, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


danielkec (approved):
Reviewed; no findings.

PR #11910, opened 2026-05-07
Preserve malformed percent sequences in URI decoding (4.x)

### LINKED ISSUE #11908: 4.x: Preserve malformed percent sequences in URI query decoding
## Description

The non-validating URI query parser uses the shared percent decoder for decoded query names and values. That decoder should consume only complete percent escapes with two hexadecimal characters. Other percent sequences should remain literal text in the decoded query output.

## Expected behavior

- Valid percent escapes continue to decode normally.
- Non-validating query parsing preserves malformed percent sequences literally in decoded names and values.
- Validating query parsing continues to reject malformed percent encoding.

## Scope

This affects `common/uri` on `helidon-4.x`.

The fix should keep the public API and SPI shape unchanged.

tomas-langer:
Resolves #11908

Updates URI percent decoding so incomplete or malformed percent sequences remain literal while valid percent escapes continue to decode normally. Adds regression coverage for URI and query decoding, and documents the non-validating behavior with pointers to strict query validation.

### DIFF
--- a/common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
@@ -172,8 +172,18 @@ public final class UriEncoding {
                 continue;
             }
             if (!isPercentEncoded(string, i, len)) {
-                sb.append(c);
-                if (++i >= len) {
+                int malformedEnd = Math.min(i + 3, len);
+                while (i < malformedEnd) {
+                    c = string.charAt(i);
+                    if (c == '[') {
+                        betweenBrackets = true;
+                    } else if (betweenBrackets && c == ']') {
+                        betweenBrackets = false;
+                    }
+                    sb.append(c);
+                    i++;
+                }
+                if (i >= len) {
                     break;
                 }
                 c = string.charAt(i);
--- a/common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,6 +46,7 @@ public final class UriEncoding {
      * <p>
      * Percent characters {@code "%s"} found between brackets {@code "[]"} are not decoded to support IPv6 literal.
      * E.g. {@code http://[fe80::1%lo0]:8080}.
+     * Malformed percent-encoded sequences are preserved literally.
      * <p>
      * See <a href="https://tools.ietf.org/html/rfc6874#section-2">RFC 6874, section 2.</a>
      *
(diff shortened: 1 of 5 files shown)


common/uri/src/test/java/io/helidon/common/uri/UriEncodingTest.java
```diff
@@ -35,4 +36,22 @@ void testSpaceDecoding() {
     void testIPv6Literal() {
         assertThat(decodeUri("http://[fe80::1%lo0]:8080"), is("http://[fe80::1%lo0]:8080"));
     }
+
+    @Test
+    void malformedHexEscapesRemainLiteral() {
+        assertAll(
```
romain-grecourt, line 42: What is the value of `assertAll` compared to individual `assertThat` statements ?

tomas-langer: you get more information in a single run, no other

danielkec (changes requested):
I found one issue that leaves part of #11908 unresolved.

The malformed escape branch preserves only the `%` and then resumes normal decoding at the next character. In query decoding, normal decoding maps `+` to a space, so malformed percent sequences containing `+` are not preserved literally. For example, `UriQuery.create("value=%+1").get("value")` decodes as `% 1` instead of `%+1`.

That conflicts with the issue's requirement that non-validating query parsing preserve malformed percent sequences literally in decoded names and values. Please adjust the malformed-escape handling to keep the whole malformed percent sequence literal, and add coverage for cases such as `%+1` in both `UriQuery.create` and `UriQueryWriteable.fromQueryString`.

common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
```diff
@@ -169,17 +171,24 @@ private static String decode(String string, boolean ignorePercentInBrackets) {
                 c = string.charAt(i);
                 continue;
             }
-            bb.clear();
-            while (true) {
-                bb.put(decode(string.charAt(++i), string.charAt(++i)));
+            if (!isPercentEncoded(string, i, len)) {
+                sb.append(c);
```
danielkec: This preserves only the `%` and then resumes normal query decoding at the next character. In query decoding, `+` is converted to a space, so a malformed sequence such as `%+1` is decoded as `% 1` instead of remaining literal as `%+1`. That leaves part of #11908 unresolved for malformed percent sequences in decoded query names and values.

Please preserve the whole malformed percent sequence literally and add coverage for cases like `%+1` through both `UriQuery.create` and `UriQueryWriteable.fromQueryString`.

common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
```diff
@@ -169,17 +171,34 @@ private static String decode(String string, boolean ignorePercentInBrackets) {
                 c = string.charAt(i);
                 continue;
             }
-            bb.clear();
-            while (true) {
-                bb.put(decode(string.charAt(++i), string.charAt(++i)));
-                if (++i >= len) {
+            if (!isPercentEncoded(string, i, len)) {
+                int malformedEnd = Math.min(i + 3, len);
```
danielkec: This can corrupt decoding when a malformed escape is immediately followed by a valid one. For example, `%G%41` currently decodes as `%G%41`, leaving the valid `%41` undecoded instead of preserving `%G` and decoding `%41` to `A`.

tomas-langer: Fixed and added regression test.

danielkec (approved):
Looks good. Previous review concerns are addressed in the current head.

PR #11930, opened 2026-05-13
Fix WebServer TLS over Unix domain sockets [27.x]

### LINKED ISSUE #11844: Fix WebServer TLS over Unix domain sockets
## Summary

WebServer TLS setup fails for Unix-domain socket listeners.

The NIO TLS path calls `SocketChannel.socket().getRemoteSocketAddress()` before creating the server `SSLEngine`. That works for Internet socket channels, but `SocketChannel.socket()` is unsupported for Unix-domain socket channels. A TLS-enabled Unix-domain listener can therefore accept a connection but fail before protocol handling starts.

## Evidence

On `origin/main`:

- `webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java:283-291`

  ```java
  private HelidonSocket createNioSocket(Tls tls, SocketChannel channel, String channelId) {
      if (tls.enabled()) {
          var address = channel.socket().getRemoteSocketAddress();

          SSLEngine engine;
          if (address instanceof InetSocketAddress isa) {
              engine = tls.sslContext().createSSLEngine(isa.getHostName(), isa.getPort());
          } else {
              engine = tls.sslContext().createSSLEngine();
          }
  ```

- `webserver/tests/unix-domain-socket` currently has only the plain Unix-domain socket test and no TLS UDS regression coverage or TLS test resources.

## Expected behavior

TLS-enabled Unix-domain socket listeners should accept and serve HTTPS requests over UDS. The server-side TLS path should only use host/port peer information when the channel remote address is an `InetSocketAddress`; Unix-domain channels should use an address-less server `SSLEngine`.

## Suggested fix

Use `SocketChannel.getRemoteAddress()` in the NIO TLS path and create the host/port `SSLEngine` only for `InetSocketAddress`. For other socket address types, create the engine without peer host/port.

Add focused coverage under `webserver/tests/unix-domain-socket` with TLS cert/key resources. At minimum, cover a TLS-enabled Unix-domain WebServer serving a request over UDS. Prefer also covering HTTP/2 ALPN and endpoint-identification behavior if the test support is practical.

## Notes

This was found while splitting issue #11811. The binding-abstraction work already included a local fix and UDS TLS tests; public issue text should use the source evidence above rather than unpublished local commit references.

tomas-langer:
Resolves #11844

Carries the WebServer TLS-over-Unix-domain-socket fix to `main`, including logical-authority preservation for WebClient UDS requests, HTTP/1 and HTTP/2 cache separation, TLS handshake timeout handling, stale TLS/UDS connection handling, focused UDS TLS coverage, and WebServer/WebClient documentation updates.


danielkec (approved):
Reviewed exact head dec84c5c14d5287fd1e67a2cd1fe5cf164aefb23; no findings.

Verdent (approved):

PR #11931, opened 2026-05-13
Apply declarative validation constraints from service interfaces (4.x)

### LINKED ISSUE #11865: 4.x: Apply declarative validation constraints from service interfaces
Declarative validation should apply constraints declared on non-private service interface methods when a service implementation is obtained from the service registry.

Currently this is not handled reliably when the constraint exists only on the interface method or parameter. This means generated validation interceptors can be missing unless the implementation repeats the constraint. Default interface methods and services discovered only through `@Service.Inject` should also be covered, because service codegen treats those implementations as registry services.

Expected behavior:
- non-private, non-static interface method and parameter constraints are applied to service implementations;
- default interface methods are validated when invoked through the service registry;
- duplicate interceptors are not generated when the implementation repeats an interface constraint;
- static interface methods remain rejected unless the interface is explicitly `@Validation.Validated`;
- inject-only service implementations are processed when constraints are declared only on their service contract.

The fix should include both codegen and functional service-registry tests proving the interceptor is generated and invoked.

tomas-langer:
Resolves #11865

Applies declarative validation constraints declared on service interfaces to registry-managed services, including factory-provided services. Also extends generated validation handling for nested type-use constraints and documents the supported service-interface behavior.

Production source files changed:

- `declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java`
- `declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidatedTypeGenerator.java`
- `declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationExtension.java`
- `declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java`
- `validation/validation/src/main/java/module-info.java`

Also changed non-test docs/snippet files:
- `declarative/README.md`
- `docs/src/main/asciidoc/se/injection/declarative.adoc`
- `docs/src/main/asciidoc/se/injection/injection.adoc`
- `docs/src/main/java/io/helidon/docs/se/inject/DeclarativeExample.java`

### DIFF
--- a/declarative/README.md
+++ b/declarative/README.md
@@ -65,7 +65,7 @@ For each Helidon feature, we need a namespace class to contain the annotations a
 | LRA              | `LRA`                                | `Lra` cannot be freed                                     |
 | Transactions     | `Tx`                                 | `Transaction` is the interface                            |
 | Service Registry | `Service`, `Interception`            | OK                                                        |
-| Validation       | `Validation`, `Check`                | OK                                                        |
+| Validation       | `Validation`                         | OK                                                        |
 
 ## Integrations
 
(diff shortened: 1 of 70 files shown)


spericas (approved):
LGTM

Verdent (approved):

PR #11935, opened 2026-05-13
Fix UDS client proxy behavior (27.x)

### LINKED ISSUE #11933: UDS client request can use configured proxy behavior
## Summary

HTTP/1 WebClient requests that use a Unix-domain socket transport can still apply configured TCP proxy behavior while writing the request.

For UDS requests, the Unix-domain socket path is the physical transport endpoint. The configured URI remains the logical HTTP authority, but the request should not be sent through an HTTP/TCP proxy and should not be formatted as a proxied TCP request.

## Evidence

PR #11930 fixed several UDS client behaviors on `main`, including logical-authority preservation, connection/cache identity, TLS handshake timeout handling, TLS failure cleanup, and WebClient UDS documentation.

The remaining proxy behavior is in the HTTP/1 request path:

- `webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java` stores the original request proxy and uses it when deciding whether the request target should be relative or absolute-form.
- `webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java` adds the `Proxy-Connection` header whenever the request proxy type is not `NONE`.

The connection path already detects `UnixDomainSocketAddress` and obtains a UDS connection, so a request can combine UDS transport with proxy-shaped HTTP/1 framing.

## Expected behavior

When the physical request address is a `UnixDomainSocketAddress`:

- the client should connect directly to the Unix-domain socket path;
- the effective proxy for request framing should be `Proxy.noProxy()`;
- HTTP/1 should use the normal relative request target, not proxy absolute-form;
- `Proxy-Connection` should not be added for the UDS request;
- configured proxy behavior should remain unchanged for normal TCP requests.

## Suggested test coverage

Add a focused WebClient HTTP/1 regression that configures a proxy and then sends a request using `address(UnixDomainSocketAddress)`. The test should prove the request reaches the UDS server directly and that the server observes normal non-proxy HTTP/1 framing.

This is a follow-up item from UDS epic #11853.

tomas-langer:
Resolves #11933

Treat Unix-domain-socket WebClient HTTP/1 requests as direct transport for request framing, even when an HTTP proxy is configured. UDS requests now use relative-form request targets and omit `Proxy-Connection`, while TCP proxy behavior is unchanged.

Adds a real UDS capture-server regression covering the direct HTTP/1 request path.

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -94,7 +94,7 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
         this.timeout = clientRequest.readTimeout();
         this.connection = clientRequest.connection().orElse(null);
         this.tls = clientRequest.tls();
-        this.proxy = clientRequest.proxy();
+        this.proxy = clientRequest.effectiveProxy();
         this.keepAlive = clientRequest.keepAlive();
         this.http1Client = clientRequest.http1Client();
         this.whenComplete = whenComplete;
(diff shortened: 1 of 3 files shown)


spericas (approved):
LGTM

PR #11941, opened 2026-05-13
Preserve UDS address across WebClient redirects (27.x)

### LINKED ISSUE #11934: Preserve UDS transport address across same-origin WebClient redirects
## Summary

WebClient requests that use a Unix-domain socket transport can lose the physical socket address when following same-origin redirects.

For UDS requests, the Unix-domain socket path is the physical transport endpoint. The configured URI remains the logical HTTP authority. A same-origin redirect should keep using the same physical UDS address while updating only the redirected logical URI.

## Evidence

The redirect code creates new request instances from the previous request, but does not propagate the optional physical socket address:

- `webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java` copies redirect state such as follow-redirect settings, TLS, and output-stream redirect state, but not `request.address()`.
- `webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientRequestImpl.java` has the same redirect-copy shape and does not copy the physical address.
- `webclient/http2/src/main/java/io/helidon/webclient/http2/RedirectionProcessor.java` and `webclient/http2/src/main/java/io/helidon/webclient/http2/Http2CallOutputStreamChain.java` construct redirected requests from the previous request, so the missing address propagation affects HTTP/2 redirect paths as well.

This is separate from the logical-authority fix completed by #11930: the original request can preserve the URI authority and carry a UDS transport address, but the redirected request can lose that transport address.

## Expected behavior

When a WebClient request has a `UnixDomainSocketAddress` physical transport address:

- same-origin redirects should continue to use that UDS address;
- the redirected URI should still determine the logical HTTP authority, path, query, cookies, and TLS peer identity;
- cross-origin redirect protections and sensitive-header filtering should remain unchanged;
- normal TCP redirects should be unaffected.

## Suggested test coverage

Add focused HTTP/1 and HTTP/2 redirect regressions using a UDS server:

- a same-origin redirect should complete over the original UDS transport;
- the redirected request should preserve the logical authority from the redirected URI;
- cross-origin redirect behavior should still apply the existing guardrails.

This is a follow-up item from UDS epic #11853.

tomas-langer:
Resolves #11934

Preserves Unix-domain-socket transport addresses for same-origin WebClient redirects in HTTP/1 and HTTP/2, while dropping UDS transport overrides when redirects cross to a different logical origin.

Adds UDS redirect regressions for per-request addresses and configured base addresses.

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
@@ -149,6 +149,8 @@ public abstract class ClientRequestBase<T extends ClientRequest<T>, R extends Ht
 
         // this must be after we set clientUri, as it is used from the method
         clientConfig.baseAddress()
+                .filter(it -> !(it instanceof UnixDomainSocketAddress)
+                        || this.redirectSourceUri == null)
                 .ifPresent(this::address);
     }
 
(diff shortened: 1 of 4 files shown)


spericas (approved):
LGTM

PR #11944, opened 2026-05-13
Document max-tcp-connections admission semantics (4.x)

### LINKED ISSUE #11942: Document max-tcp-connections pre-accept admission semantics (4.x)
The WebServer `max-tcp-connections` documentation and Javadocs describe the setting mainly as a concurrent TCP connection limit. The listener implementation reserves admission capacity before accepting a connection.

Update only documentation and Javadocs to clarify:

- the limit is a pre-accept connection admission capacity;
- Helidon reserves a permit before accepting a connection;
- the permit is released after the accepted connection finishes, or immediately if the accepted socket cannot be configured or submitted to the connection executor.

Do not change runtime behavior.

tomas-langer:
Resolves #11942

Clarifies that WebServer `max-tcp-connections` is a pre-accept admission capacity and aligns the performance tuning snippet with the listener Javadocs.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2023, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
```diff
@@ -260,10 +260,12 @@ interface ListenerConfigBlueprint {
     SocketOptions connectionOptions();
 
     /**
-     * Limits the number of connections that can be opened at a single point in time.
+     * Limits the number of connection permits that this listener may reserve before accepting sockets.
+     * A permit is acquired before each accept attempt and released after the accepted connection finishes,
```
danielkec, line 264: The release guarantee is broader than current code. `ServerListener` acquires the permit before `accept()`, but `ConnectionHandler.run()` returns without releasing it if proxy-protocol parsing or TLS/socket setup fails before it reaches the later `finally` that releases the permit. With `max-tcp-connections=1`, one malformed PROXY/TLS connection can consume the permit and leave subsequent connections waiting, so the docs would claim behavior the implementation does not provide.

tomas-langer: This is a documentation fix. The intended behavior is as documented. 
You are reporting a bug in code as a review comment for documentation...

spericas (approved):

PR #11947, opened 2026-05-13
Propagate WebServer lifecycle failures

### LINKED ISSUE #11936: WebServer Start and stop failure propagation
Part of #11854.

## Scope

WebServer start and stop should propagate lifecycle failures instead of hiding them or aborting cleanup early.

Current behavior:

- `webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java` can log listener startup failure, stop listeners, set `running` to `false`, and return from `WebServer.start()` without throwing.
- `LoomServer.stop()` loops over listeners directly, so a failure from one listener can skip later listener cleanup and leave final server state handling inconsistent.

## Expected behavior

- Collect listener startup failures.
- During failed start, still attempt cleanup for every listener that needs cleanup.
- Make `WebServer.start()` throw an `IllegalStateException` or the original runtime failure with cleanup failures suppressed.
- During stop, attempt every listener cleanup even if one listener fails.
- Set the server running state consistently before propagating stop failure.
- Keep shutdown-hook deregistration best-effort and include any deregistration failure in the final propagated failure when applicable.

## Implementation notes

- A small package-private lifecycle failure aggregation helper is acceptable if it keeps start/stop and later suspend/resume handling consistent.
- Keep this PR limited to WebServer lifecycle failure propagation; do not introduce the transport-binding SPI or other broader transport changes.

## Tests

Add focused WebServer lifecycle tests proving:

- A listener start failure is visible to the `WebServer.start()` caller.
- `isRunning()` is `false` after failed start.
- Cleanup still runs for already-started listeners after failed start.
- Stop failures are propagated only after all listeners have been asked to stop.
- Suppressed failures are preserved where multiple cleanup failures occur.

tomas-langer:
Resolves #11936

Propagates WebServer startup and shutdown lifecycle failures after attempting cleanup across all listeners.

### DIFF
--- a/docs/src/main/asciidoc/se/webserver/webserver.adoc
+++ b/docs/src/main/asciidoc/se/webserver/webserver.adoc
@@ -314,7 +314,9 @@ In this example, the `GET` handler matches requests to `/hello/subpath`.
 
 === Server Lifecycle
 
-In Helidon 4 your `HttpService` can interpose on the server lifecycle by overriding the `beforeStart` and `afterStop` methods:
+Your `HttpService` can interpose on the server lifecycle by overriding the `beforeStart`, `afterStart`, and `afterStop` methods.
+Failures from `beforeStart` or `afterStart` fail `WebServer.start()` after startup cleanup, and failures from
+`afterStop` fail `WebServer.stop()` after listener cleanup. Cleanup failures may be attached as suppressed exceptions.
 [source,java]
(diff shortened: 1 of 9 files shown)


danielkec (approved):
Reviewed at cce760580d87c8ed0bd52cfb8231705911013ae7. No findings.

PR #11950, opened 2026-05-14
Fix WebServer suspend and resume failure rollback

### LINKED ISSUE #11937: WebServer Suspend and resume failure rollback
Part of #11854.

## Scope

WebServer suspend and resume should not leave the server partially suspended or partially restored while still reported as running.

Current behavior:

- `webserver/webserver/src/main/java/io/helidon/webserver/LoomServer.java` iterates listener `suspend()` and `resume()` directly.
- If one listener fails during CRaC suspend or restore, earlier listeners can already have changed state while later listeners are untouched.
- The server can remain in an inconsistent partially stopped or partially restored state.

## Expected behavior

- On suspend failure, transition the WebServer to stopped by attempting cleanup across all listeners.
- On resume failure, transition the WebServer to stopped by attempting cleanup across all listeners.
- Set `running` to `false` after this failure cleanup.
- Rethrow the original suspend or resume failure, with cleanup failures suppressed.

## Implementation notes

- Build on the start/stop lifecycle failure handling from #11936 where possible.
- Keep the change focused on suspend/resume rollback behavior and avoid unrelated transport-binding changes.

## Tests

Add focused lifecycle tests proving:

- A suspend failure is propagated to the caller.
- A resume failure is propagated to the caller.
- The server reports `isRunning() == false` after either failure.
- Cleanup is attempted for all listeners, not only the listener that failed.
- Cleanup failures are preserved as suppressed exceptions when applicable.

tomas-langer:
Resolves #11937

Stops the WebServer after suspend or resume listener failures, preserves the original lifecycle failure, and keeps cleanup failures suppressed on it. Also keeps Unix-domain socket cleanup limited to listener-owned bound sockets during rollback.

### DIFF
--- a/docs/src/main/asciidoc/se/webserver/webserver.adoc
+++ b/docs/src/main/asciidoc/se/webserver/webserver.adoc
@@ -317,6 +317,8 @@ In this example, the `GET` handler matches requests to `/hello/subpath`.
 Your `HttpService` can interpose on the server lifecycle by overriding the `beforeStart`, `afterStart`, and `afterStop` methods.
 Failures from `beforeStart` or `afterStart` fail `WebServer.start()` after startup cleanup, and failures from
 `afterStop` fail `WebServer.stop()` after listener cleanup. Cleanup failures may be attached as suppressed exceptions.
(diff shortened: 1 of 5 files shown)


danielkec (approved):
Reviewed; no findings.

PR #11951, opened 2026-05-14
Preserve response Connection close header (3.x)

### LINKED ISSUE #11916: 3.x: MP: setting "Connection: close" on server Response does not return header to client
## Environment Details
* Helidon Version: 3.2.17
* Helidon MP
* JDK version: 17

----------

## Problem Description

If you do the following in a JAX-RS resource:

``` java
        return Response.ok(body).header("Connection", "close").header("X-My_Header", "foo").build();
```

The client will see:
```
X-My_Header: foo
```

But no `Connection: close`.

With Helidon 4.x (I tried 4.4.1 and 4.0.0) this works as expected and the client receives the "Connection: close" header.

## Steps to reproduce

Build and run the attached project. Then run:

``` bash
curl -v http://localhost:8080/simple-greet
```
and you'll see

```
* Host localhost:8080 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8080...
* Connected to localhost (::1) port 8080
> GET /simple-greet HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/8.7.1
> Accept: */*
>
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Type: text/plain
< Date: Mon, 11 May 2026 12:24:36 -0700
< X-My_Header: foo
< content-length: 12
<
* Connection #0 to host localhost left intact
Hello World!%
```

Note that the custom header "X-My_Header:" is there but "Connection: close" is not.

If you update this example to use 4.4.1 then it works as expected and you see "Connection: close".

tomas-langer:
Resolves #11916

Preserves explicit response-side `Connection: close` headers while still closing the connection after the response is written. Adds wire-level regression coverage for WebServer and MP/Jersey responses.

### DIFF
--- a/webserver/jersey/src/test/java/io/helidon/webserver/jersey/JerseyExampleResource.java
+++ b/webserver/jersey/src/test/java/io/helidon/webserver/jersey/JerseyExampleResource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2022 Oracle and/or its affiliates.
+ * Copyright (c) 2017, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


barchetta (approved):
LGTM

PR #11957, opened 2026-05-17
WebServer Accepted socket shutdown race

### LINKED ISSUE #11938: WebServer Accepted socket shutdown race
Part of #11854.

## Scope

Accepted sockets should not escape shutdown while their `ConnectionHandler` is still between accept and active connection registration.

Current behavior:

- `webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java` accepts a `SocketChannel` and submits a new `ConnectionHandler`.
- `webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java` registers an active `ServerConnection` only after socket, TLS, proxy protocol, and protocol selection setup.
- `ServerListener.suspend(...)` and shutdown paths close registered active connections, but a socket in the setup window can survive shutdown and then start request handling.

## Expected behavior

- Track accepted handlers, not only registered `ServerConnection` instances.
- Allow `ConnectionHandler.close(...)` or equivalent listener cleanup to record a close request before `ServerConnection` exists.
- Close the underlying Helidon socket or raw channel during setup when shutdown has requested closure.
- Check the close request immediately before active connection registration and before `connection.handle(...)`.
- Preserve current mainline socket/TLS behavior while adding the lifecycle race fix.

## Implementation notes

- Adapt the fix directly to current `ServerListener` and `ConnectionHandler`.
- Do not pull in the transport-binding SPI, TCP/Unix binding providers, or binding-specific port/config changes.

## Tests

Add focused tests proving:

- A graceful shutdown closes a socket paused between accept and connection registration.
- A forced shutdown closes or interrupts a handler paused in that same setup window.
- The paused handler does not reach `connection.handle(...)` after shutdown has requested closure.

tomas-langer:
Resolves #11938

Fixes shutdown handling for sockets accepted by WebServer before protocol connection registration, routing listener lifecycle and idle-timeout handling through the connection handler.

### DIFF
--- a/webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
@@ -22,6 +22,7 @@ import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketException;
 import java.net.SocketOption;
+import java.net.SocketTimeoutException;
 import java.net.UnixDomainSocketAddress;
 import java.nio.channels.SocketChannel;
 import java.nio.file.Files;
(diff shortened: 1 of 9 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
```diff
@@ -442,79 +516,134 @@ void interruptedReaderExecutorShutdownFailsStopAndForcesShutdown() throws Except
...
+            suspendThread.join(TimeUnit.SECONDS.toMillis(5));
 
-            assertThat(containsMessage(failure, "before await failed"), is(true));
-            assertThat(sharedTaskInterrupted.await(5, TimeUnit.SECONDS), is(true));
-            assertThat(server.isRunning(), is(false));
-            assertThat(service.afterStops(), is(1));
+            assertThat(suspendThread.isAlive(), is(false));
+            assertThat(suspendFailure.get(), nullValue());
+            assertThat(selector.connection().gracefulCloses(), is(0));
+            assertThat(selector.connection().forcedCloses(), is(1));
```
danielkec, line 582: CI is already hitting this race in `tests/core-linux`: the assertion expected `forcedCloses()` to be 1 but saw 0. The test only waits for `server.suspend()` to return, while suspend waits for the listener close future, not for the reader task to resume from `selector.release()` and delegate the pending forced close to the selected connection.

tomas-langer: Fixed in the follow-up commits.

The race was in the test ordering. The test released `selector` before `server.suspend()` had completed the forced-close phase, so the reader task could resume, create the delegate, close it gracefully, remove the handler, and leave the later forced-close snapshot with nothing to close.

The test now keeps the selector blocked until the suspend thread has joined and the suspend failure check has passed, then releases the selector and waits for the late-created delegate to observe `forcedCloses() == 1`. That makes the pending close request unambiguously forced when the reader resumes. The follow-up cleanup also changed `ConnectionHandler.closeChannel()` to use the existing `channelId` field directly.

Validate run 26003350287 completed successfully, including the Windows core job that exposed this race.

danielkec (approved):
Reviewed current head 1ea0f6fe183522e3d17bd7232475809f3d5cfdd6.

PR #11969, opened 2026-05-18
Allow telemetry to be disabled without startup exception

### LINKED ISSUE #11964: OpenTelemetry integration requires OTel SDK even when disabled (4.x)
## Environment Details
* Helidon Version: 4.4.1
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
Telemetry configuration allows the user to disable the feature.

But even upon doing so, Helidon still insists that an internal OTel SDK property be set.

## Steps to reproduce
Config:
```yaml
telemetry:
  enabled: false
```

At start-up:
```list
Property openTelemetrySdk must not be null
```

tjquinno:
### Description
Resolves #11964 

### Release Note
____
Helidon no longer throws an exception about a required `OpenTelemetrySdk` setting when `telemetry.enabled` is `false`.
____

The PR changes the `openTelemetrySdk` blueprint setting to an `Optional` but it has always had package-private access so there is no change to the public surface area.

The `openTelemetrySdk` setting is used only internally within the module and is not needed if the user has disabled telemetry.

A new test makes sure that disabling telemetry does not cause a failure wrt `openTelemetrySdk`.

### Documentation
No impact.

### DIFF
--- a/telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
+++ b/telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
@@ -116,6 +116,6 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
      * @return the SDK
      */
     @Option.Access("")
-    Optional<OpenTelemetrySdk> openTelemetrySdk();
+    OpenTelemetrySdk openTelemetrySdk();
 
 }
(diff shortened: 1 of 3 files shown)


telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -116,6 +116,6 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
      * @return the SDK
      */
     @Option.Access("")
-    OpenTelemetrySdk openTelemetrySdk();
+    Optional<OpenTelemetrySdk> openTelemetrySdk();
```
tomas-langer: This is a partially backward incompatible change - this method will be visible on the public type (`OpentTelemetryConfig`), and if anybody calls it, their code would break

tjquinno: Fixed.

tomas-langer (approved):

PR #11970, opened 2026-05-18
Apply container semantics to TypeHandlerOptional (4.x)

### LINKED ISSUE #11967: TypeHandlerOptional does not apply the same container semantics as TypeHandlerMap and friends
If you have a blueprint option declared as:
```
Optional<Map<String, String>> someOption();
```
...as you might find in [`UcpDataSourceConfigBlueprint`](https://github.com/helidon-io/helidon/blob/5e60b7ccb461be7ea677cf3c9488e4a7f7f3e4e1/data/sql/datasource/ucp/src/main/java/io/helidon/data/sql/datasource/ucp/UcpDataSourceConfigBlueprint.java#L154), the resulting configuration is not detached as it would be if the `Optional` part were missing.

This is due to `TypeHandlerOptional` not delegating container semantics to, e.g., `TypeHandlerMap` and related classes.

tomas-langer:
Resolves #11967

Adds optional container handling to builder codegen so `Optional<Map>`, `Optional<List>`, and `Optional<Set>` use the same config, setter/add, defensive-copy, and immutable built-value behavior as their non-optional container counterparts.

Also extends provider-backed optional collection discovery for service-loader and registry-backed providers, preserving absent versus explicit-empty container semantics.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/GenerateAbstractBuilder.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/GenerateAbstractBuilder.java
@@ -33,6 +33,7 @@ import java.util.stream.Collectors;
 import io.helidon.builder.codegen.spi.BuilderCodegenExtension;
 import io.helidon.codegen.classmodel.ClassModel;
 import io.helidon.codegen.classmodel.Constructor;
+import io.helidon.codegen.classmodel.ContentBuilder;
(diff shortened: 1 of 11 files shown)


ljnelson:
I've tested this against a UCP configuration and now I can set `connection-properties` and `connection-factory-properties` as expected. I've reviewed for obvious mistakes but am not an expert in the areas of the code involved. LGTM

ljnelson (approved):

PR #11974, opened 2026-05-18
Fix SSE comment-only events

### LINKED ISSUE #11834: SSE comment only / empty data events can not be created
## Environment Details
* Helidon Version: 4.4.1
* JDK version: 26
* OS: macOS 26

----------

## Problem Description

I'd like to send SSE events with no data and only a comment to act as a keep alive for an SSE stream.

This was addressed in https://github.com/helidon-io/helidon/issues/9165 but seems to have resurfaced as part of the refactoring that introduced `DataWriterSseSink`.

As far as I can see, the older `OutputStreamSseSink` had the following code in `emit`,

```java
Object data = sseEvent.data();
if (data != SseEvent.NO_DATA) {
  outputStream.write(SSE_DATA);
  eventConsumer.accept(data, sseEvent.mediaType().orElse(MediaTypes.TEXT_PLAIN));
  outputStream.write(SSE_NL);
}  
```

whereas `DataWriterSseSink` has the following in `emit`

```java
Object data = sseEvent.data();
if (data != null) {
  MediaType mediaType = sseEvent.mediaType().orElse(MediaTypes.TEXT_PLAIN);
  ...
  // write `data` object to the outputStream
}
```

## Steps to reproduce

Try to create an SSE event with just a comment. This will throw an exeption during the `serializeData` call within the `emit` method of the `DataWriterSseSink` as it tries to serialize the `NO_DATA` sentinel object.

## Fix

Change `DataWriterSseSink` to check for `data != SseEvent.NO_DATA` sentinel within the `emit` method

Thanks

arouel:
### Description

Fixes #11834.

SSE events without data were treated as having data because `DataWriterSseSink` checked only for `null`, while `SseEvent` uses the `SseEvent.NO_DATA` sentinel to represent absent data.

This caused comment-only SSE events, such as keep-alive events, to attempt serialization of the sentinel object instead of emitting only the comment.

The fix updates `DataWriterSseSink` to skip data serialization when `sseEvent.data()` returns `SseEvent.NO_DATA`.
A regression test was added for comment-only SSE events.

### Documentation
None

### DIFF
--- a/webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
+++ b/webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
@@ -121,7 +121,7 @@ class DataWriterSseSink implements SseSink {
                 outputStream.write(SSE_NL);
             }
             Object data = sseEvent.data();
-            if (data != null) {
+            if (data != SseEvent.NO_DATA) {
                 MediaType mediaType = sseEvent.mediaType().orElse(MediaTypes.TEXT_PLAIN);
 
                 // is it multi-line string data?
--- a/webserver/tests/sse/src/test/java/io/helidon/webserver/tests/sse/SseBaseTest.java
+++ b/webserver/tests/sse/src/test/java/io/helidon/webserver/tests/sse/SseBaseTest.java
@@ -137,6 +137,15 @@ class SseBaseTest {
         }
     }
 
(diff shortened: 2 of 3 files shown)


spericas (dismissed):

arouel:
@spericas I fixed the copyright year in the license header. Can you review again?

spericas (approved):

spericas (approved):

PR #11979, opened 2026-05-18
Fix WebServer idle timeout task lifecycle (27.x)

### LINKED ISSUE #11939: WebServer Idle timeout task lifecycle
Part of #11854.

## Scope

Idle timeout tasks should be tied to listener lifecycle and cancelled when the listener stops or suspends.

Current behavior:

- `webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java` creates an `IdleTimeoutHandler` as a local value and starts it.
- The listener does not retain the timeout task.
- Stop and suspend cleanup cannot cancel the idle timeout task explicitly.

## Expected behavior

- Retain the idle timeout handler or task as listener state.
- Start the task with listener startup.
- Cancel the task during listener stop and suspend cleanup.
- Make cleanup idempotent so repeated stop paths do not fail because the task was already cancelled.

## Implementation notes

- Keep the change limited to idle timeout lifecycle management.
- Avoid broader timer or transport refactoring unless it is required for a focused test seam.

## Tests

Add focused coverage proving:

- Listener stop cancels the idle timeout task.
- Listener suspend cancels the idle timeout task.
- Repeated stop or cleanup paths remain safe.
- The test should use a small explicit hook or test timer rather than relying on long real-time waits.

tomas-langer:
Resolves #11939

Moves the WebServer idle timeout task into the listener lifecycle so it starts with listener startup, is recreated after checkpoint resume, and is cancelled during stop, suspend, and failed-start cleanup. Adds focused coverage for stop, suspend, repeated cleanup, and resume scheduling without relying on real-time idle waits.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
@@ -318,9 +318,11 @@ class ServerListener implements ListenerContext {
         Throwable failure = null;
         // Stop listening for connections
         closeServerSocketForStop();
+        IdleTimeoutHandler cancelledIdleTimeoutHandler = null;
 
         try {
-            cancelIdleTimeoutHandler();
+            cancelledIdleTimeoutHandler = cancelIdleTimeoutHandler();
+            purgeCancelledIdleTimeoutHandler(cancelledIdleTimeoutHandler);
         } catch (RuntimeException | Error e) {
             failure = LifecycleFailures.add(failure, e);
         }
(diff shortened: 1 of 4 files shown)


romain-grecourt (dismissed):

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -316,6 +316,11 @@ private void initServerThread() {
 
     private Throwable stopResources() {
         Throwable failure = null;
+        try {
+            cancelIdleTimeoutHandler();
```
danielkec: This waits for the idle-timeout pass before the listener socket is closed. If the timer pass is blocked inside `ServerConnection.close(false)`, for example a custom connection whose graceful close waits on I/O, `stop()` or checkpoint `suspend()` can hang while the port is still bound and the accept thread is still active. The new wait test covers this blocking behavior, but not that the listener socket is closed before the wait begins.

tomas-langer: Fixed in `9f8d8f3640`. `stopResources()`, checkpoint suspend cleanup, and failed-start rollback now close/release the listener socket before waiting for idle-timeout cancellation. I also added stop and suspend regressions that block the idle-timeout close and verify the listener port rejects new connections while shutdown waits.

spericas (approved):
LGTM

PR #11988, opened 2026-05-19
Backward compatible SamplerType configuration

tomas-langer:
Adds a custom config factory so sampler-type accepts the legacy const spelling while the public SamplerType value remains CONSTANT. Also covers lowercase and uppercase constant/const configuration values in OpenTelemetry tracing config tests.


tjquinno (approved):
LGTM

PR #11992, opened 2026-05-19
4.x: Support disabling OIDC redirect attempt param 

### LINKED ISSUE #7754: Default redirect-attempt-param 'h_ra' is getting appended to url, there is no configuration to avoid appending this. param.
## Environment Details
* Helidon Version: 3.2.2
* Helidon MP
* JDK version: 17.0.3
* OS: Mac OS

----------

## Problem Description

We are using OIDC provider for SSO integration.
After successful authentication, the redirect URL is appended with param `?h_ra=1`
`https://pntiddaqy.gamma.spectra.us-phoenix-1.ocs.oc-test.com/api/ui-infrastructure/suis-index-ms/v1/applications/ui:2039e6cf-d57a-4ae4-8991-3ac7fc1a2566/redwood/rwdref?h_ra=1 `
We don't see the configuration options to not appending the param "h_ra", I think we can do one workaround by implementing Filter to remove this param.

@srusen, was expecting that configuration to be possible by setting the oidc.redirect-attempt-param as empty. We tried this option, still the default param is getting appended.

## Steps to reproduce
OidcSupport.increaseRedirectCounter(String state) method appending the 'h_ra' param.
Example to configure SSO : https://github.com/helidon-io/helidon/blob/3.2.2/examples/security/idcs-login/src/main/resources/application.yaml
My application Configuration:
```
 security:
  enabled: "${indexService.authenticate}"
  config.require-encryption: false
  properties:
    idcs-uri: "${idcs.identityServiceInstanceURL}"
    idcs-client-id: "${idcs.oauthClient.clientId}"
    idcs-client-secret: "${idcs.oauthClient.clientSecret}"
  provider-policy:
    type: "COMPOSITE"
    authentication:
      - name: "spectra-auth"
        flag: "MAY_FAIL"
      - name: "oidc-delegator"
        flag: "MAY_FAIL"
  providers:
    - spectra-auth:
        authentication:
          authenticate: ${indexService.authenticate}
    - abac:
        # OIDC needs an Authorization Provider that is why abac needs to be configured
        fail-if-none-validated: true
    - oidc:
        client-id: "${idcs.oauthClient.clientId}"
        client-secret: "${idcs.oauthClient.clientSecret}"
        identity-uri: "${idcs.identityServiceInstanceURL}"
        # support for non-public signature JWK (and maybe other IDCS specific handling)
        server-type: "idcs"
        logout-enabled: true
        redirect: true
        redirect-uri: "${indexService.path}/oidc/redirect"
        # this provider is only used for SSO
        header-use: false
        post-logout-uri: "${indexService.path}/oidc/logout"
        frontend-uri: "${environment.host}"
    - oidc-delegator:
        paths: ["/applications"]
```
This issue was discussed over slack: https://oracle-one.slack.com/archives/C9Y1RM1FG/p1696519056835549

danielkec:
Resolves #7754

### Description

Adds `redirect-attempt-counter-strategy` for OIDC redirect loop protection, with `PARAM` as the default to preserve existing behavior.

Supported strategies:

- `PARAM` appends and reads the existing `redirect-attempt-param` query parameter on the original redirect URI.
- `COOKIE` stores the redirect attempt counter in a small cookie instead of modifying the redirect URI. The cookie is
  scoped by tenant and original URI so stale counters from other flows are ignored.
- `NONE` disables redirect attempt counting.

The existing `redirect-attempt-param` setting still controls the query parameter name for `PARAM`, and the cookie name
prefix for `COOKIE`.

### Testing

- `mvn -pl security/providers/oidc-common,security/providers/oidc,!microprofile/cdi -am -Dtest=OidcConfigFromBuilderTest,OidcConfigFromConfigTest,OidcFeatureTest,TenantAuthenticationHandlerTest -Dsurefire.failIfNoSpecifiedTests=false test`


Verdent:
I dont think this is the right enhancement. That counter is there for a reason to keep track how many times did Helidon try to obtain valid information from the IDP. If we disable it and it is not saved, we are at  a high risk of infinite redirection loops.

Verdent:
I think, that better approach would be to use some very small cookie which would persist the current counter value and clearing it the moment we dont need it.

tomas-langer (approved):

tomas-langer (changes requested):
Sorry, missed David's comments

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/RedirectAttemptCookie.java
```diff
@@ -0,0 +1,66 @@
...
+                if (trimmed.startsWith(valuePrefix)) {
+                    return Optional.of(trimmed.substring(valuePrefix.length()));
+                }
+            }
+        }
+        return Optional.empty();
+    }
+
+    private static SetCookie.Builder builder(OidcConfig oidcConfig, String value) {
+        return SetCookie.builder(oidcConfig.redirectAttemptParam(), value)
```
danielkec: Using only `redirectAttemptParam()` as the cookie name makes this counter global for the browser under the shared path below. A stale failed flow for one tenant/resource can then make another flow hit `maxRedirects` early. Please scope the counter to the flow/tenant/resource, or include enough context in the cookie name/value to avoid cross-flow interference.

danielkec (commented):
I left two inline comments on the cookie-backed redirect attempt counter.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/RedirectAttemptCookie.java
```diff
@@ -0,0 +1,66 @@
...
+                .path("/")
```
danielkec: This cookie hardcodes the OIDC cookie policy instead of honoring configured attributes such as `cookie-secure`, `cookie-domain`, `cookie-path`, and `cookie-same-site`. Deployments that configure those settings will still emit the redirect-attempt cookie outside that policy; please apply the configured attributes to both set and removal cookies.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcFeature.java
```diff
@@ -538,8 +539,8 @@ private String processJsonResponse(ServerRequest req,
             }
         }
 
-        originalUri = increaseRedirectCounter(originalUri);
-        res.headers().add(HeaderNames.LOCATION, originalUri);
+        originalUri = updateRedirectCounter(req, headers, tenantName, originalUri);
```
danielkec, line 545: With `COOKIE` strategy and `query-param-use=true`, this uses `originalUri` after access/id-token query params are appended to derive the cookie name. On the next failed round trip another token param is appended, so the hash changes and the previous attempt cookie is not found; `max-redirects` can keep resetting. The cookie key should be based on the stable original URI before token propagation, or otherwise normalize these params.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/RedirectAttemptCookie.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private static SetCookie.Builder builder(OidcConfig oidcConfig, String name, String value, boolean create) {
+        SetCookie configuredCookie = create
+                ? oidcConfig.tokenCookieHandler().createCookie(value).build()
+                : oidcConfig.tokenCookieHandler().removeCookie().build();
+        SetCookie.Builder builder = SetCookie.builder(name, value);
+        configuredCookie.expires().ifPresent(builder::expires);
+        configuredCookie.maxAge().ifPresent(builder::maxAge);
+        configuredCookie.domain().ifPresent(builder::domain);
+        configuredCookie.path().ifPresent(builder::path);
```
danielkec, line 149: Copying the token cookie `Path` onto the redirect-attempt cookie can make the counter invisible on one side of the flow. For example, `cookie-path=/app` and the default callback `/oidc/redirect` means the protected request can send the cookie but the callback cannot, so the counter is recreated at 1. This counter cookie needs a scope that covers both the protected URI and redirect URI, or a strategy that does not depend on callback visibility.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -697,15 +703,18 @@ public boolean shouldRedirect() {
         return redirect;
```
danielkec, line 703: This now reads as if `redirect-attempt-param` is the exact cookie name. The cookie strategy actually uses it as a prefix and appends a per-tenant/original-URI hash, so the public config docs/Javadocs should describe it as a query parameter name or cookie-name prefix.

danielkec (commented):
Found the follow-up issues from the review.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcFeature.java
```diff
@@ -630,6 +629,24 @@ String increaseRedirectCounter(String state) {
...
+            return increaseRedirectCounter(state);
+        case COOKIE:
+            return state;
+        default:
+            throw new IllegalStateException("Unsupported redirect attempt counter strategy: "
+                                                    + oidcConfig.redirectAttemptCounterStrategy());
+        }
+    }
+
     private void processError(ServerRequest req, ServerResponse res) {
```
danielkec: When `redirect-attempt-counter-strategy=COOKIE`, this callback error path returns without expiring the scoped redirect-attempt cookie. The cookie is set before redirecting to the identity provider and is removed on successful authentication, but terminal callback failures such as an IdP error/missing code, state mismatch, or token-endpoint failure leave the counter behind. A few failed callbacks can make the next legitimate retry for the same tenant/original URI start near `max-redirects`; please clear the attempt cookie on terminal callback failures when the original URI/tenant are available, or keep that failure counter short-lived.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/RedirectAttemptCookie.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        return Optional.empty();
+    }
+
+    static String name(OidcConfig oidcConfig, String tenantId, String state) {
+        return oidcConfig.redirectAttemptParam() + "_" + hash(tenantId + "\n" + counterState(oidcConfig, state));
+    }
+
+    static String counterState(OidcConfig oidcConfig, String state) {
```
danielkec, line 76: Do we need to hand-parse query parameters here? Helidon already has `io.helidon.common.uri.UriQuery` / `UriQueryWriteable` for parsing and removing query params, which would avoid the split/indexOf logic and the custom encoded-name matching.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/RedirectAttemptCookie.java
```diff
@@ -0,0 +1,133 @@
...
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import io.helidon.common.uri.UriQueryWriteable;
```
danielkec: This class now imports `UriQueryWriteable`, but `helidon-security-providers-oidc` does not declare a direct `helidon-common-uri` dependency or `requires io.helidon.common.uri` in its module descriptor. It is currently readable transitively through other modules, but because this module uses the type directly, please add the explicit dependency/module requirement.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantAuthenticationHandler.java
```diff
@@ -516,6 +552,21 @@ private int redirectAttempt(String state) {
...
+                .map(this::parseRedirectAttemptCookie)
+                .orElse(0);
+    }
+
+    private int parseRedirectAttemptCookie(String value) {
+        try {
+            return Integer.parseInt(value);
+        } catch (NumberFormatException e) {
+            LOGGER.log(System.Logger.Level.DEBUG, "Invalid OIDC redirect attempt cookie value", e);
+            return 0;
```
danielkec: Malformed or overflowing COOKIE counter values fail open here. Returning `0` resets the guard, so a bad or injected competing cookie can keep redirect attempts below `maxRedirects`. Please fail closed, for example by treating invalid values as exhausted.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/RedirectAttemptCookie.java
```diff
@@ -0,0 +1,133 @@
...
+        int fragmentStart = state.indexOf('#', queryStart);
+        String path = state.substring(0, queryStart);
+        String query = fragmentStart < 0
+                ? state.substring(queryStart + 1)
+                : state.substring(queryStart + 1, fragmentStart);
+        String fragment = fragmentStart < 0 ? "" : state.substring(fragmentStart);
+
+        UriQueryWriteable queryParams = UriQueryWriteable.create();
+        queryParams.fromQueryString(query);
+        removeCounterQueryParam(queryParams, oidcConfig.redirectAttemptParam());
```
danielkec: For `COOKIE`, `redirectAttemptParam()` is documented as the cookie-name prefix, not the app query counter. Removing it from the original URI before hashing makes URLs such as `/orders?h_ra=1` and `/orders?h_ra=2` share the same counter cookie, so redirect counts can leak across distinct app URLs. This removal should be limited to the PARAM counter state.

docs/src/main/asciidoc/includes/security/providers/oidc.adoc
```diff
@@ -67,6 +67,12 @@ for a JWT
 5. The JWT is stored in a cookie (if cookie support is enabled, which it is by default)
 6. Helidon service redirects to original endpoint (on itself)
 
+Redirect attempts are counted to prevent infinite login redirects. By default, Helidon stores the count in the
+`redirect-attempt-param` query parameter. Set `redirect-attempt-counter-strategy` to `COOKIE` to store the counter in
+a small cookie instead, or to `NONE` to disable redirect attempt counting. The `redirect-attempt-param` value is used
```
danielkec: `NONE` disables more than storage of the count: `redirectAttempt()` always returns `0` for this strategy, so the `max-redirects` infinite-loop guard never trips. Please state explicitly that choosing `NONE` disables redirect-loop protection/max-redirect enforcement as well.

danielkec (commented):
Additional review comments from the latest redirect-attempt counter feedback triage.

tomas-langer (approved):

PR #12010, opened 2026-05-21
Fix FT bulkhead waiting metrics cleanup (4.x)

tomas-langer:
Fixes async bulkhead waiting-duration metric bookkeeping so per-invocation suppliers are cleaned up from method-scoped tracking when invocations are rejected, cancelled, or complete after queue timing is recorded.

This is a required fix to consistently pass FaultTolerance TCK on MacOS ARM machines.

Validation:
- `env JAVA_HOME=/Users/tomas/.sdkman/candidates/java/21.0.7-oracle mvn -Ptests,tck -pl microprofile/tests/tck/tck-fault-tolerance -am -DskipTests install`
- `env JAVA_HOME=/Users/tomas/.sdkman/candidates/java/21.0.7-oracle mvn -Ptests,tck -pl microprofile/tests/tck/tck-fault-tolerance test`
- `./etc/scripts/copyright.sh`

### DIFF
--- a/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
+++ b/microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
@@ -767,7 +767,7 @@ class MethodInvoker implements FtSupplier<Object> {
                 startNanos[0] = state.starts.removeFirst();
                 found[0] = true;
                 if (!state.completionObserved) {
-                    state.listenerOwnsCompletionRecord = true;
+                    state.recordedBeforeCompletion = true;
                 }
                 return state.starts.isEmpty() && state.completionObserved ? null : state;
             });
@@ -785,9 +785,11 @@ class MethodInvoker implements FtSupplier<Object> {
                     return state;
                 }
                 startNanos[0] = state.starts.removeFirst();
-                record[0] = state.completionObserved;
-                return state.starts.isEmpty() && (state.completionObserved || !state.listenerOwnsCompletionRecord)
-                        ? null : state;
+                record[0] = true;
+                if (!state.completionObserved) {
+                    state.recordedBeforeCompletion = true;
+                }
+                return state.starts.isEmpty() && state.completionObserved ? null : state;
             });
             if (record[0]) {
                 recordWaitingDuration(startNanos[0]);
@@ -805,11 +807,11 @@ class MethodInvoker implements FtSupplier<Object> {
                 }
                 if (!state.starts.isEmpty()) {
                     state.completionObserved = true;
-                    state.listenerOwnsCompletionRecord = false;
+                    state.recordedBeforeCompletion = false;
                     result[0] = true;
                     return state;
                 }
-                if (state.listenerOwnsCompletionRecord) {
+                if (state.recordedBeforeCompletion) {
                     result[0] = true;
                     return null;
                 }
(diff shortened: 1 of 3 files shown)


spericas (dismissed):

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
```diff
@@ -652,28 +649,192 @@ private void updateMetricsAfter(Throwable cause) {
...
+
+        void cancelled(Supplier<?> supplier) {
+            long[] startNanos = new long[1];
+            boolean[] record = new boolean[1];
+            states.compute(supplier, (ignored, state) -> {
+                if (state == null || state.starts.isEmpty()) {
+                    return state;
+                }
+                startNanos[0] = state.starts.removeFirst();
+                record[0] = state.completionObserved;
```
danielkec: This path seems to miss one queued wait for explicit caller cancellation when the same async retry invocation has multiple queued starts. For an `@Asynchronous @Retry @Timeout @Bulkhead` method, the same `handlerSupplier` can be queued more than once; if the returned future is cancelled while more than one start remains, this removes one start with `completionObserved == false` and does not record it, while the remaining state makes `recordedOrPending` suppress the completion-side duration record. The timeout-driven tests do not exercise that caller-cancel path.

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/MetricsTest.java
```diff
@@ -465,4 +466,69 @@ void testBulkheadMetricsAsync() throws Exception {
...
+        long startCount = waitingDuration.getCount();
+
+        CompletableFuture<String> blocker = bean.retryingBulkhead(true);
+        assertThat(bean.awaitRetryingBulkheadEntered(), is(true));
+
+        try {
+            CompletableFuture<String> retried = bean.retryingBulkhead(false);
+            Gauge<Long> executionsWaiting = BulkheadExecutionsWaiting.get(
+                    getMethodTag(bean, "retryingBulkhead"));
+            awaitGaugeValue(executionsWaiting, 3L);
```
danielkec: This waits for an exact transient queue size while the queued invocation has only a 300 ms timeout. On a loaded runner the queue can pass through `3` between the 10 ms polls and then cancel before this assertion observes it, so the test can fail without an implementation regression. The queued-timeout test below has the same timing shape for queue size `1`.

microprofile/fault-tolerance/src/main/java/io/helidon/microprofile/faulttolerance/MethodInvoker.java
tomas-langer: Resolved in 8b10d9a3f4. `cancelled(...)` now records the removed queued waiting start, even before completion observes the supplier, and `recordedBeforeCompletion` prevents a later completion-side duplicate. Added explicit caller-cancellation coverage with `testBulkheadWaitingDurationRecordsExplicitCancel`.

microprofile/fault-tolerance/src/test/java/io/helidon/microprofile/faulttolerance/MetricsTest.java
tomas-langer: Resolved in 8b10d9a3f4. The timeout/retry tests no longer wait for exact transient queue sizes; they now wait for durable outcomes: future timeout and waiting-duration histogram count. The explicit-cancel regression uses a controlled non-timed queued call and only waits for the waiting gauge to reach at least one before cancelling.

PR #12020, opened 2026-05-22
Restore ProviderRequest class shape (4.x)

### LINKED ISSUE #12019: io.helidon.security.ProviderRequest change from class to interface causes issues (4.5.0 daily)
## Environment Details
* Helidon Version: 4.5.0 (daily snapshot)
* Helidon SE
* JDK version: 25
* OS: OL9

----------

## Problem Description
In Helidon 4.5.0 the ProviderRequest class was changed to an interface.
This would normally not be an issues, as we claim to be "source" compatible (i.e. you may need to recompile).

But this becomes a problem in bigger environments with framework(s) built on top of Helidon. The coordination and orchestration task creates a lot of friction.

This is caused by a distinction in Java - it is important whether bytecode invokes method on interface or class. The same bytecode cannot do both. So if a compiled binary expected interface, and class is provided instead, it fails at runtime.

We should consider switching back to interface, so a better migration path from 4.3.x is available.

Similar to Helidon-11782

tomas-langer:
Resolves #12019

Restores `ProviderRequest` as an abstract class while keeping the static factory methods, and updates `ProviderRequestImpl` to extend it so class-compiled callers can dispatch through the restored class shape.

### DIFF
--- a/security/security/src/main/java/io/helidon/security/ProviderRequest.java
+++ b/security/security/src/main/java/io/helidon/security/ProviderRequest.java
@@ -16,6 +16,7 @@
 
 package io.helidon.security;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -36,7 +37,13 @@ import io.helidon.security.util.AbacSupport;
  * custom attributes</li>
  * </ul>
  */
-public interface ProviderRequest extends AbacSupport {
+public abstract class ProviderRequest implements AbacSupport {
+    /**
+     * Constructor for subclasses.
+     */
+    protected ProviderRequest() {
+    }
+
     /**
      * Create a new provider request.
      *
@@ -51,8 +58,8 @@ public interface ProviderRequest extends AbacSupport {
      * @param boundAttributes suppliers of bound attributes
      * @return a new provider request
      */
-    static ProviderRequest create(SecurityContext context,
-                                  Map<String, Supplier<Object>> boundAttributes) {
+    public static ProviderRequest create(SecurityContext context,
+                                         Map<String, Supplier<Object>> boundAttributes) {
         return new ProviderRequestImpl(context, boundAttributes);
     }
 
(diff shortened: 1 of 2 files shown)


security/security/src/main/java/io/helidon/security/ProviderRequest.java
```diff
@@ -36,7 +37,13 @@
  * custom attributes</li>
  * </ul>
  */
-public interface ProviderRequest extends AbacSupport {
+public abstract class ProviderRequest implements AbacSupport {
```
danielkec, line 40: This restores dispatch for bytecode compiled when `ProviderRequest` was a class, but it breaks bytecode compiled against the released 4.4.x interface shape. For example, compiling `ProviderRequest.create(null, Map.of())` or `request.env()` against `helidon-security-4.4.1.jar` emits `InterfaceMethodref`/`invokeinterface`; running those classes against this PR fails with `IncompatibleClassChangeError` (`must be Methodref constant` / `Found class ... but interface was expected`). Since 4.4.0 and 4.4.1 are public releases, this trades the 4.3.x binary break for a 4.4.x one rather than fully resolving the 4.x compatibility issue.

security/security/src/main/java/io/helidon/security/ProviderRequestImpl.java
```diff
@@ -27,7 +27,7 @@
 
 import io.helidon.security.util.AbacSupport;
 
-class ProviderRequestImpl implements ProviderRequest {
+class ProviderRequestImpl extends ProviderRequest {
```
danielkec, line 30: The changed contract is a bytecode dispatch issue, but the current tests do not exercise compiled caller bytecode against the new class shape. Existing security tests use source-level calls or Mockito mocks, so they would not catch another `class`/`interface` flip; the failing case is old compiled bytecode using `invokevirtual` vs `invokeinterface` at runtime.

danielkec (commented):
Inline review comments from ABI compatibility review.

danielkec (approved):

PR #12028, opened 2026-05-23
Fix ListContext unregister identity handling (main)

### LINKED ISSUE #11906: ListContext unregister removes replacement instance on main
## Problem

The same `ListContext.unregister(instance)` behavior tracked for `helidon-4.x` is present on `main`.

This tracks carrying the fix from #11904 forward to `main`.

## Main Relevance

`main` has the same type-based removal path in:

- `common/context/context/src/main/java/io/helidon/common/context/ListContext.java`

If one instance is registered and then another instance of the same class replaces it, unregistering the old instance removes the currently registered replacement.

Expected result after `unregister(first)`: `get(TestValue.class)` returns the replacement instance.

Actual result: the replacement is removed.

## Proposed Fix

Carry forward the 4.x fix once #11904 is resolved.

tomas-langer:
Resolves #11906

Carries the `ListContext.unregister(instance)` identity handling fix from #11907 forward to `main`.

The change keeps registration replacement by type, but makes unregister remove only stored registrations whose instance is identical to the argument, including already resolved suppliers.

### DIFF
--- a/common/context/context/src/main/java/io/helidon/common/context/Context.java
+++ b/common/context/context/src/main/java/io/helidon/common/context/Context.java
@@ -16,6 +16,7 @@
 
 package io.helidon.common.context;
 
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


danielkec (approved):
Reviewed; no findings.

PR #12029, opened 2026-05-24
Preserve malformed percent sequences in URI decoding (main)

### LINKED ISSUE #11909: Preserve malformed percent sequences in URI query decoding
## Description

Main-line tracking for the same `common/uri` percent-decoding behavior as #11908.

The non-validating URI query parser should consume only complete percent escapes with two hexadecimal characters. Other percent sequences should remain literal text in the decoded query output.

## Expected behavior

- Valid percent escapes continue to decode normally.
- Non-validating query parsing preserves malformed percent sequences literally in decoded names and values.
- Validating query parsing continues to reject malformed percent encoding.

## Scope

This affects `common/uri` on `main`.

The fix should keep the public API and SPI shape unchanged.

tomas-langer:
Resolves #11909

Carries the merged 4.x URI percent-decoding fix from #11910 to `main`, preserving malformed percent sequences literally in lenient URI and query decoding while keeping strict validation behavior unchanged.

### DIFF
--- a/common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriEncoding.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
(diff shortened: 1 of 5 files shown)


danielkec (approved):
Looks good. Reviewed exact head 8cc9b4081a4e4da65847150e94b99f2549d3b809.

PR #12030, opened 2026-05-24
Use custom uncaught exception handlers in WebServer

### LINKED ISSUE #6180: Níma: use custom uncaught exception handler
Currently we leave uncaught exception to the default handler in Java. 
We may want to add a custom uncaught exception handler to all our executor services for Níma, to be able to correctly log them.

tomas-langer:
Resolves #6180

Adds WebServer virtual-thread uncaught exception handlers for executor-service paths and logs unexpected failures from fire-and-forget WebServer work.


danielkec (approved):
Reviewed exact head c53de70b4f5ca8ff9237040aaab53293f4b07a95. No findings.

PR #12032, opened 2026-05-25
Fix intercepted field injection descriptor generation (27.x)

### LINKED ISSUE #12031: Intercepted field injection generates descriptor with undeclared invoker
The service code generator emits an invalid descriptor for an intercepted field injection when the injected field is not named `config`.

Reproducer:

1. In `service/tests/interception`, add a singleton service with a non-private field annotated with both `@Service.Inject` and an interception trigger.
2. Run `mvn -Ptests -pl service/tests/interception -am -DskipTests test-compile`.

Observed behavior:

The generated descriptor declares a field-specific invoker local such as `dependency__invoker`, but the field assignment calls `config__invoker.invoke(...)`. Compilation fails because `config__invoker` is not declared.

Expected behavior:

The generated field injection assignment should call the invoker local created for that field.

Impact:

Any intercepted `@Service.Inject` field whose source field name is not `config` can generate a service descriptor that does not compile.

tomas-langer:
Resolves #12031

Fix service descriptor generation for intercepted field injection so the generated assignment invokes the field-specific interception invoker. Add interception test coverage for a non-`config` field injection to prove the generated descriptor compiles and the field injection interceptor runs.

### DIFF
--- a/service/tests/interception/src/main/java/io/helidon/service/tests/interception/FieldInjectedDependency.java
+++ b/service/tests/interception/src/main/java/io/helidon/service/tests/interception/FieldInjectedDependency.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2026 Oracle and/or its affiliates.
+ * Copyright (c) 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


danielkec (approved):
Reviewed head 5bf48af7459aa70aa956531e8426410db937de2c. No blocking findings; focused local validation passed.

PR #12034, opened 2026-05-25
Fix service registry lookup cache poisoning (27.x)

### LINKED ISSUE #12033: Service registry lookup cache can be poisoned by mutable lookup results
`ServiceRegistry#lookupServices(Lookup)` can expose mutable lookup result lists that are also stored in the registry lookup cache when lookup caching is enabled.

Reproducer:

1. Enable service registry lookup caching.
2. Call `lookupServices(Lookup.create(SomeContract.class))` for a contract with matching services.
3. Mutate the returned list, for example by clearing it.
4. Call the same lookup again.

The repeated lookup can return the caller-mutated contents, such as an empty list, instead of the services discovered by the registry.

Expected behavior: callers should not be able to mutate the registry cache through a returned lookup result. Repeated lookups with the same `Lookup` should keep returning the discovered services regardless of caller-side mutations to an earlier returned list.

Impact: in-process code with access to the service registry can corrupt later cached lookups and hide services from consumers that rely on `lookupServices`.

tomas-langer:
Resolves #12033

Stores immutable service registry lookup cache snapshots so caller-side mutations of `lookupServices` results cannot poison later cached lookups. Adds regression coverage for both indexed and full-scan cache lookup paths.

### DIFF
--- a/service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
+++ b/service/registry/src/main/java/io/helidon/service/registry/CoreServiceRegistry.java
@@ -358,7 +358,7 @@ class CoreServiceRegistry implements ServiceRegistry, Scopes {
                     if (!result.isEmpty()) {
                         traceLookup(lookup, "by single contract", result);
                         if (cacheEnabled) {
-                            cache.put(lookup, result);
+                            return cacheLookupResult(lookup, result);
                         }
 
                         return result;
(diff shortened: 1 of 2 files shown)


danielkec (approved):
Reviewed exact head 58739d0987c173a993c1837978ec536f41c6d5e8. No findings.

PR #12039, opened 2026-05-26
Optimize JSON media writers for fixed-length entities

### LINKED ISSUE #12037: Optimize JSON media writers for fixed-length HTTP entities
Helidon JSON media writers currently write many in-memory JSON entities through streaming entity handling. For HTTP/1 responses without a known length, this results in chunked transfer encoding even when the source entity is already a materialized Java object or JSON object model.

Improve the JSON media writers so materialized JSON entities can provide their serialized bytes through the existing media instance-writer mechanism. This lets WebServer and WebClient use fixed-length entity handling where normal HTTP framing rules allow it.

The improvement should cover JSON Binding, JSON-P, JSON-B, Helidon JSON values, Jackson, and Smile. It should preserve existing content type and charset behavior, keep genuinely streaming or unbounded entities on the streaming path, and avoid adding new public API unless implementation reuse clearly requires a shared helper.

Expected benefits include fewer chunked HTTP/1 responses for ordinary JSON entities, known content lengths for eligible request and response bodies, and lower overhead in common JSON response paths.

tomas-langer:
Resolves #12037

Adds byte-backed instance writers for JSON media providers so JSON Binding, Helidon JSON, JSON-B, JSON-P, Jackson, and Smile can provide fixed-length HTTP entities when serialized bytes are materialized in memory.

This changes normal JSON and Smile request and response framing from chunked streaming to fixed-length framing when the writer can materialize bytes. The entity object is already in memory; the serialized bytes are retained long enough for the HTTP layer to send them, so large or encoded responses can temporarily hold additional materialized data.

HttpArena JSON Binding 45 second measurement:

| Variant | Framing | Throughput | Avg latency | p50 | p90 | p99 | 2xx responses |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Original | `Transfer-Encoding: chunked` | 82.93K req/s | 39.80ms | 41.00ms | 42.00ms | 43.90ms | 3,732,096 |
| Target solution | `Content-Length` | 368.91K req/s | 6.16ms | 59us | 41.00ms | 43.90ms | 16,602,357 |

Adds writer-level coverage for instance bytes, HTTP/1 JSON-B request and response framing checks, and JMH coverage for JSON Binding, JSON-B, Jackson, and Smile across small and large payloads.

### DIFF
--- a/http/media/media/src/main/java/io/helidon/http/media/ByteArrayInstanceWriter.java
+++ b/http/media/media/src/main/java/io/helidon/http/media/ByteArrayInstanceWriter.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.http.media;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.UncheckedIOException;
+import java.util.Objects;
+import java.util.OptionalLong;
+
+/**
+ * {@link InstanceWriter} for bytes already materialized in memory.
+ */
+public final class ByteArrayInstanceWriter implements InstanceWriter {
+    private final byte[] instanceBytes;
+
+    private ByteArrayInstanceWriter(byte[] instanceBytes) {
+        this.instanceBytes = Objects.requireNonNull(instanceBytes);
+    }
+
+    /**
+     * Create a new instance writer for bytes already materialized in memory.
+     *
+     * @param instanceBytes bytes to write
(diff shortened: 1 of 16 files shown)


http/media/media/src/main/java/io/helidon/http/media/ByteArrayInstanceWriter.java
```diff
@@ -0,0 +1,68 @@
...
+        return new ByteArrayInstanceWriter(instanceBytes);
+    }
+
+    @Override
+    public OptionalLong contentLength() {
+        return OptionalLong.of(instanceBytes.length);
+    }
+
+    @Override
+    public boolean alwaysInMemory() {
```
danielkec, line 51: This makes all byte-backed JSON/Smile writers take the unconditional in-memory path. `ServerResponseBase.doSend` and `Http1ClientRequestImpl.doSubmit` check `alwaysInMemory()` before applying `maxInMemoryEntity`, so setting a small `maxInMemoryEntity` and sending a larger JSON entity will still serialize the whole payload into a byte array, and the server send path can buffer it again, instead of streaming. That bypasses the configured memory limit and can turn large entities into heap pressure.

tomas-langer: We cannot check the size of the JSON in advance, and it is already in memory. This is a choice specific to JSON (and other bindings where the object is already an in-memory structure that is going to be serialized).
The intention of `maxInMemoryEntity` is for media supports such as file handling, or database loads, where we can check the size in advance.
To check the size of the JSON structure we need to serialize it - i.e. we just do it right away and keep it in memory.

danielkec (approved):

PR #12040, opened 2026-05-26
Replace JSON-P with Helidon JSON in MongoDB dbclient

### LINKED ISSUE #11378: Replace JSON-P with Helidon JSON in helidon-dbclient-mongodb

tomas-langer:
Resolves #11378

Replaces MongoDB dbclient JSON-P statement parameter handling with Helidon JSON, removes the MongoDB module's JSON-P dependency surface, and updates third-party license consumers after removing Parsson from this module.

### DIFF
--- a/dbclient/mongodb/src/main/java/io/helidon/dbclient/mongodb/MongoDbStatement.java
+++ b/dbclient/mongodb/src/main/java/io/helidon/dbclient/mongodb/MongoDbStatement.java
@@ -254,8 +254,7 @@ abstract class MongoDbStatement<S extends DbStatement<S>> extends DbStatementBas
                 || value instanceof Boolean
                 || value instanceof Character
                 || value instanceof Enum<?>
-                || value instanceof JsonValue
-                || value instanceof jakarta.json.JsonValue;
+                || value instanceof JsonValue;
     }
 
     private record NamedParameter(Object value, Map<String, Object> flattenedValues) {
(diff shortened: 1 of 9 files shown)


dbclient/mongodb/src/main/java/io/helidon/dbclient/mongodb/MongoDbStatement.java
```diff
@@ -124,6 +128,20 @@ String prepareStatement(DbClientServiceContext serviceContext) {
         return statement;
     }
 
+    @Override
+    public S params(Object... parameters) {
```
danielkec: This changes `params(Object...)` for a single `JsonObject`/`JsonArray` in a way that no longer matches the public `DbStatement` contract, which describes this method as ordered/indexed parameter binding.

Example indexed statement:

```java
{ "collection": "foo", "operation": "insert", "value": ? }
```

With:

```java
.params(JsonObject.builder().set("id", 25).build())
```

That call is still a single vararg value, but this branch treats it as named parameters. The indexed `?` is then left unresolved, and `Document.parse` fails with:

```text
Invalid JSON input ... Character: ?
```

`JsonArray` has the same ambiguity for a single `?`: the array elements are treated as separate indexed params instead of one array value.

tomas-langer: I have reverted this part of the change.

tomas-langer: Fixed in a15a9ca1c8 by removing the MongoDB-specific `params(Object...)` override. A single Helidon `JsonObject` or `JsonArray` now follows the inherited `DbStatement` varargs contract and is bound as one indexed parameter; named binding remains via named APIs such as `addParam(String, ...)` and `params(Map)`. Added focused MongoDB coverage for single JSON object/array indexed binding, including the `createGet` forwarding path.

danielkec (approved):
Reviewed current head a15a9ca1c8e1baab77dbe2893a713f6104d8165a. The previous params(Object...) finding is addressed and the focused MongoDB validation passed.

PR #12045, opened 2026-05-27
Add backward compatibility check (27.x)

### LINKED ISSUE #12044: Add backward compatibility check
Helidon main should keep a compatibility check that proves declarative modules generated with Helidon 4.4.1 remain usable from the current application runtime.

Add an integration test with a standalone Helidon 4.5.0 declarative module that has no `Main` class and is consumed by a current Helidon 27 application. The test should cover the supported declarative feature set so changes on `main` do not silently break older generated modules.

The compatibility check should include declarative HTTP server and client APIs, service registry integration, configuration/default values, headers, fault tolerance, metrics, tracing, validation, scheduling, CORS, and WebSocket server/client support.

tomas-langer:
Resolves #12044

Adds a compatibility test that compiles a declarative library with Helidon 4.5.4 and runs its existing generated classes on the current Helidon 27 runtime. The legacy library uses Java 21 bytecode and the 4.5.4 annotation processors; the current application consumes its compiled JAR without recompiling it.

Coverage includes HTTP server and REST client behavior, WebSockets, configuration, fault tolerance (including async execution and rejection paths), metrics, tracing, parameter and return-value validation, scheduling, CORS, generated builders and JSON converters, and service scopes, lifecycle, qualifiers, factories, and events. Data and declarative features introduced after 4.5.4 are outside this fixture's scope.

The service registry also recognizes contract-qualified element interceptors when older metadata omits the enclosing type. A separate interception regression covers that fallback and verifies that it does not apply to explicit enclosing types or unrelated methods; the 4.5.4 fixture already supplies enclosing-type metadata.

Older generated gauge cleanup resolves its injected meter-registry supplier during shutdown. The runtime now permits an injected supplier to return an already initialized, still-active dependency during a pre-destroy callback, restricted to the callback thread and the same scope instance. The dependency's scope must still be resolvable during cleanup. It preserves ordinary shutdown lookup rejection and does not create dependencies or wait for unfinished factory or interception initialization. Each compatibility lookup logs a warning identifying the requested service provider and injection point and recommending that cleanup dependencies be retained before shutdown.


declarative/tests/compatibility/app/src/test/java/io/helidon/declarative/tests/compatibility/app/DeclarativeCompatibilityTest.java
```diff
@@ -0,0 +1,245 @@
...
+                                                          .addQualifier(Qualifier.create(RestClient.Client.class))
+                                                          .build());
+
+        assertThat(typedClient.hello("Bob", "typed"), is("Bonjour Hello Bob typed Hello"));
+        assertThat(typedClient.entity("payload"), is("entity:payload"));
+        assertThat(typedClient.clientHeader(), is("legacy-client"));
+
+        assertThat(typedClient.fallback(), is("fallback"));
+        assertThat(typedClient.retry(), is("retry:2"));
+        HttpException exception = assertThrows(HttpException.class, typedClient::circuit);
```
danielkec, line 144: These assertions do not prove the circuit breaker, timeout, or bulkhead interceptors are active. `circuit()` always throws the same 403, `timeout(Optional.empty())` does not hit the timeout path, and `bulkhead()` has no contention, so this would still pass if those generated FT interceptors stopped being applied.

tomas-langer:
This issue must wait for 4.5.0 release

danielkec (approved):
Reviewed the compatibility fixture and service registry changes. All 89 focused local tests passed, including legacy gauge cleanup.

PR #12052, opened 2026-05-29
Forward-port service registry codegen interface-contract validation (27.x)

### LINKED ISSUE #11873: Forward-port #11872: Fix service registry codegen for interface-contract validation
Forward-port the 4.x service registry and service codegen follow-up for interface-contract validation.

The source 4.x issue is #11872.

This blocks #11866.

Expected behavior to carry forward:
- factory-provided service lookup routes provided-contract results through interception wrappers while preserving concrete provider lookup compatibility
- broad empty-contract lookup does not expose raw factory-provider descriptors that can bypass generated wrappers
- generated factory wrappers intercept provider factory SPI methods such as get(), services(), first(...), and list(...)
- supplier, services, injection-point, qualified, and optional-supplier factory wrappers preserve provider-specific first(...) and list(...) behavior without eager instance creation or null delegates
- factory adapter raw-method selection matches the expected factory SPI signatures, not only method name and arity
- direct provider-contract lookup preserves the requested provider contract in ServiceInstance.contracts()
- generated wrapper descriptors retain provider metadata such as @Weight and @Service.RunLevel
- registry codegen extension processing can opt into service-contract annotation discovery, includes contract-triggered annotations in the round type map, and avoids repeated uncached contract scans

Carry forward the focused service-registry and service-codegen regression coverage from the 4.x issue.

tomas-langer:
Resolves #11873

Forward-ports the service registry codegen interface-contract validation fix from #11875 to main. This keeps validation method constraints aligned with service contract discovery, preserves the 27.x factory-contract metadata behavior for providers, and updates validation fixtures and snippets for service-contract interfaces.

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
@@ -173,11 +173,29 @@ class InterceptorGenerator {
                 .filter(it -> ValidationHelper.needsWork(constraintAnnotations, it))
                 .toList();
 
-        return !elementsWithValidation.isEmpty()
-                && elementsWithValidation.stream()
+        if (elementsWithValidation.isEmpty()) {
+            return false;
+        }
+        if (!elementsWithValidation.stream()
                 .allMatch(it -> ElementInfoPredicates.isMethod(it)
                         && !ElementInfoPredicates.isPrivate(it)
-                        && !ElementInfoPredicates.isStatic(it));
+                        && !ElementInfoPredicates.isStatic(it))) {
+            return false;
+        }
+        return isServiceContract(type);
+    }
+
+    private boolean isServiceContract(TypeInfo type) {
+        if (roundContext.serviceContracts(type).isEligible(type)) {
+            return true;
+        }
+
+        TypeName contractType = type.typeName();
+        return roundContext.types()
+                .stream()
+                .filter(this::isService)
+                .map(roundContext::serviceContracts)
+                .anyMatch(serviceContracts -> serviceContracts.isEligible(contractType));
     }
 
     private boolean isService(TypeInfo type) {
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
@@ -167,10 +167,6 @@ class InterceptorGenerator {
         if (type.kind() != ElementKind.INTERFACE) {
             return false;
         }
-        if (!type.hasAnnotation(ServiceCodegenTypes.SERVICE_ANNOTATION_CONTRACT)) {
-            return false;
-        }
-
         List<TypedElementInfo> elementsWithValidation = type.elementInfo()
                 .stream()
                 .filter(it -> ValidationHelper.needsWork(constraintAnnotations, it))
@@ -182,7 +178,21 @@ class InterceptorGenerator {
         return elementsWithValidation.stream()
                 .allMatch(it -> ElementInfoPredicates.isMethod(it)
                         && !ElementInfoPredicates.isPrivate(it)
-                        && !ElementInfoPredicates.isStatic(it));
+                        && !ElementInfoPredicates.isStatic(it))
+                && isServiceContract(type);
+    }
+
+    private boolean isServiceContract(TypeInfo type) {
+        if (roundContext.serviceContracts(type).isEligible(type)) {
+            return true;
+        }
+
+        TypeName contractType = type.typeName();
+        return roundContext.types()
+                .stream()
+                .filter(this::isService)
+                .map(roundContext::serviceContracts)
+                .anyMatch(serviceContracts -> serviceContracts.isEligible(contractType));
     }
 
     private boolean isService(TypeInfo type) {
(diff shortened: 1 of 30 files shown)


declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
```diff
@@ -173,11 +173,29 @@ private boolean hasOnlyNonPrivateInterfaceMethodValidation(TypeInfo type) {
...
                         && !ElementInfoPredicates.isPrivate(it)
-                        && !ElementInfoPredicates.isStatic(it));
+                        && !ElementInfoPredicates.isStatic(it))) {
+            return false;
+        }
+        return isServiceContract(type);
+    }
+
+    private boolean isServiceContract(TypeInfo type) {
+        if (roundContext.serviceContracts(type).isEligible(type)) {
```
danielkec: With default `AUTO_ADD_NON_CONTRACT_INTERFACES=true`, this returns true for any interface, so a standalone constrained interface (`String validate(@NotBlank String name)`) bypasses the missing `@Validation.Validated` error. The source PR required `@Service.Contract`; the updated test only covers option=false.

tomas-langer: Thanks, good catch. I narrowed the mainline adaptation back to the source PR behavior in d87c71a6780fa72cdeeb6fa88a77b179e41a9735: the standalone constrained-interface bypass now requires an explicit @Service.Contract instead of option-sensitive service-contract eligibility, so the default AUTO_ADD_NON_CONTRACT_INTERFACES=true path no longer lets a plain constrained interface skip the @Validation.Validated diagnostic. I also removed the option override from ValidationCodegenTest.testNonServiceInterfaceMethodConstraintRequiresValidated so the default-options case is covered, and narrowed the meta-annotated positive case to an explicit service contract.

tvallin (commented):
I think this might introduce a compatibility regression around constrained service interfaces.

`InterceptorGenerator` now exempts non-service interfaces only when the interface is directly annotated with `@Service.Contract`. That looks narrower than the service registry contract discovery rules, where implemented interfaces are still treated as contracts by default via `AUTO_ADD_NON_CONTRACT_INTERFACES`, and where `@Service.ExternalContracts` is supported for interfaces whose source cannot be annotated.

So a service like this may now fail codegen unless `Api` can be changed to add `@Service.Contract`:

```java
interface Api {
    String validate(@Validation.String.NotBlank String name);
}

@Service.Singleton
class Impl implements Api {
    @Override
    public String validate(String name) {
        return name;
    }
}
```

This seems worth checking because the updated tests now cover the explicit `@Service.Contract` case, but not the default implicit-contract or `@Service.ExternalContracts` cases. If the intended behavior is to preserve default service-contract discovery, this probably needs to use the same eligibility logic as service codegen rather than requiring the direct annotation.

tvallin (changes requested):
I think this might introduce a compatibility regression around constrained service interfaces.

`InterceptorGenerator` now exempts non-service interfaces only when the interface is directly annotated with `@Service.Contract`. That looks narrower than the service registry contract discovery rules, where implemented interfaces are still treated as contracts by default via `AUTO_ADD_NON_CONTRACT_INTERFACES`, and where `@Service.ExternalContracts` is supported for interfaces whose source cannot be annotated.

So a service like this may now fail codegen unless `Api` can be changed to add `@Service.Contract`:

```java
interface Api {
    String validate(@Validation.String.NotBlank String name);
}

@Service.Singleton
class Impl implements Api {
    @Override
    public String validate(String name) {
        return name;
    }
}
```

This seems worth checking because the updated tests now cover the explicit `@Service.Contract` case, but not the default implicit-contract or `@Service.ExternalContracts` cases. If the intended behavior is to preserve default service-contract discovery, this probably needs to use the same eligibility logic as service codegen rather than requiring the direct annotation.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
```diff
@@ -167,14 +167,19 @@ private boolean hasOnlyNonPrivateInterfaceMethodValidation(TypeInfo type) {
         if (type.kind() != ElementKind.INTERFACE) {
             return false;
         }
+        if (!type.hasAnnotation(ServiceCodegenTypes.SERVICE_ANNOTATION_CONTRACT)) {
```
tvallin: I think this might introduce a compatibility regression around constrained service interfaces.

`InterceptorGenerator` now exempts non-service interfaces only when the interface is directly annotated with `@Service.Contract`. That looks narrower than the service registry contract discovery rules, where implemented interfaces are still treated as contracts by default via `AUTO_ADD_NON_CONTRACT_INTERFACES`, and where `@Service.ExternalContracts` is supported for interfaces whose source cannot be annotated.

So a service like this may now fail codegen unless `Api` can be changed to add `@Service.Contract`:

```java
interface Api {
    String validate(@Validation.String.NotBlank String name);
}

@Service.Singleton
class Impl implements Api {
    @Override
    public String validate(String name) {
        return name;
    }
}
```

This seems worth checking because the updated tests now cover the explicit `@Service.Contract` case, but not the default implicit-contract or `@Service.ExternalContracts` cases. If the intended behavior is to preserve default service-contract discovery, this probably needs to use the same eligibility logic as service codegen rather than requiring the direct annotation.

tvallin (changes requested):

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/InterceptorGenerator.java
tomas-langer: This is how it works in 4.x right now. I am working on an additional PR in 4.x and I will then forward port it to `main`

tvallin (dismissed):

tvallin (approved):

Issue #6183, opened 2023-02-13
Níma: SNI Support


tomas-langer:
WebClient:
 - there is SNI support thanks to the default SSL Socket implementation in Java, though it explicitly uses a single value (from URL), we should allow using the `Host` header value as well, if the user wants

WebServer:
 - for each server name, we should allow custom configuration of TLS (e.g. this should act similar to additional sockets, except it is on the same port, just with different TLS setup that has different SNI hosts). This would require quite big changes in WebServer setup, esp. validation of each host (for uniqueness and SNI TLS configuration)

Verdent:
Investigate proper SNI behavior

PR #12054, opened 2026-05-29
Add WebClient SNI support (27.x)

### LINKED ISSUE #4294: Allow WebClient to specify SNIHostName on outbound TLS connection
## Environment Details
* Helidon Version: 2.5.0
* Helidon SE
* JDK version: Java 11
* OS:

----------

## Problem Description
When making HTTPS requests it is not possible to "override" hostname within the TLS handshake. 

Ideally this should probably infer the correct value automatically from the HTTP "Host" header (like CURL does: https://stackoverflow.com/a/50279590).

As a minimum, caller should be able to specify the value explicitly without relation to the HTTP "Host" header.

### LINKED ISSUE #12055: WebClient SNI support
WebClient:

there is SNI support thanks to the default SSL Socket implementation in Java, though it explicitly uses a single value (from URL), we should allow using the Host header value as well, if the user wants

tomas-langer:
Part of #6183
Resolves #12055
Resolves #4294

Adds WebClient SNI configuration and routes effective TLS peer host selection through WebClient, HTTP/1, HTTP/2, gRPC, and JSON-RPC clients. The change also isolates connection caches by effective TLS identity, preserves HTTP/2 fallback lifecycle behavior, documents the shared configuration, and adds focused regression and benchmark coverage.

### DIFF
--- a/webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcBaseClientCall.java
+++ b/webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcBaseClientCall.java
@@ -33,6 +33,7 @@ import io.helidon.common.buffers.BufferData;
 import io.helidon.common.buffers.CompositeBufferData;
 import io.helidon.common.socket.HelidonSocket;
 import io.helidon.grpc.core.GrpcHeadersUtil;
+import io.helidon.http.ClientRequestHeaders;
 import io.helidon.http.Header;
 import io.helidon.http.HeaderName;
 import io.helidon.http.HeaderNames;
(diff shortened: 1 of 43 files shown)


webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcBaseClientCall.java
```diff
@@ -296,28 +298,53 @@ GrpcClientImpl grpcClient() {
...
+                    Proxy.noProxy());
+        } else {
+            connectionKey = ConnectionKey.create(
+                    clientUri,
+                    sni,
+                    clientConfig.tls(),
+                    DefaultDnsResolver.create(),
+                    DnsAddressLookup.defaultLookup(),
+                    Proxy.noProxy(),
+                    emptyHeaders());
```
danielkec: Is this intentional for gRPC? In `HOST_HEADER` mode this path builds the connection key with empty headers, so SNI falls back to `clientUri.authority()` instead of any effective gRPC `:authority`, such as `stub.withAuthority(...)`. If that limitation is expected, should the gRPC docs call it out explicitly?

danielkec (approved):
Reviewed at 6746f195d748050a0326a68a16e594897459b277. No actionable findings.

Verdent (approved):

PR #12058, opened 2026-05-30
Support pre-compressed static content (27.x)

### LINKED ISSUE #11647: Static content improvement - pre-compressed files
Our current static content handler provides files as requsted, and we depend on `helidon-http-encoding` support to compress it to the correct type (i.e. gzip, deflate etc.).

But in case the file already exists correctly compressed, we could just send the file as is.

Example:

Let's consider the following static content directory:

- `big-file.html`
- `big-file.html.gz`
- `big-file.html.br`

When a request comes for `big-file.html` with `Accept-Encoding: br;q=1, gzip;q=0.8`, we could return the `big-file.html.br` even though we do not support Brotli encoding, or when it is not on classpath.

We would need a map of content encoding name to a file suffix (i.e. br->br; gzip->gz), and have a toggle to disable the feature if it would not be correct, but otherwise this would improve performance greatly.

Relevant benchmark that could use this: https://github.com/MDA2AV/HttpArena/pull/503

---

## Design Document

# Static Content Pre-Compressed Files

Status: proposed design

Issue: helidon-io/helidon#11647

## Summary

Static content handlers should be able to serve pre-compressed resource variants
when a client advertises support for the matching content coding.

For a logical resource such as `big-file.html`, a static content directory may
contain these files:

```text
big-file.html
big-file.html.gz
big-file.html.br
```

When a request for `big-file.html` includes:

```text
Accept-Encoding: br;q=1, gzip;q=0.8
```

the handler should serve `big-file.html.br` with:

```text
Content-Type: text/html
Content-Encoding: br
Vary: Accept-Encoding
```

This is representation selection for existing files. It is not runtime
compression, and it must not depend on a Brotli or gzip encoder being present on
the classpath.

## Goals

- Serve pre-compressed filesystem and classpath static resources when the client
  accepts the content coding.
- Preserve existing static content path resolution, welcome file handling,
  single-file handlers, conditional requests, range requests, and memory or
  record caching semantics.
- Use the logical resource name for media type detection. `big-file.html.br`
  served for `big-file.html` must still be `text/html`.
- Set `Content-Encoding` on selected pre-compressed responses so the normal
  WebServer content-encoding layer does not encode the response again.
- Add `Vary: Accept-Encoding` whenever the response can vary by accepted
  encoding.
- Add proper `identity` handling for `Accept-Encoding` negotiation as a connected
  feature required by this change. Existing Helidon content encoding does not
  handle this completely today.
- Allow applications to disable this behavior for the whole static content
  feature.
- Allow applications to configure the content-coding-to-file-suffix map.
- Document the behavior in release notes because this is a new default in the
  unreleased Helidon 27 major line.

## Non-Goals

- No new runtime Brotli encoder or decoder.
- No new content-encoding provider module.
- No automatic creation, refresh, or validation of compressed sidecar files.
- No serving of `*.gz` or `*.br` sidecar files as logical resources unless the
  client requests those literal filenames.
- No broad rewrite of HTTP content negotiation beyond what this feature needs.

## Decision

Pre-compressed selection is performed inside static content handling after the
logical resource has been resolved and before the selected cached handler sends
the response.

The logical resource remains the resource named by the request and path mapper.
Pre-compressed variants are sidecar representations of that logical resource.
A variant filename is built by appending `.` and the configured suffix to the
logical resource name:

| Content coding | Suffix | Logical resource | Variant resource |
| --- | --- | --- | --- |
| `br` | `br` | `big-file.html` | `big-file.html.br` |
| `gzip` | `gz` | `big-file.html` | `big-file.html.gz` |

The default suffix map should include:

| Content coding | Suffix |
| --- | --- |
| `br` | `br` |
| `gzip` | `gz` |

Additional codings, such as `deflate`, should be configurable but not enabled by
default unless Helidon chooses a well-defined file suffix convention for them.

Configured content-coding names must be valid HTTP tokens because they are
emitted as `Content-Encoding` response header values when selected. The
configuration must reject `identity` and `*`, because `identity` represents the
absence of a content encoding and `*` is only an `Accept-Encoding` wildcard
matcher. It must also reject names containing commas, semicolons, whitespace,
controls, or any other character not allowed in an HTTP token. Coding names are
case-insensitive: store and emit the normalized lower-case form, reject
`identity` and `*` case-insensitively, and reject duplicate configured codings
after normalization.

Configured suffixes must be path-safe suffix names, not paths. A suffix value
must be non-empty and must not contain `/`, `\`, or other path separators. The
configuration value may be accepted either as `gz` or `.gz`, but the stored
normal form should be `gz` so the implementation always appends exactly one dot
between the logical resource and the suffix.

Pre-compressed selection should be enabled by default in Helidon 27. This is a
new feature in an unreleased major version, so the performance benefit outweighs
the compatibility risk that would exist for an already released line.

If this feature is ever backported to an already released Helidon line, it must
be disabled by default on that line. Backports must require explicit opt-in with
`pre-compressed-enabled: true`. The connected `Accept-Encoding` negotiation fix
for `identity`, `q=0`, and wildcard handling still applies to runtime content
encoding and must be covered by backport release notes.

Applications must be able to disable pre-compressed selection for the whole
static content feature with the top-level `pre-compressed-enabled` option. When
that option is `false`, static content does not perform sidecar lookup for any
configured filesystem, classpath, or single-file handler.

## Negotiation Rules

The handler should use the request `Accept-Encoding` header to choose a variant.

Rules:

- If `Accept-Encoding` is absent, serve the logical resource without a
  pre-compressed variant. This preserves Helidon's current default body
  selection behavior, though `Vary: Accept-Encoding` is still added while
  pre-compressed selection is enabled.
- If `Accept-Encoding` is present but empty, treat it as accepting no content
  coding. Serve only the identity resource, and do not select a sidecar or
  runtime response encoding.
- Ignore codings with `q=0`; they are not acceptable.
- Match content-coding names case-insensitively and normalize them to the
  configured coding name before selection. Multiple `Accept-Encoding` header
  values are processed as one combined field value.
- If the same coding appears more than once, use the most specific effective
  value for that coding. If duplicate concrete codings conflict, use the entry
  with the highest `q` value and preserve the first occurrence for tie-breaking.
- `identity` means the unencoded logical resource, not a sidecar variant. It is
  acceptable unless the request explicitly rejects it with `identity;q=0`, or
  rejects all otherwise-unlisted codings with `*;q=0` and does not provide a
  more specific `identity` value.
- Build a candidate set of representations the request can produce: identity
  when acceptable, concrete named sidecars whose resource exists, and concrete
  listener runtime encoders that are available for the listener.
- Choose the candidate with the best effective `Accept-Encoding` preference.
  This is analogous to `ServerRequestHeaders.bestAccepted(...)` for media types,
  but applied to content codings.
- A concrete coding or the wildcard coding `*` can select an available
  pre-compressed sidecar. A more specific coding entry overrides the wildcard,
  including a `q=0` exclusion. Because a sidecar is already an available
  encoded representation, selecting it does not require a listener runtime
  encoder.
- The wildcard coding `*` matches available sidecars and listener runtime
  encoders that are not named more specifically in the header. If only
  wildcard and identity are acceptable, identity wins over any
  wildcard-derived encoded representation.
- For the same concrete content coding, prefer an existing sidecar over runtime
  response encoding. For example, if both `big-file.html.gz` and a listener
  `gzip` encoder are available and `gzip` is the selected coding, serve the
  sidecar.
- If the selected candidate is identity, serve the logical resource as an
  identity response and do not let runtime response encoding change that choice.
- If the selected candidate is a runtime encoder, use the logical resource as
  the plain input and let the normal WebServer content-encoding layer send the
  selected encoded representation. Static content leaves `Content-Encoding`
  unset in this case.
- If no acceptable candidate can be produced, return `406 Not Acceptable`.
- Runtime content encoding must use the same parsed `Accept-Encoding` semantics
  as sidecar selection. It must not choose a coding rejected with `q=0`, and it
  must not send an identity response when identity is not acceptable.
- Sidecar selection must not require the selected coding to be supported by
  `ContentEncodingContext`. Existing bytes can be served for `br` even without a
  Brotli content-encoding provider.

Malformed `Accept-Encoding` syntax handling is not part of this feature, except
that the static-content path must not turn malformed input into an internal
server error. Any change that chooses a specific client-error status for
malformed `Accept-Encoding` belongs in a separate header-parsing task. Valid
`Accept-Encoding` headers that reject every representation this request can
produce are negotiation failures and must use `406 Not Acceptable`.

Tie handling should be deterministic. When two candidates have the same quality,
prefer the order in which the client listed them. If the client order does not
decide between sidecar candidates, compare their normalized content-coding names
alphabetically. Prefer identity over wildcard-derived runtime encoding and prefer
a sidecar over runtime encoding for the same concrete coding.

## Resource Model

The canonical logical resource should be servable by the existing static-content
rules before a sidecar variant is served. For `big-file.html`, the handler
should not serve `big-file.html.br` when `big-file.html` is missing, hidden,
unreadable, a directory without valid welcome resolution, or otherwise rejected
by the identity resource path.

Reasons:

- The logical resource supplies the default media type and user-facing resource
  identity.
- Existing not-found behavior remains predictable.
- Hidden sidecar files are not accidentally exposed as canonical content.

This canonical-resource rule applies only to negotiated sidecar lookup for the
logical resource. Direct requests for explicit filenames such as
`big-file.html.br` or `big-file.html.gz` are ordinary static file requests. They
are served or rejected using the same rules as any other static file, regardless
of whether `big-file.html` exists.

For directory requests with a welcome file, pre-compressed lookup applies to the
resolved welcome resource. A request for `/docs/` with welcome file `index.html`
may serve `/docs/index.html.br`. A request for `/docs` should still redirect to
`/docs/` before any welcome-file representation is sent.

For classpath static content, sidecar lookup follows the normal classpath
resource lookup rules, but same-origin sourcing is the default. A sidecar is
same-origin when it comes from the same classpath resource origin as the
canonical logical resource: the same JAR file for JAR resources, or the same
classpath directory root for file resources. This avoids serving a cacheable
encoded representation for one logical URL from an unrelated classpath entry.

Applications can opt in to cross-origin classpath sidecar sourcing. When this is
enabled, the identity resource and sidecar resources do not need to come from the
same JAR file or directory. An application may intentionally put plain, Brotli,
and gzip resources in different classpath entries, relying on the configured
class loader's normal lookup order for each resource name. This opt-in is not a
security boundary; applications that enable it must control their classpath order
and the sidecar resource names visible under the configured static-content root.

For single-file handlers, pre-compressed lookup applies to the configured single
file location. Any matching sidecar file is a representation of that one logical
file.

For path-mapped handlers, the path mapper produces the logical resource first.
The sidecar suffix is appended only after path mapping and normal static-content
root validation.

Helidon does not compare the freshness of the logical resource and its sidecar
variants. Build or deployment tooling owns keeping pre-compressed files in sync
with the logical resource. If a stale sidecar exists and is selected by
negotiation, Helidon serves that sidecar as the current representation.

## Response Metadata

When pre-compressed selection is enabled for the static content feature, every
static content response must merge `Vary: Accept-Encoding`, whether the selected
representation is a sidecar or the identity resource. This includes GET, HEAD,
range responses, identity fallback responses, and conditional responses such as
`304 Not Modified`. This is an intentional cache-safety tradeoff for a
feature-wide default: once a feature instance can select pre-compressed
representations, intermediaries should not reuse an identity response for a later
request that accepts an available sidecar variant.

When a sidecar variant is selected:

- `Content-Type` is detected from the logical resource filename, not from the
  sidecar filename.
- `Content-Encoding` is set to the selected content coding.
- `Content-Length` is the byte length of the selected sidecar representation.
- `Last-Modified` is based on the selected sidecar resource.
- `ETag` must distinguish the selected representation from the identity
  resource and from other encoded variants.

The `Last-Modified` source should follow the same metadata source Helidon
already uses for the selected resource type. Filesystem sidecars use the sidecar
file timestamp. Classpath `file:` URLs use the selected URL file timestamp. JAR
classpath resources keep the existing containing-JAR timestamp behavior rather
than introducing a new per-entry timestamp rule for this feature. Stream-only
classpath URLs omit `Last-Modified` when no timestamp is available.

When no sidecar is selected but the logical resource is used as input to runtime
response encoding, the selected runtime encoder owns `Content-Encoding` and
metadata that depends on the encoded bytes. Static content must not force
metadata that describes the unencoded identity bytes onto the encoded response.
In particular, runtime-encoded static responses do not send the identity
`Content-Length`.

The current static content ETag is derived from last-modified time. That is not
enough if two sidecar files and the identity file share the same timestamp.
Negotiated sidecar ETags should preserve the current static-content validator
style while adding selected-representation inputs. Use the selected sidecar's
validator source, selected content coding, and selected byte length, for example:

```text
"<selected-resource-last-modified-epoch>;encoding=br;length=<selected-length>"
```

Identity responses keep the existing logical-resource ETag shape unless changed
separately. Sidecar ETags remain strong when the existing static-content handler
would have emitted a strong ETag for that selected resource.

Runtime-encoded static responses should use a weak representation-specific ETag
when a plain logical-resource validator is available, for example
`W/"<plain-validator>;encoding=gzip"`. This avoids presenting a strong validator
for bytes that are produced dynamically while still allowing conditional
requests to revalidate the selected encoded representation. `Last-Modified` for
runtime-encoded static responses remains based on the logical resource.

Conditional request handling must run after representation selection.
`If-None-Match`, `If-Match`, `If-Modified-Since`, and
`If-Unmodified-Since` must be evaluated against the selected representation's
validators. Conditional responses such as `304 Not Modified` and
`412 Precondition Failed` should include the selected representation metadata
that applies, especially representation-specific `ETag` and
`Vary: Accept-Encoding`.

Range requests operate on the selected representation bytes. If
`big-file.html.br` is selected, a range response is a byte range over the Brotli
file, not over the decoded HTML.

`If-Range` must also use the selected representation's validator. Entity-tag
`If-Range` matching must use strong comparison; weak validators do not authorize
`206 Partial Content`. An HTTP-date authorizes a range only when it is a strong
validator for the selected representation under RFC 9110 Section 8.8.2.2 and
exactly matches that representation's `Last-Modified`. When the handler cannot
establish that date-validator strength, or when the selected validator does not
match, it ignores the range and sends the full selected representation.

When runtime response encoding is selected for static content, range handling
must not emit `206 Partial Content` for a byte range computed from the plain
file. The handler ignores `Range` and sends a full `200 OK` encoded response,
allowing the runtime encoder to produce the selected representation.

HEAD requests use the same selection logic as GET and send only headers. If HEAD
selects runtime response encoding, Helidon applies the selected encoder's
headers, such as `Content-Encoding`, without sending an entity and without
sending the identity `Content-Length`.

## Configuration

Use a top-level switch for the static content feature:

```yaml
server:
  features:
    static-content:
      pre-compressed-enabled: true
      pre-compressed-cross-origin-sourcing-enabled: false
      pre-compressed-encodings:
        br: br
        gzip: gz
      path:
        - context: "/"
          location: "/var/www/public"
          pre-compressed-enabled: false
```

Programmatic configuration should follow the same shape:

```java
StaticContentFeature.create(feature -> feature
    .preCompressedEnabled(true)
    .preCompressedCrossOriginSourcingEnabled(false)
    .putPreCompressedEncoding("br", "br")
    .putPreCompressedEncoding("gzip", "gz")
    .addPath(path -> path
        .context("/")
        .preCompressedEnabled(false)
        .location(publicDirectory)));
```

`pre-compressed-enabled` defaults to `true` in Helidon 27. If this feature is
backported to an already released Helidon line, the same option must default to
`false` on that line. Static content should expose the same pre-compressed
options on both `StaticContentConfig` and `BaseHandlerConfig`. The feature-level
values act as defaults for all handlers registered by that feature instance.
Handler-level values override those defaults. Direct services created with
`StaticContentFeature.createService(FileSystemHandlerConfig)` or
`StaticContentFeature.createService(ClasspathHandlerConfig)` use the handler
configuration directly, with the same Helidon 27 and backport defaults.

`pre-compressed-cross-origin-sourcing-enabled` defaults to `false`. It controls
only classpath static content. When `false`, classpath sidecars must come from
the same classpath resource origin as the canonical logical resource. When
`true`, sidecars may come from different classpath entries using normal class
loader lookup order.

`pre-compressed-cross-origin-sourcing-enabled` and `pre-compressed-encodings`
follow the same feature-default plus handler-override model. The suffix map on a
handler replaces the inherited feature/default map instead of merging with it.
This makes it possible to keep only one default coding, for example `gzip`,
without also keeping `br`. Applications that want to extend the defaults must
repeat the default entries they want to keep. Programmatic configuration must
expose the same distinction: a whole-map setter replaces the current map, while
singular put-style builder methods only add or replace one entry in the current
builder map.

The release notes must mention:

- pre-compressed static content is enabled by default in Helidon 27;
- any backport to an already released line is disabled by default and requires
  explicit opt-in for sidecar lookup, while the connected runtime
  `Accept-Encoding` negotiation fix still applies;
- feature-level pre-compressed settings are inherited by handlers, and
  handler-level or direct-service settings can override them;
- default sidecar suffixes are `br -> br` and `gzip -> gz`;
- existing sidecar files can now be served for requests to the logical resource;
- classpath sidecars must come from the same classpath resource origin by
  default, and cross-origin classpath sidecar sourcing is explicit opt-in;
- direct requests for literal sidecar filenames remain ordinary static-file
  requests unless the application blocks those files itself;
- static content responses merge `Vary: Accept-Encoding` while pre-compressed
  selection is enabled, including identity and conditional responses;
- encoded variants have representation-specific validators, and range requests
  operate on the selected representation bytes;
- Helidon does not verify that sidecar files are fresher than the logical
  resource;
- applications can disable the feature with `pre-compressed-enabled: false`;
- configuring `pre-compressed-encodings` replaces the default suffix map, so
  applications must repeat any default entries they want to keep.

## Implementation Implications

Relevant current code:

- `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java`
  resolves the request path and owns the per-handler record cache and the memory
  cache assigned to that handler.
- `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/BaseHandlerConfigBlueprint.java`
  should carry the handler-level pre-compressed options so direct service
  creation and feature-registered handlers use the same configuration surface.
- `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java`
  should carry feature-level pre-compressed defaults that `StaticContentFeature`
  copies into handler configs unless a handler explicitly overrides them.
- `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/FileSystemContentHandler.java`
  resolves filesystem resources and welcome files.
- `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ClassPathContentHandler.java`
  resolves classpath resources, jar entries, temporary extracted files, and
  welcome files.
- `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerPath.java`,
  `CachedHandlerJar.java`, and `CachedHandlerInMemory.java` send static
  resources and currently own conditional and range handling.
- `webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java`
  skips runtime response encoding when a response already has
  `Content-Encoding`.
- Static content can reach the listener content-encoding setup through
  `ServerRequest.listenerContext().contentEncodingContext()`. The implementation
  should use that listener context, together with the shared negotiation model,
  to decide whether a rejected-identity request can still be satisfied by
  serving the logical resource through runtime response encoding instead of
  returning `406 Not Acceptable`.
- `http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java`
  currently parses `Accept-Encoding` for runtime compression. This change must
  add a shared negotiation model that handles `q=0`, `identity`, wildcard input,
  and deterministic tie handling correctly for both static sidecar selection and
  runtime content encoding. Malformed-value status handling remains a separate
  header-parsing concern, but this feature must not turn malformed input into an
  internal server error.

Preferred implementation shape:

- Add a small static-content representation-selection layer that receives the
  logical resource, request headers, configured suffix map, and a resource lookup
  callback.
- Add or expose shared `Accept-Encoding` parsing and selection support before
  implementing sidecar selection. Static content and runtime compression must not
  use divergent parsers or different `identity`/`q=0` rules. The helper should
  follow the same candidate-ranking shape as `ServerRequestHeaders.bestAccepted`
  for media types; it can be a local helper if `ServerRequestHeaders` is not the
  right public surface.
- The shared negotiation support should expose whether a listener runtime
  encoder can satisfy the request and which coding it would use. Static content
  should not infer this only by comparing `ContentEncodingContext.encoder(...)`
  with `ContentEncoder.NO_OP`, because the selection result is also needed for
  representation metadata decisions.
- Return either the logical resource handler or an encoded variant handler with
  explicit response metadata.
- Keep root traversal protection in the existing filesystem and classpath
  resolution paths. Sidecar lookup must be derived from an already validated
  logical resource.
- Avoid caching an encoded variant under the same cache key as the identity
  resource. The record cache key must include the selected content coding, for
  example `big-file.html#encoding=br`.
- Cache successful sidecar resolutions per logical resource and content coding.
  Do not cache negative lookup results. A later warm request repeats the sidecar
  lookup so a sidecar added after the initial miss becomes available without
  evicting or clearing the handler cache. Successful sidecar entries follow the
  existing static-content record-cache lifecycle and are cleared with that cache,
  such as when the handler cache is released. Create sidecar state only after the
  canonical logical resource exists, so arbitrary missing logical resources do
  not populate sidecar cache state.
- Keep in-memory cache entries representation-specific. A cached Brotli sidecar
  must not satisfy an identity request.
- Add helper support for merging `Vary: Accept-Encoding` without dropping
  existing `Vary` values. Apply it before conditional processing can short-circuit
  into `304 Not Modified`.
- Ensure short-circuit responses carry selected representation metadata through
  the actual response path. If conditional or negotiation handling uses
  `HttpException`, attach `Vary: Accept-Encoding`, selected `ETag`, and other
  applicable representation headers to the exception rather than relying only on
  headers already set on `response.headers()`.

The feature does not require a new public module. It belongs in
`helidon-webserver-static-content`.

## Testing and Validation Coverage

Focused tests should cover:

- filesystem handler selects `.br` over `.gz` when the request prefers Brotli;
- classpath handler selects `.br` without a Brotli provider on the classpath;
- `Content-Type` comes from the logical filename;
- `Content-Encoding`, `Content-Length`, and `Vary` are correct;
- `Vary: Accept-Encoding` is present on sidecar, identity, fallback, HEAD,
  range, and conditional responses while pre-compressed selection is enabled;
- `Content-Encoding` prevents runtime gzip re-encoding;
- conditional requests are evaluated against the selected representation and
  `304` or `412` responses include the selected representation validators and
  `Vary: Accept-Encoding`;
- `If-Range` is evaluated against the selected representation and sidecar ranges
  are served only for a strong exact entity-tag match, or for an exact
  `Last-Modified` match when the handler can establish that the HTTP-date is a
  strong validator; otherwise the full selected representation is sent;
- `pre-compressed-enabled: false` serves the identity resource even when
  sidecars exist;
- absent `Accept-Encoding` serves the identity resource;
- present but empty `Accept-Encoding` serves only the identity resource and does
  not select sidecars or runtime response encoding;
- `q=0` rejects that coding;
- equal-`q` ties prefer client header order; otherwise sidecar ties use
  normalized content-coding name order;
- representation-first negotiation prefers a higher-`q` identity response over
  a lower-`q` sidecar, and a higher-`q` runtime encoding over a lower-`q`
  sidecar;
- an existing sidecar wins over runtime response encoding for the same selected
  concrete content coding;
- `identity` participates in negotiation, including `identity;q=0` returning
  `406 Not Acceptable` when no acceptable sidecar or runtime encoding can be
  produced;
- `*;q=0` rejects otherwise-unlisted codings, while a more specific `identity`
  value overrides the wildcard rule for the logical resource;
- `identity;q=0` can still return the logical resource through runtime response
  encoding when an acceptable listener content encoder is available;
- malformed `Accept-Encoding` input does not result in a static-content
  internal server error;
- runtime-encoded static responses use weak representation-specific `ETag`
  values when possible, logical-resource `Last-Modified`, and conditional
  handling based on those validators;
- HEAD responses for runtime-encoded static content apply encoder headers
  without sending an entity or identity `Content-Length`;
- range requests with runtime response encoding return a full encoded `200 OK`
  response instead of a partial response over plain-file bytes;
- runtime content encoding honors the same `identity`, wildcard, and `q=0`
  negotiation rules as static sidecar selection;
- invalid configured content-coding names are rejected, including `identity`,
  `*`, names with separators, whitespace, controls, or other non-token
  characters, and duplicate coding names after case normalization;
- invalid suffix values are rejected, including empty values and values with
  path separators, and suffix values such as `.gz` are normalized to `gz`;
- feature-level pre-compressed options are inherited by handlers, while
  handler-level and direct-service options can override them;
- a custom handler suffix map replaces the inherited/default map, so a map
  containing only `gzip: gz` does not keep the inherited/default `br` entry;
- wildcard-only requests serve identity when identity is acceptable and equally
  preferred, without selecting a wildcard-derived sidecar or runtime encoder;
- wildcard requests with `identity;q=0` select an available sidecar or listener
  runtime encoder, and return `406` only when neither can satisfy the wildcard;
- no sidecar match falls back to identity only when identity is acceptable;
- a negotiated sidecar is not served unless the canonical logical resource is
  servable by existing static-content rules;
- direct requests for literal sidecar filenames are handled as ordinary static
  file requests independent of canonical logical resource existence;
- welcome files and single-file handlers use the same representation selection;
- requests such as `/docs` redirect to `/docs/` before a pre-compressed welcome
  representation is selected;
- classpath sidecars use same-origin sourcing by default, and can be resolved
  from different classpath entries only when cross-origin classpath sidecar
  sourcing is enabled;
- path-mapped handlers append the sidecar suffix only after path mapping and
  root validation;
- cache entries do not leak encoded responses into identity responses;
- positive sidecar availability cache entries prevent repeated filesystem or
  classpath lookups on warm hits, while warm misses repeat lookup and discover
  sidecars added later;
- `Vary: Accept-Encoding` merging preserves existing `Vary` values;
- filesystem, classpath `file:` URL, JAR, and stream-only classpath resources
  follow the documented `Last-Modified` behavior;
- same-timestamp identity, gzip, and Brotli resources still produce
  representation-specific `ETag` values using selected coding and selected
  length for negotiated sidecars;
- GET, HEAD, conditional requests, and single-range requests work for selected
  sidecar files;
- generated config reference, user docs, and release notes include the
  feature-wide and handler/direct-service disable switches, default-on behavior,
  suffix-map replacement semantics, literal sidecar request behavior,
  `Vary: Accept-Encoding`, representation-specific validators, and range
  behavior over selected representation bytes.

Performance validation should include the HttpArena static-file benchmark from
the issue, comparing:

- current runtime compression;
- identity file response;
- pre-compressed `.gz`;
- pre-compressed `.br`.

Add a custom benchmark focused on the new selection and cache behavior. It
should cover warm cache hits with an existing sidecar, repeated warm sidecar
misses, filesystem resources, classpath JAR resources, Brotli absent with gzip
present, absent `Accept-Encoding`, no-sidecar identity responses, disabled
pre-compressed selection, sidecar miss with `identity;q=0` satisfied by runtime
encoding, and at least one small and one large resource. The default-on
no-sidecar identity path should be compared with feature-disabled or current
baseline behavior for common browser `Accept-Encoding` headers and for absent
`Accept-Encoding`.

The expected result is lower CPU cost and stable `Content-Length` responses for
pre-compressed variants, especially for large static files. The default-on
no-sidecar path should quantify the intentional repeated lookup cost required to
discover newly added sidecars.

## Design Decisions to Preserve

- Pre-compressed files are selected by HTTP content negotiation, not by direct
  route matching on the sidecar filename.
- The selected variant is independent of installed runtime encoders.
- The logical resource remains the resource identity for media type detection
  and not-found behavior.
- Response validators and cache keys must be representation-specific.
- The feature is opt-out, because issue #11647 describes the sidecar behavior as
  the preferred static-content performance path with a disable toggle for
  deployments where it would be incorrect.

tomas-langer:
Resolves #11647

Adds support for pre-compressed static content sidecar resources and response representation negotiation across file-system, classpath, and single-file handlers. The change also tightens `Accept-Encoding` parsing and selection, shares listener response encoders with static content, caches sidecar hits per logical resource while leaving misses uncached so newly added sidecars are discovered by later requests, preserves classpath same-origin sidecar sourcing, and updates docs, tests, and benchmarks.

HttpArena static-content comparison against the baseline before this change:

| Scenario | Before | After | Change | 5xx before/after |
| --- | ---: | ---: | ---: | ---: |
| `static/4096` | 9,651 rps | 174,882 rps | +1,712.06% | 0 / 0 |
| `static-h2/256` | 8,620 rps | 158,360 rps | +1,737.12% | 0 / 0 |

Both baseline and candidate runs reported `Helidon SE 27.0.0-SNAPSHOT`; the comparison used `origin/main` at `1a6c6c8ed5` as the baseline and this branch as the candidate.


http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +131,57 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
-            if (this == o) {
-                return true;
-            }
-            if (o == null || getClass() != o.getClass()) {
-                return false;
-            }
-            EncodingWithQ that = (EncodingWithQ) o;
-            return Double.compare(that.q, q) == 0 && encoding.equals(that.encoding);
-        }
+        return encoders.get(selected.get().coding());
```
danielkec: Here the selected token can be an alias whose encoder emits a different coding. For gzip:

```http
Accept-Encoding: x-gzip, gzip;q=0, identity;q=0
```

`selected` can be `x-gzip`, but the encoder writes `Content-Encoding: gzip`, which the client rejected with `q=0`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/FileBasedContentHandler.java
```diff
@@ -60,64 +68,137 @@ static String fileName(Path path) {
...
         if (headers.contains(HeaderNames.RANGE)) {
-            long contentLength = contentLength(path);
-            List<ByteRangeRequest> ranges = ByteRangeRequest.parse(request,
-                                                                   response,
-                                                                   headers.get(HeaderNames.RANGE).values(),
-                                                                   contentLength);
+            long contentLength = contentLength(path, knownContentLength);
+            List<ByteRangeRequest> ranges;
+            try {
+                ranges = ByteRangeRequest.parse(request,
```
danielkec: Range selection does not check `If-Range` against the selected representation before returning `206`. For a sidecar:

```http
Accept-Encoding: br
Range: bytes=0-3
If-Range: "<identity-etag>"
```

this still serves bytes from `resource.txt.br`, although the validator is not for that selected representation.

webserver/webserver/src/main/java/io/helidon/webserver/http/ErrorHandlers.java
```diff
@@ -168,7 +168,7 @@ private void handleRequestException(ConnectionContext ctx,
         }
         ctx.listenerContext()
```
danielkec, line 176: This only disables automatic encoding for direct handlers. A configured error handler can still reparse the malformed request header when it sends:

```java
error(BadRequestException.class, (req, res, e) -> res.send("bad"));
```

With `Accept-Encoding: g zip`, `res.send` goes through response content encoding and throws the same bad-header exception again.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -272,6 +317,247 @@ String welcomePageName() {
...
+        int q = Double.compare(second.quality().q(), first.quality().q());
+        if (q != 0) {
+            return q;
+        }
+        boolean firstImplicitIdentity = implicitIdentity(first);
+        boolean secondImplicitIdentity = implicitIdentity(second);
+        if (firstImplicitIdentity != secondImplicitIdentity) {
+            return firstImplicitIdentity ? 1 : -1;
+        }
+        if ((first.type() == CandidateType.IDENTITY) != (second.type() == CandidateType.IDENTITY)) {
```
danielkec: This comparator ignores the preserved client order for equal-quality candidates. Examples:

```http
Accept-Encoding: identity, gzip
Accept-Encoding: gzip, br, identity;q=0
```

The issue says equal-`q` ties prefer header order, but this branch makes identity lose to non-identity and later falls back to type/suffix/provider order.

danielkec (commented):
Inline review comments for the issues found.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
tomas-langer: Fixed in d44beb5da9: `ContentEncodingSupportImpl` now checks the coding emitted by the selected encoder headers before accepting an alias, so `x-gzip` is rejected when `gzip;q=0`. Regression coverage: `ContentEncodingSupportTest.testRuntimeEncoderRejectsRejectedEmittedCoding`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/FileBasedContentHandler.java
tomas-langer: Fixed in d44beb5da9: range handling now evaluates `If-Range` against the selected representation by passing that representation ETag, weak flag, and last-modified time into `ByteRangeRequest.parse`. Regression coverage: `StaticContentTest.testFileSystemPreCompressedIfRangeUsesSelectedRepresentationEtag` and `testFileSystemPreCompressedIfRangeAllowsMatchingSelectedRepresentationEtag`.

webserver/webserver/src/main/java/io/helidon/webserver/http/ErrorHandlers.java
tomas-langer: Fixed in d44beb5da9: configured error handlers now disable automatic response content encoding before invoking the handler, matching the direct-handler path and avoiding a second parse of the malformed `Accept-Encoding` header. Regression coverage: `BadRequestTest.testBadRequestErrorHandlerDoesNotReparseMalformedAcceptEncoding`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in d44beb5da9: `StaticContentHandler` now uses the preserved `Accept-Encoding` header order when quality ties remain after quality, concrete-vs-wildcard, and implicit-identity tie-breaks. Regression coverage: `StaticContentHandlerTest.preCompressedExplicitIdentityTieUsesHeaderOrder`, `preCompressedSidecarUsesHeaderOrderWhenQualityTies`, and `preCompressedRuntimeEncodingUsesHeaderOrderWhenQualityTies`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
```diff
@@ -33,6 +36,20 @@ record ByteRangeRequest(long fileLength, long offset, long length) {
     private static final Pattern RANGE_PATTERN = Pattern.compile("(\\d+)?-(\\d+)?(?:, )?");
 
     static List<ByteRangeRequest> parse(ServerRequest req, ServerResponse res, String headerValues, long fileLength) {
+        return parse(req, res, headerValues, fileLength, null, false, null);
```
danielkec: The legacy `parse(req, res, headerValues, fileLength)` path now reaches `ifRangeMatches` and dereferences `req.headers()` even though this overload has no selected-representation metadata. The current head fails the existing range tests with five NPEs before any range assertion runs:

```text
mvn -ntp -pl :helidon-webserver-static-content -Dtest=ByteRangeRequestTest -Dsurefire.failIfNoSpecifiedTests=false test
```

All failures go through `ByteRangeRequest.parse(...:39)` to `ifRangeMatches(...:90)`.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +132,72 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
+            String coding = quality.coding();
+            if (contentEncodingSupported(coding)) {
+                selectableEncodingIds.add(coding);
             }
         }
 
-        @Override
-        public int compareTo(EncodingWithQ o) {
-            return Double.compare(o.q, this.q);
+        Optional<AcceptEncoding.CodingQuality> selected = acceptEncoding.best(List.copyOf(selectableEncodingIds));
```
danielkec: Generic runtime response encoding still uses provider/server order for equal-quality concrete codings. `selectableEncodingIds` is seeded from `contentEncodingIds`, and `AcceptEncoding.best` breaks same-`q` ties by list index; with providers registered as `[br, gzip]`, `Accept-Encoding: gzip, br` chooses `br` (`ContentEncodingSupportTest.testRuntimeEncoderUsesProviderOrderForEqualQualityCodings`).

The linked issue requires runtime content encoding to use the same parsed `Accept-Encoding` semantics as static sidecar selection, where same-`q` ties now prefer client header order.

danielkec (commented):
Follow-up review comments for the current head. I also reproduced the copyright failure locally: `bash ./etc/scripts/copyright.sh` reports `webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java:2: Wrong year. Expected "2022, 2026", is "2022, 2023"`.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -248,7 +255,8 @@ protected byte[] entityBytes(byte[] configuredEntity) {
      */
     protected byte[] entityBytes(byte[] configuredEntity, int position, int length) {
         byte[] entity = configuredEntity;
-        if (contentEncodingContext.contentEncodingEnabled()
+        if (automaticContentEncoding
```
danielkec: This automatic response-encoding path can produce a variant selected by `Accept-Encoding` without `Vary: Accept-Encoding`. The built-in gzip/deflate encoders only add `Content-Encoding` and remove `Content-Length`, and the stream path below does the same.

A cacheable ordinary route requested with `Accept-Encoding: gzip` can therefore be stored without varying on the request header, unlike the new static-content representation path.

docs/src/main/asciidoc/se/webserver/webserver.adoc
```diff
@@ -828,6 +828,63 @@ server:
...
+the logical resource media type and sets `Content-Encoding`. Direct requests for literal sidecar filenames, such as
+`/big-file.html.gz`, are handled as ordinary static-file requests. A sidecar is selected only when the request explicitly
+accepts that concrete coding; `Accept-Encoding: *` does not select pre-compressed sidecars. Sidecar lookup hits and
+misses are cached with the resolved logical resource, so newly added sidecars are discovered after cache eviction or
+handler restart. Classpath sidecar bytes
+cached in memory still count against the configured memory cache capacity.
+
+When listener content encoding is configured, static content can dynamically encode the logical resource when that
+runtime candidate is the best acceptable representation. Wildcard values can select runtime encoders.
+Pre-compressed sidecars win ties over runtime encoders, and non-identity candidates at the same quality win over
```
danielkec: This paragraph still describes the old tie-breaking behavior. Current static-content tests now expect `Accept-Encoding: identity, gzip` to select identity and `Accept-Encoding: gzip, br, identity;q=0` to select gzip, so equal-quality candidates are no longer ordered by non-identity before identity or by `pre-compressed-encodings` before client header order.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
tomas-langer: Fixed in d4760a1052: the legacy `ByteRangeRequest.parse(req, res, headerValues, fileLength)` overload now delegates directly to the range parser and does not evaluate `If-Range`; the selected-representation overload still performs the `If-Range` check. Regression/verification: `ByteRangeRequestTest` is green, including the exact command from this thread.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
tomas-langer: Fixed in d4760a1052: `AcceptEncoding.best` now uses client `Accept-Encoding` header order for equal-quality concrete codings, while preserving provider order as the wildcard/fallback tie-break. Regression coverage: `ContentEncodingSupportTest.testRuntimeEncoderUsesHeaderOrderForEqualQualityCodings` and `testBestWildcardUsesFallbackOrder`.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
tomas-langer: Fixed in d4760a1052: `ServerResponseBase` now merges `Vary: Accept-Encoding` after automatic response encoding in both buffered and output-stream paths. Regression coverage: `ContentEncodingContextTest.testAutomaticContentEncodingAddsVaryAcceptEncoding`, `testAutomaticContentEncodingAddsVaryAcceptEncodingForOutputStream`, and reset-path assertions.

docs/src/main/asciidoc/se/webserver/webserver.adoc
tomas-langer: Fixed in d4760a1052: `docs/src/main/asciidoc/se/webserver/webserver.adoc` now describes equal-quality concrete coding ties as client-header-order based, with sidecar/provider order only as fallback, and also documents `Vary: Accept-Encoding` for automatic response encoding.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +132,72 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
+        if (AcceptEncoding.IDENTITY.equals(selectedCoding)) {
+            return ContentEncoder.NO_OP;
         }
 
-        @Override
-        public int hashCode() {
-            return Objects.hash(encoding, q);
+        ContentEncoder encoder = encoders.get(selectedCoding);
+        String emittedCoding = responseCoding(selectedCoding, encoder);
+        if (!emittedCoding.equals(selectedCoding) && acceptEncoding.match(emittedCoding, true).isEmpty()) {
```
danielkec: This still checks only whether the emitted coding is acceptable, not whether it is the best representation after aliases are resolved. For example:

```http
Accept-Encoding: x-gzip, identity;q=0.75, gzip;q=0.5
```

`best(...)` selects `x-gzip`, but the encoder emits `Content-Encoding: gzip`, so the response uses the lower-preference gzip representation instead of identity.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -272,6 +317,248 @@ String welcomePageName() {
...
 
+    boolean preCompressedCrossOriginSourcingEnabled() {
+        return preCompressedCrossOriginSourcingEnabled;
+    }
+
+    CachedHandler selectHandler(CachedHandler identityHandler,
+                                ServerRequest request,
+                                SidecarResolver sidecarResolver) throws IOException, URISyntaxException {
+        if (!preCompressedEnabled) {
+            return identityHandler;
```
danielkec: With this early return, a handler configured with `preCompressedEnabled(false)` bypasses the selected-representation wrapper and leaves response-layer automatic encoding enabled. On a real server with a gzip runtime encoder, this request shape:

```http
GET /path-disabled/resource.txt
Accept-Encoding: gzip
Range: bytes=0-3
```

returns `206` with `Content-Range` for the identity resource, but also `Content-Encoding: gzip`. That mixes range metadata for one representation with an encoded response body. The existing disabled-path test uses `Accept-Encoding: br` while the test runtime encoder is `gzip`, so it does not exercise this path.

danielkec (commented):
Follow-up review comment for the disabled pre-compressed range path.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
tomas-langer: Fixed in 289b1fae05: `ContentEncodingSupportImpl` now builds response-encoding candidates keyed by the coding the encoder actually emits, then runs `AcceptEncoding.best(...)` over those emitted codings. The `x-gzip, identity;q=0.75, gzip;q=0.5` case now selects identity instead of the alias that would emit lower-preference gzip. Regression coverage: `ContentEncodingSupportTest.testRuntimeEncoderUsesBestEmittedCoding`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in 289b1fae05: when `preCompressedEnabled(false)` and `Range` is present, `StaticContentHandler` now uses an identity representation for the selected handler, disables automatic response content encoding for the partial response, and rejects requests that explicitly refuse identity. Regression coverage: `StaticContentHandlerTest.preCompressedDisabledRangeDisablesAutomaticEncoding`, `preCompressedDisabledRangeRejectsRejectedIdentity`, and `StaticContentTest.testFileSystemPreCompressedDisabledRangeDoesNotRuntimeEncode`.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +132,89 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
-            if (o == null || getClass() != o.getClass()) {
-                return false;
-            }
-            EncodingWithQ that = (EncodingWithQ) o;
-            return Double.compare(that.q, q) == 0 && encoding.equals(that.encoding);
+        ContentEncoder encoder = encoders.get(normalized);
+        if (encoder == null) {
+            return;
         }
+        String emittedCoding = responseCoding(normalized, encoder);
```
danielkec: Here the candidate is stored under the coding emitted by the encoder. For the built-in gzip provider, `GzipEncoding.ids()` advertises `x-gzip`, but the encoder writes `Content-Encoding: gzip`, so the candidate map contains only `gzip`. That means `Accept-Encoding: x-gzip` makes `best(...)` evaluate `gzip` instead of the accepted alias and fall back to identity; `Accept-Encoding: x-gzip, identity;q=0` returns `406` even though the advertised alias is acceptable.

danielkec (commented):
Current-head review findings.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -272,6 +316,303 @@ String welcomePageName() {
...
+                               identityRepresentation,
+                               sidecarResolver);
+    }
+
+    private static List<AcceptEncoding.CodingQuality> sidecarQualities(AcceptEncoding acceptEncoding,
+                                                                       ContentEncodingContext contentEncodingContext,
+                                                                       String coding,
+                                                                       Set<String> sidecarCodings) {
+        List<AcceptEncoding.CodingQuality> result = new ArrayList<>();
+        acceptEncoding.match(coding, true).ifPresent(result::add);
```
danielkec, line 615: Passing `true` here lets `*` match a sidecar. With `asset.js`, `asset.js.br`, no runtime encoder, and `Accept-Encoding: *, identity;q=0`, the handler selects the Brotli sidecar and returns `200`. The linked design limits wildcard matching to listener runtime encoders, so this request has no acceptable representation and should return `406`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
```diff
@@ -69,6 +89,22 @@ static List<ByteRangeRequest> parse(ServerRequest req, ServerResponse res, Strin
...
+
+        String ifRange = req.headers().get(HeaderNames.IF_RANGE).get().trim();
+        if (ifRange.startsWith("\"") || StaticContentHandler.isWeakETag(ifRange)) {
+            return !weakEtag
+                    && etag != null
+                    && !StaticContentHandler.isWeakETag(ifRange)
+                    && StaticContentHandler.unquoteETag(ifRange).equals(StaticContentHandler.unquoteETag(etag));
+        }
+
+        return false;
```
danielkec, line 131: This rejects every HTTP-date form of `If-Range`. Reproducer: request a selected sidecar, copy its emitted `Last-Modified`, then repeat the request with `Range: bytes=0-3` and `If-Range: <copied value>`. Even though the representation is unchanged, the range is ignored and the response is a full `200`; the linked design says a matching selected-representation `Last-Modified` authorizes `206`.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -577,6 +578,16 @@ public void run() {
                 throw writeFailure;
             }
             // no sense in throwing an exception, as this is invoked from an executor service directly
+        } catch (BadRequestException e) {
+            if (!handleRequestException(RequestException.builder()
```
danielkec: This new path exposes the current HTTP/2 direct-handler send behavior, which uses the exception status and message instead of the selected `TransportResponse` status and entity. With a BAD_REQUEST direct response configured as `418`/`custom-direct`, an HTTP/2 request containing malformed `Accept-Encoding` returns `400` with `Invalid Accep` instead of the configured response.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ResponseRepresentation.java
```diff
@@ -0,0 +1,124 @@
...
+            return;
+        }
+        if (contentEncoding != null) {
+            headers.set(HeaderValues.create(HeaderNames.CONTENT_ENCODING, true, false, contentEncoding));
+        }
+    }
+
+    void apply(HttpException exception) {
+        if (varyAcceptEncoding) {
+            exception.header(HeaderValues.create(HeaderNames.VARY, true, false, HeaderNames.ACCEPT_ENCODING_NAME));
```
danielkec, line 84: This adds a bare exception `Vary` header. When the direct-handler bridge reapplies exception headers, `ServerResponse.header(...)` replaces the existing value. A conditional sidecar response that already has `Vary: Origin` therefore finishes with only `Vary: Accept-Encoding`, losing the `Origin` cache dimension.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +195,50 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
-    static class EncodingWithQ implements Comparable<EncodingWithQ> {
-        private final String encoding;
-        private final double q;
 
-        EncodingWithQ(String encoding, double q) {
-            this.encoding = encoding;
-            this.q = q;
+        Optional<AcceptEncoding.CodingQuality> selected = acceptEncoding.best(contentEncodingIds);
+        if (selected.isEmpty()) {
+            throw new HttpException("No acceptable response content encoding", Status.NOT_ACCEPTABLE_406, true);
```
danielkec: This exception exits before `ServerResponseBase` merges `Vary: Accept-Encoding`. With gzip enabled on the listener, an ordinary route requested using `Accept-Encoding: zstd, identity;q=0` returns `406` without `Vary`, while the same URL returns `200` to an identity-capable client. A shared cache can therefore reuse the negotiated error for the other request variant.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +195,50 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
-     *
-     * @param acceptEncoding comma-separated list of encodings
-     * @return sorted list of encodings
-     */
-    static List<EncodingWithQ> encodings(String acceptEncoding) {
-        String[] values = acceptEncoding.split(",");
-        List<EncodingWithQ> supported = new ArrayList<>(values.length);
-        for (String value : values) {
-            supported.add(EncodingWithQ.parse(value));
+        if (!contentEncodingEnabled()) {
```
danielkec: This early return, together with the response layer's `contentEncodingEnabled()` gate, skips negotiation when no response encoders are configured. An ordinary route requested with `Accept-Encoding: identity;q=0` then sends an identity `200` even though the request accepts no available representation. The linked runtime-negotiation contract explicitly includes this no-provider case.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java
```diff
@@ -151,4 +151,26 @@ interface StaticContentConfigBlueprint extends Prototype.Factory<StaticContentFe
...
     Set<String> sockets();
 
+    /**
+     * Whether pre-compressed sidecar resources should be selected by default for feature-registered handlers.
+     *
+     * @return whether pre-compressed sidecar resources should be used
+     */
+    @Option.Configured
+    @Option.DefaultBoolean(true)
+    boolean preCompressedEnabled();
```
danielkec, line 163: The feature-level sidecar defaults remain incomplete here: the linked configuration design also specifies `pre-compressed-cross-origin-sourcing-enabled`, but that option exists only on `ClasspathHandlerConfig`. With multiple classpath handlers, the documented top-level value cannot provide an inherited default and the setting must be repeated for each handler.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/SingleFileContentHandler.java
```diff
@@ -127,4 +163,21 @@ private Optional<Path> contentPath(Path path) {
...
+
+    private Optional<Path> sidecarPath(Path sidecar) {
+        Path pinnedPath = realPath.get();
+        if (pinnedPath == null || !Files.exists(sidecar)) {
+            return Optional.empty();
+        }
+
+        try {
+            Path resolvedSidecar = sidecar.toRealPath();
+            if (resolvedSidecar.startsWith(pinnedPath.getParent())) {
```
danielkec: The pinned parent is broader than the configured single file. With `/srv/public.txt`, a server-readable sibling `/srv/secret.bin`, and `/srv/public.txt.gz -> /srv/secret.bin`, `Accept-Encoding: gzip, identity;q=0` returns `Secret content` as the gzip representation. The existing coverage rejects targets outside the parent, but this exposes a different sibling through the new sidecar lookup.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/ContentEncodingSupportImpl.java
```diff
@@ -116,100 +195,50 @@ public ContentDecoder decoder(String encodingId) throws NoSuchElementException {
...
-        for (EncodingWithQ encodingWithQ : supported) {
-            if ("*".equals(encodingWithQ.encoding)) {
-                return firstEncoder;
-            }
-            if (contentEncodingSupported(encodingWithQ.encoding)) {
-                return encoders.get(encodingWithQ.encoding);
-            }
+        AcceptEncoding acceptEncoding = AcceptEncoding.create(headers);
+        if (!acceptEncoding.valid()) {
+            throw new HttpException("Invalid Accept-Encoding header", Status.BAD_REQUEST_400);
```
danielkec, line 211: Follow-up to the resolved HTTP/2 direct-handler thread: changing this to a generic `HttpException` avoids that sender path, but it now bypasses configured `BAD_REQUEST` handling altogether. With a BAD_REQUEST handler returning `418`/`custom-bad`, a normal route and `Accept-Encoding: g zip` return the default `400`/`Invalid Accept-Encoding header` on both HTTP/1.1 and HTTP/2. `ErrorHandlers` maps generic `HttpException` to `OTHER`, although an invalid header is a BAD_REQUEST event.

http/encoding/encoding/src/main/java/io/helidon/http/encoding/AcceptEncoding.java
```diff
@@ -0,0 +1,672 @@
...
+        int q = Integer.compare(bestCandidate.q(), identityQ);
+        if (q > 0) {
+            return Optional.of(codingQuality(bestCoding, bestEntry, bestCandidate.q(),
+                                           bestCandidate.order(), bestCandidate.wildcard()));
+        }
+        if (q < 0
+                || (identityOrder != IMPLICIT_IDENTITY_ORDER && identityOrder < bestCandidate.order())) {
+            return Optional.of(identityQuality(identity));
+        }
+        return Optional.of(codingQuality(bestCoding, bestEntry, bestCandidate.q(),
```
danielkec, line 304: This equal-quality fallback selects a wildcard-derived runtime encoder over implicit identity. With one test encoder and `Accept-Encoding: *`, an ordinary response is `Content-Encoding: test` with body `encoded:Hi`. #11647 and the new user guide say omitted identity must win over a runtime encoder matched only by `*`; the static-content comparator implements that rule, but this shared runtime path does not.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
```diff
@@ -33,6 +33,26 @@ record ByteRangeRequest(long fileLength, long offset, long length) {
...
+                                        ServerResponse res,
+                                        String headerValues,
+                                        long fileLength,
+                                        String etag,
+                                        boolean weakEtag) {
+        if (!ifRangeMatches(req, etag, weakEtag)) {
+            return List.of();
+        }
+
+        return parseRanges(req, res, headerValues, fileLength);
```
danielkec: The parser reached through this new selected-representation path still mishandles oversized suffix ranges. For a 50-byte sidecar, `Range: bytes=-500` calculates offset `-450`; the file path then calls `channel.position(-450)` and the in-memory path uses a negative array index. A suffix range longer than the representation should cover bytes `0-49`, so selected-sidecar range handling fails for this valid request.

docs/src/main/asciidoc/se/webserver/webserver.adoc
```diff
@@ -952,6 +952,75 @@ server:
...
+`big-file.html.br` and `big-file.html.gz` after the logical resource has been resolved by the normal static-content
+rules.
+
+The feature is enabled by default. The default sidecar suffix map is `br -> br` and `gzip -> gz`. A custom
+`pre-compressed-encodings` map replaces the defaults, so include `br` and `gzip` explicitly if they should remain
+available. An explicitly empty `pre-compressed-encodings` map disables sidecar lookup for that configuration level.
+Map keys must be concrete content codings, not `identity` or `*`, and must be valid HTTP tokens. Map values are file
+suffixes; Helidon ignores leading dots and rejects values that contain path separators.
+
+When pre-compressed negotiation runs, Helidon merges `Vary: Accept-Encoding`. When a sidecar is selected, Helidon sends
```
danielkec: This new feature section describes selection but omits the representation semantics required by #11647: selected-sidecar `Content-Length`, `Last-Modified`, and ETag; conditional response metadata; byte ranges over the selected bytes; and `If-Range` behavior. The existing documentation thread only covered equal-quality ordering, so it does not explain this gap.

danielkec (commented):
Current-head follow-up review findings.

tests/benchmark/jmh/src/main/java/io/helidon/webserver/staticcontent/StaticContentPreCompressedJmhTest.java
```diff
@@ -0,0 +1,591 @@
...
+    private Path benchmarkDirectory;
+    private Path identityPath;
+    private Path sidecarPath;
+
+    @Setup
+    public void setup() throws IOException, URISyntaxException {
+        benchmarkDirectory = Files.createTempDirectory("static-content-pre-compressed-jmh-");
+        identityPath = benchmarkDirectory.resolve(RESOURCE);
+        sidecarPath = benchmarkDirectory.resolve(RESOURCE + ".br");
+        Files.writeString(identityPath, "Content", StandardCharsets.UTF_8);
```
danielkec: This benchmark creates only a tiny identity file and a Brotli sidecar. The #11647 performance acceptance matrix is still unverified: runtime compression versus identity, `.gz`, and `.br`; JAR resources; Brotli-absent/gzip-present fallback; warm sidecar hit and negative miss; browser-style headers; and small/large resources. The PR-body HttpArena figures compare baseline and candidate endpoints, but do not isolate those representations.

tomas-langer: I will not expand the benchmark matrix in this PR. A follow-up issue can be created if this additional coverage is needed.

tomas-langer:
Temporarily converting to a draft, I will do preparatory PRs for smaller parts first, then re-open this one.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/SidecarCache.java
```diff
@@ -0,0 +1,88 @@
...
+        ReentrantLock resolutionLock = resolutionLocks.computeIfAbsent(coding, _ -> new ReentrantLock());
+        resolutionLock.lock();
+        try {
+            cachedHandler = reusable(coding);
+            if (cachedHandler != null) {
+                return Optional.of(cachedHandler);
+            }
+
+            Optional<CachedHandler> resolved = resolver.resolve(coding, suffix);
+            resolved.ifPresent(handler -> entries.put(coding, handler));
```
danielkec, line 83: `Optional.empty()` is not stored here, so two warm requests for a logical resource without a sidecar invoke the filesystem or classloader resolver twice. This contradicts #11647 and the PR body's negative-availability caching requirement, making the default-on no-sidecar path repeat lookups:

```java
cache.resolve("br", "br", miss);
cache.resolve("br", "br", miss);
assertEquals(1, lookups.get()); // actual: 2
```

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
```diff
@@ -72,37 +97,42 @@ static List<ByteRangeRequest> parse(ServerRequest req, ServerResponse res, Strin
...
-                                            false,
-                                            "bytes " + offset + "-" + last + "/" + fileLength));
-        response.contentLength(length);
-        response.status(Status.PARTIAL_CONTENT_206);
+    private static boolean ifRangeMatches(ServerRequest req, String etag, boolean weakEtag) {
+        if (!req.headers().contains(HeaderNames.IF_RANGE)) {
+            return true;
+        }
+
+        String ifRange = req.headers().get(HeaderNames.IF_RANGE).get().trim();
```
danielkec: With repeated `If-Range` field lines, this reads only the first value and can authorize `206` from one validator while ignoring the conflicting value. An exact-head probe produced:

```text
If-Range: "tag"
If-Range: "other"
Range: bytes=0-0
selected ETag: "tag"
valueCount=2 -> [offset=0, length=1]
```

`If-Range` admits one entity-tag or HTTP-date, so this combined value does not match and the range should be ignored rather than returning partial content.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -162,6 +165,21 @@ public void entityBeforeSend(Runnable listener) {
         beforeSend.add(Objects.requireNonNull(listener));
     }
 
+    @Override
+    public T automaticContentEncoding(boolean enabled) {
+        ensureContentEncodingConfigurable();
+        this.automaticContentEncoding = enabled;
```
danielkec, line 193: This new opt-out state is not honored by the existing `304` path. A route using `automaticContentEncoding(false)` returns an identity `200` without `Vary`, but the matching conditional request returns `304` with `Vary: Accept-Encoding`:

```text
GET                         -> 200, no Vary
GET + matching If-None-Match -> 304, Vary: Accept-Encoding
```

A `304` must carry the `Vary` value that the corresponding `200` would send, so the conditional response can update cache metadata with a variance the selected representation does not have.

danielkec (commented):
Current-head follow-up review findings.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ResponseRepresentation.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        return validator + ";encoding=" + contentEncoding + ";length=" + contentLength;
+    }
+
+    boolean etagRequiresContentLength() {
+        return contentEncoding != null && !runtimeEncoded();
+    }
+
+    boolean weakEtag() {
+        return contentEncoding != null;
```
danielkec: When `contentEncoding` denotes a selected sidecar, this marks its validator weak even though the selected bytes are static. A minimal current-head sequence is:

```text
GET /path/resource.txt
Accept-Encoding: br
-> ETag: W/"..."

GET /path/resource.txt
Accept-Encoding: br
Range: bytes=2-5
If-Range: W/"..."
-> 200, without Content-Range
```

Because `If-Range` uses strong comparison, a client can never use Helidon's emitted sidecar ETag to resume with `206`; #11647 requires sidecar ETags to remain strong when the underlying handler would emit a strong validator.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java
```diff
@@ -151,4 +151,36 @@ interface StaticContentConfigBlueprint extends Prototype.Factory<StaticContentFe
...
+     * handler-level entries replace these entries rather than merging with them, an explicit empty list disables sidecar
+     * lookups by default, codings must be unique concrete valid HTTP tokens other than {@code identity} and {@code *},
+     * and suffixes have leading dots ignored and must not contain path separators.
+     *
+     * @return ordered content coding and file suffix entries
+     */
+    @Option.Configured
+    @Option.DefaultMethod(type = StaticContentConfigSupport.class, value = "defaultPreCompressedEncodings")
+    @Option.Singular("preCompressedEncoding")
+    List<PreCompressedEncodingConfig> preCompressedEncodings();
```
danielkec: #11647 specifies `pre-compressed-encodings` as an ordered coding-to-suffix map with whole-map and put-style builder methods, but this prototype generates list/add-style APIs instead. A minimal configuration using the specified shape is:

```yaml
pre-compressed-encodings:
  br: br
```

Building from this configuration leaves both nested `coding` and `suffix` unset and fails validation. The mirrored handler-level `Optional<List<...>>` surface has the same incompatibility.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -410,6 +472,7 @@ protected void beforeSend() {
         beforeSend.forEach(Runnable::run);
         if (status().code() == Status.NOT_MODIFIED_304.code()
                 && contentEncodingContext.contentEncodingEnabled()
+                && automaticContentEncoding
```
danielkec, line 572: The `automaticContentEncoding(false)` case is fixed, but an explicit encoder still leaves this condition true before the encoder applies its headers. A minimal ETag route using `res.contentEncoder(testEncoder)` produces:

```text
200  Content-Encoding: test      (no Vary)
304  Content-Encoding: test
     Vary: Accept-Encoding
```

The validator-matching `304` therefore publishes cache-key metadata absent from its corresponding `200`, even though the representation is selected explicitly rather than negotiated from the request.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
```diff
@@ -231,7 +243,7 @@ public OutputStream outputStream() {
         if (noEntityResponse) {
             return new ApplicationOutputStream(outputStream, outputStream);
         }
-        OutputStream encodedOutputStream = contentEncode(outputStream);
+        OutputStream encodedOutputStream = contentEncode(outputStream, allowAutomaticEncoding);
```
danielkec, line 275: For an HTTP/2 response with both a stream filter and an explicit encoder, a `204` takes the `noEntityResponse` return above before this call can apply the encoder's headers. A minimal route is:

```java
res.streamFilter(out -> out);
res.contentEncoder(testEncoder);
res.status(Status.NO_CONTENT_204).send();
```

The HTTP/2 response omits `Content-Encoding: test`; equivalent HTTP/1 and unfiltered HTTP/2 paths retain it, so the explicit encoder contract depends on the protocol and response path.

danielkec (commented):
Current-head follow-up review findings.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java
tomas-langer: a map is not ordered, that is the reason this was changed. If you have more than one, you cannot distinguish which was specified first.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
tomas-langer: So you intentionally configure something that creates a response, and then set `NO_CONTENT`? We are not responsible of fixing user errors - if user configures data and then sets a wrong status, it is a problem of the user.

We must support this: `res.status(204).send()` that uses any components configured on the server or routing. If that works, it is sufficient.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java
tomas-langer: Checking if order is needed, if not, will use map and update documentation.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ClassPathContentHandler.java
```diff
@@ -107,18 +136,24 @@ boolean doHandle(Method method, String requestedPath, ServerRequest request, Ser
         }
 
         // we have a resource that we support, let's try to use one from the cache
-        Optional<CachedHandler> cached = cacheHandler(requestedResource);
+        Optional<CachedHandler> cached = handlerCache().get(requestedResource);
```
danielkec: When `cachedFiles("cached.txt")` is used with a bounded record cache, evicting this metadata wrapper makes the next request resolve the classpath source again even though the configured snapshot is still in `MemoryCache`; it then serves the current source bytes instead of the cached snapshot. A minimal sequence is:

```text
cachedFiles: cached.txt
record-cache-capacity: 1
GET cached.txt  -> Cached snapshot
GET other.txt   -> evicts the wrapper
replace cached.txt contents
GET cached.txt  -> Changed source (expected Cached snapshot)
```

This makes explicit in-memory caching depend on record-cache eviction and regresses the existing cache semantics.

danielkec (commented):
Current-head follow-up review finding.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java
```diff
@@ -151,4 +151,38 @@ interface StaticContentConfigBlueprint extends Prototype.Factory<StaticContentFe
...
 
+    /**
+     * Whether pre-compressed sidecar resources should be selected by default for feature-registered directory and
+     * classpath handlers; this is disabled by default and does not apply to file system handlers configured with a single
+     * file, which require handler-level opt-in.
+     *
+     * @return whether pre-compressed sidecar resources should be used
+     */
+    @Option.Configured
+    @Option.DefaultBoolean(false)
```
danielkec, line 162: With no configuration, this makes pre-compressed selection opt-in on the unreleased 27.x main line:

```java
StaticContentConfig.builder().buildPrototype().preCompressedEnabled() // false
```

#11647 requires the feature to default to enabled for Helidon 27 and reserves a false default for backports to already released lines. Existing static-content configurations therefore do not receive the issue's default-on sidecar behavior.

danielkec (commented):
Current-head review finding.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigBlueprint.java
tomas-langer: I have decided to change that. It is my issue.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
```diff
@@ -150,26 +152,32 @@ public void send(byte[] entityBytes, int position, int length) {
             byte[] actualBytes = entityBytes;
 
             Status responseStatus = status();
+            long configuredHeadLength;
             if (noEntityResponse) {
                 normalizeNoEntityHeaders(headers, responseStatus);
+                configuredHeadLength = headRequest ? headers.contentLength().orElse(-1) : -1;
+                entityBytes(BufferData.EMPTY_BYTES);
```
danielkec, line 159: When a no-entity HTTP/2 response uses an explicit encoder whose header callback sets a framing header, this call runs after `normalizeNoEntityHeaders`, so the encoder can put `Content-Length` back on a `204`. A minimal test encoder produced:

```text
status: 204
Content-Length: 23
```

`Content-Length` is forbidden on a `204`, so the emitted response no longer satisfies the no-entity framing invariant.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -559,4 +704,106 @@ private void doSend(Object entity) {
...
+        }
+
+        @Override
+        public void flush() throws IOException {
+            outputStream().flush();
+        }
+
+        @Override
+        public void close() throws IOException {
+            outputStream().close();
```
danielkec, line 750: When an explicit gzip encoder is deferred under `204`, closing this stream closes the raw response stream without instantiating or finalizing gzip, and without marking the encoder discarded. A minimal sequence is:

```text
status(204)
out = outputStream()
out.close()
status(200)
commit()
```

The resulting `200` retains `Content-Encoding: gzip` but has a zero-byte non-gzip body; opening it with `GZIPInputStream` fails with `EOFException`. The later status transition is currently accepted because no encoded bytes were written.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentConfigSupport.java
```diff
@@ -17,16 +17,77 @@
...
+    }
+
+    private static MediaType createContentTypes(Config config) {
+        return config.asString()
+                .map(MediaTypes::create)
+                .get();
+    }
+
+    private static String normalizeCoding(String configured) {
+        String coding = configured.trim().toLowerCase(Locale.ROOT);
```
danielkec, line 62: Trimming before validation lets configured content-coding names containing surrounding whitespace pass as valid. For example:

```text
normalizePreCompressedEncodings(Map.of(" gzip ", "gz"))
// actual: {gzip=gz}; expected: IllegalArgumentException
```

#11647 explicitly requires configured coding names containing whitespace to be rejected, so this silently accepts a configuration typo.

danielkec (commented):
Current-head follow-up review findings.

danielkec (approved):

PR #12060, opened 2026-06-01
Remove global tracer ownership (27.x)

### LINKED ISSUE #11485: Resolve deprecations in Tracing (27.x)
- [ ] io.helidon.tracing.Span#baggage:152
- [ ] io.helidon.tracing.Span#baggage:162
- [ ] io.helidon.tracing.Tracer#global:41
- [ ] io.helidon.tracing.Tracer#global:54

Expanding the scope of this issue to also remove deprecated opentracing modules and APIs.

`tracer-resolver`
`provider-zipkin`
`provider-opentracing`
`provider-jaeger`

tomas-langer:
Resolves #11485

Removes deprecated global tracer ownership paths and routes OpenTelemetry/Tracer ownership through the service registry. Adds explicit OpenTelemetry global publication config, updates tracing/telemetry docs and archetype output, and adds registry ownership tests plus JMH coverage.


telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetry.java
```diff
@@ -66,26 +63,16 @@ static HelidonOpenTelemetry create(Consumer<OpenTelemetryConfig.Builder> consume
...
-    static void global(OpenTelemetry openTelemetry, String serviceName, Map<String, String> tags) throws IllegalStateException {
-        GlobalOpenTelemetry.set(openTelemetry);
-        var otelTracer = openTelemetry.getTracer(serviceName);
-        var helidonTracer = io.helidon.tracing.providers.opentelemetry.HelidonOpenTelemetry.create(openTelemetry,
-                                                                                                   otelTracer,
-                                                                                                   tags);
-        Tracer.global(helidonTracer);
-        Services.set(OpenTelemetry.class, openTelemetry);
+    @Deprecated(forRemoval = true, since = "27.0.0")
+    static void global(OpenTelemetry openTelemetry, String serviceName, Map<String, String> tags) {
```
tjquinno, line 75: Consider logging a warning here. Although use of this deprecated method would be flagged by an IDE or when building a project, that might go unnoticed or be invisible to a downstream project using a library that invokes this method. 

Turning a call of this method into a no-op silently would give a consumer of a library containing such a call no hint that something that used to do something no longer does, possibly hiding an explanation for a change in behavior.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -52,15 +52,32 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
...
+     * <p>
+     * This flag is honored when this configuration is resolved by the Helidon service registry. Direct builder use creates
+     * the configured instance only; it does not publish that instance to the service registry.
      *
-     * @return true if the configured instance should be made global; false otherwise
+     * @return true if the configured instance should be application-wide when resolved by the service registry; false otherwise
      */
     @Option.Configured
     @Option.DefaultBoolean(true)
     boolean global();
```
tjquinno: Even though we will have the following setting `global-open-telemetry` that is clearly related to OTel's global instance, keeping `global` for this setting is potentially confusing. The PR is changing what `global` means; it is clearly not acting as global, it is application-wide. 

Using `registered` (or something else) instead would reinforce the direct connection to it being placed into the service registry as the app-wide instance and would remove the possible ambiguity about `global`.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -52,15 +52,32 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
...
+     * Whether the {@link io.opentelemetry.api.OpenTelemetry} instance created from this configuration should also be
+     * published as the OpenTelemetry global instance when it becomes the application-wide Helidon service and before any
+     * other code initializes {@link io.opentelemetry.api.GlobalOpenTelemetry}, because OpenTelemetry globals are JVM-wide
+     * and can be assigned only once.
+     * <p>
+     * This is disabled by default.
+     *
+     * @return true if the configured application-wide instance should be published to OpenTelemetry global; false otherwise
+     */
+    @Option.Configured("global-open-telemetry")
```
tjquinno: Why explicitly set the key that is the default anyway?

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
tjquinno, line 138: Similar comment here as with `telemetry.global`. We are no longer treating this setting as "global" but as "registered as application-wide." Having this key and the one for `telemetry` be the same--such as `registered`--emphasizes how we actual treat it.

docs/src/main/asciidoc/se/telemetry/open-telemetry.adoc
```diff
@@ -228,7 +241,10 @@ include::{rootdir}/config/io_helidon_telemetry_otelconfig_HelidonOpenTelemetry.a
 Notes:
 
 * OpenTelemetry uses default propagators of `tracecontext` and `baggage`. (See the `otel.propagators` property in link:https://opentelemetry.io/docs/languages/java/configuration/#properties-general[this OpenTelemetry guide].)
-* Setting `global` to `true` has the effect described in the <<effects-of-setting-global, section>> about global instances.
+* Setting `global` to `true` has the effect described in the <<effects-of-setting-global, section>> about application
+  instances in the Helidon service registry.
+* Setting `global-open-telemetry` to `true` also publishes that application instance to `GlobalOpenTelemetry` for
+  OpenTelemetry instrumentation that reads the JVM-wide OpenTelemetry global.
```
tjquinno, line 250: The `using-opentelemetry-api` doc page below  uses the term "active" which applies here as well, I think?

docs/src/main/asciidoc/se/telemetry/using-opentelemetry-api.adoc
```diff
@@ -62,7 +64,16 @@ If you must use the OpenTelemetry API from your code, Helidon recommends this op
...
 
-This approach prepares the OpenTelemetry runtime using OpenTelemetry's autoconfiguration instead of Helidon's integration with OpenTelemetry, but from there the programmatic use of the OpenTelemetry API is essentially the same in those two options: obtain the global `OpenTelemetry` instance and then use it to create and update spans and metrics. In this option your code must use the `GlobalOpenTelemetry` type; Helidon cannot provide the global instance via injection or Helidon services look-up.
+This approach prepares the OpenTelemetry runtime using OpenTelemetry's autoconfiguration instead of Helidon's integration with OpenTelemetry, but from there the programmatic use of the OpenTelemetry API is essentially the same in those two options: obtain the `OpenTelemetry` instance and then use it to create and update spans and metrics.
+If the external setup has already installed `GlobalOpenTelemetry` before Helidon resolves `OpenTelemetry`, and no Helidon
+telemetry or tracing owner is active, service look-up adopts that global automatically.
```
tjquinno, line 69: The page below does not define what it means, and the ownership strategy's `active` method has a very specific meaning: IIUC both `global` and `enabled`. 

Probably good to explain that here.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
```diff
@@ -52,15 +52,32 @@ interface OpenTelemetryConfigBlueprint extends Prototype.Factory<HelidonOpenTele
...
-     * @return true if the configured instance should be made global; false otherwise
+     * @return true if the configured instance should be application-wide when resolved by the service registry; false otherwise
      */
     @Option.Configured
     @Option.DefaultBoolean(true)
     boolean global();
 
+    /**
+     * Whether the {@link io.opentelemetry.api.OpenTelemetry} instance created from this configuration should also be
+     * published as the OpenTelemetry global instance when it becomes the application-wide Helidon service and before any
```
tjquinno: The PR goes to some lengths to decouple the registration of the `OpenTelemetry` instance in our registry from its "global-ness" in the OpenTelemetry API, but here we are saying that for the configured instance to be set as the OTel global one both this setting _and_ `global` (or some alternative such as `registered`) must be `true`.

It would be clearer to keep the two effects separate here as well, so `global` (`registered`) controls only registration in our service registry and `global-open-telemetry` alone controls the OTel global behavior. 

I am not sure if many users would set `global` (`registered`) to `false` and set `global-open-telemetry` to `true`, but treating them separately would be consistent with the PR's behavior elsewhere.

tjquinno (changes requested):
Some specific notes below.

Also, we might think about whether a hypothetical _metrics_ implementation using OpenTelemetry might also implement the ownership strategy and, if so, what that might imply for what module should own the ownership strategy interface.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/HelidonOpenTelemetry.java
tomas-langer: Added the warning. The deprecated method remains a no-op for the removed Helidon global tracer behavior, but it now logs at `WARNING` so downstream users can see that an old library call no longer changes tracing state.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
tomas-langer: Changed. `telemetry.registered` now controls whether the config contributes the application-wide service-registry owner, and `telemetry.global` keeps the OpenTelemetry meaning: publishing the selected application `OpenTelemetry` to `GlobalOpenTelemetry`.

The temporary `global-open-telemetry` key was removed.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
tomas-langer: Removed the explicit key. `global-open-telemetry` is gone; `global` now covers OpenTelemetry global publication again.

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprint.java
tomas-langer: Changed the tracing side the same way. `tracing.registered` controls whether tracing config contributes an application-wide ownership candidate, and `tracing.global` keeps the OpenTelemetry global-publication meaning.

docs/src/main/asciidoc/se/telemetry/open-telemetry.adoc
tomas-langer: Updated. The docs now use `registered` for the active service-registry owner and `global` only for publishing to `GlobalOpenTelemetry`.

docs/src/main/asciidoc/se/telemetry/using-opentelemetry-api.adoc
tomas-langer: Clarified this. The page now explains that a Helidon owner is active when a `telemetry` or `tracing` configuration section is enabled and its `registered` setting is `true`.

telemetry/opentelemetry-config/src/main/java/io/helidon/telemetry/otelconfig/OpenTelemetryConfigBlueprint.java
tomas-langer: Changed by removing the extra key and restoring the original meaning of `global`. The two effects are now separated as `registered` for the Helidon service-registry owner and `global` for OpenTelemetry `GlobalOpenTelemetry` publication.

tomas-langer:
For the review summary note about metrics/logging reuse: the shared OpenTelemetry ownership selection, cache, lifecycle, and global-publication handling now lives in `telemetry/opentelemetry`, and the ownership strategy contract moved to `io.helidon.telemetry.opentelemetry.spi`. That keeps the future metrics/logging path out of the tracing provider module.

The module itself is not marked as a preview feature, so classpath/module presence does not emit a preview feature warning. The exported public types are marked `@Api.Preview` so direct user use is still discouraged.

tjquinno:
Updates look good. Waiting for fixes to pipeline failures to approve.

tjquinno (approved):

PR #12064, opened 2026-06-02
Fix JSON stream parser buffer compaction

### LINKED ISSUE #12062: 4.x: Default buffer size reading JSON cannot be configured, resulting in error with bigger JSONs
Helidon Version: 4.x and main

## Enhancement Description

Currently, when JSON is read in JsonValueReader, we cannot define the buffer size, and it defaults to JsonParserStream#DEFAULT_BUFFER_SIZE = 512.

See:
https://github.com/helidon-io/helidon/blob/77190e6701f919af181ee53f1706140f605402f8/http/media/json/src/main/java/io/helidon/http/media/json/JsonValueReader.java#L56
https://github.com/helidon-io/helidon/blob/77190e6701f919af181ee53f1706140f605402f8/http/media/json/src/main/java/io/helidon/http/media/json/JsonValueReader.java#L62

Default is too low, and it needs to be configurable.

Reproducer:

```
@ServerTest
class JsonMediaHttpClientTest {
    private static final String SYNTHETIC_PADDING = "x".repeat(480);
    private static final String RESPONSE = "{\"syntheticPadding\":\""
            + SYNTHETIC_PADDING
            + "\",\"after\":\"ok\"}";

    private final WebClient client;

    JsonMediaHttpClientTest(WebClient client) {
        this.client = client;
    }

    @SetUpRoute
    static void route(HttpRouting.Builder routing) {
        routing.get("/json", (req, res) -> res.header(HeaderValues.CONTENT_TYPE_JSON)
                .send(RESPONSE));
    }

    @Test
    void shouldParseJsonResponseWithStringValueEndingAtBufferBoundary() {
        try (HttpClientResponse response = client.get("/json").request()) {
            JsonObject json = response.as(JsonObject.class);

            assertThat(json.stringValue("after", ""), is("ok"));
        }
    }
}

```

```
  java.lang.NegativeArraySizeException: -510
      at io.helidon.json.JsonParserStream.readJsonString(JsonParserStream.java:876)
      at io.helidon.json.JsonParserBase.readJsonObject(JsonParserBase.java:73)
      at io.helidon.json.JsonParserBase.readJsonValue(JsonParserBase.java:36)
      at io.helidon.http.media.json.JsonValueReader.read(JsonValueReader.java:57)
      at io.helidon.http.media.json.JsonValueReader.read(JsonValueReader.java:52)
      at io.helidon.http.media.json.JsonValueReader.read(JsonValueReader.java:33)
      at io.helidon.http.media.MediaContextImpl$CloseStreamReader.read(MediaContextImpl.java:255)
      at io.helidon.webclient.api.ClientResponseEntity.entityAs(ClientResponseEntity.java:91)
      at io.helidon.http.media.ReadableEntityBase.as(ReadableEntityBase.java:149)
      at io.helidon.http.media.ReadableEntityBase.as(ReadableEntityBase.java:144)
      at io.helidon.webclient.api.HttpClientResponse.as(HttpClientResponse.java:55)
      at io.helidon.webclient.tests.JsonMediaHttpClientTest.shouldParseJsonResponseWithStringValueEndingAtBufferBoundary(JsonMediaHttpClientTest.java:54)
```

Verdent:
### Description
Fixes #12062

### Documentation

None

### DIFF
--- a/json/json/src/main/java/io/helidon/json/JsonParserStream.java
+++ b/json/json/src/main/java/io/helidon/json/JsonParserStream.java
@@ -154,12 +154,14 @@ final class JsonParserStream extends JsonParserBase {
                 }
 
                 // 3. Read from the stream until we have enough bytes or EOF.
-                int lastRead = inputStream.read(buffer, kept, buffer.length - kept);
-                if (lastRead == -1) {
-                    finished = true;
-                    throw createException("Unexpected end of the binary JSON found");
+                while (currentIndex + amount >= bufferLength) {
+                    int lastRead = inputStream.read(buffer, bufferLength, buffer.length - bufferLength);
+                    if (lastRead == -1) {
+                        finished = true;
+                        throw createException("Unexpected end of the binary JSON found");
+                    }
+                    bufferLength += lastRead;
                 }
-                bufferLength = kept + lastRead;
             } catch (IOException e) {
                 throw new JsonException("Failed to read more Smile data from stream", e);
             }
--- a/json/json/src/main/java/io/helidon/json/JsonParserStream.java
+++ b/json/json/src/main/java/io/helidon/json/JsonParserStream.java
@@ -158,12 +158,12 @@ final class JsonParserStream extends JsonParserBase {
                     int lastRead = inputStream.read(buffer, bufferLength, buffer.length - bufferLength);
                     if (lastRead == -1) {
                         finished = true;
-                        throw createException("Unexpected end of the binary JSON found");
+                        throw createException("Unexpected end of the JSON stream found");
                     }
                     bufferLength += lastRead;
                 }
             } catch (IOException e) {
-                throw new JsonException("Failed to read more Smile data from stream", e);
+                throw new JsonException("Failed to read more JSON data from stream", e);
             }
         }
     }
--- a/json/json/src/main/java/io/helidon/json/JsonParserStream.java
+++ b/json/json/src/main/java/io/helidon/json/JsonParserStream.java
@@ -118,7 +118,7 @@ final class JsonParserStream extends JsonParserBase {
                 // 1. Compact: shift unneeded prefix out so the required bytes can fit.
                 int preserveFrom;
                 if (replayMarked && mark >= 0 && mark <= currentIndex) {
-                    preserveFrom = mark;
+                    preserveFrom = bufferingJsonValue ? Math.min(mark, jsonValueStart) : mark;
                 } else if (bufferingJsonValue) {
                     preserveFrom = jsonValueStart;
                 } else {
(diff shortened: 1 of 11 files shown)


json/json/src/main/java/io/helidon/json/JsonParserStream.java
```diff
@@ -118,7 +118,7 @@ void ensure(int amount) {
                 // 1. Compact: shift unneeded prefix out so the required bytes can fit.
                 int preserveFrom;
                 if (replayMarked && mark >= 0 && mark <= currentIndex) {
-                    preserveFrom = mark;
+                    preserveFrom = bufferingJsonValue ? Math.min(mark, jsonValueStart) : mark;
```
danielkec, line 121: This still assumes one bulk `read` supplies all bytes required by `ensure(amount)`. `InputStream.read(byte[], off, len)` may legally return fewer bytes, and callers consume the requested bytes immediately after `ensure` returns; for example, the `\u` escape path calls `ensure(4)` and then reads four bytes.

```java
class OneByteInputStream extends ByteArrayInputStream {
    OneByteInputStream(byte[] bytes) {
        super(bytes);
    }

    @Override
    public synchronized int read(byte[] b, int off, int len) {
        return super.read(b, off, Math.min(len, 1));
    }
}

byte[] bytes = "\"\\u1234\"".getBytes(StandardCharsets.UTF_8);
JsonParser.create(new OneByteInputStream(bytes), 6).readString();
```

At this head this throws `JsonException: Invalid hex digit found. Found: 0x00` instead of returning U+1234.

json/json/src/main/java/io/helidon/json/JsonValueParser.java
```diff
@@ -109,17 +109,19 @@ private void recordToReplayQueue() {
...
 
     void ensureCapacity(int capacity) {
-        if (index + capacity > values.length) {
-            JsonValue[] newValues = new JsonValue[values.length * 2];
-            System.arraycopy(values, 0, newValues, 0, index);
+        int required = Math.max(0, index) + capacity + 1;
+        if (required > values.length) {
+            int newLength = Math.max(values.length * 2, required);
+            JsonValue[] newValues = new JsonValue[newLength];
+            System.arraycopy(values, 0, newValues, 0, Math.max(0, index));
```
danielkec: `index` is still an occupied stack slot here, so this copy drops `values[index]` when a nested value grows the stack while outer tokens are still queued. The top-level large-object test passes, but this nested case still fails:

```java
StringBuilder json = new StringBuilder("{\"a\":{");
for (int i = 0; i < 120; i++) {
    if (i > 0) {
        json.append(',');
    }
    json.append("\"k").append(i).append("\":").append(i);
}
json.append("},\"z\":999}");

JsonValue value = JsonParser.create(json.toString()).readJsonValue();
JsonParser parser = JsonParser.create(value);
while (parser.hasNext()) {
    parser.nextToken();
}
```

At current head this throws `JsonException: No more JSON Values available` after the nested object instead of continuing to the outer comma/key.

Verdent: fixed

danielkec (approved):
Reviewed head 0acf01822a54b185e38f213e67a0b6f342fbc939. Previous review threads are resolved and current checks are green.

tomas-langer (approved):

PR #12065, opened 2026-06-02
Remove deprecated GC time type check (27.x)

### LINKED ISSUE #11477: Resolve deprecations in module `helidon-metrics-built-in-system-meters` (27.x)
- [ ] io.helidon.metrics.systemmeters.SystemMetersProvider#isGcTimeGauge:312

tomas-langer:
Resolves #11477

Removes the deprecated `isGcTimeGauge` helper from the built-in system meters provider and always uses the gauge registration path for `gc.time`.

### DIFF
--- a/metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/SystemMetersProvider.java
+++ b/metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/SystemMetersProvider.java
@@ -265,21 +265,11 @@ public class SystemMetersProvider implements MetersProvider {
                           GarbageCollectorMXBean::getCollectionCount,
                           Tag.create("name", poolName));
             // Express the GC time in seconds.
-            // @Deprecated(forRemoval = true) - Starting in Helidon 5 always register a gauge instead of checking which
-            // type of meter to register.
-            if (isGcTimeGauge()) {
-                registerGauge(result,
-                              metadata(GC_TIME),
-                              gcBean,
-                              bean -> (long) (bean.getCollectionTime() / 1000.0D),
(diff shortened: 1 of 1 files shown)


metrics/system-meters/src/main/java/io/helidon/metrics/systemmeters/SystemMetersProvider.java
```diff
@@ -265,21 +265,11 @@ private Metadata metadata(Metadata.Builder metadataBuilderWithCamelCaseName) {
...
-                              bean -> (long) (bean.getCollectionTime() / 1000.0D),
-                              Tag.create("name", poolName));
-            } else {
-                registerFunctionalCounter(result,
-                                          metadata(GC_TIME),
-                                          gcBean,
-                                          bean -> (long) (bean.getCollectionTime() / 1000.0D),
-                                          Tag.create("name", poolName));
-            }
+            registerGauge(result,
```
danielkec, line 268: This makes `metrics.gc-time-type=counter` a no-op while it is still public and documented. Reproducer:

```java
ConfigSources.create(Map.of("gc-time-type", "counter"));
metricsConfig.gcTimeType(); // COUNTER
```

Even with that config, `gc.time` always uses this `registerGauge(...)` path, so the runtime behavior no longer matches the accepted config.

tomas-langer: This PR is removing deprecated methods from the `systemmeters` module only. 
The reported part will be removed as cleanup of `helidon-metrics-api`

danielkec (approved):

PR #12070, opened 2026-06-04
Support OpenTelemetry HTTP server metrics semantic conventions from MP

### LINKED ISSUE #12098: Built-in HTTP server metrics should emit templated `http.route` for Helidon MP / JAX-RS resources
## Summary

Helidon 4.4.1 can emit the built-in OpenTelemetry inbound HTTP metric `http.server.request.duration` for Helidon MP applications, but for JAX-RS resources the exported metric does not include `http.route`.

For route-level RED dashboards and SLOs, the desired behavior is for `http.route` to contain the low-cardinality matched route template, such as:

```text
/demo/orders/{name}
/demo/duration-unit-probe
/greet/{name}
```

not the raw request path, such as:

```text
/demo/orders/alice
/demo/orders/bob
/greet/Joe
```

Using the route template is important because it keeps metric cardinality bounded while still allowing latency, traffic, and error analysis by API route.

## Environment

- Helidon MP 4.4.1
- JAX-RS resources
- `helidon-webserver-observe-telemetry-metrics`
- `opentelemetry-exporter-otlp`
- OTLP metrics export over gRPC
- Metrics viewed through Prometheus/Grafana via an OTLP collector

## Observed Behavior

The built-in metric is emitted, but without `http_route`:

```text
http_server_request_duration_seconds_count{
  http_request_method="GET",
  http_response_status_code="200",
  job="helidon-quickstart-mp",
  service_name="helidon-quickstart-mp",
  socket_name="@default",
  url_scheme="HTTP"
}
```

There is no `http_route` label on the live OTLP-pushed series for the MP/JAX-RS endpoints.

This makes route-level queries impossible, for example:

```promql
sum by (http_route) (
  rate(http_server_request_duration_seconds_count{service_name="helidon-quickstart-mp"}[5m])
)
```

Without `http_route`, dashboards can show overall traffic and latency, but they cannot answer which API route is slow, noisy, or failing.

## Expected Behavior

For a JAX-RS resource such as:

```java
@Path("/demo")
public class DemoApiResource {

    @GET
    @Path("/orders/{name}")
    public Response readOrder(@PathParam("name") String name) {
        ...
    }
}
```

requests to:

```text
GET /demo/orders/alice
GET /demo/orders/bob
```

should produce metrics with:

```text
http.route="/demo/orders/{name}"
```

and in Prometheus-compatible naming:

```text
http_route="/demo/orders/{name}"
```

The label should be based on the framework-matched route template, not inferred from the raw request URI.

## Why Templated Routes Matter

The `http.route` attribute is one of the most important dimensions for RED-style HTTP metrics:

- **Rate:** request volume by endpoint
- **Errors:** error percentage or burn rate by endpoint
- **Duration:** latency by endpoint

It should remain low-cardinality. A raw path label would create one time series per user ID, order ID, search term, etc. A templated route label keeps the series count bounded and operationally useful.

For example:

```text
Good: /users/{id}
Bad:  /users/12345
Bad:  /users/67890
```

## Suspected Implementation Gap

From inspecting the Helidon 4.4.1 artifacts, the OpenTelemetry HTTP metrics filter appears to populate `http.route` from:

```java
RoutingRequest.matchingPattern().orElse("")
```

The WebServer routing executor sets `matchingPattern` from the Helidon WebServer route matcher before invoking the handler.

However, the Helidon Jersey bridge registers the JAX-RS application using a WebServer `any` route. The actual JAX-RS resource matching happens later inside Jersey. As a result, the WebServer route matcher does not know the final JAX-RS resource template, so `RoutingRequest.matchingPattern()` appears to be empty for MP/JAX-RS requests.

Jersey does have the matched route templates available through `ExtendedUriInfo.getMatchedTemplates()`, but that information does not appear to be propagated back into the Helidon `RoutingRequest` matching pattern or used by the built-in OpenTelemetry metrics implementation.

## Desired Feature

For Helidon MP / JAX-RS applications, built-in HTTP server metrics should emit `http.route` using the matched JAX-RS template.

Possible implementation approaches might include:

- Have the Jersey/JAX-RS bridge propagate the matched template into `RoutingRequest.matchingPattern()`.
- Have the OpenTelemetry HTTP metrics integration consult Jersey route metadata when handling an MP/JAX-RS request.
- Provide a supported hook/SPIs so integrations can supply the canonical route template before the HTTP duration metric is recorded.

The important behavior is that `http.route` should be populated with a low-cardinality template for JAX-RS resources.

## Impact

Without `http.route`, the built-in HTTP metric is much less useful for production dashboards and SLOs:

- route-level latency panels cannot be built
- route-level error-rate panels cannot be built
- route-level request-rate panels cannot be built
- alerts cannot identify the affected API route
- teams may need to keep custom JAX-RS filters only to recover the route label

cc @tjquinno

tjquinno:
### Description
This PR makes a few "opt-in" changes in OpenTelemetry server HTTP metrics semantic conventions:
* Use correct `http.route` attribute on meters as computed by `JaxRsService` for requests in MP apps.
* Update post-request metrics data in `whenSent` rather than in a filter to more accurately account for all work done for a request.
* Govern these changes using a config setting so the default is to preserve existing behavior for compatibility.

### Release Note
____
* Helidon 4.x MP services can now push OpenTelemetry server HTTP metrics which conform to OTel's semantic conventions by adding this dependency:
   ```xml
   <dependency>
       <groupId>io.helidon.webserver.observe</groupId>
       <artifactId>helidon-webserver-observe-telemetry-metrics</artifactId>
       <scope>runtime</scope>
   </dependency>
   ```
   and setting the config value described below.
* The `http.route` attribute on the OTel HTTP semantic conventions metrics is now correct for MP apps.
* Helidon updates the OpenTelemetry server HTTP metrics timer _after the response has been sent_ (using the response `whenSent` mechanism) rather than in a filter. This more accurately accounts for all the time Helidon spends processing the request.
* These new behaviors are controlled by a new config setting:
   ```yaml
   server:
     features:
       observe:
         observers:
           metrics:
             auto-http-metrics:
               use-updated-http-metrics: true
   ```
   The new setting defaults to `false` to preserve the prior 4.x behavior for compatibility.

____

### PR summary
We already have an SE filter that adds the OTel semantic conventions, but it uses the SE `ServerRequest.matchingPattern` method which, for Jakarta REST requests, is not set.

This PR expands the API and implementations for requests to accept a `Supplier<Optional<String>>` which can provide the matching route. The new method allows `JaxRsService` to provide the matching route it detects (if any) for MP/JAX-RS requests in addition to the SE request handling providing its matching route for SE requests.

This approach reuses the existing SE filter without duplicating its code while keeping the SE code independent from MP dependencies.
____

### Documentation
PR includes an update to the common shared content on the SE and MP metrics doc pages.


webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -175,6 +176,16 @@ private void updateMetricsIfMeasured(RoutingRequest req,
             httpRequestDuration.record((endTime - startTime) / 1_000_000.0, attrBuilder.build());
         }
 
+        private String route(RoutingRequest req) {
+            /*
```
tomas-langer: JAX-RS will not be part of Helidon core in the future (it moves to Helidon Microprofile repository). 
This is a reference to a "possible" implementation, and is an in correct dependency. Either we need to design an abstraction that holds without a reference to MP here, or this code does not belong here.

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
```diff
@@ -39,6 +40,14 @@
 import io.helidon.webserver.testing.junit5.Socket;
 
 import io.opentelemetry.exporter.logging.otlp.OtlpJsonLoggingMetricExporter;
+import jakarta.ws.rs.ApplicationPath;
```
tomas-langer: You cannot use Jakarta in SE (ever).

webserver/observe/telemetry/metrics/pom.xml
```diff
@@ -78,6 +78,11 @@
             <artifactId>helidon-webserver-testing-junit5</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.helidon.jersey</groupId>
```
tomas-langer: This dependency does not belong here, create a test in `microprofile/tests` or in the Jersey module.

tomas-langer (changes requested):

tjquinno:
Ugh. Apologies. 

Yes, of course JAX-RS/Jersey stuff does not belong here. I had multiple branches in play and pushed one with the wrong changes. I've cleaned up the PR.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tomas-langer: This is still true

Either the feature is valid on its own, or do not add it.

Why does this require different handling from what we do in SE?

tomas-langer (changes requested):

tjquinno:
We have an existing SE filter that implements the HTTP server metrics semantic conventions. Ideally we would avoid writing a completely separate Jakarta REST filter that duplicated that SE filter.

Part of the semantic conventions is adding a tag for the routing pattern which matched the incoming request. For requests handled by SE, `ServerRequest.matchingPattern` works fine. 

But, when `JaxRsService` handles a request (in MP), the SE filter sees the SE `ServerRequest.matchingPattern` as an empty `Optional` even if the Jakarta REST request _did_ match a Jakarta REST routing pattern.

Given that `JaxRsService` knows about the matching pattern, and because the `JaxRsService` has access to the SE `ServerRequest`, the PR changed `JaxRsService` to set a request context value--just a `String`, with no use of JaxRS types.

The PR also changed the SE filter so it would prefer a routing value set in the request context and fall back to `ServerRequest.matchingPattern`.

This reuses the existing SE filter with a minimal change that is not at all tightly coupled to MP or JaxRs.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -343,6 +353,45 @@ private void doHandle(Context ctx, ServerRequest req, ServerResponse res) {
...
+                }
+            }
+            resource = resource.getParent();
+        }
+
+        derivedPath.push(applicationPath());
+        return String.join("", derivedPath);
+    }
+
+    private String applicationPath() {
```
tomas-langer: This seems to be happening for every request that needs this information, even though it can be computed at startup.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -318,7 +324,11 @@ private void doHandle(Context ctx, ServerRequest req, ServerResponse res) {
             kpiMetricsContext.ifPresent(KeyPerformanceIndicatorSupport.DeferrableRequestContext::requestProcessingStarted);
             appHandler.handle(requestContext);
             writer.await();
-            if (res.status() == Status.NOT_FOUND_404 && requestContext.getUriInfo().getMatchedResourceMethod() == null) {
+            boolean matchedResourceMethod = requestContext.getUriInfo().getMatchedResourceMethod() != null;
+            if (matchedResourceMethod) {
+                registerRoute(ctx, requestContext);
```
tomas-langer: This puts route derivation on every successful JAX-RS request, even when the OTel semantic-conventions metrics module is not on the classpath or auto HTTP metrics are disabled. The new `registerRoute` path allocates a `LinkedList`/nodes, joins a new route string, does the application annotation lookup, and writes into the request context for all MP traffic. A small local JMH smoke test of this route-building shape on JDK 21 measured about 47 ns/op and about 264 B/op versus about 2 ns/op for a cached route, while the SE metrics context-miss path stayed essentially allocation-free. Can we either gate this work on a consumer needing the route or precompute/cache the route template so ordinary JAX-RS requests do not pay this per-request allocation cost?

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -343,6 +353,45 @@ private void doHandle(Context ctx, ServerRequest req, ServerResponse res) {
...
+            if (resourcePath != null && !resourcePath.equals("/") && !resourcePath.isBlank()) {
+                derivedPath.push(resourcePath);
+                if (!resourcePath.startsWith("/")) {
+                    derivedPath.push("/");
+                }
+            }
+            resource = resource.getParent();
+        }
+
+        derivedPath.push(applicationPath());
```
tomas-langer: This concatenates the `@ApplicationPath` value exactly as returned by `applicationPath()` with the resource paths. Valid JAX-RS values such as `@ApplicationPath("app")` produce `app/greet/{name}` with no leading slash, and `@ApplicationPath("/foo/bar/")` with `@Path("/greet")` can produce `/foo/bar//greet`. SE routes from `matchingPattern()` are normalized slash-prefixed templates, so MP semantic-convention metrics can diverge and create separate series for the same route. Please normalize the application path/resource separators and add coverage for the non-normalized cases.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
tjquinno: Cached at init.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
tjquinno: Redesigned. 

One or more SE filters can now "request" matching route support by adding a `Consumer<Supplier<>>` to the request context, and a later filter or--in this case, the handler in `JaxRsService`--can "provide" matching route support (if it was requested upstream), passing a `Supplier` to the consumer(s) that is not invoked synchronously during the JAX-RS request handling. These details are encapsulated in a new `RoutePathSupport` type.

A requesting filter's `Consumer` saves the provided `Supplier` , if any, and then the filter retrieves the provided value (if any) from the supplier and acts accordingly. If no code invoked a requesting filter's `Consumer` then the requesting filter falls back to using SE request information to find the matching route.

This is a slightly more complicated contract but removes virtually all the additional processing from the overall request processing critical path:
* the JAX-RS processing provides a `Supplier` which computes the value rather than computing the value in-line
* the OTel semantic conventions filter already does almost of all of its work--including, now, the resolution of the JAX-RS path--asynchronously as well.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
tjquinno: Normalized; tests added.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigBlueprint.java
```diff
@@ -71,4 +72,21 @@ interface AutoHttpMetricsConfigBlueprint {
...
+     * Whether to use the updated automatic HTTP metrics behavior.
+     * <p>
+     * When enabled, OpenTelemetry automatic HTTP metrics are recorded when the response is sent, use seconds for the
+     * duration value, retain the response status when request processing throws an exception, and use the framework-
+     * provided route when one is available. The default retains the prior Helidon 4.x metric output.
+     *
+     * @return whether to use the updated automatic HTTP metrics behavior
+     * @deprecated this temporary Helidon 4.x compatibility setting is planned for removal in the next major release
+     */
+    @Deprecated(forRemoval = true, since = "4.5.2")
```
tomas-langer: I think this should not be declared deprecated. 
We need to keep this method in `main`, change the default to `true` there, and then deprecate it and mark for removal, as the default will be the "fixed" behavior

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -120,15 +127,69 @@ private MetricsRecordingFilter(DoubleHistogram httpRequestDuration, AutoHttpMetr
...
+
+            try {
+                chain.proceed();
+                chainComplete.set(true);
+                if (responseSent.get()) {
+                    recordMetrics.run();
+                }
+            } catch (Exception e) {
+                exception.set(e);
+                recordMetrics.run();
```
tomas-langer: When a downstream filter throws before sending a response, this records immediately before the outer WebServer error handler selects and sends the final response. `ServerResponseBase.status()` therefore supplies its default `200`, and the `recorded` CAS prevents `whenSent` from correcting the metric after the client receives `500` or another error status. This violates the OpenTelemetry condition that `http.response.status_code` describe the status actually sent. Delay exception-path recording until the error response completes (or omit the status if none is sent), and add an integration test covering a later filter throwing before send. [OpenTelemetry HTTP metrics contract](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestduration)

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -318,7 +320,12 @@ private void doHandle(Context ctx, ServerRequest req, ServerResponse res) {
             kpiMetricsContext.ifPresent(KeyPerformanceIndicatorSupport.DeferrableRequestContext::requestProcessingStarted);
             appHandler.handle(requestContext);
             writer.await();
-            if (res.status() == Status.NOT_FOUND_404 && requestContext.getUriInfo().getMatchedResourceMethod() == null) {
+            boolean matchedResourceMethod = requestContext.getUriInfo().getMatchedResourceMethod() != null;
+            if (matchedResourceMethod) {
+                Resource matchedResource = requestContext.getUriInfo().getMatchedModelResource();
```
tomas-langer: `getMatchedModelResource()` returns only the resource containing the finally matched method; Jersey explicitly excludes intermediary resources containing sub-resource locators. For a valid route such as `/widgets/{id}` whose locator returns a sub-resource handling `/details`, the parent walk below can emit `/app/details` and omit `/widgets/{id}`. Build the route from the complete matched-template sequence, including locator templates, then prepend the effective WebServer service mount. Add regression coverage for a sub-resource locator. [Jersey `ExtendedUriInfo` contract](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/latest31x/jersey/org/glassfish/jersey/server/ExtendedUriInfo.html)

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
```diff
@@ -343,6 +350,54 @@ private void doHandle(Context ctx, ServerRequest req, ServerResponse res) {
...
+        int start = 0;
+        int end = path.length();
+        while (start < end && path.charAt(start) == '/') {
+            start++;
+        }
+        while (end > start && path.charAt(end - 1) == '/') {
+            end--;
+        }
+
+        if (start == end) {
```
tomas-langer, line 404: This treats a matched path of `/` as an empty segment. When the JAX-RS service itself is mounted at the root, `servicePath(req)` is also empty, so a root resource such as `@Path("/")` can produce an empty supplied route and force metrics back to the less-precise WebServer matching pattern. Preserve `/` as the root route and add coverage for a root-mounted JAX-RS application with a root resource.

docs/src/main/asciidoc/mp/metrics/metrics.adoc
```diff
@@ -70,6 +81,46 @@ When you add code to your service to create a metric programmatically, the code
...
+
+[source,yaml]
+.Example configuration for a test of demo of OpenTelemetry metrics semantic conventions
+----
+telemetry:                              # <1>
+  service: mp-with-otel-sem-conv        # <2>
+  signals:
+    metrics:                            # <3>
+      exporters:
+        - type: otlp                    # <4>
```
tomas-langer, line 114: This configuration requires the concrete OTLP exporter, but the dependency instructions above add only `helidon-webserver-observe-telemetry-metrics`. Its transitive `helidon-telemetry-opentelemetry-config` dependency declares `opentelemetry-exporter-otlp` as optional, so applications following this example do not receive the exporter implementation and can fail during telemetry setup. Add `io.opentelemetry:opentelemetry-exporter-otlp` as a runtime dependency in the example, or link directly to instructions that make the extra dependency explicit.

docs/src/main/asciidoc/mp/metrics/metrics.adoc
```diff
@@ -70,6 +81,46 @@ When you add code to your service to create a metric programmatically, the code
...
+          protocol: http/proto
+          endpoint: http://localhost:9090/api/v1/otlp/v1/metrics
```
tomas-langer, line 116: `http://localhost:9090/api/v1/otlp/v1/metrics` is the Prometheus OTLP metrics receiver, not a Jaeger endpoint. Prometheus also disables this receiver by default, so the documented example will not work against a normally started local server. Name Prometheus as the backend and include its `--web.enable-otlp-receiver` setup, or replace the endpoint and prose with a metrics-capable backend that matches the example. [Prometheus OTLP guide](https://prometheus.io/docs/guides/opentelemetry/) [Jaeger APIs](https://www.jaegertracing.io/docs/2.17/architecture/apis/)

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutePathSupport.java
```diff
@@ -0,0 +1,100 @@
...
+    public static void provideRoute(Context context, Supplier<String> routeSupplier) {
+        Objects.requireNonNull(context, "Parameter 'context' is null!");
+        Objects.requireNonNull(routeSupplier, "Parameter 'routeSupplier' is null!");
+
+        context.get(ROUTE_PATH_CONSUMERS, RoutePathConsumers.class)
+                .ifPresent(consumers -> consumers.provide(memoize(routeSupplier)));
+    }
+
+    private static RoutePathConsumers routePathConsumers(Context context) {
+        return context.get(ROUTE_PATH_CONSUMERS, RoutePathConsumers.class)
```
tomas-langer: The first `get`/`register` sequence is not atomic. Two concurrent first requesters can both observe no holder, create separate `RoutePathConsumers`, and register them; `ListContext` replaces the earlier same-type registration, leaving one consumer attached to an orphaned holder that `provideRoute` never sees. `CopyOnWriteArrayList` protects each list but not holder installation. Install one canonical holder atomically or guard first initialization with an explicit lock, and add a barrier-based test that races the first `requestRoute` registrations.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: Updated completion coordinator to defer exception-path recording until final response is sent. New integration test verifies downstream failure emits status 500; no response sent ==> no metric emitted.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
tjquinno: Route computation now uses `ExtendedUriInfo.getMatchedTemplates()` and reverses Jersey’s leaf-to-root ordering, so locator templates are included. Added coverage for `/widgets/{id}/details`.

jersey/webserver/src/main/java/io/helidon/jersey/webserver/JaxRsService.java
tjquinno: Root routes now resolve to `/` rather than an empty supplied route. Added test.

docs/src/main/asciidoc/mp/metrics/metrics.adoc
tjquinno: Added `io.opentelemetry:opentelemetry-exporter-otlp` as runtime dependency in doc example.

docs/src/main/asciidoc/mp/metrics/metrics.adoc
tjquinno: Example now describes Prometheus usage the need to start it with `--web.enable-otlp-receiver`

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutePathSupport.java
tjquinno: Initialization of route-consumer holder is now atomic by synchronizing lookup and registration on the request context. Added a test.

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutePathSupport.java
```diff
@@ -73,12 +73,14 @@ public static void provideRoute(Context context, Supplier<String> routeSupplier)
     }
 
     private static RoutePathConsumers routePathConsumers(Context context) {
-        return context.get(ROUTE_PATH_CONSUMERS, RoutePathConsumers.class)
-                .orElseGet(() -> {
-                    RoutePathConsumers consumers = new RoutePathConsumers();
-                    context.register(ROUTE_PATH_CONSUMERS, consumers);
-                    return consumers;
-                });
+        synchronized (context) {
```
tomas-langer: This must not be used. We never use `synchronized` in Helidon, and never on an instance that may be locked by somebody else.

This should be guarded by a lock.

tomas-langer (changes requested):

webserver/webserver/src/main/java/io/helidon/webserver/http/RoutePathSupport.java
tjquinno: Changed.

tomas-langer (approved):

PR #12072, opened 2026-06-04
Fix FT config bootstrap metrics lookup

### LINKED ISSUE #12071: Fault tolerance config lookup recurses during config bootstrap
This is the Helidon 27/main counterpart of #12069.

Fault Tolerance currently reads the global Config from the service registry while building handlers, for example when `RetryImpl` evaluates the global metrics default through `MetricsUtils.defaultEnabled()`.

If a custom config source is used during config bootstrap and that source builds a Fault Tolerance handler, the handler setup attempts to obtain `Config` from the registry before config creation has completed. This can re-enter config bootstrap and recurse.

The 4.x report reproduces this in MP. The same setup path exists in SE when meta-configuration loads a custom config source that creates `Retry`.

Expected behavior: Fault Tolerance handlers can be constructed from config sources during config bootstrap. Global Fault Tolerance metrics configuration should only be applied once the registry Config service is already active, and must not activate or recursively create Config during bootstrap.

tomas-langer:
Resolves #12071

This fixes FT bootstrap recursion by:
- adding active-only ServiceRegistry lookup support that can observe already active services without creating Config
- marking explicit Services.set instances ACTIVE so firstActive lookup can return them while still rejecting DESTROYED instances
- using active-only config lookup for Fault Tolerance metrics defaults
- adding the SE config bootstrap reproducer and service-registry lifecycle regressions

### DIFF
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
@@ -16,6 +16,7 @@
 package io.helidon.faulttolerance;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Supplier;
 
 import io.helidon.common.LazyValue;
@@ -27,7 +28,8 @@ import io.helidon.metrics.api.Metrics;
 import io.helidon.metrics.api.MetricsFactory;
 import io.helidon.metrics.api.Tag;
 import io.helidon.metrics.api.Timer;
-import io.helidon.service.registry.Services;
+import io.helidon.service.registry.GlobalServiceRegistry;
+import io.helidon.service.registry.ServiceRegistry;
 
 import static io.helidon.faulttolerance.FaultTolerance.FT_METRICS_DEFAULT_ENABLED;
 import static io.helidon.metrics.api.Meter.Scope.VENDOR;
@@ -48,9 +50,10 @@ class MetricsUtils {
      * @return value of metrics flag
      */
     static boolean defaultEnabled() {
-        return Services.get(Config.class)
-                .get(FT_METRICS_DEFAULT_ENABLED)
-                .asBoolean()
+        return maybeExistingConfig()
+                .map(config -> config.get(FT_METRICS_DEFAULT_ENABLED)
+                        .asBoolean()
+                        .orElse(false))
                 .orElse(false);
     }
 
@@ -88,5 +91,14 @@ class MetricsUtils {
 
     static Timer timer(String name, Tag... tags) {
         return METRICS_REGISTRY.get().timer(name, List.of(tags)).orElseThrow();
+    }
+
+    private static Optional<Config> maybeExistingConfig() {
+        if (!GlobalServiceRegistry.configured()) {
+            return Optional.empty();
+        }
+
+        ServiceRegistry registry = GlobalServiceRegistry.registry();
+        return registry.firstActive(Config.class);
     }
 }
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
@@ -16,7 +16,6 @@
 package io.helidon.faulttolerance;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.function.Supplier;
 
 import io.helidon.common.LazyValue;
@@ -50,10 +49,13 @@ class MetricsUtils {
      * @return value of metrics flag
      */
     static boolean defaultEnabled() {
-        return maybeExistingConfig()
-                .map(config -> config.get(FT_METRICS_DEFAULT_ENABLED)
-                        .asBoolean()
-                        .orElse(false))
+        ServiceRegistry registry = GlobalServiceRegistry.registry();
+        if (registry.isResolvingOnCurrentThread(Config.class)) {
+            return false;
+        }
+        return registry.get(Config.class)
+                .get(FT_METRICS_DEFAULT_ENABLED)
+                .asBoolean()
                 .orElse(false);
     }
 
@@ -93,12 +95,4 @@ class MetricsUtils {
         return METRICS_REGISTRY.get().timer(name, List.of(tags)).orElseThrow();
     }
 
-    private static Optional<Config> maybeExistingConfig() {
-        if (!GlobalServiceRegistry.configured()) {
-            return Optional.empty();
-        }
-
-        ServiceRegistry registry = GlobalServiceRegistry.registry();
-        return registry.firstActive(Config.class);
-    }
 }
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
@@ -16,6 +16,7 @@
 package io.helidon.faulttolerance;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Supplier;
 
 import io.helidon.common.LazyValue;
@@ -49,13 +50,10 @@ class MetricsUtils {
      * @return value of metrics flag
      */
     static boolean defaultEnabled() {
-        ServiceRegistry registry = GlobalServiceRegistry.registry();
-        if (registry.isResolvingOnCurrentThread(Config.class)) {
-            return false;
-        }
-        return registry.get(Config.class)
-                .get(FT_METRICS_DEFAULT_ENABLED)
-                .asBoolean()
+        return maybeExistingConfig()
+                .map(config -> config.get(FT_METRICS_DEFAULT_ENABLED)
+                        .asBoolean()
+                        .orElse(false))
                 .orElse(false);
     }
 
@@ -95,4 +93,12 @@ class MetricsUtils {
         return METRICS_REGISTRY.get().timer(name, List.of(tags)).orElseThrow();
     }
 
+    private static Optional<Config> maybeExistingConfig() {
+        if (!GlobalServiceRegistry.configured()) {
+            return Optional.empty();
+        }
+
+        ServiceRegistry registry = GlobalServiceRegistry.registry();
+        return registry.firstActive(Config.class);
+    }
 }
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
@@ -16,7 +16,6 @@
 package io.helidon.faulttolerance;
 
 import java.util.List;
-import java.util.Optional;
 import java.util.function.Supplier;
 
 import io.helidon.common.LazyValue;
@@ -29,7 +28,6 @@ import io.helidon.metrics.api.MetricsFactory;
 import io.helidon.metrics.api.Tag;
 import io.helidon.metrics.api.Timer;
 import io.helidon.service.registry.GlobalServiceRegistry;
-import io.helidon.service.registry.ServiceRegistry;
 
 import static io.helidon.faulttolerance.FaultTolerance.FT_METRICS_DEFAULT_ENABLED;
 import static io.helidon.metrics.api.Meter.Scope.VENDOR;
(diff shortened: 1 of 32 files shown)


service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
```diff
@@ -66,6 +79,46 @@ ServiceInstance<T> registryInstance(Lookup lookup, QualifiedInstance<T> instance
...
     }
 
+    Optional<List<ServiceInstance<T>>> activeInstances(Lookup lookup) {
+        if (!fixedInstance && Service.PerLookup.TYPE.equals(provider.descriptor().scope())) {
+            return Optional.empty();
+        }
+
+        Activator<T> serviceActivator;
+        if (fixedInstance) {
+            serviceActivator = activatorSupplier.get();
```
danielkec: This fixed-instance path leaves the activator outside the scoped registry when `firstActive` is the only lookup path. Reproducer: call `Services.set(...)`, then `serviceRegistry.firstActive(...)`, then `manager.shutdown()`, then `serviceRegistry.firstActive(...)` again. The retained registry still returns the explicit instance because shutdown only deactivates activators held by `ScopedRegistryImpl`. The current destroyed-instance test calls `get(...)` before shutdown, which registers the activator and misses this path.

service/registry/src/main/java/io/helidon/service/registry/ServiceRegistry.java
```diff
@@ -497,6 +567,22 @@ default List<ServiceInfo> allServices(Class<?> contract) {
...
+     * create new instances for {@link Service.PerLookup} services, as per-lookup services do not have an active reusable
+     * instance. The default implementation delegates to {@link #first(Lookup)} and may trigger activation.
+     *
+     * @param lookup lookup criteria to find matching services
+     * @param <T>    type of the service, if you use any other than {@link java.lang.Object}, make sure
+     *               you have configured appropriate contracts in the lookup, as we cannot infer this
+     * @return the best active service instance matching the lookup, or an empty {@link java.util.Optional} if none match
+     */
+    default <T> Optional<T> firstActive(Lookup lookup) {
+        return first(lookup);
```
danielkec: This fallback undermines the active-only guarantee at the call site that introduced it. `MetricsUtils.maybeExistingConfig()` invokes `firstActive(Config.class)` on the public `ServiceRegistry` installed through `GlobalServiceRegistry.registry(...)`; an existing custom or wrapping implementation that inherits this default therefore calls its activating `first(Lookup)` and can re-enter Config bootstrap. A minimal proxy registry reports `first(Lookup) calls=1` from `firstActive(Config.class)`, preserving the #12071 recursion path.

service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
```diff
@@ -66,6 +79,50 @@ ServiceInstance<T> registryInstance(Lookup lookup, QualifiedInstance<T> instance
                                          instance);
     }
 
+    Optional<List<ServiceInstance<T>>> activeInstances(Lookup lookup) {
+        if (!fixedInstance && Service.PerLookup.TYPE.equals(provider.descriptor().scope())) {
```
danielkec: `fixedInstance` also bypasses this guard when the fixed object is a per-lookup provider rather than the produced contract. Reproducer: `Services.set(ServiceSupplier.class, new ServiceSupplier())`, followed by `firstActive(SuppliedContract.class)`, returns a value and invokes the supplier once (`produced=1`). The lookup therefore creates the per-lookup result that `firstActive` documents as unavailable.

service/registry/src/main/java/io/helidon/service/registry/Activators.java
```diff
@@ -72,6 +76,12 @@ static <T> Activator<T> create(ServiceProvider<T> provider, T instance) {
         };
     }
 
+    static <T> Activator<T> createActive(ServiceProvider<T> provider, T instance) {
+        Activator<T> activator = create(provider, instance);
+        activator.activate(ACTIVE_REQUEST);
```
danielkec, line 81: Activating every fixed binding here also executes factory-specific activation work during registration. With a fixed `ServicesFactory`, this reaches `services()` during `ServiceRegistryManager.registry()` or `Services.set`; a counter changes from 0 to 1 at registration on this head, while it remains 0 at the same registration points on the base. This makes bootstrap-sensitive user factory code run before its products are requested.

danielkec (commented):
Additional current-head review issues not already covered.

service/registry/src/main/java/io/helidon/service/registry/ServiceRegistry.java
tomas-langer: Fixed in efcd423ddf. The default now validates the lookup and returns an empty optional without delegating to `first`, so an implementation inheriting the default cannot activate a service or re-enter Config bootstrap. The `firstActive` Javadocs now describe the non-activating default and the explicit fixed-provider exception.

service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
tomas-langer: Fixed in efcd423ddf. `activeInstances` now rejects a per-lookup product lookup even for a fixed factory provider; only a lookup for the explicitly registered provider object may proceed. `FixedSupplierActivator` also returns that supplier object for a provider lookup without invoking `get()`. The regression reproduces this case: the product lookup is empty, the provider lookup returns the same explicit instance, and the supplier invocation count remains zero.

service/registry/src/main/java/io/helidon/service/registry/Activators.java
tomas-langer: Keeping this behavior intentionally. `createActive` is called only for an explicitly supplied instance, and its contract is to return that activator in the active phase. For a fixed `ServicesFactory`, executing `services()` is therefore part of activating the explicit binding. The incorrect `firstActive` behavior was exposing or creating per-lookup products from that active provider; efcd423ddf fixes that distinction in `ServiceManager` and the fixed supplier activator while preserving explicit-instance activation.

service/registry/src/main/java/io/helidon/service/registry/Activators.java
```diff
@@ -419,18 +429,30 @@ Optional<List<QualifiedInstance<T>>> targetInstances() {
...
         }
 
+        @SuppressWarnings("unchecked")
+        @Override
+        Optional<List<QualifiedInstance<T>>> targetInstances(Lookup lookup) {
+            if (requestedProvider(lookup, FactoryType.SUPPLIER)) {
+                return Optional.of(List.of(QualifiedInstance.create((T) instanceSupplier,
+                                                                    provider.descriptor().qualifiers())));
+            }
+            return targetInstances();
```
danielkec, line 462: This product fallback still materializes a fixed singleton `Supplier` during an active-only lookup. With an explicitly bound singleton `Supplier<Config>`, its first `get()` constructs `Retry`; `MetricsUtils` then calls `firstActive(Config.class)`, which invokes the same supplier a second time. The probe records `SUPPLIER_GET call=1`, then `call=2`; entering through `registry.get(Config.class)` deadlocks instead. This creates a product that was not already active and leaves the #12071 bootstrap recursion path open.

service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
```diff
@@ -66,6 +79,60 @@ ServiceInstance<T> registryInstance(Lookup lookup, QualifiedInstance<T> instance
...
+                    && factoryType != FactoryType.SERVICE
+                    && !Contracts.requestedProvider(lookup, descriptor, factoryType)) {
+                return Optional.empty();
+            }
+        }
+
+        Activator<T> serviceActivator;
+        if (fixedInstance) {
+            try {
+                serviceActivator = activator();
```
danielkec: Calling `activator()` here also stores a fixed activator when the active lookup returns no instances, while the empty result leaves the contract eligible for a later `Services.set`. Reproducer: bind a fixed singleton `ServicesFactory` returning no products, call `firstActive(Product.class)`, then set a replacement factory returning one product. `Services.set` succeeds, but `firstActive(Factory.class)` still returns the original factory and `firstActive(Product.class)` stays empty; a control run without the negative lookup returns the replacement. The negative lookup therefore makes the permitted later binding silently ineffective.

danielkec (commented):
Two additional exact-head issues not covered by the existing review threads.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
```diff
@@ -48,9 +50,10 @@ private MetricsUtils() {
...
      */
     static boolean defaultEnabled() {
-        return Services.get(Config.class)
-                .get(FT_METRICS_DEFAULT_ENABLED)
-                .asBoolean()
+        return maybeExistingConfig()
+                .map(config -> config.get(FT_METRICS_DEFAULT_ENABLED)
+                        .asBoolean()
+                        .orElse(false))
                 .orElse(false);
```
spericas, line 60: [Codex] This also affects ordinary early SE startup: if a Retry is created before Config is initialized, a configured global fault-tolerance.metrics.enabled=true is ignored, and RetryImpl permanently stores metrics as disabled. Limit the guard to the actual in-progress Config bootstrap case, rather than treating every absent/inactive Config as disabled.

service/registry/src/main/java/io/helidon/service/registry/Activators.java
tomas-langer: Fixed in `999da42ea6`: active-only lookup no longer invokes a fixed singleton supplier to materialize its product. The regression test covers a fixed `Supplier<Config>` whose product is still inactive.

service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
tomas-langer: Fixed in `b937ffd3c2`: an active-only miss no longer retains a fixed activator, so a later `Services.set` replacement remains effective. The regression test covers the negative factory lookup followed by replacement.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/MetricsUtils.java
tomas-langer: No code change for this comment. Issue #12071 requires fault tolerance to consult global Config only when Config is already active. Constructing `Retry` during ordinary early SE startup must not activate Config; therefore `ft.metrics.default-enabled` is best effort and is ignored for that `Retry` when Config is not yet active. The bootstrap and inverse lazy-Config tests cover this boundary.

service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
```diff
@@ -66,6 +79,82 @@ ServiceInstance<T> registryInstance(Lookup lookup, QualifiedInstance<T> instance
...
+
+        Activator<T> serviceActivator;
+        if (fixedInstance) {
+            try {
+                ScopedRegistry scopedRegistry = scopeSupplier.get().registry();
+                if (scopedRegistry instanceof ScopedRegistryImpl scopedRegistryImpl) {
+                    serviceActivator = scopedRegistryImpl.<T>existingActivator(provider.descriptor())
+                            .orElseGet(activatorSupplier);
+                } else {
+                    serviceActivator = activator();
```
danielkec, line 108: For a public custom `ScopedRegistry`, this fallback retains the fixed activator before the active lookup knows whether it has a result. Reproducer: use a custom scope registry whose `activator(...)` caches with `computeIfAbsent`; bind an empty custom-scoped singleton `ServicesFactory`, call `firstActive(Product.class)`, then replace the factory with `Services.set`. The replacement succeeds, but `firstActive(Factory.class)` returns the original factory and the replacement product remains absent. The negative lookup therefore makes the permitted replacement ineffective for custom scope implementations.

danielkec (commented):
One current-head issue not covered by the existing review threads.

service/registry/src/main/java/io/helidon/service/registry/ServiceManager.java
tomas-langer: Scope instances are created by `CoreServiceRegistry.createScope(...)`, which constructs the scoped registry and passes the resulting `Scope` to `ScopeHandler.activate(...)`. The handler's `currentScope()` returns that registry-owned scope; supplying a separate custom `ScopedRegistry` is not a supported implementation path. The reported behavior therefore does not affect supported custom scopes, so no code change is needed.

danielkec (approved):

PR #12074, opened 2026-06-04
Add WebServer SNI support

### LINKED ISSUE #12056: WebServer SNI support
WebServer:

for each server name, we should allow custom configuration of TLS (e.g. this should act similar to additional sockets, except it is on the same port, just with different TLS setup that has different SNI hosts). This would require quite big changes in WebServer setup, esp. validation of each host (for uniqueness and SNI TLS configuration)
we could then allow "virtual host" routing similar to routing setup we have now for each socket (e.g. each "virtual host" (or SNI host) would have customizable routing, defaulting to the default routing setup for the socket (or webserver if none defined per socket). Good side-effect here would be that this could be used even for regular virtual hosts (e.g. using the Host header to choose routing)

tomas-langer:
Resolves #12056

Adds WebServer listener virtual-host SNI TLS selection for NIO TLS, request authority policy enforcement for HTTP/1 and HTTP/2, and exposes SNI host metadata on ServerRequest and gRPC/JSON-RPC request contexts. Updates docs, snippets, regression tests, and SNI-related JMH parser/replay benchmarks.

SNI virtual hosts require listener TLS and the NIO socket-channel transport, which is the default.

Config example:

```yaml
server:
  tls:
    private-key:
      keystore:
        passphrase: "password"
        resource:
          resource-path: "default-server.p12"
  virtual-hosts:
    - host: "api.example.com"
      tls:
        private-key:
          keystore:
            passphrase: "password"
            resource:
              resource-path: "api-server.p12"
    - host: "*.example.org"
      tls:
        private-key:
          keystore:
            passphrase: "password"
            resource:
              resource-path: "wildcard-server.p12"
  sni:
    missing: "FALLBACK"
    unmatched: "FALLBACK"
    authority-mismatch: "REJECT"
    fallback-authority: "REJECT"
```

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
@@ -63,6 +63,8 @@ import io.helidon.webserver.ConnectionContext;
 import io.helidon.webserver.ErrorHandling;
 import io.helidon.webserver.Router;
 import io.helidon.webserver.ServerConnectionException;
+import io.helidon.webserver.SniContext;
+import io.helidon.webserver.SniRequestSupport;
 import io.helidon.webserver.http.HttpRouting;
 import io.helidon.webserver.http2.spi.Http2SubProtocolSelector;
 import io.helidon.webserver.http2.spi.SubProtocolResult;
@@ -101,6 +103,7 @@ class Http2ServerStream implements Runnable, Http2Stream {
     private boolean hasEntity = true;
     private volatile Http2Headers headers;
     private volatile Http2Priority priority;
+    private volatile boolean ignoreInboundDataAfterReset;
     // used from this instance and from connection
     private volatile Http2StreamState state = Http2StreamState.IDLE;
     private Http2SubProtocolSelector.SubProtocolHandler subProtocolHandler;
@@ -160,6 +163,9 @@ class Http2ServerStream implements Runnable, Http2Stream {
      * @throws Http2Exception in case data cannot be received
      */
     public void checkDataReceivable() throws Http2Exception {
+        if (ignoreInboundDataAfterReset) {
+            return;
+        }
         if (!DATA_RECEIVABLE_STATES.contains(state)) {
             throw new Http2Exception(Http2ErrorCode.PROTOCOL, "Received data for stream "
                     + streamId + " in state " + state);
@@ -257,6 +263,10 @@ class Http2ServerStream implements Runnable, Http2Stream {
 
     @Override
     public void data(Http2FrameHeader header, BufferData data, boolean endOfStream) {
+        if (!DATA_RECEIVABLE_STATES.contains(state)) {
+            flowControl.inbound().incrementWindowSize(header.length());
+            return;
+        }
         if (expectedLength != -1 && expectedLength < header.length()) {
             state = Http2StreamState.CLOSED;
             writeState.updateAndGet(s -> s.checkAndMove(WriteState.END));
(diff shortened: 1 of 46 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/ClientHelloPrefaceReader.java
```diff
@@ -0,0 +1,343 @@
...
+        void awaitReadable() throws IOException;
+    }
+
+    private static final class DeadlineWaiter implements ReadWaiter {
+        private static final long NANOS_PER_MILLI = 1_000_000;
+
+        private final Selector selector;
+        private final long deadlineNanos;
+
+        private DeadlineWaiter(Selector selector, Duration timeout) {
```
danielkec, line 323: `PT0S` is documented as an indefinite read timeout, but this path turns it into an already-expired deadline. `ConnectionHandler` passes `listenerConfig.connectionOptions().readTimeout()` here before TLS owns the socket, and `Duration.ZERO` makes the first nonblocking read that returns 0 fail immediately instead of waiting for the ClientHello. A listener with SNI virtual hosts and `connection-options.read-timeout: PT0S` can therefore close a valid connection before TLS starts.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -539,11 +560,56 @@ private BufferData readEntityFromPipeline() {
...
+        }
+    }
+
     private void handle() {
         Headers httpHeaders = headers.httpHeaders();
         if (httpHeaders.containsToken(HeaderValues.EXPECT_100)) {
             writeState.updateAndGet(s -> s.checkAndMove(WriteState.EXPECTED_100));
         }
+        ctx.sniContext().ifPresent(sniContext -> {
+            String authority = headers.authority();
```
danielkec, line 601: This checks only `:authority`, but raw HTTP/2 input can still carry a conflicting regular `Host` header in `headers.httpHeaders()`. For example, SNI `api.example.com`, `:authority: api.example.com`, and `host: admin.example.com` passes this SNI check, while later `Http2ServerRequest.requestedUri()` can still discover the authority from the `Host` header. That leaves request code seeing a host value that the SNI policy did not validate.

danielkec (commented):
Two review issues from the current-head pass.

webserver/webserver/src/main/java/io/helidon/webserver/ClientHelloPrefaceReader.java
tomas-langer: Fixed in 36ff9e2b63. `Duration.ZERO` is now treated as the documented indefinite timeout in the ClientHello pre-read path: the selector waits without a timeout instead of computing an already-expired deadline. Added regression coverage in `ClientHelloPrefaceReaderTest` that starts the read with `Duration.ZERO` and writes the ClientHello afterwards.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
tomas-langer: Confirmed valid and fixed in 36ff9e2b63. Raw inbound HTTP/2 can preserve a regular `Host` header alongside `:authority`, so the fix is in `Http2Headers.validateRequest()`: it rejects multiple regular `Host` headers and rejects a regular `Host` that does not normalize to the same authority as `:authority` before WebServer routing or SNI policy can observe the request. Added inbound HPACK regression coverage in `Http2HeadersTest`.

danielkec (approved):
Reviewed current head 36ff9e2b6338dda7f1c7c4eb3714d34cbf1ff485. Prior review comments are addressed; focused local validation and live checks are green.

PR #12075, opened 2026-06-05
Resolve metrics API deprecations (27.x)

### LINKED ISSUE #11476: Resolve deprecations in module `helidon-metrics-api` (27.x)
- [ ] io.helidon.metrics.api.GcTimeType.GcTimeType:21
- [ ] io.helidon.metrics.api.Metrics#globalRegistry:46
- [ ] io.helidon.metrics.api.Metrics#createMeterRegistry:58
- [ ] io.helidon.metrics.api.Metrics#createMeterRegistry:70
- [ ] io.helidon.metrics.api.Metrics#getOrCreate:86
- [ ] io.helidon.metrics.api.Metrics#getCounter:100
- [ ] io.helidon.metrics.api.Metrics#getCounter:113
- [ ] io.helidon.metrics.api.Metrics#getSummary:127
- [ ] io.helidon.metrics.api.Metrics#getSummary:140
- [ ] io.helidon.metrics.api.Metrics#getGauge:154
- [ ] io.helidon.metrics.api.Metrics#getGauge:168
- [ ] io.helidon.metrics.api.Metrics#getTimer:183
- [ ] io.helidon.metrics.api.Metrics#getTimer:196
- [ ] io.helidon.metrics.api.Metrics#get:216
- [ ] io.helidon.metrics.api.Metrics#tag:229
- [ ] io.helidon.metrics.api.Metrics#tags:242
- [ ] io.helidon.metrics.api.MetricsConfigBlueprint#createTags:74
- [ ] io.helidon.metrics.api.MetricsConfigBlueprint#createTags:86
- [ ] io.helidon.metrics.api.MetricsConfigBlueprint#restRequestEnabledShadow:176
- [ ] io.helidon.metrics.api.MetricsConfigBlueprint#gcTimeType:217

tomas-langer:
Resolves #11476

Deprecates the shared metrics factory accessors and aligns internal usage with service registry lookups or programmatic factory creation. Also updates tests and compatibility wiring for the metrics wrapper and service-managed factory lifecycle.


danielkec (commented):
Review notes for the metrics lifecycle and scoping issues.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsFactory.java
```diff
@@ -51,26 +52,33 @@ public interface MetricsFactory {
...
-     * @return current or new metrics factory
+     * @return shared metrics factory
+     * @deprecated since 27.0.0, for removal. Use
+     * {@link io.helidon.service.registry.Services#get(java.lang.Class) Services.get(MetricsFactory.class)} for the
+     * shared metrics factory, or create non-global metrics objects using the programmatic API.
      */
+    @Deprecated(since = "27.0.0", forRemoval = true)
     static MetricsFactory getInstance() {
-        return MetricsFactoryManager.getMetricsFactory();
+        return Services.get(MetricsFactory.class);
```
danielkec, line 59: Returning the service-registry singleton here means `MetricsFactory.closeAll()` can close the instance without removing or replacing it. After that, the same singleton can lazily recreate its global registry, but the Micrometer provider has already cleared its tracked factory list, so a later `closeAll()` no longer closes the revived registry. A small reproducer is: get the service factory, call `closeAll()`, call `globalRegistry()` on the retained instance, then call `closeAll()` again and observe the revived registry is not closed.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -126,6 +127,7 @@ private MMeterRegistry(io.micrometer.core.instrument.MeterRegistry delegate,
         this.clock = clock;
         this.metricsFactory = metricsFactory;
         this.metricsConfig = metricsConfig;
+        this.systemTagsManager = SystemTagsManager.create(metricsConfig);
```
danielkec: Creating a per-registry `SystemTagsManager` here also fires the global `SystemTagsManager.onChange` subscribers. `SystemTagsMeterFilterManager` responds by replacing the mutable tag list used by the Micrometer global `commonTags` filter, so creating a custom registry with different app/global tags can affect tags added to later meters in unrelated registries. This is observable because Micrometer keeps the `Iterable<Tag>` and reads it when mapping new meter IDs.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
```diff
@@ -115,7 +117,7 @@ Optional<?> output(ServerRequest req,
                        Iterable<String> nameSelection) {
         MeterRegistryFormatter formatter = chooseFormatter(meterRegistry,
                                                            mediaType,
-                                                           SystemTagsManager.instance().scopeTagName(),
+                                                           Services.get(SystemTagsManager.class).scopeTagName(),
```
danielkec: This scope tag name comes from the shared/root `SystemTagsManager`, but the meter registry above can be created from this observer's `metricsConfig`. If the observer config uses a custom scoping tag name, Prometheus `?scope=` filtering looks for the shared tag instead of the tag used when the registry registered the meters, so scoped text/OpenMetrics requests can return no matching meters. Reproducer: configure the metrics observer with a non-default scoping tag name and request `/metrics?scope=<scope>` as text/OpenMetrics.

docs/src/main/asciidoc/se/metrics/metrics.adoc
```diff
@@ -132,12 +132,24 @@ include::{sourcedir}/se/metrics/MetricsSnippets.java[tag=snippet_1, indent=0]
...
 . Use the {metric} reference returned from the `MeterRegistry` to update the {metric} or get its value.
 
 You can also use the `MeterRegistry` to remove an existing {metric}.
 
+Helidon 27 removes the former static convenience methods from `Metrics`.
+Use these replacements instead:
+
+* Replace `Metrics.globalRegistry()` with injection of `MeterRegistry` or `Services.get(MeterRegistry.class)`.
+* Replace `Metrics.createMeterRegistry()` and `Metrics.createMeterRegistry(metricsConfig)` with
+  `MeterRegistry.create()` and `MeterRegistry.create(metricsConfig)`.
```
danielkec: This migration guidance still points users to APIs deprecated for removal in this PR: both `MeterRegistry.create(...)` and `Tag.create(...)` are now deprecated. Following this list leaves callers with new 27.x deprecation warnings instead of a clean replacement path.

danielkec (commented):
One additional migration-docs note.

danielkec (commented):
Follow-up review notes for the current metrics deprecation fixes.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactoryProvider.java
```diff
@@ -28,61 +31,103 @@
...
-        Metrics.globalRegistry.config().onMeterRemoved(this::onMeterRemoved);
-    }
+        private void configure() {
+            if (configured.compareAndSet(false, true)) {
+                Metrics.globalRegistry.config().onMeterAdded(this::onMeterAdded);
+                Metrics.globalRegistry.config().onMeterRemoved(this::onMeterRemoved);
+                Metrics.globalRegistry.config().meterFilter(new MeterFilter() {
+                    @Override
+                    public Meter.Id map(Meter.Id id) {
+                        List<io.micrometer.core.instrument.Tag> tags = liveFactories().stream()
```
danielkec: This global filter still makes system tags depend on a live factory selected at meter-mapping time, not on the registry creating the meter. All programmatic registries are still built on `Metrics.globalRegistry`, and `createMeterRegistry(...)` updates the factory-wide `systemTagsManager` through `save(...)`. A small reproducer is: create registry A with `metrics.tags=a=1`, create registry B with `metrics.tags=b=2`, then register a new meter through A; the Micrometer ID/output can pick up B's tags.

docs/src/main/asciidoc/se/metrics/metrics.adoc
```diff
@@ -132,12 +132,25 @@ include::{sourcedir}/se/metrics/MetricsSnippets.java[tag=snippet_1, indent=0]
...
 . Use the `MeterRegistry` instance to register new {metrics} and look up previously-registered {metrics}.
 . Use the {metric} reference returned from the `MeterRegistry` to update the {metric} or get its value.
 
 You can also use the `MeterRegistry` to remove an existing {metric}.
 
+Helidon 27 removes the former static convenience methods from `Metrics`.
+Use these replacements instead:
+
+* Replace `Metrics.globalRegistry()` with injection of `MeterRegistry` or `Services.get(MeterRegistry.class)`.
+* Replace `Metrics.createMeterRegistry()` and `Metrics.createMeterRegistry(metricsConfig)` with
```
danielkec: The no-arg and config-taking helpers did not have the same semantics. `Metrics.createMeterRegistry()` created a new non-global registry with default config, but this grouped replacement can send those callers to `globalRegistry()`, moving isolated registry state into the shared registry. Reproducer: code that used `Metrics.createMeterRegistry()` for a private test registry follows the first replacement and starts registering meters in the shared global registry.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -857,7 +819,7 @@ private Optional<io.helidon.metrics.api.Meter> internalRemove(io.helidon.metrics
 
     private io.helidon.metrics.api.Meter.Id neutralIdWithoutSystemTags(Meter.Id micrometerId) {
         return MMeter.PlainId.create(micrometerId.getName(),
-                                     SystemTagsManager.instance().withoutSystemTags(MTag.neutralTags(micrometerId.getTags())));
+                                     systemTagsManager.withoutSystemTags(MTag.neutralTags(micrometerId.getTags())));
```
Verdent: This neutralizes IDs using this registry's system-tag key set for every Micrometer meter the registry observes. The global observer forwards each global-registry addition to every live factory, so two registries that use the same system tag key with different values collapse to the same neutral ID.

A minimal case is registry A configured with `cluster=one` and registry B with `cluster=two`: after A creates the meter, B's `getOrCreate` returns the Micrometer meter tagged `cluster=one`. Keep foreign system-tag values from collapsing into this registry's neutral key, or avoid wrapping non-originating registrations without disambiguation.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactoryProvider.java
```diff
@@ -28,61 +33,120 @@
...
+        private void configure() {
+            if (configured.compareAndSet(false, true)) {
+                Metrics.globalRegistry.config().onMeterAdded(this::onMeterAdded);
+                Metrics.globalRegistry.config().onMeterRemoved(this::onMeterRemoved);
+                Metrics.globalRegistry.config().meterFilter(new MeterFilter() {
+                    @Override
+                    public Meter.Id map(Meter.Id id) {
+                        List<io.micrometer.core.instrument.Tag> tags = CURRENT_SYSTEM_TAGS.get();
+                        if (tags == null || tags.isEmpty()) {
+                            return id;
```
Verdent: When `CURRENT_SYSTEM_TAGS` is unset, this filter returns the Micrometer ID unchanged. Direct registrations through `io.micrometer.core.instrument.Metrics.globalRegistry` do not run inside `MMeterRegistry.getOrCreate` / `withSystemTags`, so configured `metrics.tags` and app tags are no longer applied to those meter IDs.

A minimal case is initializing the global registry with `cluster=blue` and then registering a counter directly via Micrometer; the resulting Micrometer ID has no `cluster` tag. Preserve a configured fallback for direct Micrometer registrations, or add equivalent registration context and test coverage for that path.

Verdent (changes requested):
Two Micrometer integration regressions need fixes before merge. Both are in the system-tag handling path added by this PR.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
```diff
@@ -66,12 +63,20 @@ class MetricsFeature {
...
     MetricsFeature(MetricsObserverConfig config) {
         this.metricsObserverConfig = config;
-        this.metricsConfig = config.metricsConfig();
-        this.meterRegistry = config.meterRegistry().orElseGet(() -> MetricsFactory.getInstance().globalRegistry(metricsConfig));
+        Optional<MeterRegistry> configuredMeterRegistry = config.meterRegistry();
+        if (configuredMeterRegistry.isPresent()) {
+            this.metricsConfig = config.metricsConfig();
+            this.meterRegistry = configuredMeterRegistry.get();
+        } else {
+            this.metricsConfig = Services.get(MetricsFactory.class).metricsConfig();
```
tvallin: Using the factory config here discards a programmatic `MetricsObserverConfig.metricsConfig()` whenever no custom registry is supplied. This config controls `enabled`, `permitAll`, and `roles`, so an observer configured as disabled or protected can inherit permissive ambient defaults. Preserve the observer-facing settings separately from registry-scoped settings and add coverage for a custom observer config using the shared registry.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/MetricsFeature.java
```diff
@@ -66,12 +63,20 @@ class MetricsFeature {
...
     private KeyPerformanceIndicatorSupport.Metrics kpiMetrics;
 
     MetricsFeature(MetricsObserverConfig config) {
         this.metricsObserverConfig = config;
-        this.metricsConfig = config.metricsConfig();
-        this.meterRegistry = config.meterRegistry().orElseGet(() -> MetricsFactory.getInstance().globalRegistry(metricsConfig));
+        Optional<MeterRegistry> configuredMeterRegistry = config.meterRegistry();
+        if (configuredMeterRegistry.isPresent()) {
+            this.metricsConfig = config.metricsConfig();
+            this.meterRegistry = configuredMeterRegistry.get();
```
tvallin: Now that custom registries are isolated, `KeyPerformanceIndicatorMetricsImpls` still caches KPI objects using only the meter-name prefix. Two observers backed by different registries both use `requests.`, so the second reuses an object bound to the first registry and records its requests there. Include registry identity in the cache key, avoid clearing unrelated entries when one observer closes, and test two simultaneous custom registries.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsFactory.java
```diff
@@ -329,8 +355,8 @@ default Meter noOpMeter(Meter.Builder<?, ?> builder) {
         } else {
             throw new IllegalArgumentException("Unrecognized meter builder type " + builder.getClass().getName());
         }
-        SystemTagsManager.instance()
-                .effectiveScope(builder.scope())
+        builder.scope()
+                .or(() -> metricsConfig().scoping().defaultValue())
```
tvallin, line 359: `noOpMeter` uses the factory-wide default scope, but `MMeterRegistry` also invokes it for custom registries whose filtering uses registry-specific configuration. If the custom registry has a different default scope and disables a meter, the returned no-op meter reports the factory's scope. Resolve the effective scope from the target registry before constructing the no-op meter and test differing factory and registry defaults.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactory.java
```diff
@@ -353,15 +382,198 @@ private <B extends Meter.Builder<B, M>, M extends Meter> MeterRegistry applyMete
...
+            metersProviders.forEach(provider -> {
+                if (provider instanceof AutoCloseable closeable) {
+                    try {
+                        closeable.close();
+                    } catch (Exception e) {
+                        LOGGER.log(System.Logger.Level.WARNING, "Error closing metrics meter provider", e);
+                    }
+                }
+            });
+            registries.forEach(MMeterRegistry::close);
```
tvallin: This stops closing registries as soon as one `close()` throws. The `finally` block still sets `closeComplete`, so later calls cannot clean up the remaining registries and publisher resources. Attempt every close, aggregate failures, and report them after cleanup.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcChannel.java
```diff
@@ -37,6 +46,13 @@ public class GrpcChannel extends Channel {
      */
     GrpcChannel(GrpcClient grpcClient) {
         this.grpcClient = (GrpcClientImpl) grpcClient;
+        if (this.grpcClient.clientConfig().enableMetrics()) {
+            this.metricsFactory = Services.get(MetricsFactory.class);
```
tvallin: Resolve the registry first and obtain its factory from `meterRegistry.metricsFactory()`. The low-weight registry and factory services can be independently overridden, so these separate lookups can pair a custom registry with a different factory and pass foreign builders to it. Please audit the other migrated components using the same adjacent lookup pattern.

tvallin (dismissed):

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactory.java
```diff
@@ -353,15 +382,215 @@ private <B extends Meter.Builder<B, M>, M extends Meter> MeterRegistry applyMete
...
+                        closeable.close();
+                    } catch (Exception e) {
+                        LOGGER.log(System.Logger.Level.WARNING, "Error closing metrics meter provider", e);
+                    }
+                }
+            });
+            Throwable closeFailure = null;
+            for (MMeterRegistry registry : registries) {
+                try {
+                    registry.close();
```
tvallin, line 523: Continuing this outer loop does not make cleanup exhaustive within one registry. `MMeterRegistry.close()` delegates to Micrometer's `CompositeMeterRegistry.close()`, which stops closing child publishers after the first exception. Because the Helidon registry is already marked closed and removed from the factory, remaining publishers in that registry cannot be retried. Close each child publisher defensively, aggregate failures, and add coverage using one registry with multiple failing publishers.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/KeyPerformanceIndicatorMetricsImpls.java
```diff
@@ -115,7 +123,7 @@ public void onRequestReceived() {
         @Override
         public void close() {
             meters.forEach(meterRegistry::remove);
-            KPI_METRICS.clear();
+            KPI_METRICS.remove(key, this);
```
tvallin: `get` returns the same cached metrics object to observers using the same registry and prefix, but each observer independently calls `close()` during shutdown. Stopping one WebServer therefore removes the shared meters and cache entry while another server still holds the deleted object, so its requests are no longer exported. Preserve the meters until the final owner closes and add same-registry, two-observer lifecycle coverage.

tvallin (dismissed):

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatterProvider.java
```diff
@@ -44,7 +44,8 @@ public Optional<MeterRegistryFormatter> formatter(MediaType mediaType,
                                                       Optional<String> scopeTagName,
                                                       Iterable<String> scopeSelection,
                                                       Iterable<String> nameSelection) {
-        return matches(mediaType, MediaTypes.TEXT_PLAIN) || matches(mediaType, MediaTypes.APPLICATION_OPENMETRICS_TEXT)
+        return (matches(mediaType, MediaTypes.TEXT_PLAIN) || matches(mediaType, MediaTypes.APPLICATION_OPENMETRICS_TEXT))
+                && MicrometerPrometheusFormatter.prometheusMeterRegistry(meterRegistry).isPresent()
```
Verdent, line 48: `formatter` is used as a capability probe, but `prometheusMeterRegistry` unconditionally unwraps the supplied registry as a Micrometer registry. Passing a `NoOpMeterRegistry` or another non-Micrometer implementation therefore throws `ClassCastException` instead of returning `Optional.empty()`. I reproduced this with a no-op registry. Because `MetricsFeature.enabled()` probes providers during routing setup, this can prevent server startup.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcChannel.java
Verdent: The response commit corrected this ownership mismatch in `GrpcChannel` and analogous injected paths, but the declarative codegen paths still obtain `MeterRegistry` and `MetricsFactory` independently in `CountedHandler`, `TimedHandler`, and `GaugeHandler`. Using the independently overridden services covered by `GrpcChannelTest`, a no-op registry paired with the Micrometer factory fails during generated-equivalent registration because the Micrometer builder cannot be cast to the no-op builder type.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -126,124 +121,96 @@ private MMeterRegistry(io.micrometer.core.instrument.MeterRegistry delegate,
...
             meters.clear();
             buildersByPromMeterId.clear();
             scopeMembership.clear();
             metersById.clear();
         } finally {
             lock.writeLock().unlock();
+            if (notifyClosed) {
+                Throwable closeFailure = null;
+                try {
+                    if (registeredWithFactory) {
```
Verdent: The registry is removed from the factory here before its publisher registries are closed below. If a publisher close blocks, a concurrent `MicrometerMetricsFactory.close()` can snapshot no registries, set `closeComplete`, and return while publisher cleanup is still running. If the factory snapshots earlier, its second call to this registry’s `close()` returns immediately because `closed` is already true. A latch-blocked publisher reproduced both schedules.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -126,124 +121,96 @@ private MMeterRegistry(io.micrometer.core.instrument.MeterRegistry delegate,
...
 
     @Override
     public void close() {
+        boolean notifyClosed = false;
         lock.writeLock().lock();
         try {
+            if (closed) {
+                return;
+            }
+            closed = true;
```
Verdent: `closed` is consulted only by `close()`; the registration path does not check it. After this method clears the maps, `getOrCreate` can register against the closed Micrometer composite, wrap the resulting no-op meter, and insert it back into the registry. I reproduced `meters().size()` changing from 0 after close to 1 after `getOrCreate`, while the returned counter remained a no-op. An in-flight registration can also overtake the clear between the read and write locks.

tvallin (dismissed):

Verdent (dismissed):

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactory.java
```diff
@@ -353,15 +380,215 @@ private <B extends Meter.Builder<B, M>, M extends Meter> MeterRegistry applyMete
...
+        globalRegistryLock.lock();
+        boolean started = false;
+        try {
+            beginGlobalOperation();
+            started = true;
+            return operation.get();
+        } finally {
+            globalRegistryLock.unlock();
+            if (started) {
+                endGlobalOperation();
```
danielkec, line 445: When `close()` runs during this global operation, it marks the factory closed and defers cleanup because the operation is active. This return value is computed, but `endGlobalOperation()` closes it in the `finally` before the caller receives it, so `globalRegistry()` succeeds with an already-closed registry. A deterministic case is a registry-aware `MetersProvider` whose callback calls `factory.close()`:

```java
Counter.Builder builder = factory.counterBuilder("probe");
MeterRegistry registry = factory.globalRegistry(); // provider callback closes factory
registry.getOrCreate(builder); // IllegalStateException: Meter registry is closed
```

The same post-registration window occurs with concurrent service-registry shutdown, so startup can receive an unusable registry instead of a creation failure.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -941,19 +904,49 @@ public B onMeterRemoved(Consumer<io.helidon.metrics.api.Meter> listener) {
...
-                                                       clock.orElse(MClock.create(delegate.config().clock())));
-
-            onAddListener.ifPresent(result::onMeterAdded);
-            onRemoveListener.ifPresent(result::onMeterRemoved);
-
-            return (R) applyMetersProvidersToRegistry(metricsFactory, result, metersProviders);
+        MMeterRegistry buildRegistry() {
+            CompositeMeterRegistry delegate = clock
+                    .map(value -> new CompositeMeterRegistry(ClockWrapper.create(value)))
+                    .orElseGet(CompositeMeterRegistry::new);
```
danielkec, line 925: This independent composite disconnects the service-owned Helidon registry from `io.micrometer.core.instrument.Metrics.globalRegistry`, but the published Micrometer guide still creates application meters on that native global and says Helidon exposes them. On this head:

```java
var nativeRegistry = new SimpleMeterRegistry();
Metrics.addRegistry(nativeRegistry);
Metrics.globalRegistry.counter("native_only").increment();
Services.get(io.helidon.metrics.api.MeterRegistry.class)
        .meters().stream().anyMatch(m -> m.id().name().equals("native_only")); // false
```

The native counter reaches `1.0` while it remains absent from Helidon output, so existing documented applications silently stop publishing those meters. The earlier direct-registration thread covered missing system tags; this is the distinct loss of the meter itself.

danielkec (commented):
Two current-head issues remain after live semantic deduplication.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerMetricsFactory.java
tomas-langer: Will not fix this as part of this PR. This is an obscure shutdown scenario, and metrics lifecycle will receive additional work.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
tomas-langer: This should now be fixed in documentation.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/KeyPerformanceIndicatorMetricsImpls.java
```diff
@@ -77,10 +79,16 @@ static KeyPerformanceIndicatorSupport.Metrics get(MeterRegistry kpiMeterRegistry
                                                       String meterNamePrefix,
                                                       KeyPerformanceIndicatorMetricsConfig kpiConfig,
                                                       BuiltInMeterNameFormat builtInMeterNameFormat) {
-        return KPI_METRICS.computeIfAbsent(meterNamePrefix, prefix ->
-                kpiConfig.extended()
-                        ? new Extended(kpiMeterRegistry, meterNamePrefix, kpiConfig, builtInMeterNameFormat)
-                        : new Basic(kpiMeterRegistry, meterNamePrefix, builtInMeterNameFormat));
+        KpiMetricsKey key = new KpiMetricsKey(kpiMeterRegistry, meterNamePrefix);
```
danielkec, line 82: When two observers use the same registry and prefix but different KPI settings, this key aliases them and the later observer silently reuses the first observer's definition. The exact-head probe shows:

```text
get(registry, "requests.", extended=false, CAMEL)
get(registry, "requests.", extended=true, SNAKE)
same instance = true
meters = [requests.count]
```

The second observer therefore does not export its requested extended metrics; the name format and long-running threshold are likewise determined by whichever observer initializes the entry first.

danielkec (commented):
`helidon-version-is-release` is currently `false`, so the example link in `docs/src/main/asciidoc/includes/metrics/micrometer-shared.adoc:104` resolves to https://github.com/helidon-io/helidon-examples/tree/helidon-4.x/examples/integrations/micrometer/se. That live example still depends on `helidon-integrations-micrometer`, constructs `MicrometerFeature`, and verifies `/micrometer`, while this rewritten page documents the service-owned registry and `/observe/metrics`; following the link therefore sends 27.x readers to the obsolete integration path.

tomas-langer:
The 27.x examples branch does not exist yet because 27.x has not been released. Until that branch is created, the documentation attribute intentionally resolves example links to the latest released branch. There is no valid 27.x example target we can use in this PR.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/KeyPerformanceIndicatorMetricsImpls.java
tomas-langer: Sharing one meter registry between observers with conflicting registry-specific settings is not supported. A registry has one effective configuration; observers can share it only when those settings are equivalent. Observers requiring different registry behavior must use separate registries. This behavior predates this PR. I have prepared Javadoc on MetricsObserverConfig.meterRegistry() clarifying the constraint and corrected its example.

danielkec (approved):

PR #12076, opened 2026-06-05
Deprecate BufferData#asInputStream() (27.x)

### LINKED ISSUE #11788: Remove BufferData#asInputStream()
Helidon Version: Affects 4.x but likely best targeted at 27

## Enhancement Description
Remove the unused `BufferData#asInputStream()` method. It was added as part of gRPC improvements #10147 but has since been replaced in gRPC code by the gRPC module's `BufferDataInputStream` in #10797. No other Helidon code uses `asInputStream()` and the `InputStream` it produces is rather inefficient. All verifications pass without the method.

```java
default InputStream asInputStream() {
    return new InputStream() {
        @Override
        public int read() {
            return BufferData.this.available() > 0 ? BufferData.this.read() : -1;
        }
    };
}
```

It's not hurting anyone, but it's also not adding any value. I'd like to contribute the PR for this if approved, but this is a public API and I don't know if the maintainers want such a change. If so, please let me know if you want 27 only, or also for 4.x

tomas-langer:
Resolves #11788

Deprecates `BufferData#asInputStream()` for removal in 27.0.0 and documents that the method will be removed without replacement.

### DIFF
--- a/common/buffers/src/main/java/io/helidon/common/buffers/BufferData.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/BufferData.java
@@ -684,7 +684,9 @@ public interface BufferData {
      * or rewinds the buffer.
      *
      * @return an input stream backed by this buffer
+     * @deprecated this method will be removed without replacement
      */
+    @Deprecated(forRemoval = true, since = "27.0.0")
     default InputStream asInputStream() {
         return new InputStream() {
             @Override


danielkec (approved):
Reviewed exact head 48ec1198c05e5593130fa8651ddb4ad837fd8b24. No blocking findings.

PR #12079, opened 2026-06-08
Reloadable TLS

### LINKED ISSUE #10785: [4.x] Reloadable TLS
Helidon MP - We should expose WebServer#reloadTls over ServerCDIExtension
Helidon SE -  reloadTls should not accept tls config with custom sslcontext

tomas-langer:
Resolves #10785

Adds reloadable TLS material support across common TLS managers and WebServer listener APIs, including material-first reload methods, virtual-host/SNI reload support, explicit SSLContext reload rejection, documentation, and configuration-backed custom manager reload coverage.

### DIFF
--- a/common/tls/src/main/java/io/helidon/common/tls/Tls.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/Tls.java
@@ -70,7 +70,8 @@ public class Tls implements RuntimeType.Api<TlsConfig> {
     private Tls(TlsConfig config) {
         // at this time, the TlsConfigDecorator should have created SSL parameters; the SSL context is the responsibility
         // of the manager to provide
-        this.tlsConfig = Objects.requireNonNull(config);
+        config = Objects.requireNonNull(config);
+        TlsConfig configuredTls = config;
         this.sslParameters = copySslParameters(config.sslParameters().orElseThrow());
         this.enabled = config.enabled();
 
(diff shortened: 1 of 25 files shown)


common/tls/src/main/java/io/helidon/common/tls/Tls.java
```diff
@@ -268,13 +277,28 @@ public SSLParameters sslParameters() {
      * Reload reloadable {@link TlsReloadableComponent}s with the new configuration.
      *
      * @param tls new TLS configuration
+     * @deprecated use {@link #reload(TlsMaterial)}
      */
+    @Deprecated
+    @SuppressWarnings("removal")
     public void reload(Tls tls) {
```
danielkec, line 284: This deprecated overload is still public and can still receive a `Tls` built from an explicit `SSLContext`. In that case the replacement `Tls` uses `ExplicitContextTlsManager`, whose key/trust managers are empty, so `ConfiguredTlsManager.reload(Tls)` reaches this path and returns as a successful no-op instead of rejecting the custom context. Minimal shape: `current.reload(Tls.create(b -> b.sslContext(ctx)))`. That leaves the SE/WebServer `reloadTls(Tls)` path accepting exactly the explicit-`SSLContext` input that #10785 says should be rejected.

tomas-langer: Thanks, fixed in 9ce53a02ee4ee79479a2376765f1356229f3678f. Added a regression for `current.reload(Tls.create(b -> b.sslContext(ctx)))` in `ConfiguredTlsManagerTest` and validate deprecated `Tls.reload(Tls)` sources before delegating, so explicit `SSLContext` replacements now throw the same unsupported-reload exception instead of becoming an empty-material no-op.

danielkec (approved):
Reviewed current head 9ce53a02ee4ee79479a2376765f1356229f3678f. No blocking findings.

PR #12081, opened 2026-06-08
Fix HTTP/1 no-body and close-delimited responses (27.x)

### LINKED ISSUE #12080: Treat HEAD responses as empty in WebClient entity reads
`webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java` can expose a readable entity for a `HEAD` response when the server sends no `Content-Length`. If application code reads that entity, the HTTP/1 client waits for connection EOF to delimit the response body. On a persistent connection, EOF does not arrive, so the read can block until the socket read timeout even though `HEAD` responses do not carry response content.

Expected behavior: WebClient should treat responses to `HEAD` requests as having an empty entity based on the request method, regardless of whether `Content-Length` is present, and should not wait for EOF to discover the end of a non-existent response body.

Impact: callers that read or close a `HEAD` response entity can observe avoidable timeouts, and the client may close otherwise reusable connections after timing out.

tomas-langer:
Resolves #12080

Treat HEAD and other no-body HTTP/1 responses as empty when building WebClient responses, and close unframed close-delimited responses instead of returning those connections to the cache. This also preserves HEAD route coverage in WebServer tests while asserting that HEAD responses have no entity.

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
@@ -19,6 +19,7 @@ package io.helidon.webclient.http1;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.UncheckedIOException;
+import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.CompletableFuture;
@@ -409,14 +410,24 @@ class Http1CallOutputStreamChain extends Http1CallChainBase {
                     connection.readTimeout(originalRequest.readContinueTimeout());
                     responseHead = callChain.readResponseHead(connection,
                                                               reader,
-                                                              Http1CallChainBase::isPreContinueInterimResponse);
-                } catch (UncheckedIOException ignored) {
-                    // we assume this is a timeout exception, if the socket got closed, next read will throw appropriate exception
-                    // we treat this as receiving 100-Continue
-                    responseHead = null;
-                    connection.allowExpectContinue(false);
+                                                              Http1CallChainBase::isPreContinueInterimResponse,
+                                                              false);
+                } catch (UncheckedIOException e) {
+                    if (e.getCause() instanceof SocketTimeoutException) {
+                        responseHead = null;
+                        connection.allowExpectContinue(false);
+                    } else {
+                        try {
+                            connection.closeResource();
+                        } catch (Exception ex) {
+                            e.addSuppressed(ex);
+                        }
+                        throw e;
+                    }
                 } finally {
-                    connection.readTimeout(originalRequest.readTimeout());
+                    if (connection.isConnected()) {
+                        connection.readTimeout(originalRequest.readTimeout());
+                    }
                 }
 
                 Status responseStatus = responseHead == null ? Status.CONTINUE_100 : responseHead.status();
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
@@ -402,14 +402,14 @@ class Http1CallOutputStreamChain extends Http1CallChainBase {
             whenSent.complete(request);
 
             if (expects100Continue) {
-                ResponseHead responseHead;
+                ResponseHead responseHead = null;
 
                 try {
                     writer.flush();     // flush before a read
                     connection.readTimeout(originalRequest.readContinueTimeout());
                     responseHead = callChain.readResponseHead(connection,
                                                               reader,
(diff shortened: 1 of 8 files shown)


tvallin:
I think this can regress valid HTTP/1 informational response handling.

`readResponse` reads a single status/header block, and the new `mayHaveEntity` logic classifies all `1xx` responses as no-entity responses. For a valid response like `103 Early Hints` followed by `200 OK`, the client returns `103` as the final response instead of skipping interim responses and parsing the final one.

I added this test to demonstrate it; it fails on this PR with `Expected: 200 OK` but `was: 103 Early Hints`:

```java
@Test
void testInterimResponseIsSkipped() {
    String rawResponse = "HTTP/1.1 103 Early Hints\r\n"
            + "Link: </style.css>; rel=preload\r\n"
            + "\r\n"
            + "HTTP/1.1 200 OK\r\n"
            + "Content-Length: 2\r\n"
            + "\r\n"
            + "OK";

    try (Http1ClientResponse response = client.get("http://localhost:" + dummyPort + "/test")
            .connection(responseConnection(rawResponse))
            .request()) {
        assertThat(response.status(), is(Status.OK_200));
        assertThat(response.as(String.class), is("OK"));
    }
}

private static ClientConnection responseConnection(String rawResponse) {
    ArrayBlockingQueue<byte[]> response = new ArrayBlockingQueue<>(1);
    response.add(rawResponse.getBytes(StandardCharsets.US_ASCII));
    DataReader reader = DataReader.create(response::poll);
    DataWriter writer = new DataWriter() {
        @Override
        public void write(BufferData... buffers) {
        }

        @Override
        public void write(BufferData buffer) {
        }

        @Override
        public void writeNow(BufferData... buffers) {
        }

        @Override
        public void writeNow(BufferData buffer) {
        }
    };

    return new ClientConnection() {
        @Override
        public HelidonSocket helidonSocket() {
            return new FakeSocket();
        }

        @Override
        public DataReader reader() {
            return reader;
        }

        @Override
        public DataWriter writer() {
            return writer;
        }

        @Override
        public void releaseResource() {
        }

        @Override
        public void closeResource() {
        }

        @Override
        public String channelId() {
            return "preloaded-response";
        }

        @Override
        public void readTimeout(Duration readTimeout) {
        }
    };
}
```

The client should skip interim `1xx` responses until it reaches the final response, with `101 Switching Protocols` handled as the special upgrade case.

tvallin (commented):
The `Expect: 100-continue` timeout path now closes the connection before the client sends the request body. `Http1CallOutputStreamChain` calls `readResponseHead` while waiting for an interim response, and `readResponseHead` closes the connection whenever `Http1StatusParser.readStatus` throws an `UncheckedIOException`. The caller then catches that same exception as the normal "no `100 Continue` received before `readContinueTimeout`" case and proceeds as if it can send the body, but the connection has already been closed. Servers are allowed to omit `100 Continue`, so this regresses requests that should continue after the timeout. The pre-continue read needs a path that does not close the connection on timeout, plus a test where `Expect: 100-continue` times out and the final response still succeeds.

tvallin (dismissed):
Approved. The earlier blocking Expect: 100-continue timeout issue has been addressed in the latest head. I filed follow-up issue #12111 for the narrower edge case where a status line is received before readContinueTimeout but the response headers time out afterward.

tvallin (approved):
Approved. The remaining pre-continue header-timeout edge case from the review is tracked separately in #12111 and is not blocking this PR.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
```diff
@@ -422,37 +403,37 @@ private void sendPrologueAndHeader() {
...
-                } catch (UncheckedIOException ignored) {
-                    // we assume this is a timeout exception, if the socket got closed, next read will throw appropriate exception
-                    // we treat this as receiving 100-Continue
-                    responseStatus = null;
-                    connection.allowExpectContinue(false);
+                    responseHead = callChain.readResponseHead(connection,
+                                                              reader,
+                                                              Http1CallChainBase::isPreContinueInterimResponse,
+                                                              false);
+                } catch (UncheckedIOException e) {
```
Verdent, line 415: This timeout handling is unsafe once `readResponseHead(...)` has consumed any part of a response head. `readResponseHead` only wraps the status-line read; if the server sends a `100`, `103`, or final status line and then stalls while sending headers, `readHeaders(reader)` throws here after the status line has already been consumed. This catch then treats the timeout as "no continue response" and proceeds to send the request body on a desynchronized HTTP/1 connection.

Please treat status+headers as one atomic response-head read. Only map a `SocketTimeoutException` to "no continue response" when no response head bytes/status have been consumed; after status parsing starts, close the connection and propagate the failure.

tomas-langer: This is already tracked here: https://github.com/helidon-io/helidon/issues/12111

PR #12083, opened 2026-06-08
Add WebServer request-time service locator (27.x)

### LINKED ISSUE #81: Support for dynamic matching of resources
## Enhancement
There's currently support to associate a route with a service during route building. At step further would be to associate a route with a lambda that could dynamically return a service, akin to a resource locator in JAX-RS. 

For example,

```
    private static Routing createRouting() {
        return Routing.builder()
                .register("/service/{version}", path -> isV1(path) ? new ServiceV1() : new CurrentService())
                .build();
    }
```

tomas-langer:
Resolves #81

Adds request-time WebServer service location through `HttpServiceLocator`, including routing registration methods, identity-cached located service routing with lifecycle handling, WebServer docs/snippets, focused tests, and JMH coverage for static service routing and locator paths.


spericas (dismissed):
Looks good

webserver/webserver/src/main/java/io/helidon/webserver/http/HttpRules.java
```diff
@@ -39,8 +39,23 @@ public interface HttpRules {
...
+    /**
+     * Register a service locator on the current path.
+     * <p>
+     * Unlike {@link #register(Supplier)}, the locator is invoked at request time after the current path matches.
+     * If the locator returns {@link java.util.Optional#empty()}, routing continues with the next available route.
+     *
+     * @param locator service locator to register
+     * @return updated rules
+     */
+    default HttpRules register(HttpServiceLocator locator) {
```
danielkec: This overload can break existing source that uses one-argument service lambdas. With both `register(HttpService...)` and the new fixed-arity `register(HttpServiceLocator)`, Java can type a call such as `register(rules -> configure(rules))` as the locator overload, so the lambda parameter becomes `ServerRequest` instead of `HttpRules` and the previously valid service registration no longer compiles.

danielkec (commented):
Inline review issues from the exact-head pass.

webserver/webserver/src/main/java/io/helidon/webserver/http/RouteCrawler.java
```diff
@@ -93,32 +105,59 @@ boolean hasNext() {
...
+                            routes = nextRoute.routes();
+                        }
+                        subCrawler = new RouteCrawler(ctx,
+                                                      request,
+                                                      routes,
+                                                      accepts.matchedPath(),
+                                                      matchedPattern,
+                                                      accepts.unmatchedPath());
+                    } else {
+                        subCrawler = new RouteCrawler(ctx,
```
danielkec: This branch loses the parent route pattern when a locator is nested under a parameterized ordinary service route. For example, `register("/{tenant}", service)` where the service does `rules.register("/{item}", locator)` and the located service handles `/{id}` will expose `/t1/{item}/{id}` as `matchingPattern()` for `/t1/pipe/123` instead of `/{tenant}/{item}/{id}`. That leaks request values into route tags and can create high-cardinality observability data.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServiceLocatorRoute.java
```diff
@@ -0,0 +1,304 @@
...
+
+    @Override
+    public String toString() {
+        return methodPredicate + " (" + pathMatcher + ") with service locator: " + locator;
+    }
+
+    private ServiceRoute route(HttpService service) {
+        Objects.requireNonNull(service, "HttpServiceLocator must not locate a null service");
+
+        RouteEntry entry = routeEntry(service);
```
danielkec: There is a first-hit race here after an entry is returned but before this thread takes the owner read lock. Another concurrent first request for the same service can fail lazy route creation, remove the still-empty entry, and rethrow; this thread can then continue with the stale `RouteEntry`, create/start a `ServiceRoute`, and return it even though it is no longer in `routes`. That route will not be visited by `afterStop()`, so its lifecycle cleanup is skipped.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServiceLocatorRoute.java
```diff
@@ -0,0 +1,304 @@
...
+        } finally {
+            lock.readLock().unlock();
+        }
+
+        lock.writeLock().lock();
+        try {
+            if (stopped) {
+                return null;
+            }
+            return routes.computeIfAbsent(service, _ -> new RouteEntry());
```
danielkec: Every distinct service identity returned by the locator is retained in this cache until server stop. If the locator returns fresh service instances based on request-controlled path data, clients can force retained route/lifecycle state growth for each unique value. The docs recommend stable instances, but the runtime does not bound or enforce that contract.

tests/benchmark/jmh/src/test/java/io/helidon/webserver/benchmark/jmh/RoutingJmhRunnerTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.TimeValue;
+
+class RoutingJmhRunnerTest {
+    @Test
+    void run() throws RunnerException {
+        String include = System.getProperty("routing.jmh.include", ".*RoutingJmhTest.*");
```
danielkec: This benchmark appears to be included twice under the `jmh` profile. The existing aggregate runner includes `.*JmhTest`, which matches the new `RoutingJmhTest`, and this dedicated runner also includes `.*RoutingJmhTest.*`, so `-Pjmh test` can run the same routing benchmark twice.

webserver/webserver/src/main/java/io/helidon/webserver/http/RouteCrawler.java
```diff
@@ -93,32 +105,60 @@ boolean hasNext() {
...
+                                    ? matchedPattern
+                                    : parentPattern + matchedPattern;
+                            routes = nextRoute.routes(ctx, request, matchedPath, fullMatchedPattern);
+                        } else {
+                            routes = nextRoute.routes();
+                        }
+                        subCrawler = new RouteCrawler(ctx,
+                                                      request,
+                                                      routes,
+                                                      accepts.matchedPath(),
```
danielkec: This still drops ancestor path parameters for a locator nested below another located service. The current locator receives the merged parent path above, but this child crawler stores only the local `accepts.matchedPath()` and `matchedPattern`. With `/{tenant}` -> locator `/{item}` -> located service registering locator `/{part}`, the inner `part` locator sees `item` and `part`, but not `tenant`, even though the final handler later gets `tenant` after outer wrapping.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServiceLocatorRoute.java
```diff
@@ -0,0 +1,324 @@
...
+                try {
+                    if (stopped) {
+                        return null;
+                    }
+                    entry = routes.get(service);
+                    if (entry == null) {
+                        if (routes.size() >= maxServiceCacheSize) {
+                            throw new HttpException("HttpServiceLocator service cache size of "
+                                                            + maxServiceCacheSize
+                                                            + " exceeded by " + locator,
```
danielkec: This error path can return the locator's `toString()` to the client. An unhandled `HttpException` is converted into a `RequestException` with the same message, and the default direct handler sends safe messages in the response body. If an application locator overrides `toString()` with configuration or other operational state, a client that drives the cache over the limit can see it.

webserver/webserver/src/test/java/io/helidon/webserver/http/HttpServiceLocatorTest.java
```diff
@@ -0,0 +1,681 @@
...
+
+        locate(route, "/fast/1");
+
+        CompletableFuture<Void> slowCreation = CompletableFuture.runAsync(() -> locate(route, "/slow/1"));
+        if (!slow.awaitStarted()) {
+            slow.release();
+            fail("Cold service route creation did not start");
+        }
+
+        CompletableFuture<Void> fastHit = CompletableFuture.runAsync(() -> locate(route, "/fast/2"));
```
danielkec: This test still depends on common-pool parallelism. With `-Djava.util.concurrent.ForkJoinPool.common.parallelism=1`, the fast task can sit behind the intentionally blocked slow task and fail `fastHit.get(1, TimeUnit.SECONDS)` even when route-entry locking is correct.

danielkec (commented):
Current-head inline review issues.

webserver/webserver/src/test/java/io/helidon/webserver/http/HttpServiceLocatorTest.java
tomas-langer: This is a test, we control the JVM setup, and we do not change fork-join pool config in pom.xml, so it is imho not relevant.

webserver/webserver/src/main/java/io/helidon/webserver/http/RouteCrawler.java
tomas-langer: Fixed in the latest push. `RouteCrawler` now rebuilds the full ancestor matched path and matching pattern for request-aware nested routes, including locators nested below located services, while ordinary nested routes keep the local fast path. Added regression coverage for `/{tenant}` -> `/{item}` -> `/{part}`.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServiceLocatorRoute.java
tomas-langer: Fixed in the latest push. The cache-limit `HttpException` message no longer includes the locator instance, and the regression test uses a locator whose `toString()` returns sensitive text and asserts the client-safe message stays generic.

webserver/webserver/src/test/java/io/helidon/webserver/http/HttpServiceLocatorTest.java
tomas-langer: Updated in the latest push. The concurrency test now uses an explicit two-thread executor, and it passes with `-Djava.util.concurrent.ForkJoinPool.common.parallelism=1`.

danielkec (commented):
Current-head inline review issues not already reported.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServiceLocatorRoute.java
```diff
@@ -0,0 +1,324 @@
...
+
+        return pathMatcher.prefixMatch(prologue.uriPath());
+    }
+
+    @Override
+    List<HttpRouteBase> routes(ConnectionContext ctx,
+                               RoutingRequest request,
+                               RoutedPath matchedPath,
+                               String matchingPattern) {
+        request.path(matchedPath);
```
danielkec, line 176: `matchingPattern` is available here but the request only gets the matched path before `locate` runs. A locator can therefore observe a stale pattern from an earlier route attempt. For example, with `get("/{previous}/{id}", (req, res) -> res.next())` before this locator, a `registerLocator("/{item}", ...)` handling `/pipe/123` sees `/{previous}/{id}` from `request.matchingPattern()` instead of the locator's `/{item}`; without a previous nexting handler it sees no pattern.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServiceLocatorRoute.java
```diff
@@ -0,0 +1,324 @@
...
+                        routes.put(service, entry);
+                    }
+                    lock.readLock().lock();
+                    readLockHeld = true;
+                } finally {
+                    lock.writeLock().unlock();
+                }
+            }
+
+            return entry.route(this, service);
```
danielkec: Cold located-service creation can invoke user lifecycle callbacks while this route still holds the owner read lock. If the located service's `afterStart(WebServer)` calls `webServer.stop()`, the stop path reaches `ServiceLocatorRoute.afterStop()` and tries to take the owner write lock while the same request thread is still holding the read lock, so the first request can hang. A minimal reproducer is a locator returning a service whose `afterStart` calls `webServer.stop()`.

tvallin (approved):

PR #12090, opened 2026-06-10
27.x: Remove jandex dependency

barchetta:
### Description

Remove jandex dependency as it is not used.


barchetta:
Superseded by #12093

PR #12092, opened 2026-06-12
Add listener transport binding abstraction

### LINKED ISSUE #11811: Listener binding abstraction
Create a listener abstraction - bindings for TCP (http/2, http/1.1, grpc, websocket) and possible future protocols (such as QUIC).

Must be designed together with #6183

tomas-langer:
Part of #11811.

This is the initial PR for full listener transport bindings. It introduces the transport binding SPI and listener-owned binding contexts, moves the existing TCP listener path behind the new binding planning, start, stop, suspend, resume, and TLS reload flow, and keeps listener TLS, virtual-host reload, protocol selection, and lifecycle orchestration listener-scoped.

Follow-up work will split UDS into its own transport binding and then use the same binding model for QUIC/HTTP/3 support. This PR intentionally does not close #11811.


webserver/webserver/src/main/java/io/helidon/webserver/WebServerConfigSupport.java
```diff
@@ -145,6 +150,21 @@ public void decorate(ListenerConfig.BuilderBase<?, ?> target) {
             if (name == null) {
                 target.name(WebServer.DEFAULT_SOCKET_NAME);
             }
+            if (target.bindAddress().orElse(null) instanceof UnixDomainSocketAddress) {
```
danielkec, line 177: This makes existing documented UDS listener configuration fail during config build. On main, both `server.bind-address: "unix:/tmp/server.sock"` and the builder `bindAddress(UnixDomainSocketAddress.of(path))` are accepted, and the existing UDS tests use that builder path. With this check, the same app fails before start. Reproducer: keep an existing YAML listener with `bind-address: "unix:/tmp/server.sock"` and upgrade to this branch.

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
```diff
@@ -0,0 +1,707 @@
...
+    }
+
+    private void listen() {
+        ServerSocketChannel localServerSocket = serverSocket;
+        String serverChannelId = "0x" + HexFormat.of().toHexDigits(System.identityHashCode(localServerSocket));
+
+        while (running) {
+            try {
+                // this must be done before we accept, and the semaphore must be released when connection is finished
+                var outcome = connectionLimit.tryAcquireOutcome(true);
```
danielkec, line 588: This can block the other transport bindings before any connection is accepted. Each active binding has its own accept loop, but all bindings share the listener `connectionLimit`; with `maxConnections(1)`, the first loop that reaches this line can take the only permit and then block in `accept()`. Reproducer: configure a listener with default TCP plus UDS and `maxConnections(1)`, then connect first to UDS while TCP has no clients. The UDS loop waits on the permit held by the idle TCP accept loop.

danielkec (commented):
Inline review comments for the current head.

webserver/webserver/src/main/java/io/helidon/webserver/UdsTransportBindingFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public boolean required() {
+        return config.required();
+    }
+
+    @Override
+    public boolean canBind(BindingPlanContext context) {
+        Objects.requireNonNull(context, "context");
+        return config.enabled() && config.socket().isPresent();
```
danielkec, line 64: A configured UDS binding with no `socket` is silently ignored here. Because non-required bindings that cannot bind are skipped and listener normalization adds the default TCP binding when TCP was not explicitly configured, a malformed `server.bindings.uds` can start a TCP listener instead of failing the intended UDS setup. Reproducer: configure only `server.bindings.uds: {}` and omit `bindings.tcp.enabled=false`; this branch starts with TCP only.

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
```diff
@@ -73,6 +76,21 @@ interface ListenerConfigBlueprint {
...
+     * Transport bindings that serve this listener.
+     * <p>
+     * When no TCP binding is configured, the listener configuration adds the built-in TCP binding by default. Configured
+     * bindings overlay the built-in defaults and add provider-defined transport bindings.
+     *
+     * @return configured transport bindings
+     */
+    @Option.Configured
+    @Option.Singular
+    @Option.Provider(TransportBindingFactoryProvider.class)
```
danielkec: This makes transport-binding discovery opt-out instead of config-driven. `@Option.Provider` pulls in service-loaded bindings even when the listener has no `bindings` config, and `ProvidedUtil` still creates the missing provider config from the root / `Config.empty()`. Reproducer: add a third-party `TransportBindingFactoryProvider` to the classpath, leave `server.bindings` unset, and start the server; this branch still creates that binding factory and lets it participate in every listener.

webserver/webserver/src/main/java/io/helidon/webserver/WebServerConfigSupport.java
tomas-langer: Yes, this is intended and documented. UDS as we have it right now is combining with TCP listener and has a lot of loose edges.

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
tomas-langer: This is a design decision, that I already once explained when we did listeners. We must first get a permit, before we accept, as this is a security measure. If it is done the other way around, we may accept a socket even when no permit is available. So this stays as is.

webserver/webserver/src/main/java/io/helidon/webserver/UdsTransportBindingFactory.java
tomas-langer: `server.bindings.uds` never starts TCP listener. TCP listener is always started, unless explicitly disabled.
`server.bindings.uds` with no configured value will indeed not fail, and will not start a UDS.
This is basically impossible to resolve, unless we move `uds` to a separate module, which brings a lot of different issues - the main one is that `uds` and `tcp` share a huge amount of code, and if we move it to another module, this would have to become public seam. 

So we have two options:
1. accept explicit UDS with not defined socket
2. make a lot of the code for TCP/UDS split public (or duplicated), and move uds to a separate module, and make the socket required instead of optional (i.e. it would fail if on the classpath and not conifgured)

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
tomas-langer: This is working exactly the same way as protocol discovery - http/2 is also discovered simply by the means of it being available on classpath.

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
danielkec: I understand the pre-accept rationale. The remaining concern is the listener-wide shared permit across multiple bindings, not the ordering within a single binding. A minimal reproducer is a listener with default TCP still enabled, explicit UDS, and `maxConnections(1)`, with no TCP clients connecting:

```java
Path socketPath = Files.createTempDirectory("uds-repro").resolve("server.sock");

        .host("127.0.0.1")
        .port(0)
        .maxConnections(1)
        .bindingsDiscoverServices(false)
        .addBinding(UdsTransportConfig.builder()
                .socket(UnixDomainSocketAddress.of(socketPath))
                .required(true)
                .buildPrototype())
        .routing(rules -> rules.get("/test", (req, res) -> res.send("ok")))
        .build()
        .start();

WebClient.create()
        .get()
        .address(UnixDomainSocketAddress.of(socketPath))
        .path("/test")
        .request(String.class); // can hang until the test timeout fires
```

If the idle TCP accept loop acquires the single listener permit first, the UDS loop never reaches `accept()`. There are still zero active connections, but the first UDS request can block behind the idle TCP binding.

danielkec: For future reference, configuring lower maxConnections than number of binding is considered as a missconfiguration

webserver/webserver/src/main/java/io/helidon/webserver/UdsTransportBindingFactory.java
danielkec: i like option 1

config/config/src/main/java/io/helidon/config/ProvidedUtil.java
```diff
@@ -416,6 +418,19 @@ private static ConfiguredService configuredService(Config serviceConfig, boolean
         return new ConfiguredService(new TypeAndName(type, name), serviceConfig, enabled);
     }
 
+    private static void validateNoDuplicateConfiguredServices(String configKey,
```
danielkec: This validation still misses the object-form loss case. It rejects duplicate `(type, name)` pairs, but `servicesFromObject(...)` later collects object configs into a map keyed only by provider type, so two valid entries with the same type and different names still overwrite each other. Reproducer: object config with `services.first.type=weighted` and `services.second.type=weighted`; only the last `weighted` config survives, so one configured service is never created.

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
```diff
@@ -267,8 +287,22 @@ interface ListenerConfigBlueprint {
      * Defaults to {@code -1}, meaning "unlimited" - what the system allows.
      * {@code 0} has the same meaning as {@code -1} - unlimited number of connections.
      *
-     * @return pre-accept TCP connection admission capacity for this listener, regardless of protocol
+     * @return pre-accept connection admission capacity for this listener, regardless of transport binding or protocol
+     */
+    @Option.Required
```
danielkec: Marking this defaulted option as required makes the generated config metadata say every listener must configure `max-connections`, even though the runtime default and the javadoc below allow it to be omitted as `-1`/unlimited. Config/schema consumers can reject valid listener configs that rely on the documented default.

webserver/webserver/src/main/java/io/helidon/webserver/WebServerConfigSupport.java
```diff
@@ -168,6 +188,156 @@ public void decorate(ListenerConfig.BuilderBase<?, ?> target) {
                                                             .build());
             }
         }
+
+        @SuppressWarnings("removal")
+        private static void normalizeMaxConnections(ListenerConfig.BuilderBase<?, ?> target) {
+            if (target.maxConnections().isEmpty()) {
```
danielkec: This compatibility shim does not work when the builder is derived from an existing `ListenerConfig`. `ListenerConfig.builder(existing)` copies `maxConnections` before a later `.maxTcpConnections(23)` call, so `target.maxConnections()` is already present here and the deprecated value is ignored. Reproducer: `ListenerConfig.builder(ListenerConfig.builder().buildPrototype()).maxTcpConnections(23).buildPrototype().maxConnections()` still returns the inherited `-1`.

danielkec (commented):
Review comments for current head; no approval.

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
```diff
@@ -0,0 +1,707 @@
...
+    SocketTransportBinding(TransportBindingContext transportContext,
+                           String type,
+                           String name,
+                           SocketAddress configuredAddress) {
+        this.transportContext = Objects.requireNonNull(transportContext, "transportContext");
+        this.type = Objects.requireNonNull(type, "type");
+        this.name = Objects.requireNonNull(name, "name");
+        ListenerContext listenerContext = Objects.requireNonNull(transportContext.listenerContext(), "listenerContext");
+        this.listenerConfig = listenerContext.config();
+        this.socketName = listenerConfig.name();
```
danielkec, line 98: This drops the listener map-key identity for builder-defined named sockets. `putSocket("admin", socket -> socket.port(0))` builds a `ListenerConfig` without a name; `LoomServer` still has the map key and constructs `ServerListener("admin", ...)`, but this binding reads `listenerConfig.name()` instead, which has been normalized to `@default`. Custom `ServerConnectionSelectorProvider.create(socketName, ...)` implementations can now see `@default` for the `admin` listener, while main passed the listener map key directly.

config/config/src/main/java/io/helidon/config/ProvidedUtil.java
tomas-langer: yes, that is intentional - if you add a manual entry with the same type and name, it will override the discovered one. This is how it always worked, not a new feature - not part of this PR

webserver/webserver/src/main/java/io/helidon/webserver/ListenerConfigBlueprint.java
tomas-langer: `@Required` is only a marker for Helidon builder generator - and if that is used, it will generate everything correctly, as in this PR. If a third party impl of the `ListenerConfig` exists, it will return unlimited connections.

webserver/webserver/src/main/java/io/helidon/webserver/WebServerConfigSupport.java
tomas-langer: yes, and that is correct, as the `target.maxConnections() has already passed through the config support, which sets it to the correct number (i.e. 23 in your example)

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
tomas-langer: This seems like an existing bug, but one worth fixing, I will look into it.

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
Verdent, line 588: This path appears to leak a connection-limit permit. The listener acquires a token before blocking in `accept()`, but if suspend closes the server socket while `accept()` is blocked, the `AsynchronousCloseException`/`SocketException` paths below exit without completing that token.

I reproduced this locally with `maxConnections(1)`: start the server, wait until the listener thread is blocked in `accept()`, call `suspend()` then `resume()`, and then send a GET request. The client connects but times out waiting for a response, consistent with the resumed listener having no available connection permit.

webserver/webserver/src/main/java/io/helidon/webserver/SocketTransportBinding.java
```diff
@@ -0,0 +1,723 @@
...
+
+    private void listen() {
+        ServerSocketChannel localServerSocket = serverSocket;
+        String serverChannelId = "0x" + HexFormat.of().toHexDigits(System.identityHashCode(localServerSocket));
+
+        while (running) {
+            LimitAlgorithm.Token acceptToken = null;
+            try {
+                // this must be done before we accept, and the semaphore must be released when connection is finished
+                var outcome = connectionLimit.tryAcquireOutcome(true);
```
danielkec, line 588: This still leaves the equal-count multi-binding case with less usable capacity than configured. The warning only covers `bindingCount > maxConnections`, but with TCP + UDS and `maxConnections(2)`, an idle UDS listener can hold one pre-accept permit while blocked in `accept()`. If one TCP connection is already active, a second TCP connection can then wait for a permit even though the configured limit is 2. Reproducer: enable TCP + UDS, set `maxConnections(2)`, leave UDS idle, keep one TCP connection open, then open another TCP connection.

tomas-langer: yes, this is expected and as designed.

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -261,71 +277,77 @@ void start() {
...
+        Objects.requireNonNull(tls, "tls");
         if (!this.tls.enabled()) {
             throw new IllegalArgumentException("TLS is not enabled on the socket " + socketName
                                                        + " and therefore cannot be reloaded");
         }
         if (!tls.enabled()) {
             throw new UnsupportedOperationException("TLS cannot be disabled by reloading on the socket " + socketName);
         }
-        this.tls.reload(tls);
+        reloadTls(tlsMaterial(tls));
```
danielkec: This conversion drops live manager-backed key/trust state from the supplied `Tls`. For example, with a source backed by a custom `TlsManager` that exposes an `X509TrustManager` but has no static prototype material, `reloadTls(source)` succeeded through the previous `this.tls.reload(tls)` path; it now throws `IllegalArgumentException: TLS material must define private key or trust material`. The same path can reinstall stale prototype material after the source has already been reloaded.

webserver/webserver/src/main/java/io/helidon/webserver/ListenerTlsContext.java
```diff
@@ -0,0 +1,138 @@
...
+        public Tls tls() {
+            return tls;
+        }
+
+        /**
+         * Selected SNI context.
+         *
+         * @return selected SNI context
+         */
+        public SniContext sniContext() {
```
danielkec, line 109: `Selection` exposes the authority-policy result only through `SniContext`, which is marked `@Api.Internal` (and the request-time enforcement helper is internal as well). That leaves a third-party binding implementing this public TLS SPI without a supported API for preserving the listener's SNI authority policy. Reproducer: configure a virtual host for `api.example.com` with the default `authority-mismatch=REJECT`, perform the handshake through a provider binding with SNI `api.example.com`, then send `Host: admin.example.com`; built-in TCP rejects it with 421, but the provider cannot apply the same policy without depending on internal APIs.

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -261,71 +277,77 @@ void start() {
...
     boolean hasTls() {
-        return tls.enabled();
-    }
-
-    // Intended for tests that need listener state without stack walking.
-    Thread.State serverThreadState() {
-        Thread localServerThread = serverThread;
-        return localServerThread == null ? null : localServerThread.getState();
+        for (TransportBinding binding : transportBindings) {
+            if (binding.security() == TransportBinding.Security.TLS) {
```
danielkec: This excludes `TLS_EQUIVALENT` even when listener TLS is configured. The new `listenerTlsAllowsTlsEquivalentTransportBindings` test is a minimal reproducer: listener TLS is enabled, the sole active binding reports `TLS_EQUIVALENT`, and `server.hasTls()` is false, while the public API still documents this as whether TLS is configured for the socket. `HelidonServerJunitExtension` uses this value to choose `http://` vs. `https://`, so this can publish the wrong scheme for a secure endpoint.

webserver/webserver/src/main/java/io/helidon/webserver/ServerListener.java
```diff
@@ -335,200 +357,190 @@ private static InitializationContext limitContext(String socketName) {
...
+                }
+                continue;
+            }
 
-        try {
-            // Stop handling any new requests on all accepted and active connections
-            failure = LifecycleFailures.add(failure, closeOpenConnections(false));
-        } catch (RuntimeException | Error e) {
-            failure = LifecycleFailures.add(failure, e);
+            TransportBinding binding = Objects.requireNonNull(factory.create(this),
```
danielkec, line 384: This accepts a created binding whose type differs from the factory's provider/config type. Earlier normalization uses `factory.type()` to enforce one binding per type and decide whether to add default TCP, while later protocol validation uses `binding.type()`. Reproducer: a factory reports `tcp` but returns a binding reporting `custom`; default TCP is suppressed and listener planning proceeds with a non-TCP runtime binding. Distinct factory types can likewise return the same runtime type and bypass the single-type invariant.

danielkec (commented):
Inline findings not covered by the current review threads.

tvallin (dismissed):

tvallin (approved):

danielkec (approved):
Approved. Please create a follow-up issue for the deprecated connection-limit compatibility gap: ListenerConfig.builder().maxConnections(23).buildPrototype() reports 23 from maxConnections() but -1 from maxTcpConnections(), so existing consumers can observe an unlimited value while the effective listener limit is finite.

PR #12094, opened 2026-06-13
Remove WebServer internal benchmark support

tomas-langer:
Removes the WebServer `io.helidon.webserver.internal` benchmark helper package and the SNI JMH benchmark that was its only consumer.

Cleaning up an oversight when merging SNI. This has not been released, so no backward compatibility concern...


danielkec (approved):
Reviewed exact head dde0dc50f4d86795bcf807d7a0b485ce34c7d890. No blocking findings.

PR #12115, opened 2026-06-17
Fix HTTP/2 concurrent stream close accounting

tomas-langer:
Fix an HTTP/2 server race where a client could observe a response DATA END_STREAM and open another stream before the server stopped counting the completed stream against SETTINGS_MAX_CONCURRENT_STREAMS.

The fix separates active stream accounting from deferred stream-context cleanup, ignores late WINDOW_UPDATE frames for retained inactive streams, and moves final DATA END_STREAM deactivation to the concrete connection writer boundary after the raw DATA write and outbound flow-control accounting complete.

This keeps closed stream contexts available for late control-frame handling without counting them against the concurrent-stream limit after the terminal response frame has been written.

### DIFF
--- a/tests/benchmark/jmh/src/main/java/io/helidon/webserver/benchmark/jmh/HttpJmhTest.java
+++ b/tests/benchmark/jmh/src/main/java/io/helidon/webserver/benchmark/jmh/HttpJmhTest.java
@@ -50,6 +50,7 @@ import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Threads;
 import org.openjdk.jmh.infra.Blackhole;
 
 @State(Scope.Benchmark)
@@ -61,6 +62,7 @@ public class HttpJmhTest {
     private static final Header CONTENT_LENGTH = HeaderValues.createCached(HeaderNames.CONTENT_LENGTH, "13");
     private static final int LARGE_RESPONSE_SIZE = 128 * 1024;
     private static final int HTTP2_REUSE_REQUESTS = 16;
+    private static final int HTTP2_REUSE_THREADS = 1;
     private static final Header LARGE_CONTENT_LENGTH = HeaderValues.createCached(HeaderNames.CONTENT_LENGTH,
                                                                                  String.valueOf(LARGE_RESPONSE_SIZE));
     private static final Header SERVER = HeaderValues.createCached(HeaderNames.SERVER, "Helidon");
@@ -171,6 +173,7 @@ public class HttpJmhTest {
     }
 
     @Benchmark
+    @Threads(HTTP2_REUSE_THREADS)
     public void http2MaxConcurrentStreamReuseLargeResponse(Blackhole bh) {
         for (int i = 0; i < HTTP2_REUSE_REQUESTS; i++) {
             try (Http2ClientResponse response = helidonHttp2Client.get("/http2-large").request()) {
--- a/tests/benchmark/jmh/src/main/java/io/helidon/webserver/benchmark/jmh/HttpJmhTest.java
+++ b/tests/benchmark/jmh/src/main/java/io/helidon/webserver/benchmark/jmh/HttpJmhTest.java
@@ -27,17 +27,23 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpResponse;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.util.Arrays;
 
 import io.helidon.http.Header;
 import io.helidon.http.HeaderNames;
 import io.helidon.http.HeaderValues;
 import io.helidon.http.Method;
 import io.helidon.logging.common.LogConfig;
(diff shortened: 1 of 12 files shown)


danielkec (commented):
One inline review issue from the current-head pass.

tests/benchmark/jmh/src/main/java/io/helidon/webserver/benchmark/jmh/HttpJmhTest.java
```diff
@@ -137,6 +172,17 @@ public void http2(Blackhole bh) throws IOException, InterruptedException {
         bh.consume(response);
     }
 
+    @Benchmark
+    @Threads(HTTP2_REUSE_THREADS)
+    public void http2MaxConcurrentStreamReuseLargeResponse(Blackhole bh) {
```
danielkec, line 177: This leaves the regression uncovered by normal PR checks. The race described in the PR depends on a large response DATA `END_STREAM` before the client reuses the same HTTP/2 connection, but this reproducer lives only in the JMH benchmark module. That module skips Surefire by default and only runs under `-Pjmh`; I did not find any PR workflow invoking that profile, so this large-DATA path can regress while the normal checks still pass.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -591,6 +607,19 @@ void prologue(HttpPrologue prologue) {
...
+        if (connectionWriter == null) {
+            writer.writeData(frameData, flowControl.outbound());
+            if (endOfStream) {
+                closeFromLocal();
+            }
+            return frameData.header().length() + Http2FrameHeader.LENGTH;
+        }
+        return connectionWriter.writeData(frameData,
+                                          flowControl.outbound(),
+                                          endOfStream ? this::closeFromLocal : NO_OP);
```
danielkec, line 650: The `RequestException` handler above still bypasses this callback path. Empty error responses write `HEADERS END_STREAM`, and non-empty error responses write terminal DATA via `writer.writeHeaders(..., new Http2FrameData(...), flowControl.outbound())`; both call `closeRejectedStream()` only after that write returns. With `SETTINGS_MAX_CONCURRENT_STREAMS = 1`, a client can observe the terminal error response and open a replacement stream while the rejected stream is still counted active, so the close-accounting race remains for rejected requests.

danielkec (commented):
One additional inline review issue from the current-head pass.

tests/benchmark/jmh/src/main/java/io/helidon/webserver/benchmark/jmh/HttpJmhTest.java
danielkec: such test would be actually problematic as part of the pipeline

tests/benchmark/jmh/src/test/java/io/helidon/webserver/benchmark/jmh/JunitJmhRunnerTest.java
```diff
@@ -76,6 +87,17 @@ static Stream<Histogram.Benchmark> httpBenchmarks() throws RunnerException, IOEx
         return histogram.benchmarks.stream();
     }
 
+    private static ChainedOptionsBuilder optionsBuilder(String resultFile) {
+        return new OptionsBuilder()
```
danielkec, line 91: This JMH run can still pass without ever asserting the new reuse benchmark if that benchmark throws. JMH 1.23 defaults fail-on-error to false; on a `BenchmarkException` it clears the collected results and continues, so the second `new Runner(http2ReuseOptions).run()` can contribute no `RunResult` and the parameterized test simply has no case for `http2MaxConcurrentStreamReuseLargeResponse`. A simple way to see it is to make that benchmark throw during `/http2-large`: the regression benchmark is then omitted rather than failing this JUnit test.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -581,7 +597,7 @@ private void resetInvalidContentLength(int currentFrameLength, boolean endOfStre
     private void closeFromLocal() {
         if (state == Http2StreamState.HALF_CLOSED_REMOTE || state == Http2StreamState.CLOSED) {
             state = Http2StreamState.CLOSED;
-            streams.remove(this.streamId);
+            streams.deactivate(this.streamId);
```
danielkec, line 630: This deactivation now applies to every `closeFromLocal()` caller, not only the DATA path. `writeHeaders(..., endOfStream=true)` and `writeTrailers(...)` still call `closeFromLocal()` before their terminal HEADERS frame is written, so a header-only response or trailers can stop counting against `SETTINGS_MAX_CONCURRENT_STREAMS` while the final frame is still blocked on the writer/socket. With `MAX_CONCURRENT_STREAMS=1`, a second request can be admitted before the first stream is closed on the wire.

danielkec (commented):
Two remaining inline review issues from the current-head pass.

danielkec (approved):
Reviewed exact head 46e456a9101289755e7ecf6a8410ad5a23355e0a. No blocking findings.

PR #12120, opened 2026-06-18
HOCON parser fails to resolve included self-references and unresolved…

### LINKED ISSUE #12119: HOCON parser fails to resolve included self-references and unresolved merges
## Environment Details
* Helidon Version: 4.x

----------

  ### Description

  Helidon HOCON parsing fails in two related include/substitution cases.

  ### Case 1: Included self-reference
```

  include "prod"

  basicAuthEnabledServices: ${basicAuthEnabledServices} [
    "canary-unstable-alpha",
    "canary-unstable-beta",
    "canary-unstable-gamma"
  ]
```

  Expected: ${basicAuthEnabledServices} resolves from prod.conf, then appends the local entries.

  Actual: reading basicAuthEnabledServices fails with:

  IllegalStateException: Recursive update

  ### Case 2: Unresolved merge text leaks into values

  base.conf
```
  metrics.publishers.oci.project = "emailMessageTransformerApi"
  metrics.publishers.oci.fleet = ""
```

  prod.conf
```
  metrics.publishers.oci.project = ${T2_PROJECT}
  metrics.publishers.oci.fleet = ${T2_FLEET}
```

  Expected: prod.conf overrides the base values.

  Actual: values contain unresolved merge text such as: unresolved merge of 2 values follows

  ### Notes

  Both issues appear related to delayed HOCON merge/substitution values being converted into Helidon config nodes before local HOCON resolution has completed.

Spec links:

  - [Duplicate keys and object merging](https://github.com/lightbend/config/blob/main/HOCON.md#duplicate-keys-and-object-merging): later duplicate keys override earlier keys unless both values are objects, in which case objects are recursively merged.
  - [Substitutions](https://github.com/lightbend/config/blob/main/HOCON.md#substitutions): substitution processing happens after parsing, and duplicate keys should resolve to the latest assigned value, meaning the merged object or last non-object value.
  - [Include semantics: merging](https://github.com/lightbend/config/blob/main/HOCON.md#include-semantics-merging): included config keys are merged/overridden the same way as duplicate keys in one file.
  - [Config object merging and file merging](https://github.com/lightbend/config/blob/main/HOCON.md#config-object-merging-and-file-merging): merging objects from different files should behave exactly like duplicate fields in the same file.

jbescos:
Fixes #12119

This updates the HOCON parser to locally resolve delayed HOCON merge/substitution values before converting them into Helidon config nodes, while still preserving unresolved references that should be handled later by Helidon source-level resolution.

The fix covers two cases:

- Self-references that should resolve from an included/base config, such as basicAuthEnabledServices: ${basicAuthEnabledServices} [...].
- Delayed HOCON object merges where override values could previously surface as unresolved merge text.

Added regression tests for both scenarios, including the reported basicAuthEnabledServices shape and metrics override case.

### DIFF
--- a/config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
+++ b/config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
@@ -18,11 +18,15 @@ package io.helidon.config.hocon;
 
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import io.helidon.common.Api;
 import io.helidon.common.Weight;
@@ -41,6 +45,10 @@ import com.typesafe.config.ConfigList;
 import com.typesafe.config.ConfigObject;
 import com.typesafe.config.ConfigParseOptions;
 import com.typesafe.config.ConfigResolveOptions;
+import com.typesafe.config.ConfigResolver;
+import com.typesafe.config.ConfigUtil;
+import com.typesafe.config.ConfigValue;
+import com.typesafe.config.ConfigValueFactory;
 
 /**
  * Typesafe (Lightbend) Config (HOCON) {@link ConfigParser} implementation that supports following media types:
@@ -69,6 +77,8 @@ public class HoconConfigParser implements ConfigParser {
     private static final List<String> SUPPORTED_SUFFIXES = List.of("json", "conf");
     private static final Set<MediaType> SUPPORTED_MEDIA_TYPES =
             Set.of(MediaTypes.APPLICATION_HOCON, MediaTypes.APPLICATION_JSON);
+    private static final Pattern HOCON_REFERENCE = Pattern.compile("(?<!\\\\)\\$\\{(\\?)?\\s*([^}]+?)\\s*}");
+    private static final String LOCAL_RESOLVE_PATH = "helidon-local-resolution";
 
     private final boolean resolvingEnabled;
     private final ConfigResolveOptions resolveOptions;
@@ -134,9 +144,10 @@ public class HoconConfigParser implements ConfigParser {
             typesafeConfig = ConfigFactory.parseReader(readable, parseOptions);
             if (resolvingEnabled) {
                 typesafeConfig = typesafeConfig.resolve(resolveOptions);
+                return fromConfig(typesafeConfig.root());
             }
 
-            return fromConfig(typesafeConfig.root());
+            return fromConfig(typesafeConfig.root(), typesafeConfig, List.of());
         } catch (ConfigException e) {
             throw e;
         } catch (Exception e) {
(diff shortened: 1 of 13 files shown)


tomas-langer (commented):
Review findings from HOCON parser review.

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
```diff
@@ -200,11 +225,108 @@ private static ListNode fromList(ConfigList list) {
...
+        @Override
+        public ConfigValue lookup(String path) {
+            List<String> referencePath;
+            try {
+                referencePath = ConfigUtil.splitPath(path);
+            } catch (com.typesafe.config.ConfigException e) {
+                return ConfigValueFactory.fromAnyRef("${" + path + "}");
+            }
+
+            if (!requiredReferences.contains(referencePath) && optionalReferences.contains(referencePath)) {
```
tomas-langer: This resolver classifies optional vs required substitutions only by path. That breaks cases where the same path appears as both optional and required in one unresolved value, e.g. `v = ${?external}${external}`. With Typesafe `allowUnresolved`, the optional occurrence disappears and the value left for later resolution is `${external}`; with this resolver, both occurrences are treated as required because `external` is in `requiredReferences`, so the deferred Helidon value becomes `${external}${external}` and later resolves to a duplicated value. The optional/required decision needs to be tied to the substitution occurrence, or this fallback path needs to avoid rewriting such mixed optional/required values.

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
```diff
@@ -200,11 +225,108 @@ private static ListNode fromList(ConfigList list) {
...
+                return null;
+            }
+
+            return ConfigValueFactory.fromAnyRef("${" + path + "}");
```
tomas-langer: This preserves the raw HOCON path text in the placeholder handed to Helidon. For a valid HOCON reference such as `${"oracle.com"}`, the deferred value becomes `${"oracle.com"}`. Helidon value resolving treats the text inside `${...}` as a Helidon config key, so it will look for `"oracle.com"` rather than the escaped key `oracle~1com` produced when this parser converts HOCON keys. Deferred references should be converted from HOCON path syntax to Helidon key syntax before being emitted, and the regression test should cover quoted path elements/keys with dots.

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
```diff
@@ -200,11 +225,108 @@ private static ListNode fromList(ConfigList list) {
...
+    private static ConfigValue valueToConvert(ConfigValue value, Config config, List<String> path, boolean locallyResolve) {
+        if (!locallyResolve || !needsLocalResolution(value)) {
+            return value;
+        }
+
+        String hoconPath = ConfigUtil.joinPath(path);
+        try {
+            // Materialize delayed HOCON merges while preserving required references for Helidon source-merge resolution.
+            return config.withOnlyPath(hoconPath)
+                    .resolve(localResolveOptions(value))
```
tomas-langer, line 238: This means `resolvingEnabled(false)` no longer fully disables HOCON substitution resolution: the parser still performs local Typesafe resolution to materialize delayed merges/self-references, while preserving unresolved references for Helidon value resolution. That is a useful distinction for the fix, but the public docs and `HoconConfigParserBuilder.resolvingEnabled` Javadoc still say this option disables substitution resolution. Please update the docs/Javadoc to describe the new split between full HOCON resolution and this local merge/self-reference materialization path.

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
jbescos: Fixed, adding  a test for this: testMixedOptionalAndRequiredReferenceDefersOnlyRequiredOccurrence

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
jbescos: Fixed, adding tests testReferencePathRegexSupportsHoconSyntaxAndHelidonEscaping and testQuotedReferenceUsesHelidonEscapedKey

config/hocon/src/main/java/io/helidon/config/hocon/HoconConfigParser.java
jbescos: Doc updated. There was also a miss match in the builder saying that resolution was by default enabled, but actually `resolvingEnabled` is false.

tomas-langer (approved):

PR #12121, opened 2026-06-18
Fix HTTP/1 close-delimited response reuse

tomas-langer:
Fix intermittent JSON-RPC SSE failures caused by returning close-delimited HTTP/1 response connections to the keep-alive cache.

The HTTP/1 client now distinguishes real close-delimited responses from length-prefixed, chunked, and semantically body-less responses. Closing an unconsumed or fully consumed close-delimited response now closes the socket instead of reusing it, while body-less statuses and unrequested HEAD responses can still keep the connection when no stale bytes are buffered.

The regression coverage exercises abandoned SSE responses, fully consumed close-delimited responses, no-body status handling, 304 metadata, and forbidden buffered bytes after HEAD.

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -322,7 +322,6 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
         int statusCode = responseStatus.code();
         if (responseStatus.family() == Status.Family.INFORMATIONAL
                 || statusCode == Status.NO_CONTENT_204.code()
-                || statusCode == Status.RESET_CONTENT_205.code()
                 || statusCode == Status.NOT_MODIFIED_304.code()) {
             return false;
         }
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -22,6 +22,7 @@ import java.io.UncheckedIOException;
 import java.net.InetSocketAddress;
 import java.net.UnixDomainSocketAddress;
 import java.time.Duration;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Predicate;
@@ -320,14 +321,28 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
                 && statusCode != Status.NOT_MODIFIED_304.code();
     }
 
+    static boolean isChunkedFinalTransferCoding(Headers headers) {
+        if (!headers.contains(HeaderNames.TRANSFER_ENCODING)) {
+            return false;
+        }
+        List<String> values = headers.get(HeaderNames.TRANSFER_ENCODING).allValues();
+        String lastValue = values.get(values.size() - 1);
+        int lastSeparator = lastValue.lastIndexOf(',');
+        String finalCoding = lastValue.substring(lastSeparator + 1).trim();
+        return HeaderValues.TRANSFER_ENCODING_CHUNKED.get().equalsIgnoreCase(finalCoding);
+    }
+
     private static boolean mayExposeEntity(Method requestMethod, Status responseStatus, ClientResponseHeaders responseHeaders) {
         if (requestMethod == Method.HEAD) {
             return false;
         }
-        if (responseHeaders.contains(HeaderValues.CONTENT_LENGTH_ZERO)) {
+        if (!statusAllowsEntity(responseStatus)) {
             return false;
         }
-        if (!statusAllowsEntity(responseStatus)) {
+        if (responseHeaders.contains(HeaderNames.TRANSFER_ENCODING)) {
+            return true;
+        }
+        if (responseHeaders.contains(HeaderValues.CONTENT_LENGTH_ZERO)) {
             return false;
         }
         if ((
@@ -363,7 +378,10 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
             decoder = ContentDecoder.NO_OP;
         }
         InputStream inputStream;
-        if (responseHeaders.contains(HeaderNames.CONTENT_LENGTH)) {
+        if (isChunkedFinalTransferCoding(responseHeaders)) {
+            inputStream = new ChunkedInputStream(helidonSocket, reader, whenComplete, response, recvListener);
+        } else if (!responseHeaders.contains(HeaderNames.TRANSFER_ENCODING)
+                && responseHeaders.contains(HeaderNames.CONTENT_LENGTH)) {
             long length = responseHeaders.contentLength().getAsLong();
             inputStream = new ContentLengthInputStream(helidonSocket,
                                                        reader,
@@ -371,8 +389,6 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
                                                        response,
                                                        length,
                                                        recvListener);
-        } else if (responseHeaders.containsToken(HeaderValues.TRANSFER_ENCODING_CHUNKED)) {
-            inputStream = new ChunkedInputStream(helidonSocket, reader, whenComplete, response, recvListener);
         } else {
             // we assume the rest of the connection is entity (valid for HTTP/1.0, HTTP CONNECT method etc.
             inputStream = new EverythingInputStream(helidonSocket, reader, whenComplete, response, recvListener);
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -102,16 +102,20 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
         Http1ConnectionListener recvListener = http1Client.recvListener();
         WebClientServiceResponse.Builder builder = WebClientServiceResponse.builder();
         AtomicReference<WebClientServiceResponse> response = new AtomicReference<>();
+        boolean successfulConnect = isSuccessfulConnect(serviceRequest.method(), responseStatus);
 
         if (mayExposeEntity(serviceRequest.method(), responseStatus, responseHeaders)) {
             // this may be an entity (if content length is set to zero, we know there is no entity)
-            builder.inputStream(inputStream(clientConfig,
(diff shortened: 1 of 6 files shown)


webclient/http1/src/test/java/io/helidon/webclient/http1/Http1ClientTest.java
```diff
@@ -166,6 +171,106 @@ void testResponseWithoutKeepAliveHeaderReusesConnection() {
...
+
+        try (Http1ClientResponse response = client.get("http://localhost:" + dummyPort + "/test")
+                .connection(connection)
+                .request()) {
+            assertThat(response.status(), is(status));
+            assertThat(response.headers(), noHeader(HeaderNames.CONTENT_LENGTH));
+            assertThat(response.entity().hasEntity(), is(false));
+        }
+
+        assertThat(connection.releaseCount(), is(1));
```
danielkec, line 377: This test case should not include the unframed 205 response as reusable. `205` is semantically no-content, but in HTTP/1.1 it is not one of the status codes whose message is terminated by the empty line after the headers; without `Content-Length: 0`, `Transfer-Encoding`, or closing the connection, any later bytes on the socket are indistinguishable from the next response. As written this codifies reusing a connection that can still poison the next request.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -322,7 +322,6 @@ private static boolean mayHaveEntity(Method requestMethod, Status responseStatus
         int statusCode = responseStatus.code();
         if (responseStatus.family() == Status.Family.INFORMATIONAL
                 || statusCode == Status.NO_CONTENT_204.code()
-                || statusCode == Status.RESET_CONTENT_205.code()
```
tvallin, line 325: Removing 205 from the no-entity branch makes `205 Reset Content` responses visible as normal entities whenever the response does not have `Content-Length: 0` or uses non-empty length/chunked framing. That changes the existing `response.entity().hasEntity()` behavior for a status that must not carry response content, and callers can now parse protocol-invalid 205 bodies as if they were valid. The connection reuse fix can still close or reuse based on framing, but the public response should remain no-entity for 205; please keep coverage for the close/reuse cases and add assertions that 205 never exposes an entity.

tomas-langer: The previous change intentionally separated 205 from 204/304 for connection reuse, because 205 is not
 one of the HTTP/1 responses whose message body length is always determined by the header section alone. It can be explicitly framed, for example with `Content-Length: 0` or chunked framing, and otherwise we need to close rather than reuse.

But you are right that this should not change the public entity contract. 205 must still be exposed as no-entity.
I’ll split the public entity decision from the internal HTTP/1 framing/reuse decision, and add assertions that 205  never exposes an entity while preserving the close/reuse cases.

tomas-langer: Should be fixed now.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -322,7 +322,6 @@ private static boolean mayHaveEntity(Method requestMethod, Status responseStatus
...
                 || statusCode == Status.NOT_MODIFIED_304.code()) {
```
danielkec: This fixes the public `HttpClientResponse.entity()` path, but the no-entity contract still leaks through `WebClientServiceResponse`. Since `205` is no longer filtered here, `createServiceResponse` attaches an input stream, and `Http1ClientRequestImpl.invokeWithServices()` passes `serviceResponse.inputStream()` through before `Http1ClientResponseImpl` applies `entityAllowed`.

A small reproducer is a client service that does `var response = chain.proceed(request); assertThat(response.inputStream().isEmpty(), is(true));` against `HTTP/1.1 205 Reset Content\r\nConnection: keep-alive\r\n\r\n`; on this head the service sees a present stream, and reading it can block on the keep-alive connection.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -226,7 +251,7 @@ private boolean consumeUnreadEntity() {
 
     private ReadableEntity entity(ClientRequestHeaders requestHeaders,
                                   ClientResponseHeaders responseHeaders) {
-        if (inputStream == null) {
+        if (inputStream == null || !entityAllowed) {
```
danielkec, line 341: `entity()` now returns an empty entity for `205` before advancing over chunk framing, but `trailers()` later parses from the reader's current position. For a `205` with `Transfer-Encoding: chunked`, `Trailer: X-Test`, and `0\r\nX-Test: value\r\n\r\n`, calling `response.entity().hasEntity(); response.trailers()` starts trailer parsing at the `0` chunk-size line rather than `X-Test`, so trailer access fails even though the response advertised trailers.

danielkec (commented):
Two remaining issues from the current-head review.

danielkec (commented):
One additional current-head review note.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -198,16 +210,75 @@ ClientConnection connection() {
...
+        return result;
+    }
+
+    private void readNoEntityChunkedFraming() {
+        if (entityFullyRead) {
+            return;
+        }
+        DataReader reader = connection.reader();
+        int length = Http1CallChainBase.readChunkSize(reader);
+        if (length != 0) {
```
tvallin, line 273: For a no-entity `205` with chunked framing and advertised trailers, this return leaves the reader positioned at the chunk payload, but `readTrailers()` then continues by parsing headers from that same position. A response like `205` + `Transfer-Encoding: chunked` + `Trailer: X` with a non-zero chunk whose bytes look like headers can expose body bytes as trailers, mark the entity fully read, and release the connection with the real chunk terminator still buffered. We should treat `length != 0` as invalid/unreusable here instead of continuing into trailer parsing.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -162,10 +170,14 @@ public void close() {
                         || entityLength == ENTITY_LENGTH_CLOSE_DELIMITED) {
                     connection.closeResource();
                 } else {
-                    // No-body response bytes cannot be consumed as entity; buffered data makes reuse unsafe.
-                    if (inputStream == null && connection.reader().available() > 0) {
+                    if (inputStream == null && hasTrailers && !trailers.isLoaded()) {
```
tvallin: This closes any no-stream response that has a `Trailer` header, even when `entityLength == 0` and there is no chunked framing to drain, for example a `HEAD` response carrying metadata headers. That regresses safe connection reuse compared to the header-terminated no-body path. Can this be gated on actual unread chunked framing instead of just `hasTrailers`?

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -198,16 +233,83 @@ ClientConnection connection() {
...
+            }
+            invalidNoEntityChunkedFraming = true;
+            readNoEntityChunkedFraming();
+        }
+        Headers result = Http1HeadersParser.readHeaders(
+                connection.reader(),
+                protocolConfig.maxHeaderSize(),
+                protocolConfig.validateResponseHeaders()
+        );
+        entityFullyRead = true;
```
tvallin, line 262: A 205 with `Content-Length: N` and `Trailer` can reach this path without chunked framing. If the forbidden body begins with a header-shaped prefix, `readHeaders` consumes that prefix, this marks the entity fully read, and `close()` releases the connection even though body bytes remain buffered. Please reject non-chunked no-entity framing before reading trailers and add a stale-byte regression test.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -198,16 +233,83 @@ ClientConnection connection() {
...
+                    || bufferData.get(endOfChunkSize + 2) != Bytes.CR_BYTE
+                    || bufferData.get(endOfChunkSize + 3) != Bytes.LF_BYTE) {
+                return false;
+            }
+            try {
+                if (Http1CallChainBase.parseChunkSizeLine(bufferData.readString(endOfChunkSize,
+                                                                                StandardCharsets.US_ASCII)) == 0) {
+                    reader.skip(endOfChunkSize + 4);
+                    entityFullyRead = true;
+                    return true;
```
tvallin, line 318: `available` can include bytes after the zero-chunk terminator. For `0\r\n\r\nstale`, this returns true after skipping only the terminator, so `close()` caches a connection whose remaining bytes are parsed as the next response. Only mark the framing fully consumed when no buffered bytes remain, and cover it with a follow-up request.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -312,18 +313,36 @@ Http1ConnectionListener recvListener() {
...
-        if (responseHeaders.contains(HeaderValues.CONTENT_LENGTH_ZERO)) {
+        if (!statusAllowsEntity(responseStatus)) {
             return false;
         }
-        int statusCode = responseStatus.code();
-        if (responseStatus.family() == Status.Family.INFORMATIONAL
-                || statusCode == Status.NO_CONTENT_204.code()
-                || statusCode == Status.RESET_CONTENT_205.code()
-                || statusCode == Status.NOT_MODIFIED_304.code()) {
+        if (responseHeaders.contains(HeaderNames.TRANSFER_ENCODING)) {
```
tvallin, line 353: A successful `CONNECT` enters tunnel mode before normal TE/CL precedence, so those fields must be ignored. With `200 CONNECT`, `Content-Length: 0`, and `Transfer-Encoding: chunked`, this path creates a `ChunkedInputStream` and interprets tunnel bytes as HTTP chunks. Handle successful CONNECT separately so tunnel bytes remain raw and the connection is never returned to the HTTP response cache.

tvallin (approved):

PR #12123, opened 2026-06-19
Add declarative GraphQL (27.x)

### LINKED ISSUE #10728: Declarative GraphQL

- [ ] Update API (annotations, interceptors etc.)
- [ ] Implement Codegen
- [ ] Add support for new annotations to Helidon MP extension
- [ ] Add support to test new features (unless works out of the box)

See customer's issue for the same: #11120, as it contains interesting insight into what could be done better than in other frameworks.

We should also have a reasonable support for testing (which would also imply a client maybe). 
This can be folded into a single change, or separated as needed.

tomas-langer:
Resolves #10728

Adds server-side declarative GraphQL support, including shared GraphQL annotations and scalar SPI, SDL/codegen support, WebServer integration, documentation, integration tests, and benchmark coverage.


webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +125,237 @@ public static Builder builder() {
...
+                                ServerResponse res,
+                                String query,
+                                String operationName,
+                                Map<String, Object> variables,
+                                Map<String, Object> contextValues) {
 
         res.headers().contentType(MediaTypes.APPLICATION_JSON);
-        res.send(JSONB.toJson(invocationHandler.execute(query, operationName, variables)));
+        Map<String, Object> result = operationName == null
+                ? invocationHandler.executeWithContext(query, variables, contextValues)
```
danielkec: This now always routes requests through `executeWithContext(...)` while `requestContext(req)` supplies a non-empty context map. A custom `InvocationHandler` that only implements the existing `execute(String, String, Map)` contract will hit the interface default, which rejects non-empty context values, so a normal `/graphql` request fails before the handler runs.

webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityInterceptor.java
```diff
@@ -191,6 +216,79 @@ public void proceed(InterceptionContext ctx, Chain chain, ServerRequest req, Ser
...
+                        + " Endpoint: " + ctx.serviceInfo().serviceType().fqName() + ", method: "
+                        + ctx.elementInfo().elementName());
+                throw new SecurityException(securityFailureMessage());
+            }
+            success = true;
+            return result;
+        } finally {
+            try {
+                if (definition.isAudited()) {
+                    audit(ctx, success ? "OK" : "DENIED", resourceType, definition, securityContext);
```
danielkec: Because `success` remains false until after `chain.proceed(args)` returns, an audited resolver that passes authentication/authorization but then throws an application exception is recorded as `DENIED`. For example, `@Audited` on a resolver that throws `IllegalStateException` after auth succeeds will produce a GraphQL field error, but the audit record reports a security denial.

webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityInterceptor.java
```diff
@@ -496,6 +725,33 @@ private void audit(ServerRequest req,
...
+                       SecurityContext securityContext) {
+        SecurityEnvironment env = securityContext.env();
+        AuditEvent.AuditSeverity auditSeverity = "OK".equals(status)
+                ? methodSecurity.auditOkSeverity()
+                : methodSecurity.auditErrorSeverity();
+
+        SecurityAuditEvent auditEvent = SecurityAuditEvent
+                .audit(auditSeverity, methodSecurity.auditEventType(), methodSecurity.auditMessageFormat())
+                .addParam(AuditEvent.AuditParam.plain("method", env.method()))
+                .addParam(AuditEvent.AuditParam.plain("path", env.path()))
```
danielkec: `env.path()` is an `Optional<String>`, so this stores values such as `Optional[/graphql]` in the resolver audit event. The HTTP audit path above stores the raw path string, and consumers expecting the same audit shape for resolver events will see a different value type/text here.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerTypes.java
```diff
@@ -0,0 +1,348 @@
...
+        result.append('(');
+        for (int i = 0; i < arguments.size(); i++) {
+            Argument argument = arguments.get(i);
+            if (i > 0) {
+                result.append(", ");
+            }
+            result.append(argument.graphQlName())
+                    .append(": ")
+                    .append(argument.schemaType());
+            argument.defaultValue().ifPresent(it -> result.append(" = ").append(it));
```
danielkec: `@GraphQl.DefaultValue` reaches this point as a raw SDL fragment and is appended without validation. A resolver argument such as `@GraphQl.DefaultValue("\"unterminated") String name` still code-generates Java, then the generated feature fails only later when `SchemaParser.parse(schema)` runs.

danielkec (commented):
Review issues from the exact-head pass.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlEntryPointsImpl.java
```diff
@@ -0,0 +1,156 @@
...
+            this.actualDataFetcher = actualDataFetcher;
+        }
+
+        @Override
+        public Object proceed(DataFetchingEnvironment environment) throws Exception {
+            if (interceptorPos < interceptors.size()) {
+                var interceptor = interceptors.get(interceptorPos);
+                interceptorPos++;
+                try {
+                    return interceptor.proceed(ctx, this, environment);
```
danielkec, line 149: An exception thrown by an entry-point interceptor reaches GraphQL Java as a field error, and the server-side error conversion can then expose the exception text in `errors[].message` when the wrapped exception has no cause. For example, a security interceptor throwing `SecurityException("role admin required for tenant X")` can return partial data plus that internal message to the GraphQL client.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+                result.add(new ResolverParameter(parameter,
+                                                Optional.empty(),
+                                                Optional.of(source.orElseThrow().typeInfo()),
+                                                Optional.empty(),
+                                                parameterContext,
+                                                ResolverParameterKind.SOURCE));
+                continue;
+            }
+
+            Optional<ResolverParameterKind> specialParameter = specialParameterKind(parameter.typeName());
```
danielkec, line 441: Special parameters are classified before argument handling, so `@GraphQl.Argument` on one of these types is silently ignored. A resolver such as `String q(@GraphQl.Argument("env") DataFetchingEnvironment env)` generates no `env` argument and passes the environment object instead of reporting the invalid annotation/type combination.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+    private Argument toArgument(SchemaTypes schemaTypes,
+                                TypeInfo endpoint,
+                                TypedElementInfo method,
+                                TypedElementInfo parameter,
+                                Set<Annotation> annotations) {
+        ValueSchemaType valueType = schemaTypes.inputValueType(parameter.typeName(),
+                                                               endpoint,
+                                                               method,
+                                                               parameter.originatingElementValue());
+        String graphQlName = Annotations.findFirst(GRAPHQL_ARGUMENT, annotations)
```
danielkec: This prefers `@GraphQl.Argument` without checking whether the same parameter also has a conflicting `@GraphQl.Name`. For example, `@GraphQl.Argument("id") @GraphQl.Name("bookId") String id` generates the `id` argument and silently ignores `bookId`, while similar naming conflicts are rejected elsewhere in this generator.

danielkec (commented):
Additional uncovered review issues from the exact-head pass.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+
+        private Optional<String> scalarType(TypeName type, Object originatingElement) {
+            Optional<String> builtin = builtinScalarGraphQlType(type);
+            if (builtin.isPresent()) {
+                return builtin;
+            }
+
+            Optional<TypeInfo> typeInfo = roundContext.typeInfo(type.boxed())
+                    .or(() -> roundContext.typeInfo(type));
+            if (typeInfo.isPresent() && typeInfo.orElseThrow().findAnnotation(GRAPHQL_SCALAR).isPresent()) {
```
danielkec, line 1598: This accepts an annotated scalar declaration even when the use site is parameterized. For example, `@GraphQl.Scalar record Box<T>(T value) {}` used as `Box<String>` reaches scalar codegen, and the generated input/scalar adapter code then emits invalid Java such as `Box<String>.class`.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerCustomScalar.java
```diff
@@ -0,0 +1,192 @@
...
+        roundContext.addGeneratedType(generatedType,
+                                      classModel,
+                                      scalarTypeInfo.typeName(),
+                                      scalarType.originatingElement());
+    }
+
+    private static TypeName customScalarAdapterName(TypeName scalarType) {
+        return TypeName.builder()
+                .packageName(scalarType.packageName())
+                .className(scalarType.classNameWithEnclosingNames().replace('.', '_') + "Scalar__GraphQlScalar")
```
danielkec: This generated adapter name can collide for distinct scalar classes in the same package because enclosing-name separators are rewritten to underscores. For example, `pkg.Outer.Inner` and `pkg.Outer_Inner` both map to `Outer_InnerScalar__GraphQlScalar`, so one adapter is skipped and the generated feature later cannot find a `GeneratedGraphQl.CustomScalar` provider for the skipped type.

declarative/codegen/pom.xml
```diff
@@ -55,6 +55,10 @@
             <groupId>io.helidon.service</groupId>
             <artifactId>helidon-service-codegen</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.graphql-java</groupId>
+            <artifactId>graphql-java</artifactId>
```
danielkec: This adds direct `graphql-java` use from `helidon-declarative-codegen`, but the third-party license metadata still lists only `helidon-graphql-server` as the `graphql-java` consumer. The dependency graph and license consumer metadata are now inconsistent.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerTypes.java
```diff
@@ -0,0 +1,348 @@
...
+        description.ifPresent(it -> result.append(" ".repeat(indent))
+                .append('"')
+                .append(escapeDescription(it))
+                .append("\"\n"));
+    }
+
+    private static String escapeDescription(String description) {
+        return description.replace("\\", "\\\\")
+                .replace("\"", "\\\"")
+                .replace("\n", "\\n");
```
danielkec: Descriptions can still contain a raw carriage return. A valid annotation value such as `@GraphQl.Description("line1\rline2")` generates a quoted SDL description with the raw `\r`, and GraphQL Java rejects it during schema parsing. Minimal reproducer: parse `"line1\rline2"\ntype Query { hello: String }` with `graphql.parser.Parser.parse(...)`.

danielkec (commented):
Additional current-head review issues not already covered.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerCustomScalar.java
tomas-langer: This is the way we handle inner classes in Helidon codegen overall. We have to expect that developers use conding guidelines for Java, where `_` is not a character that should be used in class naming. That is why I chose it as a separator.
So yes, this case will break, but I don't think we need to cover that right now.

declarative/codegen/pom.xml
tomas-langer: This is wrong and must be removed.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -249,6 +456,47 @@ public Builder invocationHandler(Supplier<InvocationHandler> handler) {
...
+                                                                 typeQualifiers,
+                                                                 typeAnnotations,
+                                                                 POST_METHOD,
+                                                                 handler);
+            this.getEntryPoint = handler -> entryPoints.handler(descriptor,
+                                                                typeQualifiers,
+                                                                typeAnnotations,
+                                                                GET_METHOD,
+                                                                handler);
+            this.schemaEntryPoint = handler -> entryPoints.handler(descriptor,
```
Verdent, line 643: This applies the endpoint-level annotations to the synthetic schema handler. With class-level `@Authorized(explicit=true)`, security skips authorization before invoking the handler, `graphQlSchema` sends the response, and the post-invocation check only logs once the response has already been sent. Consequently, `GET /<context>/schema.graphql` can return the schema without `SecurityContext.authorize()` being called. A pure introspection query has the same behavior because GraphQL Java's introspection fetchers do not pass through the generated resolver interceptors.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+
+        roundContext.addGeneratedType(generatedType, classModel, endpointTypeName, type.originatingElementValue());
+    }
+
+    private static String featureClassName(GraphQlGroup group) {
+        if (group.endpoints().size() == 1) {
+            TypeName endpointTypeName = group.primaryEndpoint().typeInfo().typeName();
+            return endpointTypeName.classNameWithEnclosingNames().replace('.', '_') + "__GraphQlFeature";
+        }
+        return "GraphQl_" + identifierPart(group.key().listener()) + "_" + identifierPart(group.key().context())
```
Verdent: The generated name is not unique for all route strings because `identifierPart` preserves underscores while encoding `/` as `_2f_`. For example, multi-endpoint groups at `/api/v1` and `/api_2f_v1` both generate `GraphQl__40_default__2f_api_2f_v1__GraphQlFeature`. Both groups then register the same generated type, and the later registration replaces the earlier one, leaving only one group's route, schema, and resolvers in the application.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +125,237 @@ public static Builder builder() {
...
+        case Byte number -> generator.write(number);
+        case Short number -> generator.write(number);
+        case Integer number -> generator.write(number);
+        case Long number -> generator.write(number);
+        case Float number -> generator.write(number);
+        case Double number -> generator.write(number);
+        case Number number -> generator.write(new BigDecimal(number.toString()));
+        case Map<?, ?> map -> writeJsonObject(generator, map);
+        case Iterable<?> iterable -> writeJsonArray(generator, iterable);
+        default -> jsonBinding.serialize(generator, value);
```
Verdent, line 467: Non-JSON-native response values now use Helidon JSON binding instead of the previous reflective Yasson serialization. Helidon JSON requires a generated `@Json.Entity` converter or a registered serializer, but the GraphQL documentation does not state this new requirement. An existing unannotated bean in `GraphQLError.extensions`, which Yasson previously serialized, now fails response serialization with an HTTP 500. The new extension-object test succeeds specifically because `ExtensionPayload` is annotated with `@Json.Entity`.

webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityInterceptor.java
```diff
@@ -191,6 +216,81 @@ public void proceed(InterceptionContext ctx, Chain chain, ServerRequest req, Ser
...
+                        + "method was invoked and may have changed data."
+                        + " Endpoint: " + ctx.serviceInfo().serviceType().fqName() + ", method: "
+                        + ctx.elementInfo().elementName());
+                throw new SecurityException(securityFailureMessage());
+            }
+            success = true;
+            return result;
+        } finally {
+            try {
+                if (definition.isAudited()) {
```
Verdent, line 308: Class-level `@Audited` is lost before this check. `createMethodSecurity` starts with `typeSecurity.copy()`, but `HttpSecurityDefinition.copy()` does not copy the audited flag, event type, message format, or severities. A GraphQL endpoint with only type-level `@Audited` therefore emits no audit event for its request or resolver execution; the current coverage exercises method-level auditing only.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+                .map(GraphQlServerExtension::normalizePath)
+                .orElse(DEFAULT_SCHEMA_URI);
+
+        return new EndpointDeclaration(typeInfo,
+                                        typeAnnotations,
+                                        new GroupKey(listener, context),
+                                        schemaUri);
+    }
+
+    private static String normalizePath(String path) {
```
Verdent, line 208: Root and trailing-slash contexts produce a double-slash schema route because the context is preserved and later concatenated directly with `schemaUri`. `@GraphQlServer.Context("/")` registers `//schema.graphql`, while `@GraphQlServer.Context("/api/")` registers `/api//schema.graphql`. The ordinary `/schema.graphql` and `/api/schema.graphql` URLs do not match those exact routes.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+        }
+        group.schemaTypes().appendTo(result);
+        return result.toString();
+    }
+
+    private void appendType(StringBuilder result, String typeName, List<Operation> operations, SchemaTypes schemaTypes) {
+        result.append("type ")
+                .append(typeName)
+                .append(" {\n");
+        for (Operation operation : operations) {
```
Verdent, line 1313: `@GraphQl.Description` supports methods and parameters, but top-level query/mutation generation does not emit operation descriptions, and `appendArguments` does not emit argument descriptions. For `@Description("Find") @Query String find(@Description("term") String term)`, the generated SDL remains `find(term: String): String` with both descriptions omitted.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +139,293 @@ public static Builder builder() {
...
+    private static List<Object> toJavaList(JsonArray array) {
+        List<Object> result = new ArrayList<>(array.size());
+        array.values().forEach(value -> result.add(toJavaValue(value)));
+        return result;
+    }
+
+    private static Object toJavaNumber(JsonNumber number) {
+        BigDecimal value = number.bigDecimalValue();
+        if (value.scale() <= 0) {
+            BigInteger integer = value.toBigIntegerExact();
```
danielkec, line 412: `toBigIntegerExact()` materializes the full integer before GraphQL validates whether the variable is declared or used. For example, an unused exponent-form variable value such as `1e1000000` makes this 9-character number expand into a 3,321,929-bit `BigInteger`; locally this conversion took about 6.8 seconds while parsing remained cheap. A client can therefore consume disproportionate CPU and memory with a very small request.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +139,293 @@ public static Builder builder() {
...
+                // Other definitions do not contain executable selections.
+            }
+            }
+        });
+        boolean introspection = false;
+        while (!selectionSets.isEmpty() && !introspection) {
+            for (Selection<?> selection : selectionSets.removeFirst().getSelections()) {
+                switch (selection) {
+                case Field field -> {
+                    introspection = field.getName().startsWith("__");
```
danielkec: Because this scan descends into every selection set, a nested `__typename` marks an ordinary data query as an introspection request. With `@Authorized(explicit = true)` and an object-dependent policy, `query { book { id __typename } }` is implicitly authorized before the resolver can load the `Book` and call `SecurityContext.authorize(book)`, so it is denied while `query { book { id } }` reaches the resolver. Clients commonly add `__typename` automatically.

webserver/security/src/main/java/io/helidon/webserver/security/HttpSecurityDefinition.java
```diff
@@ -221,15 +226,7 @@ boolean atzExplicit() {
...
-            return annotValue.orElse(defaultValue);
-        }
-
-        if (currentValue.equals(defaultValue)) {
-            return annotValue.orElse(defaultValue);
-        }
-
-        return currentValue;
+    private <T> T annotationValue(T currentValue, Optional<T> configuredValue, T defaultValue) {
+        return configuredValue.orElse(currentValue == null ? defaultValue : currentValue);
```
danielkec: Production APT metadata includes annotation defaults, so a method-level `@Audited("method")` supplies the default `messageFormat`, `okSeverity`, and `errorSeverity` here. Those values overwrite custom settings copied from a class-level `@Audited`, even though the method declared only the event type. `HttpSecurityDefinitionTest` uses a sparse synthetic annotation and therefore does not reproduce the production representation.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +139,293 @@ public static Builder builder() {
     @Override
     public void routing(HttpRules rules) {
         // schema
-        rules.get(context + schemaUri, this::graphQlSchema);
+        String schemaPath = context.endsWith("/")
+                ? context.substring(0, context.length() - 1) + schemaUri
```
danielkec, line 149: Declarative codegen keeps `/api` and `/api/` as different endpoint groups, but this maps both services to `/api/schema.graphql`. When an application declares one endpoint at each context, the first schema handler registered for that URL wins and the other group's schema is unreachable; the route test exercises the contexts only in isolation.

etc/HELIDON_THIRD_PARTY_LICENSES.xml
```diff
@@ -357,7 +357,6 @@ ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
             <licenseName>Multiple Licenses</licenseName>
             <consumers>
                 <consumer>helidon-http-media-jsonb</consumer>
-                <consumer>helidon-webserver-graphql</consumer>
```
danielkec, line 360: This removes `helidon-webserver-graphql` only from the XML attribution, while the packaged `THIRD_PARTY_LICENSES.txt` still says `Used by: [helidon-http-media-jsonb, helidon-webserver-graphql]`. `licensing/pom.xml` copies both files under `META-INF`, so the released license metadata will disagree.

danielkec (commented):
Additional current-head review issues not already covered.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +139,293 @@ public static Builder builder() {
...
+    private void dispatchRequest(ServerRequest req, ServerResponse res, GraphQlRequest graphQlRequest) throws Exception {
+        req.context().register(graphQlRequest);
+        boolean getRequest = graphQlRequest.getRequest();
+        boolean mutation = graphQlRequest.parsedQuery().map(ParsedQuery::mutation).orElse(false);
+        if (getRequest && mutation) {
+            graphQlRequest(req, res);
+        } else if (graphQlRequest.parsedQuery().map(ParsedQuery::introspection).orElse(false)) {
+            introspectionAuthorization.handle(req, res);
+        } else {
+            graphQlRequest(req, res);
```
danielkec, line 217: An explicit-authorized GraphQL request can complete without invoking a resolver—for example, an invalid-field query or an operation whose only field has `@skip(if: true)`. This path then never calls `SecurityContext.authorize(...)`, so the outer explicit-authorization check replaces GraphQL's response with HTTP 500. On the exact head, the same query returns 200 with `skip=false` and 500 with `skip=true`; an ordinary validation error also returns 500 for an authorized caller.

danielkec (commented):
One additional current-head issue found during exact-head validation.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -249,18 +553,133 @@ public Builder invocationHandler(Supplier<InvocationHandler> handler) {
...
+                    .map(annotation -> Annotation.builder()
+                            .from(annotation)
+                            .property("explicit", false)
+                            .build());
+            TypedElementInfo schemaMethod = implicitAuthorization
+                    .map(annotation -> requestMethod("<graphql-schema>", annotation))
+                    .orElse(SCHEMA_METHOD);
+            List<Annotation> implicitTypeAnnotations = implicitAuthorization
+                    .map(_ -> typeAnnotations.stream()
+                            .filter(annotation -> !annotation.typeName().equals(AUTHORIZED_TYPE))
```
Verdent: **Schema-first access disables class-level explicit authorization**

`HttpSecurityInterceptor` caches resolved type security solely by the descriptor’s service `TypeName`. If `/schema.graphql` is the first request, these stripped type annotations populate that cache and are reused by later resolver entry points. Consequently, the original `@Authorized(explicit=true)` setting—and any configured provider—is no longer applied.

Running `DeclarativeGraphQlTest#testMetadataAuthorizationDoesNotSatisfyExplicitResolvers` by itself reproduces this: the resolver which omits explicit authorization is expected to return 500 but returns 200. The full test class masks the failure because an earlier resolver request warms the cache with the unstripped definition.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,1991 @@
...
+    }
+
+    private void addInvocationHandlerMethod(ClassModel.Builder classModel) {
+        classModel.addMethod(method -> method
+                .accessModifier(AccessModifier.PRIVATE)
+                .returnType(INVOCATION_HANDLER)
+                .name("invocationHandler")
+                .addContent("return ")
+                .addContent(INVOCATION_HANDLER)
+                .addContentLine(".create(schema());"));
```
Verdent: **Declarative endpoints bypass the invocation-handler exception configuration**

This always creates the handler with its defaults, so the documented `default-error-message`, `exception-white-list`, and `exception-black-list` configuration has no effect on generated declarative endpoints.

With the defaults, `InvocationHandlerImpl` returns checked-exception messages verbatim unless the exception is blacklisted. A resolver throwing an `IOException` or `SQLException` containing internal details therefore exposes that text in `errors[].message`, even when the application has configured those exceptions to be masked.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,1991 @@
...
+                    && parameterProvider(parameterContext(endpoint,
+                                                          method,
+                                                          parameter,
+                                                          i,
+                                                          annotations,
+                                                          OperationKind.FIELD)).isPresent()) {
+                continue;
+            }
+
+            Optional<SourceParameter> source = sourceParameter(roundContext, endpoint, method, parameter, i);
```
Verdent, line 581: **Source inference rejects valid unannotated enum and list arguments**

Every parameter without `@GraphQl.Argument` reaches source inference here. An enum argument then fails as an unsupported source because its kind is `ENUM`, even when another parameter is explicitly annotated with `@GraphQlServer.Source`. An unannotated `List<String>` is instead treated as an interface source, so `label(Book book, List<String> tags)` produces two inferred sources.

This conflicts with the documented contract: enum and list arguments are supported, and `@GraphQl.Argument` is only needed when the Java parameter name should not be used. These valid child resolvers therefore fail code generation unless unnecessary argument annotations are added.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerCustomScalar.java
```diff
@@ -0,0 +1,192 @@
...
+
+        roundContext.addGeneratedType(generatedType,
+                                      classModel,
+                                      scalarTypeInfo.typeName(),
+                                      scalarType.originatingElement());
+    }
+
+    private static TypeName customScalarAdapterName(TypeName scalarType) {
+        return TypeName.builder()
+                .packageName(scalarType.packageName())
```
Verdent: **Cross-module scalar use generates code in a dependency module’s package**

Scalar metadata can be resolved from a required module, but this places the generated adapter in that scalar’s package while annotation processing the endpoint module. With the scalar and provider in named module A and the endpoint in named module B, B attempts to generate `MoneyScalar__GraphQlScalar` in A’s package. This creates a JPMS split package and prevents compilation.

It can also duplicate the same adapter if module A already generated it. The current tests place the scalar, provider, and endpoint in one module, so they do not exercise this dependency layout.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,1991 @@
...
+                .increaseContentPadding()
+                .addContent(".webContext(")
+                .addContentLiteral(group.key().context())
+                .addContentLine(")")
+                .addContent(".schemaUri(")
+                .addContentLiteral(group.schemaUri())
+                .addContentLine(")")
+                .addContentLine(".declarativeHttpEntryPoints(httpEntryPoints, descriptor, descriptor.qualifiers(), annotations)")
+                .addContentLine(".invocationHandler(invocationHandler())")
+                .addContentLine(".build());")
```
Verdent, line 843: **Each generated route group eagerly creates an unused thread pool**

This builds `GraphQlService` without an executor override. Its default path creates a `ServerThreadPoolSupplier`, and the service constructor immediately obtains the executor. The defaults use two core threads per available processor and prestart all core threads.

`GraphQlService` never uses that executor after construction and does not shut it down. Each generated listener/context group therefore retains an idle platform-thread pool for the application lifetime—for example, 128 threads per group on a 64-vCPU host.

danielkec (commented):
Current-head follow-up review with only findings not covered by the existing review threads.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerExtension.java
```diff
@@ -0,0 +1,2000 @@
...
+                .addContentLine(".builder()")
+                .increaseContentPadding()
+                .increaseContentPadding()
+                .addContent(".webContext(")
+                .addContentLiteral(group.key().context())
+                .addContentLine(")")
+                .addContent(".schemaUri(")
+                .addContentLiteral(group.schemaUri())
+                .addContentLine(")")
+                .addContentLine(".declarativeHttpEntryPoints(httpEntryPoints, descriptor, descriptor.qualifiers(), annotations)")
```
danielkec, line 841: Current `main` now authenticates `GraphQlService` routes by default and reads the opt-out from `graphql.permit-all`. This generated `GraphQlService` builder never receives that config, so after the pending rebase an endpoint configured with `graphql.permit-all=true` still keeps the default authentication wrapper, unlike a programmatic service created from the same config. The current checks predate this combination.

graphql/server/src/main/java/io/helidon/graphql/server/InvocationHandler.java
```diff
@@ -139,6 +194,7 @@ public InvocationHandler build() {
 
             GraphQL graphQl = GraphQL.newGraphQL(schema)
                     .subscriptionExecutionStrategy(new SubscriptionExecutionStrategy())
+                    .preparsedDocumentProvider(InvocationHandlerImpl.preparsedDocumentProvider(schema))
```
danielkec: Current `main` adds the default depth (100) and complexity (1000) instrumentations to this builder, and the branch currently conflicts at this chain. If the PR-side chain is retained during conflict resolution, only the preparsed-document provider remains; for example, a query nested beyond 100 levels is no longer rejected by the default depth guard. The green checks predate this integration.

webserver/graphql/src/main/java/io/helidon/webserver/graphql/GraphQlService.java
```diff
@@ -90,68 +142,311 @@ public static Builder builder() {
...
+                .forEach(fragment -> fragments.putIfAbsent(fragment.getName(), fragment));
+        Set<String> visitedFragments = new HashSet<>();
+        ArrayDeque<SelectionSet> selectionSets = new ArrayDeque<>();
+        operationDefinition.map(OperationDefinition::getSelectionSet)
+                .ifPresent(selectionSets::addLast);
+        boolean introspection = false;
+        while (!selectionSets.isEmpty() && !introspection) {
+            for (Selection<?> selection : selectionSets.removeFirst().getSelections()) {
+                switch (selection) {
+                case Field field -> introspection = field.getName().startsWith("__");
```
danielkec: The nested-`__typename` case is fixed, but this still treats the whole operation as introspection when any top-level field starts with `__`. With resource-dependent `@Authorized(explicit=true)`, `{ __typename book { id } }` enters synthetic authorization before `book` can load the resource and call `SecurityContext.authorize(book)`, so it can be denied although `{ book { id } }` succeeds. Current tests cover nested and root-only metadata, not this mixed operation.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/graphql/server/GraphQlServerCustomScalar.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    void process(RegistryRoundContext roundContext, TypeInfo endpointTypeInfo, ScalarSchemaType scalarType) {
+        TypeInfo scalarTypeInfo = roundContext.typeInfo(scalarType.javaType().boxed())
+                .or(() -> roundContext.typeInfo(scalarType.javaType()))
+                .orElseThrow(() -> new CodegenException("Unknown Java GraphQL scalar type "
+                                                                + scalarType.javaType().fqName(),
+                                                        scalarType.originatingElement()));
+        if (generatedScalarTypes.add(scalarType.javaType())) {
+            TypeName generatedType = customScalarAdapterName(endpointTypeInfo.typeName(), scalarType.javaType());
```
danielkec: This makes the adapter consumer-owned, while `generatedScalarTypes` is local to one annotation-processing run. Minimal reproducer: place a shared `@GraphQl.Scalar("MONEY") Money` and its provider in one module, compile endpoint modules A and B that both use `Money`, then compose both modules in one application. Both generated adapters register the same GraphQL name and Java type, so startup fails with `Multiple GraphQL scalar providers found for scalar 'MONEY'...`. The cross-module test has only one consumer and does not exercise this composition.

declarative/tests/codegen/pom.xml
```diff
@@ -46,6 +46,31 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.security</groupId>
+            <artifactId>helidon-security</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.helidon.security</groupId>
+            <artifactId>helidon-security-annotations</artifactId>
```
danielkec: `helidon-security-annotations` is already declared at line 141 of this POM. A focused Maven run emits the duplicate dependency declaration warning for this module, and future Maven versions may stop accepting the malformed model.

graphql/graphql/src/main/java/io/helidon/graphql/GraphQl.java
```diff
@@ -0,0 +1,204 @@
...
+         * @return default value literal
+         */
+        String value();
+    }
+
+    /**
+     * Marks a generated GraphQL type, field, or argument as non-null.
+     */
+    @Target({
+            ElementType.TYPE_USE,
```
danielkec, line 164: This target also permits outer type-use placement, but outer nullability is derived only from declaration annotations. For example, `@GraphQl.Query java.lang.@GraphQl.NonNull String result()` compiles but generates `result: String` rather than `result: String!`; the same loss occurs for an explicitly type-annotated resolver parameter or input record component. This silently weakens the generated schema contract.

tests/benchmark/jmh/src/main/java/io/helidon/webserver/graphql/GraphQlJmhBenchmark.java
```diff
@@ -0,0 +1,210 @@
...
+
+        @Setup
+        public void setup() {
+            LogConfig.configureRuntime();
+
+            GraphQLSchema schema = schema();
+            invocationHandler = InvocationHandler.create(schema);
+            server = WebServer.builder()
+                    .host(SERVER_HOST)
+                    .routing(routing -> routing.register(GraphQlService.create(schema)))
```
danielkec: `GraphQlService.create(schema)` protects the endpoint with authentication by default, while the benchmark request below carries no credentials. The same setup is asserted to return `401` in `GraphQlServiceSecurityTest`; since this benchmark only consumes the status and body, `graphQlServiceGet` measures the authentication rejection path rather than GraphQL execution.

danielkec (commented):
Two additional current-head review issues not covered by the existing threads.

danielkec (approved):
Reviewed exact head 15aacf8f772a50f627397db6710625af2f7bbd3a; follow-up fixes verified and all review threads are resolved.

PR #12124, opened 2026-06-19
Add declarative gRPC support

### LINKED ISSUE #6748: gRPC pico based Declarative style to replace `microprofile/grpc`

tomas-langer:
Resolves #6748

Adds declarative gRPC support for generated server registrations and clients, including streaming method shapes, security, validation, interceptors, config-expression client endpoints, docs, and runtime coverage.


webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcStreams.java
```diff
@@ -0,0 +1,465 @@
...
+    public static <ReqT, ResT> StreamObserver<ReqT> bidirectionalStream(Function<Stream<ReqT>, Stream<ResT>> handler,
+                                                                        StreamObserver<ResT> responseObserver) {
+        Objects.requireNonNull(handler);
+        Objects.requireNonNull(responseObserver);
+        StreamingRequest<ReqT> request = new StreamingRequest<>(responseObserver);
+        Outbound outbound = new Outbound(responseObserver, request::cancel);
+        ContextRunner contextRunner = ContextRunner.create();
+        Thread.startVirtualThread(contextRunner.wrap(() -> {
+            try (Stream<ReqT> stream = request.stream()) {
+                sendResponses(handler.apply(stream)::iterator, responseObserver, outbound);
```
Verdent: The bidirectional response stream returned by `handler.apply(stream)` is dereferenced directly to an iterator and is never closed. For endpoints that return a resource-backed `Stream` or use `Stream.onClose()`, normal completion, cancellation, or send failure skips that cleanup path.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcStreams.java
```diff
@@ -0,0 +1,465 @@
...
+                responseObserver.onError(throwable);
+            }
+
+            @Override
+            public void onCompleted() {
+                try {
+                    responseObserver.onNext(handler.apply(List.copyOf(requests)));
+                    responseObserver.onCompleted();
+                } catch (Throwable t) {
+                    responseObserver.onError(t);
```
Verdent: This catch block sends endpoint failures through `responseObserver.onError(...)` instead of allowing them to escape through the gRPC listener path. That appears to bypass the new validation interceptor mapping for `ValidationException`, so validation failures from declarative streaming endpoints can be reported as a generic gRPC error rather than `INVALID_ARGUMENT`. The same pattern is also present in the stream-based client-streaming and bidirectional helpers.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/client/GrpcClientExtension.java
```diff
@@ -0,0 +1,732 @@
...
+        classModel.addMethod(proto -> {
+            proto.accessModifier(AccessModifier.PRIVATE)
+                    .returnType(PROTO_FILE_DESCRIPTOR)
+                    .name("proto");
+            if (protoDescriptor.descriptorType().isPresent()) {
+                proto.addContent("return ")
+                        .addContent(protoDescriptor.descriptorType().orElseThrow())
+                        .addContentLine(".getDescriptor();");
+            } else {
+                proto.addContent("return ")
```
Verdent: The client generator always emits a static `EndpointType.proto()` call for method-based `@Grpc.Proto` descriptors. The shared descriptor discovery accepts non-static `@Grpc.Proto` methods, so an interface default proto method can pass descriptor validation but produce generated client source that does not compile.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -388,17 +411,54 @@ private FlowControl.Outbound outboundFlowControl() {
...
+                halfClosePending = false;
+            }
+        } finally {
+            inboundLock.unlock();
+        }
         if (listener != null) {
-            while (!listenerQueue.isEmpty() && numMessages.getAndDecrement() > 0) {
-                listener.onMessage(listenerQueue.poll());
+            for (REQ request : requests) {
+                listener.onMessage(request);
```
tvallin: `flushQueue()` dequeues under `inboundLock` but invokes the listener after unlocking. Concurrent or reentrant `request()`/DATA paths can dispatch later messages, or `onHalfClose()`, before this batch. Please use a single-drainer mechanism that preserves callback ordering and add a concurrent demand/EOS regression test.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerRegistrationGenerator.java
```diff
@@ -0,0 +1,511 @@
...
+                .addContentLine("var method = service.findMethodByName(methodName);")
+                .addContentLine("if (method == null) {")
+                .increaseContentPadding()
+                .addContentLine("throw new IllegalArgumentException(\"Unable to find gRPC method \" + serviceName")
+                .increaseContentPadding()
+                .addContentLine("+ \"/\" + methodName + \" in proto descriptor.\");")
+                .decreaseContentPadding()
+                .decreaseContentPadding()
+                .addContentLine("}")
+                .addContentLine("if (method.isClientStreaming() != clientStreaming")
```
tvallin: Please validate the proto input/output message types against the declared request and response types, not only streaming cardinality. Unrelated protobuf classes currently pass registration, but the descriptor marshals the proto-declared classes while the generated handler expects the declared classes, causing runtime type failures. The client-side validation needs the same check.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/client/GrpcClientExtension.java
```diff
@@ -0,0 +1,732 @@
...
+                            .addContentLiteral(method.methodName())
+                            .addContentLine(", request), 0), false);");
+                    case CLIENT_STREAMING_ITERABLE -> clientMethod.addContent("return serviceClient.clientStream(")
+                            .addContentLiteral(method.methodName())
+                            .addContentLine(", requests.iterator());");
+                    case CLIENT_STREAMING_STREAM -> clientMethod.addContentLine("try (requests) {")
+                            .increaseContentPadding()
+                            .addContent("return serviceClient.clientStream(")
+                            .addContentLiteral(method.methodName())
+                            .addContentLine(", requests.iterator());")
```
tvallin: This routes a request `Stream` through the eager iterator overload, which consumes it and calls `onNext` without checking transport readiness. Large or unbounded streams can be buffered without backpressure. Please add a Stream-aware client-streaming path with readiness, cancellation, and source cleanup.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,384 @@
...
+            }
+            GrpcServerRegistrationGenerator.generate(roundContext, ctx, toEndpoint(serverEndpoint));
+        }
+    }
+
+    private GrpcEndpoint toEndpoint(TypeInfo serverEndpoint) {
+        TypeName endpointType = serverEndpoint.typeName();
+        List<GrpcMethod> methods = new ArrayList<>();
+        for (TypedElementInfo method : serverEndpoint.elementInfo()) {
+            toGrpcMethod(serverEndpoint, method).ifPresent(methods::add);
```
tvallin, line 119: Please reject duplicate `GrpcMethod.grpcName()` values while collecting endpoint methods. Two Java methods can map to the same wire method; both validate, then `GrpcServiceDescriptor.Builder` silently replaces the first registration with the second.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/client/GrpcClientExtension.java
```diff
@@ -0,0 +1,732 @@
...
+                                .addContent(", ")
+                                .addContentLiteral(parameter.name())
+                                .addContentLine(");");
+                    }
+
+                    switch (method.invocation()) {
+                    case SERVER_STREAMING_ITERABLE -> clientMethod.addContent("return () -> serviceClient.serverStream(")
+                            .addContentLiteral(method.methodName())
+                            .addContentLine(", request);");
+                    case SERVER_STREAMING_STREAM -> clientMethod.addContent("return ")
```
tvallin: The returned `Stream` has no cancellation-aware `onClose` action. Closing or short-circuiting it leaves the underlying server-streaming RPC active. Please expose a Stream-based operation that owns and cancels the `ClientCall`.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcServiceClientImpl.java
```diff
@@ -168,6 +177,55 @@ public void onCompleted() {
...
+                    responses.error(status.asRuntimeException(trailers));
+                }
+            }
+
+            @Override
+            public void onReady() {
+                ready.ready();
+            }
+        }, new Metadata());
+        Thread.startVirtualThread(() -> {
```
tvallin: The request stream is lazy, so its iterator and pipeline stages execute on this virtual thread. Neither the caller's gRPC context nor Helidon context is propagated, unlike the server-side stream path. Please capture and restore both contexts around the sender.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcEntryPointsImpl.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        @Override
+        public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(ServerCall<ReqT, RespT> call,
+                                                                     Metadata headers,
+                                                                     ServerCallHandler<ReqT, RespT> next) {
+            try {
+                return createChain(next).proceed(call, headers);
+            } catch (RuntimeException e) {
+                throw e;
```
tvallin: Runtime interceptor failures escape `GrpcProtocolHandler.init()` instead of closing only this RPC with trailers. Please preserve status and metadata for gRPC status exceptions and map other runtime interceptor failures to `INTERNAL`.

declarative/tests/grpc/src/test/java/io/helidon/declarative/tests/grpc/DeclarativeGrpcTest.java
```diff
@@ -0,0 +1,582 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+
+import io.helidon.declarative.tests.grpc.DeclarativeGrpcProto.GreetingRequest;
+import io.helidon.metrics.api.Counter;
+import io.helidon.metrics.api.MeterRegistry;
+import io.helidon.metrics.api.MetricsFactory;
+import io.helidon.metrics.api.Tag;
+import io.helidon.metrics.api.Timer;
+import io.helidon.webclient.grpc.GrpcClient;
```
tvallin: This import is unused; please remove it.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerRegistrationGenerator.java
```diff
@@ -0,0 +1,424 @@
...
+                                    GrpcMethod grpcMethod) {
+        if (security.isEmpty()) {
+            return;
+        }
+        constructor.addContent(GRPC_SECURITY)
+                .addContent(".enforce()");
+        security.authenticate().ifPresent(authenticate -> constructor.addContent(authenticate
+                                                                                         ? ".authenticate()"
+                                                                                         : ".skipAuthentication()"));
+        if (security.authenticationOptional()) {
```
tvallin: **[P1] Preserve method-level authentication defaults.** A false `authenticationOptional()` emits nothing, so a plain method-level `@Authenticated` combined with class-level `@Authenticated(optional = true, provider = "x")` retains both class settings. `Authenticated` explicitly documents that it is non-cumulative, so this unexpectedly permits anonymous access. Please emit explicit overrides, including clearing the provider, and cover this class/method combination.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,367 @@
...
+        boolean hasPermitAll = permitAll.isPresent() || rolePermitAll.isPresent();
+        boolean hasDenyAll = denyAll.isPresent();
+        boolean securityLevel = authenticated.isPresent()
+                || authorized.isPresent()
+                || audited.isPresent()
+                || hasDenyAll
+                || hasPermitAll
+                || hasRoleValidatorRoles
+                || hasAbacAnnotation
+                || !rolesAllowed.isEmpty();
```
tvallin: **[P1] Do not disable security for an empty role list.** `@RolesAllowed({})` leaves `rolesAllowed` empty and excludes the annotation from `securityLevel`; on an otherwise unsecured endpoint, no security handler is registered and the RPC is open. An empty allow-list should reject generation or deny all. Please key this check on annotation presence and add method- and type-level coverage.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcServiceClientImpl.java
```diff
@@ -131,6 +133,42 @@ public <ReqT, ResT> StreamObserver<ReqT> clientStream(String methodName, StreamO
...
+        Objects.requireNonNull(methodName, "methodName");
+        Objects.requireNonNull(requests, "requests");
+        ClientCall<ReqT, ResT> call;
+        try {
+            call = ensureMethod(methodName, MethodDescriptor.MethodType.BIDI_STREAMING);
+        } catch (RuntimeException | Error t) {
+            GrpcClientStreams.closeRequests(requests, t);
+            throw t;
+        }
+        return GrpcClientStreams.bidirectional(call, requests, grpcClient.prototype().executor());
```
tvallin: **[P1] Avoid running request production on the transport executor.** `GrpcClientCall` uses this same executor for its writer. With a single-thread executor, the first `onNext` queues the writer behind the sender, then the sender evaluates `iterator.hasNext()` or `next()`. A lockstep bidi source waiting for the first response blocks the only thread, so the first request is never written. Use a separate context-propagating task and test a response-driven source with a single-thread client executor.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,367 @@
...
+        return new GrpcSecurityDefinition(authenticate,
+                                          authenticated.flatMap(it -> it.booleanValue("optional")).orElse(false),
+                                          authenticated.flatMap(it -> it.stringValue("provider")).filter(it -> !it.isBlank()),
+                                          authorize,
+                                          authorized.flatMap(it -> it.stringValue("provider")).filter(it -> !it.isBlank()),
+                                          List.copyOf(rolesAllowed),
+                                          hasDenyAll || hasPermitAll,
+                                          audited.map(_ -> true),
+                                          audited.flatMap(Annotation::stringValue).filter(it -> !it.isBlank()),
+                                          audited.flatMap(it -> it.stringValue("messageFormat")).filter(it -> !it.isBlank()),
```
tvallin: **[P1] Keep the gRPC audit format for a default `@Audited`.** `stringValue("messageFormat")` includes `Audited.DEFAULT_MESSAGE_FORMAT` even when it was not overridden. That HTTP format references `%6$s`, while `GrpcSecurityHandler` supplies five gRPC parameters, producing a formatting-failure message. Leave the format unset when it equals the annotation default or supply compatible parameters, and assert the emitted audit text.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,367 @@
...
+        }
+
+        return new GrpcSecurityDefinition(authenticate,
+                                          authenticated.flatMap(it -> it.booleanValue("optional")).orElse(false),
+                                          authenticated.flatMap(it -> it.stringValue("provider")).filter(it -> !it.isBlank()),
+                                          authorize,
+                                          authorized.flatMap(it -> it.stringValue("provider")).filter(it -> !it.isBlank()),
+                                          List.copyOf(rolesAllowed),
+                                          hasDenyAll || hasPermitAll,
+                                          audited.map(_ -> true),
```
tvallin, line 443: **[P2] Propagate custom audit severities.** `@Audited.okSeverity()` and `errorSeverity()` are discarded, and `GrpcSecurityHandler` always uses `SUCCESS` and `FAILURE`. Please carry both values into the generated security configuration and cover successful and failed calls with non-default severities.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    private static String methodName(TypedElementInfo method, List<Annotation> annotations) {
+        Annotation concreteAnnotation = annotations.stream()
+                .filter(it -> it.typeName().equals(GRPC_METHOD) || it.hasMetaAnnotation(GRPC_METHOD))
+                .findFirst()
+                .orElseThrow();
+        String name = concreteAnnotation.typeName().equals(GRPC_METHOD)
+                ? concreteAnnotation.stringValue("name").orElse("")
+                : concreteAnnotation.stringValue().orElse("");
```
tvallin, line 315: **[P2] Honor names declared by meta-annotated `@GrpcMethod`.** For a custom annotation using `@GrpcMethod(value = UNARY, name = "SayHello")`, this reads only the concrete annotation’s unnamed value and falls back to the Java method name. The method type comes from the meta-annotation, but descriptor lookup then uses the wrong name. Read `GrpcMethod.name()` when the concrete annotation supplies no override; the client path has the same logic.

webserver/grpc/src/main/java/module-info.java
```diff
@@ -32,6 +32,7 @@
     requires io.helidon.webserver.http2;
     requires io.helidon.config;
     requires io.helidon.metrics.api;
+    requires io.helidon.service.registry;
```
tvallin: **[P2] Make the public service-registry dependency transitive.** `GrpcEntryPoint` signatures expose service-registry and common-types classes, but this requirement is not transitive. A named module requiring only `io.helidon.webserver.grpc` cannot implement those public contracts without an additional direct requirement. Please use `requires transitive`; service registry already transitively supplies common types, and add modular consumer compilation coverage.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/client/GrpcClientExtension.java
```diff
@@ -0,0 +1,582 @@
...
+                .addContentLine("}")
+                .addContentLine("declarative__clientBuilder.baseUri(uri);")
+                .addContentLine("declarative__client = declarative__clientBuilder.build();")
+                .decreaseContentPadding()
+                .addContentLine("}")
+                .addContentLine("this.serviceClient = declarative__client.serviceClient(declarative__descriptor);");
+    }
+
+    private void registryClientParameter(Parameter.Builder param, GrpcEndpoint endpoint) {
+        if (endpoint.clientName().isPresent()) {
```
Verdent, line 519: When `clientName` is absent, the generated constructor receives an unqualified `Supplier<Optional<GrpcClient>>`. Registry lookups without qualifier criteria also match named instances, so a named client can be selected instead of the endpoint’s declared URI. This contradicts the documented behavior that only an unnamed registry client should precede the URI fallback.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerRegistrationGenerator.java
Verdent: The same non-cumulative behavior also affects authorization. With class-level `@Authorized(provider = "x")` and plain method-level `@Authorized`, the method has no explicit authorizer, so `GrpcSecurityHandler.combine` retains provider `x`. The method therefore inherits the class provider even though `Authorized` documents that method-level values replace the class-level values.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -805,30 +987,50 @@ private BufferData readEntityFromPipeline() {
...
+                        && !remoteResetReceived
+                        && !resetStreamSent
+                        && (forceReset || !remoteAlreadyComplete)) {
+                    sendReset = true;
                     resetStreamSent = true;
-                    locallyResetStreamTracker.add(this.streamId, locallyResetStreamState());
-                    if (remoteCompleteAfterReset || remoteAlreadyComplete) {
-                        locallyResetStreamTracker.remoteComplete(this.streamId);
+                    if (!remoteEndOfStream) {
+                        locallyResetStreamTracker.add(this.streamId, locallyResetStreamState());
```
tvallin, line 1015: **[P1] Complete reset tracking before deactivating the stream.** The rebased base now deactivates the stream before this PR registers it as locally reset. This violates the reset-accounting invariant and deterministically fails `rejectedStreamRestoresRacingDataConnectionFlowControl` on Linux, Windows, and locally. Move `streams.deactivate(...)` after this registration block while keeping it before `writeResetStream(...)`; that preserves both tracker coverage and the base branch’s inactive-before-reset-write guarantee.

tomas-langer: Yes, fixing and reverting the test fix.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerRegistrationGenerator.java
```diff
@@ -0,0 +1,477 @@
...
+                .addContent(".builder().type(")
+                .addContent(endpointType)
+                .addContent(".class).classAnnotations(CLASS_ANNOTATIONS)");
+        if (syntheticClassDenyAll) {
+            addDenyAllAnnotation(constructor, ".addClassAnnotation(");
+        }
+        if (grpcMethod != null) {
+            constructor.addContent(".methodName(")
+                    .addContentLiteral(grpcMethod.method().elementName())
+                    .addContent(").methodAnnotations(")
```
Verdent, line 327: Security discovery includes annotations inherited from the method hierarchy, but the generated `SecurityLevel` here is populated from `METHOD_*.annotations()`, which contains only annotations declared on the concrete override. With `@DenyAll` on an interface method and an unannotated endpoint override, the generated method annotation list is empty. The ABAC provider then finds no attribute to validate and returns permit, so a method denied by its inherited contract can execute. Inherited `RoleValidator.Roles` and custom ABAC annotations are dropped by the same path.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientCall.java
```diff
@@ -97,175 +135,390 @@ public void sendMessage(ReqT message) {
...
-                socket().log(LOGGER, ERROR, e.getMessage(), e);
-                Status errorStatus = Status.UNKNOWN.withDescription(e.getMessage()).withCause(e);
-                responseListener().onClose(errorStatus, EMPTY_METADATA);
-            }
-            socket().log(LOGGER, DEBUG, "[Writing thread] exiting");
+    protected void startStreamingThreads() {
+        clientStream().onReset(rstStream -> {
+            resetFailure = new Http2Exception(rstStream.errorCode(),
+                                              "Reset of " + clientStream().streamId() + " stream received");
+            messageRequest.release();
```
Verdent: `ClientCall.request(Integer.MAX_VALUE)` is valid as effectively unbounded response demand and leaves `messageRequest` at its maximum permit count. If the peer resets before a response consumes a permit, this release overflows the semaphore and throws `Error: Maximum permit count exceeded`. Because the reset callback runs before the underlying HTTP/2 reset processing, the call can miss normal reset status delivery and remain stuck.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerRegistrationGenerator.java
```diff
@@ -0,0 +1,477 @@
...
+                .name("toString")
+                .addContent("return ")
+                .addContentLiteral("gRPC route registration for "
+                                           + endpoint.type().typeName().className()
+                                           + "(" + endpoint.serviceName() + ")")
+                .addContentLine(";"));
+    }
+
+    private static void addHandlerMethods(ClassModel.Builder classModel, GrpcEndpoint endpoint, boolean singleton) {
+        String endpointAccessor = singleton ? "endpoint." : "endpoint.get().";
```
Verdent, line 414: The non-singleton path resolves the endpoint through a service-registry supplier, but gRPC subprotocol invocation does not run through the HTTP routing filter that activates `Service.PerRequest`. A `@Service.PerRequest` endpoint using `@Grpc.ProtoDescriptor` therefore reaches this `endpoint.get()` with no active scope and throws `ScopeNotActiveException`. If the descriptor comes from an instance `@Grpc.Proto` method, the same exception occurs earlier while the registration constructor calls `proto()`. Consequently, the generated request-scoped endpoint cannot be invoked.

webserver/grpc-validation/src/main/java/io/helidon/webserver/grpc/validation/GrpcValidation.java
```diff
@@ -0,0 +1,240 @@
...
+            try {
+                super.onHalfClose();
+            } catch (ValidationException e) {
+                close(e);
+            }
+        }
+
+        @Override
+        public void onCancel() {
+            if (validationFailed.get()) {
```
Verdent: After `onMessage` or `onHalfClose` catches a `ValidationException`, `close(e)` sets `validationFailed`. Both terminal callbacks then return without invoking the delegate. The service listener consequently receives neither `onCancel` nor `onComplete` after a validation rejection, so any per-call cleanup tied to its terminal callback is skipped.

docs/src/main/asciidoc/se/injection/declarative.adoc
```diff
@@ -262,6 +264,164 @@ include::{sourcedir}/se/inject/DeclarativeExample.java[tag=snippet_3, indent=0]
...
+Generated registrations use the fully-qualified gRPC service name, including the proto package when present. Route
+registration is enabled by default and can be disabled with `server.features.grpc-route-registration.enabled=false`.
+Annotate an endpoint with `@RpcServer.Listener("admin")` to register it on a named listener. Custom
+`GrpcRouteRegistration` implementations can return a named socket. If that socket is not configured and
+`socketRequired()` returns `false`, the registration falls back to the default socket. Generated declarative
+registrations require the named listener when `@RpcServer.Listener` is used.
+
+Validation annotations on declarative gRPC server methods are enforced by generated entry point interceptors. To map a
+`ValidationException` to gRPC `INVALID_ARGUMENT`, add `helidon-webserver-grpc-validation` to the server runtime. The
+validation gRPC server service is enabled by default under `grpc.grpc-services.validation`; set
```
Verdent: The `enabled` default applies only after a validation service instance has been selected. `GrpcConfig.grpcServices()` defaults to an empty list and `grpc-services-discover-services` defaults to `false`, so adding this module alone does not instantiate the mapper. The declarative integration test explicitly sets `grpc.grpc-services.validation.enabled=true`; with the dependency-only setup described here, a `ValidationException` is not mapped to `INVALID_ARGUMENT`.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerRegistrationGenerator.java
```diff
@@ -0,0 +1,477 @@
...
+                .addContent("var builder = ")
+                .addContent(GRPC_SERVICE_DESCRIPTOR)
+                .addContent(".builder(")
+                .addContent(endpointType)
+                .addContent(".class, ")
+                .addContentLiteral(endpoint.serviceName())
+                .addContentLine(")")
+                .increaseContentPadding()
+                .increaseContentPadding()
+                .addContentLine(".proto(declarative__proto);")
```
tvallin: **[P1] Do not re-infer imported or nested protobuf classes from the service proto.** Supplying the proto here makes each subsequent `builder.<method>` call invoke `GrpcServiceDescriptor.Builder.getTypeFromMethodDescriptor`, which constructs a Java class name from the service file’s Java package/outer class plus `Descriptor.getName()`. A valid RPC using an imported type such as `google.protobuf.Empty`, or a nested message, passes the new descriptor/type validation and then registration throws `ClassNotFoundException` because that reconstruction names the wrong class. The generator already knows `method.requestType()` and `method.responseType()`; pass those classes through before inference (or resolve from the message descriptor’s own file and full nesting), and add imported/nested-message coverage.

webclient/grpc/src/main/java/io/helidon/webclient/grpc/GrpcClientStreams.java
```diff
@@ -0,0 +1,832 @@
...
+
+    private GrpcClientStreams() {
+    }
+
+    static <ReqT, ResT> Stream<ResT> serverStreaming(ClientCall<ReqT, ResT> call,
+                                                      ReqT request) {
+        Objects.requireNonNull(call);
+        Objects.requireNonNull(request);
+        var responses = new StreamingResponse<ReqT, ResT>();
+        ClientCalls.asyncServerStreamingCall(call, request, responses);
```
tvallin: **[P1] Cancel partially-started server-streaming calls.** `ClientCall.start` can fail after Helidon has opened the HTTP/2 connection and started its read task—for example when the initial header write throws. `asyncServerStreamingCall` propagates that exception without cancelling, and no response `Stream` is returned for the caller to close, so the read task/connection remain live. Please catch start failures here and cancel the call—the client-streaming and bidi paths already do equivalent cleanup—and add a start-failure lifecycle test.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcStreams.java
```diff
@@ -0,0 +1,511 @@
...
+     * @return request observer
+     */
+    public static <ReqT, ResT> StreamObserver<ReqT> clientStreaming(Function<Stream<ReqT>, ResT> handler,
+                                                                    StreamObserver<ResT> responseObserver) {
+        Objects.requireNonNull(handler);
+        Objects.requireNonNull(responseObserver);
+        StreamingRequest<ReqT> request = new StreamingRequest<>(responseObserver);
+        Outbound outbound = new Outbound(responseObserver, request::cancel);
+        ContextRunner contextRunner = ContextRunner.create();
+        Future<?> worker = contextRunner.submit(() -> {
```
Verdent, line 107: Client streaming starts this worker while `ServerCallHandler.startCall` is still constructing the request listener. If the endpoint returns a response without consuming the request stream, the worker can call `responseObserver.onCompleted()` before `GrpcProtocolHandler.init()` assigns `listener`. `ServerCall.close()` then sees `listener == null`, closes the stream, and queues no terminal callback. Once `startCall` returns, the listener is installed but receives neither `onComplete` nor `onCancel`, so interceptor and tracing cleanup tied to those callbacks does not run.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcStreams.java
```diff
@@ -0,0 +1,511 @@
...
+                source = this.source;
+                this.source = null;
+            } finally {
+                lock.unlock();
+            }
+            if (worker != null) {
+                worker.cancel(true);
+            }
+            if (source != null) {
+                THREAD_FACTORY.newThread(() -> {
```
Verdent: When the peer cancels after a response source is installed, this closes the user `Stream` on a new raw virtual thread. Normal response processing and source closure run through the captured `ContextRunner`, but this thread does not, so an `onClose` callback sees neither the call’s gRPC context nor its Helidon request context. The same cleanup hook therefore observes request context on normal completion but loses it on cancellation.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,404 @@
...
+
+        return new GrpcSecurityDefinition(authenticate,
+                                          authenticated.map(it -> it.booleanValue("optional").orElse(false)),
+                                          authenticator,
+                                          authenticated.isPresent() && authenticator.isEmpty(),
+                                          authorize,
+                                          authorizer,
+                                          authorized.isPresent() && authorizer.isEmpty(),
+                                          List.copyOf(rolesAllowed),
+                                          hasDenyAll || hasPermitAll,
```
Verdent: For a class with `@RolesAllowed("admin")` and a method with `@RoleValidator.Roles("user")`, this leaves `clearRolesAllowed` false on the method handler. Combining handlers retains the class-level direct role check, so a caller with only `user` is rejected for missing `admin` before `RoleValidator` applies its method-scope role replacement. `RoleValidator` itself processes class scope before method scope and clears/replaces class roles when method roles are present.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcRouting.java
```diff
@@ -160,27 +168,40 @@ public GrpcRouting build() {
...
+
+                boolean discoverServices = grpcConfig.get("grpc-services-discover-services")
+                        .asBoolean()
+                        .orElseGet(() -> legacyConfig.get("grpc-services-discover-services")
+                                .asBoolean()
+                                .orElse(true));
+                if (discoverServices) {
+                    Set<String> configuredProviderTypes = new LinkedHashSet<>(legacyServices.providerTypes());
+                    configuredProviderTypes.addAll(grpcServices.providerTypes());
+                    addDiscoveredGrpcServices(grpcConfig,
```
Verdent, line 192: With default service discovery, this loads the gRPC security service, which contributes a `GrpcSecurity` interceptor to `configuredInterceptors`. The documented programmatic `.intercept(GrpcSecurity.create(...))` contributes a second instance when routing interceptors are merged. They are distinct objects, so `distinct()` retains both. The discovered instance runs first and establishes `SECURITY_CONTEXT`; the programmatic instance then reuses that context, which can ignore its explicitly supplied `Security`, while security processing—and audit handling when enabled—runs twice.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/grpc/server/GrpcServerExtension.java
```diff
@@ -0,0 +1,404 @@
...
+                .flatMap(it -> it.stringValue("provider"))
+                .filter(not(String::isBlank));
+        Optional<String> authorizer = authorized
+                .flatMap(it -> it.stringValue("provider"))
+                .filter(not(String::isBlank));
+
+        if (hasDenyAll) {
+            authenticate = Optional.of(false);
+            authorize = Optional.of(true);
+        } else if (hasPermitAll) {
```
Verdent, line 418: An inherited `@PermitAll` disables security declared by the override.

Hierarchy processing can retain different annotation types from both the parent and overriding method. With `@PermitAll` on the parent and `@Authenticated` or `@Authorized` on the override, `hasPermitAll` is true here and both security flags are forced to `false`. A compiler reproduction generated `skipAuthentication()` and `skipAuthorization()`, allowing an unauthenticated call to reach the overriding endpoint without its declared security being enforced.

webclient/tests/grpc/src/test/java/io/helidon/webclient/grpc/tests/GrpcTeTrailersTest.java
```diff
@@ -90,13 +172,395 @@ void teTrailersHeaderIsSentToServer() {
...
+                .baseUri("http://localhost:" + serverPort)
+                .protocolConfig(it -> it.nextRequestWaitTime(Duration.ofMillis(50)))
+                .build();
+
+        try (Stream<Strings.StringMessage> responses = timeoutClient.serviceClient(serviceDescriptor)
+                .serverStreaming("PauseAfterTwo", Strings.StringMessage.getDefaultInstance())) {
+            Iterator<Strings.StringMessage> iterator = responses.iterator();
+            assertThat(iterator.next().getText(), is("one"));
+            assertThat("second server response sent", secondSent.await(10, TimeUnit.SECONDS), is(true));
+            Thread.sleep(500);
```
Verdent: This timeout assertion depends on client-reader scheduling.

`secondSent` confirms that the server submitted the second response, but it does not establish that the client reader has received it and entered the 50 ms no-demand wait. If that reader is delayed beyond this 500 ms sleep, the subsequent `iterator.hasNext()` supplies demand before the timeout begins and returns the second response instead of throwing `CANCELLED`, causing a slow-runner failure.

tvallin (dismissed):

tvallin (approved):

PR #12127, opened 2026-06-22
Stabilize HTTP/2 startup test handshake (27.x)

tomas-langer:
Investigates and stabilizes the HTTP/2 test failure seen on PR 12124.

The low-level HTTP/2 test connection helper now fails with an assertion when an expected frame does not arrive instead of dereferencing `null`, ACKs received non-ACK SETTINGS frames, and exposes a handshake helper that drains startup frames before a request stream is opened.

`ContentLengthTest.equalRepeatedContentLengthRejected` now completes the startup handshake before sending the malformed repeated `Content-Length` request, so the test continues to assert the intended connection-fatal protocol error rather than depending on startup-frame timing.

### DIFF
--- a/webserver/testing/junit5/http2/src/main/java/io/helidon/webserver/testing/junit5/http2/Http2TestConnection.java
+++ b/webserver/testing/junit5/http2/src/main/java/io/helidon/webserver/testing/junit5/http2/Http2TestConnection.java
@@ -39,6 +39,7 @@ import io.helidon.http.http2.Http2FrameTypes;
 import io.helidon.http.http2.Http2GoAway;
 import io.helidon.http.http2.Http2Headers;
 import io.helidon.http.http2.Http2HuffmanDecoder;
+import io.helidon.http.http2.Http2Ping;
 import io.helidon.http.http2.Http2RstStream;
 import io.helidon.http.http2.Http2Setting;
 import io.helidon.http.http2.Http2Settings;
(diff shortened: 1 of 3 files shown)


spericas (dismissed):
LGTM

webserver/testing/junit5/http2/src/main/java/io/helidon/webserver/testing/junit5/http2/Http2TestConnection.java
```diff
@@ -146,6 +157,20 @@ public Http2ConnectionWriter writer() {
         return dataWriter;
     }
 
+    /**
+     * Complete HTTP/2 startup frame exchange without opening a request stream.
+     *
+     * @param timeout timeout for each expected frame
+     */
```
tvallin, line 164: This makes `completeHandshake` depend on the server sending a connection `WINDOW_UPDATE`, but `Http2Connection.writeServerSettings()` only sends that frame when `Http2Config.initialWindowSize()` is greater than `WindowSize.DEFAULT_WIN_SIZE`. Tests can configure HTTP/2 with the default window size, in which case the next frame after server `SETTINGS` is the server `SETTINGS` ACK and this helper times out even though the handshake is valid. Could this make the window update optional, or otherwise limit/document the helper to configs that send the startup window update?

tomas-langer: Updated in 7b5a4784fd and 888961a09f. `completeHandshake` now accepts either the startup connection `WINDOW_UPDATE` followed by the server `SETTINGS` ACK, or the server `SETTINGS` ACK directly. The helper test covers `WindowSize.DEFAULT_WIN_SIZE`, and the content-length tests now complete the startup handshake before sending malformed requests so they no longer depend on unread startup frames.

tvallin (approved):

PR #12128, opened 2026-06-22
Correct the calc of OTel http server sem conv; capture end time in whenSent instead of in the filter

### LINKED ISSUE #12097: OTLP HTTP duration metric is inflated by 1000x
## Summary

In Helidon 4.4.1, the built-in OpenTelemetry inbound HTTP metric `http.server.request.duration` appears to record request duration in milliseconds, while the metric name/unit and Prometheus export indicate seconds.

This makes exported duration values and histogram quantiles appear about **1000x larger** than the actual request latency.

## Environment

- Helidon MP 4.4.1
- `helidon-webserver-observe-telemetry-metrics`
- `opentelemetry-exporter-otlp`
- OTLP metrics export over gRPC
- Metrics viewed through Prometheus/Grafana via an OTLP collector

## Observed Behavior

I created a simple JAX-RS test endpoint which always returns `200 OK` and intentionally sleeps for a small millisecond-range latency:

```java
@GET
@Path("/duration-unit-probe")
@WithSpan
public Response durationUnitProbe() {
    sleep(durationUnitProbeLatencyMillis());
    return build("duration-unit-probe", "request", Status.OK_200);
}

long durationUnitProbeLatencyMillis() {
    return 6L + (long) Math.floor(randomDoubleSupplier.getAsDouble() * 4L);
}
```

So the intentional server-side sleep is uniformly distributed across `6 ms`, `7 ms`, `8 ms`, and `9 ms`, plus normal request overhead.

A 300-request client-side `curl` timing sample showed:

```text
Client-side curl timings
samples  300 successful
min      7.485 ms (0.007485 s)
p50      11.116 ms (0.011116 s)
p90      13.083 ms (0.013083 s)
p95      13.483 ms (0.013483 s)
p99      13.971 ms (0.013971 s)
max      14.640 ms (0.014640 s)
avg      11.092 ms (0.011092 s)
```

For a seconds histogram, these requests should fall into low second buckets, mostly below `le="0.025"` seconds.

Instead, the exported `http_server_request_duration_seconds_bucket` series showed:

```text
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.005", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.01", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.025", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.05", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.075", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.1", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.25", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.5", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="0.75", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="1", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="2.5", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="5", ...} 0
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="7.5", ...} 32
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="10", ...} 751
http_server_request_duration_seconds_bucket{http_request_method="GET", http_response_status_code="200", le="+Inf", ...} 2114
```

This looks exactly like milliseconds are being recorded as if they were seconds:

```text
Actual client p50: 0.011116 s
Recorded as:        ~11.116 s

Actual client p95: 0.013483 s
Recorded as:        ~13.483 s
```

That explains the bucket behavior:

- Low second buckets such as `le="0.025"` are empty, even though they should contain nearly all requests.
- `le="7.5"` contains only the lower tail, corresponding to requests around `7.5 ms`.
- `le="10"` contains more requests, corresponding to requests up to around `10 ms`.
- Many requests are above `le="10"` and only appear in `+Inf`, matching the client-side p50/p95/max being above `10 ms`.

## Suspected Cause

Looking at the 4.4.1 implementation of the built-in HTTP metrics filter, the duration appears to be recorded roughly as:

```java
httpRequestDuration.record((endTime - startTime) / 1_000_000.0, attributes);
```

See this line of code: https://github.com/helidon-io/helidon/blob/main/webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java#L190

`System.nanoTime()` returns nanoseconds. Dividing by `1_000_000.0` converts nanoseconds to **milliseconds**, not seconds.

But the OpenTelemetry HTTP server duration metric is `http.server.request.duration`, and Prometheus exports it as:

```text
http_server_request_duration_seconds_bucket
http_server_request_duration_seconds_count
http_server_request_duration_seconds_sum
```

So the recorded value should be in seconds, which would require:

```java
(endTime - startTime) / 1_000_000_000.0
```

## Expected Behavior

A request taking about `11 ms` should be recorded as:

```text
0.011
```

not:

```text
11.0
```

For the test endpoint above, nearly all samples should land below `le="0.025"` seconds, not around `le="7.5"`, `le="10"`, or `+Inf`.

## Impact

Latency dashboards, SLOs, and alerts based on `http_server_request_duration_seconds_bucket` are misleading:

- p50/p95/p99 latency appears about **1000x too high**.
- Normal millisecond requests look like multi-second requests.
- Low-latency buckets remain empty.
- Histogram quantiles are unusable unless manually divided by 1000.

## Related PR

This issue appears to have been introduced when `Support OpenTelemetry server HTTP semantic conventions metrics` PR landed: 

https://github.com/helidon-io/helidon/pull/11168/changes#diff-013265252fe516c93a49cc078ff69b7ee3d1ab6a97a001c91a5a2b81e0a15ad8R174

## Possible Fix

Change the duration conversion from nanoseconds-to-milliseconds to nanoseconds-to-seconds, for example:

```java
httpRequestDuration.record((endTime - startTime) / 1_000_000_000.0, attributes);
```

or use `Duration`/`TimeUnit` conversion to make the intended unit explicit.

cc @tjquinno

tjquinno:
### Description
Resolves #12097 

### Release Note
____
Helidon now correctly expresses the `http.server.request.duration` OpenTelemetry metric in seconds, and correctly includes all processing (including all filters and payload preparation).
____

### PR overview
As noted in the issue, the code actually and incorrectly converted the duration to ms instead of seconds.

Further, the original code updated the timer as if all processing of the request finished in the filter, which is inaccurate. The PR changes this to compute the ending time of the duration in `whenSent` which includes all filters and time spent actually sending the response. 

### Documentation
No impact

### DIFF
--- a/webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
+++ b/webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
@@ -18,6 +18,7 @@ package io.helidon.webserver.observe.telemetry.metrics;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 
 import io.helidon.config.Config;
 import io.helidon.http.Status;
@@ -123,27 +124,24 @@ class OpenTelemetryMetricsHttpSemanticConventions implements AutoHttpMetricsProv
         @Override
         public void filter(FilterChain chain, RoutingRequest req, RoutingResponse res) {
             var startTime = System.nanoTime();
+            var exception = new AtomicReference<Exception>();
             /*
-            Duplicating the synch/async handling in the normal and exception case avoids the overhead of using an Optional to hold
-            the exception (if any) for use in a lambda.
-            */
+            Update the timer in whenSent rather than here in this filter. That way we include time spent in running succeeding
+            filters and in preparing the response entity, to more accurately capture as much as possible the full time the
+            server spent responding to the request.
+             */
+            res.whenSent(() -> {
+                try {
+                    updateMetricsIfMeasured(req, res, startTime, System.nanoTime(), exception.get());
+                } catch (Throwable e) {
+                    LOGGER.log(WARNING, "Failed to record HTTP request metrics", e);
+                }
+            });
+
             try {
                 chain.proceed();
-                Thread.ofVirtual().start(() -> {
-                    try {
-                        updateMetricsIfMeasured(req, res, startTime, System.nanoTime(), null);
-                    } catch (Throwable e) {
-                        LOGGER.log(WARNING, "Failed to record HTTP request metrics", e);
-                    }
-                });
             } catch (Exception e) {
-                Thread.ofVirtual().start(() -> {
-                    try {
-                        updateMetricsIfMeasured(req, res, startTime, System.nanoTime(), e);
-                    } catch (Throwable metricsUpdateFailure) {
-                        LOGGER.log(WARNING, "Failed to record HTTP request metrics", metricsUpdateFailure);
-                    }
-                });
+                exception.set(e);
                 throw e;
             }
         }
@@ -187,7 +185,7 @@ class OpenTelemetryMetricsHttpSemanticConventions implements AutoHttpMetricsProv
             don't currently have a way to get the HTTP version at runtime from a request.
              */
 
-            httpRequestDuration.record((endTime - startTime) / 1_000_000.0, attrBuilder.build());
+            httpRequestDuration.record((endTime - startTime) / 1_000_000_000.0, attrBuilder.build());
         }
 
         private String errorType(RoutingResponse resp, Exception exception) {
--- a/webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
+++ b/webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
@@ -168,7 +168,7 @@ class OpenTelemetryMetricsHttpSemanticConventions implements AutoHttpMetricsProv
             attrBuilder.put(AttributeKey.stringKey(HTTP_METHOD), req.prologue().method().text())
                     .put(AttributeKey.stringKey(URL_SCHEME), req.prologue().protocol())
                     .put(AttributeKey.stringKey(ERROR_TYPE), errorType(resp, exception))
-                    .put(AttributeKey.longKey(STATUS_CODE), statusCode(resp, exception))
+                    .put(AttributeKey.longKey(STATUS_CODE), resp.status().code())
                     .put(AttributeKey.stringKey(HTTP_ROUTE), req.matchingPattern().orElse(""))
                     .put(AttributeKey.stringKey(SOCKET_NAME), req.listenerContext().config().name());
 
@@ -196,10 +196,5 @@ class OpenTelemetryMetricsHttpSemanticConventions implements AutoHttpMetricsProv
                             : resp.status().codeText();
         }
 
-        private long statusCode(RoutingResponse resp, Exception exception) {
-            return (exception != null)
-                    ? 0L
-                    : resp.status().code();
-        }
     }
 }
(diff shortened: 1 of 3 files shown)


webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -123,27 +124,24 @@ private MetricsRecordingFilter(DoubleHistogram httpRequestDuration, AutoHttpMetr
...
-            Duplicating the synch/async handling in the normal and exception case avoids the overhead of using an Optional to hold
-            the exception (if any) for use in a lambda.
-            */
+            Update the timer in whenSent rather than here in this filter. That way we include time spent in running succeeding
+            filters and in preparing the response entity, to more accurately capture as much as possible the full time the
+            server spent responding to the request.
+             */
+            res.whenSent(() -> {
+                try {
+                    updateMetricsIfMeasured(req, res, startTime, System.nanoTime(), exception.get());
```
tomas-langer, line 135: Now that recording runs from `whenSent`, this exception branch can run after Helidon has already sent an error response. Returning `0` here drops the actual `resp.status().code()` for handled exceptions, so requests converted to 4xx/5xx by error handling are exported with `http.response.status_code=0` even though a status was sent. The OTel HTTP metric contract requires `http.response.status_code` when a response was sent; this should use the response status after send, not the captured exception as a blanket status override.

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
```diff
@@ -270,4 +245,113 @@ void checkAutoMetrics() {
...
+                            .map(JsonValue::asObject)
+                            .filter(dataPoint -> "/greet/{name}".equals(dataPoint.arrayValue("attributes")
+                                    .map(JsonTestUtil::asAttributesMap)
+                                    .orElseGet(Map::of)
+                                    .get(HTTP_ROUTE)))
+                            .forEach(dataPoint -> {
+                                var duration = dataPoint.doubleValue("sum").orElseThrow();
+                                var durationNanos = switch (unit) {
+                                case "s" -> Math.round(duration * 1_000_000_000.0);
+                                case "ms" -> Math.round(duration * 1_000_000.0);
```
tomas-langer: The regression test should assert the exported unit is `"s"` for `http.server.request.duration`. Accepting `"ms"` means the test can still pass if production is changed back to millisecond units consistently, which is exactly the unit regression this PR is supposed to prevent.

tomas-langer (changes requested):
Requesting changes for the two inline issues below.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: Changed.

webserver/observe/telemetry/metrics/src/test/java/io/helidon/webserver/observe/telemetry/metrics/TestOpenTelemetrySemanticConventions.java
tjquinno: Fixed. Enforces "s" now.

tomas-langer (approved):

PR #12133, opened 2026-06-24
Fix API stability record constructor errors on JDK 26 (4.x)

### LINKED ISSUE #12132: helidon-codegen-api-stability triggers invalid record canonical constructor errors on JDK 26
## Environment Details
* Helidon Version: 4.5.0
* Helidon SE or Helidon MP: N/A for the minimal reproducer; originally observed in a Helidon SE application when using `io.helidon.bundles:helidon-bundles-apt:4.5.0`.
* JDK version: Java 26, Oracle build `26+35-2893`
* OS: Linux `5.14.0-687.5.3.el9_8.x86_64` (`amd64`)

----------

## Problem Description

`io.helidon.codegen:helidon-codegen-api-stability:4.5.0` can make valid Java record source fail compilation on JDK 26 with a misleading javac error:

```text
invalid canonical constructor in record Repro
  (canonical constructor must not contain explicit constructor invocation)
```

The record canonical constructor does not contain an explicit constructor invocation. It compiles successfully when the API stability processor is not run.

This is reproducible with a minimal Maven project that uses only `helidon-codegen-api-stability` as an annotation processor. It was originally observed after upgrading a project from Helidon 4.4.1 to 4.5.0 because `helidon-codegen-api-stability` was added to `helidon-bundles-apt` in 4.5.0 and is auto-discovered by javac.

Minimal source:

```java
package example;

record Repro(String name) {
    Repro {
        name = name.trim();
    }
}
```

Failure output:

```text
[ERROR] COMPILATION ERROR :
[ERROR] .../src/main/java/example/Repro.java:[4,5] invalid canonical constructor in record Repro
[ERROR]   (canonical constructor must not contain explicit constructor invocation)
```

Expected behavior:

* The source should compile.
* The API stability processor should only report Helidon API stability diagnostics, not cause unrelated record canonical constructor errors.
* `-Ahelidon.api=ignore` should ideally avoid API stability scanning side effects, but currently the processor still scans before deciding not to report diagnostics.

Additional investigation:

The likely trigger is in `ApiStabilityScanner`, which resolves expression nodes using `Trees.getElement(path)`:

```java
@Override
public Void visitIdentifier(IdentifierTree node, Void unused) {
    var path = trees.getPath(unit, node);
    return visit(path, () -> super.visitIdentifier(node, unused));
}

private Void visit(TreePath path, Supplier<Void> superCall) {
    if (path != null) {
        var elt = trees.getElement(path);
```

A small isolated processor test showed:

* `Trees.getPath(...)` on these nodes: passes
* `Trees.getTypeMirror(...)` on these nodes: passes
* `Trees.getElement(...)` on identifiers/member-selects/new-class expressions inside a record canonical constructor body: triggers the javac error
* normal classes and non-canonical record constructors: pass

So the processor appears to hit a JDK 26/javac interaction by resolving expression trees in record canonical constructors.

## Steps to reproduce

A minimal reproducer is attached in this issue as `bug-report/reproducer`.

1. Use JDK 26.
2. Run:

   ```bash
   cd bug-report/reproducer
   mvn clean compile
   ```

3. The compile fails with:

   ```text
   invalid canonical constructor in record Repro
     (canonical constructor must not contain explicit constructor invocation)
   ```

4. Disable annotation processing and the same Java source compiles:

   ```bash
   mvn clean compile -Dmaven.compiler.proc=none
   ```

5. Suppressing Helidon API diagnostics does not avoid the failure because scanning still happens:

   ```bash
   mvn clean compile -Dmaven.compiler.compilerArgs=-Ahelidon.api=ignore
   ```

Potential fixes to consider:

* Remove `helidon-codegen-api-stability` from the default `helidon-bundles-apt` bundle, or
* skip this processor's source scan entirely when all API stability checks are configured as `ignore`, or
* avoid `Trees.getElement(...)` for expression nodes inside record canonical constructors on JDK 26, or otherwise guard that path.

[helidon-api-stability-record-reproducer.zip](https://github.com/user-attachments/files/29284772/helidon-api-stability-record-reproducer.zip)

tomas-langer:
Resolves #12132

Avoid JDK 26 javac side effects from API stability scanning in record constructors, and keep API stability diagnostics fail-open when source scanning hits an unexpected processor error.

### DIFF
--- a/codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityProcessor.java
+++ b/codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityProcessor.java
@@ -153,6 +153,13 @@ public class ApiStabilityProcessor extends AbstractProcessor {
     @Override
     public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
         try {
+            if (previewAction == Action.IGNORE
+                    && incubatingAction == Action.IGNORE
+                    && internalAction == Action.IGNORE
+                    && deprecatedAction == Action.IGNORE) {
+                return false;
+            }
+
             var trees = Trees.instance(processingEnv);
 
             for (var rootElement : roundEnv.getRootElements()) {
@@ -175,12 +182,18 @@ public class ApiStabilityProcessor extends AbstractProcessor {
             Set<Ref> deprecatedApis = new LinkedHashSet<>();
 
             for (var unit : compilationUnits) {
-                var scanner = new ApiStabilityScanner(trees, unit);
-                scanner.scan(unit, null);
-                previewApis.addAll(scanner.previewApis());
-                incubatingApis.addAll(scanner.incubatingApis());
-                privateApis.addAll(scanner.internalApis());
-                deprecatedApis.addAll(scanner.deprecatedApis());
+                try {
+                    var scanner = new ApiStabilityScanner(trees, unit);
+                    scanner.scan(unit, null);
+                    previewApis.addAll(scanner.previewApis());
+                    incubatingApis.addAll(scanner.incubatingApis());
+                    privateApis.addAll(scanner.internalApis());
+                    deprecatedApis.addAll(scanner.deprecatedApis());
+                } catch (Throwable e) {
+                    messager.printWarning("Failed to scan a source file in ApiStabilityProcessor. "
+                                                  + "API stability diagnostics may be incomplete. Exception class: "
+                                                  + e.getClass().getName() + ", message: " + e.getMessage());
+                }
             }
 
             scanModuleDirectives(modules, previewApis, incubatingApis, privateApis, deprecatedApis);
--- a/codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityScanner.java
+++ b/codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityScanner.java
@@ -34,6 +34,7 @@ import com.sun.source.tree.CompilationUnitTree;
 import com.sun.source.tree.IdentifierTree;
 import com.sun.source.tree.MemberReferenceTree;
 import com.sun.source.tree.MemberSelectTree;
+import com.sun.source.tree.MethodTree;
 import com.sun.source.tree.NewClassTree;
 import com.sun.source.tree.Tree;
 import com.sun.source.util.TreePath;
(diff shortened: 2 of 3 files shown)


Verdent (approved):

codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityScanner.java
```diff
@@ -275,7 +276,29 @@ private Element enclosingElement(TreePath node) {
...
-        return trees.getElement(node);
+        return element(node);
+    }
+
+    private Element element(TreePath path) {
+        /*
+         * JDK 26 javac can report a bogus record canonical constructor error when an annotation
+         * processor asks Trees.getElement for nodes inside record constructors.
+         */
+        return inRecordConstructor(path) ? null : trees.getElement(path);
```
danielkec, line 287: This guard is broader than the JDK 26 failure described in #12132. Because `element(...)` returns `null` for every path under any record constructor, the scanner no longer reports `@Api.Internal`/`@Api.Incubating`/`@Api.Preview`/`@Deprecated` usages in those constructor bodies. A simple repro is an overloaded record constructor that uses an `@Api.Internal` type under `-Ahelidon.api.internal=fail`; it now compiles without the source-level diagnostic even though the issue notes non-canonical record constructors were not the javac trigger.

codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityProcessor.java
```diff
@@ -153,6 +153,13 @@ public synchronized void init(ProcessingEnvironment processingEnv) {
     @Override
     public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
         try {
+            if (previewAction == Action.IGNORE
```
danielkec, line 156: This is the branch intended to avoid the issue's `-Ahelidon.api=ignore` scanning side effect, but the tests do not exercise the record compact-constructor reproducer in ignore mode. If this branch regressed, the existing global-ignore test with ordinary fields would still pass while the linked issue's ignore-mode case could fail again.

danielkec (commented):
Review comments from the exact-head pass.

codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityScanner.java
tomas-langer: This is a workaround for a JDK bug, where we need to survive an unexpected behavior. It would be too complicated to handle this without proper support of the tree. So we have a known issue that if API is used from within a custom records constructor, it will be accepted.

codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityProcessor.java
tomas-langer: Yes. We can create a  follow up issue

PR #12135, opened 2026-06-24
Fix API stability record constructor errors on JDK 26 (main)

### LINKED ISSUE #12134: Backport API stability record constructor fix to main
The 4.x fix for #12132 was merged in #12133. We need to carry the same fix to `main` / 27.x.

The original failure was caused by JDK 26 `javac` emitting an invalid canonical constructor diagnostic when the API stability annotation processor calls `Trees.getElement(...)` while scanning compact record constructors.

The 4.x fix avoids `Trees.getElement(...)` under record constructors, accepts that API stability diagnostics may be incomplete in that narrow area, and keeps user compilation fail-open if source scanning hits an unexpected processor error.

Target branch: `main`

tomas-langer:
Resolves #12134

Carries the API stability processor workaround for JDK 26 record compact constructors to `main`.

The scanner now avoids `Trees.getElement(...)` for nodes under record constructors, the processor skips source scanning when all API stability categories are ignored, and individual source scan failures no longer prevent scanning later source files. The regression tests cover both the compact-constructor reproducer and the global-ignore mode requested in the 4.x review.

### DIFF
--- a/codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityProcessor.java
+++ b/codegen/api-stability/src/main/java/io/helidon/codegen/api/stability/ApiStabilityProcessor.java
@@ -153,6 +153,13 @@ public class ApiStabilityProcessor extends AbstractProcessor {
     @Override
     public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment roundEnv) {
         try {
+            if (previewAction == Action.IGNORE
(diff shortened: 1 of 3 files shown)


danielkec (approved):
Reviewed exact head 39c02785c760c8b0e544ac5ef6d6ba369dc8f9df; no actionable findings.

Verdent (approved):

PR #12138, opened 2026-06-25
Update metrics factory documentation (4.x)

### LINKED ISSUE #12126: Check documentation for static factory methods in Metrics
We should update the documentation to make sure this is no longer recommended (tags, various metrics have the same isssue.)

Example:

```java
static Tag create(String key, String value) {
        return MetricsFactory.getInstance().tagCreate(key, value);
    }
```

This method will be deprecated in a future release, and it should not be used (as it may use a different registry than the one used by the application, this invites usage in constants, and that cannot be correctly tested.

tomas-langer:
Resolves #12126

Updates metrics API Javadocs and SE documentation snippets to use the 4.x metrics factory and service registry APIs instead of static factory shortcuts deprecated on main.

### DIFF
--- a/docs/src/main/asciidoc/se/guides/metrics.adoc
+++ b/docs/src/main/asciidoc/se/guides/metrics.adoc
@@ -1,6 +1,6 @@
 ///////////////////////////////////////////////////////////////////////////////
 
-    Copyright (c) 2019, 2025 Oracle and/or its affiliates.
+    Copyright (c) 2019, 2026 Oracle and/or its affiliates.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


docs/src/main/asciidoc/se/guides/metrics.adoc
```diff
@@ -161,7 +161,7 @@ The `Counter` {metric} is a monotonically increasing number. The following examp
 include::{sourcedir}/se/guides/MetricsSnippets.java[tag=snippet_3, indent=0]
 ----
 <1> Declare a `Counter` member field.
-<2> Create and register the `Counter` {metric} in the global meter registry`.  This `Counter` will exist for the lifetime of
+<2> Create and register the `Counter` {metric} in the global meter registry using a builder from `MetricsFactory`. This `Counter` will exist for the lifetime of
```
tjquinno, line 164: Do we also want to change uses of the term "global meter registry" to "application meter registry" instead? 

If so, there are probably additional uses of the term "global" in the docs that we would want to change at the same time so maybe that's for a separate PR.

In one of the changes below there is a new use of "shared" registry. That would work as well as "application" although IIRC  some changes already in `main` refer to the application registry.

tjquinno (commented):
LGTM with one possible change.

docs/src/main/asciidoc/se/guides/metrics.adoc
tomas-langer: The intention of this PR was to make the current docs aligned with what we will deprecate in 27. I did not intend to "coin" a new term for the globa/application/shared registry or factory.

tjquinno: The revised `MeterRegistry` line 33 introduces "shared." The other changes seem good to me so I'll approve this in case you're OK with that usage of "shared," but other places simply use "For new code, prefer..."

tjquinno (approved):

PR #12139, opened 2026-06-25
Fix HTTP/2 split header block serialization (4.x)

### LINKED ISSUE #12100: HTTP/2 split header blocks are serialized incorrectly (4.x)
Backport of #12099 to 4.x.

Large HTTP/2 header blocks serialized through `Http2ConnectionWriter.writeHeaders(...)` can produce an invalid frame sequence when the encoded header block is larger than the peer `SETTINGS_MAX_FRAME_SIZE`.

In the split path:

1. The first `HEADERS` frame is written with flags `0`, so a header-only write that originally passed `END_STREAM | END_HEADERS` loses `END_STREAM`.
2. The continuation loop writes every fragment after the first, including the final fragment.
3. The final block writes that same last fragment again with `END_HEADERS`.

For a split header block `A/B/C`, the peer receives `A, B, C, C(END_HEADERS)`. For header-only messages, the peer also does not receive the original `END_STREAM` flag.

Expected behavior:

1. The first `HEADERS` frame keeps `END_STREAM` when the original flags include it, but defers `END_HEADERS` when continuation frames follow.
2. Each continuation fragment is written exactly once.
3. Only the final `CONTINUATION` frame carries `END_HEADERS`.

Affected code:

- `http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java`

tomas-langer:
Resolves #12100

Backports the HTTP/2 split header block serialization fix from #12102 to 4.x, preserving END_STREAM on the initial HEADERS frame and emitting each continuation fragment exactly once.

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


danielkec (approved):
Reviewed exact head d75c8f6454fa56c827bcce948a4f96859146811d. No blocking findings.

PR #12143, opened 2026-06-26
Fix HTTP/2 reset stream cleanup race

tomas-langer:
A rejected HTTP/2 stream can reset before the request body is fully received. If delayed DATA arrives after the stream was removed, the connection can treat it as an unexpected stream update and prevent a replacement stream from starting. This updates reset cleanup so unread-body resets are registered before removal and adds a regression test for delayed DATA after a rejected request.

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
@@ -24,6 +24,8 @@ import java.util.Set;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -87,7 +89,7 @@ class Http2ServerStream implements Runnable, Http2Stream {
                                                   Http2Flag.DataFlags.create(Http2Flag.DataFlags.END_OF_STREAM),
                                                   0), BufferData.empty());
     private static final Runnable NO_OP = () -> { };
-    private static final long MAX_LOCALLY_RESET_STREAM_DATA = 64 * 1024;
+    private static final long MAX_LOCALLY_RESET_STREAM_HEADER_BYTES = 64 * 1024;
     private static final int MAX_LOCALLY_RESET_STREAM_HEADER_BLOCKS = 64;
     private static final System.Logger LOGGER = System.getLogger(Http2Stream.class.getName());
     private static final Set<Http2StreamState> DATA_RECEIVABLE_STATES =
@@ -95,7 +97,6 @@ class Http2ServerStream implements Runnable, Http2Stream {
 
     private final ConnectionContext ctx;
     private final Http2Config http2Config;
-    private final long maxLocallyResetStreamData;
     private final List<Http2SubProtocolSelector> subProviders;
     private final int streamId;
     private final Http2Settings serverSettings;
@@ -106,6 +107,7 @@ class Http2ServerStream implements Runnable, Http2Stream {
     private final Http2ConnectionChecks connectionAttackVectorMetrics;
     private final LocallyResetStreamTracker locallyResetStreamTracker;
     private final ArrayBlockingQueue<DataFrame> inboundData = new ArrayBlockingQueue<>(32);
+    private final ConnectionFlowControl connectionFlowControl;
     private final StreamFlowControl flowControl;
     private final Http2ConcurrentConnectionStreams streams;
     private final HttpRouting routing;
@@ -113,10 +115,8 @@ class Http2ServerStream implements Runnable, Http2Stream {
     private final ReentrantLock resetCompletionLock = new ReentrantLock();
     private boolean wasLastDataFrame = false;
     private boolean hasEntity = true;
-    private long ignoredInboundDataAfterReset;
-    private long ignoredInboundHeaderBytesAfterReset;
-    private int ignoredInboundHeaderBlocksAfterReset;
     private volatile Http2Headers headers;
+    private volatile LocallyResetStreamState locallyResetStreamState;
     private volatile Http2Priority priority;
     private boolean remoteResetReceived;
     private volatile boolean resetStreamSent;
@@ -162,7 +162,6 @@ class Http2ServerStream implements Runnable, Http2Stream {
         this.streams = streams;
         this.routing = routing;
         this.http2Config = http2Config;
-        this.maxLocallyResetStreamData = Math.max(http2Config.maxFrameSize(), MAX_LOCALLY_RESET_STREAM_DATA);
         this.subProviders = subProviders;
         this.streamId = streamId;
         this.serverSettings = serverSettings;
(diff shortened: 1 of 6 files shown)


webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -397,18 +398,17 @@ public void run() {
             }
             Http2Headers http2Headers = Http2Headers.create(headers)
                     .status(e.status());
+            boolean resetRequestBody = closeRejectedStreamState(false);
```
tvallin: Calling `closeRejectedStreamState(false)` here removes/deactivates the stream before the error response frames and the `RST_STREAM` below have actually been written. When the direct handler returns an entity, `writeHeaders(..., dataFrame, flowControl.outbound())` can block waiting for `WINDOW_UPDATE`; during that time `maxConcurrentStreams` has already been freed even though this stream is still open on the wire. Please keep the stream active until the END_STREAM response/RST is written, or add validation with a zero client stream window, a rejected request body, and a replacement stream opened before WINDOW_UPDATE.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -678,29 +679,37 @@ private BufferData readEntityFromPipeline() {
...
     }
 
-    private void closeRejectedStream(Http2ErrorCode resetCode, boolean forceReset) {
+    private boolean closeRejectedStreamState(boolean forceReset) {
         boolean resetRequestBody = forceReset || hasEntity && state != Http2StreamState.HALF_CLOSED_REMOTE;
         ignoreInboundDataAfterReset = resetRequestBody;
         this.state = Http2StreamState.CLOSED;
         writeState.updateAndGet(s -> s.checkAndMove(WriteState.END));
+        if (resetRequestBody) {
+            locallyResetStreams.accept(this.streamId);
```
tvallin: Adding rejected streams to `locallyResetStreams` only protects the later `DATA` path. If request trailers were already in flight after the server-side reset, they arrive as `HEADERS`/`CONTINUATION`; `doHeaders()` still follows normal stream lookup/header handling for the closed stream and can turn that into a connection-level `GOAWAY`. Please add handling to decode/drop trailing header blocks for locally reset streams and validate that trailing `HEADERS | END_STREAM` after a rejected request does not prevent a later stream from succeeding.

webserver/tests/http2/src/test/java/io/helidon/webserver/tests/http2/ConcurrentStreamsLimitTest.java
```diff
@@ -262,6 +263,43 @@ void rejectedRequestDoesNotConsumeConcurrentStreamLimit(Http2TestClient client)
...
+            assertErrorResponse(h2conn, 1, Status.UNSUPPORTED_MEDIA_TYPE_415);
+            Http2RstStream rstStream = h2conn.assertRstStream(1, TIMEOUT);
+            assertThat(rstStream.errorCode(), is(Http2ErrorCode.CANCEL));
+
+            h2conn.request(3, POST, BLOCKING_PATH, WritableHeaders.create(), BufferData.create(new byte[0]));
+            assertThat("Replacement stream did not start",
+                       requestsStarted.await(TIMEOUT.toSeconds(), TimeUnit.SECONDS),
+                       is(true));
+
+            h2conn.writer().writeData(new Http2FrameData(Http2FrameHeader.create(0,
```
tvallin: This regression test sends only a zero-length DATA frame with `END_STREAM`, so it does not exercise the non-empty discard path that updates connection flow control and drains delayed request-body bytes for locally reset streams. Please include a small non-empty DATA frame, or a payload frame followed by END_STREAM, so the new behavior covers delayed body data as well as the empty terminator.

tvallin (dismissed):

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -280,6 +303,17 @@ public void headers(Http2Headers headers, boolean endOfStream) {
     @Override
     public void data(Http2FrameHeader header, BufferData data, boolean endOfStream) {
         int dataLength = data.available();
+        if (ignoreInboundDataAfterReset) {
+            flowControl.inbound().incrementWindowSize(header.length());
+            if (!discardDataAfterReset(header.length())) {
```
tvallin: The 64 KiB discard cap can close the connection for protocol-compliant in-flight data. `ignoreInboundDataAfterReset` is set before the terminal response and `RST_STREAM` are written, while the default advertised receive window is 1 MiB, so the peer can already have more than 64 KiB queued. Processing that tail produces `GOAWAY(ENHANCE_YOUR_CALM)` even though it was sent before the reset was observable. Please base this bound on outstanding flow-control credit and add a case with more than 64 KiB queued before the reset.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -678,28 +758,89 @@ private BufferData readEntityFromPipeline() {
...
+                    locallyResetStreamTracker.add(this.streamId);
+                    if (remoteCompleteAfterReset || remoteAlreadyComplete) {
+                        locallyResetStreamTracker.remoteComplete(this.streamId);
+                    }
+                }
+            } finally {
+                resetCompletionLock.unlock();
+            }
+        } finally {
+            this.state = Http2StreamState.CLOSED;
```
tvallin, line 829: `hasEntity == false` does not imply that the remote half is closed. Initial HEADERS can contain `Content-Length: 0` without `END_STREAM`, leaving the stream open for terminal DATA or trailers. Here no reset or tracker is installed, but this unconditional transition to `CLOSED` removes the live remote half. A subsequent legal `END_STREAM` frame can cause `GOAWAY`, and the concurrent-stream slot is released early. Please either reset when the remote half remains open or retain it as `HALF_CLOSED_LOCAL`, and cover this sequence.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -302,14 +336,25 @@ public void data(Http2FrameHeader header, BufferData data, boolean endOfStream)
...
+    void enqueueDataAfterPrecheck(Http2FrameHeader header, BufferData data, boolean endOfStream) {
         DataFrame frame = new DataFrame(header, data);
         try {
             inboundData.put(frame);
         } catch (InterruptedException e) {
             throw new Http2Exception(Http2ErrorCode.INTERNAL, "Interrupted", e);
         }
-        if (!DATA_RECEIVABLE_STATES.contains(state) && inboundData.remove(frame)) {
-            flowControl.inbound().incrementWindowSize(header.length());
+        boolean ignoringInboundData = ignoreInboundDataAfterReset;
```
tvallin: This snapshot can become stale before the state check below. If the reader enqueues an `END_STREAM` frame and reads `false`, then the rejection thread drains that frame, sends the reset, installs the tracker, and closes the stream, this branch observes `CLOSED` but skips `remoteCompleteAfterReset()`. The tracker never records the already-received `END_STREAM` and can eventually overflow. Please re-check the reset state before deciding whether to record remote completion, or preserve `END_STREAM` while draining the queue, and add a deterministic interleaving test.

tvallin (approved):

Issue #12015, opened 2026-05-22
4.x: Jersey connector sends multipart requests without generated boundary


romain-grecourt:
## Environment Details
* Helidon Version: 4.4.1
* Helidon MP

----------

## Problem Description

Multipart requests sent with `HelidonConnectorProvider` can fail because the generated `boundary` parameter is not included in the outbound `Content-Type` header.

The same request succeeds with Jersey's `HttpUrlConnectorProvider`.

## Steps to reproduce

The following test reproduces the issue and fails with HTTP 400 before the resource returns `ok`.

```java
@ServerTest
class MultipartHeaderUpdateTest {
    private final Client client;
    private final URI uri;

    MultipartHeaderUpdateTest(URI uri) {
        var config = new ClientConfig();
        config.connectorProvider(HelidonConnectorProvider.create());
        config.register(MultiPartFeature.class);

        this.client = ClientBuilder.newClient(config);
        this.uri = uri;
    }

    @SetUpRoute
    static void routing(HttpRouting.Builder routing) {
        var resourceConfig = new ResourceConfig(MultipartResource.class)
                .register(MultiPartFeature.class);
        routing.register("/jaxrs", JaxRsService.create(Config.empty(), resourceConfig));
    }

    @AfterEach
    void closeClient() {
        client.close();
    }

    @Test
    void testMultipartBoundaryHeaderUpdatedBeforeSend() throws IOException {
        try (var multipart = new MultiPart()
                .bodyPart(new BodyPart("hello", MediaType.TEXT_PLAIN_TYPE));
             var response = client.target(uri)
                     .path("jaxrs/multipart")
                     .request()
                     .post(Entity.entity(multipart, MediaType.MULTIPART_FORM_DATA_TYPE))) {
            assertThat(response.getStatus(), is(200));
            assertThat(response.readEntity(String.class), is("ok"));
        }
    }

    @Path("/multipart")
    public static class MultipartResource {
        @POST
        @Consumes(MediaType.MULTIPART_FORM_DATA)
        @Produces(MediaType.TEXT_PLAIN)
        public String upload(MultiPart ignored) {
            return "ok";
        }
    }
}
```

oleksii-tumanov:
I prepared the fix for `main` in two parts:

1. #12153 propagates late `Content-Type` changes through WebClient
2. helidon-io/helidon-microprofile#63 refreshes Jersey's generated multipart boundary

[...]

PR #12153, opened 2026-06-29
Propagate late WebClient Content-Type updates

oleksii-tumanov:
### Description

Propagates `Content-Type` changes made by output stream handlers before request headers are sent. This lets integrations such as Jersey multipart serialization provide the generated boundary without changing framing or unrelated headers.

The change also reaches the actual HTTP/1 request when an HTTP/2 client falls back after TLS negotiation. The direct update and fallback paths are covered by WebClient tests; Checkstyle, copyright, and SpotBugs pass.

### Result

Relevant wire excerpt from the HTTP/1 fallback regression:

```http
POST /stream HTTP/1.1
Content-Type: multipart/form-data; boundary=generated

fallback-output-stream
```

The response status is `200 OK` and the response body is `http1`.

The corresponding Jersey integration and multipart tests are in helidon-io/helidon-microprofile#63.

Related to #12015

### Documentation

None

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/HttpClientRequest.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/HttpClientRequest.java
@@ -16,6 +16,9 @@
 
 package io.helidon.webclient.api;
 
+import java.io.FilterOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
 import java.net.UnixDomainSocketAddress;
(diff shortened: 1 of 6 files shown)


oleksii-tumanov:
Closing this in favor of the connector-local approach in #12155 and helidon-io/helidon-microprofile#63. The multipart boundary is now prepared before headers are mapped, so the fix no longer requires a core WebClient change.

PR #12156, opened 2026-06-29
Fix inherited declarative contract annotations

### LINKED ISSUE #12117: Generate element interceptors for inherited interface annotations
Declarative element interceptors are generated with a `@Service.Named` qualifier derived from the type that owns the annotated element. When the annotation is inherited from a service contract or declarative client interface, the generated interceptor can be qualified for the interface method, while the runtime interceptor lookup uses the effective service method owner.

Example shape:

```java
interface LegacyClient {
    @Ft.Retry
    String get();
}

// Generated service type
final class LegacyClient__DeclarativeClient implements LegacyClient {
    // descriptor method element owner is LegacyClient__DeclarativeClient
}

// Generated FT interceptor today can be named for:
//   com.example.LegacyClient.get()
// Runtime lookup asks for:
//   com.example.LegacyClient__DeclarativeClient.get()
```

Because `InterceptionMetadataImpl` matches element interceptors by the exact generated element signature (`<element enclosing type>.<method signature>`), the interceptor is not applied. A runtime fallback would only fix one lookup path and would leave future declarative features exposed to the same codegen mismatch. The generated interceptor service type and qualifier should instead use the effective service element owner so everything matches through the existing exact lookup.

Known affected declarative element-interceptor features:

- Fault tolerance: `@Ft.Retry`, `@Ft.CircuitBreaker`, `@Ft.Bulkhead`, `@Ft.Timeout`, `@Ft.Fallback`, and `@Ft.Async` handlers generate `Interception.ElementInterceptor` services.
- Metrics: `@Metrics.Counted` and `@Metrics.Timed` generate `Interception.ElementInterceptor` services. `@Metrics.Gauge` uses a registrar path, so it should be checked separately but is not the same invocation-interceptor lookup.
- Tracing: `@Traced` also generates an `Interception.ElementInterceptor` service and should be covered by the same fix/test sweep.

Validation is the useful existing model: `ValidationExtensionProvider` opts into service-contract annotations with `supportsServiceContractAnnotations()`, and the service codegen path already creates effective service elements from inherited contract annotations. The fix should use that path for element-interceptor-producing declarative extensions so inherited interface annotations are retargeted to the concrete/generated service element owner before interceptor generation.

Regression test shape:

- Add a generated declarative REST client or service-contract regression where the annotation exists only on the interface method.
- Invoke through the generated/concrete service and assert the element interceptor really fires.
- For fault tolerance, make the target fail once and then succeed so `@Ft.Retry` is observable only if the client-side interceptor runs.
- For metrics, add `@Metrics.Counted` and/or `@Metrics.Timed` only on the interface method and assert the counter/timer is registered and updated after invocation.
- For tracing, add `@Traced` only on the interface method and assert an element span is created, or at least assert generated metadata/interceptor wiring if tracing test infrastructure cannot observe spans reliably.
- Keep server-side annotations off the target endpoint in REST-client tests so the assertion cannot pass because a server-side interceptor ran instead.

Expected fix direction:

- Make the element-interceptor declarative extensions process annotations inherited from service contracts/interfaces, similar to validation.
- Generate interceptor names and service descriptors for the effective runtime owner, for example `LegacyClient__DeclarativeClient.get()`, not the source interface owner `LegacyClient.get()`, when the generated service method is the element that will be invoked.
- Avoid also generating an interface-owned element interceptor unless that interface method is itself a runtime service element, so the fix does not introduce duplicate interceptors.
- Keep the exact runtime lookup in `InterceptionMetadataImpl`; the codegen should emit the correct type names and qualifiers so the existing validation/effective-element machinery makes the pieces click together.

tomas-langer:
Resolves #12117

Carries forward the 4.x fix from #12122. Inherited annotations on service contract methods are applied to effective service elements for fault tolerance, metrics, tracing, and validation. Fault tolerance, metrics, and tracing exclude factory-provided contracts while validation preserves them; tracing also retains inherited-method precedence and generic contract signature resolution.

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/metrics/MetricsExtension.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/metrics/MetricsExtension.java
@@ -17,7 +17,6 @@
 package io.helidon.declarative.codegen.metrics;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -31,9 +30,11 @@ import io.helidon.codegen.classmodel.ClassModel;
 import io.helidon.codegen.classmodel.ContentBuilder;
 import io.helidon.common.types.Annotated;
 import io.helidon.common.types.Annotation;
+import io.helidon.common.types.ElementKind;
 import io.helidon.common.types.TypeInfo;
 import io.helidon.common.types.TypeName;
 import io.helidon.common.types.TypedElementInfo;
+import io.helidon.declarative.codegen.common.DeclarativeElementInfo;
 import io.helidon.service.codegen.RegistryCodegenContext;
 import io.helidon.service.codegen.RegistryRoundContext;
 import io.helidon.service.codegen.spi.RegistryCodegenExtension;
@@ -158,55 +159,73 @@ class MetricsExtension implements RegistryCodegenExtension {
     }
 
     private void generateCountedInterceptors(RegistryRoundContext roundContext, Map<TypeName, TypeInfo> types) {
-        Collection<TypedElementInfo> elements = roundContext.annotatedElements(ANNOTATION_COUNTED);
         Map<TypeName, AtomicInteger> counters = new HashMap<>();
         types.keySet()
                 .forEach(it -> counters.put(it, new AtomicInteger()));
 
         CountedHandler countedHandler = new CountedHandler(roundContext);
 
-        for (TypedElementInfo element : elements) {
-            TypeName enclosingType = enclosingType(element);
-            int counter = counters.computeIfAbsent(enclosingType, k -> new AtomicInteger())
-                    .getAndIncrement();
-
-            TypeInfo typeInfo = types.get(enclosingType);
-            if (typeInfo == null) {
-                typeInfo = ctx.typeInfo(enclosingType).orElseThrow(() -> new CodegenException("No type info found for type "
-                                                                                                      + enclosingType.fqName()));
+        for (TypeInfo typeInfo : types.values()) {
+            if (typeInfo.kind() == ElementKind.INTERFACE) {
+                continue;
             }
-            checkTypeIsService(roundContext, typeInfo);
-            countedHandler.handle(typeInfo, element, counter);
+            for (TypedElementInfo element : typeInfo.elementInfo()) {
+                if (!DeclarativeElementInfo.belongsToService(typeInfo, element)
+                        || !element.hasAnnotation(ANNOTATION_COUNTED)) {
+                    continue;
+                }
+
+                int counter = counters.computeIfAbsent(typeInfo.typeName(), k -> new AtomicInteger())
+                        .getAndIncrement();
+                checkTypeIsService(roundContext, typeInfo);
+                countedHandler.handle(typeInfo, element, counter);
+            }
         }
     }
 
     private void generateTimedInterceptors(RegistryRoundContext roundContext, Map<TypeName, TypeInfo> types) {
-        Collection<TypedElementInfo> elements = roundContext.annotatedElements(ANNOTATION_TIMED);
         Map<TypeName, AtomicInteger> counters = new HashMap<>();
         types.keySet()
                 .forEach(it -> counters.put(it, new AtomicInteger()));
 
         TimedHandler handler = new TimedHandler(roundContext);
 
-        for (TypedElementInfo element : elements) {
-            TypeName enclosingType = enclosingType(element);
-            int counter = counters.computeIfAbsent(enclosingType, k -> new AtomicInteger())
-                    .getAndIncrement();
-
-            TypeInfo typeInfo = types.get(enclosingType);
-            if (typeInfo == null) {
-                typeInfo = ctx.typeInfo(enclosingType).orElseThrow(() -> new CodegenException("No type info found for type "
-                                                                                                      + enclosingType.fqName()));
+        for (TypeInfo typeInfo : types.values()) {
+            if (typeInfo.kind() == ElementKind.INTERFACE) {
+                continue;
             }
-            checkTypeIsService(roundContext, typeInfo);
-            handler.handle(typeInfo, element, counter);
+            for (TypedElementInfo element : typeInfo.elementInfo()) {
+                if (!DeclarativeElementInfo.belongsToService(typeInfo, element)
+                        || !element.hasAnnotation(ANNOTATION_TIMED)) {
+                    continue;
+                }
+
+                int counter = counters.computeIfAbsent(typeInfo.typeName(), k -> new AtomicInteger())
+                        .getAndIncrement();
+                checkTypeIsService(roundContext, typeInfo);
+                handler.handle(typeInfo, element, counter);
+            }
         }
     }
 
     private void generateGaugeRegistrars(RegistryRoundContext roundContext, Map<TypeName, TypeInfo> types) {
         Map<TypeName, List<Gauge>> gaugeByType = new HashMap<>();
 
-        addGauges(roundContext, gaugeByType);
+        for (TypeInfo typeInfo : types.values()) {
+            if (typeInfo.kind() == ElementKind.INTERFACE) {
+                continue;
+            }
+            for (TypedElementInfo element : typeInfo.elementInfo()) {
+                if (!DeclarativeElementInfo.belongsToService(typeInfo, element)
+                        || !element.hasAnnotation(ANNOTATION_GAUGE)) {
+                    continue;
+                }
+
+                TypeName typeName = typeInfo.typeName();
+                var allGauges = gaugeByType.computeIfAbsent(typeName, k -> new ArrayList<>());
+                processGauge(roundContext, allGauges, typeName, element);
+            }
+        }
 
         GaugeHandler handler = new GaugeHandler(roundContext);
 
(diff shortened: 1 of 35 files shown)


service/codegen/src/main/java/io/helidon/service/codegen/ServiceRegistryCodegenExtension.java
```diff
@@ -344,12 +345,14 @@ private Set<TypeName> supportedServiceContractAnnotations(RoundContext roundCont
                                                                              supportedAnnotationsCache,
                                                                              metaAnnotationsCache,
                                                                              metaAnnotatedCache));
-        result.addAll(supportedAnnotationsOnContracts(roundContext,
-                                                      extension,
-                                                      ServiceTypes.factoryProvidedContracts(serviceContracts, serviceType),
-                                                      supportedAnnotationsCache,
-                                                      metaAnnotationsCache,
-                                                      metaAnnotatedCache));
+        if (extension.supportsFactoryProvidedServiceContractAnnotations()) {
```
tvallin, line 352: `supportsFactoryProvidedServiceContractAnnotations()` cannot be enabled independently. The early return above exits whenever `supportsServiceContractAnnotations()` is false, so a provider overriding only the new method never receives factory-provided contract annotations. Please gate the return on both flags, conditionally collect direct contracts, and test the `false/true` combination.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/tracing/TracingExtension.java
```diff
@@ -52,34 +61,114 @@ class TracingExtension implements RegistryCodegenExtension {
...
+                    .flatMap(contract -> roundContext.typeInfo(contract.type())
+                            .or(() -> roundContext.typeInfo(contract.type().genericTypeName()))
+                            .map(it -> Map.entry(contract, it))
+                            .stream())
+                    .filter(contract -> contract.getValue().hasAnnotation(ANNOTATION_TRACED))
+                    .forEach(contractEntry -> {
+                        ResolvedType resolvedContract = contractEntry.getKey();
+                        TypeInfo contract = contractEntry.getValue();
+                        List<String> typeParameters = contract.typeName().typeParameters();
+                        List<TypeName> typeArguments = resolvedContract.type().typeArguments();
```
tvallin: `resolvedContract` is not necessarily fully resolved through an intermediate generic contract. For `@Traced Base<T>`, `Mid<U> extends Base<U>`, and a service implementing `Mid<String>`, this maps `T` to `U`, so the signature does not match the effective `String` method and tracing is omitted. Please use transitive contract resolution and add a two-level generic inheritance test.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/tracing/TracingExtension.java
```diff
@@ -92,19 +181,105 @@ public void process(RegistryRoundContext roundContext) {
...
+    }
+
+    private void addTracedElement(Map<ElementSignature, TracedElement> map,
+                                  TypedElementInfo element,
+                                  Optional<Annotation> typeAnnotation) {
+        map.compute(element.signature(), (key, existing) -> {
+            if (existing == null) {
+                return new TracedElement(element, typeAnnotation);
+            }
+            if (!existing.element().hasAnnotation(ANNOTATION_TRACED) && element.hasAnnotation(ANNOTATION_TRACED)) {
```
tvallin: If both the implementation and contract have type-level `@Traced`, `existing` can contain the original implementation method while `element` contains merged contract parameter annotations. Neither replacement branch then fires, so an inherited `@Tracing.ParamTag` is discarded. Please retain the effective method metadata while preserving type-annotation precedence.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/metrics/MetricsExtension.java
```diff
@@ -158,55 +159,73 @@ private static List<Tag> toTags(List<Annotation> tags) {
...
+            for (TypedElementInfo element : typeInfo.elementInfo()) {
+                if (!DeclarativeElementInfo.belongsToService(typeInfo, element)
+                        || !element.hasAnnotation(ANNOTATION_COUNTED)) {
+                    continue;
+                }
+
+                int counter = counters.computeIfAbsent(typeInfo.typeName(), k -> new AtomicInteger())
+                        .getAndIncrement();
+                checkTypeIsService(roundContext, typeInfo);
+                countedHandler.handle(typeInfo, element, counter);
```
tvallin, line 180: The merged element can contain both repeatable annotation forms. If the implementation contributes `@Metrics.Tags` and the contract contributes `@Metrics.Counted` plus one `@Metrics.Tag`, `elementTags` selects the single inherited tag and drops the implementation tags. Please normalize and merge `Tag`/`Tags`; timed metrics and gauges are affected as well.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/tracing/TracingExtension.java
```diff
@@ -92,19 +181,105 @@ public void process(RegistryRoundContext roundContext) {
...
 
+    private TracedElements tracedElements(Map<TypeName, TracedElements> tracedElements,
+                                          TypeInfo type) {
+        return tracedElements.computeIfAbsent(type.typeName(),
+                                              key -> new TracedElements(type, new HashMap<>()));
+    }
+
+    private void addTypeTracedMethods(TracedElements tracedElementValue,
+                                      Collection<TypedElementInfo> elements,
+                                      Optional<Annotation> typeAnnotation) {
```
tvallin: Helidon Java style reserves `Optional` for return types. This new `Optional<Annotation>` parameter, and its propagation through the related methods and record, violates that rule. Please model absence using overloads or domain-specific state.

tomas-langer: This is not public API. We can violate such things in private APIs when it makes it easier.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/metrics/MetricsExtensionProvider.java
```diff
@@ -48,6 +48,19 @@ public Set<TypeName> supportedAnnotations() {
                       MetricsTypes.ANNOTATION_GAUGE);
     }
 
+    @Override
+    public boolean supportsServiceContractAnnotations() {
+        return true;
```
tvallin, line 55: Enabling contract processing here still misses split declarations where the implementation owns `@Counted`, `@Timed`, or `@Gauge` and the contract contributes only `@Metrics.Tag`/`@Metrics.Tags`. Those tag types are absent from `supportedAnnotations()`, so no effective type is built and the contract tags are lost. Please register both tag annotations and cover this case.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/tracing/TracingExtension.java
```diff
@@ -52,34 +61,123 @@ class TracingExtension implements RegistryCodegenExtension {
...
-                    .filter(Predicate.not(ElementInfoPredicates::isPrivate))
-                    .forEach(element -> map.put(element.signature(), element));
+            if (type.kind() == ElementKind.INTERFACE) {
+                continue;
+            }
+            TracedElements tracedElementValue = tracedElements(tracedElements, type);
+            TypeInfo effectiveType = effectiveTypes.getOrDefault(type.typeName(), type);
+
+            addTypeTracedMethods(tracedElementValue,
+                                 effectiveType.elementInfo(),
```
tvallin, line 80: `effectiveType` can contain methods owned by a factory-provided type when an ordinary contract triggered effective-element construction. This path does not apply `belongsToService`, so type-level `@Traced` generates interceptors for factory-provided methods despite their explicit exclusion. Please filter these elements by effective ownership.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/tracing/TracingExtension.java
```diff
@@ -52,34 +61,123 @@ class TracingExtension implements RegistryCodegenExtension {
...
+            if (type.kind() == ElementKind.INTERFACE) {
+                continue;
+            }
+
+            Set<ResolvedType> contracts = new HashSet<>();
+            ServiceContracts serviceContracts = ServiceContracts.create(ctx.options(), roundContext::typeInfo, type);
+            serviceContracts.addContracts(contracts, new HashSet<>(), type);
+
+            contracts.stream()
+                    .sorted(Comparator.comparing(contract -> contract.type().resolvedName()))
```
tvallin, line 95: Sorting contracts by FQCN makes tracing precedence lexical. If traced parent and child contracts both redeclare a method, an earlier-named parent can supply the retained type annotation instead of the more-specific child. Please order contracts by hierarchy specificity and handle unrelated ambiguities explicitly.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/tracing/TracingExtension.java
```diff
@@ -52,34 +61,123 @@ class TracingExtension implements RegistryCodegenExtension {
...
+                                            .parameterArguments(parameters)
+                                            .build()
+                                            .signature();
+                                })
+                                .collect(Collectors.toUnmodifiableSet());
+
+                        var contractAnnotation = contract.findAnnotation(ANNOTATION_TRACED);
+                        var serviceMethods = type.elementInfo()
+                                .stream()
+                                .filter(it -> contractMethods.contains(it.signature()))
```
tvallin: This comparison is sensitive to method type-variable names: contract `<T> T echo(T)` and overriding `<U> U echo(U)` are Java-equivalent, but their `ElementSignature`s differ, so type-level contract tracing is omitted. Please normalize method type variables by position or erasure before matching.

declarative/tests/tracing/src/main/java/io/helidon/declarative/tests/tracing/TypedTracingContract.java
```diff
@@ -0,0 +1,38 @@
...
+    @Http.GET
+    @Http.Path("/method")
+    @Tracing.Traced(value = "contract-method",
+                    tags = @Tracing.Tag(key = "source", value = "contract-method"),
+                    kind = Span.Kind.SERVER)
+    String contractMethod();
+
+    @Http.GET
+    @Http.Path("/type-param")
+    String typeContractParam(@Http.QueryParam("id") @Tracing.ParamTag("id") int id);
```
tvallin: The explicit `"id"` masks the default-name case. If the contract declares `@ParamTag int contractId` and the implementation names it `id`, effective-element merging retains `id`, changing the inferred tag key from the annotated contract parameter name. Please preserve the source parameter name and test differing names with the default value.

tvallin (approved):

PR #11952, opened 2026-05-15
Missing JsonObject.Builder methods added

Verdent:
### Description
Fixes #11603

### Documentation

None

PR #12158, opened 2026-06-30
Missing JsonObject.Builder methods added (#11952)

### LINKED ISSUE #12157: 27.x: JsonObject builder is missing some numeric methods
Forward port of the https://github.com/helidon-io/helidon/issues/11603

Verdent:
### Description

Forward port #11952 
Fixes https://github.com/helidon-io/helidon/issues/12157

### Documentation

None

### DIFF
--- a/json/json/src/main/java/io/helidon/json/JsonObject.java
+++ b/json/json/src/main/java/io/helidon/json/JsonObject.java
@@ -17,6 +17,7 @@
 package io.helidon.json;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -222,6 +223,68 @@ public final class JsonObject extends JsonValue {
     }
 
     /**
+     * Return the byte value associated with the specified key as an Optional.
+     *
+     * @param key the key to look up
+     * @return an Optional containing the byte value, or empty if the key is not present
+     */
+    public Optional<Byte> byteValue(String key) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return Optional.empty();
+        }
+        return Optional.of(jsonValue.asNumber().byteValue());
+    }
+
+    /**
+     * Return the byte value associated with the specified key, or the default value if the key is not present.
+     *
+     * @param key the key to look up
+     * @param defaultValue the value to return if the key is not present
+     * @return the byte value associated with the key, or the default value
+     */
+    public byte byteValue(String key, byte defaultValue) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return defaultValue;
+        }
+        return jsonValue.asNumber().byteValue();
+    }
+
+    /**
+     * Return the short value associated with the specified key as an Optional.
+     *
+     * @param key the key to look up
+     * @return an Optional containing the short value, or empty if the key is not present
+     */
+    public Optional<Short> shortValue(String key) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return Optional.empty();
+        }
+        return Optional.of(jsonValue.asNumber().shortValue());
+    }
+
+    /**
+     * Return the short value associated with the specified key, or the default value if the key is not present.
+     *
+     * @param key the key to look up
+     * @param defaultValue the value to return if the key is not present
+     * @return the short value associated with the key, or the default value
+     */
+    public short shortValue(String key, short defaultValue) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return defaultValue;
+        }
+        return jsonValue.asNumber().shortValue();
+    }
+
+    /**
      * Return the integer value associated with the specified key as an Optional.
      *
      * @param key the key to look up
@@ -253,6 +316,68 @@ public final class JsonObject extends JsonValue {
     }
 
     /**
+     * Return the long value associated with the specified key as an Optional.
+     *
+     * @param key the key to look up
+     * @return an Optional containing the long value, or empty if the key is not present
+     */
+    public Optional<Long> longValue(String key) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return Optional.empty();
+        }
+        return Optional.of(jsonValue.asNumber().longValue());
+    }
+
+    /**
+     * Return the long value associated with the specified key, or the default value if the key is not present.
+     *
+     * @param key the key to look up
+     * @param defaultValue the value to return if the key is not present
+     * @return the long value associated with the key, or the default value
+     */
+    public long longValue(String key, long defaultValue) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return defaultValue;
+        }
+        return jsonValue.asNumber().longValue();
+    }
+
+    /**
+     * Return the float value associated with the specified key as an Optional.
+     *
+     * @param key the key to look up
+     * @return an Optional containing the float value, or empty if the key is not present
+     */
+    public Optional<Float> floatValue(String key) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return Optional.empty();
+        }
+        return Optional.of(jsonValue.asNumber().floatValue());
+    }
+
+    /**
+     * Return the float value associated with the specified key, or the default value if the key is not present.
+     *
+     * @param key the key to look up
+     * @param defaultValue the value to return if the key is not present
+     * @return the float value associated with the key, or the default value
+     */
+    public float floatValue(String key, float defaultValue) {
+        ensureResolvedKeys();
+        JsonValue jsonValue = content.get(key);
+        if (jsonValue == null) {
+            return defaultValue;
+        }
+        return jsonValue.asNumber().floatValue();
+    }
+
+    /**
      * Return the double value associated with the specified key as an Optional.
      *
      * @param key the key to look up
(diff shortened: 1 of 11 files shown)


json/json/src/main/java/io/helidon/json/JsonObject.java
```diff
@@ -221,6 +222,68 @@ public String stringValue(String key, String defaultValue) {
         return jsonValue.asString().value();
     }
 
+    /**
+     * Return the byte value associated with the specified key as an Optional.
+     *
+     * @param key the key to look up
+     * @return an Optional containing the byte value, or empty if the key is not present
+     */
+    public Optional<Byte> byteValue(String key) {
```
tomas-langer, line 224: We should check that the `key` is not null.

json/json/src/main/java/io/helidon/json/JsonObject.java
```diff
@@ -283,6 +408,37 @@ public double doubleValue(String key, double defaultValue) {
...
+    }
+
+    /**
+     * Return the BigInteger value associated with the specified key, or the default value if the key is not present.
+     *
+     * @param key the key to look up
+     * @param defaultValue the value to return if the key is not present
+     * @return the BigInteger value associated with the key, or the default value
+     */
+    public BigInteger bigIntegerValue(String key, BigInteger defaultValue) {
```
tomas-langer, line 426: We should check that `key` is not null

tomas-langer (changes requested):

json/json/src/main/java/io/helidon/json/JsonValueParser.java
```diff
@@ -24,8 +24,11 @@
 class JsonValueParser implements JsonParser {
 
     private JsonValue[] values = new JsonValue[100];
+    private boolean[] valuesExpanded = new boolean[100];
```
tomas-langer: The new fixed-size expansion-state sidecars regress every `JsonParser.create(JsonValue)` call, including scalar inputs and `JsonBinding.deserialize(JsonValue, ...)`.

A three-fork base-versus-head JMH measured parser construction increasing from 44.1 ± 7.3 to 78.0 ± 5.6 ns/op and from 512 to 672 B/op. Scalar reads increased from 37.9 ± 3.2 to 79.5 ± 13.2 ns/op with the same 160 B/op allocation increase. The binding path increased from 36.2 ± 2.0 to 60.2 ± 9.6 ns/op and from 416 to 536 B/op. A common-prefix flat traversal also allocated 408 B/op more.

Allocate the expansion state lazily, or use a representation that does not add two fixed sidecar arrays to every parser.

tomas-langer: The lazy-allocation change fixes the original eager sidecar regression on parser creation, scalar reads, and binding. On the current head, those paths are back to base allocation levels (creation 520 vs 512 B/op, scalar read 520 vs 512 B/op, binding 416 vs 416 B/op) and are not slower in the current run.

However, an independent traversal-only JMH confirms that a residual traversal regression is real. I used flat arrays parameterized by element count, 3 forks, 3 x 1 s warmups, 5 x 1 s measurements, and reversed the execution order from the earlier comparison:

| Elements | Base ns/op | Head ns/op | Base B/op | Head B/op | Allocation delta |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 69.1 | 80.6 | 584 | 616 | +32 B/op |
| 8 | 217.0 | 258.4 | 800 | 904 | +104 B/op |
| 24 | 606.6 | 661.9 | 1256 | 1512 | +256 B/op |
| 64 | 1591.1 | 1673.8 | 3033 | 3609 | +576 B/op |

The allocation overhead grows with traversal length, and the 24-element result independently reproduces the prior measurement (1512 vs 1256 B/op). Latency is also slower at every tested size. The fixed per-parser cost is addressed, but the traversal-path regression remains.

Verdent: Addressed in a1347af51e. `JsonArray.values()` creates a defensive `List.copyOf`, and array expansion was invoking it four times. The parser now captures one snapshot and reuses it for sizing, the empty check, iteration, and index calculation. A local allocation probe against the former head and this commit measured 1544 -> 768 B/op for 24 elements and 3640 -> 1904 B/op for 64 elements, removing the size-dependent regression. The full `json/json` test suite (776 tests), copyright/Checkstyle, and module SpotBugs all pass.

tomas-langer (dismissed):

json/json/src/main/java/io/helidon/json/JsonValueParser.java
```diff
@@ -186,12 +144,12 @@ public boolean readBoolean() {
 
     @Override
     public byte readByte() {
-        throw new UnsupportedOperationException();
+        return current.asNumber().byteValue();
```
tomas-langer: AI Review: [P1] Value-backed integer reads silently wrap on overflow

When a parser is created from a `JsonNumber` containing `128`, `readByte()` returns `-128`; the array- and stream-backed parsers throw `JsonException` for the serialized token `128`. The same representation-dependent behavior occurs for `32768` through `readShort()` and `9223372036854775808` through `readLong()`. These new methods delegate to lossy `JsonNumber` conversions, so this PR introduces different `JsonParser` results for the same numeric value depending only on the parser source. Callers validating identifiers, sizes, or bounds can silently receive a wrapped value instead of a parse failure.

json/json/src/main/java/io/helidon/json/JsonValueParser.java
```diff
@@ -298,39 +228,207 @@ public JsonException createException(String message, Exception e) {
...
+            position = snapshot.position;
+            phase = snapshot.phase;
+            if (container.type() == JsonValueType.ARRAY) {
+                arrayValues = container.asArray().values();
+                return;
+            }
+
+            objectIterator = container.asObject().entryIterator();
+            Map.Entry<String, JsonValue> entry = null;
+            for (int i = 0; i < position; i++) {
```
tomas-langer: AI Review: [P1] Object reset cost grows with the marked position

`Frame.restore()` recreates the object iterator and advances it `position` times on every `resetToMark()`. This path is introduced by the frame-based traversal rewrite; the previous PR head restored only the tokens consumed after the mark. A mark/one-token/reset cycle at property 256 measured approximately 524 ns/op and 32 B/op at the current head versus 8.6 ns/op and 0 B/op at the previous head. Repeated lookahead while walking a large object therefore accumulates linear scans and can turn otherwise linear parsing into quadratic work.

json/json/src/test/java/io/helidon/json/JsonContainerTest.java
```diff
@@ -0,0 +1,116 @@
...
+                .set("a", "one")
+                .set("b", "two")
+                .build();
+        JsonObject parsed = JsonParser.create("{\"b\":\"two\",\"a\":\"one\"}").readJsonObject();
+        JsonObject swapped = JsonParser.create("{\"a\":\"two\",\"b\":\"one\"}").readJsonObject();
+        JsonObject extra = JsonParser.create("{\"a\":\"one\",\"b\":\"two\",\"c\":\"three\"}").readJsonObject();
+
+        assertThat(constructed.equals(parsed), is(true));
+        assertThat(parsed.equals(constructed), is(true));
+        assertThat(constructed.hashCode(), is(parsed.hashCode()));
```
tomas-langer, line 47: AI Review: [P2] Equality failures omit the compared JSON containers

These assertions evaluate `equals()` before passing the value to Hamcrest, so a regression reports only `false` versus `true`. The actual `JsonObject` instances, including the key/value or ordering difference under test, are absent from the diagnostics. Because this commit changes container equality behavior, failures in the new regression coverage will not identify which compared values diverged, making CI failures substantially harder to diagnose.

json/json/src/main/java/io/helidon/json/JsonValueParser.java
```diff
@@ -186,22 +145,38 @@ public boolean readBoolean() {
 
     @Override
     public byte readByte() {
-        throw new UnsupportedOperationException();
+        try {
+            return readBigInteger().byteValueExact();
```
tomas-langer: The primitive readers now call `readBigInteger()` before they perform the exact primitive range check. For a compact JSON number such as `1e100000000`, `readBigInteger()` converts the `BigDecimal` to a `BigInteger`, so the parser expands a tiny input into a very large integer before `byteValueExact()` can reject it. The public `JsonParser.create(JsonValue).readByte()` path can therefore spend seconds and exhaust the heap before returning the expected JSON numeric-range error. This makes hostile compact exponent values a CPU/heap denial-of-service vector in the JsonValue replay parser.

tomas-langer (dismissed):
One blocking issue found in the JsonValue primitive numeric replay path.

json/json/src/test/java/io/helidon/json/JsonValueParserTest.java
```diff
@@ -488,4 +647,19 @@ private int unscaledValueCalls() {
         }
     }
 
+    private static void assertNestedEmptyContainerSkipped(JsonValue emptyContainer, byte start, byte end) {
```
tomas-langer: The new `TrackingBigDecimal` nested helper is declared before `assertNestedEmptyContainerSkipped(...)`, so this test class now has an ordinary helper method after a nested type. That order is inconsistent with the local Helidon test layout, where nested support types are kept after the test methods and method helpers. It does not change test behavior, but it makes the class structure harder to scan.

json/json/src/main/java/io/helidon/json/JsonValueParser.java
Verdent: Addressed in f51bd6bf9f. The integral readers now check exclusive truncation bounds before converting to long, avoiding BigInteger materialization for compact extreme exponents while preserving fractional truncation. Guarded regressions cover positive and negative 100000000 exponents and primitive boundary fractions. The full json/json suite (800 tests), copyright/Checkstyle, and module SpotBugs pass.

json/json/src/test/java/io/helidon/json/JsonValueParserTest.java
Verdent: Addressed in f51bd6bf9f. The assertNestedEmptyContainerSkipped helper now precedes the nested support types.

tomas-langer (approved):

PR #12159, opened 2026-06-30
27.x: Fix JSON stream parser buffer compaction

### LINKED ISSUE #12067: 27.x: Default buffer size reading JSON cannot be configured, resulting in error with bigger JSONs
Forward port of https://github.com/helidon-io/helidon/issues/12062

Verdent:
### Description
Forward port of https://github.com/helidon-io/helidon/issues/12062
Fixes https://github.com/helidon-io/helidon/issues/12067

### Documentation

None

### DIFF
--- a/json/json/src/main/java/io/helidon/json/JsonNumber.java
+++ b/json/json/src/main/java/io/helidon/json/JsonNumber.java
@@ -18,6 +18,7 @@ package io.helidon.json;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Objects;
 
 import io.helidon.common.Api;
 
@@ -30,7 +31,7 @@ public final class JsonNumber extends JsonValue {
     private final byte[] buffer;
     private final int start;
     private final int length;
-    private final byte jsonStartChar;
+    private byte jsonStartChar;
     private BigDecimal bigDecimalValue;
 
     private JsonNumber(byte[] buffer, int start, int length) {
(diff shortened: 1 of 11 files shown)


json/json/src/main/java/io/helidon/json/JsonNumber.java
```diff
@@ -80,16 +89,42 @@ static JsonNumber create(byte[] buffer, int start, int length) {
...
+        int scale = value.scale();
+        int adjustedExponent = precision - scale - 1;
+        if (scale >= precision && adjustedExponent >= -6) {
+            return '0';
+        }
+        BigInteger unscaledValue = value.unscaledValue();
+        if (unscaledValue.bitLength() < Long.SIZE) {
+            return firstDigit(unscaledValue.longValue());
+        }
+        return (byte) unscaledValue.toString().charAt(0);
```
tomas-langer, line 117: Constructing a positive `JsonNumber` with an unscaled value larger than 63 bits now converts the entire `BigInteger` to a decimal `String` just to read one character. This makes `JsonNumber.create(BigDecimal)` O(d) in CPU and temporary allocation even when `jsonStartChar()` is never used; untrusted Smile big integers reach this path through `SmileInputStreamParser`.

 Defer token derivation until it is needed, or derive/carry the leading token without materializing and discarding the full decimal representation. Because this port inherits the same code from #12064, apply the correction consistently to 4.x and main.

tomas-langer (changes requested):
Requesting changes for the inline JsonNumber allocation and CPU regression.

tomas-langer (approved):

PR #12161, opened 2026-07-02
JSON Binding list helper methods

### LINKED ISSUE #11685: JSON Binding helper methods for list handling
Deserialization of an array into a list of values.

Current (for all types of sources - bytes, input stream, string):
```java
List<Item> items = BINDING.deserialize(bytes, new GenericType<List<Item>>() { });
```

The handling of `GenericType` is not very user-friendly, I think the following signature would be easier to handle:

```java
List<Item> items = BINDING.deserializeList(bytes, Item.class);
```

We need to keep the original method, if there are nested generics, but for this simple (and quite often used) case, we would benefit from the shortcut method(s).

Similar when serializing:

Current:
```java
BINDING.serialize(items, new GenericType<List<Item>>() { });
```

Desired:
```java
BINDING.serialize(items, Item.class);
```

If the new methods would conflict with existing APIs, we could also use `serializeList` method name (maybe this would be better either way, to make this explicit).

As the root of JSON can be either an object or an array (types that make sense for JSON binding at least), this is the only additional method that makes sense.

Verdent:
### Description
Fixes #11685

### Documentation

None

### DIFF
--- a/json/binding/src/main/java/io/helidon/json/binding/JsonBinding.java
+++ b/json/binding/src/main/java/io/helidon/json/binding/JsonBinding.java
@@ -156,6 +156,19 @@ public interface JsonBinding extends RuntimeType.Api<JsonBindingConfig> {
     <T> byte[] serializeToBytes(T obj, Class<? super T> type);
 
     /**
+     * Serializes a list of objects of a specific item type to JSON bytes encoded using UTF-8.
+     * If the provided list is null, returns the bytes of the string {@code null}.
+     *
+     * @param values   the list to serialize, this parameter may be {@code null}
+     * @param itemType the class type of each list item
+     * @param <T>      the type of each list item
+     * @return         the JSON bytes encoded using UTF-8
+     */
+    default <T> byte[] serializeListToBytes(List<T> values, Class<T> itemType) {
+        return serializeToBytes(values, listType(itemType));
+    }
+
+    /**
      * Serializes an object of a generic type to JSON bytes encoded using UTF-8.
      * If the provided object is null, returns the bytes of the string {@code null}.
      *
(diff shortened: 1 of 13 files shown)


json/binding/src/main/java/io/helidon/json/binding/JsonBinding.java
```diff
@@ -255,6 +326,18 @@ static JsonBinding create(Consumer<JsonBindingConfig.Builder> consumer) {
      */
     <T> T deserialize(byte[] bytes, Class<T> type);
 
+    /**
+     * Deserializes a JSON array from a byte array to a list of objects of the specified item type.
+     *
+     * @param bytes    the JSON data as bytes
+     * @param itemType the class type of each list item
+     * @param <T>      the type of each list item
+     * @return         the deserialized list
```
tomas-langer: A JSON null value is deserialized as Java null through the default JsonDeserializer.deserializeNull() path, but this return contract is not documented here. The existing deserialize overloads have the same undocumented behavior. Update the return Javadocs for all JsonBinding deserialize and deserializeList methods consistently to state that the result may be null when the input represents JSON null.

json/binding/src/main/java/io/helidon/json/binding/JsonBinding.java
```diff
@@ -387,4 +543,11 @@ static JsonBinding create(Consumer<JsonBindingConfig.Builder> consumer) {
      */
     <T> T deserialize(JsonParser parser, GenericType<T> type);
 
+    private static <T> GenericType<List<T>> listType(Class<? super T> itemType) {
+        return GenericType.<List<T>>builder()
```
tomas-langer: Every serializeList and deserializeList call builds a new GenericType descriptor, including its builder and generic-parameter structure. These helpers sit directly on the serialization hot path, and callers cannot reuse that generated descriptor. Reuse or cache the descriptor per itemType, or use an equivalent allocation-free strategy, and add focused performance coverage for repeated helper calls.

json/binding/src/main/java/io/helidon/json/binding/JsonBinding.java
```diff
@@ -110,6 +111,20 @@ static JsonBinding create(Consumer<JsonBindingConfig.Builder> consumer) {
...
+     * Serializes a list of objects of a specific item type to a JSON string.
+     * If the provided list is null, returns the string {@code null}.
+     * The item type determines the serializer used for each list item and may be a supertype of {@code T}.
+     *
+     * @param values   the list to serialize, this parameter may be {@code null}
+     * @param itemType the type to use for serializing each list item
+     * @param <T>      the type of each list item
+     * @return         the JSON string representation
+     */
+    default <T> String serializeList(List<T> values, Class<? super T> itemType) {
```
tomas-langer, line 124: This new user-facing convenience API is not covered by the JSON Binding guide or its snippets, so users have no discoverable example of the intended list workflow. Add an example using serializeList and deserializeList to docs/src/main/asciidoc/se/json/json.adoc and the corresponding JsonSnippets source.

tomas-langer (changes requested):
Address the three unresolved inline findings.

tomas-langer (approved):

PR #12166, opened 2026-07-02
Avoid stale HTTP/2 flow control waits

tomas-langer:
Fixes an HTTP/2 outbound flow-control race that can discard a concurrent window update and impose an unnecessary backoff. The current window is now rechecked before blocking and before treating a wait as timed out.

This was exposed by the gRPC backpressure tests in #12124.


http/http2/src/main/java/io/helidon/http/http2/WindowSizeImpl.java
```diff
@@ -170,18 +170,22 @@ public void blockTillUpdate() {
...
+                        backoff = Math.min(backoff * 2, BACKOFF_MAX);
+                    }
                 } catch (InterruptedException e) {
                     debugLog("%s OFC STR %d: Window depleted, waiting for update interrupted.", e);
                     throw new Http2Exception(Http2ErrorCode.FLOW_CONTROL, "Flow control update wait interrupted.");
                 }
-                if (System.currentTimeMillis() - startTime > timeoutMillis) {
-                    debugLog("%s OFC STR %d: Window depleted, waiting for update time-out.", null);
-                    throw new Http2Exception(Http2ErrorCode.FLOW_CONTROL, "Flow control update wait time-out.");
+                if (getRemainingWindowSize() < 1) {
```
Verdent: The timeout check now runs only while the window remains depleted. A window update arriving after the configured deadline during the current semaphore wait is therefore accepted. With a 100 ms timeout and an update after 120 ms, I reproduced three successful returns at about 122 ms; the previous implementation raised the flow-control timeout. Depending on the backoff interval, writes could resume nearly 5 seconds beyond some configured deadlines, contrary to the documented total update-wait timeout.

http/http2/src/main/java/io/helidon/http/http2/WindowSizeImpl.java
```diff
@@ -165,23 +165,32 @@ public void triggerUpdate() {
...
+                                                           TimeUnit.NANOSECONDS);
+                        // linear deterministic backoff
+                        backoff = Math.min(backoff * 2, BACKOFF_MAX);
+                    }
                 } catch (InterruptedException e) {
                     debugLog("%s OFC STR %d: Window depleted, waiting for update interrupted.", e);
                     throw new Http2Exception(Http2ErrorCode.FLOW_CONTROL, "Flow control update wait interrupted.");
                 }
-                if (System.currentTimeMillis() - startTime > timeoutMillis) {
+                if (System.nanoTime() - startTime >= timeoutNanos) {
```
tvallin: The connection window is shared by every stream, but `triggerUpdate()` releases only one permit. With two writers in the final acquire, an on-time `WINDOW_UPDATE` wakes one; the other remains parked until this check and throws although the window is already positive. I reproduced this with two virtual threads: one returned while the other timed out with 1024 bytes available. Please use broadcast/update-generation semantics or equivalent, and add a two-waiter regression.

http/http2/src/main/java/io/helidon/http/http2/WindowSizeImpl.java
```diff
@@ -153,35 +155,63 @@ public void resetWindowSize(int size) {
...
-            return n;
+            return super.decrementWindowSize(decrement);
         }
 
         @Override
         public void triggerUpdate() {
-            updatedSemaphore.release();
+            updateLock.lock();
+            try {
+                updated.signalAll();
```
Verdent, line 175: triggerUpdate now broadcasts to every waiter. FlowControlImpl.incrementStreamWindowSize also calls the shared connection window trigger for a stream-only WINDOW_UPDATE, although connection credit is unchanged. With the connection window at zero and 16 stream writers blocked on it, one stream update woke all 16; each contended for this lock, rechecked zero, and parked again. Repeating valid stream updates across N blocked streams therefore creates peer-controlled O(N^2) wake, recheck, and debug-log work without making connection progress.

http/http2/src/test/java/io/helidon/http/http2/WindowSizeTest.java
```diff
@@ -0,0 +1,207 @@
...
+            while ((blockers[0].getState() != Thread.State.TIMED_WAITING
+                    || blockers[1].getState() != Thread.State.TIMED_WAITING)
+                    && System.nanoTime() < waitDeadline) {
+                Thread.onSpinWait();
+            }
+            assertThat("first flow-control wait must start", blockers[0].getState(), is(Thread.State.TIMED_WAITING));
+            assertThat("second flow-control wait must start", blockers[1].getState(), is(Thread.State.TIMED_WAITING));
+
+            Thread.sleep(Duration.ofMillis(1700));
+            outbound.incrementWindowSize(1024);
```
Verdent, line 90: This test waits until both threads are TIMED_WAITING and then sleeps another 1.7 seconds before incrementing; the other cases update only after the timeout. None deterministically exercises the race described by this PR: an increment after the depleted-window check but before the waiter commits to await. The earlier transition test was removed, so a lost-notification regression in that window can pass this suite.

tests/benchmark/jmh/src/test/java/io/helidon/webserver/benchmark/jmh/Http2FlowControlJmhRunnerTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+import org.openjdk.jmh.runner.options.TimeValue;
+
+class Http2FlowControlJmhRunnerTest {
+
+    @Test
+    void runBenchmarks() throws RunnerException {
+        Options options = new OptionsBuilder()
+                .include(".*Http2FlowControlJmhTest.*")
```
Verdent: Http2FlowControlJmhTest already matches the existing JunitJmhRunnerTest include pattern .*JmhTest under the jmh profile. This runner therefore executes all four methods again, with annotations honored here versus the existing runner forcing eight threads, and only the existing run participates in baseline comparison. The dedicated invocation took 40.456 seconds locally, so a normal jmh-profile run produces two differently configured result sets and pays that extra run time.

tomas-langer:
Trying to fix additional discovered issues with http/2....

tvallin (approved):

PR #12171, opened 2026-07-10
4.x: Improve HTTP/1 WebClient throughput with direct bulk response reads

### LINKED ISSUE #12170: Improve HTTP/1 WebClient throughput with direct bulk response reads
Helidon Version: 4.4.1 and current helidon-4.x

## Enhancement Description

Improve `Http1ClientResponse.inputStream()` throughput for large fixed-length response entities by allowing the HTTP/1 client to read socket data directly into the caller-provided byte array.

### Environment

- Helidon: 4.4.1 and current `helidon-4.x`
- API: Helidon SE `Http1Client`
- JDK: 25
- OS: macOS on Apple Silicon

### Problem

Large HTTP/1 response bodies currently pass through an intermediate `BufferData`/entity buffering path before `InputStream.read(byte[], int, int)` copies the bytes into the caller's buffer. This extra staging and copying limits single-stream throughput for large payloads, even when the caller supplies a large destination buffer and the source is local or otherwise capable of substantially higher throughput.

The overhead is particularly visible for fixed-length binary responses. In a loopback benchmark that repeatedly reads a validated 128 MiB range from a hot file-backed response (4 GiB total per measurement), the existing Helidon HTTP/1 client path measured approximately 1.16-1.32 GiB/s. A prototype direct bulk-read path measured approximately 2.19 GiB/s on the same workload.

### Proposed Enhancement

Allow the HTTP/1 client response stream to bypass intermediate entity-buffer copies when all of the following apply:

- The caller uses `InputStream.read(byte[], int, int)`.
- The response has a known `Content-Length`.
- Any bytes already buffered while parsing the response have been consumed first.
- The underlying socket supports a direct byte-array read.

The prototype:

- Allows an HTTP client implementation to provide an optimized response `InputStream` with lifecycle behavior equivalent to the existing entity-backed stream.
- Adds an optional direct bulk-read callback to `DataReader`, while retaining the existing buffered-supplier fallback.
- Drains already-buffered parser bytes before reading directly from the socket.
- Limits each direct read to the remaining `Content-Length`, preventing consumption of bytes belonging to the next response on a persistent connection.
- Completes and releases a reusable response immediately after the final byte is consumed.
- Closes a non-reusable connection when the response stream is closed early.
- Avoids using encoded wire content length to determine completion of a decoded response stream.
- Leaves chunked, unknown-length, Unix-domain/NIO, and supplier-only paths on their existing behavior.

### Compatibility

The enhancement preserves the `InputStream` contract and existing response lifecycle semantics. Direct bulk reads are optional; transports that do not provide them continue using the current `DataReader` supplier path.

### Validation

The prototype includes coverage for:

- Direct reads into the caller's byte array.
- Draining parser-buffered bytes before direct socket reads.
- Enforcing exact content length without consuming the next response.
- Zero-length responses and supplier EOF.
- Encoded responses.
- Early stream closure versus fully consumed connection reuse.
- Idempotent completion and close behavior.
- Delegate close failures and checked `IOException` behavior.
- A real HTTP/1 persistent-connection reuse path.

A pull request implementing and testing this enhancement is ready for submission.

levinson:
### Description

Fixes #12170.

Improves HTTP/1 WebClient throughput for large fixed-length response entities by eliminating intermediate response-body copies.

`Http1ClientResponse.inputStream()` now provides a protocol-specific stream that bypasses intermediate entity staging while preserving response lifecycle behavior. Fully consumed responses release reusable connections, while early stream closure closes the connection. Encoded responses do not use the encoded wire `Content-Length` to determine decoded-stream completion.

For fixed-length entities, `DataReader` can optionally read socket data directly into caller-provided buffers after draining any bytes buffered during response parsing. Reads remain limited to the remaining `Content-Length`, preventing consumption of data belonging to the next response on a persistent connection.

The existing supplier path remains available for transports without direct bulk reads. Chunked, unknown-length, Unix-domain, and NIO response paths retain their existing behavior.

The change includes coverage for:

- Direct reads into caller-provided buffers
- Parser-buffer ordering
- Exact content-length enforcement
- Persistent connection reuse
- Early stream closure
- Encoded and zero-length responses
- Supplier EOF
- Checked `IOException` and close-failure behavior

In a loopback benchmark reading a validated 128 MiB file-backed range repeatedly, the existing HTTP/1 client path measured approximately 1.16-1.32 GiB/s. The direct bulk-read path measured approximately 2.19 GiB/s on the same workload.

### Helidon 27 follow-up

The same intermediate-copy path exists on `main`. Because the HTTP/1 response lifecycle has diverged, support for Helidon 27 would be submitted as a separate adapted PR after this approach is reviewed.

### Documentation

No user-guide changes are required. The `HttpClientResponse.inputStream()` Javadoc is updated to permit protocol-specific direct streams with equivalent response lifecycle behavior.

### Testing

- `mvn -pl webclient/http1 -am test`
- `mvn validate -Pcheckstyle`
- `mvn validate -Pcopyright`

### DIFF
--- a/common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
@@ -21,6 +21,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Supplier;
 
 /**
@@ -28,6 +29,7 @@ import java.util.function.Supplier;
  */
 public class DataReader {
     private final Supplier<byte[]> bytesSupplier;
+    private final ByteArrayReader byteArrayReader;
     private final boolean ignoreLoneEol;
     private Node head;
     private Node tail;
@@ -42,11 +44,7 @@ public class DataReader {
      */
     @Deprecated(forRemoval = true, since = "4.4.0")
     public DataReader(Supplier<byte[]> bytesSupplier) {
-        this.ignoreLoneEol = false;
-        this.bytesSupplier = bytesSupplier;
-        // we cannot block until data is actually ready to be consumed
-        this.head = new Node(BufferData.EMPTY_BYTES);
-        this.tail = this.head;
+        this(bytesSupplier, null, false);
     }
 
     /**
@@ -58,8 +56,15 @@ public class DataReader {
      */
     @Deprecated(forRemoval = true, since = "4.4.0")
     public DataReader(Supplier<byte[]> bytesSupplier, boolean ignoreLoneEol) {
+        this(bytesSupplier, null, ignoreLoneEol);
+    }
+
+    private DataReader(Supplier<byte[]> bytesSupplier,
+                       ByteArrayReader byteArrayReader,
+                       boolean ignoreLoneEol) {
         this.ignoreLoneEol = ignoreLoneEol;
-        this.bytesSupplier = bytesSupplier;
+        this.bytesSupplier = Objects.requireNonNull(bytesSupplier);
+        this.byteArrayReader = byteArrayReader;
         // we cannot block until data is actually ready to be consumed
         this.head = new Node(BufferData.EMPTY_BYTES);
         this.tail = this.head;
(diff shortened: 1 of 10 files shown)


tomas-langer (changes requested):
Hello, thanks for the contribution. I understand what you want to solve with this PR, nevertheless I would prefer to leave low level performance changes to the maintainers.

There are some conceptual issues with the approach you have chosen, that do not align with my design of this area (as this design is not public, you cannot know about it).

There are two options for this PR - either I can take over your work, and refactor it according to what I want - this will keep you as contributor, and it would keep some of the changes you have done. The other option would be to close this PR, and I will (eventually) fix the referenced issue. 

Please let me know which approach you prefer.

For Helidon team: please do not merge this PR.

levinson:
@tomas-langer Ultimately I just want to see the performance improved with HTTP/1 webclient. I'm happy to let you take over the work with whatever option you choose. I have a benchmark test in another project that can easily reproduce the issue and happy to help with validation of the fix.

PR #12182, opened 2026-07-17
Fix HTTP/2 duplex flow-control liveness

tomas-langer:
### Summary

- prevent inbound HTTP/2 `WINDOW_UPDATE` callbacks from blocking behind an unrelated transport write
- keep the uncontended write path synchronous and coalesce only contended flow-control credit
- preserve queued update ordering before `RST_STREAM` and close the owning connection if a deferred write fails
- keep the change behind the existing writer contract, with no public API additions

### Background

The gRPC failure observed in #12123 exposed a pre-existing duplex liveness cycle. A request writer could hold the HTTP/2 serialization lock while blocked in the transport, while the response reader needed that lock to return flow-control credit. The blocked reader then stopped the server from consuming more request data, so neither side could make progress.

The handoff is specific to contended `WINDOW_UPDATE` frames. Normal updates still use the existing synchronous write path. Contended increments are coalesced per stream and drained through the same serialized writer, independently of the configured connection executor.

### Performance

A three-fork bidirectional gRPC JMH comparison covered payloads of 65,530, 65,531, and 131,072 bytes with GC profiling. Throughput deltas were -0.60%, -0.47%, and +5.20%; normalized allocation deltas were +292, +438, and +278 bytes per operation. All throughput and allocation confidence intervals overlapped the `main` baseline.

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
@@ -16,6 +16,7 @@
 
 package io.helidon.http.http2;
 
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -35,6 +36,7 @@ import io.helidon.common.socket.SocketContext;
  */
 public class Http2ConnectionWriter implements Http2StreamWriter {
     private static final Runnable NO_OP = () -> { };
+    private static final int NO_RESET_STREAM = -1;
 
     private final DataWriter writer;
 
@@ -48,6 +50,9 @@ public class Http2ConnectionWriter implements Http2StreamWriter {
     private final Http2Headers.DynamicTable outboundDynamicTable;
     private final Http2HuffmanEncoder responseHuffman;
     private final BufferData headerBuffer = BufferData.growing(512);
+    // Guarded by windowUpdateLock. The generation prevents an older reset from clearing a newer reset's fence.
+    private int resetStreamId = NO_RESET_STREAM;
+    private long resetGeneration;
 
     /**
      * A new writer.
@@ -77,14 +82,17 @@ public class Http2ConnectionWriter implements Http2StreamWriter {
         // The common path stays synchronous. Only hand off when another frame is holding the serialization lock
         // during a blocking transport write, otherwise an inbound flow-control callback could stop duplex progress.
         boolean locked = false;
+        int streamId = frame.header().streamId();
         try {
             windowUpdateLock.lock();
             try {
                 throwIfWindowUpdateFailed();
+                if (streamId != 0 && streamId == resetStreamId) {
+                    return;
+                }
                 locked = streamLock.tryLock(0, TimeUnit.NANOSECONDS);
                 if (!locked) {
                     Http2WindowUpdate windowUpdate = Http2WindowUpdate.create(frame.data());
-                    int streamId = frame.header().streamId();
                     pendingWindowUpdates.merge(streamId, (long) windowUpdate.windowSizeIncrement(), Long::sum);
                 }
             } finally {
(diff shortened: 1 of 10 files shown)


http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
```diff
@@ -246,12 +297,145 @@ public void updateHeaderTableSize(long newSize) throws InterruptedException {
     private void lockedWrite(Http2FrameData frame) {
         lock();
         try {
+            throwIfWindowUpdateFailed();
+            if (frame.header().type() == Http2FrameType.RST_STREAM) {
+                Http2FrameData windowUpdate;
+                while ((windowUpdate = pollWindowUpdate(frame.header().streamId())) != null) {
```
danielkec: An update can be queued after this loop observes no more entries but before the reset frame is written. The scheduled drain then sends that same-stream `WINDOW_UPDATE` after `RST_STREAM`; a deterministic barrier probe on this head observed `[RST_STREAM, WINDOW_UPDATE]`. RFC 9113 sections 5.1 and 6.4 make the reset terminate the stream, so this produces closed-stream traffic and can trigger peer error handling.

http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
```diff
@@ -246,12 +297,145 @@ public void updateHeaderTableSize(long newSize) throws InterruptedException {
...
+                    noLockWrite(windowUpdate);
+                }
+            }
             noLockWrite(frame);
+        } catch (Throwable t) {
+            if (frame.header().type() == Http2FrameType.RST_STREAM) {
+                failWindowUpdates(t);
```
danielkec: Failures from flushing a pending update on this reset path are recorded, but the owning writer is not closed; the `write(Http2FrameData)` fast path has the same behavior. A deterministic probe made this flush throw the production `SocketWriterException` and observed `closeCalls=0`. `Http2ClientStream.reset()` catches that exception and completes the reset, leaving the connection open after its transport write has failed.

webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionTest.java
```diff
@@ -450,6 +452,70 @@ void closeNowDoesNotWaitForConnectionWriter() throws Exception {
...
+            }
+        }
+    }
+
+    @Test
+    void deferredWindowUpdateFailureClosesConnection() throws Exception {
+        try (MockedConnectionTestContext test = new MockedConnectionTestContext()) {
+            test.offerInbound(settingsFrame(1));
+            Http2ClientConnection connection = test.createConnection(false);
+            MockedConnectionTestContext.BlockedWrite blockedWrite = test.blockNextWriteNow();
```
danielkec, line 503: This can race the initial SETTINGS ACK: `handleSettingsFrame()` counts down `initialSettingsLatch` before writing the ACK, so `createConnection(false)` can return and this blocker can be consumed by the ACK instead of `activeWrite`. Once `failWrites()` is enabled, `activeWrite` fails directly and the test no longer exercises deferred failure. This reproduced within 20 single-CPU runs.

http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
tomas-langer: Fixed by fencing a resetting stream while holding the window-update lock, draining its pending WINDOW_UPDATE frames before RST_STREAM, and dropping same-stream updates while the reset fence is active. Added deterministic ordering coverage.

http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
tomas-langer: Fixed by centralizing terminal window-update failure handling in Http2ConnectionWriter. The first failure now clears pending updates, closes the underlying writer exactly once, preserves the original failure, and suppresses any close failure. Added coverage for the synchronous fast path and reset-time pending-update flush.

webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionTest.java
tomas-langer: Fixed by waiting until the preface and initial SETTINGS ACK writeNow calls have completed before arming the next-write blocker. This ensures the DATA write consumes the blocker and the deferred WINDOW_UPDATE encounters the injected failure. The test also passes with one active processor.

danielkec (approved):

PR #12183, opened 2026-07-20
Add stuck thread detection

### LINKED ISSUE #5727: Stuck Thread Detection
Currently, Helidon does not have a module/feature for stuck thread detection.

WLS has it, and many customers find this to be an important feature. We should consider implementing this support.

tomas-langer:
Resolves #5727

Adds an opt-in WebServer feature that reports request threads which remain in HTTP routing and business logic beyond a configured threshold.

The feature:

- monitors HTTP/1 request processing and HTTP/2 stream request processing using an independent filter;
- reports a request once with request and thread diagnostics, then logs recovery when the handler returns;
- remains disabled unless configured under `server.features.stuck-thread-detection` or added programmatically; and
- excludes idle connections, post-routing transport cleanup, upgraded connection lifetime, and HTTP/2 subprotocols such as gRPC.

### DIFF
--- a/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -133,6 +133,7 @@ class UriPathNoParam implements UriPath {
             rawPath = rawPath.substring(1);
         }
 
-        return URI.create(rawPath).normalize().getPath();
+        String path = URI.create(rawPath).normalize().getPath();
+        return path == null ? "" : path;
     }
 }
--- a/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
@@ -117,15 +117,15 @@ class UriPathNoParam implements UriPath {
 
          we use decoded and normalized path to match routing, so we need to resolve all of that
          */
-        if (rawPath.indexOf('/') == -1 && rawPath.indexOf(':') > 0) {
-            return "";
-        }
-
         int percent = rawPath.indexOf('%');
         int dot = rawPath.indexOf(".");
         int doubleSlash = rawPath.indexOf("//");
 
-        if (!validate && percent == -1 && doubleSlash == -1 && dot == -1) {
+        if (!validate
+                && percent == -1
+                && doubleSlash == -1
+                && dot == -1
+                && (rawPath.isEmpty() || rawPath.charAt(0) == '/' || rawPath.indexOf(':') == -1)) {
             return rawPath;
         }
 
--- a/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
+++ b/common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
@@ -117,6 +117,10 @@ class UriPathNoParam implements UriPath {
 
          we use decoded and normalized path to match routing, so we need to resolve all of that
          */
+        if (rawPath.indexOf('/') == -1 && rawPath.indexOf(':') > 0) {
+            return "";
+        }
+
         int percent = rawPath.indexOf('%');
         int dot = rawPath.indexOf(".");
         int doubleSlash = rawPath.indexOf("//");
--- a/webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
@@ -120,7 +120,7 @@ final class StuckThreadDetectionFilter implements Filter {
                                        System.nanoTime(),
                                        prologue.rawProtocol(),
                                        prologue.method().text(),
-                                       LogFormatter.escape(prologue.uriPath().rawPathNoParams()),
+                                       LogFormatter.escape(prologue.uriPath().path()),
                                        req.id(),
                                        req.serverSocketId(),
                                        req.socketId());
--- a/webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
@@ -1,0 +1,358 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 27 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
```diff
@@ -0,0 +1,358 @@
...
+            return;
+        }
+
+        Thread thread = Thread.currentThread();
+        var prologue = req.prologue();
+        var active = new ActiveRequest(thread,
+                                       System.nanoTime(),
+                                       prologue.rawProtocol(),
+                                       prologue.method().text(),
+                                       LogFormatter.escape(prologue.uriPath().rawPathNoParams()),
```
danielkec: Absolute-form request targets can leak URI userinfo into the new stuck and recovery logs. For example, send:

```http
GET http://alice:s3cr3t@localhost/slow?redacted=yes HTTP/1.1
Host: localhost
```

to a `/slow` handler that exceeds the threshold. Routing uses `/slow`, but `rawPathNoParams()` here is `http://alice:s3cr3t@localhost/slow`, so both log records include `alice:s3cr3t`; `LogFormatter.escape` does not redact it.

webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
```diff
@@ -0,0 +1,358 @@
...
+
+    StuckThreadDetectionFilter(StuckThreadDetectionConfig config, String socketName) {
+        this.checkPeriodNanos = config.checkPeriod().toNanos();
+        this.thresholdNanos = config.threshold().toNanos();
+        this.socketName = socketName;
+    }
+
+    @Override
+    public void afterStart(WebServer webServer) {
+        Thread thread = Thread.ofVirtual()
```
danielkec, line 71: This monitor uses the default virtual-thread scheduler, as do the request threads it is intended to diagnose, so CPU-bound requests can starve the detector itself. A deterministic reproducer is to start with `-Djdk.virtualThreadScheduler.parallelism=1 -Djdk.virtualThreadScheduler.maxPoolSize=1` and run a handler that busy-spins past the threshold: it occupies the only carrier, this monitor does not run, and no warning is emitted until the handler releases the carrier. The same blind spot occurs at normal parallelism when non-yielding requests occupy all carriers.

tomas-langer: Confirmed, this can occur when non-yielding request virtual threads occupy every scheduler carrier. The detector is intended as a best-effort elapsed-time diagnostic; full carrier and CPU saturation is a separate operational condition handled through CPU and thread profiling. We accept that the monitor can be delayed or miss requests under that condition and do not want to dedicate a platform thread per socket.

webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
```diff
@@ -0,0 +1,358 @@
...
+            return;
+        }
+
+        Thread thread = Thread.currentThread();
+        var prologue = req.prologue();
+        var active = new ActiveRequest(thread,
+                                       System.nanoTime(),
+                                       prologue.rawProtocol(),
+                                       prologue.method().text(),
+                                       LogFormatter.escape(prologue.uriPath().path()),
```
danielkec: Enabling this feature breaks valid HTTP/1 CONNECT authority-form requests before routing. For `CONNECT example.com:443 HTTP/1.1`, `prologue.uriPath().path()` is `null`, so this `LogFormatter.escape(...)` call throws and the request returns 500 without invoking a method-only CONNECT handler. In an exact-head control/treatment probe, the same route returned 200 without the feature and 500 with it.

danielkec (commented):
One current-head review issue.

common/uri/src/main/java/io/helidon/common/uri/UriPathNoParam.java
```diff
@@ -133,6 +133,7 @@ private static String decode(String rawPath, boolean validate) {
             rawPath = rawPath.substring(1);
         }
 
-        return URI.create(rawPath).normalize().getPath();
+        String path = URI.create(rawPath).normalize().getPath();
+        return path == null ? "" : path;
```
danielkec, line 141: This normalization applies only to the URI-parsing branch; the fast path above returns `rawPath` unchanged when it contains no `%`, `.`, or `//`. Consequently, `UriPath.create("example.com:443").path()` is empty, while `UriPath.create("localhost:443").path()` is `"localhost:443"`, although both inputs have no path under the new `UriPath.path()` contract.

tomas-langer: Fixed by sending slashless colon-containing values through URI parsing while preserving the existing fast path for ordinary paths. `example.com:443` and `localhost:443` now both produce an empty path, malformed values still fail validation, and decoded rootless colon paths retain path semantics. Added regression coverage for these cases.

webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
```diff
@@ -0,0 +1,398 @@
...
+                ActiveRequest completed;
+                while (loggedRecoveries < MAX_RECOVERY_LOGS_PER_CYCLE
+                        && (completed = completedRequests.poll()) != null) {
+                    if (running.get()) {
+                        logRecovery(completed);
+                    }
+                    loggedRecoveries++;
+                }
+            } catch (RuntimeException e) {
+                LOGGER.log(System.Logger.Level.ERROR,
```
danielkec, line 203: A persistent logging-backend failure can terminate the sole monitor here. This catch reports the recovery failure through the same logger without another guard (the scan catch at line 279 has the same pattern), so if the backend also throws for the ERROR record, the exception escapes `monitorRequests()` while `running` remains true and `monitorThread` still points to the dead thread. Reproducer: install a JUL `Handler` that throws from `publish()` for INFO, WARNING, and ERROR, trigger one recovery or scan failure, then remove the handler and block another request past several scan periods; no warning is emitted.

webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
```diff
@@ -0,0 +1,398 @@
...
+        completedRequests.clear();
+        omittedRecoveries.set(0);
+        if (interrupted) {
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    @Override
+    public void filter(FilterChain chain, RoutingRequest req, RoutingResponse res) {
+        if (!running.get()) {
```
danielkec, line 124: Requests accepted during startup can bypass detection permanently here. `ServerListener` starts its accept thread before `LoomServer.fireAfterStart()`, so while another listener delays startup, the already-bound listener can enter this branch; that request is never added to `activeRequests` after `running` becomes true. Reproducer: hold a second listener in `beforeStart`, send a blocking request to the default listener, release startup, and wait past the threshold and several scans; the request produces no warning.

danielkec (commented):
Two current-head review issues not covered by the existing threads.

webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
tomas-langer: If the logging backend starts throwing exceptions, the application is already in a fatal state. Adding another guard around the fallback logger would be over-protective, so I’m leaving this unchanged.

webserver/webserver/src/main/java/io/helidon/webserver/StuckThreadDetectionFilter.java
tomas-langer: Fixed in 615b8bae7d. Request tracking is now enabled in `beforeStart()`, while monitor-thread creation remains in `afterStart()`. Added regression coverage for a request entering between those callbacks.

danielkec (approved):

PR #12198, opened 2026-07-24
Stabilize listener port refusal lifecycle test

tomas-langer:
A WebServer lifecycle test may observe one transient loopback handshake while a closed listener's blocking `accept()` call unwinds on Linux.

Retry the port-refusal assertion with a five-second deadline while preserving the assertion that stop and suspend remain blocked until the running idle-timeout task is released. Each successful racing connection is closed immediately, and a final failure reports the number of attempts.

### DIFF
--- a/webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
@@ -693,10 +693,10 @@ class ServerListenerLifecycleTest {
 
             assertThat(stopStarted.await(5, TimeUnit.SECONDS), is(true));
             assertThat(connection.awaitForcedClose(), is(true));
+            assertPortEventuallyRefusesConnections(port);
             assertThat("stop should wait for the running idle timeout task",
                        stopDone.getCount(),
                        is(1L));
-            assertPortRefusesConnections(port);
 
             connection.releaseIdleClose();
             stopThread.join(TimeUnit.SECONDS.toMillis(5));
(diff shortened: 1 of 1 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
```diff
@@ -934,13 +934,24 @@ private static void assertPortCanBind(int port) throws Exception {
...
-            throw new AssertionError("Listener socket still accepted connections on port " + port);
-        } catch (ConnectException | SocketTimeoutException _) {
-            // validates that the listener socket is closed while shutdown waits for idle-timeout cleanup
+        long deadline = System.nanoTime() + TimeUnit.SECONDS.toNanos(5);
+        int attempts = 0;
+        while (true) {
+            try (Socket socket = new Socket()) {
+                attempts++;
+                socket.connect(new InetSocketAddress(InetAddress.getLoopbackAddress(), port), 250);
+            } catch (ConnectException | SocketTimeoutException _) {
```
danielkec, line 945: `SocketTimeoutException` does not prove that the listener closed after this loop has made successful connections. If the listener remains bound but its accept loop stops, these retries can fill the accept backlog; a later connect then times out and this returns while the listener is still open. With an open loopback listener using backlog 128 and no `accept()`, this exact 250 ms/10 ms loop completed 129 connections, then timed out on attempt 130 with `listener.isClosed() == false`. The previous one-shot check failed on its first successful connection. Because the effective backlog is platform-capped, this can mask the lifecycle regression the test is meant to detect.

tomas-langer: I think this test is "good enough" for the task it is trying to do.

tvallin (approved):

webserver/webserver/src/test/java/io/helidon/webserver/ServerListenerLifecycleTest.java
danielkec: Agree

danielkec (approved):

PR #12212, opened 2026-07-30
Register client filter only when CDI can resolve Helidon Tracer to allow Jersey client to work

### LINKED ISSUE #12211: 4.x Helidon telemetry filter causes failure in Jersey client (non-CDI environment)
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
The `helidon-microprofile-telemetry` component exposes a Jersey `AutoDiscoverable` that unconditionally registers `HelidonTelemetryClientFilter`.

For a plain Jersey client using HK2 outside a Helidon MP/CDI runtime, HK2 attempts to construct the filter but cannot resolve its CDI dependencies:

- `io.helidon.tracing.Tracer`
- telemetry-helper `Instance`s

A normal `ClientBuilder.newBuilder().build()` can therefore fail on its first request merely because the MP telemetry artifact is on the classpath.

### Expected Behavior
Standalone Jersey clients should continue without MP telemetry when no suitable CDI runtime is active. Helidon MP applications should retain automatic outgoing-client telemetry.

## Steps to reproduce
See attached user-provided project.

[filter-bug-with-jersey-client.zip](https://github.com/user-attachments/files/30563306/filter-bug-with-jersey-client.zip)

tjquinno:
### Description
Resolves #12211

### Release Note
____
Plain Jersey clients that include `helidon-microprofile-telemetry` but do not run in a compatible CDI runtime no longer fail on their first request. In that environment, automatic MP telemetry for Jersey client requests is skipped.
Helidon MP/CDI applications retain automatic outgoing Jersey client telemetry.
____

### Overview
Prevents plain Jersey/HK2 clients from failing when `helidon-microprofile-telemetry `is present without an active Helidon MP/CDI runtime.
Jersey auto-discovery now registers the telemetry client filter only when CDI can resolve a Helidon `Tracer`. The filter is registered as a CDI-constructed instance, avoiding HK2 attempts to inject CDI dependencies.

#### Tests

- Added auto-discovery decision-path tests.
- Added a plain Jersey client regression test.
- Verified existing MP client span-nesting coverage remains successful.
- 

### Documentation
No impact.


microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
```diff
@@ -15,28 +15,65 @@
...
+                           "Skipping Jersey client telemetry because CDI cannot resolve a Helidon tracer");
+                return;
+            }
+
+            Instance<HelidonTelemetryClientFilterHelper> helpers = cdi.select(HelidonTelemetryClientFilterHelper.class);
+            ctx.register(new HelidonTelemetryClientFilter(tracers.get(), helpers));
+        } catch (IllegalStateException e) {
+            LOGGER.log(System.Logger.Level.TRACE,
+                       "Skipping Jersey client telemetry because no active CDI container is available");
+        } catch (RuntimeException e) {
```
tomas-langer: Once CDI is active and the `Tracer` is resolvable, this catch also swallows failures from `tracers.get()`, helper materialization in the filter constructor, and `FeatureContext.register`. The client then continues without spans or trace propagation, masking a broken MP telemetry setup. Limit graceful fallback to the no-CDI or unresolvable-tracer cases and allow failures after successful resolution to surface. Add coverage for a post-resolution dependency or registration failure.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
```diff
@@ -15,28 +15,65 @@
...
+            CDI<Object> cdi = cdiSupplier.get();
+            Instance<Tracer> tracers = cdi.select(Tracer.class);
+            if (!tracers.isResolvable()) {
+                LOGGER.log(System.Logger.Level.TRACE,
+                           "Skipping Jersey client telemetry because CDI cannot resolve a Helidon tracer");
+                return;
+            }
+
+            Instance<HelidonTelemetryClientFilterHelper> helpers = cdi.select(HelidonTelemetryClientFilterHelper.class);
+            ctx.register(new HelidonTelemetryClientFilter(tracers.get(), helpers));
```
tomas-langer: This instance registration bypasses CDI/JAX-RS lifecycle management for the filter and its dependencies. The filter obtains a dependent-scoped `Tracer` and materializes helper beans, but neither their CDI handles nor a destruction hook are retained for `Client.close()`. Repeated client construction can retain dependent helper resources until CDI shutdown. Keep class-based filter registration after confirming CDI/tracer availability, or explicitly tie CDI handle destruction to client closure.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
```diff
@@ -15,28 +15,65 @@
...
+    private final Supplier<CDI<Object>> cdiSupplier;
 
     /**
      * For service loading.
      */
     public TelemetryAutoDiscoverable() {
+        this(CDI::current);
+    }
+
+    TelemetryAutoDiscoverable(Supplier<CDI<Object>> cdiSupplier) {
```
tomas-langer: The `Supplier<CDI<Object>>` field and package-private constructor are used only to replace `CDI.current()` in `TelemetryAutoDiscoverableTest`; the service-loaded production path always uses the no-arg constructor. This adds a test-only production seam. Cover the no-CDI and MP/CDI paths through observable integration behavior and remove the supplier-backed constructor and field.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
tjquinno: Catch of `RuntimeException` removed.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
tjquinno: Fixed.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
tjquinno: Fixed.

tomas-langer (approved):
Reviewed current head bda852e7c1a3f5f452c63791d588ddb2ee466070; no PR-caused findings remain.

PR #12218, opened 2026-07-31
OIDC cookie compression 4.x

Verdent:
### Description
Fixes https://github.com/helidon-io/helidon/issues/12223

### Documentation

For a top-level `server-type=idcs`, the access-token and ID-token cookies are GZIP-compressed by default when
compression reduces their size. Compression is disabled by default for other server types and can be enabled explicitly with `cookie-compression-enabled` for the access token and `cookie-compression-id-enabled` for the ID token.  Compression is applied before encryption when encryption is enabled, and uses a cookie-safe encoded form otherwise. 

During a rolling upgrade from a version that does not understand compressed token cookies, set both options to `false` on all nodes before upgrading. New nodes still read compressed cookies while these settings are disabled, but write the older uncompressed format. Re-enable compression after all nodes have been upgraded and the rollback window has closed. Older nodes cannot read compressed cookies. Rolling back after compression has been re-enabled therefore requires invalidating affected sessions or waiting for the compressed cookies to expire and users to authenticate again.

### DIFF
--- a/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcCookieHandler.java
+++ b/security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcCookieHandler.java
@@ -47,6 +47,7 @@ public class OidcCookieHandler {
     private static final System.Logger LOGGER = System.getLogger(OidcCookieHandler.class.getName());
     private static final byte COMPRESSED_BASE64_VALUE = 0;
     private static final byte COMPRESSED_RAW_VALUE = 1;
+    private static final String COMPRESSED_VALUE_PREFIX = "~";
     private static final int COMPRESSION_BUFFER_SIZE = 1024;
     private static final int MAX_COOKIE_VALUE_SIZE = 64 * 1024;
     private static final int MAX_DECOMPRESSED_BASE64_VALUE_SIZE = MAX_COOKIE_VALUE_SIZE / 4 * 3;
@@ -119,8 +120,39 @@ public class OidcCookieHandler {
             this.decryptFunction = it -> new String(decompress(cookieEncryption.decrypt(it)),
                                                     StandardCharsets.UTF_8);
         } else {
-            this.encryptFunction = Function.identity();
-            this.decryptFunction = Function.identity();
+            if (builder.compressionEnabled) {
+                this.encryptFunction = it -> {
+                    byte[] valueBytes = it.getBytes(StandardCharsets.UTF_8);
+                    byte[] compressed = compress(valueBytes);
+                    if (compressed == valueBytes) {
+                        return it;
+                    }
+                    String encoded = COMPRESSED_VALUE_PREFIX + Base64.getEncoder().encodeToString(compressed);
+                    return encoded.length() < valueBytes.length ? encoded : it;
+                };
+            } else {
+                this.encryptFunction = Function.identity();
+            }
+            if (builder.compressionConfigured) {
+                // OIDC token values use Base64 or Base64URL, neither of which contains the compression prefix.
+                this.decryptFunction = it -> {
+                    if (!it.startsWith(COMPRESSED_VALUE_PREFIX)) {
+                        return it;
+                    }
+                    try {
+                        byte[] compressed = Base64.getDecoder().decode(it.substring(COMPRESSED_VALUE_PREFIX.length()));
+                        byte[] decompressed = decompress(compressed);
+                        if (decompressed == compressed) {
+                            throw new CryptoException("OIDC compressed cookie has an unsupported format");
+                        }
+                        return new String(decompressed, StandardCharsets.UTF_8);
+                    } catch (IllegalArgumentException e) {
+                        throw new CryptoException("OIDC cookie decompression failed", e);
+                    }
+                };
+            } else {
+                this.decryptFunction = Function.identity();
+            }
         }
 
         if (LOGGER.isLoggable(Level.TRACE)) {
@@ -167,7 +199,7 @@ public class OidcCookieHandler {
 
     /**
      * Locate cookie in a map of headers and return its value.
-     * If the cookie is encrypted, decrypts the cookie value.
+     * If the cookie is encrypted or compressed, restores the original cookie value.
      *
      * @param headers headers to process
      * @return cookie value, or empty if the cookie could not be found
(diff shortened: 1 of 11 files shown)


security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcCookieHandler.java
```diff
@@ -92,11 +108,51 @@ private OidcCookieHandler(Builder builder) {
...
+                    String encoded = COMPRESSED_VALUE_PREFIX + Base64.getEncoder().encodeToString(compressed);
+                    return encoded.length() < valueBytes.length ? encoded : it;
+                };
+            } else {
+                this.encryptFunction = Function.identity();
+            }
+            if (builder.compressionConfigured) {
+                // OIDC token values use Base64 or Base64URL, neither of which contains the compression prefix.
+                this.decryptFunction = it -> {
+                    if (!it.startsWith(COMPRESSED_VALUE_PREFIX)) {
```
tomas-langer, line 139: `findCookie` can throw `CryptoException` here before `TenantAuthenticationHandler.authenticate` enters the catch that handles invalid ID-token cookies. With ID-token encryption disabled, a request such as `JSESSIONID_2=~not-base64` now escapes authentication and can become an HTTP 500 instead of a controlled authentication failure. Catch cookie-restoration failures around the `findCookie` call, matching the access-token cookie path, and add a provider-level regression test for a malformed unencrypted `~` cookie.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcCookieHandler.java
```diff
@@ -188,6 +244,82 @@ String cookieValuePrefix() {
...
+            if (decompressed.length > maxDecompressedValueSize) {
+                throw new CryptoException("OIDC decompressed cookie exceeds the maximum supported size");
+            }
+            return base64Value ? Base64.getEncoder().encode(decompressed) : decompressed;
+        } catch (IOException e) {
+            throw new CryptoException("OIDC cookie decompression failed", e);
+        }
+    }
+
+    private static final class BestCompressionGzipOutputStream extends GZIPOutputStream {
```
tomas-langer: This private nested type is interleaved with method declarations: `createCookieDirectValue` follows it, and the package-private `Builder` follows that. Helidon declaration order keeps nested member types in the final section of the class, ordered from most visible to least visible. Move `BestCompressionGzipOutputStream` after `Builder`.

tomas-langer (changes requested):
Two blocking findings are attached.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcCookieHandler.java
```diff
@@ -92,11 +108,51 @@ private OidcCookieHandler(Builder builder) {
...
+                    if (!it.startsWith(COMPRESSED_VALUE_PREFIX)) {
+                        return it;
+                    }
+                    try {
+                        byte[] compressed = Base64.getDecoder().decode(it.substring(COMPRESSED_VALUE_PREFIX.length()));
+                        byte[] decompressed = decompress(compressed);
+                        if (decompressed == compressed) {
+                            throw new CryptoException("OIDC compressed cookie has an unsupported format");
+                        }
+                        return new String(decompressed, StandardCharsets.UTF_8);
```
tomas-langer, line 148: Unencrypted compressed cookie values are decoded into arbitrary UTF-8 here, and `OidcFeature.logoutWithTenant` appends the restored ID-token value directly to the `Location` header (`OidcFeature.java:352-374`) without validating it or URL-encoding `id_token_hint`. A cookie-safe `~` + Base64/GZIP value can therefore restore CR/LF: with normal response-header validation the logout becomes HTTP 500, and with `validate-response-headers=false` HTTP/1 can write injected headers. Validate the restored value as an ID token and URL-encode it before constructing `Location`, and add a logout regression for an unencrypted compressed cookie containing invalid/header-control data.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcFeature.java
```diff
@@ -349,7 +352,27 @@ private void logoutWithTenant(ServerRequest req,
                                   String encryptedIdToken,
                                   String stateQuery) {
         try {
-            String idToken = idTokenCookieHandler.decrypt(encryptedIdToken);
+            String idToken;
+            try {
+                idToken = idTokenCookieHandler.decrypt(encryptedIdToken);
+                JwtHeaders jwtHeaders = JwtHeaders.parseToken(idToken);
+                if (jwtHeaders.encryption().isPresent()) {
+                    EncryptedJwt.parseToken(jwtHeaders, idToken);
```
tomas-langer, line 360: This passes up to 64 KiB of decompressed, attacker-controlled cookie input to `EncryptedJwt.parseToken`. Its five-segment matcher uses repeated `\\S+` groups separated by dots; a valid JWE header followed by a long near-match can cause catastrophic backtracking and `StackOverflowError`. Because ID-token cookie encryption may be disabled, an unauthenticated request can reach this path, and the new `catch` does not catch `Error`. Bound and validate the exact five base64url segments with a linear parser before calling `EncryptedJwt.parseToken` (or make that parser linear), and add an adversarial compressed-cookie regression.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcFeature.java
```diff
@@ -349,7 +352,27 @@ private void logoutWithTenant(ServerRequest req,
...
+            } catch (CryptoException
+                     | IllegalArgumentException
+                     | JwtException
+                     | Errors.ErrorMessagesException e) {
+                if (LOGGER.isLoggable(Level.TRACE)) {
+                    LOGGER.log(Level.TRACE, "Invalid OIDC logout ID token cookie", e);
+                }
+                res.status(Status.BAD_REQUEST_400)
+                        .send();
+                return;
```
tomas-langer, line 374: The 400 response still returns before `clearLocalOidcCookies`, so a corrupt or malformed ID-token cookie makes logout fail while leaving the access, ID, tenant, and refresh cookies active. The unknown-tenant and successful paths both clear the local session cookies. Clear the local OIDC cookies before this return (or in a failure/finally path), and extend the regression to assert the removal cookies as well as the 400 status.

tomas-langer (approved):

PR #12221, opened 2026-08-01
Skip unknown HTTP/2 SETTINGS identifiers when parsing frames

### LINKED ISSUE #12196: HTTP/2: unknown SETTINGS identifier is not skipped, corrupting parsing of the remaining settings (peer INITIAL_WINDOW_SIZE dropped → outbound stream stalls at 65535 bytes)
## Environment Details
* Helidon Version: 4.5.0
* Helidon SE 
* JDK version:  JDK 25
* OS: Linux

----------

## Problem Description

Helidon's HTTP/2 `SETTINGS` frame parser does not skip the value of a setting whose **identifier it
does not recognize**. It reads the 2-byte identifier but, for an unknown/unsupported setting, it
never consumes the 4-byte value. Since every setting is 6 bytes and the loop is bounded by
`available / 6`, the read cursor becomes misaligned and **every setting after the unknown one is
parsed from the wrong offset** and effectively lost.

**Current behavior:** if a peer sends `SETTINGS_INITIAL_WINDOW_SIZE (0x4)` **after** an unknown
setting in the same frame, Helidon silently drops the new initial window and keeps the default
65535. The server's outbound (send) flow-control window for a stream therefore stays at 65535, so
**any response body larger than 65535 bytes stalls forever at exactly 65535 bytes**, waiting for a
`WINDOW_UPDATE` that a well-behaved peer has no reason to send (it believes it already granted a
much larger window).

**Expected behavior:** per **RFC 9113 §6.5.2** — *"An endpoint that receives a SETTINGS frame with
any unknown or unsupported identifier MUST ignore that setting."* — the unknown setting (including
its value) must be skipped and the remaining settings, notably `INITIAL_WINDOW_SIZE`, must still be
applied. The full response should be delivered.

**Easily reproducible:** yes, 100% deterministic (see below).

### Root cause (exact location)

`http/http2/src/main/java/io/helidon/http/http2/Http2Settings.java`, method `create(BufferData)`:

```java
int settingCount = available / 6; // each setting is 6 bytes
Map<Integer, SettingValue> values = new HashMap<>();

for (int i = 0; i < settingCount; i++) {
    int identifier = frame.readInt16();          // consumes 2 bytes

    Http2Setting<?> http2Setting = Http2Setting.BY_ID.get(identifier);
    if (http2Setting != null) {
        values.put(identifier, new SettingValue(http2Setting, http2Setting.read(frame))); // consumes 4 bytes
    }
    // BUG: when http2Setting == null (unknown identifier) the 4-byte value is NOT consumed,
    //      so the next iteration's readInt16() reads from the middle of this setting's value.
}
```

For an unknown identifier the loop consumes only 2 of the setting's 6 bytes, so the cursor is 4
bytes short for every subsequent iteration and the rest of the frame is mis-parsed.

### Suggested fix

Consume (skip) the 4-byte value of an unknown setting so the cursor stays aligned:

```java
Http2Setting<?> http2Setting = Http2Setting.BY_ID.get(identifier);
if (http2Setting != null) {
    values.put(identifier, new SettingValue(http2Setting, http2Setting.read(frame)));
} else {
    frame.skip(4); // ignore unknown setting, but still consume its value (RFC 9113 §6.5.2)
}
```

### Real-world trigger and impact

An **Envoy** proxy (gRPC / gRPC-Web) in front of a Helidon gRPC backend advertises
`SETTINGS_ENABLE_CONNECT_PROTOCOL` (`0x8`, RFC 8441) **before** `SETTINGS_INITIAL_WINDOW_SIZE` on
its upstream HTTP/2 connection. Helidon therefore ignores Envoy's larger initial window, and every
unary gRPC response bigger than 65535 bytes deadlocks:

```
Envoy → Helidon:  SETTINGS[ HEADER_TABLE_SIZE, ENABLE_PUSH, ENABLE_CONNECT_PROTOCOL(0x8), MAX_CONCURRENT_STREAMS, INITIAL_WINDOW_SIZE=64MiB ]
                  WINDOW_UPDATE(stream 0, +128MiB)
Helidon → Envoy:  DATA … total exactly 65535 bytes, then no more (stream never ends, no trailers)
```

Envoy reaches its route timeout, resets the stream, and the gRPC client (grpc-web / grpc-dart)
never receives trailers, so its call never completes. The defect is in the shared
`io.helidon.http.http2` `SETTINGS` parser, so **plain (non-gRPC) HTTP/2 responses larger than
65535 bytes are affected too** — which is how the minimal reproduction below is built.

This is **not** #9589 / hashgraph/pbj#329 (outbound window drifting negative on sustained
streaming), which was fixed in 4.1.6/4.2.0 (PR #9520) and is already present in 4.5.0. This is a
distinct `SETTINGS`-parsing defect.

## Steps to reproduce

Minimal, self-contained, **plain HTTP/2 (no gRPC needed)**. Both the server and the client are
inline below (the client is in the collapsible block at the end).

**1. Server** — a ~15-line Helidon 4.5.0 SE `WebServer` (`GET /big?size=N` returns N `'x'` bytes):

```java
import io.helidon.webserver.WebServer;

public final class Http2SettingsBugServer {
    public static void main(String[] args) throws Exception {
        int port = args.length > 0 ? Integer.parseInt(args[0]) : 8080;
                .port(port)
                .routing(r -> r.get("/big", (req, res) -> {
                    int size = req.query().first("size").map(Integer::parseInt).orElse(70000);
                    res.send("x".repeat(size));
                }))
                .build()
                .start();
        System.out.println("listening on :" + server.port());
        Thread.currentThread().join();
    }
}
```

Compile/run with `helidon-webserver` + `helidon-webserver-http2` (4.5.0) on the classpath, JDK 25:
`java -cp <helidon-classpath>:. Http2SettingsBugServer 8080`.

**2. Client** — `repro_client.py`, a dependency-free HTTP/2 (h2c prior-knowledge) client. It issues a
single `GET /big?size=70000` and sends **no** per-stream `WINDOW_UPDATE`, so the server relies on the
advertised `INITIAL_WINDOW_SIZE`. The only difference between modes is the `SETTINGS` frame it sends:

```
python3 repro_client.py 127.0.0.1 8080 --mode minimal   # SETTINGS[ INITIAL_WINDOW_SIZE=64MiB ]
python3 repro_client.py 127.0.0.1 8080 --mode ok        # SETTINGS[ INITIAL_WINDOW_SIZE, 0x8 ]  (unknown AFTER)
python3 repro_client.py 127.0.0.1 8080 --mode stall     # SETTINGS[ 0x8, INITIAL_WINDOW_SIZE ]  (unknown BEFORE)
```

**Observed output** (Helidon 4.5.0, response body = 70000 bytes):

```
mode=minimal body_bytes= 70000 end_stream=True  -> complete
mode=ok      body_bytes= 70000 end_stream=True  -> complete
mode=stall   body_bytes= 65535 end_stream=False -> STALLED
```

Only `--mode stall` (an unknown setting **before** `INITIAL_WINDOW_SIZE`) stalls, at exactly 65535
bytes with no `END_STREAM`. Any unknown id (`0x7`, `0x63`, …) reproduces it; using only known ids
before `INITIAL_WINDOW_SIZE` does not — confirming the trigger is the unskipped value of an unknown
setting, not the specific id `0x8`.

<details>
<summary><code>repro_client.py</code> (dependency-free, ~100 lines — click to expand)</summary>

```python
#!/usr/bin/env python3
"""
Minimal, dependency-free HTTP/2 (h2c, prior knowledge) client that demonstrates the
Helidon 4.5.0 SETTINGS-parsing bug (helidon-io/helidon issue: unknown SETTINGS id not skipped).

It performs a single GET whose response body is larger than 65535 bytes and only varies the
SETTINGS frame it sends. It sends NO per-stream WINDOW_UPDATE, so the server must rely on the
advertised INITIAL_WINDOW_SIZE to send the whole body.

    python3 repro_client.py <host> <port> [--mode stall|ok|minimal] [--path /big?size=70000]

Modes (SETTINGS parameters sent, in frame order):
  stall    : <unknown id 0x8>, INITIAL_WINDOW_SIZE=64MiB   -> server stalls at 65535 bytes (BUG)
  ok       : INITIAL_WINDOW_SIZE=64MiB, <unknown id 0x8>   -> server delivers the full body
  minimal  : INITIAL_WINDOW_SIZE=64MiB                     -> server delivers the full body

Exit: prints bytes received and whether the stream completed (END_STREAM seen).
"""
import socket, struct, sys, time

PREFACE = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
BIG_WINDOW = 64 * 1024 * 1024          # 64 MiB, advertised via SETTINGS_INITIAL_WINDOW_SIZE (id 0x4)
UNKNOWN_ID = 0x8                        # SETTINGS_ENABLE_CONNECT_PROTOCOL (RFC 8441); Helidon 4.5.0 doesn't know it


def frame(ftype, flags, stream_id, payload=b""):
    return struct.pack(">I", len(payload))[1:] + bytes([ftype, flags]) + struct.pack(">I", stream_id) + payload


def hpack_int(value, prefix_bits):
    mask = (1 << prefix_bits) - 1
    if value < mask:
        return bytes([value])
    out = bytes([mask]); value -= mask
    while value >= 128:
        out += bytes([(value & 0x7F) | 0x80]); value >>= 7
    return out + bytes([value])


def hpack_str(s):
    b = s.encode()
    return hpack_int(len(b), 7) + b            # H=0 (no Huffman)


def header_indexed(index):                     # fully indexed header (static table)
    return bytes([0x80 | index])


def header_named(name_index, value):           # literal, name from static table, value literal
    return hpack_int(name_index, 4) + hpack_str(value)


def settings_payload(mode):
    init = struct.pack(">HI", 0x4, BIG_WINDOW)             # INITIAL_WINDOW_SIZE
    unknown = struct.pack(">HI", UNKNOWN_ID, 0)           # an id Helidon does not recognize
    if mode == "stall":
        return unknown + init                             # unknown BEFORE the window -> window is dropped
    if mode == "ok":
        return init + unknown                             # unknown AFTER the window -> window applied
    return init                                           # minimal


def main():
    host, port = sys.argv[1], int(sys.argv[2])
    mode = "stall"; path = "/big?size=70000"
    for i, a in enumerate(sys.argv):
        if a == "--mode": mode = sys.argv[i + 1]
        if a == "--path": path = sys.argv[i + 1]

    s = socket.create_connection((host, port), timeout=10); s.settimeout(10)
    s.sendall(PREFACE)
    s.sendall(frame(0x4, 0, 0, settings_payload(mode)))               # our SETTINGS
    s.sendall(frame(0x8, 0, 0, struct.pack(">I", BIG_WINDOW)))        # grow the connection window too
    # request headers (GET), pseudo-headers first
    hb = (header_indexed(2)                        # :method GET
          + header_indexed(6)                      # :scheme http
          + header_named(4, path)                  # :path
          + header_named(1, f"{host}:{port}"))     # :authority
    s.sendall(frame(0x1, 0x4 | 0x1, 1, hb))        # HEADERS: END_HEADERS | END_STREAM (no body)

    body = 0; end = False; started = time.time(); buf = b""
    def readn(n):
        nonlocal buf
        while len(buf) < n:
            d = s.recv(65536)
            if not d: raise ConnectionError("closed")
            buf += d
        r, buf = buf[:n], buf[n:]
        return r
    outcome = "complete"
    try:
        while not end:
            hd = readn(9)
            ln = int.from_bytes(hd[:3], "big"); ftype = hd[3]; flags = hd[4]
            pl = readn(ln) if ln else b""
            if ftype == 0x4 and not (flags & 1):        # peer SETTINGS -> ACK
                s.sendall(frame(0x4, 1, 0))
            elif ftype == 0x6 and not (flags & 1):      # PING -> PONG
                s.sendall(frame(0x6, 1, 0, pl))
            elif ftype == 0x0:                          # DATA
                body += ln
                if flags & 1: end = True
            elif ftype in (0x1, 0x3, 0x7) and (flags & 1 or ftype != 0x1):
                end = True
    except (socket.timeout, ConnectionError):
        outcome = "STALLED"
    dt = time.time() - started
    print(f"mode={mode:7s} body_bytes={body:6d} end_stream={end} -> {outcome} (t={dt:.1f}s)")


if __name__ == "__main__":
    main()
```

</details>

arimu1:
### Description

Fixes #12196.

HTTP/2 SETTINGS parsing in `Http2Settings.create(BufferData)` did not consume the 4-byte value of an unknown/unsupported setting identifier. Because each setting is 6 bytes (2-byte id + 4-byte value) and the loop is bounded by `available / 6`, the read cursor became misaligned and **every setting after the unknown one was parsed from the wrong offset** and lost.

In particular, when a peer sent an unknown setting (e.g. Envoy’s `SETTINGS_ENABLE_CONNECT_PROTOCOL` / `0x8`) **before** `SETTINGS_INITIAL_WINDOW_SIZE`, Helidon dropped the peer’s larger initial window and kept the default **65535**. Outbound responses larger than 65535 bytes then stalled forever waiting for a `WINDOW_UPDATE` the peer had no reason to send.

**Fix:** per [RFC 9113 §6.5.2](https://www.rfc-editor.org/rfc/rfc9113.html#section-6.5.2) — *“An endpoint that receives a SETTINGS frame with any unknown or unsupported identifier MUST ignore that setting.”* — skip the 4-byte value for unknown identifiers so subsequent settings stay aligned and are applied.

```java
if (http2Setting != null) {
    values.put(identifier, new SettingValue(http2Setting, http2Setting.read(frame)));
} else {
    frame.skip(4); // ignore unknown setting, still consume its value
}
```

Unit tests cover:
- known settings only
- unknown **before** `INITIAL_WINDOW_SIZE` (regression for #12196)
- unknown **after** `INITIAL_WINDOW_SIZE`
- multiple unknown ids before known settings
- only-unknown frames leave defaults in place

### Documentation

None

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Settings.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Settings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -78,6 +78,10 @@ public final class Http2Settings implements Http2Frame<Http2Flag.SettingsFlags> 
             Http2Setting<?> http2Setting = Http2Setting.BY_ID.get(identifier);
             if (http2Setting != null) {
                 values.put(identifier, new SettingValue(http2Setting, http2Setting.read(frame)));
+            } else {
+                // RFC 9113 §6.5.2: ignore unknown/unsupported identifiers, but still consume the value
+                // so subsequent settings stay aligned (each setting is 2-byte id + 4-byte value).
+                frame.skip(4);
             }
         }
         return new Http2Settings(values);
--- a/http/http2/src/test/java/io/helidon/http/http2/Http2SettingsTest.java
+++ b/http/http2/src/test/java/io/helidon/http/http2/Http2SettingsTest.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.http.http2;
+
+import io.helidon.common.buffers.BufferData;
+
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+class Http2SettingsTest {
+
+    private static final int UNKNOWN_ID = 0x8; // SETTINGS_ENABLE_CONNECT_PROTOCOL (RFC 8441)
+    private static final long BIG_WINDOW = 64L * 1024 * 1024;
+
+    @Test
+    void knownSettingsAreParsed() {
+        BufferData frame = settingsBuffer(
+                setting(Http2Setting.INITIAL_WINDOW_SIZE.identifier(), BIG_WINDOW),
+                setting(Http2Setting.MAX_FRAME_SIZE.identifier(), 32_768L));
+
+        Http2Settings settings = Http2Settings.create(frame);
+
+        assertThat(settings.value(Http2Setting.INITIAL_WINDOW_SIZE), is(BIG_WINDOW));
+        assertThat(settings.value(Http2Setting.MAX_FRAME_SIZE), is(32_768L));
+        assertThat(settings.hasValue(Http2Setting.INITIAL_WINDOW_SIZE), is(true));
+    }
+
+    @Test
+    void unknownSettingBeforeInitialWindowSizeIsSkipped() {
+        // #12196: unknown id before INITIAL_WINDOW_SIZE must not drop the window setting
+        BufferData frame = settingsBuffer(
+                setting(UNKNOWN_ID, 0L),
+                setting(Http2Setting.INITIAL_WINDOW_SIZE.identifier(), BIG_WINDOW));
+
+        Http2Settings settings = Http2Settings.create(frame);
+
+        assertThat(settings.hasValue(Http2Setting.INITIAL_WINDOW_SIZE), is(true));
+        assertThat(settings.value(Http2Setting.INITIAL_WINDOW_SIZE), is(BIG_WINDOW));
+    }
+
+    @Test
+    void unknownSettingAfterInitialWindowSizeIsSkipped() {
+        BufferData frame = settingsBuffer(
+                setting(Http2Setting.INITIAL_WINDOW_SIZE.identifier(), BIG_WINDOW),
+                setting(UNKNOWN_ID, 0L));
+
+        Http2Settings settings = Http2Settings.create(frame);
+
+        assertThat(settings.value(Http2Setting.INITIAL_WINDOW_SIZE), is(BIG_WINDOW));
+    }
+
+    @Test
+    void multipleUnknownSettingsDoNotCorruptLaterKnownSettings() {
+        BufferData frame = settingsBuffer(
+                setting(0x7, 1L),
+                setting(UNKNOWN_ID, 0L),
+                setting(0x63, 99L),
+                setting(Http2Setting.MAX_CONCURRENT_STREAMS.identifier(), 100L),
+                setting(Http2Setting.INITIAL_WINDOW_SIZE.identifier(), BIG_WINDOW));
+
+        Http2Settings settings = Http2Settings.create(frame);
+
+        assertThat(settings.value(Http2Setting.MAX_CONCURRENT_STREAMS), is(100L));
+        assertThat(settings.value(Http2Setting.INITIAL_WINDOW_SIZE), is(BIG_WINDOW));
+    }
+
+    @Test
+    void onlyUnknownSettingsYieldsEmptyPresentValues() {
+        BufferData frame = settingsBuffer(
+                setting(UNKNOWN_ID, 1L),
+                setting(0x63, 2L));
+
+        Http2Settings settings = Http2Settings.create(frame);
+
+        assertThat(settings.hasValue(Http2Setting.INITIAL_WINDOW_SIZE), is(false));
+        assertThat(settings.value(Http2Setting.INITIAL_WINDOW_SIZE),
+                   is((long) WindowSize.DEFAULT_WIN_SIZE));
+        assertThat(settings.presentValue(Http2Setting.ENABLE_PUSH).isEmpty(), is(true));
+    }
+
+    private static byte[] setting(int identifier, long value) {
+        BufferData tmp = BufferData.create(6);
+        tmp.writeInt16(identifier);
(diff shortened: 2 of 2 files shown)


tomas-langer:
I am closing this PR, as OCA was not signed, and I will work on the issue myself.

Issue #12226, opened 2026-08-04
Helidon SE guide for Metrics contains an unresolved AsciiDoc attribute reference: {h1-prefix}


bhatpmk:
Helidon Version: 4.5.0
Documentation URL: https://helidon.io/docs/v4/se/guides/metrics

## Problem Description
The guide for Metrics contains an unresolved AsciiDoc attribute reference: {h1-prefix}

`This guide describes how to create a sample Helidon {h1-prefix} project that can be used to run some basic examples using both built-in and custom meters with Helidon.`

PR #12220, opened 2026-07-31
4.x: Convert AsciiDoc documentation to Markdown

romain-grecourt:
### Description

Convert AsciiDoc documentation to Markdown.

The Markdown is flavored using HTML comment directives so the source still
renders as vanilla Markdown.

The configuration documentation generator now outputs Markdown instead of
AsciiDoc, with simpler anchor handling for included pages. Generated
configuration docs are now checked in under `docs/config`, and a new CI job
regenerates them and fails if the checked-in output is out of date.

Aggregate Javadocs are now generated from the new `javadocs/pom.xml` module
instead of `docs/pom.xml`.

### Documentation

This pull request is the documentation conversion.

### Merge order

This is the fourth and final pull request in the documentation transition.
It must merge after #12216, #12217, and #12213. After those changes land,
this branch must be updated onto the resulting `helidon-4.x`, `docs/config`
regenerated, and validation rerun before merge.

### Validation

Validation included:

- focused configuration-reference generator tests;
- the aggregate Javadocs build;
- the CI freshness check for checked-in `docs/config` output;
- checkstyle, copyright, and whitespace checks; and
- structural callout, local-link, Markdown-width, and snippet-parity checks.

PR #12227, opened 2026-08-04
Fixes an unresolved AsciiDoc attribute in the Helidon SE Metrics Guide

bhatpmk:
### Description
The intro-project-name attribute was assigned from {h1-prefix} before h1-prefix was defined, leaving an unresolved attribute in the generated documentation. This change references {h1-prefix} directly, consistent with the rest of metrics.adoc.

### Documentation
None

### DIFF
--- a/docs/src/main/asciidoc/includes/guides/metrics.adoc
+++ b/docs/src/main/asciidoc/includes/guides/metrics.adoc
@@ -20,7 +20,7 @@ ifndef::rootdir[:rootdir: {docdir}/../..]
 
 // tag::intro[]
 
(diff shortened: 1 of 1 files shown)


romain-grecourt:
#12226 is filed against v4 but this PR targets main which is v27.

The docs are being converted from asciidoc to markdown as part of #12220 this change will be a direct conflict.

The descriptions says "None" for documentation but this PR is doing an actual documentation change.

tomas-langer:
This PR is no longer relevant. The markdown document is fixed, and the asciidoc no longer exists.

PR #12228, opened 2026-08-04
Fix resolved URI query parameters (4.x)

### LINKED ISSUE #8566: WebClient - Uri resolved without query params when path and query params are mixed.

tomas-langer:
Resolves #8566

Carries forward Kay Werndli’s contribution from #8614 onto the current 4.x line and preserves request-added query parameters when resolving URI templates.

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
@@ -295,14 +295,17 @@ public abstract class ClientRequestBase<T extends ClientRequest<T>, R extends Ht
 
     @Override
     public T queryParam(String name, String... values) {
-        if (uriTemplate != null) {
-            if (uriTemplateQueryParamNames == null) {
-                uriTemplateQueryParamNames = new HashSet<>();
-                uriTemplateQuerySnapshot = UriQueryWriteable.create().from(clientUri.query());
-            }
+        UriQueryWriteable querySnapshot = uriTemplate != null && uriTemplateQueryParamNames == null
+                ? UriQueryWriteable.create().from(clientUri.query())
+                : null;
+        clientUri.writeableQuery().set(name, values);
+        if (querySnapshot != null) {
+            uriTemplateQueryParamNames = new HashSet<>();
+            uriTemplateQuerySnapshot = querySnapshot;
+        }
+        if (uriTemplateQueryParamNames != null) {
             uriTemplateQueryParamNames.add(name);
         }
-        clientUri.writeableQuery().set(name, values);
         return identity();
     }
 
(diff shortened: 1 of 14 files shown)


webclient/api/src/main/java/io/helidon/webclient/api/ClientRequestBase.java
```diff
@@ -573,11 +585,27 @@ protected MediaContext mediaContext() {
...
+                uri = URI.create(resolved);
             } else {
-                toResolve.resolve(URI.create(UriEncoding.encodeUri(resolved)));
+                uri = URI.create(UriEncoding.encodeUri(resolved));
+            }
+            toResolve.resolve(uri);
+
+            if (requestQueryParams != null) {
+                UriQueryWriteable query = toResolve.writeableQuery();
+                requestQueryParams.forEach(query::set);
```
danielkec: [P2] Equivalent encoded names produce different signed and transmitted targets

This leaves the template's raw query entry behind when its percent-encoded name decodes to the same request parameter name. For example:

```java
.uri("https://www.example.com/{path}?%61=template")
.pathParam("path", "p")
.skipUriEncoding(true)
.queryParam("a", "request")
```

This resolves with `query().all("a") == ["request"]` and `pathWithQueryAndFragment() == "/p?a=request"`, but `query().rawValue() == "%61=template&a=request"`. WebClient Security snapshots that raw value for HTTP-signature `(request-target)`, while HTTP/1 and HTTP/2 send `pathWithQueryAndFragment()`. The signed target can therefore differ from the transmitted target, causing signature authentication to fail.

danielkec (commented):
One issue found.

tvallin (approved):

danielkec (approved):

PR #12240, opened 2026-08-06
Fix listener and provider configuration compatibility

tomas-langer:
Follow-up to #12092.

Keeps the deprecated listener connection-limit accessor synchronized with `maxConnections`, and represents provider configurations that accept either object or list form in generated configuration metadata.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
@@ -423,6 +423,9 @@ class SchemaGenerator {
             if (configForm == Option.Provider.ConfigForm.LIST) {
                 return "LIST";
             }
+            if (configForm == Option.Provider.ConfigForm.OBJECT_OR_LIST) {
+                return "MAP_OR_LIST";
+            }
         }
 
         var typeName = optionInfo.declaredType();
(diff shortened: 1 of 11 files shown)


builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
```diff
@@ -423,6 +423,9 @@ private String optionKind(OptionInfo optionInfo) {
             if (configForm == Option.Provider.ConfigForm.LIST) {
                 return "LIST";
             }
+            if (configForm == Option.Provider.ConfigForm.OBJECT_OR_LIST) {
+                return "MAP_OR_LIST";
```
danielkec, line 427: **[P1] This new kind aborts LSP metadata loading**

The current official `helidon-build-tools` LSP consumer defines only `VALUE`, `LIST`, and `MAP` and parses `kind` with `ConfiguredOptionKind.valueOf`. A minimal `ConfiguredOptionKind.valueOf("MAP_OR_LIST")` throws `IllegalArgumentException: No enum constant ...MAP_OR_LIST`. Exact-head WebServer metadata emits this value for `protocols`, `concurrency-limit`, and `features`; because the exception is uncaught in dependency collection, configuration metadata loading fails for the entire project.

tomas-langer: Tracked in helidon-io/helidon-build-tools#1166. We will keep `MAP_OR_LIST` in this PR and add language-server support in build tools.

danielkec (approved):
Reviewed at exact head e2754e24a417851da796059209cc51bb99b7d085.

PR #12245, opened 2026-08-07
Restrict non-standard HTTP method usage

tjquinno:
### Release note
____
Automatic HTTP metrics report only configured known HTTP methods by name, defaulting to the standard HTTP method names plus `LIST`; all other methods are recorded as `_OTHER`. This applies to both legacy and updated automatic HTTP metrics. The optional auto-http metrics config setting `known-methods` replaces its default list, so configurations must also list standard methods they intend to retain.
____

### Overview
The OTel HTTP semantic conventions implementation in Helidon now by default uses only the standard HTTP method names plus `LIST` in building the HTTP method tag that is part of a meter's ID, but users can assign a new config setting to override that set and prescribe their own.

### Documentation
No impact


webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -278,4 +294,47 @@ private long legacyStatusCode(RoutingResponse resp, Exception exception) {
...
 
     }
+
+    private static final class NonstandardHttpMethods {
+        private final Set<String> methods = new HashSet<>();
+        private final Lock lock = new ReentrantLock();
+        private final AtomicBoolean limitWarningLogged = new AtomicBoolean();
+
+        private String normalize(Method method) {
+            if (method == Method.GET
```
tomas-langer: We must support additional "known" HTTP methods, that must never be excluded (i.e. `LIST`)

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -278,4 +294,47 @@ private long legacyStatusCode(RoutingResponse resp, Exception exception) {
...
+            }
+
+            if (limitWarningLogged.compareAndSet(false, true)) {
+                LOGGER.log(WARNING,
+                           "Maximum of {0} non-standard HTTP methods has been reached; {1} and subsequent "
+                                   + "previously unseen non-standard methods will be recorded as UNKNOWN",
+                           MAX_NONSTANDARD_METHODS,
+                           methodName);
+            }
+            return UNKNOWN_METHOD;
```
tomas-langer: Overflow methods are recorded as `UNKNOWN`, but the OpenTelemetry HTTP semantic conventions used by this module define `_OTHER` and require an override for the known-method list when valid methods might otherwise be collapsed. Because membership here depends on encounter order, any extension method arriving after the 16th distinct value is collapsed to a non-standard sentinel; `UNKNOWN` is also itself a valid HTTP method token. Use a deterministic known-method list, emit `_OTHER`, and expose the required known-method override. See https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/model/http/registry.yaml.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -221,10 +237,10 @@ private void updateMetricsIfMeasured(RoutingRequest req,
                     .ifPresent(route -> attrBuilder.put(AttributeKey.stringKey(HTTP_ROUTE), route));
 
             if (isOptedIn(config, SERVER_ADDRESS)) {
-                attrBuilder.put(AttributeKey.stringKey(SERVER_ADDRESS), req.requestedUri().host());
+                attrBuilder.put(AttributeKey.stringKey(SERVER_ADDRESS), req.localPeer().host());
```
tomas-langer: `server.address` and `server.port` must describe the original logical server host from `Forwarded`, `:authority`, or `Host`, not the local listener. Using `localPeer()` collapses virtual hosts and proxied names to a binding IP and also changes the legacy/default output. Keep `requestedUri()` here and in the legacy path, and cover differing logical and local endpoints in both modes. See https://github.com/open-telemetry/semantic-conventions/blob/v1.37.0/docs/http/http-spans.md.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -278,4 +294,47 @@ private long legacyStatusCode(RoutingResponse resp, Exception exception) {
...
+                    || method == Method.DELETE
+                    || method == Method.CONNECT
+                    || method == Method.OPTIONS
+                    || method == Method.TRACE
+                    || method == Method.PATCH) {
+                return method.text();
+            }
+
+            String methodName = method.text();
+            lock.lock();
```
tomas-langer: `normalize` acquires the singleton provider-wide lock on every non-standard method, including already admitted names and all overflow names. Concurrent extension-method traffic therefore serializes during response completion. Make the steady-state lookup lock-free and lock only when admitting a previously unseen method.

tomas-langer (changes requested):
Requesting changes for the three inline findings: align method fallback and known-list handling with OpenTelemetry 1.37, preserve logical server address and port semantics, and remove steady-state global lock contention for extension methods.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: Addressed by creating the new `known-methods` auto-metrics config setting which defaults to the standard HTTP methods plus `LIST`.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: Fixed.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: Restored. That change should not have been part of the PR in the first place.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: No longer applies now that we have an allow-list for the HTTP methods to be accepted.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
```diff
@@ -246,7 +254,7 @@ private void updateLegacyMetricsIfMeasured(RoutingRequest req,
             }
             AttributesBuilder attrBuilder = Attributes.builder();
 
-            attrBuilder.put(AttributeKey.stringKey(HTTP_METHOD), req.prologue().method().text())
+            attrBuilder.put(AttributeKey.stringKey(HTTP_METHOD), httpMethod(req.prologue().method()))
```
tomas-langer, line 257: `useUpdatedHttpMetrics=false` no longer preserves the earlier 4.x method-label behavior here: custom methods outside `known-methods` are now recorded as `_OTHER` in the legacy path. The configuration docs still say the default retains prior 4.x output, and the PR description still describes the superseded 16-method/`UNKNOWN` implementation. This can silently break dashboards or alerts keyed by custom HTTP method labels. Update the compatibility documentation and the release note/overview to state that method normalization applies in both modes.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigBlueprint.java
```diff
@@ -72,6 +72,19 @@ interface AutoHttpMetricsConfigBlueprint {
...
+    /**
+     * HTTP methods reported with their actual automatic metrics method attribute;
+     * methods not in this list are reported as {@code _OTHER}, conforming to the OpenTelemetry semantic conventions.
+     * <p>
+     * Method names are canonicalized using Helidon's HTTP method model.
+     *
+     * @return known HTTP methods
+     */
+    @Option.Configured
+    @Option.DefaultCode("new java.util.ArrayList<>(AutoHttpMetricsConfigSupport.DEFAULT_KNOWN_METHODS)")
```
tomas-langer, line 89: `known-methods` is a full replacement, but the public Javadoc and generated configuration reference neither list the ten defaults nor say that configuring this property replaces them. A user adding one extension method can therefore turn standard methods such as `GET` into `_OTHER`. Document the exact default set and the full-override semantics here, then regenerate the configuration reference.

webserver/observe/telemetry/metrics/src/main/java/io/helidon/webserver/observe/telemetry/metrics/OpenTelemetryMetricsHttpSemanticConventions.java
tjquinno: Fixed.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigBlueprint.java
tjquinno: Fixed.

tomas-langer: I do not see the fix for this. There should be a list of methods used as the default ones, and there should be a clear statement that this replaces the default list if configured.

tomas-langer (approved):

Issue #3686, opened 2021-12-01
HTTP/3 Support


tomas-langer:
Investigate support for HTTP/3.
Should be available with Netty 5

zimmi:
This would allow helidon to function as a [WebTransport](https://developer.mozilla.org/en-US/docs/Web/API/WebTransport) server.

barchetta:
See also #5629

tomas-langer:
Plan (to be updated):

- create listener abstraction - bindings for `TCP` (http/2, http/1.1, grpc, websocket) and `QUIC` (http/3 and possible additional protocols)
- create abstraction for quic (quic module with streams, packets etc.), webclient/quic, webserver/quic
- create support for quic "sub-protocols", http/3 being the first (and only one for now) implemented 
- create abstraction for http/3 (http/http3, webclient/http3, webserver/http3)
- move shared code to common (such as huffman encoding that is used by both http/2 and http/3)

Goals:
- have an HTTP/3 server
- have an HTTP/3 client
- support for `Alt-Svc` handling (both server and client)
- shared routing between HTTP versions
- single listener (host + port) serving either TCP, QUIC, or both

Non-Goals:
- additional protocols (WebTransport, Media over QUIC) - these would be follow ups)
- plaintext (only TLS)
- additional datagram options (i.e. only HTTP/3 will be exposed to users for routing for now)

Blockers:
- JDK does not expose security primitives for quic TLS establishment. We would need to implement this ourself - this is a major task requiring a lot of reviews, with uncertainty if we can fit such a module into Helidon

tomas-langer:
Relevant issue - will be also handled as a pre-requisite: https://github.com/helidon-io/helidon/issues/7278

PR #12246, opened 2026-08-08
Fix cross-package blueprint schema resolution

tomas-langer:
Pre-requisite for #3686

Uses the declared option type's package when resolving a same-round generated blueprint for configured schema metadata. Retains the enclosing blueprint package fallback for unqualified runtime-factory types and adds a cross-package regression test.

### DIFF
--- a/builder/tests/configured/src/test/java/io/helidon/builder/tests/configured/SchemaGeneratorTest.java
+++ b/builder/tests/configured/src/test/java/io/helidon/builder/tests/configured/SchemaGeneratorTest.java
@@ -1689,14 +1689,14 @@ class SchemaGeneratorTest {
     }
 
     @Test
-    void testPrototypeForwardReference() throws IOException {
+    void testPrototypeCrossPackageForwardReference() throws IOException {
         var result = TestCompiler.builder()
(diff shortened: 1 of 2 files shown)


builder/tests/configured/src/test/java/io/helidon/builder/tests/configured/SchemaGeneratorTest.java
```diff
@@ -1689,14 +1689,14 @@ interface AcmeLoggingConfigBlueprint extends Prototype.Factory<AcmeLogging> {
     }
 
     @Test
-    void testPrototypeForwardReference() throws IOException {
+    void testPrototypeCrossPackageForwardReference() throws IOException {
```
romain-grecourt, line 1692: Is the old test not relevant anymore ?

tomas-langer: The test is just expanded to use a different package, originally it was using the same package. So ittests both the original assertions, and a new one that it works when in another package.

romain-grecourt (approved):

PR #12250, opened 2026-08-09
Accept parsed authority for SNI validation

tomas-langer:
Pre-requisite for #3686

Use parsed `UriAuthority` values for SNI validation while retaining the string-based helper for HTTP/1 and HTTP/2. This lets protocol implementations pass parsed authority metadata without reparsing it.


danielkec (approved):
Reviewed exact head 62388e2e9750a686165d521c5570e25338dbb08c; no introduced blocking findings.

PR #12251, opened 2026-08-10
Support protocol-owned SSE entity streams

tomas-langer:
Pre-requisite for #3686

Allow SSE sinks to use a protocol-supplied entity output stream when the protocol cannot expose its connection data writer directly.

Let the protocol run final response preparation immediately before constructing response filters and content encoding.

Close the entity stream before committing the response so encoding wrappers can emit their final bytes, while preserving the existing HTTP/1 behavior.

### DIFF
--- a/webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
+++ b/webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
@@ -73,6 +73,8 @@ class DataWriterSseSink implements SseSink {
     private final Runnable closeRunnable;
     private final ConnectionContext ctx;
     private final OutputStream outputStream;
+    private final boolean closeServerSocket;
+    private final boolean closeOutputStreamBeforeCommit;
 
     DataWriterSseSink(SinkProviderContext context) {
         this.response = context.serverResponse();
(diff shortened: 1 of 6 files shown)


webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
```diff
@@ -73,30 +73,38 @@ class DataWriterSseSink implements SseSink {
...
         this.ctx = context.connectionContext();
         this.mediaContext = ctx.listenerContext().mediaContext();
         this.closeRunnable = context.closeRunnable();
 
-        // output stream to write headers
-        OutputStream headersOutputStream = new DataWriterOutputStream(ctx.dataWriter());
+        prepareResponseHeaders(true);
+        Optional<OutputStream> entityOutputStream = context.entityOutputStream(() -> prepareResponseHeaders(false));
+        if (entityOutputStream.isPresent()) {
+            this.outputStream = entityOutputStream.orElseThrow();
```
tvallin, line 88: Please preserve eager SSE header emission on this path. `entityOutputStream(...)` promises a decorated entity stream but does not require headers to be emitted before returning, and this branch has no header-flush hook. Helidon streaming output streams can defer HEADERS until body data or commit, so an SSE handler waiting before its first event leaves the client waiting for the initial 200 response. The 4.x SSE-over-HTTP/2 path uses an explicit `flushHeaders()` hook for this case. Please add equivalent SPI support and delayed-first-event coverage.

webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java
```diff
@@ -184,6 +190,26 @@ void writeStatusAndHeaders(OutputStream headersOutputStream) {
...
+                || ct != null && !CONTENT_TYPE_EVENT_STREAM.values().equals(ct.mediaType().text())) {
+            throw new IllegalStateException("ServerResponse instance cannot be used to create SseSink");
+        }
+
+        if (ct == null) {
+            headers.add(CONTENT_TYPE_EVENT_STREAM);
+        }
+        headers.set(CACHE_NO_CACHE_ONLY);
+        if (addDateHeader && !headers.contains(HeaderNames.DATE)) {
+            headers.set(create(HeaderNames.DATE, DateTime.rfc1123String()));
```
tvallin: Please create this as a changing header: `create(HeaderNames.DATE, true, false, DateTime.rfc1123String())`. The two-argument overload sets `changing=false`, so protocol encoders treat each generated Date value as indexable. This churns HPACK/QPACK dynamic tables; all existing `Http2ServerResponse` Date paths use `changing=true`.

tvallin (approved):

PR #12252, opened 2026-08-10
4.x: Bump LangChain4j to 1.18.1

### LINKED ISSUE #9389: 4.x: test: LraDisabledDiscoveryResourceTest.testLraCompensate
Seen on helidon-4.1.x branch:

```
2024-10-14T16:18:11.2569645Z [INFO] Running io.helidon.microprofile.testing.lra.NonLraResourceTest
2024-10-14T16:18:13.2095294Z [INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.946 s - in io.helidon.microprofile.testing.lra.NonLraResourceTest
2024-10-14T16:18:13.2101245Z [INFO] Running io.helidon.microprofile.testing.lra.LraDisabledDiscoveryResourceTest
2024-10-14T16:18:13.8506329Z [ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.637 s <<< FAILURE! - in io.helidon.microprofile.testing.lra.LraDisabledDiscoveryResourceTest
2024-10-14T16:18:13.8509086Z [ERROR] io.helidon.microprofile.testing.lra.LraDisabledDiscoveryResourceTest.testLraCompensate  Time elapsed: 0.317 s  <<< FAILURE!
2024-10-14T16:18:13.8510604Z java.lang.AssertionError: 
2024-10-14T16:18:13.8510955Z 
2024-10-14T16:18:13.8511147Z Expected: is <Cancelled>
2024-10-14T16:18:13.8511646Z      but: was <Cancelling>
2024-10-14T16:18:13.8512458Z 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
2024-10-14T16:18:13.8513470Z 	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:8)
2024-10-14T16:18:13.8515173Z 	at io.helidon.microprofile.testing.lra.LraDisabledDiscoveryResourceTest.testLraCompensate(LraDisabledDiscoveryResourceTest.java:117)
2024-10-14T16:18:13.8517429Z 	at io.helidon.microprofile.testing.lra.LraDisabledDiscoveryResourceTest$Proxy$_$$_WeldClientProxy.testLraCompensate(Unknown Source)
```

danielkec:
### Description

Backport of https://github.com/helidon-io/helidon-extensions/pull/114 to `helidon-4.x`.

Bumps LangChain4j core to `1.18.1` and community integrations to `1.18.0-beta28`.

Compatibility additions for this backport:

- Preserves the deprecated OpenAI image `style` and `response-format` configuration surface as no-op compatibility options.
- Preserves legacy Cohere `Proxy` configuration, including SOCKS, while preferring an explicitly configured `HttpClientBuilder`.
- Updates generated model lifecycle handling while retaining caller ownership of supplied OCI clients.
- Adapts Agentic code generation to the new `AgentConfigurator` API.

### Additional fix

- Fixes [#9389](https://github.com/helidon-io/helidon/issues/9389) by waiting for asynchronous LRA compensation to reach its terminal state before asserting the result.

### Documentation

No separate documentation change.

Previously generated Agentic service bytecode must be regenerated because LangChain4j removed the upstream static
overload it invokes. Direct callers of the removed LangChain4j OpenAI image builder methods must migrate; the Helidon
configuration API remains compatible.

### Testing

- Full 17-module LangChain4j reactor with tests, Javadocs, and SpotBugs
- OpenAI: 9 tests; Cohere: 8 tests; OCI GenAI: 12 tests
- LRA: complete 8-test module and 2 fresh compensation-test runs under JDK 21
- JApiCmp against published Helidon 4.5.2 provider artifacts: no removed public or protected APIs and no binary incompatibilities
- Copyright, checkstyle, and `git diff --check`


dependencies/pom.xml
```diff
@@ -109,8 +109,8 @@
         <version.lib.kafka>3.9.2</version.lib.kafka>
         <!-- Kotlin: dependency convergence requirement, we never use this directly -->
         <version.lib.kotlin>1.9.10</version.lib.kotlin>
-        <version.lib.langchain4j>1.12.1</version.lib.langchain4j>
-        <version.lib.langchain4j-community>1.12.1-beta21</version.lib.langchain4j-community>
+        <version.lib.langchain4j>1.18.1</version.lib.langchain4j>
```
danielkec, line 113: This upgrade brings incompatible JNA versions into the root reactor:

```text
langchain4j-embeddings:1.18.1-beta28 -> ai.djl:api:0.36.0 -> jna:5.17.0
coherence-hnsw:25.03.1 -> jna:5.14.0
```

`mvn -T8 -DskipTests -Ptests install` therefore fails Maven Enforcer dependency convergence in `helidon-all`, so the build cannot complete.

integrations/langchain4j/providers/open-ai/src/main/java/io/helidon/integrations/langchain4j/providers/openai/OpenAiImageLc4jProvider.java
```diff
@@ -0,0 +1,51 @@
...
+
+@AiProvider.ConfigKey("open-ai")
+@AiProvider.ModelConfig(value = OpenAiImageModel.class)
+interface OpenAiImageLc4jProvider {
+
+    /**
+     * Legacy image style option retained for configuration compatibility.
+     *
+     * @return configured image style
+     * @deprecated this option is no longer supported by LangChain4j and is ignored
```
danielkec, line 34: `docs/se/ai/langchain4j/open-ai.md` still describes `style` and image `response-format` as effective request options, although both compatibility properties are ignored here. Users can therefore configure values that have no effect; regenerating the configuration docs moves both under “Deprecated Options” and leaves the tracked docs dirty, so the docs clean-tree gate will also fail once the reactor reaches it.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/ModelFactoryCodegen.java
```diff
@@ -200,32 +221,219 @@ public List<Service.QualifiedInstance<OciGenAiChatModel>> services() {
...
+                .addContent(LIST)
+                .addContentLine(".of();")
+                .decreaseContentPadding()
+                .addContentLine("}")
+                .addContentLine("if (modelsToClose == null) {")
+                .increaseContentPadding()
+                .addContentLine("return;")
                 .decreaseContentPadding()
+                .addContentLine("}")
+                .addContentLine("var failure = closeModels(modelsToClose, null);")
```
Verdent: [P1] Closing models synchronously here can deadlock registry shutdown for an owned OCI streaming model. `ScopedRegistryImpl.deactivate()` holds `serviceProvidersLock.writeLock()` while invoking `@PreDestroy`, while LangChain4j community 1.18.0-beta28 waits without a timeout for active streaming callbacks to finish. If a callback performs `registry.get(...)` after shutdown takes the write lock, the callback waits for the read lock and shutdown waits for the callback. Reproduction: pause an owned streaming handler, start `ServiceRegistryManager.shutdown()`, then call `registry.get(...)` from the handler; neither thread completes.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/ModelFactoryCodegen.java
```diff
@@ -200,32 +221,219 @@ public List<Service.QualifiedInstance<OciGenAiChatModel>> services() {
...
+                .addContentLine("}")
+                .addContentLine("return services;")
+                .decreaseContentPadding()
+                .addContentLine("}")
+                .build();
+    }
+
+    private static Method preDestroyMethod() {
+        return Method.builder()
+                .addAnnotation(Annotation.create(SERVICE_ANNOTATION_PRE_DESTROY))
```
Verdent: [P1] This `@PreDestroy` belongs to the generated model factory, whose default descriptor weight is 98. `ScopedRegistryImpl` orders equal-run-level shutdown by ascending weight, so this closes models before ordinary weight-100 consumers are destroyed. A consumer that uses its injected model from its own `@PreDestroy` receives an already-closed model. `OciGenAiModelFactoryLifecycleTest` invokes the factory directly and does not exercise registry-managed consumer ordering.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/ModelFactoryCodegen.java
```diff
@@ -200,32 +221,219 @@ public List<Service.QualifiedInstance<OciGenAiChatModel>> services() {
...
+                .addContentLine("services = completedServices;")
+                .addContentLine("ownedModels = completedModels;")
+                .decreaseContentPadding()
+                .addContent("} catch (")
+                .addContent(RuntimeException.class)
+                .addContent(" | ")
+                .addContent(Error.class)
+                .addContentLine(" e) {")
+                .increaseContentPadding()
+                .addContentLine("closeModels(createdModels, e);")
```
Verdent: [P1] When rollback cleanup itself fails, this catch leaves `services` null and discards the `createdModels` reference. A later `services()` call re-enters initialization, so repeated resolution can construct more owned models while each failed-close instance becomes unreachable. `OciGenAiModelFactoryLifecycleTest#closesEarlierOwnedModelsWhenLaterConstructionFails` calls `services()` only once and therefore does not cover this retry.

integrations/langchain4j/providers/cohere/src/main/java/io/helidon/integrations/langchain4j/providers/cohere/CohereScoringLc4jProvider.java
```diff
@@ -23,19 +23,48 @@
...
+
+    /**
+     * Customizes the model builder to preserve the legacy proxy configuration through the new LangChain4j HTTP client
+     * abstraction.
+     *
+     * @return partially configured LangChain4j model builder
+     */
+    default CohereScoringModel.CohereScoringModelBuilder configuredBuilder() {
+        var modelBuilder = CohereScoringModel.builder();
+        httpClientBuilder()
```
Verdent, line 67: [P1] `httpClientBuilder()` can be populated implicitly by service discovery, not only by explicit configuration. This precedence therefore ignores an explicitly named legacy proxy whenever any `HttpClientBuilder` service is discoverable, bypassing the configured proxy/egress policy. `testCustomProxy` already runs with `MockHttpClientFactory` registered, but checks only that the proxy property exists and that model construction succeeds; it never verifies the selected transport.

docs/config/io.helidon.integrations.langchain4j.providers.cohere.CohereEmbeddingModelConfig.md
```diff
@@ -30,25 +30,25 @@ Configuration for LangChain4j model CohereEmbeddingModel
...
 </td>
 <td>
 </td>
-<td>Generated from <code>dev.<wbr>langchain4j.<wbr>model.<wbr>cohere.<wbr>Cohere<wbr>Embedding<wbr>Model.<wbr>Cohere<wbr>Embedding<wbr>Model<wbr>Builder#<wbr>logRequests(<wbr>java.<wbr>lang.<wbr>Boolean)</code></td>
+<td>Generated from <code>dev.<wbr>langchain4j.<wbr>model.<wbr>cohere.<wbr>Cohere<wbr>Embedding<wbr>Model.<wbr>Cohere<wbr>Embedding<wbr>Model<wbr>Builder#<wbr>http<wbr>Client<wbr>Builder(<wbr>dev.<wbr>langchain4j.<wbr>http.<wbr>client.<wbr>Http<wbr>Client<wbr>Builder)</code></td>
 </tr>
 <tr>
 <td>
-<code>api-<wbr>key</code>
+<code>listeners</code>
```
Verdent, line 44: [P2] This documented option is not usable from configuration. `IntrospectionBlueprintBuilder` classifies `ChatModelListener` but not `EmbeddingModelListener` as `@Option.RegistryService`, so generated code calls `cfg.as(EmbeddingModelListener.class)`. The repository has no mapper for that interface; a non-empty `listeners` value fails during config mapping. The OpenAI embedding configuration has the same generated behavior.

docs/config/io.helidon.integrations.langchain4j.providers.oci.genai.OciGenAiStreamingChatModelConfig.md
```diff
@@ -244,6 +267,17 @@ Configuration for LangChain4j model OciGenAiStreamingChatModel
 </tr>
 <tr>
 <td>
+<code>executor-<wbr>service</code>
```
Verdent, line 282: [P2] This documented option is emitted as an ordinary configured value because `ExecutorService` is not classified as a registry service by `IntrospectionBlueprintBuilder`. Generated code calls `Config.as(ExecutorService.class)`, and no mapper for that interface exists, so a configured executor cannot be resolved. The Cohere OCI streaming model has the same behavior.

integrations/langchain4j/providers/oracle-genai/src/main/java/io/helidon/integrations/langchain4j/providers/oci/genai/OciGenAiLc4jProvider.java
```diff
@@ -81,6 +78,11 @@ interface OciGenAiLc4jProvider {
     @Option.RegistryService
     Optional<GenerativeAiInferenceClient> genAiClient();
 
+    @Override
+    default boolean closeModelOnShutdown() {
+        return genAiClient().isEmpty();
```
Verdent: [P2] This makes generated shutdown close internally constructed synchronous models, but LangChain4j community 1.18.0-beta28 accounts for active operations only in `ociChatAsync`; `ociChat` has no equivalent lifecycle guard. Concurrent registry shutdown can therefore close the OCI client beneath an in-flight synchronous chat request.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/LlmModelBlueprintBuilder.java
```diff
@@ -79,12 +80,15 @@ protected String configRoot() {
 
     @Override
     protected Map<String, TypedElementInfo> resolveOverriddenProperties() {
-        return parentTypeInfo()
-                .elementInfo()
-                .stream()
+        var lineage = new ArrayList<TypeInfo>();
+        ModelType.allParents(parentTypeInfo(), lineage);
```
Verdent, line 84: [P2] `allParents()` returns depth-first traversal order, which is not globally most-specific for an interface diamond. For `Provider extends Left, Right` where both inherit `Shared` and `Right` overrides `Shared.endpoint`, the lineage visits `Shared` through `Left` before `Right`; the merge below keeps `Shared` and discards the more-specific `Right` metadata. Generated `@CustomBuilderMapping`, nested-config metadata, or requiredness can therefore come from the wrong declaration.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/ModelFactoryCodegen.java
danielkec: Fixed in b527019458. Scoped registry shutdown now marks the scope inactive and snapshots activators under the write lock, then invokes deactivation after releasing it. A deterministic regression test forces an uncached registry lookup during @PreDestroy and verifies both lookup and shutdown complete.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/ModelFactoryCodegen.java
danielkec: Fixed in b527019458. Generated model cleanup now belongs to a separate lifecycle coordinator at minimum run level and maximum weight, so consumers deactivate before owned models close. Registry-managed tests cover ordinary consumers, higher-run-level consumers, and same-run-level higher-weight consumers.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/ModelFactoryCodegen.java
danielkec: Fixed in b527019458. A rollback close failure is retained as terminal CLEANUP_FAILED state; subsequent services() and shutdown calls rethrow the stored failure without rebuilding models or retrying close. The lifecycle suite covers repeated access and shutdown after the failure.

integrations/langchain4j/providers/cohere/src/main/java/io/helidon/integrations/langchain4j/providers/cohere/CohereScoringLc4jProvider.java
danielkec: Fixed in a95aca177c. Cohere scoring now tracks explicit versus discovered proxy/HTTP-client values across setter order: explicit HTTP wins, explicit proxy beats discovered HTTP, and discovery-versus-discovery retains HTTP precedence. The tests cover direct and named values, both-default discovery, setter ordering, and a real SOCKS request.

docs/config/io.helidon.integrations.langchain4j.providers.cohere.CohereEmbeddingModelConfig.md
danielkec: Fixed in a95aca177c. EmbeddingModelListener, including List<EmbeddingModelListener>, is now classified as a registry service. Named-listener tests verify the exact registry instance, and regenerated configuration references include listener service discovery.

docs/config/io.helidon.integrations.langchain4j.providers.oci.genai.OciGenAiStreamingChatModelConfig.md
danielkec: Fixed in a95aca177c. ExecutorService is now classified as a registry service, so generated streaming configuration uses registry resolution instead of Config.as(...). A named-executor test verifies the exact service instance, and both OCI streaming configuration references were regenerated.

integrations/langchain4j/providers/oracle-genai/src/main/java/io/helidon/integrations/langchain4j/providers/oci/genai/OciGenAiLc4jProvider.java
danielkec: Fixed in a95aca177c. Both synchronous OCI configurations now opt out of generated factory closing because upstream does not account for synchronous chat/doChat operations. Streaming models retain owned cleanup and their active-operation coordination.

integrations/langchain4j/codegen/src/main/java/io/helidon/integrations/langchain4j/codegen/LlmModelBlueprintBuilder.java
danielkec: Fixed in a95aca177c. ModelType.allParents now deduplicates nodes and uses deterministic child-before-parent topological ordering, so more-specific sibling metadata precedes shared ancestors. Deep-hierarchy and interface-diamond tests verify traversal and the selected configured-property metadata.

service/registry/src/main/java/io/helidon/service/registry/ScopedRegistryImpl.java
```diff
@@ -77,18 +80,29 @@ public void activate() {
...
-            serviceProvidersLock.writeLock().lock();
+            while (deactivating) {
+                deactivationCompleted.awaitUninterruptibly();
+            }
             if (!active) {
                 return;
             }
 
-            List<Activator<?>> toShutdown = activators.values()
+            active = false;
```
tvallin: [P1] Marking the scope inactive here also rejects dependency lookups made by the `@PreDestroy` callbacks below. This already breaks both failed CI jobs: the declarative metrics `GaugeRegistrar.preDestroy()` resolves an injected `ServiceSupply`, which reaches `checkActive()` and throws `ScopeNotActiveException`. Please preserve access to existing dependencies for the deactivation thread while still rejecting outside concurrent lookups, and add a regression test covering a lookup from within `preDestroy()`.

integrations/langchain4j/providers/cohere/src/main/java/module-info.java
```diff
@@ -26,16 +26,22 @@
...
 @Features.Preview
 module io.helidon.integrations.langchain4j.providers.cohere {
+    requires java.net.http;
+
     requires static io.helidon.common.features.api;
     requires static io.helidon.config.metadata;
 
     requires langchain4j.cohere;
     requires langchain4j.core;
+    requires langchain4j.http.client;
```
tvallin: [P2] The newly generated public `CohereEmbeddingModelConfig` and `CohereScoringModelConfig` APIs expose `dev.langchain4j.http.client.HttpClientBuilder`. A downstream named module requiring only this Cohere module will not read `langchain4j.http.client` through this plain `requires`. Please make this `requires transitive` per the JPMS guideline and add or extend a module-path consumer compile test.

service/registry/src/main/java/io/helidon/service/registry/ScopedRegistryImpl.java
danielkec: Fixed in 54d1917203. The service-registry shutdown changes are reverted, so @PreDestroy callbacks retain the existing registry behavior and the declarative metrics regression is removed. The original OCI streaming deadlock is now handled inside the LangChain4j integration by opting both OCI streaming provider configurations out of registry-owned model close. Verified with the declarative metrics test (6/6), the 47-module service-registry reactor including registry tests (16/16), and the full 14-module LangChain4j reactor.

integrations/langchain4j/providers/cohere/src/main/java/module-info.java
danielkec: Fixed in cb624d1718. The Cohere module now re-exports langchain4j.http.client with requires transitive. A Maven Invoker regression compiles a standalone named module that depends on and requires only the Cohere provider, then consumes HttpClientBuilder through both generated configuration APIs. Before the descriptor change it failed because the consumer did not read langchain4j.http.client; afterward clean verify passes all 13 Cohere tests and the JPMS consumer build.

tomas-langer (commented):
I found the following issues in the current PR head:

1. `io.helidon.integrations.langchain4j.codegen.IntrospectionBlueprintBuilder:146` now makes HTTP clients, embedding listeners, and executors registry-backed by default, but the generated factory at `io.helidon.integrations.langchain4j.codegen.ModelFactoryCodegen:167` receives no owning `ServiceRegistry` and builds model configs without one at line 933. `RegistryBuilderSupport` consequently falls back to `GlobalServiceRegistry`. A factory owned by an isolated registry can therefore resolve an HTTP client, listener, or executor from an unrelated global registry, crossing registry and lifecycle boundaries and potentially routing model data or authentication-bearing requests through the wrong service.

2. `io.helidon.integrations.langchain4j.codegen.ModelFactoryCodegen:73` assigns generated lifecycle coordinators `Double.MIN_VALUE`. This is the smallest positive value, while `io.helidon.service.registry.ScopedRegistryImpl:158` deactivates services in descending run-level order. A consumer at run level `0.0` or below is consequently destroyed after the coordinator, so its `@PreDestroy` can observe an already-closed model. The source PR now uses a genuinely negative terminal value, making the target backport behavior divergent.

3. The generated `closeModels` method at `io.helidon.integrations.langchain4j.codegen.ModelFactoryCodegen:732` catches every throwable from `model.close()` and aggregates it without restoring the interrupt flag when the throwable is `InterruptedException`. A blocking close normally clears that flag before throwing, so registry shutdown or rollback can consume cancellation state and continue with `Thread.currentThread().isInterrupted()` false.

4. `io.helidon.integrations.langchain4j.providers.oci.genai.OciGenAiLc4jProvider:81` and the corresponding Cohere and streaming providers unconditionally return `false` from `closeModelOnShutdown()`. This includes configurations where no synchronous or asynchronous OCI client was supplied and the model created its own resources. Those internally owned clients/models are omitted from generated shutdown and have no cleanup owner, allowing OCI connection and execution resources to survive registry shutdown.

5. `io.helidon.integrations.langchain4j.codegen.ModelFactoryCodegen:388` still constructs every named model while `ServicesFactory.services()` advertises instances. If one model construction fails, `ServicesFactoryActivator` can already be in its active phase with no target instances installed, causing later registry lookups for every model from that factory to remain empty instead of retrying or isolating the bad model. The current source PR advertises lazy qualified instances and includes registry-level retry coverage, but that lifecycle change is absent from this backport.

6. `docs/se/ai/langchain4j/langchain4j.md:447` describes Agentic workflows without disclosing the migration caused by the dependency and generator changes. `io.helidon.integrations.langchain4j.codegen.AgentCodegen:153` now emits the new `AgentConfigurator` API because the previous upstream overload was removed. Previously generated Agentic artifacts can therefore fail after upgrading unless they are regenerated, but that compatibility requirement is absent from user-facing migration documentation.

7. `docs/se/ai/langchain4j/cohere.md:79` still omits the new `http-client-builder` option and the preserved, deprecated `proxy` path. The production code adds SOCKS adaptation and non-trivial precedence between explicitly configured and discovered HTTP clients and proxies. Existing 4.x users cannot determine from the authored guide whether their proxy remains supported, what replaces it, or which transport setting wins when both are present.

8. `docs/se/ai/langchain4j/oci-genai.md:155` documents only the synchronous OCI client for streaming models, while the PR adds public `gen-ai-async-client` and `executor-service` configuration. The same change alters model/client shutdown ownership. Users cannot determine which client is consumed by streaming models or whether Helidon owns and closes the configured resources.

9. `docs/config/io.helidon.langchain4j.providers.OpenAiConfig.md:433` presents provider-level `response-format` as an active JSON-mode option, while the same merged page includes the image configuration for which this PR makes `response-format` deprecated and ignored. The aggregate generated reference therefore assigns one key two incompatible user-visible contracts and can direct image-model users to expect behavior that no longer occurs.

10. `io.helidon.integrations.langchain4j.providers.cohere.CohereScoringConfigSupport:43` introduces intrinsic monitor locking around a static `WeakHashMap` in five methods. No third-party contract requires monitor locking here, so the new code violates `DEV-GUIDELINES.md` Rules 1.4 and 1.4.1 and makes configuration precedence depend on a global monitor and mutable side table.

11. `io.helidon.integrations.langchain4j.providers.cohere.CohereHttpClientSupport:173` catches `IOException` and throws `new RuntimeException(e)`. Direct use of `RuntimeException` violates `DEV-GUIDELINES.md` Rule 1.1.1 and removes module-specific diagnostic meaning from failures in the new SOCKS transport adapter.

12. `io.helidon.integrations.langchain4j.providers.cohere.CohereHttpClientSupport:116` uses fully qualified `okhttp3.OkHttpClient` names in ordinary code despite having no simple-name conflict. The new production code therefore violates `DEV-GUIDELINES.md` Rule 3.1.

13. `io.helidon.integrations.langchain4j.codegen.ModelTypeTest:52` uses JUnit `assertEquals` and `assertSame` for its new assertions. Helidon permits only `assertAll` and `assertThrows` as JUnit assertion exceptions and otherwise requires Hamcrest, so this violates `DEV-GUIDELINES.md` Rules 10.4, 10.5.1, and 10.6 and diverges from the source PR's current test.

danielkec:
Addressed Tomas review items 10-13 in 930ab9a7b6.

- 10: replaced intrinsic monitor locking around Cohere scoring setter provenance with an explicit ReentrantLock.
- 11: wrap non-timeout SOCKS I/O failures in LangChain4jException while preserving the cause.
- 12: import and use the simple OkHttpClient name.
- 13: use Hamcrest assertions in ModelTypeTest and declare the test-scoped Hamcrest dependency.

Validation passed under JDK 21: codegen 2/2, Cohere 13/13, JPMS consumer 1/1, the full 14-module LangChain4j reactor, copyright, checkstyle, and git diff --check.

These findings were posted in a general review body rather than inline threads, so GitHub provides no individual thread-resolution control for them.

danielkec:
Addressed Tomas review item 2 in e3cb1f7e77.

This is the exact semantic port of helidon-extensions commit 8a7a3f8: generated lifecycle coordinators now use the finite terminal run level -Double.MAX_VALUE, emitted as a numeric annotation value, while retaining Double.MAX_VALUE weight. Regression coverage now checks descriptor metadata, a real run-level 0.0 consumer, and equal-run-level weight ordering.

Validation passed under JDK 21: the generated descriptor contains @RunLevel(-1.7976931348623157E308D), OCI lifecycle tests pass 17/17, the clean 14-module LangChain4j install succeeds, and copyright, checkstyle, and git diff --check pass.

This finding is part of a general review body rather than an inline thread, so GitHub provides no individual thread-resolution control.

danielkec:
Addressed Tomas review items 3, 5, 7, 8, and 9 in d93b76162a.

- 3: generated model cleanup now restores the interrupt flag after collecting an `InterruptedException`, with rollback and shutdown coverage.
- 5: generated factories advertise stable lazy qualified instances, omit disabled models, retry failed initialization, atomically publish an immutable model map, and resolve named models with an O(1) lookup. Registry retry, concurrent publication, retained-reference, disabled-model, and many-model tests cover the behavior.
- 7: the Cohere guide now documents `http-client-builder`, the deprecated direct/HTTP/SOCKS `proxy` compatibility path, discovery controls, and explicit-versus-discovered precedence.
- 8: the OCI guide and streaming provider metadata now document asynchronous client and executor configuration, client selection, resource ownership, and the 4.x shutdown limitation.
- 9: OpenAI chat/image `response-format` metadata now states both contracts, and all affected individual and aggregate config references were regenerated.

Item 4 remains intentionally unresolved. Copying the forward-port ownership predicates into 4.x would let synchronous close race an untracked active call and let streaming close wait for a callback that still needs the registry. All four OCI providers therefore remain opted out of registry-driven model close, and the guide now documents explicit close responsibility for internally created clients.

Validation passed under JDK 21: clean 14-module LangChain4j install, OCI lifecycle 23/23, Cohere 13/13, Cohere JPMS consumer 1/1, Coherence integration 1/1, config-doc generation, copyright, checkstyle, and `git diff --check`.

These findings are part of a general review body rather than inline threads, so GitHub provides no individual thread-resolution control.

danielkec:
Addressed Tomas review item 6 in 310d21c158.

The authored Agents guide now requires a clean rebuild, such as `mvn clean package`, after upgrading to the LangChain4j 1.18.1-based integration. It explains that stale generated source can fail compilation and stale compiled Agentic services can throw `NoSuchMethodError` because they invoke the removed `AgenticServices.createAgenticSystem(...)` overload; current code generation uses `AgentConfigurator`.

Validation passed under JDK 21: clean Agentic tests 4/4, clean Agentic MCP tests 4/4, all seven regenerated `*__AiAgent.java` sources use `new AgenticServices.AgentConfigurator(...)`, and copyright, checkstyle, and `git diff --check` pass.

This finding is part of a general review body rather than an inline thread, so GitHub provides no individual thread-resolution control.

danielkec:
Addressed Tomas review item 1 in 600da0c413.

Generated model factories now constructor-inject the `ServiceRegistry` that owns the factory and pass it to each generated model configuration builder. Generated service descriptors declare and resolve that registry dependency, so `RegistryBuilderSupport` no longer falls back to an unrelated global registry for HTTP clients, listeners, executors, or other registry-backed options.

The regression installs a deliberately wrong global registry alongside a separate non-global owning registry. It verifies that both default and named `ExecutorService` discovery use the owning registry, that the generated descriptor declares the registry dependency, and that the previous global registry is restored after isolated shutdown.

Validation passed under JDK 21: focused OCI lifecycle 24/24, the full 17-module LangChain4j `-Ptests` clean install, all 25 generated model factories contain registry injection and builder propagation, copyright, checkstyle, and `git diff --check`.

This finding is part of a general review body rather than an inline thread, so GitHub provides no individual thread-resolution control.

tomas-langer (approved):

PR #12253, opened 2026-08-10
Add class-level declarative media types

### LINKED ISSUE #12091: Add MediaType support at class level
Helidon Version:

## Enhancement Description

When using the HTTP Declarative API, to specify what Media type is consumed and produced by a method in the , the right annotations must be added e.g.

```
@Http.POST
       @Http.Consumes(MediaTypes.APPLICATION_JSON_VALUE)
       @Http.Produces(MediaTypes.APPLICATION_JSON_VALUE)
       @RestServer.Status(Status.CREATED_201_CODE)
       Order createOrder(@Http.Entity CreateOrderRequest request) {
           return service.createOrder(request);
       }
```
However, these annotations are marked as method only, forcing the developer to repeatedly write the same tedious annotations for every method in that class that consumes and produces JSON. 

A useful enhancement is to be able to define these once at the class level such that this becomes the default. A user should also be able to explicitly override this default at the method level.

tomas-langer:
Resolves #12091

Allows `@Http.Consumes` and `@Http.Produces` to define type-level defaults for declarative server and client APIs. Method annotations replace those defaults, including explicit clearing, and type-level `Consumes` applies only to entity-bearing methods.

### DIFF
--- a/declarative/README.md
+++ b/declarative/README.md
@@ -126,6 +126,8 @@ Annotations on type:
 - `@RestServer.Header` - a header to be sent with every response from the server, defined either on the endpoint class, or on a
   method (repeatable)
 - `@RestServer.ComputedHeader` - a header to be sent with every response computed from a service
+- `@Http.Produces` - default media types produced by endpoint methods, individually overridable on a method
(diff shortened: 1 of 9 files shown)


tjquinno (approved):
LGTM

PR #12254, opened 2026-08-10
Add server Alt-Svc advertisement

tomas-langer:
Pre-requisite for #3686

Add preview Alt-Svc configuration and runtime support for HTTP/1.1 and HTTP/2 responses.

Keep advertisement opt-in, preserve application-provided headers, scope generated state per listener, and avoid Alt-Svc work on disabled response paths.

Support buffered, streaming, SSE, h2c, and error/reset paths, including listener-port fallback and explicit ports for non-port bindings.


webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConnectionSelector.java
```diff
@@ -111,7 +123,19 @@ public Set<String> supportedApplicationProtocols() {
...
+            return new Http1Connection(ctx, config.config(), upgradeProviderMap);
+        }
+
+        AltSvcConfig altSvcPrototype = altSvcConfig.get();
+        int listenerPort = ctx.localPeer().port();
+        if (altSvcPrototype.port().isEmpty() && (listenerPort < 1 || listenerPort > 65_535)) {
+            return new Http1Connection(ctx, config.config(), upgradeProviderMap);
+        }
+
+        Header header = altSvcByListener.computeIfAbsent(ctx.listenerContext(),
```
tvallin, line 136: This cache assumes a `ListenerContext` has one immutable port. Checkpoint suspend keeps the listener, context, and selector alive, but `SocketTransportBinding` clears its bound port; a `port(0)` listener can therefore bind a different ephemeral port on resume. `computeIfAbsent` then keeps advertising the pre-checkpoint endpoint. Please key or invalidate the cache using the effective port—the same pattern appears in `Http2ConnectionSelector:103` and `Http2Upgrader:102`—and add a dynamic-port suspend/resume test.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -1226,8 +1258,15 @@ private void handle() {
                                                                    outcome,
                                                                    ctx.listenerContext().config().maxPayloadSize(),
                                                                    http2Config.maxBufferedEntitySize().toBytes());
-            Http2ServerResponse response = new Http2ServerResponse(this, request,
-                                                                   http2Config.validateResponseHeaders());
+            Http2ServerResponse response;
+            if (altSvcHeader == null) {
+                response = new Http2ServerResponse(this, request, http2Config.validateResponseHeaders());
+            } else {
+                response = new Http2AltSvcServerResponse(this,
```
tvallin, line 1265: This wrapper is created only when `selectedSubProtocol == null`; selected subprotocols take the branch at line 1317. gRPC is such a subprotocol and writes its own successful `:status 200` HEADERS, so the configured `http_2.alt-svc` advertisement is absent from gRPC responses despite being protocol-level HTTP/2 configuration. Please propagate the advertisement through the subprotocol path, or explicitly narrow the documented contract, and add gRPC/custom-subprotocol coverage.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConfigBlueprint.java
```diff
@@ -113,6 +116,18 @@ interface Http1ConfigBlueprint extends HttpConfig, ProtocolConfig {
...
     List<Http1ConnectionListener> receiveListeners();
 
+    /**
+     * Explicit {@code Alt-Svc} advertisement.
+     *
+     * @return alternative service advertisement
+     */
+    @Api.Preview
+    @Option.Configured
+    @Option.PrototypedBy("io.helidon.webserver.http.AltSvcConfig")
```
tvallin, line 126: `@Option.PrototypedBy` emits the `alt-svc` option type as `AltSvcConfig`, while the factory blueprint metadata is registered under its runtime type, `AltSvc`; no `AltSvcConfig` metadata type is emitted. Because `CmResolver` uses exact type names, it cannot descend into `enabled`, `protocol`, `port`, `max-age`, and `persist`, leaving this option opaque in generated configuration reference/schema output. Please make the generated type names agree—the same issue occurs in `Http2ConfigBlueprint:168`—and add resolver-level metadata coverage.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
```diff
@@ -132,6 +131,35 @@ public void send(byte[] entityBytes, int position, int length) {
             headers.setIfAbsent(HeaderValues.create(HeaderNames.DATE, true,
                                                     false,
                                                     DateTime.rfc1123String()));
+
+            int initialStatusCode = status().code();
+            beforeSend();
```
tvallin: `isSent` is now set only after this callback, and `outputStream()` similarly sets `streamingEntity` after its callback. A listener that calls `send()` or `outputStream()` therefore sees both guards clear: an unconditional listener recurses to `StackOverflowError`, while a one-shot listener can commit an inner response before the outer path emits a second HEADERS/DATA sequence. Previously the state flags rejected this re-entry. Please establish a separate preparation/reentrancy guard before callbacks, roll it back on callback failure, and cover both nested paths.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1AltSvcServerResponse.java
```diff
@@ -0,0 +1,152 @@
...
+    protected void beforeSend() {
+        super.beforeSend();
+        responsePrepared = true;
+        reconcileAltSvcHeader();
+    }
+
+    @Override
+    protected Optional<OutputStream> sinkEntityOutputStream(Runnable responsePreparation) {
+        Objects.requireNonNull(responsePreparation);
+        beforeSend();
```
tvallin: This makes application `beforeSend` listeners execute during SSE sink construction only when Alt-Svc is configured. If a listener uses the documented ability to change status—for example, to 400—`DataWriterSseSink` rejects the resulting non-200 SSE response and normal error handling emits 500. Without Alt-Svc, the same path bypasses the callback and returns 200. Please move sink response preparation into the base HTTP/1 lifecycle so configured and unconfigured responses behave consistently, and test both configurations.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConnectionSelector.java
tomas-langer: If somebody uses random ports with suspend/resume, they are out of luck. I consider this use case invalid, as you cannot rebind same port, when it is random.

tvallin (approved):

PR #12258, opened 2026-08-11
Add read-only buffer slicing support

tomas-langer:
Add an internal buffer operation that consumes an exact source range and returns an independently positioned read-only view. Read-only array data shares backing storage, while other buffer implementations use an exact defensive copy.

Pre-requisite for #3686.

### DIFF
--- a/common/buffers/src/main/java/io/helidon/common/buffers/BufferData.java
+++ b/common/buffers/src/main/java/io/helidon/common/buffers/BufferData.java
@@ -82,11 +82,11 @@ public interface BufferData {
      * Consume a range from the source and return an independently positioned read-only view of exactly that range.
      *
      * <p>Sources created with
-     * {@link io.helidon.common.buffers.BufferData#createReadOnly(byte[], int, int)} share their backing array with the
+     * {@link #createReadOnly(byte[], int, int)} share their backing array with the
      * returned view. The caller must not modify that array while either buffer is in use. All other sources use an exact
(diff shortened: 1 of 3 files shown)


common/buffers/src/main/java/io/helidon/common/buffers/BufferData.java
```diff
@@ -76,6 +78,43 @@ static BufferData createReadOnly(byte[] bytes, int offset, int length) {
...
+     * @throws IndexOutOfBoundsException if length is negative or exceeds the available source bytes
+     */
+    @Api.Internal
+    static BufferData readOnlySlice(BufferData source, int length) {
+        Objects.requireNonNull(source, "source");
+        Objects.checkFromIndexSize(0, length, source.available());
+        if (length == 0) {
+            return ReadOnlyArrayData.EMPTY;
+        }
+        if (source instanceof ReadOnlyArrayData arrayData) {
```
tvallin, line 103: `ReadOnlyArrayData` is also produced by `DataReader.getBuffer(...)` for contiguous input and by this method itself, so both sources enter this branch even though the Javadoc promises a defensive copy for anything not created with `createReadOnly(...)`. For `DataReader`, storage ownership then varies with input fragmentation: contiguous input shares the supplier array, while composite input is copied. A caller relying on the documented copy guarantee can therefore observe later mutations. Please align the implementation and contract, and cover DataReader-derived and re-sliced sources.

danielkec (approved):

PR #12259, opened 2026-08-11
Fix static content byte ranges

tomas-langer:
Pre-requisite for #12058

Correct static-content byte-range responses for ordinary cached and file-backed resources.

- Clamp an oversized suffix range to the full representation.
- Clamp an explicit range end to the representation length.
- Use the correct exclusive array end when slicing cached content at a non-zero offset.

This keeps `Content-Range`, `Content-Length`, and the emitted payload consistent.

### DIFF
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, 2023 Oracle and/or its affiliates.
+ * Copyright (c) 2022, 2026 Oracle and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/ByteRangeRequest.java
```diff
@@ -54,10 +54,10 @@ static List<ByteRangeRequest> parse(ServerRequest req, ServerResponse res, Strin
             if (secondGroup != null) {
                 long second = Long.parseLong(secondGroup);
                 if (firstGroup == null) {
-                    from = fileLength - second;
+                    from = Math.max(0, fileLength - second);
```
tvallin, line 69: For `fileLength == 0`, a valid non-zero suffix such as `bytes=-1` now produces `from == 0` and `last == -1`, so `create` reports 416. [RFC 9110 §14.1.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-14.1.2) defines such a suffix range as satisfiable even for an empty representation, and §14.2 permits ignoring `Range` when there is no content. Please return the normal empty 200 response instead and cover both file-backed and in-memory cached resources, while keeping `bytes=-0` and `bytes=0-` unsatisfiable.

tomas-langer: Addressed: zero-length resources now treat a non-zero suffix range as satisfiable and ignore the Range header, so file-backed and in-memory cached resources return the normal empty 200 response. `bytes=-0` and `bytes=0-` remain 416. The zero-length path also classifies range members before numeric conversion, so an oversized value in a mixed range set cannot escape as an internal error.

tvallin (approved):

PR #12260, opened 2026-08-11
Fix static content conditional requests

tomas-langer:
Pre-requisite for #12058

Correct static-content conditional request handling:

- require strong validator comparison for If-Match
- evaluate If-Match before If-None-Match
- ignore If-Modified-Since whenever If-None-Match is present

Add regression coverage for weak validators, wildcard handling, and conditional-header precedence.

### DIFF
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerUrlStream.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerUrlStream.java
@@ -45,13 +45,12 @@ record CachedHandlerUrlStream(MediaType mediaType, URL url) implements CachedHan
 
         URLConnection urlConnection = url.openConnection();
         long lastModified = urlConnection.getLastModified();
+        Instant modified = lastModified == 0 ? null : Instant.ofEpochMilli(lastModified);
 
-        if (lastModified != 0) {
-            StaticContentHandler.processPreconditions(String.valueOf(lastModified),
-                                                      Instant.ofEpochMilli(lastModified),
-                                                      request.headers(),
-                                                      response.headers());
-        }
+        StaticContentHandler.processPreconditions(modified == null ? null : String.valueOf(lastModified),
+                                                  modified,
+                                                  request.headers(),
+                                                  response.headers());
 
         response.headers().contentType(mediaType);
 
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerUrlStream.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerUrlStream.java
@@ -47,8 +47,10 @@ record CachedHandlerUrlStream(MediaType mediaType, URL url) implements CachedHan
         long lastModified = urlConnection.getLastModified();
 
         if (lastModified != 0) {
-            StaticContentHandler.processEtag(String.valueOf(lastModified), request.headers(), response.headers());
-            StaticContentHandler.processModifyHeaders(Instant.ofEpochMilli(lastModified), request.headers(), response.headers());
+            StaticContentHandler.processPreconditions(String.valueOf(lastModified),
+                                                      Instant.ofEpochMilli(lastModified),
+                                                      request.headers(),
+                                                      response.headers());
         }
 
         response.headers().contentType(mediaType);
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
@@ -23,6 +23,7 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.time.chrono.ChronoZonedDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -114,6 +115,7 @@ abstract class StaticContentHandler implements HttpService {
         }
 
         if (modified != null) {
+            modified = modified.truncatedTo(ChronoUnit.SECONDS);
             // Last-Modified
             setModified.accept(responseHeaders, modified);
             // If-Unmodified-Since
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
@@ -93,10 +93,11 @@ abstract class StaticContentHandler implements HttpService {
         boolean ifMatchPresent = requestHeaders.contains(HeaderNames.IF_MATCH);
         if (newEtag != null && ifMatchPresent) {
             // Process If-Match header
-            List<String> ifMatches = requestHeaders.get(HeaderNames.IF_MATCH).allValues();
+            List<String> ifMatches = requestHeaders.values(HeaderNames.IF_MATCH);
             if (!ifMatches.isEmpty()) {
                 boolean ifMatchChecked = false;
                 for (String ifMatch : ifMatches) {
+                    ifMatch = ifMatch.trim();
                     boolean wildcard = "*".equals(ifMatch);
                     boolean weak = ifMatch.startsWith("W/") || ifMatch.startsWith("w/");
                     ifMatch = unquoteETag(ifMatch);
@@ -131,8 +132,9 @@ abstract class StaticContentHandler implements HttpService {
         // Process If-None-Match header
         boolean ifNoneMatchPresent = requestHeaders.contains(HeaderNames.IF_NONE_MATCH);
         if (newEtag != null && ifNoneMatchPresent) {
-            List<String> ifNoneMatches = requestHeaders.get(HeaderNames.IF_NONE_MATCH).allValues();
+            List<String> ifNoneMatches = requestHeaders.values(HeaderNames.IF_NONE_MATCH);
             for (String ifNoneMatch : ifNoneMatches) {
+                ifNoneMatch = ifNoneMatch.trim();
                 ifNoneMatch = unquoteETag(ifNoneMatch);
                 if ("*".equals(ifNoneMatch) || ifNoneMatch.equals(etag)) {
                     // using exception to handle normal flow (same as in reactive static content)
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
@@ -166,10 +166,6 @@ abstract class StaticContentHandler implements HttpService {
                 if (entityStart < fieldLength && fieldValue.charAt(entityStart) == '"') {
                     int entityEnd = fieldValue.indexOf('"', entityStart + 1);
                     if (entityEnd < 0) {
-                        if (fieldLength - entityStart == 1) {
-                            // Preserve the existing malformed-tag failure until invalid entity tags are handled separately.
-                            unquoteETag(fieldValue.substring(entityStart));
-                        }
                         break;
                     }
 
--- a/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
+++ b/webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
@@ -150,6 +150,8 @@ abstract class StaticContentHandler implements HttpService {
     }
 
     private static boolean matchesEntityTag(Header header, String etag, boolean strongComparison) {
+        boolean matched = false;
+        boolean singleValue = header.valueCount() == 1;
         for (String fieldValue : header.allValues()) {
             int tokenStart = 0;
             int fieldLength = fieldValue.length();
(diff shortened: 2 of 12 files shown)


webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -121,6 +110,19 @@ static void processEtag(String etag, ServerRequestHeaders requestHeaders, Server
                 }
             }
         }
+
+        // Process If-None-Match header
```
tvallin: **Medium — Enforce the complete precondition order**

Every static-content handler calls `processEtag` before `processModifyHeaders`, making the effective order `If-Match` → `If-None-Match` → `If-Modified-Since` → `If-Unmodified-Since`. [RFC 9110 §13.2.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-13.2.2) requires `If-Match` → `If-Unmodified-Since` only without If-Match → `If-None-Match` → `If-Modified-Since` only without If-None-Match. Consequently, matching If-Match plus stale If-Unmodified-Since returns 412 instead of 200, while stale If-Unmodified-Since plus matching If-None-Match returns 304 instead of 412. Please coordinate the complete evaluation order and cover both combinations.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -90,27 +90,16 @@ static void processEtag(String etag, ServerRequestHeaders requestHeaders, Server
...
-            }
-        }
-
         if (requestHeaders.contains(HeaderNames.IF_MATCH)) {
             // Process If-Match header
             List<String> ifMatches = requestHeaders.get(HeaderNames.IF_MATCH).allValues();
             if (!ifMatches.isEmpty()) {
                 boolean ifMatchChecked = false;
                 for (String ifMatch : ifMatches) {
+                    boolean wildcard = "*".equals(ifMatch);
```
tvallin: **Medium — Parse entity-tag lists before comparing**

`Header.allValues()` preserves each field-line value; it does not parse the comma-separated `#entity-tag` list within one field. Against current tag `"123"`, the valid wire form `If-Match: "old", "123"` reaches this loop as one string and returns 412; the equivalent If-None-Match request returns 200 instead of 304. Please parse every list member—quote-aware because commas are valid opaque-tag characters—and add client/parser-level tests. See [RFC 9110 §§13.1.1–13.1.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.1).

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -121,6 +110,19 @@ static void processEtag(String etag, ServerRequestHeaders requestHeaders, Server
                 }
             }
         }
+
+        // Process If-None-Match header
+        if (requestHeaders.contains(HeaderNames.IF_NONE_MATCH)) {
+            List<String> ifNoneMatches = requestHeaders.get(HeaderNames.IF_NONE_MATCH).allValues();
+            for (String ifNoneMatch : ifNoneMatches) {
+                ifNoneMatch = unquoteETag(ifNoneMatch);
+                if ("*".equals(ifNoneMatch) || ifNoneMatch.equals(etag)) {
```
tvallin: **Medium — Distinguish quoted `*` tags from the wildcard**

Only bare `*` is the wildcard. Calling `unquoteETag` before this check makes the valid entity-tags `"*"` and `W/"*"` indistinguishable from it. Against current ETag `"123"`, either quoted value incorrectly returns 304 instead of 200. Preserve the wildcard decision before unquoting, as the new If-Match path does, and cover both quoted forms. See the [If-None-Match and entity-tag grammar](https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.2).

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -134,11 +136,13 @@ static void processModifyHeaders(Instant modified,
...
-        Optional<Instant> ifModSince = requestHeaders
-                .ifModifiedSince()
-                .map(ChronoZonedDateTime::toInstant);
-        if (ifModSince.isPresent() && !ifModSince.get().isBefore(modified)) {
-            throw new HttpException("Not valid for If-Modified-Since header", Status.NOT_MODIFIED_304, true);
+        if (!requestHeaders.contains(HeaderNames.IF_NONE_MATCH)) {
+            Optional<Instant> ifModSince = requestHeaders
+                    .ifModifiedSince()
+                    .map(ChronoZonedDateTime::toInstant);
+            if (ifModSince.isPresent() && !ifModSince.get().isBefore(modified)) {
```
danielkec, line 146: This comparison uses the original subsecond `modified` instant, while `Last-Modified` is emitted at whole-second precision. In a real-server round trip that copied the emitted `Last-Modified` value, `If-Modified-Since` returned 200 instead of 304; the analogous `If-Unmodified-Since` check returned 412 instead of 200.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -134,11 +136,13 @@ static void processModifyHeaders(Instant modified,
...
         setModified.accept(responseHeaders, modified);
         // If-Modified-Since
-        Optional<Instant> ifModSince = requestHeaders
-                .ifModifiedSince()
-                .map(ChronoZonedDateTime::toInstant);
-        if (ifModSince.isPresent() && !ifModSince.get().isBefore(modified)) {
-            throw new HttpException("Not valid for If-Modified-Since header", Status.NOT_MODIFIED_304, true);
+        if (!requestHeaders.contains(HeaderNames.IF_NONE_MATCH)) {
+            Optional<Instant> ifModSince = requestHeaders
+                    .ifModifiedSince()
```
danielkec: An invalid conditional date reaches `DateTime.parse`, and the outer handler converts the parse exception into a 500. Real requests with `If-Modified-Since: nope` or `If-Unmodified-Since: nope` returned `500 Cannot access static resource`; invalid date conditions are ignored, so these should take the ordinary 200 path.

danielkec (commented):
Two additional conditional-date findings confirmed on exact head `5b3f4acfb6229e502ca5c9283b35e283e74f06b2`.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in 64bcde8968. Static content now evaluates all four preconditions in RFC 9110 order through one coordinated processPreconditions call. The regression tests cover both precedence combinations described here.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in 1da998052d and optimized in 58f5f8c583. The matcher scans every raw repeated field and every comma-delimited member without splitting commas inside quoted opaque tags. Client/server tests cover matching If-Match and If-None-Match lists containing a quoted comma tag.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in 245730faea. Wildcard handling is decided before entity-tag payload comparison, so only a bare asterisk is a wildcard; quoted and weak quoted asterisks remain ordinary entity tags. Both forms have focused regression coverage.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in 33a54c16ec. The resource modification instant is truncated to HTTP-date whole-second precision before If-Modified-Since and If-Unmodified-Since comparisons. Regression tests copy the emitted Last-Modified value through the real server path.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
tomas-langer: Fixed in ac6a22d25b. Conditional date parsing now catches DateTimeParseException and treats an invalid value as an absent condition. Real-server tests cover invalid If-Modified-Since and If-Unmodified-Since values returning the ordinary 200 response.

danielkec (commented):
Three inline findings and one inherited response-finalization finding were confirmed on exact head `58f5f8c5836f8a31e042d6759c6f6929c6b3ca60`.

Static-content 304 responses currently carry `Content-Length: 0` for non-empty representations. For a seven-byte resource, matching `If-Modified-Since` or `If-None-Match` requests over both GET and HEAD return:

```http
HTTP/1.1 304 Not Modified
Content-Length: 0
```

The corresponding 200 response has `Content-Length: 7`. [RFC 9110 section 8.6](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6) permits a Content-Length on 304 only when it equals the size of that corresponding 200 response. The zero is injected by shared response finalization in `DirectHandlers` / `Http1ServerResponse`, outside this PR's changed files, so there is no honest inline anchor here.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -71,97 +72,147 @@ abstract class StaticContentHandler implements HttpService {
...
-     * @param responseHeaders an HTTP response headers
-     * @throws io.helidon.http.RequestException if (un)modify since header is checked
-     */
-    static void processModifyHeaders(Instant modified,
-                                     ServerRequestHeaders requestHeaders,
-                                     ServerResponseHeaders responseHeaders) {
-        processModifyHeaders(modified, requestHeaders, responseHeaders, ServerResponseHeaders::lastModified);
+    private static Optional<Instant> conditionalDate(Supplier<Optional<ZonedDateTime>> dateSupplier) {
+        try {
+            return dateSupplier.get().map(ChronoZonedDateTime::toInstant);
```
danielkec, line 232: When the request contains more than one `If-Modified-Since` or `If-Unmodified-Since` field line, the supplied accessors parse `Header.get()`, so this delegates only the first value and enforces the condition. The combined field value is not a valid single HTTP-date and should be ignored; on the exact head these cases return 304 and 412 instead of the ordinary 200. A minimal IMS case is:

```http
If-Modified-Since: Wed, 21 Oct 2099 07:28:00 GMT
If-Modified-Since: Wed, 21 Oct 2015 07:28:00 GMT
```

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerUrlStream.java
```diff
@@ -47,8 +47,10 @@ public boolean handle(LruCache<String, CachedHandler> cache,
         long lastModified = urlConnection.getLastModified();
 
         if (lastModified != 0) {
-            StaticContentHandler.processEtag(String.valueOf(lastModified), request.headers(), response.headers());
-            StaticContentHandler.processModifyHeaders(Instant.ofEpochMilli(lastModified), request.headers(), response.headers());
+            StaticContentHandler.processPreconditions(String.valueOf(lastModified),
```
danielkec: Because this call remains inside the `lastModified != 0` guard, an existing classpath resource with no reported timestamp skips every entity-tag precondition. Exact-head requests return:

```text
If-None-Match: *         -> 200 (expected 304)
If-Match: "not-current" -> 200 (expected 412)
```

These conditions depend on representation existence, not timestamp availability; the other three cached handlers have the same gate.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -71,97 +72,147 @@ abstract class StaticContentHandler implements HttpService {
...
+                boolean weak = fieldLength - start >= 2
+                        && (fieldValue.charAt(start) == 'W' || fieldValue.charAt(start) == 'w')
+                        && fieldValue.charAt(start + 1) == '/';
+                int entityStart = weak ? start + 2 : start;
+                if (entityStart < fieldLength && fieldValue.charAt(entityStart) == '"') {
+                    int entityEnd = fieldValue.indexOf('"', entityStart + 1);
+                    if (entityEnd < 0) {
+                        if (fieldLength - entityStart == 1) {
+                            // Preserve the existing malformed-tag failure until invalid entity tags are handled separately.
+                            unquoteETag(fieldValue.substring(entityStart));
```
danielkec: When a conditional entity-tag field contains only a DQUOTE, this call passes a one-character string to `unquoteETag`, which throws `StringIndexOutOfBoundsException`; the handler converts the request-controlled failure into a 500. A minimal case is:

```http
If-None-Match: "
```

`If-Match: "` and the corresponding `W/"` forms fail the same way; malformed conditional input should not produce an internal-server error.

tvallin (approved):

PR #12263, opened 2026-08-11
Support disabled protocol configurations

tomas-langer:
Add a common protocol enablement contract that preserves the enabled behavior of existing configurations. Disabled protocols no longer participate in transport binding validation or selector construction, and disabled WebSocket configurations are ignored when selecting connection settings.

Pre-requisite for #3686.

### DIFF
--- a/webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
+++ b/webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
@@ -91,7 +91,7 @@ public class WsConnection implements ServerConnection, WsSession {
                                       .config()
                                       .protocols()
                                       .stream()
-                                      .filter(p -> p instanceof WsConfig)
+                                      .filter(p -> p instanceof WsConfig && p.enabled())
                                       .findFirst()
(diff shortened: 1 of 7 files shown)


danielkec (approved):

webserver/websocket/src/main/java/io/helidon/webserver/websocket/WsConnection.java
```diff
@@ -91,7 +91,7 @@ private WsConnection(ConnectionContext ctx,
                                       .config()
                                       .protocols()
                                       .stream()
-                                      .filter(p -> p instanceof WsConfig)
+                                      .filter(p -> p instanceof WsConfig && p.enabled())
```
tvallin, line 94: This lookup can select an enabled WebSocket configuration belonging to a different transport binding. With a disabled UDS config followed by enabled TCP and UDS configs using different frame limits, the UDS upgrader is built from the enabled UDS config, but this scan selects the TCP config and applies its frame limit. I reproduced this with a raw UDS WebSocket test: it closes with 1009 at the base revision but echoes the oversized frame at this head. Please carry the exact `WsConfig` used by `WsUpgrader` into `WsConnection` instead of resolving it again from the listener-wide list, and add a multi-binding regression test.

PR #12264, opened 2026-08-11
Clear HTTP/1 read timeout after h2c upgrade

tomas-langer:
Clear the HTTP/1 request-scoped socket read timeout when an upgraded connection is handed to the HTTP/2 client, before its permanent reader starts. HTTP/2 continues to apply response deadlines per stream, while normal, ALPN, and prior-knowledge connection setup remains unchanged.

This is an intermittently failing test, and this change should remove the failures, this time caused by a production code.

I have added a fix for a flaky grpc test as well, as it failed on this PR directly.

### DIFF
--- a/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
+++ b/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
@@ -17,6 +17,7 @@
 package io.helidon.webclient.http2;
 
 import java.io.UncheckedIOException;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -165,7 +166,15 @@ public class Http2ClientConnection {
                                         Consumer<Http2ClientConnection> closeListener) {
 
         Http2ClientConnection h2conn = new Http2ClientConnection(http2Client, connection, closeListener);
-        return create(h2conn, http2Client, sendSettings);
+        return create(h2conn, http2Client, sendSettings, false);
+    }
+
+    static Http2ClientConnection createUpgraded(Http2ClientImpl http2Client,
+                                                ClientConnection connection,
+                                                Consumer<Http2ClientConnection> closeListener) {
+
+        Http2ClientConnection h2conn = new Http2ClientConnection(http2Client, connection, closeListener);
+        return create(h2conn, http2Client, false, true);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnectionHandler.java
```diff
@@ -252,9 +252,7 @@ private Http2ConnectionAttemptResult httpX(Http2ClientImpl http2Client,
                                                       http2Client.protocolConfig());
             if (upgradeResponse.isUpgraded()) {
                 result.set(Result.HTTP_2);
-                Http2ClientConnection conn = createHttp2Connection(http2Client,
-                                                                   upgradeResponse.connection(),
-                                                                   false);
+                Http2ClientConnection conn = createUpgradedHttp2Connection(http2Client, upgradeResponse.connection());
```
danielkec: When the peer's `MAX_CONCURRENT_STREAMS` is exhausted, this initial upgraded connection can be displaced by the replacement created for a concurrent request. This path stores it only in `activeConnection`, never in `allConnections`, so `Http2Client.closeResource()` cannot reach it; since `createUpgraded` now clears its socket timeout, the socket and reader can remain alive past client shutdown without a local timeout. A real-server probe with a limit of 1, one held request, and a concurrent second request produced:

```text
closed [initial, replacement]: [false, true]
initial SO_TIMEOUT after >2s: 0
```

danielkec (commented):
One inline lifecycle finding.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
```diff
@@ -753,6 +771,12 @@ private boolean handleSettingsFrame(int streamId, Http2FrameHeader frameHeader,
         recvListener.frame(ctx, streamId, receivedSettings);
         serverSettings = mergeSettings(serverSettings, receivedSettings);
         updatePeerSettings(receivedSettings);
+        if (!initialSettingsReceived && clearReadTimeoutAfterInitialSettings) {
+            // Preserve the HTTP/1.1 request timeout through the upgrade handshake. Once the peer's initial
+            // SETTINGS arrive, HTTP/2 owns a permanent connection reader and applies timeouts to individual
+            // streams, so the inherited socket timeout must be cleared before the next connection read.
+            connection.readTimeout(Duration.ZERO);
```
danielkec, line 793: When the client is closed after the h2c `101` but before the peer's initial `SETTINGS`, `createUpgraded` is still waiting and this connection has not yet been added to the handler registries, so `closeResource()` removes an empty handler. When `SETTINGS` then reaches this line, the socket timeout is cleared and the waiter publishes the connection into that already-removed handler, leaving its socket and reader alive past client shutdown. A loopback probe reproduced:

```text
peer: 101 -> hold initial SETTINGS
client: closeResource()
peer: send SETTINGS
result: no socket EOF after 5 s
```

danielkec (commented):
One additional current-head lifecycle finding.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
tomas-langer: Fixed in `aecd83a55e5c08a14780b552ad91951d234e1fa4`. Upgraded h2c connections are now registered before startup, drained during shutdown, and cannot be activated after the handler closes. Shutdown before the client preface completes closes the transport without emitting an out-of-order GOAWAY. Added deterministic unit and raw-loopback regressions for both startup windows.

danielkec (approved):
Approved.

PR #12266, opened 2026-08-11
Expose active WebClient TCP protocols

tomas-langer:
Expose the active TCP protocol IDs in configured preference order through the WebClient contract, allowing protocol clients to preserve TCP fallback without an implementation-specific downcast. Existing implementations default to an empty immutable list.

Pre-requisite for #3686.

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/LoomClient.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/LoomClient.java
@@ -159,7 +159,10 @@ class LoomClient implements WebClient {
     @Override
     public List<String> tcpProtocolIds() {
         List<String> protocolIds = tcpProtocolIds;
-        return protocolIds == null ? List.of() : protocolIds;
+        if (protocolIds == null) {
+            throw new IllegalStateException("TCP protocol IDs are not available during WebClient construction");
+        }
+        return protocolIds;
     }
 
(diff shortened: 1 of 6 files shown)


webclient/api/src/main/java/io/helidon/webclient/api/LoomClient.java
```diff
@@ -156,6 +156,12 @@ public void closeResource() {
         }
     }
 
+    @Override
+    public List<String> tcpProtocolIds() {
+        List<String> protocolIds = tcpProtocolIds;
+        return protocolIds == null ? List.of() : protocolIds;
```
danielkec: When a service-loaded provider calls `client.tcpProtocolIds()` from `HttpClientSpiProvider.protocol(...)`, this getter runs before `LoomClient` assigns `tcpProtocolIds`, so the provider can permanently retain an empty TCP-fallback snapshot even though the completed client exposes the configured protocol. A minimal provider that snapshots the value produces:

```text
provider snapshot: []
post-build getter: [probe]
provider snapshot after build: []
```

This leaves the constructed protocol client without the fallback inventory this contract is intended to expose.

danielkec (approved):
Reviewed exact head 13bf7e159eb36f40c372b0b484bda47ac4d98cd9; no actionable findings.

PR #12271, opened 2026-08-12
Add Vary for automatic content encoding

tomas-langer:
Pre-requisite for #12058

Automatic WebServer content encoding selects a response representation using `Accept-Encoding`, but the response does not currently advertise that selection to shared caches.

Add `Vary: Accept-Encoding` for both buffered and streaming automatic encoding paths. Append the value so dimensions configured by other components remain intact.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
@@ -301,6 +301,12 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
      */
     protected void beforeSend() {
         beforeSend.forEach(Runnable::run);
+        if (status().code() == Status.NOT_MODIFIED_304.code()
+                && contentEncodingContext.contentEncodingEnabled()
+                && !headers().contains(HeaderNames.CONTENT_ENCODING)
+                && !headers().containsToken(VARY_ACCEPT_ENCODING)) {
+            headers().add(VARY_ACCEPT_ENCODING);
+        }
     }
 
     /**
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
@@ -67,6 +67,8 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
      */
     protected static final Header STREAM_TRAILERS =
             HeaderValues.create(HeaderNames.TRAILER, STREAM_RESULT_NAME.defaultCase());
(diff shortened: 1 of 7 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -272,6 +274,7 @@ protected byte[] entityBytes(byte[] configuredEntity, int position, int length)
             }
             entity = baos.toByteArray();
             encoder.headers(headers());
+            headers().add(VARY_ACCEPT_ENCODING);
```
danielkec, line 277: For a buffered HTTP/2 response, this value is added before `Http2ServerResponse` invokes `beforeSend`, so a supported callback that sets `Vary` afterward removes `Accept-Encoding` after gzip has already been selected. A minimal route is:

```java
res.beforeSend(() -> res.headers().set(VARY_ORIGIN));
res.send("hello"); // HTTP/2 request with Accept-Encoding: gzip
```

This returns `Content-Encoding: gzip` with only `Vary: Origin`, allowing a shared cache to reuse the compressed representation for a request that did not advertise gzip.

danielkec (commented):
One automatic streaming path has no honest changed-line anchor: direct HTTP/1 SSE responses bypass both changed `ServerResponseBase` methods. `DataWriterSseSink` negotiates and applies its encoder directly; the following request to a gzip-enabled SSE route receives `Content-Encoding: gzip` without `Vary: Accept-Encoding`:

```http
GET /events HTTP/1.1
Accept-Encoding: gzip
```

`Cache-Control: no-cache` still permits storage and revalidation, so a shared cache lacks the representation-selection dimension. The relevant path is `webserver/sse/src/main/java/io/helidon/webserver/sse/DataWriterSseSink.java:93-105`.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
danielkec, line 277: Outside the static-content-specific conditional handling in #12058, a normal route that returns an empty `304` never reaches this addition because `length == 0`. With a route that sends `"hello"` normally and calls `res.status(NOT_MODIFIED_304).send()` when `If-None-Match` is present, the wire behavior is:

```http
GET /conditional
Accept-Encoding: gzip
If-None-Match: "v1"

=> 304 without Vary: Accept-Encoding
```

The corresponding `200` includes the field, and [RFC 9110 section 15.4.5](https://www.rfc-editor.org/rfc/rfc9110#section-15.4.5) requires the `304` to include the `Vary` field that the `200` would have sent, so this conditional response is missing representation metadata needed for cache revalidation.

danielkec (approved):
Reviewed and validated at dc37e423edd9107b85b37e2bf5e7d83664b544f5.

PR #12272, opened 2026-08-12
Support char[] in @Configuration.Value injection

### LINKED ISSUE #12267: Support char[] in @Configuration.Value injection
Helidon Version: 4.5.1 (the same mapping gap is present on current `main`)

## Enhancement Description

`@Configuration.Value` supports `Supplier<T>` injection, but a scalar configuration value cannot be mapped to `char[]`. Configured blueprints already support both `char[]` and `Supplier<char[]>` using `asString().as(String::toCharArray)`; injected configuration values should provide the same type support.

### Reproducer

With `app.password=secret-value`:

```java
@Service.Singleton
final class SecretAwareService {
    private final Supplier<char[]> password;

    @Service.Inject
    SecretAwareService(
            @Configuration.Value("app.password")
            Supplier<char[]> password) {
        this.password = password;
    }

    char[] password() {
        return password.get();
    }
}
```

The application compiles and the service is constructed, but the first `password.get()` fails with:

```text
io.helidon.config.ConfigMappingException: Mapping of value stored under key
'app.password' to type 'class [C' has failed: No mapper configured
```

Direct `@Configuration.Value("app.password") char[] password` injection reaches the same unsupported mapping path.

### Expected behavior

Both `char[]` and `Supplier<char[]>` should resolve scalar string configuration without requiring a custom mapper. The supplier form should retain the existing deferred-resolution behavior of `Supplier<T>` injection.

This gives application code a mutable character array that it can clear after use. The configuration value still passes through `String`, so this is not an end-to-end memory-erasure guarantee.

Configured-blueprint supplier support was introduced by #7323. The adjacent generic mapper fix in #12104 does not add `char[]` conversion.

danielkec:
### Description

Fixes #12267

Adds a built-in scalar `char[]` configuration mapper so direct `@Configuration.Value` and `Supplier<char[]>` injection work without requiring a custom mapper.

Supplier resolution remains deferred, and user-provided `char[]` mappers continue to take precedence.

### Documentation

None

### Testing

- `mvn -Ptests -pl config/config,config/tests/service-registry -am -Dtest=ConfigurationValueFactoryTest,ConfigurationValueInjectionTest,ConfigMappersTest,ConfigMappersFailingTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -pl config/config -am test`
- `mvn -Ptests -pl config/tests/service-registry -am test`
- `git diff --check`

### DIFF
--- a/config/config/src/main/java/io/helidon/config/ConfigMappers.java
+++ b/config/config/src/main/java/io/helidon/config/ConfigMappers.java
@@ -128,6 +128,7 @@ public final class ConfigMappers {
                              Map.entry(Double.class, wrap(ConfigMappers::toDouble)),
                              Map.entry(Boolean.class, wrap(ConfigMappers::toBoolean)),
                              Map.entry(Character.class, wrap(ConfigMappers::toChar)),
+                             Map.entry(char[].class, wrap(String::toCharArray)),
                              //java.lang
                              Map.entry(Class.class, wrap(ConfigMappers::toClass)),
                              //javax.math
--- a/config/config/src/test/java/io/helidon/config/ConfigurationValueFactoryTest.java
+++ b/config/config/src/test/java/io/helidon/config/ConfigurationValueFactoryTest.java
@@ -39,6 +39,7 @@ import org.junit.jupiter.api.Test;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
 class ConfigurationValueFactoryTest {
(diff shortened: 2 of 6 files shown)


config/config/src/main/java/io/helidon/config/ConfigMappers.java
```diff
@@ -128,6 +128,7 @@ private static OptionalInt optionalIntEssential(Config node) {
                              Map.entry(Double.class, wrap(ConfigMappers::toDouble)),
                              Map.entry(Boolean.class, wrap(ConfigMappers::toBoolean)),
                              Map.entry(Character.class, wrap(ConfigMappers::toChar)),
+                             Map.entry(char[].class, wrap(String::toCharArray)),
```
tomas-langer, line 131: `@Default.Value` still cannot supply a `char[]` here. When a `@Configuration.Value` key is absent, `ConfigurationValueFactory` delegates to `DefaultResolverService`, which uses the common mapper registry; that registry has `String -> char`/`Character` but no `String -> char[]`. As a result, `@Configuration.Value("app.password") @Default.Value("secret") char[] password` throws `MapperException`, while the inline `${...:...}` form added by these tests works. Please add the common `String -> char[]` mapper and cover the annotation-default path with a regression test.

config/config/src/test/java/io/helidon/config/ConfigurationValueFactoryTest.java
```diff
@@ -59,6 +60,39 @@ void literalKeyUsesDirectLookup() {
...
+
+        assertThat(values, hasSize(1));
+        assertArrayEquals("secret-value".toCharArray(), (char[]) values.getFirst().get());
+    }
+
+    @Test
+    void literalKeyUsesCustomCharArrayMapper() {
+        Config config = Config.builder()
+                .sources(ConfigSources.create(Map.of("app.password", "secret-value")))
+                .addMapper(char[].class, it -> "custom-value".toCharArray())
```
tomas-langer: `it` is intentionally unused. On this Java 26 line, please use the unnamed lambda parameter (`_ -> "custom-value".toCharArray()`) so the test follows Helidon source style and does not imply the config node affects this precedence-only mapper.

tomas-langer (approved):

PR #12273, opened 2026-08-12
Expose TLS reload generation

tomas-langer:
Pre-requisite for #3686

Adds a TLS manager generation contract so protocol runtimes can detect changes to key and trust material while retaining a stable `SSLContext`. Existing `TlsManager` implementations remain compatible through a default generation of zero; generation-aware managers can override it.

`ConfiguredTlsManager` serializes reload publication, provides stable generation reads, and derives exposed key and trust managers from the same reloadable delegates used by the `SSLContext`. Automatically configured managers remain owned by each `Tls` instance, while explicitly supplied managers can be shared.

### DIFF
--- a/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
@@ -118,7 +118,15 @@ public class ConfiguredTlsManager implements TlsManager {
     @SuppressWarnings("removal")
     public void reload(Tls tls) {
         Tls.validateReloadSource(tls);
-        reload(tls.keyManager(), tls.trustManager());
+        long sourceGeneration;
+        Optional<X509KeyManager> keyManager;
+        Optional<X509TrustManager> trustManager;
+        do {
+            sourceGeneration = tls.generation();
+            keyManager = tls.keyManager();
+            trustManager = tls.trustManager();
+        } while (sourceGeneration != tls.generation());
+        reload(keyManager, trustManager);
     }
 
     @Override // TlsManager
--- a/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
+++ b/common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
@@ -50,6 +50,7 @@ import javax.net.ssl.TrustManagerFactory;
 import javax.net.ssl.X509KeyManager;
 import javax.net.ssl.X509TrustManager;
 
+import io.helidon.common.Api;
 import io.helidon.common.LazyValue;
 
 /**
@@ -65,6 +66,8 @@ public class ConfiguredTlsManager implements TlsManager {
     private final ReentrantLock stateLock = new ReentrantLock();
 
     private volatile TlsManagerState state = new TlsManagerState();
+    // Even values are stable. An odd value exists only while stateLock is held for material publication.
+    private volatile long generationState;
 
     ConfiguredTlsManager() {
         this("@default", "tls-manager");
@@ -99,7 +102,15 @@ public class ConfiguredTlsManager implements TlsManager {
 
     @Override // TlsManager
     public void init(TlsConfig tlsConfig) {
-        sslContext(tlsConfig);
+        Objects.requireNonNull(tlsConfig, "tlsConfig");
+        stateLock.lock();
+        try {
+            if (state.sslContext == null) {
+                sslContext(tlsConfig);
+            }
+        } finally {
+            stateLock.unlock();
+        }
     }
 
     @Override // TlsManager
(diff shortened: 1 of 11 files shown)


common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
```diff
@@ -99,7 +102,15 @@ public SSLContext sslContext() {
 
     @Override // TlsManager
     public void init(TlsConfig tlsConfig) {
-        sslContext(tlsConfig);
+        Objects.requireNonNull(tlsConfig, "tlsConfig");
+        stateLock.lock();
+        try {
+            if (state.sslContext == null) {
```
danielkec, line 108: When one `ConfiguredTlsManager` subclass is shared by two `Tls` instances, this guard silently retains the first initialization even if the second configuration has a different trust policy. A minimal sequence is:

```java
var first = Tls.builder().manager(manager).trustAll(true).build();
var second = Tls.builder().manager(manager).trustAll(false).build();
```

At this head, `second.prototype().trustAll()` is `false`, but `second.sslContext() == first.sslContext()` and the installed trust manager still accepts an untrusted chain. This can make a stricter client or listener run with the earlier trust-all policy.

common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
```diff
@@ -121,18 +132,38 @@ public void reload(TlsMaterial material) {
...
+            return generationState >>> 1;
+        } finally {
+            stateLock.unlock();
+        }
+    }
+
     @Override // TlsManager
     public Optional<X509KeyManager> keyManager() {
-        return Optional.ofNullable(state.keyManager);
+        return Optional.ofNullable(state.reloadableKeyManager.current());
```
danielkec, line 161: `reload(Tls)` reads this accessor and `trustManager()` sequentially without checking the source generation. With this live-delegate access, if source publication pauses after key `K1` is installed but before trust `T1`, a target copy observes:

```text
source before: K0/T0
source in publication: K1/T0
target.reload(sourceTls): K1/T0 at a stable target generation
```

A latch-based 100-iteration probe reproduced this 100/100 times; the same paused window returned `K0/T0` on the base. The older sequential copy could already straddle a completed outer-state swap, but this adds the in-publication window and can leave paired mTLS identity and trust material mismatched until a later reload.

danielkec (commented):
I found two issues at this exact head.

common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
tomas-langer: Reusing one manager with conflicting `Tls` configurations is not a supported use case. The inconsistency already exists on `main`: repeated initialization replaces the shared manager state, so the second initialization wins while the earlier `Tls` remains attached to its previous context. This PR does not make one manager support independent trust policies.

common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
tomas-langer: Good catch. `reload(Tls)` was reading the source key and trust managers independently and could therefore copy different generations. It now follows the generation snapshot protocol and retries until both values come from a stable generation. A deterministic regression covers the crossing-generation case.

common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
```diff
@@ -121,18 +140,38 @@ public void reload(TlsMaterial material) {
...
+    @Api.Internal
+    public long generation() {
+        long current = generationState;
+        if ((current & 1L) == 0L) {
+            return current >>> 1;
+        }
+
+        stateLock.lock();
+        try {
+            return generationState >>> 1;
```
danielkec, line 153: When the publisher thread re-enters `generation()` from a synchronous logging callback during reload, `stateLock.lock()` succeeds reentrantly and this returns the previous generation while publication is still between the key and trust updates. A handler invoked at that point can call `target.reload(sourceTls)` and both generation reads match:

```text
source callback: generation 0/0, K1/T0
target after copy: K1/T0 at generation 1
```

The retry therefore accepts an incoherent pair as stable; a deterministic 100-iteration handler probe reproduced this 100/100 times.

danielkec (commented):
I found one current-head follow-up issue.

common/tls/src/main/java/io/helidon/common/tls/Tls.java
```diff
@@ -77,7 +77,14 @@ private Tls(TlsConfig config) {
 
         if (config.enabled()) {
             if (config.sslContext().isEmpty()) {
-                this.tlsManager = config.manager();
+                TlsManager configuredManager = config.manager();
+                if (configuredManager.getClass() == ConfiguredTlsManager.class) {
+                    configuredManager = new ConfiguredTlsManager();
+                    configuredTls = TlsConfig.builder(config)
```
tvallin: `Tls` implements `RuntimeType.Api`, whose `prototype()` contract is the prototype *as received*. This rebuild means `Tls.create(input).prototype()` is no longer `input` on the default-manager path, and it is not even equal because the manager identity changes. Please preserve the received `TlsConfig` and track the per-instance manager separately.

common/tls/src/main/java/io/helidon/common/tls/ConfiguredTlsManager.java
tomas-langer: The synchronous logger itself is legal; the unsupported behavior is a logging handler re-entering TLS reload
operations while handling an internal reload message. We should not complicate ConfiguredTlsManager to
accommodate that scenario.

danielkec (approved):

PR #12275, opened 2026-08-12
Add shared HTTP transport observability

tomas-langer:
Pre-requisite for #3686

Adds a protocol-neutral internal observer for HTTP connection, handshake, protocol-selection, and stream lifecycle events. Observer composition isolates callback failures and preserves exact lifecycle completion under concurrent and reentrant callbacks.


http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,521 @@
...
+            lifecycleLock.lock();
+            try {
+                if (closed || Objects.equals(this.protocol, selectedProtocol)) {
+                    return;
+                }
+                this.protocol = selectedProtocol;
+                drain = enqueueTransitionLocked(() -> {
+                    for (ConnectionObservation observation : observations) {
+                        try {
+                            observation.protocolSelected(selectedProtocol);
```
danielkec, line 210: When the first observer re-enters `protocolSelected` from its handshake-close callback, this transition reaches every observer before handshake-close dispatch resumes for the second observer. The exact-head event order is:

```text
first-handshake-SUCCESS
first-protocol-http/2
second-protocol-http/2
second-handshake-SUCCESS
```

The second observer is therefore told that the connection is established and usable before it is told that TLS succeeded, contradicting the interface contract and corrupting protocol-scoped metrics or traces. The same causal-order gap lets a reentrant `streamOpened` reach the second observer before its protocol-selection callback.

http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,521 @@
...
+            List<ConnectionObservation> observations = new ArrayList<>(observers.size());
+            for (HttpTransportObserver observer : observers) {
+                try {
+                    ConnectionObservation observation = Objects.requireNonNull(
+                            observer.connectionOpened(role, transport, handshake),
+                            "connection observation");
+                    if (observation != NOOP_CONNECTION) {
+                        observations.add(observation);
+                    }
+                } catch (Throwable failure) {
```
danielkec: Catching every `Throwable` here also treats fatal JVM conditions as recoverable observer failures; the same pattern recurs for all lifecycle callbacks. On this head, an observer throwing `new OutOfMemoryError("probe")` is logged, the next observer is still invoked, and `connectionOpened` returns normally. Suppressing these resource/termination signals—and attempting to log a stack trace during resource exhaustion—can worsen an availability failure.

http/http/src/main/java/io/helidon/http/HttpTransportObserver.java
```diff
@@ -0,0 +1,346 @@
...
+         *
+         * @return handshake observation
+         */
+        HandshakeObservation handshakeStarted();
+
+        /**
+         * Reports the currently selected HTTP protocol.
+         *
+         * <p>The first selection marks the physical connection as established and usable for that protocol. A publisher
+         * must therefore report the first selection only after any required transport security handshake has succeeded.
```
danielkec: This contract cannot faithfully represent accepted TLS 1.3 early data: an HTTP client can send requests before the handshake completes, while the first protocol selection is forbidden until success and is defined as the point when the connection becomes usable. The observer must therefore emit an early stream before its protocol/usable state, delay or omit that lifecycle, or violate this contract. This applies directly to the planned HTTP/3/QUIC use case; [RFC 8470](https://www.rfc-editor.org/rfc/rfc8470.html) describes HTTP requests sent without waiting for TLS completion.

http/http/src/main/java/io/helidon/http/HttpTransportObserver.java
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    /**
+     * Connection termination outcome.
+     */
+    enum ConnectionOutcome {
+        /**
+         * Connection completed normally.
+         */
+        NORMAL,
```
danielkec, line 162: These outcomes overlap without a documented precedence: a graceful connection shutdown is both normal and locally initiated for one endpoint (remotely initiated for its peer), so `NORMAL` and one of the side-specific values describe the same event. Likewise, a refused multiplexed stream can be both `REJECTED` and `RESET`. Publishers can therefore tag identical lifecycle events differently, making metrics produced through this shared API incomparable.

danielkec (commented):
Four exact-head findings are included inline.

http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,541 @@
...
+        public HandshakeObservation handshakeStarted() {
+            CompositeHandshakeObservation result;
+            boolean open = false;
+            lifecycleLock.lock();
+            try {
+                if (closed) {
+                    return NOOP_HANDSHAKE;
+                }
+                if (handshake == null) {
+                    handshake = new CompositeHandshakeObservation(this);
```
danielkec, line 149: A second concurrent `handshakeStarted()` call can return this shared observation while the first call is still inside an underlying observer's `handshakeStarted()`. Closing that return value with `SUCCESS` and then selecting HTTP/2 produces:

```text
first-handshake-start
first-protocol-http/2
second-protocol-http/2
first-handshake-SUCCESS
second-handshake-start
second-handshake-SUCCESS
```

This uses supported concurrency, not callback reentry. The second observer receives protocol selection before its handshake has even started, so metrics or traces can record an impossible lifecycle.

http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,541 @@
...
+                    }
+                } finally {
+                    lifecycleLock.unlock();
+                }
+                try {
+                    transition.run();
+                } catch (RuntimeException failure) {
+                    observerFailed("connection transition", failure);
+                } catch (Throwable failure) {
+                    throw failure;
```
danielkec: When a queued observer callback throws an `Error`, this exits the drainer while `transitionDraining` remains set. After the caller catches the error, selecting another protocol and closing the connection produces:

```text
caught: probe
recordedProtocols=0, recordedCloses=0
```

Both later transitions remain queued indefinitely, so the remaining observers silently miss protocol and teardown events.

danielkec (commented):
Two current-head lifecycle findings are included inline.

http/http/src/main/java/io/helidon/http/HttpTransportObserver.java
```diff
@@ -0,0 +1,364 @@
...
+         * represent HTTP streams accepted as TLS or QUIC early data.
+         *
+         * <p>The caller owns the returned observation and closes it when the stream terminates.
+         * The returned observation must not be {@code null}.
+         *
+         * @param direction stream directionality
+         * @param initiator stream initiator
+         * @return stream observation
+         */
+        StreamObservation streamOpened(Direction direction, Initiator initiator);
```
tvallin, line 306: This callback does not yet define a protocol-neutral unit. HTTP/1.1 has request/response exchanges rather than multiplexed streams, while HTTP/3 has request, push, control, and QPACK streams that `Direction` and `Initiator` cannot distinguish. Different publishers can therefore emit incomparable lifecycle counts. Please define exactly which protocol events must call this method—for example, application request/response exchanges only—or add an extensible stream-kind identifier before integrations depend on this signature.

http/http/src/main/java/io/helidon/http/HttpTransportObserver.java
```diff
@@ -0,0 +1,364 @@
...
+     */
+    String TRANSPORT_QUIC = "quic";
+    /**
+     * Known protocol identifier for HTTP/1.1, aligned with the HTTP transport metrics tag value.
+     */
+    String PROTOCOL_HTTP_1_1 = "http/1.1";
+    /**
+     * Known protocol identifier for HTTP/2, aligned with the HTTP transport metrics tag value.
+     */
+    String PROTOCOL_HTTP_2 = "http/2";
```
tvallin, line 55: This introduces a second HTTP/2 identifier namespace: existing Helidon protocol-selection APIs and ALPN use `h2`, while this observer constant uses `http/2`. Since `protocolSelected` accepts any non-blank protocol identifier and does not require canonicalization to these metrics constants, publishers can report the same protocol under either value and split observer tags. Please either reuse the existing IDs or explicitly require known protocols to use the observer constants.

http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,549 @@
...
+            }
+        }
+
+        private void closeHandshake(CompositeHandshakeObservation handshake,
+                                    HandshakeOutcome outcome) {
+            List<HandshakeObservation> observations;
+            lifecycleLock.lock();
+            try {
+                observations = handshake.prepareClose(outcome);
+                if (observations == null) {
```
danielkec, line 331: When one thread is still dispatching `HandshakeObservation.close(SUCCESS)`, an idempotent duplicate close returns here immediately. The publisher can then satisfy the documented close-return precondition and call `protocolSelected`, producing:

```text
before release: [first-handshake-SUCCESS, first-protocol-http/2, second-protocol-http/2]
after release:  [first-handshake-SUCCESS, first-protocol-http/2, second-protocol-http/2, second-handshake-SUCCESS]
```

This uses one sequential `handshakeStarted()` call and no callback reentry, so the second observer still receives protocol selection before handshake completion under supported concurrency.

http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,549 @@
...
+                        break;
+                    }
+                } finally {
+                    lifecycleLock.unlock();
+                }
+                try {
+                    transition.run();
+                } catch (RuntimeException failure) {
+                    observerFailed("connection transition", failure);
+                } catch (Error failure) {
```
danielkec, line 392: Although the new drainer recovery handles `Error`, a callback can still throw another non-runtime `Throwable`, for example a checked exception from Kotlin or a generic sneaky throw. It bypasses this catch and exits before `transitionDraining` is cleared; after the caller catches it, the exact-head result is:

```text
caught=CheckedFailure:probe
recordedProtocols=0, recordedCloses=0
```

Subsequent protocol and teardown events remain queued indefinitely, contradicting the documented handling of any other `Throwable`.

http/http/src/main/java/io/helidon/http/HttpTransportObservers.java
```diff
@@ -0,0 +1,549 @@
...
+                            observerFailed("protocol selection", failure);
+                        } catch (Throwable failure) {
+                            throw failure;
+                        }
+                    }
+                });
+            } finally {
+                lifecycleLock.unlock();
+            }
+            if (drain) {
```
danielkec, line 221: When a second `protocolSelected` call arrives while another thread is draining, its transition is queued and this caller returns without draining it. If that queued observer callback throws an `AssertionError`, the exact-head result is:

```text
secondCallerFailure=null
firstCallerFailure=java.lang.AssertionError: second-transition
```

The caller whose transition failed continues normally while the fatal failure surfaces on the unrelated original drainer thread.

danielkec (commented):
Three distinct current-head lifecycle findings are included inline.

tvallin (approved):

PR #12277, opened 2026-08-12
Reset response entity before error handling

tomas-langer:
Pre-requisite for #12058

Reset unsent response entity state before invoking error handling so replacement error responses do not inherit stale content metadata, trailers, or entity-specific filters and listeners.

Preserve infrastructure response filters and listeners across entity replacement. Keep the new `RoutingResponse` hooks internal, with `resetEntity()` delegating to the existing `resetStream()` by default for implementation compatibility.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
@@ -25,6 +25,7 @@ import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.function.UnaryOperator;
 
+import io.helidon.common.Api;
 import io.helidon.common.GenericType;
 import io.helidon.common.buffers.BufferData;
 import io.helidon.common.uri.UriPath;
@@ -154,6 +155,7 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
         return (T) this;
     }
 
+    @Api.Internal
     @Override
     public void entityBeforeSend(Runnable listener) {
         beforeSend.add(Objects.requireNonNull(listener));
@@ -233,6 +235,7 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
         return hasEntity() || isNexted() || shouldReroute();
     }
 
+    @Api.Internal
     @Override
     public boolean resetEntity() {
         if (!resetStream()) {
@@ -274,6 +277,7 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
         streamFilter = addStreamFilter(streamFilter, filterFunction);
     }
 
+    @Api.Internal
     @Override
     public void entityStreamFilter(UnaryOperator<OutputStream> filterFunction) {
         checkStreamFilter(filterFunction);
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
@@ -86,8 +86,8 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
     private UriQuery rerouteQuery;
     private String reroutePath;
     private Consumer<ServerResponseTrailers> beforeTrailers;
-    private Runnable persistentBeforeSend;
-    private UnaryOperator<OutputStream> persistentStreamFilter;
+    private Runnable responseBeforeSend;
+    private UnaryOperator<OutputStream> responseStreamFilter;
     private UnaryOperator<OutputStream> streamFilter;
 
     /**
@@ -142,19 +142,19 @@ public abstract class ServerResponseBase<T extends ServerResponseBase<T>> implem
 
     @Override
     public ServerResponse beforeSend(Runnable listener) {
-        beforeSend.add(Objects.requireNonNull(listener));
-        return (T) this;
-    }
-
-    @Override
-    public void persistentBeforeSend(Runnable listener) {
         Objects.requireNonNull(listener);
-        Runnable current = persistentBeforeSend;
-        persistentBeforeSend = current == null ? listener : () -> {
+        Runnable current = responseBeforeSend;
+        responseBeforeSend = current == null ? listener : () -> {
             current.run();
             listener.run();
         };
         beforeSend.add(listener);
+        return (T) this;
+    }
+
+    @Override
+    public void entityBeforeSend(Runnable listener) {
+        beforeSend.add(Objects.requireNonNull(listener));
     }
 
     @Override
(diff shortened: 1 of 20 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -211,6 +229,48 @@ public boolean isResponseHandled() {
...
+        headers.remove(REPR_DIGEST_NAME);
+        headers.remove(HeaderNames.ETAG);
+        headers.remove(HeaderNames.LAST_MODIFIED);
+        headers.remove(HeaderNames.ACCEPT_RANGES);
+        beforeSend.clear();
+        if (persistentBeforeSend != null) {
+            beforeSend.add(persistentBeforeSend);
+        }
+        beforeTrailers = null;
+        streamFilter = persistentStreamFilter;
```
danielkec: `resetEntity()` treats every registration made through the public `beforeSend` and `streamFilter` APIs as entity-scoped. An application/global filter can also use these hooks for cross-cutting policy; if the downstream route throws before committing, these lines remove the hooks before the replacement error is sent, so a 500 can lose a CSP/security header or body signing/encryption. A minimal header case is:

```java
response.beforeSend(() -> response.header("Content-Security-Policy", "default-src 'none'"));
chain.proceed(); // downstream route throws before sending
```

Before this change, `resetStream()` retained both registrations. The persistent alternatives are `@Api.Internal`, leaving existing application middleware without a supported public way to preserve this coverage across error replacement.

tomas-langer: Fixed. Public `beforeSend` and `streamFilter` hooks now remain response-scoped across `resetEntity()`. New internal entity-scoped hooks are cleared with the discarded entity. HTTP/1 and HTTP/2 regression tests cover retained headers and body filters on replacement error responses.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -213,6 +231,51 @@ public boolean isResponseHandled() {
...
+        if (status != null && status.code() == Status.PARTIAL_CONTENT_206.code()) {
+            status = null;
+        }
+        headers.remove(HeaderNames.CONTENT_TYPE);
+        headers.remove(HeaderNames.CONTENT_ENCODING);
+        headers.remove(HeaderNames.CONTENT_LANGUAGE);
+        headers.remove(HeaderNames.CONTENT_LOCATION);
+        headers.remove(HeaderNames.CONTENT_DISPOSITION);
+        headers.remove(CONTENT_DIGEST_NAME);
+        headers.remove(REPR_DIGEST_NAME);
```
tvallin, line 257: `resetEntity()` still leaves the legacy entity-integrity fields `Digest` and `Content-MD5` attached to the replacement body. If a route sets either field, obtains an output stream without committing bytes, and then throws, the reset succeeds and the error handler sends a different entity with the original digest still present. Please clear these fields along with `Content-Digest`/`Repr-Digest` and cover the reset path, otherwise clients or proxies which still validate them will reject the error response.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -228,6 +291,26 @@ protected Consumer<ServerResponseTrailers> beforeTrailers() {
         return beforeTrailers;
     }
 
+    /**
+     * Whether this response has any output stream filters.
+     *
+     * @return whether an output stream filter is configured
+     */
+    protected final boolean hasStreamFilter() {
```
tvallin, line 305: Please avoid adding these helpers as `final` protected methods on `ServerResponseBase` (the same applies to `applyStreamFilters(...)` below). This class is public in an exported package, so an existing compiled subclass that already declares either signature will fail to load after upgrading with `IncompatibleClassChangeError: ... overrides final method`. Making the helpers non-final (or keeping this state out of the subclass API) avoids that linkage break.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -135,10 +142,21 @@ public void send(Object entity) {
...
+        responseBeforeSend = current == null ? listener : () -> {
+            current.run();
+            listener.run();
+        };
         beforeSend.add(listener);
         return (T) this;
     }
 
+    @Override
+    public void entityBeforeSend(Runnable listener) {
```
tvallin, line 160: Please repeat `@Api.Internal` on this override and on `resetEntity()` and `entityStreamFilter(...)`. Method annotations are not inherited from `RoutingResponse`, and the API-stability scanner checks the selected method itself. A caller whose static receiver type is the public `ServerResponseBase` therefore sees these new infrastructure hooks as ordinary API, defeating the intended internal contract.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
```diff
@@ -213,6 +231,51 @@ public boolean isResponseHandled() {
...
+        headers.remove(REPR_DIGEST_NAME);
+        headers.remove(HeaderNames.ETAG);
+        headers.remove(HeaderNames.LAST_MODIFIED);
+        headers.remove(HeaderNames.ACCEPT_RANGES);
+        beforeSend.clear();
+        if (responseBeforeSend != null) {
+            beforeSend.add(responseBeforeSend);
+        }
+        beforeTrailers = null;
+        streamFilter = responseStreamFilter;
```
danielkec, line 266: Restoring the public filter here while the reset clears its eagerly registered representation metadata and trailer callback changes existing cross-cutting middleware behavior. For example:

```java
res.header(HeaderNames.CONTENT_ENCODING, "base64");
res.header(HeaderNames.TRAILER, RESPONSE_SIGNATURE_NAME.defaultCase());
res.beforeTrailers(t -> t.set(RESPONSE_SIGNATURE_NAME, "signed"));
res.streamFilter(Base64.getEncoder()::wrap);
chain.proceed(); // downstream throws before committing
```

With this sequence, the replacement error is still Base64-transformed on both HTTP/1.1 and HTTP/2, but it carries neither `Content-Encoding` nor the signature trailer, so clients cannot interpret or verify it. Before this PR, `resetStream()` retained these registrations together; the new `beforeSend` requirement therefore leaves existing public-API registrations with a wire-level compatibility regression.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
tomas-langer: Fixed in `9ffe61d4a4`: `resetEntity()` now removes both `Digest` and `Content-MD5`, and the reset regression covers both headers.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
tomas-langer: I am keeping these helpers `final`. `ServerResponseBase` is an internal implementation base used by frameworks and HTTP extensions, not a general application SPI. This is a major-version transition, so those implementations must be updated for 27.x. The helpers are now marked `@Api.Internal` to make that classification explicit.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
tomas-langer: Fixed in `9ffe61d4a4`: `@Api.Internal` is repeated on `entityBeforeSend(...)`, `resetEntity()`, and `entityStreamFilter(...)`. The newly exposed protected stream-filter helpers are classified the same way.

webserver/webserver/src/main/java/io/helidon/webserver/http/ServerResponseBase.java
tomas-langer: The filter registration is response-scoped, while representation metadata and trailer state belong to the entity being discarded and cannot safely be copied to the replacement. Middleware that transforms whichever entity is sent must pair the public `streamFilter(...)` with response-scoped `beforeSend(...)`, which re-establishes the matching headers and trailer callback for the actual entity. The Javadoc/docs and HTTP/1.1 and HTTP/2 tests now cover that contract.

danielkec (approved):
Reviewed the current head. The response entity reset contract is documented and covered for HTTP/1.1 and HTTP/2.

tvallin (approved):

PR #12286, opened 2026-08-13
Avoid CDI tracing lookup during OIDC metadata load

### LINKED ISSUE #12285: IDCS role mapper can fail during MP startup when OIDC metadata loading uses CDI-backed tracing
After upgrading toHelidon MP 4.5.2, applications using `idcs-role-mapper` can fail during startup when MP telemetry/tracing is present.

The IDCS role mapper eagerly resolves the OIDC token endpoint during MP security initialization. If well-known metadata discovery is enabled, this triggers a startup-time WebClient request to:

`<identity-uri>/.well-known/openid-configuration`

That WebClient request goes through tracing. In 4.5.x, tracing can resolve through the Helidon service registry into CDI-backed telemetry producers while CDI initialization is still in progress. Weld then rejects the lookup with:

`WELD-001332: BeanManager method getReference() is not available during application initialization`

Verdent:
Fixes #12285

Run well-known metadata discovery with a request-local no-op tracer when no tracer is already available in the Helidon context. This prevents MP startup-time metadata loading from resolving CDI-backed tracing producers before the CDI container is ready.

Add an MP telemetry and IDCS role mapper startup regression test.

### DIFF
--- a/security/providers/idcs-mapper/src/main/java/io/helidon/security/providers/idcs/mapper/IdcsRoleMapperProvider.java
+++ b/security/providers/idcs-mapper/src/main/java/io/helidon/security/providers/idcs/mapper/IdcsRoleMapperProvider.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
+import io.helidon.common.LazyValue;
 import io.helidon.common.context.Context;
 import io.helidon.common.context.Contexts;
 import io.helidon.config.Config;
@@ -48,10 +49,9 @@ import io.helidon.webclient.api.HttpClientRequest;
 public class IdcsRoleMapperProvider extends IdcsRoleMapperProviderBase implements SubjectMappingProvider {
     private final EvictableCache<String, List<Grant>> roleCache;
     private final String asserterUri;
-    private final URI tokenEndpointUri;
 
     // caching application token (as that can be re-used for group requests)
-    private final AppToken appToken;
+    private final LazyValue<AppToken> appToken;
 
     /**
      * Constructor that accepts any {@link IdcsRoleMapperProvider.Builder} descendant.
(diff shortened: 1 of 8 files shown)


tests/integration/mp-idcs-tracing-startup/src/test/java/io/helidon/tests/integration/idcstracing/IdcsRoleMapperStartupTest.java
```diff
@@ -0,0 +1,92 @@
...
+            try (OutputStream output = exchange.getResponseBody()) {
+                output.write(metadata);
+            }
+        });
+        metadataServer.start();
+
+        Map<String, String> properties = new HashMap<>();
+        properties.put("server.port", "0");
+        properties.put("otel.sdk.disabled", "true");
+        properties.put("security.providers.0.idcs-role-mapper.multitenant", "false");
```
tomas-langer: This regression test covers only the single-tenant path, while the PR also removes eager client initialization from the default multitenant endpoints. Please add a multitenant test with well-known discovery enabled that verifies provider construction makes no metadata request and the first token request triggers discovery.

tests/integration/mp-idcs-tracing-startup/src/test/java/io/helidon/tests/integration/idcstracing/IdcsRoleMapperStartupTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.sun.net.httpserver.HttpServer;
```
tomas-langer: The import blocks do not follow the repository's declared layout. Non-static imports are ordered as `java`, `javax`, `io.helidon`, then all other packages. Here `com.sun` precedes the Helidon imports and `org.junit` is separated from the same catch-all block, leaving the new test inconsistent with the repository's source style despite the current Checkstyle job passing.

security/providers/idcs-mapper/src/main/java/io/helidon/security/providers/idcs/mapper/IdcsRoleMapperProvider.java
```diff
@@ -189,7 +179,19 @@ protected List<? extends Grant> getGrantsFromServer(Subject subject) {
 
     // Added for testing purposes
     Optional<String> getAppToken(RoleMapTracing tracing) {
-        return appToken.getToken(tracing);
+        return appToken.get().getToken(tracing);
```
tomas-langer, line 181: [P2] First-use metadata failures can serialize concurrent requests into repeated timeouts

This moves synchronous well-known metadata discovery onto the first role-mapping request. If discovery fails, `LazyValueImpl` resets the lazy value to its initial state, so each queued caller can become the next initializer and repeat the same network timeout; the multitenant `computeIfAbsent` path similarly retains no entry when initialization throws. With N concurrent first mappings while metadata is unavailable, the requests can therefore perform up to N serialized discovery attempts, and the kth caller can wait roughly k times the metadata-client timeout while request threads or virtual threads remain queued. The added tests cover only one successful first-use initialization, so they do not exercise this outage behavior.

tomas-langer (approved):

PR #12291, opened 2026-08-13
Cache static content metadata

tomas-langer:
Pre-requisite for #12058

Cache static-content representation metadata with each resolved resource so HEAD and conditional requests do not repeatedly inspect or open the source.

Files added at runtime remain discoverable because misses are not cached. Disk-backed cache records are evicted after a removal is observed, allowing a later re-add to create a fresh record. Modification or replacement of an existing resource remains unsupported and is documented as a resource-lifecycle limitation.


webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerPath.java
```diff
@@ -22,117 +22,77 @@
...
+    }
+
+    @Override
+    public boolean handle(LruCache<String, CachedHandler> cache,
+                          Method method,
+                          ServerRequest request,
+                          ServerResponse response,
+                          String requestedResource) throws IOException {
+
+        if (!Files.exists(sourcePath)) {
```
tvallin, line 75: `Files.exists` alone no longer preserves the former per-request accessibility checks. An existing-but-unreadable file or a deletion race before `newByteChannel` now produces a 500 without evicting the cached metadata; HEAD can return 200 with stale metadata without opening the file at all. Please restore consistent validation and eviction for GET and HEAD, and cover access loss plus deletion/re-add races.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerPath.java
```diff
@@ -22,117 +22,77 @@
...
-            if (followLinks && secureRoot == null) {
-                lastModified = lastModified().apply(resolvedPath).orElse(null);
-            } else {
-                lastModified = FileBasedContentHandler.lastModified(resolvedPath, followLinks, secureRoot).orElse(null);
-            }
-        } catch (IOException e) {
+        Instant modified = attributes.lastModifiedTime().toInstant();
+        return new CachedHandlerPath(path,
+                                     resolvedPath,
+                                     StaticContentMetadata.create(mediaType, modified, attributes.size()),
```
tvallin, line 63: This metadata snapshot can survive a classpath handler stop/start cycle. `ClassPathContentHandler.releaseCache()` does not call `super.releaseCache()`, so dynamically discovered `file:` entries remain cached. If the file changes while stopped, the restarted service sends new bytes using the old Content-Length, ETag, and range size. Please clear the superclass caches and add a stop → modify → restart regression test.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/StaticContentHandler.java
```diff
@@ -398,30 +437,11 @@ LruCache<String, CachedHandler> handlerCache() {
...
-            inMemoryResource = new CachedHandlerInMemory(contentType,
-                                                         lastModified.get(),
-                                                         (headers, instant) -> headers.set(lastModifiedHeader),
-                                                         bytes,
-                                                         contentLength,
-                                                         contentLengthHeader);
-        }
+        CachedHandlerInMemory inMemoryResource =
+                new CachedHandlerInMemory(StaticContentMetadata.create(contentType,
+                                                                       lastModified.orElse(null),
```
tvallin: `cacheInMemory` still accepts `Optional<Instant>` as an input, and this newly added line converts absence to `null`. Helidon’s mandatory Optional style reserves `Optional` for return values and disallows null sentinels. Please use present/absent overloads, unwrap the two callers, and provide a no-last-modified metadata overload.

webserver/static-content/src/main/java/io/helidon/webserver/staticcontent/CachedHandlerPath.java
```diff
@@ -22,118 +22,99 @@
...
-                    }
-                } else {
-                    contentLength = Files.size(resolvedPath);
-                }
-                response.headers().set(HeaderValues.create(HeaderNames.CONTENT_LENGTH, contentLength));
-            } catch (IOException e) {
-                cache.remove(requestedResource);
-                throw new ForbiddenException("File is not accessible", e);
+        try (SeekableByteChannel openChannel = channel) {
+            processPreconditions(metadata, request.headers(), response.headers());
```
danielkec, line 108: While request A retains this cached handler, request B can observe the file missing and evict it; after the file is re-added, A can open the new file but still reaches this check with the pre-removal `metadata`. This can return 304 for the old ETag or later advertise/use the old length for the new bytes, even though the removal was observed:

```text
cache H for "old" (3 bytes); A retains H
delete file; B requests it and evicts H after observing the miss
write "new content" (11 bytes)
A resumes H.handle(HEAD) -> Content-Length: 3
```

tvallin (approved):

danielkec (approved):
Reviewed exact head 7a4eb4a2273db082dd068264dc4a29df9b3b3914.

PR #12292, opened 2026-08-13
Stabilize gRPC heartbeat rejection test

tomas-langer:
Stabilizes `GrpcClientCallLifecycleTest` by deterministically saturating its caller-runs executor before the heartbeat is scheduled. This prevents the setup task from being rejected and blocking the JUnit thread indefinitely, while still verifying heartbeat work runs on the dispatch thread.

Adds an outer test timeout so future lifecycle regressions fail within the test.


tvallin (approved):

danielkec (approved):
Reviewed exact head 2ea666b973007ef95aa5e7be01effccd74a7c198; no actionable findings.

PR #12294, opened 2026-08-13
Optimize header token matching

tomas-langer:
Optimize `Headers.containsToken` to scan raw header values without allocating token collections or trimmed strings.

Add semantic coverage for repeated values, quoted commas, whitespace, empty tokens, and case-insensitive matching, together with focused JMH coverage for the common HTTP/1 and HTTP/2 paths.


danielkec (approved):
Reviewed at exact head 342845d7b3a2009596d1e5236b6d114d3b228156.

PR #12295, opened 2026-08-13
Preserve direct handler response metadata

tomas-langer:
Pre-requisite for #12058

Preserve direct-handler response metadata through HTTP/1.1 and HTTP/2 response processing. Apply normal response transformations to HEAD while suppressing response data, and keep no-content response framing consistent.

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
@@ -105,7 +105,7 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
             if (hasStreamFilter()) {
                 // in this case we must honor user's request to filter the stream
                 try (OutputStream os = outputStream()) {
-                    if (!isNoEntityStatus(status())) {
+                    if (!outputStream.noEntityResponse) {
                         os.write(entityBytes, position, length);
                     }
                 } catch (IOException e) {
@@ -126,26 +126,14 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
                                                     false,
                                                     DateTime.rfc1123String()));
 
-            int initialStatusCode = status().code();
-            prepareResponse();
+            boolean noEntityResponse = prepareResponse();
 
             int actualLength = length;
             int actualPosition = position;
             byte[] actualBytes = entityBytes;
 
             Status responseStatus = status();
-            int statusCode = responseStatus.code();
-            boolean statusChanged = statusCode != initialStatusCode;
-            boolean noEntityResponse = isNoEntityStatus(responseStatus);
-            if (statusChanged && responseStatus.family() == Status.Family.INFORMATIONAL) {
-                LOGGER.log(System.Logger.Level.ERROR,
-                           "Attempt to send a final informational response. "
-                                   + "Server responded with Internal Server Error.");
-                status(Status.INTERNAL_SERVER_ERROR_500);
-                headers.set(HeaderValues.CONTENT_LENGTH_ZERO);
-                headers.remove(HeaderNames.TRAILER);
-                noEntityResponse = true;
-            } else if (noEntityResponse) {
+            if (noEntityResponse) {
                 normalizeNoEntityHeaders(headers, responseStatus);
             } else {
                 // handle content encoding
@@ -228,27 +216,38 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
             headers.add(STREAM_TRAILERS);
         }
 
-        prepareResponse();
+        boolean noEntityResponse = prepareResponse();
         streamingEntity = true;
 
         outputStream = new BlockingOutputStream(request, this, () -> this.isSent = true, () -> {
             this.isSent = true;
             afterSend();
-        }, beforeTrailers());
-        if (isNoEntityStatus(status())) {
+        }, beforeTrailers(), noEntityResponse);
+        if (noEntityResponse) {
             return new ApplicationOutputStream(outputStream, outputStream);
         }
         OutputStream encodedOutputStream = contentEncode(outputStream);
         return new ApplicationOutputStream(applyStreamFilters(encodedOutputStream), outputStream);
     }
 
-    private void prepareResponse() {
+    private boolean prepareResponse() {
         preparingResponse = true;
         try {
             beforeSend();
         } finally {
             preparingResponse = false;
         }
+        if (status().family() == Status.Family.INFORMATIONAL) {
+            LOGGER.log(System.Logger.Level.ERROR,
+                       "Attempt to send a final informational response. "
+                               + "Server responded with Internal Server Error.");
+            status(Status.INTERNAL_SERVER_ERROR_500);
+            headers.set(HeaderValues.CONTENT_LENGTH_ZERO);
+            headers.remove(HeaderNames.TRANSFER_ENCODING);
+            headers.remove(HeaderNames.TRAILER);
+            return true;
+        }
+        return isNoEntityStatus(status());
     }
 
     @Override
@@ -364,6 +363,7 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         private final Http2ServerResponse response;
         private final Http2ServerStream stream;
         private final boolean headRequest;
+        private final boolean noEntityResponse;
 
         private BufferData firstBuffer;
         private boolean closed;
@@ -377,7 +377,8 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
                                      Http2ServerResponse response,
                                      Runnable responseSentRunnable,
                                      Runnable responseCloseRunnable,
-                                     Consumer<ServerResponseTrailers> beforeTrailers) {
+                                     Consumer<ServerResponseTrailers> beforeTrailers,
+                                     boolean noEntityResponse) {
             this.request = request;
             this.response = response;
             this.headers = response.headers;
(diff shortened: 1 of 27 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -118,6 +119,7 @@ static void nonEntityBytes(ServerResponseHeaders headers,
 
         status = status == null ? Status.OK_200 : status;
 
+        normalizeNoContentHeaders(headers, status);
```
danielkec: When an early HTTP/1 request exception reaches a custom direct handler that returns `204` with a non-empty entity, this removes `Content-Length` before the no-entity validation, while `Http1Connection.handleRequestException` still appends the entity after the header section. With entity `X`, the exact-head response is (Date omitted):

```text
HTTP/1.1 204 No Content\r\n
Connection: close\r\n
\r\n
X
```

A 204 ends at the header terminator, so `X` is outside valid response framing. The merge base instead trips the existing guard and emits a framed 500, making this a regression in this change.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
```diff
@@ -434,6 +471,13 @@ private void write(BufferData buffer) throws IOException {
             if (closed) {
                 throw new IOException("Stream already closed");
             }
+            if (status.code() == Status.NO_CONTENT_204.code() && buffer.available() > 0) {
```
tvallin: `send()` routes responses with an installed `streamFilter` through `outputStream()`, even for an empty 204. With `Accept-Encoding: gzip`, constructing `GZIPOutputStream` immediately writes framing bytes to this sink, so this guard throws while serving a valid `status(204); send()`. Tracing installs such an output filter, making this reachable in normal deployments. Please bypass filters and encoders for 204 as the HTTP/1 path does, or discard transport-generated encoder bytes, and add an HTTP/2 test covering 204, a stream filter, and negotiated gzip.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -834,6 +889,7 @@ private void sendFirstChunkOnly() {
 
             // at this moment, we must send headers
             Status usedStatus = status.get();
+            normalizeNoContentHeaders(headers, usedStatus);
```
tvallin: `isNoEntityStatus` is captured when the output stream is created, but HTTP/1 allows the status to change afterward. If a route obtains the stream at 200, buffers bytes, and then changes the status to 204, this removes the framing while `sendFirstChunkOnly()` still appends `firstBuffer` below. On a persistent connection, those bytes can corrupt the next response because a 204 ends after its headers. Please re-check the final status before writing buffered data, or reject late status changes, and add coverage for `outputStream()` followed by `status(204)` and commit.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
```diff
@@ -434,6 +471,13 @@ private void write(BufferData buffer) throws IOException {
...
+                throw new IllegalStateException("Attempting to write data on a response with status " + status);
+            }
+            if (headRequest) {
```
danielkec, line 499: When a HEAD route streams from a long-lived source, this branch makes every write return after only counting the bytes, while `flush()` also returns without emitting HEADERS. Because routing commits only after the handler finishes, the client receives no response until the source ends and the producer no longer encounters HTTP/2 flow-control backpressure; repeated HEAD requests can therefore keep producer work running indefinitely. The HTTP/1 streaming path has the same count-and-return behavior. A minimal case is:

```java
var output = response.outputStream();
while (source.hasNext()) {
    output.write(source.next());
    output.flush();
}
```

[RFC 9110 section 9.3.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.2) permits omitting fields determined only while generating content and notes that avoiding content generation is preferable for HEAD.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -574,6 +584,14 @@ private static boolean isNoEntityStatus(Status status) {
                 || code == Status.NOT_MODIFIED_304.code();
     }
 
+    private static void normalizeNoContentHeaders(ServerResponseHeaders headers, Status status) {
+        if (status.code() == Status.NO_CONTENT_204.code()) {
+            headers.remove(HeaderNames.CONTENT_LENGTH);
```
danielkec, line 579: When HTTP/1 response buffering is disabled with `writeBufferSize(0)`, removing `Content-Length` here makes a 204 output stream enter the chunked path. This zero-length sequence:

```java
response.status(Status.NO_CONTENT_204);
OutputStream out = response.outputStream();
out.write(new byte[0]);
out.write(new byte[0]);
response.commit();
```

writes `0\r\n\r\n0\r\n\r\n` after the 204 header section without `Transfer-Encoding`, so the extra bytes corrupt the next response on a persistent connection because the 204 ended at the header terminator.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -698,13 +698,20 @@ private boolean handleRequestException(RequestException exception) {
                                                                   exception.responseHeaders(),
                                                                   message);
 
+        Status responseStatus = response.status();
         ServerResponseHeaders headers = response.headers();
-        byte[] entity = response.entity().orElse(BufferData.EMPTY_BYTES);
-        if (entity.length != 0) {
+        byte[] entity;
+        if (responseStatus.code() == Status.NO_CONTENT_204.code()) {
```
danielkec: The direct-handler response can also use 205 or 304, but this 204-only condition lets a non-empty entity fall through to an END_STREAM DATA frame for those no-content statuses. A focused variant of `rejectedStreamSuppressesNoContentEntityAndFraming` using:

```java
.status(Status.RESET_CONTENT_205)
.entity("ignored")
```

recorded `Content-Length: 7` followed by a 7-byte END_STREAM DATA frame. The merge base kept the original request-exception status (400 for this trigger), so this invalid HTTP/2 response is introduced by preserving the direct-handler status.

common/testing/http-junit5/src/main/java/io/helidon/common/testing/http/junit5/SocketHttpClient.java
```diff
@@ -172,12 +173,12 @@ public static ClientResponseHeaders headersFromResponse(String response) {
         WritableHeaders<?> headers = WritableHeaders.create();
 
         assertThat(response, notNullValue());
-        int index = response.indexOf("\n\n");
+        int index = response.indexOf(END_OF_HEADERS);
```
danielkec: `receive()` now preserves LF-only response line endings, while this parser only accepts the CRLF delimiter. For example:

```java
String response = "HTTP/1.1 204 No Content\nConnection: close\n\n";
SocketHttpClient.headersFromResponse(response);
```

This throws `AssertionError: Missing end of headers in response!`, and `entityFromResponse(response, true)` fails the same way. The merge base accepted this LF response, so a response returned by `receive()` can no longer be consumed by the class's own public parsing helpers.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -698,13 +698,29 @@ private boolean handleRequestException(RequestException exception) {
...
+            }
+            headers.remove(HeaderNames.TRANSFER_ENCODING);
+            headers.remove(HeaderNames.TRAILER);
+        } else {
+            entity = response.entity().orElse(BufferData.EMPTY_BYTES);
             headers.set(HeaderValues.create(HeaderNames.CONTENT_LENGTH, String.valueOf(entity.length)));
         }
         Http2Headers http2Headers = Http2Headers.create(headers)
-                .status(exception.status());
+                .status(responseStatus);
```
tvallin, line 739: `TransportResponse.status()` can be informational, but this path now writes that status as the final response. Because 1xx statuses are not included in `noEntityResponse`, an empty entity produces HEADERS with END_STREAM below, while a non-empty entity also produces DATA. HTTP/2 informational HEADERS must not end the stream or carry response content and require a subsequent final response; 101 is prohibited entirely. The base path used the request exception’s final status, so this is introduced by preserving the handler status. Please reject or convert informational statuses before framing and add a rejected-stream test using 103.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -698,13 +698,29 @@ private boolean handleRequestException(RequestException exception) {
...
+            if (noContentResponse) {
+                headers.remove(HeaderNames.CONTENT_LENGTH);
+            } else if (responseStatusCode == Status.RESET_CONTENT_205.code()) {
+                headers.set(HeaderValues.CONTENT_LENGTH_ZERO);
+            }
+            headers.remove(HeaderNames.TRANSFER_ENCODING);
+            headers.remove(HeaderNames.TRAILER);
+        } else {
+            entity = response.entity().orElse(BufferData.EMPTY_BYTES);
             headers.set(HeaderValues.create(HeaderNames.CONTENT_LENGTH, String.valueOf(entity.length)));
```
tvallin: This unconditional assignment still loses valid metadata for an early rejected HEAD request. If the direct handler supplies no entity but sets `Content-Length: 23`, the entity is empty and this overwrites the representation length with 0. The merge base only replaced Content-Length for a non-empty entity, and the regular `DirectHandlers.handle` path explicitly preserves this absent-entity HEAD case. Please apply the same rule using `exception.request().method()` and add a rejected HEAD test with an explicit Content-Length and no entity.

danielkec (approved):

tvallin (approved):

PR #12297, opened 2026-08-14
4.x: Support OCI-managed private-key certificate bundles

### LINKED ISSUE #12296: 4.x: Support OCI-managed private-key certificate bundles in the OCI TLS manager
Helidon Version: 4.x

## Task Description

Backport [helidon-extensions PR #106](https://github.com/helidon-io/helidon-extensions/pull/106) to `helidon-4.x`.

The 4.x OCI TLS Certificates manager has the same two-source design described by
[helidon-extensions issue #105](https://github.com/helidon-io/helidon-extensions/issues/105): it downloads the public
certificate and chain from OCI Certificates, then obtains the private key separately through KMS `ExportKey`. This
requires an exportable software-protected Vault key and prevents applications from directly using an OCI-issued,
OCI-managed leaf certificate with OCI Certificates automatic renewal.

The limitation was reproduced with Helidon 4.3.2 and is still present on `helidon-4.x`. The required OCI SDK APIs are
already available in the 4.x dependency set, so no OCI SDK upgrade is needed.

## Backport scope

- Add an `oci-certificate-bundle-tls-manager` provider that retrieves the current certificate, chain, matching private
  key, and optional passphrase from one `CERTIFICATE_CONTENT_WITH_PRIVATE_KEY` bundle.
- Preserve the existing `oci-certificates-tls-manager` Vault export path and configuration unchanged.
- Support RSA and EC PKCS#8 private keys and verify that the downloaded private key matches the leaf certificate.
- Reload identity and CA trust transactionally, retain the last working TLS material after refresh failures, and
  serialize concurrent refreshes.
- Use change detection for certificate bundles by default, with `always-reload` available to force rebuilding on every
  poll. Preserve the existing Vault manager's reload behavior.
- Backport the focused configuration, downloader, parsing, rotation, concurrency, retry, and Vault-regression tests.
- Update the 4.x module documentation with both configuration modes, IAM requirements, renewal behavior, and private-key
  handling.

## Compatibility

Existing valid Vault YAML and programmatic builder usage must continue to work. The backport should adapt the forward
change to the released 4.x configuration API where necessary instead of blindly copying 27.x API changes.

The forward PR changes three configuration getters to `Optional`, but those getters are released public API on 4.x.
The backport therefore uses a separate provider/configuration for the new bundle path, retaining the existing
`URI vaultCryptoEndpoint()`, `String keyOcid()`, and `Supplier<char[]> keyPassword()` descriptors. The generic lazy
config-factory metadata limitation remains tracked separately by #12288 and is not part of this backport.

## Backport PR

- https://github.com/helidon-io/helidon/pull/12297

## Source

- Forward issue: https://github.com/helidon-io/helidon-extensions/issues/105
- Forward PR: https://github.com/helidon-io/helidon-extensions/pull/106
- Merged forward commit: https://github.com/helidon-io/helidon-extensions/commit/c3a09c3854ac513cfa102e5054dd4234aa7a5491

danielkec:
Fixes #12296

### Description

Backport of https://github.com/helidon-io/helidon-extensions/pull/106 to `helidon-4.x`.

Adds a separate `oci-certificate-bundle-tls-manager` provider that retrieves the current
`CERTIFICATE_CONTENT_WITH_PRIVATE_KEY` bundle from OCI Certificates. It supports encrypted and unencrypted RSA/EC
PKCS#8 keys and verifies at the shared downloader-factory boundary that the private key matches the leaf certificate.

Polling checks the public current version before retrieving private material, rejects public/private version races, and
installs certificate-version and CA changes as one complete TLS generation. A stable manager context routes each new
socket or engine to the latest fully initialized generation, while in-flight connections remain consistent. Each update
also replaces the TLS session caches so a client authenticated by a retired CA cannot resume its old session. Failed
refreshes retain the last working material for retry.

Initialization is serialized and idempotent for provider-cached managers. Polling uses Helidon's managed scheduler and
is stopped by `TaskManager.shutdown()`. Incompatible outer TLS context settings on a shared manager and the inherited
non-atomic external `Tls.reload` path fail explicitly.

The separate provider and configuration are intentional for 4.x. The forward change can update unreleased
configuration getter return types to `Optional`, but doing that on the released 4.x API would break existing clients.
The existing `oci-certificates-tls-manager` Vault implementation and public configuration descriptors remain unchanged.
The downloader SPI extension is additive and binary compatible.

### Documentation

Updates the OCI TLS Certificates README with both provider keys, configuration examples, the required OCI SDK HTTP
provider and JPMS requirement, IAM conditions, automatic renewal behavior, reload semantics, and private-key handling.
Generated metadata declares the keyed provider prefix and `TlsManager` contract, and generated configuration docs link
the provider from all TLS-manager usages.

### Testing

- Clean JDK 21 OCI TLS module verify with tests and Javadocs: 33 tests passed.
- Public-behavior coverage includes:
  - a provider-shared manager used by two `Tls` instances with one stable context and one scheduled task;
  - cached socket-factory RSA-to-EC rotation with atomic identity/trust publication;
  - TLS 1.2 mTLS session resumption before CA rotation and rejection after replacing the trust anchor;
  - public/private version-race rejection and retry retention;
  - managed scheduler shutdown;
  - a custom downloader returning a mismatched key pair;
  - a downstream named-module compilation requiring only this module.
- JDK 21 SpotBugs: zero findings and zero errors.
- Full 466-project tests-skipped install plus docs generation: committed OCI/TLS generated docs reproduced exactly.
- Existing SE-1 native-image jobs now include the OCI TLS module and explicitly initialize the former
  eager-`SecureRandom` downloader class at image build time; the local GraalVM build and both native integration tests pass.
- `git diff --check`


integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+        super(name, TYPE);
+        this.cfg = Objects.requireNonNull(cfg);
+
+        if (config instanceof Config watchableConfig) {
+            watchableConfig.onChange(this::config);
+        }
+    }
+
+    @Override
+    public void init(TlsConfig tls) {
```
tvallin, line 81: `TlsManagerProvider.getOrCreate` can share this manager between equal configurations, but every `init` builds a new inherited `SSLContext` and starts another poller. On 4.x, an earlier `Tls` retains the old context while later refreshes update only the replacement context, so the first socket never receives certificate rotations. Please make initialization serialized and idempotent with one stable context and scheduler, and add a shared-manager, two-`Tls` rotation test.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/spi/OciCertificatesDownloader.java
```diff
@@ -63,6 +80,36 @@ static Certificates create(String version,
...
+
+        X509Certificate[] certificateCopy = Objects.requireNonNull(certificates, "Certificates are required").clone();
+        if (certificateCopy.length == 0) {
+            throw new IllegalArgumentException("At least one certificate is required");
+        }
+        if (Arrays.stream(certificateCopy).anyMatch(Objects::isNull)) {
+            throw new NullPointerException("Certificates must not contain null elements");
+        }
+
+        return new CertificatesWithPrivateKey(version,
```
tvallin, line 108: The matching check exists only in `DefaultOciCertificatesDownloader`; this public factory accepts a mismatched leaf/private key from any custom downloader despite documenting the opposite. PKCS12 and the default `KeyManagerFactory` accept such a pair, so the manager publishes it, records the version as installed, and subsequent TLS handshakes fail without retrying that version. Please enforce the match at the shared factory or manager boundary and add a custom-downloader mismatch test.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificatesDownloader.java
```diff
@@ -47,6 +54,8 @@
 @Service.Provider
 @Weight(Weighted.DEFAULT_WEIGHT - 10)
 class DefaultOciCertificatesDownloader implements OciCertificatesDownloader {
+    private static final SecureRandom RANDOM = new SecureRandom();
```
tvallin, line 57: This eager `SecureRandom` is incompatible with Helidon native-image initialization. MP native-image configuration initializes `io.helidon` at build time, and existing TLS code deliberately wraps `SecureRandom` in `LazyValue` because an instance cannot be stored in the image heap. Please initialize it lazily at runtime—or use a fixed challenge, because unpredictability is unnecessary for key-pair matching—and add native-image coverage.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+        this.tlsConfig = tls;
+        ServiceRegistry registry = GlobalServiceRegistry.registry();
+        this.certDownloader = registry.supply(OciCertificatesDownloader.class);
+        ScheduledExecutorService asyncExecutor = Executors.newSingleThreadScheduledExecutor();
```
tvallin, line 85: This creates a non-daemon executor that is neither retained nor shut down. `Cron.close()` and `TaskManager.shutdown()` cancel the task but do not stop a caller-supplied executor, and `TlsManager` has no close hook. The thread can therefore outlive the server and prevent JVM termination; the tests manually shut it down, masking the production leak. Please use managed scheduling or introduce an owned lifecycle and test shutdown.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /**
+     * Triggers a reload after a backing configuration change.
+     *
+     * @param config the updated config
+     */
+    void config(io.helidon.common.config.Config config) {
+        Objects.requireNonNull(config);
+        if (loadContext(false)) {
```
tvallin, line 138: This callback ignores the updated `config`; `loadContext` continues reading the constructor-captured immutable `cfg`. Changes to `cert-ocid`, `ca-ocid`, `always-reload`, or `schedule` therefore reload the old configuration. Please apply the new prototype atomically, including rescheduling when necessary, or remove the change listener. Add a test that changes the OCIDs and verifies the downloader receives them.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/OciCertificateBundleTlsManagerConfigBlueprint.java
```diff
@@ -0,0 +1,61 @@
...
+package io.helidon.integrations.oci.tls.certificates;
+
+import io.helidon.builder.api.Option;
+import io.helidon.builder.api.Prototype;
+
+/**
+ * Blueprint configuration for {@link OciCertificateBundleTlsManager}.
+ */
+@Prototype.Blueprint
+@Prototype.Configured
```
tvallin, line 26: This configured runtime type declares neither its keyed prefix nor its provider contract. Builder codegen warns that the configured provider is undeclared, and generated metadata contains neither the `oci-certificate-bundle-tls-manager` prefix nor `TlsManagerProvider` in `provides`, so config tooling cannot associate it with `TlsConfig.manager`. Please use keyed, non-root `@Prototype.Configured(...)`, add `@Prototype.Provides(TlsManagerProvider.class)`, and assert both metadata fields.

integrations/oci/tls-certificates/README.md
```diff
@@ -1,35 +1,93 @@
...
-## Usage
-Integrating with OCI's Certificates Service from Helidon is a simple matter of configuration.
+Add the module to the application:
 
-First, use OCI's Certificates Service to create your certificates. Follow the directions [here](https://docs.oracle.com/en-us/iaas/Content/certificates/home.htm).
+```xml
+<dependency>
+    <groupId>io.helidon.integrations.oci</groupId>
+    <artifactId>helidon-integrations-oci-tls-certificates</artifactId>
+</dependency>
```
tvallin, line 18: Following this dependency snippet does not add an OCI SDK HTTP provider. The module keeps Jersey 3 test-scoped, while the OCI service dependencies provide only the HTTP API, so `CertificatesClient.builder().build(...)` fails before its first request unless users independently know to select a provider. Please document the required application dependency here, including the JPMS requirement.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @param initialLoad whether this is the initial TLS installation
+     * @return whether TLS material was installed
+     */
+    boolean loadContext(boolean initialLoad) {
+        reloadLock.lock();
+        try {
+            OciCertificatesDownloader downloader = certDownloader.get();
+            OciCertificatesDownloader.CertificatesWithPrivateKey identity =
+                    downloader.loadCertificatesWithPrivateKey(cfg.certOcid());
```
tvallin, line 154: Every poll exports and parses the private key before checking whether the identity version changed. With the documented 30-second schedule, unchanged keys are repeatedly transmitted and materialized in heap. Please perform a public/metadata version check first, retrieve the private bundle only for a candidate update, and verify the private bundle’s version before installing it to handle rotation races.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/OciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.function.Consumer;
+
+import io.helidon.builder.api.RuntimeType;
+import io.helidon.common.tls.TlsManager;
+import io.helidon.config.Config;
+
+/**
+ * A {@link TlsManager} that loads a certificate and its matching private key from an OCI-managed certificate bundle.
+ */
+public interface OciCertificateBundleTlsManager
```
tvallin, line 28: This new exported API and its generated configuration expose types from `common.tls`, `builder.api`, `config`, `common.config`, and `common`, but `module-info.java` keeps those requirements non-transitive. Exact-head compilation emits JPMS export warnings, contrary to the 4.x development guideline requiring transitive dependencies for public API types. Please update the descriptor and add a downstream named-module compilation test.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+    private Supplier<OciCertificatesDownloader> certDownloader;
+    private TlsConfig tlsConfig;
+
+    DefaultOciCertificateBundleTlsManager(OciCertificateBundleTlsManagerConfig cfg) {
+        this(cfg, "@default", null);
+    }
+
+    DefaultOciCertificateBundleTlsManager(OciCertificateBundleTlsManagerConfig cfg,
+                                          String name,
+                                          io.helidon.common.config.Config config) {
```
tvallin, line 71: Please remove usage of FQCN.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+                    .map(X509TrustManager.class::cast)
+                    .findFirst();
+            if (trustManager.isEmpty()) {
+                throw new IllegalStateException("Unable to find X.509 trust manager in download: " + cfg.certOcid());
+            }
+
+            if (initialLoad) {
+                initSslContext(tlsConfig, secureRandom, kmf.getKeyManagers(), tmf.getTrustManagers());
+            } else {
+                reload(keyManager, trustManager);
```
tomas-langer, line 195: `reload` publishes the new key manager and trust manager sequentially through `ConfiguredTlsManager.reload`, while the live `SSLContext` calls the two reloadable delegates independently. A handshake that overlaps this call can select an alias from one identity generation and read its key or chain from the next, or authenticate the peer with a CA generation different from the server identity. The `reloadLock` only serializes refresh callers; it does not serialize JSSE callbacks. This contradicts the documented "one complete TLS update" and can produce intermittent handshake failures or mixed mTLS policy during rotation.

tomas-langer, line 195: This reload retains the existing `SSLContext` and its session cache. A TLS 1.2 mTLS client authenticated under the previous CA can resume that session after the trust-manager update without invoking the new trust manager, so the retired CA can remain effective until session expiry (24 hours by default). The PR presents CA rotation as complete once this call succeeds, but resumed connections can continue using the old authentication decision.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    /**
+     * Downloads and installs the current OCI-managed identity and CA certificate.
+     *
+     * @param initialLoad whether this is the initial TLS installation
+     * @return whether TLS material was installed
+     */
+    boolean loadContext(boolean initialLoad) {
```
tomas-langer, line 149: `loadContext` and the related reload/parser helpers are package-visible so the new tests can invoke production internals directly; `DefaultOciCertificatesDownloader.toCertificatesWithPrivateKey` is exposed in the same way. The resulting tests bind regression coverage to implementation decomposition instead of the public TLS manager and downloader behavior, while also enlarging the package-level production surface. This conflicts with `DEV-GUIDELINES.md` rules 4.1.4, 10.1, and 10.3.

integrations/oci/tls-certificates/src/test/java/io/helidon/integrations/oci/tls/certificates/OciCertificatesTlsManagerTest.java
```diff
@@ -101,5 +163,319 @@ void configIsMonitoredForChange() throws Exception {
...
+
+    @Test
+    void legacyVaultReloadsWhenVersionIsUnchanged() {
+        DefaultOciCertificatesTlsManager manager = newVaultManager();
+
+        assertThat(TestOciCertificatesDownloader.callCount_loadCertificates, equalTo(1));
+        assertThat(TestOciCertificatesDownloader.callCount_loadCACertificate, equalTo(1));
+        assertThat(TestOciPrivateKeyDownloader.callCount, equalTo(1));
+
+        assertTrue(manager.loadContext(false));
```
tomas-langer, line 188: The new tests use JUnit `assertTrue`, `assertFalse`, and `assertNotNull` assertions. `DEV-GUIDELINES.md` rules 10.5 and 10.6 require Hamcrest assertions and limit direct JUnit assertions to `assertAll` and `assertThrows`, so the changed tests violate the repository's assertion and diagnostic conventions.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/DefaultOciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        @Override
+        public String[] getSupportedCipherSuites() {
+            return delegate().getSupportedCipherSuites();
+        }
+
+        @Override
+        public ServerSocket createServerSocket() throws IOException {
+            return delegate().createServerSocket();
```
tomas-langer, line 628: `SwitchingSslServerSocketFactory` selects the current `TlsMaterial` only while creating the `SSLServerSocket`. The returned provider socket remains bound to that generation's `SSLContext`; later `installedMaterial` swaps do not reload the wrappers inside that old context. This differs from core `ConfiguredTlsManager`, which keeps one context and updates its reloadable key/trust-manager delegates. Consequently, a long-lived listener created through `Tls.createServerSocket()` continues using the retired identity, trust, and session cache until the socket is recreated. The new rotation test creates a fresh server socket for every handshake, so it does not cover this public API path. Helidon WebServer's per-connection `SSLEngine` path is not affected.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/OciCertificateBundleTlsManager.java
tomas-langer, line 32: The new manager exposes the `TlsManager` contract, but external `Tls.reload(Tls)` calls always throw `UnsupportedOperationException`. This intentional restriction appears in the PR description but is absent from the public API documentation and README, so users following the published reload contract encounter an undocumented exception.

integrations/oci/tls-certificates/src/main/java/io/helidon/integrations/oci/tls/certificates/OciCertificateBundleTlsManager.java
```diff
@@ -0,0 +1,93 @@
...
+        return OciCertificateBundleTlsManagerConfig.builder();
+    }
+
+    /**
+     * Creates a manager configured by a builder consumer.
+     *
+     * @param consumer builder consumer
+     * @return a configured instance
+     */
+    static OciCertificateBundleTlsManager create(Consumer<OciCertificateBundleTlsManagerConfig.Builder> consumer) {
```
tomas-langer, line 92: The new `create(Consumer)` factory and `DefaultOciCertificateBundleTlsManager.init(TlsConfig)` reject null through incidental dereferences rather than explicit boundary validation required by [Rule 1.2.2](https://github.com/helidon-io/helidon/blob/main/DEV-GUIDELINES.md#rule-1-2-2). In `init`, the private configuration helper dereferences the argument before the later null check can run. This is a guideline issue rather than acceptance of null input.

tomas-langer (approved):

PR #12300, opened 2026-08-18
Upgrade Kotlin and Okio dependencies

tvallin:
### Description

- Upgrade Kotlin dependency to 2.3.21 version
- Upgrade Okio dependency to 3.18.1 version

### DIFF
--- a/dependencies/pom.xml
+++ b/dependencies/pom.xml
@@ -106,7 +106,9 @@
         <version.lib.junit>5.12.2</version.lib.junit>
         <version.lib.kafka>3.9.2</version.lib.kafka>
         <!-- Kotlin: dependency convergence requirement, we never use this directly -->
-        <version.lib.kotlin>1.9.10</version.lib.kotlin>
+        <version.lib.kotlin>2.3.21</version.lib.kotlin>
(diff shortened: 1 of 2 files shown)


dependencies/pom.xml
```diff
@@ -981,11 +981,6 @@
...
                 <version>${version.lib.kotlin}</version>
             </dependency>
-            <dependency>
-                <groupId>org.jetbrains.kotlin</groupId>
-                <artifactId>kotlin-stdlib-common</artifactId>
-                <version>${version.lib.kotlin}</version>
-            </dependency>
             <dependency>
                 <groupId>org.jetbrains.kotlin</groupId>
                 <artifactId>kotlin-stdlib-jdk8</artifactId>
```
tomas-langer, line 987: This removes `org.jetbrains.kotlin:kotlin-stdlib-common` from the published `helidon-dependencies` management set. A 4.x application or downstream BOM that imports `io.helidon:helidon-dependencies` and declares this coordinate without a version will now fail Maven model validation, while transitive users lose Helidon’s Kotlin-family convergence. This is therefore a user-facing 4.x compatibility break caused by this PR, not merely a dependency version upgrade.

tvallin:
We are going to upgrade kotlin to 2.4.20 directly in September.

PR #12301, opened 2026-08-19
Share resolved WebClient targets

tomas-langer:
Pre-requisite for #3686.

## Intent

WebClient currently lets each HTTP implementation derive connection information inside its own request flow. HTTP/3 needs to make transport and cache decisions while preserving the same logical origin, TLS peer identity, SNI configuration, proxy decision, and DNS result already used by HTTP/1.1 and HTTP/2. Those values cannot be represented safely by one host and port: the authority used for HTTP and TLS can differ from the proxy or resolved socket peer, and services can finalize the request URI and `Host` header before a connection is acquired.

This change establishes a protocol-neutral target boundary before HTTP/3 is added. It keeps logical origin and TLS identity separate from route selection and physical resolution, so all protocol implementations can use the same connection identity without coupling HTTP/3 to the HTTP/1.1 or HTTP/2 implementations.

## Target model

The internal model separates three concerns:

- `ClientConnectionTarget` is the immutable logical and cache identity created after WebClient services have finalized the URI and authority. It captures TLS identity and reload generation, SNI, DNS policy, the selected proxy route, transport, and local binding.
- `ProxyRoute` records the effective direct, HTTP forward, HTTP CONNECT tunnel, or SOCKS route and its capabilities.
- `ResolvedClientTarget` describes one physical connection attempt, including the concrete peer and proxy destination, without changing the logical HTTP or TLS identity.

A DNS-free lookup key also lets a configured IP-based `no-proxy` policy reuse a healthy cached connection before performing DNS again. A cache miss still resolves current addresses and binds a direct connection to the address that caused the bypass decision.

## HTTP/1.1 and HTTP/2 adoption

HTTP/1.1 and HTTP/2 now construct their connection targets from the final service-mutated request state and cache connections by the complete target. This preserves several invariants needed by the later HTTP/3 implementation:

- TLS endpoint identification and SNI use the logical origin, never the DNS or proxy peer.
- Selected routes survive protocol probing, HTTP/1 fallback, and same-origin redirects, but a reused request re-evaluates dynamic proxy policy on its next top-level invocation.
- IP-based `no-proxy` matching uses one resolved address for both the bypass decision and the physical connection attempt; healthy cache hits do not require another DNS lookup.
- HTTP/1.1 emits `Proxy-Connection` only for an actual forward-proxy exchange.
- HTTP/2 honors `connection-cache-size` as the number of reusable physical connections retained per target and gracefully drains displaced connections without allowing stale failures to remove a replacement handler.

The corresponding cache and shared-cache Javadocs now describe their per-protocol, per-target retention semantics and distinguish retained connections from concurrent requests, streams, and temporarily draining sockets.

## Scope

The change does not add HTTP/3 or QUIC; it supplies the shared routing and cache foundation used by the follow-on implementation.
It adds no configuration keys and does not enable Alt-Svc discovery or alternative routing by itself.

### DIFF
--- a/tests/benchmark/jmh/src/main/java/io/helidon/webclient/benchmark/jmh/WebClientConnectionTargetBenchmark.java
+++ b/tests/benchmark/jmh/src/main/java/io/helidon/webclient/benchmark/jmh/WebClientConnectionTargetBenchmark.java
@@ -78,6 +78,14 @@ public class WebClientConnectionTargetBenchmark {
         }
     }
 
+    private static String[] targetUris(int targetCount, int serverPort) {
+        String[] result = new String[targetCount];
+        for (int i = 0; i < targetCount; i++) {
+            result[i] = "http://target-" + i + ".example:" + serverPort + PATH;
+        }
+        return result;
+    }
+
     @State(Scope.Benchmark)
     public static class ServerState {
         private WebServer server;
@@ -179,13 +187,5 @@ public class WebClientConnectionTargetBenchmark {
         private String targetUri(int index) {
             return targetUris[index];
         }
-    }
-
-    private static String[] targetUris(int targetCount, int serverPort) {
-        String[] result = new String[targetCount];
-        for (int i = 0; i < targetCount; i++) {
-            result[i] = "http://target-" + i + ".example:" + serverPort + PATH;
-        }
-        return result;
     }
 }
--- a/tests/benchmark/jmh/src/main/java/io/helidon/webclient/benchmark/jmh/WebClientConnectionTargetBenchmark.java
+++ b/tests/benchmark/jmh/src/main/java/io/helidon/webclient/benchmark/jmh/WebClientConnectionTargetBenchmark.java
@@ -40,6 +40,7 @@ import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.infra.BenchmarkParams;
 import org.openjdk.jmh.infra.Blackhole;
 import org.openjdk.jmh.infra.ThreadParams;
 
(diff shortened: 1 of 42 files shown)


webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ConnectionCache.java
```diff
@@ -75,14 +96,66 @@ boolean supports(ConnectionKey ck) {
...
-    void remove(ConnectionKey connectionKey) {
-        if (!closed.get()) {
-            closeAndRemove(connectionKey);
+    void remove(ClientConnectionTarget connectionTarget, Http2ClientConnectionHandler expectedHandler) {
+        cacheLock.lock();
+        try {
+            if (cache.remove(connectionTarget, expectedHandler)) {
+                insertionOrder.remove(connectionTarget, expectedHandler);
+                removeLookupHandler(connectionTarget, expectedHandler);
+                http2Supported.remove(connectionTarget.connectionKey());
```
danielkec: When two `ClientConnectionTarget`s share one `ConnectionKey`, removing one timed-out handler here also clears the key-wide `http2Supported` marker even if the sibling handler remains cached. A minimal cache-state probe is: prime handlers for Host A and B under one key, call `remove(A, handlerA)`, and observe that `supports(key)` becomes false while B is still present. Generic cleartext WebClient discovery therefore falls back to HTTP/1.1 for B (and an h2c-only endpoint can fail) despite its reusable HTTP/2 handler.

tests/benchmark/jmh/src/main/java/io/helidon/webclient/benchmark/jmh/WebClientConnectionTargetBenchmark.java
```diff
@@ -0,0 +1,320 @@
...
+            }
+        }
+    }
+
+    @Benchmark
+    @OperationsPerInvocation(REQUESTS_PER_INVOCATION)
+    public void http1RotatingTargetsPerThread(Http1State state, ThreadParams threadParams, Blackhole blackhole) {
+        int firstTargetIndex = threadParams.getThreadIndex() * TARGETS_PER_THREAD;
+        for (int i = 0; i < REQUESTS_PER_INVOCATION; i++) {
+            int targetIndex = firstTargetIndex + i % TARGETS_PER_THREAD;
```
danielkec, line 105: With the declared `prefilledTargetCount=1` parameter, this rotating workload computes target index 1 on its second request even though setup allocated a one-element `targetUris` array, so this benchmark configuration aborts before producing a score. The HTTP/2 rotating variant has the same mismatch. A minimal case with `WebClientConnectionTargetJmhRunnerTest` is:

```text
webclient.target.jmh.include=.*http1RotatingTargetsPerThread
webclient.target.jmh.targetCounts=1
webclient.target.jmh.proxyModes=NONE
webclient.target.jmh.threads=1
-> ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
```

danielkec (approved):

PR #12316, opened 2026-08-22
Use TypeName class names for JSON converter field suffixes

### LINKED ISSUE #11399: JSON Binding generator uses `toString()` on TypeName
`io.helidon.json.codegen.JsonConverterGenerator#ensureUpperStart(io.helidon.common.types.TypeName)` uses `toString` on type name, which may cause issues.

This can lead to fields named `serializerMyTestPerson>` (this was a symptom of another problem, but can be reproduced by adding `@Json.Entity` on a Helidon `BuilderBase` class.

tanvir-ux:
Fixes #11399

`JsonConverterGenerator.ensureUpperStart` built field suffixes from `TypeName.toString()`. That string includes generic syntax, so types such as `BuilderBase<MyTestPerson>` became identifiers like `serializerMyTestPerson>`.

Build the suffix from class name, enclosing names, type arguments, and array flag instead, and drop characters that are not valid in a Java identifier.

Tested with `JsonConverterGeneratorTest` (`mvn -Ptests -pl tests/codegen test -Dtest=JsonConverterGeneratorTest` from `json/`).

### DIFF
--- a/json/codegen/src/main/java/io/helidon/json/codegen/JsonConverterGenerator.java
+++ b/json/codegen/src/main/java/io/helidon/json/codegen/JsonConverterGenerator.java
@@ -1248,13 +1248,45 @@ class JsonConverterGenerator {
         }
     }
 
-    private static String ensureUpperStart(TypeName typeName) {
-        String className = typeName.toString();
-        int index = className.lastIndexOf(".");
-        if (index > -1) {
-            className = className.substring(index + 1);
+    /**
+     * Build a Java identifier suffix for converter fields from a type name.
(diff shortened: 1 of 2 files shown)


tanvir-ux:
OCA update: I signed in to https://oca.opensource.oracle.com/ and confirmed my Individual Agreement is submitted and currently **Under review** (submission date 2026-08-22; GitHub username `tanvir-ux`, email `tanviralamshuvo@gmail.com`). Happy to wait on Oracle approval — please ping if anything else is needed on my side.

barchetta:
@tanvir-ux thanks for the contribution. We are following up on the OCA.

PR #12320, opened 2026-08-24
Update development guidelines

tomas-langer:
Update the development guidelines to make code-review rules available to contributors.

Stable rule identifiers let contributors, reviewers, and related repositories cite or override individual rules without
copying their text or depending on headings that may change. Removed identifiers remain as tombstones so existing references
cannot silently acquire a different meaning.

The guidelines now:

- clarify Java, API, test, package, builder, and synchronization conventions;
- assign stable rule identifiers and anchors, with tombstones for removed rules;
- preserve legacy section anchors; and
- treat configured Checkstyle and copyright failures as guideline violations.

### DIFF
--- a/DEV-GUIDELINES.md
+++ b/DEV-GUIDELINES.md
@@ -16,6 +16,11 @@ not renumber or reuse them when rule names, text, or chapter organization change
 <a id="rule-0-3"></a>**Rule 0.3 — Preserve removed rules.** When removing a rule, keep its identifier and anchor in place
 with the text `Removed. This identifier is reserved.` An optional note may identify replacement rules.
 
+<a id="rule-0-4"></a>**Rule 0.4 — Treat repository checks as guidelines.** A failure from the repository-configured
+Checkstyle or copyright validation is a guideline violation even when this document does not enumerate the underlying
+requirement.
+
+<a id="general-coding-rules"></a>
 <a id="chapter-1"></a>
 ## 1. General coding rules
 
@@ -85,6 +90,7 @@ type name unless a name conflict or Java syntax requires a fully qualified name.
 <a id="rule-3-2"></a>**Rule 3.2 — Do not import types only for Javadoc.** Use the fully qualified name in Javadoc
 instead.
 
+<a id="package-and-module-structure"></a>
 <a id="chapter-4"></a>
 ## 4. Package and module structure
 
@@ -117,6 +123,7 @@ different concerns separate.
   `io.helidon.${project_module}.${module_name}`, such as `io.helidon.security` or
   `io.helidon.security.providers.common`.
 
+<a id="configuration-and-programmatic-api"></a>
 <a id="chapter-5"></a>
 ## 5. Configuration and programmatic API
 
@@ -146,6 +153,7 @@ node containing the component's configuration, such as `ServerConfiguration` in 
 
 See [Builders](#chapter-8) and the [helidon-builder documentation](builder/README.md).
 
+<a id="getters-and-setters"></a>
 <a id="chapter-6"></a>
 ## 6. Getters and setters
 
@@ -160,6 +168,7 @@ use `port(int newPort)` and `int port()`.
 
 Example: [io.helidon.security.providers.oidc.common.OidcConfig](security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java).
 
+<a id="fluent-api"></a>
 <a id="chapter-7"></a>
 ## 7. Fluent API
 
@@ -169,6 +178,7 @@ Example: [io.helidon.security.providers.oidc.common.OidcConfig](security/provide
 - <a id="rule-7-1-2"></a>**Rule 7.1.2.** Use fluent APIs for control methods, such as
   `Server server = Server.create().start()`.
 
+<a id="builders"></a>
 <a id="chapter-8"></a>
 ## 8. Builders
 
@@ -216,6 +226,7 @@ follow these rules:
 
 Example: [io.helidon.faulttolerance.RetryConfigBlueprint](fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/RetryConfigBlueprint.java).
 
+<a id="jpms"></a>
 <a id="chapter-9"></a>
 ## 9. JPMS
 
@@ -229,6 +240,7 @@ Example: [io.helidon.faulttolerance.RetryConfigBlueprint](fault-tolerance/fault-
 <a id="rule-9-4"></a>**Rule 9.4 — Qualify service types.** Use fully qualified class names for provided and used
 services.
 
+<a id="testing"></a>
 <a id="chapter-10"></a>
 ## 10. Testing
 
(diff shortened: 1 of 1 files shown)


tvallin (dismissed):

DEV-GUIDELINES.md
```diff
@@ -1,206 +1,346 @@
...
+switch code.
+
+- <a id="rule-1-6-1"></a>**Rule 1.6.1.** Make switches exhaustive for enums and sealed hierarchies.
+
+<a id="chapter-2"></a>
+## 2. Class member order
+
+<a id="rule-2-1"></a>**Rule 2.1 — Order field groups.** Place static fields before instance fields.
+
+- <a id="rule-2-1-1"></a>**Rule 2.1.1.** Within each group, order fields by visibility: public, protected, package-private,
```
danielkec: Rule 2.1.1 conflicts with the repository's configured IDEA arrangement, which places final fields before mutable fields inside both the static and instance groups. For example, the configured order is:

```java
private static final Object CONSTANT = new Object();
public static Object mutable;
```

while this rule orders the public field first. Contributors can therefore satisfy the new citable rule or the project formatter, but not both.

DEV-GUIDELINES.md
```diff
@@ -1,206 +1,346 @@
...
+  then private.
+
+<a id="rule-2-2"></a>**Rule 2.2 — Order public methods.** Place public static methods before public instance methods.
+
+<a id="rule-2-3"></a>**Rule 2.3 — Put nested types last.** Place nested member types after fields, initializer blocks,
+constructors, and methods.
+
+- <a id="rule-2-3-1"></a>**Rule 2.3.1.** Order nested types by visibility: public, protected, package-private, then private.
```
danielkec: Rule 2.3.1 conflicts with the configured arrangement, which groups nested declarations by kind (`enum`, `interface`, static class, then class), not by visibility. With:

```java
private enum State {}
public interface Api {}
```

IDE rearrange puts the enum first, while this rule requires the public interface first, so the review guidance and formatter disagree.

DEV-GUIDELINES.md
```diff
@@ -1,206 +1,346 @@
...
+- <a id="rule-8-4-1"></a>**Rule 8.4.1.** Use a hidden constructor, private or protected, preserving the option to switch
+  to an interface.
+- <a id="rule-8-4-2"></a>**Rule 8.4.2.** Provide `public static Builder builder()`.
+- <a id="rule-8-4-3"></a>**Rule 8.4.3.** Declare every field obtained from the builder as `final`.
+- <a id="rule-8-4-4"></a>**Rule 8.4.4.** A builder may provide `builder(...)` overloads for mandatory or very common
+  parameters.
+    - <a id="rule-8-4-4-1"></a>**Rule 8.4.4.1.** Provide at most two `builder(...)` methods per class.
+- <a id="rule-8-4-5"></a>**Rule 8.4.5.** A no-argument `create()` factory may delegate to `builder().build()`.
+- <a id="rule-8-4-6"></a>**Rule 8.4.6.** A `create(io.helidon.config.Config)` factory may delegate to
+  `builder().config(config).build()`.
```
danielkec: These two `may delegate` rules weaken the previous contract: the factories were optional, but when present their implementation was specified as builder-backed. Under this wording, the following is now compliant even though it bypasses builder defaults, normalization, and validation, so factory construction can diverge from builder/config construction:

```java
public static Foo create() {
    return new Foo(DEFAULT);
}
```

spericas (dismissed):

DEV-GUIDELINES.md
tomas-langer: it is not in conflict, it just does not specify it in this detail. I will try to make this rule exhaustive

DEV-GUIDELINES.md
```diff
@@ -1,206 +1,358 @@
...
+15. Static nested classes.
+16. Non-static nested classes.
+
+Within one of these groups, no additional ordering by name or visibility is required unless the group explicitly defines
+a visibility order.
+
+<a id="chapter-3"></a>
+## 3. Imports and Javadoc
+
+<a id="rule-3-1"></a>**Rule 3.1 — Use imported simple names outside Javadoc.** Outside Javadoc, use an imported simple
```
danielkec, line 98: Rule 3.1 and Rule 9.4 give mutually exclusive instructions for `module-info.java`. A minimal case is:

```java
import p.Service;
module example {
    uses Service;
}
```

The simple name satisfies this rule but violates Rule 9.4; changing it to `uses p.Service;` reverses the violation, so the same declaration cannot comply with both rules.

DEV-GUIDELINES.md
```diff
@@ -1,206 +1,358 @@
...
+- <a id="rule-8-4-5"></a>**Rule 8.4.5.** A no-argument `create()` factory, when present, must delegate to
+  `builder().build()`.
+- <a id="rule-8-4-6"></a>**Rule 8.4.6.** A `create(io.helidon.config.Config)` factory, when present, must delegate to
+  `builder().config(config).build()`.
+- <a id="rule-8-4-7"></a>**Rule 8.4.7.** Other factories for predefined instances must use the builder internally.
+- <a id="rule-8-4-8"></a>**Rule 8.4.8.** Name a nested builder class `Builder`.
+    - <a id="rule-8-4-8-1"></a>**Rule 8.4.8.1.** A top-level builder name must identify the built type, such as
+      `FooBarBuilder`.
+
+Example: [io.helidon.faulttolerance.RetryConfigBlueprint](fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/RetryConfigBlueprint.java).
```
danielkec: The cited `RetryConfigBlueprint` contains constants and a nested `BuilderDecorator` implementation helper, which are exactly the Blueprint members Rule 8.2 prohibits. A minimal excerpt is:

```java
@Prototype.Blueprint(decorator = RetryConfigBlueprint.BuilderDecorator.class)
interface RetryConfigBlueprint {
    int DEFAULT_CALLS = 3;
    class BuilderDecorator { }
}
```

Following the document's example therefore violates the document's own rule.

DEV-GUIDELINES.md
```diff
@@ -1,206 +1,358 @@
...
+
+<a id="rule-1-6"></a>**Rule 1.6 — Prefer enhanced switches.** Prefer switch expressions and arrow labels for new
+switch code.
+
+- <a id="rule-1-6-1"></a>**Rule 1.6.1.** Make switches exhaustive for enums and sealed hierarchies.
+
+<a id="chapter-2"></a>
+## 2. Class member order
+
+<a id="rule-2-1"></a>**Rule 2.1 — Order class members consistently.** Declare class members in the following order,
```
danielkec, line 72: The follow-up replaces Rules 2.1.1, 2.2, 2.3, and 2.3.1 with this expanded Rule 2.1, but none of the four removed IDs remains as a tombstone. A minimal case is:

```text
389b637: #rule-2-2 -> Order public methods
62c9d92: #rule-2-2 -> no target
```

An existing Rule 2.1 citation also changes from static/instance field ordering to the complete 16-group member order. This is the silent reference breakage the PR body and Rules 0.2-0.3 say tombstones prevent.

danielkec (approved):
Reviewed exact head 75209e7453b74dd536fd64b83cc3fca929cefa95.

Issue #12321, opened 2026-08-24
Remove static service lookups from registry-managed services


tomas-langer:
Registry-managed service implementations must not use the static `io.helidon.service.registry.Services` facade for dependency lookup. They should inject required dependencies, or inject `ServiceRegistry` when programmatic selection is required.

Audit the core repository for registry-created paths which can reach `Services` calls. Replace confirmed usages while preserving static lookup only on direct-construction paths which cannot be reached through a registry-managed instance.

Some subsystems, including observability and configured clients, might require registry-based lifecycle and configuration design rather than passing `ServiceRegistry` through unrelated runtime contexts.

The API Javadoc restriction will be handled in a separate pull request.

PR #12322, opened 2026-08-24
Document service static lookup restriction

tomas-langer:
Documents that registry-managed service implementations must use injected dependencies, or an injected `ServiceRegistry` for programmatic lookup, rather than the static `Services` facade throughout their lifecycle.

Related to #12321.

### DIFF
--- a/service/registry/src/main/java/io/helidon/service/registry/Service.java
+++ b/service/registry/src/main/java/io/helidon/service/registry/Service.java
@@ -36,9 +36,9 @@ import io.helidon.common.types.TypeName;
 /**
  * A set of annotations (and APIs) required to declare a service.
  * <p>
- * A service managed by the service registry must never call methods on
+ * Any service instance created by the service registry must never call methods on
  * {@link io.helidon.service.registry.Services}. Required dependencies must be injected into the service. When
- * programmatic lookup is required, inject
+ * programmatic lookup is required during normal operation, inject
  * {@link io.helidon.service.registry.ServiceRegistry} instead. This restriction applies throughout the service lifecycle,
  * including construction, post-construct, business methods, and pre-destroy.
  */
@@ -164,9 +164,10 @@ public final class Service {
      * This also implies that instances that are NOT created within a scope cannot have their pre-destroy methods
      * invoked, as we do not control their lifecycle.
      * <p>
-     * A pre-destroy method must not call methods on {@link io.helidon.service.registry.Services}. Dependencies used during
-     * shutdown must be injected into the service. When programmatic lookup is required, inject
-     * {@link io.helidon.service.registry.ServiceRegistry} instead.
+     * A pre-destroy method is intended to release resources held by the service instance, not to interact with other services.
+     * It must not invoke methods that may instantiate another service. For example, it must not obtain a value from an
+     * injected {@link java.util.function.Supplier} or call an instance lookup method on an injected
+     * {@link io.helidon.service.registry.ServiceRegistry}.
      * <p>
      * The method must not have any parameters and must be accessible (not {@code private}).
      */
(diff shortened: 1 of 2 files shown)


service/registry/src/main/java/io/helidon/service/registry/Service.java
```diff
@@ -35,6 +35,12 @@
 
 /**
  * A set of annotations (and APIs) required to declare a service.
+ * <p>
+ * A service managed by the service registry must never call methods on
```
danielkec: `@Service.PerLookup` is explicitly documented below as producing instances that are "not managed," so this wording can be read as excluding that supported registry-created scope. For example:

```java
@Service.PerLookup
class Consumer {
    @Service.Inject
    Consumer() {
        Services.get(Dependency.class);
    }
}

localRegistry.get(Consumer.class);
```

The registry still constructs, injects, and post-constructs `Consumer`, while `Services.get(...)` queries `GlobalServiceRegistry`; with a non-global registry this can resolve a different dependency from the one that created the service.

service/registry/src/main/java/io/helidon/service/registry/Service.java
```diff
@@ -158,6 +164,10 @@ private Service() {
      * This also implies that instances that are NOT created within a scope cannot have their pre-destroy methods
      * invoked, as we do not control their lifecycle.
      * <p>
+     * A pre-destroy method must not call methods on {@link io.helidon.service.registry.Services}. Dependencies used during
+     * shutdown must be injected into the service. When programmatic lookup is required, inject
+     * {@link io.helidon.service.registry.ServiceRegistry} instead.
```
danielkec: An injected `ServiceRegistry#get(...)` can still activate services during shutdown. `ScopedRegistryImpl.deactivate()` snapshots active services before invoking pre-destroy methods, so a first lookup here creates a service after that snapshot and the new service never receives its own pre-destroy callback:

```java
@Service.PreDestroy
void stop() {
    registry.get(LazyService.class); // first lookup
}
```

This can leave resources owned by `LazyService` active after the registry has shut down.

danielkec (commented):
Two documentation-contract issues found at the current head.

service/registry/src/main/java/io/helidon/service/registry/Service.java
tomas-langer: I will clarify the javadoc, that this is for all services whose instances are created by the registry.

service/registry/src/main/java/io/helidon/service/registry/Service.java
tomas-langer: Also I will clarify the javadoc, that pre destroy method cannot invoke "things" that could instantiate a service (injected suppliers, or any getters from the `ServiceRegistry` instance)

service/registry/src/main/java/io/helidon/service/registry/Service.java
```diff
@@ -158,6 +164,11 @@ private Service() {
      * This also implies that instances that are NOT created within a scope cannot have their pre-destroy methods
      * invoked, as we do not control their lifecycle.
      * <p>
+     * A pre-destroy method is intended to release resources held by the service instance, not to interact with other services.
```
danielkec, line 167: `WebServerStarterService.preDestroy()` already calls `stop()` on its injected singleton `LoomServer`:

```java
if (server != null) {
    server.stop();
}
```

Because this dependency was instantiated before pre-destroy, the interaction cannot cause the shutdown activation leak, but this sentence still describes that core shutdown path as contrary to `@PreDestroy`'s intent. The next sentence and the earlier discussion limit the restriction to operations that may instantiate another service.

danielkec (commented):
One current-head documentation-contract issue found.

service/registry/src/main/java/io/helidon/service/registry/Service.java
tomas-langer: no, this is exactly the intention - the resource that was opened is the webserver and its resources. This is not a contradiction.
Also the rules are describing what should be - there are cases where we must fix this. There will be a follow up as mentioned in the description (https://github.com/helidon-io/helidon/issues/12321)

danielkec (approved):

PR #12323, opened 2026-08-24
Permits a JsonGenerator to work with Appendables

ljnelson:
This PR adds the ability for a `JsonGenerator` to be used with `Appendable` implementations. This reduces array allocations and opens the door for [certain structured logging situations](https://logging.apache.org/log4j/2.12.x/manual/garbagefree.html) among probably lots of other use cases.

### DIFF
--- a/json/json/src/main/java/io/helidon/json/JsonGenerator.java
+++ b/json/json/src/main/java/io/helidon/json/JsonGenerator.java
@@ -76,7 +76,28 @@ public interface JsonGenerator extends AutoCloseable {
      * @return new Generator instance
      */
     static JsonGenerator create(Writer writer, boolean prettyPrint) {
-        return new JsonGeneratorWriter(writer, prettyPrint);
+        return create((Appendable) writer, prettyPrint);
+    }
+
+    /**
+     * Create a {@link JsonGenerator} instance to write to the provided {@link Appendable}.
+     *
+     * @param appendable {@link Appendable} to write to
+     * @return new Generator instance
+     */
+    static JsonGenerator create(Appendable appendable) {
+        return create(appendable, false);
+    }
+
+    /**
+     * Create a {@link JsonGenerator} instance to write to the provided {@link Appendable}.
+     *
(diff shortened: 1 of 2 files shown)


json/json/src/main/java/io/helidon/json/JsonGenerator.java
```diff
@@ -76,7 +76,28 @@ static JsonGenerator create(Writer writer) {
...
+        return create((Appendable) writer, prettyPrint);
+    }
+
+    /**
+     * Create a {@link JsonGenerator} instance to write to the provided {@link Appendable}.
+     *
+     * @param appendable {@link Appendable} to write to
+     * @return new Generator instance
+     */
+    static JsonGenerator create(Appendable appendable) {
```
Verdent, line 88: `PrintStream` extends `OutputStream` and implements `Appendable`, so adding this overload makes existing source such as `JsonGenerator.create(System.out)` ambiguous between `create(OutputStream)` and `create(Appendable)`. The two-argument overload has the same problem. Compiling either call against this PR produces `reference to create is ambiguous`; both calls resolved to the `OutputStream` overload before this change.

ljnelson: Right, and without it I think there would no longer be binary compatibility. It's a tradeoff. If we don't need backwards binary compatibility, because this is a 27 PR and not a 4.x PR, then I can more drastically change the signatures.

tomas-langer: Or use a different method name in 4.x and then remove Writer and keep only APpendable in 27

tomas-langer: although it would require a deprecated method in 27 with the different name and it could only be removed in 28

tomas-langer: Or use a different method name in 4.x and then remove Writer and keep only Appendable in 27.
although it would require a deprecated method in 27 with the different name and it could only be removed in 28

PR #12326, opened 2026-08-25
Close HTTP/1.1 connections after Connection: close (4.x)

### LINKED ISSUE #12324: WebServer does not close HTTP/1.1 connections after Connection: close
## Problem

Helidon WebServer recognizes an HTTP/1.1 request containing `Connection: close` and sends a correct response that also contains `Connection: close`, but it does not close the TCP connection after sending the complete response. A client waiting for the promised EOF blocks until its own timeout.

This reproduces with Helidon 4.4.1 and the latest release, 4.5.3, using both the standard and tuned HttpArena applications.

## Reproducer

Start a WebServer route on port 8080 that returns `55` from `GET /test`, then run:

```python
import socket
import time

request = (
    "GET /test HTTP/1.1\r\n"
    "Host: localhost\r\n"
    "Connection: close\r\n"
    "\r\n"
).encode("ascii")

with socket.create_connection(("localhost", 8080), timeout=3) as sock:
    sock.settimeout(3)

    # Fragmentation is not required, but verifies that partial reads work.
    sock.sendall(request[:1])
    time.sleep(0.2)
    sock.sendall(request[1:])

    response = b""
    while b"\r\n\r\n55" not in response:
        response += sock.recv(4096)

    print(response.decode("ascii"))
    print(sock.recv(1))
```

The complete response arrives immediately and contains:

```text
HTTP/1.1 200 OK
Connection: close
Content-Length: 2

55
```

Expected: the final `recv(1)` returns `b''` promptly.

Actual: the final `recv(1)` times out. Sending the entire request in one `sendall` has the same result.

## HttpArena impact

[HttpArena PR #1289](https://github.com/MDA2AV/HttpArena/pull/1289) temporarily disabled both Helidon entries after its exhaustive TCP-fragmentation validator reported every request shape as failed. The validator splits nine valid requests at every byte offset and reads until EOF because each request specifies `Connection: close`.

The full check reproduced locally on 4.4.1 and 4.5.3:

- 0 of 9 request shapes passed
- all 1,047 split offsets timed out
- both the standard and tuned 4.5.3 applications behaved identically

The reported fragmentation diagnosis is misleading: Helidon parses the fragmented request and returns the correct status and body immediately. The missing connection close is what makes the validator time out.

## Source analysis

`Http1Connection` creates `Http1ServerResponse` with keep-alive disabled when the request contains `Connection: close`, and `Http1ServerResponse` correctly writes the response header. After normal routing completes, however, the HTTP/1 connection loop continues waiting for another request instead of exiting and closing the socket. `Http1ServerResponse.keepConnectionOpen()` is consulted by an entity-consumption error path, but not by the normal successful request path.

The relevant incoming HTTP/1 read/parser sources are effectively unchanged between tags 4.4.1 and 4.5.3, and no later release fixes this behavior.

tomas-langer:
Resolves #12324

Ensures HTTP/1.1 responses selecting `Connection: close` are fully flushed before the server closes the socket, without holding the request-limit permit. Adds raw-socket coverage for request- and response-driven closure, entities, streaming responses, and rejected routed upgrades. Also avoids tokenizing absent headers when checking connection options.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
@@ -111,6 +111,7 @@ public class Http1Connection implements ServerConnection, InterruptableTask<Void
     private volatile Thread myThread;
     private volatile boolean canRun = true;
     private volatile boolean closeInterrupt;
+    private volatile boolean closingConnection;
     private volatile boolean currentlyReadingPrologue;
     private volatile ZonedDateTime lastRequestTimestamp;
     private volatile ServerConnection upgradeConnection;
@@ -145,7 +146,7 @@ public class Http1Connection implements ServerConnection, InterruptableTask<Void
     @Override
     public boolean canInterrupt() {
         if (upgradeConnection == null) {
-            return currentlyReadingPrologue;
+            return currentlyReadingPrologue || closingConnection;
         }
         return true;
     }
@@ -216,6 +217,7 @@ public class Http1Connection implements ServerConnection, InterruptableTask<Void
                                     LimitAlgorithm.Token permit = accepted.token();
                                     ServerConnection routedUpgradeConnection = null;
                                     boolean routeNormally = false;
+                                    boolean keepConnectionOpen = true;
                                     boolean permitCompleted = false;
                                     try {
                                         this.lastRequestTimestamp = DateTime.timestamp();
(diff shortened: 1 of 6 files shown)


spericas (dismissed):
LGTM

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Connection.java
```diff
@@ -145,7 +146,7 @@ public class Http1Connection implements ServerConnection, InterruptableTask<Void
     @Override
     public boolean canInterrupt() {
         if (upgradeConnection == null) {
-            return currentlyReadingPrologue;
+            return currentlyReadingPrologue || closingConnection;
```
tvallin: Including `closingConnection` here makes the response flush interruptible during graceful `close(false)`. `ServerListener.suspend` invokes `close(false)`, which interrupts the handler whenever `canInterrupt()` returns true. If this races with `writer.flush()`, the async writer’s interruptible wait throws and queued response data can be discarded during connection cleanup. Please keep graceful flush non-interruptible—forced shutdown already uses `close(true)`—and add a stop-during-flush regression test.

romain-grecourt (approved):

PR #12328, opened 2026-08-25
Reduce indexed HTTP header allocation

tomas-langer:
Reduces per-header collection allocation by moving five obsolete, deprecated, or unused names out of the indexed header enum while preserving the public constants.

Also avoids creating the custom-header map for read-only misses and adds regression and allocation benchmark coverage.


romain-grecourt (approved):
No findings. Residual risk is limited to runtime allocation and benchmark behavior.

PR #12329, opened 2026-08-25
Reject relative HTTP/1.1 origin-form request-targets with 400 (#12319)

### LINKED ISSUE #12319: Relative HTTP/1.1 request-target causes 500 instead of 400 with SecurityFeature
## Environment Details
* Helidon Version: 4.5.3 (also reproducible with 4.3.4 and current `helidon-4.x` / 4.5.4-SNAPSHOT)
* Helidon SE
* JDK version: 26.0.1
* OS: macOS 26.6.2

----------

## Problem Description

Helidon WebServer accepts a relative HTTP/1.1 request-target without a leading `/`, even when path validation is enabled. When `SecurityFeature` is installed, `SecurityContextFilter` converts the reconstructed requested URI using `req.requestedUri().toUri()`. The relative path cannot be combined with an absolute URI authority, so the request fails with HTTP 500:

```text
java.lang.IllegalArgumentException: UriInfo cannot be used to create a URI:
UriInfo{scheme=http,host=localhost,port=8080,path=boards/,query=,fragment=}
    at io.helidon.common.uri.UriInfoBlueprint.toUri(UriInfoBlueprint.java:106)
    at io.helidon.webserver.security.SecurityContextFilter.filter(SecurityContextFilter.java:64)
Caused by: java.net.URISyntaxException: Relative path in absolute URI:
http://localhost:8080boards/
```

A malformed request-target should be rejected at the HTTP protocol boundary with HTTP 400, rather than being accepted and failing later inside the security filter with HTTP 500.

This is the same behavior previously reported in #1218. PR #1233 changed the old `helidon-1.x` WebServer to return HTTP 400 for invalid URLs, but the Helidon 4 WebServer implementation does not preserve that behavior.

The issue remains reproducible in the 4.5.3 release. The relevant code on the current `helidon-4.x` branch is also unchanged.

## Steps to reproduce

1. Start a Helidon SE WebServer with `SecurityFeature` installed and any route. For example:

```java
Security security = Security.builder().build();

        .port(8080)
        .addFeature(SecurityFeature.builder()
                            .security(security)
                            .build())
        .routing(routing -> routing.get("/*", (req, res) -> res.send("ok")))
        .build()
        .start();
```

2. Send an HTTP/1.1 request with a relative request-target:

```shell
curl --http1.1 http://localhost:8080 --request-target boards/ -v
```

3. Observe HTTP 500 and the `Relative path in absolute URI` exception.

Control request:

```shell
curl --http1.1 http://localhost:8080 --request-target /boards/ -v
```

The control request is processed normally.

At the component level, the behavior can also be seen by passing `boards/` to `HttpPrologue.create(..., validatePath=true)`: it is accepted and remains relative through `uriPath().absolute().path()`. The default requested-URI discovery then copies the relative path into `UriInfo`, and `toUri()` throws the exception above.

## Expected behavior

Reject a relative HTTP/1.1 origin-form request-target such as `boards/` during request-line validation and return HTTP 400. A valid target such as `/boards/` should continue to be accepted.

wadhwaroh-lang:
Fixes #12319

### Description

Helidon WebServer accepted a relative HTTP/1.1 origin-form request-target such as `boards/`. With `SecurityFeature` installed, `SecurityContextFilter` later called `req.requestedUri().toUri()` and failed with HTTP 500 (`Relative path in absolute URI`).

A malformed request-target should be rejected at the HTTP protocol boundary with HTTP 400. This restores the Helidon 1.x behavior from #1218 / #1233.

When path validation is enabled, `HttpPrologue.create` now rejects relative origin-form targets. CONNECT authority-form, `OPTIONS *`, and absolute-form remain valid. `Http1Prologue` already maps the resulting `IllegalArgumentException` to HTTP 400.

### Documentation

None. This is a protocol-boundary bug fix with no API or configuration change.


tomas-langer:
Hi, thanks for the contribution. In general protocol changes are done only by maintainers. I will take this PR over. 
I will keep your contribution visible in git.

http/http/src/main/java/io/helidon/http/HttpPrologue.java
```diff
@@ -297,6 +302,41 @@ public String toString() {
...
     }
 
+    private static void validateRequestTarget(Method method, String rawPath) {
+        if (rawPath.isEmpty()) {
+            throw new IllegalArgumentException("Relative path in HTTP request-target");
+        }
+        if (rawPath.charAt(0) == '/') {
+            return;
+        }
+        if ("*".equals(rawPath) || Method.CONNECT.equals(method) || isAbsoluteForm(rawPath)) {
```
tvallin: `"*"` and `CONNECT` are exempted without checking the applicable request-target form, and `isAbsoluteForm` is also reused for HTTP/2. Consequently malformed targets such as `GET *`, `OPTIONS *?q=1`, `CONNECT boards/`, and HTTP/2 `:path: http://example/a` bypass the new 400/RST handling. Please validate the complete request-target using protocol- and method-specific rules and add negative HTTP/1 and HTTP/2 regression tests.

tomas-langer:
I took over this PR and expanded the original fix so request-target validation follows the actual protocol boundaries.

- Rather than introducing another IP-literal parser, this update brings over the `UriValidator` improvements already present on `main` from #12183. That includes uppercase IPvFuture support, exact IPv6 word accounting, correct two-word handling for embedded IPv4, rejection at the ninth IPv6 word, and bounded scanning of malformed input. I also added the missing rejection for ambiguous multi-digit embedded IPv4 octets with leading zeroes.
- The shared `HttpPrologue` now handles only protocol-neutral parsing and general URI path validation. HTTP/1 owns its request-target forms: CONNECT authority-form, `OPTIONS *`, origin-form, and absolute-form. HTTP/2 owns its `:path` rules and resets only the malformed stream with `PROTOCOL_ERROR`, keeping the connection usable.
- Literal fragments are now rejected in HTTP/1 origin-form and absolute-form targets. Percent-encoded `%23` remains ordinary path or query data.
- WebClient now retains a fragment in `ClientUri` and exposes it to client services, but omits it when serializing the HTTP/1 request line or HTTP/2 `:path`, because fragments are client-side URI components and must not be sent to the server. HTTP/1 send diagnostics were aligned with the actual fragment-free wire target as well.
- Regression coverage now includes the malformed method/form combinations from the review, IPv6/IPvFuture and embedded-IPv4 edge cases, fragment handling, WebClient service propagation, both WebClient protocols, and HTTP/1 send logging. Focused matched JMH coverage was also added; it found no material common-path throughput, latency, or allocation regression.

This keeps the original goal—rejecting invalid relative request targets—while making the rules complete for each HTTP protocol and reusing the validator behavior already established on `main`.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Prologue.java
```diff
@@ -158,6 +159,88 @@ private static boolean isPostMethod(byte[] bytes, int index) {
...
     }
 
+    private static void validateRequestTarget(Method method, String requestTarget) {
+        if (Method.CONNECT.equals(method)) {
+            if (isAuthorityForm(requestTarget)) {
+                return;
+            }
+        } else if ("*".equals(requestTarget)) {
+            if (Method.OPTIONS.equals(method)) {
+                return;
```
tvallin: [P2] Preserve special request-target forms after validation

This return makes `OPTIONS *` valid only at the parser boundary. Both accepted special forms are still stored as an ordinary `UriPath`. With `SecurityFeature`, `req.requestedUri().toUri()` then returns HTTP 500: `OPTIONS *` fails with `Relative path in absolute URI`, while `CONNECT example.com:443` produces a null decoded path and an NPE. I reproduced both against this head. HTTP/2 `OPTIONS *` has the same downstream representation. Please ensure these accepted forms survive requested-URI construction and add full-server security integration tests rather than checking only the prologue/header layer.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Prologue.java
```diff
@@ -158,6 +159,88 @@ private static boolean isPostMethod(byte[] bytes, int index) {
...
+            }
+        } else {
+            if (requestTarget.indexOf('#') >= 0) {
```
tvallin: [P2] Preserve the `validate-prologue=false` fragment opt-out

This check rejects every literal fragment whenever `validatePath` is enabled, before `Http1Connection` consults `validatePrologue()`. Existing applications using `validate-prologue=false` previously accepted `/#fragment`, as its public documentation promises, while retaining path validation. The only workaround now is `validate-path=false`, which disables all request-target validation. This is a configuration compatibility change, contrary to the PR description. Please keep fragment rejection controlled by `validatePrologue`, or explicitly handle this as a user-facing compatibility change with updated contract and migration guidance.

danielkec (dismissed):
Reviewed, thank you.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
```diff
@@ -893,6 +913,22 @@ private void doHeaders(Limit limit) {
...
 
+    private boolean validateRequestTarget(Method method, String requestTarget) {
+        boolean specialRequestTarget = Method.OPTIONS.equals(method) && "*".equals(requestTarget);
+        if (!http2Config.validatePath()) {
+            return specialRequestTarget;
+        }
+        if (specialRequestTarget) {
+            return true;
+        }
+        if (!requestTarget.isEmpty()
```
tvallin, line 928: [P2] Validate the query portion of `:path`

This accepts every leading-slash target without validating its query. `HttpPrologue.create()` strips the query before `UriPath.validate()`, and HTTP/2 has no later equivalent of HTTP/1’s `validatePrologue()`. Consequently targets such as `/ok?q=%GG` reach routing instead of receiving `RST_STREAM(PROTOCOL_ERROR)`, while `requestedUri().toUri()` re-encodes the malformed value and presents security code with a different target. Please validate the raw query when path validation is enabled and verify that the connection remains usable afterward.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1Prologue.java
```diff
@@ -231,12 +327,43 @@ Read HTTP Version (we only support HTTP/1.1
...
+                            .request(DirectTransportRequest.create(protocol, method.text(), path))
+                            .message("CONNECT IP address mechanism is not supported")
+                            .safeMessage(true)
+                            .build();
+                }
+                return HttpPrologue.create(protocol,
+                                           "HTTP",
+                                           "1.1",
+                                           method,
+                                           UriPath.createRelative(UriPath.root(), requestTarget),
```
tvallin: [P2] Preserve the decoded value of CONNECT authorities

`UriPath.createRelative` expects a decoded value and stores its argument as both the raw and decoded path. Passing the wire target means `CONNECT example%2Ecom:443` exposes `uriPath.path()` as `example%2Ecom:443`, not `example.com:443`. Custom path matchers and filters therefore treat encoded and unencoded equivalent authorities differently. Please preserve the raw authority while supplying the correctly decoded value—keeping `+` literal for reg-name semantics—and add a `path()`/matcher regression.

danielkec (approved):
Reviewed, thank you.

PR #12331, opened 2026-08-25
Allow SE filter to get matching http route from elsewhere and more accurately record request processing time

### LINKED ISSUE #12222: 27 Forward port - Built-in HTTP server metrics should emit templated `http.route` for Helidon MP / JAX-RS resources  #12098
Helidon Version: 27

## Enhancement Description

Forward-port of #12098

tjquinno:
### Description
Resolves #12222 

Forward-ports the Helidon core/SE portion of #12098 to Helidon 27, allowing external frameworks—particularly MicroProfile, for example—to provide the matched HTTP route which is used in OpenTelemetry HTTP server metrics as well as potentially other places.

### Release Note

____

Helidon now records `http.server.request.duration` after the response is sent, in seconds, using the final response status. The `http.route` attribute is included only when Helidon provides a non-blank matching route. The deprecated `use-updated-http-metrics` setting remains accepted for configuration compatibility with 4.x, but Helidon 27 ignores its value and always uses the updated behavior.

____

### Summary

- Adds lazy, cached matching-route suppliers to HTTP/1 and HTTP/2 requests.
- Allows an external framework to provide its route template after Helidon routing has completed (for example, Helidon MP).
- Records OpenTelemetry HTTP request metrics after the response is sent instead of in the filter itself.
- Reports duration in seconds and preserves the final HTTP response status.
- Emits `http.route` only when a non-blank matching route is available.
- Ensures metrics are recorded once regardless of filter-chain and response-completion ordering.
- Logs metric-recording failures without interfering with request processing.
- Retains the deprecated `use-updated-http-metrics` configuration property so migrated 4.x configuration remains valid, but ignores its value in Helidon 27.
- Adds HTTP/1, HTTP/2, configuration, route-matching, timing, and failure-handling tests.
- Updates automatic HTTP metrics documentation.

### Compatibility

Existing configuration containing `use-updated-http-metrics` remains valid. Both `true` and `false` select the Helidon 27 behavior: response-sent timing, durations in seconds, final response status, and `http.route` only for a non-blank matching route.

MicroProfile-specific integration is handled separately in the Helidon MicroProfile repository.

### Documentation

Documentation updates are included in the PR.


webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/AutoHttpMetricsConfigBlueprint.java
```diff
@@ -71,4 +72,17 @@ interface AutoHttpMetricsConfigBlueprint {
...
+    /**
+     * Whether to use the updated automatic HTTP metrics behavior.
+     *
+     * @return whether to use updated automatic HTTP metrics behavior
+     * @deprecated Set this only to {@code false} as a temporary compatibility workaround for legacy metric output.
+     */
+    @Deprecated
+    @Option.Configured
+    @Option.DefaultBoolean(true)
+    default boolean useUpdatedHttpMetrics() {
```
tomas-langer, line 85: The deprecated `use-updated-http-metrics` option must remain in Helidon 27 for configuration compatibility, but it is a no-op on `main`. This implementation still branches on its value and restores the 4.x legacy metrics behavior when it is `false`; the Blueprint Javadoc likewise describes it as an active compatibility switch. Existing configuration can therefore change Helidon 27 telemetry semantics instead of being safely ignored, and the generated configuration reference documents the wrong contract.

tjquinno: Fixed. Setting is now permitted but ignored.

tomas-langer (approved):

PR #12332, opened 2026-08-25
27/x: Remove dependencyManagement for jaxb artifacts

barchetta:
### Description

Remove management of jaxb artifacts.


ljnelson (approved):
LGTM; as far as I remember this was only managed to deal with JPA and Eclipselink and Hibernate discrepancies and transitive stuff.

PR #12335, opened 2026-08-25
Remove static service lookups from registry-managed services

### LINKED ISSUE #12321: Remove static service lookups from registry-managed services

tomas-langer:
Resolves #12321

Remove static service lookups from registry-managed paths by supplying dependencies from the owning registry while preserving direct-construction fallbacks. This also provides registry-based observability discovery with ServiceLoader compatibility and clarifies the related configuration and lifecycle contracts.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryPrototypeInfo.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryPrototypeInfo.java
@@ -578,9 +578,103 @@ final class FactoryPrototypeInfo {
     }
 
     private static boolean registrySupport(TypeInfo blueprint) {
-        return blueprint.findAnnotation(Types.PROTOTYPE_SERVICE_REGISTRY)
+        Optional<Annotation> annotation = blueprint.findAnnotation(Types.PROTOTYPE_SERVICE_REGISTRY);
+        if (annotation.isPresent()) {
+            return annotation.get().booleanValue().orElse(true);
+        }
+        return superBlueprintDefinition(blueprint)
+                .map(FactoryPrototypeInfo::registrySupport)
+                .orElse(false);
+    }
+
+    static Optional<String> inheritedServiceRegistryAccessor(TypeInfo blueprint) {
+        return superBlueprintDefinition(blueprint)
+                .flatMap(FactoryPrototypeInfo::serviceRegistryOwner)
+                .map(FactoryPrototypeInfo::serviceRegistryAccessor);
+    }
+
+    static Optional<String> inheritedConfigAccessor(TypeInfo blueprint) {
+        return superBlueprintDefinition(blueprint)
+                .flatMap(FactoryPrototypeInfo::configOwner)
+                .map(FactoryPrototypeInfo::configAccessor);
+    }
+
+    private static Optional<TypeInfo> serviceRegistryOwner(TypeInfo blueprint) {
+        Optional<TypeInfo> inheritedOwner = superBlueprintDefinition(blueprint)
+                .flatMap(FactoryPrototypeInfo::serviceRegistryOwner);
+        if (inheritedOwner.isPresent()) {
+            return inheritedOwner;
+        }
+        return declaresServiceRegistry(blueprint) ? Optional.of(blueprint) : Optional.empty();
+    }
+
+    private static Optional<TypeInfo> configOwner(TypeInfo blueprint) {
+        Optional<TypeInfo> inheritedOwner = superBlueprintDefinition(blueprint)
+                .flatMap(FactoryPrototypeInfo::configOwner);
+        if (inheritedOwner.isPresent()) {
+            return inheritedOwner;
+        }
+        return blueprint.hasAnnotation(PROTOTYPE_CONFIGURED) ? Optional.of(blueprint) : Optional.empty();
+    }
+
+    private static String serviceRegistryAccessor(TypeInfo owner) {
+        return recordStyleAccessors(blueprintAnnotation(owner)) ? "serviceRegistry" : "getServiceRegistry";
+    }
+
+    private static String configAccessor(TypeInfo owner) {
+        return recordStyleAccessors(blueprintAnnotation(owner)) ? "config" : "getConfig";
+    }
+
+    private static Optional<TypeInfo> superBlueprintDefinition(TypeInfo blueprint) {
+        return superBlueprintDefinition(blueprint, new HashSet<>());
+    }
+
+    private static Optional<TypeInfo> superBlueprintDefinition(TypeInfo inProgress, Set<TypeName> processed) {
+        for (TypeInfo superInterface : inProgress.interfaceTypeInfo()) {
+            if (!processed.add(superInterface.typeName())) {
+                continue;
+            }
+            if (superInterface.hasAnnotation(PROTOTYPE_BLUEPRINT)) {
+                return Optional.of(superInterface);
+            }
+            Optional<TypeInfo> prototypeBlueprint = superInterface.interfaceTypeInfo()
+                    .stream()
+                    .filter(it -> it.hasAnnotation(PROTOTYPE_BLUEPRINT))
+                    .findFirst();
+            if (prototypeBlueprint.isPresent()) {
+                return prototypeBlueprint;
+            }
+            Optional<TypeInfo> inheritedBlueprint = superBlueprintDefinition(superInterface, processed);
+            if (inheritedBlueprint.isPresent()) {
+                return inheritedBlueprint;
+            }
+        }
+        return Optional.empty();
+    }
+
+    private static boolean declaresServiceRegistry(TypeInfo blueprint) {
+        boolean registrySupport = blueprint.findAnnotation(Types.PROTOTYPE_SERVICE_REGISTRY)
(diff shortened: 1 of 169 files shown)


webclient/api/src/main/java/io/helidon/webclient/api/WebClientConfigFactory.java
```diff
@@ -0,0 +1,70 @@
...
+        Config defaultConfig = configuredClients.remove(Service.Named.DEFAULT_NAME);
+        result.add(configured(Service.Named.DEFAULT_NAME, defaultConfig == null ? Config.empty() : defaultConfig));
+        configuredClients.forEach((name, clientConfig) -> result.add(configured(name, clientConfig)));
+        return List.copyOf(result);
+    }
+
+    private Service.QualifiedInstance<WebClientConfig> configured(String name, Config config) {
+        WebClientConfig webClientConfig = WebClientConfig.builder()
+                .config(config)
+                .serviceRegistry(serviceRegistry)
```
danielkec, line 66: When this factory builds a registry-managed client, `WebClientSecurityProvider` reaches the new three-argument provider path but inherits the default fallback to `create(config, name)`. That creates `WebClientSecurity` from ambient/global context rather than the registry passed here, so a request without an ambient `SecurityContext` proceeds without the owning registry's outbound security:

```text
managed-outbound-provider-calls=0
managed-wire-x-probe=<absent>
explicit-outbound-provider-calls=1
explicit-wire-x-probe=explicit
```

config/config/src/main/java/io/helidon/config/ProvidedUtil.java
```diff
@@ -774,25 +642,67 @@ private record TypeAndName(String type, String name) {
...
 
         @Override
-        public <T> List<T> all(Class<T> type) {
-            return Services.all(type);
+        public List<T> all() {
+            Map<String, T> providers = new LinkedHashMap<>();
+            HelidonServiceLoader.create(providerType)
+                    .forEach(provider -> providers.putIfAbsent(provider.configKey(), provider));
+            Services.all(providerType)
+                    .forEach(provider -> providers.putIfAbsent(provider.configKey(), provider));
```
danielkec: On a direct builder, `Services.set(WebClientServiceProvider.class, replacement)` makes the replacement the registry's only provider, but the classpath entry is inserted first and these `putIfAbsent` calls discard the registry replacement when both use the `security` key. A fresh-JVM probe produced:

```text
replacement-creates=0
services=[security:io.helidon.webclient.security.WebClientSecurity]
```

The builder therefore uses the classpath provider instead of the explicitly bound replacement.

webserver/grpc/src/main/resources/META-INF/helidon/manifest
```diff
@@ -0,0 +1,2 @@
+#HELIDON MANIFEST#
+META-INF/helidon/io.helidon.webserver.grpc/service.loader
```
danielkec, line 2: After a clean package generates the registry metadata entries, a second successful incremental package copies this two-line source manifest while javac reports `Nothing to compile`. The jar still contains `service-registry.json` and `feature-registry.json`, but this manifest lists only `service.loader`, so registry discovery skips the generated gRPC services; `webserver/observe/observe` has the same overwrite.

```text
mvn -pl webserver/grpc -am -DskipTests clean package
mvn -pl webserver/grpc -DskipTests package
unzip -p webserver/grpc/target/helidon-webserver-grpc-27.0.0-SNAPSHOT.jar META-INF/helidon/manifest
#HELIDON MANIFEST#
META-INF/helidon/io.helidon.webserver.grpc/service.loader
```

builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryPrototypeInfo.java
```diff
@@ -578,9 +590,91 @@ private static Set<TypeName> providerProvides(TypeInfo blueprint) {
...
+            if (!processed.add(superInterface.typeName())) {
+                continue;
+            }
+            if (superInterface.hasAnnotation(PROTOTYPE_BLUEPRINT)) {
+                return Optional.of(superInterface);
+            }
+            Optional<TypeInfo> prototypeBlueprint = superInterface.interfaceTypeInfo()
+                    .stream()
+                    .filter(it -> it.hasAnnotation(PROTOTYPE_BLUEPRINT))
+                    .findFirst();
```
danielkec, line 645: When an unannotated child blueprint extends the generated prototype of a `@Prototype.Blueprint(detach = true)` parent with `@RegistrySupport`, this lookup cannot recover the parent blueprint because the detached prototype does not implement it. The child therefore generates the ServiceLoader path even when its inherited builder receives an explicit registry, contrary to the new parent-prototype inheritance contract.

```text
PARENT_PRESENT=true
CHILD_PRESENT=false
```

webserver/grpc/src/main/resources/META-INF/helidon/manifest
tomas-langer: This is valid and is broader than the two manifests added here. I reproduced the incremental overwrite in gRPC and
Observe, and found the same checked-in-manifest/generated-metadata combination in seven Helidon modules. The source
manifest should not need to enumerate files created by annotation processors.

The shared build-tools fix is tracked in
[helidon-io/helidon-build-tools#1167](https://github.com/helidon-io/helidon-build-tools/issues/1167). The intended design
is a post-compilation Maven reconciliation step that rebuilds `META-INF/helidon/manifest` from the metadata actually
present in the module output on every build, while retaining APT generation for direct javac and non-Maven use. I will
not hardcode generated JSON paths into these module manifests.

webclient/api/src/main/java/io/helidon/webclient/api/WebClientConfigFactory.java
tomas-langer: Fixed by e9497991a3, 81a2a430e8, and 2e6303ef99.

Registry-created WebClient security now resolves `Security` from the owning registry. That instance continues to select the outbound provider, while the authenticated request `SecurityContext` supplies the current user and service subjects plus its correlation ID. Outbound audit is dispatched through the owning `Security` as well.

The regressions verify that the owning outbound provider is invoked, receives both authenticated subjects, and emits audit only through the client-owning security. The final affected-module run passed 1,409 Config tests (6 skipped), 66 Security tests, and 7 WebClient security tests.

config/config/src/main/java/io/helidon/config/ProvidedUtil.java
tomas-langer: Fixed by 418aa8c092 and 2e6303ef99.

Direct configured builders now keep classpath provider ordering and overlay explicit registry bindings by configuration key. When multiple explicit registry providers share a key, the first and highest-priority registry provider remains selected.

The fail-first regressions cover both replacement of the classpath `security` provider and priority preservation for two explicit same-key providers. The final Config suite passed all 1,409 tests with 6 skipped.

builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryPrototypeInfo.java
tomas-langer: Fixed by beeab7aaef and 9c32a206e4.

A detached generated parent is now resolved through its existing `@Generated.trigger` metadata in `RoundContext`, and the recovered blueprint participates in the same registry-support, registry-ownership, and inherited-accessor traversal as attached parents.

The fail-first regression uses a child-only provider contract to verify explicit-registry lookup. It also asserts that the generated `META-INF/helidon/service.loader` contains that unique contract, independently covering automatic metadata creation. The final builder run passed 121 tests with 1 skipped.

webclient/security/src/main/java/io/helidon/webclient/security/WebClientSecurity.java
```diff
@@ -52,13 +52,15 @@ public class WebClientSecurity implements WebClientService {
...
     private final Security security;
+    private final boolean defaultOutboundSecurity;
 
     private WebClientSecurity() {
         this(null);
     }
 
     private WebClientSecurity(Security security) {
         this.security = security;
+        this.defaultOutboundSecurity = security != null && !security.resolveOutboundProvider(null).isEmpty();
```
danielkec, line 63: When the owning registry supplies a programmatically disabled `Security` that still has an outbound provider, this flag treats it as active because it checks only provider presence. A registry-managed WebClient then invokes that provider even though security is disabled, so it can still add outbound credentials. A fail-first probe produced:

```text
security-enabled=false
outbound-provider-calls=1
```

The current inactive-security regression covers an empty provider list, not a disabled `Security` with providers.

danielkec (approved):

PR #12340, opened 2026-08-26
Support Alt-Svc in WebClient

### LINKED ISSUE #7278: Support for `Alt-Svc` header in WebClient
When we receive a response from a server, we may be informed that  a "better" protocol version is available (such as an HTTP/1.1 endpoint can return that HTTP/2 is availalbe even on plaintex endpoint, without ALPN, or in the future, that HTTP/3 is available).

When using the `WebClient` API that is not protocol version specific, we should check the response, and if it contains such a header, we should mark the endpoint as supporting a newer HTTP version, so a next request to that endpoint will use that newer protocol version

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Alt-Svc

tomas-langer:
Resolves #7278
Pre-requisite for https://github.com/helidon-io/helidon/issues/3686

This change adds preview, opt-in client Alt-Svc discovery to WebClient. Its primary intent is to let a generic WebClient learn an HTTP/2 alternative from a network response and route later requests to that endpoint without changing the request's logical origin, authority, or TLS identity. It also establishes protocol-neutral configuration, parsing, and response-notification foundations that can be reused by the HTTP/3 client.

### Configuration and protocol selection

The new optional `ClientAltSvcConfig` belongs to the common HTTP client configuration rather than an individual protocol configuration.

When the configuration is absent, Alt-Svc handling is disabled. Within a present configuration, `enabled` defaults to `true`. The optional `protocols` collection is an exact, case-sensitive ALPN allow-list; an empty collection permits every available protocol provider that supports Alt-Svc.

Protocol preference continues to control provider ordering. Alt-Svc discovery only makes an advertised protocol available for selection.

### HTTP/2 alternative routing

The HTTP/2 provider can learn an `h2` alternative from an HTTP/1.1 or HTTP/2 response and use it for a later request. The alternative changes only the physical connection endpoint:

- The original request URI and authority are preserved.
- TLS peer identity, SNI, and the configured TLS policy remain those of the logical origin.
- Alternative requests include the generated `Alt-Used` header.
- Connection pools remain partitioned by their concrete targets while compatible DNS-host casing shares advertisement state.
- Shared connection caches share Alt-Svc discovery state; private caches remain isolated.

The implementation preserves response ordering across redirects and HTTP/1.1 fallback, including output-stream and Expect/Continue flows.

### Advertisement lifecycle

The client parser handles multiple Alt-Svc field values, opaque protocol identifiers, percent escapes, `ma`, `persist`, `Age`, apparent age from `Date`, and the complete-field `clear` value.

Discovery state is bounded and tied to the connection-cache lifecycle. Replacement, `clear`, current-alternative `421` responses, TLS reload, network changes, and failed alternative establishment invalidate or suppress the appropriate generation. An expired advertisement cannot create a new alternative connection, but an already-opening or established compatible connection can remain reusable.

If an alternative fails before the request can have been processed, WebClient falls back to the origin. Failures after possible transmission are not automatically replayed.

### Security, proxy, and current scope

Initial client support is deliberately limited to:

- Alt-Svc response headers from `https` origins
- Same-host alternatives, optionally on a different port
- TLS HTTP/2 using ALPN `h2`
- Clients whose proxy type is explicitly `none`

Any configured proxy policy—including the default system proxy policy and configured no-proxy exceptions—causes the HTTP/2 provider to ignore the alternative rather than bypass that policy.

WebClient honors the configured TLS policy as-is, including custom TLS managers, custom SSL contexts, disabled endpoint identification, and `trust-all`. A permissive TLS configuration therefore makes Alt-Svc steering equally permissive.

This change does not support cleartext `h2c`, upgrading plain `http` origins, RFC 8164 opportunistic HTTP, cross-host alternatives, or the HTTP/2 ALTSVC frame.

### DIFF
--- a/webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
@@ -422,6 +422,7 @@ public final class AltSvcHeader {
         int emptyElements = initialEmptyElements;
         boolean quoted = false;
         boolean escaped = false;
+        boolean malformed = emptyElements > MAX_EMPTY_LIST_ELEMENTS;
         boolean tooMany = false;
         boolean clear = false;
         for (int index = 0; index <= value.length(); index++) {
@@ -448,9 +449,11 @@ public final class AltSvcHeader {
 
             String element = trimOws(value.substring(elementStart, index));
             if (element.isEmpty()) {
-                emptyElements++;
+                if (emptyElements <= MAX_EMPTY_LIST_ELEMENTS) {
+                    emptyElements++;
+                }
                 if (emptyElements > MAX_EMPTY_LIST_ELEMENTS) {
-                    return new AlternativeListResult(clear, true, tooMany, emptyElements);
+                    malformed = true;
                 }
             } else if ("clear".equals(element)) {
                 clear = true;
@@ -461,7 +464,7 @@ public final class AltSvcHeader {
             }
             elementStart = index + 1;
         }
-        return new AlternativeListResult(clear, false, tooMany, emptyElements);
+        return new AlternativeListResult(clear, malformed, tooMany, emptyElements);
     }
 
     private static Optional<List<String>> splitValues(String value, char delimiter) {
--- a/webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
@@ -149,15 +149,16 @@ public final class AltSvcHeader {
     }
 
     private static GrammarResult parseGrammarSnapshot(List<String> fieldLines) {
+        if (fieldLines.size() == 1 && "clear".equals(trimOws(fieldLines.getFirst()))) {
+            return CLEAR_GRAMMAR;
+        }
+
         List<String> encodedAlternatives = new ArrayList<>();
         boolean malformed = false;
         boolean tooMany = false;
         int emptyElements = 0;
         for (String headerValue : fieldLines) {
             AlternativeListResult result = splitAlternatives(headerValue, encodedAlternatives, emptyElements);
-            if (result.clear()) {
-                return CLEAR_GRAMMAR;
-            }
             malformed |= result.malformed();
             tooMany |= result.tooMany();
             emptyElements = result.emptyElements();
@@ -422,7 +423,6 @@ public final class AltSvcHeader {
         int emptyElements = initialEmptyElements;
         boolean quoted = false;
         boolean escaped = false;
-        boolean clear = false;
         boolean tooMany = false;
         for (int index = 0; index <= value.length(); index++) {
             if (index < value.length()) {
@@ -443,17 +443,15 @@ public final class AltSvcHeader {
                     continue;
                 }
             } else if (quoted || escaped) {
-                return new AlternativeListResult(false, true, tooMany, emptyElements);
+                return new AlternativeListResult(true, tooMany, emptyElements);
             }
 
             String element = trimOws(value.substring(elementStart, index));
             if (element.isEmpty()) {
                 emptyElements++;
                 if (emptyElements > MAX_EMPTY_LIST_ELEMENTS) {
-                    return new AlternativeListResult(false, true, tooMany, emptyElements);
+                    return new AlternativeListResult(true, tooMany, emptyElements);
                 }
-            } else if ("clear".equals(element)) {
-                clear = true;
             } else if (result.size() == MAX_ALTERNATIVES) {
                 tooMany = true;
             } else {
@@ -461,7 +459,7 @@ public final class AltSvcHeader {
             }
             elementStart = index + 1;
         }
-        return new AlternativeListResult(clear, false, tooMany, emptyElements);
+        return new AlternativeListResult(false, tooMany, emptyElements);
     }
 
     private static Optional<List<String>> splitValues(String value, char delimiter) {
@@ -669,7 +667,7 @@ public final class AltSvcHeader {
     private record AlternativeTemplate(String protocolId, String host, int port, long maxAge, boolean persist) {
     }
 
-    private record AlternativeListResult(boolean clear, boolean malformed, boolean tooMany, int emptyElements) {
+    private record AlternativeListResult(boolean malformed, boolean tooMany, int emptyElements) {
     }
 
     private record DateMemo(String value, Instant parsedDate) {
--- a/webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
+++ b/webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
@@ -1,0 +1,686 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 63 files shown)


webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
```diff
@@ -0,0 +1,684 @@
...
+            fieldLineSnapshot = List.copyOf(fieldLines);
+        }
+
+        GrammarResult result = parseGrammarSnapshot(fieldLineSnapshot);
+        grammarMemo = new GrammarMemo(fieldLineSnapshot, result);
+        return result;
+    }
+
+    private static GrammarResult parseGrammarSnapshot(List<String> fieldLines) {
+        if (fieldLines.size() == 1 && "clear".equals(trimOws(fieldLines.getFirst()))) {
```
danielkec: [RFC 7838 section 3](https://www.rfc-editor.org/rfc/rfc7838.html#section-3) says a field value containing the case-sensitive `clear` token invalidates all cached alternatives, including an invalid reply containing both `clear` and alternatives. This sole-field check makes the following return `Optional.empty()`; `Http2ClientImpl.responseReceived` then skips the cache update, leaving the previous route eligible:

```text
Alt-Svc: h2=":9443", clear
```

Later requests can therefore continue to a service the origin attempted to withdraw.

webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
```diff
@@ -0,0 +1,684 @@
...
+        }
+        return new GrammarResult(false, false, List.copyOf(alternatives));
+    }
+
+    private static long responseAge(ClientResponseHeaders headers, Instant receivedAt) {
+        long responseAge = 0;
+        if (headers.contains(HeaderNames.AGE)) {
+            long parsedAge = parseDeltaSeconds(headers.get(HeaderNames.AGE).get());
+            if (parsedAge >= 0) {
+                responseAge = parsedAge;
```
danielkec, line 187: [RFC 7838 section 3.1](https://www.rfc-editor.org/rfc/rfc7838.html#section-3.1) delegates freshness to response age; [RFC 7234 section 4.2.3](https://www.rfc-editor.org/rfc/rfc7234.html#section-4.2.3) calculates corrected age as `Age + (response_time - request_time)`. Here the parsed `Age` becomes the response age without that delay, so this case is treated as fresh for another 10 seconds even though its corrected age is already 65 seconds:

```text
request_time = 0s
response_time = 15s
Age: 50
Alt-Svc: h2=":8443"; ma=60
Date: absent or unusable
```

During that interval the client can create a new connection after the advertised lifetime has ended.

webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
```diff
@@ -0,0 +1,686 @@
...
+                }
+            } else if (quoted || escaped) {
+                return new AlternativeListResult(clear, true, tooMany, emptyElements);
+            }
+
+            String element = trimOws(value.substring(elementStart, index));
+            if (element.isEmpty()) {
+                emptyElements++;
+                if (emptyElements > MAX_EMPTY_LIST_ELEMENTS) {
+                    return new AlternativeListResult(clear, true, tooMany, emptyElements);
```
danielkec: `clear` can still be hidden by the empty-element bound. `splitAlternatives` returns on the 33rd empty element before it reaches a later `clear`, so this case produces `Optional.empty()` and `responseReceived` leaves the previously cached alternative eligible; placing the same `clear` on a second field line does clear it:

```java
AltSvcHeader.create(headers(",".repeat(33) + "clear"), now) // Optional.empty()
```

This is a bounded residual of the [earlier mixed-`clear` discussion](https://github.com/helidon-io/helidon/pull/12340#discussion_r3862705503); RFC 7838 requires a field containing `clear` to invalidate all alternatives.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2AltSvcCache.java
```diff
@@ -0,0 +1,1172 @@
...
+                        : normalizeHost(target.originAuthority().host().value());
+                alternative = selectAlternative(originHost, header, now);
+            }
+            boolean withdrawal = alternative == null;
+            Instant latestObservation = previous == null ? tombstones.get(routeKey) : previous.observedAt;
+            // Equal-time withdrawals win; advertisements must be strictly newer than state and network barriers.
+            if (observedAt.isBefore(networkChangedAt)
+                    || (!withdrawal && observedAt.equals(networkChangedAt))
+                    || (latestObservation != null
+                            && (observedAt.isBefore(latestObservation)
```
danielkec: When a `clear` response observed before a local connection failure is delivered after that failure, `recordFailure` has advanced `previous.observedAt`, so this stale check drops the withdrawal. A minimal sequence is:

```text
t0  advertise h2=":8443"; ma=3600
t5  observe clear; delay callback
t10 recordFailure(selection)
    record(clear, observedAt=t5)
t+5m select(...) -> :8443
```

The old route becomes eligible again after the backoff even though the origin withdrew it.

webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
```diff
@@ -0,0 +1,689 @@
...
+            }
+            alternatives.add(alternative.get());
+        }
+        return new GrammarResult(false, false, List.copyOf(alternatives));
+    }
+
+    private static long responseAge(ClientResponseHeaders headers, Instant receivedAt) {
+        long responseAge = 0;
+        if (headers.contains(HeaderNames.AGE)) {
+            long parsedAge = parseDeltaSeconds(headers.get(HeaderNames.AGE).get());
```
danielkec: Independently of the response-delay case, `Age` can be received as a list such as `50, 10`; [RFC 9111 section 5.1](https://www.rfc-editor.org/rfc/rfc9111.html#section-5.1) says to use the first member. Passing the whole value to `parseDeltaSeconds` rejects it and leaves `responseAge` at zero. With no usable `Date` and `Alt-Svc: h2=":8443"; ma=60`, the alternative expires at `receivedAt + 60s` rather than `receivedAt + 10s`, leaving it eligible for 50 seconds past its lifetime.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2AltSvcCache.java
```diff
@@ -0,0 +1,1171 @@
...
+            if (!header.clear()) {
+                originHost = memoized
+                        ? previous.originHost
+                        : normalizeHost(target.originAuthority().host().value());
+                alternative = selectAlternative(originHost, header, now);
+            }
+            boolean withdrawal = alternative == null;
+            Instant latestObservation = previous == null ? tombstones.get(routeKey) : previous.observedAt;
+            // Withdrawals always win; advertisements must be strictly newer than state and network barriers.
+            if (!withdrawal
```
danielkec: The delayed-`clear`/failure case is fixed, but when a `clear` callback observed before a newer advertisement is delivered afterward, `withdrawal` bypasses the `latestObservation` check and removes the newer route. A minimal sequence is:

```text
t5  observe clear; delay callback
t10 record h2=":9443"; ma=60
    record(clear, observedAt=t5)
    select(...) -> null  // expected :9443
```

This lets an older response overwrite the origin's newer Alt-Svc state and forces origin fallback until another advertisement arrives.

webclient/api/src/main/java/io/helidon/webclient/api/AltSvcHeader.java
tomas-langer: This is not a time sensitive feature. We do not need to take care of every millisecond.
You will announce Alt-Svc because you want clients to use a newer version of http (i.e. http/3). Not to control what to use every five seconds (at least that seems to be the intent of the spec).
In case there is a good use case to have quick changes, we can introduce proper accounting for timing of request

danielkec (approved):

PR #12350, opened 2026-08-26
Allow MP Telemetry span to include response write

### LINKED ISSUE #12344: Helidon MP server span ends  before Streaming output and WriterInterceptor to complete
### Environment

- Helidon version: 4.5.1
- Helidon flavor: MP
- Helidon obtained through:
  oracle.spectra.bundles:helidon-mp:2610.0.9
- Java version: 25
- JAX-RS implementation: Jersey through Helidon MP
- Telemetry: OpenTelemetry

### Problem

For a JAX-RS endpoint returning StreamingOutput, Helidon's automatic server
span ends when the resource method returns.

The actual response is serialized later through WriterInterceptor and
StreamingOutput.write(...). Application spans created during this processing
therefore continue after the automatic server span has ended.

Example observed timing:

- WebServer access-log duration: 21.647 ms
- Automatic GET server span: 1.65 ms
- permission materialization child span: 15.84 ms

Response materialization is part of the HTTP request lifecycle, but the child
span ends after its parent server span.

### Minimal reproduction

```java
@Path("/stream")
public class StreamingResource {
    @GET
    public Response stream() {
        StreamingOutput output = stream -> {
            try (var span = startChildSpan("response.materialization")) {
                // Simulate delayed response generation.
                Thread.sleep(20);
                stream.write("{\"result\":\"ok\"}".getBytes(StandardCharsets.UTF_8));
            }
        };
        return Response.ok(output).build();
    }
}
```

Expected behavior

  The automatic server span should remain active until response entity writing
   has completed or failed, including WriterInterceptor and StreamingOutput
   processing.
  The server span should contain response-materialization child spans and should
   reflect the complete application response lifecycle.

Actual behavior

  The automatic server span ends in the ContainerResponseFilter before
   WriterInterceptor and StreamingOutput serialization execute.

tjquinno:
### Description
Resolves helidon-io/helidon#12344 

## Release Note

____
Optional response-preparation coverage for REST spans using MicroProfile Telemetry

Helidon MP adds the new deprecated configuration setting
`telemetry.span.includes-response-write`. The default value `false` preserves
existing Helidon 4 behavior in which the automatic incoming REST span ends
before response-entity serialization.

Setting it to `true` keeps the span open through Jersey response processing,
including response serialization and encoding into Helidon’s response stream.
If an asynchronous resource method is still executing when Jersey finishes
processing the response, the span ends when that resource method returns.

The span does not include the later WebServer socket commit, network delivery,
or client receipt or acknowledgement.

The setting is planned for removal in a future major release, after which
automatic incoming REST spans will always include response preparation through
Jersey completion.
____

## Summary

Adds the deprecated compatibility setting
`telemetry.span.includes-response-write`, which controls whether an automatic
incoming REST server span includes response preparation.

The setting defaults to `false`, preserving existing Helidon 4 behavior. When
set to `true`, the span remains open through Jersey response processing, after
entity serialization and encoding have populated Helidon’s response stream. If
an asynchronous resource method is still executing at Jersey `FINISHED`, the
span remains open until that method returns.

This boundary does not include the later WebServer socket commit, network
delivery, or client acknowledgement. After the setting is removed in a future
major release, automatic spans will always include response preparation through
Jersey completion.

## Design/Implementation

The opt-in lifecycle uses phase-specific, thread-owned scope activation:

- The initial request-filter scope is closed on its owning dispatcher thread
  when the exact OpenTelemetry context established by the automatic span is
  current.
- If a later request filter has established a nested span or baggage context,
  initial-scope closure is deferred until the telemetry response filter runs
  after the paired response filter has restored the original context.
- Separate scopes make the automatic span current during resource-method
  execution and exception mapping, including managed asynchronous execution.
- The automatic span is not forced current across arbitrary response filters,
  preserving the nesting established by paired request and response filters.
- An outer writer interceptor makes the span current during response-entity
  materialization, so spans created by writer interceptors, message-body
  writers, and `StreamingOutput` become children of the automatic server span.
- Jersey `FINISHED` records completion of response processing and makes the
  span eligible to end. If a resource method is still executing, completion is
  deferred until `RESOURCE_METHOD_FINISHED`.
- The response filter records a provisional HTTP status, and Jersey `FINISHED`
  replaces it with the final response status when available. The lifecycle
  applies that final status immediately before ending the span, so later
  response-filter changes are authoritative and an interim 5xx status does not
  remain sticky.
- Serialization failures, response-processing exceptions, and unsuccessful
  response writes mark the span as erroneous and record the available failure.
- VarHandle-based atomic state transitions ensure the span ends only once,
  regardless of asynchronous event ordering.
- The writer interceptor removes the lifecycle request property after initial
  entity processing, and the request listener releases its lifecycle reference
  after terminal completion. This avoids retaining ended telemetry state for
  long-lived `ChunkedOutput` and SSE responses.
- The request listener and writer interceptor are registered only for server
  runtimes when response-write inclusion is enabled and no OpenTelemetry Java
  agent is present. They add no per-request processing to the default,
  Java-agent, or Jersey-client paths.

The existing synchronous behavior remains unchanged when the setting is omitted
or explicitly set to `false`. The container filter retains its Java-agent guard
as defensive protection, and requests suppressed by a telemetry filter helper
remain unchanged.

The telemetry documentation describes this setting and the existing deprecated
span-naming compatibility settings.

## Testing

Added or updated coverage for:

- Omitted and explicitly disabled configuration.
- Successful responses and entity-less responses.
- HTTP 500 responses and mapped exceptions.
- Final HTTP status after a failed entity write.
- Writer and response-filter failures.
- Writer-interceptor and `StreamingOutput` child-span relationships.
- Exception-mapper child-span relationships.
- Managed asynchronous and suspended resource methods.
- Scope ownership across dispatcher, managed-executor, exception-mapper, and
  entity-writer threads.
- Both possible orderings of resource-method completion and Jersey `FINISHED`.
- Atomic and idempotent span completion.
- Child-span and baggage-only nesting in paired request and response filters.
- Prevention of stale trace and baggage context between sequential requests.
- Removal of the lifecycle request property after successful and failed writes.
- Release of the request listener’s lifecycle reference after synchronous and
  asynchronous completion.
- Absence of the opt-in providers when configuration is omitted or `false`.
- Server-only registration of the opt-in providers, including their absence
  from Jersey clients.

Verification completed successfully:

- `mvn -pl microprofile/telemetry -Pcheckstyle verify`
  - 51 tests passed across the default and specialized Surefire executions.
  - 0 Checkstyle violations.
- `etc/scripts/copyright.sh`
  - `COPYRIGHT OK`.
- `git diff --check`
  - Passed.

### Documentation

Update included in PR.

### DIFF
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -480,19 +480,26 @@ naming convention unconditionally.
 
 When `telemetry.span.includes-response-write` is `false`, Helidon ends the
 span before serializing the response entity, preserving the behavior of earlier
-Helidon 4 releases. When it is `true`, Helidon ends the span after response
+Helidon 4 releases. When it is `true`, Helidon keeps the span open until response
 serialization and encoding have populated Helidon's response stream and Jersey
-has finished processing the response. The span remains open during response
+has finished processing the response. If an asynchronous resource method is
+still executing at that point, the span ends when the resource method returns.
+The span remains open during response
 filtering, but Helidon makes it current only during discrete, thread-owned
-phases: request filtering, resource-method execution, and entity
-materialization. This includes asynchronous JAX-RS responses. Spans started by
-writer interceptors, message-body writers, or streaming output are therefore
-children of the automatic span, without leaving its scope attached to a thread
-between phases. If response processing or writing fails, Helidon records the
-available failure and ends the automatic span when Jersey finishes processing
-the failed request. An application exception which Jersey successfully maps to
-a response is not itself treated as a response-writing failure; the resulting
-HTTP status determines the automatic span status.
+phases: request filtering, resource-method execution, exception mapping, and
+entity materialization. This includes asynchronous JAX-RS responses. Spans
+started by exception mappers, writer interceptors, message-body writers, or
+streaming output are therefore children of the automatic span, without leaving
+its scope attached to a thread between phases. Helidon does not force the
+automatic span to be current across arbitrary response filters because a
+request filter can establish a nested scope which its paired response filter
+must close in nesting order. The automatic span still remains open during that
+work. If response processing or writing fails, Helidon records the available
+failure and ends the automatic span after Jersey finishes processing the failed
+request and any still-running asynchronous resource method returns. An
+application exception which Jersey successfully maps to a response
+is not itself treated as a response-writing failure; the resulting HTTP status
+determines the automatic span status.
 
 This setting is deprecated for removal in a future major release. After its
 removal, automatic incoming REST spans will always include response preparation
@@ -503,7 +510,9 @@ server-side work of preparing the response. It does not measure the later
 WebServer socket commit, network delivery, or wait for the client to receive or
 acknowledge the response. A downstream transport failure after JAX-RS processing
 has finished might therefore not be recorded on the span, but it cannot leave
-the span unfinished.
+the span unfinished. For long-lived `ChunkedOutput` and server-sent event
+responses, the span covers initial JAX-RS response processing but not the later
+serialization and transmission of individual chunks or events.
 
 ### OpenTelemetry Java Agent
 
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -467,12 +467,16 @@ When `telemetry.span.includes-response-write` is `false`, Helidon ends the
 span before serializing the response entity, preserving the behavior of earlier
 Helidon 4 releases. When it is `true`, Helidon ends the span after response
 serialization and encoding, when the last byte has been buffered for writing to
-the socket. The automatic span remains current while the response entity is
-written, so spans started by writer interceptors, message-body writers, or
-streaming output are children of the automatic span. If response writing fails,
-Helidon records the failure and ends the automatic span when Jersey finishes
-processing the failed request. This setting is deprecated for removal in a
-future major release.
+the socket. Helidon propagates the automatic span to the thread which processes
+the response, including asynchronous JAX-RS responses. The span remains current
+during response filtering and entity materialization, so spans started by writer
+interceptors, message-body writers, or streaming output are children of the
+automatic span. If response processing or writing fails, Helidon records the
+available failure and ends the automatic span when Jersey finishes processing
+the failed request. An application exception which Jersey successfully maps to
+a response is not itself treated as a response-writing failure; the resulting
+HTTP status determines the automatic span status. This setting is deprecated
+for removal in a future major release.
 
 For `telemetry.span.includes-response-write`, `true` measures the
 server-side work of preparing the response. It does not measure network delivery
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -482,16 +482,21 @@ When `telemetry.span.includes-response-write` is `false`, Helidon ends the
 span before serializing the response entity, preserving the behavior of earlier
 Helidon 4 releases. When it is `true`, Helidon ends the span after response
 serialization and encoding have populated Helidon's response stream and Jersey
-has finished processing the response. Helidon propagates the automatic span to
-the thread which processes the response, including asynchronous JAX-RS
-responses. The span remains current during response filtering and entity
-materialization, so spans started by writer interceptors, message-body writers,
-or streaming output are children of the automatic span. If response processing
-or writing fails, Helidon records the available failure and ends the automatic
-span when Jersey finishes processing the failed request. An application
-exception which Jersey successfully maps to a response is not itself treated as
-a response-writing failure; the resulting HTTP status determines the automatic
-span status. This setting is deprecated for removal in a future major release.
+has finished processing the response. The span remains open during response
+filtering, but Helidon makes it current only during discrete, thread-owned
+phases: request filtering, resource-method execution, and entity
+materialization. This includes asynchronous JAX-RS responses. Spans started by
+writer interceptors, message-body writers, or streaming output are therefore
+children of the automatic span, without leaving its scope attached to a thread
+between phases. If response processing or writing fails, Helidon records the
+available failure and ends the automatic span when Jersey finishes processing
+the failed request. An application exception which Jersey successfully maps to
+a response is not itself treated as a response-writing failure; the resulting
+HTTP status determines the automatic span status.
+
+This setting is deprecated for removal in a future major release. After its
+removal, automatic incoming REST spans will always include response preparation
+through the end of Jersey response processing.
 
 For `telemetry.span.includes-response-write`, `true` measures the
 server-side work of preparing the response. It does not measure the later
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -452,8 +452,23 @@ settings for automatic incoming REST request spans:
 
 | Property | Type | Default | Description |
 |----------|------|---------|-------------|
+| `telemetry.span.full.url` | `Boolean` | `false` | **Deprecated.** Whether the span name uses the absolute request path instead of the matched JAX-RS route. |
 | `telemetry.span.name-includes-method` | `Boolean` | `false` | **Deprecated.** Whether the span name includes the HTTP request method. |
 | `telemetry.span.includes-response-write` | `Boolean` | `false` | **Deprecated.** Whether the span includes preparing and writing the response entity. |
+
+By default, Helidon uses the matched, low-cardinality JAX-RS route in the span
+name, such as `/greet/{name}`. Setting `telemetry.span.full.url` to `true`
+uses the resolved absolute request path instead, such as
+`http://localhost:8080/greet/Joe`. The absolute path does not include the query
+string. Because resolved paths can contain user-supplied path values and
+therefore create high-cardinality span names, leave this setting `false` unless
+compatibility with an application that expects the older full-URL naming format
+requires it. The setting is deprecated because a future major release will use
+low-cardinality route-based span names unconditionally.
+
+The `telemetry.span.full.url` and `telemetry.span.name-includes-method` settings
+are independent. For example, setting both to `true` produces a span name such
+as `GET http://localhost:8080/greet/Joe`.
 
 Earlier Helidon 4 releases used OpenTelemetry semantic conventions which did
 not include the HTTP method in automatic incoming REST span names. Setting
@@ -466,21 +481,24 @@ naming convention unconditionally.
 When `telemetry.span.includes-response-write` is `false`, Helidon ends the
 span before serializing the response entity, preserving the behavior of earlier
 Helidon 4 releases. When it is `true`, Helidon ends the span after response
-serialization and encoding, when the last byte has been buffered for writing to
-the socket. Helidon propagates the automatic span to the thread which processes
-the response, including asynchronous JAX-RS responses. The span remains current
-during response filtering and entity materialization, so spans started by writer
-interceptors, message-body writers, or streaming output are children of the
-automatic span. If response processing or writing fails, Helidon records the
-available failure and ends the automatic span when Jersey finishes processing
-the failed request. An application exception which Jersey successfully maps to
-a response is not itself treated as a response-writing failure; the resulting
-HTTP status determines the automatic span status. This setting is deprecated
-for removal in a future major release.
+serialization and encoding have populated Helidon's response stream and Jersey
+has finished processing the response. Helidon propagates the automatic span to
+the thread which processes the response, including asynchronous JAX-RS
+responses. The span remains current during response filtering and entity
+materialization, so spans started by writer interceptors, message-body writers,
+or streaming output are children of the automatic span. If response processing
+or writing fails, Helidon records the available failure and ends the automatic
+span when Jersey finishes processing the failed request. An application
+exception which Jersey successfully maps to a response is not itself treated as
+a response-writing failure; the resulting HTTP status determines the automatic
+span status. This setting is deprecated for removal in a future major release.
 
 For `telemetry.span.includes-response-write`, `true` measures the
-server-side work of preparing the response. It does not measure network delivery
-or wait for the client to receive or acknowledge the response.
+server-side work of preparing the response. It does not measure the later
+WebServer socket commit, network delivery, or wait for the client to receive or
+acknowledge the response. A downstream transport failure after JAX-RS processing
+has finished might therefore not be recorded on the span, but it cannot leave
+the span unfinished.
 
 ### OpenTelemetry Java Agent
 
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -501,6 +501,10 @@ application exception which Jersey successfully maps to a response
 is not itself treated as a response-writing failure; the resulting HTTP status
 determines the automatic span status.
 
+This setting controls only automatic spans created by Helidon. When the
+OpenTelemetry Java Agent is present, the agent owns the automatic server span,
+so `telemetry.span.includes-response-write` has no effect.
+
 This setting is deprecated for removal in a future major release. After its
 removal, automatic incoming REST spans will always include response preparation
 through the end of Jersey response processing.
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -467,8 +467,12 @@ When `telemetry.span.includes-response-write` is `false`, Helidon ends the
 span before serializing the response entity, preserving the behavior of earlier
 Helidon 4 releases. When it is `true`, Helidon ends the span after response
 serialization and encoding, when the last byte has been buffered for writing to
-the socket. This setting is also deprecated for removal in a future major
-release.
+the socket. The automatic span remains current while the response entity is
+written, so spans started by writer interceptors, message-body writers, or
+streaming output are children of the automatic span. If response writing fails,
+Helidon records the failure and ends the automatic span when Jersey finishes
+processing the failed request. This setting is deprecated for removal in a
+future major release.
 
 For `telemetry.span.includes-response-write`, `true` measures the
 server-side work of preparing the response. It does not measure network delivery
--- a/docs/mp/telemetry.md
+++ b/docs/mp/telemetry.md
@@ -54,7 +54,9 @@ telemetry.span.name-includes-method = true
 
 The ability to use the older format is deprecated, and you should plan for its
 removal in a future major release of Helidon. For that reason Helidon logs a
-warning message if you use the older REST span naming convention.
+warning message if you use the older REST span naming convention. See
+[Helidon automatic span compatibility](#helidon-automatic-span-compatibility)
+for details about this setting and the response-writing compatibility setting.
 
 ## Usage
 
@@ -442,6 +444,35 @@ use a different exporter set `otel.traces.exporter` in your configuration to the
 appropriate value. The `zipkin` value is deprecated because OpenTelemetry
 stopped publishing its Zipkin exporter in 1.65.0; use `otlp` for traces. See
 the [examples](#examples) section below.
+
+### Helidon Automatic Span Compatibility
+
+Helidon supports the following deprecated vendor-specific compatibility
+settings for automatic incoming REST request spans:
+
+| Property | Type | Default | Description |
+|----------|------|---------|-------------|
+| `telemetry.span.name-includes-method` | `Boolean` | `false` | **Deprecated.** Whether the span name includes the HTTP request method. |
+| `telemetry.span.includes-response-write` | `Boolean` | `false` | **Deprecated.** Whether the span includes preparing and writing the response entity. |
+
+Earlier Helidon 4 releases used OpenTelemetry semantic conventions which did
+not include the HTTP method in automatic incoming REST span names. Setting
+`telemetry.span.name-includes-method` to `true` selects the current convention,
+which includes the method. Leaving it unset or setting it to `false` preserves
+the older span names for compatibility and causes Helidon to log a warning. The
+setting is deprecated because a future major release will use the current span
+naming convention unconditionally.
+
+When `telemetry.span.includes-response-write` is `false`, Helidon ends the
+span before serializing the response entity, preserving the behavior of earlier
+Helidon 4 releases. When it is `true`, Helidon ends the span after response
+serialization and encoding, when the last byte has been buffered for writing to
+the socket. This setting is also deprecated for removal in a future major
+release.
+
+For `telemetry.span.includes-response-write`, `true` measures the
+server-side work of preparing the response. It does not measure network delivery
+or wait for the client to receive or acknowledge the response.
 
 ### OpenTelemetry Java Agent
 
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -214,7 +214,7 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
                     request.setProperty(ServerSpanLifecycle.PROPERTY, lifecycle);
                 }
                 ServerSpanLifecycle finalLifecycle = lifecycle;
-                serverResponse.whenSent(() -> finalLifecycle.complete(span, scope, true));
+                serverResponse.whenSent(() -> finalLifecycle.responseSent(span, scope));
             } else {
                 span.end();
             }
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -28,7 +28,6 @@ import io.helidon.tracing.Scope;
 import io.helidon.tracing.Span;
 import io.helidon.tracing.SpanContext;
 import io.helidon.tracing.providers.opentelemetry.HelidonOpenTelemetry;
-import io.helidon.webserver.http.ServerResponse;
 
 import io.opentelemetry.semconv.ServerAttributes;
 import jakarta.enterprise.inject.Instance;
@@ -93,9 +92,6 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
 
     @jakarta.ws.rs.core.Context
     private ResourceInfo resourceInfo;
-
-    @jakarta.ws.rs.core.Context
-    private ServerResponse serverResponse;
 
     @Inject
     HelidonTelemetryContainerFilter(io.helidon.tracing.Tracer helidonTracer,
@@ -207,19 +203,10 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
             if (response.getStatusInfo().getFamily().compareTo(Response.Status.Family.SERVER_ERROR) == 0) {
                 span.status(Span.Status.ERROR);
             }
-            if (autoSpanIncludesResponseWrite) {
-                ServerSpanLifecycle lifecycle = (ServerSpanLifecycle) request.getProperty(ServerSpanLifecycle.PROPERTY);
-                if (lifecycle == null) {
-                    lifecycle = new ServerSpanLifecycle();
-                    request.setProperty(ServerSpanLifecycle.PROPERTY, lifecycle);
-                }
-                ServerSpanLifecycle finalLifecycle = lifecycle;
-                serverResponse.whenSent(() -> finalLifecycle.responseSent(span, scope));
-            } else {
+            // Response-write-inclusive spans end at Jersey's FINISHED request event.
+            if (!autoSpanIncludesResponseWrite) {
                 span.end();
             }
-
-
         } finally {
             if (!autoSpanIncludesResponseWrite) {
                 request.removeProperty(SPAN);
--- a/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
+++ b/microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
@@ -28,6 +28,7 @@ import io.helidon.tracing.Scope;
 import io.helidon.tracing.Span;
 import io.helidon.tracing.SpanContext;
 import io.helidon.tracing.providers.opentelemetry.HelidonOpenTelemetry;
+import io.helidon.webserver.http.ServerResponse;
 
 import io.opentelemetry.semconv.ServerAttributes;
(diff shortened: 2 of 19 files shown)


microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -192,7 +205,11 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
             if (response.getStatusInfo().getFamily().compareTo(Response.Status.Family.SERVER_ERROR) == 0) {
                 span.status(Span.Status.ERROR);
             }
-            span.end();
+            if (autoSpanIncludesResponseWrite) {
```
tomas-langer, line 204: When this option is enabled, the automatic server span remains unended but is no longer current during entity serialization. `scope.close()` has already run before this branch, so spans started by a `WriterInterceptor` or `StreamingOutput` inherit the previously active context instead of the automatic server span. The new tests access the span through `TestSpanListener` and explicitly assert that its scope is closed, so they do not exercise issue #12344's required response-materialization parent/child relationship. The option therefore extends the recorded duration while leaving the broken trace hierarchy unchanged.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -192,7 +205,11 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
...
+                serverResponse.whenSent(span::end);
```
tomas-langer: `whenSent` runs only after a response is sent successfully. The HTTP/1 and HTTP/2 implementations reach `afterSend()` only after their response writes succeed; a `WriterInterceptor` or message-body exception, client disconnect, HTTP/2 reset, or socket-write failure exits without invoking `span::end`. Because the request properties are removed immediately after registering this callback, no remaining path owns the span. With this option enabled, failed or cancelled response writes therefore leave automatic server spans unfinished and unexported.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Addressed by the recent commits.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Addressed.

tjquinno: Reworking the design to use JAX-RS only (not `ServerResponse.whenSent`) for clarity and simplicity. Commit coming shortly.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
```diff
@@ -0,0 +1,157 @@
...
+        boolean endSpan;
+        synchronized (this) {
+            resourceMethodFinished = true;
+            closeRequestScope = claimRequestScope();
+            endSpan = claimSpanEnd();
+        }
+        closeAndEnd(span, requestScope, closeRequestScope, endSpan);
+    }
+
+    void responseProcessingStarted(Span span) {
```
tomas-langer: With `telemetry.span.includes-response-write=true`, `RESP_FILTERS_START` activates a second automatic-server scope and keeps it open until `FINISHED`. A paired request/response filter which activated an ingress scope before the telemetry request filter then closes that earlier scope while the new automatic scope is current. OpenTelemetry ignores the out-of-order close; closing the automatic scope later restores the already-ended ingress context. A reused server thread can therefore carry the previous request's trace and baggage into a later request, and `HelidonTelemetryContainerFilter.parentSpanContext` prefers that stale current span over the next request's headers.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
```diff
@@ -0,0 +1,157 @@
...
+
+    void resourceMethodFinished(Span span, Scope requestScope) {
+        boolean closeRequestScope;
+        boolean endSpan;
+        synchronized (this) {
+            resourceMethodFinished = true;
+            closeRequestScope = claimRequestScope();
+            endSpan = claimSpanEnd();
+        }
+        closeAndEnd(span, requestScope, closeRequestScope, endSpan);
```
tomas-langer: The request filter activates and stores the automatic span scope on the dispatcher thread. For a Jersey `@ManagedAsync` resource, `RESOURCE_METHOD_FINISHED` runs on the managed-executor thread and reaches this close path for the dispatcher-thread-owned scope. OpenTelemetry scopes are thread-affine: default context storage can ignore the cross-thread close and leave the ended span current on the dispatcher thread, while strict context storage throws and can fail the request. The new async test uses `@Suspended`, whose resource method still finishes on the dispatcher thread, so it does not cover this execution mode.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
```diff
@@ -0,0 +1,92 @@
...
+            }
+            break;
+        default:
+            break;
+        }
+    }
+
+    static void cleanupIfFinished(ContainerRequest request, ServerSpanLifecycle lifecycle) {
+        if (lifecycle.isFinished()) {
+            request.removeProperty(HelidonTelemetryContainerFilter.SPAN);
```
tomas-langer: `cleanupIfFinished` removes telemetry properties from the `ContainerRequest` obtained from a Jersey `RequestEvent`. Jersey's `RequestEvent#getContainerRequest` contract says that listeners must not modify the returned request. Because composite request listeners receive the same object sequentially, this listener makes the visible request state depend on listener ordering and on unsupported mutation of monitoring-event data.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -95,6 +99,8 @@ class HelidonTelemetryContainerFilter implements ContainerRequestFilter, Contain
                                     Instance<HelidonTelemetryContainerFilterHelper> helpersInstance) {
         this.helidonTracer = helidonTracer;
         isAgentPresent = HelidonOpenTelemetry.AgentDetector.isAgentPresent(MpConfig.toHelidonConfig(mpConfig));
+        autoSpanIncludesResponseWrite = mpConfig.getOptionalValue(AUTO_SPAN_INCLUDES_RESPONSE_WRITE, Boolean.class)
```
tomas-langer, line 104: The PR description says configuring `telemetry.span.includes-response-write` logs a deprecation warning, but the current implementation only reads the value here. The revised head removed the earlier `DeprecatedConfig.get(..., AUTO_SPAN_INCLUDES_RESPONSE_WRITE)` call, and the only remaining telemetry deprecation warning is for `telemetry.span.name-includes-method`. Operators configuring this deprecated key therefore do not receive the migration warning advertised by the PR and release note.

docs/mp/telemetry.md
```diff
@@ -443,6 +445,61 @@ appropriate value. The `zipkin` value is deprecated because OpenTelemetry
...
+has finished processing the response. Helidon propagates the automatic span to
+the thread which processes the response, including asynchronous JAX-RS
+responses. The span remains current during response filtering and entity
+materialization, so spans started by writer interceptors, message-body writers,
+or streaming output are children of the automatic span. If response processing
+or writing fails, Helidon records the available failure and ends the automatic
+span when Jersey finishes processing the failed request. An application
+exception which Jersey successfully maps to a response is not itself treated as
+a response-writing failure; the resulting HTTP status determines the automatic
+span status. This setting is deprecated for removal in a future major release.
```
tomas-langer: The new compatibility setting is marked for removal in a future major release, but the documentation does not state which timing contract remains after removal. The adjacent deprecated span-name settings explain their future unconditional behavior; this setting does not say whether automatic spans will always include response writing or will retain the earlier pre-write end point. Users therefore cannot plan trace duration, child-span parenting, and failure-recording behavior across the upgrade.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
```diff
@@ -45,6 +45,8 @@ public TelemetryAutoDiscoverable() {
     @Override
     public void configure(FeatureContext ctx) {
         ctx.register(HelidonTelemetryContainerFilter.class);
+        ctx.register(HelidonTelemetryWriterInterceptor.class);
```
tomas-langer: The writer interceptor and request-event listener are registered unconditionally even though `telemetry.span.includes-response-write` defaults to `false`. Every Jersey request event and every serialized response entity therefore traverses the new hooks for users who never enable this deprecated compatibility option. The PR provides correctness tests but no targeted old-versus-current JMH evidence bounding the added default-path latency, throughput, or allocation cost.

tomas-langer (commented):
This PR targets `helidon-io/helidon` branch `helidon-4.x`, while the practical current-line destination for this MicroProfile telemetry behavior is `helidon-io/helidon-microprofile` `main`. Bounded searches found no corresponding carry-forward issue in either repository. Without durable carry-forward tracking, the new production behavior, tests, and documentation can be lost from the forward-development line and omitted from release planning.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
```diff
@@ -0,0 +1,157 @@
...
+    private boolean responseProcessingStarted;
+    private boolean requestFinished;
+    private boolean responseWritten;
+    private boolean requestScopeClosed;
+    private boolean responseScopeClosed;
+    private boolean spanEnded;
+    private Throwable failure;
+    private Scope responseScope;
+
+    synchronized void resourceMethodStarted() {
```
tomas-langer: This new Helidon-owned lifecycle state machine uses intrinsic monitor locking throughout its methods and blocks, but no third-party contract requires monitor ownership. This conflicts with Development Guidelines Rule 1.4 and Rule 1.4.1, which prohibit `synchronized` in Helidon-owned Java code except where an external contract specifically requires it. The lifecycle consequently introduces hidden monitor ownership into the MP telemetry request path contrary to the project's concurrency convention.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
```diff
@@ -0,0 +1,92 @@
...
+    private static void onRequestEvent(RequestEvent event) {
+        ContainerRequest request = event.getContainerRequest();
+        ServerSpanLifecycle lifecycle = (ServerSpanLifecycle) request.getProperty(ServerSpanLifecycle.PROPERTY);
+        if (lifecycle == null) {
+            return;
+        }
+
+        Span span = (Span) request.getProperty(HelidonTelemetryContainerFilter.SPAN);
+        Scope scope = (Scope) request.getProperty(HelidonTelemetryContainerFilter.SPAN_SCOPE);
+        switch (event.getType()) {
```
tomas-langer: This new lifecycle dispatch uses colon labels and `break` statements despite having no fall-through behavior. Development Guidelines Rule 1.6 directs new switch code toward enhanced switches and arrow labels. Using the older form in the central event-state transition increases edit-time fall-through risk and departs from the contributor-facing Java convention.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -57,8 +57,8 @@
 @Provider
 class HelidonTelemetryContainerFilter implements ContainerRequestFilter, ContainerResponseFilter {
     private static final System.Logger LOGGER = System.getLogger(HelidonTelemetryContainerFilter.class.getName());
-    private static final String SPAN = Span.class.getName();
-    private static final String SPAN_SCOPE = Scope.class.getName();
+    static final String SPAN = Span.class.getName();
```
tomas-langer, line 62: The changed declarations do not follow Development Guidelines Rule 2.1 member ordering. The package-private `SPAN` and `SPAN_SCOPE` constants appear after private static finals here; `HelidonTelemetryRequestEventListener.cleanupIfFinished` appears after a private static method; `ServerSpanLifecycle.isCurrent` appears after private instance methods; and `AutoSpanIncludesWriteTestBase` places package-private static finals after a private static final and nested annotation interfaces after static nested classes. The new lifecycle and test scaffolding therefore diverge from the repository's contributor and IDE ordering contract in several changed files.

tjquinno:
There is a pre-existing forward-port issue: https://github.com/helidon-io/helidon-microprofile/issues/100

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
tjquinno: Fixed. `RESP_FILTERS_START` no longer activates a response-wide scope. The initial request scope is closed at `REQUEST_FILTERED`, and the writer interceptor activates the automatic span only around entity materialization. Added strict-context, sequential-request coverage to verify that trace and baggage state from one request cannot become the next request’s parent.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
tjquinno: Fixed. The dispatcher-thread scope is now closed at `REQUEST_FILTERED`. A separate scope is activated at `RESOURCE_METHOD_START` and closed at `RESOURCE_METHOD_FINISHED` on the resource-execution thread. Added strict-context `@ManagedAsync` coverage to verify scope ownership and that the automatic span remains current during resource execution.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
tjquinno: Fixed. Removed `cleanupIfFinished` and all request-property mutation from the monitoring listener. The listener now reads the telemetry properties only; Jersey retains and releases the request and its properties as part of its normal lifecycle.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Addressed by correcting the PR description and release-note text rather than restoring the `DeprecatedConfig` lookup. The setting remains documented and declared as deprecated, but the revised text no longer claims that using it emits a runtime deprecation warning.

docs/mp/telemetry.md
tjquinno: Fixed. The documentation now states that, after this compatibility setting is removed in a future major release, automatic incoming REST spans will always include response preparation through the end of Jersey response processing. Later WebServer socket commit, network delivery, and client acknowledgement remain outside this span boundary.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
tjquinno: Fixed by eliminating the additional hooks from the default path. The request-event listener and writer interceptor are now registered only when `telemetry.span.includes-response-write=true`. Tests inspect the server configuration and verify that both providers are absent when the setting is omitted or `false`, and present when it is `true`.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
tjquinno: Fixed. `ServerSpanLifecycle` no longer uses intrinsic monitor locking. It now uses an atomic state bitmask and atomic references to coordinate scope closure, failure capture, and idempotent span completion.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
tjquinno: Fixed. The request-event dispatch now uses enhanced switch syntax with arrow labels. All currently defined Jersey request-event types are handled explicitly; the default remains to satisfy the Helidon 4 Checkstyle configuration and safely ignore future event types.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Fixed. Reordered the affected production and test members according to Development Guidelines Rule 2.1. I also audited the complete PR diff with the repository Checkstyle configuration; the module now reports zero Checkstyle violations.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
```diff
@@ -0,0 +1,84 @@
...
+        case REQUEST_FILTERED -> {
```
tomas-langer: AI Review: With `telemetry.span.includes-response-write=true`, this `REQUEST_FILTERED` callback can close the automatic server scope out of order. Jersey emits the event only after all request filters have run, so a later-priority filter can have a child tracing scope current and retain it until its paired response filter. `ServerSpanLifecycle.requestFiltered` then closes the server scope while that child is current. OpenTelemetry ignores the out-of-order delegate close, but the Helidon scope wrapper marks itself closed. When the child later closes it restores the automatic server context; the writer interceptor sees that span already current and creates no detach scope, and `FINISHED` ends the span while leaving its context attached to the reused request thread. A subsequent request can therefore inherit the previous request's trace and baggage.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
```diff
@@ -0,0 +1,84 @@
...
+            if (span != null) {
+                lifecycle.resourceMethodStarted(span);
+            }
+        }
+        case RESOURCE_METHOD_FINISHED -> {
+            if (span != null) {
+                lifecycle.resourceMethodFinished(span);
+            }
+        }
+        case RESP_FILTERS_START -> lifecycle.responseProcessingStarted();
```
tomas-langer: AI Review: In inclusive mode the automatic server span remains open but is not current during `ExceptionMapper` and `ContainerResponseFilter` callbacks. `REQUEST_FILTERED` closes the initial scope, the resource scope closes at method completion, and this event only records state; `RESP_FILTERS_FINISHED`, `EXCEPTION_MAPPER_FOUND`, and `EXCEPTION_MAPPING_FINISHED` are ignored. Application or library spans started in these standard Jakarta REST response-pipeline providers therefore attach to an outer context or start a separate trace instead of becoming children of the automatic REST span. This changes causal trace topology for response preparation that the enabled mode is intended to cover.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -192,15 +205,26 @@ public void filter(final ContainerRequestContext request, final ContainerRespons
...
-
-
+            // Response-write-inclusive spans end at Jersey's FINISHED request event.
+            if (!autoSpanIncludesResponseWrite) {
+                span.end();
+            }
         } finally {
-            request.removeProperty(SPAN);
-            request.removeProperty(SPAN_SCOPE);
+            if (!autoSpanIncludesResponseWrite) {
```
tomas-langer, line 226: AI Review: Inclusive mode skips removal of `SPAN` and `SPAN_SCOPE`, and the `FINISHED` listener also leaves those properties and `ServerSpanLifecycle.PROPERTY` on the request. Jersey retains the `ContainerRequest` for long-lived `ChunkedOutput` and SSE responses after `FINISHED`, while subsequent chunk writes bypass writer interceptors. Each open stream therefore retains the ended span, its baggage and context, the closed scope, and the lifecycle object for the stream's lifetime. Many long-lived streams can amplify this avoidable retention into material heap pressure.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
```diff
@@ -158,6 +166,9 @@ public void filter(ContainerRequestContext requestContext) {
 
         requestContext.setProperty(SPAN, helidonSpan);
         requestContext.setProperty(SPAN_SCOPE, helidonScope);
+        if (autoSpanIncludesResponseWrite) {
+            requestContext.setProperty(ServerSpanLifecycle.PROPERTY, new ServerSpanLifecycle());
```
tomas-langer: AI Review: Enabled mode allocates a `ServerSpanLifecycle` for every traced request, including its `AtomicInteger` and two `AtomicReference` instances, then performs repeated atomic read-modify-write and request-property operations across request, resource, and response phases. Because the listener and writer interceptor are registered application-wide, helper-excluded requests and entity writes also pay their fast-path callback and property checks. This is request and serialization hot-path overhead with allocation, fence, and asynchronous cache-line-transfer costs, but the PR contains no targeted throughput, latency, or allocation measurements establishing its impact.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
tjquinno: Addressed. `ServerSpanLifecycle` now records both the owning thread and the exact OpenTelemetry context established by the automatic request scope. It closes that scope only when both still match.

If a later request filter has a child span, baggage, or another nested context current at `REQUEST_FILTERED`, closure is deferred. The telemetry response filter retries after higher-priority paired response filters have restored the original context, with `FINISHED` as a final safe retry point.

Strict-context tests now cover both a distinct child span and a baggage-only nested context, followed by sequential requests which verify that no stale trace or baggage becomes the next request’s parent.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryRequestEventListener.java
tjquinno: Addressed where Jersey exposes a safe, phase-owned boundary. The request-event listener now activates the automatic span at `EXCEPTION_MAPPER_FOUND` and closes that scope at `EXCEPTION_MAPPING_FINISHED`. A test verifies that a span created by an exception mapper is a child of the automatic server span.

The implementation intentionally does not activate one scope across all arbitrary response filters. Jersey exposes only whole-phase events for response filters, and such a scope would recreate the out-of-order nesting problem when a paired request/response filter retains its own scope. The automatic span remains open during response filtering, and the outer writer interceptor makes it current during entity materialization for writer interceptors, message-body writers, and `StreamingOutput`. This distinction is now documented.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Addressed without modifying Jersey’s monitored request from the request-event listener.

Inclusive mode now stores only `ServerSpanLifecycle.PROPERTY`; it does not also store the older `SPAN` and `SPAN_SCOPE` properties. The outer writer interceptor removes the lifecycle property in `finally` after initial entity processing, for both successful and failed writes. This removes telemetry state before Jersey retains the request for later `ChunkedOutput` chunks or SSE events.

The request listener captures the lifecycle once so it can still process `FINISHED`, then releases its own reference after both Jersey completion and any outstanding asynchronous resource method have completed. Unit tests verify both request-property removal and listener-reference release, including the asynchronous event ordering.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/HelidonTelemetryContainerFilter.java
tjquinno: Addressed structurally.

The request-event listener and writer interceptor are now registered only for server runtimes when `telemetry.span.includes-response-write=true`. They are absent when the setting is omitted or `false`, and they are not registered with Jersey clients, so the default path incurs no new per-request event or entity-write callbacks.

Enabled mode now uses one lifecycle object containing volatile state fields and static `VarHandle` operations. This removes the earlier per-request `AtomicInteger` and `AtomicReference` helper allocations. The request listener captures the lifecycle property once, and the writer interceptor performs one unavoidable property lookup.

Tests verify provider absence for omitted and explicit-`false` configuration, presence for enabled server configuration, and absence from Jersey clients. No new JMH harness was added for this opt-in compatibility path; the overhead reduction and default-path isolation are established structurally.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    void writerFailure(Throwable throwable) {
+        if (throwable != null) {
+            FAILURE.compareAndSet(this, null, throwable);
+        }
+    }
+
+    void responseStatus(int status) {
```
tomas-langer, line 112: `responseStatus` marks the span `ERROR` as soon as it observes a 5xx, but the inclusive path invokes it once from the telemetry response filter and again from Jersey `FINISHED`. A response filter which runs after the telemetry filter can replace an initial 5xx with a final 2xx, 3xx, or 4xx. `FINISHED` then overwrites the `http.status_code` attribute, but the earlier error status remains sticky; the Helidon OpenTelemetry wrapper also does not reset status for `UNSET`. The exported span can therefore carry a final non-5xx status code while still reporting `ERROR`. The added status-transition test covers only `200 -> 500`, not the inverse final-status correction.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
```diff
@@ -45,6 +47,10 @@ public TelemetryAutoDiscoverable() {
     @Override
     public void configure(FeatureContext ctx) {
         ctx.register(HelidonTelemetryContainerFilter.class);
+        if (ctx.getConfiguration().getRuntimeType() == RuntimeType.SERVER && includesResponseWrite()) {
```
tomas-langer: These response-write providers are registered based only on server runtime and `telemetry.span.includes-response-write`. When the OpenTelemetry Java agent is present, `HelidonTelemetryContainerFilter` returns without creating either a Helidon automatic span or a `ServerSpanLifecycle`, but the registered application event listener still allocates a `RequestListener` for every request and the writer interceptor still performs its property lookup for every serialized entity. In the supported agent-present configuration, these hooks cannot affect the automatic span and add per-request allocation, event dispatch, and entity-write-path CPU overhead with no telemetry effect.

microprofile/telemetry/src/test/java/io/helidon/microprofile/telemetry/ServerSpanLifecycleTest.java
```diff
@@ -0,0 +1,363 @@
...
+        lifecycle.responseStatus(200);
+        ContainerRequest request = containerRequest();
+        request.setProperty(ServerSpanLifecycle.PROPERTY, lifecycle);
+        ContainerResponse response = new ContainerResponse(request, Response.status(503).build());
+        RequestEvent finishedEvent = requestEvent(request, response, RequestEvent.Type.FINISHED, false);
+        RequestEventListener requestListener = new HelidonTelemetryRequestEventListener().onRequest(finishedEvent);
+
+        requestListener.onEvent(finishedEvent);
+
+        Field lifecycleField = requestListener.getClass().getDeclaredField("lifecycle");
```
tomas-langer: These new tests use `getDeclaredField("lifecycle")` and `setAccessible(true)` to assert the private state of `RequestListener`. This violates `DEV-GUIDELINES.md` Rule 10.2 (`#rule-10-2`), which prohibits reflection into private state, and Rule 10.1 (`#rule-10-1`), which requires tests to verify observable behavior. The assertions couple the tests to the private listener/lifecycle split and can fail after an internal refactoring even when span completion and retention behavior remain unchanged.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/ServerSpanLifecycle.java
tjquinno: The lifecycle now applies only the final Jersey response status when ending the span. Tests cover both 200 → 500 and 500 → 200, including an end-to-end response-filter rewrite.

microprofile/telemetry/src/main/java/io/helidon/microprofile/telemetry/TelemetryAutoDiscoverable.java
tjquinno: The listener and writer interceptor are now registered only when response-write inclusion is enabled and no OpenTelemetry Java agent is present. An isolated agent-mode test verifies both providers are omitted.

microprofile/telemetry/src/test/java/io/helidon/microprofile/telemetry/ServerSpanLifecycleTest.java
tjquinno: Removed all reflection into private lifecycle state. The tests now verify observable behavior such as final status, error state, exactly-once completion, property immutability, and asynchronous completion.

tomas-langer (approved):

PR #12352, opened 2026-08-26
Fail active HTTP/2 streams when the connection closes

### LINKED ISSUE #12345: Fail active HTTP/2 streams when the connection closes
Active HTTP/2 streams are not failed when the connection reader terminates because of EOF, an I/O failure, or a protocol failure. `Http2ClientConnection` closes the transport and removes the connection handler, but it does not fail the entries in its active stream map. A request thread blocked in `Http2ClientStream.readHeaders()` can therefore remain there until its configured read timeout.

This was first observed as an unrelated `tests/core-windows` failure in PR #12073: `UnixDomainSocketTest.http2UnixSocketSameOriginRedirectPreservesTransportAddress` reached its 10-second JUnit timeout and was interrupted while waiting for response headers.

The waiter gap reproduces deterministically with `Http2ClientConnectionTest.MockedConnectionTestContext`:

1. Establish an HTTP/2 connection and complete stream 1.
2. Open stream 3 and begin waiting for its response headers.
3. Terminate the mocked inbound connection.

The connection closes, but the stream 3 header future remains pending instead of failing promptly with the connection-closure cause.

Connection-level termination should propagate its failure to every active stream and wake header, continue, and entity waiters while preserving the underlying cause.

tomas-langer:
Resolves #12345

Propagates unexpected HTTP/2 connection failures to active stream waiters and unblocks outbound flow control. Preserves completed responses and writes ordered GOAWAY frames before closing when the transport remains usable.

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/FlowControl.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/FlowControl.java
@@ -100,6 +100,14 @@ public interface FlowControl {
         void blockTillUpdate();
 
         /**
+         * Notify flow control that the connection has closed.
+         * <p>
+         * Releases threads blocked waiting for a window update.
+         */
+        default void connectionClosed() {
+        }
+
+        /**
          * MAX_FRAME_SIZE setting last received from the other side or default.
          * @return MAX_FRAME_SIZE
          */
--- a/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
+++ b/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
@@ -582,7 +582,8 @@ public class Http2ClientConnection {
             try {
                 closeTransport();
             } finally {
-                failedStreams.forEach(stream -> stream.completeTrailersFailure(actualFailure));
+                failedStreams.forEach(stream ->
+                        Thread.startVirtualThread(() -> stream.completeTrailersFailure(actualFailure)));
             }
         }
     }
--- a/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
+++ b/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
@@ -516,7 +516,10 @@ public class Http2ClientConnection {
             closeConnection(failure);
             return;
         }
-        closeOrderingLock.lock();
+        if (!closeOrderingLock.tryLock()) {
+            closeConnection(failure);
+            return;
+        }
         try {
             closeFailure.compareAndSet(null, failure);
             Http2ErrorCode errorCode = failure instanceof Http2Exception http2Exception
--- a/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
+++ b/webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
@@ -108,6 +108,7 @@ public class Http2ClientConnection {
     private final ReentrantLock reservedStreamsLock = new ReentrantLock();
     private final CountDownLatch initialSettingsLatch = new CountDownLatch(1);
     private final AtomicReference<State> state = new AtomicReference<>(State.OPEN);
+    private final AtomicReference<RuntimeException> closeFailure = new AtomicReference<>();
     private volatile int lastStreamId;
     private volatile long expectedPingAck = NO_PING_ACK;
     private volatile long peerMaxConcurrentStreams = Http2Setting.MAX_CONCURRENT_STREAMS.defaultValue();
@@ -275,6 +276,13 @@ public class Http2ClientConnection {
                 .writeInt64(pingId);
     }
 
+    private static RuntimeException connectionFailure(Throwable failure) {
+        if (failure instanceof RuntimeException runtimeException) {
+            return runtimeException;
+        }
+        return new IllegalStateException("HTTP/2 connection failed", failure);
+    }
+
     Http2ConnectionWriter writer() {
         return writer;
     }
@@ -362,12 +370,25 @@ public class Http2ClientConnection {
      * @param stream the stream
      */
     public void addStream(int streamId, Http2ClientStream stream) {
+        RuntimeException failure = null;
         Lock lock = streamsLock.writeLock();
         lock.lock();
         try {
-            this.streams.put(streamId, stream);
+            State currentState = state.get();
+            if (currentState == State.CLOSED || currentState == State.GO_AWAY) {
+                failure = closeFailure.get();
+                if (failure == null) {
+                    failure = new IllegalStateException("HTTP/2 connection is closed");
+                }
+            } else {
+                this.streams.put(streamId, stream);
+            }
         } finally {
             lock.unlock();
+        }
+        if (failure != null) {
+            stream.connectionClosed(failure);
+            throw failure;
         }
     }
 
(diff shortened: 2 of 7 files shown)


webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
```diff
@@ -494,40 +507,131 @@ void retire() {
...
+    }
+
+    private void close(RuntimeException failure) {
         initialSettingsLatch.countDown();
-        closeConnection();
+        if (!clientPrefaceSent) {
+            closeConnection(failure);
+            return;
+        }
+        closeOrderingLock.lock();
```
tvallin: `closeOrderingLock` remains held while `goAway()` waits for `Http2ConnectionWriter.streamLock`, which is held across blocking `DataWriter.writeNow`. If a socket write stalls, `closeTransport()` cannot close the socket to unblock it. A second normal `close()` now also blocks here; previously it could skip the duplicate GOAWAY and reach transport closure. Please keep transport closure able to break a stuck write, or bound the GOAWAY attempt, and add a test that does not manually release the blocked write.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientStream.java
```diff
@@ -719,14 +755,49 @@ void inboundHeaders(Http2Headers headers, boolean endOfStream) {
...
 
+    private void connectionClosed(RuntimeException failure, boolean notifyConnectionFlowControl) {
+        recordConnectionFailure(failure);
+        RuntimeException actualFailure;
+        StreamBuffer streamBuffer;
+        boolean inboundComplete;
+        inboundStateLock.lock();
+        try {
+            actualFailure = connectionFailure;
+            inboundComplete = readState == ReadState.END || inboundEndQueued;
```
tvallin, line 775: Treating `inboundEndQueued` as complete preserves the buffer after `closeTransport()` closes the socket. Draining that DATA later calls `data()` and restores inbound flow-control credit; the default strategy immediately writes a `WINDOW_UPDATE` through the closed socket, so the supposedly preserved body can fail. The mock writer still accepts post-close writes, hiding this. Please suppress post-close flow-control writes while retaining local accounting, and make the test writer reject writes after transport closure.

http/http2/src/main/java/io/helidon/http/http2/FlowControl.java
```diff
@@ -99,6 +99,14 @@ interface Outbound extends FlowControl {
          */
         void blockTillUpdate();
 
+        /**
+         * Notify flow control that the connection has closed.
+         * <p>
+         * Releases threads blocked waiting for a window update.
+         */
+        default void connectionClosed() {
```
tvallin, line 107: Could this lifecycle hook remain package-private on `StreamFlowControl` or the owned implementation? `FlowControl.Outbound` is exported public API, making this a permanent implementor contract, but existing/custom implementations inherit a no-op that does not release blocked waiters as documented. The production path owns the concrete flow-control implementation, so the public API expansion appears unnecessary.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
tomas-langer: Fixed in fa0bf8134b. close(RuntimeException) now uses tryLock; when another close is already performing GOAWAY, the concurrent close immediately closes the transport so it can unblock a stalled write. Added secondCloseBreaksBlockedConnectionWriter, which verifies that the second normal close closes the transport and completes without manually releasing the blocked write.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientStream.java
tomas-langer: Fixed in 71e89bed6d. writeWindowsUpdate now skips writes once the connection state is CLOSED, while the local flow-control accounting still occurs before the callback. The mock writer now rejects writes after transport closure, and connectionLossPreservesQueuedResponseEntity continues to verify that the buffered body drains successfully.

http/http2/src/main/java/io/helidon/http/http2/FlowControl.java
tomas-langer: The default method is intentional for compilation compatibility with Helidon test and benchmark implementations. FlowControl.Outbound is not designed as an external implementation SPI. The API Javadoc describes the behavior users get from the Helidon production implementation, whose override releases both connection- and stream-window waiters. We will keep this unchanged.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
```diff
@@ -1060,13 +1182,39 @@ private void ackSettings() {
...
+            try {
+                writeGoAway(streamId, errorCode, msg);
+            } finally {
+                goAwayWriteComplete.countDown();
+                if (errorCode != Http2ErrorCode.NO_ERROR) {
+                    errorGoAwayWriteComplete.countDown();
+                }
+            }
+        } else if (state.get() == State.GO_AWAY) {
+            await(goAwayWriteComplete);
```
danielkec, line 1203: The `tryLock()` escape does not cover retirement: `finishRetirement()` starts its GOAWAY before taking `closeOrderingLock`, so a concurrent `close()` acquires that lock and waits here instead of reaching `closeTransport()`. In a deterministic probe, blocking the retirement GOAWAY and then calling `close()` left the transport open and the close future timed out after 1 second; it completed only after the blocked write was manually released. This is the retirement-specific residual of the earlier stalled-write thread.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
```diff
@@ -494,40 +507,134 @@ void retire() {
...
+
+    private void closeConnection(RuntimeException failure) {
+        List<Http2ClientStream> failedStreams = beginClose(failure);
+        if (failedStreams != null) {
+            RuntimeException actualFailure = closeFailure.get();
+            try {
+                closeTransport();
             } finally {
-                closeListener.accept(this);
+                failedStreams.forEach(stream -> stream.completeTrailersFailure(actualFailure));
```
danielkec: `CompletableFuture.completeExceptionally()` runs non-async dependents inline, so this serial loop can stop on an application callback attached to the first trailers future. With two streams holding pending trailers, a blocking `whenComplete` on whichever future completed first left only 1 of 2 source futures done after transport closure. The sibling trailers waiter therefore does not receive the connection failure until the unrelated callback returns.

webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionTest.java
```diff
@@ -1087,6 +1836,11 @@ private MockedConnectionTestContext() {
             when(clientConnection.reader()).thenReturn(DataReader.create(this::nextInboundFrame));
             when(clientConnection.writer()).thenReturn(dataWriter);
             when(clientConnection.helidonSocket()).thenReturn(socket);
+            doAnswer(_ -> {
+                transportClosed.set(true);
+                failBlockedWrite();
```
danielkec, line 1912: `closeResource()` now calls `failBlockedWrite()`, so `closeNow()` can release and complete the cancel task before `closeNowDoesNotWaitForConnectionWriter` reaches its `assertFalse(cancel.isDone())`. The focused exact-head test failed with `expected: <false> but was: <true>`, and the current-main synthetic full-class run hit the same ordering. This makes the new regression coverage timing-dependent.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
tomas-langer: Fixed by `7135d8af5f`, `1177835ed8`, and `1be43d6749`. Retirement now publishes a distinct `RETIREMENT_GO_AWAY` state atomically, so normal close reaches transport closure to break a blocked retirement write. Protocol-error GOAWAY remains ordered and is not bypassed. The paired retirement and error-GOAWAY regressions cover both paths.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientConnection.java
tomas-langer: Fixed by `57f0bbe27f` and `20cc4b098e`. Each failed stream now completes pending trailers on its own virtual thread after transport closure, so a blocking dependent cannot delay sibling futures. The two-stream regression waits for both callback entries while the first callback remains blocked.

webclient/http2/src/test/java/io/helidon/webclient/http2/Http2ClientConnectionTest.java
tomas-langer: Fixed by `163aa10aa3`. The test now runs `closeNow()` asynchronously with a bounded completion wait, verifies transport closure, and then waits for the cancel task to finish from the transport-induced writer release. It no longer relies on the timing-dependent `cancel.isDone()` assertion or manual release.

spericas (approved):

PR #12353, opened 2026-08-26
Reject missing declarative request entities (27.x)

### LINKED ISSUE #12149: Carry forward #12147: reject missing declarative request entities
Carry forward the Helidon 4.x fix from issue #12147 and PR #12148 to `main`.

The 4.x fix makes generated declarative server bindings reject a required `@Http.Entity` parameter before decoding when the request has no entity, including `Content-Length: 0`. This prevents parameter binding from failing with `IllegalStateException` when user code requests an `InputStream` entity for an empty request.

Target line: `main` / 27.x.

tomas-langer:
Resolves #12149

Carries forward the missing declarative request entity handling from #12148 to `main`. Missing required `@Http.Entity` request bodies now return HTTP 400; the narrow target-line adaptation applies the same check to `@Http.RequestParams` entity components supported on `main`.

This is a source-faithful carry-forward with only the adaptation required for the target line. It is not an attempt to resolve inherited findings in the source change.


danielkec (approved):
Reviewed at 54ca7f319f1d4af97543a69ed66bac9889c3a70d. No findings.

PR #11506, opened 2026-04-01
[4.x] Add support for SSE over HTTP/2

spericas:
### Description

Use response output streams for SSE on both HTTP/1.1 and HTTP/2 while preserving correct close semantics and eager header flushing.

  - add flushHeaders to SinkProviderContext and wire it through HTTP/1.1 and HTTP/2 responses
  - keep DataWriterSseSink on response.outputStream() and close via response commit instead of closing the underlying connection
  - add HTTP/2 sink provider support in Http2ServerResponse
  - mark sink-backed responses as sent in both HTTP/1.1 and HTTP/2
  - add SSE tests for delayed header flushing and HTTP/2 sink completion

See #11053 and #9164. Initial version of PR provided here #11325.

PR #12364, opened 2026-08-27
Support SSE over HTTP/2 (27.x)

### LINKED ISSUE #12360: Carry forward #11053: support SSE over HTTP/2 on main
Carry forward the SSE-over-HTTP/2 behavior from #11053 and merged 4.x PR #11506 (`bca3e5262bd4f176108e60852ba425822f6330db`) to `main`.

Current `main` contains the protocol-owned SSE output-stream groundwork from PR #12251, including `SinkProviderContext.entityOutputStream(...)` and header flushing, but `Http2ServerResponse` still does not provide sink creation. Creating an `SseSink` for an HTTP/2 request therefore remains unsupported.

Adapt the 4.x behavior to the current response lifecycle rather than transplanting the older response plumbing:

- create SSE sinks over the HTTP/2 protocol-owned entity stream;
- send response headers before a delayed first event without ending the stream;
- preserve exact SSE event framing and prompt flushing;
- close only the HTTP/2 stream, leaving the connection and concurrent streams usable;
- preserve content encoding, stream filters, trailers, flow control, response lifecycle callbacks, reset, cancellation, and error handling;
- retain HTTP/1 chunked, compressed, and connection-reuse behavior;
- re-enable and execute the disabled SSE client coverage.

Add regression coverage beyond the source PR for:

- TLS with ALPN-negotiated HTTP/2;
- multiple concurrent SSE streams on one HTTP/2 connection with independent completion;
- client cancellation / `RST_STREAM` without affecting sibling streams or the connection;
- sustained output under HTTP/2 flow-control pressure and a slow consumer;
- delayed first-event header delivery on HTTP/2;
- encoding and stream-filter finalization before `END_STREAM`;
- trailers and response lifecycle callbacks occurring exactly once;
- HTTP/1 chunking, compression finalization, and connection reuse.

Related 4.x issues: #9164 and #11298.

tomas-langer:
Resolves #12360

Carries the SSE-over-HTTP/2 behavior from #11506 forward to `main`, adapting it to the current protocol-owned response entity-stream lifecycle. HTTP/2 responses now create `SseSink` instances through `SinkProviderContext`, flush headers without ending the stream, and finalize encoding, filters, trailers, and lifecycle callbacks when the sink closes.

Adds deterministic HTTP/2 transport coverage for ALPN, multiplexed streams and connection reuse, cancellation isolation, flow control, delayed events, trailers, and finalization; preserves HTTP/1 chunking, compression, and connection reuse; and re-enables the SSE client tests.

This is a source-faithful carry-forward with target-line adaptation. It does not attempt to resolve inherited findings from the source implementation.

### DIFF
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
@@ -353,12 +353,6 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         return headers.contains(HeaderNames.TRAILER);
     }
 
-    private void flushHeaders() {
-        if (outputStream != null) {
-            outputStream.flushHeaders();
-        }
-    }
-
     private void validateResponse(Http2Headers http2Headers) {
         http2Headers.validateResponse();
         if (validateResponseHeaders) {
@@ -388,6 +382,12 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         }
         headers.remove(HeaderNames.TRANSFER_ENCODING);
         headers.remove(HeaderNames.TRAILER);
+    }
+
+    private void flushHeaders() {
+        if (outputStream != null) {
+            outputStream.flushHeaders();
+        }
     }
 
     private static class BlockingOutputStream extends OutputStream {
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
@@ -347,6 +347,12 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         }
     }
 
+    void flushHeaders() {
+        if (outputStream != null) {
+            outputStream.flushHeaders();
+        }
+    }
+
     private static boolean sendTrailers(ServerResponseHeaders headers) {
         return headers.contains(HeaderNames.TRAILER);
     }
@@ -380,12 +386,6 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         }
         headers.remove(HeaderNames.TRANSFER_ENCODING);
         headers.remove(HeaderNames.TRAILER);
-    }
-
-    void flushHeaders() {
-        if (outputStream != null) {
-            outputStream.flushHeaders();
-        }
     }
 
     private static class BlockingOutputStream extends OutputStream {
--- a/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
+++ b/webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
@@ -252,9 +252,7 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         }
         OutputStream encodedOutputStream = contentEncode(outputStream);
         OutputStream applicationOutputStream = applyStreamFilters(encodedOutputStream);
-        return outputStream.headRequest
-                ? new ApplicationOutputStream(applicationOutputStream, outputStream)
-                : applicationOutputStream;
+        return new ApplicationOutputStream(applicationOutputStream, outputStream);
     }
 
     private boolean prepareResponse() {
@@ -409,6 +407,8 @@ class Http2ServerResponse extends ServerResponseBase<Http2ServerResponse> {
         private boolean headResponseSent;
         private long bytesWritten;
         private long headRepresentationLength;
+        private boolean discardWrites;
+        private RuntimeException writeFailure;
         private final Consumer<ServerResponseTrailers> beforeTrailers;
 
         private BlockingOutputStream(Http2ServerRequest request,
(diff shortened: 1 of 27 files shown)


webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerResponse.java
```diff
@@ -89,6 +93,16 @@ public Http2ServerResponse header(Header header) {
         return this;
     }
 
+    @Override
+    public <X extends Sink<?>> X sink(GenericType<X> sinkType) {
+        return createSink(findSinkProvider(sinkType, request),
+                          request,
+                          ctx,
+                          responsePreparation -> Optional.of(outputStream(responsePreparation)),
```
danielkec, line 101: After this new SSE bridge has flushed response headers, a client can reset the stream and the handler still enters the normal outbound flow-control path for its next `emit`. With a zero stream window, that call remains blocked until the flow-control timeout instead of completing after `RST_STREAM`, retaining the cancelled handler and its resources.

```text
SETTINGS { INITIAL_WINDOW_SIZE = 0 }
GET /controlled (stream 1, Accept: text/event-stream)
<- response HEADERS
RST_STREAM(CANCEL, stream 1)
release the handler to emit its next event
```

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -795,6 +769,40 @@ void commit() {
...
+            if (firstBuffer != null) {
+                try {
+                    write(BufferData.empty());
+                } catch (IOException e) {
+                    throw new ServerConnectionException("Failed to flush server response headers.", e);
+                }
+                responseSentRunnable.run();
+                return;
+            }
+            sendHeadersAndPrepare();
```
danielkec, line 806: With the supported `writeBufferSize(0)` configuration, this eager flush puts the new SSE response into chunked mode before the event is serialized. `SseEvent.create("")` performs a valid zero-length data-value write, which is encoded as `0\r\n\r\n` (the HTTP/1.1 terminal chunk); the client therefore receives only `data:` and the following SSE newlines corrupt persistent-connection framing.

```java
try (SseSink sink = response.sink(SseSink.TYPE)) {
    sink.emit(SseEvent.create(""));
}
```

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
```diff
@@ -373,53 +363,37 @@ public <X extends Sink<?>> X sink(GenericType<X> sinkType) {
...
+                          },
+                          this::flushHeaders);
     }
 
     protected Optional<OutputStream> sinkEntityOutputStream(Runnable responsePreparation) {
         Objects.requireNonNull(responsePreparation);
         beforeSend();
-        return Optional.empty();
+        responsePreparation.run();
+        return Optional.of(outputStream(false));
```
danielkec, line 390: A route or response filter can supply `Content-Length`, but this new reusable SSE stream commits an under-length body without detecting the missing bytes. For example, `Content-Length: 13` with `SseEvent.create("x")` emits only `data:x\n\n` (8 bytes); on the next request over the same HTTP/1.1 connection, the client consumes `HTTP/` from the next status line as the remaining entity bytes and response parsing is desynchronized.

```java
response.header(HeaderNames.CONTENT_LENGTH, "13");
try (SseSink sink = response.sink(SseSink.TYPE)) {
    sink.emit(SseEvent.create("x"));
}
```

danielkec (commented):
Three inline findings follow.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -278,10 +278,11 @@ public boolean rstStream(Http2RstStream rstStream) {
         resetCompletionLock.lock();
         try {
             remoteResetReceived = true;
-            rapidReset = writeState.get() == WriteState.INIT;
+            rapidReset = writeState.getAndSet(WriteState.END) == WriteState.INIT;
         } finally {
             resetCompletionLock.unlock();
         }
+        flowControl.outbound().streamClosed();
```
danielkec, line 285: When the server resets an SSE stream locally—for example, after a request declares `Content-Length: 0` and then sends DATA—this notification is not reached because it only runs in the peer `RST_STREAM` path. With outbound stream credit exhausted, `resetInvalidContentLength()` sends `RST_STREAM(PROTOCOL)` and removes the stream, but the SSE writer remains blocked until the flow-control timeout, retaining the route handler after the reset. A minimal sequence is:

```text
SETTINGS { INITIAL_WINDOW_SIZE = 64 }
HEADERS GET /sse, Content-Length: 0 (no END_STREAM)
<- response HEADERS + 64 DATA bytes
DATA "x" + END_STREAM
-> RST_STREAM(PROTOCOL); route still blocked after 500 ms
```

danielkec (commented):
One current-head follow-up finding.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ServerResponse.java
danielkec, line 806: When an SSE request contains `TE: trailers` but the route does not add a response `Trailer` header, this eager flush commits `Transfer-Encoding: chunked` without declaring `stream-result`. `commit()` still writes that trailer after the zero chunk; Helidon's HTTP/1 client therefore leaves those bytes unread and releases the pooled connection, so its next request fails while parsing the status line.

```text
GET /sse
TE: trailers

response: Transfer-Encoding: chunked (no Trailer)
body tail: 0\r\nstream-result: \r\n\r\n
next pooled request: HTTP Response did not contain HTTP status line
```

danielkec (commented):
One inline finding follows.

webserver/http2/src/main/java/io/helidon/webserver/http2/Http2ServerStream.java
```diff
@@ -603,12 +601,15 @@ public void run() {
                     ? h2Exception.code()
                     : Http2ErrorCode.STREAM_CLOSED;
             Http2RstStream rst = new Http2RstStream(errorCode);
-            try {
-                writeResetStream(rst, serverSettings);
-                resetSent = true;
-            } catch (SocketWriterException | UncheckedIOException writeFailure) {
-                connectionFailed = true;
-                throw writeFailure;
+            boolean sendReset = claimResetStreamSent();
```
danielkec, line 604: The blocked-writer part of the [earlier peer-reset report](https://github.com/helidon-io/helidon/pull/12364#discussion_r3873131406) is addressed, but when a client resets an SSE stream after the response headers, the peer reset is recorded in `remoteResetReceived` while this guard only considers `resetStreamSent`. Cleanup can therefore send `RST_STREAM(STREAM_CLOSED)` back on the already-reset stream. A minimal wire sequence is:

```text
C -> S: HEADERS(stream=1, END_STREAM)
S -> C: HEADERS(stream=1)
C -> S: RST_STREAM(stream=1, CANCEL)
S -> C: RST_STREAM(stream=1, STREAM_CLOSED)
```

[RFC 9113 Sections 5.1 and 6.4](https://www.rfc-editor.org/rfc/rfc9113.html#section-5.1) prohibit sending frames on a closed stream and specifically prohibit answering `RST_STREAM` with another reset; a peer can treat the extra frame as a connection error, affecting sibling streams.

danielkec (commented):
One current-head follow-up finding.

danielkec (approved):
Reviewed at exact head 52c8ec43341b059b23d928659f6d95571136f3fa; all prior findings are addressed and focused validation is green.

PR #12368, opened 2026-08-27
Accept legacy const sampler type configuration (27.x)

### LINKED ISSUE #12367: Accept legacy const sampler type configuration (27.x)
Carry forward the tracing sampler configuration compatibility fix from [4.x PR #11988](https://github.com/helidon-io/helidon/pull/11988) and commit [`49edbfa0`](https://github.com/helidon-io/helidon/commit/49edbfa0da4728c5d25ebeab8dda242cd2ffc5d4) to `main`.

Current `main` documentation still uses `sampler-type: const`, but configuration conversion targets `SamplerType.CONSTANT` and rejects the legacy `const` spelling. Existing configurations can therefore fail during tracing startup.

Accept `const` as a case-insensitive configuration alias for `CONSTANT` while keeping the public enum unchanged. Preserve the standard `constant` and `ratio` spellings and add regression coverage for lowercase and uppercase `const` and `constant` values.

This is a source-faithful carry-forward, not a broader tracing configuration cleanup.

tomas-langer:
Resolves #12367

This is a source-faithful carry-forward of [4.x PR #11988](https://github.com/helidon-io/helidon/pull/11988) and commit [`49edbfa0`](https://github.com/helidon-io/helidon/commit/49edbfa0da4728c5d25ebeab8dda242cd2ffc5d4) to `main`.

Adds a configuration factory so the legacy `const` sampler type maps to `SamplerType.CONSTANT` without adding another public enum value. Covers lowercase and uppercase `const` and `constant` configurations.

This carry-forward intentionally preserves inherited source behavior, including existing invalid-value diagnostics and generated config-metadata representation; it is not an attempt to resolve those findings.

### DIFF
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
@@ -390,6 +390,9 @@ class SchemaGenerator {
         if (optionInfo.prototypedBy().isPresent()) {
             return optionInfo.prototypedBy().get();
         }
+        if (!optionInfo.allowedValues().isEmpty()) {
+            return typeName.boxed();
+        }
 
         // check configured factory method
         var configuredDeclaredTypeName = optionInfo.configured()
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
@@ -240,14 +240,18 @@ class SchemaGenerator {
         if (typeInfo == null || typeInfo.kind() != ElementKind.ENUM) {
             return defaultValues;
         }
+        var allowedValues = optionInfo.allowedValues()
+                .stream()
+                .map(OptionAllowedValue::value)
+                .toList();
         return defaultValues.stream()
-                .map(defaultValue -> optionInfo.allowedValues()
-                        .stream()
-                        .map(OptionAllowedValue::value)
-                        .filter(value -> value.equals(value.toLowerCase(Locale.ROOT)))
-                        .filter(value -> value.equals(defaultValue.toLowerCase(Locale.ROOT)))
-                        .findFirst()
-                        .orElse(defaultValue))
+                .map(defaultValue -> allowedValues.contains(defaultValue)
+                        ? defaultValue
+                        : allowedValues.stream()
+                                .filter(value -> value.equals(value.toLowerCase(Locale.ROOT)))
+                                .filter(value -> value.equals(defaultValue.toLowerCase(Locale.ROOT)))
+                                .findFirst()
+                                .orElse(defaultValue))
                 .toList();
     }
 
--- a/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
+++ b/builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
@@ -19,6 +19,7 @@ import java.lang.System.Logger.Level;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Locale;
 
 import io.helidon.builder.api.Option;
 import io.helidon.codegen.CodegenContext;
@@ -187,7 +188,7 @@ class SchemaGenerator {
             builder.property("kind", enumValue);
         }
 
-        var defaultValue = defaultValue(optionInfo);
+        var defaultValue = defaultValue(typeInfo, optionInfo);
         if (defaultValue != null) {
             builder.property("value", defaultValue);
         }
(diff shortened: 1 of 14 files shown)


tvallin (dismissed):

danielkec (dismissed):

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprintSupport.java
```diff
@@ -0,0 +1,35 @@
...
+import io.helidon.config.Config;
+
+class ExtendedTracerConfigBlueprintSupport {
+
+    private ExtendedTracerConfigBlueprintSupport() {
+    }
+
+    @Prototype.ConfigFactoryMethod("samplerType")
+    static SamplerType createSamplerType(Config config) {
+        String samplerType = config.asString().get().toUpperCase(Locale.ROOT);
```
danielkec: When the supplied `Config` has a registered `SamplerType` mapper, this raw-string conversion bypasses it, so existing custom sampler aliases that worked before this change now fail during tracing startup. A minimal case is:

```java
Config config = Config.builder(ConfigSources.create(Map.of(
        "service", "x",
        "sampler-type", "custom")))
        .addMapper(SamplerType.class, ignored -> SamplerType.RATIO)
        .build();
ExtendedTracerConfig.create(config).samplerType();
```

This returns `RATIO` on `46ce1b0`, but this head throws `IllegalArgumentException`, despite programmatically registered mappers having documented priority.

danielkec: When `sampler-type` is a hybrid node, this reads its scalar value even though the node also has children. As a result, a malformed layered configuration that the base rejected now starts with `RATIO` and silently retains the default `samplerParam` of `1.0` (100% sampling), ignoring the intended `0.0`:

```java
var config = Config.just(ConfigSources.create(Map.of(
        "service", "x",
        "sampler-type", "ratio",
        "sampler-type.param", "0.0")));
var tracing = ExtendedTracerConfig.create(config);
// tracing.samplerType() == RATIO; tracing.samplerParam() == 1.0
```

`46ce1b0` throws `ConfigMappingException: config node must be a leaf but is not` for the same input.

tracing/tracing/src/main/java/io/helidon/tracing/ExtendedTracerConfigBlueprintSupport.java
```diff
@@ -0,0 +1,42 @@
...
+    private static final System.Logger LOGGER = System.getLogger(ExtendedTracerConfigBlueprintSupport.class.getName());
+
+    private ExtendedTracerConfigBlueprintSupport() {
+    }
+
+    @Prototype.ConfigFactoryMethod("samplerType")
+    static SamplerType createSamplerType(Config config) {
+        try {
+            return config.as(SamplerType.class).get();
+        } catch (ConfigMappingException e) {
```
danielkec, line 33: When a registered `SamplerType` mapper rejects `const` with `ConfigMappingException`, this catch treats that mapper failure as the built-in enum miss and returns `CONSTANT`, so the application's higher-priority mapping policy is bypassed and tracing can become always-on. A minimal case is:

```java
Config config = Config.builder(ConfigSources.create(Map.of(
        "service", "x",
        "sampler-type", "const")))
        .addMapper(SamplerType.class, node -> {
            throw new ConfigMappingException(node.key(), "rejected by application mapper");
        })
        .build();
ExtendedTracerConfig.create(config); // base throws; this head returns CONSTANT
```

builder/codegen/src/main/java/io/helidon/builder/codegen/SchemaGenerator.java
```diff
@@ -227,11 +228,29 @@ private String defaultValue(OptionInfo optionInfo) {
...
+                                           List<String> defaultValues) {
+        if (typeInfo == null || typeInfo.kind() != ElementKind.ENUM) {
+            return defaultValues;
+        }
+        return defaultValues.stream()
+                .map(defaultValue -> optionInfo.allowedValues()
+                        .stream()
+                        .map(OptionAllowedValue::value)
+                        .filter(value -> value.equals(value.toLowerCase(Locale.ROOT)))
+                        .filter(value -> value.equals(defaultValue.toLowerCase(Locale.ROOT)))
```
danielkec: When an enum has case-distinct constants, this lower-case lookup can replace an exact declared default with a different constant in generated metadata. For example:

```java
enum Mode { MODE, mode }

@Option.Default("MODE")
@Option.AllowedValue(value = "mode", description = "Lower")
@Option.AllowedValue(value = "MODE", description = "Upper")
Mode mode();
```

The base metadata emits `MODE`, but this head emits `mode`; Helidon's enum mapper prefers exact names, so metadata consumers can select `Mode.mode` while the generated builder defaults to `Mode.MODE`.

danielkec (commented):
Two current-head review findings.

tvallin (approved):

Issue #73, opened 2018-09-20
Display friendly message if port is in use


spericas:
## Environment Details
* Helidon Version: 0.10.0
* Helidon SE 
* JDK version:
* OS:
* Docker version (if applicable):

----------

## Problem Description
When a Helidon app starts and finds that it cannot bind to a port in use, it aborts without reporting a friendly message to the user. We should detect the condition and print a message explaining the conflict.

## Steps to reproduce
Using quickstart-se and having a process already listening to 8080 (default port).

spericas:
The cause is now printed as part of the error message.

PR #12371, opened 2026-08-27
Guard null context class loader lookups (27.x)

### LINKED ISSUE #12370: Guard null context class loader lookups (27.x)
Carry forward the core-owned portion of [4.x PR #12078](https://github.com/helidon-io/helidon/pull/12078) and commit [`55db2d30`](https://github.com/helidon-io/helidon/commit/55db2d3015c62521a254718d1c3d9f29fe502be9) to `main`.

Java permits a thread context class loader to be `null`. Several current core paths still dereference the TCCL while loading classpath resources, scanning features, configuring JUL, synthesizing annotations, reading OpenAPI documents, or configuring static content. These paths can fail with a null dereference instead of falling back to the defining class loader.

Apply the source behavior only to current core-owned paths: use the owning class loader when a lookup requires a non-null class loader and the TCCL is null, while preserving existing behavior for non-null TCCLs. Leave intentionally nullable propagation and restoration paths unchanged, omit modules that moved or were removed, and adapt regression coverage to the current core implementation.

The MicroProfile-owned portion is already complete in [helidon-microprofile PR #73](https://github.com/helidon-io/helidon-microprofile/pull/73). No matching Extensions-owned work remains.

This is a source-faithful carry-forward, not a broader class-loader propagation redesign and not an attempt to fix an incorrect but non-null TCCL.

tomas-langer:
Resolves #12370

This is a source-faithful carry-forward of the current core-owned portion of [4.x PR #12078](https://github.com/helidon-io/helidon/pull/12078) and commit [`55db2d30`](https://github.com/helidon-io/helidon/commit/55db2d3015c62521a254718d1c3d9f29fe502be9) to `main`.

Falls back to each owning class loader when a lookup requires a non-null class loader and the TCCL is null. Annotation proxy synthesis always uses the annotation type’s class loader so an unrelated non-null TCCL cannot make the annotation interface invisible. Other source semantics remain unchanged, including intentionally nullable propagation and restoration paths.

The target adaptation omits moved and removed monorepo paths and updates the static-content regression for the current feature-based implementation. The MicroProfile portion is already complete in [helidon-microprofile PR #73](https://github.com/helidon-io/helidon-microprofile/pull/73); Extensions has no remaining matching work.

This carry-forward otherwise preserves inherited source behavior and documentation and is not an attempt to resolve inherited findings.

### DIFF
--- a/metadata/reflection/src/main/java/io/helidon/metadata/reflection/AnnotationFactory.java
+++ b/metadata/reflection/src/main/java/io/helidon/metadata/reflection/AnnotationFactory.java
@@ -103,14 +103,6 @@ public final class AnnotationFactory {
                                                       new AnnotationInvocationHandler(annotationType, annotation)));
     }
 
-    private static ClassLoader classLoader() {
-        ClassLoader loader = Thread.currentThread().getContextClassLoader();
-        if (loader == null) {
-            return loader;
-        }
-        return AnnotationFactory.class.getClassLoader();
-    }
-
     // basically the same semantics as in `AptAnnotationFactory` (and scan based annotation factory)
     private static Optional<Annotation> createAnnotation(java.lang.annotation.Annotation annotation,
                                                          Set<TypeName> processedTypes) {
@@ -227,6 +219,14 @@ public final class AnnotationFactory {
             return result;
         }
         throw new IllegalArgumentException("Unknown primitive type: " + componentType.getName());
+    }
+
+    private static ClassLoader classLoader() {
+        ClassLoader loader = Thread.currentThread().getContextClassLoader();
+        if (loader == null) {
+            return AnnotationFactory.class.getClassLoader();
+        }
+        return loader;
     }
 
     private static class AnnotationInvocationHandler implements InvocationHandler {
--- a/metadata/reflection/src/main/java/io/helidon/metadata/reflection/AnnotationFactory.java
+++ b/metadata/reflection/src/main/java/io/helidon/metadata/reflection/AnnotationFactory.java
@@ -98,7 +98,7 @@ public final class AnnotationFactory {
         }
 
         // type is available, we can synthesize the annotation
-        return Optional.of((T) Proxy.newProxyInstance(classLoader(),
+        return Optional.of((T) Proxy.newProxyInstance(annotationType.getClassLoader(),
                                                       new Class[] {annotationType},
                                                       new AnnotationInvocationHandler(annotationType, annotation)));
     }
@@ -219,14 +219,6 @@ public final class AnnotationFactory {
             return result;
         }
         throw new IllegalArgumentException("Unknown primitive type: " + componentType.getName());
-    }
-
-    private static ClassLoader classLoader() {
-        ClassLoader loader = Thread.currentThread().getContextClassLoader();
-        if (loader == null) {
-            return AnnotationFactory.class.getClassLoader();
-        }
-        return loader;
     }
 
     private static class AnnotationInvocationHandler implements InvocationHandler {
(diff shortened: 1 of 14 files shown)


metadata/reflection/src/main/java/io/helidon/metadata/reflection/AnnotationFactory.java
```diff
@@ -229,6 +221,14 @@ private static Object toPrimitiveArray(Class<?> componentType, Object value) {
         throw new IllegalArgumentException("Unknown primitive type: " + componentType.getName());
     }
 
+    private static ClassLoader classLoader() {
+        ClassLoader loader = Thread.currentThread().getContextClassLoader();
+        if (loader == null) {
+            return AnnotationFactory.class.getClassLoader();
+        }
+        return loader;
```
danielkec: When the current thread has a non-null isolated TCCL, this returns that loader even though `TypeFactory.toClass(...)` resolves the annotation through Helidon's defining loader. `Proxy.newProxyInstance` then rejects the resolved interface, so `AnnotationFactory.synthesize` fails in container or plugin threads with an unrelated TCCL. A minimal case is:

```java
@Retention(RUNTIME)
@interface ProbeAnnotation {}

Thread.currentThread().setContextClassLoader(
        new URLClassLoader(new URL[0], null));
AnnotationFactory.synthesize(
        Annotation.create(ProbeAnnotation.class));
```

The base returns the synthesized annotation; this head throws `IllegalArgumentException: ... is not visible from class loader`.

metadata/reflection/src/main/java/io/helidon/metadata/reflection/AnnotationFactory.java
```diff
@@ -98,19 +98,11 @@ public static <T extends java.lang.annotation.Annotation> Optional<T> synthesize
         }
 
         // type is available, we can synthesize the annotation
-        return Optional.of((T) Proxy.newProxyInstance(classLoader(),
+        return Optional.of((T) Proxy.newProxyInstance(annotationType.getClassLoader(),
```
danielkec, line 101: This change fixes synthesis when the calling thread has an unrelated non-null TCCL, but the metadata/reflection tests do not exercise that case. The unchanged `AnnotationTest` passes on `9fe7aebe`, where an isolated TCCL still makes `synthesize` throw `IllegalArgumentException`, so the test suite would not detect a recurrence.

metadata/reflection/src/test/java/io/helidon/metadata/reflection/AnnotationTest.java
```diff
@@ -108,6 +108,23 @@ public void testCreateAnnotation() {
         assertThat(annotation.stringValues("singletonList"), optionalValue(is(List.of("value"))));
     }
 
+    @Order(1)
+    @Test
+    public void testSyntheticAnnotationWithUnrelatedContextClassLoader() {
+        Thread thread = Thread.currentThread();
+        ClassLoader original = thread.getContextClassLoader();
+        try {
+            thread.setContextClassLoader(ClassLoader.getPlatformClassLoader());
```
danielkec, line 134: This covers an unrelated non-null TCCL, but the platform-loader case still passes against base `46ce1b0`, where a null TCCL is passed to `Proxy.newProxyInstance` and annotation synthesis throws `IllegalArgumentException`. The two prior implementations fail opposite cases (`46ce1b0`: null fails/platform passes; `9fe7aebe`: null passes/platform fails), so the checked-in regression would not detect a return to the PR's original null-TCCL failure. A minimal reproducer for that distinct path is:

```java
ClassLoader original = thread.getContextClassLoader();
try {
    thread.setContextClassLoader(null);
    AnnotationFactory.synthesize(helidonAnnotation).orElseThrow();
} finally {
    thread.setContextClassLoader(original);
}
```

danielkec (approved):

PR #12376, opened 2026-08-27
Fix nullable cascading validation [4.x]

### LINKED ISSUE #12375: Declarative @Validation.Valid invokes nested validators for null values [4.x]
## Description

Helidon declarative validation generates an unconditional nested-validator call for values annotated with `@Validation.Valid`.

When the value is null, the generated type validator can dereference it while validating its properties. This makes `@Validation.Valid` behave as though the value were also required, even though cascading validation and nullability should be independent concerns.

This is the core issue preventing helidon-extensions#118 from preserving optional and nullable OpenAPI request bodies: https://github.com/helidon-io/helidon-extensions/issues/118

## Affected code

`declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java`

The generated code currently resembles:

```java
childValidator.check(validationContext, value);
```

This call is made even when `value` is null.

## Expected behavior

`@Validation.Valid` should recursively validate a value only when it is present:

```java
if (value != null) {
    childValidator.check(validationContext, value);
}
```

Nullability should be controlled separately:

- `@Validation.Valid` means "recursively validate this value when present."
- `@Validation.NotNull` means "reject null."
- `@Validation.Valid @Validation.NotNull` must continue to reject null.

## Actual behavior

A null value annotated only with `@Validation.Valid` is passed to its generated type validator. The generated validator can then dereference the null value while accessing validated properties.

This affects cascading validation on:

- Method parameters
- Method return values
- Validated fields and properties
- Collection and array elements
- Map values

## Minimal reproduction

```java
@Validation.Validated
record Child(
        @Validation.String.Length(min = 3)
        String code) {
}

interface ChildService {
    String process(@Validation.Valid Child child);
}
```

Calling the intercepted service with a null value:

```java
service.process(null);
```

should be accepted because the parameter is not annotated with `@Validation.NotNull`. Instead, the generated validation interceptor invokes the `Child` type validator with null.

The same problem occurs for null container elements:

```java
String processAll(List<@Validation.Valid Child> children);
```

For example:

```java
service.processAll(Collections.singletonList(null));
```

## Proposed fix

Guard only the generated nested-validator call:

```java
if (value != null) {
    childValidator.check(validationContext, value);
}
```

Do not guard the entire validation block, because independent constraints such as `@Validation.NotNull` must still receive and validate the null value.

## Acceptance criteria

- A null `@Validation.Valid` method parameter is skipped.
- A null `@Validation.Valid` return value is skipped.
- A null validated field or property is skipped.
- Null collection and array elements are skipped.
- Null map values are skipped.
- Present values continue to receive cascading validation.
- `@Validation.Valid @Validation.NotNull` continues to reject null.
- Existing `Optional<@Validation.Valid T>` behavior remains unchanged.

## OpenAPI impact

Fixing this allows OpenAPI generators to preserve optional and nullable request bodies while still generating cascading validation.

Marking such request bodies as `required: true` is not an appropriate workaround because it changes the source API contract.

arjav-desai:
Fixes #12375.

`@Validation.Valid` now skips cascading validation when the value is null. This allows nullable and optional values to remain optional while still recursively validating values that are present.

An independent `@Validation.NotNull` constraint continues to reject null values.

Regression tests cover method parameters, return values, fields, collection and array elements, and map values.

### Validation

- 36 focused tests passed
- 0 failures or errors
- `git diff --check` passed

### DIFF
--- a/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java
+++ b/declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java
@@ -83,10 +83,14 @@ final class ValidationHelper {
     }
 
     static void addValidationOfValid(ContentBuilder<?> contentBuilder, String validatorField, String varName) {
-        contentBuilder.addContent(validatorField)
+        contentBuilder.addContent("if (")
+                .addContent(varName)
+                .addContentLine(" != null) {")
+                .addContent(validatorField)
                 .addContent(".check(validation__ctx, ")
                 .addContent(varName)
-                .addContentLine(");");
+                .addContentLine(");")
+                .addContentLine("}");
     }
 
     static ValidationContext validationContext(TypeName generatedType,
--- a/validation/tests/validation/src/test/java/io/helidon/validation/tests/validation/TypeValidationTest.java
+++ b/validation/tests/validation/src/test/java/io/helidon/validation/tests/validation/TypeValidationTest.java
@@ -57,7 +57,7 @@ public class TypeValidationTest {
         var value = new NullableCascadingType(null,
                                               Collections.singletonList(null),
                                               map);
-        var arrayValue = new NullableCascadingArrayType(new ValidatedType[]{null});
+        var arrayValue = new NullableCascadingArrayType(new ValidatedType[] {null});
 
         assertThat(validator.validate(NullableCascadingType.class, value).valid(), is(true));
         assertThat(validator.validate(NullableCascadingArrayType.class, arrayValue).valid(), is(true));
--- a/validation/tests/validation/src/test/java/io/helidon/validation/tests/validation/TypeValidationTest.java
+++ b/validation/tests/validation/src/test/java/io/helidon/validation/tests/validation/TypeValidationTest.java
@@ -17,6 +17,8 @@
 package io.helidon.validation.tests.validation;
 
 import java.math.BigDecimal;
+import java.util.Collections;
+import java.util.HashMap;
 
 import io.helidon.testing.junit5.Testing;
 import io.helidon.validation.TypeValidation;
@@ -46,6 +48,19 @@ public class TypeValidationTest {
         response = validator.validate(ValidatedType.class, validatedType);
         assertThat(response.valid(), is(false));
         assertThat(response.message(), is("41 is less than 42"));
+    }
+
+    @Test
+    public void validateNullableCascadingBoundaries() {
+        var map = new HashMap<String, ValidatedType>();
+        map.put("null", null);
+        var value = new NullableCascadingType(null,
+                                              Collections.singletonList(null),
+                                              map);
+        var arrayValue = new NullableCascadingArrayType(new ValidatedType[]{null});
+
+        assertThat(validator.validate(NullableCascadingType.class, value).valid(), is(true));
+        assertThat(validator.validate(NullableCascadingArrayType.class, arrayValue).valid(), is(true));
     }
 
     @Test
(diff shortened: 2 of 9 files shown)


declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java
```diff
@@ -83,10 +83,14 @@ static String addTypeValidator(FieldHandler fieldHandler, TypeName validatedType
     }
 
     static void addValidationOfValid(ContentBuilder<?> contentBuilder, String validatorField, String varName) {
-        contentBuilder.addContent(validatorField)
+        contentBuilder.addContent("if (")
```
tomas-langer, line 86: PR 12376 targets `helidon-4.x`, but the same unconditional `addValidationOfValid` call remains on `main`, and there is no carry-forward issue referencing #12375. Without forward-port tracking, 4.5.4 can ship this fix while 27.x/main remains affected, including the stated `helidon-extensions#118` unblock.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java
```diff
@@ -83,10 +83,14 @@ static String addTypeValidator(FieldHandler fieldHandler, TypeName validatedType
...
+                .addContent(varName)
+                .addContentLine(" != null) {")
```
tomas-langer, line 88: This changes the public `@Validation.Valid` behavior so cascaded validation skips null, while null rejection remains the independent `@Validation.NotNull` contract. The `Validation.Valid` Javadoc and `docs/se/injection/declarative.md` still only say the object is validated and do not describe that null is accepted and skipped. As a result, users cannot discover when `@NotNull` is required, including nullable OpenAPI request-body users.

validation/tests/validation/src/test/java/io/helidon/validation/tests/validation/TypeValidationTest.java
```diff
@@ -48,6 +50,19 @@ public void validateType() {
...
     }
 
+    @Test
+    public void validateNullableCascadingBoundaries() {
+        var map = new HashMap<String, ValidatedType>();
+        map.put("null", null);
+        var value = new NullableCascadingType(null,
+                                              Collections.singletonList(null),
+                                              map);
+        var arrayValue = new NullableCascadingArrayType(new ValidatedType[]{null});
```
tomas-langer: This new initializer is `new ValidatedType[]{null}`, but repository codestyle sets `SPACE_BEFORE_ARRAY_INITIALIZER_LBRACE=true`, and neighboring validation tests use `new String[] {"valid"}`. The new test therefore diverges from the repository's Java formatting convention.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java
arjav-desai: I just filed https://github.com/helidon-io/helidon/issues/12377 and will raise a PR for that once this is merged.

declarative/codegen/src/main/java/io/helidon/declarative/codegen/validation/ValidationHelper.java
arjav-desai: Updated javadoc and use docs.

validation/tests/validation/src/test/java/io/helidon/validation/tests/validation/TypeValidationTest.java
arjav-desai: Added space before initializer.

tomas-langer (approved):

PR #12381, opened 2026-08-28
Add non-mutating LRU cache lookup (4.x)

tomas-langer:
## Summary

- Add `LruCache.peek(K key, V defaultValue)` for lookup without changing LRU order.
- Guard Helidon's implementation with the existing read lock.
- Preserve compatibility for custom implementations with an unsupported default method.

Existing cache operations and their locking remain unchanged.

### DIFF
--- a/common/common/src/main/java/io/helidon/common/LruCache.java
+++ b/common/common/src/main/java/io/helidon/common/LruCache.java
@@ -15,6 +15,7 @@
  */
 package io.helidon.common;
 
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Supplier;
 
@@ -64,6 +65,22 @@ public interface LruCache<K, V> {
      * @return value if present or empty
      */
     Optional<V> get(K key);
+
+    /**
+     * Get a value from the cache without marking it as recently used.
+     * Custom implementations may not support this operation.
+     *
+     * @param key          key to retrieve
+     * @param defaultValue value to return if the key is not present
+     * @return cached value if present, otherwise {@code defaultValue}
+     * @throws NullPointerException if {@code key} or {@code defaultValue} is {@code null}
+     * @throws UnsupportedOperationException if this cache does not support peeking
+     */
+    default V peek(K key, V defaultValue) {
+        Objects.requireNonNull(key);
+        Objects.requireNonNull(defaultValue);
+        throw new UnsupportedOperationException("Peeking is not supported by this cache");
+    }
 
     /**
      * Remove a value from the cache.
--- a/common/common/src/main/java/io/helidon/common/LruCacheImpl.java
+++ b/common/common/src/main/java/io/helidon/common/LruCacheImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, 2025 Oracle and/or its affiliates.
+ * Copyright (c) 2024, 2026 Oracle and/or its affiliates.
  *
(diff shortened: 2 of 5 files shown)


common/common/src/main/java/io/helidon/common/LruCache.java
```diff
@@ -15,13 +15,18 @@
...
 /**
  * Least recently used cache.
- * This cache has a capacity. When the capacity is reached, the oldest record is removed from the cache when a new one
- * is added.
+ * This cache has a capacity. When adding a record would exceed the capacity, the least recently used record is removed.
+ * A capacity of zero disables caching while preserving the behavior of methods that compute or return values.
+ * <p>
+ * Implementations must be thread safe. Getting or replacing a value marks it as recently used; peeking does not.
+ * Cache keys must not call methods on the same cache instance, directly or indirectly, from their
+ * {@link Object#hashCode()} or {@link Object#equals(Object)} implementations. Such reentrant access is not supported.
```
Verdent: Documenting this as unsupported does not preserve the existing behavior of this `@Api.Stable` API. A key whose `hashCode()` calls `cache.size()` completes on the current 4.x implementation, but with this lock change `cache.put(key, value)` never returns because it holds the write stamp while `size()` waits for a read stamp. That is an incompatible runtime change for a maintained public API (Rule 4.1.5).

common/common/src/main/java/io/helidon/common/LruCacheImpl.java
```diff
@@ -33,98 +32,94 @@
  */
 final class LruCacheImpl<K, V> implements LruCache<K, V> {
     private final SequencedMap<K, V> backingMap = new LinkedHashMap<>();
-    private final ReadWriteLock rwLock = new ReentrantReadWriteLock(true);
-    private final Lock readLock = rwLock.readLock();
-    private final Lock writeLock = rwLock.writeLock();
-
+    private final StampedLock lock = new StampedLock();
```
Verdent: The previous lock was fair, while `StampedLock` allows a new reader to enter after a writer is queued. With one `peek` blocked in a valid key `hashCode()`, a second `peek` can acquire another read stamp; relaying readers this way keeps `get`, `put`, `remove`, and `clear` blocked indefinitely. On JDK 21 I reproduced this: the queued writer ran only after the final reader released its stamp.

common/common/src/main/java/io/helidon/common/LruCache.java
```diff
@@ -48,50 +53,73 @@ static <K, V> LruCache<K, V> create() {
...
     /**
-     * Either return a cached value or compute it and cache it.
+     * Either return a cached value or compute it and cache it if the capacity is greater than zero.
      * In case this method is called in parallel for the same key, the value actually present in the map may be from
      * any of the calls.
-     * This method always returns either the existing value from the map, or the value provided by the supplier. It
-     * never returns a result from another thread's supplier.
+     * This method always returns either the value observed in the cache by this invocation, or the value provided by
+     * this invocation's supplier. A value supplied by a concurrent invocation may be stored in the cache, but it is not
+     * returned in place of this invocation's supplied value.
```
Verdent: All current `computeValue` tests are sequential, so none exercises this same-key concurrent guarantee after the synchronization rewrite. The suite would still pass if this refactor returned a value supplied by another caller.

danielkec (approved):

PR #12384, opened 2026-08-28
Reject relative request targets at the protocol boundary (27.x)

### LINKED ISSUE #12334: Reject relative request targets at the protocol boundary (27.x)
Blocked by #12329. Once that PR is merged, carry the request-target validation fix forward to `main`.

This tracks the `main` follow-up for #12319. Preserve the protocol-specific behavior from the source change:

- Reject a relative HTTP/1.1 origin-form request-target such as `boards/` at the protocol boundary with HTTP 400.
- Treat a malformed HTTP/2 `:path` as a stream-level `PROTOCOL_ERROR`, leaving the connection usable by later streams.
- Continue to accept CONNECT authority-form, `OPTIONS *`, and absolute-form request-targets.

tomas-langer:
Resolves #12334

Carries forward the request-target validation from #12329 to main, preserving protocol-owned HTTP/1.1 and HTTP/2 behavior. It also aligns ordinary HTTP/2 CONNECT framing with RFC 9113 while returning 501 because HTTP/2 tunnel lifecycle is not supported.

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -361,6 +361,11 @@ public class Http2Headers {
         if (pseudoHeaders.size() != 0) {
             throw new Http2Exception(Http2ErrorCode.PROTOCOL, "Pseudo header in trailers");
         }
+        for (Header header : headers) {
+            if (header.headerName().isPseudoHeader()) {
+                throw new Http2Exception(Http2ErrorCode.PROTOCOL, "Pseudo header in trailers");
+            }
+        }
     }
 
     /**
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -353,9 +353,9 @@ public class Http2Headers {
     }
 
     /**
-     * Validate request or response trailers.
+     * Validate that request or response trailers do not contain pseudo-headers.
      *
-     * @throws Http2Exception in case the trailers are invalid
+     * @throws Http2Exception if the trailers contain a pseudo-header
      */
     public void validateTrailers() throws Http2Exception {
         if (pseudoHeaders.size() != 0) {
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -574,7 +574,7 @@ public class Http2Headers {
     }
 
     private static boolean authoritiesMatch(String scheme, String authority, String host) {
-        if (scheme == null && authority.equals(host)) {
+        if (scheme == null && authority.equalsIgnoreCase(host)) {
             return true;
         }
         try {
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -396,10 +396,10 @@ public class Http2Headers {
                 throw new Http2Exception(Http2ErrorCode.PROTOCOL, "Missing :scheme pseudo header");
             }
             if (!pseudoHeaders.hasPath()) {
-                throw new Http2Exception(Http2ErrorCode.PROTOCOL, "Missing :path pseudo header");
+                throw new Http2Exception(Http2ErrorCode.PROTOCOL, "Missing :path pseudo header", true);
             }
             if (pseudoHeaders.path().isEmpty()) {
-                throw new Http2Exception(Http2ErrorCode.PROTOCOL, ":path pseudo header has empty value");
+                throw new Http2Exception(Http2ErrorCode.PROTOCOL, ":path pseudo header has empty value", true);
             }
         }
         List<String> hostValues = headers.all(HeaderNames.HOST, List::of);
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -26,6 +26,7 @@ import java.util.Set;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
+import io.helidon.common.buffers.Ascii;
 import io.helidon.common.buffers.BufferData;
 import io.helidon.common.uri.UriAuthority;
 import io.helidon.http.Header;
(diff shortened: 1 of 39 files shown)


webserver/http2/src/main/java/io/helidon/webserver/http2/Http2Connection.java
```diff
@@ -906,7 +951,9 @@ private void doHeaders(Limit limit) {
             return;
         }
 
-        headers.validateRequest();
+        if (!validateRequestHeaders(headers, stream, streamId, newStream, endOfStream)) {
```
tvallin, line 954: [P1] Keep malformed requests scoped to their stream

This reset boundary misses two `:path` cases. A non-CONNECT request with no `:path` reaches `validateRequestHeaders()`, whose CONNECT-only catch rethrows; an explicitly empty `:path` is rejected by `Http2Headers.create()` before this call. Both `PROTOCOL_ERROR`s reach `Http2Connection.handle()`, which sends GOAWAY and aborts unrelated streams. Please move or delay request-specific validation so missing and empty `:path` reset only the affected stream—without swallowing true HPACK `COMPRESSION_ERROR`s—and test a healthy follow-up stream.

http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
```diff
@@ -543,7 +556,7 @@ private static Http2Headers create(ServerRequestHeaders httpHeaders, PseudoHeade
 
     private static boolean authoritiesMatch(String scheme, String authority, String host) {
         try {
-            int defaultPort = defaultPort(scheme);
+            int defaultPort = scheme == null ? UriAuthority.UNDEFINED_PORT : defaultPort(scheme);
```
tvallin, line 588: [P2] Accept matching URI-host forms on CONNECT

The null-scheme path still parses both `:authority` and `Host` with `UriAuthority`, but that type rejects URI-host forms accepted by the new CONNECT validator. Identical valid pairs such as `[Vf.foo-bar]:443`, `service+name:443`, or `service%2Dname:443` are therefore reset instead of reaching the intended 501 response. Please compare CONNECT values using the same URI-host grammar, or accept validated exact matches before normalization, and add wire-level tests with both fields present.

tvallin (dismissed):

http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
```diff
@@ -542,8 +574,11 @@ private static Http2Headers create(ServerRequestHeaders httpHeaders, PseudoHeade
     }
 
     private static boolean authoritiesMatch(String scheme, String authority, String host) {
+        if (scheme == null && authority.equals(host)) {
```
danielkec: [P2] Keep normalized-equivalent CONNECT authorities valid

The earlier thread covered byte-identical values, but when a CONNECT `Host` and `:authority` identify the same authority while differing only in host or percent-hex case, this exact-match shortcut falls through to `UriAuthority.create()`, which rejects these otherwise valid forms. Both cases below reset the stream with `PROTOCOL_ERROR` instead of reaching the intended 501 response:

```text
:authority: [Vf.foo-bar]:443
Host: [vf.foo-bar]:443

:authority: service%2Dname:443
Host: service%2dname:443
```

RFC 9113 section 8.3.1 requires the fields to be normalized for comparison; URI host and percent-encoding hex case are insensitive.

webserver/webserver/src/main/java/io/helidon/webserver/http1/Http1ConfigBlueprint.java
```diff
@@ -63,7 +63,8 @@ interface Http1ConfigBlueprint extends HttpConfig, ProtocolConfig {
     int maxHeadersSize();
 
     /**
-     * If set to false, any query and fragment is accepted (even containing illegal characters).
+     * Disables query and fragment syntax validation when set to {@code false}; intended only for closed systems receiving
```
danielkec, line 66: [P1] Generated configuration documentation is out of date

These description changes make the current merge regenerate `docs/config/io.helidon.webserver.http1.Http1Config.md` and `docs/config/io.helidon.webserver.http2.Http2Config.md`, but those generated diffs are absent. The live `docs` job prints the three resulting description changes and exits with `Generated configuration documentation is out of date`, so this head cannot pass the required docs check against current `main`; source PR #12329 carried the same generated artifacts.

http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
```diff
@@ -542,8 +575,11 @@ private static Http2Headers create(ServerRequestHeaders httpHeaders, PseudoHeade
     }
 
     private static boolean authoritiesMatch(String scheme, String authority, String host) {
+        if (scheme == null && Ascii.toLowerCase(authority).equals(Ascii.toLowerCase(host))) {
```
danielkec: [P2] Equivalent CONNECT authorities still reset the stream

The new case-folding path handles the earlier case-only examples, but it still treats a literal unreserved character and its percent-encoded form as different. Both wire orderings below fall through to `UriAuthority`, throw `PROTOCOL_ERROR`, and reset the stream instead of reaching the intended 501 response:

```text
:authority: service-name:443
Host: service%2Dname:443
```

The reverse ordering fails the same way. RFC 9113 section 8.3.1 requires normalization before considering the fields different, and RFC 3986 section 6.2.2.2 treats percent-encoded unreserved octets as equivalent to their literal form.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2ClientStream.java
```diff
@@ -731,17 +732,34 @@ void inboundHeaders(Http2Headers headers, boolean endOfStream) {
...
+                    headers.validateTrailers();
+                }
+            } catch (Http2Exception e) {
+                if (trailerBlock) {
+                    int discardedDataLength = buffer.failAndDiscard(e);
+                    try {
+                        failInboundLocked(e);
+                        incrementInboundWindowSizeLocked(discardedDataLength);
+                    } finally {
+                        trailers.completeExceptionally(e);
```
danielkec: [P1] Trailer failure callbacks can block sibling frame dispatch

`inboundHeaders` runs on the connection's single frame-dispatch thread, and this exceptional completion executes non-async dependents inline. For responses declaring trailers, `WebClientServiceResponse.trailers()` is derived synchronously from this future, so one blocking callback stops the reader from processing every sibling stream.

A minimal two-stream probe is:

```text
serviceResponse.trailers().whenComplete((_, _) -> blockOnLatch());
peer -> stream 1: END_STREAM trailers containing :path
peer -> stream 3: final response headers
```

Stream 3 timed out while the latch was held and returned `200 OK` immediately after release; the merge-base path processed stream 3 without entering the callback on the reader.

danielkec (approved):

PR #12390, opened 2026-08-29
Stabilize gRPC request backpressure test (27.x)

tomas-langer:
An unrelated Windows validation failure in #12384 exposed that the request-side gRPC backpressure test also echoed its 32 KiB request payloads, adding response-side flow-control pressure to the same call.

This keeps the large requests used to verify that production stops without server demand, but emits small responses after demand resumes. The test still requires all 128 requests to be produced and consumed and all 128 responses to arrive.


tomas-langer:
I have too many failures on this tests. Merging, as it only touches test classes and should stabilize the pipeline

PR #12478, opened 2026-09-01
Add support for Json Schema "default" basic metadata

### LINKED ISSUE #12476: Add support for Json Schema "default" basic metadata
Helidon Version: 27

## Enhancement Description

See https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01#section-9.2

tvallin:
### Description

Fixes #12476

### Documentation

Documentation is updated as part of this PR

### DIFF
--- a/json/schema/codegen/src/main/java/io/helidon/json/schema/codegen/SchemaInfo.java
+++ b/json/schema/codegen/src/main/java/io/helidon/json/schema/codegen/SchemaInfo.java
@@ -35,8 +35,10 @@ import io.helidon.common.types.TypeInfo;
 import io.helidon.common.types.TypeName;
 import io.helidon.common.types.TypeNames;
 import io.helidon.common.types.TypedElementInfo;
+import io.helidon.json.JsonException;
 import io.helidon.json.JsonObject;
 import io.helidon.json.JsonParser;
+import io.helidon.json.JsonValue;
 
 import static java.util.function.Predicate.not;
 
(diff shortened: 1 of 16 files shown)


json/schema/schema/src/main/java/io/helidon/json/schema/JsonSchema.java
```diff
@@ -84,6 +84,26 @@ private JsonSchema() {
...
+     * Default JSON value associated with a JSON schema.
+     * <p>
+     * The annotation value must contain JSON text. For example, a JSON string default can be declared as
+     * {@code @JsonSchema.Default("\"value\"")}. Any JSON value is supported, including {@code null}, arrays, and objects.
+     * A value that validates against the associated schema is recommended, but not required. The value is annotation metadata
+     * and is not used to fill in missing instance values.
+     */
+    @Target({ElementType.TYPE, ElementType.METHOD, ElementType.FIELD})
+    @Retention(RetentionPolicy.CLASS)
+    public @interface Default {
```
tomas-langer, line 95: `@JsonSchema.Default` accepts a Java `String` but interprets it as serialized JSON, so the source syntax does not identify the generated JSON type: `@JsonSchema.Default("0")` emits a numeric default, while a string default requires `@JsonSchema.Default("\"0\"")`. This makes common call sites visually ambiguous and forces a second encoding layer for strings; it also differs from the established `Option.Default*` annotations, which make the value type explicit. Because this is a new public annotation, releasing this contract locks in that ambiguity—changing the interpretation of existing source such as `@Default("0")` later would silently change generated schemas.

json/schema/codegen/src/main/java/io/helidon/json/schema/codegen/SchemaInfo.java
```diff
@@ -87,9 +90,27 @@ private static void processCommonAnnotations(JsonObject.Builder builder, Annotat
...
+        annotatedType.findAnnotation(SchemaTypes.JSON_SCHEMA_DEFAULT)
+                .flatMap(it -> it.stringValue())
+                .map(it -> parseDefaultValue(annotatedType, it))
+                .ifPresent(it -> builder.set("default", it));
         annotatedType.findAnnotation(SchemaTypes.JSON_SCHEMA_REQUIRED).ifPresent(it -> required.set(true));
     }
 
+    private static JsonValue parseDefaultValue(Annotated annotated, String value) {
+        JsonParser parser = JsonParser.create(value);
+        JsonValue result = parser.readJsonValue();
```
danielkec: `@JsonSchema.Default("01")` is accepted even though `01` is not valid JSON, and the generated provider silently emits `"default": 1`. This means malformed annotation input is normalized instead of rejected:

```java
@JsonSchema.Schema
@JsonSchema.Default("01")
class Example {
}
```

tvallin:
The PR has some functionality gaps, and requires additional design consideration.

PR #12489, opened 2026-09-03
Batch terminal HTTP/2 response writes

tomas-langer:
Improves HTTP/2 response write efficiency and hardens connection-writer failure handling.

- Batches terminal response HEADERS with immediately writable DATA into one transport write.
- Limits batching to single-frame headers and DATA payloads of at most 16 KiB; fragmented or oversized headers retain the prior per-frame write path.
- Preserves flow-control, pending WINDOW_UPDATE ordering, continuation-frame behavior, and end-stream callbacks when DATA must be split or deferred.
- Makes header and subsequent DATA transport/accounting failures terminal so the connection writer cannot be reused after its shared state might be uncertain.
- Keeps stream-local flow-control cancellation and timeout failures from terminating the shared connection writer.
- Adds focused regression coverage and targeted JMH cases for funded, exhausted, partial-window, and fragmented-header paths.

A short balanced, source-bound HttpArena `baseline-h2` comparison at 32 connections and 100 streams averaged 2.462 million requests/second for the bounded-batching candidate versus 1.231 million requests/second for its exact `main` parent (+99.98%), with no request failures. In an exact-base targeted JMH comparison of the final success paths, exhausted-window throughput improved by 16.38% single-threaded and 49.03% under contention, while partial-window throughput improved by 5.74% and 44.92%, respectively. The 99.9% throughput confidence intervals were disjoint in the candidate's favor; allocation decreased on the exhausted-window paths and increased by at most 40 bytes per operation on the partial-window paths.

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2ConnectionWriter.java
@@ -28,7 +28,6 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import io.helidon.common.buffers.BufferData;
-import io.helidon.common.buffers.CompositeBufferData;
 import io.helidon.common.buffers.DataWriter;
 import io.helidon.common.socket.SocketContext;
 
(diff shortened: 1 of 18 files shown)


danielkec (dismissed):
Reviewed current head fc9342a6.

danielkec (approved):
Reviewed current head 7d8155ab.

PR #12494, opened 2026-09-03
Upgrade from Prometheus simpleclient to Java client 1.x

### LINKED ISSUE #9258: 27.x Complete update to latest Micrometer release
## Environment Details
* Helidon Version: 5.x
* Helidon SE or Helidon MP
* JDK version:
* OS:

----------

## Problem Description
See also #9208 

Micrometer introduced some breaking changes between 1.11.3 and 1.13.3 that could be mitigated (temporarily) by using a bridge library. 

PR #9225 updated the Micrometer release to 1.13.3 and used the bridge library. (See https://github.com/micrometer-metrics/micrometer/wiki/1.13-Migration-Guide)

This issue represents the need to stop using the bridge library and make whatever changes in Helidon are needed, possibly including breaking changes, to adopt the latest Micrometer release.

### LINKED ISSUE #12487: 27 - Update Prometheus client
Helidon Version: 27

## Task Description

Prometheus's legacy client (`simpleclient`) is no longer under active development and has been supplanted by the Prometheus Java Client (currently at release 1.7.0).

This issue is to track the need to update the client.

Note that the newer client does not support an option exposed in Helidon 4.x. #12496 marks that option as deprecated in 4.x and 27 will continue to accept the option but will ignore it and warn when it is set.

tjquinno:
## Release Note
____
Helidon now uses Prometheus Java Client 1.7.0 instead of the deprecated `simpleclient` integration. The new client’s metric-name normalization applies by default; applications requiring naming compatible with earlier Helidon releases can set `metrics.publishers.prometheus.naming-convention.non-letter-prefix` to `m_`. The legacy `metrics.prometheus.histogramFlavor` setting remains accepted but is ignored with a warning.
____

## Summary

Fixes #12487 by replacing the deprecated Prometheus `simpleclient` integration with Prometheus Java Client 1.7.0 through Micrometer’s `micrometer-registry-prometheus` module.

Resolves #9258.

The new Prometheus Java client uses updated normalization rules for metric names that begin with non-letter characters and strips reserved suffixes such as _total, _created, _bucket, and _info before applying type-specific suffixes. With this PR, Helidon's default is to adopt the new Prometheus Java client rules. Most applications using "well-behaved" meter names will see no difference in the Prometheus output. Applications that need output naming compatible with earlier Helidon releases can opt into legacy normalization using the new `naming-convention.non-letter-prefix` setting.

## Configuration

Adds `PrometheusNamingConventionConfig` with:

- `timer-suffix`, defaulting to `""`
- Optional `non-letter-prefix`

For example, the following selects the earlier Helidon/simpleclient naming behavior:

```yaml
metrics:
  publishers:
    prometheus:
      naming-convention:
        non-letter-prefix: "m_"
```

Equivalent programmatic configuration:

```java
PrometheusPublisher.builder()
        .namingConvention(builder -> builder.nonLetterPrefix("m_"))
        .build();
```

Valid non-letter prefixes must match `[A-Za-z][A-Za-z0-9_]*`.

If `non-letter-prefix` is absent, Helidon uses the new client’s naming convention. Setting it enables legacy-compatible normalization and also preserves user-supplied reserved suffixes such as `_total`, `_created`, `_bucket`, and `_info`.

The existing publisher `prefix` remains unchanged and controls Micrometer property lookup, not exported metric-name prefixing. The documentation also notes that the legacy `prometheus.histogramFlavor` property is no longer supported, although Helidon accepts it if present and logs a warning that it is ignored.

## Implementation

- Replaces `micrometer-registry-prometheus-simpleclient` and Prometheus simpleclient dependencies with the new Prometheus client model, registry, tracer, and exposition modules.
- Updates Maven dependency management, JPMS declarations, Javadocs, third-party licenses, and dependency-check metadata.
- Centralizes Prometheus registry construction in `PrometheusPublisher`.
- Adds a Helidon-owned legacy naming convention compatible with simpleclient behavior for supported Helidon meters.
- Preserves legacy counter, gauge, timer, and distribution-summary output suffixes.
- Prevents the new registry from interpreting legacy gauges ending in `.info` as Prometheus info metrics.
- Rejects primary and generated metric-family name collisions during registration so they cannot cause scrape failures.
- Updates selective scraping to match original Helidon meter names while deriving selectable families only from meters accepted by the Prometheus child registry, including separate `_max` families.
- Uses the new Prometheus text and OpenMetrics writers, with created-timestamp output disabled by default.
- Retains the existing public behavior of `normalizeNameToPrometheus(String)`.

## Exemplars

- Migrates from the simpleclient exemplar APIs to `io.prometheus.metrics.tracer.common.SpanContext`.
- Adds sampling-state reporting to Helidon `SpanContext`.
- Reports the actual OpenTelemetry sampling decision.
- Uses the propagated Helidon context consistently for trace ID, span ID, sampling state, and exemplar marking.

## Compatibility scope

Legacy-compatible naming covers meter types created by Helidon:

- Counters and functional counters
- Gauges
- Timers
- Distribution summaries

Long-task timers and arbitrary custom Micrometer meters are outside the compatibility scope.

## Testing

Added or updated tests covering:

- Default new-client normalization
- Legacy-compatible metric and tag naming
- Reserved-suffix preservation
- `.info` gauge behavior and naming collisions
- Registration-time collisions involving generated metric-family names
- Rejected normalized aliases remaining unavailable to selective scrape
- Counter, gauge, timer, and distribution-summary output
- Prometheus text and OpenMetrics formats
- Base units, descriptions, tags, histograms, and scopes
- Selective scraping and separate `_max` families
- Config and programmatic configuration
- Valid and invalid custom prefixes
- Custom timer suffixes
- Exemplar output, sampling state, and propagated-context marking
- Absence of simpleclient artifacts and use of Prometheus Java Client 1.7.0

Focused metrics, tracing, and exemplar tests pass, together with compilation, dependency convergence, Checkstyle, copyright validation, Javadocs, and whitespace checks.

### Documentation
Updated

### DIFF
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
@@ -125,7 +125,7 @@ class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry 
     protected LongTaskTimer newLongTaskTimer(Meter.Id id,
                                              DistributionStatisticConfig distributionStatisticConfig) {
         return register(id,
-                        distributionNames(id, distributionStatisticConfig),
+                        longTaskTimerNames(id, distributionStatisticConfig),
                         () -> super.newLongTaskTimer(id, distributionStatisticConfig));
     }
 
@@ -170,6 +170,17 @@ class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry 
                       conventionName + CREATED_SUFFIX);
     }
 
+    private Set<String> longTaskTimerNames(Meter.Id id,
+                                           DistributionStatisticConfig distributionStatisticConfig) {
+        Set<String> result = new LinkedHashSet<>(distributionNames(id, distributionStatisticConfig));
+        if (distributionStatisticConfig.isPublishingHistogram()) {
+            String conventionName = expositionName(id);
+            result.add(conventionName + "_gcount");
+            result.add(conventionName + "_gsum");
+        }
+        return Set.copyOf(result);
+    }
+
     private String expositionName(Meter.Id id) {
         return PrometheusNameSupport.expositionName(id, config().namingConvention());
     }
@@ -202,7 +213,7 @@ class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry 
                                                                + "' is produced by both " + claim.owner + " and " + owner);
                 }
             }
-            names.forEach(name -> claims.compute(name, (ignored, claim) -> claim == null
+            names.forEach(name -> claims.compute(name, (_, claim) -> claim == null
                     ? new Claim(owner)
                     : claim.increment()));
             return new Reservation(names);
@@ -233,7 +244,7 @@ class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry 
     }
 
     private void releaseLocked(Reservation reservation) {
-        reservation.names.forEach(name -> claims.computeIfPresent(name, (ignored, claim) -> claim.decrement()));
+        reservation.names.forEach(name -> claims.computeIfPresent(name, (_, claim) -> claim.decrement()));
     }
 
     private static final class Claim {
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
@@ -1,0 +1,225 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.metrics.providers.micrometer;
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
+import java.util.function.ToDoubleFunction;
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.Timer;
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
+import io.micrometer.core.instrument.distribution.pause.PauseDetector;
+import io.micrometer.prometheusmetrics.PrometheusConfig;
+import io.micrometer.prometheusmetrics.PrometheusMeterRegistry;
+import io.prometheus.metrics.model.registry.PrometheusRegistry;
+import io.prometheus.metrics.tracer.common.SpanContext;
+
+/**
+ * Prometheus meter registry which detects collisions involving generated Prometheus sample names.
+ *
+ * <p>The Prometheus client validates collector names at registration, but the Micrometer collector used by the
+ * Micrometer Prometheus registry does not describe the secondary names it generates. This registry accounts for those
+ * names for the meter types Helidon creates.</p>
+ */
+class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry {
+
+    private static final String CREATED_SUFFIX = "_created";
+    private static final String TOTAL_SUFFIX = "_total";
+
+    private final Lock reservationLock = new ReentrantLock();
+    private final Map<String, Claim> claims = new HashMap<>();
+    private final Map<Meter.Id, Reservation> reservations = new HashMap<>();
+
+    CollisionDetectingPrometheusMeterRegistry(PrometheusConfig config) {
+        super(config);
+        config().onMeterRemoved(meter -> release(meter.getId()));
+    }
+
+    CollisionDetectingPrometheusMeterRegistry(PrometheusConfig config,
+                                               PrometheusRegistry registry,
+                                               Clock clock,
+                                               SpanContext spanContext) {
+        super(config, registry, clock, spanContext);
+        config().onMeterRemoved(meter -> release(meter.getId()));
+    }
+
+    @Override
+    public Counter newCounter(Meter.Id id) {
+        return register(id, counterNames(id), () -> super.newCounter(id));
+    }
+
+    @Override
+    public DistributionSummary newDistributionSummary(Meter.Id id,
+                                                      DistributionStatisticConfig distributionStatisticConfig,
+                                                      double scale) {
+        return register(id,
+                        distributionNames(id, distributionStatisticConfig),
+                        () -> super.newDistributionSummary(id, distributionStatisticConfig, scale));
+    }
+
+    @Override
+    protected Timer newTimer(Meter.Id id,
+                             DistributionStatisticConfig distributionStatisticConfig,
+                             PauseDetector pauseDetector) {
+        return register(id,
+                        distributionNames(id, distributionStatisticConfig),
+                        () -> super.newTimer(id, distributionStatisticConfig, pauseDetector));
+    }
+
+    @Override
+    protected <T> Gauge newGauge(Meter.Id id, T obj, ToDoubleFunction<T> valueFunction) {
+        return register(id, gaugeNames(id), () -> super.newGauge(id, obj, valueFunction));
+    }
+
+    @Override
+    protected <T> FunctionCounter newFunctionCounter(Meter.Id id, T obj, ToDoubleFunction<T> countFunction) {
+        return register(id, counterNames(id), () -> super.newFunctionCounter(id, obj, countFunction));
+    }
+
+    private static String owner(Meter.Id id) {
+        return "'" + LegacyPrometheusMeterFilter.originalGaugeName(id.getName()) + "' (type=" + id.getType()
+                + ", baseUnit=" + Objects.toString(id.getBaseUnit(), "<none>") + ")";
+    }
+
+    private Set<String> counterNames(Meter.Id id) {
+        String conventionName = getConventionName(id);
+        String counterName = conventionName.endsWith(TOTAL_SUFFIX) ? conventionName : conventionName + TOTAL_SUFFIX;
+        String baseName = counterName.substring(0, counterName.length() - TOTAL_SUFFIX.length());
+        return Set.of(baseName, counterName, baseName + CREATED_SUFFIX);
+    }
+
+    private Set<String> gaugeNames(Meter.Id id) {
+        String conventionName = getConventionName(id);
+        return id.getName().endsWith(".info")
+                ? Set.of(conventionName, conventionName + "_info")
+                : Set.of(conventionName);
+    }
+
+    private Set<String> distributionNames(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig) {
+        String conventionName = getConventionName(id);
+        Set<String> result = new LinkedHashSet<>();
+        result.add(conventionName);
+        result.add(conventionName + "_count");
+        result.add(conventionName + "_sum");
+        result.add(conventionName + CREATED_SUFFIX);
+        if (distributionStatisticConfig.isPublishingHistogram()) {
+            result.add(conventionName + "_bucket");
+        }
+        result.add(conventionName + "_max");
+        return Set.copyOf(result);
+    }
+
+    private <M extends Meter> M register(Meter.Id id, Set<String> names, Supplier<M> registration) {
+        Reservation reservation = reserve(id, names);
+        try {
+            M meter = registration.get();
+            reservationLock.lock();
+            try {
+                reservations.put(id, reservation);
+            } finally {
+                reservationLock.unlock();
+            }
+            return meter;
+        } catch (RuntimeException | Error e) {
+            release(reservation);
+            throw e;
+        }
+    }
+
+    private Reservation reserve(Meter.Id id, Set<String> names) {
+        String owner = owner(id);
+        reservationLock.lock();
+        try {
+            for (String name : names) {
+                Claim claim = claims.get(name);
+                if (claim != null && !claim.owner.equals(owner)) {
+                    throw new IllegalArgumentException("Prometheus metric name collision: '" + name
+                                                               + "' is produced by both " + claim.owner + " and " + owner);
+                }
+            }
+            names.forEach(name -> claims.compute(name, (ignored, claim) -> claim == null
+                    ? new Claim(owner)
+                    : claim.increment()));
+            return new Reservation(names);
+        } finally {
+            reservationLock.unlock();
+        }
+    }
+
+    private void release(Meter.Id id) {
+        reservationLock.lock();
+        try {
+            Reservation reservation = reservations.remove(id);
+            if (reservation != null) {
+                releaseLocked(reservation);
+            }
+        } finally {
+            reservationLock.unlock();
+        }
+    }
+
+    private void release(Reservation reservation) {
+        reservationLock.lock();
+        try {
+            releaseLocked(reservation);
+        } finally {
+            reservationLock.unlock();
+        }
+    }
+
+    private void releaseLocked(Reservation reservation) {
+        reservation.names.forEach(name -> claims.computeIfPresent(name, (ignored, claim) -> claim.decrement()));
+    }
+
+    private static final class Claim {
+        private final String owner;
+        private int count = 1;
+
+        private Claim(String owner) {
+            this.owner = owner;
+        }
+
+        private Claim increment() {
+            count++;
+            return this;
+        }
+
+        private Claim decrement() {
+            return --count == 0 ? null : this;
+        }
+    }
+
+    private static final class Reservation {
+        private final Set<String> names;
+
+        private Reservation(Set<String> names) {
+            this.names = names;
+        }
+    }
+}
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
@@ -21,16 +21,20 @@ import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Supplier;
 import java.util.function.ToDoubleFunction;
+import java.util.function.ToLongFunction;
 
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.FunctionCounter;
+import io.micrometer.core.instrument.FunctionTimer;
 import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.LongTaskTimer;
 import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.Timer;
 import io.micrometer.core.instrument.distribution.DistributionStatisticConfig;
@@ -102,27 +106,50 @@ class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry 
         return register(id, counterNames(id), () -> super.newFunctionCounter(id, obj, countFunction));
     }
 
+    @Override
+    protected <T> FunctionTimer newFunctionTimer(Meter.Id id,
+                                                 T obj,
+                                                 ToLongFunction<T> countFunction,
+                                                 ToDoubleFunction<T> totalTimeFunction,
+                                                 TimeUnit totalTimeFunctionUnit) {
+        return register(id,
+                        functionTimerNames(id),
+                        () -> super.newFunctionTimer(id,
+                                                     obj,
+                                                     countFunction,
+                                                     totalTimeFunction,
+                                                     totalTimeFunctionUnit));
+    }
+
+    @Override
+    protected LongTaskTimer newLongTaskTimer(Meter.Id id,
+                                             DistributionStatisticConfig distributionStatisticConfig) {
+        return register(id,
+                        distributionNames(id, distributionStatisticConfig),
+                        () -> super.newLongTaskTimer(id, distributionStatisticConfig));
+    }
+
     private static String owner(Meter.Id id) {
         return "'" + LegacyPrometheusMeterFilter.originalGaugeName(id.getName()) + "' (type=" + id.getType()
                 + ", baseUnit=" + Objects.toString(id.getBaseUnit(), "<none>") + ")";
     }
 
     private Set<String> counterNames(Meter.Id id) {
-        String conventionName = getConventionName(id);
+        String conventionName = expositionName(id);
         String counterName = conventionName.endsWith(TOTAL_SUFFIX) ? conventionName : conventionName + TOTAL_SUFFIX;
         String baseName = counterName.substring(0, counterName.length() - TOTAL_SUFFIX.length());
         return Set.of(baseName, counterName, baseName + CREATED_SUFFIX);
     }
 
     private Set<String> gaugeNames(Meter.Id id) {
-        String conventionName = getConventionName(id);
+        String conventionName = expositionName(id);
         return id.getName().endsWith(".info")
                 ? Set.of(conventionName, conventionName + "_info")
                 : Set.of(conventionName);
     }
 
     private Set<String> distributionNames(Meter.Id id, DistributionStatisticConfig distributionStatisticConfig) {
-        String conventionName = getConventionName(id);
+        String conventionName = expositionName(id);
         Set<String> result = new LinkedHashSet<>();
         result.add(conventionName);
         result.add(conventionName + "_count");
@@ -133,6 +160,18 @@ class CollisionDetectingPrometheusMeterRegistry extends PrometheusMeterRegistry 
         }
         result.add(conventionName + "_max");
         return Set.copyOf(result);
+    }
+
+    private Set<String> functionTimerNames(Meter.Id id) {
+        String conventionName = expositionName(id);
+        return Set.of(conventionName,
+                      conventionName + "_count",
+                      conventionName + "_sum",
+                      conventionName + CREATED_SUFFIX);
+    }
+
+    private String expositionName(Meter.Id id) {
+        return PrometheusNameSupport.expositionName(id, config().namingConvention());
     }
 
     private <M extends Meter> M register(Meter.Id id, Set<String> names, Supplier<M> registration) {
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/LegacyPrometheusMeterFilter.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/LegacyPrometheusMeterFilter.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.metrics.providers.micrometer;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.config.MeterFilter;
+
+/**
+ * Prevents the new Prometheus registry from treating a legacy gauge name ending in {@code .info} as an info metric.
+ * The marker also preserves the original gauge name through Micrometer's pre-filter ID mapping so the Prometheus
+ * registry can detect distinct names which normalize to the same name.
+ */
+class LegacyPrometheusMeterFilter implements MeterFilter {
+    private static final String GAUGE_MARKER = "\uE000helidon-legacy-gauge:";
+
+    @Override
+    public Meter.Id map(Meter.Id id) {
+        if (id.getType() != Meter.Type.GAUGE) {
+            return id;
+        }
+        return id.withName(GAUGE_MARKER + id.getName() + GAUGE_MARKER);
+    }
+
+    static String originalGaugeName(String name) {
+        return name.startsWith(GAUGE_MARKER) && name.endsWith(GAUGE_MARKER)
+                ? name.substring(GAUGE_MARKER.length(), name.length() - GAUGE_MARKER.length())
+                : name;
+    }
+}
--- a/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/LegacyPrometheusMeterFilter.java
+++ b/metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/LegacyPrometheusMeterFilter.java
@@ -27,17 +27,17 @@ import io.micrometer.core.instrument.config.MeterFilter;
 class LegacyPrometheusMeterFilter implements MeterFilter {
     private static final String GAUGE_MARKER = "\uE000helidon-legacy-gauge:";
 
+    static String originalGaugeName(String name) {
+        return name.startsWith(GAUGE_MARKER) && name.endsWith(GAUGE_MARKER)
+                ? name.substring(GAUGE_MARKER.length(), name.length() - GAUGE_MARKER.length())
+                : name;
+    }
+
     @Override
     public Meter.Id map(Meter.Id id) {
         if (id.getType() != Meter.Type.GAUGE) {
             return id;
         }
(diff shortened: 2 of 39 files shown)


metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
```diff
@@ -105,19 +112,24 @@ public String type() {
...
-            spanContextSupplierProvider instanceof NoOpSpanContextSupplierProvider
-                ? new PrometheusMeterRegistry(prometheusConfig(lookupFunction))
-                : new PrometheusMeterRegistry(prometheusConfig(lookupFunction),
-                                         new CollectorRegistry(),
-                                         io.micrometer.core.instrument.Clock.SYSTEM,
-                                         new DefaultExemplarSampler(spanContextSupplierProvider.get()));
+        return (lookupFunction, spanContextSupplierProvider) -> {
+            PrometheusConfig prometheusConfig = prometheusConfig(lookupFunction);
+            warnIfLegacyHistogramFlavorConfigured(prometheusConfig);
+            PrometheusMeterRegistry registry = spanContextSupplierProvider instanceof NoOpSpanContextSupplierProvider
```
tomas-langer, line 124: AI_Findings: **[P1] Secondary metric-family collisions now fail during scrape.** `io.helidon.metrics.providers.micrometer.PrometheusPublisher:124` — A timer named `request` followed by a gauge named `request_seconds_max` was previously rejected during gauge registration. The new registry accepts both, then the metrics scrape throws `DuplicateLabelsException`, making `/observe/metrics` fail.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
```diff
@@ -173,65 +159,45 @@ public Optional<Object> formatMetadata() {
...
-                    .forEach(units -> allSuffixesForMeterName
-                            .forEach(suffix -> result.add(normalizedMeterName + units + suffix)));
+
+            io.micrometer.core.instrument.Meter micrometerMeter =
+                    meter.unwrap(io.micrometer.core.instrument.Meter.class);
+            io.micrometer.core.instrument.Meter.Id meterId = micrometerMeter.getId();
+            String conventionName = prometheusMeterRegistry.config()
+                    .namingConvention()
+                    .name(meterId.getName(), meterId.getType(), meterId.getBaseUnit());
+            result.add(conventionName);
```
tomas-langer, line 196: AI_Findings: **[P2] Selective scraping can return another meter's data.** `io.helidon.metrics.providers.micrometer.MicrometerPrometheusFormatter:180` — After `jobs_created` normalizes to the existing `jobs` family and its Prometheus collector is rejected, the Helidon composite registry can retain the logical meter. Selecting `name=jobs_created` then recomputes the family name as `jobs` and returns the accepted meter's data instead of an empty result.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
```diff
@@ -173,65 +159,45 @@ public Optional<Object> formatMetadata() {
...
      */
     Set<String> meterNamesOfInterest(PrometheusMeterRegistry prometheusMeterRegistry,
                                      Set<String> scopes,
                                      Set<String> names) {
 
         Set<String> result = new HashSet<>();
 
-        for (Meter meter : prometheusMeterRegistry.getMeters()) {
-            String meterName = meter.getId().getName();
+        for (io.helidon.metrics.api.Meter meter : meterRegistry.meters()) {
```
tomas-langer, line 180: AI_Findings: **[P2] No usable performance evidence covers the new metrics hot paths.** `io.helidon.metrics.providers.micrometer.PrometheusPublisher:124`, `io.helidon.metrics.providers.micrometer.MicrometerPrometheusFormatter:147` — The migration changes per-observation work and full/selective scrape serialization. The only branch-local performance probe is disabled and explicitly ad hoc, so the change has no reproducible base-versus-head evidence bounding recording overhead, allocation, throughput, or scrape latency at realistic cardinalities.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
```diff
@@ -152,4 +164,22 @@ public Duration step() {
...
+    }
+
+    private void configureNaming(PrometheusMeterRegistry registry) {
+        PrometheusNamingConventionConfig namingConfig = config.namingConvention()
+                .orElseGet(() -> PrometheusNamingConventionConfig.builder().build());
+
+        namingConfig.nonLetterPrefix().ifPresentOrElse(prefix -> {
+            registry.config().meterFilter(new LegacyPrometheusMeterFilter());
+            registry.config().namingConvention(new LegacyPrometheusNamingConvention(namingConfig.timerSuffix(), prefix));
+        }, () -> registry.config().namingConvention(new PrometheusNamingConvention(namingConfig.timerSuffix())));
```
tomas-langer, line 183: AI_Findings: **[P2] Shipped migration docs omit the default Prometheus naming break.** `docs/modules/metrics/metrics.md:710` — The PR description explains the new default and legacy opt-in, but the authored migration guide does not. Existing dashboards, alerts, recording rules, and clients using legacy metric or tag names can silently stop matching after upgrade.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/spi/SpanContextSupplierProvider.java
```diff
@@ -15,18 +15,18 @@
...
 public interface SpanContextSupplierProvider {
 
     /**
-     * Returns a {@link io.prometheus.client.exemplars.tracer.common.SpanContextSupplier} for supporting exemplars.
+     * Returns a {@link io.prometheus.metrics.tracer.common.SpanContext} for supporting exemplars.
      *
      * @return a span context supplier
      */
-    SpanContextSupplier get();
+    SpanContext get();
```
tomas-langer, line 31: AI_Findings: **[P2] The exported Prometheus SPI migration is undocumented.** `io.helidon.metrics.providers.micrometer.spi.SpanContextSupplierProvider:31` — `get()` changes from the simpleclient `SpanContextSupplier` type to the new client `SpanContext`, and `PrometheusPublisher.prometheusRegistry()` exposes a different Micrometer registry package. Existing external providers and callers are source/binary incompatible, but the shipped migration docs do not describe the changed types or impact.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
```diff
@@ -26,21 +27,27 @@
...
+    private static final String HISTOGRAM_FLAVOR_PROPERTY = "histogramFlavor";
+
     private final PrometheusPublisherConfig config;
 
     private PrometheusPublisher(PrometheusPublisherConfig config) {
-        this.config = config;
+        this.config = Objects.requireNonNull(config);
+        config.namingConvention()
+                .flatMap(PrometheusNamingConventionConfig::nonLetterPrefix)
+                .ifPresent(LegacyPrometheusNamingConvention::validatePrefix);
```
tomas-langer, line 50: AI_Findings: **[P2] The new prefix option omits its validation contract.** `docs/config/io.helidon.metrics.providers.micrometer.PrometheusNamingConventionConfig.md:26` — The implementation accepts only `[A-Za-z][A-Za-z0-9_]*`, but neither the blueprint Javadoc nor generated config reference documents that constraint or the startup failure for invalid values. Plausible values such as `_` or `1_` therefore fail unexpectedly.

tracing/tracing/src/main/java/io/helidon/tracing/SpanContext.java
```diff
@@ -33,6 +33,17 @@ public interface SpanContext {
...
     String spanId();
 
+    /**
+     * Whether the associated span is sampled.
+     * Implementations should override this method to report the tracer's sampling decision. The default returns
+     * {@code true} for compatibility with existing implementations.
+     *
+     * @return whether the span is sampled
+     */
+    default boolean sampled() {
```
tomas-langer, line 43: AI_Findings: **[P2] Exemplar docs omit sampled-span eligibility.** `docs/modules/metrics/exemplar.md:77` — The candidate now suppresses exemplars when the current span is not sampled, while the documentation still describes exemplar support as automatic after adding the dependencies. `docs/includes/tracing/common-callbacks.md:83` also omits the new public `SpanContext.sampled()` operation, leaving the changed behavior and API undiscoverable.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/LegacyPrometheusMeterFilter.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    public Meter.Id map(Meter.Id id) {
+        if (id.getType() != Meter.Type.GAUGE) {
+            return id;
+        }
+        return id.withName(GAUGE_MARKER + id.getName() + GAUGE_MARKER);
+    }
+
+    static String originalGaugeName(String name) {
```
tomas-langer, line 38: AI_Findings: **[P3] New helper methods violate Helidon member ordering.** `io.helidon.metrics.providers.micrometer.LegacyPrometheusMeterFilter:38` and `io.helidon.metrics.providers.micrometer.LegacyPrometheusNamingConvention:73` declare non-private static methods after instance methods, contrary to `DEV-GUIDELINES.md` Rule 2.1 (`#rule-2-1`).

metrics/providers/micrometer/src/test/java/io/helidon/metrics/providers/micrometer/TestPrometheusNaming.java
```diff
@@ -0,0 +1,284 @@
...
+
+    @Test
+    @SuppressWarnings("unchecked")
+    void testLegacySupportedMeterOutputInBothFormats() {
+        PrometheusPublisher publisher = PrometheusPublisher.builder()
+                .namingConvention(builder -> builder.nonLetterPrefix("m_"))
+                .build();
+        PrometheusMeterRegistry registry = registry(publisher);
+        try {
+            io.micrometer.core.instrument.Counter.builder("jobs")
```
tomas-langer, line 187: AI_Findings: **[P3] New test code violates imported-simple-name style.** `io.helidon.metrics.providers.micrometer.TestPrometheusNaming:187` uses the fully qualified `io.micrometer.core.instrument.Counter` despite having no simple-name conflict, contrary to `DEV-GUIDELINES.md` Rule 3.1 (`#rule-3-1`).

metrics/providers/micrometer/src/test/java/io/helidon/metrics/providers/micrometer/TestPrometheusNaming.java
```diff
@@ -0,0 +1,284 @@
...
+        PrometheusMeterRegistry registry = PrometheusPublisher.create()
+                .prometheusRegistry()
+                .apply(_ -> null, provider);
+        try {
+            registry.counter("exemplar.counter").increment();
+
+            assertThat(scrape(registry),
+                       allOf(containsString("trace_id=\"0123456789abcdef0123456789abcdef\""),
+                             containsString("span_id=\"0123456789abcdef\"")));
+            assertThat("Current span was marked as an exemplar", markedAsExemplar.get());
```
tomas-langer, line 271: AI_Findings: **[P3] A new test uses Hamcrest's boolean overload without a matcher.** `io.helidon.metrics.providers.micrometer.TestPrometheusNaming:271` calls `assertThat` with a reason and a boolean expression rather than an actual value and matcher, weakening failure diagnostics and violating `DEV-GUIDELINES.md` Rule 10.5.1 (`#rule-10-5-1`).

tomas-langer:
Tim, the PR already covers the dependency, client API, configuration, and exemplar migration. The remaining work on top of the current changes is:

1. Handle secondary metric-family collisions during registration so a collision cannot make the metrics endpoint fail during scrape.
2. Derive selective-scrape families only from meters actually accepted by the Prometheus registry, so a rejected normalized alias cannot return another meter's data.
3. Add authored migration documentation covering the default naming change and legacy opt-in, the public Prometheus API/SPI type changes, the `non-letter-prefix` validation contract, and sampled-span eligibility for exemplars.
4. Apply the small source- and test-style corrections noted in the inline comments.
5. Add `Resolves #9258` to the PR body so the original issue closes together with #12487.

The current dependency/client conversion and associated metadata work do not need to be repeated.

tomas-langer:
1. Handle secondary metric-family collisions during registration so a collision cannot make the metrics endpoint fail during scrape.
2. Derive selective-scrape families only from meters actually accepted by the Prometheus registry, so a rejected normalized alias cannot return another meter's data.
3. Add authored migration documentation covering the default naming change and legacy opt-in, the public Prometheus API/SPI type changes, the non-letter-prefix validation contract, and
      sampled-span eligibility for exemplars.
4. Add Resolves #9258 to the PR body so the original issue closes together with #12487.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
tjquinno: Addressed in 16ccb7f824. PrometheusPublisher now constructs CollisionDetectingPrometheusMeterRegistry, which reserves both the base family and every generated exposition name during meter registration. Timers and distribution summaries reserve _count, _sum, _max, optional _bucket, and _created; counters reserve the base, _total, and _created; gauges reserve their emitted name. The reported request_seconds_max name is therefore claimed by timer request, so the gauge is rejected at registration rather than causing scrape failure. The registry also releases claims on removal. TestPrometheusNaming.testGeneratedNameCollisionsAreRejectedDuringRegistration covers registration-time secondary-name collisions, successful scrape afterward, reverse registration order, and claim release.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
tjquinno: Addressed in 16ccb7f824. MicrometerPrometheusFormatter.meterNamesOfInterest now iterates prometheusMeterRegistry.getMeters(), so it derives selectable families only from meters accepted by the Prometheus child registry rather than from logical meters in the Helidon composite. PrometheusPublisher also enables throwExceptionOnRegistrationFailure, preventing a rejected child registration from leaving the alias in the composite. TestPrometheusFormatting.testRejectedPrometheusMeterIsNotAvailableForSelectiveScrape verifies registration rejection, absence from the Helidon registry, and an empty selective result for the rejected name.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
tjquinno: Assessed the two paths separately against the Micrometer 1.17.1 old and new implementations. The latest Helidon collision and naming code adds no per-observation work: its locking, name derivation, and reservation maps run only when a distinct meter is registered or removed. Without exemplars, counter, timer, and distribution-summary recording still use the same adders, TimeWindowMax, and optional Micrometer histogram operations. With exemplars, the new client rate-limits sampling; its usual high-throughput path is an AtomicBoolean check, whereas the old sampler checked the clock and tracing state on each observation. I therefore do not expect an appreciable recording regression.

Scraping is the credible risk. The 1.x integration builds typed Labels, data-point snapshots, metric snapshots, sorted MetricSnapshots collections, and a duplicate-merge view before writing. Histograms also convert cumulative counts to non-cumulative counts which the writer accumulates again. The latest Helidon selective-name pass improves from a possible repeated registry scan to one linear pass, but Micrometer 1.17.1 does not publish descriptors for its MultiCollector, so the Prometheus 1.x registry can collect unrelated families before discarding them for a narrow selection. Full-scrape allocation and narrow selective-scrape latency could therefore be appreciably higher at high cardinality even though ordinary recording should not be.

No base-versus-head benchmark evidence has been added, so I am not claiming this concern is eliminated. Proportionate evidence should focus on full and narrow selective scrapes at realistic series, label, and histogram cardinalities, measuring both time and allocation; a small recording benchmark would serve mainly to confirm the source-level parity.

As scraping is on-demand as requested from a client, and because scraping would normally be requested infrequently, the performance risk, while real, does not seem critical.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
tjquinno: Addressed in 16ccb7f824 in docs/modules/metrics/metrics.md under Metrics Configuration Migration. The shipped guide now states that Helidon 27 uses Prometheus Java Client 1.7.0, explains the new default leading-character and reserved-suffix normalization, and shows the naming-convention.non-letter-prefix: m_ configuration for retaining earlier Helidon output. It also explains that setting non-letter-prefix selects legacy normalization as a whole, including reserved-suffix preservation, so applications can preserve the names used by existing dashboards, alerts, rules, and clients.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/spi/SpanContextSupplierProvider.java
tjquinno: Addressed in 16ccb7f824 in the shipped Metrics Configuration Migration section. It calls out code that accesses Prometheus-specific Helidon APIs or implements the exemplar SPI, then documents both source-incompatible changes: PrometheusPublisher.prometheusRegistry() now exposes io.micrometer.prometheusmetrics.PrometheusMeterRegistry instead of io.micrometer.prometheus.PrometheusMeterRegistry, and SpanContextSupplierProvider.get() now returns io.prometheus.metrics.tracer.common.SpanContext instead of the simpleclient SpanContextSupplier. The text also describes the new trace ID, span ID, sampled-state, and exemplar-marking responsibilities.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/PrometheusPublisher.java
tjquinno: Addressed in 16ccb7f824. PrometheusNamingConventionConfigBlueprint.nonLetterPrefix now documents that the prefix must be non-empty, must match [A-Za-z][A-Za-z0-9_]*, and causes publisher construction to fail when invalid. The validation contract is kept in the first Javadoc sentence so docs generation includes it. The regenerated docs/config/io.helidon.metrics.providers.micrometer.PrometheusNamingConventionConfig.md now contains the same regex and failure behavior, and the authored migration guide repeats the allowed pattern.

tracing/tracing/src/main/java/io/helidon/tracing/SpanContext.java
tjquinno: Addressed in 16ccb7f824. docs/modules/metrics/exemplar.md now states that Helidon supplies exemplars only from the current span when that span is sampled. docs/includes/tracing/common-callbacks.md now lists SpanContext.sampled() and describes it as reporting whether the associated span is sampled. The exemplar tests also cover sampled-state reporting and marking the current propagated span as an exemplar.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/LegacyPrometheusMeterFilter.java
tjquinno: Addressed in 16ccb7f824. LegacyPrometheusMeterFilter.originalGaugeName and LegacyPrometheusNamingConvention.validatePrefix are now declared in the non-private static-method section before the instance overrides and private helpers, matching DEV-GUIDELINES.md Rule 2.1.

metrics/providers/micrometer/src/test/java/io/helidon/metrics/providers/micrometer/TestPrometheusNaming.java
tjquinno: Addressed in 16ccb7f824. TestPrometheusNaming now imports io.micrometer.core.instrument.Counter and uses the simple name at the call site, matching DEV-GUIDELINES.md Rule 3.1.

metrics/providers/micrometer/src/test/java/io/helidon/metrics/providers/micrometer/TestPrometheusNaming.java
tjquinno: Addressed in 16ccb7f824. The exemplar-marking assertion now passes the actual boolean value to assertThat with the Hamcrest is(true) matcher, providing matcher diagnostics and satisfying DEV-GUIDELINES.md Rule 10.5.1.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+
+    private Set<String> counterNames(Meter.Id id) {
+        String conventionName = getConventionName(id);
+        String counterName = conventionName.endsWith(TOTAL_SUFFIX) ? conventionName : conventionName + TOTAL_SUFFIX;
+        String baseName = counterName.substring(0, counterName.length() - TOTAL_SUFFIX.length());
+        return Set.of(baseName, counterName, baseName + CREATED_SUFFIX);
+    }
+
+    private Set<String> gaugeNames(Meter.Id id) {
+        String conventionName = getConventionName(id);
```
tomas-langer: AI_Findings: **[P1] Collision reservations and selective scraping use a pre-exposition family name.** `io.helidon.metrics.providers.micrometer.CollisionDetectingPrometheusMeterRegistry:118`, `io.helidon.metrics.providers.micrometer.MicrometerPrometheusFormatter:192` — `getConventionName` can retain characters introduced by `baseUnit(String)` which Prometheus Java Client later canonicalizes for its stored and exposed family name. For example, an accepted gauge `payload` with base unit `bytes/second` is reserved and selected as `payload_bytes/second`, while the client stores and exposes `payload_bytes_second`. Selecting `payload` therefore returns no family, and a second gauge named `payload_bytes_second` bypasses the reservation despite colliding at exposition, which can still make scrape throw `DuplicateLabelsException`.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
```diff
@@ -0,0 +1,225 @@
...
+                        () -> super.newTimer(id, distributionStatisticConfig, pauseDetector));
+    }
+
+    @Override
+    protected <T> Gauge newGauge(Meter.Id id, T obj, ToDoubleFunction<T> valueFunction) {
+        return register(id, gaugeNames(id), () -> super.newGauge(id, obj, valueFunction));
+    }
+
+    @Override
+    protected <T> FunctionCounter newFunctionCounter(Meter.Id id, T obj, ToDoubleFunction<T> countFunction) {
```
tomas-langer, line 105: AI_Findings: **[P1] Standard Micrometer timer families bypass or contradict the new family accounting.** `io.helidon.metrics.providers.micrometer.CollisionDetectingPrometheusMeterRegistry:101`, `io.helidon.metrics.providers.micrometer.MicrometerPrometheusFormatter:196` — `FunctionTimer` and `LongTaskTimer` remain reachable through the underlying Micrometer registry but do not pass through the reservation overrides. A `FunctionTimer` has `Meter.Type.TIMER` but no `_max` family, so selection adds `_max` and can expose an unrelated gauge; a `LongTaskTimer` does emit secondary and `_max` families but neither reserves nor selects them. Collisions can still fail an unfiltered scrape, and selective output can omit or include the wrong family.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
```diff
@@ -203,35 +179,23 @@ Set<String> meterNamesOfInterest(PrometheusMeterRegistry prometheusMeterRegistry
         Set<String> result = new HashSet<>();
 
         for (Meter meter : prometheusMeterRegistry.getMeters()) {
```
tomas-langer, line 202: AI_Findings: **[P2] Default-scoped direct Micrometer meters are omitted from scoped output.** `io.helidon.metrics.providers.micrometer.MicrometerPrometheusFormatter:181` — The new loop checks the requested scope only against the Prometheus child meter’s physical tag. `MMeterRegistry` assigns directly registered Micrometer meters Helidon’s configured default scope even when their Micrometer ID has no scope tag; the previous loop used that wrapper scope. Selecting the default scope now skips those accepted meters, so the scoped metrics endpoint silently omits them.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
tomas-langer: AI_Findings: **[P2] This performance concern remains unresolved.** The response confirms that no base-versus-head benchmark evidence was added and identifies full-scrape allocation and narrow selective-scrape latency as credible risks. A qualitative expectation of recording parity does not bound the changed recording and scrape hot paths, so the original finding still applies at head `16ccb7f8240614f3471305affe46b996afbfc91c`.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
tjquinno: Addressed in 9824d2d310. The new package-private PrometheusNameSupport.expositionName method applies the active Micrometer naming convention and then PrometheusNaming.prometheusName, matching the final character canonicalization used for exposition. CollisionDetectingPrometheusMeterRegistry now uses that result for every reservation, and MicrometerPrometheusFormatter uses the same result for selected-family names and family-to-tag mappings. Therefore a base unit such as bytes/second is consistently treated as bytes_second for registration, selection, and output. TestPrometheusNaming.testExpositionNameCollisionsAreRejectedDuringRegistration and TestPrometheusFormatting.testSelectiveScrapeUsesExpositionName cover the reported collision and selective-scrape cases.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
tjquinno: Addressed in 9824d2d310. CollisionDetectingPrometheusMeterRegistry now overrides newFunctionTimer and newLongTaskTimer. Function timers reserve their base, count, sum, and created names without claiming a max family; long-task timers use the distribution-family reservation including the separate max family and optional bucket family. MicrometerPrometheusFormatter now decides max-family inclusion from the actual Timer, DistributionSummary, or LongTaskTimer interface instead of Meter.Type.TIMER, so FunctionTimer does not accidentally select an unrelated max family and LongTaskTimer does select its own. Tests cover registration-time collisions and both selective-scrape behaviors in TestPrometheusNaming and TestPrometheusFormatting.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
tjquinno: The rebase makes this pre-rebase finding no longer applicable. Upstream main removed active metrics scope support, and the rebased formatter does not store, inspect, or filter by scope. Its deprecated scopeSelection and scopeTagName builder methods validate non-null arguments but are otherwise no-ops, and the deprecated provider overload is retained only as an inert compatibility bridge. Tag selection is now the active filtering API. TestPrometheusFormatting.testScopeSelectionIsIgnored verifies that the inherited deprecated scope arguments have no effect, so the PR does not reintroduce scope behavior.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
tjquinno: Addressed with reproducible base-versus-head JMH coverage in 9824d2d310. MetricsRecordingJmhBenchmark and its runner measure counter, timer, and distribution-summary recording throughput and allocation. MetricsFormatterJmhBenchmark now covers unfiltered, tag-selected, distinct-family name-only, and name-plus-tag scrapes at cardinalities 10, 100, and 1000, with GCProfiler allocation data. Using identical three-fork runs, recording remained effectively unchanged: counter +0.3%, timer +1.5%, and distribution summary +2.4%, with essentially zero allocation. At 1000, unfiltered scrape latency changed from 199 to 836 microseconds, name-only from 34 to 693 microseconds, and name-plus-tag from 35 to 694 microseconds. The results confirm additional scrape allocation and latency from the new client snapshot model, especially because the Micrometer MultiCollector does not publish descriptors that would let the registry skip unrelated collectors. Scraping is normally performed only once per monitoring interval, commonly every tens of seconds, whereas metric recording occurs throughout ordinary request processing; even the 1000-family cases remain below one millisecond in these tests. The change therefore leaves the normal server-operation hot path unaffected while supplying repeatable measurements for the less-frequent scrape path.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
```diff
@@ -0,0 +1,264 @@
...
+                                                     countFunction,
+                                                     totalTimeFunction,
+                                                     totalTimeFunctionUnit));
+    }
+
+    @Override
+    protected LongTaskTimer newLongTaskTimer(Meter.Id id,
+                                             DistributionStatisticConfig distributionStatisticConfig) {
+        return register(id,
+                        distributionNames(id, distributionStatisticConfig),
```
tomas-langer: AI_Findings: **[P2] Long-task histograms allow duplicate output samples.** `io.helidon.metrics.providers.micrometer.CollisionDetectingPrometheusMeterRegistry:128` omits the emitted `_gcount` and `_gsum` names from its reservations. A histogram-enabled timer named `operation` and a gauge named `operation_seconds_gcount` both register, producing invalid duplicate samples. [OpenMetrics suffix contract](https://prometheus.io/docs/specs/om/open_metrics_spec/#suffixes).

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
```diff
@@ -291,60 +254,105 @@ private static Map<String, Set<String>> meterTagNamesByFamily(PrometheusMeterReg
...
 
-        while (metricFamilySamples.hasMoreElements()) {
-            Collector.MetricFamilySamples family = metricFamilySamples.nextElement();
+        for (MetricSnapshot snapshot : snapshots) {
             Set<String> meterTagNames = selectsGeneratedLabel
-                    ? meterTagNamesByFamily.getOrDefault(family.name, Set.of())
-                    : commonLabelNames(family.samples);
-            List<Collector.MetricFamilySamples.Sample> matchingSamples = family.samples.stream()
-                    .filter(sample -> matchesTagSelection(prometheusMeterRegistry, meterTagNames, sample))
+                    ? meterTagNamesByFamily.getOrDefault(snapshot.getMetadata().getPrometheusName(), Set.of())
```
tomas-langer, line 317: AI_Findings: **[P2] Legacy counter tag selection drops matching metrics.** `io.helidon.metrics.providers.micrometer.MicrometerPrometheusFormatter:320` looks up snapshot key `jobs_total` in a map keyed by `jobs`. With legacy naming enabled, selecting a counter's real `le=actual` tag therefore omits it. The same applies to `quantile`, `statistic`, and `vmrange`.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
```diff
@@ -0,0 +1,264 @@
...
+        reservationLock.lock();
+        try {
+            for (String name : names) {
+                Claim claim = claims.get(name);
+                if (claim != null && !claim.owner.equals(owner)) {
+                    throw new IllegalArgumentException("Prometheus metric name collision: '" + name
+                                                               + "' is produced by both " + claim.owner + " and " + owner);
+                }
+            }
+            names.forEach(name -> claims.compute(name, (ignored, claim) -> claim == null
```
tomas-langer: AI_Findings: **[P3] Unused lambda bindings violate Rule 1.5.** `io.helidon.metrics.providers.micrometer.CollisionDetectingPrometheusMeterRegistry:205` and `:236` name deliberately unused parameters `ignored`. [Rule 1.5](https://github.com/helidon-io/helidon/blob/main/DEV-GUIDELINES.md#rule-1-5).

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
tjquinno: Addressed in 61818480f0. newLongTaskTimer now uses longTaskTimerNames, which adds the emitted _gcount and _gsum names to the collision reservations whenever histogram publication is enabled, while retaining the existing distribution-family reservations. The updated testLongTaskTimerGeneratedNameCollisionsAreRejectedDuringRegistration registers a histogram-enabled operation timer, verifies that gauges named operation_seconds_gcount and operation_seconds_gsum are rejected during registration, and confirms that the timer emits those samples.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
tjquinno: Addressed in 61818480f0. meterTagNamesByFamily now keys the actual-tag schema by the exact exposition family name and no longer strips _total from legacy counters. The jobs_total snapshot therefore finds the jobs_total tag schema. The new testLegacyCounterActualGeneratedNameTagIsSelectable verifies a legacy counter with a real le=actual tag; the corrected lookup is generic and applies equally to quantile, statistic, and vmrange.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/CollisionDetectingPrometheusMeterRegistry.java
tjquinno: Addressed in 61818480f0. Both deliberately unused map-computation key parameters now use the unnamed variable _, in accordance with Rule 1.5. The Micrometer provider checkstyle run reports zero violations.

tomas-langer (approved):

PR #12496, opened 2026-09-03
Mark metrics.prometheus.histogramFlavor as deprecated

### LINKED ISSUE #12495: 4.x - Mark `metrics.prometheus.histogramFlavor` as deprecated
Helidon Version:4.x

## Task Description
When we adopt a more recent Prometheus client in 27, Prometheus will no longer honor the setting corresponding to our `metrics.prometheus.histogramFlavor` setting. In 27 we will continue to allow the setting to appear in config (to avoid breaking apps that currently set it) but Helidon will ignore it.

tjquinno:
### Description
Resolves #12495 

In 27 (see #12487) we will adopt a newer Prometheus client. The newer client will no longer support the histogram flavor setting, so in 27 we will ignore the corresponding Helidon config setting. Mark it as deprecated now.

### Documentation
Updated


metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigBlueprint.java
```diff
@@ -146,6 +146,17 @@ static List<Tag> createTags(String pairs) {
...
+    /**
+     * Histogram flavor to use for Prometheus output.
+     *
+     * @return configured Prometheus histogram flavor
+     * @deprecated This Micrometer setting will be removed in a future release of Helidon with no replacement.
+     */
+    @Deprecated(since = "4.5.5", forRemoval = true)
+    @Option.Configured("prometheus.histogramFlavor")
+    @Option.Access("")
+    Optional<String> prometheusHistogramFlavor();
```
tomas-langer: `@Option.Access("")` changes only the visibility of generated builder methods; prototype getters remain public. Consequently, this option adds the public abstract `MetricsConfig#prometheusHistogramFlavor()` method in 4.5.5. Direct implementations and test doubles of `MetricsConfig` must add the method to recompile, and the new API is deprecated for removal from its first release, even though #12495 asks only to deprecate the existing configuration key. This turns config metadata into a new public API and compatibility obligation outside the issue's scope.

tjquinno: Yes. In fact, we did not have an explicit setting for this previously but we would pass it through to Prometheus as-is.

There is no need to add this setting to our blueprint at all. There is surely a better way to alert users who do use the histogram flavor setting that they should consider it deprecated. Working on that.

tjquinno: Fixed.

metrics/api/src/main/java/io/helidon/metrics/api/MetricsConfigSupport.java
```diff
@@ -132,6 +136,12 @@ public void decorate(MetricsConfig.BuilderBase<?, ?> builder) {
             if (builder.config().isEmpty()) {
                 builder.config(Services.get(Config.class).get(MetricsConfigBlueprint.METRICS_CONFIG_KEY));
             }
+            if (builder.config().orElseThrow().get(PROMETHEUS_HISTOGRAM_FLAVOR_CONFIG_KEY).exists()) {
+                LOGGER.log(Level.WARNING,
```
tomas-langer, line 144: `MetricsConfigSupport.BuilderDecorator` runs on every `MetricsConfig` build, and normal factory initialization builds the config once in `MetricsFactoryManager#createCurrentMetricsFactory` and again in `applyOverrides`. A single configured `metrics.prometheus.histogramFlavor` therefore emits two identical warnings during startup, and any later rebuild can emit more. This deprecation warning should be logged at most once during the lifetime of the VM. The new test exercises only one direct `MetricsConfig.create(config)` call, so it does not cover the duplicate-warning path.

tjquinno: Fixed.

tomas-langer (approved):

PR #12504, opened 2026-09-05
Limit metrics scope to Helidon MP only

### LINKED ISSUE #11540: Limit metrics scope to Helidon MP only
Helidon’s `scope` concept aligns with MicroProfile Metrics, where `base`, `vendor`, and `application` are part of the model and, in current spec form, are represented as an `mp_scope` tag. That gives Helidon continuity with MP-era semantics and preserves compatibility for users who think in those terms.

At the same time, the broader metrics ecosystem Helidon interoperates with, especially Micrometer and Prometheus, does not natively require this scope taxonomy. In those systems, metric identity is fundamentally name plus tags/labels, and any `scope` label is framework-defined rather than a standard convention.

Similarly, in OpenTelemetry metrics, timeseries identity is based on metric name plus attributes, along with resource and instrumentation scope at the OTLP model level; OTel’s “scope” means instrumentation scope such as the emitting library/module name and version, not a classification of metrics into MP-style buckets.

OpenTelemetry also standardizes metrics primarily through semantic conventions and attribute naming, not through a mandatory three-way taxonomy like MicroProfile Metrics.

Helidon's priority should be alignment with the modern portable observability stacks such as Micrometer, Prometheus, and OpenTelemetry. As such, `scope` should be treated as legacy compatibility metadata rather than a core cross-platform concept. 

We should plan on the following:

- Short term: keep internal support for `scope` to avoid compatibility breaks.
- Medium term: demote it conceptually. Document it as a Helidon/MP compatibility feature, not a universal metrics primitive.

tomas-langer:
Resolves #11540

Removes scope-specific behavior from core metrics while retaining the existing scope APIs as deprecated compatibility
no-ops. Adds provider-neutral meter-origin customization and formatter tag selection so integrations can implement their
own classification and filtering without core depending on a scope model.


metrics/api/src/main/java/io/helidon/metrics/api/Meter.java
```diff
@@ -55,11 +56,15 @@ public interface Meter extends Wrapper {
...
+     * Always returns empty because core metrics does not assign scopes.
      *
-     * @return scope
+     * @return empty
+     * @deprecated Core metrics does not assign scopes, and this method will be removed.
      */
-    Optional<String> scope();
+    @Deprecated(forRemoval = true, since = "27.0.0")
+    default Optional<String> scope() {
+        return Optional.empty();
```
danielkec, line 66: When current `helidon-microprofile` main at `f7d087e` runs against this change, its application/base/vendor registries cannot create meters: the MP builders still call `scope(...)`, and the synchronous add listener routes the new delegate using `delegate.scope()`, which now always returns empty. The listener records the delegate in a null-scope registry, so a bean invoking `registry.counter("probe").inc()` on an injected `MetricRegistry` fails during startup, and MP output also loses its `mp_scope` tag. The selected MP metrics tests pass against base `bbd3371` but report 1 failure and 10 errors on this head. There is currently no companion `helidon-microprofile` PR covering this migration, so the MP metrics runtime remains broken.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
```diff
@@ -267,14 +254,59 @@ private static String normalizeUnit(String unit) {
...
+            throw new UncheckedIOException("Error preparing Prometheus metrics output", e);
+        }
+        return result.toString();
+    }
+
+    private boolean matchesTagSelection(PrometheusMeterRegistry prometheusMeterRegistry,
+                                        Collector.MetricFamilySamples.Sample sample) {
+        for (Map.Entry<String, Set<String>> selection : tagSelection.entrySet()) {
+            String tagName = prometheusMeterRegistry.config().namingConvention().tagKey(selection.getKey());
+            int labelIndex = sample.labelNames.indexOf(tagName);
```
danielkec, line 346: `sample.labelNames` also contains generated Prometheus labels such as `le` and `quantile`, not only meter tags. With a distribution summary that has SLOs `1, 2` and no `le` tag, `tagSelection(Map.of("le", Set.of("+Inf")))` returns only `dist_bucket{le="+Inf"} 1.0` and drops `_count` and `_sum`, although no meter tag matched. This lets a tag filter match unintended data and emits an incomplete metric family.

webserver/observe/metrics/src/main/java/io/helidon/webserver/observe/metrics/JsonFormatter.java
```diff
@@ -357,6 +282,25 @@ private boolean matchesName(String metricName) {
         return false;
     }
 
+    private boolean matchesTags(Meter.Id meterId) {
+        if (tagSelection.isEmpty()) {
+            return true;
+        }
+        for (Map.Entry<String, Set<String>> selection : tagSelection.entrySet()) {
+            boolean matches = false;
+            for (Tag tag : meterId.tags()) {
```
danielkec: `Meter.Id.tags()` excludes configured system tags, while the Prometheus formatter sees those tags as common labels. With a configured `env=prod` system tag, an otherwise untagged counter, and `tagSelection(Map.of("env", Set.of("prod")))`, JSON returns empty while Prometheus returns the counter. The new provider-neutral filter therefore changes meaning with the requested media type.

metrics/api/src/main/java/io/helidon/metrics/api/Meter.java
tomas-langer: Resolved by helidon-io/helidon-microprofile#103.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MicrometerPrometheusFormatter.java
```diff
@@ -267,14 +254,75 @@ private static String normalizeUnit(String unit) {
         return unit == null ? "" : unit;
     }
 
+    private static Set<String> commonLabelNames(List<Collector.MetricFamilySamples.Sample> samples) {
+        if (samples.isEmpty()) {
+            return Set.of();
+        }
+        Set<String> result = new HashSet<>(samples.getFirst().labelNames);
```
danielkec, line 267: The earlier `le`/`quantile` cases are now rejected, but this intersection still classifies a generated label as a meter tag when that label occurs on every sample. For example, this meter's only actual ID tag is `real=yes`:

```java
Meter.builder("probe", Meter.Type.OTHER, List.of(
        new Measurement(() -> 1, Statistic.COUNT),
        new Measurement(() -> 2, Statistic.TOTAL)))
        .tag("real", "yes")
        .register(delegate);
// tagSelection(Map.of("statistic", Set.of("COUNT")))
```

Micrometer generates `statistic=COUNT/TOTAL`; the selection returns only the COUNT sample and drops `probe_sum`, so a selection matching no meter tag still emits an incomplete family.

metrics/providers/micrometer/src/main/java/io/helidon/metrics/providers/micrometer/MMeterRegistry.java
```diff
@@ -310,15 +295,15 @@ public Clock clock() {
...
+        try {
+            checkOpen();
+        } finally {
+            lock.readLock().unlock();
         }
-
-        // If this is not "one of ours" then we need to create a new builder, based on the one passed in but with the correct
-        // Micrometer delegate builder assigned.
-        return (HM) getOrCreateUntyped(convertNeutralBuilder(builder));
+        metricsFactory.customize(builder);
```
danielkec, line 305: When `getOrCreate` is already in the customizer chain and another thread completes `close()`, a later customizer still starts after the registry is closed. A two-customizer schedule where the first callback waits, `close()` returns, and then the first resumes produces:

```text
second before release=0
creation failure=java.lang.IllegalStateException: Meter registry is closed
second invocations=1, observed closeReturned=true
```

This invokes user service code after registry resources have been closed for a registration that is then rejected.

danielkec (commented):
Two remaining current-head issues.

danielkec (approved):

tjquinno (approved):

PR #12507, opened 2026-09-05
Generalize fault tolerance retry handling

tomas-langer:
Add contextual retry invocation to the core fault-tolerance API so callers such as messaging connectors can share the existing retry implementation.

- Expose per-attempt `RetryContext` and terminal `RetryOutcome` through `RetryException`.
- Allow caller-controlled waits while preserving `FtHandler` chaining and the existing `Supplier` contract.
- Add exponential delay with absolute or relative jitter and an optional maximum delay, including declarative and configuration parity.
- Preserve caller interruption provenance and bound retained failures.
- Move blueprint decorators into package-private fault-tolerance support.

### DIFF
--- a/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FtBuilderSupport.java
+++ b/fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FtBuilderSupport.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2026 Oracle and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.helidon.faulttolerance;
+
+import java.time.Duration;
+
+import io.helidon.builder.api.Prototype;
+
+final class FtBuilderSupport {
+    private FtBuilderSupport() {
+    }
+
+    static class AsyncBuilderDecorator implements Prototype.BuilderDecorator<AsyncConfig.BuilderBase<?, ?>> {
+        @Override
+        public void decorate(AsyncConfig.BuilderBase<?, ?> target) {
+            if (target.name().isEmpty()) {
+                target.config()
+                        .ifPresent(cfg -> target.name(cfg.name()));
+            }
+        }
+    }
+
+    static class BulkheadBuilderDecorator implements Prototype.BuilderDecorator<BulkheadConfig.BuilderBase<?, ?>> {
+        @Override
+        public void decorate(BulkheadConfig.BuilderBase<?, ?> target) {
+            if (target.name().isEmpty()) {
+                target.config()
+                        .ifPresent(cfg -> target.name(cfg.name()));
+            }
+        }
+    }
+
+    static class CircuitBreakerBuilderDecorator
+            implements Prototype.BuilderDecorator<CircuitBreakerConfig.BuilderBase<?, ?>> {
+        @Override
+        public void decorate(CircuitBreakerConfig.BuilderBase<?, ?> target) {
+            if (target.name().isEmpty()) {
+                target.config()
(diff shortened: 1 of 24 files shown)


fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/Retry.java
```diff
@@ -158,11 +302,30 @@ public Optional<Long> nextDelayMillis(long firstCallMillis, long lastDelay, int
                 return Optional.empty();
             }
 
-            if (call == 0 || lastDelay == 0) {
-                return Optional.of(delayMillis);
+            long delay = multiply(delayMillis, Math.pow(delayFactor, Math.max(0, call - 1)));
```
danielkec: When `delayFactor` is fractional, recomputing each delay from the base changes existing no-jitter schedules and can turn a successful retry sequence into a timeout. A minimal exact base/head case is:

```text
delay=1ms, factor=1.5, calls=18, overallTimeout=100ms
base: 17 x 1ms -> attempt 18 succeeds
head: 1,1,2,3,5,7,11,17,25,... -> RetryTimeoutException after attempt 10
```

This also differs from the Javadoc above, which describes multiplying the previous delay by the factor.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/RetryImpl.java
```diff
@@ -33,8 +34,12 @@
 import io.helidon.service.registry.ServiceRegistry;
 
 class RetryImpl implements Retry {
+    private static final int MAX_RETAINED_FAILURES = 16;
```
danielkec, line 37: The retention cap itself is described in the PR body, but it makes the public fallback contract inaccurate: `Ft.Fallback` still says that all intermediate exceptions are added as suppressed. With 20 failed retry attempts, the base exposes 19 suppressed failures while this head exposes only 15, so consumers inspecting the fallback cause silently lose the earliest failures despite the API promise.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FtBuilderSupport.java
```diff
@@ -0,0 +1,141 @@
...
+
+    static class RetryBuilderDecorator implements Prototype.BuilderDecorator<RetryConfig.BuilderBase<?, ?>> {
+        @Override
+        public void decorate(RetryConfig.BuilderBase<?, ?> target) {
+            validate(target);
+            if (target.name().isEmpty()) {
+                target.config()
+                        .ifPresent(cfg -> target.name(cfg.name()));
+            }
+            if (target.retryPolicy().isEmpty()) {
```
danielkec, line 71: When a config is copied with `RetryConfig.builder(existing)`, the generated `from(...)` carries over the already-derived `retryPolicy`, so this branch skips rebuilding it after the new delay settings change. A minimal exact-head case is:

```java
var original = RetryConfig.builder().delay(Duration.ofMillis(100)).buildPrototype();
var changed = RetryConfig.builder(original).maxDelay(Duration.ofMillis(50)).buildPrototype();
changed.retryPolicy().orElseThrow().nextDelayMillis(0, 100, 2); // Optional[200]
```

`changed.maxDelay()` is `PT0.05S`, but the copied config shares the original policy and its advertised maximum-delay cap is silently ignored.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/FtBuilderSupport.java
```diff
@@ -0,0 +1,178 @@
...
```
danielkec, line 71: The earlier `RetryConfig.builder(original).maxDelay(...)` order now works, but setting the optional override before `from(...)` still leaves the config and copied policy inconsistent:

```java
var original = RetryConfig.builder().delay(Duration.ofMillis(100)).buildPrototype();
var changed = RetryConfig.builder()
        .maxDelay(Duration.ofMillis(50))
        .from(original)
        .buildPrototype();
```

`changed.maxDelay()` is `PT0.05S`, while the copied policy returns delays of 100 ms and then 200 ms. A caller layering a prototype after this optional override therefore silently runs without the advertised maximum-delay cap.

tomas-langer: This is correct. Whatever is called later must win. So in this case whatever is in `original` will override `maxDelay` - this is a design followed by Helidon builders.

danielkec: agree

danielkec (approved):

PR #12512, opened 2026-09-06
Restore default-port CORS origin matching (4.x)

### LINKED ISSUE #12190: Unexpected CORS behaviour change with default https ports
## Environment Details
* Helidon Version: 4.4.1
* JDK version: 21
* OS: Linux x86-64

----------

## Problem Description
Breakage noticed between Helidon 4.2.4 and 4.4.1 for CORS behaviour.  Specifically, Helidon no longer treats `https://somehost` and `https://somehost:443` as equivalents.  An cors.path origin configuration that has the latter used to handle both these client-provided origins in 4.2.4, but does an exact match as of 4.4.1.

Easily reproducible.

In internal discussion, this has been traced to commit `7cf9b9f354` “CORS updated and declarative CORS” - changing path validation from `CorsSupportHelper.compareOrigins` (which knows these urls are equivalent) to an exact match `allowedOriginsExact.contains`.

## Steps to reproduce
Setup a hello world service on Helidon; configure cors to allow origin "https://myhostname:443".  Hit the endpoint with a browser that suppresses the default port (or just use curl with ORIGIN injected in that manner).  With 4.2.4, this goes through.  With 4.4.1, there is a 403 response.

n/a

tomas-langer:
Resolves #12190

Normalize explicit HTTP and HTTPS default ports in configured exact CORS origins so they match browser-serialized `Origin` headers.

### DIFF
--- a/webserver/cors/src/main/java/io/helidon/webserver/cors/CorsPathValidator.java
+++ b/webserver/cors/src/main/java/io/helidon/webserver/cors/CorsPathValidator.java
@@ -17,6 +17,7 @@
 package io.helidon.webserver.cors;
 
 import java.lang.System.Logger.Level;
+import java.net.URI;
 import java.time.Duration;
 import java.util.HashSet;
 import java.util.List;
@@ -234,12 +235,15 @@ class CorsPathValidator {
 
     private static String normalizeDefaultPort(String origin) {
         int authorityOffset;
+        int defaultPort;
         int portLength;
         if (origin.startsWith("http://") && origin.endsWith(":80")) {
             authorityOffset = "http://".length();
+            defaultPort = 80;
             portLength = 3;
         } else if (origin.startsWith("https://") && origin.endsWith(":443")) {
             authorityOffset = "https://".length();
+            defaultPort = 443;
             portLength = 4;
         } else {
             return origin;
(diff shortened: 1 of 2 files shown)


webserver/cors/src/main/java/io/helidon/webserver/cors/CorsPathValidator.java
```diff
@@ -78,7 +78,7 @@ private CorsPathValidator(CorsPathConfig config) {
                     // this is a pattern
                     originPatterns.add(Pattern.compile(origin, Pattern.CASE_INSENSITIVE));
                 } else {
-                    exactMatchOrigins.add(origin);
+                    exactMatchOrigins.add(normalizeDefaultPort(origin));
```
danielkec, line 84: Normalizing only the configured value makes the same explicit-port origin stop matching. With:

```text
allow-origins: [https://example.com:443]
Origin: https://example.com:443
```

the base allows the request, but this head stores `https://example.com` while the header is still compared verbatim, so it now returns 403. Issue #12190 describes this configuration as previously handling both forms, while the added test covers only the no-port request.

webserver/cors/src/main/java/io/helidon/webserver/cors/CorsPathValidator.java
```diff
@@ -232,6 +232,28 @@ Result flight(ServerRequest req, ServerResponse res) {
...
+            return origin;
+        }
+        int portOffset = origin.length() - portLength;
+        if (portOffset == authorityOffset
+                || origin.indexOf('/', authorityOffset) != -1
+                || origin.indexOf('?', authorityOffset) != -1
+                || origin.indexOf('#', authorityOffset) != -1) {
+            return origin;
+        }
+        return origin.substring(0, portOffset);
```
danielkec, line 262: This textual suffix removal can turn a malformed allowlist entry into a different valid origin. A minimal case is:

```text
allow-origins: [https://example.com:8443:443]
Origin: https://example.com:8443
```

The base returns 403, but this head strips the final `:443`, accepts the request, and echoes the valid origin in `Access-Control-Allow-Origin`. This makes malformed configuration authorize an origin that was not validly configured.

danielkec (approved):

PR #12178, opened 2026-07-15
Add support for RFC 10008 (The HTTP QUERY Method)

desiderantes:
### Description

Resolves #12174

This change adds support for [RFC 10008](https://www.rfc-editor.org/info/rfc10008), including the `Accept-Query` header and redirection handling for the QUERY method.

This addition would also enable future OpenAPi 3.2 support.

PR #12518, opened 2026-09-06
Support RFC 10008 QUERY method (27.x)

### LINKED ISSUE #12174: Support for QUERY HTTP Method
see https://datatracker.ietf.org/doc/html/rfc10008

tomas-langer:
Resolves #12174

Adds RFC 10008 `QUERY` support across HTTP, WebClient, WebServer, declarative clients and servers, and HTTP metrics.

- Adds `Method.QUERY` and `@Http.QUERY`; declarative code generation uses the existing generic client and routing APIs without adding QUERY convenience methods.
- Validates QUERY `Content-Type` requirements in WebServer and WebClient.
- Preserves the QUERY method and entity across 301, 302, 307, and 308 redirects, changes to GET for 303, and retains cross-origin replay protection.
- Adds RFC 9651 `Accept-Query` constants and typed response-header helpers backed by a package-private parser.
- Includes documentation and targeted redirect benchmarks.

This replaces #12178.

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
@@ -185,6 +185,10 @@ class Http1ClientRequestImpl extends ClientRequestBase<Http1ClientRequest, Http1
             }
         }
 
+        if (method() == Method.QUERY && !headers().contains(HeaderNames.CONTENT_TYPE)) {
+            throw new IllegalArgumentException("Content-Type header is required for method '" + Method.QUERY + "'");
+        }
+
         if (followRedirects()) {
             return RedirectionProcessor.invokeWithFollowRedirects(this, entityBytes);
         }
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
@@ -46,9 +46,14 @@ import io.helidon.webclient.api.WebClientServiceResponse;
 class Http1ClientRequestImpl extends ClientRequestBase<Http1ClientRequest, Http1ClientResponse>
         implements Http1ClientRequest {
     private static final System.Logger LOGGER = System.getLogger(Http1ClientRequestImpl.class.getName());
-    // RFC 9110, section 8.1: these define the replayed representation data's format and encoding.
-    private static final Set<HeaderName> REPRESENTATION_HEADERS = Set.of(HeaderNames.CONTENT_TYPE,
-                                                                         HeaderNames.CONTENT_ENCODING);
+    private static final Set<HeaderName> REPLAYABLE_HEADERS = Set.of(HeaderNames.ACCEPT,
+                                                                     HeaderNames.ACCEPT_CHARSET,
+                                                                     HeaderNames.ACCEPT_ENCODING,
+                                                                     HeaderNames.ACCEPT_LANGUAGE,
+                                                                     HeaderNames.CONTENT_ENCODING,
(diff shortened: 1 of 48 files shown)


desiderantes:
About the OCA, I already signed it, but because I live in Brazil, it requires review, so it's still in Under Review status

webclient/http1/src/main/java/io/helidon/webclient/http1/RedirectionProcessor.java
```diff
@@ -36,17 +36,20 @@ static boolean redirectionStatusCode(Status status) {
             && status.family() == Status.Family.REDIRECTION;
     }
 
-    static boolean keepsMethodAndEntity(Status status) {
+    static boolean keepsMethodAndEntity(Method method, Status status) {
         int statusCode = status.code();
         return statusCode == Status.TEMPORARY_REDIRECT_307.code()
-                || statusCode == Status.PERMANENT_REDIRECT_308.code();
+                || statusCode == Status.PERMANENT_REDIRECT_308.code()
+                || (Method.QUERY.equals(method)
```
danielkec, line 43: When a QUERY request with a per-request `Accept` header follows a same-origin 301 or 302, this preservation branch recreates it through a redirect request that copies only `Content-Type` and `Content-Encoding`. A minimal 302 case is:

```java
client.method(Method.QUERY)
        .uri("/source")
        .accept(MediaTypes.APPLICATION_JSON)
        .contentType(MediaTypes.create("application", "sql"))
        .submit("select 1".getBytes());
```

In both HTTP/1 and HTTP/2, the target receives QUERY, the body, and `Content-Type: application/sql`, but no `Accept` (`Content-Language` is dropped likewise), so the redirect can change response negotiation or query interpretation. RFC 10008 section 2.5 describes replaying a similar QUERY, while RFC 9110 section 15.4 removes content-specific fields when the method changes to GET or HEAD; this branch keeps QUERY. The equivalent HTTP/2 branch in `webclient/http2/.../RedirectionProcessor.java` has the same behavior.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientRequestImpl.java
```diff
@@ -46,9 +46,14 @@
 class Http1ClientRequestImpl extends ClientRequestBase<Http1ClientRequest, Http1ClientResponse>
         implements Http1ClientRequest {
     private static final System.Logger LOGGER = System.getLogger(Http1ClientRequestImpl.class.getName());
-    // RFC 9110, section 8.1: these define the replayed representation data's format and encoding.
-    private static final Set<HeaderName> REPRESENTATION_HEADERS = Set.of(HeaderNames.CONTENT_TYPE,
-                                                                         HeaderNames.CONTENT_ENCODING);
+    private static final Set<HeaderName> REPLAYABLE_HEADERS = Set.of(HeaderNames.ACCEPT,
```
danielkec, line 49: The earlier `Accept` and `Content-Language` cases are covered now, but this finite whitelist still drops other caller fields that RFC 10008 assigns semantics to QUERY. A same-origin case is:

```text
QUERY /source
If-None-Match: "query-v1"
Range: bytes=0-9
Content-Type: application/sql

/source -> 302 Location: /target
/target receives QUERY + Content-Type, but neither conditional/range field
```

Both HTTP/1 and HTTP/2 reproduce this, which can turn a conditional or partial result into an unconditional, full result; RFC 10008 sections 2.6 and 2.8 define these behaviors for QUERY.

tomas-langer: This is a `SHOULD` not `MUST` from the point of view of the RFC. We will keep the minimal headers copied for now.

danielkec (approved):

PR #12521, opened 2026-09-07
Avoid startup failure when JWT provider JWKS is unavailable (27.x)

### LINKED ISSUE #11927: Avoid startup failure when JWT provider JWKS is unavailable
Helidon Version: 4.x

## Enhancement Description
  When Helidon JWT authentication is enabled with signature verification, the application currently fails during startup if the configured verification JWKS resource is missing, invalid, or temporarily unreachable.

  Expected behavior: startup should not fail solely because JWKS data is unavailable. The JWT provider should initialize successfully, but JWT authentication should fail at request time while verification keys are unavailable.

  Requested behavior:
  1. Do not fail application startup if security.providers[].jwt.atn-token.jwk.resource cannot be loaded.
  2. Keep verify-signature enabled by default.
  3. If no verification JWKS is available, JWT signature validation should fail authentication.
  4. Preserve the 'optional' behavior: if JWT auth is optional, the missing JWKS should result in abstain rather than a startup crash.
  5. Log a warning when JWKS loading fails, including enough detail to diagnose the resource issue.

  This would make applications more resilient to temporary JWKS endpoint outages during startup while still preserving secure request-time behavior.

tomas-langer:
Resolves #11927

Add resilient, lazy verification-key loading for JWT, OIDC, and IDCS-backed providers. Remote and file-backed resources use configured Fault Tolerance circuit breaker, retry, and per-attempt timeout policies, while fixed sources continue to validate at startup.

Also add shared cached-value and timeout-aware resource loading, concurrency and recovery coverage, generated configuration reference updates, and focused benchmarks.


security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
```diff
@@ -178,14 +197,21 @@ private AuthenticationResponse authenticateToken(String token) {
             //invalid token
             return failOrAbstain("Invalid token" + e);
         }
+        Jwt jwt = signedJwt.getJwt();
```
danielkec: `getJwt()` materializes typed claims before signature verification and outside the parse catch. This syntactically valid token has an invalid signature and a string `exp`, and it throws `JsonException` here rather than returning normal `FAILURE`/`ABSTAIN`:

```text
eyJhbGciOiJSUzI1NiIsImtpZCI6InZlcmlmeS1yc2EifQ.eyJzdWIiOiJqb2UiLCJleHAiOiJvb3BzIn0.AA
```

Consequently, unauthenticated malformed input can escape the provider's authentication handling, including in optional mode.

security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
```diff
@@ -679,6 +720,11 @@ public static final class Builder implements io.helidon.common.Builder<Builder,
                 .build();
         private OutboundConfig outboundConfig = OutboundConfig.builder().build();
         private JwkKeys verifyKeys;
+        private ResourceConfig verifyKeysResource;
+        private ResilientValue<JwkKeys> verifyKeysLoader;
+        private Retry jwkRetry = defaultJwkRetry();
+        private CircuitBreaker jwkCircuitBreaker = defaultJwkCircuitBreaker();
```
danielkec: This stateful circuit breaker is initialized once on the builder and passed to every dynamic loader produced by subsequent `build()` calls, so reused builders also reuse failure history. With `volume(1)`, authenticate once while a path-based JWK file is absent, create the file, and build a second provider from the same builder: the second provider is still rejected by the first provider's open circuit, while an equivalent fresh builder succeeds. A new provider can therefore remain unavailable because of an earlier provider's load.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -63,6 +78,10 @@ public abstract class BaseBuilder<B extends BaseBuilder<B, T>, T> implements Bui
     private URI tokenEndpointUri;
     private Duration clientTimeout = Duration.ofSeconds(DEFAULT_TIMEOUT_SECONDS);
     private JwkKeys signJwk;
+    private ResourceConfig signJwkResource;
+    private Retry jwkRetry = defaultJwkRetry();
+    private CircuitBreaker jwkCircuitBreaker = defaultJwkCircuitBreaker();
```
danielkec: This stateful circuit breaker is likewise initialized once on the OIDC base builder and copied to every configuration built from it. With `volume(1)`, fail the first configuration's lazy signing-JWK load against a missing file, create the file, and build another configuration from the same builder: the new configuration remains rejected until the first configuration's breaker delay expires, while a fresh builder succeeds. Reusing a `TenantConfig.Builder` can therefore couple otherwise independent tenants.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValueImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        try {
+            return retry.invoke(this::loadWithTimeout);
+        } catch (RetryTimeoutException e) {
+            throw new ResilientValue.UnavailableException(description + " did not become available before the retry timeout", e);
+        }
+    }
+
+    private T loadWithTimeout() {
+        try {
+            return timeout.invoke(loader);
```
danielkec, line 176: When a caller interrupts a stalled load, this current-thread timeout clears the thread's interrupted status before the outer retry observes the failure, so another attempt starts instead of terminating for cancellation. A loopback URI reproducer that interrupts the first stalled request observed:

```text
accepted=2, loads=2, result=ok, interrupted=false
```

Request or shutdown cancellation can therefore leave authentication resource loading running after the caller has cancelled it.

security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
```diff
@@ -985,7 +1237,194 @@ private static String requireText(String value, String description) {
...
+                                                         jwkCircuitBreaker,
+                                                         jwkTimeout);
+            } else {
+                verifyKeys = JwkKeys.builder()
+                        .resource(Resource.create(resourceConfig))
+                        .build();
+            }
+        }
+
+        static Retry defaultJwkRetry() {
```
danielkec: The three `defaultJwk*` methods are accessible outside `Builder`, although every production call is within the nested class; their only external callers are direct assertions in `JwtProviderJwkLoadingTest`. This relaxes production visibility solely for tests, contrary to Development Guideline 4.1.4, and exposes implementation-only defaults to the package.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -585,4 +844,144 @@ String name() {
...
     }
+
+    private static JwkKeys requireSigningKeys(JwkKeys keys) {
+        if (keys.keys().isEmpty()) {
+            throw new IllegalArgumentException("Configured OIDC signing JWK must contain at least one usable key");
+        }
+        return keys;
+    }
+
+    static Retry defaultJwkRetry() {
```
danielkec: These new package-private static methods appear after the class's non-private instance methods and a private static helper. Development Guideline 2.1 places non-private static methods immediately after constructors and before non-private instance methods, so the introduced layout does not follow the repository member order.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/OidcConfig.java
```diff
@@ -455,13 +444,23 @@ private OidcConfig(Builder builder) {
...
         this.webClientBuilderSupplier = builder.webClientBuilderSupplier;
-        this.defaultTenant = LazyValue.create(() -> Tenant.create(this, this));
+        this.defaultTenant = tenantSupplier(this);
         this.outboundType = builder.outboundType;
         this.clientCredentialsConfig = builder.clientCredentialsConfig;
 
         LOGGER.log(Level.TRACE, () -> "Redirect URI with host: " + frontendUri + redirectUri);
     }
 
+    private Supplier<Tenant> tenantSupplier(TenantConfig tenantConfig) {
```
danielkec: `tenantSupplier` is a private instance method placed before the public static `builder()` and `create(Config)` factories. Development Guideline 2.1 orders non-private static methods before private instance methods, so this introduced placement does not match the repository member order.

fault-tolerance/fault-tolerance/src/test/java/io/helidon/faulttolerance/ResilientValueTest.java
```diff
@@ -0,0 +1,777 @@
...
+        assertThrows(ResilientValue.UnavailableException.class, value::get);
+        assertThrows(ResilientValue.UnavailableException.class, value::get);
+        assertThat(executor.runNext(), is(true));
+        assertThat(value.get(), is("loaded"));
+
+        assertThat(handler.messages(Level.WARNING),
+                   is(List.of("logged test value is unavailable; retries are exhausted: safe failure detail")));
+        assertThat(handler.messages(Level.INFO), is(List.of("logged test value is available again")));
+        assertThat(String.join("\n", handler.messages()),
+                   org.hamcrest.CoreMatchers.not(org.hamcrest.CoreMatchers.containsString("raw cause")));
```
danielkec: `org.hamcrest.CoreMatchers.not` and `containsString` are fully qualified in executable code even though there is no name conflict. Development Guideline 3.1 requires imported simple names outside Javadoc.

danielkec (commented):
Review findings are in the inline comments.

common/configurable/src/test/java/io/helidon/common/configurable/ResourceTest.java
```diff
@@ -123,4 +132,103 @@ void testConfigContent() {
...
+                        .uri(uri)
+                        .buildPrototype();
+
+                long beforeLoad = System.nanoTime();
+                assertThrows(ResourceException.class,
+                             () -> Resource.create(resourceConfig, Duration.ofMillis(50)).bytes());
+                long loadMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - beforeLoad);
+
+                assertThat("test server accepted the load", accepted.await(5, TimeUnit.SECONDS), is(true));
+                assertThat("stalled load respected its read timeout", loadMillis < 5_000, is(true));
```
Verdent, line 165: The configured timeout is 50 ms, but this assertion accepts any load below five seconds. A one-second read timeout therefore passes this test, so the regression coverage does not establish that the configured duration is honored.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/Tenant.java
```diff
@@ -76,16 +87,26 @@ public static Tenant create(OidcConfig oidcConfig, TenantConfig tenantConfig) {
         Errors.Collector collector = Errors.collector();
 
         URI identityUri = tenantConfig.identityUri();
-        OidcMetadata oidcMetadata = OidcMetadata.builder()
+        Duration jwkTimeout = tenantConfig.tenantLoadingLazy()
```
Verdent, line 90: `tenantLoadingLazy()` is deliberately false for classpath and inline `ResourceConfig` sources, so `jwkTimeout` is null here. A custom `TenantConfig` exposing one of those fixed sources through `oidcMetadataResource()` or `tenantSignJwkResource()` then reaches `Resource.create(config, null)` and throws before the fixed resource can be loaded. Such a tenant fails provider/feature startup, or a finder-backed request, with an escaping `NullPointerException` instead of accepting the valid fixed resource.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/TimeoutImpl.java
```diff
@@ -130,6 +131,7 @@ public <T> T invoke(Supplier<? extends T> supplier) {
                 }
                 return result;
             } catch (Throwable t) {
+                restoreInterrupt = !interrupted.get() && hasInterruptedCause(t);
```
Verdent: A nested `InterruptedException` does not establish that this caller was interrupted. With `SupplierHelper.toSyncSupplier(() -> CompletableFuture.failedFuture(new InterruptedException()), ...)`, the failed stage leaves the invoking thread un-interrupted, but this sets `restoreInterrupt` and the `finally` block interrupts it. An enclosing retry then terminates as caller cancellation instead of applying its normal failure policy, and the thread exits with a spurious interrupt flag.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -124,7 +183,16 @@ public B config(Config config) {
         config.get("token-endpoint-uri").as(URI.class).ifPresent(this::tokenEndpointUri);
         config.get("logout-endpoint-uri").as(URI.class).ifPresent(this::logoutEndpointUri);
 
-        config.get("sign-jwk.resource").as(Resource::create).ifPresent(this::signJwk);
+        config.get("sign-jwk.resource").as(ResourceConfig::create).ifPresent(this::signJwk);
+        config.get("jwk-loader.retry")
+                .as(it -> RetryConfig.builder(DEFAULT_JWK_RETRY_CONFIG).config(it).buildPrototype())
```
Verdent: `tenantFromConfig` layers tenant configuration over the base OIDC configuration, and the documented multi-tenant contract says omitted tenant properties inherit from that base. Rebuilding this value from `DEFAULT_JWK_RETRY_CONFIG` discards the base values whenever a tenant overrides any retry field: for example, a base delay of 7 ms plus a tenant-only `calls: 2` produces the built-in 200 ms delay. The timeout and circuit-breaker mappings below behave the same way, so partial tenant overrides silently change the configured JWK outage policy.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValueImpl.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Override
+    public T get() {
+        if (loaded) {
+            return value;
+        }
+
+        LazyValue<Outcome<T>> currentAttempt = attempt.get();
+        Outcome<T> outcome = currentAttempt.get();
```
Verdent: Concurrent followers wait here through `LazyValue`, whose contended path uses `Semaphore.acquireUninterruptibly()`. With one thread blocked in the loader, interrupting a follower leaves it parked until that load ends, after which it returns the shared outcome. Canceled authentication or shutdown tasks can therefore remain retained for the full JWK load duration, and indefinitely when the underlying I/O does not unwind.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValueImpl.java
```diff
@@ -0,0 +1,151 @@
...
+        });
+    }
+
+    private T load() {
+        T loadedValue;
+        try {
+            loadedValue = circuitBreaker.invoke(this::loadWithRetry);
+        } catch (CircuitBreakerOpenException e) {
+            throw new ResilientValue.UnavailableException(description + " is temporarily unavailable", e);
+        } catch (UnavailableException e) {
```
Verdent, line 139: A caller cancellation is now excluded from retry and circuit-breaker accounting, but it still reaches this `UnavailableException` branch. The `callerInterruptStopsRetries` path therefore sets `failed` and logs that retries were exhausted; the next successful load logs that the value is available again, even though the source never failed and the breaker remained closed.

common/configurable/src/main/java/io/helidon/common/configurable/ResourceUtil.java
```diff
@@ -106,6 +108,60 @@ static InputStream toIs(URI uri, Proxy proxy) {
...
+        connection.setReadTimeout(timeoutMillis);
+        connection.setUseCaches(false);
+        return connection.getInputStream();
+    }
+
+    private static int timeoutMillis(Duration timeout) {
+        long timeoutMillis;
+        try {
+            timeoutMillis = timeout.toMillis();
+        } catch (ArithmeticException e) {
```
Verdent: This deliberately unused catch binding violates Development Guideline 1.5, which requires `_` for unused exception parameters.

security/providers/jwt/src/main/java/io/helidon/security/providers/jwt/JwtProvider.java
```diff
@@ -172,20 +193,28 @@ public AuthenticationResponse authenticate(ProviderRequest providerRequest) {
...
         } catch (Exception e) {
             //invalid token
             return failOrAbstain("Invalid token" + e);
         }
         if (verifySignature) {
-            Errors errors = signedJwt.verifySignature(verifyKeys, defaultJwk);
+            JwkKeys keys;
+            try {
+                keys = verificationKeys(jwt);
+            } catch (ResilientValue.UnavailableException e) {
```
Verdent: This caught exception is deliberately ignored after selecting the availability response. The named binding violates Development Guideline 1.5, which requires `_` for unused exception parameters.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -585,4 +934,165 @@ String name() {
...
+            try {
+                URI uri = URI.create(value);
+                if (!uri.isAbsolute()) {
+                    collector.fatal("OIDC metadata field \"" + key + "\" must be an absolute URI");
+                } else if (!isHttpUri(uri)) {
+                    collector.fatal("OIDC metadata field \"" + key + "\" must use HTTP or HTTPS");
+                } else if (uri.getHost() == null) {
+                    collector.fatal("OIDC metadata field \"" + key + "\" HTTP URI must include a host");
+                }
+            } catch (IllegalArgumentException e) {
```
Verdent: This `IllegalArgumentException` binding is never read. Development Guideline 1.5 requires `_` for deliberately unused exception parameters.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcProvider.java
```diff
@@ -154,35 +158,39 @@ public AuthenticationResponse authenticate(ProviderRequest providerRequest) {
...
-    private Optional<TenantAuthenticationHandler> cachedTenantAuthenticationHandler(String tenantId) {
-        Optional<TenantAuthenticationHandler> cachedHandler = tenantAuthHandlers.get(tenantId);
-        if (cachedHandler.isPresent()) {
-            return cachedHandler;
+        try {
+            return cachedTenantAuthenticationHandler(tenantId)
+                    .map(Supplier::get)
+                    .map(handler -> handler.authenticate(tenantId, providerRequest))
+                    .orElseGet(this::unknownTenantResponse);
+        } catch (ResilientValue.UnavailableException e) {
```
Verdent: This caught exception is deliberately ignored after selecting the availability response. The named binding violates Development Guideline 1.5, which requires `_` for unused exception parameters.

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcJwkLoadingTest.java
```diff
@@ -0,0 +1,846 @@
...
+
+        assertThat(config.signJwk().keys().size(), is(1));
+    }
+
+    @Test
+    void missingDiscoveredJwkUriCanRecover() {
+        AtomicInteger metadataRequests = new AtomicInteger();
+        AtomicInteger jwkRequests = new AtomicInteger();
+        AtomicInteger serverPort = new AtomicInteger();
+        WebServer server = WebServer.builder()
```
Verdent, line 663: These four new test servers (also at lines 660, 697, and 734) leave `host` at the WebServer default, `0.0.0.0`. That publishes the unauthenticated metadata and token routes on every interface during the test run, contrary to the test-isolation requirement that real listeners bind to loopback and port 0.

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcJwkLoadingTest.java
```diff
@@ -0,0 +1,846 @@
...
+                    .jwkRetry(singleCallRetry())
+                    .jwkCircuitBreaker(twoFailureCircuitBreaker())
+                    .build();
+
+            assertThrows(ResilientValue.UnavailableException.class, config::signJwk);
+            assertThat(config.signJwk().keys().size(), is(1));
+            assertThat(metadataRequests.get(), is(2));
+            assertThat(jwkRequests.get(), is(1));
+        } finally {
+            server.stop();
```
Verdent, line 696: Stopping the peer server does not release this test's `OidcConfig.generalWebClient()`. The test opens real HTTP connections through that client, so its protocol and pooled-connection resources survive the test; the same cleanup gap is present in the three following server-backed tests.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValue.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param loader supplier that loads the value
+     * @param retry retry handler
+     * @param circuitBreaker circuit breaker handler
+     * @param timeout timeout handler applied to each load attempt
+     * @param <T> type of the loaded value
+     * @return a new resilient value
+     * @throws IllegalArgumentException if the timeout is not positive, exceeds the retry overall timeout, or does not
+     *                                  execute on the current thread
+     */
+    static <T> ResilientValue<T> create(String description,
```
Verdent: `ResilientValue` is a public type, but this introduces a five-argument positional construction API. Rule 8.3 requires builders for parameterized construction, and there is no documented Rule 0.1 exception here. This makes future construction options require incompatible overload growth instead of extending a builder-backed contract.

common/configurable/src/main/java/io/helidon/common/configurable/ResourceUtil.java
```diff
@@ -106,6 +108,60 @@ static InputStream toIs(URI uri, Proxy proxy) {
...
+            throw new ResourceException("Failed to open stream to configured URI", e);
+        }
+    }
+
+    private static InputStream toIs(URLConnection connection, Duration timeout) throws IOException {
+        int timeoutMillis = timeoutMillis(timeout);
+        connection.setConnectTimeout(timeoutMillis);
+        connection.setReadTimeout(timeoutMillis);
+        connection.setUseCaches(false);
+        return connection.getInputStream();
```
Verdent: `HttpURLConnection.getInputStream()` throws for a 4xx/5xx response after the response connection and error stream exist. This new path then drops the `URLConnection` without consuming or closing that stream or disconnecting it; JDK error-stream buffering is disabled by default. A JWKS endpoint returning errors across retries and half-open probes can therefore leave sockets and connection slots retained until GC.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValueImpl.java
```diff
@@ -0,0 +1,160 @@
...
+            throw new IllegalArgumentException("Timeout must not exceed retry overall timeout");
+        }
+    }
+
+    private FutureTask<Outcome<T>> newAttempt() {
+        return new FutureTask<>(() -> {
+            try {
+                return new Outcome<>(load(), null);
+            } catch (RuntimeException | Error e) {
+                return new Outcome<>(null, e);
```
Verdent: `FutureTask` shares this `Outcome`, including the load leader's caller-specific interruption. If one request wins `run()` and is interrupted while another request is waiting in `get()`, both receive the same `UnavailableException`; the unaffected follower does not use the replacement attempt. Canceling the first JWKS request can therefore fail every concurrent authentication waiting on that load, despite cancellation being excluded from breaker accounting.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/CircuitBreakerImpl.java
```diff
@@ -150,13 +150,22 @@ Future<Boolean> schedule() {
...
+    }
+
     private <U> U executeTask(Supplier<? extends U> supplier) {
         try {
             U result = supplier.get();
             results.update(ResultWindow.Result.SUCCESS);
             return result;
         } catch (Throwable t) {
             Throwable throwable = SupplierHelper.unwrapThrowable(t);
+            if (callerCancelled(t)) {
```
Verdent: This bypasses the configured `applyOn` policy. With `volume(1)`, `errorRatio(100)`, and `addApplyOn(InterruptedException.class)`, an interrupted caller whose supplier throws `SupplierException(new InterruptedException(...))` leaves a closed breaker closed; in half-open state it remains half-open. `CircuitBreakerConfig.applyOn()` says these throwables are considered failures, but this path never consults `errorChecker`.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValue.java
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @return {@code true} if the value is loaded
+     */
+    boolean loaded();
+
+    /**
+     * Exception indicating that a source may become available later.
+     */
+    @Api.Internal
+    class UnavailableException extends RuntimeException {
```
Verdent: `UnavailableException` is a new public exception but extends `RuntimeException` directly. Development Guideline 1.1.1 requires a module-appropriate descendant, and this module already exposes `FaultToleranceException` as its common exception family. Callers handling fault-tolerance failures through that hierarchy will not catch resilient-value outages.

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/TenantCache.java
```diff
@@ -0,0 +1,182 @@
...
+                    continue;
+                }
+                if (resolved.isEmpty()) {
+                    return Optional.empty();
+                }
+                ResolvedTenantConfig resolvedTenant = resolved.orElseThrow();
+                tenantConfig = resolvedTenant.tenantConfig();
+                value = values.get(resolvedTenant.cacheKey())
+                        .orElseGet(() -> {
+                            Supplier<T> newValue = valueFactory.apply(tenantConfig);
```
danielkec, line 103: With global FT metrics enabled, each fresh lazy tenant created here obtains a default Retry, Timeout, and CircuitBreaker whose identity-derived names immediately register six distinct meters. Invalidation and LRU eviction discard only the supplier, so the registry retains those meters after the tenant is gone. A minimal lifecycle check is:

```text
resolve N fresh lazy tenant configs (or invalidate and reload one ID N times)
drop the returned suppliers and force GC
meter count increase = 6 * N
```

On this head, 10 discarded handler triplets retained 60 meters and 100 retained 600, so long-lived or request-derived tenant churn can grow registry memory and scrape cardinality beyond the cache bound.

danielkec (commented):
One review finding is in the inline comment.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/CircuitBreakerImpl.java
```diff
@@ -78,6 +79,8 @@ private CircuitBreakerImpl(CircuitBreakerConfig config,
         this.results = new ResultWindow(config.volume(), config.errorRatio());
         this.executor = config.executor().orElseGet(FaultTolerance.executor());
         this.errorChecker = ErrorChecker.create(config.skipOn(), config.applyOn());
+        this.callerCancellationIsFailure = !config.applyOn().isEmpty()
+                && !errorChecker.shouldSkip(new InterruptedException());
```
danielkec: The direct `InterruptedException` case from the earlier thread is fixed, but `callerCancellationIsFailure` still classifies every cancellation using a synthetic `InterruptedException`, so the actual unwrapped failure's configured policy can be inverted. With `volume(1)`, `errorRatio(100)`, the caller interrupt flag set, and a supplier throwing `SupplierException(new IOException("I/O", new InterruptedException("cancelled")))`, current-head results are:

```text
applyOn(Throwable.class) + skipOn(IOException.class) -> OPEN (expected CLOSED)
applyOn(IOException.class) -> CLOSED (expected OPEN)
```

Both the closed and half-open paths therefore bypass `errorChecker` for the actual `IOException`, which can reject later healthy calls despite `skipOn` or omit a configured failure from breaker accounting.

security/providers/oidc-common/src/main/java/io/helidon/security/providers/oidc/common/BaseBuilder.java
```diff
@@ -566,6 +869,55 @@ JwkKeys signJwk() {
...
+        Supplier<? extends Retry> supplier = jwkRetrySupplier;
+        RetryConfig prototype = jwkRetryConfig;
+        return LazyValue.create(() -> {
+            if (instance != null) {
+                return instance;
+            }
+            if (supplier != null) {
+                return Objects.requireNonNull(supplier.get());
+            }
+            return RetryConfig.builder(prototype).build();
```
danielkec: The stable default names fix the earlier default-policy cache-churn case, but an unnamed prototype accepted by `jwkRetry(RetryConfig)` is cloned here into an identity-named handler; the analogous timeout and circuit-breaker paths do the same. Each fresh dynamic tenant therefore adds six meter series which invalidation or eviction does not unregister. With global FT metrics enabled, ten triplets from the named defaults left 6 series, while ten additional unnamed-prototype triplets left 66 after GC, so customized tenant churn can still grow registry memory and scrape cardinality beyond the cache bound.

security/providers/oidc/src/test/java/io/helidon/security/providers/oidc/HostHeaderTenantDiscoveryTest.java
```diff
@@ -357,17 +879,46 @@ public void close() {
 
     private static final class MockIdpServer implements AutoCloseable {
         private final AtomicInteger wellKnownHits = new AtomicInteger();
+        private final AtomicInteger jwkHits = new AtomicInteger();
+        private final CountDownLatch metadataRequest = new CountDownLatch(1);
         private final JsonObject[] metadataHolder = new JsonObject[1];
+        private volatile boolean metadataAvailable = true;
+        private volatile boolean jwkAvailable = true;
+        private volatile CountDownLatch metadataResponseGate;
```
danielkec, line 889: Within `MockIdpServer`, these three new non-final instance fields are declared before the existing final `server` and `identityUri` fields. Development Guideline 2.1 orders final instance fields before non-final instance fields, including in nested test types, so this added layout does not follow the repository member order.

danielkec (commented):
Three follow-up review findings are in the inline comments.

danielkec (dismissed):

security/providers/oidc/src/main/java/io/helidon/security/providers/oidc/OidcProvider.java
```diff
@@ -102,10 +108,12 @@ private OidcProvider(Builder builder, OidcOutboundConfig oidcOutboundConfig) {
         this.jwtGroupsSeparator = builder.jwtGroupsSeparator;
         this.outboundConfig = oidcOutboundConfig;
 
-        tenantConfigFinders = List.copyOf(builder.tenantConfigFinders);
         tenantIdFinders = List.copyOf(builder.tenantIdFinders);
 
-        tenantConfigFinders.forEach(tenantConfigFinder -> tenantConfigFinder.onChange(tenantAuthHandlers::remove));
+        oidcConfig.validateForAuthentication();
```
Verdent: This also validates inbound signing keys when the provider is used exclusively through `Security.builder().addOutboundSecurityProvider(...)`. A client-credentials configuration with client ID/secret, explicit identity and token endpoints, and `oidcMetadataWellKnown(false)` now fails construction with `A signing JWK source is required when JWT validation with JWK is enabled`. The same construction succeeds with the base provider implementation. Outbound client-credentials exchange and token propagation do not verify inbound JWT signatures, so this prevents an existing outbound-only setup from starting.

fault-tolerance/fault-tolerance/src/main/java/io/helidon/faulttolerance/ResilientValueImpl.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    private T load() {
+        T loadedValue;
+        try {
+            loadedValue = circuitBreaker.invoke(this::loadWithRetry);
+        } catch (CircuitBreakerOpenException e) {
+            throw new ResilientValue.UnavailableException(description + " is temporarily unavailable", e);
+        } catch (SupplierException e) {
+            if (Thread.currentThread().isInterrupted() && SupplierHelper.interrupted(e) != null) {
```
Verdent, line 135: A `SupplierException` can wrap an `Error` here, so the interrupt check also classifies a non-cancellation load failure as caller cancellation. With real Retry/CircuitBreaker/current-thread Timeout handlers, a loader that restores its interrupt flag and throws `AssertionError(new InterruptedException(...))` causes the leader to receive `UnavailableException` and an unaffected waiting follower to retry and succeed (`loads=2`, `follower=recovered`). The `interruptedErrorFromLoadLeaderIsShared` test bypasses this path by using a custom breaker that throws the Error directly; the actual handler composition does not preserve that sharing behavior.

security/providers/oidc-common/src/test/java/io/helidon/security/providers/oidc/common/OidcJwkLoadingTest.java
```diff
@@ -0,0 +1,903 @@
...
+
+    @Test
+    void missingDiscoveredJwkUriCanRecover() {
+        AtomicInteger metadataRequests = new AtomicInteger();
+        AtomicInteger jwkRequests = new AtomicInteger();
+        AtomicInteger serverPort = new AtomicInteger();
+        WebServer server = WebServer.builder()
+                .host("localhost")
+                .routing(routing -> routing
+                        .get("/identity/.well-known/openid-configuration", (req, res) -> {
```
Verdent: The `req` parameter is unused in this new handler and the adjacent `/jwks` handler, with the same pattern in the timeout tests below. These deliberately unused lambda bindings violate Development Guideline 1.5, which requires `_`.

security/providers/oidc/src/test/java/io/helidon/security/providers/oidc/HostHeaderTenantDiscoveryTest.java
```diff
@@ -357,17 +879,46 @@ public void close() {
...
+                                        throw new IllegalStateException("Interrupted while blocking metadata response", e);
+                                    }
+                                }
+                                if (metadataAvailable) {
+                                    res.send(metadataHolder[0]);
+                                } else {
+                                    res.status(Status.SERVICE_UNAVAILABLE_503).send();
+                                }
+                            })
+                            .get("/jwks", (req, res) -> {
```
Verdent: The new `/jwks` handler never reads `req`; its body only updates the counter and writes the response. This deliberately unused lambda binding violates Development Guideline 1.5.

danielkec (approved):
Reviewed the current changes and follow-up fixes. Focused clean validation passed on the PR head and against current main; no outstanding findings.

Verdent (approved):

Issue #11724, opened 2026-04-21
Declarative API stability


tomas-langer:
Helidon Version: 4.x

## Task Description
Declarative APIs must have the modules marked as production (i.e. no Features.Preview or Features.Incubating).

The APIs will be declared as `Api.Preview`, as we plan to keep them, and we will not remove them.

In addition, within 4.x. branch, we will not do backward incompatible API changes in declarative modules, except for possible bugfixes. We still have the possibility to do backward incompatible changes between 4.x and 27 if we need to.

Part of #11565

PR #12523, opened 2026-09-08
Remove declarative ignore-incubating option (27.x)

tomas-langer:
Removes the Declarative startup warning and its `declarative.ignore-incubating` suppression option.

The primitive configuration-expression regression remains covered under a neutral test key, and unrelated test configurations no longer carry the removed setting.

### DIFF
--- a/docs/modules/injection/declarative.md
+++ b/docs/modules/injection/declarative.md
@@ -19,9 +19,9 @@ Our declarative approach has the following advantages:
   not require additional dependencies)
 
 > [!NOTE]
-> Helidon Declarative is a preview feature. The APIs shown here are subject to
-> change between major releases without deprecation, but they are intended for
-> supported external use.
+> Helidon Declarative is a preview feature. It is ready for production use. Its
+> APIs will remain backward compatible within a major version, but may change
+> without the usual deprecation process in a new major version.
 
 ## Usage
 
--- a/docs/modules/injection/declarative.md
+++ b/docs/modules/injection/declarative.md
@@ -19,8 +19,9 @@ Our declarative approach has the following advantages:
   not require additional dependencies)
 
 > [!NOTE]
-> Helidon Declarative is an incubating feature. The APIs shown here are subject
-> to change. These APIs will be finalized in a future release of Helidon.
+> Helidon Declarative is a preview feature. The APIs shown here are subject to
+> change between major releases without deprecation, but they are intended for
+> supported external use.
 
 ## Usage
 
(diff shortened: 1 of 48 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/WebServerStarterService.java
```diff
@@ -16,33 +16,20 @@
...
-                            boolean ignoreIncubating) {
+    WebServerStarterService(LoomServer server) {
         this.server = server;
-        this.ignoreIncubating = ignoreIncubating;
     }
 
     @Service.PostConstruct
     void postConstruct() {
         server.start();
-        if (!ignoreIncubating) {
```
danielkec, line 40: When a 27.x application starts WebServer through the service registry without `declarative.ignore-incubating`, this removes the runtime notice even though `HttpEntryPoint` and `RestServer` on this branch are still marked `@Api.Incubating`, so users can adopt APIs that may change incompatibly or be removed without seeing the warning. On 4.x, #11727 promoted these APIs and the Declarative documentation to Preview before #11858 removed the warning, so this forward-port omits that prerequisite. A minimal case is a registry-managed server configured only with `server.port=0`: the target branch logs the incubating warning, while this head starts silently.

docs/modules/injection/declarative.md
```diff
@@ -19,8 +19,9 @@ Our declarative approach has the following advantages:
   not require additional dependencies)
 
 > [!NOTE]
-> Helidon Declarative is an incubating feature. The APIs shown here are subject
-> to change. These APIs will be finalized in a future release of Helidon.
+> Helidon Declarative is a preview feature. It is ready for production use. Its
+> APIs will remain backward compatible within a major version, but may change
```
danielkec, line 23: `@Api.Preview` and `@Features.Preview` on this branch allow backward-incompatible changes between minor releases, while this text guarantees compatibility throughout 27.x. #11724 documented that stronger guarantee specifically for 4.x and allowed an incompatible 4.x-to-27 transition; is the same stronger guarantee intended for all 27.x Declarative and Data APIs? As written, 27.x users also receive compiler diagnostics that Preview APIs may change between minor versions, so the two contracts give conflicting upgrade guidance.

tomas-langer: This is giving a stronger guarantee that the annotation. This is correct. We will also update the warnings, as we are aligning Preview with how we treat it for declarative

danielkec (approved):

PR #12524, opened 2026-09-08
Add sealed support for prototype generated interfaces

### LINKED ISSUE #12520: Add sealed support for prototype generated interfaces
Helidon Version: 27.x

## Enhancement Description

Add a `@Prototype.Sealed` marker annotation that causes a prototype blueprint to generate its public prototype API as a Java sealed interface.

### Problem

A prototype blueprint currently generates an ordinary public interface:

```java
import io.helidon.builder.api.Prototype;

@Prototype.Blueprint
interface RetryConfigBlueprint {
    int maxRetries();
}
```

Conceptually, the generated API is:

```java
public interface RetryConfig extends RetryConfigBlueprint, Prototype.Api {
}
```

Although Helidon generates and controls the implementation, external code can implement RetryConfig directly. Such implementations can bypass generated validation, equality, hash code, string representation, and future generated behavior.

Some prototype APIs are intended to be closed value types whose instances can only be created through their generated builders. The blueprint author cannot currently express this because the public interface is generated.

### Proposed API

```java
Add a new marker annotation to Prototype:

/**
 * Generate the prototype API as a sealed interface.
 * <p>
 * This annotation must be used together with {@link Prototype.Blueprint}.
 */
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.CLASS)
public @interface Sealed {
}
```

A blueprint opts in explicitly:

```java
import io.helidon.builder.api.Prototype;

@Prototype.Blueprint
@Prototype.Sealed
interface RetryConfigBlueprint {
    int maxRetries();
}
```

Using a separate marker makes this significant extensibility decision visible at the blueprint declaration and avoids adding another flag to `@Prototype.Blueprint`.

### Generated API

The generated API would conceptually become:

```java
public sealed interface RetryConfig
        extends RetryConfigBlueprint, Prototype.Api
        permits RetryConfig.BuilderBase.RetryConfigImpl {

    abstract class BuilderBase<BUILDER, PROTOTYPE> {

        protected static final class RetryConfigImpl
                implements RetryConfig {
            // Generated implementation
        }
    }
}
```

Calling code remains unchanged:

```java
RetryConfig config = RetryConfig.builder()
        .maxRetries(3)
        .build();
```

A direct external implementation would no longer compile:

```java
// Compilation error: RetryConfig is sealed.
final class CustomRetryConfig implements RetryConfig {
}
```

### Scope

For a blueprint annotated with both `@Prototype.Blueprint` and `@Prototype.Sealed`, code generation should:

  - Generate the prototype API as a sealed interface.
  - Permit the generator-owned prototype implementation.
  - Generate that implementation as final.
  - Continue allowing custom subclasses of BuilderBase; custom builders still produce the generator-owned implementation.
  - Support configuration, factory prototypes, generics, detach, and generated factory methods.
  - Report a clear code-generation error for incompatible prototype inheritance.

`@Prototype.Sealed` should not be inherited implicitly. Each generated prototype must explicitly opt into sealing.

### Prototype inheritance

A sealed prototype should initially be treated as a leaf prototype.

A generated child prototype normally extends its parent prototype interface directly. Such a child cannot extend a sealed parent unless it is included in the parent’s permitted-subtype list, but the parent processor cannot reliably discover future children from other modules.

The initial enhancement should therefore reject known attempts to extend a sealed prototype with an actionable diagnostic:

Prototype ChildConfig cannot extend sealed prototype ParentConfig. Sealed prototypes must be leaf prototypes.

Support for an explicit user-defined permits list can be considered separately.

### Compatibility

Existing blueprints remain unchanged because sealing is opt-in.

Adding @Prototype.Sealed to an existing public blueprint is a source-compatibility change for consumers that implement the generated interface. This should be documented and normally limited to new APIs or major-version changes.

Removing `@Prototype.Sealed` is not expected to break callers, but it broadens the set of permitted implementations.

tvallin:
### Description

Fixes #12520

### Documentation

Updated as part of this PR.


builder/api/src/main/java/io/helidon/builder/api/Prototype.java
```diff
@@ -172,6 +172,16 @@ public interface Factory<T> {
         boolean detach() default false;
     }
 
+    /**
+     * Generate the prototype API as a sealed interface.
```
tomas-langer, line 176: I think there should be more here, as this introduces the following limitations:

- the generated prototype will not be extensible
- the implementation class will not be protected (must be private final)

builder/codegen/src/main/java/io/helidon/builder/codegen/BuilderCodegen.java
```diff
@@ -657,6 +657,11 @@ private void generatePrototype(RoundContext ctx,
                                                  prototype))
                 .accessModifier(prototypeInfo.accessModifier());
 
+        if (blueprint.hasAnnotation(Types.PROTOTYPE_SEALED)) {
+            classModel.isSealed(true)
+                    .addPermittedSubclass(prototype.fqName() + ".BuilderBase." + ifaceName + "Impl");
```
tomas-langer: We should have the implementation `TypeName` available somewhere, it should not be constructed more than once. If it is not available, we should make it available.

codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ClassBase.java
```diff
@@ -501,6 +556,17 @@ public B isAbstract(boolean isAbstract) {
             return identity();
         }
 
+        /**
+         * Whether this type is sealed.
+         *
+         * @param isSealed whether this type is sealed
+         * @return updated builder instance
+         */
+        public B isSealed(boolean isSealed) {
```
tomas-langer: rename to `sealed` (all the accessors). The `is` is redundant here

builder/codegen/src/main/java/io/helidon/builder/codegen/FactoryPrototypeInfo.java
```diff
@@ -629,6 +630,19 @@ private static Optional<TypeInfo> superBlueprintDefinition(RoundContext ctx, Typ
         return superBlueprintDefinition(ctx, blueprint, new HashSet<>());
     }
 
+    private static void validateSealedSuperPrototype(RoundContext ctx, TypeInfo blueprint) {
```
tomas-langer, line 628: `validateSealedSuperPrototype` only walks the blueprint's Java supertype hierarchy, while `@Prototype.Implement` entries are added separately to the generated interface's supertype set. A blueprint using `@Prototype.Implement("com.acme.ParentConfig")` can therefore make its generated prototype extend a sealed generated parent without this validation seeing it. Compilation then fails against generated source with javac's sealed-type error instead of the actionable leaf-prototype `CodegenException` required by #12520. This leaves one supported supertype-declaration path outside the new validation and regression coverage.

builder/README.md
```diff
@@ -227,6 +227,7 @@ Annotations:
 | Annotation                           | Required | Description                                                                                                                                                        |
 |--------------------------------------|----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `Prototype.Blueprint`                | `true`   | Annotation on the blueprint interface is required to trigger annotation processing                                                                                 |
+| `Prototype.Sealed`                   | `false`  | Generate a sealed prototype interface whose only permitted implementation is the generated implementation                                                          |
```
tomas-langer, line 230: The PR adds the new annotation and compatibility guidance only to this repository README, while the published v27 Builder guide at `docs/modules/builder.md` still omits `Prototype.Sealed` from its annotation table and contains no leaf or compatibility guidance. Users following the product documentation cannot discover the opt-in or its source and binary compatibility impact, despite the PR's documentation claim.

codegen/class-model/src/main/java/io/helidon/codegen/classmodel/ClassBase.java
tvallin: I spotted it as well and I decided to keep it because of consistency with other accessors (I will update them). Is it something to add to the dev-guideline maybe ?

tomas-langer (approved):

PR #12525, opened 2026-09-08
Drain direct HTTP/1 error responses before close (27.x)

### LINKED ISSUE #12428: Drain direct HTTP/1 error responses before closing the connection
## Description

Direct error handling queues an asynchronous response and immediately returns to connection closure, whose writer can discard queued buffers. The response should be drained or completion-coordinated before close; otherwise clients can receive EOF or reset instead of the complete 4xx or 5xx response.

tomas-langer:
Resolves #12428

Flush direct HTTP/1 error responses before connection shutdown so queued buffers are sent. Add deterministic regression coverage for the response-drain ordering.

### DIFF
--- a/webserver/webserver/src/test/java/io/helidon/webserver/http1/Http1ConnectionTest.java
+++ b/webserver/webserver/src/test/java/io/helidon/webserver/http1/Http1ConnectionTest.java
@@ -26,6 +26,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import io.helidon.common.buffers.BufferData;
 import io.helidon.common.buffers.DataReader;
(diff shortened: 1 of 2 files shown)


webserver/webserver/src/test/java/io/helidon/webserver/http1/Http1ConnectionTest.java
```diff
@@ -117,6 +117,43 @@ void rejectedGetStillSendsHandlerEntity() throws InterruptedException {
...
+                                                      writer,
+                                                      directHandlers,
+                                                      Router.empty());
+        try {
+            Future<?> connectionTask = executor.submit(() -> {
+                connection.handle(limit);
+                return null;
+            });
+
+            assertThat("Direct error response did not reach the final flush",
```
danielkec, line 145: This test currently proves only that `flush()` starts before `handle()` returns, not that the response was queued before that flush. As a minimal mutation, moving `flushBeforeClose()` above `writer.write(buffer)` in `handleRequestException` still passes because `BlockingDataWriter.write(...)` records nothing, while `flush()` trips the latch and blocks until release. With a real asynchronous writer, the response would then be queued after the drain and could be discarded when `ConnectionHandler` closes the writer, reproducing the missing error response.

danielkec (approved):

Issue #11500, opened 2026-04-01
Annotate Helidon public APIs with API Stability Annotations


tomas-langer:
Follow-up of #11124 and #11184.

This issue tracks the remaining work to annotate Helidon public APIs with API stability annotations on `main`.

## Scope and completion criteria

All current production JAR modules are in scope unless they are explicitly retired or excluded below. Aggregator, bundle, application, test, and testing-support modules are not part of this checklist.

A module is complete when:

- every public top-level production type, including generated-source templates, declares exactly one of `@Api.Stable`, `@Api.Preview`, `@Api.Incubating`, or `@Api.Internal`; and
- its normal compile config runs `helidon-codegen-api-stability-enforcer`, so future unannotated public top-level types fail the build.

`helidon-codegen-api-stability` is a mechanism exception to the second criterion because annotation processing cannot enforce that processor while it is being built. `helidon-codegen-api-stability-enforcer` is an exception to both criteria: it deliberately has no Helidon dependencies and compiles with annotation processing disabled.

The current `main` branch is the source of truth. Modules added later are automatically in scope even when they are not named below.

## Completed

- [x] `helidon-codegen-api-stability` — API stability processor introduced by #11184
- [x] `helidon-common-mapper` — annotated and enforced by #11184
- [x] `helidon-common` — all current public top-level types are annotated and the enforcer runs during normal compilation

`helidon-codegen-api-stability-enforcer` is the enforcement mechanism itself and is covered by the mechanism exception above.

## Fully annotated; enforcement or an explicit mechanism exception still pending

- [ ] `helidon-codegen-apt` — 1/1 current public top-level type annotated by #11802
- [ ] `helidon-config-metadata-model` — 3/3 current public top-level types annotated by #11802
- [ ] `helidon-json-binding` — 12/12 current public top-level types annotated by #11766

These remain open until their normal compile configuration enforces the annotations or a necessary mechanism exception is documented.

## Current focused slice

- [ ] `helidon-http-encoding-gzip` — PR #12380
- [ ] `helidon-http-encoding-deflate` — PR #12380

Each module has two public top-level types. The encoding and provider types are stable APIs; the ServiceLoader-only public provider constructors remain internal as established by #11758.

## Partial progress already merged

- #11758 marked ServiceLoader-only public constructors internal across the affected modules, but did not classify their containing top-level types.
[...]

PR #12527, opened 2026-09-08
Annotate Common APIs with API stability (27.x)

tomas-langer:
Part of #11500

Annotates public APIs across all deployed Common modules with `Stable`, `Preview`, or `Internal` status and enables API-stability enforcement during normal compilation. This includes deployed Common testing artifacts and preserves generator-specific compiler configuration.

Marks `Bytes` as Preview and suppresses the resulting diagnostic in JSON-generated converters, so Helidon and 27.x examples compile without `Bytes` API-stability warnings.


common/buffers/src/main/java/io/helidon/common/buffers/DataReader.java
```diff
@@ -23,9 +23,12 @@
 import java.util.List;
 import java.util.function.Supplier;
 
+import io.helidon.common.Api;
+
 /**
  * Data reader that can pull additional data.
  */
+@Api.Stable
```
danielkec, line 31: When `DataReader.readLazyString(...)` is used from a normal consumer, the newly Stable `DataReader` exposes the newly Internal `LazyString`, so compilation with the default API-stability policy fails. A minimal case is:

```java
LazyString value = DataReader.create(() -> "x".getBytes(StandardCharsets.UTF_8))
        .readLazyString(StandardCharsets.UTF_8, 1);
```

The processor reports `LazyString is internal API` for the declared return type, making this part of the Stable surface unusable under normal enforcement.

json/codegen/src/main/java/io/helidon/json/codegen/JsonConverterGenerator.java
```diff
@@ -84,6 +85,10 @@ static void generateConverter(ClassBase.Builder<?, ?> classBuilder, ConvertedTyp
         Map<String, TypeToConfigure> toConfigure = new HashMap<>();
         classBuilder.name(converterInfo.converterType().className())
                 .addInterface(converterInterfaceType)
+                .addAnnotation(Annotation.builder()
+                                       .type(SuppressWarnings.class)
+                                       .addParameter("value", Api.SUPPRESS_PREVIEW)
+                                       .build())
```
danielkec: When the source entity is generic, this suppression is emitted on the private nested converter, while the `Bytes` import belongs to the unsuppressed outer `_BindingFactory`. A minimal trigger is:

```java
@Json.Entity
public class Container<T> {
    private T value;
    public T getValue() { return value; }
    public void setValue(T value) { this.value = value; }
}
```

Compiling the generated `Container_BindingFactory.java` with `ApiStabilityProcessor` still reports `io.helidon.common.buffers.Bytes is preview API` at the import, so this generated-code path remains unsuppressed.

tomas-langer: generated code must suppress all Helidon API stability warnings, not just preview.

danielkec (approved):

PR #12529, opened 2026-09-08
Avoid service-factory activation and registry-shutdown lock inversion (27.x)

### LINKED ISSUE #12471: Avoid service-factory activation and registry-shutdown lock inversion
## Description

Registry shutdown holds the scope write lock while deactivating services, while factory activation holds an instance lock while invoking user factory lookup. Lock ordering should not call user lookup under the opposing lock; otherwise activation and shutdown can deadlock indefinitely.

tomas-langer:
Resolves #12471

Snapshot services under the scoped-registry lock, then release it before lifecycle callbacks. Preserve access to already-created services during shutdown, reject new activation, synchronize scope activation with shutdown state, document the deactivation concurrency constraint, and cover the lock inversion deterministically.

### DIFF
--- a/service/registry/src/main/java/io/helidon/service/registry/Activators.java
+++ b/service/registry/src/main/java/io/helidon/service/registry/Activators.java
@@ -168,6 +168,7 @@ final class Activators {
 
         volatile ActivationPhase currentPhase = ActivationPhase.INIT;
         private boolean deactivationRequested;
+        private boolean postConstructCompleted;
         private boolean activationCleanupPending;
         private ScopedRegistryImpl scopedRegistry;
 
@@ -398,9 +399,9 @@ final class Activators {
             if (!deactivationRequested && (scopedRegistry == null || scopedRegistry.activationAllowed())) {
                 return false;
             }
-            if (currentPhase.eligibleForDeactivation()) {
-                // Activation stopped at a callback boundary. Preserve its phase so queued deactivation can clean up
-                // after the activating thread releases the instance lock, without publishing the instance.
+            if (postConstructCompleted && currentPhase.eligibleForDeactivation()) {
+                // Preserve queued cleanup only after post construct returned normally. Earlier cancellation must not
+                // invoke pre destroy on an instance whose lifecycle initialization is incomplete.
                 activationCleanupPending = true;
             } else {
                 currentPhase = ActivationPhase.DESTROYED;
@@ -468,6 +469,7 @@ final class Activators {
                     return response.build();
                 }
                 postConstruct(response);
+                postConstructCompleted = true;
                 if (activationInterrupted(response)) {
                     return response.build();
                 }
--- a/service/registry/src/main/java/io/helidon/service/registry/Activators.java
+++ b/service/registry/src/main/java/io/helidon/service/registry/Activators.java
@@ -168,6 +168,7 @@ final class Activators {
 
         volatile ActivationPhase currentPhase = ActivationPhase.INIT;
         private boolean deactivationRequested;
+        private boolean activationCleanupPending;
         private ScopedRegistryImpl scopedRegistry;
 
         BaseActivator(ServiceProvider<T> provider, DependencyContext dependencyContext) {
@@ -261,9 +262,9 @@ final class Activators {
                 if (phase.ordinal() < ActivationPhase.ACTIVATION_FINISHING.ordinal()) {
                     deactivationRequested = true;
                 }
-                if (phase == ActivationPhase.CONSTRUCTING
+                if (!activationCleanupPending && (phase == ActivationPhase.CONSTRUCTING
                         || phase == ActivationPhase.INJECTING
-                        || phase == ActivationPhase.POST_CONSTRUCTING) {
+                        || phase == ActivationPhase.POST_CONSTRUCTING)) {
                     // A JVM shutdown hook can run while the activating thread waits for shutdown to complete. Taking the
                     // instance lock here would deadlock, so interrupt activation and let the activating thread terminate.
                     var response = ActivationResult.builder()
@@ -397,8 +398,14 @@ final class Activators {
             if (!deactivationRequested && (scopedRegistry == null || scopedRegistry.activationAllowed())) {
                 return false;
             }
-            currentPhase = ActivationPhase.DESTROYED;
-            response.finishingActivationPhase(ActivationPhase.DESTROYED)
+            if (currentPhase.eligibleForDeactivation()) {
+                // Activation stopped at a callback boundary. Preserve its phase so queued deactivation can clean up
+                // after the activating thread releases the instance lock, without publishing the instance.
+                activationCleanupPending = true;
+            } else {
+                currentPhase = ActivationPhase.DESTROYED;
+            }
+            response.finishingActivationPhase(currentPhase)
                     .success(false)
                     .error(new IllegalStateException("Activation interrupted by deactivation request"));
             return true;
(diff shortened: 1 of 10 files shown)


service/registry/src/main/java/io/helidon/service/registry/ScopedRegistryImpl.java
```diff
@@ -115,30 +128,33 @@ public void deactivate() {
...
 
     @SuppressWarnings("unchecked")
     @Override
     public <T> Activator<T> activator(ServiceInfo descriptor, Supplier<Activator<T>> activatorSupplier) {
         try {
             serviceProvidersLock.readLock().lock();
-            checkActive();
             Activator<?> activator = activators.get(descriptor);
-            if (activator != null) {
+            if (activator != null && state != RegistryState.INACTIVE) {
```
danielkec: [P1] An existing `INIT` activator can start after the shutdown snapshot

When a lookup is preempted between `ServiceManager.activator()` and `Activator.instances()`, this branch returns its already-mapped `INIT` activator during `DEACTIVATING`, even though the shutdown snapshot omitted it because `INIT` is not eligible for deactivation. The resumed lookup activates the service after the snapshot, so shutdown reaches `INACTIVE` without invoking that service's `@PreDestroy`; a latch probe produced:

```text
merge base: lookup blocked, then ScopeNotActiveException
this head: phase after shutdown=ACTIVE, PreDestroy calls=0
```

service/registry/src/main/java/io/helidon/service/registry/Activators.java
```diff
@@ -354,11 +368,46 @@ private void stateTransitionStart(ActivationResult.Builder res, ActivationPhase
...
+        }
+
+        private boolean activationInterruptedLocked(ActivationResult.Builder response) {
+            if (currentPhase == ActivationPhase.ACTIVE) {
+                return false;
+            }
+            if (!deactivationRequested && (scopedRegistry == null || scopedRegistry.activationAllowed())) {
+                return false;
+            }
+            currentPhase = ActivationPhase.DESTROYED;
```
danielkec: [P1] Interrupted activation becomes terminal without cleanup

When a lower-run-level service finishes `postConstruct` after shutdown has snapshotted it, but while shutdown is still deactivating a higher-run-level service, this assignment makes the lower service terminal before its queued `deactivate()` call. That call then treats `DESTROYED` as ineligible and skips `preDestroy`, so resources created during activation remain open. A latch-controlled reproducer gives:

```text
merge base: phase=DESTROYED, PreDestroy calls=1
this head: phase=DESTROYED, PreDestroy calls=0
```

service/registry/src/main/java/io/helidon/service/registry/ScopedRegistryImpl.java
```diff
@@ -173,8 +197,30 @@ private static Comparator<? super Activator<?>> shutdownComparator() {
...
     }
+
+    private Activator<?> scopedActivator(Activator<?> activator) {
+        if (activator instanceof Activators.BaseActivator<?> baseActivator) {
+            baseActivator.scopedRegistry(this);
+        }
+        return activator;
+    }
+
+    boolean activationAllowed() {
```
danielkec: [P2] This new package-private instance method is declared after the private `shutdownComparator`, `checkActive`, and `scopedActivator` helpers. `DEV-GUIDELINES.md` Rule 2.1 requires non-private instance methods to precede private static and private instance methods, so this declaration order violates the project's mandatory member-order rule.

service/registry/src/main/java/io/helidon/service/registry/Activators.java
```diff
@@ -354,11 +369,52 @@ private void stateTransitionStart(ActivationResult.Builder res, ActivationPhase
...
+        }
+
+        private boolean activationInterruptedLocked(ActivationResult.Builder response) {
+            if (currentPhase == ActivationPhase.ACTIVE) {
+                return false;
+            }
+            if (!deactivationRequested && (scopedRegistry == null || scopedRegistry.activationAllowed())) {
+                return false;
+            }
+            if (currentPhase.eligibleForDeactivation()) {
```
danielkec: [P1] Queued cleanup can run before post-construct

When shutdown snapshots a lower-run-level activator in `CONSTRUCTING` or `INJECTING` and is delayed by another service, this branch marks cleanup pending as soon as that callback returns. Queued `deactivate()` then invokes `preDestroy` although `postConstruct` never ran, so cleanup can access uninitialized lifecycle state. A latch-controlled construction probe gives:

```text
merge base: PostConstruct=1, PreDestroy=1, shutdown succeeds
this head:  PostConstruct=0, PreDestroy=1, shutdown fails
```

The generated gauge registrar demonstrates the consequence: its `meterRegistry` field is assigned only in `postConstruct`, so this ordering makes its `preDestroy` throw `NullPointerException`.

service/registry/src/main/java/io/helidon/service/registry/ScopedRegistryImpl.java
```diff
@@ -115,18 +132,21 @@ public void deactivate() {
...
-                return;
-            }
-            ServiceRegistryException failure = new ServiceRegistryException("Deactivation failed");
-            exceptions.forEach(failure::addSuppressed);
-            throw failure;
         } finally {
-            serviceProvidersLock.writeLock().unlock();
+            try {
+                serviceProvidersLock.writeLock().lock();
+                state = RegistryState.INACTIVE;
```
danielkec, line 138: [P2] Fatal cleanup makes the remaining services permanently unreachable

When one `preDestroy` throws an `Error`, the loop exits because it catches only `Exception`, but this `finally` marks the registry `INACTIVE`. Lower-run-level services remain `ACTIVE` and later `deactivate()` calls return immediately, so their cleanup cannot be retried. A callback that throws `AssertionError` once gives:

```text
merge base retry: fatal=2/DESTROYED, lower=1/DESTROYED
this head retry:  fatal=1/PRE_DESTROYING, lower=0/ACTIVE
```

tomas-langer: Error is a terminal shutdown failure here: it propagates immediately, stops remaining callbacks, and leaves the scope inactive. We do not support retrying lifecycle cleanup after an Error. The previous implementation happened to allow a later call to retry, but that behavior was not a lifecycle guarantee. INACTIVE means the scope is no longer usable; it does not assert that every cleanup callback succeeded.

danielkec (approved):

service/registry/src/main/java/io/helidon/service/registry/ScopedRegistryImpl.java
```diff
@@ -158,13 +178,21 @@ public <T> Activator<T> activator(ServiceInfo descriptor, Supplier<Activator<T>>
     <T> Optional<Activator<T>> existingActivator(ServiceInfo descriptor) {
         try {
             serviceProvidersLock.readLock().lock();
+            Activator<?> activator = activators.get(descriptor);
+            if (activator != null && availableForActiveLookup(activator)) {
```
Verdent, line 182: An explicitly registered instance still becomes invisible to `firstActive()` and `allActive()` during shutdown, even when its existing activator is `ACTIVE`. `ServiceManager.activeInstances()` gets that activator here, but its fixed-instance path then calls `activator()` again to retain it; the new `DEACTIVATING` state makes that call throw `ScopeNotActiveException`, and the successful lookup is replaced with an empty result.

A public-API probe using `ServiceRegistryConfig.builder().putServiceInstance(...)`, a prior successful lookup, and a paused higher-run-level cleanup callback gives:

```text
before shutdown: fixed=true, managed=true
during shutdown: fixed=false, allFixed=0, managed=true
```

The same active lookup during a base-version cleanup callback returns the fixed instance. Shutdown-time factory lookups can therefore lose an already-bound dependency even while ordinary active services remain available.

tomas-langer: I honestly do not care that this happens during shutdown. Services are not allowed to lookup services from service registry in PreDestroy

PR #12532, opened 2026-09-08
Fix HTTP parsing and redirect handling (4.x)

### LINKED ISSUE #12526: Fix HTTP parsing and redirect handling (4.x)
Backport #12505 to the `helidon-4.x` branch.

The source PR fixes HTTP parsing and redirect correctness, including:

- parsing and serializing media types according to RFC 9110;
- preserving HTTP field-value octets with a one-to-one ISO-8859-1 mapping instead of lossy US-ASCII decoding;
- propagating HTTP/2 parser mode and validating header blocks before HPACK mutation; and
- correcting method-preserving redirects and HTTP/1 connection cleanup.

Keep the backport source-faithful to merge commit `d421a3c9a84ee561c301f32b523c5d6b2b65a99b`.

tomas-langer:
Resolves #12526

Backports #12505 to Helidon 4.x, preserving HTTP field-value octets and aligning HTTP parsing and redirect handling with main.

Includes 4.x-specific adaptations for early final responses and redirect-response lifecycle management.

Follow up issue: https://github.com/helidon-io/helidon/issues/12533 - please do not report findings mentioned in the follow up issue

### DIFF
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -256,24 +256,27 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
                                           DataReader reader) {
 
         Status responseStatus;
-        try {
-            responseStatus = Http1StatusParser.readStatus(reader, protocolConfig.maxStatusLineLength());
-        } catch (UncheckedIOException e) {
-            // if we get a timeout or connection close, we must close the resource (as otherwise we may receive
-            // data of this request on the next use of this connection
+        ClientResponseHeaders responseHeaders;
+        do {
             try {
-                connection.closeResource();
-            } catch (Exception ex) {
-                e.addSuppressed(ex);
+                responseStatus = Http1StatusParser.readStatus(reader, protocolConfig.maxStatusLineLength());
+            } catch (UncheckedIOException e) {
+                // if we get a timeout or connection close, we must close the resource (as otherwise we may receive
+                // data of this request on the next use of this connection
+                try {
+                    connection.closeResource();
+                } catch (Exception ex) {
+                    e.addSuppressed(ex);
+                }
+                throw e;
             }
-            throw e;
-        }
 
-        recvListener.status(connection.helidonSocket(), responseStatus);
+            recvListener.status(connection.helidonSocket(), responseStatus);
 
-        ClientResponseHeaders responseHeaders = readHeaders(reader);
+            responseHeaders = readHeaders(reader);
 
-        recvListener.headers(connection.helidonSocket(), responseHeaders);
+            recvListener.headers(connection.helidonSocket(), responseHeaders);
+        } while (originalRequest.outputStreamRedirect() && isPreContinueInterimResponse(responseStatus));
 
         return createServiceResponse(http1Client,
                                      serviceRequest,
@@ -290,6 +293,12 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
 
     Http1ConnectionListener recvListener() {
         return recvListener;
+    }
+
+    private static boolean isPreContinueInterimResponse(Status responseStatus) {
+        return responseStatus.family() == Status.Family.INFORMATIONAL
+                && responseStatus.code() != Status.CONTINUE_100.code()
+                && responseStatus.code() != Status.SWITCHING_PROTOCOLS_101.code();
     }
 
     private static String requestTarget(ClientUri uri) {
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -316,8 +316,10 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
         if (responseHeaders.contains(HeaderValues.CONTENT_LENGTH_ZERO)) {
             return false;
         }
-        // Why is NOT_MODIFIED_304 not added here too?
-        if (responseStatus.code() == Status.NO_CONTENT_204.code()) {
+        int statusCode = responseStatus.code();
+        if (statusCode == Status.NO_CONTENT_204_CODE
+                || statusCode == Status.RESET_CONTENT_205_CODE
+                || statusCode == Status.NOT_MODIFIED_304_CODE) {
             return false;
         }
         if ((
--- a/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
+++ b/webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
@@ -149,19 +149,30 @@ abstract class Http1CallChainBase implements WebClientService.Chain {
     public WebClientServiceResponse proceed(WebClientServiceRequest serviceRequest) {
         // either use the explicit connection, or obtain one (keep alive or one-off)
         effectiveConnection = connection == null ? obtainConnection(serviceRequest) : connection;
-        effectiveConnection.readTimeout(this.timeout);
+        try {
+            effectiveConnection.readTimeout(this.timeout);
 
-        DataWriter writer = effectiveConnection.writer();
-        DataReader reader = effectiveConnection.reader();
-        ClientUri uri = serviceRequest.uri();
-        ClientRequestHeaders headers = serviceRequest.headers();
+            DataWriter writer = effectiveConnection.writer();
+            DataReader reader = effectiveConnection.reader();
+            ClientUri uri = serviceRequest.uri();
+            ClientRequestHeaders headers = serviceRequest.headers();
 
-        writeBuffer.clear();
-        originalRequest.sanitizeRedirectHeaders(uri, headers);
-        prologue(effectiveConnection, writeBuffer, serviceRequest, uri);
-        headers.setIfAbsent(HeaderValues.create(HeaderNames.HOST, uri.authority()));
+            writeBuffer.clear();
+            originalRequest.sanitizeRedirectHeaders(uri, headers);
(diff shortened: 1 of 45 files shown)


webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
```diff
@@ -587,16 +589,20 @@ private void redirect(Status lastStatus, WritableHeaders<?> headerValues) {
...
-                    if (!sendEntity || sendEmptyEntity) {
-                        //OS changed its state to interrupted, that means other usage of this OS will result in NOOP actions.
-                        this.interrupted = true;
-                        this.response = response;
-                        //we are not sending anything by this OS, we need to interrupt it.
-                        throw new OutputStreamInterruptedException();
-                    } else {
+                    if (sendEntity
+                            && !sendEmptyEntity
+                            && response.status().code() == Status.CONTINUE_100.code()) {
```
danielkec, line 597: When a method-preserving redirect target sends `103 Early Hints` before `100 Continue`, this equality treats `103` as the terminal probe response, so the caller receives `103` and the redirected request body is never sent. A raw response sequence reproduces this on the PR head:

```text
307 -> 103 -> 100 -> 200
PR head: returned-status=103, target-body=<EOF>
4.x base: returned-status=200, target-body=redirect-body
```

The source PR's redirect-probe reader skips pre-continue informational responses, so this is also a behavioral divergence from the source change.

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2CallOutputStreamChain.java
```diff
@@ -567,27 +569,25 @@ private void redirect(Status lastStatus, Headers headerValues) {
...
                     } else {
-                        if (!sendEntity || sendEmptyEntity) {
-                            //OS changed its state to interrupted, that means other usage of this OS will result in NOOP actions.
-                            this.interrupted = true;
-                            this.response = response;
-                            //we are not sending anything by this OS, we need to interrupt it.
-                            throw new OutputStreamInterruptedException();
-                        }
-                        return;
+                        //OS changed its state to interrupted, that means other usage of this OS will result in NOOP actions.
```
danielkec, line 586: When a redirect probe receives a terminal response with no entity, such as `204` with `END_STREAM`, this path returns while the upload side is still open, and fully reading the zero-byte entity cannot trigger response cleanup. The HTTP/2 stream remained `HALF_CLOSED_REMOTE` with one reserved slot until explicit `response.close()`; the analogous HTTP/1 branch left its acquired socket open after `status=204, entity-bytes=0`. This violates `HttpClientResponse`'s implicit-close-after-full-read contract and can retain connection or stream capacity.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallOutputStreamChain.java
```diff
@@ -351,13 +352,15 @@ WebClientServiceResponse serviceResponse() {
...
-                                         response.headers(),
-                                         whenComplete);
+            WebClientServiceResponse result = createServiceResponse(http1Client,
+                                                                    request,
+                                                                    response.connection(),
+                                                                    response.connection().reader(),
+                                                                    response.status(),
+                                                                    response.headers(),
+                                                                    whenComplete);
+            response.closeIfNoEntity();
```
danielkec: When an early `301`/`302`/`303` changes the redirected output-stream request to GET and that GET returns `204` or `Content-Length: 0`, lines 355-361 transfer the raw connection to the caller-facing response, but this call closes the separate inner response and releases that same connection immediately. Closing the public response releases it again. A base/head release-counter probe shows:

```text
b25d889: before public close=0, after=1
ff5f168: before public close=1, after=2
```

The HTTP/1 cache accepts both entries, so two concurrent requests can be handed the same live socket and their HTTP messages or responses can become interleaved.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1ClientResponseImpl.java
```diff
@@ -186,7 +190,17 @@ public ClientUri lastEndpointUri() {
...
-        return connection;
+        return closeConnectionOnClose ? new CloseOnReleaseClientConnection(connection) : connection;
+    }
+
+    void closeConnectionOnClose() {
+        closeConnectionOnClose = true;
+    }
+
+    void closeIfNoEntity() {
+        if (inputStream == null) {
```
danielkec: The earlier `204` cleanup case is addressed, but header-terminated `205 Reset Content` and `304 Not Modified` responses still receive an `EverythingInputStream`, so this condition does not close them. Exact two-hop probes reproduced the residual behavior:

```text
307 -> 304: status=304, entity read=BLOCKED
307 -> 205: status=205, entity read=BLOCKED
after public close: service callbacks started=2, completed=1
```

The peer remains open until explicit close, and the inner redirect request's completion never fires even then. The source change classifies `205` and `304` as body-free; the new 4.x coverage exercises only `204`.

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -296,13 +316,20 @@ private static boolean mayHaveEntity(Status responseStatus, ClientResponseHeader
         if (responseHeaders.contains(HeaderValues.CONTENT_LENGTH_ZERO)) {
             return false;
         }
-        // Why is NOT_MODIFIED_304 not added here too?
-        if (responseStatus.code() == Status.NO_CONTENT_204.code()) {
+        int statusCode = responseStatus.code();
+        if (statusCode == Status.NO_CONTENT_204_CODE
+                || statusCode == Status.NOT_MODIFIED_304_CODE) {
```
danielkec, line 324: `304` now returns no input stream, so `Http1ClientResponseImpl` leaves the response length at zero and releases the connection without checking buffered bytes. If a peer replies `304 Content-Length: 42` and immediately sends a 42-byte response before the next request, that queued response becomes the result of the next pooled request; at `ff5f168` those bytes were consumed or the socket was retired. The new metadata test contains no post-header bytes, so it does not exercise this response-desynchronization path.

```text
HTTP/1.1 304 Not Modified
Content-Length: 42

HTTP/1.1 200 OK
Content-Length: 4

evil
```

webclient/http1/src/main/java/io/helidon/webclient/http1/Http1CallChainBase.java
```diff
@@ -296,13 +316,20 @@ private static boolean mayHaveEntity(Status responseStatus, ClientResponseHeader
...
+        if (statusCode == Status.RESET_CONTENT_205_CODE
+                && !responseHeaders.contains(HeaderNames.CONTENT_LENGTH)
+                && !responseHeaders.contains(HeaderNames.TRANSFER_ENCODING)) {
+            // Preserve header-terminated 205 compatibility, but honor declared framing so it is consumed before reuse.
             return false;
         }
-        if ((
-                responseHeaders.contains(HeaderNames.UPGRADE)
-                        && !responseHeaders.containsToken(HeaderValues.TRANSFER_ENCODING_CHUNKED))) {
+        if (statusCode == Status.SWITCHING_PROTOCOLS_101_CODE
```
danielkec, line 333: Restricting the Upgrade shortcut to `101` makes a regular HEAD request parse `426 Upgrade Required` metadata as content. A valid HEAD response with `Connection: Upgrade`, `Upgrade: h2c`, and `Content-Length: 7` has no payload, but current head creates a seven-byte stream and a read waits until timeout; closing it also discards the keep-alive connection. At `ff5f168`, the same response returned EOF immediately, and the new coverage exercises only GET.

webclient/http1/src/test/java/io/helidon/webclient/http1/Http1ClientTest.java
```diff
@@ -1073,6 +1403,410 @@ private void webServerHandle() {
...
+                default -> throw new IllegalStateException("Unexpected header parser state");
+                };
+            }
+        }
+    }
+
+    private record UpgradeRequiredServer(ServerSocket server,
+                                         CompletableFuture<Void> completion) implements AutoCloseable {
+        private static final byte[] UPGRADE_RESPONSE = ("HTTP/1.1 426 Upgrade Required\r\n"
+                + "Upgrade: h2c\r\n"
```
danielkec, line 1896: This 426 fixture sends `Upgrade: h2c` without a `Connection` header naming `Upgrade`. Upgrade is connection-specific, so this is not a valid HTTP/1.1 exchange and intermediaries can discard the field; the test therefore does not cover the valid 426 response form it is meant to exercise.

Verdent (approved):

webclient/http2/src/main/java/io/helidon/webclient/http2/Http2CallChainBase.java
```diff
@@ -256,7 +257,9 @@ static Http2Headers readHeaders(Http2ClientStream stream) {
 
     static Http2Headers readHeaders(Http2ClientStream stream, Duration readTimeout) {
         try {
-            return readTimeout == null ? stream.readHeaders() : stream.readHeaders(readTimeout);
+            Http2Headers headers = readTimeout == null ? stream.readHeaders() : stream.readHeaders(readTimeout);
+            stream.finishNoContent();
```
danielkec, line 261: When an output-stream redirect receives final HEAD/304 headers but wire `END_STREAM` does not arrive before the read timeout, this new drain throws `StreamTimeoutException`. That exception bypasses the `Http2Exception` cleanup; the output-stream redirect path then treats it as a missing `100 Continue`, so upload processing resumes while the stream remains registered and its peer-concurrency slot stays reserved. Exact-head probe after the timeout:

```text
timeout=StreamTimeoutException
streamStillRegistered=true
peerSlotStillReserved=true
```

PR #12534, opened 2026-09-08
Align gRPC inbound message size handling

### LINKED ISSUE #11720: gRPC server sends CANCELLED/UNKNOWN instead of RESOURCE_EXHAUSTED for oversized inbound messages

spericas:
### Description

- Increase the default gRPC inbound read buffer from 2 MiB to 4 MiB.
- Return `RESOURCE_EXHAUSTED` with a size-neutral description when an inbound message exceeds the configured limit.
- Enforce the configured limit before narrowing the unsigned message length or reusing the initial read buffer.
- Log oversized-message diagnostics at `DEBUG` without a stack trace, keeping the configured and declared sizes server-side.
- Preserve `ERROR` logging with the throwable for unrelated `RESOURCE_EXHAUSTED` failures from request parsing or listener callbacks.
- Add unit coverage for limits below the initial buffer capacity, unsigned lengths above `Integer.MAX_VALUE`, and both logging paths.
- Update integration coverage for the new default and oversized-message response.

Fixes #11720

### Documentation

- Regenerate the gRPC configuration reference for the 4 MiB default.

### Testing

- `mvn -f webserver/grpc/pom.xml -Pcheckstyle,spotbugs verify`
- `mvn -f webserver/tests/grpc/pom.xml -Pcheckstyle,spotbugs -Dtest=UploadServiceTest verify`
- `mvn -f webserver/grpc/pom.xml -Pcopyright validate`
- `mvn -f webserver/tests/grpc/pom.xml -Pcopyright validate`


tomas-langer (dismissed):

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
```diff
@@ -392,18 +394,35 @@ public void data(Http2FrameHeader header, BufferData data) {
                 throw new ServerConnectionException("gRPC call cancelled by remote peer", e);
             }
             closeOnException(e, header);
-            LOGGER.log(ERROR, "Failed to process grpc request, data bytes: " + data.available(), e);
+            Status status = Status.fromThrowable(e);
+            String description = status.getDescription();
+            if (description == null) {
+                description = e.getMessage() == null ? "Failed to process gRPC request" : e.getMessage();
+            }
+            if (status.getCode() == Status.Code.RESOURCE_EXHAUSTED) {
```
tomas-langer: Every `RESOURCE_EXHAUSTED` exception reaching this catch is treated as an oversized-message rejection. The catch also encloses request parsing and listener callbacks invoked by `flushQueue()`, which can throw an unrelated `StatusRuntimeException` with this status. Such failures are now logged only at DEBUG, without the throwable, and with misleading maximum/declared-message-size fields. The wire status remains correct, but the server diagnostics misidentify and hide non-size resource-exhaustion failures.

webserver/grpc/src/test/java/io/helidon/webserver/grpc/GrpcProtocolHandlerTest.java
```diff
@@ -774,6 +794,40 @@ public ServerCall.Listener<String> startCall(ServerCall<String, String> call, Me
...
+        data.write(0);
+        data.writeUnsignedInt32(messageSize);
+        Http2FrameHeader header = Http2FrameHeader.create(data.available(),
+                                                          Http2FrameTypes.DATA,
+                                                          Http2Flag.DataFlags.create(Http2Flag.END_OF_STREAM),
+                                                          1);
+
+        handler.data(header, data);
+
+        assertAll(
```
tomas-langer: The new tests assert the client status and description but do not observe logging. Issue #11720 also requires oversized inbound messages to stop producing ERROR-level stack traces. This unit test and the integration test would still pass if the handler returned `RESOURCE_EXHAUSTED` while regressing to the previous ERROR log, so the logging behavior changed by this PR has no regression protection.

webserver/grpc/src/test/java/io/helidon/webserver/grpc/GrpcProtocolHandlerTest.java
```diff
@@ -774,6 +794,40 @@ public ServerCall.Listener<String> startCall(ServerCall<String, String> call, Me
         assertThat(grpcConnectionContext.get().sniMatchedHost(), is(Optional.of("*.example.com")));
     }
 
+    private void assertOversizedMessage(long messageSize, int maxReadBufferSize) {
```
tomas-langer: `assertOversizedMessage(...)` is a private instance method declared before private static helpers beginning with `route(...)` at line 848. `DEV-GUIDELINES.md` Rule 2.1 requires private static methods before private instance methods. This newly added member therefore violates the published class-member ordering rule.

webserver/grpc/src/main/java/io/helidon/webserver/grpc/GrpcProtocolHandler.java
spericas: Logging location has changed

webserver/grpc/src/test/java/io/helidon/webserver/grpc/GrpcProtocolHandlerTest.java
spericas: New tests

webserver/grpc/src/test/java/io/helidon/webserver/grpc/GrpcProtocolHandlerTest.java
spericas: Should be fixed

tomas-langer (approved):

Issue #12533, opened 2026-09-08
Complete HTTP header and redirect follow-up work


tomas-langer:
Follow-up work from #12505, #12526, and the 4.x backport #12532. Keep `main`
and `helidon-4.x` aligned where applicable:

- Reject `null` in `HttpMediaType.withCharset(String)` on both lines, and align
  its Javadoc and tests with Helidon's non-null-by-default public API contract.
- Align the corresponding `HttpTokenTest` on `main` by invoking validation
  directly instead of wrapping it in `assertDoesNotThrow`; 4.x is already
  corrected in #12532.
- Carry the explicit `Header` Javadocs from #12532 to `main`, documenting that
  characters from U+0080 through U+00FF represent opaque field-value octets,
  not UTF-8 or ISO-8859-1 text, and that characters above U+00FF are rejected.
- Document which request headers and entities WebClient preserves or filters
  for method-preserving, method-rewriting, same-origin, and cross-origin
  redirects on both lines.
- Investigate and reduce repeated HTTP field-value validation and encoding
  passes while preserving the octet-mapping and rejection behavior. Matched
  JMH measurements for #12505 showed uncached HTTP/1 ASCII serialization rising
  from 15.929 +/- 1.093 to 26.220 +/- 3.325 ns/op (about 65%), HPACK request
  blocks from 357.770 +/- 31.480 to 503.596 +/- 28.809 ns/op (about 41%), and
  HPACK response blocks from 279.278 +/- 10.247 to 423.994 +/- 15.445 ns/op
  (about 52%).

This follow-up should not change the source-faithful correctness scope of
#12532.

tomas-langer:
Reject null charsets - this would be a backward incompatible change in 4.x, preserving existing behavior
Optimization - not going to backport into a tail version

Closing this issue, as `main` is updated as it should be, and there is no intention to backport the changes to 4.x

PR #12538, opened 2026-09-08
Complete HTTP header and redirect follow-up (27.x)

tomas-langer:
Addresses #12533

Rejects null charset updates and clarifies the HTTP header octet and WebClient
redirect contracts.

Reduces repeated HPACK field validation by rolling back partial encodings and
dynamic-table updates when encoding fails, while preserving the header-name
validation introduced by #12505.

### DIFF
--- a/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
+++ b/http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
@@ -89,6 +89,9 @@ public class Http2Headers {
     private static final String HTTPS = "https";
     private static final String PATH_SLASH = "/";
     private static final String PATH_INDEX = "/index.html";
+    private static final int SHOULD_INDEX = 1;
+    private static final int NEVER_INDEX = 1 << 1;
+    private static final int VALIDATE_NAME = 1 << 2;
 
     private final Headers headers;
     private final PseudoHeaders pseudoHeaders;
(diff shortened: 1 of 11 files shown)


http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
```diff
@@ -1459,6 +1453,35 @@ int currentTableSize() {
...
+
+        private void commitEncoding() {
+            encoding = false;
+            encodingSnapshot = null;
+        }
+
+        private void rollbackEncoding() {
+            if (encodingSnapshot != null) {
+                headers.clear();
+                headers.addAll(encodingSnapshot);
```
danielkec, line 1471: This restoration branch is not exercised with a non-empty snapshot. The pre-populated-table test rejects `\u0100` before `DynamicTable.add()` checkpoints or mutates anything, while every mutation-before-failure test starts with an empty table, so omitting the entry/size restoration would still leave the suite green. A minimal uncovered sequence is: prepopulate a table, encode a block that first inserts or evicts a valid field and then fails on an out-of-range value, and verify the prior entry/size and the next indexed block remain usable; otherwise a rollback regression can desynchronize the connection-scoped HPACK state.

tests/benchmark/jmh/src/main/java/io/helidon/http/http2/HttpCodecJmhBenchmark.java
```diff
@@ -144,6 +167,47 @@ public static class EncodingState {
...
+
+    @State(Scope.Thread)
+    public static class PopulatedTableInsertionState {
+        private final Http2Headers existingHeaders = indexedHeaders(ASCII_VALUE);
+        private final Http2Headers newHeaders = indexedHeaders(OTHER_ASCII_VALUE);
+        private final Http2HuffmanEncoder huffman = Http2HuffmanEncoder.create();
+        private final BufferData buffer = BufferData.growing(64);
+        private Http2Headers.DynamicTable table;
+
+        @Setup(Level.Invocation)
```
danielkec, line 207: `@Setup(Level.Invocation)` runs before the per-operation timing, but `GCProfiler` snapshots only at iteration boundaries. This setup therefore adds the new table allocation—and, in the populated case, a complete `existingHeaders.write()` including the changed snapshot path—to every reported `gc.alloc.rate.norm` value. A short runner probe reported roughly 428 B/op for the empty case and 652 B/op for the populated case; those allocation figures mix setup with the operation under test, so they cannot quantify the allocation cost of the new insertion path even though the primary `ns/op` score remains valid.

http/http2/src/main/java/io/helidon/http/http2/Http2Headers.java
danielkec: not really a blocker

tests/benchmark/jmh/src/main/java/io/helidon/http/http2/HttpCodecJmhBenchmark.java
danielkec: not a blocker

danielkec (dismissed):

danielkec (approved):

PR #12542, opened 2026-09-09
Log expected WebServer connection termination at trace

tomas-langer:
Carry forward #12541 to main.

Handle peer EOF, socket writer failures, and connection interruption as expected connection termination so these paths are logged at trace instead of warning.

Also select fixed-length protocols once their advertised byte count is available and normalize truncated HTTP/1 headers and HTTP/2 prefaces to connection closure.

### DIFF
--- a/webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java
+++ b/webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java
@@ -48,6 +48,7 @@ import io.helidon.common.socket.NioSocket;
 import io.helidon.common.socket.PeerInfo;
 import io.helidon.common.socket.PlainSocket;
 import io.helidon.common.socket.SocketWriter;
+import io.helidon.common.socket.SocketWriterException;
 import io.helidon.common.socket.TlsNioSocket;
 import io.helidon.common.socket.TlsSocket;
 import io.helidon.common.task.InterruptableTask;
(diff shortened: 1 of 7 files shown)


webserver/webserver/src/main/java/io/helidon/webserver/ConnectionHandler.java
```diff
@@ -353,6 +354,12 @@ private void run(String channelId) {
         } catch (CloseConnectionException e) {
             // end of request stream - safe to close the connection, as it was requested by our client
             helidonSocket.log(LOGGER, TRACE, "connection close requested", e);
+        } catch (DataReader.InsufficientDataAvailableException | SocketWriterException e) {
```
danielkec, line 357: `SocketWriterException` is not limited to peer I/O: `SocketWriterAsync` catches any `Throwable` from `writeBatch` and later wraps it; the existing `asyncWriterFlushPropagatesWriteFailureWithoutSuccessRace` test demonstrates this with `RuntimeException("boom")`. This catch therefore moves an internal writer failure from the previous WARNING-level `unexpected exception` path to TRACE, which can hide a server defect from normal operational logs.

A minimal reproducer through the new test helper is:

```java
assertConnectionFailureLevel(
        new SocketWriterException(new RuntimeException("boom")),
        Level.WARNING);
```

tomas-langer: This classification is aligned with the behavior already merged into main in #12358: HTTP/2 reset/window-update write paths wrap `SocketWriterException` in `ServerConnectionException`, which is logged at TRACE without inspecting the cause. We want main and 4.x to retain functionally equivalent behavior, so we are keeping that classification here. The wrapped internal-failure concern also applies to those existing paths; changing it should be handled consistently across both lines.

tomas-langer: The expected failure here is that the writer can no longer write to its socket, so I would retain the connection-termination classification at this boundary. The cited test injects a `RuntimeException` into a mocked socket to verify failure propagation; it does not demonstrate a separate internal failure during normal socket-writer operation.

I agree that `catch (Throwable)` can also wrap an unexpected defect, so there is a diagnostic tradeoff. I don't think that possibility alone warrants cause-based handling in this PR. This keeps the behavior aligned with the existing HTTP/2 classification from #12358 and the 4.x change in #12541. If we identify a concrete internal failure needing separate treatment, the socket-writer layer would be the place to establish that distinction consistently.

danielkec (approved):

PR #12553, opened 2026-09-11
Documentation for Data JDBC

bhatpmk:
### Description
Provides documentation for Helidon Data JDBC

### Documentation
The documentation presents Data JDBC as a provider with two complementary programming models:

- Declarative programming using generated repository implementations
- Imperative programming using JdbcClient


docs/modules/data-jdbc/README.md
romain-grecourt, line 1: README.md are for navigation exclusively.
If you want the route to be `/data-jdbc` name this file `data-jdbc.md` and use index.page in the README.md frontmatter.

docs/modules/data-jdbc/data-jdbc.md
```diff
@@ -0,0 +1,291 @@
...
+    <groupId>io.helidon.data.jdbc</groupId>
+    <artifactId>helidon-data-jdbc</artifactId> <!-- Data JDBC API and runtime -->
+  </dependency>
+  <dependency>
+    <groupId>io.helidon.config</groupId>
+    <artifactId>helidon-config-yaml</artifactId> <!-- support for application.yaml -->
+  </dependency>
+  <dependency>
+    <groupId>com.mysql</groupId>
+    <artifactId>mysql-connector-j</artifactId> <!-- MySQL JDBC driver -->
```
tomas-langer, line 57: This snippet leaves `com.mysql:mysql-connector-j` unversioned, but the linked dependency-management guide supports applications that import only `helidon-bom`, which does not manage that dependency. Those applications receive a Maven missing-version error when they copy this setup. The version is inherited only when using the Helidon application parent.

bhatpmk: I will provide a link to guides/maven.md#dependency-management, which suggests to import io.helidon:helidon-dependencies if the application is not using Helidon application parent. The mysql-connector-j version is defined in helidon/dependencies/pom.xml.

docs/modules/data-jdbc/README.md
bhatpmk: Updated as suggested.

docs/modules/data-jdbc/imperative.md
```diff
@@ -0,0 +1,360 @@
...
+
+Resources are also released when database access, mapping, or result validation
+fails. [Errors and JDBC Warnings](data-jdbc.md#errors-and-jdbc-warnings)
+describes the resulting diagnostic information.
+
+## Transaction Participation
+
+A client obtained from the Service Registry can participate in a local JDBC
+transaction. Transaction annotations take effect when the application invokes
+a service instance obtained from the Service Registry. They do not take effect
```
tomas-langer: The same-instance limitation is incorrect for registry-managed services. Helidon generates an intercepted subclass, so a call from one service method to another intercepted method on `this` still reaches the transaction interceptor. For example, an internal call to an intercepted `@Tx.New` or `@Tx.Unsupported` method can start a new transaction or suspend the current one. Private methods are not intercepted and their transaction annotations do not take effect; directly annotated private methods can also be rejected during annotation processing. The current blanket statement can therefore lead applications to rely on the wrong transaction boundary.

tomas-langer (approved):

PR #12570, opened 2026-09-14
Optimize HTTP/2 buffered responses and fix byte ranges

tomas-langer:
Buffered HTTP/2 responses copy their DATA payload before the synchronous transport write. Reuse a read-only view of the buffered entity and flatten the combined HEADERS/DATA batch to reduce copying and allocation. Other buffer sources retain a defensive copy, and media serialization remains protocol-independent.

Preserves binary and hex diagnostics for read-only slices and partially consumed buffers, and sizes their diagnostic storage from the logical byte count instead of the backing-array offset. Adds regression coverage for byte ranges, flow-control splits, completion ordering, caller-buffer reuse and listener diagnostics, plus JMH cases for representative payload sizes and mutable-buffer fallback.

Corrects the default `ServerResponse.send(byte[], position, length)` implementation to treat `length` as a byte count. Implementations inheriting the default now send the requested range instead of truncating it or rejecting valid nonzero offsets; tests cover full, partial and empty ranges.

In the synchronous writer microbenchmark on JDK 26, a 3,519-byte read-only payload took about 27% less time, with allocation reduced from approximately 3,944 to 272 bytes per operation. Focused tests and production-module static checks passed.


danielkec (approved):
Reviewed buffer ownership, byte ranges, diagnostics, flow control, and completion ordering. All 86 focused tests passed. Base/head JMH comparisons also confirm the intended read-only payload allocation and writer-time improvements.

PR #12571, opened 2026-09-15
Fix MP metrics docs: use @Initialized(ApplicationScoped) instead of @RuntimeStart

### LINKED ISSUE #11163: 4.x Doc incorrectly suggests observing `@RuntimeStart` (for obtaining metric registry for example) when it should suggesting observing `@Initialized(ApplicationScoped.class)` instead
## Environment Details
* Helidon Version: 4.x
* Helidon SE or Helidon MP MP
* JDK version:
* OS:

----------

## Problem Description
Places in the doc (definitely MP metrics, maybe elsewhere) incorrectly state that user code can observe the `@RuntimeStart` event to make sure they don't access things like the MP `MetricRegistry` before it is properly initialization.

In fact, those places need to suggest observing `@Initialized(ApplicationScoped.class)` instead.

Observing `@RuntimeStart` is too soon for this purpose.

tanvir-ux:
## Description
Fixes incorrect CDI lifecycle guidance in the MP metrics docs.

`docs/mp/metrics/metrics.md` told users to observe Helidon’s `@RuntimeStart` before working with `MetricRegistry`. That is too early; metrics may not yet reflect configuration. The docs now recommend `@Initialized(ApplicationScoped.class)`, consistent with `docs/mp/guides/metrics.md`.

Fixes #11163

## Testing
Docs-only change; reviewed the CDI extension example and surrounding prose.


docs/mp/metrics/metrics.md
```diff
@@ -1310,33 +1310,37 @@ curl -H "Accept: application/json"  'http://localhost:8080/metrics?scope=applica
...
+                 MetricRegistry metricRegistry) {                               // <2>
+        metricRegistry.counter("myCounter");                                    // <3>
     }
 }
 ```
-1. Declares that your observer method responds to the `RuntimeStart` event. By
-   this time, Helidon has initialized the metrics system.
+1. Declares that your observer method responds to the
+   `@Initialized(ApplicationScoped.class)` event. By this time, Helidon has
+   initialized the metrics system. Observing Helidon’s `@RuntimeStart` is too
```
tjquinno, line 1328: I know this is correcting the earlier, incorrect note about RuntimeStart. But would it be clearer just to mention the correct way to do it, rather than also describing that RuntimeStart _does not_ work?

docs/mp/metrics/metrics.md
```diff
@@ -1310,33 +1310,37 @@ curl -H "Accept: application/json"  'http://localhost:8080/metrics?scope=applica
...
 3. Uses the injected registry to register a metric (a counter in this case).
 <!--@mdc :: -->
 
-Helidon does not prevent you from working with metrics earlier than the
-`RuntimeStart` event, but, if you do so, then Helidon might ignore certain
-configuration settings that would otherwise control how metrics behaves.
+Helidon does not prevent you from working with metrics earlier than
+`@Initialized(ApplicationScoped.class)`, but, if you do so, then Helidon might
+ignore certain configuration settings that would otherwise control how metrics
+behaves.
```
tjquinno, line 1337: The side effects can be even worse, because registering metrics too early can trigger the creation of two sets instead of one set of underlying data structures, and the early-registered metrics might not even appear in the output.

tjquinno (changes requested):

PR #12574, opened 2026-09-15
Fix byte-range sends and buffer diagnostics (4.x)

### LINKED ISSUE #12573: Fix byte-range sends and buffer diagnostics (4.x)
The default `ServerResponse.send(byte[], position, length)` implementation treats `length` as an end index instead of a byte count. For example, sending three bytes at offset two can truncate the response; valid ranges with larger offsets can throw an exception.

Binary and unread hex diagnostics for read-only buffers can omit bytes when a slice has a nonzero offset or some bytes have already been consumed. Diagnostic storage is also sized from the backing-array endpoint instead of the number of bytes displayed.

Backport the byte-range and buffer-diagnostic fixes and their regression tests from #12570 to `helidon-4.x`. The HTTP/2 allocation optimization is excluded.

tomas-langer:
Resolves #12573

Backport the byte-range and buffer-diagnostic fixes from #12570. The default `ServerResponse.send(byte[], position, length)` now sends the requested byte count. Binary and hex diagnostics preserve the selected and unread buffer ranges and size their storage from the displayed byte count. Includes the corresponding regression tests.


danielkec (approved):
The backport matches the intended fixes from #12570 and resolves #12573. The 13 focused tests and 3,393 buffer-range probe cases pass; no introduced findings.

Verdent (approved):

PR #12587, opened 2026-09-17
Fix service registry bindings, event lifecycle, and factory validation

tomas-langer:
Preserve equally weighted late-bound service instances and stable listener snapshots during event dispatch. Shut down registry-owned event executors without waiting for submitted tasks, and reject null factory inputs.

Add regression coverage for registry lifecycle, qualified instances, and interception metadata, plus a brief note on the JNDI module's supporting role for Helidon Data.


docs/modules/injection/injection.md
```diff
@@ -1178,6 +1183,12 @@ var registryManager = ServiceRegistryManager.create();
 registryManager.shutdown();
 ```
 
+## JNDI support
+
+The `helidon-service-jndi` module is a supporting module for Helidon Data.
```
romain-grecourt: I disagree with intersecting features this way.

If the module is located under service/jndi, then it must be completely decoupled from Helidon Data.
Otherwise it should reside under data/jndi.

Remove the section or document the current "limited" JDNI support.

danielkec (approved):

